PR #12093, opened 2020-05-06
openssh-server: deprecate the ecdsa HostKey

pprindeville:
Maintainer: @tripolar 
Compile tested: x86_64, generic, master HEAD
Run tested: same

Generated `openssh-server` package and installed it (no pam).  Had to:

```
root@OpenWrt:~# mv -f /etc/ssh/sshd_config-opkg /etc/ssh/sshd_config
root@OpenWrt:~# /etc/init.d/sshd restart
```

Logged into server.

Confirmed in logs that there's no error message.

```
root@OpenWrt:~# tail /var/log/messages | grep sshd
root@OpenWrt:~# 
```

### DIFF
--- a/net/openssh/Makefile
+++ b/net/openssh/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openssh
 PKG_VERSION:=8.2p1
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


BKPepe:
This was not approved by @tripolar and he is maintainer. He should have the final word about that!

champtar:
@pprindeville you can wait a bit more than 22h ;)

BKPepe:
I think you should take a look at [CONTRIBUTING.md](https://github.com/openwrt/packages/blame/master/CONTRIBUTING.md#L70):

> * Do NOT commit to other maintainer's packages without their consent.

Which you basically did, right?

PR #12103, opened 2020-05-07
shadowsock-libev: reload only when running

zhanhb:
Signed-off-by: Huangbin Zhan <zhanhb88@gmail.com>

Maintainer: @yousong 
Compile tested: 
Run tested: 

Description:
Don't reload service if it's is stopped

### DIFF
--- a/net/shadowsocks-libev/files/firewall.ss-rules
+++ b/net/shadowsocks-libev/files/firewall.ss-rules
@@ -1,2 +1,4 @@
 #!/bin/sh
-/etc/init.d/shadowsocks-libev reload
+if /etc/init.d/shadowsocks-libev running; then
+    /etc/init.d/shadowsocks-libev rules
+fi
--- a/net/shadowsocks-libev/files/shadowsocks-libev.init
+++ b/net/shadowsocks-libev/files/shadowsocks-libev.init
@@ -52,9 +52,10 @@ ss_mkjson_ss_tunnel_conf() {
 	json_add_string tunnel_address "$tunnel_address"
 }
 
-ss_xxx() {
+ss_xxx() { # <cfg> <cfgtype> <skip_instance>
 	local cfg="$1"
 	local cfgtype="$2"
+	local skip_instance="$3"
 	local bin="$ss_bindir/${cfgtype/_/-}"
 	local confjson="$ss_confdir/$cfgtype.$cfg.json"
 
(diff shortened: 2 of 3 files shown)


net/shadowsocks-libev/files/firewall.ss-rules
```diff
@@ -1,2 +1,7 @@
 #!/bin/sh
-/etc/init.d/shadowsocks-libev reload
+if /etc/init.d/shadowsocks-libev running; then
+    /etc/init.d/shadowsocks-libev reload
+else
+    /usr/bin/ss-rules -f
+    /usr/bin/ss-rules -6 -f
+fi
```
yousong, line 7: `service_running` at the moment only checks a single instance's running status.  To be more precise, just check value of uci option `shadowsocks-libev.ss_rules.disabled`.

The `ss-rules` flush call should be not needed.  In normal case `firewall.ss-rules` will only be invoked by firewalld to populate ss-rules, i.e. the flush is already done.

yousong:
The commit title should be `shadowsocks-libev: ss-rules: xxx`

net/shadowsocks-libev/files/firewall.ss-rules
zhanhb: Maybe the service is stopped by user with command `service shadowsocks-libev stop`, shoud it be reloaded then?

yousong: The way it was written is intended to make state of uci config to have the final say on which instances are to be run.  Constantly syncing that information to procd should make the integration more resilient while to a degree being less efficient (redundant).

If we look at the "reload" call alone, the default is to "stop then start" or just "start".  In this pull request, it's firewall.ss-rules we are talking about, it could be any other scripts.  This is how it was supposed to work.  If we are to indicate to other players that the shadowsocks-libev instances should not be run, we should make it clear.  Runtime status of procd service instances is not for this.

net/shadowsocks-libev/files/shadowsocks-libev.init
```diff
@@ -65,26 +66,28 @@ ss_xxx() {
...
+		[ -z "$user" ] || json_add_string user "$user"
+		json_dump -i >"$confjson"
+
+		procd_open_instance "$cfgtype.$cfg"
+		procd_set_param command "$bin" -c "$confjson"
+		[ "$verbose" = 0 ] || procd_append_param command -v
+		[ -z "$bind_address" ] || procd_append_param command -b "$bind_address"
+		procd_set_param file "$confjson"
+		procd_set_param respawn
+		procd_close_instance
```
yousong, line 89: This indented hunk is idempotent.  procd will compare current and to-be-applied specs of instances, do nothing if they match.  I am afraid adding a new knob is not worth the sweat.

PR #12104, opened 2020-05-07
readsb: new package

stintel:
Maintainer: me
Compile tested: OpenWrt master r13172-2b31f143f9 on bcm27xx/bcm2709
Run tested: OpenWrt master r13172-2b31f143f9 on bcm27xx/bcm2709

This package is based on dump1090-fa with many modifications. A big
advantage over dump1090-fa is that it can connect to multiple services
like adsbexchange, piaware or fr24feed.

As the idea for this package is mostly to feed other services, the HTML
files to serve via a webserver have not been included. The service(s)
that will be fed by readsb can offer such a webpage instead.

Finally, the init script does not use local variables, as local is
undefined in POSIX.


utils/readsb/Makefile
```diff
@@ -0,0 +1,66 @@
...
+
+PKG_NAME:=readsb
+PKG_VERSION:=3.8.3
+PKG_RELEASE:=1
+
+PKG_SOURCE_URL:=https://github.com/Mictronics/readsb/archive/v$(PKG_VERSION)
+PKG_HASH:=bf59b8ecd0ff66945b210c55a5b824aa63ff8cbb1704249528d30a4902e716b2
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+
+PKG_BUILD_DEPENDS:=!USE_GLIBC:argp-standalone
```
neheb, line 16: No idea if PKG_INSTALL would work. PKG_BUILD_PARALLEL probably would.

utils/readsb/Makefile
```diff
@@ -0,0 +1,66 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/readsb/default
+  SECTION:=utils
+  CATEGORY:=Utilities
+  DEPENDS:=+libncurses
+  TITLE:=Mode-S/ADSB/TIS decoder for various devices
+  URL:=https://github.com/Mictronic/readsb
+  MAINTAINER:=Stijn Tintel <stijn@linux-ipv6.be>
```
neheb, line 26: I recommend using PKG_MAINTAINER instead. Also adding PKG_LICENSE information in the same group.

utils/readsb/Makefile
```diff
@@ -0,0 +1,66 @@
...
+endef
+
+MAKE_FLAGS += \
+        BLADERF=no \
+        CFLAGS="$(TARGET_CFLAGS)" \
+	RTLSDR=yes \
+        UNAME="Linux"
+
+TARGET_CFLAGS += -ffunction-sections -fdata-sections -flto
+TARGET_LDFLAGS += -Wl,--gc-sections,--as-needed -largp
```
neheb, line 47: -largp probably needs to be conditional. See https://github.com/openwrt/packages/blob/master/libs/libv4l/Makefile#L81

utils/readsb/Makefile
```diff
@@ -0,0 +1,66 @@
...
+
+define Package/viewadsb
+  $(call Package/readsb/default)
+  TITLE+= (viewadsb)
+endef
+
+MAKE_FLAGS += \
+        BLADERF=no \
+        CFLAGS="$(TARGET_CFLAGS)" \
+	RTLSDR=yes \
```
neheb, line 43: Whitespace issue?

utils/readsb/Makefile
```diff
@@ -0,0 +1,66 @@
...
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=readsb
+PKG_VERSION:=3.8.3
+PKG_RELEASE:=1
+
+PKG_SOURCE_URL:=https://github.com/Mictronics/readsb/archive/v$(PKG_VERSION)
```
neheb, line 12: I think codeload should be used here.

stintel: My package vallumd doesn't use this either, so I prefer to keep them consistent. If using codeload is *really* a must, we should update everything that doesn't use it, and then auto-reject PKG_SOURCE_URL starting with https://github.com/ via CI with a friendly message to use codeload, and how/why.

utils/readsb/Makefile
stintel: Fixed.

utils/readsb/Makefile
stintel: Good catch. Fixed.

utils/readsb/Makefile
stintel: Changed to PKG_MAINTAINER, added PKG_LICENSE{,_FILES}.

utils/readsb/Makefile
stintel: I don't see the need for PKG_INSTALL. It's just 2 binaries in the root of the source. Using PKG_INSTALL would only result in a bigger Makefile.

Added PKG_BUILD_PARALLEL.

utils/readsb/Makefile
neheb: Eh no.

The typical pattern with github is to use codeload for packages that have proper versioning and local tarballs for packages that do not.

Note that that URL redirects to codeload anyway. I think redirects are to be avoided here.

stintel: Changed to codeload.

PR #12115, opened 2020-05-08
oniguruma: bump to 6.9.5_rev1, switch to cmake

cotequeiroz:
Maintainer: me
Compile tested: mvebu, wrt3200acm, openwrt master
Run tested: plan to test this with seafile as soon as able, using the wrt3200acm above

Description:
Besides the version bump, this backports a patch so that cmake uses the
same so version information as the current autotools, allowing the
switch over to cmake.

6.9.5_rev1 Changelog:

- Look-behind bug: if the look-behind contains a branch with a character
  length of 0 and an anchor is included in the branch, the whole
  look-behind is ignored
- POSIX API disabled by default -- enabled in openwrt, as at least
  libevhtp needs it
- Update Unicode version 13.0.0
- NEW: Code point sequence notation \x{HHHH HHHH ...}, \o{OOOO OOOO ...}
- NEW API: retry limit in search functions
- NEW API: maximum nesting level of subexp call
- Fixed behavior of isolated options in Perl and Java syntaxes.
  /...(?i).../

---

@neheb You wanted to use cmake from the start, now I finally took the time to update it.  The patch was merged upstream.

### DIFF
--- a/libs/oniguruma/Makefile
+++ b/libs/oniguruma/Makefile
@@ -5,12 +5,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=oniguruma
-PKG_VERSION:=6.9.4
+PKG_VERSION:=6.9.5_rev1
 PKG_RELEASE:=1
 
-PKG_SOURCE:=onig-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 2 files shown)


neheb:
Great.

Issue #5550, opened 2018-02-01
libssh2: no crypto backend choice


vampik:
Maintainer: @micmac1 
Environment: x86_64, OpenWRT trunk

Description:

I use a custom OpenWRT trunk build with openssl. Unfortunately, one of the recent commits changed the crypto backend for libssh2 from openssl to mbedtls. Now libssh2 pulls in libmbedtls which is unneeded since I already have openssl.

Could you please make an option to choose libssh2 crypto backend as it is done in all other packages?

micmac1:
Hi Andrey,

I'll send a PR to Jiri this weekend.

Regards,
Seb

micmac1:
Please close this issue when you get around to it.

PR #12118, opened 2020-05-08
libssh2: add mbedtls to PKG_BUILD_DEPENDS

### LINKED ISSUE #12108: libssh2: compile error
Maintainer: @jslachta 
Environment: OpenWrt master r13174-73fa1aba94 on octeon

Description:
Building an image for my ERL fails, due to a compile error in libssh2, which is not even enabled in my .config.

```
stijn@taz ~/Development/LEDE/source $ grep ssh2 .config
# CONFIG_PACKAGE_libssh2 is not set
```

micmac1:
This adds to PKG_BUILD_DEPENDS to prevent built failures. A comment is
left in case anybody stumbles over this in the future.

Closes #12108

Maintainer: @jslachta 
Compile tested: ath79 master
Run tested: No change to package

Description:
Hi all,

With this change mbedtls gets pulled in when needed (fixes build failures like described in the referenced issue). In case the user selects the symbol LIBSSH2_OPENSSL mbedtls doesn't get pulled in, because it's not needed.

I did not bump the revision because package remains unchanged.

Kind regards,
Seb


micmac1:
Had to fix spelling in comment and commit. "built" vs "build" was driving me mad %) I hope spelling is correct now.

micmac1:
And another misspelling removed :)

neheb:
@neheb is more in favor of removing openssl support.

jslachta:
The ability to choose crypto dependency is causing me headaches.

I would like to drop openssl support.

micmac1:
>@neheb is more in favor of removing openssl support.

We had that for a (short) while, until people wanted it back :-)

https://github.com/openwrt/packages/issues/5550

neheb:
Why doesn't libcurl hit this problem?

micmac1:
Maybe if done the other way around (drop mbetls) nobody would notice/complain. Sounds funny, but might be true. So far I can only remember people asking for openssl.

neheb:
Good point. None of the dependencies depend on mbedtls. Some depend on OpenSSL.

micmac1:
> Why doesn't libcurl hit this problem?

You mean the problem nmap has? Because curl uses
"+LIBCURL_SSH2:libssh2". If you mean openssl vs. mbedtls (vs. wolfssl
vs. gnutls), I'd say because curl can also be built without SSL support
(--without-wolfssl, --without-gnutls, --without-ssl, --without-mbedtls).
libssh2 needs a TLS lib on the other hand.

micmac1:
Good morning,

I'm sure you saw the idea from @cotequeiroz in the other issue. I like this better. So I changed this pull request accordingly (PKG_BUILD_DEPENDS not needed anymore).

I also switched the default to OpenSSL, to try it on for size. This way mbedtls is still available as an option, but as @neheb said yesterday, all packages that depend on libssh2 _and_ need their own TLS lib depend on OpenSSL.

What do you think about this?

libs/libssh2/Makefile
```diff
@@ -53,10 +66,10 @@ CMAKE_OPTIONS += \
 	-DENABLE_ZLIB_COMPRESSION=ON \
 	-DCLEAR_MEMORY=ON
 
-ifeq ($(CONFIG_LIBSSH2_OPENSSL),y)
-	CMAKE_OPTIONS += -DCRYPTO_BACKEND=OpenSSL
-else
+ifeq ($(CONFIG_LIBSSH2_MBEDTLS),y)
 	CMAKE_OPTIONS += -DCRYPTO_BACKEND=mbedTLS
+else
+	CMAKE_OPTIONS += -DCRYPTO_BACKEND=OpenSSL
```
neheb, line 72: I assume this can be simplified to

```
-DCRYPTO_BACKEND=$(if $(CONFIG_LIBSSH2_MBEDTLS),mbedTLS,OpenSSL)
```

and placed with the main group.

micmac1:
> >  	CMAKE_OPTIONS += -DCRYPTO_BACKEND=mbedTLS
> +else
> +	CMAKE_OPTIONS += -DCRYPTO_BACKEND=OpenSSL
>
> I assume this can be simplified to
>
> ```
> -DCRYPTO_BACKEND=$(if $(CONFIG_LIBSSH2_MBEDTLS),mbedTLS,OpenSSL)
> ```
>
> and placed with the main group.

Sure, just pushed it.

jslachta:
Merged, thank you!

PR #12121, opened 2020-05-09
acme: Update acme.sh URL, add support for challenge & domain alises.

4wrxb:
Signed-off-by: Will O'Neill <0100wrxb@gmail.com>

Maintainer: @tohojo 
Compile tested: Snapshot, default target (for makefile download of new URL)
Run tested: ramips_mt7621_DEVICE_ubnt_edgerouter-x (for shell scritp run)

Description:
1) The acme.sh project has moved on github to [acmesh-official](https://github.com/acmesh-official/acme.sh)
2) I have added some simple logic in the script to support challenge and domain aliases. These related functions allow using CNAME records from the main domain to point to a different domain where the script can modify text records.

I am preparing an associated LUCI commit for the lua file.

### DIFF
--- a/net/acme/Makefile
+++ b/net/acme/Makefile
@@ -9,10 +9,10 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=acme
 PKG_VERSION:=2.8.6
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://codeload.github.com/Neilpang/acme.sh/tar.gz/$(PKG_VERSION)?
+PKG_SOURCE_URL:=https://github.com/acmesh-official/acme.sh/archive/$(PKG_VERSION).tar.gz
 PKG_HASH:=fd36cb749466296ded521ceacda3fc841ec76be620900d1116e0492d171c1d9f
 PKG_BUILD_DIR:=$(BUILD_DIR)/acme.sh-$(PKG_VERSION)
 
--- a/net/acme/files/acme.config
+++ b/net/acme/files/acme.config
@@ -3,14 +3,31 @@ config acme
        option account_email 'email@example.org'
        option debug 0
 
+config cert 'example_wildcard'
+       option enabled 0
+       option use_staging 1
+       option keylength 2048
+       option update_uhttpd 1
+       option update_nginx 1
+       list domains example.org
+       list domains sub.example.org
+       list domains *.sub.example.org
+       # option user_setup "path-to-custom-setup.script"
+       # option user_cleanup "path-to-custom-cleanup.script"
+       option dns "dns_freedns"
+       list credentials 'FREEDNS_User="ssladmin@example.org"'
+       list credentials 'FREEDNS_Password="1234"'
+       option calias "example.com"
+       option dalias "dalias.example.com"
+
 config cert 'example'
        option enabled 0
        option use_staging 1
        option keylength 2048
        option update_uhttpd 1
        option update_nginx 1
+       list domains example.org
+       list domains sub.example.org
        option webroot ""
-       option dns ""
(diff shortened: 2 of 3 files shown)


net/acme/Makefile
```diff
@@ -9,10 +9,10 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=acme
 PKG_VERSION:=2.8.6
-PKG_RELEASE:=1
+PKG_RELEASE:=1.1
```
tohojo, line 12: PKG_RELEASE is just an integer, so just bump this to '2'

net/acme/files/run.sh
```diff
@@ -268,6 +270,16 @@ issue_cert()
     if [ -n "$dns" ]; then
         log "Using dns mode"
         acme_args="$acme_args --dns $dns"
+        if [ -n "$dalias" ]; then
+            log "Using domain alias for dns mode"
+            acme_args="$acme_args --domain-alias $dalias"
```
tohojo, line 275: Is this going to work when there's multiple domains? The examples on the wiki seem to indicate that the alias args need to go after the first domain definition - not sure if this is strictly a requirement, but if it is, your patch will end up with it in the wrong place...

net/acme/Makefile
4wrxb: Oops, will amend

net/acme/files/run.sh
4wrxb: I read through the options parser and logic of the top-level acme.sh when changing this. Both the challenge alias and domain alias arguments go into a single ordered list (regardless of where they appear within the arguments overall).

I also tested this concept with multiple domains against both single and multiple alias options on my PC shell in various orders with no surprises. The only "odd" part is that the behavior for more domains than challenge/domain aliases currently does not use any alias. So if more than one alias is specified they must all be specified.

To avoid any issues I've simply limited the uci config to one of either challenge alias or domain alias so that only one element ends up in this list. That way acme.sh will apply it to all domains. If anything, the user can simply do one certificate per alias.

Let me know if you would still like it changed, but it would likely take some heavier refactoring.

tohojo: Nah, that's fine; just wanted to make sure you had checked :)

4wrxb:
Isn't the UCI change covered by https://github.com/openwrt/luci/pull/4038 ? I don't see an acme uci file in this repo.

4wrxb:
@tohojo Isn't the UCI change covered by https://github.com/openwrt/luci/pull/4038 ? I don't see an acme uci file in this repo.

tohojo:
It's `files/acme.config`

4wrxb:
Oops, missed that somehow.

I modeled an example off my current current and added it as a distinct wildcard example using DNS. I hope it's more illustrative, but let me know if it's too long.

net/acme/files/acme.config
```diff
@@ -3,14 +3,31 @@ config acme
...
+       list domains example.org
+       list domains sub.example.org
+       list domains *.sub.example.org
+       # option user_setup "path-to-custom-setup.script"
+       # option user_cleanup "path-to-custom-cleanup.script"
+       option dns "dns_freedns"
+       list credentials 'FREEDNS_User="ssladmin@example.org"'
+       list credentials 'FREEDNS_Password="1234"'
+       option calias "exampleorg-acmechallenge.mooo.com"
+       option dalias "exampleorg-acmechallenge.mooo.com"
```
tohojo, line 21: Let's keep this to example.TLD instead of squatting on something that could technically be someone's domain. You can just use example.com as the distinct other domain :)

4wrxb: Good point. I have updated it.

Technically a domain alias should point to a sub domain due to reasons (the docs aren't clear to me... it's either an acme API limitation or an acme.sh DNS API limitation). Sometimes it works to use ".example.com" or "@.example.com". Since run.sh does nothing to try and make it work I just used "dalias.example.com" rather than give a potential problematic example.

PR #11663, opened 2020-03-25
mini_snmpd: Add sysDescr option

vKnmnn:
Maintainer: Marcin Jurkowski 
Compile tested: no need, i think
Run tested: it runs for me on `19.07.2 r10947-65030d81f3 / LuCI openwrt-19.07 branch git-20.081.60866-4778aa6` and `mini_snmpd v1.4-rc1`

Description:
Some monitoring tools will run into errors whe sysDescr field is empty. This was the case for cacti1.2.8 that i'm using to monitor my network. Ideally the sysDescr field would contain some information from the overview page of luci, e.g. the Model and Firmware and Kernel-Version fields, but i'm a stranger to openWRT so i'm unable to do that.

What i did do is enable us to set the sysDescr field as a config option to mini_snmpd and populate it with the default value "OpenWRT"

PR #12123, opened 2020-05-09
mini_snmpd: Add sysDescr option.

vKnmnn:
retrying since i messed up with #11663.

Some monitoring tools will run into errors when sysDescr field is empty. This was the case for cacti 1.2.8 that i'm using to monitor my network. Ideally the sysDescr field would contain some information from the overview of luci, e.g. the Model, Firmware and Kernel-Version fields, but i am stranger to OpenWRT, so I'm unable to do that.

Signed-off-by: Robert von Könemann <lordtaifleh@gmail.com>

Maintainer: @marcin1j
Run tested: OpenWRT 19.07

### DIFF
--- a/net/mini_snmpd/Makefile
+++ b/net/mini_snmpd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mini_snmpd
 PKG_VERSION:=1.4-rc1
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 PKG_MAINTAINER:=Marcin Jurkowski <marcin1j@gmail.com>
 PKG_LICENSE:=GPL-2.0
 PKG_LICENSE_FILES:=COPYING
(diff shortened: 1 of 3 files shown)


vKnmnn:
ah, by the way, @marcin1j i spotted a little bug on line 75. There's an unmatched `]` there.

damadmai:
Do you mean this line?
https://github.com/openwrt/packages/pull/12123/files#diff-78f09756d8ba17d489df983f03361f5eL79

vKnmnn:
yes that's the line
sorry, it was line 75 in a previous commit

marcin1j:
LGTM

Thanks for spotting unmatched bracket. The script surely needs a major rewrite but **the mistake you spotted (which by the way can be traced back to 2016) doesn't affect its behavior**.

In this case closing square bracket is nothing special, just an extra argument (file name to check) to `grep`. Since file `]` doesn't exist, `grep` will emit `ENOENT` error message (silently ignored by `procd`) and grep only in `/proc/net/dev`, which is what author of this part wanted.
That's why this typo was unnoticed for such a long time.

You can open another PR to fix this typo but it's only a matter of code style which is ugly anyway.

neheb:

marcin1j:
Right. @vKnmnn please bump `PKG_RELEASE` in `Makefile`, amend the commit and force-push to your PR branch.

PR #12127, opened 2020-05-10
ModemManager and libqmi stable releases in 19.07

aleksander0m:
Maintainer: @nickberry17


aleksander0m:
@neheb I've also cherry-picked the uClibc fix you did (adding nls include path) even if I believe it's not strictly necessary, given that glib2 wasn't bumped in 19.07. Let me know if that's fine, or if you prefer to skip that one.

neheb:
It’s fine.

PR #12133, opened 2020-05-10
treewide: Update ARM NEON/VFP detection

jefferyto:
Maintainer: @jow- (libpng), @thess & @antonlacon (opus & ffmpeg), @wigyori (mpg123), @tripolar (pulseaudio)
Compile tested: mvebu-cortexa9/armvirt-32, 2020-05-10 snapshot sdk
Run tested: none

Description:
With openwrt/openwrt@8dcc1087602e2dd606e4f6e81a06aee62cfd4f4c, the ARM FPU compiler options are no longer part of `CONFIG_TARGET_OPTIMIZATION`.

This updates various packages that look for NEON/VFP support to search `CONFIG_CPU_TYPE` instead.

### DIFF
--- a/libs/libpng/Makefile
+++ b/libs/libpng/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libpng
 PKG_VERSION:=1.6.37
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@SF/libpng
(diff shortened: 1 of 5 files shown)


neheb:
It's an open question whether the originals are correct. NEON is part of ARM64 aka aarch64.

The bug fix is correct though. Merging.

jefferyto:
I see... is there any documentation that says VPF/NEON are always present for ARMv8/aarch64? I see the `fp` and `simd` [feature modifiers in gcc](https://gcc.gnu.org/onlinedocs/gcc/AArch64-Options.html#aarch64-feature-modifiers) are on by default, but I would feel better if there is some official documentation before making this change.

neheb:
I found this: https://developer.android.com/ndk/guides/cpu-arm-neon

No idea about official documentation.

jefferyto:
From https://developer.arm.com/docs/den0024/a/aarch64-floating-point-and-neon:

> Both floating-point and NEON are required in all standard ARMv8 implementations. However, implementations targeting specialized markets may support the following combinations:
> 
>   * No NEON or floating-point.
>   * Full floating-point and SIMD support with exception trapping.
>   * Full floating-point and SIMD support without exception trapping.

So FP/NEON are required except when they are not required :joy:

neheb:
I assume all OpenWrt devices do. At least the toolchain does :).

jefferyto:
#12152 :wink:

PR #12140, opened 2020-05-11
gerbera: set config file to 600 permissions

neheb:
Only gerbera needs to read the file.

Added config file to conffiles to fix opkg update support.

Added several extra options to the UCI config. Manual editing is still required.

Maintainer: nobody
Compile tested: N/A

It's fairly difficult to get this to option parity with minidlna.

I don't know of a way using sh to insert XML that is not present. It needs to be inserted in the proper location.

### DIFF
--- a/multimedia/gerbera/Makefile
+++ b/multimedia/gerbera/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gerbera
 PKG_VERSION:=1.5.0
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/gerbera/gerbera/tar.gz/v$(PKG_VERSION)?
@@ -41,7 +41,6 @@ endef
 
 define Package/gerbera/conffiles
 /etc/config/gerbera
-/etc/gerbera/config.xml
 endef
 
 CMAKE_OPTIONS += \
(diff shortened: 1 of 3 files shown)


micmac1:
Interesting that the init script runs gerbera (as root) to setup the "$home" dir (configuration I suppose). I'm not sure what xmledit does. How about using "su" instead? Like, if "$home" doesn't exist create the folder, hand it over to gerbera user, then:

su -s /bin/sh -c "gerbera -m \""$home"\" -f '' --create-config > \""$home/config.xml"\" gerbera

(more or less)

Whether xmledit is need afterward I can't say as I don't know what it does.

I've added "su -" to the freeswitch init script recently. To make sure su is available I added "+!BUSYBOX_DEFAULT_SU:shadow-su". It's just a few kilo bytes, so I don't think it matters.

Just an idea :)

Kind regards,
Seb

micmac1:
You could change the umask before the pipe to something more private (and put it back to default after), saving you the chmod command.

neheb:
That xmledit thing doesn’t work. The idea is to edit an XML field.

I think it needs to be redone. Maybe switch to a samba style configuration where a config template is used.

micmac1:
Right. Maybe it's not necessary? I mean users would want this in a place that doesn't get wiped on reboot anyway, right? So they can setup the file while they figure stuff out ;-)

neheb:
The goal is to be able to generate a config straight from UCI and optionally use your own. Currently, the latter must be done.

neheb:
Then again, a lot of configuration is done through the web interface...

micmac1:
>Then again, a lot of configuration is done through the web interface...

True. Depends on user, though. First thing I do on fresh install is turn off uhttpd ;-)

PR #12143, opened 2020-05-11
django: bump to version 3.0.6

commodo:
Maintainer: me
Compile tested: https://github.com/openwrt/openwrt/commit/58bec0d172  x86
Run tested: https://github.com/openwrt/openwrt/commit/58bec0d172 x86

### DIFF
--- a/lang/python/django/Makefile
+++ b/lang/python/django/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=django
-PKG_VERSION:=3.0.5
-PKG_RELEASE:=3
+PKG_VERSION:=3.0.6
+PKG_RELEASE:=1
 
 PYPI_NAME:=Django
-PKG_HASH:=d4666c2edefa38c5ede0ec1655424c56dc47ceb04b6d8d62a7eac09db89545c1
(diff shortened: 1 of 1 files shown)


commodo:
@peter-stadler 
would you be interested in being a co-maintainer on this package?
you seem to be interested in it a bit more than me;

peter-stadler:
Thank you, I am happy if I can help. Although I am not as experienced and you are doing a great job, I see that it is a lot of work and I think I can take care of some parts. So I would be happy co-maintaining this package :-)

commodo:
> Thank you, I am happy if I can help. Although I am not as experienced and you are doing a great job, I see that it is a lot of work and I think I can take care of some parts. So I would be happy co-maintaining this package :-)

Well, you'll see overtime, that maintaining some packages is not a big effort.
Django usually isn't.
But in my case I have plenty of python packages I maintain, and I feel like I should probably distribute them.

Now, in the past I would try to ask someone to completely take over a package, but that doesn't work.
What seems to work, is to propose a co-maintainer role, and see about maybe withdrawing at a later point in time.
I feel that I should say this now, so that things are clearer.
Naturally, this can happy in a few months, a few years. Let's see.
But I don't want to abandon packages just like that.

commodo:
Changelog v1 -> v2:
* add `Peter Stadler <peter.stadler@student.uibk.ac.at>` as co-maintainer;

peter-stadler:
This is good for me.

PR #12146, opened 2020-05-11
ifstat: make dependency on libnetsnmp conditional

kobboi:
The upstream package supports it with a configure switch.
Defaulting to 'with SNMP support' as it was before.

Signed-off-by: Christophe Lermytte <gentoo@lermytte.be>

Maintainer: unknown
Compile tested with IFSTAT_SNMP set and unset and ELF-checked the resulting ifstat binary

### DIFF
--- a/net/ifstat/Makefile
+++ b/net/ifstat/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ifstat
 PKG_VERSION:=1.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=http://gael.roualland.free.fr/ifstat/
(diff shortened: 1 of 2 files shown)


neheb:
Maintainer is @nikil 

This requires a PKG_RELEASE bump.

kobboi:
@neheb Thanks for your feedback. PKG_RELEASE bumped.

net/ifstat/Makefile
```diff
@@ -36,6 +36,16 @@ define Package/ifstat/description
 	interfaces using SNMP.
 endef
 
+define Package/ifstat/config
+	source "$(SOURCE)/Config.in"
+endef
+
+ifeq ($(CONFIG_IFSTAT_SNMP),y)
```
neheb, line 43: This can be replaced with CONFIGURE_ARGS += --with$(if $(CONFIG_IFSTAT_SNMP),,out)-snmp

kobboi: Nifty. Updated.

Issue #12139, opened 2020-05-11
nginx: incorrect worker_processes invocation in init.d script


dl12345:
The init.d script tries to invoke nginx with a number of worker_processes equal to the number of cores. This is incorrect, as the master process is included in the maximum number of processes. The number of worker_processes should be $NCPUS-1

```
May 11 01:31:45 openwrt nginx[6063]: 2020/05/11 01:31:45 [alert] 6063#0: no more than 8 processes can be spawned
May 11 01:31:45 openwrt nginx[6063]: 2020/05/11 01:31:45 [alert] 6063#0: no more than 8 processes can be spawned
May 11 01:31:45 openwrt nginx[6063]: 2020/05/11 01:31:45 [alert] 6063#0: no more than 8 processes can be spawned
May 11 01:31:45 openwrt nginx[6063]: 2020/05/11 01:31:45 [alert] 6063#0: no more than 8 processes can be spawned
May 11 01:31:45 openwrt nginx[6063]: 2020/05/11 01:31:45 [alert] 6063#0: no more than 8 processes can be spawned
May 11 01:31:45 openwrt nginx[6063]: 2020/05/11 01:31:45 [alert] 6063#0: no more than 8 processes can be spawned
May 11 01:31:45 openwrt nginx[6063]: 2020/05/11 01:31:45 [alert] 6063#0: no more than 8 processes can be spawned
May 11 01:31:45 openwrt nginx[6063]: 2020/05/11 01:31:45 [alert] 6063#0: no more than 8 processes can be spawned
```

The following code
```
start_service() {
   ...
    procd_set_param command /usr/sbin/nginx -c "${NGINX_CONF}" \
        -g "daemon off; worker_processes $NCPUS;"
```

should read

```
start_service() {
   ...
    procd_set_param command /usr/sbin/nginx -c "${NGINX_CONF}" \
        -g "daemon off; worker_processes $(expr $NCPUS - 1);"
```

cc @Ansuel 
cc @peter-stadler

pprindeville:
The number of CPU's should be doubled if the cores are hyper-threaded...

dl12345:
Then surely the code needs to check for that? It just tries to blindly instantiate a number of worker processes equal to the number of cores. 

On a non-hyper threaded system, this behaviour results in the errors in my first comment. But even on a hyper threaded system, it would still need to spawn one less process than the total number of virtual cores due to the master process being counted in the maximum allowable total.

dl12345:
[...]

PR #12159, opened 2020-05-13
nginx: do not reduce max processes

peter-stadler:
Maintainer: @Ansuel 
Compile tested: MIPS 74K, Asus RT-N16, master snapshot
Run tested: MIPS 74K, Asus RT-N16, master snapshot, run luci-ssl-nginx

Description: Use the upstream value for NGX_MAX_PROCESSES (=1024) again (fixing #12139). So it is safe to use `worker_processes auto;` (which sets it to the number of CPUs). Also use this directive in the conf instead of calculating the number of CPUs in the init and passing it as argument.
(**Edit:** With safe I mean that there is no error message, I do not see a vulnerability there.)


peter-stadler:
(rebased)

dl12345:
bump....anyone going to review this PR? this fixes a problem that prevents nginx from starting on systems with 8 or more cores. Probably not many of those at the moment, but I'm at least one person who has one. 

@Ansuel

PR #12162, opened 2020-05-13
npupnp updates

neheb:
Maintainer: @ignisf 
Compile tested: ath79

I tried to convert the first to a static library. Unfortunately, it gives linking errors and complains that compilation with -fPIC is needed. Sounds bogus since all libraries are compiled with -fPIC anyway.

ping @medoc92

edit: sorry the second. libnpupnp.

### DIFF
--- a/libs/libnpupnp/Makefile
+++ b/libs/libnpupnp/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libnpupnp
-PKG_VERSION:=2.2.1
+PKG_VERSION:=4.0.3
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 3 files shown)


medoc92:
Couuld you be more specific about the static link errors ? A build log maybe ?

neheb:
https://gist.github.com/neheb/a76ff305eccd9dabe1277aa7654fb248

medoc92:
Are you trying to build a shared libupnpp from a static libnpupnp ?

I knew about the the multiple defined stringlowercmp() etc. and this is
fixed in current git.

About the NetIF symbols: probably the file needs to be at the exact same
version in both libs and it is not.

With the current git code from both libs, it's possible to use the static
versions, but it will remain a bit tricky as they share several object
files (which have to be mostly identical else both may need to be pulled
from the static libs and you get multiply defined symbols).

I am not sure how to fix this. 


Rosen Penev writes:

PR #12175, opened 2020-05-15
collectd: Add collector for lantiq dsl metrics

jekkos:
Signed-off-by: Jeroen Peelaerts <jeroen.peelaerts@gmail.com>

Compile tested: wr9980b
Run tested: wr9980b. Using the metrics in a grafana dashboard to monitor VDSL status

Description:
This addition fetches the lantiq vdsl statistics and sends them to collectd.


jekkos:
Screenshot from 2020-05-15 13-58-59

hnyman:
Your commit also removes the restriction that the exec scripts are not run as root.
At the first glance, I do not like that removal.

You do not even mention that in the commit message or the PR explanation. Sneaky.

Can't you do like the SQM stats module did? (adjust the script's ownership and then run it as nobody)  Or is that because the script actually runs an init script from /etc/init.d  ???

jekkos:
@hnyman Indeed I removed the restriction to not run as root. I have tried to use the nobody user but indeed, it's unable to run the init.d script that dumps the vdsl statistics. 

The rationale behind that decision is that as other frontends (like luci) are using this script, they are probably also running as root. So I'm not sure whether there is an easy way to reuse this without removing this limitation. 

So I'd like to get your input on that, what would be the better way to solve this? I checked whether a setuid on this script was an option, but I'm not sure if that is a good option.

Another way would be to try and fetch the values directly from the dsl_pipe, but again I'm not certain whether that would work as a non root user.

feckert:
I have already implemented this with the [collectd lua extention](https://github.com/openwrt/packages/blob/master/utils/collectd/Makefile#L255). So we could use directly the output from [`/etc/init.d/dsl_control lucistat`](https://github.com/openwrt/openwrt/blob/d1072096f49823eb39357f9555d7854a9c91bcfb/package/network/config/ltq-adsl-app/files/dsl_control). That are only some lines of code. I will post this lines on monday when i am back in the office. I think this is a better alternative.

jekkos:
@feckert Ok great, curious to see how you tackled that!

feckert:
This is my implementation
```
-- Copyright 2020 Florian Eckert <fe@dev.tdt.de>
-- Licensed to the public under the GNU General Public License v2.

local ubus = require("ubus")
local fs = require("nixio.fs")

local hostname_file = "/proc/sys/kernel/hostname"

local informations= {
        {
                name = "signal_attenuation",
                type = "signal_quality",
                unit = "dB"
        },
        {
                name = "noise_margin",
                type = "signal_quality",
                unit = "dB"
        },
        {
                name = "data_rate",
                type = "bitrate",
                unit = "bit/s"
        },
        {
                name = "latency",
                type = "ping",
                unit = "10^-3ms"
        }
}

local function get_values(conn, hostname)
        local host = hostname:sub(1, -2)
        local status = conn:call("luci-rpc", "getDSLStatus", {})

        for _, info in pairs(informations) do
                for _, direction in pairs({"up", "down"}) do
                        local value = string.format("%s_%s", info["name"], direction)
                        if status and status[value] then
                                local t = {
                                        host = host,
                                        plugin = 'dsl',
                                        type = info["type"],
                                        type_instance = value,
                                        values = {status[value]}
                                }
                                collectd.dispatch_values(t)
                        end
                end
        end
end

local function read()
        local hostname = fs.readfile(hostname_file)

        local conn = ubus.connect()
        if not conn then
                return 0
        end

        get_values(conn, hostname)

        return 0
end

collectd.register_read(read)
```

Configuration for collectd
```
LoadPlugin "lua"
<Plugin "lua">
        Script "/usr/lib/collectd-mod-lua/dsl.lua"
</Plugin>
```

jekkos:
Ok the workaround could be to ask for the values through ubus instead? Can it be called w/o running as root? Also which solution would be more efficient? The lua mode does require to have a lua interpreter installed.

feckert:
> Ok the workaround could be to ask for the values through ubus instead?

I think this is the simplest way.

> Can it be called w/o running as root?

I think this is not necessary collectd is running as root anyway

> The lua collectd does require to have a lua interpreter + collectd lua plugin installed, while exec only requires to have collectd-mod-exec.

Size of collectd-mod-lua vs. collect-mod-exec ipkg packages stack on x86_64

exec stack

- collectd-mod-exec = 25KB

lua stack
- collectd-mod-lua = 8KB
- liblua = 73KB
- lua = 6KB

#

Size of collectd-mod-lua vs. collectd-mod-exec ipkg packages stack on lantiq_xrx200

exec stack
- collectd-mod-exec = 26KB

lua stack
- collectd-mod-lua = 8KB
- liblua = 63KB
- lua = 6KB

Normally we have installed a lua interpreter with the liblua on openwrt if we use LuCI or any package that needs lua. So I think that we have a smaller footprint in the end, because we only have to compare **collecd-mod-exec** vs. **collectd-mod-lua**  and so the lua stack footprint is smaller.

> Also which solution would be more efficient?

I think lua because we do not fork so often as in the exec plugin all calls are done with lua

jekkos:
Ah cool nice to hear that! I think most data from dsl driver are digits and in this case i map them in grafana instead. If you check the scripts in openwrt then you can find out what those constants mean.

Regarding cpu i found that my approach is quite heavy, i had to disable some metric collection to get cpu load avg down from 0.7 to 0.3. I'm curious to see how your version would work out.

feckert:
> Regarding cpu i found that my approach is quite heavy, i had to disable some metric collection to get cpu load avg down from 0.7 to 0.3. I'm curious to see how your version would work out.

@jekkos The problem is not only that you use exec. It is also because the data must always be requested from the dsl subsystem via a pipe. And that takes time, it is not very performant.

On my system I use the following lua code [dsl.txt](https://github.com/openwrt/packages/files/5261577/dsl.txt)

You have to change line 58 to use another ubus call. I have written a new ubus call to cache the dsl values. Because I have also other service that need dsl values and the have slowed down my whole system.

jekkos:
@feckert FYI I have applied a recent patch for the dsl_control_pipe data retrieval and I must say that this significantly speeds up the retrievals. You can find it [here](https://patchwork.ozlabs.org/project/openwrt/list/?series=218621)

feckert:
@jekkos I have just seen in [patchwork](http://patchwork.ozlabs.org/project/openwrt/list/?series=220002) that this changes significantly speeds up the execution time to this [change](https://patchwork.ozlabs.org/project/openwrt/list/?series=218621).

dhewg:
Yupp, that set was made prometheus/collectd ;)
Please give it a spin, it should also be possible to get rid of your uid hack

dhewg:
@feckert I've just noticed your name in the ltq-adsl-app commit history. Can you test my patchset with that daemon on supported hardware?

feckert:
> @feckert I've just noticed your name in the ltq-adsl-app commit history. Can you test my patchset with that daemon on supported hardware?

I'm already on it. But first I have to start a build. Would report on Monday

dhewg:
Nice, thanks!

dhewg:
Is the cpe device also called `/dev/dsl_cpe_api0` on that hardware? That wasn't obvious in the define maze...
If not, search for `DSL_CPE_DEVICE_NAME` in `package/network/config/ltq-vdsl-app/src/src/dsl_cpe_ubus.c`, it might need changes then

feckert:
@dhewg My colleague @sch-m has taken on the matter.
He knows more about DSL in our company.
That's why I'm out of here.

jekkos:
I have updated the collector now after flashing the latest changes from @dhewg .. works a lot better now.
Also got rid of the uid hack.. @feckert would there be a chance for this change to be included? Is there any config to be done in case the collector is enabled?

I did need to compile mod-lua into the upgrade image. Installing it afterwards didn't work. Probably because enabling it toggles a linker flag (-llua) on the collectd build. Not sure how these things should be handled within the build system (I have now added a dependency on collectd-mod-lua).

utils/collectd/files/exec-scripts/dslstat.lua
```diff
@@ -0,0 +1,150 @@
...
+	{
+		name = "mode",
+		type = "gauge",
+	},
+	{
+		name = "state",
+		type = "gauge"
+	},
+	{
+		name = "power_state",
```
feckert, line 92: Power state is now a string.

utils/collectd/files/exec-scripts/dslstat.lua
```diff
@@ -0,0 +1,150 @@
...
+	{
+		name = "uptime",
+		type = "uptime"
+	}
+}
+
+
+local function get_values(conn, hostname, variables, status, direction)
+    local host = hostname:sub(1, -2)
+    local is_global = direction ~= ''
```
feckert, line 104: When it comes to the global section. Can you please make it simpler?
Some who read the code have to look twice to understand the code.
For example, you can pass `global` and build your metric with an if.

Pseudo code:
```
if direction = global then
  metric = value
else
  metric = string.format("%s_%s", value, direction)
end

```

feckert (changes requested):
Those are my changes I would like to see for now.
Can you please also use tabs for the whitespaces?
So far you are mixing tabs and space

utils/collectd/files/exec-scripts/dslstat.lua
jekkos: Ok I have rewritten this a bit

utils/collectd/files/exec-scripts/dslstat.lua
jekkos: I tried to change it to string but then saw some collectd errors. Not sure if this a default type in the `types.db`?

utils/collectd/files/exec-scripts/dslstat.lua
```diff
@@ -0,0 +1,150 @@
...
+	},
+	{
+		name = "uptime",
+		type = "uptime"
+	}
+}
+
+
+local function get_values(conn, hostname, variables, status, direction)
+    local host = hostname:sub(1, -2)
```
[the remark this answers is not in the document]
jekkos: Ok moved it to the caller

utils/collectd/files/exec-scripts/dslstat.lua
```diff
@@ -0,0 +1,150 @@
...
+local function read()
+    local hostname = fs.readfile(hostname_file)
+
+    local conn = ubus.connect()
+    if not conn then
+        collectd.log_error("Failed to connect to ubus")
+        return 0
+    end
+
+    local status = conn:call(string.format("dsl", name), "metrics", {})
```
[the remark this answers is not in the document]
jekkos: Removed this

utils/collectd/files/exec-scripts/dslstat.lua
```diff
@@ -0,0 +1,150 @@
...
+
+    local near_errors = status["errors"]["near"]
+    local far_errors = status["errors"]["far"]
+    local down_line = status["downstream"]
+    local up_line = status["upstream"]
+
+    get_values(conn, hostname, errors, near_errors, "near")
```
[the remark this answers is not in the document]
jekkos: Removed the variable

jekkos:
> Those are my changes I would like to see for now.
> Can you please also use tabs for the whitespaces?
> So far you are mixing tabs and space

I tried to fix all the comments. Let me know if there's anything else. Testing the updated version on my wr9980 and looks good for now.

jekkos:
@feckert Any chance to have a second look? Thanks!

feckert (changes requested):
I have now looked at it again.
I am not entirely happy that we are using the `BuildScriptPlugin` install macro here.
I think this should be used for plugins that use by collectd-mod-exec only.

I would therefore introduce a new install macro for lua scripts named `BuildScriptLuaPlugin` and all collectd lua modules install there files into `/usr/lib/collectd-mod-lua/`
With this change we could then add to collectd the following config section 

```
LoadPlugin "lua"
<Plugin "lua">
       BasePath "/usr/lib/collectd-mod-lua"
</Plugin>
```
so all module gets installed and are loaded.

jekkos:
@feckert in the collectd.conf I find `/usr/share/collect-mod-lua` as Basedir. Did you mean that one?

jekkos:
I have added the new macro for LuaScripts now.

utils/collectd/Makefile
```diff
@@ -501,5 +527,5 @@ $(eval $(call BuildPlugin,vmem,virtual memory usage input,vmem,))
 $(eval $(call BuildPlugin,wireless,wireless status input,wireless,))
 $(eval $(call BuildPlugin,write-graphite,Carbon/Graphite output,write_graphite,+PACKAGE_collectd-mod-write-graphite:libpthread))
 $(eval $(call BuildPlugin,write-http,HTTP POST output,write_http,+PACKAGE_collectd-mod-write-http:libcurl))
-
 $(eval $(call BuildScriptPlugin,sqm,SQM/qdisc collection,sqm_collectd,+PACKAGE_collectd-mod-sqm:collectd-mod-exec))
+$(eval $(call BuildScriptLuaPlugin,ltq-dsl,Lantiq DSL collection,dslstat,+PACKAGE_collectd-mod-ltq-dsl:collectd-mod-lua luci-lib-nixio))
```
feckert, line 531: I think you have missed a `+` before luci-lib-nixio
Also the lua ubus dependency is also missing

jekkos: Are you sure this + is needed? The string is inserted in `(4)` below

```
+  $$(call Package/collectd/Default)                                                                                                                                                                                                                                                                                          
+    TITLE:=$(2) plugin                                                                                                                                                                                                                                                                                                       
+    DEPENDS:= $(4)                                                                                                                                                                                                                                                                                                           
+  endef                                                                                                                                                                                                                                                                                                                      
+         
```

jekkos: I tested on my device and this setup is able to run the script correctly, so it seems there is no need to select anything additional for the ubus import?

feckert: Please have look on this line https://github.com/openwrt/packages/blob/15f5340a0c43e06e29a25082627c19185a0dccfb/utils/collectd/Makefile#L479 I think we do here the same as in this line

jekkos: Ok changes were applied.

utils/collectd/Makefile
```diff
@@ -501,5 +527,5 @@ $(eval $(call BuildPlugin,vmem,virtual memory usage input,vmem,))
 $(eval $(call BuildPlugin,wireless,wireless status input,wireless,))
 $(eval $(call BuildPlugin,write-graphite,Carbon/Graphite output,write_graphite,+PACKAGE_collectd-mod-write-graphite:libpthread))
 $(eval $(call BuildPlugin,write-http,HTTP POST output,write_http,+PACKAGE_collectd-mod-write-http:libcurl))
-
 $(eval $(call BuildScriptPlugin,sqm,SQM/qdisc collection,sqm_collectd,+PACKAGE_collectd-mod-sqm:collectd-mod-exec))
+$(eval $(call BuildScriptLuaPlugin,ltq-dsl,Lantiq DSL collection,dslstat,+PACKAGE_collectd-mod-ltq-dsl:collectd-mod-lua +luci-lib-nixio))
```
feckert, line 531: I pulled this into my staging tree and noticed that this packages makes only sense for lantiq xrx200 targets.
So I think we should update the dependency with the additional target dependencies `@TARGET_lantiq_xrx200`

I think we also need to get rid of the dependency against `luci-lib-nixio` in the lua script.
The problem here I see is, that there are people who don't use the LuCI feed and so do not have this package.

We only need this for `local hostname = fs.readfile(hostname_file)` could you have a look to replace this with lua std lib calls ?
https://www.tutorialspoint.com/lua/lua_file_io.htm

jekkos: I managed to remove the `luci-lib-nixio` dependency now. Also I added the dependency on the `xrx200` target, it seems to work fine in the build menu.

jekkos:
The build seems to fail because of this, which I do not think is related to my changes?

 ```
2021-10-01T07:51:38.3697085Z /home/build/openwrt/staging_dir/toolchain-mips_24kc_gcc-11.2.0_musl/bin/../lib/gcc/mips-openwrt-linux-musl/11.2.0/../../../../mips-openwrt-linux-musl/bin/ld: cannot find -llibwebpmux
2021-10-01T07:51:38.3699378Z /home/build/openwrt/staging_dir/toolchain-mips_24kc_gcc-11.2.0_musl/bin/../lib/gcc/mips-openwrt-linux-musl/11.2.0/../../../../mips-openwrt-linux-musl/bin/ld: cannot find -llibwebpmux
2021-10-01T07:51:38.3700756Z collect2: error: ld returned 1 exit status
2021-10-01T07:51:38.3701365Z ninja: build stopped: subcommand failed.
```

**Comment by feckert** [comment]
This has nothing to do with your change
 It has already been `https://github.com/openwrt/packages/commit/401291ba7e97a696d7f7657f2b483781dee980a2`.
@jekkos Can you rebase and push again?

**Comment by feckert** [comment]
@jekkos could you please bump the [PKG_RELEASE](https://github.com/openwrt/packages/blob/master/utils/collectd/Makefile#L12) number again?
Other wise from my point of view the requested change should fit now.
@hnyman @jow- as the package maintainers, do you have any comments on this change?

File: utils/collectd/Makefile [hunk 8]
```diff
@@ -501,5 +527,5 @@ $(eval $(call BuildPlugin,vmem,virtual memory usage input,vmem,))
 $(eval $(call BuildPlugin,wireless,wireless status input,wireless,))
 $(eval $(call BuildPlugin,write-graphite,Carbon/Graphite output,write_graphite,+PACKAGE_collectd-mod-write-graphite:libpthread))
 $(eval $(call BuildPlugin,write-http,HTTP POST output,write_http,+PACKAGE_collectd-mod-write-http:libcurl))
-
```
hnyman, line 504: Please leave the empty line between the regular plugin and these new exec script plugins

feckert:
@jekkos I have now quickly put something together
[dsl.txt](https://github.com/openwrt/packages/files/7337439/dsl.txt)

Just as I would imagine how this should work.
If that works for you, then you can update the pullrequest with my file.
Please check again if the data you need are included

PR #12178, opened 2020-05-16

xinxijishuwyq:

Description:

### DIFF
--- a/lang/perl/patches/999-fix-build-failure-against-gcc-10.patch
+++ b/lang/perl/patches/999-fix-build-failure-against-gcc-10.patch
@@ -1,0 +1,78 @@
+Index: perl-5.28.1/Configure
+===================================================================
+--- perl-5.28.1.orig/Configure	2018-11-02 06:07:20.000000000 +0800
++++ perl-5.28.1/Configure	2020-05-16 12:53:09.219976591 +0800
+@@ -4689,7 +4689,7 @@
+ fi
+ $rm -f try try.*
+ case "$gccversion" in
+-1*) cpp=`./loc gcc-cpp $cpp $pth` ;;
++1.*) cpp=`./loc gcc-cpp $cpp $pth` ;;
+ esac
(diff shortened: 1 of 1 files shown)


xinxijishuwyq:
I used the perl source code in this project (openwrt official) to add the patch, put it in openwrt 19.07 mod by Lean and compiled successfully under the latest version of archlinux.

neheb:
ping @pprindeville

pprindeville:
Please fix the initial comment.

neheb:
Tested on Fedora 32. All perl host packages failed to compile without this.

xinxijishuwyq:
> Tested on Fedora 32. All perl host packages failed to compile without this.


In arch Linux there is the same problem.

xinxijishuwyq:
> I'm not seeing specific handling for `10.*` anywhere... and I don't like the idea of having it default unless it can be shown that this is the right thing to do.


In fact, I don't know how it works. I just made reference to other people's patches. And, it's work in my test environment.

https://github.com/Perl/perl5/pull/17295/files

https://bugs.gentoo.org/attachment.cgi?id=612830

zacharyrs:
At a glance, it looks like the issue was a lack of specificity in the switch cases.
Basically the case checks for version 1 matched version 10, whereas now they keep the period to distinguish between `x.y` and `x0.y`.

That is, prior code was like:
```
case "$gccversion" in
1*)
```

Is now:
```
case "$gccversion" in
1.*)
```

zacharyrs:
@pprindeville - I'm pretty clueless about the right way, but I think this is one of those cases it makes sense.

xinxijishuwyq:
#12262

pprindeville:
> Please fix the initial comment.

@xinxijishuwyq Waiting on this. You can click on "..." on the top right side of the comment box and select "edit" to make changes.

ynezz:
Superseded by https://github.com/openwrt/packages/pull/12391

PR #12180, opened 2020-05-16
libcap-ng: add package libcap-ng

lucize:

Maintainer: me
Tested x86

Needed for dropping privileges at least on libreswan

### DIFF
--- a/libs/libcap-ng/Makefile
+++ b/libs/libcap-ng/Makefile
@@ -1,0 +1,90 @@
+#
+# Copyright (C) 2020 Lucian Cristian
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libcap-ng
+PKG_VERSION:=0.7.10
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://people.redhat.com/sgrubb/libcap-ng
+PKG_HASH:=a84ca7b4e0444283ed269b7a29f5b6187f647c82e2b876636b49b9a744f0ffbf
+
+PKG_MAINTAINER:=Lucian CRISTIAN <lucian.cristian@gmail.com>
+PKG_LICENSE:=GPL-2.0-or-later LGPL-2.1-or-later
+PKG_LICENSE_FILES:=License
+
+PKG_INSTALL:=1
+PKG_BUILD_PARALLEL:=1
+PKG_FIXUP:=autoreconf
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/libcap-ng/Default
+  TITLE:=POSIX capabilities library
+  SECTION:=libs
+  CATEGORY:=Libraries
+  URL:=https://people.redhat.com/sgrubb/libcap-ng/index.html
+endef
+
+define Package/libcap-ng/description/Default
+  POSIX capabilities library
+endef
+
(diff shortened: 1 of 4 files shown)


BKPepe:
We should be careful about adding new libraries as some packages can detect them. That's case of package knot. 

```
Package knot is missing dependencies for the following libraries:
libcap-ng.so.0
```
More details:
https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a53/packages/knot/compile.txt

I will fix it.

libs/libcap-ng/Makefile
```diff
@@ -0,0 +1,90 @@
...
+PKG_VERSION:=0.7.10
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://people.redhat.com/sgrubb/libcap-ng
+PKG_HASH:=a84ca7b4e0444283ed269b7a29f5b6187f647c82e2b876636b49b9a744f0ffbf
+
+PKG_MAINTAINER:=Lucian CRISTIAN <lucian.cristian@gmail.com>
+PKG_LICENSE:=GPL-2.0-or-later LGPL-2.1-or-later
+PKG_LICENSE_FILES:=License
```
BKPepe, line 20: Should be `COPYING`

libs/libcap-ng/Makefile
```diff
@@ -0,0 +1,90 @@
...
+  URL:=https://people.redhat.com/sgrubb/libcap-ng/index.html
+endef
+
+define Package/libcap-ng/description/Default
+  POSIX capabilities library
+endef
+
+define Package/libcap-ng
+  $(call Package/libcap-ng/Default)
+  TITLE += library
```
BKPepe, line 41: I'm not sure if this is all right as this is the same as in `libcap` Makefile.

image

lucize: is the same as is the -ng version, the bin files are different
capsh  getcap  getpcaps  setcap
vs
captest  filecap  netcap  pscap

lucize:
Yes, I didn't compile all the packages but if there is need I will address all the changes

libs/libcap-ng/Makefile
BKPepe: I mean that there is twice library in TITLE.

lucize: Ah, I see, I copied it as the template

neheb:
> We should be careful about adding new libraries as some packages can detect them.
No, we should be careful when adding packages such as knot to disable everything that can potentially picked up. I try to do that with all my packages.

The main reason behind that is that traditionally, packages would sometimes pick up host libraries. Today, that does not seem to be the case.

> That's case of package knot.
> 
> ```
> Package knot is missing dependencies for the following libraries:
> libcap-ng.so.0
> ```
> 
> More details:
> https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a53/packages/knot/compile.txt
> 
> I will fix it.

BKPepe:
With those patches, there is a fuzz and offset.

```
Applying ./patches/add_library_if_header_found.patch using plaintext: 
patching file configure.ac

Applying ./patches/fix-includes.patch using plaintext: 
patching file src/cap-ng.c
Hunk #1 succeeded at 33 with fuzz 2 (offset -1 lines).

Applying ./patches/fix_build_without_threads.patch using plaintext: 
patching file src/cap-ng.c
Hunk #1 succeeded at 167 (offset 1 line).

```

lucize:
I'll address the issues, the patches are from github, after the 0.7.10 tag

PR #12201, opened 2020-05-17
ddns-scripts: Revmove afraid.org-v2-token

youxiaojie:
ddns-scripts: Revmove afraid.org-v2-token
Signed-off-by: You Xiaojie / 尤晓杰 <yxj790222@163.com>
Bad uri format 404 error.

Description:

### DIFF
--- a/net/ddns-scripts/files/services
+++ b/net/ddns-scripts/files/services
@@ -46,7 +46,7 @@
 "afraid.org-basicauth"	"https://[USERNAME]:[PASSWORD]@freedns.afraid.org/nic/update?hostname=[DOMAIN]&myip=[IP]"
 "afraid.org-keyauth"	"https://freedns.afraid.org/dynamic/update.php?[PASSWORD]&address=[IP]"
 "afraid.org-v2-basic"	"https://[USERNAME]:[PASSWORD]@sync.afraid.org/u/?h=[DOMAIN]&ip=[IP]"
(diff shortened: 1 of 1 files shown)


tigerw:
I'm using this...

youxiaojie:
> 
> 
> I'm using this...
you are right. I see the usage in homepage.

https://sync.afraid.org/u/eHhUemFiVlByaWxhYU95aVhuamRYY2VqOjE5MTIwNjcw
Error 404 : Page not found 

interesting. I see the usage in homepage ,but I try it in firefox, got 404.

tigerw:
From what I can tell you also need the `/?address=[IP]` part at the end

youxiaojie:
> From what I can tell you also need the `/?address=[IP]` part at the end

Update Interface v2 TIPS!
Sync v2 is easy, portable, and native IPv6 friendly

Providing the IP is optional in all update URL styles. If you do not include it, automatic server side IP detection will be used.

tigerw:
Ok, it must have the trailing slash, and the secret token you give must be valid, otherwise it will 404. But other than that, this form of update-URL is indeed valid.

youxiaojie:
> 
> 
> Ok, it must have the trailing slash, and the secret token you give must be valid, otherwise it will 404. But other than that, this form of update-URL is indeed valid.

I add a / at the end of url, also 404. the token is valid.fuck afraid.org

PR #11116, opened 2020-01-25
ooniprobe: new packages

ja-pa:
Maintainer: me
Compile tested: Turris Omnia (TOS5), OpenWrt master
Run tested: Turris Omnia (TOS5), OpenWrt master

Description:
This PR adds ~~two packages probe-cli~~ ooniprobe package related to network measurements for OONI (Open Observatory of Network Interference).
It helps researchers collect information about malicious behavior in the public networks (us of DPI, middleboxes, censorship,etc.)

Project website https://ooni.org/post/ https://ooni.org

PR #12217, opened 2020-05-19
measurement-kit: Add static build dependencies

jefferyto:
Maintainer: @ja-pa 
Compile tested: armvirt-64, 2020-05-17 snapshot sdk
Run tested: none

Description:
When measurement-kit is a build dependency (in `PKG_BUILD_DEPENDS`) of another package, this package's build dependencies (in `DEPENDS`) may not be selected and so may not be built.

(This package does not produce a shared library; the `measurement-kit` target package contains a program that is statically linked to the measurement-kit library. Other packages may include `measurement-kit` in their `PKG_BUILD_DEPENDS` to link to the static library but not include `measurement-kit` as a run-time dependency.)

This adds `PKG_BUILD_DEPENDS` to this package to ensure that its dependencies are built in this case.

This also adds `MEASUREMENT_KIT_BUILD_DEPENDS` that dependant packages should select to ensure that this package's dependencies are correctly built. (`libevent2-openssl` and `libevent2-pthreads` need to be selected for the necessary functionality to be enabled when compiling libevent2.)

This also adds `openssl` as an explicit dependency; it appears to be used directly by this package.

### DIFF
--- a/libs/measurement-kit/Makefile
+++ b/libs/measurement-kit/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=measurement-kit
 PKG_VERSION:=0.10.11
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


ja-pa:
Thx for changes, now I'm getting a better picture of what you meant in https://github.com/openwrt/packages/pull/11116#issuecomment-623438762

LGTM

Issue #11977, opened 2020-04-27
libmraa not building


neheb:
Maintainer: @nxhack 
Environment: at least mips64: https://downloads.openwrt.org/snapshots/faillogs/mips64_octeonplus/packages/libmraa/compile.txt

This seems like a node.js API discrepancy.

neheb:
Related: https://github.com/eclipse/mraa/issues/969
https://github.com/swig/swig/issues/1520

I think we should disable building of swig. ping @jefferyto

jefferyto:
I don't have a strong opinion on this, though swig/swig#1746 looks promising.

PR #12225, opened 2020-05-19
libmraa: disable swig

neheb:
swig as used in libmraa is not compatible with node 12.

This is causing build failures on all platforms.

Maintainer: @nxhack


neheb:
Fixes: https://github.com/openwrt/packages/issues/11977

PR #10633, opened 2019-11-25
iputils: Move the iputils tools to the packages feed

nmeyerhans:
Per discussion in openwrt/openwrt#1804, iputils is moving from the main openwrt repository to the packages feed, and is switching from the abandoned skbuff.net upstream to github.com/iputils/iputils

This change import iputils into packages. A subsequent change to openwrt will drop iputils from that repository once this one is merged.

Maintainer: me
Compile tested: armv7l (linksys 1900acs)
Run tested: armv7l (linksys 1900acs), openwrt master

PR #12229, opened 2020-05-20
mwan3: ipv6 improvements

aaronjg:
- Support iputils ping by using `ping6' command
- Don't add link-local routes to tables
- Reduce calls to get routing table by half
- Handle failed ping when quality monitoring

Maintainer: @feckert 
Compile tested: No compile
Run tested: OpenWrt 19.07.2 r10947-65030d81f3 

Description:

- IPV6 pinging was broken using iputils-ping, since it will need an explicit call to ping6 when pinging an ipv6 address.
- ping can also return a non-zero exit code for various reasons, which broke the quality monitoring code. We check for this now and specify max latency
- mwan3 was trying to add all of the ipv6 link local addresses into the various routing tables, which failed with the error "RTNETLINK answers: File exists."
- save the output of `ip list route ...` so we don't need to call it as often

### DIFF
--- a/net/mwan3/files/lib/mwan3/mwan3.sh
+++ b/net/mwan3/files/lib/mwan3/mwan3.sh
@@ -890,12 +890,30 @@ mwan3_set_user_iptables_rule()
 	config_get timeout $1 timeout 600
 	config_get ipset $1 ipset
 	config_get proto $1 proto all
-	config_get src_ip $1 src_ip 0.0.0.0/0
-	config_get src_port $1 src_port 0:65535
-	config_get dest_ip $1 dest_ip 0.0.0.0/0
-	config_get dest_port $1 dest_port 0:65535
+	config_get src_ip $1 src_ip
+	config_get src_iface $1 src_iface
+	network_get_device src_dev $src_iface
+	config_get src_port $1 src_port
+	config_get dest_ip $1 dest_ip
+	config_get dest_port $1 dest_port
 	config_get use_policy $1 use_policy
 	config_get family $1 family any
+
+	[ -z "$dest_ip" ] && unset dest_ip
+	[ -z "$src_ip" ] && unset src_ip
+	[ -z "$ipset" ] && unset ipset
+	[ -z "$src_port" ]  && unset src_port
+	[ -z "$dest_port" ]  && unset dest_port
+	[ "$proto"  != 'tcp' ]  && [ "$proto" != 'udp' ] && {
+		[ -n "$src_port" ] && {
+			$LOG warn "src_port set to '$src_port' but proto set to '$proto' not tcp or udp. src_port will be ignored"
+		}
+		[ -n "$dest_port" ] && {
+			$LOG warn "dest_port set to '$dest_port' but proto set to '$proto' not tcp or udp. dest_port will be ignored"
+		}
+		unset src_port
+		unset dest_port
+	}
 
 	config_get rule_logging $1 logging 0
 	config_get global_logging globals logging 0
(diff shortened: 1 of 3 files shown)


feckert:
Thanks for your suggested changes.
They all sound plausible and are welcome to improve mwan3.
But can you please break them down into three commits?
I'm not a fan of having multiple improvments or issue fixes mixed up in one commit.

Also update the PKG_VERSION:=2.8.5 to PKG_VERSION=2.8.6 and please fix the indentation
* https://github.com/openwrt/packages/pull/12229/commits/665ab20f3235e84e0de8b2ba92af74879ff45bd4#diff-87268a842bab5effbe181f9f18b2d348R117
* https://github.com/openwrt/packages/pull/12229/commits/665ab20f3235e84e0de8b2ba92af74879ff45bd4#diff-87268a842bab5effbe181f9f18b2d348R155

aaronjg:
No problem. Let me know what you think now.

feckert:
Thanks for the rebase. Now the question is if we want to have a dependency on **iputils-ping**.
It seems that the option for ping from the busybox is different for ipv4 and ipv6.
```
BusyBox v1.30.1 () multi-call binary.

Usage: ping [OPTIONS] HOST

Send ICMP ECHO_REQUEST packets to network hosts

        -4,-6           Force IP or IPv6 name resolution
        -c CNT          Send only CNT pings
        -s SIZE         Send SIZE data bytes in packets (default 56)
        -i SECS         Interval
        -A              Ping as soon as reply is recevied
        -t TTL          Set TTL
        -I IFACE/IP     Source interface or IP address
        -W SEC          Seconds to wait for the first response (default 10)
                        (after all -c CNT packets are sent)
        -w SEC          Seconds until ping exits (default:infinite)
                        (can exit earlier with -c CNT)
        -q              Quiet, only display output at start
                        and when finished
         -p HEXBYTE      Pattern to use for payload
```

```
BusyBox v1.30.1 () multi-call binary.

Usage: ping6 [OPTIONS] HOST

Send ICMP ECHO_REQUEST packets to network hosts

        -c CNT          Send only CNT pings
        -s SIZE         Send SIZE data bytes in packets (default 56)
        -i SECS         Interval
        -A              Ping as soon as reply is recevied
        -I IFACE/IP     Source interface or IP address
        -q              Quiet, only display output at start
                        and when finished
        -p HEXBYTE      Pattern to use for payload
```

@aaronjg What version auf ping do you use busybox or iputils-ping

aaronjg:
sendto(0, "\10\0\212\210N\16\0\0\305\267Y\261\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 64, 0, {sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("10.95.129.17")}, 28) = 64
0


sendmsg(3, {msg_name={sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("10.95.129.17")}, msg_namelen=16, msg_iov=[{iov_base="\10\0BJ\vq\0\1n\366\307^z\353\16\0\10\t\n\v\f\r\16\17\20\21\22\23\24\25\26\27"..., iov_len=64}], msg_iovlen=1, msg_control=[{cmsg_len=24, cmsg_level=SOL_IP, cmsg_type=IP_PKTINFO, cmsg_data={ipi_ifindex=if_nametoindex("vpn-vvpnc"), ipi_spec_dst=inet_addr("0.0.0.0"), ipi_addr=inet_addr("0.0.0.0")}}], msg_controllen=24, msg_flags=0}, 0) = 64



In busybox `ping6` is just an alias to `ping -6`,  and the other options liwke `-w` still work. I'm not sure why they aren't documented. 

https://git.busybox.net/busybox/tree/networking/ping.c#n40

Older versions of iputils don't have `ping -6`, so rely on the `ping6` invocation.  (Although newer versions of iputils do not make the ping6 symlink, so this should be changed.)

FURTHERMORE, the version of iputils-ping available from opkg in the 19.0.2 is very old (20101006). 
https://openwrt.org/packages/pkgdata/iputils-ping

In this version, pinging from a source interface is broken. So we can't reliably tell if an interface is down or not. 
https://bugs.openwrt.org/index.php?do=details&task_id=1486
https://github.com/iputils/iputils/issues/55

It looks like iputils has moved from the main library to the packages in the release that came out earlier this week. 

https://github.com/openwrt/packages/pull/10633

I will test everything out there and update the PR accordingly.

aaronjg:
It looks like the iputils package isn't moving out of base and into packages until the next major release. We should force mwan3 to use the busybox ping for now. 

It looks like if a user has iputils installed and is using the master mwan3 version, they pings aren't being sent out of the correct interface and could be misreported as up or down depending on the configuration of the routing table.

feckert:
> It looks like the iputils package isn't moving out of base and into packages until the next major release. We should force mwan3 to use the busybox ping for now.
> 

So I think we should pin the path so we use the busybox ping in `/bin`

> It looks like if a user has iputils installed and is using the master mwan3 version, they pings aren't being sent out of the correct interface and could be misreported as up or down depending on the configuration of the routing table.

If we pin the use of ping busybox this should get fixed.

aaronjg:
Great. So did 35a86bd address the concerns?

I think busybox ping also has issues binding to devices on ipv6, but so does iputils ping, so this may actually be just a fundamental way binding to a device is handled in linux on ipv6 vs ipv4. 

https://github.com/openwrt/packages/issues/12309

aaronjg:
I've made a few more improvements - I can split into a new PR if you prefer:

- Don't mangle outgoing ICMPv6 echo packets from internal IP addresses. This fixes the issue I raised here: https://github.com/openwrt/packages/issues/12309.  

- Add an option for adding rules based on source interface.  I removed the default 0.0.0.0/0 src and destination ip addresses. I don't see how the 'any' family of rules would have worked because it seems like it always required an ipv4 or ipv6 address. With this change, the any family will work again.  I also cleaned up a bunch of repeated code around adding the iptables rules for ipv4/ipv6/any in making the change.

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -969,17 +973,17 @@ mwan3_set_user_iptables_rule()
...
+			[ "$family" == "ipv4" ] && [ "$IPT" == "$IPT6" ] && continue
+			[ "$family" == "ipv6" ] && [ "$IPT" == "$IPT4" ] && continue
+			case $proto in
+				tcp|udp)
 					[ "$global_logging" = "1" ] && [ "$rule_logging" = "1" ] && {
 						$IPT -A mwan3_rules \
 							-p $proto \
-							-s $src_ip \
-							-d $dest_ip $ipset \
+							${src_ip+-s} $src_ip \
```
feckert, line 984: I think there's a colon missing?

feckert:
Could you also add in the commit description https://github.com/openwrt/packages/pull/12229/commits/b4131022913c8eda9c46ec1fb80f8b3611dc720c the comment from https://github.com/openwrt/packages/pull/12229#issuecomment-635751688. So we know why we did it.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: good catch - it actually works without one, which is why I missed it, but it should be there.

aaronjg:
Sure, I also was able to clean up a bunch more repeated code. Please check again.

aaronjg:
How would you feel about deleting the /dev/null redirect on the iptables rules? These should not write anything to stdout or stderr if they are working properly. If they do, something has gone wrong that the user should know about.

aaronjg:
@feckert how does this look now?

feckert:
> How would you feel about deleting the /dev/null redirect on the iptables rules? These should not write anything to stdout or stderr if they are working properly. If they do, something has gone wrong that the user should know about.

I would prefer if we could do this in a second pullrequest.

> @feckert how does this look now?

I'll have to look at it again next week. I'm on holiday at the moment and have no chance to test your improvement.
Thanks for your work. It all looks very good for now without a run time test.

aaronjg:
Sounds good. Enjoy your holiday!

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -339,6 +351,12 @@ mwan3_set_general_iptables()
 					-p ipv6-icmp \
 					-m icmp6 --icmpv6-type 137 \
 					-j RETURN
+				# do not mangle outgoing echo request
+				$IPT6 -A mwan3_hook \
+				      -m set --match-set mwan3_source_v6 src \
+				      -p ipv6-icmp -m icmp6 --icmpv6-type 128 \
```
feckert, line 357: Could we move  `-m icmp6 --icmpv6-type 128` to a new line?

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -872,13 +890,31 @@ mwan3_set_user_iptables_rule()
 	config_get timeout $1 timeout 600
 	config_get ipset $1 ipset
 	config_get proto $1 proto all
-	config_get src_ip $1 src_ip 0.0.0.0/0
-	config_get src_port $1 src_port 0:65535
-	config_get dest_ip $1 dest_ip 0.0.0.0/0
-	config_get dest_port $1 dest_port 0:65535
+	config_get src_ip $1 src_ip
+	config_get src_iface $1 src_iface
```
feckert, line 894: Could we please declare `src_iface` and `src_dev` as a local variable, on the beginning of this function?

feckert:
@aaronjg Can you please change these things? 
When that's done, we can merge them.

Thanks for your work :+1:

aaronjg:
thanks for reviewing. I just made the changes you suggested.

feckert:
@aaronjg Thanks for your work.
I have merged the pullrequest with the commti https://github.com/openwrt/packages/pull/12229/commits/38be40843b97ca3af9ebe37aae8ebfda7b6af65c to get this into mwan3.
But for the future I would suggest to changes this directly and not add an additional change commit.

Issue #12209, opened 2020-05-18
python-openssl: Add dependency on ca-certificates


tiran:
Maintainer: @jefferyto
Environment: all

Description:

The ``python3-openssl`` package and ``python-openssl`` package should depend on  ``ca-certificates`` package. Since Python 2.7.9 and 3.4.0 the ssl module encourages developers to use ``ssl.create_default_context()`` to create a working and securely configured SSL context object. The implementation assumes that the platform has a correctly configured OpenSSL libssl that can load the default trust anchors (root CA certificates) with ``SSL_CTX_set_default_verify_paths()``.

Also see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=960869 for my Debian related bug.

jefferyto:
@commodo Any opinion on this? This sounds reasonable to me though I'm far from an OpenSSL expert.

hnyman:
> should depend on ca-certificates 

Why should it be ca-certificates, and not the pseudo-package "ca-certs" ????

so that both ca-certificates and the newer ca-bundle can fulfill the dependency, and the user is not forced to install also ca-certificates...

If the dependency would be introduced, I suggest that the newish "ca-certs" is used.

https://github.com/openwrt/openwrt/commit/dd299805ad18472a8245b4524a25e4381e166057

tiran:
I didn't know about ``ca-certs``. A generic provider sounds even better to me.

commodo:
`ca-certs` sounds good to me as well;
i think we may need to open this as far back as OpenWrt 18.06?

jefferyto:
@commodo I think updating 18.06 for this makes sense - I'll open some PRs soon.

hnyman:
Note that the newish ca-certs may not exist in the relase branches.

commodo:
We can probably close this now.

PR #7154, opened 2018-10-08
Python: Fix compile of host modules

jan-kardell:
Add -rpath linker option to host build, pointing to staging/hostpkh/lib.
It's needed to find the correct host libs during runtime, without it the
hosts libs may be used instaead, causing failures.

Signed-off-by: Jan Kardell <jan.kardell@telliq.com>

Maintainer: @commodo
Compile tested: (omap, OpenWrt 18.06.01 with packages master on OpenSUSE LEAP 15.0)
Run tested: (put here arch, model, OpenWrt version, tests done)

Description:

PR #7129, opened 2018-10-01
Python3: Fix host build on OpenSUSE

jan-kardell:
The linker option -rpath is required to find libs in staging_dir. Now it
is included when building host modules. Without it the import test of
the _ctypes module would fail, as it uses system libs. The _ctypes module uses
libffi.so.6 from staging, but OpenSUSE LEAP 15 has libffi.so.7.
It will also fail on LEAP 42.x, Fedora28 and 29 and future or old
versions of Ubuntu.

Fix needed in master and 18.06 branches.

Signed-off-by: Jan Kardell <jan.kardell@telliq.com>

Maintainer: me / @\<github-user> (find it by checking history of the package Makefile)
Compile tested: (omap, OpenWrt 18.06.01 with packages master on OpenSUSE LEAP 15.0)
Run tested: (put here arch, model, OpenWrt version, tests done)

Description:

PR #12235, opened 2020-05-20
[openwrt-18.06] python-openssl,python3-openssl: Add dependency on ca-bundle

jefferyto:
Maintainer: me, @commodo 
Compile tested: armvirt-64, 18.06.8 sdk
Run tested: none

Description:
The ssl module assumes OpenSSL can load the default trust anchors (root CA certificates).

From https://github.com/openwrt/packages/issues/12209

This also backports two commits to fix builds for host Python (#7129, #7154). Local builds of python3 fail for me without this. (Not sure if anything is fixed by the commit for python but I don't think it hurts to include it.)


lang/python/python/files/python-package-openssl.mk
```diff
@@ -8,7 +8,7 @@
 define Package/python-openssl
 $(call Package/python/Default)
   TITLE:=Python $(PYTHON_VERSION) SSL module
-  DEPENDS:=+python-light +libopenssl
+  DEPENDS:=+python-light +libopenssl +ca-bundle
```
commodo, line 11: i am little confused by the existence of both `ca-bundle` and `ca-certificates`
but i guess this is fine;

PR #12241, opened 2020-05-20
lxc fixes

neheb:
Maintainer: @ratkaj 
Compile tested: ath79, arc700

This is currently failing on all targets as Lua support was removed upstream.

ping @cotequeiroz I'm not sure the second commit is correct.

### DIFF
--- a/utils/lxc/Makefile
+++ b/utils/lxc/Makefile
@@ -46,24 +46,25 @@ define Package/lxc/Default
   CATEGORY:=Utilities
   TITLE:=LXC userspace tools
   URL:=https://linuxcontainers.org/
-  DEPENDS:=@!arc
+  DEPENDS:=lxc
 endef
 
 define Package/lxc
   $(call Package/lxc/Default)
+  DEPENDS:=@!arc
   MENU:=1
 endef
 
 define Package/lxc-auto
   $(call Package/lxc/Default)
   TITLE:= (initscript)
-  DEPENDS+=lxc-start +lxc-stop
+  DEPENDS+=+lxc-start +lxc-stop
 endef
 
 define Package/lxc-auto/description
  LXC is the userspace control package for Linux Containers, a lightweight
  virtual system mechanism sometimes described as "chroot on steroids".
- This package adds and initscript for starting and stopping the containers
+ This package adds an initscript for starting and stopping the containers
  on boot and shutdown.
 endef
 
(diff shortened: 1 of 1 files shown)


BKPepe:
> Reworked configure options. Several were outdated

Should be in separated commit as it does not belong in lua support commit

BKPepe:
ping @robimarko as he updated LXC in master

dangowrt:
Apparently we also need to disable MIPS16...
```
af_unix.c: In function 'lxc_abstract_unix_set_sockaddr':
af_unix.c:52:1: sorry, unimplemented: -fstack-check=specific not implemented for MIPS16
 }
 ^
af_unix.c:52: confused by earlier errors, bailing out
```

neheb:
Weird. Locally, it compiles with musl. With uClibc-ng, I get the same error. I wonder what's happening...

cotequeiroz:
It seems that disabling mips16 by itself is not enough as it can't be linked against libcap built with mips16.  Perhaps adding this solves the mips16 problem:
```
ifdef CONFIG_USE_MIPS16
  CFLAGS+= -minterlink-mips16
endif
```
I'm wondering 🤔  if this should not be added to `include/package.mk` instead/also.  It may possibly occur in any package that links to libraries using mips16:
```
'-minterlink-compressed'
'-mno-interlink-compressed'
     Require (do not require) that code using the standard
     (uncompressed) MIPS ISA be link-compatible with MIPS16 and
     microMIPS code, and vice versa.

     For example, code using the standard ISA encoding cannot jump
     directly to MIPS16 or microMIPS code; it must either use a call or
     an indirect jump.  '-minterlink-compressed' therefore disables
     direct jumps unless GCC knows that the target of the jump is not
     compressed.

'-minterlink-mips16'
'-mno-interlink-mips16'
     Aliases of '-minterlink-compressed' and
     '-mno-interlink-compressed'.  These options predate the microMIPS
     ASE and are retained for backwards compatibility.
```

I must say I haven't tried this (perhaps later tonight).

@neheb, while I have looked at all 3 commits and found nothing major wrong (description of the commit removing mips16: `s/PKG_NO_MIPS16/PKG_USE_MIPS16=0/`), it was not clear to me which commit to pay special attention to, or why.

neheb:
@cotequeiroz the arc one. One of the config options has !arc. Not sure if that’s problematic.

utils/lxc/Makefile
```diff
@@ -57,7 +57,7 @@ endef
 define Package/lxc-auto
   $(call Package/lxc/Default)
   TITLE:= (initscript)
-  DEPENDS:=+lxc-start +lxc-stop
+  DEPENDS+=lxc-start +lxc-stop
```
cotequeiroz, line 60:
```suggestion
  DEPENDS+=+lxc-start +lxc-stop
```
I had made this suggestion at the old commit, so it was not shown here in the PR.

cotequeiroz:
BTW, I have compile-tested the `-minterlink-mips16` flag using `mipsel-74kc`, and it links OK.

cotequeiroz:
One more thing: luci-app-lxc must depend on @!arc as well, since it is selecting lxc.  Alternatively, it may depend on lxc, without selecting it (change `+lxc` to `lxc` in `LUCI_DEPENDS`), but then it will only be shown if lxc is selected first.

neheb:
Applied changes.

cotequeiroz:
I don't think this was quite ready for prime time.  Sorry, it is my fault, but the mips-specific CFLAGS were not properly guarded.  I'm in the middle of testing a global `-minterlink-mips16` flag--so the proper fix would depend on its fate.  Right now guarding it would work, and later, if the global flag is admitted, we can remove it here.

PR #11793, opened 2020-04-07
python3: symlink 'python' to '/usr/bin/python3'

commodo:
Maintainer: me, @jefferyto 
Compile tested: x86 https://github.com/openwrt/openwrt/commit/02f07c640159ed1ef0698f3a4198cf5363680edd
Run tested: x86 https://github.com/openwrt/openwrt/commit/02f07c640159ed1ef0698f3a4198cf5363680edd

-------------------------------

Time to make it more official.
Running 'python' should get you an interpreter running, and that one is now
Python3.

Signed-off-by: Alexandru Ardelean <ardeleanalex@gmail.com>

PR #12249, opened 2020-05-21
python-packages: Add usr/bin symlinks without "3" suffix

jefferyto:
Maintainer: me, @commodo 
Compile tested: armvirt-64, 2020-05-17 snapshot sdk
Run tested: none

Description:
Previously, binaries installed by Python packages will have a non-suffixed Python 2 version and a suffixed Python 3 version, e.g. `pip` and `pip3`. With the removal of Python 2, the non-suffixed names are no longer taken.

This adds symlinks for the non-suffixed names linking to the suffixed scripts (or in the case of `pip`, `easy_install`, and `python-config`, to the fully-versioned scripts).


jefferyto:
This is basically #11793 but for other Python packages.

PR #12257, opened 2020-05-23
nss: fix build for mips64

lucize:
disable arm32_neon on unsupported target

Maintainer: me 
Tested mvebu
Compile tested octeonplus


neheb:
Normally, --disbale-werror is used with configure scripts. Is that not the case here?

lucize:
the old option was `export NSS_ENABLE_WERROR=0`
nss doesn't have a dynamic configure script, should I re enable the old way ?

neheb:
maybe that would be better.

PR #12263, opened 2020-05-23
Add Google Test packages

Rustaphor:
Append **Google Test** framework. Into target machine will install only shared object libraries `libgtest`, `libgmock`. All headers stay in the _SDK_ that is very convenient.


ja-pa:
You have to sign your commits. Also, your commits should start with ```<package name>: description```
For more info see [CONTRIBUTING.md](https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#commits-in-your-pull-requests-should)

Last note. Please use English in your commit message. OpenWrt is an international project, A lot of people don't understand the Russian language. If we start mixing other languages in commit messages,  commit history will not be easily readable for everyone.

PR #12264, opened 2020-05-23
ctop: don't build on mipsel

cotequeiroz:
Maintainer: @jmarcet 
Compile tested: not working on ramips, mipsel_74kc
Run tested: N/A

Description:
The current version fails to build on mips and also on mipsel, but the
the latter is not masked.

---
I'm not bumping `PKG_RELEASE`, as it should not change any other target.


neheb:
Also fails here: https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a53/packages/ctop/compile.txt
https://downloads.openwrt.org/snapshots/faillogs/arm_arm1176jzf-s_vfp/packages/ctop/compile.txt

PR #12268, opened 2020-05-23
[19.07] gerbera: add taglib support

neheb:
I also backported a libupnp change. Gerbera suggests enabling reuseaddr in libupnp.

I tried backporting libmatroska. Got linking errors.

I also tried to patch Gerbera 1.5.0 to work with GCC7. It's a lot more annoying that I thought. experimental/filesystem are not quite the same. The former lacks some features that the latter has.

Maintainer: nobody
Compile tested: aarch64

ping @BKPepe


neheb:
Erm wrong branch.

PR #12277, opened 2020-05-24
ddns-scripts: Added api token method for cloudflare ddns script v4

xjboss:
ddns-scripts: Added api token method for cloudflare ddns script v4

Maintainer: xjboss
Compile tested: mt7621at, gehua_ghl-r-001, OpenWrt R20.5.20 (unoffical fork https://github.com/coolsnowwolf/lede)
Run tested: mt7621at, gehua_ghl-r-001, OpenWrt R20.5.20)

Description: 
I clone update_cloudflare_com_v4.sh script and modify it' authorized method to api token as update_cloudflare_com_v4_token.sh. I test it on my router it is working. It will coexist with original ddns script.

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -12,7 +12,7 @@ PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.7.8
 # Release == build
 # increase on changes of services files or tld_names.dat
-PKG_RELEASE:=20
+PKG_RELEASE:=21
 
 PKG_LICENSE:=GPL-2.0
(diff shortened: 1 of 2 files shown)


neheb:
Needs a PKG_RELEASE bump.

jefferyto:
I don't think cloning update_cloudflare_com_v4.sh just to change 3 lines of code is a good idea.

Another user has opened #12762 - on that PR I have suggested removing update_cloudflare_com_v4_token.sh and using their change. Interested parties may want to comment on that PR.

PR #12280, opened 2020-05-24
sslh: add transparent proxy support

sgabe:
Compile with `USELIBCAP=1` to make use of POSIX capabilities. This will save the required capabilities needed for transparent proxying for unprivileged processes. Furthermore, upstream will drop support for the `ssl` option in the next future version, hence we should use `tls` instead.

Signed-off-by: Gabor Seljan <sgabe@users.noreply.github.com>

Maintainer: @jmccrohan
Compile tested: MT7621, RB750Gr3, v19.07.3
Run tested: MT7621, RB750Gr3, v19.07.0

Description:
- Transparent proxying is supported with the `--transparent` option added in v1.15 of sslh. This PR updates the init script and the configuration file to support transparent proxying.
- Capabilities are supported since v1.16 of sslh using the `USELIBCAP` compiler flag. This PR adds the `libcap` package as a dependency and enables the `USELIBCAP` flag. Capabilities support is especially needed for transparent proxying while running sslh as the unprivileged nobody user.
- The `ssl` option is deprecated and the legacy support will be removed in sslh v1.21. This PR replaces the old `ssl` option with the new `tls` option that should be used instead.

### DIFF
--- a/net/sslh/Makefile
+++ b/net/sslh/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=sslh
 PKG_VERSION:=v1.20
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://rutschle.net/tech/sslh/
--- a/net/sslh/Makefile
+++ b/net/sslh/Makefile
@@ -26,7 +26,7 @@ define Package/sslh
   CATEGORY:=Network
   SUBMENU:=Routing and Redirection
   TITLE:=SSL/SSH multiplexer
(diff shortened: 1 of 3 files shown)


neheb:
Needs a PKG_RELEASE bump. Might want to put all your changes in one PR.

sgabe:
I have opened separate PRs because this way the proposed changes can be discussed and accepted or rejected independently and I could increment the `PKG_RELEASE` in the order the PRs are merged.

neheb:
All of the changes look fine. I’d put them in one PR. Easier that way.

sgabe:
As per your request, I have pulled all changes into this PR.

BKPepe:
Shouldn't it be cherry-picked into 19.07?

BKPepe:
Done - https://github.com/openwrt/packages/commit/267b49087864adb295834764cb1a04bc64dd9b4a

Fixes warning in OpenWrt 19.07:
`/etc/init.d/sslh start Usage of 'ssl' setting is deprecated and will be removed in v1.21. Please use 'tls' instead `

PR #12288, opened 2020-05-25
libtasn1: fix host build under old GCC

neheb:
prereq-build.mk says that GCC 4.8 and above is supported. GCC 4.8
defaults to std=gnu89, breaking the build. GCC5 defaults to gnu99.

Maintainer: @nmav 
Compile tested: CentOS7


nmav:
LGTM

PR #12294, opened 2020-05-25
[19.07] backport kea changes

neheb:
Maintainer: @rosysong @hbl0307106015
Compile tested: CentOS7

This contains several fixes for the host build that are needed to get it to compile.


hbl0307106015:
ACK, But it seems the build failed on the internal compiler.

neheb:
That’s just boost failing to compile from too little RAM. The build it can handle it.

PR #12295, opened 2020-05-25
ola: fix compilation with GCC4

neheb:
The configure script confusingly sets CXXFLAGS to gnu++11 but does
not use that to check the important stuff. Fix that.

Maintainer: @bk138 
Compile tested: CentOS7


neheb:
I think https://github.com/openwrt/openwrt/commit/3c1d1d4332c7fbaccea01b92b28f6d96f7222492 broke compilation here.

@Cynerd can you take a look?

Cynerd:
I am not sure about https://github.com/openwrt/openwrt/commit/3c1d1d4332c7fbaccea01b92b28f6d96f7222492 breaking this. It might be wrong but there is just few usages of `AX_` macros. In reality there are two uses. They use `AX_HAVE_EPOLL` and `AX_PYTHON_MODULE`.

Is it possible that failure is with python module compilation? In that case I could understand that.

At the same time I do not like following patch: https://github.com/openwrt/packages/blob/master/net/ola/patches/050-always-force-gnu%2B%2B11.patch
As I see it first `PKG_CHECK_MODULES` fails so it continues to second one.. that is going to fail as well because `cppunit` I suspect is not installed. Result is message `OLA requires std::auto_ptr support.` and variable `require_gnu_plus_plus_11` set to default `no`. Subsequent code is because of that skipped:
```
AS_IF([test "x$require_gnu_plus_plus_11" = xyes],
      [CXXFLAGS="$CXXFLAGS -std=gnu++11"])
```
I also feel that this is soo much fragile. If there would be `cppunit` package compile before `ola` then it would pass and enabled gnu++11 correctly.

I would try to just replace whole that check with:
```
require_gnu_plus_plus_11="yes"
```

neheb:
All I know is, build breakage was recent (within 2 weeks). I have no idea what caused it.

Cynerd:
> All I know is, build breakage was recent (within 2 weeks). I have no idea what caused it.

I think that inclusion of `AX_` macros just surfaced broken patch.

I am going to test if my fix helps. If fixing patch I pointed out helps with compilation.

Cynerd:
Eh wtf.. just adding host CFLAG does not help according to my tests. I am stuck with (on host compilation):
```
checking for resolv.h... (cached) yes
checking for res_init() in -lresolv... no
checking for res_init() in -resolv... no
configure: error: Missing resolv, please install it
```
Before anyone asks I am compiling on glibc based system and `libresolv.so` is in `/usr/lib` as expected.

Looking in to configure log it turns our that `CXX` variable is omitted or empty as most of the tests fail with `-o: command not found`.
```
configure:15717:  -o conftest -g -O2 -std=gnu++11 -Wno-error=deprecated-declarations  -I/home/cynerd/openwrt/staging_dir/host/include -I/home/cynerd/openwrt/staging_dir/hostpkg/include -I/home/cynerd/openwrt/staging_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/host/include -std=gnu++11 -L/home/cynerd/openwrt/staging_dir/host/lib -L/home/cynerd/openwrt/staging_dir/hostpkg/lib -L/home/cynerd/openwrt/staging_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/host/lib -Wl,-rpath=/home/cynerd/openwrt/staging_dir/hostpkg/lib -lresolv conftest.cpp  -ldl  >&5
./configure: line 2066: -o: command not found
```

Cynerd:
Ok I see it. It is caused by https://github.com/openwrt/openwrt/commit/3c1d1d4332c7fbaccea01b92b28f6d96f7222492. There is `ACX_PTHREAD` that is aliased to `AX_PTHREAD`. Ola ships with some old version (an probably modified one). The question is why `ax_pthread` is prefered over exact match. Probably a problem of inclusion.

neheb:
I assume @jow-

neheb:
Read it again. Option 1 is probably better. This is the only package that fails. autotools have been a thorn in here for quite a while...

I just pushed an update getting rid of the c++11 patch.

neheb:
I pushed the first patch to fix CentOS7. I have no solution to the actual build failure.

neheb:
I'll merge the second commit as well. Why not.

PR #12298, opened 2020-05-25
linuxptp: fix build failure

pffmachado:
Maintainer: Wojciech Dubowik <Wojciech.Dubowik@neratec.com>
Compile tested: OpenWrt x86 target, glibc, master branch

Description:
 
Build is failing because HWTSTAMP_TX_ONESTEP_P2P is defined anymore on linux net_tstamp.h.

Moreover, the usual way of linuxptp build is looking for system includes  unless user defines differently. That also was tried to fix.


neheb:
Please bump the PKG_RELEASE

neheb:
Maintainer is @wodu

wodu:
I haven't seen any issues with musl build but I haven't yet tried glibc. I will have a look at it today.

wodu:
@pffmachado 
I can't reproduce it with x86 and glibc build. Can you give me your config i.e. .config file? What's your build host?

pffmachado:
It is a host issue! I'm using arch linux with 5.6.14-arch1-1 #1 SMP PREEMPT Wed, 20 May!
On latest linux versions HWTSTAMP_TX_ONESTEP_P2P isn't define anymore!
Anyway, incdefs.sh by default it looks for system includes, I don't think tha is correct! What do you think?

pffmachado:
I have a fedora 30 with linux kernel  5.4.17-100! and it works!
The issue seams to be related with HWTSTAMP_TX_ONESTEP_P2P, which on new kernel version is defined!

wodu:
It shouldn't take includes from the host. I have to have a look at it.

wodu:
Could you try to use
```
diff --git a/net/linuxptp/Makefile b/net/linuxptp/Makefile
index d96808159..4d768dd83 100644
--- a/net/linuxptp/Makefile
+++ b/net/linuxptp/Makefile
@@ -6,6 +6,7 @@
 #
 
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=linuxptp
 PKG_VERSION:=2.0
@@ -39,7 +40,8 @@ endef
 EXTRA_CFLAGS += -DHAVE_CLOCK_ADJTIME -DHAVE_POSIX_SPAWN -DHAVE_ONESTEP_SYNC
 
 MAKE_VARS += \
-       EXTRA_CFLAGS="$(TARGET_CFLAGS) $(TARGET_CPPFLAGS) $(EXTRA_CFLAGS)"
+       EXTRA_CFLAGS="$(TARGET_CFLAGS) $(TARGET_CPPFLAGS) $(EXTRA_CFLAGS)" \
+       KBUILD_OUTPUT="$(LINUX_DIR)"
 
 define Package/linuxptp/install
        $(INSTALL_DIR) $(1)/usr/sbin

```
that's what linuxptp suggests in documentation. Toolchain dir also has linux headers but I am not sure whether they are rebuilt every time.
Then you don't need to patch sources as folder structure is correct. So you could drop the extra patch.

I also see some redefinitions during compilation. As there is no new release I will just update to the latest upstream version. It seems a lot has been fixed there.

pffmachado:
For that work "properly" we need to change the tstamp for tstamp=/include/uapi/linux/net_tstamp.h !!
or some similar as the net_tstam.h only lies on the following locations:

linux-4.19.123/include/uapi/linux/net_tstamp.h:92
or on  this 
linux-4.19.123/user_headers/include/linux/net_tstamp.h:92

wodu:
Yes, I can see now. There is a change in folder structure between kernel versions. So your patch is correct as it support both old and new. So just adapt and fire. I guess uapi is ok.

I will try a new version of linuxptp but probably not today as I want to make sure that it really synchronizes clocks.

pffmachado:
Ok! Just updated!! Have a look!

wodu:
It's ok. Thanks.

wodu:
Ack.

PR #12299, opened 2020-05-25
Patch 2

firewire10000:
libpam is required to prevent errors when compiling


hnyman:
Erroneous. You are trying to push 250 commits to 19.07.

Additionally, you apparent actual addition to samba4 is erroneous (= instead of +)

```
  TITLE+= libs
  DEPENDS:= +libtirpc +libreadline +libpopt +libcap +zlib +libgnutls +libtasn1 +libuuid +libopenssl +USE_GLIBC:libpthread \
  DEPENDS:= =libpam +libtirpc +libreadline +libpopt +libcap +zlib +libgnutls +libtasn1 +libuuid +libopenssl +USE_GLIBC:libpthread \
```
That looks like completely untested.

firewire10000:
This is the post that refers to `libpam.so` missing.

https://forum.openwrt.org/t/samba-4-11-x-package-support-thread/10841/302

I still have to go into `make menuconfig > Libraries` and check `libpam.so` so get past the errors when compiling firmware.

Issue #11383, opened 2020-02-19
ModemManager: MTU not being set based on Bearer info


jonesmz:
Maintainer: @nickberry17
Description:

When Netifd brings a ModemManager based connection up, the bearer informs Netifd of the MTU that the connection supports.

For example, an AT&T cellphone connection frequently has an MTU of below 1500 (e.g. https://www.digi.com/support/knowledge-base/recommended-mtu-mru-settings-on-cellular-networks)

For example:
```
root@:~# mmcli -b 31
  --------------------------------
  General            |  dbus path: /org/freedesktop/ModemManager1/Bearer/31
                     |       type: default
  --------------------------------
  Status             |  connected: yes
                     |  suspended: no
                     |  interface: wwan0
                     | ip timeout: 20
  --------------------------------
  Properties         |    roaming: allowed
  --------------------------------
  IPv4 configuration |     method: static
                     |    address: 10.52.41.111
                     |     prefix: 27
                     |    gateway: 10.52.41.112
                     |        dns: 172.26.38.1, 10.0.0.252
                     |        mtu: 1430
```

However, netifd is not using that information to set the MTU of the wwan0 interface that it brings up.

This results in some traffic getting silently dropped.

jonesmz:
I see in the netifd script that this is marked as a "TODO"
```
modemmanager_connected_method_static_ipv4() {
        local interface="$1"
        local wwan="$2"
        local address="$3"
        local prefix="$4"
        local gateway="$5"
        local mtu="$6"
        local dns1="$7"
        local dns2="$8"
        local metric="$9"

        local mask=""

        [ -n "${address}" ] || {
                proto_notify_error "${interface}" ADDRESS_MISSING
                return
        }

	[ -n "${prefix}" ] || {
                proto_notify_error "${interface}" PREFIX_MISSING
                return
        }

        mask=$(cdr2mask "${prefix}")

        # TODO: mtu reporting in proto handler

        proto_init_update "${wwan}" 1
        proto_set_keep 1
        echo "adding IPv4 address ${address}, netmask ${mask}"
        proto_add_ipv4_address "${address}" "${mask}"
        [ -n "${gateway}" ] && {
                echo "adding default IPv4 route via ${gateway}"
                proto_add_ipv4_route "0.0.0.0" "0" "${gateway}" "${address}"
        }
	[ -n "${dns1}" ] && {
                echo "adding primary DNS at ${dns1}"
                proto_add_dns_server "${dns1}"
        }
[...]

PR #12302, opened 2020-05-25
modemmanager: set interface MTU based on bearer settings

aleksander0m:
Using the same method used by other protocol handlers like uqmi.

Fixes https://github.com/openwrt/packages/issues/11383

Signed-off-by: Aleksander Morgado <aleksander@aleksander.es>

Maintainer: @nickberry17

### DIFF
--- a/net/modemmanager/Makefile
+++ b/net/modemmanager/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=modemmanager
 PKG_VERSION:=1.12.10
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=ModemManager-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://www.freedesktop.org/software/ModemManager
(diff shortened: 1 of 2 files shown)


jonesmz:
Is this really the right approach?

Doesn't this result in a race condition between netifd and the proto script when it comes to creating the interface and setting the parameters?

Also @aleksander0m this does not address the MTU for the pppd call, which you'll see is addressed in some of the patches that I was experimenting with in https://github.com/openwrt/packages/issues/11383

aleksander0m:
> Is this really the right approach?
> 

I don't know, it's the best thing I've found around other protocol handlers needing the same thing.

> Doesn't this result in a race condition between netifd and the proto script when it comes to creating the interface and setting the parameters?
> 

The interface already exists in the system, it's the network interface exposed by the kernel, so there is no race condition at all. The MTU would be configured before netifd knows anything about the IP settings to configure. As long as netifd doesn't reset the MTU while configuring the IP settings, which it doesn't right now, there should be no problem.

> Also @aleksander0m this does not address the MTU for the pppd call, which you'll see is addressed in some of the patches that I was experimenting with in #11383

That is not needed, because when PPP is used, the bearer object in ModemManager won't report an explicit MTU to use. I originally also thought the same and had a similar patch ready, but it really isn't required.

So, all in all, I do believe this patch solved the issue, at least for the static IP settings case. The only "gray area" is when the modem reports that DHCP needs to be used to setup the interface; in this case we're totally relying on the MTU to be configured via DHCP as well. I haven't validated that we're actually requesting MTU config via DHCP in our side, but that's on my TODO list.

PR #12313, opened 2020-05-26
modemmanager: set interface MTU based on bearer settings

aleksander0m:
Using the same method used by other protocol handlers like uqmi.

Fixes https://github.com/openwrt/packages/issues/11383

Signed-off-by: Aleksander Morgado <aleksander@aleksander.es>
(cherry picked from commit 41552c1cc225066bd1f35685ccd5dcf83a21b7e5)

Maintainer: @nickberry17


aleksander0m:
@neheb this is the fix for the stable 19.07 branch

PR #12274, opened 2020-05-24
openfortivpn: update to 1.14.1

qweaszxcdf:
Maintainer: @lucize 
Compile tested: bcm53xx,phicomm-k3,snapshot
Run tested: bcm53xx,phicomm-k3,snapshot, tests done

PR #12324, opened 2020-05-26
openfortivpn: Use netifd for script rather than init script

aaronjg:
By using the netifd for open fortivpn we are able to set up multiple
VPN connections and manage them through the netifd toolset.

This also adds support for binding an openfortivpn client to a given
interface, in which case when that interface comes online, the vpn
will be initiated via a hotplug script.

This is a breaking commit and configurations will need to be migrated
from openfortivpn.config into the /etc/config/networks.

Maintainer: @lucize
Compile tested: 19.07.3-ipq806x
Run tested: 19.07.3-ipq806x


net/openfortivpn/files/openfortivpn.sh
```diff
@@ -0,0 +1,139 @@
...
+        [ -n "$username" ] && append_args -u "$username"
+        [ -n "$password" ] && {
+                umask 077
+                mkdir -p /var/etc
+                pwfile="/var/etc/openfortivpn-$config.passwd"
+                echo "$password" > "$pwfile"
+        }
+
+        [ -n "$local_ip" ] || local_ip=192.0.2.1
+        mkdir -p '/etc/ppp/peers'
```
lucize, line 100: I think this could be also on /var so no flash writes

aaronjg: Will openwrt `pppd` read from /var? I know on other installations, it is fixed to read from /etc/ppp/peers for security reasons.  Would be nice to avoid the flash writes though.

lucize:
so, something looks strange

```
openfortivpn: vvpnc: adding host dependency for 10.160.1.2 on wan at vvpnc
openfortivpn: vvpnc: executing 'openfortivpn '10.160.1.2' '--pppd-ifname=vpn-vvpnc' '--use-syslog' '-c' '/dev/null' '--set-dns=0' '--no-routes' '--pppd-use-peerdns=0' '--ifname=eth1' '--trusted-cert=xxx' '-u' 'xxx' '--pppd-call=vvpnc''
openfortivpn: vvpnc: metric is  10

```
doesn't do anything on forti
added 
```
diagnose debug application sslvpn -1
diagnose debug enable
```

but manually running with the arguments: 
```
[...]
INFO:   Interface vpn-vvpnc is UP.
INFO:   Tunnel is up and running.

```
`[9216:FW01:8]sslvpn_ppp_associate_fd_to_ipaddr:280 associate 10.212.134.200 to tun (ssl.FW01:38)`



```
config interface 'vvpnc'
        option proto 'openfortivpn'
        option server '10.160.1.2'
        option username 'xxx'
        option password 'xxx'
        # optional arguments follow
        option local_ip '192.0.5.1'
        option port '10443'
        option iface_name 'wan'
        option trusted_cert 'xxx'
        option set_dns '0'
        option pppd_use_peerdns '0'
        option metric '10'

```

lucize:
great work ! waiting for the luci-app :)
LGTM

lucize:
@neheb this one can be merged in favor of https://github.com/openwrt/packages/pull/12274

neheb:
Weird that the first commit does not mention a version update.

lucize:
@aaronjg can you amend the first commit ?

aaronjg:
Sure. I inserted a new commit that just has the version update. I also made a change to write the peers file to `/var` rather than `/etc` to eliminate the flash write.

qweaszxcdf:
```
        [ -n $iface_name ] && {
            ping -I $iface_device_name -c 1 -w 10 $server_ip > /dev/null 2>&1 || {
                logger -t "openfortivpn" "$config: failed to ping $server_ip on $iface_device_name"
                sleep 10
                proto_notify_error  "$config" "failed to ping $server_ip on $iface_device_name"
                proto_setup_failed "$config"
                exit 1
            }
        }
```
pingcheck should be optional since some server may not allow ping
and iface_l3_device_name should be used instead of iface_device_name as for pppoe interface ping will always fail

PR #12331, opened 2020-05-26
avahi: fix reliance on config symbol SSP_SUPPORT

dl12345:
The config symbol SSP_SUPPORT is ambiguous and means different
things to different packages: either "toolchain is compiled
with ssp support" or "toolchain uses gcc libssp". The use of the
symbol should be deprecated and the appropriate symbol used
instead.

Signed-off-by: Ian Cooper <iancooper@hotmail.com>

Maintainer: @thess 
Compile tested: mips musl, x86_64 glibc, arc ucLibc with stack protection on and off in all cases
Run tested: x86_64 glibc

Musl uses libssp_nonshared.a, which is automatically linked and does not need to be explicitly specified. If using SSP with glibc or uClibc, they currently use gcc's libssp, which has a similarly named static variant of libssp.

```
 config SSP_SUPPORT
        default y if USE_MUSL || GCC_LIBSSP
        bool
```

However, if we were to switch to using the native ssp implementations of glibc and uClibc, then this will cause an attempted link of a non-existent libssp_nonshared.a. See [here](https://patchwork.ozlabs.org/project/openwrt/patch/KU1PR01MB2022BE75254218CF30E1C9B7ADB00@KU1PR01MB2022.apcprd01.prod.exchangelabs.com/)

This patch switches to using the right symbol needed to base a link decision on and thus will be ignored if gcc's libssp is not being used. It preserves current behaviour intact.

cc @neheb

### DIFF
--- a/libs/avahi/Makefile
+++ b/libs/avahi/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=avahi
 PKG_VERSION:=0.8
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://github.com/lathiat/avahi/releases/download/v$(PKG_VERSION) \
@@ -251,7 +251,7 @@ $(call Package/avahi/Default/description)
 endef
 
 TARGET_CFLAGS += $(FPIC) -DGETTEXT_PACKAGE
-TARGET_LDFLAGS += $(if $(CONFIG_SSP_SUPPORT),-lssp_nonshared)
+TARGET_LDFLAGS += $(if $(CONFIG_GCC_LIBSSP),-lssp_nonshared)
 
 CONFIGURE_ARGS += \
 	--enable-shared \


neheb:
```
~/d/packages (master)> git grep SSP_SUPPORT | wc -l
3
```

```
~/d/packages (master)> git grep CONFIG_GCC_LIBSSP | wc -l
2
```

No real opinion on this.

dl12345:
> No real opinion on this.

It doesn't change current behaviour but will allow us to remove gcc libssp entirely without breaking the compilation of avahi. If you look at the patchworks link I included, that will add some colour to it. There's a cover letter explanation in the patch set sent to the list yesterday you may want to look at (unless you've binned it already in which case I can copy/paste it here)

neheb:
I like removing stuff.

I assume this was tested on problematic platforms like PPC or i386?

dl12345:
> I like removing stuff.
> 
> I assume this was tested on problematic platforms like PPC or i386?

It was tested with all 3 clibs: musl, glibc and uclibc in both modes: where libssp is required and where it's not. The platforms I used were arc, mips, x86_64. uclibc is currently only used for arc.

I think the actual platform is not material, since it's a linking decision of a c library component that will be the same regardless of platform. 

All this change does is to not specify the library explicitly in TARGET_LDFLAGS when using musl (since it's already linked in - see `toolchain/musl/patches/200-add_libssp_nonshared.patch`) and continue to specify it in TARGET_LDFLAGS when using gcc libssp.

So the net result is identical, but the ambiguity between "is stack protection specified for userspace packages" and "link gcc/musl's libssp_nonshared.a" is resolved.

neheb:
Should this be reverted? GCC_LIBSSP is no longer present.

dl12345:
It should not be reverted, The entire line `TARGET_LDFLAGS += $(if $(CONFIG_GCC_LIBSSP),-lssp_nonshared)` should be removed with a new patch. 

The CONFIG_SSP_SUPPORT symbol was previously used in an ambiguous manner, so reverting it will just back out to the ambiguous usage and try to link the non-existent library static libssp_nonshared.a if stack protection is enabled for userspace packages. 

This case was the outlier use of the symbol which was using it to check if it needs to link libssp. It was changed to prevent package breakage when the libssp patch is applied to the main repo. The other uses of CONFIG_SSP_SUPPORT use it to mean "ssp support is enabled"

neheb:
That's what I thought. I'm testing it on i386 with musl and glibc (a platform which needed this flag). So far nothing is breaking.

dl12345:
None of the packages using CONFIG_SSP_SUPPORT or CONFIG_GCC_LIBSSP should break, although we should clean them up and remove these references.

Those packages that use CONFIG_SSP_SUPPORT (tor and openssh) should be patched to refer instead to `!CONFIG_PKG_CC_STACK_PROTECTOR_NONE`

neheb:
Look at https://github.com/neheb/packages/tree/mangix

All those packages build fine. All these SSP workarounds are gone with my branch,

The only thing left to test is my libcxx version bump. It includes -lssp_nonshared currently.

dl12345:
Your libcxx version bump needs to maintain the link of libssp_nonshared.a as it is actually for musl's variant and not for gcc's variant.

It's a coincidence, but musl also puts its ssp into an external library where the static variant has exactly the same name as gcc's static libssp (which is why the avahi Makefile actually previously worked ok before I patched it but would have broken once the libssp patch was applied to the main repo). 

In musl's case, the external ssp library is linked automatically via a patch to the musl toolchain. However, the libcxx bump uses cmake and doesn't pick up the default libs as a result. So keep it the way it is

dl12345:
> 
> All these SSP workarounds are gone with my branch,
> 
much prettier

neheb:
The patch needs to be maintained or the link flag in g++-libcxx?

dl12345:
The 010-ssp.patch needs to stay in

dl12345:
Can you link me your repo with the cxx bump?

neheb:
https://github.com/neheb/openwrt/commits/mangix

dl12345:
It's this that needs to remain in

```
+  if (LIBCXX_HAS_MUSL_LIBC)
+    target_link_libraries(${target} PRIVATE ssp_nonshared)
+  endif()
```

See comment above: https://github.com/openwrt/packages/pull/12331#issuecomment-645691212

neheb:
I removed lssp_nonshared from g++-libcxx. A bunch of packages started throwing chk_fail linker errors. It stays looks like.

dl12345:
> I removed lssp_nonshared from g++-libcxx. A bunch of packages started throwing chk_fail linker errors. It stays looks like.

On a musl build yes, that would be the case

neheb:
Wonder if I should bother trying to limit it to musl in g++-libcxx.

uClibc-ng is marked as BROKEN and glibc is not really supported in any official capacity. It's mainly used for testing purposes AFAIK.

dl12345:
No reason to limit it - you have to make the patch anyway for musl, so I don't see what's gained by limiting it to musl in this way. 

I'm also not sure why uclibc is broken. I actually test built a bunch of uclibc builds for different platforms (after editing the toolchain makefiles to allow this of course) and they all worked fine. Currently uclibc is the only supported libc for ARC builds. 

Anyway, it used to be the case that glibc required a fair bit of effort to get everything building properly. In my 15.05 build I had about 60 different patches to make it build properly. Today, I have a total of 9 commits on top of master and they're all for specific things unrelated to glibc, so it works very nicely.

Glibc is good for x86_64 or machines which are not resource constrained. It's a bit faster than musl. My experience on current master is that glibc works just as well as musl.

neheb:
I don't mean the patch. I mean the g++-libcxx file. If you look at uClibc++, it does this: https://github.com/openwrt/openwrt/blob/master/package/libs/uclibc%2B%2B/Makefile#L54

SSP_LIB is evaluated for the g++-uc file.

uclibc-ng is only in the tree for ARC. Otherwise it would be removed.

I assume it's marked as BROKEN as a massive amount of packages were not building with it until I cleaned it up. I also assume nobody really cares for it. Synopsis really needs to release ARC patches for musl.

The glibc speed thing is not that clear. It is the case with x86 and ARM but on MIPS and other exotic platforms, it's not so clear.

dl12345:
Ah, gotcha. I'd surmise that

```
ifeq ($(CONFIG_USE_MUSL),y)
SSP_LIB=-lssp_nonshared
endif
```
might not even be required. What happens on a musl build if you remove it? I suspect it will link ok, since the musl ALL_LIBS symbol already contains libssp_nonshared.a

dl12345:
> 
> The glibc speed thing is not that clear. It is the case with x86 and ARM but on MIPS and other exotic platforms, it's not so clear.

Just saw the old comments on https://forum.openwrt.org/t/musl-vs-glibc-what-do-you-prefer/12626. I think that about sums it up

neheb:
> Ah, gotcha. I'd surmise that
> 
> ```
> ifeq ($(CONFIG_USE_MUSL),y)
> SSP_LIB=-lssp_nonshared
> endif
> ```
> 
> might not even be required. What happens on a musl build if you remove it? I suspect it will link ok, since the musl ALL_LIBS symbol already contains libssp_nonshared.a

this will be a pain to test. I'll try on my VM later.

With libcxx, removing lssp_nonshared from g++-libcxx causes at least aircrack-ng, powertop, and aria2 to fail to compile under musl.

dl12345:
> 
> With libcxx, removing lssp_nonshared from g++-libcxx causes at least aircrack-ng, powertop, and aria2 to fail to compile under musl.

Yep, but that's with a cmake build. uclibc++ isn't using cmake, so it should not break without this. 

Otherwise every package would need an explicit link to libssp_nonshared.a, which we know they don't

neheb:
SSP_LIB is used to generate the g++-uc file. g++-libcxx is based on g++-uc. It's not used for anything else.

neheb:
tested on i386 musl. Commenting that out breaks compilation on my VM. latest OpenWrt master.

PR #12337, opened 2020-05-27
Lxml bump 4 5 1

commodo:
Maintainer: me, @jslachta 
Compile tested: https://github.com/openwrt/openwrt/commit/4661b053908f2a955a70fcb8a691736a90ce6308 x86
Run tested: https://github.com/openwrt/openwrt/commit/4661b053908f2a955a70fcb8a691736a90ce6308  x86

--------------------------------------------

Took a while to get this running.
It turns out, the fix is mostly in libxslt.
lxml uses the xsltGetProfileInformation() function, which is disabled from libxslt via --without-profiler.
This causes a runtime error, since it cannot find the symbol. 

### DIFF
--- a/lang/python/python-lxml/Makefile
+++ b/lang/python/python-lxml/Makefile
@@ -8,11 +8,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-lxml
-PKG_VERSION:=4.4.2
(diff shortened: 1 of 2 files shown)


hnyman:
@jslachta 
Any objections?

jslachta:
No objections. Thank you!

PR #12346, opened 2020-05-28
acme: remove cron entry on package removal

wt:
Previously, the cron config was not removed on uninstall. This change fixes that.

Signed-off-by: Wren Turkal <wt@penguintechs.org>

Maintainer: unknown
Compile tested: Turris Omnia
Run tested: Turris Omnia

### DIFF
--- a/net/acme/Makefile
+++ b/net/acme/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=acme
 PKG_VERSION:=2.8.6
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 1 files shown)


tohojo:
Pretty sure busybox cron doesn't read /etc/cron.d, so this will only work with a full cron daemon (such as vixie-cron which seems to be what the Omnia distribution installs)

wt:
I have updated the change to simply remove the cron entry on removal of the package. Is this acceptable?

tohojo:
Sure! But please bump the package revision as well :)

wt:
Do you mean the version number of the package defined in the Makefile?

wt:
I think I changed what you want. Please check.

tohojo:
Yup, looks good, thanks!

PR #12348, opened 2020-05-28
[19.07] treewide: use relative include paths for python Makefiles

adrianschmutzler:
This updates the include paths for python(3)-package.mk to be relative
to the package Makefile. If not, in certain cases this will print errors
like the following one:

ERROR: please fix feeds/openwrt/net/freeradius3/Makefile
   - see logs/feeds/openwrt/net/freeradius3/dump.txt for details

In the dump.txt there is the following:

Makefile:42: /mylocalpath/feeds/packages/lang/python/python3-package.mk: No such file or directory
make[1]: *** No rule to make target '/mylocalpath/feeds/packages/lang/python/python3-package.mk'.  Stop.

The relative path is used already in 19.07 for most of the packages, and
has been updated for the packages at hand in master as well:

302f4d17e336 ("libmraa,libupm: Disable default Python package build recipe")
1bc2f4f3c621 ("treewide: Remove Python variants for non-Python packages")

### DIFF
--- a/libs/libmraa/Makefile
+++ b/libs/libmraa/Makefile
@@ -26,8 +26,8 @@ PKG_USE_MIPS16:=0
 
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/cmake.mk
-include $(TOPDIR)/feeds/packages/lang/python/python-package.mk
-include $(TOPDIR)/feeds/packages/lang/python/python3-package.mk
+include ../../lang/python/python-package.mk
+include ../../lang/python/python3-package.mk
 
 CMAKE_OPTIONS=-DENABLEEXAMPLES=0 \
 	-DFIRMATA=ON
--- a/libs/libupm/Makefile
+++ b/libs/libupm/Makefile
@@ -25,8 +25,8 @@ PKG_USE_MIPS16:=0
 
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/cmake.mk
-include $(TOPDIR)/feeds/packages/lang/python/python-package.mk
-include $(TOPDIR)/feeds/packages/lang/python/python3-package.mk
+include ../../lang/python/python-package.mk
+include ../../lang/python/python3-package.mk
 
 UPM_MODULES:= \
 	a110x abp ad8232 adafruitms1438 adafruitss adc121c021 adis16448 ads1x15 adxl335 adxl345 \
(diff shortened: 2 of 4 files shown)


jefferyto:
Apologies for the late response - I think this was fine to merge, but I would have liked to investigate the cause of the errors, since in theory `$(TOPDIR)/feeds/packages/lang/python/python-package.mk` and `../../lang/python/python-package.mk` should have referred to the same file.

hnyman:
> should have referred to the same file.

Not in case the packages repo is mounted differently on somebody's private build system, or in a downstream distro that has modified the build system structure somehow.

Avoiding the explicit assumption of the location of the local repo in the local file system is good (as the reference is internal to this repo).

jefferyto:
I agree this is a good change in general - I misread the original error message:

> Makefile:42: /mylocalpath/feeds/packages/lang/python/python3-package.mk: No such file or directory
make[1]: *** No rule to make target '/mylocalpath/feeds/packages/lang/python/python3-package.mk'. Stop.

to say that the packages repo is named `packages` in feeds.conf, when in fact it is saying the file isn't found in `feeds/packages/...` :man_facepalming:

adrianschmutzler:
@jefferyto This has indeed been encountered in a downstream project, and since I need no python packages at all there I just ignored the errors for quite some time. However, yesterday I became curious and found that it's easy to fix. I don't really understand why it didn't work before myself, but since this is "fixed" in master anyway and it doesn't touch my project at all, I didn't consider it worth further investigation.

FYI: I grepped through 19.07, so after this patch it should be relative-path-only (the majority was relative already).

Despite, it still recommends the old scheme here (also in master):
https://github.com/openwrt/packages/blob/master/lang/python/README.md

Maybe this should be adjusted as well, but I didn't want to do it, as I lack the deeper understanding of the python stuff's organization in OpenWrt.

jefferyto:
A few notes about the readme:

* The section that mentions `include $(TOPDIR)/feeds/packages/lang/python/python3-package.mk` is called [Using Python in external/other package feeds](https://github.com/openwrt/packages/blob/master/lang/python/README.md#using-python-in-externalother-package-feeds)
  * External feeds need this syntax to refer to the Python makefiles in the packages feed
  * The section goes into more detail about how this syntax is fragile and has sample code (using `$(wildcard ...)`) for a more robust method

* In [Building a Python package](https://github.com/openwrt/packages/blob/master/lang/python/README.md#building-a-python-package) / [Include python3-package.mk](https://github.com/openwrt/packages/blob/master/lang/python/README.md#include-python3-packagemk), you will see that it tells you to use `include ../python3-package.mk`.

* You can say that the above section is for Python packages in `lang/python` and perhaps there should be another section of instructions for other packages in the feed. I don't think there is enough demand to justify this work but patches are welcome.

PR #12357, opened 2020-05-29
uacme: remove cron entry on package removal

wt:
Previously, the cron config was not removed on uninstall. This change fixes
that.

Signed-off-by: Wren Turkal <wt@penguintechs.org>

Maintainer: @pesintta 
Compile tested: Turris Omnia
Run tested: None

Description:
The same diff was applied to the acme package as well. I jut noticed that this package had the same issue.


pesintta:
Maintainer for uacme is @lucize 

FWIW the pull-request seems ok to me

lucize:
package bump and LGTM

PR #10797, opened 2019-12-15
smartdns: add new package

pymumu:
Maintainer: me / @\<github-user> (find it by checking history of the package Makefile)
Compile tested: (put here arch, model, OpenWrt version)
ARCH: arm, mips, mipsel, x86, x86_64, aarch64
Model: K3, K3P
Openwrt: 18.0x, 19.0x
Run tested: (put here arch, model, OpenWrt version, tests done)
Model: K3, K3P
Openwrt: 18.0x, 19.0x

Description:
SmartDNS is a local DNS server which accepts DNS query requests from local network clients,
get DNS query results from multiple upstream DNS servers concurrently, and returns the fastest IP to clients.
Unlike dnsmasq's all-servers, smartdns returns the fastest IP. 

Signed-off-by: Nick Peng <pymumu@gmail.com>

PR #12360, opened 2020-05-30
nextdns: Update to version 1.6.4

rs:
Update to version 1.6.4


BKPepe:
Changelog? Indication of compile test and run testing?

rs:
This is automated push of new versions. Nothing changes in the package itself other than the code for the installed binary.

rs:
Added: 

> Refactor the package so init script is installed from the binary instead of an init script embedded in the package.

BKPepe:
> 
> 
> Added:
> 
> > Refactor the package so init script is installed from the binary instead of an init script embedded in the package.

This change is no-go for me. Same reasoning as here:
https://github.com/openwrt/packages/pull/10797#pullrequestreview-335591352

rs:
The current package is broken, it breaks users config on upgrade, so it’s an urgent fix.

Our software deal with many different router firmware and OS. It is unpractical to maintain init scripts and configs in packages for each platform, so the binary can install itself and manage its config, this way the package is just about installing the binary and call install.

To change settings, we tell our users to use `nextdns config set ...` which will interact with the platform config system (icu in the case of openwrt).

OpenWRT was the first firmware we supported, that is why we had a different approach when we first created this package.

Regarding GPL, we are MIT, so I don’t see any incompatibility.

BKPepe:
ping @openwrt/package-maintainers - we are no tracking init changes in this repository anymore?

hnyman:
> we are no tracking init changes in this repository anymore?

There are also other packages that either do not have init scripts or have the init script elsewhere. For example, of the widely used packages SQM (sqm-scripts) has its init script in author's own repo, not here and also the whole install process is defined in the files in the outside repo.

That is rare, but not forbidden.

Having init scripts, default config etc. here makes it easier to track changes or notice possible problems with new global features. Some of the apps with the "outside files" approach have got burned in recent uci/luci changes, as their files have been outside the repo, so that treewide modifications due to new features have not reached those apps. 

So, I would prefer having init scripts etc here, but that is not an absolute requirement.

PR #12361, opened 2020-05-30
[19.07] nextdns: Update to version 1.6.4

rs:
Update to version 1.6.4


rs:
Added: 

> Refactor the package so init script is installed from the binary instead of an init script embedded in the package.

PR #12164, opened 2020-05-14
admin/netdata: Update to 1.22.1

diizzyy:
Maintainer: myself and @BKPepe 
Compile tested: ipq40xx, Linksys EA6350v3, OpenWrt master
Run tested: ipq40xx, Linksys EA6350v3, OpenWrt master

Description:
Update netdata to 1.22.1
Disable cloud functionality
Shoehorn patches from FreeBSD's ports repo
Remove cloud notifications and netdata's self-update feature
json-c is no longer optional

Signed-off-by: Daniel Engberg <daniel.engberg.lists@pyret.net>

PR #12375, opened 2020-06-01
update netdata to 1.22.1

snoopotic:
Hi,
first step is to bump this makefile to version 1.22.1

next is to get the agent and the agent-register script compiled and running as well... then We can watch our router status on nextdata.cloud.
You may help? :)
Thanks!

Compile tested: on debian10 machine cross compield for armv7 (as entware)
Run tested: on an armv7 mipsel asus rt-ac88u

Description:
bumping to 1.22.1 

Looks like it needs some further rework to get the netw cloud integration working. I will try things out. maybe you guys will try, too? :)

### DIFF
--- a/admin/netdata/Makefile
+++ b/admin/netdata/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=netdata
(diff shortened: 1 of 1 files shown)


diizzyy:
Duplicate of #12164

PR #12381, opened 2020-06-02
remove pointless host builds

neheb:
these are completely unused

Maintainer: @srdgame @val-kulkov @jow- 
Compile tested: fedora 32, centos 7.


NeoRaider:
The lua-cjson host build is used by Gluon.

neheb:
Removed commit.

champtar:
> The lua-cjson host build is used by Gluon.

Maybe we can add a comment / link ?

PR #12383, opened 2020-06-02
[19.07] update pigeonhole

neheb:
Maintainer: @flyn-org 

Recently @ja-pa backported dovecot which contains API changes that broke pigeonhole. This updates pigeonhole to fix.


flyn-org:
Looks good. Recommend merge.

PR #12388, opened 2020-06-02
dnsdist: fix compilation with libcxx 10

neheb:
string_view is available with both boost and std.

Maintainer: @James-TR 
Compile tested: ath79

### DIFF
--- a/net/dnsdist/Makefile
+++ b/net/dnsdist/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsdist
 PKG_VERSION:=1.4.0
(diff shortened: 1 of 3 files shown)


James-TR:
There seems to be a bit of contention on this going on with the 1.5.0 alphas where they're fixing this for openbsd. I wonder what the preferred solution would be. It seems upstream would prefer to use std over boost where possible.

neheb:
I found a better alternative to this patch: https://github.com/PowerDNS/pdns/pull/9207

neheb:
Added extra patch to fix compilation with the new 010-libcxx patch. It's a backport.

neheb:
Merging as these are upstream backports.

PR #12389, opened 2020-06-03
xinetd: add uci support

he-ma:
This patch allows xinetd to be used with UCI configuration for services.
The _defaults_ section in the generated xinetd config file is generated as an empty section and the "includedir /etc/xinetd.d" is included too to keep compatibility with existing configurations sitting here.

A (only commented) example config is included.


net/xinetd/files/xinetd.init
```diff
@@ -1,15 +1,119 @@
...
-SERVICE_USE_PID=1
+USE_PROCD=1
 
-start() {
-	service_start /usr/sbin/xinetd -pidfile /var/run/xinetd.pid
+PROG="/usr/sbin/xinetd"
+
+PIDFILE="/var/run/xinetd.pid"
+
+CONF_FILE="/etc/config/xinetd"
```
feckert, line 13: Is this variable needed?
I can only find it once in code https://github.com/openwrt/packages/pull/12389/files#diff-6f5884fed5ef0d2076baa0e53c31f1a9R94

he-ma: Oh yes. I think you are right.
This is obviously a relict from former implementation.
Maybe we can remove it.

feckert:
@neheb I am nor sure if @jmccrohan is still the maintainer of this package.
He has not been active on openwrt for a long time!

If it's all right from your side, I would like to merge this.

neheb:
Go ahead.

PR #12390, opened 2020-06-03
unbound: expose interface-auto to UCI

blocktrron:
Maintainer: @EricLuehrsen 
Compile tested: ipq40xx / AVM 4040
Run tested: ipq40xx / AVM 4040

Description:

This exposes the interface-auto option to UCI.

By default, interface-auto is disabled.
This leads to the DNS reply possibly originating from
a different address then the request was sent to.

Devices with a packet filter might not receive the reply in this case.

Enabling interface-auto ensures the reply is sent with the
source-address the request was sent to.

### DIFF
--- a/net/unbound/files/README.md
+++ b/net/unbound/files/README.md
@@ -259,6 +259,10 @@ config unbound
     Boolean. If enabled version.server, version.bind, id.server, and
     hostname.bind queries are refused.
 
+  option interface_auto '0'
+    Boolean. If enabled DNS replies will have the same source address as
+    the request was sent to.
+
   option listen_port '53'
     Port. Incoming. Where Unbound will listen for queries.
 
--- a/net/unbound/files/unbound.sh
+++ b/net/unbound/files/unbound.sh
@@ -36,6 +36,7 @@ UB_B_NTP_BOOT=1
 UB_B_QUERY_MIN=0
 UB_B_QRY_MINST=0
 UB_B_AUTH_ROOT=0
+UB_B_IF_AUTO=0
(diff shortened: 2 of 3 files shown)


dibdot:
Did you get a confirmation from the maintainer for your proposed change outside of this repo?

EricLuehrsen:
Honestly I was not contacted. If this option has been *tested* to actually work (Unbound documentation is equivocal), then I would actually make this default 'on' or not even an option. Its morally incorrect to respond on an address not queried. I assume OpenWrt later kernel picks don't have errors on the API. I would like to see some verification on common hardware. Also when fixing some GPL-violators drivers things break.

blocktrron:
@EricLuehrsen This has nothing to do with the Kernel, it's just unbound not binding to a specific address when sending out the response. This also happens on Debian.

Imagine a bridge with two addresses. `2001:67c:2ed8:100e::1` and `2001:67c:2ed8:100e::2`. Sending queries to `2001:67c:2ed8:100e::1` results in them potentially being answered from `2001:67c:2ed8:100e::2`, which in a typical client setup will lead to the reply being dropped by the packet filter.

The warning in unbound documentation does not affect us, most likely some other UNIX derivates do not fully support this setting.

The change is working (rolled it out to multiple APs already). I chose the default value to reflect unbounds default so there's not settings changed unexpectedly. However IMHO it would make sense to enabled it by-default, especially as it resembles the behavior of dnsmasq.

BKPepe:
Still it should be approved by the maintainer.

PR #12391, opened 2020-06-03

dwmw2:
The perl Configure file was matching GCC 10 against "1*" and treating it
as GCC 1, causing ABI breakage and segfaults.

Cherry-pick the upstream patch which fixes it to check against (e.g)
"1.*" instead, which will make it work for hundreds more GCC versions
to come.

https://github.com/Perl/perl5/commit/6bd6308fcea3541

Maintainer: @pprindeville @Naoir 
Compile tested: Asus RT-AC85P, OpenWRT master
Run tested: x

This supersedes #12178 in which we're being cruel to a potential contributor by asking them to "fix" their commit message without telling them how. If they knew what we were asking for, they might have done it in the first place.

@xinxijishuwyq here's what I think you were being asked for:
 • Make it clearer that the patch is cherry-picked from upstream, and provide the reference
 • Use the precise patch that was merged upstream (there are a couple of minor differences)
 • Include an explanation of the problem in the OpenWRT commit message.

Thanks for your contribution to OpenWRT!

### DIFF
--- a/lang/perl/patches/999-fix-build-failure-against-gcc-10.patch
+++ b/lang/perl/patches/999-fix-build-failure-against-gcc-10.patch
@@ -1,0 +1,103 @@
+From 6bd6308fcea3541e505651bf8e8127a4a03d22cd Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Petr=20P=C3=ADsa=C5=99?= <ppisar@redhat.com>
+Date: Tue, 12 Nov 2019 09:19:18 +0100
+Subject: [PATCH] Adapt Configure to GCC version 10
+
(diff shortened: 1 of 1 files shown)


pprindeville:
> This supersedes #12178 in which we're being cruel to a potential contributor by asking them to "fix" their commit message without telling them how. If they knew what we were asking for, they might have done it in the first place.

Wow. You mean telling them like I did here?

>> Please fix the initial comment.
> 
> @xinxijishuwyq Waiting on this. You can click on "..." on the top right side of the comment box and select "edit" to make changes.

That was 4 days ago and from what I can tell, you've done nothing with that information.

PR #9797, opened 2019-08-22
[18.06]numpy for Python

jpc:
Maintainer: @commodo
Compile & run tested: OpenWRT 18.06 target-aarch64_cortex-a53_musl NanoPi Neo Core 2

Description:
These are some pretty big hacks required to cross compile numpy. This required both changes to python distutils and hacking platform detection in numpy compilation scripts.

Issues:
1. It compiles and simple use cases work but I have not tested it extensively yet (I did not run the test suite on the device).
2. python2 will probably need a similar distutils fix (I did not need python2 support so I did touch this)
3. I am not sure if this should be merged as is but maybe somebody finds it useful (I would) while we figure out a way to make it less hacky.

PR #12404, opened 2020-06-04
numpy: add new package

commodo:
Maintainer: me
Compile tested: x86 https://github.com/openwrt/openwrt/commit/712e00877d69a6ec5a694c0773c764abab455b6c   ar71xx (also)
Run tested: x86 https://github.com/openwrt/openwrt/commit/712e00877d69a6ec5a694c0773c764abab455b6c

Essentially, this is a re-spin from
  https://github.com/openwrt/packages/pull/9797/

But a really trimmed down version.
Only the Py3 variant is added now, which makes the Makefile really small
now.

Cython is needed on the host, to cythonize some files.

### DIFF
--- a/lang/python/numpy/Makefile
+++ b/lang/python/numpy/Makefile
@@ -1,0 +1,48 @@
+#
+# Copyright (C) 2019 Jakub Piotr Cłapa <jpc@loee.pl>
+# Copyright (C) 2020 Alexandru Ardelean <ardeleanalex@gmail.com>
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=numpy
+PKG_VERSION:=1.18.5
+PKG_RELEASE:=1
+
+PYPI_NAME:=$(PKG_NAME)
+PKG_HASH:=34e96e9dae65c4839bd80012023aadd6ee2ccb73ce7fdf3074c62f301e63120b
+
+PKG_MAINTAINER:=Alexandru Ardelean <ardeleanalex@gmail.com>
+
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE.txt
+PKG_CPE_ID:=cpe:/a:numpy:numpy
+
+# yes, zip... sigh
+PYPI_SOURCE_EXT:=zip
+HOST_PYTHON3_PACKAGE_BUILD_DEPENDS="Cython==0.29.19"
+
+include ../pypi.mk
+include $(INCLUDE_DIR)/package.mk
+include ../python3-package.mk
+
+define Package/python3-numpy
+  SUBMENU:=Python
+  SECTION:=lang
+  CATEGORY:=Languages
+  TITLE:=The fundamental package for scientific computing with Python
+  URL:=https://www.numpy.org/
+  DEPENDS:=@!SOFT_FLOAT +INSTALL_GFORTRAN:libgfortran +python3
+endef
+
+define Package/python3-numpy/description
+NumPy is the fundamental package for array computing with Python.
+
+By default, this package is built without some modules.
+For some modules to be available, the INSTALL_GFORTRAN symbol needs
+to be enabled in the OpenWrt core/toolchain.
+endef
+
+$(eval $(call Py3Package,python3-numpy))
+$(eval $(call BuildPackage,python3-numpy))
+$(eval $(call BuildPackage,python3-numpy-src))


commodo:
@bhack   sorry for the delay on this; i got overloaded and delayed for a while;

@jpc @bhack   do you guys want to be co-maintainers on this?
this would be the first package with 3 co-maintainers; but let's see; this could open-up the room for more scipy stuff;

bhack:
Thanks seems very simplified now with python3 only so I think it is quite easy to maintain.

I guess if we could find a way to merge the  Openblas accelleration PR.

https://towardsdatascience.com/is-your-numpy-optimized-for-speed-c1d2b2ba515

commodo:
> Thanks seems very simplified now with python3 only so I think it is quite easy to maintain.
> 
> I guess if we could find a way to merge the Openblas accelleration PR.
> 
> https://towardsdatascience.com/is-your-numpy-optimized-for-speed-c1d2b2ba515

So, do you want to be added as co-maintainer or not? :)
either is fine from my side; i just need confirmation and an email address [if you do];

jpc:
I am not currently testing this since the project in which I needed numpy went in another direction for now. I also have scipy-ndimage package (only this submodule, not the whole scipy) working on ARM.

I did not worry about BLAS since I was more interested in simple signal/image processing and not linear algebra and matrix inversions. numpy without acceleration may be quite interesting in other similar use cases.

commodo:
i'm reading a "no" from both; though i would have preferred a more direct answer :)
will add myself as single-maintainer for now;

commodo:
Changelog v1 -> v2:
* added myself as single maintainer
* made version of cython fixed; to prevent weird stuff as cython upgrades;
* made `libgfortran` dependency optional; added in description about it

bhack:
What I told Is that I want to see if Openblas PR could land in the repo cause in that case probably I need to co-maintain that. 
I still don't know if I could finalize that PR cause I have not an overview of the total amount of work needed.

neheb:
```
numpy/core/src/npymath/ieee754.c.src:794:33: error: 'FE_DIVBYZERO' undeclared (first use in this function); did you mean 'FD_ZERO'?
     int fpstatus = fetestexcept(FE_DIVBYZERO | FE_OVERFLOW |
                                 ^~~~~~~~~~~~
                                 FD_ZERO
numpy/core/src/npymath/ieee754.c.src:794:33: note: each undeclared identifier is reported only once for each function it appears in
numpy/core/src/npymath/ieee754.c.src:794:48: error: 'FE_OVERFLOW' undeclared (first use in this function); did you mean 'EOVERFLOW'?
     int fpstatus = fetestexcept(FE_DIVBYZERO | FE_OVERFLOW |
                                                ^~~~~~~~~~~
                                                EOVERFLOW
numpy/core/src/npymath/ieee754.c.src:795:33: error: 'FE_UNDERFLOW' undeclared (first use in this function); did you mean 'NPY_FPE_UNDERFLOW'?
                                 FE_UNDERFLOW | FE_INVALID);
                                 ^~~~~~~~~~~~
                                 NPY_FPE_UNDERFLOW
numpy/core/src/npymath/ieee754.c.src:795:48: error: 'FE_INVALID' undeclared (first use in this function); did you mean 'EINVAL'?
                                 FE_UNDERFLOW | FE_INVALID);
                                                ^~~~~~~~~~
                                                EINVAL
numpy/core/src/npymath/ieee754.c.src: In function 'npy_clear_floatstatus_barrier':
numpy/core/src/npymath/ieee754.c.src:814:23: error: 'FE_DIVBYZERO' undeclared (first use in this function); did you mean 'FD_ZERO'?
         feclearexcept(FE_DIVBYZERO | FE_OVERFLOW |
                       ^~~~~~~~~~~~
                       FD_ZERO
numpy/core/src/npymath/ieee754.c.src:814:38: error: 'FE_OVERFLOW' undeclared (first use in this function); did you mean 'EOVERFLOW'?
         feclearexcept(FE_DIVBYZERO | FE_OVERFLOW |
                                      ^~~~~~~~~~~
                                      EOVERFLOW
numpy/core/src/npymath/ieee754.c.src:815:23: error: 'FE_UNDERFLOW' undeclared (first use in this function); did you mean 'NPY_FPE_UNDERFLOW'?
                       FE_UNDERFLOW | FE_INVALID);
                       ^~~~~~~~~~~~
                       NPY_FPE_UNDERFLOW
numpy/core/src/npymath/ieee754.c.src:815:38: error: 'FE_INVALID' undeclared (first use in this function); did you mean 'EINVAL'?
                       FE_UNDERFLOW | FE_INVALID);
                                      ^~~~~~~~~~
                                      EINVAL
numpy/core/src/npymath/ieee754.c.src: In function 'npy_set_floatstatus_divbyzero':
numpy/core/src/npymath/ieee754.c.src:824:19: error: 'FE_DIVBYZERO' undeclared (first use in this function); did you mean 'FD_ZERO'?
     feraiseexcept(FE_DIVBYZERO);
                   ^~~~~~~~~~~~
                   FD_ZERO
numpy/core/src/npymath/ieee754.c.src: In function 'npy_set_floatstatus_overflow':
numpy/core/src/npymath/ieee754.c.src:829:19: error: 'FE_OVERFLOW' undeclared (first use in this function); did you mean 'EOVERFLOW'?
     feraiseexcept(FE_OVERFLOW);
                   ^~~~~~~~~~~
                   EOVERFLOW
numpy/core/src/npymath/ieee754.c.src: In function 'npy_set_floatstatus_underflow':
numpy/core/src/npymath/ieee754.c.src:834:19: error: 'FE_UNDERFLOW' undeclared (first use in this function); did you mean 'NPY_FPE_UNDERFLOW'?
     feraiseexcept(FE_UNDERFLOW);
                   ^~~~~~~~~~~~
                   NPY_FPE_UNDERFLOW
numpy/core/src/npymath/ieee754.c.src: In function 'npy_set_floatstatus_invalid':
numpy/core/src/npymath/ieee754.c.src:839:19: error: 'FE_INVALID' undeclared (first use in this function); did you mean 'EINVAL'?
     feraiseexcept(FE_INVALID);
                   ^~~~~~~~~~
                   EINVAL
```

Needs some fixes.

commodo:
hmm; weird; i'll take a look at those errors;
i did not get them on x86;

neheb:
No these are specific to soft float targets. x86 is hard float.

bhack:
E.g. on MIPS https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mips/bits/fenv.h;h=58acb9cf07631a78459e71c0b38e7e90ca6e51f8;hb=HEAD#l70

commodo:
Changelog v2 -> v3:
* adding package dependency on SOFT_FLOAT symbol; numpy won't be selectable for soft-float targets (like MIPS); this requires a special MIPS build with hard-float; i don't think we should bother [for now] building numpy on targets with no hard-float support

neheb:
I will note that these MIPS routers do not have hardware float support at all. I've bricked mine by changing the toolchain like that.

neheb:
ping @jefferyto

jefferyto:
I'm not sure `SOFT_FLOAT` is the right symbol to depend on, mostly because it is a [config option](https://github.com/openwrt/openwrt/blob/d93da0d0169b2ca90b888e4bb782c259e5ddf2fe/toolchain/Config.in#L16) and the user can change it no matter what the actual hardware support is.

If you really want hard float, I would suggest [`HAS_FPU`](https://github.com/openwrt/openwrt/blob/9a477b833ab2aea96b9eee55acb5f9e7b01b36d8/target/Config.in#L13).

But if it was up to me, I would allow "soft float" devices to install this package (even if it runs a little slower), provided that the kernel FPU emulator is enabled. (This is the real issue with MIPS devices; the FPU emulator is disabled by default in OpenWrt to save space.) For this I would test for [`KERNEL_MIPS_FP_SUPPORT`](https://github.com/openwrt/openwrt/blob/d93da0d0169b2ca90b888e4bb782c259e5ddf2fe/config/Config-kernel.in#L53).

neheb:
SOFT_FLOAT is correct. It's directly responsible for the build failure.

I don't think the issue is the FPU as much as it is getting this thing to compile.

And yes, the lack of kernel FP emulation was why I bricked my router when using a hard float toolchain.

commodo:
well, we can add `@(HAS_FPU||KERNEL_MIPS_FPU_EMULATOR)`
before deciding to go with SOFT_FLOAT, i took a look at how other packages do it, and this was one variant;
i have no strong opinion on one or the other;

commodo:
> well, we can add `@(HAS_FPU||KERNEL_MIPS_FPU_EMULATOR)`
> before deciding to go with SOFT_FLOAT, i took a look at how other packages do it, and this was one variant;
> i have no strong opinion on one or the other;

i was wrong;
@neheb is right;
if taking a look at the code in musl-libc in `./arch/mips/bits/fenv.h`
```
#ifdef __mips_soft_float
#define FE_ALL_EXCEPT 0
#define FE_TONEAREST  0
#else
#define FE_INEXACT    4
#define FE_UNDERFLOW  8
#define FE_OVERFLOW   16
#define FE_DIVBYZERO  32
#define FE_INVALID    64

#define FE_ALL_EXCEPT 124

#define FE_TONEAREST  0
#define FE_TOWARDZERO 1
#define FE_UPWARD     2
#define FE_DOWNWARD   3
#endif

typedef unsigned short fexcept_t;

typedef struct {
        unsigned __cw;
} fenv_t;

#define FE_DFL_ENV      ((const fenv_t *) -1)

```

so, we either patch musl-libc: i wouldn't at this point
or patch numpy: i also wouldn't at this point;

making hard-float a requirement seems to be the easiest for now;
we can extend later;

lang/python/numpy/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+PKG_MAINTAINER:=Alexandru Ardelean <ardeleanalex@gmail.com>
+
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE.txt
+PKG_CPE_ID:=cpe:/a:numpy:numpy
+
+# yes, zip... sigh
+PYPI_SOURCE_EXT:=zip
+HOST_PYTHON3_PACKAGE_BUILD_DEPENDS="Cython==0.29.19"
```
jefferyto, line 23:
```suggestion
HOST_PYTHON3_PACKAGE_BUILD_DEPENDS="Cython>=0.29.14"
```
From [INSTALL.rst.txt](https://github.com/numpy/numpy/blob/v1.18.5/INSTALL.rst.txt#prerequisites).

jefferyto:
Also, 1.19.0 was just released - would be nice if this can be updated as well.

commodo:
> Also, 1.19.0 was just released - would be nice if this can be updated as well.

pfff, last week it was still at 1.18.5
oh well   ¯\_(ツ)_/¯

lang/python/numpy/Makefile
commodo: umm, i would still lock it at a certain version;
no preference which;
while i was reading for LXML: https://lxml.de/build.html#cython   it kinda made me paranoid about allowing \_\_any\_\_ version to be  `>= __version_x_y__`
they're saying:
```
We even encourage you to not install Cython for a normal release build,
as the generated C code can vary quite heavily between Cython versions,
which may or may not generate correct code for lxml. The pre-generated 
release sources were tested and therefore are known to work.
```

so, for numpy they're not that strict, but i versions being updated underneath me, which i can't control, sounds a bit familiar;
also, if the scope of reproduce-able builds comes back around, a fixed cython version makes this build a bit more reproduce-able;

thoughts/opinions?

jefferyto: I think reproducibility is a good thing, but if this dependency is locked at a specific version, I doubt it will be updated with any regularity (if at all) when there are bug fixes, upgrades, etc. (Especially since there will be no uscan emails notifying of upgrades.) For instance, 0.29.20 is already available.

I would prefer updated but not reproducible builds over reproducible but outdated (and potentially buggy) builds, but this is just my personal preference.

I suppose if you add a comment near `PKG_VERSION` reminding us that cython should be checked for upgrades as well, that could also be sufficient.

commodo: > I would prefer updated but not reproducible builds over reproducible but outdated (and potentially buggy) builds, but this is just my personal preference.

we can argue the buggy element both ways; i've had issues where `new` was `buggy`;
so, there was X that was working, it updated underneath me, and X + 1 was buggy;
had to wait for X + 2, to be fixed again, while temporarily patching X+1 to work;

> I suppose if you add a comment near `PKG_VERSION` reminding us that cython should be checked for upgrades as well, that could also be sufficient.


though, one thing that could work, is to have a global `CYTHON_HOST:="Cython>=0.29.14"` variable, as a sort of global version switch for all Cython host builds;
and keep it in `python-package.mk`, or  `cython.mk` or whichever;

the obvious benefit of this is uniformity for Cythonized builds; and better global control over updates, for updating locked versions;



> I think reproducibility is a good thing, but if this dependency is locked at a specific version, I doubt it will be updated with any regularity (if at all) when there are bug fixes, upgrades, etc. (Especially since there will be no uscan emails notifying of upgrades.) For instance, 0.29.20 is already available.

coming back to this;
i really liked the reproduce-ability project;
too bad, no more effort was spent on this;
and i tried to help with it; i even sent some patches to upstream Python for this;
so, i would like to keep things reproce-able as much as possible;

jefferyto: I think "working" and "not working" are not binary states. A version of cython that works with numpy may not work with lxml; a version that works with both may not work for another package. Within one package, a version that works with feature A may not work with feature B. (I don't think simple run-tests actually tell us if a package is fully functional.)

I don't want to over-complicate this (for now anyway :joy:). I'm fine with your choice of `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` :smile:

jefferyto:
1.19.0 was [released](https://pypi.org/project/numpy/#history) 3 weeks after 1.18.5.

commodo:
> I suppose the version bump can come later.

oh;
i was working on this; but there are many things that have piled up, so i am slower these days;
will do a version-bump PR;
i usually try to run-test it on an x86 VM, and this MIPS hard-float slowed things down a bit;

PR #12407, opened 2020-06-04
Update Domoticz to 2020.2, and its dependencies

dwmw2:
Maintainer: @dwmw2 
Compile tested: mipsel_24kc / Asus RT-AC85P / OpenWRT master
Run tested: mipsel_24kc / Asus RT-AC85P / OpenWRT master. Run Domoticz with Onkyo and ZWave devices.

Description:
Take ownership of Domoticz and Open-ZWave now that @stintel doesn't want them any more. I do still use them on OpenWRT in production.

Update Domoticz to 2020.2, and fix up its dependencies somewhat. Including the new minizip and cereal packages since the Domoticz tarball doesn't include those any more.


libs/minizip/Makefile
```diff
@@ -0,0 +1,84 @@
...
+
+PKG_SOURCE:=$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/nmoinvaz/minizip/archive/
+PKG_HASH:=f64b2d228a03673a8ec36a53402e2108437226fd9170d49b7f0c0bca94708b93
+
+PKG_MAINTAINER:=David Woodhouse <dwmw2@infradead.org>
+PKG_LICENSE:=Zlib
+PKG_LICENSE_FILES:=LICENSE
+
+CMAKE_INSTALL:=1
```
neheb, line 21: Erm CMAKE_INSTALL is PKG_INSTALL + InstallDev handling.

You're including a custom InstallDev section. Either that should be removed or CMAKE_INSTALL (PKG_INSTALL is already default).

libs/minizip/Makefile
```diff
@@ -0,0 +1,84 @@
...
+PKG_BUILD_PARALLEL:=1
+
+CMAKE_OPTIONS:=-DINSTALL_INC_DIR=/usr/include/minizip -DMZ_BZIP2=OFF -DBUILD_SHARED_LIBS=yes
```
neheb, line 24: for consistency, shared libs should be set to ON.

that inc dir is curious. Do you mean it's required to be installed on the host?

libs/minizip/Makefile
```diff
@@ -0,0 +1,84 @@
...
+  minizip is a zip manipulation library written in C that is supported on Windows, macOS, and Linux
+endef
+
+define Package/minizip-dev/description
+  minizip is a zip manipulation library written in C that is supported on Windows, macOS, and Linux
+  This package includes the development support files.
+endef
+
+define Build/InstallDev
+	mkdir -p $(1)/usr/include/
```
neheb, line 56: $(INSTALL_DIR) should be used instead of mkdir -p.

libs/minizip/Makefile
```diff
@@ -0,0 +1,84 @@
...
+	$(CP)   $(PKG_INSTALL_DIR)/usr/include/minizip \
+		$(1)/usr/include/
+	mkdir -p $(1)/usr/lib
+	$(CP)   $(PKG_INSTALL_DIR)/usr/lib/libminizip.so* \
+		$(1)/usr/lib/
+	mkdir -p $(1)/usr/lib/pkgconfig
+	$(CP)   $(PKG_INSTALL_DIR)/usr/lib/pkgconfig/minizip.pc \
```
neheb, line 63: One problem I've seen with CMake packages is faulty pkgconfig file handling. See https://github.com/openwrt/packages/blob/master/libs/pugixml/Makefile#L47 for what I mean. Make sure the pkgconfig files are set up correct. That is, no libdir=/usr/lib or such.

Basically, I have to add that InstallDev section instead of just using CMAKE_INSTALL.

libs/cereal/Makefile
```diff
@@ -26,7 +26,9 @@ CMAKE_OPTIONS += -DCMAKE_CXX_FLAGS=-latomic -DWITH_WERROR=OFF -DSKIP_PORTABILITY
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/cmake.mk
 
-define Package/cereal-dev
+
+define Package/cereal
+  HIDDEN:=1
```
neheb, line 31: I'd rather use BUILDONLY. It accomplishes the same thing I think.

libs/minizip/Makefile
dwmw2: No, it does indeed get installed in the staging dir, we want it in ${staging_dir}/usr/include/minizip not ${staging_dir}/usr/include which is the default. Partly because that's just cleaner, and partly because Domoticz has a bug where it doesn't actually honour the pkg-config at all and just *assumes* it's in `minizip/` (which I'm fixing in https://github.com/domoticz/domoticz/pull/4213 but which we don't need to wait for).

libs/minizip/Makefile
dwmw2: I don't think I need the custom InstallDev section. Will drop it.

libs/minizip/Makefile
dwmw2: So... it's OK for `prefix` and `exec_prefix` to be set explicitly to `/usr` (because they're overridden?) but other things like `includedir` have to be specified in terms of `${prefix}` instead of absolute paths?

libs/cereal/Makefile
dwmw2: OK.

stintel:
Thanks for taking over. As I'm no longer maintainer, you do not need my ACK, but you have it anway.
Please squash the 3 cereal commits into 1 cereal commit, the 2 minizip commits into another single minizip commit, etc. You get the idea.

dwmw2:
Yeah, I'll squash them shortly but wanted to leave them like that for incremental review. Thanks.

dwmw2:
Squashed. What is the process for requesting commit access? And in the meantime can someone merge it (unless you have further comments) please?

libs/minizip/Makefile
neheb: Absolute paths can be used, but they cannot point to the host. pkgconf is more strict about this than pkg-config.

libs/cereal/Makefile
```diff
@@ -0,0 +1,44 @@
+#
+# Copyright (C) 2020 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
```
BKPepe, line 5: https://github.com/openwrt/packages/blame/master/CONTRIBUTING.md#L30

libs/minizip/Makefile
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright (C) 2020 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
```
BKPepe, line 5: https://github.com/openwrt/packages/blame/master/CONTRIBUTING.md#L30

dwmw2: Well, these were mostly lifted from existing files © OpenWRT but I don't object to claiming ownership of them. Fixed.

dwmw2:
> What is the process for requesting commit access?

Let me turn that implicit request into an explicit one: Please can I have commit access?

BKPepe:
This is the second pull request and I would be rather careful.

stintel:
> This is the second pull request from you and I would be rather careful. There is nothing wrong to send a pull request to see if it was built by CircleCI or not.

```
git clone https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git
cd linux
git log --author="David Woodhouse"
```

I think he knows what he is doing and granted the commit access request.

PR #12408, opened 2020-06-04
adguardhome: initial package recipe

dobo90:
AdGuardHome is a network-wide ads and trackers blocking DNS server.
After installing it with opkg, start it like every service:
`/etc/init.d/adguardhome start`

In order to complete the installation visit http://{YOUR_ROUTERS_IP}:3000.

Then you can setup dnsmasq to forward DNS traffic to AdGuardHome:
```
uci -q delete dhcp.@dnsmasq[0].server
uci add_list dhcp.@dnsmasq[0].server=127.0.0.1#{PORT_SET_DURING_INSTALL}
uci set dhcp.@dnsmasq[0].noresolv=1
uci commit dhcp
/etc/init.d/dnsmasq restart
```

Signed-off-by: Dobroslaw Kijowski <dobo90@gmail.com>

Maintainer: me (if someone more experienced than me volunteers I will happily give this package for maintenance)  
Compile tested: OpenWrt latest master, mipsel_24kc, Xiaomi Redmi AC2100 
Run tested: OpenWrt latest master, mipsel_24kc, Xiaomi Redmi AC2100

Description:
This is my first time creating recipe for an OpenWrt package so forgive me if I've missed something. It was a little bit tricky to create it as it uses compile-time generated assets like .js files by npm and .go files by packr.
As mentioned above if someone wants to make it proper or maintain it feel free to take it.

### DIFF
--- a/net/adguardhome/Makefile
+++ b/net/adguardhome/Makefile
@@ -1,0 +1,66 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=adguardhome
+PKG_VERSION:=0.103.3
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_VERSION:=v$(PKG_VERSION)
+PKG_SOURCE_URL:=https://github.com/AdguardTeam/AdGuardHome
+PKG_MIRROR_HASH:=edf00ede9d208b7f78574e267ad2617dd4d4f06548cb2e6369cc1ad702427d9a
+
+PKG_LICENSE:=GPL-3.0-only
+PKG_LICENSE_FILES:=LICENSE.txt
+PKG_MAINTAINER:=Dobroslaw Kijowski <dobo90@gmail.com>
+
+PKG_BUILD_DEPENDS:=golang/host node/host packr/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/AdguardTeam/AdGuardHome
+GO_PKG_EXCLUDES:=dhcpd/standalone
+GO_PKG_LDFLAGS_X:=main.version=$(PKG_VERSION) main.channel=release main.goarm=$(GO_ARM)
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/adguardhome
+	SECTION:=net
+	CATEGORY:=Network
+	TITLE:=Network-wide ads and trackers blocking DNS server
+	URL:=https://github.com/AdguardTeam/AdGuardHome
+	DEPENDS:=$(GO_ARCH_DEPENDS) +ca-bundle
+endef
+
+define Package/adguardhome/conffiles
+/etc/adguardhome.yaml
+endef
+
+define Package/adguardhome/description
+Free and open source, powerful network-wide ads and trackers blocking DNS server.
+endef
+
+define Build/Compile
+	( \
+		pushd $(PKG_BUILD_DIR) ; \
+		npm --prefix client ci ; \
+		npm --prefix client run build-prod ; \
+		packr -z -v -i home ; \
+		popd ; \
+		$(call GoPackage/Build/Compile) ; \
+	)
+endef
+
+define Package/adguardhome/install
+	$(call GoPackage/Package/Install/Bin,$(1))
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_BIN) ./files/adguardhome.init $(1)/etc/init.d/adguardhome
+endef
+
+$(eval $(call GoBinPackage,adguardhome))
+$(eval $(call BuildPackage,adguardhome))
--- a/net/adguardhome/files/adguardhome.init
+++ b/net/adguardhome/files/adguardhome.init
@@ -1,0 +1,17 @@
+#!/bin/sh /etc/rc.common
+
+PROG=/usr/bin/AdGuardHome
+USE_PROCD=1
+
+# starts after network starts
+START=21
+# stops before networking stops
+STOP=89
+
+start_service() {
+  procd_open_instance
+  procd_set_param command "$PROG" -c /etc/adguardhome.yaml -w /tmp/adguardhome --no-check-update
+  procd_set_param stdout 1
+  procd_set_param stderr 1
+  procd_close_instance
+}


neheb:
ping @jefferyto

net/adguardhome/Makefile
```diff
@@ -0,0 +1,67 @@
...
+PKG_LICENSE_FILES:=LICENSE.txt
+PKG_MAINTAINER:=Dobroslaw Kijowski <dobo90@gmail.com>
+
+PKG_BUILD_DEPENDS:=golang/host node/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/AdguardTeam/AdGuardHome
+GO_PKG_EXCLUDES:=dhcpd/standalone
+GO_PKG_LDFLAGS:=-s -w
```
jefferyto, line 27: The build system strips binaries by default so these flags are not necessary.

net/adguardhome/Makefile
```diff
@@ -0,0 +1,67 @@
...
+GO_PKG_LDFLAGS_X:=main.version=$(PKG_VERSION) main.channel=release
```
jefferyto, line 28: The upstream Makefile [also sets](https://github.com/AdguardTeam/AdGuardHome/blob/bab7b2dc65004d7c9661a590965f710584f063b9/Makefile#L29) `main.goarm` - do you know if this is necessary here?

net/adguardhome/Makefile
```diff
@@ -0,0 +1,67 @@
...
+define Package/adguardhome/description
+Free and open source, powerful network-wide ads and trackers blocking DNS server.
+endef
+
+define Build/Compile
+( \
+  cd $(PKG_BUILD_DIR) ; \
+  npm --prefix client ci ; \
+  npm --prefix client run build-prod ; \
+  GOOS=$$$$(go env GOOS) GOARCH=$$$$(go env GOARCH) GO111MODULE=off go get -v github.com/gobuffalo/packr/... ; \
```
jefferyto, line 54: I would prefer adding a separate package (with a known version/commit) for packr and using it as a host tool instead of downloading and building it here.

net/adguardhome/files/adguardhome.init
```diff
@@ -0,0 +1,15 @@
...
+USE_PROCD=1
+
+# starts after network starts
+START=21
+# stops before networking stops
+STOP=89
+
+start_service() {
+  procd_open_instance
+  procd_set_param command "$PROG" -c /etc/adguardhome.yaml -w /tmp/adguardhome --no-check-update
```
jefferyto, line 13: I see that the config file [will be created](https://github.com/AdguardTeam/AdGuardHome/wiki/Configuration#configuration-file) if it does not exist, but are there any sensible values we can provide in a default config file that would make setting up the service easier? (I haven't looked into this software enough to know if there are any default values that make sense.)

net/adguardhome/Makefile
dobo90: GOARM isn't set anywhere in Makefile so I've skipped it.

net/adguardhome/Makefile
dobo90: Done.

net/adguardhome/Makefile
dobo90: GOARM isn't set anywhere in Makefile so I've skipped it.

net/adguardhome/Makefile
dobo90: I've tried to do it but I'm sure that I haven't done it properly (you can check it here https://gist.github.com/dobo90/4c0a78dff4bd30358f38aa5cd95727b8). I haven't found any host package using golang. So I've copy paste a lot of variables form golang-package.mk and golang-values.mk. Any suggestion how to make it look good are welcome.

On the other side: I saw that recent version of AdGuardHome has introduced usage of go generate. I've updated the version of the software and made Makefile use the go generate. There was an issue that it was compiling packr for the target architecture but I've pushed a separate commit to address this.
I've made it download specified version of packr. Does this meet your requirements?

net/adguardhome/files/adguardhome.init
dobo90: The software generates the config file after first run. I can run AdGuardHome manually then copy the config file into repository. But IMHO it would be painful in maintenance.

net/adguardhome/Makefile
jefferyto: For release builds, `GOARM` is set by [goreleaser](https://github.com/AdguardTeam/AdGuardHome/blob/39f2d5c4ae0f04b3598e7f96720d8872f90f1ccf/.goreleaser.yml#L32-L35).

When using make, you are expected to set the various `GO*` variables on the command line when [cross building](https://github.com/AdguardTeam/AdGuardHome#building), including `GOARM`.

net/adguardhome/Makefile
jefferyto: I can see packr here would be pinned to a known version, but I'm still not crazy about (re-)building a host tool every time this package is built.

I took a brief look at your packr Makefile gist - thanks for attempting it. I think I thought host Go packages were already possible but obviously I was wrong :joy:

I think going forward, there may be more host tools written in Go. Let me think about how it can be done in a similar manner as target Go packages (probably with something like a golang-host.mk include file).

net/adguardhome/files/adguardhome.init
jefferyto: If there is only a partial config in the config file, is it smart enough to fill in the missing options with their default values? If so, then we can specify only the options we want to override; this should be easier to maintain.

net/adguardhome/Makefile
dobo90: You're perfectly right! I've missed that. It should be fine now.

net/adguardhome/files/adguardhome.init
dobo90: The thing is that when the config file isn't preset, the application starts setup wizard where you setup login/password, listening IPs/ports, etc. It would be hard to deploy a default config (especially with hardcoded user and encrypted password).
I've created a patch which makes AdGuardHome accept an empty configuration file. So, when the /etc/adguardhome.yaml is empty the application will start setup wizard. Hopefully this solution would be enough for you. Now, we have a config file (it's empty but at least it is present) inside the package archive.

graysky2:
@dobo90 - I haven't dug into your PR closely.  Is this setup to allow for an easy way to define the directory which holds frequently updated files?  Ideally, one should be able to specify a USB stick mount point to avoid lots of I/O to the internal flash memory.

dobo90:
@graysky2 in the current state of this PR, working directory is set to a directory located in tmpfs (see /etc/init.d/adguardhome).

It seems that AdGuardHome doesn't keep any open descriptors besides working directory:
```root@OpenWrt:~# ls -l /proc/$(pidof AdGuardHome)/fd
lr-x------    1 root     root            64 Sep  1 18:08 0 -> /dev/null
l-wx------    1 root     root            64 Sep  1 18:08 1 -> pipe:[495]
lrwx------    1 root     root            64 Sep  1 18:08 10 -> /tmp/adguardhome/data/sessions.db
lrwx------    1 root     root            64 Sep  1 18:08 11 -> /tmp/adguardhome/data/stats.db
lrwx------    1 root     root            64 Sep  1 18:08 12 -> socket:[2922]
lrwx------    1 root     root            64 Sep  1 18:08 13 -> socket:[2920]
lrwx------    1 root     root            64 Sep  1 18:08 14 -> socket:[2923]
lr-x------    1 root     root            64 Sep  1 18:08 15 -> /tmp/adguardhome/data/filters/1.txt
lrwx------    1 root     root            64 Sep  1 18:08 19 -> socket:[143903]
l-wx------    1 root     root            64 Sep  1 18:08 2 -> pipe:[496]
lr-x------    1 root     root            64 Sep  1 18:08 3 -> anon_inode:inotify
lrwx------    1 root     root            64 Sep  1 18:08 4 -> anon_inode:[eventpoll]
lr-x------    1 root     root            64 Sep  1 18:08 5 -> pipe:[2866]
l-wx------    1 root     root            64 Sep  1 18:08 6 -> pipe:[2866]
lrwx------    1 root     root            64 Sep  1 18:08 7 -> anon_inode:[eventpoll]
lr-x------    1 root     root            64 Sep  1 18:08 8 -> pipe:[2912]
l-wx------    1 root     root            64 Sep  1 18:08 9 -> pipe:[2912]
```

The size of the temporary files doesn't seem to be very large:
```root@OpenWrt:~# du -h -d0 /tmp/adguardhome/
780.0K  /tmp/adguardhome/
```

I'm happily running this on a MIPS 1004Kc processor with 128 MB of RAM (I've got Redmi Router AC2100).

neheb:
Is this ready? Please rebase against master. We got new buildbots.

net/adguardhome/Makefile
jefferyto: I've added support for building host Go programs in #13570. Once that is merged, I will open another PR to add host packr, then this package can add `packr/host` to `PKG_BUILD_DEPENDS`. You can either patch out the `go:generate go install` line and call `go generate` or just call `packr -z` directly during `Build/Configure`.

net/adguardhome/files/adguardhome.init
jefferyto: My point isn't to include an empty config file just to have the file present, it's to have sensible/usable defaults (even if partial, e.g. without username/password) for the user already set. (Also this is not a requirement, just a nice-to-have.) If this isn't currently possible then it's okay to not have this at this time.

jefferyto:
@neheb As noted in [my comment](https://github.com/openwrt/packages/pull/12408#discussion_r500286587), I would prefer to have this package use a packaged packr (that I will add soon).

dobo90:
I'll update and rebase the commit when #13570 is merged into master. @jefferyto please poke me when it's done.

zdz2019:
@dobo90 golang is merged, need your help!

dobo90:
Gah, I've just noticed that @jefferyto has also made a packr recipe. I'll just wait one more time, now until #13593 is merged.

neheb:
It is.

jefferyto (approved):
LGTM - thanks!

Issue #12410, opened 2020-06-05
dawn not compiling on 64-bit


neheb:
Maintainer: @PolynomialDivision 
Environment: aarch64

https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a53/packages/dawn/compile.txt

%zu is the proper format.

PolynomialDivision:
Thanks I will update that. :)

dwmw2:
We can't use size_t as the on-wire format anyway.

PR #12412, opened 2020-06-05
dawn: fix x86 compiling

PolynomialDivision:
Fixes: #12410


BKPepe:
Compiled & run tested?

PolynomialDivision:
> Compiled & run tested?
Yes. ;)

neheb:
This is the only change: https://github.com/berlin-open-wireless-lab/DAWN/commit/2151db28ef24c856be1ffad836a7d20afdd9fba1
 
No real change.

BKPepe:
I think that this still does not explain, why if there is pull request template it gets removed each time and there are just fixes or changelog.

Issue #12307, opened 2020-05-25
elektra not compiling with gcc10


neheb:
```
[ 98%] Linking C executable ../bin/benchmark_large
/usr/bin/ld: CMakeFiles/benchmark_large.dir/large.c.o:(.bss+0x0): multiple definition of `large'; CMakeFiles/benchmark_large.dir/benchmarks.c.o:(.bss+0x0): first defined here
collect2: error: ld returned 1 exit status
```

This is the host build. Tested with:

```
make package/elektra/host/compile V=s
```

Maintainer: @haraldg
Environment: fedora 32, gcc10

haraldg:
Thanks, I will have a look.

haraldg:
@markus2330 Is this issue fixed in any released version?

(We probably should disable building benchmarks, but might be a nice excuse to update the package anyway.)

mpranj:
@haraldg we have fixed this problem, and the fix is released with [elektra 0.9.2](https://www.libelektra.org/news/0.9.2-release).

PR #12415, opened 2020-06-05
elektra: update to 0.9.2

haraldg:
Notable changes:
* Move some plugins from libelektra-plugins to libelektra-cpp because
  they got reimplemented in C++.
* New package libelektra-zmq
* Move the crypto plugin from libopenssl to libgcrypt
* Disable host build: elektra doesn't need it anymore and nobody uses
  kdb during package builds for now.

Closes: #12307

Signed-off-by: Harald Geyer <harald@ccbib.org>

Maintainer: me
Compile tested: openwrt-sdk-mxs_gcc-8.4.0_musl_eabi.Linux-x86_64 from 2020-05-28
Run tested: none


libs/elektra/Makefile
```diff
@@ -149,7 +146,7 @@ endef
 define Package/libelektra-crypto
   $(call Package/libelektra/Default)
   TITLE:=Elektra crypto plugin
-  DEPENDS:=+libelektra-core +libopenssl
+  DEPENDS:=+libelektra-core +libgcrypt
```
neheb, line 149: Curious why.

haraldg: It's an upstream decision. They don't support libopenssl anymore. Maybe @markus2330 can fill in the details?

It seems CI fails after 45 minutes, when it is still building boost. Just runs out of time?

I hope this is not a problem for merging the PR.

neheb: No it’s a RAM problem. I ran into the same issue locally with my VM. I’ll merge later today.

markus2330: We needed to put too much effort into maintenance, so we only kept the best plugin as evaluated by the author (@petermax2): https://www.libelektra.org/ftp/elektra/publications/nirschl2018cryptographic.pdf

PR #12419, opened 2020-06-06
ipmitool: fix CVE-2020-5208

lynxis:
Maintainer: me
Compile tested: ar71xx
Run tested: ar71xx

All patches has been applied upstream.

Signed-off-by: Alexander Couzens <lynxis@fe80.eu>


neheb:
I suggest adding PKG_CPE_ID to the Makefile.

PR #12421, opened 2020-06-06
pixman: update to 0.40.0

neheb:
Add some spacing and minor adjustments for consistency between packages

Switched to using meson for compilation. meson compiles faster.

Maintainer: @yousong 
Compile tested: ath79


yousong:
I splited the commit into three.  Run-tested within qemu machine with qemu-5.0 installed.

Thanks for the heads-up

PR #12430, opened 2020-06-07
sane-backends: update to 1.0.30

neheb:
Minor cleanups.

Maintainer: @luizluca 
Compile tested: ath79

### DIFF
--- a/utils/sane-backends/Makefile
+++ b/utils/sane-backends/Makefile
@@ -9,11 +9,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=sane-backends
-PKG_VERSION:=1.0.29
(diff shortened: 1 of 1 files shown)


luizluca:
I did my extended compile test without any issues (arc_generic,armvirt_64,ath79_generic,powerpc,ramips_mt7620,x86_64,x86_generic).

However, I'm currently away from a scanner for two weeks. So, there is no way to do a real run test. It just need some basic test like scanning a page. It would be enough to test the openwrt specific features like granting usb device permission to saned user and service control.

neheb:
Accidentally deleted my branch during cleanup.

luizluca:
I just did the runtime test. All good. Thanks

PR #12447, opened 2020-06-08
Jinja2: update to version 2.11.2

BKPepe:
Maintainer: @dangowrt 

Description:
- Update to version [2.11.2](https://jinja.palletsprojects.com/en/2.11.x/changelog/#version-2-11-2)
- Update project website

### DIFF
--- a/lang/python/Jinja2/Makefile
+++ b/lang/python/Jinja2/Makefile
@@ -5,11 +5,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=Jinja2
-PKG_VERSION:=2.10.3
-PKG_RELEASE:=2
+PKG_VERSION:=2.11.2
+PKG_RELEASE:=1
 
 PYPI_NAME:=$(PKG_NAME)
-PKG_HASH:=9fe95f19286cfefaa917656583d020be14e7859c6b0252588391e47db34527de
+PKG_HASH:=89aab215427ef59c34ad58735269eb58b1a5808103067f7bb9d5836c651b3bb0
 
 PKG_MAINTAINER:=Daniel Golle <daniel@makrotopia.org>
 PKG_LICENSE:=BSD-3-Clause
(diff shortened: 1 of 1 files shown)


lang/python/Jinja2/Makefile
```diff
@@ -25,7 +25,7 @@ define Package/python3-jinja2
   CATEGORY:=Languages
   SUBMENU:=Python
   TITLE:=Jinja2
-  URL:=http://jinja.pocoo.org/
+  URL:=https://jinja.palletsprojects.com/en/2.11.x/
```
karlp, line 28: I'd say https://jinja.palletsprojects.com/ is a better option here, even if it's currently redirecting to the versioned sub tree.  or even https://palletsprojects.com/p/jinja/ which links "jinja.palletsproject.com" as the url for docs

BKPepe: Thanks @karlp for review. I was not quite sure about it here.
Like you are saying there is an versioned subtree, but there could be released version 2.12 and on the router, there is installed 2.11 and it woudn't match. 

The website I was thinking for that and skipped that. Don't know the reason now, but maybe that the previous URL pointed to doc.

Anyway, I don't mind putting there https://palletsprojects.com/p/jinja/ .
Is that OK for you?

karlp: works for me, I just meant that in general we link to the _project_ page, not a particular version.

BKPepe: Changed. If you agree, please resolve the discussion and I will merge it.

dangowrt:
@BKPepe Thanks for taking care of that. I have stopped using those Python packages myself (Jinja2, Flask, Werkzeug) and would appreciate transferring maintainership to you or eg. @aparcar (or anyone else actually using them). Interested?

BKPepe:
I think for Turris routers I'm interested in Flask and Werkzeug as we are using it for our redesigned interface, which is the development. Jinja2 is still used for the previous one, so for some time I can send updates, but as being maintainer don't know for this one.

lang/python/Jinja2/Makefile
karlp: oh, sorry, I had only meant to comment, not "start a review" :)

PR #12458, opened 2020-06-08
cgroupfs-mount: enable use_hierarchy on the root cgroup

jmarcet:
Maintainer: Gerard Ryan G.M0N3Y.2503@gmail.com
Compile tested: master x86-64
Run tested: master x86-64

Description:
When using docker I could see on my dmesg:

```
[   99.156810] cgroup: runc (21352) created nested cgroup for controller "memory" which has incomplete hierarchy support. Nested cgroups may change behavior in the future.
[   99.157978] cgroup: "memory" requires setting use_hierarchy to 1 on the root
```

OpenWRT mounts the root sysfs cgroup in /etc/init.d/cgroupfs-mount, which has this comment:
```
# Procd mounts non-hierarchical cgroupfs so unmount first before cgroupfs-mount
```
So the intention, if I understand correctly, is already to have hierarchical cgroups. Anyhow, the cgroup memory controller needs it.

Writing `1` to `/sys/fs/cgroup/memory/memory.use_hierarchy` right after the mount, fixes the problem.

### DIFF
--- a/utils/cgroupfs-mount/files/cgroupfs-mount.init
+++ b/utils/cgroupfs-mount/files/cgroupfs-mount.init
@@ -9,4 +9,5 @@ boot() {
 	fi
 
 	cgroupfs-mount
+	echo 1 > /sys/fs/cgroup/memory/memory.use_hierarchy
 }


neheb:
ping @G-M0N3Y-2503

G-M0N3Y-2503:
[This doesn't seem like a OpenWrt specific issue](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=940713), Also [lxc sets this themselves](https://github.com/lxc/lxc/issues/236).
[I've submitted an upstream PR](https://github.com/tianon/cgroupfs-mount/pull/15) as I'm thinking this should be best suited there, or if rejected in `docker-ce`.

peterwillcn:
@neheb ping

neheb:
up to maintainer.

G-M0N3Y-2503:
I'm not sure how long it will take for the upstream to accept the change.
So if you wanted, I'd be willing to accept [a similar patch](https://github.com/tianon/cgroupfs-mount/pull/15/files) like lxc for the [dockerd.init](https://github.com/openwrt/packages/blob/master/utils/docker-ce/files/dockerd.init) in the interim.

jmarcet:
So you propose patching openwrt's docker-ce init instead of openwrt's copy of cgroups-mount?

Why don't we use the same patch you're pushing upstream while it is not accepted/merged?

peterwillcn:
@neheb This is the solution to my problem. Please merge！

peterwillcn:
@neheb It's just a small change that doesn't have to take so long, Merge or not merge this is a problem？ any idear?
I'd like to merge first！

neheb:
Up to the maintainer.

neheb:
Since https://github.com/openwrt/packages/pull/12618 was merged, this should be closed I take it?

jmarcet:
> Since #12618 was merged, this should be closed I take it?

You're right.

Issue #12443, opened 2020-06-07
zstd: meson-based build requires `python3-distutils` presence in host system


ryzhovau:
Can be reproduced on clean Debian 10 with [Buildroot requirements](https://oldwiki.archive.openwrt.org/doc/howto/buildroot.exigence#install_procedure_on_linux) installed.

```
$ make V=s package/zstd/compile
...
The Meson build system
Version: 0.54.2
Source dir: /home/ryzhovau/Entware/build_dir/target-mipsel_mips32r2_glibc-2.31/zstd-1.4.4/build/meson
Build dir: /home/ryzhovau/Entware/build_dir/target-mipsel_mips32r2_glibc-2.31/zstd-1.4.4/build/meson/openwrt-build
Build type: cross build
Using 'PKG_CONFIG_PATH' from environment with value: '/home/ryzhovau/Entware/staging_dir/target-mipsel_mips32r2_glibc-2.31/opt/lib/pkgconfig:/home/ryzhovau/Entware/staging_dir/target-mipsel_mips32r2_glibc-2.31/opt/share/pkgconfig'
WARNING: Unknown options: "bin_control"
The value of new options can be set with:
meson setup <builddir> --reconfigure -Dnew_option=new_value ...
Project name: zstd
Project version: DUMMY
C compiler for the build machine: gcc (gcc 8.3.0 "gcc (Debian 8.3.0-6) 8.3.0")
C linker for the build machine: gcc ld.bfd 2.31.1
C++ compiler for the build machine: g++ (gcc 8.3.0 "g++ (Debian 8.3.0-6) 8.3.0")
C++ linker for the build machine: g++ ld.bfd 2.31.1
Using 'CFLAGS' from environment with value: ''
Using 'LDFLAGS' from environment with value: ''
C compiler for the host machine: mipsel-openwrt-linux-gnu-gcc (gcc 8.4.0 "mipsel-openwrt-linux-gnu-gcc (OpenWrt GCC 8.4.0 r1223-5ceb6995) 8.4.0")
C linker for the host machine: mipsel-openwrt-linux-gnu-gcc ld.bfd 2.31.1
Using 'LDFLAGS' from environment with value: ''
C++ compiler for the host machine: mipsel-openwrt-linux-gnu-g++ (gcc 8.4.0 "mipsel-openwrt-linux-gnu-g++ (OpenWrt GCC 8.4.0 r1223-5ceb6995) 8.4.0")
[...]

PR #12469, opened 2020-06-09
zstd: fix compilation without host distutils

neheb:
This is the case in debian.

Maintainer: @dhewg 
Compile tested: debian 10

Fixes: https://github.com/openwrt/packages/issues/12443


dhewg:
Nice, looks good, thanks!

PR #12473, opened 2020-06-10
syslog-ng: remove network source config from the default setting

ja-pa:
maintainer: @BKPepe 
Compile tested: Turris Omnia (TOS5), OpenWrt master
Run tested: Turris Omnia (TOS5), OpenWrt master

Description:
This PR disables network source config in the default setting.

The reason for this change is, that users can have a different setting
(like disabled IPv6 etc) in which case this will result in syslog
crash. It makes more sense to set this config by user when it's really
needed and don't set it as default.


neheb:
ping @BKPepe

BKPepe:
Superseded by https://github.com/openwrt/packages/pull/12619

PR #12281, opened 2020-05-24
bind: add nslookup alternative to busybox nslookup

dl12345:
Add the bind9 nslookup utility as an alternative to busybox nslookup

Signed-off-by: Ian Cooper <iancooper@hotmail.com>

Maintainer: @nmeyerhans 
Compile tested:  x86_64 glibc, mips24k_c uClibc, mips24k_c musl
Run tested: x86_64 glibc master/trunk

Busybox nslookup is broken. Offer an alternative that isn't broken.

Where a host does not have an ipv6 address record, busybox's nslookup returns an error

```
root@openwrt:/# nslookup www.google.com
Server:		192.168.1.1
Address:	192.168.1.1#53

Name:      www.google.com
Address 1: 216.58.210.36
Address 2: 2a00:1450:4009:800::2004

root@openwrt:/# nslookup whatsmyipaddress.com
Server:		192.168.2.1
Address:	192.168.2.1#53

Name:      whatsmyipaddress.com
Address 1: 35.186.238.101
*** Can't find whatsmyipaddress.com: No answer
````

@neheb

PR #12489, opened 2020-06-12
kmod: fix ALTERNATIVES

kuanyili:
ALTERNATIVES shall be defined in the package where /sbin/kmod resides --
otherwise kmod tools will break if we install only the library as these
tools will be linked to a nonexistent executable.

Maintainer: @jdub
Compile tested: aarch64, Raspberry Pi 4 Model B Rev 1.1, OpenWrt SNAPSHOT r13548-8f3b176e86, pciutils installed
Run tested: as above

Description: An example is when we install `pciutils`, while it installs `libkmod` as dependency, it does not install `kmod`, thus there will be no `/sbin/kmod` to link to. And kmod tools break in this case before this fix is applied.

### DIFF
--- a/utils/kmod/Makefile
+++ b/utils/kmod/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=kmod
 PKG_VERSION:=20
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@KERNEL/linux/utils/kernel/kmod/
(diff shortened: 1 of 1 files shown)


BKPepe:
ping @zhanhb

zhanhb:
Good catch. I put ALTERNATIVES in the wrong place.

jdub:
btw, if any of you would like to change the maintainer field, feel free to do so 😄

dl12345:
See this comment on naming convention and location of the binary: https://github.com/openwrt/packages/pull/12281#issuecomment-634116902

zhanhb:
@dl12345 IMO, we don't need to move `/sbin/kmod` to `/usr/libexec`, for we might execute `kmod` directly such as `kmod list`. Or another name other than `kmod`, `lsmod`, `rmmod`, `insmod`, `modinfo`, `modprobe`, `depmod`?

kuanyili:
@dl12345 Debian `kmod` package put `kmod` executable in `/bin`, and OpenWrt `ubox` link kernel module tools to `kmodloader` in `/sbin`. Personally, I would leave `kmod` executable in `/sbin` as it used to be -- current layout looks good to me, no need to risk breaking something that depends on it.

PR #12492, opened 2020-06-12
adblock: update 4.0.5-6

amteza:
New options under YouTube SafeSearch: restrictive, moderate or disabled. This allows for better YouTube filter control, as for some even moderate filtering is too intrusive, e.g. with music. This will allow to disable completely YouTube filtering and keeping SafeSearch enabled for the other search engines.

Signed-off-by: Alberto Martínez Álvarez <amteza@gmail.com>

Maintainer: Dirk Brenken <dev@brenken.org> @dibdot    @\<github-user> (find it by checking history of the package Makefile)
Compile tested: (aarch64, RPi4, OpenWrt master)
Run tested: (aarch64, RPi4, OpenWrt master, tested all options on interface)

### DIFF
--- a/multimedia/ffmpeg/Makefile
+++ b/multimedia/ffmpeg/Makefile
@@ -9,12 +9,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ffmpeg
-PKG_VERSION:=4.2.2
-PKG_RELEASE:=4
+PKG_VERSION:=4.2.3
+PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://ffmpeg.org/releases/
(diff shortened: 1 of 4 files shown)


hnyman:
As you are pushing from master, the PR contains unrelated ffmpeg files.

dibdot:
I don't like this "youtube-only" way. Why should we limit a disabled safesearch provider functionality to youtube? I'll provide a more generic approach with the next update.

dibdot:
generic approach implemented in #12494 and https://github.com/openwrt/luci/pull/4166

amteza:
> As you are pushing from master, the PR contains unrelated ffmpeg files.

Apologies, messed up.

amteza:
> generic approach implemented in #12494 and [openwrt/luci#4166](https://github.com/openwrt/luci/pull/4166)

Thanks, I like your generic approach better. Only YouTube has an explanation, usually you want to filter searching in all the engines; however, YouTube filters are very restrictive in their nature, so even under moderate you will find that —for example— common music videos get blocked, which is more an annoyance than anything else. Same thing happens to other contents.

dibdot:
The generic variant has been merged - closing this.

Thanks!

PR #12500, opened 2020-06-13
[19.07] libtirpc: update to 1.2.6

Andy2244:
Maintainer: me
Compile tested: arm (19.07)
Run tested: arm/qemu (19.07)

Description:
* update to 1.2.6
* Add HOST_BUILD_PARALLEL for faster compilation

### DIFF
--- a/libs/libtirpc/Makefile
+++ b/libs/libtirpc/Makefile
@@ -1,19 +1,20 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libtirpc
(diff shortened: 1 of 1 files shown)


neheb:
What patch was removed?

Andy2244:
copy&paste error from your master notes, seems 19.07 never had "010-xdr_float-do-not-include-bits-endian.h.patch"

neheb:
Ah. Typically when doing these backports, the preference here is to use git cherry-pick -x

PR #12509, opened 2020-06-13

lipnitsk:
The perl Configure file was matching GCC 10 against "1*" and treating it
as GCC 1, causing ABI breakage and segfaults.

Cherry-pick the upstream patch which fixes it to check against (e.g)
"1.*" instead, which will make it work for hundreds more GCC versions
to come.

https://github.com/Perl/perl5/commit/6bd6308fcea3541
 "Adapt Configure to GCC version 10"

Also includes the previous commit just adding GCC 8 and 9 to one case:
https://github.com/Perl/perl5/commit/ae195500577d707
 "Add gcc-8 and gcc-9 for FORTIFY_SOURCE"

(cherry picked from commit 65578a43f0d12c02888df00b6fdc90c73a02875c)


neheb:
Works for me.

PR #12516, opened 2020-06-14
FreeBSD build fix

rozhuk-im:
Maintainer: Jo-Philipp Wich <jo@mein.io>
Compile tested: FreeBSD 12.1 amd64, with clang 10, D-Link DIR-510L, current master
Run tested: no

Description:
Allow to build on non Linux OS.


neheb:
Please bump the PKG_RELEASE.

diizzyy:
...and this change doesn't break when using busybox?

rozhuk-im:
@neheb Are you sure that PKG_RELEASE must be updated?
There is no changes for users that already have it.

@diizzyy If some one does not have /bin/sh - then openwrt will not build.
Why you asking about busybox? (I do not see how it connected)

diizzyy:
Because this can potentially break on platforms that are known working, also changes like this should go upstream not in package repo as dropping requirement of bash is beneficial on all platforms.

rozhuk-im:
If you look into package, then you see that in other make files almost same content except ```SHELL = /bin/sh``` - already set.

neheb:
Yes I am sure about PKG_RELEASR. You're changing a Makefile.

neheb:
```
=== Checking commit 'd65218327bec821b06f9c74516eaada22f5b5ead'
 
Author name (Rozhuk Ivan) seems ok
 
Commit subject line MUST start with '<package name>: ' (FreeBSD build fix)
 
Signed-off-by match author
 
```

Issue #12403, opened 2020-06-04
PHP: segmentation fault when exiting


barhom:
When running any PHP script the script finishes successfully however exits with a "Segmentation fault".

Running master from 2020-06-04.
Packages installed:
```
opkg list-installed | grep php
php7 - 7.4.6-1
php7-cli - 7.4.6-1
php7-mod-filter - 7.4.6-1
php7-mod-json - 7.4.6-1
php7-mod-mbstring - 7.4.6-1
php7-mod-openssl - 7.4.6-1
```
The php script:
```
# cat test.php 
<?php
echo "HELLO WORLD\n";
```

The output:
```
# php-cli test.php 
HELLO WORLD
Segmentation fault
```

barhom:
@mhei do you know what is causing this?

Thanks

mhei:
I traced it down with gdb to be a problem in oniguruma:

```
Program received signal SIGSEGV, Segmentation fault.
0xb6e37f70 in i_free_callout_name_entry (key=0x369040, e=0x0, arg=0x37afe0) at oniguruma-6.9.5_rev1/src/regparse.c:1297
```

I'll create a PR with a fix, see https://github.com/kkos/oniguruma/pull/196.

PR #12521, opened 2020-06-14
oniguruma: add patch to fix NULL pointer exception

mhei:
Maintainer: @cotequeiroz
Compile tested: mxs
Run tested: mxs

Description:
I propose to carry this patch until a new upstream
release includes it. For forther references see:

https://github.com/openwrt/packages/issues/12403
and
https://github.com/kkos/oniguruma/pull/196


cotequeiroz:
This was merged upstream, but I have voiced my concern about this hiding a problem somewhere else (perhaps a call with bad parameters?)
It does provide a quick way out of the original problem, so I'll merge this, while keeping an eye upstream.

Issue #12490, opened 2020-06-12
[utils/shadow] bcrypt should not be compiled in


n8v8R:
Maintainer: unknown (none stated)

Environment: tree wide

Description: `crypt_blowfish` is being removed in the toolchain https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob;f=toolchain/musl/patches/901-crypt_size_hack.patch and therefore it does not make sense to compile `bcrypt` into the package https://github.com/openwrt/packages/blob/master/utils/shadow/Makefile#L44

neheb:
In that package, I adjust the default based on strength and what is available: https://github.com/openwrt/packages/blob/master/utils/shadow/Makefile#L131

You want https://github.com/openwrt/packages/blob/master/utils/shadow/Makefile#L44 to be changed based on availability?

n8v8R:
Checking on algo availability is more flexible and probably beneficial for users compiling their own kernels and retaining the algo.

neheb:
I can't find any reference to crypt_blowfish in shadow-utils.

n8v8R:
shadow does not provide the algo but the kernel

neheb:
Simple test:

```
passwd: failed to crypt password with salt '$2a$13$w8EJ0Yfz5bGsG4U/0m7bk/': Function not implemented
The password for root is unchanged.
```

Sounds like /etc/login.defs should be changed as needed.

n8v8R:
> edit: and no. shadow-utils does not use the kernel.

Yes, it does, https://forum.openwrt.org/t/passwd-a-sha256-sha512-less-secure-than-default/13117/3 and that is with `crypt_blowfis` being stripped from the kernel it does not work. If it would leverage any other crypto library, e.g. OpenSSL or WolfSSL salting with SHA|Blowfish would work.
___
> Sounds like /etc/login.defs should be changed as needed.

That is not an issue, except being left with DES or MD5 which are neither suitable for salting password hashes. 

https://openwrt-devel.openwrt.narkive.com/AtxRVHLp/sha512-sha256-hashes-truncated-busybox-weirdness

> All password hash algorithms except for MD5 (default) and DES (for compatibility reasons) have been stripped from libc to cut down on useless bloat.
>
>If you care about password storage security, it's better to use a function that was intentionally designed NOT to be fast, e.g. PBKDF2 or bcrypt.
___
Therefore compiling `bcrypt` into shadow is entirely pointless.

neheb:
[...]

PR #12530, opened 2020-06-16
shadow: don't compile bcrypt for unsupported platforms

neheb:
If musl has no bcrypt:

passwd: failed to crypt password with salt '$2a$13$w8EJ0Yfz5bGsG4U/0m7bk/': Function not implemented
The password for root is unchanged.

glibc output as it has no bcrypt:

passwd: failed to crypt password with salt '$2a$13$xbpmAYmq6Q/rZN5jOlNxJZ': Invalid argument
The password for root is unchanged.

--without-bcrypt output:

Invalid ENCRYPT_METHOD value: 'BCRYPT'.
Defaulting to DES.
passwd: password changed.

The solution was tested on glibc despite using a musl specific variable
Still works.

Maintainer: nobody
Compile tested: malta musl/glibc mips-be

Fixes: https://github.com/openwrt/packages/issues/12490

The proper fix to these hacks is to introduce https://github.com/besser82/libxcrypt

ping @Cynerd @BKPepe

### DIFF
--- a/utils/shadow/Makefile
+++ b/utils/shadow/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=shadow
 PKG_VERSION:=4.8.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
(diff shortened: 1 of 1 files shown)


Cynerd:
For our case with `CONFIG_MUSL_DISABLE_CRYPT_SIZE_HACK` enabled it stays the same so I have not problem with that. The behavior with enabled size hack is not something I can assess.

PR #12533, opened 2020-06-17
mwan3: improvements

feckert:
Maintainer: me 
Compile tested: not needed only script change
Run tested: x86_64 and lantiq_xrx200, APU3, OpenWrt master

Description:

- Use procd to handle service reload on uci config change
- Set ubus output for **status** info to unknown if this information is not available

### DIFF
--- a/net/mwan3/files/usr/libexec/rpcd/mwan3
+++ b/net/mwan3/files/usr/libexec/rpcd/mwan3
@@ -77,7 +77,7 @@ get_mwan3_status() {
 	local online=0
 	local offline=0
 	local up="0"
-	local enabled pid device time_p time_n time_u time_d
+	local enabled pid device time_p time_n time_u time_d status
 
 	network_get_device device $1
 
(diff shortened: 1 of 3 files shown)


net/mwan3/files/usr/libexec/rpcd/mwan3
```diff
@@ -111,6 +111,12 @@ get_mwan3_status() {
 		network_get_uptime uptime "$iface"
 		network_is_up "$iface" && up="1"
 
+		if [ -f "$MWAN3TRACK_STATUS_DIR/${iface}/STATUS" ]; then
+			status="$(cat "$MWAN3TRACK_STATUS_DIR/${iface}/STATUS")"
+		else
+			status="unknown"
```
aport, line 117: Can this instead read the interface status in `/var/run/mwan3/iface_status/${iface}` ?

The problem is that when tracking is disabled on an interface, `mwan3 status` will still report correctly but the ubus call will not.

feckert: This has nothing to do with my change.
It's just an improvement.
If the status is unknown, because the file doesn't exist, then he should return `unknowon` and not just nothing.

I think the value you need is already present https://github.com/openwrt/packages/blob/master/net/mwan3/files/usr/libexec/rpcd/mwan3#L125

PR #12536, opened 2020-06-17
oniguruma: build using autotools

### LINKED ISSUE #12403: PHP: segmentation fault when exiting

cotequeiroz:
Maintainer: me
Compile tested: mvebu, WRT3200ACM, openwrt master
Run tested: as above, running php-cli with a "hello world" script

Description:
Cmake currently fails to compile properly.  While there's a fix upstream, we can avoid carrying any patches here, and use the more stable build system, despite being slower.

---

I thought about using https://github.com/kkos/oniguruma/commit/c07fa0aeae09ea7a1564539de92dd9b5ae078592 to keep using cmake, but I feel we're jumping through hoops to do it.
I know autotools uses more resources to build, so I'm not committing this right away, to see if anyone disagrees with this.  Since it is currently broken, we should not wait too long.
Fixes #12403 
See also kkos/oniguruma#196 and kkos/oniguruma#197
Ping @neheb @mhei


neheb:
My preference is to keep using CMake and patch it as necessary.

I don't know if it's fair to call autotools a more stable build system.

As a sidenote, I also had this PR in base for even faster compilation: https://github.com/openwrt/openwrt/pull/2874

cotequeiroz:
When I dared to compare the build systems, I only meant for the onig package itself.  I asked upstream if cmake support is on par with autotools, but I **think** I already know the answer.  I just don't want to wait too long to have this bug fixed.  We can always make the switch after the next release.

cotequeiroz:
I'm quoting upstream from kkos/oniguruma#196:

> For a long time, the only thing I had to check about cmake was that it didn't give any errors on compilation.

> Certainly no more tested in cmake than in autotools.
> cmake is an alternative for those who cannot or do not want to use autotools.

I'm going to merge this now.  We can reevaluate it for next release.

PR #12548, opened 2020-06-18
powertop: fix compilation with glibc

neheb:
It seems several libraries need to be linked.

Maintainer: @lucize 
Compile tested: i386 glibc

The stack protector thing is not needed as of https://github.com/openwrt/openwrt/commit/b933f9cf0cb254e368027cad6d5799e45b237df5


lucize:
LGTM

PR #12549, opened 2020-06-18
libextractor: remove ffmpeg plugin

neheb:
FFmpeg has patented functionality.

Maintainer: @dangowrt 
Compile tested: ath79

### DIFF
--- a/libs/libextractor/Makefile
+++ b/libs/libextractor/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libextractor
 PKG_VERSION:=1.10
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 # ToDo:
 # - package missing optional dependencies: libexiv2, gsf, librpm, smf, tidy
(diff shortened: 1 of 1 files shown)


dangowrt:
FFmpeg *may* contain patented functionality if compiled with it. It doesn't need to. Nor do the header files used by libextractor use any patented functionality.
I don't see why we should remove the ffmpeg plugin entirely (if one doesn't want to install ffmpeg he can just install libextractor without that plugin).

neheb:
I asked on the ffmpeg IRC channel before. ffmpeg developers have a different opinion:

```
mangix@armbiaan ~/i/Freenode> cat "#ffmpeg.log" | grep -i patented
15:20 < mangix> oh fun question: does ffmpeg offer a way to disable patented AAC functionality?
15:33 <@cehoyos> mangix: Just to make sure, even with --disable-all we assume that (a lot of) "patented" algorithms are still compiled
```

I can make this change dependent on BUILD_PATENTED if desired.

diizzyy:
I don't see any need to cripple software just for the sake if it.

neheb:
Note that BUILD_PATENTED is default. Again, it's fairly easy to make this conditional.

karlp:
@neheb Please link to where you have a project agreement or even a discussion on these sweeping changes to remove patented sw.

neheb:
Reworked to use CONFIG_BUILD_PATENTED.

neheb:
Please don't merge until there's a decision here: http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030629.html

neheb:
The minimal discussion indicates that this should be merged.

Issue #12358, opened 2020-05-29
docker-ce: firewall rules not flushed on "/etc/init.d/dockerd stop"


dl12345:
The title of the issue says it all. Probably makes sense to do a `fw3 restart` when the service is stopped. 

@G-M0N3Y-2503

feckert:
This raises the question of who is now responsible for this to clean up the mess

- dockerd binary ?
or
- dockerd init script ?

I would expect the this is the task for dockerd binary. But the way I see it, that's not the case. Then we have no choice but to do this in the [init script](https://github.com/openwrt/packages/blob/master/utils/docker-ce/files/dockerd.init) 

I think the whole fw3  firewall restart is dangerous.

feckert:
I have updated the dockerd init script in this pullrequest [commit](https://github.com/openwrt/packages/pull/12556/commits/2c33d39a5e0f69d6de295ea7aa279b8f3b2a2545) to cleanup all docker related rules that have been created on my system.

@dl12345 does this solve your issue?

dl12345:
Yeah, that should work. Definitely better than doing a `fw3 restart`

G-M0N3Y-2503:
FYI it looks like [libnetwork](https://github.com/moby/libnetwork/blob/master/drivers/bridge/setup_ip_tables.go) is the one responsible for modifying the firewall, Though I'm not familiar with `docker-proxy`s interactions with `dockerd` so I'm not sure where a binary fix would go.

feckert:
I think we should start by getting the firewall rules out on service stop first (https://github.com/openwrt/packages/pull/12556)

We could set the iptables key to false in the Docker engine’s configuration file at /etc/docker/daemon.json.
And do everything ourselves with the fw3?

G-M0N3Y-2503:
> 1. We could set the iptables key to false in the Docker engine’s configuration file at /etc/docker/daemon.json.
>    And do everything ourselves with the fw3?

To me, this seems like a lot of ongoing maintenance, and I'm not sure what benefits we would gain by doing this...
I imagine most people will either have `dockerd` installed and running, in which case the inserted rules are being used, or just not have it installed. I don't think we'd have many users with `dockerd` installed run once and then not have it running for the majority of the time, and even if we do I think https://github.com/openwrt/packages/pull/12556 is enough.

> 2\. Another solution is to patch/update the docker-ce go program to delete the firewall rules on binary stop.
>     But I don't think I'm gonna be able to do it that fast.

[...]

PR #12556, opened 2020-06-19
docker-ce: add uci support

feckert:
Maintainer: @G-M0N3Y-2503 

Description:
Recently we also have the application [luci-app-dockerman](https://github.com/openwrt/luci/tree/master/applications/luci-app-dockerman) in the LuCI. In this app we have also an service which reads the configuration from[ /etc/docker/daemon.json](https://github.com/openwrt/packages/blob/master/utils/docker-ce/files/daemon.json) and compares this against his own uci configuration in [/etc/config/dockerman](https://github.com/openwrt/luci/blob/master/applications/luci-app-dockerman/root/etc/config/dockerman). To reconcile configuration differences a [script](https://github.com/openwrt/luci/blob/master/applications/luci-app-dockerman/root/usr/share/dockerman/dockerd-config.lua) is used, so that everything works again.
To simplify things I think it is better to move the uci into the docker-ce package. So that we have a clear responsibility.

@lisaac To get everything working again we have to make some changes to the luci-app-dockerman as well.

Since docker's configuration options are very large, I don't think we can display all configuration options here. But I guess it doesn't hurt that we can configure at least a few things.

### DIFF
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -58,3 +58,36 @@ start_service() {
 	procd_set_param limits nofile="${nofile} ${nofile}"
 	procd_close_instance
 }
+
+ip4tables_remove_nat() {
+	iptables -t nat -D OUTPUT ! -d 127.0.0.0/8 -m addrtype --dst-type LOCAL -j DOCKER
+	iptables -t nat -D PREROUTING -m addrtype --dst-type LOCAL -j DOCKER
+
+	iptables -t nat -F DOCKER
+	iptables -t nat -X DOCKER
+}
+
+ip4tables_remove_filter() {
+	iptables -t filter -D FORWARD -j DOCKER-USER
+	iptables -t filter -D FORWARD -j DOCKER-ISOLATION-STAGE-1
+	iptables -t filter -D FORWARD -o docker0 -j DOCKER
+
+	iptables -t filter -F DOCKER
+	iptables -t filter -F DOCKER-ISOLATION-STAGE-1
+	iptables -t filter -F DOCKER-ISOLATION-STAGE-2
+	iptables -t filter -F DOCKER-USER
+
+	iptables -t filter -X DOCKER
+	iptables -t filter -X DOCKER-ISOLATION-STAGE-1
+	iptables -t filter -X DOCKER-ISOLATION-STAGE-2
+	iptables -t filter -X DOCKER-USER
+}
+
+ip4tables_remove() {
+	ip4tables_remove_nat
+	ip4tables_remove_filter
+}
+
+stop_service() {
+	ip4tables_remove
+}
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -3,12 +3,58 @@
 USE_PROCD=1
 START=25
 
+DOCKERD_CONF="/tmp/dockerd/daemon.json"
+
+json_add_array_string() {
+	json_add_string "" "$1"
+}
+
+process_config() {
+	local alt_config_file data_root log_level
+
+	rm -f "$DOCKERD_CONF"
+
+	[ -f /etc/config/dockerd ] || {
+		# Use the daemon default configuration
+		DOCKERD_CONF=""
+		return 0
+	}
+
+	config_load 'dockerd'
+
+	config_get alt_config_file globals alt_config_file
+	[ -n "$alt_config_file" ] && [ -f "$alt_config_file" ] && {
+		ln -s "$alt_config_file" "$DOCKERD_CONF"
+		return 0
+	}
+
+	config_get data_root globals data_root "/opt/docker/"
+	config_get log_level globals log_level "warn"
+
+	. /usr/share/libubox/jshn.sh
+	json_init
+	json_add_string "data-root" "$data_root"
+	json_add_string "log-level" "$log_level"
+	json_add_array "registry-mirrors"
+	config_list_foreach globals registry_mirror json_add_array_string
+	json_close_array
+
+	mkdir -p /tmp/dockerd
+	json_dump > "$DOCKERD_CONF"
+}
+
 start_service() {
 	local nofile=$(cat /proc/sys/fs/nr_open)
 
+	process_config
+
 	procd_open_instance
 	procd_set_param stderr 1
-	procd_set_param command /usr/bin/dockerd
+	if [ -z "$DOCKERD_CONF" ]; then
+		procd_set_param command /usr/bin/dockerd
+	else
+		procd_set_param command /usr/bin/dockerd --config-file="$DOCKERD_CONF"
+	fi
 	procd_set_param limits nofile="${nofile} ${nofile}"
 	procd_close_instance
 }
--- a/utils/docker-ce/files/etc/config/dockerd
+++ b/utils/docker-ce/files/etc/config/dockerd
@@ -1,0 +1,7 @@
+
+config globals 'globals'
+#	option alt_config_file "/etc/docker/daemon.json"
+	option data_root "/opt/docker/"
+	option log_level "warn"
+#	list registry_mirror "https://<my-docker-mirror-host>"
+#	list registry_mirror "https://hub.docker.com"
--- a/utils/docker-ce/Makefile
+++ b/utils/docker-ce/Makefile
@@ -124,9 +124,6 @@ define Package/docker-ce/install
 	$(INSTALL_DIR) $(1)/etc/init.d
 	$(INSTALL_BIN) ./files/dockerd.init $(1)/etc/init.d/dockerd
 
-	$(INSTALL_DIR) $(1)/etc/docker
-	$(INSTALL_CONF) ./files/daemon.json $(1)/etc/docker/
-
 	$(INSTALL_DIR) $(1)/etc/config
 	$(INSTALL_CONF) ./files/etc/config/dockerd $(1)/etc/config/dockerd
 
--- a/utils/docker-ce/Makefile
+++ b/utils/docker-ce/Makefile
@@ -49,6 +49,10 @@ define Package/docker-ce
   MENU:=1
 endef
 
+define Package/docker-ce/conffiles
+/etc/config/dockerd
+endef
+
 define Package/docker-ce/description
   Docker Engine is used by millions enables containerized applications
   to run anywhere consistently on any infrastructure.
@@ -122,6 +126,9 @@ define Package/docker-ce/install
 
 	$(INSTALL_DIR) $(1)/etc/docker
 	$(INSTALL_CONF) ./files/daemon.json $(1)/etc/docker/
+
+	$(INSTALL_DIR) $(1)/etc/config
+	$(INSTALL_CONF) ./files/etc/config/dockerd $(1)/etc/config/dockerd
 
 	# Must be after systcl 11-br-netfilter.conf from kmod-br-netfilter
 	$(INSTALL_DIR) $(1)/etc/sysctl.d
--- a/utils/docker-ce/Makefile
+++ b/utils/docker-ce/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=docker-ce
 PKG_VERSION:=19.03.12
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=components/cli/LICENSE components/engine/LICENSE
 
--- a/utils/docker-ce/files/daemon.json
+++ b/utils/docker-ce/files/daemon.json
@@ -1,4 +1,0 @@
-{
-    "data-root": "/opt/docker/",
-    "log-level": "warn"
-}


utils/docker-ce/files/dockerd.init
```diff
@@ -3,12 +3,41 @@
 USE_PROCD=1
 START=25
 
+DOCKERD_CONF="/tmp/dockerd/daemon.json"
+
+handle_registry_mirror() {
```
G-M0N3Y-2503, line 8: This looks like it would be useful in a general sense when more settings are added as well. 
we could probably rename it to something like `json_add_array_string`

G-M0N3Y-2503:
Otherwise, thanks for the work, sounds like a great idea going forward!

lisaac:
Sounds great!

Now, we can get rid of `dockerd-config.lua` for the luci-app-dockerman.

@feckert And how about the access control?

feckert:
@lisaac I have to see how you do it first.
But I could imagine that we add a [firewall object](https://github.com/openwrt/packages/blob/6df49a634af7eeb37a1f1c6166ba8aa92949e47c/net/net-snmp/files/snmpd.init#L330) to the docker-ce service and then send a reload to the firewall, which should take care of setting the right rules.
The advantage would be that we have no dependency on the firewall tools.
But we can work with the openwrt tool set.

feckert:
I have now added a proof of concept firewall ac handling in the dockerd service with this [commit](https://github.com/openwrt/packages/pull/12556/commits/f5c947a1f9a38acd13474ce14ec6fca3bba71e3c). The advantage we do not need any iptables commands. The firewall handling is abstracted by procd.

If we set the following option in the dockerd uci config 
```
config globals 'globals'
        list ac_allowed_interface 'lan'
```
then the following rule is added to the firewall by fw3.
```
iptables-save | grep docker0
-A zone_lan_input -i docker0 -p tcp -m comment --comment "!fw3: ubus:dockerd[instance1] rule 0" -j ACCEPT
```

Unfortunately, it is not yet clear to me what you @lisaac want to achieve with ac.
* AC to the containers
or
* AC to the config interface over the network.

The way I see it, this has nothing to do with the configuration interface. Did I get that right?

If so is that firewall handling really necessary in docker?
I think we could add the following configuration section to **/etc/config/network** and **/etc/config/firewall** and then we achieve the same. Or did I get something wrong.

**/etc/config/network**

```
config interface 'docker0'
        option proto 'none'
        option ifname 'docker0'
        option force_link '0
```

**/etc/config/firewall**
```
config zone
        option name 'lan'
        option input 'ACCEPT'
        option output 'ACCEPT'
        option forward 'ACCEPT'
        option network 'lan docker0'
```

lisaac:
Without the access control, WAN can easily access the ports exposed by the container by `WANIP:PORT`, which is big security problems.

On another hand, sometimes, you would like some ports exposed by containers could be access from WAN.

For more details, please see https://github.com/lisaac/blog/issues/7 , https://docs.docker.com/network/iptables/

In order to let those works, I add the access control in luci-app-dockerman.

> -A zone_lan_input -i docker0 -p tcp -m comment --comment "!fw3: ubus:dockerd[instance1] rule 0" -j ACCEPT 
@feckert I don't think it will work, and maybe we could let dockerman achieve with ac, let docker-ce do his own job

G-M0N3Y-2503:
The [docker documentation](https://docs.docker.com/network/iptables/) @lisaac linked leads me to believe that this kind of setup is outside the scope of `docker-ce`, at least in general. However, I think it is also inside the scope of this package to get `docker-ce` running on OpenWrt.

It could be argued either way really, but since [one of the reasons to use OpenWrt is "Security"](https://openwrt.org/#why_use_openwrt) and being able to access from the WAN by default seems to go against that. I lean slightly in favour of configuring it from `docker-ce`. Though the only case I can think of would be where someone doesn't have luci-app-dockerman for whatever reason...
But, I'm open to suggestions.

Also, if we decide to do AC here, it looks like this PR would close https://github.com/openwrt/packages/issues/12358

feckert:
@G-M0N3Y-2503 thanks for the docker iptables link. That was so not clear to me that the IP packages that are for docker are not handled by the fw3 of openwrt firewall. Then we can forget about the procd (firewall_array)?

@lisaac If I see this correctly, would it be better to add the rules for the docker to DOCKER-USER and not do the new  DOCKER-MAN chain?

lisaac:
@feckert 
`iptables -I DOCKER-USER -j DOCKER-MAN`

Use a custom chain to facilitate maintenance

feckert:
I have update the pullrequest so that on docker stop the related pullrquests for dockerd are removed.
The next step would be to update luci-app-dockerman to add the rule `iptables -I DOCKER-USER -j DOCKER-MAN` to iptables on luci-app-dockerman service start. @lisaac Is there anything else we need to consider?

jmarcet:
Hi everybody,

My two cents with respect to this PR, with apologies for the length, but there
is a lot to be said.

I have been using docker extensively in openwrt for more than 6 months. I am
the maintainer of the docker-compose package. With that said, I would like to
do some observations:

 - I love luci-app-dockerman, for me it was a surprise when I saw it merged,
 and a pretty good one. The only thing I needed to change in my systems was
 launching dockerman instead of dockerd.
 
 - I am all in favor of a proper integration of docker in openwrt.
 
 - In principle I like the idea of controlling dockerd parameters (the ones
 which end up in /etc/docker/daemon.json) from uci, *but* there is a
 plethora of potentially interesting options. How do you envision supporting
 them all?
 
 - dockerd does its own iptables handling. The fact that you can end up
 exposing ports on wan easily is the standard dockerd behavior. Instead of
 writing a new set of iptables rules outside of docker, there are other two
 sensible ways to avoid it, all while keeping the standard docker behavior.
 You can either map ports to an internal IP (be it localhost or lan) container
 by container, or you can change the default docker bridge binding address,
 which defaults to 0.0.0.0, with a dockerd parameter. Hence I favor not doing
 any new iptables outside of docker by default.
 
 ```
 $ dockerd --help | grep bind
 --ip ip  Default IP when binding container ports (default 0.0.0.0)
 ```

 Or in a docker-compose.yml, instead of doing:

```
    ports:
      - '8087:80'
```

 We would do something like:

```
    ports:
      - '127.0.0.1:8087:80'
```

jmarcet:
- There is quite a lot of room for improvement in other areas to have a cleaner
 integration in openwrt. Namely the integration of docker networks in openwrt's
 interfaces. This could come in as a new type of network interface for externally
 managed networks, as well as offering an easy way to use specially crafted docker
 networks like macvlan bridges.
 
 Let me give you a few examples. I currently have 20+ containers running,
 connected to different network bridges. By default, dockerd assigns (partially)
 random mac and ip addresses to those bridges, and the interfaces are not
 considered to be up  by openwrt.
 
 All those random addresses are a PITA from a security and administration point
 of view. Instead of having a controlled list of mac addresses in the network,
 you end up with an ever growing list. Using something like nlbwmon turns really
 ugly.
 
 I first tried adding the bridges to `/etc/config/network`, and it sort of
 worked, I could even set static mac addresses there and have the bridges always
 use the same macs. However dockerd often failed to start some containers at
 boot saying the network already existed.
 
 I ended up adding the networks this way:
 
 ```
/etc/docker/daemon.json
{
   "debug": false,
   "data-root": "/opt/docker",
   "log-level": "warn",
   "bip": "10.XX.YY.1/24"
}
 ```
 
 ```
 /etc/config/network
 config interface 'docker'
        option ifname 'docker0'
        option proto 'static'
        option ipaddr '10.XX.YY.1'
        option auto '0'

 config interface 'nextcloud'
        option ifname 'br-nextcloud'
        option proto 'static'
        option ipaddr '10.AA.BB.1'
        option auto '0'
```

```
/etc/config/firewall
config zone
        option network 'docker'
        option input 'REJECT'
        option output 'ACCEPT'                                                                                                         
        option forward 'REJECT'
        option name 'docker'

config zone
        option network 'nextcloud'
        option input 'REJECT'
        option output 'ACCEPT'                                                                                                         
        option forward 'REJECT'
        option name 'nextcloud'
```

```
/etc/hotplug.d/net/11-docker-bridges
#!/bin/sh

IFACES="docker0 br-nextcloud"

[ -n "$INTERFACE" -a -n "$( echo $IFACES | grep $INTERFACE )" ] || exit 0

if [ "$ACTION" = add ]; then
        MAC="$( grep $INTERFACE /etc/ethers | cut -d' ' -f1 )"
        [ "$INTERFACE" = "docker0" ] && IFACE=docker || IFACE="${INTERFACE/br-}"
        logger -t docker-bridges "Setting $IFACE network up"
        ifup $IFACE
        logger -t docker-bridges "Setting $INTERFACE mac address to $MAC"
        ifconfig $INTERFACE hw ether $MAC
fi
```

```
nextcloud's docker-compose.yml

networks:
  default:
    driver: bridge
    ipam:
      driver: default
      config:
        - subnet: 10.AA.BB.0/24
    driver_opts:
      com.docker.network.bridge.name: br-nextcloud
```

```
/etc/ethers

ZZ:ZZ:ZZ:ZZ:00:00 docker0
ZZ:ZZ:ZZ:ZZ:60:00 br-nextcloud
```

With this I have a nice view of all the interfaces (real and virtual) on my
network from luci. I also get proper metrics from all the bridges with
node_exporter.

docker0, lan and the docker bridges are in different zones. Access between
them is not allowed by default. If I want to allow something I can easily
add a new firewall rule (all from luci if you like). So this adds to docker
infrastructure instead of duplicating it. 

It could be much cleaner if we could set the bridges as externally managed
in `/etc/config/network`. That would erase the need to add the hotplug
script and the macs to `/etc/ethers`.

With respect to the access control, we could probably change the default
binding address from 0.0.0.0 to the router lan's ip. Anyhow this all should
be well documented in the wiki, since the docker way (from reading blogs
everywhere about it) would be specifying an IP when doing any port mapping.

**Comment by jmarcet** [comment]
Last, we could add support for defining macvlan bridges directly in
`/etc/config/network`. At the moment you have to manually add the bridge if you
want to use it from docker. So, for example, say you want to have a container
directly on your lan, with its own mac and an ip address within your lan's
subnet. You can define it in docker-compose like this (where 192.168.1.1/24
is your lan subnet and eth0 your lan interface):


```
docker-compose.yml
version: '2.4'                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                               
services:                                                                                                                                                                                                                                                                      
  someservice:
    [...]
    networks:
      somelan:
        ipv4_address: 192.168.1.65

networks:
  somelan:
    driver: macvlan
    driver_opts:
      parent: eth0
      macvlan_mode: bridge
    ipam:
      config:
        - subnet: 192.168.1.64/29
```


But you need to set the macvlan bridge yourself:

```
ip link add veth link eth0 type macvlan mode bridge
```

This can be added as a hotplug script to make it automatic.

Again, this could be much nicer if we had a macvlan bridge option in openwrt's
interfaces.

For reference, you can see the complete set of docker-compose files I am using
in my openwrt managed home lan [here](https://github.com/jmarcet/dockers)

What do you think?

**Comment by feckert** [comment]
@jmarcet Thanks for your feedback and your thoughts
> Last, we could add support for defining macvlan bridges directly in
> `/etc/config/network`. At the moment you have to manually add the bridge if you
> want to use it from docker.

I think this is already possible neifd could setup this. I have the following uci config in /etc/config/network to setup a macvlan




> ```
> docker-compose.yml
> version: '2.4'                                                                                                                                                                                                                                                                 
>                                                                                                                                                                                                                                                                                
> services:                                                                                                                                                                                                                                                                      
>   someservice:
>     [...]
>     networks:
>       somelan:
>         ipv4_address: 192.168.1.65
> 
> networks:
>   somelan:
>     driver: macvlan
>     driver_opts:
>       parent: eth0
>       macvlan_mode: bridge
>     ipam:
>       config:
>         - subnet: 192.168.1.64/29
> ```
> 
> But you need to set the macvlan bridge yourself:
> 
> ```
> ip link add veth link eth0 type macvlan mode bridge
> ```
> 
> This can be added as a hotplug script to make it automatic.
> 
> Again, this could be much nicer if we had a macvlan bridge option in openwrt's
> interfaces.
> 
> For reference, you can see the complete set of docker-compose files I am using
> in my openwrt managed home lan [here](https://github.com/jmarcet/dockers)
> 
> What do you think?

**Comment by jmarcet** [comment]
@feckert Wow, that was fast!

> @jmarcet Thanks for your feedback and your thoughts
> I still have to look at the other commentaries
> 
> But for the last one:
> 
> > Last, we could add support for defining macvlan bridges directly in
> > `/etc/config/network`. At the moment you have to manually add the bridge if you
> > want to use it from docker.
> 
> I think this is already possible. The service netifd could setup this. I have the following uci config in /etc/config/network to setup a macvlan docker setup.
> 

Really interesting. This just needs to be documented then.

**Comment by lisaac** [comment]
@jmarcet @feckert 
Thanks for your sharing!

> this could be much nicer if we had a macvlan bridge option in openwrt's
interfaces.

We do have macvlan bridge option in it, and if you want, you can use dockerman to add macvlan network, dockerman will add macvlan interface automatically(with the option)

**Comment by G-M0N3Y-2503** [comment]
> * In principle I like the idea of controlling dockerd parameters (the ones
>   which end up in /etc/docker/daemon.json) from uci, _but_ there is a
>   plethora of potentially interesting options. How do you envision supporting
>   them all?

I Imagine we'll only cover the options that are frequently used, at least until someone adds them all or writes something to serialize/deserialize the UCI to the `daemon.json` or CLI arguments.

But now I think of it, we should have a way to not use the UCI for advanced configurations.
If I'm not mistaken @feckert, the only way we can do this with this PR is by deleting `/etc/config/dockerd` and placing out own `/tmp/dockerd/daemon.json`. However, we probably want a persistent directory where a config file can be placed for advanced configurations. maybe we optionally add ` --config-file="$DOCKERD_CONF"` only if a `/etc/config/dockerd` exists? that way people can just use the default configuration file.

Otherwise, it looks like we agree that `docker-ce` will focus on being manually configurable and expose aspects through the UCI to make integration easier with other applications.

File: utils/docker-ce/files/dockerd.init [hunk 2]
```diff
@@ -3,12 +3,101 @@
...
 	procd_open_instance
 	procd_set_param stderr 1
-	procd_set_param command /usr/bin/dockerd
+	procd_set_param command /usr/bin/dockerd --config-file="$DOCKERD_CONF"
 	procd_set_param limits nofile="${nofile} ${nofile}"
 	procd_close_instance
 }
+
+ip4tables_remove_nat() {
+	iptables -t nat \
```
G-M0N3Y-2503, line 46: We are using a fair amount of vertical space for these commands, IMO we could just have one command per line.
Also, are these all the commands that get added?

feckert:
> But now I think of it, we should have a way to not use the UCI for advanced configurations.

That's what I thought too, that we won't get all configuration options into the uci.
I have now designed it in a way that nothing changes for the users who already use it.
If there is the config option in the globals uci section `alt_config_file` then this will be used.
This is default and we use `/etc/daemon.json`.

@G-M0N3Y-2503 I would like to finish this PR to get this in.
After that we can discuss the next steps.
https://github.com/openwrt/packages/pull/12556#issuecomment-648867398
https://github.com/openwrt/packages/pull/12556#issuecomment-648866347
https://github.com/openwrt/packages/pull/12556#issuecomment-648865084

utils/docker-ce/files/dockerd.init
```diff
@@ -3,12 +3,83 @@
 USE_PROCD=1
 START=25
 
+DOCKERD_CONF="/tmp/dockerd/daemon.json"
+
+json_add_array_string() {
+	json_add_string "" "$1"
+}
+
+process_config() {
```
G-M0N3Y-2503, line 12: I think it makes sense for `process_config()` to either return the config file or set `DOCKERD_CONF` directly, rather than using symlinks.

if `/etc/config/dockerd` doesn't exist `DOCKERD_CONF` equals `""` because the daemon will just startup with its defaults and not mislead anyone looking at the process command line.

if `alt_config_file` is set in the UCI, `DOCKERD_CONF` will equal that if it exists or `""` because it's most likely an error.

otherwise, we set `DOCKERD_CONF` to the one the UCI will generate `"/tmp/dockerd/daemon.json"`

G-M0N3Y-2503:
This Implementation looks a little more complicated than it needs to be, perhaps I wasn't clear about what I think is important.

I think longterm we want the UCI config to be the default configuration file as it will probably be enough for most users. So we will have to break the previous configuration for this, preferably before a mainline release. I'm leaning towards ripping off the band-aid with this PR, but I'm open to suggestions as to when.

So with that, It might be nicer from a maintainers perspective to just have a default `/etc/docker/daemon.json` and the one generated by the UCI config. The simplest implementation I can think of would be to `[ -f /etc/docker/daemon.json ]` and use it if it exists otherwise we use the UCI generated one, then we could default to UCI by not installing a `/etc/docker/daemon.json`.

If we still want an `alt_config_file` that's fine, I'm just wondering what the use case is that justifies this extra complication is...
But with that, the breaking change we'd have would be by not having `alt_config_file` defined by default.

utils/docker-ce/files/dockerd.init
```diff
@@ -3,12 +3,83 @@
...
+	local alt_config_file data_root log_level
+
+	[ -f /etc/config/dockerd ] || return 0
+
+	rm -f "$DOCKERD_CONF"
+	config_load 'dockerd'
+
+	config_get alt_config_file globals alt_config_file
```
G-M0N3Y-2503, line 20: This might get reworked depending on how you want to address the previous comments, so FYI, this looks like it should have been
```suggestion
	config_get alt_config_file globals ""
```

utils/docker-ce/files/etc/config/dockerd
```diff
@@ -0,0 +1,7 @@
+
+config globals 'globals'
+	option alt_config_file "/etc/daemon.json"
```
G-M0N3Y-2503, line 3: This might get reworked depending on how you want to address the previous comments, so FYI, this looks like it should have been
```suggestion
	option alt_config_file "/etc/docker/daemon.json"
```

feckert: I think we should do the link.
I've seen this before in openWrt.

- https://github.com/openwrt/packages/blob/master/net/keepalived/files/keepalived.init#L500
- https://github.com/openwrt/packages/blob/master/net/stunnel/files/stunnel.init#L342

feckert:
I should also mention that from my point of view it is important that we store the uci generated configuration in tmp.
This reduces the write cycles, OpenWrt is made for embedded systems in mind.

G-M0N3Y-2503:
hmmm, I think I may not have been clear, I wasn't talking about whether the config should be in `/tmp` or not. 
I've created a patch to help explain what I think should be done at a minimum.

In summary:
* Use `/etc/config/dockerd` as the default configuration going forward.
* If no config can be found let the daemon use its defaults.

```diff
diff --git a/utils/docker-ce/Makefile b/utils/docker-ce/Makefile
index 8ed184c40..f8b1d13e0 100644
--- a/utils/docker-ce/Makefile
+++ b/utils/docker-ce/Makefile
@@ -124,9 +124,6 @@ define Package/docker-ce/install
 	$(INSTALL_DIR) $(1)/etc/init.d
 	$(INSTALL_BIN) ./files/dockerd.init $(1)/etc/init.d/dockerd
 
-	$(INSTALL_DIR) $(1)/etc/docker
-	$(INSTALL_CONF) ./files/daemon.json $(1)/etc/docker/
-
 	$(INSTALL_DIR) $(1)/etc/config
 	$(INSTALL_CONF) ./files/etc/config/dockerd $(1)/etc/config/dockerd
 
diff --git a/utils/docker-ce/files/daemon.json b/utils/docker-ce/files/daemon.json
deleted file mode 100644
index 53c03211f..000000000
--- a/utils/docker-ce/files/daemon.json
+++ /dev/null
@@ -1,4 +0,0 @@
-{
-    "data-root": "/opt/docker/",
-    "log-level": "warn"
-}
diff --git a/utils/docker-ce/files/dockerd.init b/utils/docker-ce/files/dockerd.init
index c2e5c4dcc..c892c203b 100644
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -12,14 +12,18 @@ json_add_array_string() {
 process_config() {
 	local alt_config_file data_root log_level
 
-	[ -f /etc/config/dockerd ] || return 0
-
 	rm -f "$DOCKERD_CONF"
+
+	[ -f /etc/config/dockerd ] || {
+		# Use the daemon default configuration
+		DOCKERD_CONF=""
+		return 0
+	}
+
 	config_load 'dockerd'
 
 	config_get alt_config_file globals alt_config_file
 	[ -n "$alt_config_file" ] && [ -f "$alt_config_file" ] && {
-		rm -f "$DOCKERD_CONF"
 		ln -s "$alt_config_file" "$DOCKERD_CONF"
 		return 0
 	}
diff --git a/utils/docker-ce/files/etc/config/dockerd b/utils/docker-ce/files/etc/config/dockerd
index f49860c6a..b89d5e4bf 100644
--- a/utils/docker-ce/files/etc/config/dockerd
+++ b/utils/docker-ce/files/etc/config/dockerd
@@ -1,7 +1,6 @@
-
 config globals 'globals'
-	option alt_config_file "/etc/daemon.json"
-#	option data_root "/opt/docker/"
-#	option log_level "warn"
+#	option alt_config_file "/etc/docker/daemon.json"
+	option data_root "/opt/docker/"
+	option log_level "warn"
 #	list registry_mirror "https://<my-docker-mirror-host>"
 #	list registry_mirror "https://hub.docker.com"
```
[patch.txt](https://github.com/openwrt/packages/files/4856589/patch.txt)

feckert:
I understand now what you mean.

I agree with you that we use the uci variant as default.
So that we start with the uci when on next openwrt release.

I added your suggestion, so we return if no uci config for docker is found, and use the docker config default location.

The one thing I did not change is the installation of the daemon.json file you suggested.
Why is this necessary?

@G-M0N3Y-2503 please check my changes. If I get a thumb up I will merge this.

utils/docker-ce/files/dockerd.init
```diff
@@ -3,12 +3,92 @@
...
+		# Use the daemon default configuration
+		DOCKERD_CONF=""
+		return 0
+	}
+
+	config_load 'dockerd'
+
+	config_get alt_config_file globals alt_config_file
+	[ -n "$alt_config_file" ] && [ -f "$alt_config_file" ] && {
+		rm -f "$DOCKERD_CONF"
```
G-M0N3Y-2503, line 27: This could be removed if you wanted to as it's redundant, it won't exist at this point because of line 15

G-M0N3Y-2503:
> The one thing I did not change is the installation of the daemon.json file you suggested.
> Why is this necessary?

It isn't necessary exactly, just that the same config can now be achieved by the default or empty `/etc/config/dockerd`, so `/etc/docker/daemon.json` isn't needed as far as I can see.

But this PR looks good to me either way.

feckert:
@G-M0N3Y-2503 Thanks for your input and review :+1:

lisaac:
@feckert @G-M0N3Y-2503 
I just review the code.

In order to allow manually configures `daemon.json`, can we compare `daemon.json` with the UCI configuration, and modify the `daemon.json`, just like `dockerd-config.lua` did.

G-M0N3Y-2503:
I don't think I understand @lisaac...

Do you mean that we should have the UCI override parts of the daemon.json?
If so, it might be easier just to add options to the UCI when there are commonly needed or just take ownership of an entire `daemon.json` and set `alt_config_file` to point to it.

lisaac:
@G-M0N3Y-2503 
>Do you mean that we should have the UCI override parts of the daemon.json?
Yes, exactly.

>If so, it might be easier just to add options to the UCI when there are commonly needed or just take ownership of an entire daemon.json and set alt_config_file to point to it.
Yes, it works, just different way.

PR #12558, opened 2020-06-20
iOS stuff

neheb:
Maintainer: me
Compile tested: i386 musl

Upstream added official tarballs. Switched to those for speed.

Most of these are updates for iOS 13.5.1.


neheb:
Tested going in and out of recovery mode. Good enough.

PR #12569, opened 2020-06-21
sqm-scripts: Update Makefile

DocMAX:
remove +kmod-sched-cake-virtual

### DIFF
--- a/net/sqm-scripts/Makefile
+++ b/net/sqm-scripts/Makefile
@@ -25,8 +25,7 @@ define Package/sqm-scripts
   SECTION:=net
   CATEGORY:=Base system
   DEPENDS:=+tc +kmod-sched-core +kmod-ifb +iptables \
-	+iptables-mod-ipopt +iptables-mod-conntrack-extra \
-	+kmod-sched-cake-virtual
+	+iptables-mod-ipopt +iptables-mod-conntrack-extra
   TITLE:=SQM Scripts (QoS)
   PKGARCH:=all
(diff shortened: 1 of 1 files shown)


feckert:
Please sign off your commit and add a valid header for your commit message.
For example:
`sqm-scripts: remove kmod-sched-cake-virtual dependency`

hnyman:
What are your reasons for this proposal of removing cake dependency?

I do not see very likely that the cake dependency would be removed from SQM.


cc @tohojo @ldir-EDB0

tohojo:
Yeah, what @hnyman said - what's the rationale for this?

DocMAX:
Create index file './feeds/telephony.index'
WARNING: Makefile 'package/feeds/packages/sqm-scripts/Makefile' has a dependency on 'kmod-sched-cake-virtual', which does not exist
Installing all packages from feed packages.
Installing all packages from feed luci.
Installing all packages from feed routing.

hnyman:
> WARNING: Makefile 'package/feeds/packages/sqm-scripts/Makefile' has a dependency on 'kmod-sched-cake-virtual', which does not exist

You should not reveice that warning, as the SQM scripts only contains that dependency in master, where the new kmod-sched-cake-virtual also exists.

You should only receive that if you
* have dirty, unupdated repo
* try to use master packages with 19.07 OpenWrt

DocMAX:
you are right. i have a dirty tree. sorry, closed.

PR #12576, opened 2020-06-22
squeezelite: remove WMA support

neheb:
WMA uses ffmpeg, which has patented functionality.

WMA is also an old format.

Maintainer: @thess 
Compile tested: ath79

### DIFF
--- a/sound/squeezelite/Makefile
+++ b/sound/squeezelite/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=squeezelite
 PKG_VERSION:=1.9.6-1210
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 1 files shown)


karlp:
Where's the policy document on openwrt and patented software that supports removing things?

karlp:
I mean, we _have_ ffmpeg, so removing features from a package because it relies on ffmpeg seems.... circular at best.  Is the goal to remove everything that uses ffmpeg and then say, "look, nothing uses it, we can drop it" ?

neheb:
The goal is to remove ffmpeg support as BUILD_PATENTED is enabled by default.

An alternative way is to make WMA support conditional on BUILD_PATENTED.

And further to make BUILD_PATENTED on by default since I believe OpenWrt is not in a position to pay for licensing.

karlp:
Where is this discussion on removing ffmpeg and build patented taking place?  That's all relevant to this PR

neheb:
reworked to depend on BUILD_PATENTED.

Note that this changes nothing. SQUEEZELITE_WMA is disabled by default.

PR #12577, opened 2020-06-23
added libxcrypt and used for shadow-utils

neheb:
Maintainer: nobody
Compile tested: ath79
Run tested: malta

Made libxcrypt a static library in a non-standard location as other packages were picking it up and breaking.

ping @n8v8R


n8v8R:
Thank you for the effort, LGTM, hoping that it will get accepted this way.

Too bad 

>Made libxcrypt a static library in a non-standard location as other packages were picking it up and breaking.

Maybe maintainers could be notified about the availability of this library and one day it could be a dynamically linked from a standard location instead.

PR #12590, opened 2020-06-25
numpy: bump to version 1.19.0

commodo:
Maintainer: me
Compile tested: x86, ath79  https://github.com/openwrt/openwrt/commit/870588b6ebd8d818f65e5784bc866a8d67de0a31
Run tested: x86  https://github.com/openwrt/openwrt/commit/870588b6ebd8d818f65e5784bc866a8d67de0a31

--------------------------------------------

Bump host Cython version as well.
Add note near PKG_VERSION to remember to periodically update it.

With this version, we're also trying support for MIPS soft-float.

### DIFF
--- a/lang/python/numpy/Makefile
+++ b/lang/python/numpy/Makefile
@@ -6,11 +6,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=numpy
-PKG_VERSION:=1.18.5
-PKG_RELEASE:=2
+# Note: make sure to periodically update the Cython version in HOST_PYTHON3_PACKAGE_BUILD_DEPENDS
+PKG_VERSION:=1.19.0
+PKG_RELEASE:=1
 
 PYPI_NAME:=$(PKG_NAME)
-PKG_HASH:=34e96e9dae65c4839bd80012023aadd6ee2ccb73ce7fdf3074c62f301e63120b
+PKG_HASH:=76766cc80d6128750075378d3bb7812cf146415bd29b588616f72c943c00d598
 
 PKG_MAINTAINER:=Alexandru Ardelean <ardeleanalex@gmail.com>
 
@@ -21,7 +22,7 @@ PKG_CPE_ID:=cpe:/a:numpy:numpy
 # yes, zip... sigh
 PYPI_SOURCE_EXT:=zip
 PKG_BUILD_PARALLEL:=0
-HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:="Cython==0.29.19"
+HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:="Cython==0.29.20"
 
 include ../pypi.mk
 include $(INCLUDE_DIR)/package.mk


commodo:
@bhack    can you try this version?
i'm trying soft-float support, which may be the default for OpenWrt, however I don't have a MIPS board around to test this;i
i build-tested for MIPS soft-float;

i run-tested on x86; 

if that's confirmed, we can merge this from my side;

bhack:
Sorry I don't have the MIPS board under the hand anymore. Can be tested with openwrt qemu?

lang/python/numpy/patches/001-stub-fe-values-mips-soft-float.patch
```diff
@@ -0,0 +1,20 @@
...
+ /* General GCC code, should work on most platforms */
+ #  include <fenv.h>
+ 
++/* stub these for MIPS */
++#ifdef __mips_soft_float
++#define FE_INEXACT	0
++#define FE_UNDERFLOW	0
++#define FE_OVERFLOW	0
++#define FE_DIVBYZERO	0
++#define FE_INVALID	0
```
jefferyto, line 15: I think these would be available when compiling with glibc? Or is `__mips_soft_float` only defined by musl?

neheb: What's the point of stubbing when !soft_float is set in depends

commodo: Oops
Pushed wrong branch

commodo: > I think these would be available when compiling with glibc? Or is `__mips_soft_float` only defined by musl?

no idea; but if would make sense to do (even under the __mips_soft_float block)
```
#ifndef FE_UNDERFLOW
#define FE_UNDERFLOW 0
#endif 
```

commodo:
Changelog v1 -> v2:
* pushed correct branch with @!SOFT_FLOAT dep removed
* added `#ifndef` guards for the FE defines in case they are defined for some other libcs;

lang/python/numpy/patches/001-stub-fe-values-mips-soft-float.patch
bhack: Probably these are defined in c++11 

https://en.cppreference.com/w/cpp/numeric/fenv/FE_exceptions

bhack: See https://sourceware.org/git/?p=glibc.git;a=tree;f=sysdeps/mips/nofpu;hb=HEAD

commodo: ok; will remove the `#ifndef __mips_soft_float`  block

commodo:
Changelog v2 -> v3:
* removed `__mips_soft_float` guard

jefferyto:
I haven't tested this with mips yet, but is it safe to define the missing `FE_*` macros?

From [cppreference.com](https://en.cppreference.com/w/cpp/numeric/fenv/FE_exceptions) (emphasis mine):

> All these macro constants (except FE_ALL_EXCEPT) expand to integer constant expressions that are distinct powers of 2, which uniquely identify all supported floating-point exceptions. **Each macro is only defined if it is supported.**

From the [glibc manual](https://www.gnu.org/software/libc/manual/html_node/Status-bit-operations.html) (emphasis mine):

> These constants represent the various IEEE 754 exceptions. Not all FPUs report all the different exceptions. **Each constant is defined if and only if the FPU you are compiling for supports that exception,** so you can test for FPU support with ‘`#ifdef`’. They are defined in fenv.h.

Getting this to compile is nice but I would be interested to see if the package actually works on soft float platforms.

commodo:
> I haven't tested this with mips yet, but is it safe to define the missing `FE_*` macros?
> 
> From [cppreference.com](https://en.cppreference.com/w/cpp/numeric/fenv/FE_exceptions) (emphasis mine):
> 
> > All these macro constants (except FE_ALL_EXCEPT) expand to integer constant expressions that are distinct powers of 2, which uniquely identify all supported floating-point exceptions. **Each macro is only defined if it is supported.**
> 
> From the [glibc manual](https://www.gnu.org/software/libc/manual/html_node/Status-bit-operations.html) (emphasis mine):
> 
> > These constants represent the various IEEE 754 exceptions. Not all FPUs report all the different exceptions. **Each constant is defined if and only if the FPU you are compiling for supports that exception,** so you can test for FPU support with ‘`#ifdef`’. They are defined in fenv.h.
> 
> Getting this to compile is nice but I would be interested to see if the package actually works on soft float platforms.

I was hoping there would be a bit more interest in the soft-float part.
Initially [when I introduced numpy], I felt a slight disappointment when I added the SOFT_FLOAT dependency.
I guess, since there isn't anyone able to test this (on soft-float), we can defer it for another time and make this just a version bump.

commodo:
Changelog v3 -> v4:
* leaving SOFT_FLOAT limitation in place; just doing a version bump

will rebase this, if https://github.com/openwrt/packages/pull/12653 is merged before this

jefferyto:
> I guess, since there isn't anyone able to test this (on soft-float), we can defer it for another time and make this just a version bump.

You can compile this for malta-be and test with qemu.

I've compiled v3 for malta-be. The module loads but I don't know which functions I should call to test floating point operations. (I don't have time right now to trace through numpy source code to find out.)

neheb:
https://github.com/openwrt/packages/pull/12653 was merged.

commodo:
rebased on top of new master

commodo:
any objections on this as-is?
it's just a version update now;

PR #12593, opened 2020-06-25
python-requests: bump to version 2.24.0

commodo:
Maintainer: @BKPepe , me
Compile tested: https://github.com/openwrt/openwrt/commit/870588b6ebd8d818f65e5784bc866a8d67de0a31   x86
Run tested:  https://github.com/openwrt/openwrt/commit/870588b6ebd8d818f65e5784bc866a8d67de0a31   x86

### DIFF
--- a/lang/python/python-requests/Makefile
+++ b/lang/python/python-requests/Makefile
@@ -8,8 +8,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-requests
-PKG_VERSION:=2.23.0
-PKG_RELEASE:=2
+PKG_VERSION:=2.24.0
+PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


BKPepe:
@commodo Should we update it in OpenWrt 19.07?

commodo:
> @commodo Should we update it in OpenWrt 19.07?

no idea;
i still don't have a clear view/idea on 19.07 getting updates;
i can do a PR if you want;

hnyman:
> i still don't have a clear view/idea on 19.07 getting updates;

19.07 should get security updates /fixes
19.07 can get backports of other fixes, assuming that they do not affect other packages or invalidate current uci config etc.

commodo:
I can't tell if this version bump really has only security updates.
Maybe for 19.07 it may not make sense to update?

PR #12601, opened 2020-06-26
libpam: update to 1.4.0

neheb:
Remove upstreamed patch and add new ones to fix compilation.

Add some more configure options.

Maintainer: @nmav 
Compile tested: ath79


libs/libpam/patches/020-backport-musl-patch.patch
```diff
@@ -0,0 +1,273 @@
...
+--- a/libpam/libpam.map
++++ b/libpam/libpam.map
+@@ -82,3 +82,8 @@ LIBPAM_1.4 {
+   global:
+     pam_start_confdir;
+ } LIBPAM_1.0;
++
++LIBPAM_MODUTIL_1.4.1 {
++  global:
++    pam_modutil_check_user_in_passwd;
```
nmav, line 69: Is this patch necessary? It seems to be introducing a new API/ABI for the library which may not appear, or may appear modified in 1.4.1.

libs/libpam/patches/020-backport-musl-patch.patch
```diff
@@ -0,0 +1,273 @@
...
+ int
+ pam_sm_authenticate(pam_handle_t *pamh, int flags UNUSED,
+ 		    int argc, const char **argv)
+@@ -173,7 +91,7 @@ pam_sm_authenticate(pam_handle_t *pamh, int flags UNUSED,
+ 		return rc == PAM_CONV_AGAIN ? PAM_INCOMPLETE : rc;
+ 	}
+ 
+-	return check_user_in_passwd(pamh, user_name, file_name);
++	return pam_modutil_check_user_in_passwd(pamh, user_name, file_name);
+ }
```
nmav, line 271: Maybe we replicate the function's actions here so we don't introduce the API?

libs/libpam/patches/030-musl.patch
```diff
@@ -0,0 +1,73 @@
...
+-
+-	fclose (fp);
+-
+-	if (errn != 0 && errn != ENOENT) {
+-		pam_syslog(pamh, LOG_ERR, "unable to enumerate local accounts: %m");
+-		return -1;
+-	} else {
+-		return found;
+-	}
++	return pam_modutil_check_user_in_passwd(pamh, user, NULL) == PAM_SUCCESS;
```
nmav, line 70: Same here

nmav:
Thank you for getting that ready and bringing the right patches. It looks good to me, but it introduces a new API and if the final version modifies it we end up with a forked version and an incompatible ABI. I'd feel more comfortable including this if we didn't export the symbol.

libs/libpam/Makefile
```diff
@@ -45,12 +45,16 @@ CONFIGURE_ARGS += \
...
+	--disable-debug \
+	--disable-doc \
+	--disable-econf \
 	--disable-lckpwdf \
 	--disable-nis \
+	--disable-prelude \
 	--disable-regenerate-docu \
 	--disable-rpath \
 	--disable-selinux \
+	--disable-Werror \
```
nmav, line 57: This seems good in general. What kind of warnings are you trying to avoid?

neheb: I think this is already default. I just made it explicit.

Format warnings have typically broken builds.

libs/libpam/patches/030-musl.patch
neheb: Could work since this function is static. I'll see about that later.

libs/libpam/patches/020-backport-musl-patch.patch
neheb: Sort of. I figured that the simplest solution was to backport everything.

neheb:
Copied the static function and adjusted it to the new one. Builds now.

PR #12604, opened 2020-06-26
v4l2rtpspserver: update to 0.2.0

neheb:
Updated live555 to 2020.06.23.

Maintainer: @roger- 
Compile tested: ath79


roger-:
Looks like OpenSSL support needs to be disabled when building the new live555.

neheb:
I'll update this when the new version comes out. It seems git has important fixes.

PR #12619, opened 2020-06-26
syslog-ng: detect disabled IPv6 on loopback and fallback to IPv4

Cynerd:
Maintainer: @BKPepe 
Compile tested: 
Run tested: Turris Omnia (OpenWrt 19.07)

Description:
Binding in default to IPv6 is preferable but it can be disabled in
kernel and that prevents syslog-ng to start. This setup should not be
that common but syslog is very important service and should survive
that.

This introduces new plugin defining source generator
`network_localhost`. This is used instead of original network source.

### DIFF
--- a/admin/syslog-ng/Makefile
+++ b/admin/syslog-ng/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=syslog-ng
 PKG_VERSION:=3.27.1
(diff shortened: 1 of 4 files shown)


BKPepe:
Cross-reference: https://gitlab.labs.nic.cz/turris/turris-build/issues/158#note_160429

Thanks! It works. For the next time, I will increase my test coverage.

PR #12631, opened 2020-06-27
miniupnpd: remove libcap(-ng) dependency

neheb:
Not needed.

Maintainer: nobody
Compile tested: ath79
Run tested: Archer C7v2


neheb:
Specifically I tested it with libcap-ng. Is there a point to libcap-ng?

neheb:
Oh I see. Because this is running as root, libcap(-ng) is used to limit capabilities.

neheb:
Fixes: https://github.com/openwrt/packages/issues/12625

PR #12633, opened 2020-06-28
freetype: update to 2.10.2

neheb:
Switched to CMake to simplify the Makefile and to remove autotools
hacks.

Replaced InstallDev with CMAKE_INSTALL.

Added bzip2 capability.

Maintainer: @val-kulkov 
Compile tested: ath79

### DIFF
--- a/libs/freetype/Makefile
+++ b/libs/freetype/Makefile
@@ -8,22 +8,24 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freetype
-PKG_VERSION:=2.10.1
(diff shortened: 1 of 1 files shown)


val-kulkov:
@neheb : can you please check that the switch to CMake does not break compilation of packages that depend on libfreetype? Such as, for example, libgd and php7-mod-gd?

neheb:
Turns out fontconfig fails with bzip2 enabled. The reason is that the BZip2 package in OpenWrt does not include a pkgconfig file.

I'd rather not deal with that so disabled.

PR #12637, opened 2020-06-28
tiff: switch to CMake

neheb:
Allows removing autotools hacks. Also simplified InstallDev section.

Added ABI_VERSION parameter for proper rebuilds. Also used in the
pkgconfig path fixup.

Small cleanups for consistency between packages.

Maintainer: @jslachta 
Compile tested: ath79

### DIFF
--- a/libs/tiff/Makefile
+++ b/libs/tiff/Makefile
@@ -9,31 +9,27 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tiff
 PKG_VERSION:=4.1.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://download.osgeo.org/libtiff
 PKG_HASH:=5d29f32517dadb6dbcd1255ea5bbc93a2b54b94fbf83653b4d65c7d6775b8634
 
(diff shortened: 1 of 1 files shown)


micmac1:
Hi all,

It looks like this needs

`TARGET_CFLAGS+=$(TARGET_CPPFLAGS)`

to make use of fortify headers.

The ABI_VERSION should be 5, as that's the SONAME of the libs (4 is just the version of the tiff source code). It'd make sense to use $(ABI_VERSION) in Package/libtiff/install and Package/libtiffxx/install to notice SONAME changes in the future. The libtiff.so and libtiffxx.so symlinks don't need to be installed on targets, so something like

`$(CP) $(PKG_INSTALL_DIR)/usr/lib/libtiff.so.$(ABI_VERSION)* $(1)/usr/lib`

should do.

Kind regards,
Seb

neheb:
Updated.

micmac1:
>Updated.

Can you up ABI_VERSION to 5?

neheb:
Yeah. Builds now.

micmac1:
On Mon, Jun 29, 2020 at 09:23:38PM +0000, Rosen Penev wrote:
> Yeah. Builds now.

Thanks!

Small nitpick: Package/libtiffxx/install still installs the symlink
libtiffxx.so to target.

Otherwise LGTM.

Kind regards,
Seb

jslachta:
Thank you, guys!

PR #12647, opened 2020-06-29
gdbm: fix GCC 10 multiple def compilation error

neheb:
Signed-off-by: Syrone Wong <wong.syrone@gmail.com>

Maintainer: @Naoir 
Compile tested: ath79

ping @wongsyrone


wongsyrone:
LGTM.

PR #12650, opened 2020-06-29
gstreamer updates

neheb:
Maintainer: @flyn-org 
Compile tested: ath79

Added several plugins. Removed deinterlace plugin since it wasn't compiling for some reason.


flyn-org:
Looks good. Recommend merge.

Issue #12617, opened 2020-06-26
python-attrs: compilation error


feckert:
Maintainer: @jefferyto 
Environment: x86_64, APU3, OpenWrt master

Description:

I get the following error if I enable docker-compose package.
This error does not happen all the time on every build.
I thought it has to do with parallel builds, so I disable this in the Makefile for the package with `PKG_BUILD_PARALLEL:=0`.
But that didn't help either.
@jefferyto Any hints for this issue would be create!

```
Traceback (most recent call last):
  File "setup.py", line 105, in <module>
    setup(
  File "/home/bbworker/bbworker/owrt_master_x86_64/System6/build/openwrt/staging_dir/hostpkg/lib/python3.8/site-packages/setuptools/__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "/home/bbworker/bbworker/owrt_master_x86_64/System6/build/openwrt/staging_dir/target-x86_64_musl/usr/lib/python3.8/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/home/bbworker/bbworker/owrt_master_x86_64/System6/build/openwrt/staging_dir/target-x86_64_musl/usr/lib/python3.8/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/home/bbworker/bbworker/owrt_master_x86_64/System6/build/openwrt/staging_dir/target-x86_64_musl/usr/lib/python3.8/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "/home/bbworker/bbworker/owrt_master_x86_64/System6/build/openwrt/staging_dir/hostpkg/lib/python3.8/site-packages/setuptools/command/install.py", line 61, in run
    return orig.install.run(self)
  File "/home/bbworker/bbworker/owrt_master_x86_64/System6/build/openwrt/staging_dir/target-x86_64_musl/usr/lib/python3.8/distutils/command/install.py", line 545, in run
    self.run_command('build')
  File "/home/bbworker/bbworker/owrt_master_x86_64/System6/build/openwrt/staging_dir/target-x86_64_musl/usr/lib/python3.8/distutils/cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "/home/bbworker/bbworker/owrt_master_x86_64/System6/build/openwrt/staging_dir/target-x86_64_musl/usr/lib/python3.8/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
[...]

Issue #12557, opened 2020-06-20
pythom3-cffi: packages/cffi/cffi_opcode.py': No such file or directory


peterwillcn:
pythom3-cffi

```
t/master/openwrt/staging_dir/target-aarch64_generic_musl" _python_prefix="/usr" _python_exec_prefix="/usr" ; fi; )
    Running setup.py install for cffi: finished with status 'error'
    Running setup.py install for cffi: finished with status 'error'
    ERROR: Command errored out with exit status 1:
    copying cffi/error.py -> build/lib.linux2-3.8/cffi
    copying cffi/_cffi_errors.h -> build/lib.linux2-3.8/cffi
    c/_cffi_backend.c:6245:5: warning: 'ffi_prep_closure' is deprecated: use ffi_prep_closure_loc instead [-Wdeprecated-declarations]
    error: could not delete '/home/openwrt/master/openwrt/staging_dir/hostpkg/lib/python3.8/site-packages/cffi/cffi_opcode.py': No such file or directory
    ERROR: Command errored out with exit status 1:
    copying cffi/error.py -> build/lib.linux2-3.8/cffi
    copying cffi/_cffi_errors.h -> build/lib.linux2-3.8/cffi
    c/_cffi_backend.c:6245:5: warning: 'ffi_prep_closure' is deprecated: use ffi_prep_closure_loc instead [-Wdeprecated-declarations]
    copying build/lib.linux2-3.8/cffi/error.py -> /home/openwrt/master/openwrt/staging_dir/hostpkg/lib/python3.8/site-packages/cffi
    copying build/lib.linux2-3.8/cffi/_cffi_errors.h -> /home/openwrt/master/openwrt/staging_dir/hostpkg/lib/python3.8/site-packages/cffi
    error: [Errno 2] No such file or directory
ERROR: Command errored out with exit status 1: /home/openwrt/master/openwrt/staging_dir/hostpkg/bin/python3.8 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/openwrt/master/openwrt/tmp/pip-install-6kx8xxye/cffi/setup.py'"'"'; __file__='"'"'/home/openwrt/master/openwrt/tmp/pip-install-6kx8xxye/cffi/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"
[...]

PR #12653, opened 2020-06-30
python-packages: Disable parallel build when host pip is needed

jefferyto:
Maintainer: multiple (ping @commodo)
Compile tested: armvirt-64, 2020-06-29 snapshot sdk
Run tested: none

Description:
This adds `PKG_BUILD_PARALLEL:=0` to packages that depend on host Python packages (`HOST_PYTHON3_PACKAGE_BUILD_DEPENDS`), because installing packages with multiple concurrent pip processes can lead to [errors or unexpected results][1].

This also:

* Move `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` definitions to before python3-package.mk is included

* Update Python folder readme to include `PKG_BUILD_PARALLEL:=0`

[1]: https://github.com/pypa/pip/issues/2361


jefferyto:
I believe this fixes #12557; the log seems to indicate that both bcrypt and cryptography packages are installing host cffi at the same time and both trying to delete the same file. I have not been able to reproduce the error though and the issue reporter has not responded.

Also apologies to @commodo - I know this conflicts with #12590. I'm not sure when that PR may be merged so I thought I would go ahead and submit this one.

feckert:
@jefferyto could this also fix https://github.com/openwrt/packages/issues/12617 ?

jefferyto:
@feckert I don't think so - python-attrs doesn't use `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` and you have already tried adding `PKG_BUILD_PARALLEL:=0`.

commodo:
i can't think of another easy fix for this

PR #12656, opened 2020-06-30
liburing: new package: liburing v0.6

gladiac1337:
Maintainer: Christian Lachner <gladiac@gmail.com>
Compile tested: mipsel, mips74k, ipq806x, x86, arc700

Description: New package: liburing v0.6
This is the liburing library. liburing provides helpers to setup and
teardown io_uring instances, and also a simplified interface for
applications that don't need (or want) to deal with the full kernel
side implementation.

This package was made primarily to support a [new VFS module](https://wiki.samba.org/index.php/Samba_4.12_Features_added/changed#.27io_uring.27_vfs_module) for the samba4 package and has been run-tested by @Andy2244 who is a maintainer for samba4 in openwrt.

This PR does the following:
- Add initial Makefile
- Add patch to fix inttypes.h inclusion when __kernel_timespec is unavailable

### DIFF
--- a/libs/liburing/Makefile
+++ b/libs/liburing/Makefile
@@ -33,9 +33,7 @@ endef
 
 CONFIGURE_ARGS:=--prefix=$(CONFIGURE_PREFIX) --cc=${TARGET_CC}
 
-ifeq ($(CONFIG_USE_MUSL),y)
-	TARGET_CFLAGS+=-Dloff_t=off_t
-endif
+TARGET_CFLAGS += $(if $(CONFIG_USE_MUSL),-Dloff_t=off_t)
 
 define Build/Compile
 	$(MAKE) $(PKG_BUILD_DIR) \
--- a/libs/liburing/Makefile
+++ b/libs/liburing/Makefile
@@ -20,7 +20,7 @@ include $(INCLUDE_DIR)/package.mk
 define Package/liburing
   SECTION:=libs
   CATEGORY:=Libraries
-  TITLE:=liburing
+  TITLE:=io_uring library
   URL:=https://git.kernel.dk/cgit/liburing
 endef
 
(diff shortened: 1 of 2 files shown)


karlp:
Have you sent the patch upstream? it seems relevant tothem, not just to us

gladiac1337:
> Have you sent the patch upstream? it seems relevant tothem, not just to us

Not yet, but it is on my todo list.

libs/liburing/Makefile
```diff
@@ -0,0 +1,65 @@
...
+endef
+
+CONFIGURE_ARGS:=--prefix=$(CONFIGURE_PREFIX) --cc=${TARGET_CC}
+
+ifeq ($(CONFIG_USE_MUSL),y)
+	TARGET_CFLAGS+=-Dloff_t=off_t
+endif
+
+ifeq ($(CONFIG_USE_UCLIBC),y)
+	TARGET_CFLAGS+=-fno-stack-protector
```
neheb, line 41: Is this really necessary?

https://github.com/openwrt/openwrt/commit/b933f9cf0cb254e368027cad6d5799e45b237df5 should have fixed this.

Wipe and rebuild your toolchain to make sure.

gladiac1337: I pulled from the master branch and also updated the feeds before I started working on this. However, I will check if -fno-stack-protecter is still necessary today. I will start from zero to make sure. Maybe reusing old .config-files is causing issues here...

gladiac1337: @neheb Ok, I just checked and you were right. I will reenable the stack-protector for uclibc targets.

libs/liburing/Makefile
```diff
@@ -0,0 +1,65 @@
...
+  liburing provides helpers to setup and teardown io_uring instances,
+  and also a simplified interface for applications that don't need
+  (or want) to deal with the full kernel side implementation.
+  For more info on io_uring, please see: https://kernel.dk/io_uring.pdf
+endef
+
+CONFIGURE_ARGS:=--prefix=$(CONFIGURE_PREFIX) --cc=${TARGET_CC}
+
+ifeq ($(CONFIG_USE_MUSL),y)
+	TARGET_CFLAGS+=-Dloff_t=off_t
```
neheb, line 37: I would rewrite this as

TARGET_CFLAGS += $(if $(CONFIG_USE_MUSL),-Dloff_t=off_t)

gladiac1337: Ok, done!

PR #12658, opened 2020-06-30
mini_snmpd: Allow specifying raw netdev names

NightTsarina:
Maintainer: Marcin Jurkowski <marcin1j@gmail.com>
Compile tested: no
Run tested: mips_24kc, TP-LINK TL-WR703N v1, 17.01.7, tested running the daemon with these changes

Description:
There are many cases where you want to specify a network interface name directly, instead of an UCI name. For example, I have a LAN bridge where I want to monitor the traffic of each interface separately. This change allows to specify either kind of name.

### DIFF
--- a/net/mini_snmpd/files/mini_snmpd.config
+++ b/net/mini_snmpd/files/mini_snmpd.config
@@ -21,6 +21,6 @@ config mini_snmpd 'default'
 	list disks '/tmp'
 	# enable basic network statistics on specified interface
 	# 4 interfaces maximum per instance, as named in /etc/config/network and luci
-	# not physical device names
+	# OR physical device names
 	list interfaces 'lan'
 	list interfaces 'wan'
--- a/net/mini_snmpd/files/mini_snmpd.init
+++ b/net/mini_snmpd/files/mini_snmpd.init
@@ -76,6 +76,10 @@ append_interface() {
 	[ -z $netdev_count ] && netdev_count=0
 	# for the purposes of snmp monitoring it doesn't need to be up, it just needs to exist in /proc/net/dev
 	network_get_device netdev "$name"
+	# Also accept raw device names.
+	if [ -z "$netdev" ] && grep -qF "$name" /proc/net/dev ]; then
+		netdev="$name"
+	fi
 	if [ -n "$netdev" ] && grep -qF "$netdev" /proc/net/dev ]; then
 		[ $netdev_count -ge 8 ] && {
 			_err "$cfg: too many network interfaces configured, ignoring $name"


neheb:

neheb:
```
Author name (Martina Ferrari) seems ok
 
Commit subject line seems ok (mini_snmpd: Allow specifying raw netdev names)
 
Signed-off-by is missing or doesn't match author (should be 'Signed-off-by: Martina Ferrari <tina@tina.pm>')
 ```

Fix by doing git commit --amend --author='Martina Ferrari <tina@debian.org>'

NightTsarina:
Thanks for spotting that. Fixed the email disparity.

neheb:
ping @marcinj1

marcin1j:
OpenWRT makes distinction between network interfaces (logical) and network devices (physical). Mixing them in one variable is a bad idea IMO. This approach also makes it hard to spot obvious configuration mistakes as misspelled interface name is silently treated as a network device name. Lastly, it's entirely legal (albeit stupid) for a different logical interface and physical device to share the same name.

How about a separate config variable for physical device names? This would make script longer but consistent with OpenWRT network interface model, allow properly handling all the strange configurations and seems to be more future-proof.

NightTsarina:
I agree that makes more sense, but I am not sure my knowledge of (l)uci is enough to implement it properly :)

marcin1j:
This script is a bit convoluted which make everything seem complex but actually it's quite easy. Your current solution is efficient with minimal amount of code added - of course that's an advantage. The only problem is that if we introduce it now, we can't change it later without breaking users' config.  I'm sure you can properly implement the complex variant too:) A bit more code but not that much.

Let me offer you a bunch of tips:

1. You need a function similar to `append_interface`, operating on a device level. Let's name it `append_device`.
2. Pick a config list name for network devices, say `devices`. You need to iterate over it in `start_instance` function and call `append_device` for each list entry. There's already such a loop for `interfaces` list:
`config_list_foreach "$cfg" 'interfaces' append_interface`
Similarly, you need something like this to iterate over `devices` list:
`config_list_foreach "$cfg" 'devices' append_device`
3. To fill function `append_device` you just need to move a part of the code from `append_interfaces`. I would move everything nested in the first `if` branch.
4. There seem to be an error with `netdev_count`: this variable is supposed to store the total number of daemon arguments and keep it between function calls but it's declared local. So either move ``[ -z $netdev_count ] && netdev_count=0` logic from `append_interfaces` to `append_device` or explicitly declare global variable `netdev_count=0`. Either way remove `local netdev_count` declaration!
5. In `append_interface` function you call the new `append_device` function passing `netdev` as an argument.

Of course it's only a suggestion (don't pay attention to names; pick whatever is sane and suits you) but that's all you need. There are other ways to do it, this was just the easiest to explain.

PR #12662, opened 2020-06-30
tiff: declare ABI_VERSION in libtiffxx as well

micmac1:
ABI_VERSION is used in Package/libtiffxx/install but not defined, so we
don't get SONAME treatment for libtiffxx. This commit adds the missing
ABI_VERSION.

Maintainer: @jslachta 
Compile tested: ath79 master
Run tested: no runtime-change

Description:
Hi Jiri,

This adds a missing var.

Kind regards,
Seb

### DIFF
--- a/libs/tiff/Makefile
+++ b/libs/tiff/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tiff
 PKG_VERSION:=4.1.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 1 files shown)


neheb:
Curious how this compiles in the first place since the tiffxx install section is using this variable. Is it being picked up from the previous location?

micmac1:
>Curious how this compiles in the first place since the tiffxx install section is using this variable. Is it being picked up from the previous location?

Because of the wildcard following the var. And no to 2nd part. Hence also no soname name suffix for the cxx lib pkg. If var would leak then utils pkg would have soname suffix as well.

PR #12671, opened 2020-07-01
treewide: alternatives instead of postinst, {pre,post}rm

zhanhb:
Maintainer: @RussellSenior, @neheb 
Compile tested: ath79/nand, mpc85xx/p2020, latest master
Run tested: ath79/nand, tests done

Description:
Make it work fine to upgrade package gzip.

### DIFF
--- a/utils/grep/Makefile
+++ b/utils/grep/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=grep
 PKG_VERSION:=3.4
-PKG_RELEASE:=1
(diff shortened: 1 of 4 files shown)


karlp:
Where is this ALTERNATIVES documented for package authors?  Ican't find a single mention of it on the www.openwrt.org page, especiallynot on the "creating packages" page, which has so farbeentheprimary source of guidance on creating packages.

zhanhb:
I can't find anything about ALTERNATIVES either. I just read the source code of opkg-lede and thought what ALTERNATIVES should be.

PR #12676, opened 2020-07-01
[19.07] bind: update to version 9.16.x

BKPepe:
Maintainer: @nmeyerhans 
Compile tested: Turris 1.1, powerpc_8540, OpenWrt 19.07.3
Run tested: Turris 1.1, powerpc_8540, OpenWrt 19.07.3

Description:

In OpenWrt 19.07, there's a version of bind 9.14, which is not supported anymore. It reached End of Life since May 2020.
It means that there are not going to be any security updates and as this is a DNS resolver, we should be using versions, which are supported. There was a possibility to use 9.11, but it would be a downgrade from 9.14. Let's use 9.16, which is used in OpenWrt master branch. I cherry-picked commits from OpenWrt master branch, compile it and run tested it on 19.07.

This perhaps should solve issue which was reported on Turris forum:
https://forum.turris.cz/t/bind-server-hangs-after-some-time-1-day-avg/13286?u=pepe

### DIFF
--- a/net/bind/Makefile
+++ b/net/bind/Makefile
@@ -9,7 +9,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bind
-PKG_VERSION:=9.16.2
+PKG_VERSION:=9.16.3
 PKG_RELEASE:=1
 USERID:=bind=57:bind=57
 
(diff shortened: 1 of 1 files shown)


BKPepe:
In some PRs we agreed that timeout for stalled PRs is 2 weeks and this passed. I'm merging this.

PR #12688, opened 2020-07-01
openfortivpn: Remove pingcheck and use l3_device instead

qweaszxcdf:
Maintainer: @lucize

Description:
some server may not allow ping, therefore pingcheck shoud be optional or removed

l3_device is better considering of pppoe connections


lucize:
should be ok once the checks are also ok

PR #12695, opened 2020-07-02
mc: fix mouse handling

BKPepe:
Maintainer: @dibdot 
Compile tested: Turris Omnia, mvebu (cortex-a9), OpenWrt master (kernel: 5.4.48)
Run tested: Konsole - KDE Terminal emulator version 20.04.2 and mouse clicking in mc

Description:
similar to https://github.com/openwrt/packages/pull/12694, but in OpenWrt master is different version of mc and I rather tested it on OpenWrt master as well to try to avoid any circumstances instead of doing just cherry-pick.

### DIFF
--- a/utils/mc/Makefile
+++ b/utils/mc/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mc
 PKG_VERSION:=4.8.24
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_MAINTAINER:=Dirk Brenken <dev@brenken.org>
 PKG_LICENSE:=GPL-3.0-or-later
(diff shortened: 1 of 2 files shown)


dibdot:
Hi Josef, personally I've never used mc with a mouse ... hotkeys are the way to go ... ;-))
Anyway, patch looks OK and you've tested that - good enough IMHO. Many thanks!

BKPepe:
Time to time, I am using mouse in mc, but I didn't care about it if it does not work, but someone reported it to us in the Czech thread [1] on Turris forum, so I just go ahead, took the patch from the issue, compile it and tested it.

[1] https://forum.turris.cz/t/nefunkcni-mys-v-mc-pres-putty-a-dalsi/13269

PR #12700, opened 2020-07-03
shadowsocks-libev: fix compilation with GCC10

neheb:

Maintainer: @yousong 
Compile tested: ath79


yousong:
Thanks for taking care of this.

For archive purposes, the following links should suffice for documentation

- https://gcc.gnu.org/gcc-10/porting_to.html#common
- `-fcommon`, https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html

PR #12709, opened 2020-07-03
mwan3: Fix mwan3 start not doing anything

m01:
Maintainer: @feckert
Compile tested: armv7l, Turris Omnia, Turris OS 5.0.2 (based on OpenWrt 19.07.3 I believe; note that this is a minimal script-only change)
Run tested: armv7l, Turris Omnia, Turris OS 5.0.2 (based on OpenWrt 19.07.3 I believe), mwan3 start/stop/restart, switch wan link from A->B

Description:
Due to a missing config load function call, mwan3 start runs ifup for an empty
list of interfaces, thus not calling ifup at all.

This PR introduces the missing config_load call.

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.8.7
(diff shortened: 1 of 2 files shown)


feckert:
Thanks :- )

PR #12711, opened 2020-07-04
liburcu: install pkgconfig files

neheb:
Newer versions of lttng* need these.

Maintainer: @salzmdan 
Compile tested: ath79


salzmdan:
Thanks!

PR #12716, opened 2020-07-04
c-ares: switch to CMake

neheb:
Allows simplifying the Makefile. Also allows faster compilation.

Maintainer: @karlp 
Compile tested: ath79


karlp:
Not at all convinced that it's simplified the makefile, it jsut spewed more magic cmake internals, but it sure is faster getting rid of the autohell stage, so that's a definite win!  I've merged this after rebasing in 9bd81604b thanks!

```
original
real	1m29.687s
user	1m12.567s
sys	0m14.702s

cmake
real	0m37.739s
user	0m33.837s
sys	0m8.341s
```

PR #12720, opened 2020-07-05
fwknop: fix compilation with GCC10

neheb:
Added PKG_LICENSE_FILES.

Simplified configure arguments.

Small cleanups for consistency between packages.

Maintainer: @jp-bennett 
Compile tested: ath79


DazzyWalkman:
gnupg as one of its optional dependency, will fail with GCC 10.

PR #12721, opened 2020-07-05
forked-daapd: fix compilation with GCC10

neheb:

Maintainer: @ejurgensen 
Compile tested: ath79


ejurgensen:
Thanks, looks good to me. I have also fixed it upstream (though slightly differently) so that the next version of forked-daapd won't require a patch.

PR #12724, opened 2020-07-05
add gl-mifi-mcu: GL.iNet GL-MiFI Power monitoring support

nunojpg:
Signed-off-by: Nuno Goncalves <nunojpg@gmail.com>

Maintainer: me
Compile tested: GL.iNet GL-MiFI, v19.07, master
Run tested: GL.iNet GL-MiFI, v19.07, master. Kernel module load and unload without crashes

Description: provides identical functionality to downstream GL.iNet opkg package gl-mifi-mcu but with much lower footprint.

This communicates with a MCU on board via a soft UART to retrieve battery SOC, temperature and charge status.

GL.iNet used a soft uart kernel module derived from https://github.com/adrianomarto/soft_uart, and a userland binary that would use tty interface to talk to the uart and obtain the values and then write them to /tmp/mcu_data.

I instead did a kernel module that every 5 seconds does the soft uart part to query and retrieve the data and make it available over /proc/gl_mifi_mcu.

### DIFF
--- a/utils/gl-mifi-mcu/Makefile
+++ b/utils/gl-mifi-mcu/Makefile
@@ -1,0 +1,38 @@
+include $(TOPDIR)/rules.mk
+include $(INCLUDE_DIR)/kernel.mk
+
+PKG_NAME:=gl-mifi-mcu
+PKG_RELEASE:=1
+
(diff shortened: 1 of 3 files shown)


nunojpg:
replaced with #12733

Issue #12725, opened 2020-07-05
libpng: compile fail


anomeome:
libpng 1.6.37 compile issue with master:

```
make[6]: Entering directory '/home/dev/wrtpac/source/build_dir/target-arm_cortex-a9+vfpv3_musl_eabi/libpng-1.6.37'
[  1%] Generating pnglibconf.out
/home/dev/wrtpac/source/staging_dir/host/bin/ccache: invalid option -- 'E'
Usage:
```

@neheb is this from the push ~yesterday?

neheb:
Indeed. How did you get this builds failure?

anomeome:
mvebu multiple targets (all the wrtpac devices). I cleaned up (dirclean) and removed all relevant directories, but error persists. I assume that is to be handed through to GCC?

anomeome:
PR resolves, thanks much.

PR #12729, opened 2020-07-05
libpng: fix compilation with ccache

neheb:
It seems an extra CMake variable is needed for one of the scripts.

Maintainer: @jow- 
Compile tested: ath79

Fixes: https://github.com/openwrt/packages/issues/12725


neheb:
Wow that's annoying. It interprets an empty variable as an actual file...

neheb:
Ah quoting error. Never mind.

PR #12733, opened 2020-07-06
gl-mifi-mcu: add GL.iNet GL-MiFI power monitoring kernel module

nunojpg:
Maintainer: me
Compile tested: GL.iNet GL-MiFI, v19.07, master
Run tested: GL.iNet GL-MiFI, v19.07, master. Kernel module load and unload without crashes

Description: provides identical functionality to downstream GL.iNet opkg package gl-mifi-mcu but with much lower footprint.

This communicates with a MCU on board via a soft UART to retrieve battery SOC, temperature and charge status.

GL.iNet used a soft uart kernel module derived from https://github.com/adrianomarto/soft_uart, and a userland binary that would use tty interface to talk to the uart and obtain the values and then write them to /tmp/mcu_data.

I instead did a kernel module that every 5 seconds does the soft uart part to query and retrieve the data and make it available over /proc/gl_mifi_mcu.

### DIFF
--- a/utils/gl-mifi-mcu/Makefile
+++ b/utils/gl-mifi-mcu/Makefile
@@ -1,0 +1,36 @@
+include $(TOPDIR)/rules.mk
+include $(INCLUDE_DIR)/kernel.mk
+
+PKG_NAME:=gl-mifi-mcu
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=Nuno Goncalves <nunojpg@gmail.com>
+PKG_LICENSE:=GPL-3.0-or-later
+
+include $(INCLUDE_DIR)/package.mk
+
+define KernelPackage/gl-mifi-mcu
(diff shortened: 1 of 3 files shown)


utils/gl-mifi-mcu/Makefile
```diff
@@ -0,0 +1,38 @@
+include $(TOPDIR)/rules.mk
+include $(INCLUDE_DIR)/kernel.mk
+
+PKG_NAME:=gl-mifi-mcu
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=Nuno Goncalves <nunojpg@gmail.com>
+PKG_LICENSE:=GPL-3.0
```
neheb, line 8: Should be -or-later.

BKPepe:
I wonder why this kernel package is compiled for each router supported in OpenWrt.

nunojpg:
@BKPepe because I don't know better. What do you suggest?

PR #12736, opened 2020-07-06
libqmi: enable selection of message collections

mips171:
Maintainer: me 
Compile tested: ath79 Telco T1 master
Run tested: ath79 Telco T1 master, checked basic functions

Description:
* Added new message collections support, so that users can select which messages to include in the built library during configure with the `--enable-collection` option. Three predefined sets are given:
   * `minimal`: the bare minimum messages required to control connectivity.
   * `basic`: all messages and indications that ModemManager requires. (selected here by default)
   * `full`: all supported messages and indications.

More information: https://lists.freedesktop.org/archives/libqmi-devel/2020-June/003341.html

### DIFF
--- a/libs/libqmi/Config.in
+++ b/libs/libqmi/Config.in
@@ -1,0 +1,21 @@
+menu "Configuration"
+depends on PACKAGE_libqmi
+
+	config LIBQMI_COLLECTIONS_MINIMAL
+		bool "Build with minimal message collection support"
+		default n
+		help
+		  the bare minimum messages required to control connectivity
+
+	config LIBQMI_COLLECTIONS_BASIC
+		bool "Build with basic message collection support"
+		default y
+		help
+		  all messages and indications that ModemManager requires
+
+	config LIBQMI_COLLECTIONS_FULL
+		bool "Build with full message collection support"
+		default n
+		help
+		  all supported messages and indications
+endmenu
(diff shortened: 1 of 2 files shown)


neheb:
Needs a PKG_RELEASE bump.

nickberry17:
> Needs a PKG_RELEASE bump.

Thanks, done

dangowrt:
Imho it'd be nicer to use package variants, so users can select at run-time rather than compile-time.

nickberry17:
> Imho it'd be nicer to use package variants, so users can select at run-time rather than compile-time.

I have been looking at, and have a couple of questions.  If we have three variants, libqmi-minimal/basic/full, how do we satisfy ModemManager's dependency on libqmi?  Is there a way to create a wrapper package somehow so that MM is always satisfied?  ModemManager depends on at least libqmi-basic, it will not work with libqmi-minimal.

neheb:
Needs a rebase.

feckert:
@nickberry17 You could work with PROVIDES please have a look to https://github.com/openwrt/packages/blob/60e60fc38f26ef3397411488dfe4a1117f04592a/net/mosquitto/Makefile#L46 who this works

BKPepe:
@nickberry17 Kind reminder, please.

Neustradamus:
It has been replaced by:
- https://github.com/openwrt/packages/pull/24873
- https://github.com/openwrt/packages/pull/18154

PR #12740, opened 2020-07-06
treewide: add and fix conffiles

zhanhb:
Maintainer: 
Compile tested: only package openssh, i2pd and sane-backends tested. (ath79/nand, mips_24kc, latest snapshot)
Run tested: only package openssh, i2pd and sane-backends tested.(ath79/nand, mips_24kc, latest snapshot)

**Description:**

The original configuration of package openssh-server which depends the files in dir `/etc/ssh` of host machine is incorrect and varies on different machines. Maybe the source code of [openwrt/include/package-ipkg.mk](https://github.com/openwrt/openwrt/blob/4c8eb4ad9eea29707b377c987b6f3be927d8e741/include/package-ipkg.mk#L238-L249) needs a patch to support wildcard conffiles.
```bash
wget -qO- https://downloads.openwrt.org/snapshots/packages/mips_24kc/packages/openssh-server_8.3p1-1_mips_24kc.ipk | \
    tar -zxOf- ./data.tar.gz | tar -zxOf- ./lib/upgrade/keep.d/openssh-server
/etc/ssh/ssh_host_dsa_key
/etc/ssh/ssh_host_ecdsa_key
/etc/ssh/ssh_host_rsa_key
/etc/ssh/ssh_host_dsa_key.pub
/etc/ssh/ssh_host_ecdsa_key.pub
/etc/ssh/ssh_host_rsa_key.pub
```
There's neither conffiles nor keepd generated for other packages.

### DIFF
--- a/admin/ipmitool/Makefile
+++ b/admin/ipmitool/Makefile
@@ -39,9 +39,6 @@ define Package/ipmitool/install
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/src/ipmitool $(1)/usr/sbin/
 endef
 
-define Package/ipmitool/conffiles
-endef
-
 CONFIGURE_ARGS += \
 	--enable-intf-lan \
 	--enable-intf-lanplus \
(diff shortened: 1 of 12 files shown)


neheb:
Wow. That OpenSSH thing is serious.

Issue #12741, opened 2020-07-07
getdns: missing dependency for libbsd.so.0


bogi788:
Maintainer: @jonathanunderwood
Environment: mips_24kc, TP-Link Archer C7 v2, OpenWRT Development Snapshot

Since #12679 getdns fails to build exactly as in #7019 due to missing its libbsd dependency:
https://downloads.openwrt.org/snapshots/faillogs/mips_24kc/packages/getdns/compile.txt

---

> the update seems to have brought #7019 back

Please open a new issue and provide more details including what version of OpenWRT you're using, the output you see etc.

_Originally posted by @jonathanunderwood in https://github.com/openwrt/packages/issues/12529#issuecomment-654674510_

bogi788:
#6960 fixed the problem previously

neheb:
Adding -DCMAKE_DISABLE_FIND_BSD=ON or something similar should fix it.

jonathanunderwood:
> Adding -DCMAKE_DISABLE_FIND_BSD=ON or something similar should fix it.

That's already there. I guess it's not working, and is probably the wrong flag.

jonathanunderwood:
> Adding -DCMAKE_DISABLE_FIND_BSD=ON or something similar should fix it.

That's already there. I guess it's not working, and is probably the wrong flag.

bogi788:
in #6960 they used CONFIGURE_VARS += LIBBSD_LIBS=-lc

jonathanunderwood:
> in #6960 they used CONFIGURE_VARS += LIBBSD_LIBS=-lc

Yes; the move to 1.6.0 also changed the buildsystem to CMake so that fix no longer applies.

jonathanunderwood:
> Adding -DCMAKE_DISABLE_FIND_BSD=ON or something similar should fix it.

Note that CMAKE_DISABLE_FIND_XXXXX works for dependency packages that are discovered by CMake using `find_package`. But, `libbsd` searching is not done using `find_package` but `find_library`, so the -DCMAKE_DISABLE_FIND_BSD=ON is a no-op.

PR #12745, opened 2020-07-07
getdns: properly remove libbsd support

neheb:
Also fixed compilation without deprecated OpenSSL APIs.

Maintainer: @jonathanunderwood 
Compile tested: ath79

Fixes: https://github.com/openwrt/packages/issues/12741

### DIFF
--- a/libs/getdns/Makefile
+++ b/libs/getdns/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=getdns
 PKG_VERSION:=1.6.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_LICENSE:=BSD-3-Clause
 PKG_LICENSE_FILES:=LICENSE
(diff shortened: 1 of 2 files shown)


jonathanunderwood:
This looks promising, thanks for beating me to it!

But it's unclear to me why that openssl patch is necessary. Can we break that out into a separate commit with a bit more of an explanation to aid future maintenance?

neheb:
It's just a missing header.

jonathanunderwood:
> It's just a missing header.

But I wasn't seeing any build breakage without it, hence my ask for a bit more detail and a separate commit, so in the future we know why it was needed.

 Has that patch been submitted upstream?

neheb:
https://github.com/getdnsapi/getdns/pull/482

Doubt it'll get merged since it seems it's an unbound tool.

Split up the commits.

jonathanunderwood:
Thanks @neheb really appreciate it.

**LGTM**

jonathanunderwood:
Aside: I wonder why are we enabling DSA in OpenWRT - it's known to be broken and insecure.

neheb:
https://github.com/getdnsapi/getdns/blob/develop/CMakeLists.txt#L140

jonathanunderwood:
> https://github.com/getdnsapi/getdns/blob/develop/CMakeLists.txt#L140

Interesting - I hadn't spotted that. Actually tempted to disable DSA in a future package release.

neheb:
I think that should be done.

PR #12746, opened 2020-07-07
qrencode: switch to CMake

neheb:
Allows simplifying the Makefile. Also faster compilation.

Added libpng dependency as it seems difficult to get rid of with CMake.

Maintainer: @jp-bennett 
Compile tested: ath79

### DIFF
--- a/libs/qrencode/Makefile
+++ b/libs/qrencode/Makefile
@@ -17,12 +17,13 @@ PKG_HASH:=c9cb278d3b28dcc36b8d09e8cad51c0eca754eb004cb0247d4703cb4472b58b4
 
 PKG_MAINTAINER:=Jonathan Bennett <JBennett@incomsystems.biz>
 PKG_LICENSE:=LGPL-2.1-or-later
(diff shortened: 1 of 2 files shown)


jp-bennett:
What does pulling libpng do to the overall install size? When qrencode was added to Openwrt, we added that option specifically to be able to squeeze it onto a 4 meg router. I guess Openwrt no longer officially targets the 4 meg platforms, so maybe it's a moot point.

neheb:
Increase in ipk size from 10638 11885 .

I can try removing libpng again if desired.

jp-bennett:
@neheb Don't forget the size of the libpng package in that calculation. cmakelists.txt includes a WITHOUT_PNG option, so it should be possible.

neheb:
Fixed it. Size of ipk is now 10790

jp-bennett:
@neheb Perfect, looks good to me, then.

neheb:
Added LTO and gc-sections. Size is now 10593. I think that's good enough.

PR #12747, opened 2020-07-07
tesseract/leptonica: switch to CMake

neheb:
Maintainer: @vk496 
Compile tested: ath79


neheb:
I placed the cmake files in a different directory. Based on my local installation of Fedora, I think /usr/lib/cmake is the correct directory.

PR #12753, opened 2020-07-08
modemmanager: remove unnecessary autoreconf 

nickberry17:
Signed-off-by: Nicholas Smith <nicholas.smith@telcoantennas.com.au>

Maintainer: me 
Compile tested: ath79, Telco T1, master
Run tested: ath79, Telco T1, master, checked basic functions
Description:
Autoreconf is unnecessary and is actually causing build errors on some systems for MM 1.14 and above.  Also, the correct glib-mkenums and gdbus-codegen are already being found correctly by the regular build system, so the setting the directory in Build/Prepare are not required.

More information: https://lists.freedesktop.org/archives/modemmanager-devel/2020-July/008005.html

### DIFF
--- a/net/modemmanager/Makefile
+++ b/net/modemmanager/Makefile
@@ -20,7 +20,6 @@ PKG_MAINTAINER:=Nicholas Smith <nicholas.smith@telcoantennas.com.au>
 PKG_LICENSE:=GPL-2.0-or-later
 PKG_LICENSE_FILES:=COPYING
 
-PKG_FIXUP:=autoreconf
 PKG_INSTALL:=1
 PKG_BUILD_PARALLEL:=1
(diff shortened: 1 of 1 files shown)


jow-:
A broken `autoreconf` usually is a very strong indication for a broken autoconf setup. It means that shoipped configure scripts are not reproducable. You should fix the autoreconf problems instead.

neheb:
Seems fine to me. Nothing is being patched currently.

micmac1:
:)

And it fails again on the build bots. Looks like comments of commit fa8ad6e are still valid.

neheb:
I didn't look at it carefully enough. This commit is indeed totally wrong. Probably @nickberry17 has those tools installed locally.

nickberry17:
Revert it

Issue #12680, opened 2020-07-01
[19.07] Some Python packages are not built due to ValueError: not enough values to unpack


BKPepe:
Maintainers: @jefferyto , @commodo 
Env: all
After update of Python3 in OpenWrt 19.07 some packages like zipp, cryptography, dateutil fails with the following error:

```
  File "/builder/shared-workdir/build/sdk/staging_dir/hostpkg/lib/python3.7/site-packages/pip/_vendor/packaging/tags.py", line 642, in _linux_platforms
    _, arch = linux.split("_", 1)
ValueError: not enough values to unpack (expected 2, got 1)
Makefile:35: recipe for target '/builder/shared-workdir/build/sdk/build_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/pypi/python-zipp-python3/zipp-0.6.0/.built' failed
```

Faillog: https://downloads.openwrt.org/releases/faillogs-19.07/arm_cortex-a9_vfpv3-d16/packages/python-zipp/python3/compile.txt

BKPepe:
I think that this could be related to this issue https://github.com/pypa/setuptools/issues/1934

BKPepe:
I downgraded setuptools in `staging_dir/hostpkg/lib/python3.7/site-packages` and installed `setuptools-scm` by this command: `build/staging_dir/hostpkg/bin/pip3.7 install setuptools-scm` and then run `make package/python-zipp/{cleancompile}` it worked.

jefferyto:
I think the actual issue is pypa/packaging#259. I'm looking into why we are setting `_PYTHON_HOST_PLATFORM=linux2` now, if we can do what Alpine is doing or if we should patch pip's copy of packaging to bypass this.

commodo:
> I think the actual issue is [pypa/packaging#259](https://github.com/pypa/packaging/issues/259). I'm looking into why we are setting `_PYTHON_HOST_PLATFORM=linux2` now, if we can do what Alpine is doing or if we should patch pip's copy of packaging to bypass this.

That setting is pretty old.
I think I just copied it from the old Python ~2.7.3 without giving it much thought.
I think it may have been put there by the first person who packaged Python for OpenWrt (before 2015). In the “oldpackages” repo, if that still exists anywhere these days.

jefferyto:
Well...

[...]

PR #12754, opened 2020-07-08
python3: Use default _PYTHON_HOST_PLATFORM

jefferyto:
Maintainer: me, @commodo 
Compile tested: armvirt-64, 2020-07-05 snapshot sdk
Run tested: armvirt-64 (qemu), 2020-07-05 snapshot

Description:
This lets the Python build process set `_PYTHON_HOST_PLATFORM` instead of forcing an explicit value.

Also:

* Save the target `_PYTHON_HOST_PLATFORM` value during `Build/InstallDev` for use when building target Python packages (in python3-package.mk).

* Use the (mostly) default `PYTHON_FOR_BUILD` value, instead patch configure to remove the platform triplet from the sysconfigdata file name.

* Remove the `CROSS_COMPILE=yes` make variable (there is no indication that this variable is necessary).

* Force host pip to build packages from source instead of downloading binary wheels.

  Previously, host pip can download universal (platform-independent) wheels but not platform-specific wheels, because of the custom `_PYTHON_HOST_PLATFORM` value. (Packages that do not have universal wheels would be compiled from source.)

  With a correct `_PYTHON_HOST_PLATFORM`, host pip can install platform-specific wheels as well. However, the pre-built shared object (.so) files in these wheels will have the host's platform triplet in their file names. When target Python packages are built (using the target's `_PYTHON_HOST_PLATFORM`), Python will not use these shared object files.

  By forcing host pip to build packages from source, the built shared object files will not have the platform triplet in their file names. (Host Python has been patched to remove the platform triplet from file names.) This allows these packages to be used when building target Python packages.

  (The net effect of this complete change is that platform-dependent packages will continue to be compiled from source, while platform-independent packages will now also be compiled from source.)


jefferyto:
This is #12755 but for master, because:

* Usually changes need to be added to master first, then back-ported to release branches
* Eventually pip in master will be upgraded to a newer version, and #12680 will affect master as well

I should note that removing `PYTHON_FOR_BUILD` is not strictly necessary, but looking at where it was originally added (2244ad9f56d2ea8d2965e2a1f42ccf5937eb21a2) it seems the original intent is to remove the platform triplet from the sysconfigdata file name, and patching it out is easier than keeping `PYTHON_FOR_BUILD` and also "fixing" `_PYTHON_HOST_PLATFORM` within it.

jefferyto:
Also, an unrelated note: I noticed there are lots of build-specific file paths in _sysconfigdata.py. We may need to look at it in the future to improve reproducibility of the Python package itself.

PR #12755, opened 2020-07-08
[openwrt-19.07] python3: Use default _PYTHON_HOST_PLATFORM

jefferyto:
Maintainer: me, @commodo 
Compile tested: armvirt-64, 19.07.3 sdk
Run tested: armvirt-64 (qemu), 19.07.3

Description:
This lets the Python build process set `_PYTHON_HOST_PLATFORM` instead of forcing an explicit value.

Also:

* Save the target `_PYTHON_HOST_PLATFORM` value during `Build/InstallDev` for use when building target Python packages (in python3-package.mk).

* Use the (mostly) default `PYTHON_FOR_BUILD` value, instead patch configure to remove the platform triplet from the sysconfigdata file name.

* Remove the `CROSS_COMPILE=yes` make variable (there is no indication that this variable is necessary).

* Force host pip to build packages from source instead of downloading binary wheels.

  Previously, host pip can download universal (platform-independent) wheels but not platform-specific wheels, because of the custom `_PYTHON_HOST_PLATFORM` value. (Packages that do not have universal wheels would be compiled from source.)

  With a correct `_PYTHON_HOST_PLATFORM`, host pip can install platform-specific wheels as well. However, the pre-built shared object (.so) files in these wheels will have the host's platform triplet in their file names. When target Python packages are built (using the target's `_PYTHON_HOST_PLATFORM`), Python will not use these shared object files.

  By forcing host pip to build packages from source, the built shared object files will not have the platform triplet in their file names. (Host Python has been patched to remove the platform triplet from file names.) This allows these packages to be used when building target Python packages.

  (The net effect of this complete change is that platform-dependent packages will continue to be compiled from source, while platform-independent packages will now also be compiled from source.)

Fixes https://github.com/openwrt/packages/issues/12680.

### DIFF
--- a/lang/python/python3-host.mk
+++ b/lang/python/python3-host.mk
@@ -50,8 +50,7 @@ define host_python3_settings
 	LDSHARED="$(HOSTCC) -shared" \
 	CFLAGS="$(HOST_CFLAGS)" \
 	CPPFLAGS="$(HOST_CPPFLAGS) -I$(HOST_PYTHON3_INC_DIR)" \
-	LDFLAGS="$(HOST_LDFLAGS) -lpython$(PYTHON3_VERSION) -Wl$(comma)-rpath=$(STAGING_DIR_HOSTPKG)/lib" \
-	_PYTHON_HOST_PLATFORM=linux2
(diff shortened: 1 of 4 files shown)


BKPepe:
If you would like, I can use Turris infrastructure to test it with the full build.

jefferyto:
@BKPepe I'm somewhat confident this will work :joy: but more testing is better (usually). If you do go ahead with a full build, you may want to pull in #12756 as well.

PR #11797, opened 2020-04-07
lang: python: flup: drop package

commodo:
Maintainer: me
Compile tested: n/a
Run tested: n/a

-----------------------------------------

Was mostly required for Seafile.
Now it isn't.
It's also pretty ancient.
   https://www.saddi.com/software/flup/

Page was last updated in 2005.

Signed-off-by: Alexandru Ardelean <ardeleanalex@gmail.com>

PR #12758, opened 2020-07-08
Bring back python-flup

BKPepe:
Maintainers: @commodo and @me
Compile tested: N/A
Run tested: n/a

Description:
as discussed here https://github.com/openwrt/packages/pull/11797#issuecomment-655411983

I was thinking if in the revert commit I shouldnt add me as maintainer, but there isnt anything wrong with 2 commits? I don't have strong opinion on this one.

### DIFF
--- a/lang/python/flup/Makefile
+++ b/lang/python/flup/Makefile
@@ -1,0 +1,40 @@
+#
+# Copyright (C) 2007-2016 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=flup
+PKG_VERSION:=1.0.3
+PKG_RELEASE:=4
+
+PYPI_NAME:=flup
+PKG_HASH:=5eb09f26eb0751f8380d8ac43d1dfb20e1d42eca0fa45ea9289fa532a79cd159
+
+PKG_MAINTAINER:=Alexandru Ardelean <ardeleanalex@gmail.com>, Josef Schlehofer <pepe.schlehofer@gmail.com>
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=PKG-INFO
+
+include ../pypi.mk
+include $(INCLUDE_DIR)/package.mk
+include ../python3-package.mk
+
+define Package/python3-flup
+  SECTION:=lang
+  CATEGORY:=Languages
+  SUBMENU:=Python
+  TITLE:=Random assortment of WSGI servers
+  URL:=https://www.saddi.com/software/flup/
+  DEPENDS:=+python3-light +python3-logging 
+endef
+
+define Package/python3-flup/description
+  Random assortment of WSGI servers.
+endef
+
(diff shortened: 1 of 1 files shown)


commodo (approved):

lang/python/flup/Makefile
```diff
@@ -0,0 +1,49 @@
...
+  SUBMENU:=Python
+  TITLE:=Random assortment of WSGI servers
+  URL:=https://www.saddi.com/software/flup/
+endef
+
+define Package/python3-flup
+$(call Package/flup/Default)
+  DEPENDS+= \
+      +PACKAGE_python3-flup:python3-light \
+      +PACKAGE_python3-flup:python3-logging
```
commodo, line 37: so a few quick things maybe: can you remove prefix `+PACKAGE_python3-flup:`

lang/python/flup/Makefile
```diff
@@ -14,7 +14,7 @@ PKG_RELEASE:=3
 PYPI_NAME:=$(PKG_NAME)
 PKG_HASH:=5eb09f26eb0751f8380d8ac43d1dfb20e1d42eca0fa45ea9289fa532a79cd159
 
-PKG_MAINTAINER:=Alexandru Ardelean <ardeleanalex@gmail.com>
+PKG_MAINTAINER:=Alexandru Ardelean <ardeleanalex@gmail.com>, Josef Schlehofer <pepe.schlehofer@gmail.com>
```
commodo, line 17: maybe bump PKG_RELEASE with this change?

commodo: i.e. in comit `python-flup: add me as co-maintainer `

lang/python/flup/Makefile
```diff
@@ -0,0 +1,49 @@
...
+
+PKG_MAINTAINER:=Alexandru Ardelean <ardeleanalex@gmail.com>
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=PKG-INFO
+
+include ../pypi.mk
+include $(INCLUDE_DIR)/package.mk
+include ../python3-package.mk
+
+define Package/flup/Default
```
commodo, line 25: maybe also do a quick merge of this Default into the main `define Package/python3-flup`?

commodo:
so this is fine from to merge as-is from my side;
only 1 comment may be a bit more important [about PKG_RELEASE]

rest we can do later;

BKPepe:
I thought that reverting it would be fine, my bad. I will do all the changes as requested. You are right that we should polish it before merging.

commodo:
> I thought that reverting it would be fine, my bad. I will do all the changes as requested. You are right that we should polish it before merging.

i probably removed it sometime after plenty of reworks & removing of Python2 variants happened

BKPepe:
Sorry took me some time during holidays. Now, it should be done. Would you please check it once again? I compile and run tested it on Turris Omnia, OpenWrt master.

commodo (approved):

BKPepe:
I think, now it is the right time to drop this. :) According to 
https://gitlab.nic.cz/turris/os/packages/-/merge_requests/1355 

cc: @shenek and @aleksan4eg

Issue #12750, opened 2020-07-08
make Error


snakwu:
identifier could be generated
onsgmls:/home/snakwu/openwrt-master/openwrt/build_dir/hostpkg/expat-2.2.9/doc/xmlwf.xml:97:51: entity was defined here
onsgmls:I: maximum number of errors (200) reached; change with -E option
Can't call method "value" on an undefined value at /usr/share/perl5/sgmlspl-specs/docbook2man-spec.pl line 250, <STDIN> line 1.
make[6]: *** [CMakeFiles/expat.dir/build.make:131: libexpat.so.1.6.11] Error 8
make[6]: Leaving directory '/home/snakwu/openwrt-master/openwrt/build_dir/hostpkg/expat-2.2.9'
make[5]: *** [CMakeFiles/Makefile2:109: CMakeFiles/expat.dir/all] Error 2
make[5]: Leaving directory '/home/snakwu/openwrt-master/openwrt/build_dir/hostpkg/expat-2.2.9'
make[4]: *** [Makefile:158: all] Error 2
make[4]: Leaving directory '/home/snakwu/openwrt-master/openwrt/build_dir/hostpkg/expat-2.2.9'
make[3]: *** [Makefile:57: /home/snakwu/openwrt-master/openwrt/build_dir/hostpkg/expat-2.2.9/.built] Error 2
make[3]: Leaving directory '/home/snakwu/openwrt-master/openwrt/feeds/packages/libs/expat'
time: package/feeds/packages/expat/host-compile#0.21#0.06#0.30
make[2]: *** [package/Makefile:111: package/feeds/packages/expat/host/compile] Error 2
make[2]: Leaving directory '/home/snakwu/openwrt-master/openwrt'
make[1]: *** [package/Makefile:107: /home/snakwu/openwrt-master/openwrt/staging_dir/target-x86_64_musl/stamp/.package_compile] Error 2
make[1]: Leaving directory '/home/snakwu/openwrt-master/openwrt'
make: *** [/home/snakwu/openwrt-master/openwrt/include/toplevel.mk:231: world] Error 2

snakwu:
dentifier could be generated
onsgmls:/home/snakwu/openwrt-master/openwrt/build_dir/hostpkg/expat-2.2.9/doc/xmlwf.xml:97:11: entity was defined here
[...]

PR #12759, opened 2020-07-08
Fix libexpat make error #12750

kongfl888:

Description:


kongfl888:
pull misstake....

PR #12762, opened 2020-07-08
ddns-scripts: Support Bearer Token for Cloudflare

ngrsdhbw:
Signed-off-by: Norman Gehrsitz <gehrsitz.norman@student.dhbw-kahlsruhe.de>

Maintainer: @neheb 
Compile tested: n/a
Run tested:  `armv7`, `Netgear R8000` ,`OpenWrt 19.07.1`

Description: This commit simplifies the ability to use API Tokens with restricted permissions for Cloudflare-v4. If the username equals "Bearer" the password is used as the Token. Since "Bearer" is not a valid email this should not cause any conflicts. Documenting this on the luci configuration page could be useful. Previously the Script for using a Bearer Token was simply duplicated and only modified in the authentication section.


jefferyto:
I'm not the maintainer, but it looks like API token support was added in #12277 (and not backported to 19.07, at least not yet).

TBH I like your change more than bf1d880c24023a5a260f7cd57564a425fd7b1e74 - I would prefer to see this change also  remove update_cloudflare_com_v4_token.sh, but again I am not the maintainer.

neheb:
Needs a PKG_RELEASE bump.

ngrsdhbw:
@jefferyto Thanks for finding this i did not notice it was already present on master when creating the pull request since i was testing on 19.07.
@neheb I rebased my changes on master and applied them as a overhaul to the existing implementation.

But i am not sure what the correct design is from a user experience standpoint. For this change to be useful people would have to know about the Bearer user which would need at least some form of documentation.
A better way to keep the two files approach would be keeping the main logic in one file but moving the setting of authorization headers in a separate function. The _token file could then simply source the base file but overwrite the authorization function.

jefferyto:
It would be good if after this is merged you can update the [DDNS Client](https://openwrt.org/docs/guide-user/services/ddns/client) page on the wiki (perhaps add a section under [Provider specifics](https://openwrt.org/docs/guide-user/services/ddns/client#provider_specifics)) documenting how to use bearer tokens with this change.

jefferyto:
BTW there is a merge conflict (probably with `PKG_RELEASE`), there have been some other changes lately.

PR #12773, opened 2020-07-10
xinetd: start service in foreground for procd

feckert:
Maintainer: @neheb the origin maintainer @jmccrohan is not active anymore so please have a look. If it is OK from your side I would also register as maintainer, because I use this package in my own setup.

Run tested: x86_64, APU3, OpenWRT master, latest master

Description:
Since we use now procd for xinetd, we have to start the service with the
option `-dontfork` to make procd happy.


neheb:
Seems correct. Feel free to add yourself.

PR #12779, opened 2020-07-10
strongswan: run through shellcheck

neheb:

Maintainer: @stintel


stintel:
The commit subject seems strange to me. Sure, you ran the scripts through shellcheck, but that on its own doesn't fix anything, does it? So a more appropriate commit subject would be "fix shellcheck warnings".

Also, did you runtime test it? This is a kind of software that if something breaks, it could cause people to be unable to access some remote devices, which really should be avoided at all cost, especially in times like this with many travel restrictions. So please, runtime test it and add this info to the PR.

PR #12782, opened 2020-07-10
mosquitto: switch to CMake

neheb:
Allows faster compilation. Also added PKG_BUILD_PARALLEL for faster
compilation.

Small Makefile cleanups.

Maintainer: @karlp 
Compile tested: ath79


karlp:
nak on this one, the cmake file is only provided for windows/mac, and the makefile is the only canonically supported one.

PR #12784, opened 2020-07-11
lighttpd: switch to meson

neheb:
Simplifies the Makefile and allows faster compilation with Ninja.

Added patch to fix libmariadb dependency.

Added extra modules.

Speed Before:

time make package/lighttpd/compile -j 12
________________________________________________________
Executed in   47.91 secs   fish           external
   usr time   41.83 secs  384.00 micros   41.83 secs
   sys time   10.79 secs   37.00 micros   10.79 secs

Speed After:

time make package/lighttpd/compile -j 12
________________________________________________________
Executed in   19.67 secs   fish           external
   usr time   42.79 secs  377.00 micros   42.79 secs
   sys time    8.56 secs   37.00 micros    8.56 secs

Tested with fish shell.

Maintainer: @flyn-org 
Compile tested: ath79

ping @BKPepe I know you guys use this in TurrisOS.

### DIFF
--- a/net/lighttpd/Makefile
+++ b/net/lighttpd/Makefile
@@ -20,16 +20,16 @@ PKG_LICENSE:=BSD-3-Clause
 PKG_LICENSE_FILES:=COPYING
 PKG_CPE_ID:=cpe:/a:lighttpd:lighttpd
 
-PKG_FIXUP:=autoreconf
 PKG_INSTALL:=1
+PKG_BUILD_DEPENDS:=meson/host
+PKG_CONFIG_DEPENDS:=CONFIG_LIGHTTPD_SSL $(patsubst %,CONFIG_PACKAGE_lighttpd-mod-%,$(REBUILD_MODULES))
 
 REBUILD_MODULES=authn_gssapi authn_ldap authn_mysql cml magnet mysql_vhost trigger_b4_dl webdav
-
-PKG_CONFIG_DEPENDS:=CONFIG_LIGHTTPD_SSL $(patsubst %,CONFIG_PACKAGE_lighttpd-mod-%,$(REBUILD_MODULES))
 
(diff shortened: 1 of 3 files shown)


neheb:
There are more modules that can be built. Not sure what to do with those.

neheb:
Added those extra modules. Removed pointless libpcre depends as the main lighttpd depends on it.

Left out memcached since it seems to conflict with libgdbm, maxminddb since it seems the meson script is broken with it, and the geoip plugin since there's no libgeoip in OpenWrt.

neheb:
Added maxminddb. Fixing meson was actually pretty easy.

flyn-org:
Wow---thank you! Recommend merge.

I am glad we got Meson in OpenWrt (originally, for the glib2-related stuff). I would have never guessed lighttpd would adopt it so quickly.

BKPepe:
> ping @BKPepe I know you guys use this in TurrisOS.

@neheb There is a small possibility that I can test it tomorrow, but if I don't make it, I would be able to get to it at least on Wednesday. Maybe @ja-pa would be faster.

neheb:
Added lighttpd-mod-vhostdb_dbi.

Used the upstream patches to clarify that they are now upstream backports.

ja-pa:
I just run tested it on Turris OS and it works ok (we use slightly changed [Makefile](https://gitlab.nic.cz/turris/turris-os-packages/-/blob/master/net/lighttpd/Makefile) )

list of tested packages:
* lighttpd
* lighttpd-mod-alias
* lighttpd-mod-cgi
* lighttpd-mod-fastcgi
* lighttpd-mod-openssl
* lighttpd-mod-proxy
* lighttpd-mod-redirect
* lighttpd-mod-setenv

PR #12785, opened 2020-07-11
cache-domains: Improved configure reliability

G-M0N3Y-2503:
Maintainer: @G-M0N3Y-2503

Description:
I've noticed that in deployment on some reboots it won't configure, this seems to be due to running before the system/network is ready.

* `cache-domains configure` exits with `1` on failure
* Retrying on failure every 30s for 3 attempts 
* Removing `/var/cache-domains` on failure so subsequent hotplug events can retry

### DIFF
--- a/utils/cache-domains/Makefile
+++ b/utils/cache-domains/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cache-domains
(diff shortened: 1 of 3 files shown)


neheb:
I see ==. Please also run that script through shellcheck.

G-M0N3Y-2503:
Thanks for the tip, wasn't aware of `shellcheck`

PR #7323, opened 2018-11-06
rng-tools: version 6.6 is now available

nwf:
Work around automake demanding README with a symlink.
Add new dependencies: libsysfs, libxml, curl, openssl.
  We still remove gcrypt and jitterentropy is not packaged.
Remove patch that no longer cleanly applies.
  If newlines in log messages are a problem, we should take it up with
  upstream.

Signed-Off-By: Nathaniel Wesley Filardo <nwfilardo@gmail.com>

Maintainer: me
Compile and run tested: kirkwood HEAD

PR #12788, opened 2020-07-11
rng-tools: bump to 6.10, require libopenssl

nwf:
Upstream now requires libopenssl unconditionally

Given the new requirement of `libopenssl`, it might be ideal to extend OpenWRT's urngd to support reading from files (in addition to jitterentropy) and either remove `rng-tools` or enable its advanced functionality (nistbeacon, pkcs11, rtl-sdr).

Maintainer: me
Compile tested: kirkwood OpenWRT HEAD
Runtime tested: kirkwood OpenWRT HEAD

### DIFF
--- a/utils/rng-tools/Makefile
+++ b/utils/rng-tools/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=rng-tools
-PKG_VERSION:=6.7
+PKG_VERSION:=6.10
 PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/nhorman/rng-tools
 PKG_SOURCE_VERSION:=v$(PKG_VERSION)
-PKG_MIRROR_HASH:=05cb68b8600025f362ea0875f5966b60f8195f91ed89b431996a48cd88b1e5b0
+PKG_MIRROR_HASH:=853341a3be05164616ba11a8cf983847477e02d6d43a259fc2352dbec262d649
 
 PKG_MAINTAINER:=Nathaniel Wesley Filardo <nwfilardo@gmail.com>
 PKG_LICENSE:=GPL-2.0-or-later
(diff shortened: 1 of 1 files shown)


neheb:
Need a signoff.

neheb:
I also assume this will not compile with GCC10. I see this commit upstream: https://github.com/nhorman/rng-tools/pull/94/commits/eec8ee57a34800abff5732991503a5100120fb8a

Use TARGET_CFLAGS += -fno-common to simulate failure with previous versions of GCC.

nwf:
We disable the nistbeacon module, so we shouldn't be including that second definition.

neheb:
WARNING: PKG_MIRROR_HASH does not match rng-tools-6.10.tar.xz hash 853341a3be05164616ba11a8cf983847477e02d6d43a259fc2352dbec262d649

neheb:
Actually, if you could, post the tarball that you have. The build system generating different tarballs is interesting to look at.

nwf:
Oh, huh.  I think I naively assumed that I could use the hash of the tar.gz from https://github.com/nhorman/rng-tools/releases/tag/v6.10 .  I guess it wants a .xz and so it's making its own rather than grabbing the gz from github.  Sorry, will fix.

neheb:
For that, you'd need to convert to using codeload.

hnyman:
> Upstream now requires libopenssl unconditionally

Why is libopenssl actually required?  For which purpose? Did you look into specifics?

We discussed that with the 6.6 version bump, and then it could still be disabled.
https://github.com/openwrt/packages/pull/7323#issuecomment-436306288

Intuitively I see no reason for using openssl in managing kernel entropy pool.

nwf:
I've not looked under the covers save to note that `configure.ac` now says `OpenSSL`'s required:
https://github.com/nhorman/rng-tools/blob/02f34d69b04492726cc51b8dd587af1534e38417/configure.ac#L134


That said, I'd very much like to give up maintainership of `rng-tools` once https://patchwork.ozlabs.org/project/openwrt/patch/20200720105304.1090741-1-nwfilardo@gmail.com/ and friends land in `urngd`, since I don't use the more advanced features of `rng-tools` (e.g., nist beacon) and those would be all that's left for motivation.

hnyman:
Apparently openssl has replaced libgcrypt, but the not-so-funny part is that we have earlier disbaled the libgcrypt here... So that functionality has not been used here. But apparently openssl is now forced to be included in any case.

Please fix the PKG_MIRROR_HASH, so that the circleci test passes.

PR #12798, opened 2020-07-12
unbound: improve startup and dhcp script robustness

EricLuehrsen:
Maintainer: me
Tested: Linksys WRT3200ACM
Description:
- prevent rapid overlap in DHCP script updates
- check and allow localhost forwards with specific applications
- add option for rate limiting inbound queries
- change UCI list to table format with Unbound conf references

enhances: #12581
resolves: OpenWrt [Forum 66354](https://forum.openwrt.org/t/issues-with-unbound-and-odhcp-setup/66354/19)


neheb:
Why remove those quotes in [] ? Shellcheck complains about it.

quotes are not necessary with [[]] but this is not bash.

EricLuehrsen:
Pedantic quotes on well defined variables should be avoided. Quotes on math should be avoided. They should be used for user string and file selection checks (-n, -z, -f, ...) where odd characters, globs, and splitting occur. SC2086 is a green check instead of orange or red.

EricLuehrsen:
Pedantic quotes on well defined variables should be avoided. Quotes on math should be avoided. They should be used for user string and file selection checks (-n, -z, -f, ...) where odd characters, globs, and splitting occur. It in the same readability and maintainability noise as unnecessary `${symbol}` when `$symbol` is sufficient and correct, situation dependent. SC2086 is a green check instead of orange or red.

PR #12799, opened 2020-07-12
openfortivpn: use standard 'peerdns' config option on netifd script

aaronjg:

updated to use the standard 'peerdns' config option as @jow- pointed out here

https://github.com/openwrt/luci/pull/4175

Maintainer: @lucize
Compile tested: 19.07.3-ipq806x
Run tested: 19.07.3-ipq806x

### DIFF
--- a/net/openfortivpn/Makefile
+++ b/net/openfortivpn/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openfortivpn
 PKG_VERSION:=1.14.1
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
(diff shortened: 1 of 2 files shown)


lucize:
LGTM

aaronjg:
It looks like `--set-dns` should always be set to 0. When it is set to true, it updates /etc/resolv.conf. The dns is better handled by `--pppd-use-peerdns` which then uses the `/lib/netifd/ppp-up` script to set the dns.

aaronjg:
@lucize I just rebased off of master to resolve the merge conflict. I gave it a new version bump since I had though this would have been merged first, so we are skipping pkg release 1.14.1-3. 

I also found a few other things like this where I was doing things in a way that was not standard to OpenWRT, so I'll make a new PR for those.

lucize:
I understand, LGTM

PR #12804, opened 2020-07-13
[RFC] dockerd: add iptables access control for docker0

feckert:
Maintainer: @G-M0N3Y-2503 

Description:

With this change you can determine who is allowed to access the docker0
interface. For this to take effect a new section access_control with the
option iface must be added to the uci dockerd configuration.

config access_control
	option iface "br-lan"

This is now maintained in the [luci-app-dockerman](https://github.com/openwrt/luci/blob/master/applications/luci-app-dockerman/root/etc/init.d/dockerman). To be able to
discard the init script in the LuCI feed the same functionality must be
included and edited in the dockerd init script.

We have already add uci handling to make it luci-app-dockerman easy to change the needed values.
But it is still to be adjusted by removeing the dockerd-config in the init script and update the cbi handling sources of luci-app-dockerman do use /etc/config/dockerd and not /etc/config/dockerman

@lisaac could you also please verify if this works for you?

### DIFF
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -43,10 +43,39 @@ process_config() {
 	json_dump > "$DOCKERD_CONF"
 }
 
+process_ac_allowed() {
+	local cfg="$1"
+
+	local iface
+
+	config_get iface "$cfg" iface
+	iptables -A DOCKER-ACCESS-CONTROL -i "$iface" -o docker0 -j RETURN
+
+}
+
+process_ac() {
+	[ -f /etc/config/dockerd ] || {
+		return 0
+	}
+
+	# Add access control
+	iptables -N DOCKER-ACCESS-CONTROL
+
+	config_foreach process_ac_allowed access_control
+
+	iptables -A DOCKER-ACCESS-CONTROL -m conntrack --ctstate ESTABLISHED,RELATED -o docker0 -j RETURN
+	iptables -A DOCKER-ACCESS-CONTROL -m conntrack --ctstate NEW,INVALID -o docker0 -j DROP
+	iptables -A DOCKER-ACCESS-CONTROL -j RETURN
+
+	iptables -N DOCKER-USER
+	iptables -I DOCKER-USER -j DOCKER-ACCESS-CONTROL
+}
+
 start_service() {
 	local nofile=$(cat /proc/sys/fs/nr_open)
 
 	process_config
+	process_ac
 
 	procd_open_instance
 	procd_set_param stderr 1
@@ -76,11 +105,13 @@ ip4tables_remove_filter() {
 	iptables -t filter -F DOCKER-ISOLATION-STAGE-1
 	iptables -t filter -F DOCKER-ISOLATION-STAGE-2
 	iptables -t filter -F DOCKER-USER
+	iptables -t filter -F DOCKER-ACCESS-CONTROL
 
 	iptables -t filter -X DOCKER
 	iptables -t filter -X DOCKER-ISOLATION-STAGE-1
 	iptables -t filter -X DOCKER-ISOLATION-STAGE-2
 	iptables -t filter -X DOCKER-USER
+	iptables -t filter -X DOCKER-ACCESS-CONTROL
 }
 
 ip4tables_remove() {


utils/docker-ce/files/dockerd.init
```diff
@@ -43,10 +43,39 @@ process_config() {
...
+
+	iptables -N DOCKER-USER
+	iptables -I DOCKER-USER -j DOCKER-ACCESS-CONTROL
+}
+
 start_service() {
 	local nofile=$(cat /proc/sys/fs/nr_open)
 
 	process_config
+	process_ac
```
G-M0N3Y-2503, line 78: Logically, AC is defined in the configuration, so maybe we should call this function in `process_config`?
We could also rely on the `if [` at line 17 if we call it somewhere  like line 33 then remove the duplicate at line 57 that way.

utils/docker-ce/files/dockerd.init
```diff
@@ -43,10 +43,39 @@ process_config() {
...
+		return 0
+	}
+
+	# Add access control
+	iptables -N DOCKER-ACCESS-CONTROL
+
+	config_foreach process_ac_allowed access_control
+
+	iptables -A DOCKER-ACCESS-CONTROL -m conntrack --ctstate ESTABLISHED,RELATED -o docker0 -j RETURN
+	iptables -A DOCKER-ACCESS-CONTROL -m conntrack --ctstate NEW,INVALID -o docker0 -j DROP
```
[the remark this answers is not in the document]
feckert: Thanks for the link. But if i read this correct,
I have to add for each interface (ext_if) an iptables rule.
`iptables -I DOCKER-USER -i ext_if ! -s 192.168.1.1 -j DROP`
On My system I have the following interfaces _eth0_, _eth1_, _eth2_ and _wwan_.
That are only the real interface and not the virtual interface like pppoe, wireguard and openvpen.

utils/docker-ce/files/dockerd.init
feckert: Sounds good. Will do update my P/R.

feckert:
@G-M0N3Y-2503 An additional question is whether we should give the ac interface its own section like in this P/R, or whether we should add it as a list element in the global section, like we originally did in [luci-app-dockerman](https://github.com/openwrt/luci/blob/master/applications/luci-app-dockerman/root/etc/config/dockerman#L10) config.

lisaac:
It's ok for me, once it merged, we could remove this part in dockerman.

feckert:
@lisaac please have a look to me comment in https://github.com/openwrt/packages/pull/12804#issuecomment-658655920

lisaac:
@feckert 
On my view, we should give AC an additional section, since may be we have more other ac in it.

Such as: https://github.com/lisaac/luci-app-dockerman/pull/54, I'm still thinking about how to solve this elegantly, maybe @jmarcet had given the answer at https://github.com/openwrt/packages/pull/12556#issuecomment-648866347, in this way, we can easily control it by firewall rule. But it's a little bit complex. So what you think about it? @feckert

utils/docker-ce/files/dockerd.init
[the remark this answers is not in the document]
G-M0N3Y-2503: Oh, I think I was misreading the link, I didn't realise that they said to do it for all interfaces. This way makes sense, thanks for the clarification.

G-M0N3Y-2503:
I agree with @lisaac, if well need a couple of settings for AC a new section would be good. Also it may make the configuration file a little easier to read as the option names won't need `ac` in them if they are under a `access-controll` section

utils/docker-ce/files/dockerd.init
```diff
@@ -43,10 +43,39 @@ process_config() {
 	json_dump > "$DOCKERD_CONF"
 }
 
+process_ac_allowed() {
+	local cfg="$1"
+
+	local iface
+
+	config_get iface "$cfg" iface
```
G-M0N3Y-2503, line 51: Do we want to name this option `allowed_iface`?
Also, We may want to add a commented out example of this to the default config file.

feckert:
> Such as: [lisaac/luci-app-dockerman#54](https://github.com/lisaac/luci-app-dockerman/pull/54), I'm still thinking about how to solve this elegantly, maybe @jmarcet had given the answer at [#12556 (comment)](https://github.com/openwrt/packages/pull/12556#issuecomment-648866347), in this way, we can easily control it by firewall rule. But it's a little bit complex. Or just add a button "Could allowed interface can visit the containers directly"?
> So what you think about it? @feckert


I have read through this and I find this a very good approach. Perhaps we should continue to work in this direction, even if it is a little more complex. But for that it is implemented in a very good way and in the docker-ce package.

So I think we have to do the following steps:
1. Integrate the suggestion from @jmarcet into docker-ce -> close this RFC and open a new one.
2. Remove ac handling from luci-app-dockerman
3. Remove uci handlng from luci-app-dockerman
4. Because of removing **uci** and **ac** from luci-app-dockerman we could remove entirely `/etc/init.d/dockerman` and` /etc/config/dockerman`

Because of upcoming next openwrt Release we should go on this. Otherwise we have to take care of upgrade script, if we don't change it till the next Openwrt stable release.

feckert:
This is replaced by https://github.com/openwrt/packages/pull/12897/commits

peterwillcn:
Error: Network device is not present
Screen Shot 2020-08-29 at 10 55 43 AM

```
config interface 'docker'
	option ifname 'docker0'
	option proto 'static'
	option auto '0'
	option ipaddr '172.18.0.1'
	option metric '20'

config device 'docker0'
	option type 'bridge'
	option name 'docker0'
	list ifname 'docker0'
```

PR #12810, opened 2020-07-14
mwan3: fix some shellcheck warnings

feckert:
Maintainer: me

Description:
With this pullrequest I want to fix some shellcheck warnings.
I think more will follow!

### DIFF
--- a/net/mwan3/files/lib/mwan3/mwan3.sh
+++ b/net/mwan3/files/lib/mwan3/mwan3.sh
@@ -1146,8 +1146,6 @@ mwan3_report_policies_v6()
 
 mwan3_report_connected_v4()
 {
-	local address
-
 	if [ -n "$($IPT4 -S mwan3_connected 2> /dev/null)" ]; then
 		$IPS -o save list mwan3_connected_v4 | grep add | cut -d " " -f 3
 	fi
@@ -1155,8 +1153,6 @@ mwan3_report_connected_v4()
 
 mwan3_report_connected_v6()
 {
-	local address
-
 	if [ -n "$($IPT6 -S mwan3_connected 2> /dev/null)" ]; then
 		$IPS -o save list mwan3_connected_v6 | grep add | cut -d " " -f 3
 	fi
--- a/net/mwan3/files/lib/mwan3/mwan3.sh
+++ b/net/mwan3/files/lib/mwan3/mwan3.sh
@@ -192,7 +192,7 @@ mwan3_get_iface_id()
 	mwan3_get_id()
 	{
 		let _iface_count++
-		[ "$1" == "$_iface" ] && _tmp=$_iface_count
+		[ "$1" = "$_iface" ] && _tmp=$_iface_count
 	}
 	config_foreach mwan3_get_id interface
 	export "$1=$_tmp"
@@ -396,7 +396,7 @@ mwan3_create_iface_iptables()
 
 	[ -n "$id" ] || return 0
 
-	if [ "$family" == "ipv4" ]; then
+	if [ "$family" = "ipv4" ]; then
 		$IPS -! create mwan3_connected list:set
 
 		if ! $IPT4 -S mwan3_ifaces_in &> /dev/null; then
@@ -428,7 +428,7 @@ mwan3_create_iface_iptables()
 			-j mwan3_iface_in_$1
 	fi
 
-	if [ "$family" == "ipv6" ]; then
+	if [ "$family" = "ipv6" ]; then
 		$IPS -! create mwan3_connected_v6 hash:net family inet6
 
 		if ! $IPT6 -S mwan3_ifaces_in &> /dev/null; then
@@ -462,7 +462,7 @@ mwan3_delete_iface_iptables()
 {
 	config_get family $1 family ipv4
 
-	if [ "$family" == "ipv4" ]; then
+	if [ "$family" = "ipv4" ]; then
 
 		$IPT4 -D mwan3_ifaces_in \
 			-m mark --mark 0x0/$MMX_MASK \
@@ -471,7 +471,7 @@ mwan3_delete_iface_iptables()
 		$IPT4 -X mwan3_iface_in_$1 &> /dev/null
 	fi
 
-	if [ "$family" == "ipv6" ]; then
+	if [ "$family" = "ipv6" ]; then
 
 		$IPT6 -D mwan3_ifaces_in \
 			-m mark --mark 0x0/$MMX_MASK \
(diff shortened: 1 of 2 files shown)


net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -929,92 +917,92 @@ mwan3_set_user_iptables_rule()
...
-				     ${dest_ip:+-d} $dest_ip\
+				     -p "$proto" \
+				     ${src_ip:+-s} "$src_ip" \
+				     ${src_dev:+-i} "$src_dev" \
+				     ${dest_ip:+-d} "$dest_ip" \
 				     $ipset \
-				     ${src_port:+-m} ${src_port:+multiport} ${src_port:+--sports} $src_port \
-				     ${dest_port:+-m} ${dest_port:+multiport} ${dest_port:+--dports} $dest_port \
+				     ${src_port:+-m} ${src_port:+multiport} ${src_port:+--sports} "$src_port" \
+				     ${dest_port:+-m} ${dest_port:+multiport} ${dest_port:+--dports} "$dest_port" \
```
aaronjg, line 990: This appears to break my configuration. If the src_port, etc,  is not set you end up with extra blank arguments that iptables does not like

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -929,92 +917,92 @@ mwan3_set_user_iptables_rule()
...
+			     ${src_dev:+-i} "$src_dev" \
+			     ${dest_ip:+-d} "$dest_ip" \
 			     $ipset \
-			     ${src_port:+-m} ${src_port:+multiport} ${src_port:+--sports} $src_port \
-			     ${dest_port:+-m} ${dest_port:+multiport} ${dest_port:+--dports} $dest_port \
+			     ${src_port:+-m} ${src_port:+multiport} ${src_port:+--sports} "$src_port" \
+			     ${dest_port:+-m} ${dest_port:+multiport} ${dest_port:+--dports} "$dest_port" \
 			     -m mark --mark 0/$MMX_MASK \
-			     -j $policy &> /dev/null
+			     -j "$policy" &> /dev/null
```
aaronjg, line 1005: policy can be multiple arguments, eg set on line 922, so quotes are not correct here

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -391,148 +391,136 @@ mwan3_create_iface_iptables()
...
-			route_args="$route_args metric $metric"
-		fi
+		network_get_metric metric "$1"
 
-		$IP4 route flush table $id
-		$IP4 route add table $id default $route_args dev $2
+		$IP4 route flush table "$id"
+		$IP4 route add table "$id" default \
+				${via:+via} "$via" \
+				${metric:+metric} "$metric" \
```
aaronjg, line 505: quotes are not needed here and cause problems if metric or via is not set

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -46,7 +46,7 @@ mwan3_rtmon_ipv4()
 		idx=$((idx+1))
 		tid=$idx
 		[ "$(uci get mwan3.@interface[$((idx-1))].family)" = "ipv4" ] && {
-			tbl=$($IP4 route list table $tid)
+			tbl=$($IP4 route list table $tid 2>/dev/null)
```
aaronjg, line 49: What's the reasoning here? I've found mwan3 already has a tendency to send too many errors to /dev/null, which can make debugging difficult. When does ip route list output to stderr? Should it really be suppressed here?

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -391,148 +391,136 @@ mwan3_create_iface_iptables()
...
 		fi
 
-		$IPT4 -F mwan3_iface_in_$1
-		$IPT4 -A mwan3_iface_in_$1 \
-			-i $2 \
+		$IPT4 -F "mwan3_iface_in_$1"
+		$IPT4 -A "mwan3_iface_in_$1" \
+			-i "$2" \
 			-m set --match-set mwan3_connected src \
 			-m mark --mark 0x0/$MMX_MASK \
```
aaronjg, line 414: Why is $MMX_MASK quoted above but not here?

feckert: I haven't quoted all the quotation that shellcheck recommends.
I didn't want to change too much.

net/mwan3/files/lib/mwan3/mwan3.sh
feckert: You're right that's hard to debug, but I always get the following output `dump terminated` coming from that line, if i do an `mwan3 restart` on the command line and my wwan interface is not connected. Maybe we have to consider if we don't write all output to a log. At least the one from stderr.

net/mwan3/files/lib/mwan3/mwan3.sh
feckert: Thanks for pointing that out, I'm gonna fix that right now.
So we can merge my pullrequest and then you could [rebase](https://github.com/openwrt/packages/pull/12814) yours.

net/mwan3/files/lib/mwan3/mwan3.sh
feckert: Thanks for pointing that out, I'm gonna fix that right now.
So we can merge my pullrequest and then you could [rebase](https://github.com/openwrt/packages/pull/12814) yours.

net/mwan3/files/lib/mwan3/mwan3.sh
feckert: Thanks for pointing that out, I'm gonna fix that right now.
So we can merge my pullrequest and then you could [rebase](https://github.com/openwrt/packages/pull/12814) yours.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: Any idea why you are getting the `dump terminated` message? Sounds like something else is wrong, and it would be better to fix that than to mask the error here.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: Okay - it probably makes sense to at least be consistent with what is quoted. Although, it probably doesn't matter, if somehow a space got into the MMX_MASK one of the previous iptables command would have failed.

net/mwan3/files/lib/mwan3/mwan3.sh
feckert: This is because if the table is empty this will output.
On my setup it's right.
The interface has not yet been get connected  by the netifd, so if you restart mwan3 with `mwan3 restart' the table will be empty.

feckert: The problem is also that if you start this on the cmd you get this message every interval of rtmon! That is boring

PR #12833, opened 2020-07-16
golang: Add GO_PKG_TAGS for Go packages

jefferyto:
Maintainer: me, @aparcar (for syncthing)
Compile tested: armvirt-64, 2020-07-12 snapshot sdk
Run tested: none

Description:
This adds a new Makefile variable, `GO_PKG_TAGS`, for Go packages. When set, the value is passed as the parameter of the `-tags` option for `go install`.

This also updates syncthing to use this variable.

### DIFF
--- a/lang/golang/golang-package.mk
+++ b/lang/golang/golang-package.mk
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2018 Jeffery To
+# Copyright (C) 2018-2020 Jeffery To
 #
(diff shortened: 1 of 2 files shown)


aparcar:
LGTM

Does it need a release bump? Shouldn't it be ideally the same checksum?

jefferyto:
There should be no difference in compiled output, but bumping `PKG_RELEASE` is safer than not bumping it :joy:

PR #12849, opened 2020-07-17
docker-ce: add reload handling

feckert:
Maintainer: @G-M0N3Y-2503 

Description:

If the uci configuration is changed send dockerd a SIGHUP to reload the
generated daemon.json file with the new configuration.

This is also needed to switch luci-app-dockerman to use the docker-ce uci handling.
We do not want to start on every change the docker daemon. 
So send them a SIGHUP to reload their configuration.

### DIFF
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -59,6 +59,15 @@ start_service() {
 	procd_close_instance
 }
 
+reload_service() {
+	process_config
+	procd_send_signal dockerd
+}
+
+service_triggers() {
+	procd_add_reload_trigger 'dockerd'
+}
+
 ip4tables_remove_nat() {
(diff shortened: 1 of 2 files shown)


utils/docker-ce/files/dockerd.init
```diff
@@ -59,6 +59,16 @@ start_service() {
 	procd_close_instance
 }
 
+reload_service() {
+	process_config
+	#SIGHUP is used by dockerd to do init.d reload
```
G-M0N3Y-2503, line 64: I'm not sure what this comment means.
The link below suggests that `SIGHUP` will cause the daemon to reload the configuration.
https://docs.docker.com/config/daemon/#enable-debugging

feckert: I copied this from the keepaliveed package.
But you're right.
I'll delete it.

PR #12857, opened 2020-07-17
rsyslog: update to 8.2006.0

neheb:
Fixes compilation with GCC10.

Maintainer: none
Compile tested: ath79


neheb:
Looks like rsyslog.com is down.

PR #12860, opened 2020-07-18
bfdd: fix compilation with GCC 10

neheb:

Maintainer: @lucize 
Compile tested: ath79


lucize:
LGTM

PR #12863, opened 2020-07-18
flashrom: switch to building with meson

neheb:
Added a patch to fix meson compilation.

Reworked flashrom.mk to work with meson.

Several minor Makefile cleanups.

Maintainer: @Noltari 
Compile tested: ath79


neheb:
This is more or less ready.

The i386 patch is concerning. I think meson should be fixed.

neheb:
Removed i386 patch as that was fixed in https://github.com/openwrt/packages/pull/12870

PR #12864, opened 2020-07-18
openfortivpn: add host up test back

aaronjg:
Maintainer: @lucize
Compile tested: netifd changes only 
Run tested: 19.07.3-ipq806x

Description:
PR: https://github.com/openwrt/packages/pull/12692 removed some functionality from the package that was critical for my setup. The openfortivpn package used to ping the server to see if it was up before trying to connect, then sleep for a time if it was unavailable. This prevented continually trying to connect to the down server.  I have added it back in. It is off by default with a new 'ping_check' flag to turn the functionality on.

Please be careful resolving merge conflicts with https://github.com/openwrt/packages/pull/12799 when merging.

cc  @qweaszxcdf

### DIFF
--- a/net/openfortivpn/Makefile
+++ b/net/openfortivpn/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openfortivpn
 PKG_VERSION:=1.14.1
-PKG_RELEASE:=2
+PKG_RELEASE:=4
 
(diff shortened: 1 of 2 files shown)


aaronjg:
Actually, this check can be done with curl - which is a better indication if the vpn server is up nad reachable since the https request will be part of the fortivpn connection process anyway. I will add code for that.

aaronjg:
curl may not be installed, so making this a configurable option to use curl, ping or do no check at all

lucize:
LGTM

PR #12865, opened 2020-07-18
prometheus-node-exporter-lua: Add feature to listen on all local addre…

margau:
Description:
This PR introduces a new feature: When the 'listen_all'-option is true, the exporter listens on all local addresses, both IPv4 and IPv6.
The feature uses an TCP6-Socket with IPV6_V6ONLY disabled, therefore also accepting IPv4 connections.
If 'listen_all' is true, 'listen_interface' and 'listen_ipv6' are ignored.

Signed-off-by: Marvin Gaube <dev@marvingaube.de>

Maintainer: @champtar
Compile tested: -
Run tested: x86/64, OpenWrt 19.07-SNAPSHOT r11098+11-9cafcbe0bd, new features working, old behavior with 'listen_all' disabled is stable

### DIFF
--- a/utils/prometheus-node-exporter-lua/files/etc/config/prometheus-node-exporter-lua
+++ b/utils/prometheus-node-exporter-lua/files/etc/config/prometheus-node-exporter-lua
@@ -1,4 +1,5 @@
 config prometheus-node-exporter-lua 'main'
+	option listen_all '0'
 	option listen_interface 'loopback'
 	option listen_ipv6 '0'
 	option listen_port '9100'
--- a/utils/prometheus-node-exporter-lua/files/etc/init.d/prometheus-node-exporter-lua
+++ b/utils/prometheus-node-exporter-lua/files/etc/init.d/prometheus-node-exporter-lua
@@ -11,12 +11,13 @@ _log() {
 start_service() {
 	. /lib/functions/network.sh
 
-	local interface ipv6 port bind
+	local interface ipv6 port bind all
 
 	config_load prometheus-node-exporter-lua.main
 	config_get interface "main" listen_interface "loopback"
 	config_get_bool ipv6 "main" listen_ipv6 0
 	config_get port "main" listen_port 9100
+	config_get_bool all "main" listen_all 0
 
 	if [ "$interface" = "*" ]; then
 		[ "$ipv6" = 1 ] && bind="::" || bind="0.0.0.0"
@@ -38,6 +39,9 @@ start_service() {
 	procd_set_param command /usr/bin/prometheus-node-exporter-lua
 	procd_append_param command --bind ${bind}
 	procd_append_param command --port ${port}
+	if [ "$all" = 1 ]; then
+		procd_append_param command -a
+	fi
 
 	procd_set_param stdout 1
 	procd_set_param stderr 1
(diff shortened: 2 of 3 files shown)


dhewg:
Wouldn't it make more sense to make `listen_ipv6` a tristate?
Instead of `0` or `1`, make it `0` or `1` or `both`, so you can combine 4+6 on the chosen `listen_interfaces` - which would work with `*` too then to archive your goal.

margau:
With the current architecture, the combined listener only works for "any" adress ('::'). 
listen_interfaces internally uses an IP address, which would mean we need separate sockets for IPv4 and IPv6. 
That is no easy change, because it'd need another architecture to have more than one server socket (at least one for each family).
Because of this, I've chosen the way with another option, and dual stack only available for all interfaces (at least for now).

dhewg:
Hm, right... The required changes shouldn't be too bad, but I get your point.
It's just unfortunate to expose another option, the limiting factor atm is just an implementation detail.

How about making it a tristate anyway and error out if `both` is given but `listen_interface` isn't `*`? That way the code could be adapted later to support multiple interfaces without breaking any config.

margau:
The Code ist fully backward compatible now. 
I personally would prefer two booleans over an tristate, and the config with `listen_all` is probably easier to understand? Are there other suggestions about this?

dhewg:
Well, it does fit into the scheme of the two existing knobs. A third one circumvents the other two. That's not just confusing for the user but weird for automation (hardcoded special case for e.g. luci).

And I meant forward compatibility, which a third option most likely is not.

margau:
I personally have problems with switching an existing knob from boolean to non-boolean. Could you post a "matrix" (table) stating what should happen at which (existing) option configuration?

NoctivagusObitus:
Hi
Just stumbling across this and I like to comment from a naive users perspective. I had some real hassle with the mutually exclusiveness of the two options. What sure is mostly because of my custom set up and therefore my fault... 
After reading up a bit on the handling of IPv4 and IPv6 dual stack setup it looked to me like the intended way of handling such a situation would be listening on IPv6 because it is preferred and having ipv4 be some kind of fallback.

Anyway I am really not competent in any way to comment on actual implementation. I just like to suggest to have the new behaviour proposed by @margau  be the default since that is what at least I would expect.

Apart from that. Thanks for nice pice of software and for pushing it further. Makes for some nice little dashboards :)

champtar:
I have to think more about it, but as OpenWrt is going to ship an ssl lib by default, maybe we should do a breaking change and use a proper http server (the same as luci) instead of hand coded http 1.0 supporting one request at a time

NoctivagusObitus:
Just in case you are interested I had thought kind of in that direction... For my use case it is all about scarping the metrics from an external Prometheus in the internet without having the metrics exposed to the internet.
https://forum.openwrt.org/t/node-exporter-with-uhttpd/64661

To give you more inside: I am currently scraping 5 routers (number growing) on 2 sides where one is a small business. And the insight you can get from a simple Grafana dashboard is quite awesome :)

Again just to give some feedback. I would appreciate a breaking change as proposed by @champtar.

neheb:
What's the status here?

champtar:
After a long delay ... #18360 should fix your issue

PR #12870, opened 2020-07-18
meson: fix arch mapping

neheb:
meson uses a different architecture mapping than CONFIG_ARCH does.

Fix the problematic ones and fall back to CONFIG_ARCH.

Maintainer: @dhewg 
Compile tested: i386

https://mesonbuild.com/Reference-tables.html#cpu-families


neheb:
Forgot about powerpc64.

dhewg:
Does this fix an actual error?

Going by `target/Config.in` you missed: aarch64_be arceb i686 sh4eb
And then there's sh3 and sh3eb which isn't in meson's list at all...

neheb:
Yes it does. See the i386 patch here: https://github.com/openwrt/packages/pull/12863

That patch is needed to avoid issues. It becomes useless with this PR.

Those other ones are not used at all. Neither is armeb and powerpc64 currently. Nevertheless, I can add them to the list.

PR #12872, opened 2020-07-18
spice: convert to meson

neheb:
Faster compilation.

Removed autotools patches.

Maintainer: @yousong 
Compile tested: ath79


neheb:
Odd. I don't get this error locally.

neheb:
This might be more trouble than it's worth...

@dhewg do you know how to pass MESON_ARGS to subprojects?

dhewg:
Nope, I didn't use subproject() myself yet. Passing all of MESON_ARGS sounds wrong though, there might be main project specific args in there. What's the issue, which args are missing?

neheb:
In the Makefile I have an empty \ with the arguments after meant for the sub project.

dhewg:
I think you need support of the parent project to pass those on. Either patch those in or expose a parent arg to set the sub args:
https://gitlab.freedesktop.org/spice/spice/-/blob/master/meson.build#L59

neheb:
@dhewg according to IRC, the format is -Dsubproject:option=x. Tested locally. Seems to work.

neheb:
Abandoning this. It requires multiple upstream backports. Maybe next version...

dhewg:
> @dhewg according to IRC, the format is -Dsubproject:option=x. Tested locally. Seems to work.

Huh, good to know, didn't know that existed.

PR #12879, opened 2020-07-20
mwan3: update to version 2.8.11

### LINKED ISSUE #12807: mwan3: switching to procd init script causes status/overview, system/startup loading failure
Maintainer: @feckert 
Environment: snapshot mwan3_2.8.8-1
Description:
switching to procd init script causes status/overview, system/startup loading failure

openwrt/luci#4228

feckert:
Maintainer: me

Description:

- [Commit](https://github.com/openwrt/packages/commit/058a2b6f3057d849cbe4f2c4fda258d32d6da2fd) fixes #12807 
- Add online time and uptime to detail output
- Remove pending lock process on mwan3 stop
- [Commit](https://github.com/openwrt/packages/commit/bcd13ba95cd5b96d04e490b576faab54404cbf69) fixes output `uci: Entry not found`

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.8.10
+PKG_VERSION:=2.8.11
 PKG_RELEASE:=1
 PKG_MAINTAINER:=Florian Eckert <fe@dev.tdt.de>
 PKG_LICENSE:=GPL-2.0
--- a/net/mwan3/files/etc/init.d/mwan3
+++ b/net/mwan3/files/etc/init.d/mwan3
@@ -10,17 +10,22 @@ boot() {
 	rc_procd start_service
 }
 
+# FIXME
+# fd 1000 is an inherited lock file descriptor for preventing concurrent
+# init script executions. Close it here to prevent the mwan3 daemon from
+# inheriting it further to avoid holding the lock indefinitely.
+
 reload_service() {
-	/usr/sbin/mwan3 restart
+	/usr/sbin/mwan3 restart 1000>&-
 }
 
 start_service() {
 	[ -n "${mwan3_boot}" ] && return 0
-	/usr/sbin/mwan3 start
+	/usr/sbin/mwan3 start 1000>&-
 }
 
 stop_service() {
-	/usr/sbin/mwan3 stop
+	/usr/sbin/mwan3 stop 1000>&-
 }
 
 service_triggers() {
--- a/net/mwan3/files/lib/mwan3/mwan3.sh
+++ b/net/mwan3/files/lib/mwan3/mwan3.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+. /usr/share/libubox/jshn.sh
+
 IP4="ip -4"
 IP6="ip -6"
 IPS="ipset"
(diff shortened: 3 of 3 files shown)


feckert:
@aaronjg @neheb I would suggest to backport this also to openwrt-19.07 to fix the procd [issue](https://github.com/openwrt/packages/issues/12868). 

The Problem is that we have to pre backport additional the  [P/R ](https://github.com/openwrt/packages/pull/12814) and [P/R](https://github.com/openwrt/packages/pull/12810) too, so that the patches can be applied correctly. Do you see any problems, because the branch is a stable release. 

The other option is that we only backport the procd fix. The problem there is then, we do not have the same version string. This makes the maintainability more difficult.

I would prefer if we backport the missing P/R´s and so have the same mwan3 version in openwrt-19.07 (last stable) and master branch.

aaronjg:
Hi @feckert, I have been testing this with your latest commits on a 19.03 build.  I did notice one issue with locking: If you run `mwan3 restart` when one of the monitoring scripts has triggered a reload, it can leave the routing table and firewall rules in a non working state. 

I think the issue comes from the sleep statement introduced here https://github.com/openwrt/packages/pull/12879/commits/cde2a77ed3b3b0df9e693e121dccdc97ef163156  in the mwan3_lock_clean command, so you can get in a scenario like this:

1) Process A secures the lock and starts doing work
2) Process B attempts to get the lock but is blocked by A
3) Process A finishes and begins cleaning up the locks
4) Process B starts and secures the locks
5) Process A finishes clearing the locks and deletes process B's lock, and kills process B in the process 

Looking at the code, there was already a race condition before the commit, but it was much less likely to be hit because there was no sleep condition, and it was not as bad because it wasn't killing the other processes. It likely would have needed three restarts running at once to trigger, but now it seems very easy to trigger the race condition.

So the lock file should be left in place after mwan3 is stopped, rather than being deleted. Or if it must be deleted, it needs to be done much more carefully

https://stackoverflow.com/questions/17708885/flock-removing-locked-file-without-race-condition

feckert:
@aaronjg thanks for the hint! I hadn't noticed that.

Pleas try my following patch if this solve your issue.

[0001-mwan3-fix-race-condition-on-mwan3-restart.txt](https://github.com/openwrt/packages/files/4958486/0001-mwan3-fix-race-condition-on-mwan3-restart.txt)


I also noticed the following output `sh: 0: unknown operand` if I do an restart. With the `set -x` I saw that this has to do with the NO_IPV6 check. So I changed it into

[0002-mwan3-fix-ipv6-command-check-test.txt](https://github.com/openwrt/packages/files/4958487/0002-mwan3-fix-ipv6-command-check-test.txt)


And the message is gone.

Could you please try this two patches on top of the current mwan3 version?
On my system and setup everything works.

aaronjg:
Can you describe a bit more when you were getting that error with the the NO_IPV6 test? I am not able to reproduce with a simple `mwan3 restart`, and I want to test this on a router compiled without IPv6 support to make sure the test still evaluates correctly in that case.

We can still have a race condition with that patch. 
If you are running in process A: `mwan3 stop`, and then in B: `mwan3 start`, it is possible for A to obtain the lock between B's  `mwan3_unlock "command" "mwan3"` and `mwan3_lock_clean` in the stop function.

Then B will start messing with process A in the `mwan3_lock_clean.`

I'm guessing you saw some sort of issue that was fixed with adding the kill commands to the lock clean? What was that? Maybe there is another way to solve it.  

Also, I don't think we should be deleting the lock file at all, since that is an easy way for a race condition to occur (http://www.guido-flohr.net/never-delete-your-pid-file/). However, that was already broken so not something that needs to be fixed in the backport.  It is possible to safely delete the lock files, but more challenging (https://stackoverflow.com/questions/17708885/flock-removing-locked-file-without-race-condition)

feckert:
> Can you describe a bit more when you were getting that error with the the NO_IPV6 test? I am not able to reproduce with a simple `mwan3 restart`, and I want to test this on a router compiled without IPv6 support to make sure the test still evaluates correctly in that case.

Funny, now I can't even reproduce it after restarting the router! Let's just say that was a one-time thing.

> We can still have a race condition with that patch.
> If you are running in process A: `mwan3 stop`, and then in B: `mwan3 start`, it is possible for A to obtain the lock between B's `mwan3_unlock "command" "mwan3"` and `mwan3_lock_clean` in the stop function.
> 
> Then B will start messing with process A in the `mwan3_lock_clean.`

I think that's unlikely, but while we're at it, let's get it right

> I'm guessing you saw some sort of issue that was fixed with adding the kill commands to the lock clean? What was that? Maybe there is another way to solve it.

I have seen several lock process. But that was before I changed the locks(see patch). Maybe this is not necessary anymore.

> 
> Also, I don't think we should be deleting the lock file at all, since that is an easy way for a race condition to occur (http://www.guido-flohr.net/never-delete-your-pid-file/). However, that was already broken so not something that needs to be fixed in the backport. It is possible to safely delete the lock files, but more challenging (https://stackoverflow.com/questions/17708885/flock-removing-locked-file-without-race-condition)

Can you try if you don't call the function `mwan3_lock_clean` in mwan3 stop call and apply the new lock patch, if this will solve your problems?

feckert:
@aaronjg I would suggest that we undo the change in [mwan3_lock_clean](https://github.com/openwrt/packages/pull/12879/commits/cde2a77ed3b3b0df9e693e121dccdc97ef163156) so that it behaves the same as before and is harder to provoke the race condition.

So we should live with this race condition first and only apply the new lock patch improvement from my [comment](https://github.com/openwrt/packages/pull/12879#issuecomment-662292607).

I want to have the mwan3 in a working condition as soon as possible.
Not that we suddenly have a new 19.07 release and the mwan3 is broken.
I would like to make a pullrequest with the changes. 
After that mwan3 in master and openwrt-19.07 is on the same version, we have time to fix the race condition on mwan3 command execution

aaronjg:
That sounds fine to me. I have been testing with mwan3_lock_clean removed entirely and I have been unable to provoke the race condition. I have not yet done a formal analysis to determine if a race condition is possible though.  

So either undoing the kill commands in mwan3_lock_clean or removing it entirely would be a good fix for the master branch. As maintainer, it's your call to what you think is safer.

I would also suggest moving 
`mwan3_flush_conntrack "$INTERFACE" "$ACTION"'`
to before the lock is released in `net/mwan3/files/etc/hotplug.d/iface/16-mwan3`

PR #12897, opened 2020-07-21
docker-ce: add default bridge to openwrt uci backend

feckert:
Maintainer: @G-M0N3Y-2503 
Run tested: x86_64, APU3, OpenWrt master, startup test

Description:

- I have added the **bip** docker option so we could customize the default bridge. This was descriped by @jmarcet in his [comment](https://github.com/openwrt/packages/pull/12556#issuecomment-648866347).
- I just added the default bridge handling, with the new init.d targets uciadd and ucidel.
- I also replaced the SIGHUB on config change with a start and stop of the whole dockerd service, so the new bip config option get applied.

### DIFF
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -122,6 +122,9 @@ process_config() {
 	json_add_array "registry-mirrors"
 	config_list_foreach globals registry_mirror json_add_array_string
 	json_close_array
+	json_add_array "hosts"
+	config_list_foreach globals hosts json_add_array_string
+	json_close_array
 
 	mkdir -p /tmp/dockerd
 	json_dump > "$DOCKERD_CONF"
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -3,14 +3,96 @@
 USE_PROCD=1
 START=25
 
+EXTRA_COMMANDS="uciadd ucidel"
+EXTRA_HELP="\
+	uciadd  Add default bridge configuration to network and firewall uci config
+	ucidel  Delete default bridge configuration from network and firewall uci config"
+
 DOCKERD_CONF="/tmp/dockerd/daemon.json"
+
+uci_quiet() {
+	uci -q ${@} >/dev/null
+}
 
 json_add_array_string() {
 	json_add_string "" "$1"
 }
 
+uciupdate() {
+	local net="$1"
+
+	uci -q get network.docker >/dev/null || {
+		logger -t "dockerd-init" -p warn "No network uci config section for docker default bridge (docker0) found"
+		return
+	}
+
+	[ -z "$net" ] && {
+		logger -t "dockerd-init" -p notice "Removing network uci config options for docker default bridge (docker0)"
+		uci_quiet delete network.docker.netmask
+		uci_quiet delete network.docker.ipaddr
+		uci_quiet commit network
+		return
+	}
+
+	eval "$(ipcalc.sh "$net")"
+	logger -t "dockerd-init" -p notice "Updating network uci config option \"$net\" for docker default bridge (docker0)"
+	uci_quiet set network.docker.netmask="$NETMASK"
+	uci_quiet set network.docker.ipaddr="$IP"
+	uci_quiet commit network
+}
+
+uciadd() {
+	/etc/init.d/dockerd running && {
+		echo "Please stop dockerd service first"
+		exit 0
+	}
+
+	# Add network interface
+	if ! uci -q get network.docker >/dev/null; then
+		logger -t "dockerd-init" -p notice "Adding docker default bridge to network uci config (docker0)"
+		uci_quiet add network interface
+		uci_quiet rename network.@interface[-1]="docker"
+		uci_quiet set network.docker.ifname="docker0"
+		uci_quiet set network.docker.proto="static"
+		uci_quiet set network.docker.auto="0"
+		uci_quiet commit network
+	fi
+
+	# Add firewall zone
+	if ! uci -q get firewall.docker >/dev/null; then
+		logger -t "dockerd-init" -p notice "Adding docker default bridge firewall zone (docker0)"
+		uci_quiet add firewall zone
+		uci_quiet rename firewall.@zone[-1]="docker"
+		uci_quiet set firewall.docker.network="docker"
+		uci_quiet set firewall.docker.input="REJECT"
+		uci_quiet set firewall.docker.output="ACCEPT"
+		uci_quiet set firewall.docker.forward="REJECT"
+		uci_quiet set firewall.docker.name="docker"
+		uci_quiet commit firewall
+	fi
+
+	reload_config
+}
+
+ucidel() {
+	/etc/init.d/dockerd running && {
+		echo "Please stop dockerd service first"
+		exit 0
+	}
+
+	logger -t "dockerd-init" -p notice "Deleting docker default bridge network from network uci config (docker0)"
+	uci_quiet delete network.docker
+	uci_quiet commit network
+
+	logger -t "dockerd-init" -p notice "Deleting docker default bridge firewall zone from firewall uci config (docker0)"
+	uci_quiet delete firewall.docker
+	uci_quiet commit firewall
+
+	reload_config
+}
+
 process_config() {
-	local alt_config_file data_root log_level
+	local alt_config_file data_root log_level bip
 
 	rm -f "$DOCKERD_CONF"
 
@@ -30,17 +112,21 @@ process_config() {
 
 	config_get data_root globals data_root "/opt/docker/"
 	config_get log_level globals log_level "warn"
+	config_get bip globals bip ""
 
 	. /usr/share/libubox/jshn.sh
 	json_init
 	json_add_string "data-root" "$data_root"
 	json_add_string "log-level" "$log_level"
+	[ -z "$bip" ] || json_add_string "bip" "$bip"
 	json_add_array "registry-mirrors"
 	config_list_foreach globals registry_mirror json_add_array_string
 	json_close_array
 
 	mkdir -p /tmp/dockerd
 	json_dump > "$DOCKERD_CONF"
+
+	uciupdate "$bip"
 }
 
 start_service() {
@@ -77,19 +163,25 @@ ip4tables_remove_nat() {
 }
 
 ip4tables_remove_filter() {
-	iptables -t filter -D FORWARD -j DOCKER-USER
+	# Chain DOCKER-USER is only present,
+	# if bip option is NOT set, so >/dev/null 2>&1
+	iptables -t filter -D FORWARD -j DOCKER-USER >/dev/null 2>&1
 	iptables -t filter -D FORWARD -j DOCKER-ISOLATION-STAGE-1
 	iptables -t filter -D FORWARD -o docker0 -j DOCKER
 
 	iptables -t filter -F DOCKER
 	iptables -t filter -F DOCKER-ISOLATION-STAGE-1
 	iptables -t filter -F DOCKER-ISOLATION-STAGE-2
-	iptables -t filter -F DOCKER-USER
+	# Chain DOCKER-USER is only present,
+	# if bip option is NOT set, so >/dev/null 2>&1
+	iptables -t filter -F DOCKER-USER >/dev/null 2>&1
 
 	iptables -t filter -X DOCKER
 	iptables -t filter -X DOCKER-ISOLATION-STAGE-1
 	iptables -t filter -X DOCKER-ISOLATION-STAGE-2
-	iptables -t filter -X DOCKER-USER
+	# Chain DOCKER-USER is only present,
+	# if bip option is NOT set, so >/dev/null 2>&1
+	iptables -t filter -X DOCKER-USER >/dev/null 2>&1
 }
 
 ip4tables_remove() {
@@ -98,5 +190,8 @@ ip4tables_remove() {
 }
 
 stop_service() {
-	ip4tables_remove
+	if /etc/init.d/dockerd running; then
+		service_stop "/usr/bin/dockerd"
+		ip4tables_remove
+	fi
 }
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -18,6 +18,11 @@ json_add_array_string() {
 	json_add_string "" "$1"
 }
 
+boot() {
+	uciadd
+	rc_procd start_service
+}
+
 uciupdate() {
 	local net="$1"
 
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -120,7 +120,7 @@ process_config() {
 	json_add_string "log-level" "$log_level"
 	[ -z "$bip" ] || json_add_string "bip" "$bip"
 	json_add_array "registry-mirrors"
-	config_list_foreach globals registry_mirror json_add_array_string
+	config_list_foreach globals registry_mirrors json_add_array_string
 	json_close_array
 	json_add_array "hosts"
 	config_list_foreach globals hosts json_add_array_string
--- a/utils/docker-ce/Makefile
+++ b/utils/docker-ce/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=docker-ce
 PKG_VERSION:=19.03.12
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=components/cli/LICENSE components/engine/LICENSE
 
--- a/utils/docker-ce/files/etc/config/dockerd
+++ b/utils/docker-ce/files/etc/config/dockerd
@@ -7,5 +7,5 @@ config globals 'globals'
 	# If the bip option is changed, dockerd must be restarted.
 	# A service reload is not enough.
 	option bip "172.18.01./24"
-#	list registry_mirror "https://<my-docker-mirror-host>"
-#	list registry_mirror "https://hub.docker.com"
+#	list registry_mirrors "https://<my-docker-mirror-host>"
+#	list registry_mirrors "https://hub.docker.com"
--- a/utils/docker-ce/files/etc/config/dockerd
+++ b/utils/docker-ce/files/etc/config/dockerd
@@ -3,5 +3,6 @@ config globals 'globals'
 #	option alt_config_file "/etc/docker/daemon.json"
 	option data_root "/opt/docker/"
 	option log_level "warn"
+	option hosts "unix://var/run/docker.sock"
 #	list registry_mirror "https://<my-docker-mirror-host>"
 #	list registry_mirror "https://hub.docker.com"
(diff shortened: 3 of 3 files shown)


utils/docker-ce/files/dockerd.init
```diff
@@ -30,17 +87,21 @@ process_config() {
 
 	config_get data_root globals data_root "/opt/docker/"
 	config_get log_level globals log_level "warn"
+	config_get bip globals bip "172.17.0.1/16"
```
G-M0N3Y-2503, line 90: We need a default because we add the bridge to the UCI right?
I'm just asking because we don't have a way of having of adding config options without a default yet.
we'd need something like `[ -n "$option" ] && json_add_string "option" "$option"` or something nicer so that we aren't maintaining defaults over the ones build into the daemon.

utils/docker-ce/files/dockerd.init
```diff
@@ -9,8 +9,65 @@ json_add_array_string() {
...
 
+docker_bridge_add_config() {
+	local net="$1"
+
+	eval "$(ipcalc.sh "$net")"
+
+	# Add network interface
+	if ! uci -q get network.docker >/dev/null; then
+		uci -q batch >/dev/null <<EOF
+add network interface
```
G-M0N3Y-2503, line 20: I'm not really a big fan of the "here document" redirection, it tends to make the code harder to read in my opinion.  
maybe we could just have a wrapper function that silences `uci` maybe
```sh
uci_quiet() {
    uci -q ${@} >/dev/null
}
uci_quiet add network interface
uci_quiet rename network.@interface[-1]=docker
...
```

utils/docker-ce/files/dockerd.init
```diff
@@ -9,8 +9,65 @@ json_add_array_string() {
...
+set firewall.docker.forward='REJECT'
+set firewall.docker.name='docker'
+commit firewall
+EOF
+	logger -t "dockerd" -p notice "Firewall zone docker added"
+
+	reload_config
+}
+
+docker_bridge_remove_config() {
```
G-M0N3Y-2503, line 56: do we want to be able to manually adjust `network.docker` or `firewall.docker` outside of this setup?

utils/docker-ce/files/dockerd.init
feckert: I think so too. I'll fix it.

utils/docker-ce/files/dockerd.init
feckert: I do not understand you exactly? Sorry
But the code can be described as follows:
If no config option **bip** is set in die uci config (/etc/config/dockerd) then use `172.17.0.1/16` as default bip value.

utils/docker-ce/files/dockerd.init
feckert: I have discussed recently with @lisaac about this topic to add and remove docker networks to the network and firewall uci configuration.
The result was, that only the init script should add the default docker0 to the network and firewall configuration.
If we add additional network configuration this is done by luci-app-dockerman and not by the init script.

utils/docker-ce/files/dockerd.init
G-M0N3Y-2503: I mean `daemon.json` has a lot of options and if we have to figure out the default for each we one we add, then also maintain its the same as the default built into the daemon. e.g making sure `{ "bip": "172.17.0.1/16" }` in the `daemon.json` has the same effect as leaving it out of the `daemon.json`. it'll quickly add up to a lot of maintenance, unless I'm mistaken.

But from what you said in the other comments, it sounds like we need to know the default or otherwise so we can set up `docker0`.
This is just a consideration for options other than these is what I'm thinking.

feckert: > But from what you said in the other comments, it sounds like we need to know the default or otherwise so we can set up `docker0`.
> This is just a consideration for options other than these is what I'm thinking.

That's right. Also, it seems that if I set the option `bip`, the firewall chain `DOCKER-USER` is not created.
I agree with you about the other option. We should not maintain the default.

utils/docker-ce/files/dockerd.init
feckert: I have also thought about whether we should make the creation of the network and firewall configuration configurable?

After thinking about this I do not like my implementation to delete the configuration and commit theme on service stop. We could introduce a separate init target next to stop/start/restart that would delete the configuration. Maybe clean or purge. The new target should only be executed if f the dockerd service is not running.

G-M0N3Y-2503: I'm not sure if I understood all that. 
do you mean the following?
1. we would only add the network to the UCI if it doesn't exist already and just update `network.docker.netmask` and `network.docker.ipaddr` if `bip` changes.
2.  we would only add the firewall to the UCI if it doesn't exist already
3. add a cleanup target that is only manually called to remove the firewall and network

That sounds like a good way to be flexible so anyone can have exactly what they want.

feckert: Yeah, that's what I meant.
Gonna update my pullrequest.
Thanks for your feedback.

feckert:
@G-M0N3Y-2503 @jmarcet 
I have now updated the pullrequest as I imagine it and as far as I have come before weekend.
But I have still some issues.
I think nevertheless we need a hotplug script, to update the MAC.
If I change the IP (bip) in the uci config the daemon.json gets update but the docker0 gets not changed!
Could it be that the porcd hook does not work?

@G-M0N3Y-2503 I wonder if it is not possible to create the device in the netifd so that we already have a docker0 interface on boot or on dockerd start (docker will setup the uci in network and firewall).
Or does the dockerd have a problem with an already configured and present interface?

utils/docker-ce/files/dockerd.init
```diff
@@ -30,17 +93,22 @@ process_config() {
 
 	config_get data_root globals data_root "/opt/docker/"
 	config_get log_level globals log_level "warn"
+	config_get_bool bip_enable globals bip_enable "1"
+	[ "$bip_enable" = "1" ] && config_get bip globals bip "172.17.0.1/24"
```
G-M0N3Y-2503, line 97: Wouldn't the following do the same?
```
config_get bip globals bip ""
```

G-M0N3Y-2503:
> If I change the IP (bip) in the uci config the daemon.json gets update but the docker0 gets not changed!
> Could it be that the porcd hook does not work?

The procd hook only does a reload, doesn't it? the following link suggests that a full restart has to be done to change the default bridge.
https://docs.docker.com/network/bridge/#configure-the-default-bridge-network

utils/docker-ce/files/dockerd.init
feckert: Yeah, we can do that. But then I would write the default value for the bip option in the uci.

utils/docker-ce/files/dockerd.init
```diff
@@ -30,17 +93,22 @@ process_config() {
...
 	json_add_string "log-level" "$log_level"
+	[ -z "$bip" ] || json_add_string "bip" "$bip"
 	json_add_array "registry-mirrors"
 	config_list_foreach globals registry_mirror json_add_array_string
 	json_close_array
 
 	mkdir -p /tmp/dockerd
 	json_dump > "$DOCKERD_CONF"
+
+	[ -z "$bip" ] || ucisetup "$bip"
```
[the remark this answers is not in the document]
feckert: I have now moved this to the function

utils/docker-ce/files/dockerd.init
```diff
@@ -3,14 +3,77 @@
...
+	# Add firewall zone
+	if ! uci -q get firewall.docker >/dev/null; then
+		logger -t "dockerd" -p notice "Add firewall zone docker config"
+		uci_quiet add firewall zone
+		uci_quiet rename firewall.@zone[-1]="docker"
+	fi
+
+	logger -t "dockerd" -p notice "Update firewall zone docker config"
+	uci_quiet set firewall.docker.network="docker"
+	uci_quiet set firewall.docker.input="REJECT"
```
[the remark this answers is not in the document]
feckert: done

utils/docker-ce/files/dockerd.init
```diff
@@ -3,14 +3,77 @@
...
+	if ! uci -q get network.docker >/dev/null; then
+		logger -t "dockerd" -p notice "Add network interface docker config"
+		uci_quiet add network interface
+		uci_quiet rename network.@interface[-1]="docker"
+	fi
+
+	logger -t "dockerd" -p notice "Update network interface docker config"
+	uci_quiet set network.docker.ifname="docker0"
+	uci_quiet set network.docker.proto="static"
+	uci_quiet set network.docker.netmask="$NETMASK"
```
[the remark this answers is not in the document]
feckert: done

feckert:
@G-M0N3Y-2503 @jmarcet  can you watch it again?

feckert:
>   I thought we wanted to add the UCI config if it didn't exist on startup so that it would be set for everyone at least initially.

I once had a discussion with @jow-, which is about automatically adding configuration with scripts that the user has not set is difficult. The problem here is that the user might have already set configuration and it would be saved automatically with the script config changes. But the user did not want that.
To solve this we could add a boot target here that writes the configuration during boot.
The user definitely has no configuration staged!

```
boot() {
        uciadd
        rc_procd start_service
}
```

>   It looks like we've tried this way because of the required restart if bridge settings are changed. But I don't think we should have every reload be a restart just so that the bridge settings are propagated. If we don't want to do something more complex where we manage the state, I think that logging and putting a comment in the config that bridge settings require a restart would be fine.

I wasn't sure about that either. But I like your suggestion to add a comment here that the service has to be restarted if you change the option bip.

G-M0N3Y-2503:
> The problem here is that the user might have already set configuration and it would be saved automatically with the script config changes. But the user did not want that.

I don't think we have a problem with this. 
We only add `network.docker` and `firewall.docker` if it hasn't been created yet.
That means the user could create `network.docker` and `firewall.docker` themselves and it would only be changed if the config if `bip` has been set and even then they could just remove it from the config. 
Unless I'm missing something, it should be safe to add to `start`.

feckert:
I'm not sure what the best solution in this case is either.
I just wanted to point out that if someone has already changed something in the network configuration and it is not yet committed, it will be committed at docker start.
The docker init script changes the network configuration and saves the other settings in the network configuration as well.
I'm not sure we want that.

I suggest we start out that way, and if it turns out we can always change it.

G-M0N3Y-2503:
Hmmmm, I didn't realize the commits would commit the whole network/firewall, but I appreciate the information.

Otherwise, this PR looks good to me, thanks for the work!

peterwillcn:
But, i get can't work.
```
Sat Aug  8 19:09:02 2020 daemon.err dockerd[2561]: time="2020-08-08T11:09:02.033225252Z" level=warning msg="Your kernel does not support cgroup blkio weight"
Sat Aug  8 19:09:02 2020 daemon.err dockerd[2561]: time="2020-08-08T11:09:02.033533292Z" level=warning msg="Your kernel does not support cgroup blkio weight_device"
Sat Aug  8 19:09:03 2020 daemon.err dockerd[2561]: failed to start daemon: Devices cgroup isn't mounted
```

G-M0N3Y-2503:
> But, i get can't work.
> 
> ```
> Sat Aug  8 19:09:02 2020 daemon.err dockerd[2561]: time="2020-08-08T11:09:02.033225252Z" level=warning msg="Your kernel does not support cgroup blkio weight"
> Sat Aug  8 19:09:02 2020 daemon.err dockerd[2561]: time="2020-08-08T11:09:02.033533292Z" level=warning msg="Your kernel does not support cgroup blkio weight_device"
> Sat Aug  8 19:09:03 2020 daemon.err dockerd[2561]: failed to start daemon: Devices cgroup isn't mounted
> ```

That issue seems to be unrelated to this PR, it looks to be https://github.com/openwrt/packages/issues/13052 instead

peterwillcn:
Intrusive modification to solve the bridge? Simple things become more and more complicated, leading to a great decrease in stability ？

peterwillcn:
> CONFIG_KERNEL_CGROUP_DEVICE=y
> CONFIG_KERNEL_CGROUP_FREEZER=y
> CONFIG_KERNEL_CGROUP_NET_CLASSID=y
> CONFIG_KERNEL_CGROUP_NET_PRIO=y
> CONFIG_KERNEL_NET_CLS_CGROUP=y

> I'm not sure what the best solution in this case is either.
> I just wanted to point out that if someone has already changed something in the network configuration and it is not yet committed, it will be committed at docker start.
> The docker init script changes the network configuration and saves the other settings in the network configuration as well.
> I'm not sure we want that.
> 
> I suggest we start out that way, and if it turns out we can always change it.

Intrusively modifying network firewalls is not a good design

peterwillcn:
Error: Network device is not present
Screen Shot 2020-08-29 at 10 57 03 AM


```
config interface 'docker'
	option ifname 'docker0'
	option proto 'static'
	option auto '0'
	option ipaddr '172.18.0.1'
	option metric '20'

config device 'docker0'
	option type 'bridge'
	option name 'docker0'
	list ifname 'docker0'
```

feckert:
When docker starts this for the first time, the interface docker and the device docker0 is created in the configuration under `/etc/config/network.` Also the correct interface (docker0) will be created in the system when docker starts. It seems to me that the docker service is not running.

peterwillcn:
> @peterwillcn When docker starts this for the first time, the interface docker and the device docker0 is created in the configuration under `/etc/config/network.` Also the correct interface (docker0) will be created in the system when docker starts. It seems to me that the docker service is not running.

Yes, you are right,
What causes this? 
```
Tue Sep  1 00:33:40 2020 daemon.err dockerd[2990]: failed to start daemon: Error initializing network controller: Error creating default "bridge" network: Pool overlaps with other one on this address space
Tue Sep  1 00:34:04 2020 user.notice dockerd-init: Removing network uci config options for docker default bridge (docker0)
Tue Sep  1 00:34:05 2020 daemon.err dockerd[3156]: time="2020-08-31T16:34:05.627224515Z" level=warning msg="failed to load plugin io.containerd.snapshotter.v1.btrfs" error="path /opt/docker/containerd/daemon/io.containerd.snapshotter.v1.btrfs must be a btrfs filesystem to be used with the btrfs snapshotter"
Tue Sep  1 00:34:05 2020 daemon.err modprobe: failed to find a module named aufs
Tue Sep  1 00:34:05 2020 daemon.err dockerd[3156]: time="2020-08-31T16:34:05.643189910Z" level=warning msg="failed to load plugin io.containerd.snapshotter.v1.aufs" error="modprobe aufs failed: "": exit status 255"
Tue Sep  1 00:34:05 2020 daemon.err dockerd[3156]: time="2020-08-31T16:34:05.644921607Z" level=warning msg="could not use snapshotter btrfs in metadata plugin" error="path /opt/docker/containerd/daemon/io.containerd.snapshotter.v1.btrfs must be a btrfs filesystem to be used with the btrfs snapshotter"
Tue Sep  1 00:34:05 2020 daemon.err dockerd[3156]: time="2020-08-31T16:34:05.645033630Z" level=warning msg="could not use snapshotter aufs in metadata plugin" error="modprobe aufs failed: "": exit status 255"
Tue Sep  1 00:34:05 2020 daemon.err dockerd[3156]: time="2020-08-31T16:34:05.645089642Z" level=warning msg="could not use snapshotter zfs in metadata plugin" error="path /opt/docker/containerd/daemon/io.containerd.snapshotter.v1.zfs must be a zfs filesystem to be used with the zfs snapshotter: skip plugin"
Tue Sep  1 00:34:05 2020 daemon.err dockerd[3156]: time="2020-08-31T16:34:05.761457583Z" level=warning msg="Your kernel does not support cgroup blkio weight"
Tue Sep  1 00:34:05 2020 daemon.err dockerd[3156]: time="2020-08-31T16:34:05.761777317Z" level=warning msg="Your kernel does not support cgroup blkio weight_device"
Tue Sep  1 00:34:06 2020 daemon.info modprobe: nf_conntrack is already loaded
Tue Sep  1 00:34:06 2020 daemon.info modprobe: nf_conntrack_netlink is already loaded
```

I temporarily started from modifying /etc/config/dockerd to comment #option BIp "172.18.0.1/24"

utils/docker-ce/files/dockerd.init
```diff
@@ -3,14 +3,101 @@
...
+		uci_quiet commit network
+	fi
+
+	# Add firewall zone
+	if ! uci -q get firewall.docker >/dev/null; then
+		logger -t "dockerd-init" -p notice "Adding docker default bridge firewall zone (docker0)"
+		uci_quiet add firewall zone
+		uci_quiet rename firewall.@zone[-1]="docker"
+		uci_quiet set firewall.docker.network="docker"
+		uci_quiet set firewall.docker.input="REJECT"
```
G-M0N3Y-2503, line 72: @feckert, I've been trying to have [docker use the local DNS server by default](https://github.com/openwrt/packages/commit/c21cfa44d44807803535c0e35a78a1540e4546e0) but I've noticed the traffic is being blocked on the input zone. 
It seems to me now that we have the safety of the `blocked_interfaces` preventing someone accidentally opening a port on the `wan`, as a default. 
We should be able to treat `docker0` as any other local interface and not need a firewall rule by default, or perhaps treat it like another `lan` zone.
I'd be interested to hear your thoughts or if you recall anything about the original intention of this.

G-M0N3Y-2503: I have a PR draft for this https://github.com/openwrt/packages/pull/12897#discussion_r579615391

feckert: @G-M0N3Y-2503 I wanted the docker0 interface to be in its own firewall zone. Until now, it was possible for all interfaces, including the wan interface, to access the docker containers without a firewall controller. Until now, there have been two methods to control the docker interface handling.
* With the option **blocked_interfaces** ( no integration into the fw3 )
* And then other via **/etc/init.d/docker uciadd \<interface\>**

I always use the second one via uciadd. This integrates my docker interfaces/firewwallzone and thus the containers into the OpenWrt firewall handling. With REJECT I also wanted to control access from LAN to the Conteiner and forwarding should also be prevented. If someone wants to access your container, they have to open the port in the firewall the container is listening on. I therefore did not want to put the docker interface in the lan zone. It may be that not everyone should access the container. I wanted to think of it more as a different network.

G-M0N3Y-2503: In researching what an expected default might be I found that docker already has [Integration with Firewalld](https://docs.docker.com/network/iptables/#integration-with-firewalld). It also appears that the zone is [set to `ACCEPT` by default](https://github.com/moby/libnetwork/blob/fa125a3512ee0f6187721c88582bf8c4378bd4d7/iptables/firewalld.go#L204). This default appears to be more like [my proposed changed](https://github.com/openwrt/packages/pull/14807), if I'm interpreting [the setting](https://firewalld.org/documentation/zone/options.html) correctly.

I tend to think that most people will primarily be running containers for services either for their LAN or WAN running them in OpenWrt. But I also agree that we should be mindful of the WAN given that it's atypical of other places docker might run. Given that there is this precedent set upstream, I'm inclined to think that we should be more allowing for people to quickly set up a service for their LAN.

With the [proposed changes](https://github.com/openwrt/packages/pull/14807) I've tested and containers have no WAN access in or out with fw3 or docker with iptables just on device and LAN access.

feckert: @G-M0N3Y-2503 Ok from my side. I can also live with that :+1:

PR #12899, opened 2020-07-22
openfortivpn: use a more 'OpenWRT' paradigm throughout the code

aaronjg:
Maintainer: @lucize 
Compile tested: no compiled changes
Run tested: 19.07.3

Description:
In the initial implementation, I had done many things in non-standard and potentially broken ways, such as calling 'uci' to get configuration parameters (which would cause it to operate on uncommitted configuration changes), or 'ifstatus' to get interface status. This PR cleans a lot of that up to use the built in openwrt function libraries where appropriate.

I also enhanced the hotplug script to not try to start interfaces that were set to not autostart.

I improved the wrapper script to block restarting of the interface if authentication fails due to a bad password, which prevents the openwrt from continually trying to login to the fortigate with bad credentials.

### DIFF
--- a/net/openfortivpn/Makefile
+++ b/net/openfortivpn/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openfortivpn
 PKG_VERSION:=1.14.1
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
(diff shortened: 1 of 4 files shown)


lucize:
LGTM

aaronjg:
Thanks @lucize. I just made one more small change, apparently the explicit handling of peerdns and defaultroute is not needed, since netifd will take care of that. This whole proto/netifd stuff could be much better documented...

PR #8341, opened 2019-03-04
libp11: Fix dependency breakage

neheb:
This requires a +.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @dangowrt 
Compile tested: ath79

PR #12902, opened 2020-07-22
libsodium: convert into package variants

dangowrt:
Maintainer: @damianorenfer 
Compile tested: x86/64
Run tested: x86/64

Description:
Instead of configuring at build-time whether to build libsodium with `--enable-minimal` or `--disable-minimal`, convert this choice into build-variants resulting in two binary packages (libsodium having `--enable-minimal` set and libsodium-full setting `--disable-minimal`).
This is necessary for packages which require the full version of libsodium as the buildbot currently builds libsodium with `--enable-minimal` set by default.

### DIFF
--- a/libs/libsodium/Makefile
+++ b/libs/libsodium/Makefile
@@ -26,14 +26,26 @@ PKG_BUILD_PARALLEL:=1
 
 include $(INCLUDE_DIR)/package.mk
 
-define Package/libsodium
+define Package/libsodium/default
   SECTION:=libs
   CATEGORY:=Libraries
   TITLE:=P(ortable|ackageable) NaCl-based crypto library
   URL:=https://github.com/jedisct1/libsodium
 endef
 
-define Package/libsodium/description
+define Package/libsodium
+  $(call Package/libsodium/default)
+  VARIANT:=minimal
+  CONFLICTS:=libsodium-full
+endef
+
+define Package/libsodium-full
+  $(call Package/libsodium/default)
+  VARIANT:=full
+  PROVIDES:=libsodium
+endef
+
+define Package/libsodium/description/default
   NaCl (pronounced "salt") is a new easy-to-use high-speed software library for network communication, encryption, decryption, signatures, etc.
   NaCl's goal is to provide all of the core operations needed to build higher-level cryptographic tools.
   Sodium is a portable, cross-compilable, installable, packageable fork of NaCl (based on the latest released upstream version nacl-20110221), with a compatible API.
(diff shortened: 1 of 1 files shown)


neheb:
What package needs the full version?

neheb:
Oh I see, gnunet.

How about having a depends on !@LIBSODIUM_MINIMAL . Better yet, change the symbol to LIBSODIUM_FULL and depend on @LIBSODIUM_FULL .

One irritating thing about variants is what happens to InstallDev when both are selected as M.

dangowrt:
GNUnet since version 0.13.

see https://downloads.openwrt.org/snapshots/faillogs/mipsel_24kc_24kf/packages/gnunet/compile.txt
```
OpenWrt-libtool: link: warning: library `/builder/shared-workdir/build/sdk/staging_dir/toolchain-mipsel_24kc+24kf_gcc-8.4.0_musl/lib/libatomic.la' was moved.
/builder/shared-workdir/build/sdk/staging_dir/toolchain-mipsel_24kc+24kf_gcc-8.4.0_musl/bin/../lib/gcc/mipsel-openwrt-linux-musl/8.4.0/../../../../mipsel-openwrt-linux-musl/bin/ld: .libs/crypto_ecc.o: in function `GNUNET_CRYPTO_ecdsa_key_get_public':
crypto_ecc.c:(.text+0x1d4): undefined reference to `crypto_scalarmult_ed25519_base_noclamp'
Makefile:1728: recipe for target 'libgnunetutil.la' failed
make[7]: *** [libgnunetutil.la] Error 1
```

dangowrt:
> Oh I see, gnunet.
> 
> How about having a depends on !@LIBSODIUM_MINIMAL . Better yet, change the symbol to LIBSODIUM_FULL and depend on @LIBSODIUM_FULL .

That will practically disable GNUnet binaries to be build by the buildbots and provided via downloads.openwrt.org as it will not
select @LIBSODIUM_FULL (or !@LIBSODIUM_MINIMAL) unless we change the default to build the full version,

> 
> One irritating thing about variants is what happens to InstallDev when both are selected as M.

headers are identical so even if installed twice that should not matter.

neheb:
Reminds me of this: https://github.com/openwrt/packages/pull/8341

Anyway, a variant will mess up InstallDev for sure. Depending on which gets built last, compilation of gnunet will either fail or succeed.

It's not just about the headers. The library matters as well.

dangowrt:
Hm, so all we could do is guarding InstallDev to only happen for the full variant and make all packages depending on libsodium also build-depend on libsodium-full. Not exactly elegant... Any better ideas?

neheb:
Install the full version to a different location and for depending packages, add an -rpath-link=. Not elegant either.

That being said, this is exactly what I did for shadow and libxcrypt.

neheb:
In the case of libsodium minimal, the headers are different as well:

```
crypto_ecc.c:181:3: warning: implicit declaration of function 'crypto_scalarmult_ed25519_base_noclamp'; did you mean 'crypto_scalarmult_curve25519_base'? [-Wimplicit-function-declaration]
   crypto_scalarmult_ed25519_base_noclamp (pub->q_y, priv->d);
```

neheb:
Alternatively, you could revert this upstream commit: https://github.com/jedisct1/libsodium/commit/f5076db5f8ef2799bd049677b4b4020286efe397

dangowrt:
> Alternatively, you could revert this upstream commit: [jedisct1/libsodium@f5076db](https://github.com/jedisct1/libsodium/commit/f5076db5f8ef2799bd049677b4b4020286efe397)

I guess that's the best solution then. I've opened a new PR for that #12994

PR #12909, opened 2020-07-24
meson: update to 0.55.0

neheb:
Remove upstreamed patch and add distutils one.

Maintainer: @dhewg 
Compile tested: ath79

The meson developers kind of sort of fixed the stupid rpath removal that was being done. Requires an explicit -Wl,-rpath but that should be fine.


neheb:
To clarify, the patch is from https://github.com/mesonbuild/meson/pull/7027

The goal is to avoid patching every single package that uses find_installation(). I recently found out that jsoncpp also suffers from this problem, not just zstd and lz4.

dhewg:
That patch was rejected and is not upstream.

But why bother with this crap at all? It's not openwrt's problem, it's a distro's one because they chose to package a broken python. Just check for distutils in include/prereq-build.mk and bail out!

jefferyto:
@dhewg The main repo requires Python to run the scripts in [scripts](https://github.com/openwrt/openwrt/tree/master/scripts), not to help packages. If you want to convince them to add distutils as a requirement, here (the packages repo) is not the right place to do so.

dhewg:
True, I just don't think that breaking meson to work around a broken python is the way to go forward.

neheb:
@dhewg sure. but these errors will keep being reported. Those zstd and lz4 pathes are a consequence of this. Turns out jsoncpp fails in the same way.

dhewg:
I know, and they're starting to annoy me too :)
But I would bail out and report the broken python issue to the user, and not try to patch around it.
I suspect that with your patch and the next package building a python module fails in a similar fashion without that minimal python.

jefferyto:
The minimal Python isn't "broken" it's minimal. We have a "working" host Python here, you just chose not to use it.

dhewg:
It's broken as there just is no official minimal python.
Now you want to introduce a dependency to build a whole blown python just to work around distro choice breaking stuff? That's totally backwards. Bail out on broken setups and don't punish users with a working setup.
What's the point of switching to meson to gain a few seconds if you insist on building python everywhere?

jefferyto:
No, I want to carry this patch. Don’t punish the user by forcing them to install something (full Python) they had no other need to.

dhewg:
Ok, good, your earlier comment sounded like it.
I don't know the consequences of that patch, the upstream issue sounds like it really is not a good idea.
But if you guys think that this fixes the annoying issues for now I won't object. But please don't forget about it, I bet this one backfires at one point...

neheb:
So far it's only one patch. I noticed openembedded and buildroot carry more. I ported them over locally but saw no benefit.

jefferyto:
I've thought about this a bit more and I think it would be better if we don't apply this patch.

These projects that use meson (zstd, lz4, etc.) would not build on out-of-the-box Debian (i.e. if the user installed meson from source then tried to build these projects directly) if we hadn't contributed fixes upstream. If we apply this patch, we would work around the problem for us but not for anyone else.

It will be more work (mainly for @neheb :joy:) and it would be nice if this was documented in the meson FAQ or somewhere, but this would be better for everyone. (It would be best if meson fixed the issue on their end but I don't have much hope for that; for a project that claims to be mindful of dependencies they really like this distutils dependency.)

neheb:
Somewhat related: https://github.com/mesonbuild/meson/pull/7461

It'll take a while for them to remove it :P.

I'll have to rebuild my VM to see which meson packages fail to compile without the patch. I mentioned that this is currently the case with jsoncpp. That'll take a while though...

PR #12917, opened 2020-07-24
netwhere: fix compilation with newer libmicrohttpd

neheb:
API change.

Cleaned up Makefile.

Fixed license information.

Remove pointless CMAKE_INSTALL. This is not needed in InstallDev.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @benhsmith 
Compile tested: arc700

PR #12920, opened 2020-07-25
meson: backport no setuptools patch

neheb:
It seems this was cut after the release. Fixes CircleCI compilation.

Maintainer: @dhewg 
Compile tested: https://github.com/openwrt/packages/pull/12917


neheb:
Turns out, all meson packages are now failing: https://downloads.openwrt.org/snapshots/faillogs/arm_cortex-a7/packages/

It also turns out that the distutils patch does nothing. jsoncpp still fails. Removing and merging.

PR #12926, opened 2020-07-25
gstreamer suite: update to 1.17.2

neheb:
Maintainer: @flyn-org 
Compile tested: arc700

All were patched to remove a host python3-distutils dependency. This should probably be followed upstream...


flyn-org:
Looks good. Recommend merge. Thank you!

PR #12929, opened 2020-07-25
nginx: enable ssl for all variants and remove nginx-util w/o ssl

peter-stadler:
Maintainer: @Ansuel, @heil and me
Run tested: x86_64, x86_64 qemu, master, installing/upgrading the packages

Description: Make the plain `nginx` version the same as `nginx-ssl`. In a further change we will replace `nginx` by a dummy package, depending on `+nginx-ssl` instead. This allows a smooth upgrade.

For `nginx-util` leave the ssl and the ssl-nopcre versions only.
For `nginx-mod-luci` remove the duplicate ssl version.
Replace the removed packages by dummies for a smooth upgrade. The dummy packages will be removed after a transition period.

Additionally, enable the http2 module for `nginx-ssl` by default and cleanup the `nginx.init` file.


neheb:
I like this.

PR #12940, opened 2020-07-26
openvswitch: adds new UCI section ovs_switch

SimonKinane:
This new config section in package openvswitch
supports creating a named bridge, and setting
its' OpenFlow controller end-point.

An example config is included in /rom/etc/config/openvswitch

Signed-off-by: Simon Kinane <skinane@fb.com>

Maintainer: me
Compile tested: (mips, GL-MT300Nv2, OpenWrt 18.06.1)
Run tested: (mips, GL-MT300Nv2, OpenWrt version, add-br name; set controller; set alt-controller; set controller:port)

Description:

### DIFF
--- a/net/openvswitch/files/openvswitch.config
+++ b/net/openvswitch/files/openvswitch.config
@@ -6,3 +6,8 @@ config ovn_northd north
 
 config ovn_controller controller
 	option disabled 1
+
+config ovs_bridge
+	option disabled 1
+	option name 'my-bridge'
+	option controller 'tcp:192.168.0.1'
--- a/net/openvswitch/files/openvswitch.init
+++ b/net/openvswitch/files/openvswitch.init
@@ -36,6 +36,8 @@ ovs_action() {
 	for cfgtype in ovs ovn_northd ovn_controller; do
 		config_foreach "ovs_xx" "$cfgtype" "$action" "$cfgtype"
 	done
+
+	config_foreach ovs_bridge_init "ovs_bridge"
 }
 
 ovs_xx() {
@@ -51,7 +53,7 @@ ovs_xx() {
 		status|stop) ;;
 		*)
 			config_get_bool disabled "$cfg" disabled 0
-			[ "$disabled" -le 0 ] || return
+			[ "$disabled" == "0" ] || return
 			;;
 	esac
 
@@ -64,4 +66,22 @@ ovs_xx() {
 			"$ovn_ctl" "${action}_${cfgtype#ovn_}"
 			;;
 	esac
+}
+
+ovs_bridge_init() {
+	local cfg="$1"
+
+	local disabled
+	local name
+	local controller
+
+	config_get_bool disabled "$cfg" disabled 0
+	[ "$disabled" == "0" ] || return
+
+	config_get name "$cfg" name $cfg
+	ovs-vsctl --may-exist add-br "$name"
+
+	config_get controller "$cfg" controller
+	[ -n "$controller" ] && \
+		ovs-vsctl set-controller "$name" "$controller"
 }
--- a/net/openvswitch/Makefile
+++ b/net/openvswitch/Makefile
@@ -17,7 +17,7 @@ include ./openvswitch.mk
 #
 PKG_NAME:=openvswitch
 PKG_VERSION:=$(ovs_version)
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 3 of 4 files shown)


net/openvswitch/files/openvswitch.config
```diff
@@ -6,3 +6,8 @@ config ovn_northd north
 
 config ovn_controller controller
 	option disabled 1
+
+config ovs_bridge
+	option name 'my-bridge'
+	option controller '192.168.0.1'
+	option controller_port '6653' # optional
```
yousong, line 13: I think `controller` alone should suffice.  I suppose that's what regular ovs users are familiar with.  I prefer for the moment the init script do not prefix it with "tcp:" but we can do that later if that is needed.  The point is try to stick with what people are already familiar with and less burden of new trivial things.

yousong:
Hi, @SimonKinane ,

`PKG_RELEASE` in the Makefile needs to be incremented to reflect the change.

Please also add a section in the README.md documenting the new uci options.  A subsection under "How to use them" may do

net/openvswitch/files/openvswitch.config
yousong: Another thing is that this section will need a `disabled` option.  Value of that should be 1 in the sample config.

SimonKinane: > I think `controller` alone should suffice. I suppose that's what regular ovs users are familiar with. I prefer for the moment the init script do not prefix it with "tcp:" but we can do that later if that is needed. The point is try to stick with what people are already familiar with and less burden of new trivial things.

The transport prefix (e.g. "tcp:") is mandatory, so
I've added this to the sample config. If we omit the 
transport protocol, we get an error like :
"ovs|00002|vsctl|WARN|target type "1.2.3.4" is possibly erroneous".

net/openvswitch/files/openvswitch.init
```diff
@@ -65,3 +67,22 @@ ovs_xx() {
...
 	esac
 }
+
+ovs_bridge_init() {
+	local cfg="$1"
+
+	local disabled
+	local bridge_name
+	local of_controller
+	local of_controller_port
```
yousong, line 77: This variable can be removed now

net/openvswitch/files/openvswitch.init
```diff
@@ -65,3 +67,22 @@ ovs_xx() {
...
+
+	config_get_bool disabled "$cfg" disabled 0
+	[ "$disabled" -le 0 ] || return
```
yousong, line 80: `config_get_bool` will set it to either `1` or `0`, that's part of the API and convention.  `[ "$disabled" = "0" ]` should do

net/openvswitch/files/openvswitch.init
```diff
@@ -65,3 +67,22 @@ ovs_xx() {
 			;;
 	esac
 }
+
+ovs_bridge_init() {
+	local cfg="$1"
+
+	local disabled
+	local bridge_name
+	local of_controller
```
yousong, line 76: Would help reading if variable name matches uci option name

net/openvswitch/files/openvswitch.init
```diff
@@ -65,3 +67,22 @@ ovs_xx() {
...
+
+	config_get_bool disabled "$cfg" disabled 0
+	[ "$disabled" -le 0 ] || return
+
+	config_get bridge_name "$cfg" name $cfg
+	ovs-vsctl --may-exist add-br $bridge_name
+
+	config_get of_controller "$cfg" controller
+	[ -n "$of_controller" ] && \
+		ovs-vsctl set-controller $bridge_name $of_controller
```
yousong, line 87: Always surround shell variable refs with quotes is also the convention.

yousong:
After all is done, please squash the commits into a single one.  They all serve a single purpose.

net/openvswitch/files/openvswitch.init
SimonKinane: Agreed. Should I make this change for all the options "disabled" in package openvswitch ?

yousong: Do as you will ;)

SimonKinane: > `config_get_bool` will set it to either `1` or `0`, that's part of the API and convention. `[ "$disabled" = "0" ]` should do

It may be desirable to apply this globally.
The "-le" operator is arithmetic, so the string argument presumable is parsed as an int.
The "==" operator is a string comparison, so should be computationally cheaper for these single char options.

( I won't make any global changes in this pull request... )

yousong:
Thank you!

Issue #12930, opened 2020-07-25
<v4l2-utils> v4l2-ctl segfault when setting uvc controls


mgondium:
v4l2-ctl -c sharpness=16
Segmentation fault

Currently broken in trunk, works in openwrt 18, not sure in openwrt 19.

I've read a similar problem that  has to do with some hacking regarding musl, not sure if it's already fixed and just needs a pull request.

neheb:
I bumped the version to the latest.

Please run strace and post the log. I have a feeling it's a uClibc++ issue.

mgondium:
Freshly built.
OpenWrt SNAPSHOT, r13952-ac9730c495
```
root@Openwrt:~# strace v4l2-ctl -c sharpness=16
execve("/usr/bin/v4l2-ctl", ["v4l2-ctl", "-c", "sharpness=16"], 0x7fd57438 /* 13 vars */) = 0
set_thread_area(0x77e0aec8)             = 0
set_tid_address(0x77e01ff8)             = 1851
open("/etc/ld-musl-mips-sf.path", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = -1 ENOENT (No such file or directory)
open("/lib/libv4l2.so.0", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = -1 ENOENT (No such file or directory)
open("/usr/local/lib/libv4l2.so.0", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = -1 ENOENT (No such file or directory)
open("/usr/lib/libv4l2.so.0", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = 3
fcntl64(3, F_SETFD, FD_CLOEXEC)         = 0
fstat64(3, {st_mode=S_IFREG|0755, st_size=45816, ...}) = 0
read(3, "\177ELF\1\2\1\0\0\0\0\0\0\0\0\0\0\3\0\10\0\0\0\1\0\0\17\300\0\0\0004"..., 936) = 936
mmap2(NULL, 114688, PROT_READ|PROT_EXEC, MAP_PRIVATE, 3, 0) = 0x77d3e000
mmap2(0x77d55000, 20480, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED, 3, 0x7000) = 0x77d55000
close(3)                                = 0
[...]

PR #12944, opened 2020-07-26
v4l-utils: bring back getsubopt patch

neheb:
During the update to 1.20.0, this patch was mistakenly removed.

Maintainer: @thess 
Compile tested: ath79

Fixes: https://github.com/openwrt/packages/issues/12930


neheb:
Looking at this patch, I don't like it. It makes parse_next_subopt always return false, which never executes common_usage(). I guess it's fine though.

Issue #12931, opened 2020-07-25
shadowsocks-libev: Allow bind to local IPv4/v6 addresses


wyf88:
Maintainer: @yousong 
Environment: any

Description:

Dear developer, 

Is it possible to allow bind to both IPv4/v6 addresses? The configs are `local_ipv4_address` and `local_ipv6_address`. See also
https://github.com/shadowsocks/shadowsocks-libev/commit/4a42da641b6e0039497998614e84c94205939c24

Thanks a lot!

yousong:
Hi @wyf88 , please consider giving #12945 a try and see if it works for you.

yousong:
Merged both pull requests for shadowsocks-libev and luci-app-shadowsocks-libev.  Should appear in the downloads site in the following days.

PR #12945, opened 2020-07-27
shadowsocks-libev: support ss-server option local_{ipv4,ipv6}_address

yousong:
Maintainer: me
Compile tested: x86/64, qemu
Run tested: x86/64, qemu

Description:

At the moment ss-server seems to be the only component using these two
options.  It also accepts "local_address" of either ip4 or ip6 address,
but the meaning is different from that of ss-local, ss-tunnel etc.
where it is for listen bind

Link: https://github.com/shadowsocks/shadowsocks-libev/commit/4a42da641b6e0039497998614e84c94205939c24
Link: https://github.com/openwrt/packages/issues/12931

### DIFF
--- a/net/shadowsocks-libev/Makefile
+++ b/net/shadowsocks-libev/Makefile
@@ -14,7 +14,7 @@ include $(TOPDIR)/rules.mk
 #
 PKG_NAME:=shadowsocks-libev
 PKG_VERSION:=3.3.4
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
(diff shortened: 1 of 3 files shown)


wyf88:
@yousong thanks a lot for your quick action! I will give it a try in the weekend and let you know.

yousong:
v2 <- v1

 - Option names are now `local_{ipv4,ipv6}_address`, instead of `local_address_{v4,v6}`, to align with json config field names in upstream project.
 - Deprecate ss_server `bind_address`.  The equivalent is now `local_address`

PR #12948, opened 2020-07-27
mwan3: update to version 2.8.12

feckert:
Maintainer: me 

Description:

- Fix race condition on mwan3 restart
- Revert: also cleanup lock on mwan3 stop
- Remove lock file cleanup entirely
- Unify variable check

### DIFF
--- a/net/mwan3/files/etc/hotplug.d/iface/15-mwan3
+++ b/net/mwan3/files/etc/hotplug.d/iface/15-mwan3
@@ -9,13 +9,13 @@
 [ -n "$INTERFACE" ] || exit 2
 
 if [ "$ACTION" == "ifup" ]; then
-        [ -n "$DEVICE" ] || exit 3
+	[ -n "$DEVICE" ] || exit 3
 fi
 
 mwan3_lock "$ACTION" "$INTERFACE"
 config_load mwan3
 config_get_bool enabled globals 'enabled' '0'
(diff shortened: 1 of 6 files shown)


net/mwan3/files/etc/hotplug.d/iface/15-mwan3
```diff
@@ -9,24 +9,29 @@
...
+	exit 0
+}
 
-mwan3_lock "$ACTION" "$INTERFACE"
 mwan3_init
 mwan3_set_connected_iptables
 mwan3_set_custom_ipset
-mwan3_unlock "$ACTION" "$INTERFACE"
 
 config_get enabled $INTERFACE enabled 0
```
aaronjg, line 27: Can this be "config_get_bool" here to be consistent with the global enabled check?

net/mwan3/files/usr/sbin/mwan3rtmon
```diff
@@ -7,6 +7,7 @@ LOG="logger -t $(basename "$0")[$$] -p"
 
 clean_up() {
 	$LOG notice "Stopping mwan3rtmon..."
+	mwan3_unlock "service" "mwan3rtmon"
```
aaronjg, line 10: I think this could free a lock not owned by mwan3rtmon. When the process exits, if it is holding the lock it should be released anway.

PR #12950, opened 2020-07-27
mwan3: backport version 2.8.12 to stable branch openwrt-19.07

feckert:
Maintainer: me 

Description:
- IPv6 improvements
- Add check if IPv6 is enabled on the system
- Fix some shellcheck warnings
- Fix procd blocking introduced with commit https://github.com/openwrt/packages/commit/f7749c663b4ac4880cbd1ed7217073d8b03ea072#diff-34590d11545c84b961d1992ce3aefdeb

If [P/R ](https://github.com/openwrt/packages/pull/12948) is merged into master, then we should also backport and add the commits to this pull request. The pull request in master fixes the race condition between mwan3 command and hotplug-script execution.

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.8.11
+PKG_VERSION:=2.8.12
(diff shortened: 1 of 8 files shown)


aaronjg:
I have been testing with #12948, and it appears stable - so once that is merged in, this looks good for the stable branch 👍.

PR #12966, opened 2020-07-29
openvswitch: fix build with kernel 4.14.187 and handle libunwind dependency

yousong:
Maintainer: me
Compile tested: x86/64
Run tested: x86/64, qemu, ldd, kconfig, and run check

Description:

### DIFF
--- a/net/openvswitch/Makefile
+++ b/net/openvswitch/Makefile
@@ -156,7 +156,7 @@ ovs_kmod_openvswitch-lisp-intree_depends:= +kmod-openvswitch-intree
 ovs_kmod_openvswitch-lisp-intree_files:= $(ovs_kmod_intree_dir)/vport-lisp.ko
 $(eval $(call OvsKmodPackageTemplate,openvswitch-lisp-intree))
 
-ovs_common_depends:= +libatomic +libunbound +libunwind
+ovs__common_depends:= +libatomic +libunbound +libunwind
 
 # Dependency review
 #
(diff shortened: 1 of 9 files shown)


yousong:
v2 <- v1 Quote commit message of "openvswitch: only libopenvswitch depends on libunbound, libunwind"

> This is to make explicit the conditional select of libunwind is only
> needed for libopenvswitch.  If we spill it over other packages, the
> generated kconfig will have recursive dependency issue.  2 new patches
> were made for this goal
> 
> The other thing is that "+libunwind" will cause it to be built if any of the
> packages defined in this Makefile is enabled (y or m).  This is at the moment
> by-design of the build system.
> 
> Libunwind does not support architectures like arc.  Use conditional select To
> avoid (libunwind) build failures like the following,
> 
>   checking for ELF helper width... configure: error: Unknown ELF target: arc
>   make[3]: *** [Makefile:65: /data/openwrt/build_dir/target-arc_arc700_uClibc/
>   libunwind-1.3.1/.configured_68b329da9893e34099c7d8ad5cb9c940] Error 1
> 
> Things like "+PACKAGE_openvswitch-libopenvswitch:libunwind" will also result in
> recursive deps error for chains of 3 nodes.  Kconfig construct like the
> following will be made
> 
> 	config A
> 		tristate
> 		select B
> 		depends on !(C) || (x)
> 
> 	config B
> 		tristate
> 		select C
> 
> 	config C
> 		tristate
> 
> 	config x
> 		bool
> 
> Other changes include
> 
>  - Shared use of variable ovs__common_depends was removed
>  - Ovn doc build was patched out

PR #11439, opened 2020-02-26
runc/containerd: Fix incomplete cleanup in Build/InstallDev

tete1030:
Maintainer:  @G-M0N3Y-2503 
Compile tested: x86_64, OpenWrt master
Run tested: x86_64, OpenWrt master

Description:
Currently it only cleans up binaries when executing `Build/InstallDev` without deleting .built stamp file.

This leads to wrong information about existence of built runc/containerd binaries and causes error when executing `package/{runc,containerd}/install` twice.

Related PR/Issues (in Chinese): 
* coolsnowwolf/packages#34 (PR)
* coolsnowwolf/lede#3222
* coolsnowwolf/lede#3282
* coolsnowwolf/lede#3213
* coolsnowwolf/lede#1759

PR #12967, opened 2020-07-29
runc/containerd/libnetwork: Refactored Makefiles

G-M0N3Y-2503:
Maintainer: @G-M0N3Y-2503

Description:
Refactored packages to help address broken workflows mentioned in https://github.com/openwrt/packages/pull/11439, I wasn't able to reproduce their issue per sey, but this simplifies the makefiles regardless.

### DIFF
--- a/utils/containerd/Makefile
+++ b/utils/containerd/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=containerd
 PKG_VERSION:=1.2.13
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=LICENSE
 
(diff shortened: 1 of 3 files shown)


neheb:
ping @jefferyto 

Didn't know golang packages have a default InstallDev section like cmake.mk.

G-M0N3Y-2503:
In this case, I believe the `InstallDev` isn't even used, the initial motivation was to put it in an expected state if it was going to be used, but as mentioned in this comment https://github.com/openwrt/packages/pull/11439#issuecomment-658755750, it probably doesn't matter if there are binaries or not.

aparcar:
I think @dangowrt works on a implementation of runc/crun or whatever it's called to run containers on OpenWrt withouth the need of any additional packages. Progress [here](https://git.openwrt.org/project/procd.git)

jefferyto:
My apologies, I didn't have time to review this (it looks fine to me btw).

Just wanted to comment on `Build/InstallDev`: Before Go had modules, Go library packages needed to share their source code in `Build/InstallDev` so that Go programs can find them; now with modules this is mostly unnecessary. At some point this "legacy support" for non-module programs can be removed, but it's not much of a maintenance burden for now and technically usage of modules for Go programs is still optional.

PR #12971, opened 2020-07-30
glib2: add back gettext dependencies

neheb:
The libintl check is broken in meson.build. Patched it and added back
the gettext build dependencies. Avoids the proxy-intl thing that glib2
likes to use. Also fixes several problems under uClibc-ng.

Updated configure options.

Refreshed patches.

Maintainer: @tripolar 
Compile tested: arc700 with irqbalance


nobk:
still got libintl link error.
And I find my config file there only syslog-ng depend on glib2, I unselect syslog-ng and select rsyslog.

Issue #12938, opened 2020-07-26
Nginx problem under master branch!!


zxlhhyccc:
![image](https://user-images.githubusercontent.com/45259624/88482659-4bfa6800-cf95-11ea-91a2-f3640374da91.png)

neheb:
ping @peter-stadler

peter-stadler:
Thank you very much.

peter-stadler:
Does commenting `PROVIDES:=nginx-mod-luci-ssl` as in the draft PR solves it? (I cannot test it right now)

neheb:
yes

peter-stadler:
Thank you :-) I created a new PR as I did not find a way to squash commits on github's online editor (I have only access to my main system in a week again).

garypang13:
```
2020-07-28T04:04:42.0442810Z Collected errors:
2020-07-28T04:04:42.0444723Z  * satisfy_dependencies_for: Cannot satisfy the following dependencies for luci-ssl-nginx:
2020-07-28T04:04:42.0446928Z  * 	nginx-util (>= 1.4-2)
2020-07-28T04:04:42.0448520Z  * opkg_install_cmd: Cannot install package luci-ssl-nginx.
2020-07-28T04:04:42.0449776Z make[2]: *** [package/install] Error 255
2020-07-28T04:04:42.0451939Z package/Makefile:65: recipe for target 'package/install' failed
```

peter-stadler:
Hmm, right now I can test it only on my router (where it works). I tried to avoid this problem by adding the dependency statement to the same Makefile containing the depended package, too. What is `opkg list | grep nginx-util` showing?

rmandrad:
Why are you releasing code you can't test!!!! can you just undo what you have done until you properly test ?

peter-stadler:
I did quite some testing before making the PR. I am sorry that I overlooked an error for `make menuconfig`. This problem should be solved now: the change was just to remove the `PROVIDES:=nginx-mod-luci-ssl` (I could not test that change, but I think it should not make a problem elsewhere).

Unfortunately I cannot use my main system (where I do the compilation and testing in a VM) until Monday. I did the installation test again on my router (two days ago) and there I did not see the dependency problem for `nginx-util (>= 1.4-2)` ...

Can you give me some details on the problem: 
1. On which system do you experience it?
2. Does it happen on a fresh install and/or an upgrade of `luci-ssl-nginx`?
[...]

PR #12983, opened 2020-07-31
nginx-util: fix dependency

peter-stadler:
Maintainer: me
Run tested: x86_64, x86_64 qemu, master, `make` and run with `luci-ssl-nginx`

Description: Add DEPENDS since EXTRA_DEPENDS is not used by `make menuconfig`. Issue [#12938 no. 2](https://github.com/openwrt/packages/issues/12938#issuecomment-664788535).


peter-stadler:
(rebased using the right `signed-off-by` and tested)

PR #12991, opened 2020-08-01
libgpg-error: Fix compilation with GAWK 5.0

jjm2473:
Patch from Upstream.

Maintainer: W. Michael Petullo <mike@flyn.org>
Compile tested: Host MacOS, Openwrt 18.06
Run tested: 

Description:

### DIFF
--- a/libs/libgpg-error/patches/020-gawk5-support.patch
+++ b/libs/libgpg-error/patches/020-gawk5-support.patch
@@ -1,0 +1,158 @@
+From 7865041c77f4f7005282f10f9b6666b19072fbdf Mon Sep 17 00:00:00 2001
+From: NIIBE Yutaka <gniibe@fsij.org>
+Date: Mon, 15 Apr 2019 15:10:44 +0900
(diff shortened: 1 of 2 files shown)


neheb:
ping @flyn-org

neheb:
Just noticed this needs a PKG_RELEASE bump.

jjm2473:
@neheb It just fix the compile error, does it really needs a PKG_RELEASE bump?

neheb:
Hmm I guess not. These all look like Makefile fixes.

The rule regarding PKG_RELEASE is that if any installed file changes, PKG_RELEASE gets bumped so that opkg install works properly. opkg checks the ipkg filesize to see if it is correct.

flyn-org:
This looks fine to me, assuming @neheb is satisfied with the PKG_RELEASE. I believe the corresponding work upstream is in https://github.com/gpg/libgpg-error/commit/7865041c77f4f7005282f10f9b6666b19072fbdf.

PR #12992, opened 2020-08-01
libdvbcsa: add new package

Rafostar:
Libdvbcsa is a free implementation of the DVB Common Scrambling Algorithm DVB/CSA - with encryption and decryption capabilities.

OpenWrt packages like `tvheadend` and `minisatip` can benefit from it.

Signed-off-by: Rafał Dzięgiel <rafostar.github@gmail.com>

Maintainer: me
Compile tested: x86_64
Run tested: x86_64 (tested with tvheadend)

### DIFF
--- a/libs/libdvbcsa/Makefile
+++ b/libs/libdvbcsa/Makefile
@@ -1,0 +1,67 @@
+#
+# Copyright (C) 2020 Rafał Dzięgiel <rafostar.github@gmail.com>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libdvbcsa
+PKG_RELEASE:=1
+
+PKG_SOURCE_VERSION:=bc6c0b164a87ce05e9925785cc6fb3f54c02b026
+PKG_HASH:=2d761c9e094642f2c9aa7e66534c6147a59d0d0bc709ec0f2fdbb34bf020d8ec
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_SOURCE_VERSION).tar.gz
+PKG_SOURCE_URL:=https://code.videolan.org/videolan/libdvbcsa/-/archive/$(PKG_SOURCE_VERSION)
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_SOURCE_VERSION)
+
+PKG_MAINTAINER:=Rafał Dzięgiel <rafostar.github@gmail.com>
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=COPYING
+
+PKG_FIXUP:=autoreconf
+PKG_INSTALL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/libdvbcsa
+  SECTION:=libs
+  CATEGORY:=Libraries
+  TITLE:=DVB Common Scrambling Algorithm Library
+  URL:=https://www.videolan.org/developers/libdvbcsa.html
+endef
+
+define Package/libdvbcsa/description
+ Libdvbcsa is a free implementation of the DVB Common Scrambling
+ Algorithm DVB/CSA - with encryption and decryption capabilities.
+endef
+
+define Package/libdvbcsa/config
+  menu "Configuration"
+  depends on PACKAGE_libdvbcsa
+  source "$(SOURCE)/Config.in"
+  endmenu
+endef
+
+CONFIGURE_ARGS += \
+	--$(if $(CONFIG_LIBDVBCSA_DEBUG),en,dis)able-debug \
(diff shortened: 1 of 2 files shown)


libs/libdvbcsa/Makefile
```diff
@@ -0,0 +1,70 @@
...
+endef
+
+define Package/libdvbcsa/config
+  menu "Configuration"
+  depends on PACKAGE_libdvbcsa
+  source "$(SOURCE)/Config.in"
+  endmenu
+endef
+
+ifeq ($(CONFIG_LIBDVBCSA_DEBUG),y)
```
neheb, line 47: All of these can be simplified by eg:

```
CONFIGURE_ARGS += \
    --$(if $(CONFIG_LIBDVBCSA_DEBUG)),en,dis)able-debug
```

libs/libdvbcsa/Makefile
```diff
@@ -0,0 +1,70 @@
...
+  CONFIGURE_ARGS += --enable-mmx
+else ifeq ($(CONFIG_LIBDVBCSA_SSE2),y)
+  CONFIGURE_ARGS += --enable-sse2
+else ifeq ($(CONFIG_LIBDVBCSA_ALTIVEC),y)
+  CONFIGURE_ARGS += --enable-altivec
+else ifeq ($(CONFIG_LIBDVBCSA_NEON),y)
+  CONFIGURE_ARGS += --enable-neon
+endif
+
+define Build/InstallDev
```
neheb, line 61: This is a bit broad. Please be more verbose in this section.

libs/libdvbcsa/Makefile
```diff
@@ -0,0 +1,70 @@
...
+	$(CP) $(PKG_INSTALL_DIR)/* $(1)/
+endef
+
+define Package/libdvbcsa/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libdvbcsa.so* $(1)/usr/lib/
```
neheb, line 67: .so is not useful. Install .so.*

neheb:
I notice this is not in Fedora's repositories but in rpmfusion. Is DVB/CSA patented?

libs/libdvbcsa/Makefile
Rafostar: Only one can be used at a time. In `configure.ac` the used CPU extension is selected inside long `if... else if...` statement. Selecting more than one would not break the compiling, but we would end up with library optimized for the CPU on top of the `if-else` statement.

This too can be simplified with `--$(if $(CONFIG_LIBDVBCSA_SSE2),en,dis)able-sse2` like you pointed above, because I made the `Config.in` in a way that prevents selecting more then one option per arch.

Rafostar:
> I notice this is not in Fedora's repositories but in rpmfusion. Is DVB/CSA patented?

According to [https://www.videolan.org/legal.html](https://www.videolan.org/legal.html):
> Yes, you may distribute an original or a modified version of a piece of VideoLAN software as long as you comply with its license terms. 
> ...
> Therefore, software patents licenses do not apply on VideoLAN software.

This lib is licensed under GPLv2+.

libs/libdvbcsa/Makefile
neheb: Not what I meant. The InstallDev section is not verbose as it is in other packages.

neheb:
ping @openwrt/package-maintainers 

Should this be marked with BUILD_PATENTED?

libs/libdvbcsa/Makefile
Rafostar: @neheb I updated this part in PR. Is this what you wanted?

neheb: Yes

neheb:
Can you rebase against master? We have new buildbots.

Rafostar:
>Can you rebase against master? We have new buildbots.

Rebased

neheb:
:). both powerpc are failing. One of them should work as it supports floating point.

Rafostar:
Seems like they fail only because of unused variable/function. Aren't the new buildbots a little too aggressive? I mean failing because of unused variable when 99.9% of OpenWrt packages are externally developed is a little exaggeration.

neheb:
No. The main buildbots would fail the same way.

Rafostar:
OK. I will have some more free time to work on this next week.

BTW: Is there a way to test compile packages with new buildbots before submitting a PR or do I still have to clone, set arch and compile locally?

neheb:
Just force push until fixed.

neheb:
I just looked through the code. This altivec implementation looks broken.

Browsing through git history reveals that it's been last touched 7 years ago. I'd disable it by default.

neheb:
Manually merged.

BKPepe:
For reference merged commit: https://github.com/openwrt/packages/commit/51c5a8b4bcb5ba4d7447bd6ce77ddc41a46570aa

PR #12993, opened 2020-08-01
libextractor: update to 1.10

neheb:
Made two compile options dependent on whether or not the plugins are
selected.

Removed autoreconf as autools files are not being patched.

Added patch to fix compilation with libjpeg-turbo. libjpeg-turbo can be
fixed to include this API, but it's probably better for size reasons to
patch libextractor instead.

Added -liconv since it seems to be needed for some reason.

Maintainer: @dangowrt 
Compile tested: ath79


dangowrt:
Thanks!

PR #12994, opened 2020-08-01
libsodium: compile with scalarmult_ed25519_ref10

dangowrt:
Maintainer: @damianorenfer 
Compile tested: x86/64
Run tested: x86/64

Description: compile with scalarmult_ed25519_ref10

scalarmult_ed25519_ref10 was excluded from 'minimal' variant (which is
what we are building) by upstream commit f5076db5f8ef27.
Revert that to allow downstream projects to make use of it (eg. gnunet)

### DIFF
--- a/libs/libsodium/Makefile
+++ b/libs/libsodium/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libsodium
 PKG_VERSION:=1.0.18
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://download.libsodium.org/libsodium/releases \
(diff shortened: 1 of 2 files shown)


jefferyto:
Patching APIs back into the minimal build every time a downstream project needs them doesn't sound like a sustainable solution.

dangowrt:
@jefferyto @neheb Any ideas how we could resolve that long-standing regression and improve this sad situation?
The only alternative I can see which doesn't involve patching back the now missing ed25519 functions into the minimal build is simply not using the minimal build **by default**. That would result in buildbots building the full version of libsodium and hence also allow building GNUnet.
However, as that is a waste of space and most of the functions are righteously removed using the `minimal` option, I would still prefer simply revert upstream commit jedisct1/libsodium@f5076db5f8ef27 . While I agree in principle that we should carry as few patches as possible and stay close to upstream when ever that can work -- in this case it doesn't and the change in the patch is very trivial and easy to maintain.

PR #13009, opened 2020-08-03
modemmanager: Added options

valexi7:
Maintainer: @valexi7
Compile tested: ar71xx, MikroTik RouterBoard >= 128MB NAND, OpenWrt master
Run tested: Mikrotik rb912uag-5hpnd, OpenWRT master

Description:

Added:
- Signal refresh rate. This options is reset on every ModemManager restart, so we will set it if it's defined on proto setup. You can ask signal info from modem (mmcli -m 0 --signal-get). This is useful for some scripts or other applications.
- Loglevel check. If you enable debugmode (for sending AT commands to modem) it will automatically set loglevel to debug. With this we can change loglevel to ERR before connecting to avoid cluttered logging.

Question:
- Can we add signal info to netifd proto? It would be great to see SINR, RSSI, RSRQ, RSRP info in interface view.

### DIFF
--- a/net/modemmanager/README.md
+++ b/net/modemmanager/README.md
@@ -23,6 +23,7 @@ Once installed, you can configure the 2G/3G/4G modem connections directly in
         option pincode     '7423'
         option iptype      'ipv4'
         option lowpower    '1'
+        option signalrate  '30'
 
 Only 'device' and 'proto' are mandatory options, the remaining ones are all
 optional.
@@ -34,3 +35,6 @@ allowing all protocols.
 
 The 'iptype' option supports any of these values: 'ipv4', 'ipv6' or 'ipv4v6'.
 It will default to 'ipv4' if not given.
+
+The 'signalrate' option set's the signal refresh rate (in seconds) for the device.
+You can call signal info with command: mmcli -m 0 --signal-get
--- a/net/modemmanager/files/modemmanager.proto
+++ b/net/modemmanager/files/modemmanager.proto
@@ -343,6 +343,7 @@ proto_modemmanager_init_config() {
 	proto_config_add_string	 password
 	proto_config_add_string	 pincode
 	proto_config_add_string	 iptype
+	proto_config_add_int	 signalrate
 	proto_config_add_boolean lowpower
 	proto_config_add_defaults
 }
@@ -354,11 +355,11 @@ proto_modemmanager_setup() {
 	local bearermethod_ipv4 bearermethod_ipv6 auth cliauth
 	local operatorname operatorid registration accesstech signalquality
 
-	local device apn allowedauth username password pincode iptype metric
+	local device apn allowedauth username password pincode iptype metric signalrate
 
 	local address prefix gateway mtu dns1 dns2
 
-	json_get_vars device apn allowedauth username password pincode iptype metric
+	json_get_vars device apn allowedauth username password pincode iptype metric signalrate
 
 	# validate sysfs path given in config
 	[ -n "${device}" ] || {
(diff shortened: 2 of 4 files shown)


net/modemmanager/README.md
```diff
@@ -34,3 +37,12 @@ allowing all protocols.
 
 The 'iptype' option supports any of these values: 'ipv4', 'ipv6' or 'ipv4v6'.
 It will default to 'ipv4' if not given.
+
```
aleksander0m, line 40: You're mixing device-level options (e.g. signalrate) with daemon-level options (debugmode and loglevel). If you only have one single modem, this may work, but it won't work ifyou have multiple modems with one settings section for each. You cannot have daemon-level options in `/etc/config/network`

net/modemmanager/files/modemmanager.init
```diff
@@ -26,7 +33,12 @@ start_service() {
 
 	# Setup ModemManager service
 	procd_open_instance
-	procd_set_param command /usr/sbin/ModemManager
+	# If debugmode is enabled we start ModemManager in debug mode. This enables AT commands to modem.
```
aleksander0m, line 36: Instead of this debugmode, if you want AT command support via the DBus Command() interface, you can build MM using the `--with-at-command-via-dbus` configure option. This makes much more sense, if you know you're going to use those AT commands.

net/modemmanager/files/modemmanager.proto
```diff
@@ -392,6 +395,12 @@ proto_modemmanager_setup() {
 		cliauth="${cliauth}${cliauth:+|}$auth"
 	done
 
+	# set logging level
+	[ -n "${loglevel}" ] && {
+		echo "Setting logging level to: ${loglevel}"
+		mmcli -G ${loglevel}
```
aleksander0m, line 401: As said earlier, this won't work properly as soon as you have more than one modem in the system. The daemon log level is a daemon option that applies to all modems.

net/modemmanager/files/modemmanager.proto
```diff
@@ -402,6 +411,45 @@ proto_modemmanager_setup() {
...
+	# check if Signal refresh rate is set
+	if [ -n "${signalrate}" ] && [ "${signalrate}" -eq "${signalrate}" ] 2>/dev/null; then
+		echo "setting signal refresh rate to ${signalrate} seconds"
+		mmcli --modem="${device}" --signal-setup="${signalrate}"
+		
+		# log signal information
+		local modemsignal
+		modemsignal=$(mmcli --modem="${device}" --signal-get --output-keyvalue)
+		
+		cdma1x_rssi=$(modemmanager_get_field "${modemsignal}" "modem.signal.cdma1x.rssi")
```
aleksander0m, line 424: I would not do this. There's no point in printing a snapshot of signal information in the log as soon as the extended signal support is enabled. Having an option to enable extended signal support would be fine, but this print is a bit useless IMO :)

net/modemmanager/files/modemmanager.init
aleksander0m: Also, if you build using `--with-at-command-via-dbus` the log level is NOT changed to debug by default, so your concern about getting debug level when running using `--debug` no longer applies. A new MM build option in the openwrt setup makes more sense definitely.

neheb:
This requires a PKG_RELEASE bump.

ping @aleksander0m

aleksander0m:
> This requires a PKG_RELEASE bump.
> 
> ping @aleksander0m

Oh, I missed this MR update. All the new changes look good to me!

@valexi7 could you please update the PKG_RELEASE?

@nickberry17 what's your opinion on these changes?

nickberry17:
I like the functionality.  Does anyone have a better suggestion on where to put the `signalrate` variable?  I cannot think of a better place at the moment, because although it is not connectivity related (in the pure netifd sense) it relates to a specific modem.  If nobody has any objections or better suggestions, LGTM.

nickberry17 (approved):

aleksander0m:
> I like the functionality. Does anyone have a better suggestion on where to put the `signalrate` variable? I cannot think of a better place at the moment, because although it is not connectivity related (in the pure netifd sense) it relates to a specific modem. If nobody has any objections or better suggestions, LGTM.

Yes, that's also a minor thing I saw; the "place" for the signal rate config isn't really connection related, and it could also be enabled e.g. when the modem is not connected but just registered; but, well, that usecase is not a primary one in OpenWRT I'd say, so I'd say it's fine where it is.

aleksander0m:
Hey @neheb but the package version bump was not done yet

neheb:
Did so separately.

aleksander0m:
> Did so separately.

Oh! ok thanks :D

PR #13010, opened 2020-08-03
python-pytest-xdist: add new package

ja-pa:
Maintainer: me
Compile tested: Turris Omnia (TOS5), OpenWrt master
Run tested: Turris Omnia (TOS5), OpenWrt master

Description:
This PR adds new package **python-pytest-xdist** and its dependencies **python-apipkg** and **python-execnet**
It's python-pytest package wich provides usefull test execution modes

```
The pytest-xdist plugin extends pytest with some unique test execution modes:

    test run parallelization: if you have multiple CPUs or hosts you can use those for a combined test run. This allows to speed up development or to use special resources of remote machines.
    --looponfail: run your tests repeatedly in a subprocess. After each run pytest waits until a file in your project changes and then re-runs the previously failing tests. This is repeated until all tests pass after which again a full run is performed.
    Multi-Platform coverage: you can specify different Python interpreters or different platforms and run tests in parallel on all of them.
```

### DIFF
--- a/lang/python/python-execnet/Makefile
+++ b/lang/python/python-execnet/Makefile
@@ -1,0 +1,45 @@
+#
+# Copyright (C) 2020 CZ.NIC, z. s. p. o. (https://www.nic.cz/)
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=python-execnet
+PKG_VERSION:=1.7.1
+PKG_RELEASE:=1
+
+PYPI_NAME:=execnet
+PKG_HASH:=cacb9df31c9680ec5f95553976c4da484d407e85e41c83cb812aa014f0eddc50
+
+PKG_MAINTAINER:=Jan Pavlinec <jan.pavlinec@nic.cz>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_PARALLEL:=0
+HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
+
+include ../pypi.mk
+include $(INCLUDE_DIR)/package.mk
+include ../python3-package.mk
+
+define Package/python3-execnet
+  SUBMENU:=Python
+  SECTION:=lang
+  CATEGORY:=Languages
+  TITLE:=Library for distributed interact mechanism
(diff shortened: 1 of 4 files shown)


BKPepe:
cc: @jefferyto and @commodo

lang/python/python-pytest-xdist/Makefile
```diff
@@ -0,0 +1,49 @@
...
+  SECTION:=lang
+  CATEGORY:=Languages
+  TITLE:=Pytest plugin for parallel test running
+  URL:=https://github.com/pytest-dev/pytest-xdist
+  DEPENDS:= \
+	+python3-light \
+	+python3-pytest \
+	+python3-execnet \
+	+python3-apipkg \
+	+python3-multiprocessing
```
jefferyto, line 39: pytest-xdist [does not depend](https://github.com/pytest-dev/pytest-xdist/blob/v1.34.0/setup.py#L3) on apipkg directly. It also depends on pytest-forked and six.

Also, it [uses multiprocessing](https://github.com/pytest-dev/pytest-xdist/blob/v1.34.0/src/xdist/plugin.py#L15-L18) only when it cannot import `os.cpu_count` (because `os.cpu_count` was [added](https://docs.python.org/3/library/os.html#os.cpu_count) in Python 3.4 but it supports older versions). Since our Python always has `os.cpu_count`, multiprocessing is not necessary.

lang/python/python-execnet/Makefile
```diff
@@ -0,0 +1,45 @@
...
+include $(INCLUDE_DIR)/package.mk
+include ../python3-package.mk
+
+define Package/python3-execnet
+  SUBMENU:=Python
+  SECTION:=lang
+  CATEGORY:=Languages
+  TITLE:=Library for distributed interact mechanism
+  URL:=https://execnet.readthedocs.io
+  DEPENDS:= +python3-light +python3-ctypes
```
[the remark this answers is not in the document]
ja-pa: Fixed to python3-apipkg

lang/python/python-pytest-xdist/Makefile
ja-pa: Removed python3-multiprocessing and python3-apipkg 
Added python3-pytest-forked python3-six

ja-pa:
@jefferyto Thanks for review. I fixed problems with dependecies and also added new package python3-pytest-forked

ja-pa:
I squashed fixups and rebase it to master. Thanks for your review!

PR #13017, opened 2020-08-04
docker-ce: improvements / changes

feckert:
Maintainer: @G-M0N3Y-2503 

Description:

- Add the bridge device to the uci network config
- Change uci hosts config parameter type this is not an option this is a list element
- Add postinst script to enable, uciadd and start docker-ce on package installation
- Add prerm script to disable and stop docker-ce on package remove

### DIFF
--- a/utils/docker-ce/Makefile
+++ b/utils/docker-ce/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=docker-ce
 PKG_VERSION:=19.03.12
-PKG_RELEASE:=6
+PKG_RELEASE:=7
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=components/cli/LICENSE components/engine/LICENSE
 
--- a/utils/docker-ce/Makefile
+++ b/utils/docker-ce/Makefile
@@ -133,4 +133,22 @@ define Package/docker-ce/install
 		$(1)/etc/sysctl.d/12-br-netfilter-ip.conf
 endef
 
+define Package/docker-ce/postinst
+#!/bin/sh
+[ -n "$$IPKG_INSTROOT" ] || {
+	/etc/init.d/dockerd enable
+	/etc/init.d/dockerd uciadd
+	/etc/init.d/dockerd start
+}
+endef
+
+define Package/docker-ce/prerm
+#!/bin/sh
+[ -n "$$IPKG_INSTROOT" ] || {
+	/etc/init.d/dockerd disable
+	/etc/init.d/dockerd stop
+	/etc/init.d/dockerd ucidel
+}
+endef
+
 $(eval $(call BuildPackage,docker-ce))
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -54,7 +54,7 @@ uciadd() {
 
 	# Add network interface
 	if ! uci -q get network.docker >/dev/null; then
-		logger -t "dockerd-init" -p notice "Adding docker default bridge to network uci config (docker0)"
+		logger -t "dockerd-init" -p notice "Adding docker default interface to network uci config (docker)"
 		uci_quiet add network interface
 		uci_quiet rename network.@interface[-1]="docker"
 		uci_quiet set network.docker.ifname="docker0"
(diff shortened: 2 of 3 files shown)


utils/docker-ce/files/dockerd.init
```diff
@@ -85,11 +96,15 @@ ucidel() {
...
+	logger -t "dockerd-init" -p notice "Deleting docker default bridge device from network uci config (docker0)"
+	uci_quiet delete network.docker0
+	uci_quiet commit network
+
+	logger -t "dockerd-init" -p notice "Deleting docker default interface from network uci config (docker)"
 	uci_quiet delete network.docker
 	uci_quiet commit network
 
-	logger -t "dockerd-init" -p notice "Deleting docker default bridge firewall zone from firewall uci config (docker0)"
+	logger -t "dockerd-init" -p notice "Deleting docker firewall zone from firewall uci config (docker"
```
G-M0N3Y-2503, line 107:
```suggestion
	logger -t "dockerd-init" -p notice "Deleting docker firewall zone from firewall uci config (docker)"
```

utils/docker-ce/files/dockerd.init
```diff
@@ -52,9 +52,20 @@ uciadd() {
 		exit 0
 	}
 
+	# Add docker bridge device
```
G-M0N3Y-2503, line 55: I'm not super familiar with the network UCI, so why do we need a `bridge` type interface as well as a normal interface?
I would have thought that we could just `uci_quiet set network.docker.type="bridge"` on the one we already have.

utils/docker-ce/Makefile
```diff
@@ -133,4 +133,21 @@ define Package/docker-ce/install
...
+	/etc/init.d/dockerd uciadd
+	/etc/init.d/dockerd start
+}
+endef
+
+define Package/docker-ce/prerm
+#!/bin/sh
+[ -n "$$IPKG_INSTROOT" ] || {
+	/etc/init.d/dockerd disable
+	/etc/init.d/dockerd stop
```
G-M0N3Y-2503, line 149: Perhaps it should be left until later. 
but it'd be nice if we could prompt and run `ucidel` if the user selects `y` or run it by default if there isn't anything in `uci changes`

feckert: The uci changes are tricky because there are different staging areas for them.
The LuCI uses something different than when we run the uci config via cmd.
If we want to prompt the user if the config should be deleted, then this does not work if we remove the package over the LuCI. So I think the only solution is either do not remove the uci or remove them with out asking as it is now.

utils/docker-ce/files/dockerd.init
feckert: Recently there is a staged [pullrequest](https://github.com/openwrt/luci/pull/4307) from @jow- to create and manage devices in the LuCI.
So from the beginning an interface works on a device.
So I think we should add also the docker as a device to the uci because this is a bridge.
And on top of them we give them an ip.
This better reflects the reality on the system.

G-M0N3Y-2503: Oh, `config device` is something that was added recently, thanks for the info

utils/docker-ce/Makefile
G-M0N3Y-2503: > remove them with out asking as it is now

Just to clarify, you mean with `ucidel` right? because I can see it on the package removal.

In the future, it might be nice to look into using `uci changes firewall` to detect if we can make changes and commit. because that way we could remove all the docker UCI additions on the package removal.

feckert: > > remove them with out asking as it is now
> 
> Just to clarify, you mean with `ucidel` right? because I can see it on the package removal.

Yes

> In the future, it might be nice to look into using `uci changes firewall` to detect if we can make changes and commit. because that way we could remove all the docker UCI additions on the package removal.

That's what I would do with the command ucidel?

feckert:
@G-M0N3Y-2503 I just did another git push force.

With the following changes:

- Call ucidel on a docker remove in the prerm script https://github.com/openwrt/packages/pull/13017/commits/c7c593dcfebe51ecc992f3c5b902eb18e2d7d791#diff-7eea1abdaf7c48c6f04ed7980fb62c0eR150
- Add the interface docker to the new uci bridge device docker0 https://github.com/openwrt/packages/pull/13017/commits/46bf9606abe795c60e40899e5533502117e57e23#diff-d2f590ffe4d65867a6968c82a4a2ed5aR75

If get a :+1: i will merge this.

utils/docker-ce/files/dockerd.init
```diff
@@ -52,20 +52,33 @@ uciadd() {
...
 	if ! uci -q get network.docker >/dev/null; then
-		logger -t "dockerd-init" -p notice "Adding docker default bridge to network uci config (docker0)"
+		logger -t "dockerd-init" -p notice "Adding docker default interface to network uci config (docker)"
 		uci_quiet add network interface
 		uci_quiet rename network.@interface[-1]="docker"
 		uci_quiet set network.docker.ifname="docker0"
 		uci_quiet set network.docker.proto="static"
 		uci_quiet set network.docker.auto="0"
+		# Add docker interface to docker0 bridge device
+		uci_quiet add_list network.docker0.ifname="docker0"
```
G-M0N3Y-2503, line 75: Would it make sense to just swap these code blocks around?
That is, we add the interface first then when we create the bridge device we add the interface to it?

Also, does this change mean we should not `uci set network.docker0.bridge_empty="1"` anymore?

feckert: > Would it make sense to just swap these code blocks around?
> That is, we add the interface first then when we create the bridge device we add the interface to it?

Thanks for the advice. I did.

> Also, does this change mean we should not `uci set network.docker0.bridge_empty="1"` anymore?

Also done.

G-M0N3Y-2503:
Looks good to me

PR #13027, opened 2020-08-05
inadyn: update to 2.7

neheb:
Remove pic hack.

Fix license information.

Maintainer: @openwrt/package-maintainers 
Compile tested: ath79

### DIFF
--- a/net/inadyn/Makefile
+++ b/net/inadyn/Makefile
@@ -8,15 +8,15 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=inadyn
-PKG_VERSION:=2.5
+PKG_VERSION:=2.7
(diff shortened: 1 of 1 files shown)


net/inadyn/Makefile
```diff
@@ -41,16 +41,13 @@ define Package/inadyn/conffiles
...
 endef
 
-TARGET_CFLAGS += $(FPIC)
-
 CONFIGURE_ARGS += \
 	--enable-shared \
 	--disable-static \
 	--enable-openssl \
-	--without-pic
+	--with-pic
```
aparcar, line 48: Why this change?

neheb: Removed the CFLAGS hack and replaced with this.

PR #13036, opened 2020-08-05
acme: fix alignment space vs tabs

feckert:
Maintainer: @tohojo 
Run tested: no

Description:
This is a cosmetic change that optimizes the memory space of the script. I replaced all spaces for indenting the script with tabs. I know that there is no function change and that it must not matter.
But there are targets where every byte counts that can be saved.
Here it makes the sum of the scripts that are used.

Size of acme before change:

```
1109 acme.config
495 acme.init
10384 run.sh
```

Size of acme after change:

```
941 acme.config
459 acme.init
9054 run.sh 
```

This is a memory saving of

- 15,1% acme.config 
- 7,2% acme.init 
- 12,8% run.sh

### DIFF
--- a/net/acme/Makefile
+++ b/net/acme/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=acme
 PKG_VERSION:=2.8.6
-PKG_RELEASE:=3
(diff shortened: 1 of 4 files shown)


tohojo:
Heh, sure, why not!

PR #13038, opened 2020-08-05
json-glib: fix compilation with full language support

neheb:

Maintainer: @micmac1 
Compile tested: ath79

### DIFF
--- a/libs/json-glib/Makefile
+++ b/libs/json-glib/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=json-glib
 PKG_VERSION:=1.4.4
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@GNOME/json-glib/1.4
@@ -47,6 +47,8 @@ MESON_ARGS += \
 	-Ddocs=false \
 	-Dintrospection=false \
 	-Dman=false
+
+TARGET_LDFLAGS += $(if $(INTL_FULL),-lintl)
 
 # Disable installed-tests; this also indirectly removes
 # build_aux/gen-installed-test.py calls (copied from Gentoo)


micmac1:
Hello Rosen,

The problem is not in json-glib. The problem is in glib2. It's not json-glib that pulls in the functions that require "-lintl", but glib2.

Back when autotools were used to build glib2 there were no hacks needed. glib2 would detect that it needed to add "-lintl". It added it to it's pkg-config file as well:

autotools glib2 pkg-config file: Libs: -L${libdir} -lglib-2.0 -lintl
meson glib2 pkg-config file: Libs: -L${libdir} -lglib-2.0 -pthread -liconv -lm

This way the packages that depend on glib2 and use pkg-config are all set.

When using autotools glib2 relies on glib-gettext.m4 to add the proper flags. The meson build obviously doesn't. In the glib2 source there's actually a comment regarding these differences:

_"FIXME: glib-gettext.m4 has much more checks to detect broken/uncompatible
implementations. This could be extended if issues are found in some platforms." (meson.build)_

So in my opinion glib2 should have CONFIG_BUILD_NLS in PKG_CONFIG_DEPENDS _and_ add "-lintl" to it's pkg-config file when needed. This way one doesn't need to add workarounds to packages that depend on glib2.

Kind regards,
Seb

neheb:
Eh no. I'm not touching glib2's libintl handling again. Last time I did it caused massive buildbot failure where all depending packages picked up libintl from gettext-full even when full language support was disabled.

micmac1:
On Thu, Aug 06, 2020 at 08:21:11PM +0000, Rosen Penev wrote:
> Eh no. I'm not touching glib2's libintl handling again. Last time I
> did it caused massive buildbot failure where all depending packages
> picked up libintl from gettext-full even when full language support
> was disabled.

Your observation seems unrelated to what I wrote. On the build bots
CONFIG_BUILD_NLS is not set.

micmac1:
I looked at a few other packages that depend on glib2 to see how they do with CONFIG_BUILD_NLS. So I tried to compile grilo and grilo-plugins.

libsoup:
`soup-init.c:(.text.startup+0x1e): undefined reference to `libintl_bindtextdomain'`

grilo:
`grilo.c:(.text+0x162): undefined reference to `libintl_bindtextdomain'`

gstreamer:
`gsterror.c:(.text+0x12a): undefined reference to `libintl_dgettext'`

gst1-plugins-base:
`tags.c:(.text+0x26): undefined reference to `libintl_bindtextdomain'`

I'm guessing there are more of these cases. Adding a workaround to all of them seems tedious.

Instead something like this would bring back one autotools behavior, adding '-lintl' to the glib-2.0.pc file when CONFIG_BUILD_NLS is set that is.

```
diff --git a/libs/glib2/Makefile b/libs/glib2/Makefile
index 434a40dd9..505c841cf 100644
--- a/libs/glib2/Makefile
+++ b/libs/glib2/Makefile
@@ -26,6 +26,8 @@ PKG_BUILD_DEPENDS:=meson/host gettext-full libiconv/host
 HOST_BUILD_DEPENDS:=meson/host gettext-full/host libiconv/host libffi/host
 PKG_INSTALL:=1

+PKG_CONFIG_DEPENDS:=CONFIG_BUILD_NLS
+
 include $(INCLUDE_DIR)/nls.mk
@@ -111,6 +113,10 @@ define Build/InstallDev
                $(1)/usr/lib/

+ifneq ($(INTL_FULL),)
+       $(SED) '/^Libs:/s/$$$$/ -lintl/' \
+               $(PKG_INSTALL_DIR)/usr/lib/pkgconfig/glib-2.0.pc
+endif
        $(INSTALL_DATA) \
                $(PKG_INSTALL_DIR)/usr/lib/pkgconfig/*.pc \
                $(1)/usr/lib/pkgconfig
```

Apart from this I found something else that doesn't work with glib2. After reconfigure it doesn't clean the build dir and then may trip over the remnants of the previous compile, so you need to run an explicit "clean" on the glib2 package. But that's a different topic and maybe it's not even particularly related to glib2, but to Ninja/Meson. I don't know.

neheb:
Figured it out. The issue is here: https://github.com/GNOME/glib/blob/master/meson.build#L1989

My initial patch turned the initial if to if false , which took the else path. Unfortunately, that adds a dependency to gettext-full. I don't see a good way to modify the meson file so your pkgconfig hack seems correct.

Issue #11336, opened 2020-02-14
Lang: add rust support


peterwillcn:
Add [rust-lang](https://www.rust-lang.org/) support

neheb:
Why?

klukonin:
I'll let myself to answer.

First of all, some packages such as Suricata require Rust to work properly.

https://github.com/openwrt/packages/pull/8659

champtar:
Feature requests are rarely effective, maybe try the forum or IRC
To respond to @neheb, rust-lang for embedded make more sense than golang or python, better performance and size, when the architecture is supported :(

neheb:
Sure. python is relatively huge.

I know go is large. No idea how big rust is.

champtar:
I don't have much experience in rust
Just did some tests and a stripped hello world is ~200k with the rust stdlib
There is no stable ABI yet in rust, so it seems that you have static binaries with everything in it except libc which is inked
no idea how hard it is to have separate libs, as for OpenWRT the solution around unstable ABI can be just rebuild the firmware

champtar:
Some people already played with rust on OpenWRT
https://medium.com/althea-mesh/cross-compiling-complex-rust-programs-for-openwrt-targets-80897fcf7648
https://github.com/althea-net/althea-packages/blob/master/althea/althea-rust-binaries/Makefile
https://github.com/japaric/rust-cross

champtar:
Not sure how slow and painfull it would be to rebuild the rust compiler & cargo like we do for golang.

jefferyto:
I've thought about adding Rust support for a while... My main concerns are:

* musl is in their [second tier](https://forge.rust-lang.org/release/platform-support.html#tier-2) of support, with most musl platforms only getting build tested with the standard library only. I don't know how stable that is, though I know almost nothing about Rust and maybe it is already very stable.

* I don't have a lot of free time to do this :joy: Building the Go build system took a while, even though I basically copied Debian's system. Plus I'm already busy enough supporting Python and Go, not sure I can add another language.

On the other hand, since I'm the one that brought Go packages into the feed, I feel like I need to make Rust packages possible to bring balance to the Force :joy:

diizzyy:
Not to mention that ~90%+ of all targets wont be able to run it anyway?

Grommish:
[...]

PR #13040, opened 2020-08-06
rustup: Bring up rustup rustc/cargo language installer

Grommish:

Maintainer: me
Compile tested: None
Run tested: None

Description:
Work in progress. Input/Changes would be exceedingly helpful.

### DIFF
--- a/lang/rustup/Makefile
+++ b/lang/rustup/Makefile
@@ -1,0 +1,75 @@
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=rustup
+PKG_VERSION:=1.22.1
+PKG_RELEASE:=1
+PKG_LICENSE:=Apache-2.0 MIT
+PKG_LICENSE_FILES:=LICENSE-APACHE LICENSE-MIT
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/rust-lang/rustup.git
+PKG_SOURCE_DATE:=2020-07-08
+PKG_SOURCE_VERSION:=995002f2b88e8eea1cdf4879fa9e7cdc402b4938
+PKG_HASH:=f37a7b56b36e4acf24675d6d3794059c228b750f780654900d3f8a5cecd52695
+PKG_HOST_ONLY:=1
+
+include $(INCLUDE_DIR)/host-build.mk
+include $(INCLUDE_DIR)/package.mk
+include ./rustc-triple.mk
+
+define Package/rustup
+  	SECTION:=lang
+  	CATEGORY:=Languages
+  	TITLE:=Rust Programming Language Installer
+  	URL:=https://www.rust-lang.org/
+endef
+
+define Package/rustup/description
+  The rustup package installs the Rust Programming Language (rustc/cargo)
+  for use with packages that require Rust to compile.
+
+  For Rust Support, add the following to the package:
+  ***************************************************
+  PKG_BUILD_DEPENDS:=rustup/host
+  include ../../lang/rustup/rustc-triple.mk
+
+  CONFIGURE_VARS += \
+  	CARGO_HOME=$(BUILD_DIR_HOST)/.cargo \
+  	RUSTUP_HOME=$(BUILD_DIR_HOST)/.rustup \
+   	ac_cv_path_CARGO="$(BUILD_DIR_HOST)/.cargo/bin/cargo" \
+    	ac_cv_path_RUSTC="$(BUILD_DIR_HOST)/.cargo/bin/rustc"
+
+endef
+
+RUSTUP_INIT_ARGS:= \
+  	--profile default \
+	--default-toolchain stable \
+  	--target $(RUSTC_TARGET_ARCH)
+
+define Host/Configure
+	CARGO_HOME=$(BUILD_DIR_HOST)/.cargo \
+	RUSTUP_HOME=$(BUILD_DIR_HOST)/.rustup \
+	$(BUILD_DIR_HOST)/$(PKG_SOURCE_SUBDIR)/rustup-init.sh -y -v $(RUSTUP_INIT_ARGS)
+endef
+
+define Host/Compile
+	true
+endef
+
+define Host/Install
+	CARGO_HOME=$(BUILD_DIR_HOST)/.cargo \
+	RUSTUP_HOME=$(BUILD_DIR_HOST)/.rustup \
+	$(BUILD_DIR_HOST)/.cargo/bin/rustup toolchain install stable -t $(RUSTC_TARGET_ARCH)
+endef
+
+define Package/rustup/compile
+endef
+
+define Package/rustup/install
+endef
+
+$(eval $(call HostBuild))
+$(eval $(call BuildPackage,rustup))
--- a/lang/rustup/rustc-triple.mk
+++ b/lang/rustup/rustc-triple.mk
@@ -1,0 +1,74 @@
+# To enable rustc/cargo support in your packages, add:
+# include $(TOPDIR)/feeds/packages/lang/rustup/rustc-triple.mk
+# below include $(TOPDIR)/rules.mk
+#
+# You will need to add --host=$(RUSTC_TARGET_ARCH) to your packages
+# CONFIGURE_VARS to cross-compile for the target architecture
+
+RUSTC_ARCH_TARGETS:= \
+aarch64-unknown-linux-gnu aarch64-unknown-linux-musl \
+\
+arm-unknown-linux-gnueabi arm-unknown-linux-gnueabihf \
+arm-unknown-linux-musleabi arm-unknown-linux-musleabihf \
+\
+armv4t-unknown-linux-gnueabi \
+\
+armv5te-unknown-linux-gnueabi armv5te-unknown-linux-musleabi \
+\
+armv7-unknown-linux-gnueabi armv7-unknown-linux-gnueabihf \
+armv7-unknown-linux-musleabi \
+\
+hexagon-unknown-linux-musl \
+\
+i586-unknown-linux-gnu i586-unknown-linux-musl \
+\
+i686-unknown-linux-gnu i686-unknown-linux-musl \
+\
+mips-unknown-linux-gnu mips-unknown-linux-musl mips-unknown-linux-uclibc \
+\
+mips64-unknown-linux-gnuabi64 mips64-unknown-linux-muslabi64 \
+\
+mips64el-unknown-linux-gnuabi64 mips64el-unknown-linux-muslabi64 \
+\
+mipsel-unknown-linux-gnu mipsel-unknown-linux-musl \
+mipsel-unknown-linux-uclibc \
+\
+mipsisa32r6-unknown-linux-gnu mipsisa32r6el-unknown-linux-gnu \
+\
+mipsisa64r6-unknown-linux-gnuabi64 mipsisa64r6el-unknown-linux-gnuabi64 \
+\
+powerpc-unknown-linux-gnu powerpc-unknown-linux-musl \
+\
+powerpc64-unknown-linux-gnu powerpc64-unknown-linux-musl \
+\
+powerpc64le-unknown-linux-gnu powerpc64le-unknown-linux-musl \
+\
+riscv64gc-unknown-linux-gnu \
+\
+s390x-unknown-linux-gnu \
+\
+sparc-unknown-linux-gnu \
+\
+sparc64-unknown-linux-gnu \
+\
+thumbv7neon-unknown-linux-gnueabihf thumbv7neon-unknown-linux-musleabihf \
+\
+x86_64-unknown-linux-gnu x86_64-unknown-linux-musl
+
+ifeq ($(CONFIG_LIBC_USE_MUSL),y)
+RUSTC_TARGET_ARCH:=$(strip $(foreach v,$(filter $(ARCH)-%, $(RUSTC_ARCH_TARGETS)), $(if $(findstring -musl,$v),$v)))
+else ifeq ($(CONFIG_LIBC_USE_GLIBC),y)
+RUSTC_TARGET_ARCH:=$(strip $(foreach v,$(filter $(ARCH)-%, $(RUSTC_ARCH_TARGETS)), $(if $(findstring -gnu,$v),$v)))
+else
+RUSTC_TARGET_ARCH:=$(strip $(foreach v,$(filter $(ARCH)-%, $(RUSTC_ARCH_TARGETS)), $(if $(findstring -uclibc,$v),$v)))
+endif
+
+# More than one triple-target remains.
+ifneq ($(word 2, $(RUSTC_TARGET_ARCH)),)
+$(error RUSTC ERROR: Unsupported or Unknown Target Triple: $(RUSTC_TARGET_ARCH))
+endif
+
+# These are environment variables that are used by other packages to
+# define where rustc/cargo are kept.
+CARGO_HOME=$(BUILD_DIR_HOST)/.cargo
+RUSTUP_HOME=$(BUILD_DIR_HOST)/.rustup


lang/rustup/Makefile
```diff
@@ -0,0 +1,66 @@
...
+define Host/Compile
+	set RUSTUP_HOME=$(HOST_RUST_DIR)
+	set CARGO_HOME=$(HOST_CARGO_DIR)
+
+	$(BUILD_DIR_HOST)/$(PKG_NAME)-$(PKG_VERSION)/rustup-init.sh -y -v --profile default --target $(RUSTC_TARGET_ARCH)
+endef
+
+define Host/Install
+	$(HOST_CARGO_DIR)/bin/rustup target add $(RUSTC_TARGET_ARCH)
+	set PATH=$(HOST_CARGO_DIR)/bin:$(PATH)
```
Grommish, line 59: This fails to set.  if I manually call host/configure, it seems fine.  but then the PATH doesn't hold and by the time rustc is needed (in a suricata package I'm also working on), it can't be found.

Grommish:
This passed my ARCH test, but would need to be tested by others.  I will be sending up a Suricata 5.0.3 package that can be used for testing, as it requires rust to compile.

test package (suricata 5.0.3) PR can be found: https://github.com/openwrt/packages/pull/13170

adschm:
@Grommish I don't do packages, but there might be many people around that are familiar with the requirements for package feed I'm sure.

lang/rustup/Makefile
```diff
@@ -0,0 +1,70 @@
+#
+# Copyright (C) 2020 OpenWrt.org
```
karlp, line 2: You can't assign copyright to openwrt.org, you either need to drop it, or copyright yourself.  (Unless you actually have a copyright assignment on file with ... SPI or SFC or whoevers' in charge of ti these days)

lang/rustup/Makefile
```diff
@@ -0,0 +1,70 @@
...
+    TITLE:=Rust Programming Language Installer
+    URL:=https://www.rust-lang.org/
+endef
+
+define Package/rustup/description
+ rustup installs The Rust Programming Language from the official release
+ channels, enabling you to easily switch between stable, beta, and nightly
+ compilers and keep them updated. It makes cross-compiling simpler with binary
+ builds of the standard library for common platforms. And it runs on all
+ platforms Rust supports, including Windows.
```
karlp, line 39: It might be more appropriate to put somethign here more relevant to rustup on openwrt?

lang/rustup/Makefile
Grommish: Dropped.

lang/rustup/Makefile
Grommish: Thank you for the critique - Updated!

lang/rustup/Makefile
```diff
@@ -0,0 +1,64 @@
...
+    CATEGORY:=Languages
+    TITLE:=Rust Programming Language Installer
+    URL:=https://www.rust-lang.org/
+endef
+
+define Package/rustup/description
+ The rustup package installs the Rust Programming Language (rustc/cargo)
+ for use with packages that require Rust to compile.
+
+ This is a HOST ONLY package; no installed files to the target device.
```
BKPepe, line 35: What about BUILDONLY? ;-) Two spaces as indentation here as well.

lang/rustup/rustc-triple.mk
```diff
@@ -0,0 +1,76 @@
+# To enable rustc/cargo support in your packages, add:
+# include $(TOPDIR)/feeds/packages/lang/rustup/rustc-triple.mk
```
BKPepe, line 2: What if packages feed is named differently?

lang/rustup/Makefile
```diff
@@ -0,0 +1,64 @@
...
+PKG_NAME:=rustup
+PKG_VERSION:=1.22.1
+PKG_RELEASE:=1
+PKG_LICENSE:=MIT
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/rust-lang/rustup.git
+PKG_SOURCE_DATE:=2020-07-18
+PKG_SOURCE_VERSION:=ac1a8c415b5c2886275ca92a3a8f3a4cd0587711
+PKG_HASH:=46c1cf800e48f43bf0b3e19bf1eb2ed414106e5ff2647cf7858286489454acf2
```
BKPepe, line 15: Why you are not using tagged releases?

lang/rustup/Makefile
```diff
@@ -0,0 +1,64 @@
...
+
+define Host/Compile
+	true
+endef
+
+define Build/InstallDev
+	RUSTUP_HOME=$(HOST_RUST_DIR) CARGO_HOME=$(HOST_CARGO_DIR) $(HOST_CARGO_DIR)/bin/rustup target add $(RUSTC_TARGET_ARCH)
+endef
+
+# Rustup is a Host-only Package.  Nothing to be done in Package/
```
BKPepe, line 56: I think this is clear. Can be dropped

lang/rustup/Makefile
```diff
@@ -0,0 +1,64 @@
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=rustup
+PKG_VERSION:=1.22.1
+PKG_RELEASE:=1
+PKG_LICENSE:=MIT
```
BKPepe, line 9: I see in their repo, there is also APACHE license. What about to add PKG_LICENSE_FILES?

BKPepe (changes requested):
I didn't look at it much, some quick notes about it. I didn't do compile neither run test. Still thinking that this does not look right because it should be done similar way as go or python in OpenWrt.

Take a look at this comment https://github.com/openwrt/packages/issues/11336#issuecomment-589975725 I think @jefferyto should have word on this as well.

lang/rustup/Makefile
```diff
@@ -0,0 +1,70 @@
...
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/rust-lang/rustup.git
+PKG_SOURCE_DATE:=2020-07-18
+PKG_SOURCE_VERSION:=ac1a8c415b5c2886275ca92a3a8f3a4cd0587711
+PKG_HASH:=46c1cf800e48f43bf0b3e19bf1eb2ed414106e5ff2647cf7858286489454acf2
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)
```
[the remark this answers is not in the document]
Grommish: I think the PKG_BUILD_DIR/PKG_SOURCE_SUBDIR was vestigial from when I was still working it up.  I'm sure they don't need to be there. I'll test and correct

lang/rustup/Makefile
```diff
@@ -0,0 +1,70 @@
...
+PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
```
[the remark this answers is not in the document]
Grommish: the .tar.gz was just arbitrary since I didn't know there is a preferred format.  I'll correct

lang/rustup/rustc-triple.mk
Grommish: This is something I seriously struggled with, since I don't know another way to ultimately pass $(RUSTC_TARGET_ARCH) to the packages that need the rust variables. What is the best way to do this?

lang/rustup/Makefile
jefferyto: There is also `BUILDONLY:=1`; it seems host-only tools include both of these.

lang/rustup/rustc-triple.mk
jefferyto: There is a similar dilemma with the Python include files ([Using Python in external/other package feeds](https://github.com/openwrt/packages/tree/master/lang/python#using-python-in-externalother-package-feeds)). For packages in this repo, relative paths are best ([Include python3-package.mk](https://github.com/openwrt/packages/tree/master/lang/python#include-python3-packagemk)).

lang/rustup/Makefile
```diff
@@ -0,0 +1,64 @@
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=rustup
```
jefferyto, line 6: `rustup` is the name of the installer for Rust. I don't think this package should be named after the installer.

lang/rustup/Makefile
```diff
@@ -0,0 +1,64 @@
...
+ The rustup package installs the Rust Programming Language (rustc/cargo)
+ for use with packages that require Rust to compile.
+
+ This is a HOST ONLY package; no installed files to the target device.
+endef
+
+define Host/Configure
+	RUSTUP_HOME=$(HOST_RUST_DIR) \
+	CARGO_HOME=$(HOST_CARGO_DIR) \
+	$(BUILD_DIR_HOST)/$(PKG_SOURCE_SUBDIR)/rustup-init.sh -y -v --profile default --target $(RUSTC_TARGET_ARCH)
```
jefferyto, line 41: I believe when `Host/Configure` is called, the working directory is already `$(BUILD_DIR_HOST)/$(PKG_SOURCE_SUBDIR)` so it should not be necessary to repeat the path.

lang/rustup/Makefile
```diff
@@ -0,0 +1,64 @@
...
+	true
+endef
+
+define Host/Compile
+	true
+endef
+
+define Build/InstallDev
+	RUSTUP_HOME=$(HOST_RUST_DIR) CARGO_HOME=$(HOST_CARGO_DIR) $(HOST_CARGO_DIR)/bin/rustup target add $(RUSTC_TARGET_ARCH)
+endef
```
jefferyto, line 54: I don't think you can rely on `Build/InstallDev` to be called before rustc is used by a package build, e.g. if host tools are built, then the target is changed with `make menuconfig`, if this (rustup) package is not selected (and it should not be), then `Build/InstallDev` would not be run.

I think it would be better to add a pre-build hook (either `Hooks/Configure/Pre` or `Hooks/Build/Pre`) in rustc-triple.mk to add the target if necessary.

lang/rustup/rustc-triple.mk
Grommish: @jefferyto What about if (example) ./net/suricata5 needed to call it?  Just something like:

include ../../lang/rustc-triple.mk?

lang/rustup/rustc-triple.mk
```diff
@@ -0,0 +1,76 @@
...
+\
+s390x-unknown-linux-gnu \
+\
+sparc-unknown-linux-gnu \
+\
+sparc64-unknown-linux-gnu \
+\
+thumbv7neon-unknown-linux-gnueabihf thumbv7neon-unknown-linux-musleabihf \
+\
+x86_64-unknown-linux-gnu x86_64-unknown-linux-musl
```
jefferyto, line 56: Where does this list come from, and I wonder if there is a more maintainable way of organizing this list, e.g. `unknown-linux` is present in every item.

lang/rustup/rustc-triple.mk
```diff
@@ -0,0 +1,76 @@
...
+ifeq ($(CONFIG_LIBC_USE_MUSL),y)
+RUSTC_TARGET_ARCH:=$(strip $(foreach v,$(filter $(ARCH)-%, $(RUSTC_ARCH_TARGETS)), \
+  $(if $(findstring -musl,$v),$v)))
+else ifeq ($(CONFIG_LIBC_USE_GLIBC),y)
+RUSTC_TARGET_ARCH:=$(strip $(foreach v,$(filter $(ARCH)-%, $(RUSTC_ARCH_TARGETS)), \
+  $(if $(findstring -gnu,$v),$v)))
+else
+RUSTC_TARGET_ARCH:=$(strip $(foreach v,$(filter $(ARCH)-%, $(RUSTC_ARCH_TARGETS)), \
+  $(if $(findstring -uclibc,$v),$v)))
+endif
```
jefferyto, line 67: Perhaps you can save `$(filter $(ARCH)-%, $(RUSTC_ARCH_TARGETS))` to a variable so you don't have to repeat it multiple times.

lang/rustup/rustc-triple.mk
```diff
@@ -0,0 +1,76 @@
...
+
+ifeq ($(word 1, $(RUSTC_TARGET_ARCH)),)
```
jefferyto, line 69: You are testing the first word in `$(RUSTC_TARGET_ARCH)` to see if it is the empty string; why not just test if `$(RUSTC_TARGET_ARCH)` is empty?

lang/rustup/rustc-triple.mk
```diff
@@ -0,0 +1,76 @@
...
+$(info RUSTC ERROR: Unsupported or Unknown Target Triple: $(RUSTC_TARGET_ARCH))
```
jefferyto, line 70: If `$(RUSTC_TARGET_ARCH)` is empty, then printing it in the error message does nothing.

Also, I believe when the target is invalid, this message will be printed for all packages that include this file (during the preparation stage when makefiles are scanned for information), even if the packages are not selected.

Also, this does not stop the package from attempting to be built and exiting with an error. If there is a limitation on which platforms are supported or not supported, it would be better to provide a way for packages to set their `DEPENDS` value accordingly (e.g. golang-values.mk sets [`GO_ARCH_DEPENDS`](https://github.com/openwrt/packages/blob/f7041ed34ac82a26736f9046999b832e6ad672f5/lang/golang/golang-values.mk#L192) which packages can [reference](https://github.com/openwrt/packages/blob/f7041ed34ac82a26736f9046999b832e6ad672f5/net/obfs4proxy/Makefile#L35)).

lang/rustup/rustc-triple.mk
```diff
@@ -0,0 +1,76 @@
...
+endif
+
+# These are environment variables that are used by other packages to
+# define where rustc/cargo are kept.
+HOST_RUST_DIR:=$(BUILD_DIR_HOST)/.rustup
+HOST_CARGO_DIR:=$(BUILD_DIR_HOST)/.cargo
```
jefferyto, line 76: `STAGING_DIR_HOSTPKG` is a better place to store host files than `BUILD_DIR_HOST`. (`$(STAGING_DIR_HOSTPKG)/bin` is also in the `PATH` during target package build.)

It can be argued that `HOST_CARGO_DIR` would be better placed in the download or tmp folder, but `STAGING_DIR_HOSTPKG` would still be better than `BUILD_DIR_HOST`.

jefferyto (commented):
If my understanding is correct, rustup downloads and installs prebuilt binaries. I don't really like this approach because AFAIK:
* The files it downloads are not cached (offline installation is also not possible)
* It doesn't validate any signatures of the files it downloads
* We have to trust that the binaries they built are safe

lang/rustup/rustc-triple.mk
jefferyto: Yes that should work.

lang/rustup/rustc-triple.mk
jefferyto: Also, it seems that this list of targets is used for two purposes, to test if the target platform is supported, and to have a target to pass to rustc.

My previous comment (organizing the list in a more maintainable way) goes towards testing if the target platform is supported.

To have a target to pass to rustc, I think it would be nice if [`GNU_TARGET_NAME` or `REAL_GNU_TARGET_NAME`](https://github.com/openwrt/openwrt/blob/bcf0704bd2ca852b1cbfa481095afb73e78287f9/rules.mk#L125-L126) can be used instead (although I have no idea if rustc will accept it or not).

lang/rustup/Makefile
jefferyto: On second thought, since this installs a `rustup` tool, perhaps the name is appropriate.

Grommish:
> If my understanding is correct, rustup downloads and installs prebuilt binaries. I don't really like this approach because AFAIK:
> 
>     * The files it downloads are not cached (offline installation is also not possible)
> 
>     * It doesn't validate any signatures of the files it downloads
> 
>     * We have to trust that the binaries they built are safe

I'm all for building from source, when possible.  Rustup does validate signatures, afaik, but yes, if there is a better way, I'm all for it

lang/rustup/rustc-triple.mk
Grommish: This list comes from ./rustup.sh --list-targets and has been trimmed to only include MUSL, Glibc and uclibc targets.

At least in the Suricata example, the rustc "triple" (totally misnamed these days) has to be passed in the CONFIGURE_ARGS as a --target=$(RUSTC_TARGET_ARCH) so it knows it's cross-compiling.  The triple is based on the above list.  So, yes, it does do dual duty of not only seeing what the target arch is based on the CONFIG_ flags, but then passes that onto whatever is going to use rustc/cargo to build with.  

Originally, i was looking for a way to just export RUSTC_TARGET_ARCH to the build system so I could pull it, but since that didn't work, I ended up going with the rustc-triple.mk file.

lang/rustup/Makefile
Grommish: the other packages that use rustup would need a PKG_HOST_DEPENDS/PKG_BUILD_DEPENDS for rustup/host.  Would that cause a race against Build/InstallDev?

detly:
I would think the appropriate build-from-source method would be to compile Rust toolchain + Cargo as host packages, and use Cargo for all other rust packages. I'm not sure how bootstrapping would work though, since rustc is in Rust.

lang/rustup/Makefile
jefferyto: `Build/InstallDev` is part of the target package build (for this package). `PKG_BUILD_DEPENDS=rustup/host` is host package build. (There is no `PKG_HOST_DEPENDS`.) Host and target package builds take place independently of each other (unless the target package depends on the host package).

jefferyto:
I would try looking at / replicating https://github.com/dtolnay/bootstrap.

> Be aware that the complete bootstrap uses a lot of disk space. Bootstrapping up through rustc 1.40 requires 120 GB.

This is a concern, but perhaps after compiling a version X+1, version X can be discarded, this might save disk space. This process might take a long time though.

I would also suggest looking at how other distros are handling this, although I'm not sure they are doing a complete bootstrap from source either.

detly:
This is also an interesting issue: [Rust #48707](https://github.com/rust-lang/rust/issues/48707) mainly because of the links to other info on bootstrapping strategies (including the mrustc approach @jefferyto posted).

Grommish:
I'm looking at the mrustc package.  I think I can shrink the footprint.  It creates every version of rustc from 1.29 thru 1.45, and probably doesn't need to.  However, it also relies on a debian-type system to work and tries to setup a sandbox vm to build in, which may or may not be a deal-breaker.  I don't seem to need it under Ubuntu, but not everyone will be using ubuntu to build with.  I'll report back when i see what is actually needed

lang/rustup/rustc-triple.mk
Grommish: GNU_TARGET_NAME/GNU_HOST_NAME triples do not seem to be compat with what the rust triples are.

lang/rustup/Makefile
Grommish: PKG_HOST_ONLY it is.. BUILDONLY I don't believe will call Compile/Install

Grommish:
Since mrustc is not an option, and the current rustc/cargo source code not only downloads a pre-compiled binary to compile the latest version, but also has an issue that keeps it from being usable (https://github.com/rust-lang/rust/issues/75841) in it's present form.  Because of this, I'm requesting to push on with rustup.

Grommish:
rustup in now abandoned, as it is ill-equipped to deal with the OpenWrt environment. 

New PR for Rust-from-source is https://github.com/openwrt/packages/pull/13916

Grommish:
For anyone in the future finding this because a search for `openwrt` and `rustup`, let me save you some time.

As of this comment, `rustup` is not `-msoft-float`, and there for cannot be used by Openwrt MUSL.

Check the status of https://github.com/openwrt/packages/pull/13916 for status on rust lang being available in OpenWrt.

Issue #12972, opened 2020-07-30
mwan3: Multi-wan copying route problem. 


CsHeng:
Maintainer: @feckert 
Environment: [OpenWrt 19.07.3 r11063-85e04e9f46](https://downloads.openwrt.org/releases/19.07.3/targets/x86/64/)
mwan3: [mwan3_2.8.12-1_all.ipk](https://downloads.openwrt.org/releases/19.07.3/packages/x86_64/packages/mwan3_2.8.12-1_all.ipk)
Description:

Multi-wan copying route problem. 

by checking the [mwan3.sh source file](https://github.com/openwrt/packages/blob/facf8ea299b185a82ae803d68e2b54a9599d8c14/net/mwan3/files/lib/mwan3/mwan3.sh):

**i found `$IP4 route add table $tid $line` will only add the first line in `ip route list table main` with same subnet/proto kernel**

related mwan3 source code below
```bash
# return true(=0) if has any mwan3 interface enabled
# otherwise return false
mwan3_rtmon_ipv4()
{
	local tid=1
	local idx=0
	local ret=1
	local tbl=""
	mkdir -p /tmp/mwan3rtmon
	($IP4 route list table main  | grep -v "^default\|linkdown" | sort -n; echo empty fixup) >/tmp/mwan3rtmon/ipv4.main
	while uci get mwan3.@interface[$idx] >/dev/null 2>&1 ; do
		idx=$((idx+1))
		tid=$idx
		[ "$(uci get mwan3.@interface[$((idx-1))].family)" = "ipv4" ] && {
			tbl=$($IP4 route list table $tid 2>/dev/null)
			if echo "$tbl" | grep -q ^default; then
				(echo "$tbl"  | grep -v "^default\|linkdown" | sort -n; echo empty fixup) >/tmp/mwan3rtmon/ipv4.$tid
				cat /tmp/mwan3rtmon/ipv4.$tid | grep -v -x -F -f /tmp/mwan3rtmon/ipv4.main | while read line; do
					$IP4 route del table $tid $line
				done
				cat /tmp/mwan3rtmon/ipv4.main | grep -v -x -F -f /tmp/mwan3rtmon/ipv4.$tid | while read line; do
					$IP4 route add table $tid $line
				done
			fi
		}
		if [ "$(uci get mwan3.@interface[$((idx-1))].enabled)" = "1" ]; then
			ret=0
		fi
	done
	rm -f /tmp/mwan3rtmon/ipv4.*
	return $ret
[...]

Issue #12924, opened 2020-07-25
mwan3: rtmon_ip4/6 functions use uncommited uci changes


aaronjg:
Maintainer: @feckert
Environment: 19.07.3

Description:

mwan3_rtmon_ip4/6 functions call `uci get` which returns uncommitted changes staged using uci. These should 'config_get' functions to get only committed changes. 

I'm working on a fix and will have a P/R in the next few days.

feckert:
The problem her is, if we use `config_get` in `mwan3_rtmon_ip4/6` then we have to load the whole config with `config_load mwan3` every time.
This is very time and memory intensive.

I see her two possibilities to achieve this:

- Add the variable on rtmon [start](https://github.com/openwrt/packages/blob/master/net/mwan3/files/lib/mwan3/mwan3.sh#L667)

or

- Before the rtmon [process](https://github.com/openwrt/packages/blob/master/net/mwan3/files/usr/sbin/mwan3rtmon#L24) loop starts. 

I think the second one is the best, because we to an `config_load mwan3` already.

aaronjg:
It looks like the config_load takes about 80ms, which is not great, but not terrible. On the other hand, the calls to `uci`, which we are currently doing, are even slower.

I have a fix for this ready here:
https://github.com/aaronjg/openwrt-packages/tree/mwan3-rtmon-uci-fix

Time to run an `mwan3 restart` is comparable 

new code
```
real    0m 22.97s
user    0m 9.75s
sys     0m 4.61s
```

old code
```
real    0m 23.35s
user    0m 9.69s
sys     0m 4.87s
```

performance checking of `uci get` vs. config_load.
```shell
~ # cat test_config_load.sh; time ./test_config_load.sh 
#!/bin/sh
. /lib/functions.sh
for i in `seq 1 100`; do
        config_load mwan3
        for section in ${CONFIG_SECTIONS}; do
                config_get cfgtype "$section" TYPE
                [ "$cfgtype" != "interface" ] && continue
                tid=$((tid+1))
                config_get family "$section" family ipv4
                config_get enabled "$section" enabled 0
        done
done
real    0m 7.73s
user    0m 6.62s
sys     0m 1.00s

```


```shell

~ # cat test_config_load2.sh; time ./test_config_load2.sh 
#!/bin/sh
for i in `seq 1 100`; do
        idx=0
        while uci get mwan3.@interface[$idx] >/dev/null 2>&1 ; do
                idx=$((idx+1))
[...]

PR #13041, opened 2020-08-06
mwan3: startup performance improvements and other fixes

aaronjg:
Maintainer: @feckert 
Compile tested: Scripts only
Run tested: 19.0.7.3

Description:

Improve startup and runtime performance by

1) moving common startup procedures out of hotplug script when called
from mwan3 start
2) reducing calls to iptables to check status of rules
3) consolidating iptables updates and updating with iptables-restore
4) do not wait for kill if nothing was killed
5) running interface hotplug scripts in parallel
6) eliminate operations in hotplug script that check status on every single interface unnecessarily
7) consolidate how mwan3track makes hotplug calls
8) do not restart mwan3track on connected events (in addition to the speedup, cutting out the mwan3track restarts is going to make it easier to move to true procd compliant service model).

This is a significant refactor, but should not result in any breaking
changes or require users to update their configurations.

Also 
- use functions.sh functions rather than uci command line tool to find interfaces for routing table.
- use ip monitor route to detect routing changes rather than polling the routing table

Fixes https://github.com/openwrt/packages/issues/12924, and, as part of the rtmon cleanup, fixes https://github.com/openwrt/packages/issues/12972

I have been running this fine for a few days, and it is working well for my configuration. Startup time is reduced from ~25s to ~5s. 

Testing with additional configurations would be very helpful.

cc @brianjmurrell @jamesmacwhite


feckert:
@aaronjg thanks for your work :+1:

I'll check out the next week.

ptpt52:
Looks great improvement

net/mwan3/files/etc/hotplug.d/iface/15-mwan3
```diff
@@ -4,95 +4,93 @@
...
+		if [ -n "${trackpid}" ]; then
+                        device_pid=$(pgrep -f "mwan3track $INTERFACE $DEVICE")
+                        if [ "$device_pid" = "$trackpid" ]; then
+				[ "$ACTION" = ifup ] && kill -USR2 "$trackpid"
+                        else
+            			mwan3_track $INTERFACE $DEVICE "$binary_status" "$src_ip"
+				LOG notice "Restarted tracker [$!] on interface $INTERFACE (${DEVICE:-unknown})"
+                        fi
+                else
+			mwan3_track $INTERFACE $DEVICE "$binary_status" "$src_ip"
```
ptpt52, line 79: whitespace and tab indent mixed here

ptpt52:
May squash commits into one?

ptpt52:
and many indentation issues

ptpt52:
I found that `ip monitor` could not handle `ip change`
for example
```
ip route add 1.2.3.4 via 192.168.1.1
ip change 1.2.3.4 via 192.168.2.1

ip monitor route output:
1.2.3.4 via 192.168.1.1 dev br-lan
1.2.3.4 via 192.168.2.1 dev eth0.2
```

aaronjg:
> I found that `ip monitor` could not handle `ip change` or `ip replace`
> for example
> 
> ```
> ip route add 1.2.3.4 via 192.168.1.1
> ip change 1.2.3.4 via 192.168.2.1
> 
> ip monitor route output:
> 1.2.3.4 via 192.168.1.1 dev br-lan
> 1.2.3.4 via 192.168.2.1 dev eth0.2
> ```



> I found that `ip monitor` could not handle `ip change` or `ip replace`

Excellent observation. Thank you. I will look into what the best way to handle this is.

aaronjg:
> Looks great improvement
Thank you.

> May squash commits into one?
As this is WIP, it is better to keep the commits separate. However, it can be squashed into one before merging.

Change and replace events are now handled properly: d9dc527

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -39,85 +39,50 @@ NO_IPV6=$?
 
 # return true(=0) if has any mwan3 interface enabled
 # otherwise return false
-mwan3_rtmon_ipv4()
+mwan3_rtmon_route()
 {
-	local idx=0
+	config_load mwan3
```
feckert, line 61: We have two config_load mwan3 calls on this line and this https://github.com/openwrt/packages/pull/13041/commits/bbe32d3473207764ea7f6d5c51d0f1024bdfee33

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -39,85 +39,50 @@ NO_IPV6=$?
...
-
-		family="$(uci -q get mwan3.@interface[$idx].family)"
-		[ -z "$family" ] && family="ipv4"
-
-		enabled="$(uci -q get mwan3.@interface[$idx].enabled)"
-		[ -z "$enabled" ] && enabled="0"
+	[ $NO_IPV6 -eq 0 ] && ($IP6 route list table main  | grep -v "^default\|^::/0\|^fe80::/64\|^unreachable" | sort -n; echo empty fixup) >/tmp/mwan3rtmon/ipv6.main
+	config load mwan3
+	for section in ${CONFIG_SECTIONS}; do
+		config_get cfgtype "$section" TYPE
```
feckert: Why is the vairalbe TYPE written in capital letters

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -39,85 +39,50 @@ NO_IPV6=$?
...
-		tid=$((idx+1))
-
-		family="$(uci -q get mwan3.@interface[$idx].family)"
-		[ -z "$family" ] && family="ipv4"
-
-		enabled="$(uci -q get mwan3.@interface[$idx].enabled)"
-		[ -z "$enabled" ] && enabled="0"
+	[ $NO_IPV6 -eq 0 ] && ($IP6 route list table main  | grep -v "^default\|^::/0\|^fe80::/64\|^unreachable" | sort -n; echo empty fixup) >/tmp/mwan3rtmon/ipv6.main
+	config load mwan3
+	for section in ${CONFIG_SECTIONS}; do
```
feckert: We could move this for section into a function and could then do a config_foreach call.
See https://github.com/openwrt/packages/blob/master/net/mwan3/files/lib/mwan3/mwan3.sh#L217

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -613,12 +639,14 @@ mwan3_delete_iface_ipset_entries()
 
 mwan3_rtmon()
 {
-	pid="$(pgrep -f mwan3rtmon)"
-	if [ "${pid}" != "" ]; then
-		kill -USR1 "${pid}"
-	else
-		[ -x /usr/sbin/mwan3rtmon ] && /usr/sbin/mwan3rtmon &
-	fi
+	local protocal
```
feckert: typo

net/mwan3/files/usr/sbin/mwan3rtmon
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-
+set -o pipefail
```
feckert: This is only for debug?

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -195,17 +247,39 @@ mwan3_init()
...
+	config_get family $INTERFACE family ipv4
+	if [ "$family" = "ipv4" ]; then
+		network_get_ipaddr _src_ip ${INTERFACE}
+		[ -n "$_src_ip" ] || _src_ip="0.0.0.0"
+	elif [ "$family" = "ipv6" ]; then
+		network_get_ipaddr6 _src_ip ${INTERFACE}
+		[ -n "$_src_ip" ] || _src_ip="::"
+	fi
+	export "$1=$_src_ip"
+}
```
feckert, line 292: new line

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -36,15 +38,32 @@ MM_UNREACHABLE=""
...
+	local facility=$1; shift
+	# in development, we want to show 'debug' level logs
+	# when this release is out of beta, the comment in the line below
+	# should be removed
+	# [ "$facility" = "debug" ] && continue
+	logger -t "$SCRIPTNAME[$$]" -p $facility "$*"
+}
+
+
+mwan3_push_update()
```
feckert, line 53: i do not know this syntax. Can you please tell me exactly what you're doing here.
current and update?

aaronjg: Sure. It's a poor-man's stringbuilder. Rather than making a bunch of separate `iptables` calls, it is faster to create everything in an update variable and then do one `iptables restore` call per function.  The syntax looks a bit weird here because I need to put in a new line after each update, and (as far as I can tell) `ash` doesn't allow escaped new lines in string literals, so there is an actual new line there. I can add a comment there to make things clearer.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: fixed in 169982d. We can squash everything into one commit before merging.

net/mwan3/files/usr/sbin/mwan3rtmon
aaronjg: Removed in 3543df28e. We can squash all the commits before merging.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: Fixed in 3543df28e.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: yes, thank you.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: good catch. Thank you.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: Removed one of them here: 3543df28e

feckert:
Which commits would you like to squash?

aaronjg:
> Which commits would you like to squash?

potentially everything in this P/R, but if the `ip route monitor` stuff looks okay, then at least: bbe32d3473207764ea7f6d5c51d0f1024bdfee33 db425b756d031b6b0e9800636e5fba6660ca92c8 23f0361eca81a32cdfe8a3c0d22aae05f9a9e024 3543df28e15987b329705c179f63f3cd5d351635.

I first got rid of the `uci` calls, but then it was a refactor to use the `ip route monitor,` so if the `ip route monitor` stuff looks okay to you, then we don't need to keep the `uci` removal stuff around as a separate commit.

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -352,6 +352,46 @@ mwan3_set_custom_ipset()
...
+	local has_ipv4=0
+	for dir in src dest; do
+		config_get ips $rule ${dir}_ip
+		if [ -z "$ips" ] || [ -n "${ips##*,*}" ]; then
+			continue
+		fi
+		if [ "$family" = any ]; then
+			mwan3_push_update create mwan3_rule_${rule}_${dir} list:set
+			mwan3_push_update create mwan3_rule_${rule}_${dir}_ipv4 hash:ip
+			mwan3_push_update create mwan3_rule_${rule}_${dir}_ipv6 hash:ip family inet6
```
ptpt52, line 372: if use ipset, src ip may be `ip/cidr`, so we may use hash:net
and please set the ipset hash size, default is 1024, may eat memory size.

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -352,6 +352,46 @@ mwan3_set_custom_ipset()
...
 
+mwan3_load_ipset_rule()
+{
+	local family
+	local rule=$1
+
+	config_get family $rule family 'any'
+	local dir ips
+	local has_ipv6=0
+	local has_ipv4=0
```
ptpt52, line 363: what these vars(`has_ipv6/has_ipv4`) for?

aaronjg: thank you. These are not needed.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: How much memory do you think a 1024 hash uses?  I changed this (and elsewhere in the codebase that uses hash:net), but the overall memory footprint will be unnoticeable. I saw elsewhere you are using luci - if you are on a very restricted memory device, you may consider turning off uhttpd, which will save a fair bit of RAM.

net/mwan3/files/usr/sbin/mwan3
```diff
@@ -81,8 +81,16 @@ ifup()
 	if [ "$up" = "1" ] \
 		&& [ -n "$l3_device" ] \
 		&& [ "$enabled" = "1" ]; then
-		ACTION=ifup INTERFACE=$1 DEVICE=$l3_device /sbin/hotplug-call iface
```
feckert, line 84: Sorry this does not work! I want to have this.
All hotplug script must be called under /etc/hotplug.d/iface I have additional scripts there that should also get notified.

net/mwan3/files/usr/sbin/mwan3rtmon
```diff
@@ -1,38 +1,28 @@
...
+	local IP family
 
 	config_load mwan3
-	config_get rtmon_interval globals rtmon_interval '5'
-
-	sleep 3
-	while true; do
+	family=$1
+	[ -z $family ] && family=ipv4
+	if [ "$family" == ipv6 ]; then
```
feckert, line 13: Thats a string check? so `=` and `"ipv6"`

net/mwan3/files/usr/sbin/mwan3
feckert: We could move MWAN3_STARTUP into **/var/run/mwan3/startup** ? so we do not needed to change the hotplug.d call

aaronjg: @feckert Can you describe your setup a bit more? My problem was that I had a lot of stuff in the hotplug directory that shouldn't have been called on an `mwan3 start`.  

Connecting an interface for the first time requires a different response than initiating mwan3. So it didn't really make sense to me to cause a full ifup hotplug event when a user runs '/init.d/mwan3/restart.'  For example, I have a hotplug script that brings up my VPN when the router connects to the 4G modem and gets an ifup event. With the current configuration, it will try to restart the VPN when mwan3 is restarted. This slowed down the mwan3 startup process and seemed unnecessary.

>We could also add this check to the ubus backend.

I don't understand, could you explain a bit more?

> We could move MWAN3_STARTUP into /var/run/mwan3/startup ?

Could you explain a bit more? It would be nice to avoid the MWAN3_STARTUP semaphore if possible.

net/mwan3/files/usr/sbin/mwan3rtmon
aaronjg: yes, thank you :)

net/mwan3/files/usr/sbin/mwan3
feckert: > @feckert Can you describe your setup a bit more? My problem was that I had a lot of stuff in the hotplug directory that shouldn't have been called on an `mwan3 start`.

Okay, now I get it. I missed you explaining why you wanted to do it that way.

> Connecting an interface for the first time requires a different response than initiating mwan3. So it didn't really make sense to me to cause a full ifup hotplug event when a user runs '/init.d/mwan3/restart.' For example, I have a hotplug script that brings up my VPN when the router connects to the 4G modem and gets an ifup event. With the current configuration, it will try to restart the VPN when mwan3 is restarted. This slowed down the mwan3 startup process and seemed unnecessary.

Sounds reasonable. Fits from my side.

net/mwan3/files/usr/sbin/mwan3
```diff
@@ -81,8 +81,16 @@ ifup()
...
+		MWAN3_STARTUP=$MWAN3_STARTUP ACTION=ifup INTERFACE=$1 DEVICE=$l3_device sh /etc/hotplug.d/iface/15-mwan3
+		MWAN3_STARTUP=$MWAN3_STARTUP ACTION=ifup INTERFACE=$1 DEVICE=$l3_device sh /etc/hotplug.d/iface/16-mwan3-user
+	fi &
```
feckert, line 86: Only this `&`  I think it's weird.
I know you want to put this in the background.
But it's easy to miss!
Does we have for this another solution?
Move this in a function?
mwan3_startup &

aaronjg: Yes, that would be much clearer. I'll work on that change this evening.

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -37,89 +38,163 @@ MM_UNREACHABLE=""
...
+mwan3_push_update()
+{
+	# helper function to build an update string to pass on to
+	# IPTR or IPS RESTORE. Modifies the 'update' variable in
+	# the local scope.
+	update="$update
+$*";
+}
+
+mwan3_rtmon_route_handle()
```
feckert, line 62: Should we not move this line to rtmon?

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -37,89 +38,163 @@ MM_UNREACHABLE=""
 command -v ip6tables > /dev/null
 NO_IPV6=$?
 
-# return true(=0) if has any mwan3 interface enabled
-# otherwise return false
-mwan3_rtmon_ipv4()
+SCRIPTNAME="$(basename "$0")"
+LOG()
```
feckert, line 42: We could also move the LOG function to common.sh.
So we have them in each mwan3 script if we include /lib/mwan3/common.sh

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: Yes, better to put it there. Will fix this evening.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: Yes, good idea. Then we don't need to redefine it in mwan3track.

aaronjg:
@feckert per your suggestion, I removed the ipset improvements, and put them in their own branch.  

https://github.com/aaronjg/openwrt-packages/tree/mwan3-ipset

feckert:
@aaronjg could you please rebase this P/R There's something wrong

net/mwan3/files/lib/mwan3/mwan3.sh
ptpt52: remove it(mwan3_rtmon_route_handle) from mwan3.sh? now dup

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -37,89 +38,164 @@ MM_UNREACHABLE=""
...
+
+	update_table()
+	{
+		let _tid++
+		config_get family "$section" family ipv4
+		network_get_device device "$section"
+		[ -z "$device" ] && continue
+		config_get enabled "$section" enabled
+		[ "$enabled" -eq 0 ] && continue
+		curr_table=$(eval "echo	 \"\$mwan3_dev_tbl_${family}\"")
```
ptpt52, line 144: here use `continue` or `return` ?

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -37,89 +38,164 @@ MM_UNREACHABLE=""
...
+{
+	local _tid section family cfgtype curr_table _mwan3_iface_tbl
+	mwan3_iface_tbl=" "
+	update_table()
+	{
+		let _tid++
+		export mwan3_iface_tbl="${mwan3_iface_tbl}${1}=$_tid "
+	}
+	config_foreach update_table interface
+}
```
ptpt52, line 161: `local _tid section family cfgtype curr_table _mwan3_iface_tbl`
many unused vars?

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -506,44 +617,110 @@ mwan3_delete_iface_iptables()
...
 {
-	local id via metric V V_ IP
+	local id via metric V V_ IP family
+	local iface device cmd true_iface
 
-	config_get family "$1" family ipv4
-	mwan3_get_iface_id id "$1"
+	iface=$1
+	device=$2
+	config_get family "$iface" family ipv4 device
```
ptpt52, line 625: typo ` device` here?

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -506,44 +617,110 @@ mwan3_delete_iface_iptables()
...
+	     dev $2"
+	$cmd || LOG warn "ip cmd failed $cmd"
+
+}
+
+mwan3_add_non_default_iface_route()
+{
+	local tid route_line family IP family id
+	config_get family "$1" family ipv4 device
+	mwan3_get_iface_id id "$1"
```
ptpt52, line 660: dup `family` vars?
and typo `device`

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -195,17 +271,40 @@ mwan3_init()
...
 mwan3_get_iface_id()
 {
 	local _tmp _iface _iface_count
+	[ -z "$mwan3_iface_tbl" ] && mwan3_update_iface_to_table
+	_tmp="${mwan3_iface_tbl##* ${2}=}"
+	_tmp=${_tmp%% *}
+	export "$1=$_tmp"
+	new_val=$_tmp
+	unset _tmp
+	return
```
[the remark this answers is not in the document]
aaronjg: thank you. I was testing out a better implementation. Old version after return is no longer needed.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: yes, thank you for the all comments. Just pushed fixes.

aaronjg:
Thank you both for the helpful comments.  I just pushed fixes, as well as a fix for race condition that came up with mwan3's faster startup on device start.

net/mwan3/files/lib/mwan3/mwan3.sh
ptpt52: this line
```
config_get family "$iface" family ipv4 device
```
should remove the `device`?

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -37,89 +38,85 @@ MM_UNREACHABLE=""
...
+	ubus call "network.interface.${iface}_${V}" status &>/dev/null && _true_iface="${iface}_${V}"
+	export "$1=$_true_iface"
+}
+
+mwan3_route_line_dev()
+{
+	# must have mwan3 config already loaded
+	# arg 1 is route device
+	local section enabled device route_line route_device
+	local family route_family
```
ptpt52, line 103: many unused vars

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -506,44 +526,110 @@ mwan3_delete_iface_iptables()
...
+		[ "$family" != "$ipv" ] && return
+		$IP route list table $tid 2>/dev/null | grep -q "^default\|^::/0" && {
+			active_tbls="$active_tbls${tid} "
+		}
+	}
+
+	add_route()
+	{
+		let tid++
+		config_get family "$section" family ipv4
```
ptpt52, line 606: this line unused
`config_get family "$section" family ipv4`

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -834,189 +914,230 @@ mwan3_set_policies_iptables()
...
+	fi
 
-				for IPT in "$IPT4" "$IPT6"; do
-					[ "$IPT" = "$IPT6" ] && [ $NO_IPV6 -ne 0 ] && continue
-					if ! $IPT -S "$policy" &> /dev/null; then
-						$IPT -N "$policy"
-					fi
+	[ "$ipv" = ipv6 ] && [ $NO_IPV6 -ne 0 ] && return
+	[ "$family" = "ipv4" ] && [ "$ipv" = "ipv6" ] && return
+	[ "$family" = "ipv6" ] && [ "$ipv" = "ipv4" ] && return
```
ptpt52, line 1022: what is `ipv` here? unset?

ptpt52: sorry I misunderstood here

feckert: @ptpt52 You got that right. I don't understand it either at the first place.
I can't find the variable `$ipv`.
But it is a global Variable if so we should capitalize them but I do not like this.
I think the best is to set it as an argument to the function call.
@aaronjg Could we please pass this variable as an function argument.
So we have
`config_foreach mwan3_set_user_iptables_rule rule "$ipv"`
And this is the first argument of the function
```
mwan3_set_user_iptables_rule() {
local ipv="$1"
```

Also you forgot the quotation marks in `ipv6` check in this line with `$ipv`
so we have `[ "$ipv" = "ipv6" ] && [ $NO_IPV6 -ne 0 ] && return`

feckert:
@aaronjg 
Please add the variables as arguments after the section when calling the function `config_foreach`.
`config foreach function_name2 uci_section "$variable1" "$variable"`
So we have in the function call the following.
```
 function_name() {
local uci_section_name="$1"
local variable1="$2"
local variable2="$3"
```
This makes reading the code easier and we don't have to work with global variables.
The are a lot of examples in this package feed.

aaronjg:
>  This makes reading the code easier and we don't have to work with global variables.

Thank, you. That is much cleaner. I pushed an update.

net/mwan3/files/lib/mwan3/mwan3.sh
ptpt52: @aaronjg  hi, forget to resolve here?

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -506,44 +523,109 @@ mwan3_delete_iface_iptables()
...
 		V_=""
 		IP="$IP4"
 	elif [ "$family" = "ipv6" ]; then
-		V=6
 		V_=6
 		IP="$IP6"
-	else
-		return
 	fi
+	mwan3_get_true_iface true_iface $iface
```
ptpt52, line 544: call `mwan3_get_true_iface` twice here?

net/mwan3/files/usr/sbin/mwan3
```diff
@@ -37,52 +38,64 @@ ifdown()
...
+
+		[ ${enabled} -gt 0 ] || {
+			echo "The service mwan3 is global disabled."
+			echo "Please execute \"/etc/init.d/mwan3 start\" first."
+			exit 1
+		}
+	else
+		enabled=1
+	fi
+	mwan3_get_true_iface true_iface $interface
```
ptpt52, line 75: it seems call `mwan3_get_true_iface` without `config_load mwan3`

net/mwan3/files/lib/mwan3/mwan3.sh
ptpt52: ok, no problem, config_load before call `ifup()`

net/mwan3/files/usr/sbin/mwan3
aaronjg: config_load mwan3 will be called on line 64 or on line 158

ptpt52: this ok, no problem, config_load before call ifup()

net/mwan3/files/lib/mwan3/common.sh
```diff
@@ -4,3 +4,14 @@ get_uptime() {
...
 }
+
+SCRIPTNAME="$(basename "$0")"
+LOG()
+{
+	local facility=$1; shift
+	# in development, we want to show 'debug' level logs
+	# when this release is out of beta, the comment in the line below
+	# should be removed
+	[ "$facility" = "debug" ] && continue
```
ptpt52, line 15: here Please use `return`, not `continue`

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -195,36 +189,47 @@ mwan3_init()
...
+	INTERFACE=$2
+	unset "$1"
+	config_get family "$INTERFACE" family ipv4
+	if [ "$family" = "ipv4" ]; then
+		network_get_ipaddr _src_ip "$INTERFACE"
+		[ -n "$_src_ip" ] || _src_ip="0.0.0.0"
+	elif [ "$family" = "ipv6" ]; then
+		network_get_ipaddr6 _src_ip "$INTERFACE"
+		[ -n "$_src_ip" ] || _src_ip="::"
+	fi
```
ptpt52, line 212: here
```
network_get_ipaddr or network_get_ipaddr6
```
we must use true_iface of `"$INTERFACE"`
or we may get the wrong ip

ptpt52: maybe use `mwan3_get_true_iface` to fetch true_iface and then use it.

net/mwan3/files/usr/sbin/mwan3
```diff
@@ -37,52 +38,64 @@ ifdown()
...
 
-	status=$(ubus -S call network.interface.$1 status)
 	[ -n "$status" ] && {
 		json_load "$status"
 		json_get_vars up l3_device
 	}
+	hotplug_startup()
+	{
+		MWAN3_STARTUP=$MWAN3_STARTUP ACTION=ifup INTERFACE=$true_iface DEVICE=$l3_device sh /etc/hotplug.d/iface/15-mwan3
+		MWAN3_STARTUP=$MWAN3_STARTUP ACTION=ifup INTERFACE=$true_iface DEVICE=$l3_device sh /etc/hotplug.d/iface/16-mwan3-user
```
ptpt52, line 85: here
`INTERFACE=$true_iface`
should be "$interface" ?

aaronjg: Yes. I think you are correct here. The current implementation is actually a bit broken. If the rule is defined on e.g. 'WAN' but the actual interface is named 'WAN_6' then when 'WAN_6' has a hotplug event the mwan3 hotplug script will not run. If WAN has a hotplug event and WAN_6 does not, the hotplug event will run unnecessarily.

It actually can be even worse than that, if a user has WAN and WAN_4 that are different interfaces (not sure why, maybe they have WAN, WAN_2, WAN_3 and WAN_4?), then mwan3 will return the ip address for WAN_4 for the WAN interface. To be clear, this is already broken in the current version, and not a new issue with this P/R. I made an issue for it, but I think fixing that is beyond the scope of this P/R. I made an issue for it here #13133 

Best practice is to have the mwan3 interface name match the openwrt interface name. Still, it is bad user experience to break things, even if the features are undocumented. This P/R is focusing on performance improvements - having the extra `ubus` calls has a very minor performance hit, but it is probably worth the performance penalty to not break existing configurations.

Priorities for P/Rs should be:
1) Not introducing new bugs for people running things set up following the documentation
2) Fixing existing bugs
3) Adding new features and improvements
4) Not breaking undocumented behavior that people could be using

ptpt52: ok for now.

net/mwan3/files/usr/sbin/mwan3
```diff
@@ -37,52 +38,64 @@ ifdown()
...
 
-	status=$(ubus -S call network.interface.$1 status)
 	[ -n "$status" ] && {
 		json_load "$status"
 		json_get_vars up l3_device
 	}
+	hotplug_startup()
+	{
+		MWAN3_STARTUP=$MWAN3_STARTUP ACTION=ifup INTERFACE=$INTERFACE DEVICE=$l3_device TRUE_INTERFACE=$true_iface sh /etc/hotplug.d/iface/15-mwan3
+		MWAN3_STARTUP=$MWAN3_STARTUP ACTION=ifup INTERFACE=$INTERFACE DEVICE=$l3_device TRUE_INTERFACE=$true_iface sh /etc/hotplug.d/iface/16-mwan3-user
```
ptpt52, line 85: should be `INTERFACE=$interface` ?

interface is local var and set by `interface=$1` in line 59

aaronjg: yes. ty

ptpt52:
great, looks all ok.

feckert:
@aaronjg What did you changed in your last force push?

aaronjg:
Create the routes on "connected" events rather than just "ifup" events

feckert:
@aaronjg thanks again. Great job :+1:
A did some test and noticed that if I have one interface then I have 4 rtmon service running.
```
][root@G3400-V ~ # ps | grep mwan3
26741 root      1300 S    /bin/sh /usr/sbin/mwan3track wan eth0 offline 172.16.2.171 8.8.8.8 1.1.1.1 9.9.9.9 update.tdt.de
26832 root      1496 S    /bin/sh /usr/sbin/mwan3rtmon ipv4
26834 root      1496 S    /bin/sh /usr/sbin/mwan3rtmon ipv6
26865 root      1496 S    /bin/sh /usr/sbin/mwan3rtmon ipv4
26880 root      1496 S    /bin/sh /usr/sbin/mwan3rtmon ipv6
27164 root      1136 S    grep mwan3
```
I don't think that's okay, is it?
Only two should run for all interfaces one for ipv4 and the ipv6 no matter how much interface I add to mwan3?

ptpt52:
Because pipe is used, there will be two processes for each (ipv4/ipv6)
It is ok, I think.

feckert:
Looks great.
Many thanks for this excelend improvment.

Issue #11826, opened 2020-04-11
mwan3: Does not respect "compile with(out) IPV6 support"


anonymous-one:
Looks like mwan3 doesn't respect if OpenWRT has been compiled with the enabled / disable IPV6 build setting.

I have a custom OpenWRT build with IPV6 disabled (via Build Options in make menuconfig), and this is what I get when getting mwan3 status:

```
/usr/sbin/mwan3: line 1214: ip6tables: not found
```

A simple check if iptables6 exists would suffice.

Although this is completly fine as thus far I have only seen it pop up on mwan3 status... It does full the log with the ip6tables not found message when viewing via LuCI.

feckert:
That's right. It wasn't supposed to happen like this until now.
If you have the time and inclination, you can submit a pullrequest that will fix the problem. I have it on my list but due to time constraints I will not be able to implement it any time soon.

feckert:
This was solved by @aaronjg (thanks for the work).
Please try the latest version 2.8.12

- [master](https://github.com/openwrt/packages/commit/a0d66d4eebefe6e89c582df2b4dc972ea7c5c7f5#diff-34590d11545c84b961d1992ce3aefdeb) 
- [openwrt-19.07](https://github.com/openwrt/packages/commit/dc25b38f1317ad5a849a6ef12b46f4edfe53a7e4#diff-34590d11545c84b961d1992ce3aefdeb
)

PR #13048, opened 2020-08-07
[WIP] mwan3:  dualstack and ipv4 or ipv6 single support

ptpt52:
Maintainer: me / @feckert 
Compile tested: ipq40xx
Run tested: ipq40xx

Description:

When the ipv4 interface or ipv6 interface does not exist, we do not set ipv4 or ipv6 rules
what if not, if we setup mwan3 for ipv4 only, ipv6 would be blocked, which is not expected.

Suppose we have two wans, we set up mwan3 ipv4 load balancing, and ipv4 works as expected. At this time, ipv6 did not do any mwan3 settings. We expected it (ipv6) to be able to connect to the Internet at least, but the result did not. Because mwan3 sets ip rule and others, ipv6 is blocked.

Also I make `mwan3track` work with `gateway` as `track_ip`, so we could set `gateway` as tracking ip

### DIFF
--- a/net/mwan3/files/etc/hotplug.d/iface/15-mwan3
+++ b/net/mwan3/files/etc/hotplug.d/iface/15-mwan3
@@ -17,6 +17,7 @@ fi
 [ "$MWAN3_STARTUP" = 1 ] || mwan3_lock "$ACTION" "$INTERFACE"
 
 config_load mwan3
+mwan3_check_family_needs
 config_get_bool enabled globals 'enabled' '0'
 [ "${enabled}" -gt 0 ] || {
 	[ "$MWAN3_STARTUP" = 1 ] || mwan3_unlock "$ACTION" "$INTERFACE"
@@ -25,10 +26,20 @@ config_get_bool enabled globals 'enabled' '0'
 	exit 0
 }
 
-$IPT4 -S mwan3_hook &>/dev/null || {
-	mwan3_unlock "$ACTION" "$INTERFACE"
-	LOG warn "hotplug called on $INTERFACE before mwan3 has been set up"
-	exit 0
+[ $NEED_IPV4 -ne 0 ] && {
+	$IPT4 -S mwan3_hook &>/dev/null || {
+		mwan3_unlock "$ACTION" "$INTERFACE"
+		LOG warn "hotplug called on $INTERFACE(ipv4) before mwan3 has been set up"
+		exit 0
+	}
+}
+
+[ $NEED_IPV6 -ne 0 ] && {
+	$IPT6 -S mwan3_hook &>/dev/null || {
+		mwan3_unlock "$ACTION" "$INTERFACE"
+		LOG warn "hotplug called on $INTERFACE(ipv6) before mwan3 has been set up"
+		exit 0
+	}
 }
 
 mwan3_init
--- a/net/mwan3/files/etc/hotplug.d/iface/15-mwan3
+++ b/net/mwan3/files/etc/hotplug.d/iface/15-mwan3
@@ -48,57 +48,59 @@ config_get_bool enabled $INTERFACE 'enabled' '0'
 	LOG notice "mwan3 hotplug on $INTERFACE not called because interface disabled"
 	exit 0
 }
+[ "$FAMILY" = "ipv4" ] || [ "$FAMILY" = "ipv6" ] || FAMILY="ipv4 ipv6"
 
-trackpid=$(pgrep -f "mwan3track $INTERFACE ")
+for family in $FAMILY; do
+
+trackpid=$(pgrep -f "mwan3track $family $INTERFACE ")
 
 if [ "$initial_state" = "offline" ]; then
-	status=$(cat $MWAN3TRACK_STATUS_DIR/$INTERFACE/STATUS 2>/dev/null || echo unknown)
+	status=$(cat $MWAN3TRACK_STATUS_DIR/$INTERFACE.$family/STATUS 2>/dev/null || echo unknown)
 else
 	status=online
 fi
 
-[ -z "$TRUE_INTERFACE" ] && mwan3_get_true_iface TRUE_INTERFACE $INTERFACE
-
 binary_status=$status
 [ "$binary_status" = "online" ] || binary_status=offline
 
-LOG notice "Execute "$ACTION" event on interface $INTERFACE (${DEVICE:-unknown})"
+LOG notice "Execute "$ACTION" event on interface $INTERFACE($family) (${DEVICE:-unknown})"
 
 case "$ACTION" in
 	ifup|connected)
-		mwan3_create_iface_iptables $INTERFACE $DEVICE
-		mwan3_create_iface_rules $INTERFACE $DEVICE
-		mwan3_create_iface_route $INTERFACE $DEVICE
-		[ "$MWAN3_STARTUP" != 1 ] && mwan3_add_non_default_iface_route $INTERFACE $DEVICE
-		mwan3_set_iface_hotplug_state $INTERFACE "$binary_status"
+		mwan3_create_iface_iptables $INTERFACE $DEVICE $family
+		mwan3_create_iface_rules $INTERFACE $DEVICE $family
+		mwan3_create_iface_route $INTERFACE $DEVICE $family
+		[ "$MWAN3_STARTUP" != 1 ] && mwan3_add_non_default_iface_route $INTERFACE $DEVICE $family
+		mwan3_set_iface_hotplug_state $INTERFACE "$binary_status" $family
 
-		mwan3_get_src_ip src_ip "$TRUE_INTERFACE"
 		if [ -n "${trackpid}" ]; then
-			device_pid=$(pgrep -f "mwan3track $INTERFACE $DEVICE ")
+			device_pid=$(pgrep -f "mwan3track $family $INTERFACE $DEVICE ")
 			if [ "$device_pid" = "$trackpid" ]; then
 				[ "$ACTION" = ifup ] && kill -USR2 "$trackpid"
 			else
-				mwan3_track $INTERFACE $DEVICE "$binary_status" "$src_ip"
-				LOG notice "Restarted tracker [$!] on interface $INTERFACE (${DEVICE:-unknown})"
+				mwan3_track $INTERFACE $DEVICE "$binary_status" $family
+				LOG notice "Restarted tracker [$!] on interface $INTERFACE($family) (${DEVICE:-unknown})"
 			fi
 		else
-			mwan3_track $INTERFACE $DEVICE "$binary_status" "$src_ip"
-			LOG notice "Started tracker [$!] on interface $INTERFACE (${DEVICE:-unknown})"
+			mwan3_track $INTERFACE $DEVICE "$binary_status" $family
+			LOG notice "Started tracker [$!] on interface $INTERFACE($family) (${DEVICE:-unknown})"
 		fi
-		[ "$MWAN3_STARTUP" != 1 ] && [ "$binary_status" == "online" ] && mwan3_set_policies_iptables
+		[ "$MWAN3_STARTUP" != 1 ] && [ "$binary_status" == "online" ] && mwan3_set_policies_iptables $family
 
 	;;
 	ifdown|disconnected)
-		mwan3_set_iface_hotplug_state $INTERFACE "offline"
-		mwan3_delete_iface_ipset_entries $INTERFACE
-		mwan3_delete_iface_rules $INTERFACE
-		mwan3_delete_iface_route $INTERFACE
-		mwan3_delete_iface_iptables $INTERFACE
+		mwan3_set_iface_hotplug_state $INTERFACE "offline" $family
+		mwan3_delete_iface_ipset_entries $INTERFACE $family
+		mwan3_delete_iface_rules $INTERFACE $family
+		mwan3_delete_iface_route $INTERFACE $family
+		mwan3_delete_iface_iptables $INTERFACE $family
 		if [ "$ACTION" = "ifdown" ]; then
 			[ -n "$trackpid" ] && kill -USR1 "$trackpid"
 		fi
 		mwan3_set_policies_iptables
 	;;
 esac
+
+done
 [ "$MWAN3_STARTUP" = 1 ] || mwan3_unlock "$ACTION" "$INTERFACE"
 exit 0
--- a/net/mwan3/files/usr/sbin/mwan3rtmon
+++ b/net/mwan3/files/usr/sbin/mwan3rtmon
@@ -54,8 +54,8 @@ mwan3_rtmon_route_handle()
 	}
 	handle_route_cb(){
 		let tid++
-		config_get family "$section" family ipv4
-		[ "$family" != "$route_family" ] && return
+		config_get family "$section" family "any"
+		[ "$family" != "$route_family" ] && [ "$family" != "any" ] && return
 		handle_route
 	}
 
--- a/net/mwan3/files/etc/hotplug.d/iface/16-mwan3-user
+++ b/net/mwan3/files/etc/hotplug.d/iface/16-mwan3-user
@@ -7,6 +7,7 @@
 	[ "$MWAN3_STARTUP" = 1 ] || mwan3_lock "$ACTION" "$DEVICE-user"
 
 	config_load mwan3
+	mwan3_check_family_needs
 	config_get_bool enabled globals 'enabled' '0'
 	[ "${enabled}" -gt 0 ] || {
 		[ "$MWAN3_STARTUP" = 1 ] || mwan3_unlock "$ACTION" "$DEVICE-user"
--- a/net/mwan3/files/lib/mwan3/common.sh
+++ b/net/mwan3/files/lib/mwan3/common.sh
@@ -15,3 +15,6 @@ LOG()
 	[ "$facility" = "debug" ] && return
 	logger -t "$SCRIPTNAME[$$]" -p $facility "$*"
 }
+
+MWAN3_STATUS_DIR="/var/run/mwan3"
+MWAN3TRACK_STATUS_DIR="/var/run/mwan3track"
--- a/net/mwan3/files/lib/mwan3/mwan3.sh
+++ b/net/mwan3/files/lib/mwan3/mwan3.sh
@@ -36,6 +36,30 @@ MM_UNREACHABLE=""
 
 command -v ip6tables > /dev/null
 NO_IPV6=$?
+NEED_IPV4=0
+NEED_IPV6=0
+
+mwan3_check_family_needs()
+{
+	local enabled family
+	check_family()
+	{
+		config_get enabled "$1" enabled 0
+		config_get family "$1" family "any"
+
+		[ "$enabled" = "0" ] && return
+		if [ "$family" = "any" ]; then
+			NEED_IPV4=1
+			NEED_IPV6=1
+		elif [ "$family" = "ipv4" ]; then
+			NEED_IPV4=1
+		elif [ "$family" = "ipv6" ]; then
+			NEED_IPV6=1
+		fi
+	}
+	config_foreach check_family interface
+	[ $NO_IPV6 -ne 0 ] && NEED_IPV6=0
+}
 
 mwan3_push_update()
 {
@@ -320,9 +344,20 @@ mwan3_set_connected_iptables()
 {
 	local connected_network_v6 source_network_v6 error
 	local update=""
-	mwan3_set_connected_ipv4
 
-	[ $NO_IPV6 -eq 0 ] && {
+	if [ $NEED_IPV6 -ne 0 ] || [ $NEED_IPV4 -ne 0 ]; then
+		mwan3_push_update -! create mwan3_connected list:set
+		mwan3_push_update flush mwan3_connected
+	fi
+
+	[ $NEED_IPV4 -ne 0 ] && {
+		mwan3_set_connected_ipv4
+		mwan3_push_update -! add mwan3_connected mwan3_connected_v4
+		mwan3_push_update -! create mwan3_dynamic_v4 hash:net
+		mwan3_push_update -! add mwan3_connected mwan3_dynamic_v4
+	}
+
+	[ $NEED_IPV6 -ne 0 ] && {
 		mwan3_push_update -! create mwan3_connected_v6 hash:net family inet6
 		mwan3_push_update flush mwan3_connected_v6
 
@@ -334,18 +369,12 @@ mwan3_set_connected_iptables()
 		for source_network_v6 in $($IP6 addr ls | sed -ne 's/ *inet6 \([^ \/]*\).* scope global.*/\1/p'); do
 			mwan3_push_update -! add mwan3_source_v6 "$source_network_v6"
 		done
+
+		mwan3_push_update -! add mwan3_connected mwan3_connected_v6
+		mwan3_push_update -! create mwan3_dynamic_v6 hash:net family inet6
+		mwan3_push_update -! add mwan3_connected mwan3_dynamic_v6
 	}
 
-	mwan3_push_update -! create mwan3_connected list:set
-	mwan3_push_update flush mwan3_connected
-	mwan3_push_update -! add mwan3_connected mwan3_connected_v4
-	[ $NO_IPV6 -eq 0 ] && mwan3_push_update -! add mwan3_connected mwan3_connected_v6
-
-	mwan3_push_update -! create mwan3_dynamic_v4 hash:net
-	mwan3_push_update -! add mwan3_connected mwan3_dynamic_v4
-
-	[ $NO_IPV6 -eq 0 ] && mwan3_push_update -! create mwan3_dynamic_v6 hash:net family inet6
-	[ $NO_IPV6 -eq 0 ] && mwan3_push_update -! add mwan3_connected mwan3_dynamic_v6
 	error=$(echo "$update" | $IPS restore 2>&1) || LOG error "set_connected_iptables: $error"
 }
 
@@ -354,7 +383,8 @@ mwan3_set_general_rules()
 	local IP
 
 	for IP in "$IP4" "$IP6"; do
-		[ "$IP" = "$IP6" ] && [ $NO_IPV6 -ne 0 ] && continue
+		[ "$IP" = "$IP4" ] && [ $NEED_IPV4 -eq 0 ] && continue
+		[ "$IP" = "$IP6" ] && [ $NEED_IPV6 -eq 0 ] && continue
 		RULE_NO=$(($MM_BLACKHOLE+2000))
 		if [ -z "$($IP rule list | awk -v var="$RULE_NO:" '$1 == var')" ]; then
 			$IP rule add pref $RULE_NO fwmark $MMX_BLACKHOLE/$MMX_MASK blackhole
@@ -371,7 +401,8 @@ mwan3_set_general_iptables()
 {
 	local IPT current update error
 	for IPT in "$IPT4" "$IPT6"; do
-		[ "$IPT" = "$IPT6" ] && [ $NO_IPV6 -ne 0 ] && continue
+		[ "$IPT" = "$IPT4" ] && [ $NEED_IPV4 -eq 0 ] && continue
+		[ "$IPT" = "$IPT6" ] && [ $NEED_IPV6 -eq 0 ] && continue
 		current="$($IPT -S)"
 		update="*mangle"
 		if [ -n "${current##*-N mwan3_ifaces_in*}" ]; then
@@ -467,13 +498,12 @@ mwan3_create_iface_iptables()
 	[ "$family" = "any" ] && family="$3"
 	[ "$family" = "$3" ] || return 0
 
-	if [ "$family" = "ipv4" ]; then
+	if [ "$family" = "ipv4" ] && [ "$NEED_IPV4" -ne 0 ]; then
 		connected_name=mwan3_connected
 		IPT="$IPT4"
 		IPTR="$IPT4R"
 		$IPS -! create $connected_name list:set
-
-	elif [ "$family" = "ipv6" ] && [ $NO_IPV6 -eq 0 ]; then
+	elif [ "$family" = "ipv6" ] && [ $NEED_IPV6 -ne 0 ]; then
 		connected_name=mwan3_connected_v6
 		IPT="$IPT6"
 		IPTR="$IPT6R"
@@ -526,10 +556,9 @@ mwan3_delete_iface_iptables()
 	[ "$family" = "any" ] && family="$2"
 	[ "$family" = "$2" ] || return
 
-	if [ "$family" = "ipv4" ]; then
+	if [ "$family" = "ipv4" ] && [ $NEED_IPV4 -ne 0 ]; then
 		IPT="$IPT4"
-	elif [ "$family" = "ipv6" ]; then
-		[ $NO_IPV6 -ne 0 ] && return
+	elif [ "$family" = "ipv6" ] && [ $NEED_IPV6 -ne 0 ]; then
 		IPT="$IPT6"
 	else
 		return
@@ -634,11 +663,12 @@ mwan3_add_all_nondefault_routes()
 
 	mwan3_update_dev_to_table
 	for ipv in ipv4 ipv6; do
-		[ "$ipv" = "ipv6" ] && [ $NO_IPV6 -ne 0 ] && continue
-		if [ "$ipv" = "ipv4" ]; then
+		if [ "$ipv" = "ipv4" ] && [ $NEED_IPV4 -ne 0 ]; then
 			IP="$IP4"
-		elif [ "$ipv" = "ipv6" ]; then
+		elif [ "$ipv" = "ipv6" ] && [ $NEED_IPV6 -ne 0 ]; then
 			IP="$IP6"
+		else
+			continue
 		fi
 		tid=0
 		active_tbls=" "
@@ -664,12 +694,12 @@ mwan3_delete_iface_route()
 	[ "$family" = "any" ] && family="$2"
 	[ "$family" = "$2" ] || return
 
-	if [ "$family" = "ipv4" ]; then
+	if [ "$family" = "ipv4" ] && [ $NEED_IPV4 -ne 0 ]; then
 		$IP4 route flush table "$id"
-	fi
-
-	if [ "$family" = "ipv6" ] && [ $NO_IPV6 -eq 0 ]; then
+	elif [ "$family" = "ipv6" ] && [ $NEED_IPV6 -ne 0 ]; then
 		$IP6 route flush table "$id"
+	else
+		return
 	fi
 }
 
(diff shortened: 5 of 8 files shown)


net/mwan3/files/usr/sbin/mwan3rtmon
```diff
@@ -25,10 +25,12 @@ main() {
...
 	while true; do
 		mwan3_lock "service" "mwan3rtmon"
-		mwan3_rtmon_ipv4 || mwan3_rtmon_ipv6
+		mwan3_rtmon_ipv4
 		ret=$?
+		mwan3_rtmon_ipv6
+		ret=$ret$?
 		mwan3_unlock "service" "mwan3rtmon"
-		[ "$ret" = "0" ] || break
+		[ "$ret" = "11" ] || break
```
aaronjg, line 33: it appears you flipped the behavior here. What is the reason for that change?

ptpt52:
if none interface is setup for ipv6, I am expecting it to be bypass.
if one or more interface is/are setup for ipv6, I am expecting it work as mwan3 policy/rule

Actual case:
I have two wan(s), and I setup balance on them for ipv4, and nothing to do with ipv6
I am expecting ipv4 traffic go as mwan3 policy/rule
but ipv6 just route via main table only

aaronjg:
Why not just add something like this to your mwan3 config?

```
config rule 'default'
       option family 'any'
       option proto 'all'
       option sticky '0'
       option use_policy 'default'
```

ptpt52:
update:
fix in `mwan3: make sure run both mwan3_rtmon_ipv4/6`
add new commit `mwan3: daul stack interface support `

ptpt52:
for interface, If family is "" means both ipv4 and ipv6

ptpt52:
> Why not just add something like this to your mwan3 config?
> 
> ```
> config rule 'default'
>        option family 'any'
>        option proto 'all'
>        option sticky '0'
>        option use_policy 'default'
> ```

I am prefer to this idea:
if ipv4 interface is set, we make mwan3 works for ipv4 only, and leave ipv6 alone
if ipv6 interface is set, we make mwan3 works for ipv6 only, and leave ipv4 alone

ptpt52:
and I add 1 changes in mwan3 luci https://github.com/openwrt/luci/pull/4347

net/mwan3/files/etc/hotplug.d/iface/15-mwan3
```diff
@@ -32,7 +32,8 @@ config_get_bool enabled $INTERFACE 'enabled' '0'
 }
 
 if [ "$ACTION" = "ifup" ]; then
-	config_get family $INTERFACE family ipv4
+	config_get familys $INTERFACE family "ipv4 ipv6"
```
aaronjg, line 35: In the extant code base, "any" is used to indicate both ipv4 and ipv6.

ptpt52: @aaronjg  as I can see,
in https://github.com/openwrt/luci/blob/master/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js#L217,
empty family is "ipv4 + ipv6"

aaronjg: The value in "_('XXX')" is the helper text. It actually leaves the field blank. The default value is then "any", as interpreted by fw3. Config files can also explicitly use the 'any' family, eg: https://git.openwrt.org/?p=project/firewall3.git;a=blob;f=helpers.conf;h=b8ae857cb8ce0424eed59e79b876c68048e1b957;hb=HEAD

Which is the same way that mwan3 handles the default in `mwan3_set_user_iptables_rule`
https://github.com/openwrt/packages/blob/master/net/mwan3/files/lib/mwan3/mwan3.sh#L881

aaronjg:
> I am prefer to this idea:
> if ipv4 interface is set, we make mwan3 works for ipv4 only, and leave ipv6 alone
> if ipv6 interface is set, we make mwan3 works for ipv6 only, and leave ipv4 alone

Here is how it can have unexpected behavior:
- ipv4 rules for interfaces wana and wanb. 
- load balancing set up for most traffic between wana (metric 10) and wanb (metric 20)
- Host 192.168.1.100 routes has a rule to route all traffic out of wanb
- user runs "uci set mwan3.wanb.enabled=0", now 192.168.1.100 traffic is not routed out (as expected)
- user runs "uci set mwan3.wana.enabled=0", now 192.168.1.100 traffic now is routed out of wana, because it was the last ipv4 interface enabled. This is unexpected! Disabling wana in mwan3 caused traffic from 192.168.1.100 to be routed out of wana!

If you want to disable mwan3 on ipv4/ipv6 completely, it is better to have a global configuration option, rather than trying to infer this from which interfaces are enabled. This would also allow a user to easily switch off mwan3 for ipv4 or ipv6 when that is there intention without toggling all of the interfaces.

ptpt52:
>user runs "uci set mwan3.wana.enabled=0", now 192.168.1.100 traffic now is routed out of wana, because it was the last ipv4 interface enabled. This is unexpected! Disabling wana in mwan3 caused traffic from 192.168.1.100 to be routed out of wana!

that is expected, because you disable all ipv4 interface, that means, traffic must follow the main table, it is reasonable.

net/mwan3/files/etc/hotplug.d/iface/15-mwan3
ptpt52: no, We never set family to the value `'any'` explicitly
and if family is unset or empty, it is dual stack

ptpt52: for code in lib/mwan3/mwan3.sh#L881
that is, if we get empty then the default use 'any' as value, 
This is the result required by the shell code itself, not the value of the family itself

aaronjg: it is possible to edit the settings using uci or editing the config file directly. The example you showed as a model was from fw3, which can have family be set to `any` in a dual stack setting. mwan3 already supports having family set to any, just like fw3 does. I don't understand why you want to break this.

ptpt52: You don't seem to understand?

family in config file is "ipv4" "ipv6" or ""/none
we nevert save family to "any"

ptpt52:
I will redo the changes after this PR(https://github.com/openwrt/packages/pull/13041) get merged

net/mwan3/files/etc/hotplug.d/iface/15-mwan3
feckert: @ptp52 the line is wrong I agree with @aaronjg 
`config_get familys $INTERFACE family "ipv4 ipv6"`
this should be
`config_get family $INTERFACE family ""`

feckert:
I have marked this pullrequest as WIP because and work on this again if the [P/R](https://github.com/openwrt/packages/pull/13041) is merged

ptpt52:
we may stop review on this PR for now, I will redo these changes after the PR (https://github.com/openwrt/packages/pull/13041)

aaronjg:
> we may stop review on this PR for now, I will redo these changes after the PR (#13041)

Thank you. When you do the refactor on this, please think carefully about different possible mwan3 configurations, and carefully consider if the changes will alter the behavior of mwan3 for users. 

Sometimes it is reasonable to make changes that break existing configurations if their is an overall benefit. However, it is always better for an upgrade is seamless for existing users.

Not all users are running the same configuration as you, and not all users set up their configuration using LuCI, so there will be valid configurations that have been deployed that are not possible to achieve using LuCI. This includes explicitly setting default options for various config values. 

You often use 'we' when referring to how you set up your configuration. Please keep in mind that mwan3 is a widely used package with heterogeneous use cases and not everyone will have set up their configuration or have the same expectations of default behavior that you do.

ptpt52:
I now re-push new changes for dual stack support
1. if we setup for ipv4 interface only, no touch for ipv6
2. if we setup for ipv6 interface only, no touch for ipv4
3. interface track_ips can be ipv4 or ipv6, or hostname for both
4. do a masq help for ipv6

ptpt52:
also I have to change luci to support mwan3 dual stack interface
https://github.com/openwrt/luci/pull/4347

ptpt52:
for usage case example:
```
wana is pppoe and virtual wana_6 is up for ipv6
wana is another pppoe and virtual wana_6 is up for ipv6

route v6 in main:
default from aaaa::/64 via aaaa:xxxx::1
default from bbbb::/64 via bbbb:xxxxx::1

we setup mwan3 on interfaces:
1. wana family "any"
track_ip: 1.1.1.1, 88::88,

2. wanb family "any"
track_ip: 88::88
no ipv4 ip for track

mwan3 status:
Interface status:
 interface wana(ipv4) is online 10h:44m:40s, uptime 10h:44m:41s and tracking is active
 interface wana(ipv6) is online 12h:44m:40s, uptime 12h:44m:42s and tracking is active
 interface wanb(ipv4) is online 00h:00m:00s, uptime 12h:44m:41s and tracking is not enabled (:no ip for track)
 interface wanb(ipv6) is online 6h:44m:40s, uptime 6h:44m:42s and tracking is active
...

route in table 1 v6:
default dev pppoe-wana metric 1024 pref medium

route in table 2 v6:
default dev pppoe-wanb metric 1024 pref medium


MASQ for ipv6
-A POSTROUTING -o pppoe-wana -m set ! --match-set mwan3_wana_ipv6_src_from src -m comment --comment masq-help-wana-dev -j MASQUERADE
-A POSTROUTING -o pppoe-wanb -m set ! --match-set mwan3_wanb_ipv6_src_from src -m comment --comment masq-help-wanb-dev -j MASQUERADE

ipset in mwan3_wana_ipv6_src_from:
aaaa::/64

ipset in mwan3_wanb_ipv6_src_from:
bbbb::/64

```
we need this masq help, if not, if traffic comes from aaaa::xyz and dst is dst::dst, and if mwan3 rules match and select route table 2, and it is going out via wanb, but src is aaaa::xyz, so it will not work.

feckert:
For the ipv6 nat could we use her the procd firewall?
Look at the way this is being handled in net-snmp.
https://github.com/openwrt/packages/blob/3e33919d1025e26760b200a3442b39bf1f4f0fc9/net/net-snmp/files/snmpd.init#L330

I have also found the following packages that make use of this.
```
package/feeds/packages/simple-adblock/files/simple-adblock.init:                json_add_array firewall
package/feeds/packages/net-snmp/files/snmpd.init:       json_add_array firewall
package/network/ipv6/ds-lite/files/dslite.sh:   json_add_array firewall
package/network/ipv6/464xlat/files/464xlat.sh:  json_add_array firewall
package/network/ipv6/map/files/map.sh:  json_add_array firewall
package/network/services/igmpproxy/files/igmpproxy.init:        json_add_array firewall
package/network/services/omcproxy/files/omcproxy.init:  json_add_array firewall
package/network/utils/wireguard-tools/files/wireguard.init:             json_add_array firewall
```
The fw3 firewall of openwrt does support nat see [commit](https://git.openwrt.org/?p=project/firewall3.git;a=commit;h=31456301f514e3e04f61930bb00f6b2d99b4d067) 
I haven't tried it myself yet, but I would try it like this.
The advantage would be that the Openwrt Firewall takes care of the whole thing and mwan3 only requests this.

ptpt52:
This masq rule(s) is dynamic, so we have to monitor them our self
1. if dhcpv6 get new ip and change ip/route, then we have to update masq rules

aaronjg:
I agree that you need some sort of additional handling (potentially masquerading) to get IPv6 to work, just like you need masquerading for IPv4 to work. However with IPv6, there are other configurations to make this work. I'm not sure it is within the scope of mwan3 to take care of this.

Other configurations could be:
1) Using network prefix translation (RFC 7157) (I believe Brian Murrell has something like this set up).
2) Having failover and sending RECONFIGURE messages to the clients to alert them of the new prefix
3) Having a provider independent address space (though likely out of reach of 99.9% of SOHO setups, so wouldn't be too much of a problem to not support this.).

I believe @jamesmacwhite has already set this up with the standard openwrt NAT66 setup (https://openwrt.org/docs/guide-user/network/ipv6/ipv6.nat6), and he wrote up a blog post about it https://medium.com/@jamesmacwhite/using-nat66-with-mwan3-on-openwrt-f8ebe4f21044.

Is there something that this doesn't do that requires deeper integration into mwan3? 

Either way, mwan3 should not be tied to masquerading on ipv6.

I will review the P/R once we get the 5.4.x issues fixed. Stylistically it is much improved, but it still seems like there are unhandled edge cases, and changes that could break existing configurations.

ptpt52:
as we know, the mwan3 would break the ipv6 if no nat setup for it
but if we setup nat for all ipv6 to make mwan3 work, it is no sense, because  the client use global ipv6 addr to send traffic out, and the router do nat on it, this is unreasonable.

for the example above:
if client with addr6 `aaaa:xyz` send traffic to `some_dst_addr6`, and if mwan3 use balanced policy rule, 
there will be two situations:
1. the traffic goes out via wana (control by mwan3), we don't need to do masq nat for it
2. the traffic goes out via wanb (also selected by mwan3), we have to do masq nat for it

for these situations, what should be done is proper?
what we do before is, setup nat for ipv6 for all traffic comes from `lan` to `wan*`, so that mwan3 work

but now I think about another way to archive this.
if traffic goes out via `wana` no masq nat needed
if traffic goes out via `wanb`, then do masq nat on it.

ptpt52:
anyway, the first two commits are likely ready to get merged
```
net/mwan3: support dual stack interface
net/mwan3: skip setup for unused interface family
```
and for commit `net/mwan3: help masq for ipv6`
We may have disputes, need more discussion, maybe there is a better way

jamesmacwhite:
Just my take on this given I do this already with IPv6 and mwan3, but using the NAT6 firewall script here: https://openwrt.org/docs/guide-user/network/ipv6/ipv6.nat6 which looks similar to the proposed firewall up script shipping with mwan3 here.

So mwan3 doing IPv6 masquerading is going be subjective for a couple of reasons. It isn't the only way to achieve IPv6 routing with mwan3, there are alternatives such as NETMAP or NPTv6 but at this time this will require advanced/additional configuration not part of mwan3. 

So if mwan3 went down forcing a specific IPv6 masquerading solution that's probably pushing mwan3 down a specific configuration route, which we'd need to be careful of. However, if it's something optional or able to be conditionally enabled/disabled then it is more a case of providing helper configuration if required.

As @aaronjg mentioned the NAT6 firewall script above, already allows IPv6 masquerading to work with mwan3. You don't need to do anything in mwan3 is you use the NAT6 script above, as it's applied to the firewall zones with the `option masq6 '1'` option, it's all handled higher up. I'd be interested to know what the difference would be between the two, given I already use the firewall script outside of mwan3.

aaronjg:
> To be honest, according to my current testing situation, this helper is very useful.

While it's unlikely there will be a one-size-fits-all solution for IPv6 that can be enabled in mwan3 by default, we can certainly add optional helpers to make setup on IPv6 easier. If you think this would be widely useful, you could write up a bit more about what it does.  There was no commit message on 3d8ebc6 so it is not clear what it is doing without going through all the code.

1) What is the problem that 3d8ebc6 is trying to solve? 
2) What are the standard alternatives? (eg RFC 6296, RFC 7157, RFC 7934, RFC 8415 , RIPE 738§7)
3) Why are these inadequate for mwan3?
4) How does 3d8ebc6 address these inadequacies? Is it implementing any RFC standard?
5) What scenarios would 3d8ebc6 be inappropriate for?

luizluca:
As a mwan3 IPv6 user, I will give my 2 cents.

NPT is not an option. It is designed to be a light weight translator but it only works (and it only makes sense) if you do not use conntrack. So, no stateful firewall and no mwan3. The alternative is NETMAP, which is the one I use. A single /64 could be shared amongst multiple internal subnets without collision as interface-id are mostly unique.

I do not use NAT66 as it is just forcing all connections through the same IP. I would only use it as a last resort when ISP didn't provide a delegated prefix. NAT is for the lack of addresses, while in ipv6 we have plenty of them.

I'm not distributing my ISP IPv6 addresses internally. All clients uses only ULA/48 addresses (a suggestion from RFC 6296). I have two /48 from my ISP and I netmap them 1x1 to ULA addresses. It might be a bad design as ULA are supposed to be local. Using ULA to access internet has a bad side effect because /etc/gai.conf default conf prefer IPv4 over IPv6 ULA to internet. As all my traffic goes through a proxy server, it is currently a minor problem, which might bite me in the future.

All other RFC can be summarized as "select outgoing ISP already in clients and, if applicable, are outside mwan3 business. They normally are better implemented using multiple routers as a single router is a single point of failure.

My suggestion: consider NETMAP by default, NAT66 as a not recommend fallback. Forget about NPT and any other current RFC which solve this problem before mwan3.

aaronjg:
Thanks @luizluca. That makes a lot of sense to me. Sounds like NETMAP and NAT66 should be done outside of mwan3 because they have other use cases. 

If the ISP isn't providing a delegated prefix, it sounds like the NAT66 would have the same issue as NETMAP with IPv4 preferred over IPv6 ULA? 

For faliover use cases, what do you think about sending RECONFIGURE according to RFC 8415?

luizluca:
> Thanks @luizluca. That makes a lot of sense to me. Sounds like NETMAP and NAT66 should be done outside of mwan3 because they have other use cases.

mwan3 could provide firewall scripts but not enable them by default. They could be added to config/firewall but check if config/mwan3 enable either one of them. I prefer to have them added on demand (specially using luci) as I'm very conservative on touching config files without an effective change.

There might be little usage for NETMAP/NAT66 outside multihoming. I use NETMAP at home to provide access to IPv6-only services for secondary subnet as my ISP just give me a single /64 delegated prefix. Although, they have the issue of using ULA-only addresses. It's not mwan3 specific.
 
> If the ISP isn't providing a delegated prefix, it sounds like the NAT66 would have the same issue as NETMAP with IPv4 preferred over IPv6 ULA?

Exactly. IPv4 will have precedence over ULA (fc00::/7) by default. It has a different label from (::/0) while both private and public IPv4 shares the same label (::ffff:0:0/96). Matching labels are checked before precedence. The only way to change it for Linux machines is manually editing /etc/gai.conf, setting your ULA prefix label to match the one used by ::/0. I never checked how or even if it is possible to do it for other OS. There is a proposed RFC 7078 to change that using DHCPv6 but I know no client that currently consider it. So, mwan3 for IPv6 will be mostly useless without a globally routable prefix on client side. It will only be used by IPv6 only clients (that might use DNS46, NAT64), to reach IPv6 only services or for those who manually edit their config files. In my (enterprise) particular case, all internet access goes through a HTTP Proxy, which makes changing gai.conf feasible. 

We need a "random generated globally routable private prefix" exclusively for NPT/NETMAP/NAT66 cases. It should not be routed by any ISP (just like IPv4 private addresses and ULA) but clients would try to route them. We could also propose a RFC to use fc00::/8 as ULA currently only uses fd00::/8 but it would take ages to widepread the new rules. It would be better to simply reserve a new /8 for that case. The IPv6 advocates does not really like NPT/NETMAP/NAT66 and it might be a hard path to follow.

I don't know what solution will get widely available first: a way to reconfigure Address Selection Policy via DHCPv6? A new private prefix sharing the label with ::/0? IPv4 being turned off? Cliente-side multipath protocols (MPTCP, QUIC)? Network-wide ISP state notification?

> For faliover use cases, what do you think about sending RECONFIGURE according to RFC 8415? Would NPT also be suitable in failover cases?

If you enable NPT, there is no stateful firewall. You'll need to simply disable firewall (allowing forwarding from/to internet without restriction) or filter solely based on flags (TCP) and creating rules for both from/to flows for other protocols. There is no "related", "connected" states. And forget about multiport protocols like FTP, TFTP, SIP... It's worse than NAT. 

The reason NPT does not work with conntrack is because if you are already spending CPU for controlling connection, using NPT instead of NAT/NETMAP will save you nothing and NETMAP can preserve subnet id and interface id while translating addresses. You can even NETMAP to a smaller prefix. I never tested it but I guess if you map to a /128 network, it is functionally equivalent to NAT66. If it does not use more CPU, we can drop NAT66 as MASQUERADE might be a special case of NETMAP.

If you really really need to use NPT, you'll need to have a router dedicated to translating and another one doing mwan3 (which is based on conntrack) and filtering. It could be inside the same machine using netns but it is way more complex than any other solution. NTP is simply incompatible with mwan3.

You can use DHCP RECONFIGURE with or without NPT but only if you do not use mwan3 and you have another firewall solution. Anyway, it is out of topic for mwan3.

ptpt52:
How to make mwan3 on ipv6 work properly?
As you know, MASQ NAT needs to be set up, right?

Here is another way, masq helper to help more suitable MASQ
This means that mwan3 can work perfectly on ipv6 in two aspects:
1. We do masq if needed
2. We don’t make masq if not needed

When is it needed and when is it not needed?
1. client use GUA(Global Unicast Address), and assume GUA is fetch from `wana`,
and traffic goes out via `wana`, this case we do not make masq

2. client use address-not-from-wana, but goes out via `wana`, this case we have to make masq

aaronjg:
> There might be little usage for NETMAP/NAT66 outside multihoming. I use NETMAP at home to provide access to IPv6-only services for secondary subnet as my ISP just give me a single /64 delegated prefix. Although, they have the issue of using ULA-only addresses. It's not mwan3 specific.

Yes. This makes sense. Some VPN providers also only give a /128 and do NAT66 on there side, so may be needed even without mwan3.

> If you enable NPT, there is no stateful firewall. You'll need to simply disable firewall (allowing forwarding from/to internet without restriction) or filter solely based on flags (TCP) and creating rules for both from/to flows for other protocols. There is no "related", "connected" states. And forget about multiport protocols like FTP, TFTP, SIP... It's worse than NAT.

I see. So probably very little use case for NPT, and NAT66/NETMAP would be preferred.

> We need a "random generated globally routable private prefix" exclusively for NPT/NETMAP/NAT66 cases. It should not be routed by any ISP (just like IPv4 private addresses and ULA) but clients would try to route them. We could also propose a RFC to use fc00::/8 as ULA currently only uses fd00::/8 but it would take ages to widepread the new rules. It would be better to simply reserve a new /8 for that case. The IPv6 advocates does not really like NPT/NETMAP/NAT66 and it might be a hard path to follow.

Yes, this seems like an ideal solution, but as you say, it would likely take a long time to get something like this, so NAT66/NETMAP on the router would be a good solution (likely configured outside of mwan4).

> You can use DHCP RECONFIGURE with or without NPT but only if you do not use mwan3 and you have another firewall solution. 

I was thinking that DHCP RECONFIGURE could be used with mwan3 failover if you have two ISPs that give you different globally routable prefixes. If mwan3 detects that the main link becomes unreliable, you could send a DHCP RECONFIGURE requests to all the clients to tell them to use the other prefix. Or would that not work?

luizluca:
> I was thinking that DHCP RECONFIGURE could be used with mwan3 failover if you have two ISPs that give you different globally routable prefixes. If mwan3 detects that the main link becomes unreliable, you could send a DHCP RECONFIGURE requests to all the clients to tell them to use the other prefix. Or would that not work?

It might work but controlling DHCPv6 might be out of topic for mwan3 and even be too complex for odhcpd (I didn't try it). You'll need to make it forget one prefix and use other one (without touching the persistent configuration files). And you cannot simply bring down the interface as you might need to use it for detecting when the link is up again or even for connections not controlled by mwan3 like from internet to an internal server. After that, you'll need to read the leases files and send the RECONFIGURE. You could send it "on demand" detecting those clients still sending data through the dead link. Anyway, the implementation might be dependent on which DHCPv6 the user is using. And I'm not even considering that clients might use a different device as DHCPv6. A SMB, which would be very interesting in mwan3, might prefer to use MS DHCP. My point is that it is more a DHCPv6 add-on than a mwan3 solution. You'll need to have a odhcpd-mwan3, dnsmasq-mwan3, etc.

mwan3 has two jobs: 1) detect if the link is working and 2) route traffic to one or other link. The 1) part could be an independent service, announcing link events to other interesting services or devices, specially DHCPv6 services and any other client-side link probe service. However, the 2) rules used to generate "ip(6)tables mangles" cannot be used for DHCPv6 world. You'll need different rules consumed by different code. It might be better to have a different package.

jamesmacwhite:
I think mwan3 shouldn't just enable NAT6, unless it's a configurable option, not on by default. Whatever your viewpoint on IPv6 and NAT, I think we generally agree there needs to be some form of IPv6 masquerading to make it work in mwan3 but I think it's ultimately a user configuration decision on what this is, as it's subjective. Right here we have NETMAP vs NAT6 just as an example.

I think the concern is mwan3 pushing the configuration down a specific route. If it was to include helper scripts, they should be default off and allowed to be conditionally enabled or disabled. e.g. right now if mwan3 just did masquerading because IPv6 is present, this would conflict with my NAT6 firewall config already running as I'm doing it at the firewall level now. I already have masq6 enabled on my various WAN firewall zones.

My reason for NAT6 is purely for multihoming. I actually have two IPv6 WANs that delegate a /48, but then I have another 4G/LTE backup connection which doesn't have a PD and I also use Wireguard which only uses ULA as that's NAT6 already, so NAT6 is implemented on outbound to allow IPv6 traffic to be balanced/failover within mwan3 for all interfaces, it isn't necessarily about not having enough IPv6 addresses, more a case of masquerading IPv6 so the traffic can be managed by mwan3 properly.

feckert:
@jamesmacwhite I agree with your suggestion that we should do the NAT6 or NETMAP for IPv6 not in mwan3. I see the task more in the firewall.

For IPv4 it is easy, we have everything we need and it works immediately we always do NAT. That is the default use case we have for IPv4.

For IPv6 we have only NETMAP and NAT6. And as @jamesmacwhite already mention not everyone has the same scenario on their interfaces so we should not do always NAT6 to mirror the IPv4 behavior.

The question is how does the firewall (fw3) in openwrt supports this for IPv6?
And another question is also does this configuration could be done in the LuCI?
Maybe we have to make some changes in the fw3 of openwrt to make NETMAP work with ipv6.
NAT6 should work.
But I have not tried that yet. I have to deal more with IPv6 in my setup. Germany is a desert for IPv6.

@jamesmacwhite could you please post your network and mwan3 configuration. And could you also post your package that are needed to do NETMAP or NAT6 for IPv6.

jamesmacwhite:
@feckert I can't comment on NETMAP, but maybe @lucize can perhaps provide some information on the setup. OpenWrt doesn't have much information in the docs on NETMAP, only NAT6 seems to be covered in some detail. Perhaps that's something that can be added IPv6 and NETMAP.

I've posted the key configuration elements for my setup with NAT6 here: https://gist.github.com/jamesmacwhite/58757c67cf6566c3d6cff46ece2fea32

Doing NAT6 with fw3 and LuCI is broken, you can't really configure it with LuCI alone. The firewall.nat6 script is needed to basically fill in that problem.

Package wise you only need `kmod-ipt-nat6` and of course the IPv6 packages themselves, but these will likely be installed already.

luizluca:
> @feckert I can't comment on NETMAP, but maybe @lucize can perhaps provide some information on the setup.

You mean @luizluca?

For outgoing traffic, it is basically a single line code:

* internal is the ULA or wanA delegated prefix (with mask) or a subnet of it (or even a single address);
* external is the delegated prefix (with mask) from wanB connected through DEVICE;
* wangw is the gateway used by wanB from "ip -6 route show dev $DEVICE default";
* internal_mapped is a subset of internal (or internal itself when it fits) where you can and want to receive incoming connections

```
ip6tables --table nat --append POSTROUTING --out-interface $DEVICE --source $internal -j NETMAP --to $external
```

internal can be larger than external. I didn't test but I guess external can be even a/128 address, turning NETMAP into MASQUERADE. In this case, external could be one IPv6 address from the delegated prefix or the wan ipv6 address.

You normally need to add a new route as the one you get from DHCPv6 normally include a condition "from <delegated prefix>" which will prevent ULA from reaching internet. It does not happen with static IPv6.

```
ip -6 route add default from $internal via $wangw dev $DEVICE metric 512
```

You just need to loop for each wan interface you want to netmap.

Just remember that outgoing mapping is connection based, just like NAT as the external address might be shared between multiple internal clients (although it's almost impossible to conflict when SLAAC is used in a /64 prefix). So, don't expect to reach an internal client pinging the same wanB mapped address it used to reach internet. You need to add another rule for incoming connections. 

For incoming connections, you just reverse the process:

```
ip6tables --table nat --append PREROUTING --destination $external --in-interface $DEVICE --jump NETMAP --to $internal_mapped
```

internal_mapped cannot be larger than external. If your "internal network" is larger than external, you need to select which part of your internal subnet would be accessible from internet.

And you are good to go.

For those with static addresses, you can just throw those rules inside firewall.user. For dynamic addresses, you might want to use iface triggers (for hot setup), and a firewall script or even a dedicated service (for cold setup).

jamesmacwhite:
Sorry @luizluca. My apologies. A mistake on my part there.

Thanks for the insight into NETMAP..

ptpt52:
@feckert @aaronjg  any progress

feckert:
@ptpt52 for now I am still waiting for https://github.com/openwrt/packages/pull/13169 to get CI and some error added by this change fixed by @aaronjg. Please hold on. We have done it soon.

CI is not a problem because this is not related to the pullrequest. But [this](https://github.com/openwrt/packages/pull/13169#issuecomment-696086200) must get fixed before merge

aaronjg:
@ptpt52  Thank you for your hard work on this. I can look at this more closely once it is rebased, but looking at the current version there are a few general points to address:

- Please test to make sure that everything (rules, routes, iptables, ipsets, status files, etc.) is properly cleared when mwan3 is turned off after ipv4 or ipv6 stack gets disabled
- If you want to change any defaults, please consider adding a migration script to the package to maintain behavior for existing configurations. If you change a default, and it makes sense to change existing configurations for users, please make sure that all such changes are documented in the commit messages.
- Please add checks for ipv4/ipv6 mwan3 rules when determining if a stack should be turned off. This will minimize breaking changes in existing configurations
- Be mindful of the memory footprint of mwan3track. Many copies of this can be running, so importing all of mwan3.sh could have a moderate increase in memory usage. If there are some functions that are needed, those could be moved from mwan3.sh to common.sh
- Body of for loops should be indented
- Please test on a router without ipv6 support. This changes a lot of things behavior of the ipv6 stack, and we should confirm that there is no regression for  #11826

ptpt52:
@aaronjg @feckert 

I am going to close this P/R, it is out of date.

and for now I fork another mwan3(https://github.com/x-wrt/com.x-wrt/tree/master/net/mwan3plus) code and developed another path, using in my product.

The simple description about the difference:
it use the iptables rule policy select the line,
and then
Only need to match the routing tables to select the routes

and drop the matching rules for `mwan3_connected` set

for connected networks, it just match in the routing table
This is my goal of introducing `mwan3rtmon` in the past

All traffic control is controlled by routing tables
Every routing tables will include the direct routes, that is what `mwan3rtmon` do

take an example to explain:
```
the main table:
default via 192.168.17.1 dev eth0.1 proto static metric 40 
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
default via 192.168.16.1 dev wlan0 proto static src 192.168.16.189 metric 50 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.0/24 dev wlan0 proto static scope link metric 50 
192.168.17.0/24 dev eth0.1 proto static scope link metric 40 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.27 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.28 

the number 3 table:
default via 192.168.18.1 dev pppoe-xwan02 metric 42 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 metric 256 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 metric 256 
192.168.16.0/24 dev wlan0 proto static scope link metric 50 
192.168.17.0/24 dev eth0.1 proto static scope link metric 40 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.28 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.27 metric 2 

the number 2 table:
default via 192.168.18.1 dev pppoe-xwan01 metric 41 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 metric 256 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 metric 256 
192.168.16.0/24 dev wlan0 proto static scope link metric 50 
192.168.17.0/24 dev eth0.1 proto static scope link metric 40 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.27 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.28 metric 3 

...

for client---->target 192.168.18.1:
1. match iptables rules, balanced policy, and select table 2(could be 1, 2, 3) in random,

2. lookup table 2 and get outdev pppoe-xwan01 as result and that works

for any flow come in from wan, from any source, lookup table of that wan, 
and the reply direction traffic also lookup that table
```

we now don't care about it is connected (by matching ipset) or not, 
This is suitable for most network tops/cases

aaronjg:
Thanks for the update. If you are designing for a more constrained ecosystem, it is likely that you could remove quite a bit of complexity from the package that has been introduced to accommodate different types of connections and work around various bugs in other packages. You will have a lot more freedom to explore changes since you now no longer have legacy stakeholders to support, which sounds like a big advantage for the development of your router firmware. Best of luck!

feckert:
@ptpt52 @aaronjg 
I find it a pity that there is no agreement here.

@ptpt52 is it really necessary to start a fork here?
Time has shown that a fork only brings disadvantages, for both existing project and for the fork itself.
1. The testers are splitted
2. The developers are splitted

@ptpt52 Could you please mention again exactly and in detail what you do not like about the current mwan3 implementation.
I would like to see your approach with ipv6 in mwan3.
But currently other topics were more important.

I think it is really better to pull together than to split

ptpt52:
@feckert hi
I need a working version that matches my network topology. The current mwan3 is not perfect as I expect, and it is difficult for me to implement the functions I need based on the current version.
The required features include:
1. Dual stack support
2. The backhaul routing is working properly. (issue https://github.com/openwrt/packages/issues/13907)

I will continue to pay attention to the development of mwan3, wait until it is suitable, and when I have time, I will do it again.
Now I need a ready-to-use version in x-wrt, so I fork a branch separately

Thanks to @aaronjg  for the performance optimization of mwan3 and the function of WRAP.

But the way to implement a procd resident process in mwan3 is not yet acceptable to me. 
Let’s observe and take a look at the follow-up development, we can’t give a conclusion at present

aaronjg:
It would be nice to add dual stack support so users don't need to create additional device aliases to run that. Aside from the merge conflicts, your P/R still would need some work to be able to handle this correctly in all cases. If you'd like to resolve the merge conflicts and incorporate feedback, I can review again. Though, if you decide to do so, please separate the dual stack configuration P/R from the run only one stack at a time P/R to make it easier to review.

Regarding the routing from WAN initatiated traffic. Your assessment is correct that this will not work properly in mwan3. The situation you describe in the P/R is very specialized. It is two WANs that share a common gateway and traffic is originating at the gateway. I don't fully understand what having this setup accomplishes, but I trust that you have a good reason to use it. 

The issue happens because the traffic is originated from a member of the directly connected ipset, so it does not get the fwmark for the interface it comes in on. While it appears to be a simple fix, to resolve the issue in mwan3, we will likely need to make some changes to the mwan3_connected rules. However, this part of the code is very old ([goes beginning of git history](https://github.com/openwrt/packages/commit/7afdbd28cf453d02c28c933b434fdb93e3cd84f2#diff-c5a0098c4f4698e13dd6dd29799b432ef56193583dcee1b3f7a5a29334be00ccR91)) so in making the changes, we will need to be very careful to not break existing configurations.

Thus, for the time being, I would rather have a known bug in this very specialized case, rather than risk introducing new, unknown bugs by changing this very old bit of code, and potentially breaking many other configurations. 

The issue described in #13909, is similar in appearance, but is a more intermittent issue and seems to be triggered by an issue in a different part of the code. I am hoping to fix that, but I am waiting on more data from James to help recreate the issue.

Issue #12928, opened 2020-07-25
Minidlna under master branch not enabled but show service running!!


zxlhhyccc:
![image](https://user-images.githubusercontent.com/45259624/88454680-d19deb00-cea3-11ea-8643-98f10b02e2f3.png)

Issue #12837, opened 2020-07-16
miniupnpd: Causes basename command output to be shown when mwan3 is restarted


jamesmacwhite:
Maintainer: @neheb (Sorry I don't know who's the true maintainer?)
Environment: 19.07.3 Linksys WRT3200ACM

Description:

The summary of #12806 is that when I use mwan3 and use a command like `mwan3 restart` I was getting the the following output below constantly in console. Originally it looked like it was mwan3 itself, but it is actually coming from miniupnpd. I determined this after echoing the initscript $1 variable in /etc/rc.common and watched the output. This led me to discover it was miniupnpd causing the output and confirmed this when I completely remove the package with opkg, the console output does not occur.

```
BusyBox v1.30.1 () multi-call binary.

Usage: basename FILE [SUFFIX]

Strip directory path and .SUFFIX from FILE
uci: Entry not found
uci: Entry not found
```

I'm not sure why a basename call is essentially empty, but I now know it is tied to miniupnpd and procd maybe?

neheb:
The miniupnpd init script does not use basename. No idea what your issue is.

neheb:
Try:

```
--- a/net/miniupnpd/files/miniupnpd.init
+++ b/net/miniupnpd/files/miniupnpd.init
@@ -164,8 +164,7 @@ upnpd() {
 
 		[ -z "$uuid" ] && {
 			uuid="$(cat /proc/sys/kernel/random/uuid)"
-			uci set upnpd.config.uuid="$uuid"
-			uci commit upnpd
+			config_set uuid config uuid
 		}
 
 		[ "$uuid" = "nocli" ] || echo "uuid=$uuid"
```

jamesmacwhite:
@neheb Yes it is very strange, I still don't know exactly what is causing this, updating the init with that change does not resolve it. Perhaps miniupnpd is triggering something else, but it goes away when I remove the package.

neheb:
Fixed.

PR #13051, opened 2020-08-07
net/miniupnpd: check by /etc/init.d/miniupnpd running

ptpt52:
use `/etc/init.d/miniupnpd running` API

### DIFF
--- a/net/miniupnpd/Makefile
+++ b/net/miniupnpd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=miniupnpd
 PKG_VERSION:=2.1.20200510
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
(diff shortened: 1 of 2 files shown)


neheb:
Please bump PKG_RELEASE.

Does this fix https://github.com/openwrt/packages/issues/12928 ?

ptpt52:
when I run `mwan3 restart`, some error msg out:
```
BusyBox v1.31.1 () multi-call binary.

Usage: basename FILE [SUFFIX]

Strip directory path and .SUFFIX from FILE
```
where the error msg comes from `proc_lock()` in `/lib/functions/procd.sh`, which is invoked by the hotplug script `50-miniupnpd`

I found this is because the `initscript` variable is empty
I’m not very clear but I know it’s not correct to use `procd.sh` this way

> Does this fix #12928 ?

not related

neheb:
Sounds like it fixes https://github.com/openwrt/packages/issues/12837

PR #13058, opened 2020-08-09
netperf: fix compilation with GCC10

guidosarducci:
**Maintainer:** @tohojo 
**Compile tested:** malta/mips32-be on master branch
**Run tested:**  malta/mips32-be on master branch, tested under QEMU using a speedtest

**Description:**
GCC10 defaults to -fno-common, which breaks compilation when there are
multiple definitions of implicit "extern" variables. Remove extra definitions.

Thanks!


neheb:
Huh cool. I tried fixing this locally but couldn't figure it out.

PR #13059, opened 2020-08-09
treewide: replace `which` with `command -v`

aparcar:
Fix shellcheck SC2230
> which is non-standard. Use builtin 'command -v' instead.

Once applied to everything concerning OpenWrt we can disable the busybox
feature `which` and save 3.8kB.

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -12,7 +12,7 @@ PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.7.8
 # Release == build
 # increase on changes of services files or tld_names.dat
-PKG_RELEASE:=21
+PKG_RELEASE:=22
 
 PKG_LICENSE:=GPL-2.0
 PKG_MAINTAINER:=
--- a/net/ddns-scripts/files/ddns.defaults
+++ b/net/ddns-scripts/files/ddns.defaults
@@ -33,7 +33,7 @@ timer2dhms() {
 
 # using function to not confuse function calls with existing ones inside /lib/functions.sh
 update_config() {
-	uc_uci="$(which uci) -q"	# ignore errors
+	uc_uci="$(command -v uci) -q"	# ignore errors
 	uc_cfg=""
 	uc_name=""
 	uc_var=""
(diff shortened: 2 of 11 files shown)


aparcar:
For `pcuutils` which uses `which` in a script I proposed an [update](https://github.com/pciutils/pciutils/pull/44)

neheb:
ping @openwrt/package-maintainers

luizluca:
That might break dozens of scripts. Is it really worth it?
Nobody tests if which is installed (as it is a tool to check if something is available).
It's normally considered universally available.

I do have scripts that use it and I would not bother to change which to command -v. However it might be tons of references, both in scripts visible inside git but also from external sources downloaded during the build. Even binaries might use it (as it's easier to call a program than a shell built-in command from anything but a shell).

It might be a better way to deprecate it. Patch it t log to klog when it is used? Use a simple wrap script during transition? As 4M support is gone, I think it is too much trouble for less than 4k. Compare with how long it took to get rid of ifconfig.

aparcar:
We should start rejecting patches/PR including `which` and instead only accept `command -v`, easily done by running Shellcheck over all packages.

Even with 8MB of flash we should try to keep OpenWrt as slim as possible and remove nonsense - in my opinion.

A wrapper script would look like that and could be added to `base-files`

```shell
cat /bin/which
#!/bin/sh
command -v "$@"
```

Which is somewhat around 27 Bytes

champtar:
Removal of which applet likely will be NACKed because of the risk of silent breakage of many downstream distro. I'm ok with this PR.

aparcar:
Well I think so too, but what about the wrapper script? You don't think that's enough? I never saw anyone using which with the -a param.

The wrapper could also include a stderr message mentioning that it's outdated.

champtar:
I would chase other topics, like ca certs vs ca bundle, have most ssl enabled package have a wolfssl option if possible now that it's shipped by default, that will make a bigger positive impact. 4M is dead, but 8M is getting full pretty fast, and I'm sure there are bigger low hanging fruit than 4k (uncompressed ?). I'm not the one you need to convince, and small gain for possible silent breakage is always hard to sell. Don't get me wrong I like your work, you can try to send a patch on the ML, but don't fight too long for it ;)

hnyman:
"which" is so common and traditional that I feel strange if it would be missing (from busybox).

It may not be official POSIX, but I have been using "which" since year 1989 and I had never heard about "command -v" before this discussion...

diizzyy:
This is in the wrong order, first get approval by upstream then we can work on it.

neheb:
Works for me.

aparcar:
> This is in the wrong order, first get approval by upstream then we can work on it.

It's been approved upstream [before](https://github.com/openwrt/openwrt/commit/56f813674a912490df327304033bf667b285930a#diff-905b668d89bc96e038b343065bf462b2) :)

PR #13062, opened 2020-08-10
mpd - remove ffmpeg support and add opus

neheb:
Maintainer: @thess 
Compile tested: ath79

ffmpeg in MPD is decode only. There's no encoding support. Pursuant to the recent BUILD_PATENTED discussion on the mailing list, remove it.

Added libopus support. ffmpeg seems to be missing this. opus is a popular format nowadays.

### DIFF
--- a/sound/mpd/Makefile
+++ b/sound/mpd/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mpd
 PKG_VERSION:=0.21.25
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://www.musicpd.org/download/mpd/0.21/
@@ -18,6 +18,7 @@ PKG_LICENSE:=GPL-2.0-or-later
 PKG_LICENSE_FILES:=COPYING
 
 PKG_CONFIG_DEPENDS:= \
+	CONFIG_BUILD_PATENTED \
 	CONFIG_IPV6 \
 
 PKG_INSTALL:=1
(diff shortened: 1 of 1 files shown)


karlp:
-1, I'd say "without a clear resolution on the list, make no changes"  You're making changes to establish a new ground, _without_ that consensus.

neheb:
I actually question whether ffmpeg support is even useful. The full list of formats is here: https://github.com/MusicPlayerDaemon/MPD/blob/master/src/decoder/plugins/FfmpegDecoderPlugin.cxx#L665

Some of those are compile time disabled by the current ffmpeg package, meaning they won't work at all.

I added AAC support through libfaad2, which ffmpeg does not have. ffmpeg is also not compiled with opus support. Also no musepack.

Also note that for the longest time, the MPD ffmpeg plugin was a bit broken: https://github.com/MusicPlayerDaemon/MPD/commit/0914644d2b9d380587762bb2acfc7d3fa6c32704#diff-e0b48c578aeeb87a844fdf924bd67755
https://github.com/MusicPlayerDaemon/MPD/commit/50003f6ad2465c0612594ec5635da73096e7d7cf#diff-e0b48c578aeeb87a844fdf924bd67755

thess:
@neheb - I disagree with your assessment of the utility of ffmpeg with MPD. However, with that said, I would really like it if you would remove my name as maintainer. If you do not wish to do that, I will do it directly myself in a couple of days. [In fact, I will set aside some time and remove myself from a number of other packages as well.]

neheb:
Removed.

karlp:
So, on the mailing list, you _yourself_ say you would like a decision, "I'd like a decision on this issue. Debian to my knowledge does not
sell anything whereas Red Hat is the most profitable Linux company out
there, which makes sense as to why they would want to shield
themselves from litigation."

But no, you're just going to make the decision on the ground real first, you don't actually want any consensus, you don't really seem to care about any differing concerns here at all!

PR #13064, opened 2020-08-10
linknx: fix compilation without sys/cdefs

neheb:
sys/cdefs.h is not included with musl. It's also deprecated.

Maintainer: @tru7 
Compile tested: ath79


tru7:
fine with me. I'll watch and remove the patch when source code is changed upstream

PR #13065, opened 2020-08-10
knxd: fix compilation without sys/cdefs

neheb:
sys/cdefs.h does not come with musl. It's also deprecated.

Maintainer: @tru7 
Compile tested: ath79


tru7:
I check with the developper if this could be changed upstream.

tru7:
fine with me. I'll watch and remove the patch when source code is changed upstream

PR #13069, opened 2020-08-10
micropython: fix compilation without sys/cdefs

neheb:
sys/cdefs.h is not provided by musl. micropython also includes its own
copy.

Maintainer: @roger- 
Compile tested: ath79


karlp:
For this, and allll the other cdefs patches, have you filed them all upstream as well?  Especially this one, where they even have their own cdefs, should be a no-brainer

neheb:
Seems like upstream is somewhat aware of the issue: https://github.com/pfalcon/berkeley-db-1.xx/commit/3f80a8c422a2aa40e71403b9c26d492d49a9ab82

PR #13071, opened 2020-08-10
quota: fix compilation without sys/cdefs

neheb:
sys/cdefs.h does not come included with musl. It's also deprecated.

Maintainer: @flyn-org 
Compile tested: ath79


flyn-org:
Looks good. Recommend merge.

PR #13080, opened 2020-08-10
minidlna: fix GCC 10 multiple def compilation error

neheb:
Diffrent file use its own setjmp_buffer, thus
we have two global variables called setjmp_buffer

I am not sure if we should use only one instance of it.

The patch sent upstream uses a similar approach.

https://sourceforge.net/p/minidlna/bugs/327/

Signed-off-by: Syrone Wong <wong.syrone@gmail.com>

Maintainer: @openwrt/package-maintainers 
Compile tested: ath79

ping @wongsyrone

### DIFF
--- a/multimedia/minidlna/Makefile
+++ b/multimedia/minidlna/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=minidlna
-PKG_RELEASE:=7
+PKG_RELEASE:=8
 
 PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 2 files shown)


wongsyrone:
As I stated in the commit message, I am not sure if we should use only one instance of it. The patch sent upstream just rename one of them into a different name. I choose to limit the variable to a compilation-unit scope via static keyword.

neheb:
The comment states:

```
/* For libjpeg error handling */
```

Does it really matter?

wongsyrone:
It seems minidlna works as expected.

PR #13090, opened 2020-08-11
glib2: add full language support hack to fix compilation

neheb:
Otherwise, dependent packages need to all add -lintl to LDFLAGS.

Maintainer: @tripolar 
Compile tested: ath79

ping @micmac1

### DIFF
--- a/libs/glib2/Makefile
+++ b/libs/glib2/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=glib2
 PKG_VERSION:=2.65.0
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 PKG_SOURCE:=glib-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@GNOME/glib/2.65
@@ -24,6 +24,7 @@ PKG_BUILD_DIR:=$(BUILD_DIR)/glib-$(PKG_VERSION)
 HOST_BUILD_DIR:=$(BUILD_DIR_HOST)/glib-$(PKG_VERSION)
 PKG_BUILD_DEPENDS:=meson/host gettext-full libiconv/host
 HOST_BUILD_DEPENDS:=meson/host gettext-full/host libiconv/host libffi/host
+PKG_CONFIG_DEPENDS:=CONFIG_BUILD_NLS
 PKG_INSTALL:=1
 
 include $(INCLUDE_DIR)/host-build.mk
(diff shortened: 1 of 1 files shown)


neheb:
@micmac1 regarding that ICONV_FULL hack.

I see that -liconv is under Libraries.private. Should this be done here as well?

micmac1:
> @micmac1 regarding that ICONV_FULL hack.
> 
> I see that -liconv is under Libraries.private. Should this be done here as well?

Hi Rosen, I removed "-lintl" from the glib2 pc file and then got this when attempting to compile json-glib:

```
json-glib-validate.c.o: undefined reference to symbol 'libintl_textdomain'
libintl.so.8: error adding symbols: DSO missing from command line
```

Adding "-lintl" to "Libs.private" didn't prevent the above from happening.

Interestingly, when "-lintl" is used, the resulting libjson-glib.so doesn't link to libintl. Probably because of the use of "--as-needed".

To summarize, to me it looks like the "-lintl" should stay for now. There's probably some bug somewhere, but I don't know where I'd look for it.

Kind regards,
Seb

micmac1:
Hi Rosen,

I checked a bit more. glib2's meson.build runs check `cc.has_function('ngettext', args : osx_ldflags)` and, if successful, it never looks for non-libc libintl. For musl this test is always successful. For glibc as well, I tested on Debian:

```
$ cat test.c
#include <stdio.h>

int main(void)
{
	ngettext();
}
$ gcc -o test test.c 
test.c: In function ‘main’:
test.c:5:2: warning: implicit declaration of function ‘ngettext’ [-Wimplicit-function-declaration]
    5 |  ngettext();
      |  ^~~~~~~~
$ echo $?
0
```

glib2 unconditionally sets ENABLE_NLS, so during compile `<libintl.h>` gets included always. But then we have a disconnect when the OpenWrt pkg is being built with BUILD_NLS=y, because the `<libintl.h>` will be from libintl-full but glib2 will not link to libintl-full.

With BUILD_NLS=n there's no problem, because the `<libintl.h>` will be from libc.

In lieu of proper libintl detection in glib2 meson build I think removing the SED from Makefile together with the below patch would do:

```
cat feeds/packages/libs/glib2/patches/007-openwrt-libintl.patch 
--- a/meson.build
+++ b/meson.build
@@ -2045,8 +2045,11 @@ endif
 # proxy-libintl subproject.
 # FIXME: glib-gettext.m4 has much more checks to detect broken/uncompatible
 # implementations. This could be extended if issues are found in some platforms.
+#
+# 007-openwrt-libintl.patch: with BUILD_NLS=y use libintl-full, else use libc
+#
 libintl_deps = []
-if cc.has_function('ngettext', args : osx_ldflags)
+if get_option('nls').disabled() and cc.has_function('ngettext', args : osx_ldflags)
   have_bind_textdomain_codeset = cc.has_function('bind_textdomain_codeset')
 else
   # First just find the bare library.
```

Compiles fine with BUILD_NLS y and n here.

On another note, `include/nls.mk` still adds libintl-stub include and lib path with BUILD_NLS=n, even though the gettext package was removed. This is mere cosmetic, unless somebody still has the remnants lying around in their build dir. That should also be cleaned up.

Kind regards,
Seb

neheb:
Sounds great. Would love a PR.

PR #13096, opened 2020-08-12
unbound: support hostnames that are already FQDNs

stuartpb:
Maintainer: @EricLuehrsen
Compile tested: no
Run tested: no

Description: The current version of unbound's odhcpd integration script inadvertently drops records for DHCP leases that specify a multi-level hostname containing dots (ie. a fully-qualified domain name).

There are a few ways this could be handled - ideally, based on configuration variables:

- interpret any multi-level hostnames as fully-qualified domains
  - this is the approach taken by this PR
- interpret multi-level hostnames as subdomains of the local domain
- switch between subdomain/FQDN behavior based on the presence of a trailing `.`
  - I believe this is the approach taken by dnsmasq
- switch between subdomain/FQDN behavior based on pattern matching
  - eg. interpret `.internal` hostnames as FQDNs, and everything else as subdomains of the local domain
- rewrite or block multi-level hostnames based on pattern matching
  - eg. accept leases for `.internal`, discard leases for `.com`, treat everything else as subdomains

I don't have a testbed environment set up, so this patch is just something I threw together via the GitHub patch UI - definitely needs maintainer review.


EricLuehrsen:
I have a concern on use case and scope creep. So like dnsmasq, unbound+odhcpd is a cheap small LAN configuration to build pseudo-authoritative DNS and DHCP all into one small package. With OpenWrt it should all flash under 8MB. When these cases come up, then it starts to grow into a situation where dhcpd, kea, bind, or nsd are needed in some combination. RFC4702 is complex and odhcpd doesn't seem to handle all the nuances. If odhcpd is given a domain to assign (the controlling factor here), then it should override the S bit to zero and set the O bit. It should prune the FQDN.

In a normal small business or residence a computer is just "fred" and "fred" gets added to "fred.internal." The domain "internal." is authoritative to the local site and its all nice and tidy.

In a data center run by "example.net." all the physical machines have their cabinet numbers like "host-4gb21.example.net." However the virtual machines are spun up within them for other companies, such as their offsite web servers, "www.example.com." and "www.example.org." They are within the same numerical subnet "192.2.0.0/16" but have authoritative host entries in completely different servers. "example.org." may not be authoritative at this data center, and so the DHCP server here cannot add "www.example.org." anywhere. It just knows "192.2.11.101" was assigned to it.

EricLuehrsen:
I have a concern on use case and scope creep. So like dnsmasq, unbound+odhcpd is a cheap small LAN configuration to build pseudo-authoritative DNS and DHCP all into one small package. With OpenWrt it should all flash under 8MB. When these cases come up, then it starts to grow into a situation where dhcpd, kea, bind, or nsd are needed in some combination. RFC4702 is complex and odhcpd doesn't seem to handle all the nuances. If odhcpd is given a domain to assign (the controlling factor here), then it should override the S bit to zero and set the O bit. It should prune the FQDN.

In a normal small business or residence a computer is just "fred" and "fred" gets added to "fred.internal." The domain "internal." is authoritative to the local site and its all nice and tidy. If "fred" thinks he is still "fred.someplace.else." then the DHCP server will correct him (S and O bits).

In a data center run by "example.net." the physical machines may be cabinet numbers like "host-4gb21.example.net." However the virtual machines are spun up for other companies such as their offsite web servers "www.example.com." and "www.example.org." Physical and virtual are within the same numerical subnet "192.2.0.0/16" but have authoritative host entries in completely different home. "example.org." may not be SOA at this data center, and so the DHCP server here cannot add "www.example.org." here. It just knows "192.2.11.101" was assigned to it. Of course the pointer records within "192.2.0.0/16" have SOA in the data center. So there are CNAME involved or CDIR PTR are delegated back to the customers authoritative server... more... Ugly.

stuartpb:
> If odhcpd is given a domain to assign (the controlling factor here), then it should override the S bit to zero and set the O bit. It should prune the FQDN.

It uses the domain that ships with the base firmware as `.local`, right? How should it be handling the removal of the domain to assign - check if the string is empty?

> In a normal small business or residence a computer is just "fred" and "fred" gets added to "fred.internal." The domain "internal." is authoritative to the local site and its all nice and tidy. If "fred" thinks he is still "fred.someplace.else." then the DHCP server will correct him (S and O bits).

In that case, shouldn't unbound still be configured to return records for fred.internal (the corrected address it sounds like odhcpd would be returning to the client here)?

> In a data center[...]

I understand the concerns about scope creep; I can put together a user story over the weekend of a reasonable scheme for a small business / residence to want to handle.

EricLuehrsen:
Any update on what you would like to see here?

BKPepe:
Closing, once you wants to continue on this PR, lets reopen this.

stuartpb:
Yeah - after initially authoring this PR, I looked into it a little more, and though I can't really articulate the exact problems, it became increasingly clear that FQDNs don't really mix with the common expectations of hostnames, in not just DHCP but the kernel/libc and consumers thereof (someone more qualified than me probably ought to write a blog post elaborating on this).

PR #13098, opened 2020-08-13
clamsmtp: remove

neheb:
Judging from https://salsa.debian.org/clamav-team/clamsmtp/-/tags ,
it seems that this was last updated 11 years ago. The domain is also
gone.

Maintainer: @lucize 

Is this package needed?


lucize:
no, clam-milter is now available so you can drop this

PR #13099, opened 2020-08-13
gkermit: replace PKG_SOURCE_URL

neheb:
It's been down for quite a while. HTTP is also more reliable.

Maintainer: @nwf


nwf:
:+1: LGTM

PR #13105, opened 2020-08-13
coreutils: don't use non-portable path to bash

pstef:
As /bin/bash is not guaranteed to exist, use the less non-portable /bin/sh.
I thought of replacing it with `env bash`, but apparently, sh works here too.
Build-tested on FreeBSD.

### DIFF
--- a/utils/coreutils/Makefile
+++ b/utils/coreutils/Makefile
@@ -134,7 +134,7 @@ CONFIGURE_ARGS += \
 define Build/Compile
 	$(MAKE) -C $(PKG_BUILD_DIR) \
 		DESTDIR="$(PKG_INSTALL_DIR)" \
-		SHELL="/bin/bash" \
+		SHELL="/bin/sh" \
 		all install
 endef
 


utils/coreutils/Makefile
```diff
@@ -134,7 +134,7 @@ CONFIGURE_ARGS += \
 define Build/Compile
 	$(MAKE) -C $(PKG_BUILD_DIR) \
 		DESTDIR="$(PKG_INSTALL_DIR)" \
-		SHELL="/bin/bash" \
+		SHELL="/bin/sh" \
```
karlp, line 137: can you just drop the SHELL= bit altogether?  It would seem that it was only added because someone felt it _needed_ bash?

pstef: I just tested and that works for me as well.

neheb: Wonder if this entire section can be removed and replaced with PKG_INSTALL.

karlp: yeah, I was starting to look it looked pretty generic

neheb: Already did this.

PR #13107, opened 2020-08-13
adblock: change polish filter list

matx1002:
Change polish filter list to general list in adblock package. Previous list was made for polish scam sites.


matx1002:
OK, as you think. Two lists will be fine. The previous list only blocked some Polish scam sites. Instead, I thought that ad blocking list would be better for this project, but if you want to add both of them, it'll be great.

dibdot:
superseded by #13124

PR #13109, opened 2020-08-14
lcd4linux: fix samsung driver compilation

neheb:
Turns out, it uses an API that is not provided by libjpeg.

Add compatibility code from v4l-utils.

Maintainer: @feckert 
Compile tested: ath79

### DIFF
--- a/utils/lcd4linux/Makefile
+++ b/utils/lcd4linux/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lcd4linux
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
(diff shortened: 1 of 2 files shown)


neheb:
ping @feckert 

The patch is identical to https://github.com/openwrt/packages/commit/26b1cc700a838920e490cec1b5111e01850dc538#diff-06a8edd35a8146b1e29d160f8bb1f28f .

neheb:
Merging. There's nothing special here.

feckert:
@neheb Thank you. And sorry I didn't answer. I was on vacation.

Since the upstream project is no longer maintained by lcd4linux, and we use my fork. Is the question if we don't throw the patches from openwrt on my fork?

neheb:
Feel free to do so.

PR #13110, opened 2020-08-14
u2pnpd: remove index API usage

neheb:
(r)index is deprecated and replaced with str(r)chr.

Maintainer: @mhei 
Compile tested: ath79

### DIFF
--- a/net/u2pnpd/Makefile
+++ b/net/u2pnpd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=u2pnpd
 PKG_VERSION:=0.5
-PKG_RELEASE:=2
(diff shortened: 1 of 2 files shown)


karlp:
Have you pushed this upstream?  Is there an option in musl that will let you drop (r)index?  Otherwise this is just adding patches locally that do nothing at all.

neheb:
@mhei is upstream.

mhei:
I added this upstream, but I have one other change pending which needs more testing before creating a new upstream release. Since I do not see any urgent importance of this "fix", I'm closing this PR since it will be part of next version bump.

mhei:
For reference: https://github.com/mhei/u2pnpd/commit/eee14f2183cb0bc59a43e7ec761b6c52c714b5d6

PR #13113, opened 2020-08-14
net-snmp: fix compilation without sigblock

neheb:
sigblock is deprecated. Use newer APIs.

Maintainer: @stintel 
Compile tested: ath79


stintel:
NACK, upstream behaviour changed in 5.9. If this is not good enough, please submit your changes upstream first, and request backport once someone found the time to bump net-snmp to 5.9. I've just returned from a 2 week holiday, it might be a while before I get to it.

neheb:
I did submit it upstream. They applied a different solution. Was going to update this PR. I'll resend some other time.

PR #13114, opened 2020-08-14
tracertools: remove usleep usage

neheb:
usleep is deprecated.

Modernized the Makefile to modern standards.

Maintainer: @dangowrt 
Compile tested: ath79


dangowrt:
I've pushed the change into the upstream repo instead of adding a local patch.
https://github.com/openwrt/packages/commit/52ea87c5d7175825d48fbeccd0b9b608ad036bae
Thank you for taking care of that, I'm far from any hardware to test changes (doesn't look like this change could break anything where you'd need to have the hardware to find out).

Issue #12812, opened 2020-07-14
canutils:  Subcomponents do not build after commit 5f9a918c4fe8d3e6e3a29f653204683881ae6f07


Entropy512:
Maintainer: @neheb 
Environment: Latest git master, built on amd64, target for ar750s

Description:

The listed commit states "Revert b70cb0e : that commit was misleading

How is the following misleading?
```
Collected errors:
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for canutils-asc2log:
 *      canutils
 * opkg_install_cmd: Cannot install package canutils-asc2log.
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for canutils-bcmserver:
 *      canutils
 * opkg_install_cmd: Cannot install package canutils-bcmserver.
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for canutils-canbusload:
 *      canutils
 * opkg_install_cmd: Cannot install package canutils-canbusload.
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for canutils-can-calc-bit-timing:
 *      canutils
 * opkg_install_cmd: Cannot install package canutils-can-calc-bit-timing.
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for canutils-candump:
 *      canutils
 * opkg_install_cmd: Cannot install package canutils-candump.
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for canutils-canfdtest:
 *      canutils
 * opkg_install_cmd: Cannot install package canutils-canfdtest.
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for canutils-cangen:
 *      canutils
 * opkg_install_cmd: Cannot install package canutils-cangen.
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for canutils-cangw:
 *      canutils
```
Above repeats for every canutils subcomponent that is selected unless DEPENDS:=canutils is removed again

neheb:
Removing that causes a pointless coreutils metapackage with all of the subpackages to be outside of it.

Entropy512:
No it doesn't.  The current DEPENDS line causes a dependency on a "pointless canutils metapackage" that, in fact, prevents any of the canutils packages from getting installed (as seen in the log snippet shown) because the metapackage is not built/does not exist.

Remove the DEPENDS line and:
No metapackage is built (no change)
No metapackage is required either (change that allows individual canutils packages to be installed)

[...]

PR #13126, opened 2020-08-14
canutils: Fix CAN Utilities and Tools dependency.

pffmachado:
Instead added a Submenu where CAN utilities could be placed

PKG_RELEASE bumped to 3.

Signed-off-by: Paulo Machado <pffmachado@yahoo.com>

Compile tested: x86, generic, OpenWrt master branch

### DIFF
--- a/utils/canutils/Makefile
+++ b/utils/canutils/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=canutils
 PKG_VERSION:=2020.02.04
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/linux-can/can-utils/tar.gz/v$(PKG_VERSION)?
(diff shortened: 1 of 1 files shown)


neheb:
I don't like this.

What's the current problem? Installing a can utility works just fine currently.

pffmachado:
No! It doesn't work, it is breaking the build!
I think the solution is the more readable and makes sense compared with comparable situations!

You can ignore this PR, but please find a solution for it!

neheb:
Yes it does. Fix your config.

I just tested installing canutils-asc2log. Works perfectly fine.

pffmachado:
Well! I think there is a Issue open with the same problem https://github.com/openwrt/packages/issues/12812

I've a couple of Build Systems doing the build from scratch, clean state, and all are failing with that issue for x86-generic targets!
Turns out the dependency on canutils make it break at install stage!

neheb:
Oh, that's embarrassing. I have it fixed locally but did not push it here.

neheb:
Just pushed. Please test again.

neheb:
https://github.com/openwrt/packages/commit/d34430f02f33765770e4fc748094a0a047c73265

pffmachado:
Yes! It is working now!

Thanks for fixing it!

PR #13127, opened 2020-08-14
openvswitch: backport patch to fix build against kernel 4.14.193

NeoRaider:
(and possibly other recent stable kernels)

Maintainer: @yousong
Compile tested: ar71xx/generic
Run tested: none

After fixing the build against kernel 4.14.187, it broke again for kernel 4.14.193...


yousong:
> After fixing the build against kernel 4.14.187, it broke again for kernel 4.14.193...

Yeah, it's not uncommon with openvswitch kmod code.  Thanks for taking care of this ;)

PR #13129, opened 2020-08-15
mpg123: update to 1.26.3

neheb:
Switched to CMake. CMake compiles faster.

Added libsyn123 library as it's needed since version 1.26.

Before:

time make package/mpg123/compile -j 12
Executed in   16.05 secs   fish           external
   usr time   26.65 secs    0.00 micros   26.65 secs
   sys time    4.82 secs  833.00 micros    4.81 secs

After:

time make package/mpg123/compile -j 12
Executed in   12.12 secs   fish           external
   usr time   19.35 secs    0.00 micros   19.35 secs
   sys time    3.22 secs  752.00 micros    3.22 secs

Maintainer: @wigyori 
Compile tested: ath79

Setting as draft since other platforms probably need to be tested.

@jefferyto optimized defaults are available for several platforms. Do you think there's any point in applying those or just keep them generic as is? Currently only NEON, Aarch64, and x86 generates optimized code. There's also code for PPC(with/out-FPU) and ARM(with/out FPU).

### DIFF
--- a/sound/mpg123/Makefile
+++ b/sound/mpg123/Makefile
@@ -8,21 +8,22 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mpg123
-PKG_VERSION:=1.25.13
-PKG_RELEASE:=3
+PKG_VERSION:=1.26.3
+PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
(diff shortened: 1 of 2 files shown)


jefferyto:
I don't know what the performance improvements would be, but I think enabling optimizations would be good whenever possible.

neheb:
I tried adding altivec optimizations for the 464fp target. Seems it's more complicated than adding platform definitions.

Let's just keep it as is for now.

PR #13130, opened 2020-08-15
exfat-nofuse: remove

neheb:
Since the introduction of the official exfat-oot package in base, this
is now pointless.

Maintainer: @yousong 
Compile tested: ath79

https://github.com/openwrt/openwrt/commit/cd41234d2f63c484eeec619e0fb82ae54fe8979e


yousong:
Original plan was to deprecate it first and drop it on first break.  No need to wait with a pull request ready ;)

PR #13131, opened 2020-08-15
python3: Use hash-checking mode when installing host pip packages

jefferyto:
Maintainer: me, @commodo 
Compile tested: armvirt-32, 2020-08-11 snapshot sdk
Run tested: none

Description:
In [hash-checking mode][1], pip will verify downloaded package archives (source tarballs in our case) against known SHA256 hashes before installing the packages.

As a consequence, this requires the use of [requirements files][2] and pinning packages to known versions.

The syntax for package Makefiles has changed slightly; `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` no longer accepts requirement specifiers like `"foo>=1.0"`, only requirements file names (which are the same as package names in the most common case).

This also updates affected packages, in particular:

* python-zipp (ping @ja-pa): `setuptools_scm[toml]` has been split into `setuptools-scm toml` to reuse the requirements file for setuptools-scm (the extra depends installed by `setuptools_scm[toml]` is toml).

* python-pycparser: This previously used ply 3.10, whereas the requirements file will now install 3.11.

[1]: https://pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode
[2]: https://pip.pypa.io/en/stable/user_guide/#requirements-files


jefferyto:
Some additional notes:
* I spent a lot of time (too much time) thinking about how the requirements files should be organized, to maximize ease of use/ease of maintenance/flexibility/etc. I'm still not 100% happy with the current setup, so I'm open for suggestions.
* The requirements files duplicate information (version number and sha256 hash) from existing package Makefiles. I wish there was a way to (auto-)generate them from package Makefiles but I haven't figured out a good way to do this yet.
* I also want a way to run `pip list --outdated` (for host pip) to check for updates (since uscan won't be checking these packages for updates), but again I haven't found a good way to add a target for this in make.
* I think there is still a chance for package conflicts (e.g. two OpenWrt packages depend on conflicting versions of the same Python package), although our current usage is simple and packages that use host pip will not run in parallel (thus each OpenWrt package can override any previously-installed Python package). It may be worth using virtualenv in the future to isolate each OpenWrt package build.

@commodo I converted the Cython dependency for numpy into a "specific" version requirements file as I think you prefer to review each Cython update. I wonder if you would be open to changing it to a "latest" version requirements file (just named "Cython.txt"), since now all Python packages are pinned anyway and need to be manually updated.

commodo:
>     * I spent a lot of time (too much time) thinking about how the requirements files should be organized, to maximize ease of use/ease of maintenance/flexibility/etc. I'm still not 100% happy with the current setup, so I'm open for suggestions.

this looks fine from my side; if we find that too many files are too granular, we may change; but i doubt it;
considering the noise on this repo, granular files may be fine, as it would be easier to track history for certain changes;


>     * The requirements files duplicate information (version number and sha256 hash) from existing package Makefiles. I wish there was a way to (auto-)generate them from package Makefiles but I haven't figured out a good way to do this yet.

you could extend [later] the build rule to dump out a temp-file generated from the package's Makefile


>     * I also want a way to run `pip list --outdated` (for host pip) to check for updates (since uscan won't be checking these packages for updates), but again I haven't found a good way to add a target for this in make.

i would not worry too much about updating host-side packages; target-side packages are more important; 


>     * I think there is still a chance for package conflicts (e.g. two OpenWrt packages depend on conflicting versions of the same Python package), although our current usage is simple and packages that use host pip will not run in parallel (thus each OpenWrt package can override any previously-installed Python package). It may be worth using virtualenv in the future to isolate each OpenWrt package build.

i think we should see these version conflicts a bit more differently;
this offers some convergence for those packages [or package versions];
most of the time this won't cause issues (using version X when Y is recommended by the package authors);
if it does, we report it to them, ask to fix it, propose a fix, etc; or worst case, decide to drop the package, if the upstream people don't take care of their stuff too well;
i feel that virtualenv may be a bit too much; i'm not against it completely, but ¯\_(ツ)_/¯ 



> @commodo I converted the Cython dependency for numpy into a "specific" version requirements file as I think you prefer to review each Cython update. I wonder if you would be open to changing it to a "latest" version requirements file (just named "Cython.txt"), since now all Python packages are pinned anyway and need to be manually updated.

fine for `Cython.txt`; i just want us to have control over which version of Cython gets installed; and not `pip`; if we don't pin it, it updates underneath us and get breakage we have to deal with;
dealing with breakage is fine; but i'd like that we get more control over when it breaks, not pip

jefferyto:
Thanks @commodo - I've renamed `Cython-0.29.21.txt` to `Cython.txt`.

PR #13138, opened 2020-08-16
sqlite3: use eaccess to check permission

jjm2473:
The 'access' syscall uses REAL uid to check aceess,
but we should use EFFECTIVE uid with 'eaccess' call.
See man page for their difference.

Maintainer: not found
Compile tested: MacOS 10.15.6, OpenWRT 18.06
Run tested: aarch64 rtd1296, z9s, OpenWRT 18.06, tested by forked-daapd

Description:
Issue in forked-daapd,
Before patched, forked-daapd scanning multi files or doing database vacuum, 
 'unable to open database file' may appears in log,
Using strace to trace it, shows that:
```
setresgid(-1, 3003, -1)                 = 0
...
setresuid(-1, 190, -1)                  = 0
...
newfstatat(AT_FDCWD, "/var/tmp", {st_mode=S_IFDIR|0755, st_size=40, ...}, 0) = 0
faccessat(AT_FDCWD, "/var/tmp", W_OK|X_OK) = 0
faccessat(AT_FDCWD, "/var/tmp/etilqs_d4c84886ce52e72a", F_OK) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/var/tmp/etilqs_d4c84886ce52e72a", O_RDWR|O_CREAT|O_EXCL|O_LARGEFILE|O_NOFOLLOW|O_CLOEXEC, 0600) = -1 EACCES (Permission denied)
openat(AT_FDCWD, "/var/tmp/etilqs_d4c84886ce52e72a", O_RDONLY|O_EXCL|O_LARGEFILE|O_NOFOLLOW|O_CLOEXEC) = -1 ENOENT (No such file or directory)
```
As shown above, forked-daapd was run by root, then it sets its effective uid to 190, 
then sqlite trying to check '/var/tmp' writeable with `faccessat(AT_FDCWD, "/var/tmp", W_OK|X_OK)`, kernel says ok, but failed when actually create file with `openat`.
Because `faccessat` using real uid to check access by default, but `openat` using effective uid.
PS, The information of '/var/tmp' :
```
drwxr-xr-x    2 root     root            40 Aug 16 10:29 tmp
```


neheb:
Needs a PKG_RELEASE bump.

micmac1:
Hello @jjm2473 

Is this a bug in sqlite3? Or is this something that only affects OpenWrt?

Kind regards,
Seb

champtar:
Not sure we should carry such local patch that isn't in any other distro. I think we should either have it accepted by upstream sqlite3, or fix forked-daapd.

jjm2473:
@micmac1 Is bug of sqlite3, but the `eaccess` is **NOT** portable for other platforms. In `eaccess` man page, it says that `The eaccess() function was added to glibc in version 2.4.`.

I wrote a simple test code:
```
#define _GNU_SOURCE
#include <unistd.h>

int main() {
	return eaccess("/tmp", 07);
}
```
I tested these code in MacOS, then it link failed.
In Linux is OK.
In OpenWrt is OK.

I have reported the bug to sqlite3, maybe they have better solution.

champtar:
Can you link to the bug report ?

jjm2473:
@champtar https://www.sqlite.org/forum/forumpost/bf6271887e

neheb:
Needs to be rebased.

@champtar what's the verdict here?

champtar:
I'm still not a fan of having this patch as it impact all sqlite3 users, for a cosmetic problem.

I understand that forked-daapd runs on Linux/FreeBSD, so this bug is not specific to OpenWRT but more specific to forked-daapd.
Either you could see with sqlite to remove all 'access' usage as in any cases it's prone to TOCTOU
```
man eaccess
...
Warning: Using this function to check a process's permissions on a file before performing some operation based on that information leads to race conditions: the file permissions may change between the two steps. Generally, it is safer just to attempt the desired operation and handle any permission error that occurs. 
```

or you can talk to forked-daapd to see if they can set the real UID also to 190 ?

neheb:
ping @ejurgensen

PR #13139, opened 2020-08-16
openfortivpn: fix routes passed to netifd

aaronjg:
Maintainer: @lucize 
Compile tested: openwrt-sdk-19.07.3-ipq806x-generic_gcc-7.5.0_musl_eabi.Linux-x86_64
Run tested: 19.07.3

Description:

The openfortivpn routes are a bit different than the standard ppp
routes so we need to handle them with a custom ppp-up script.

Gateway should not be set, and src should be set to the PPP local ip
address.

### DIFF
--- a/net/openfortivpn/files/openfortivpn-ppp-up
+++ b/net/openfortivpn/files/openfortivpn-ppp-up
@@ -1,0 +1,20 @@
+#!/bin/sh
+PPP_IPPARAM="$6"
+
+. /lib/netifd/netifd-proto.sh
+proto_init_update "$IFNAME" 1 1
+proto_set_keep 1
+[ -n "$PPP_IPPARAM" ] && {
+	[ -n "$IPLOCAL" ] && proto_add_ipv4_address "$IPLOCAL" 32 "" "$IPREMOTE"
+	[ -n "$IPREMOTE" ] && proto_add_ipv4_route 0.0.0.0 0 "" "$IPLOCAL"
+	[ -n "$DNS1" ] && proto_add_dns_server "$DNS1"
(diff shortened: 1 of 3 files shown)


net/openfortivpn/files/openfortivpn-ppp-up
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+PPP_IPPARAM="$6"
+
+. /lib/netifd/netifd-proto.sh
+proto_init_update "$IFNAME" 1 1
+proto_set_keep 1
+[ -n "$PPP_IPPARAM" ] && {
+	[ -n "$IPLOCAL" ] && proto_add_ipv4_address "$IPLOCAL" 32 "" "${IPREMOTE:-2.2.2.2}"
```
feckert, line 8: Why do we set the IP to 2.2.2.2 if $IPREMOTE is not set?

aaronjg: I'm not sure. It does seem weird. I copied this from the ppp-up script. The only change was on line 9.

feckert: That is the commit that adds this address.
https://github.com/openwrt/openwrt/commit/5caa45d0b396972b313fb3d99908ba978b684570#diff-d7672a29ff6fe8d75e5edc9d5b7e95a7
The maintainer is not active anymore. So maybe @dedeckeh could help use hier?

lucize:
@neheb ping

net/openfortivpn/files/openfortivpn-ppp-up
aaronjg: Openfortivpn defaults to 192.0.2.1 which is an RFC 3330 "TEST-NET" ip address. Perhaps that would be a good choice here?

PR #13140, opened 2020-08-16
generate-ipv6-address: add ipv6 generator

PolynomialDivision:
Generates IPv6 addresses from a given prefix and either a given MAC-48 address (an Ethernet hardware address) or a randomly drawn host number.

### DIFF
--- a/ipv6/generate-ipv6-address/Makefile
+++ b/ipv6/generate-ipv6-address/Makefile
@@ -1,0 +1,38 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=generate-ipv6-address
+PKG_VERSION:=0.1
+PKG_RELEASE:=1
+
+PKG_SOURCE:=generate-ipv6-address-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


ipv6/generate-ipv6-address/Makefile
```diff
@@ -0,0 +1,37 @@
...
+PKG_NAME:=generate-ipv6-address
+PKG_VERSION:=0.1
+PKG_RELEASE:=1
+
+PKG_SOURCE:=generate-ipv6-address-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://www.irif.fr/~jch/software/files/
+PKG_HASH:=e1356d245d5f891fa39b796a8a2deefdaa89f08130dc97a378738ca5ed9a20fa
+PKG_INSTALL:=1
+
+PKG_MAINTAINER:=Nick Hainke <vincent@systemli.org>
```
neheb, line 12: no license?

PolynomialDivision: I searched now for the header text in a c file and I think it is MIT. I will ad this. but what is then the `PKG_LICENSE_FILES`?

PR #13151, opened 2020-08-17
mpg123: revert back to using autotools

neheb:
CMake build seems to be experimental and not ready for primetime.

Added altivec support.

Fixed installing static output libraries to the target.

Maintainer: @wigyori 
Compile tested: powerpc 464fp, aarch64


neheb:
ping @micmac1

PR #13152, opened 2020-08-17
glib2: update to 2.65.1

neheb:
Add a small hack to glib2 to avoid using proxy-intl and break
compilation under uClibc-ng.

Maintainer: @tripolar 
Compile tested: ath79 both musl and uClibc-ng


neheb:
Tested with all packages. No problems.

PR #13158, opened 2020-08-18
logrotate: remove selinux support

neheb:

Maintainer: @bk138 
Compile tested: ath79


bk138:
If OpenWrt wants to go without SELinux, I'm fine with this change.

PR #13166, opened 2020-08-18
boost: Package Revision - Fix C++ standard flag

ClaymorePT:
Maintainer: me / @ClaymorePT 
Compile tested: Raspberry Pi 4B

Description:
This commit fixes the use of the C++20 standard when GCC 10.x is used.


ClaymorePT:
@neheb Sorry for this late fix. The previous PR (https://github.com/openwrt/packages/pull/13160) where I updated the boost package, was not using the C++20 standard when GCC 10 was selected.

PR #13173, opened 2020-08-19
bandwidthd: don't build in parallel

neheb:
There are intermittent build failures on the buildbots because of this.
I see the same build failures locally as well.

Maintainer: @padre-lacroix 
Compile tested: ath79

https://downloads.openwrt.org/snapshots/faillogs/mips_mips32/packages/bandwidthd/


padre-lacroix:
If build in parallel causes problem, then obviously it has to be disabled!  If it is more reliable that way, sure!
I approve this commit.

PR #13183, opened 2020-08-21
ddns-scripts: use ip_source as bind_network default

mrjoel:
Run tested: OpenWRT snapshot with IPv6 connectivity

Description:
Fixes: #13182

### DIFF
--- a/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh
+++ b/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh
@@ -707,6 +707,9 @@ do_transfer() {
 
 	[ $# -ne 1 ] && write_log 12 "Error in 'do_transfer()' - wrong number of parameters"
 
+	# Use ip_network as default for bind_network if not separately specified
+	[ -z "$bind_network" ] && [ "$ip_source" = "network" ] && [ "$ip_network" ] && bind_network="$ip_network"
+
 	# lets prefer GNU Wget because it does all for us - IPv4/IPv6/HTTPS/PROXY/force IP version
 	if [ -n "$WGET_SSL" -a $USE_CURL -eq 0 ]; then 			# except global option use_curl is set to "1"
 		__PROG="$WGET_SSL --hsts-file=/tmp/.wget-hsts -nv -t 1 -O $DATFILE -o $ERRFILE"	# non_verbose no_retry outfile errfile
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_LICENSE:=GPL-2.0
 


mrjoel:
Sorry for the delay, rebased onto current master and added the PKG_RELEASE bump.

neheb:
ping @feckert

net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh
```diff
@@ -707,6 +707,9 @@ do_transfer() {
 
 	[ $# -ne 1 ] && write_log 12 "Error in 'do_transfer()' - wrong number of parameters"
 
+	# Use ip_network as default for bind_network if not separately specified
+	[ -z "$bind_network" -a "$ip_source" = "network" -a "$ip_network" ] && bind_network="$ip_network"
```
feckert, line 711: I am aware that in the whole script -a or -o is used.
But can you please change that to && this is more POSIX comform.
`[ -z "$bind_network" ] && [ "$ip_source" = "network" ] && [ "$ip_network" ] && bind_network="$ip_network"`
But I think this is wrong----------------------------------------------------------------^  ?

I also do not verify if this is correct, because I do not use this.
@Ansuel Can you say something about this

Ansuel: A better question would be...
@mrjoel what does this fix? Why this change?

Ansuel: Found the issue... It would be good to add to the commit description
"Fixes: #13182"

Ansuel: @feckert looks like this fix a specific problem...
bind_network is indeed a separate ddns config but is needed when ip_network is used...

So i think we can accept this when the `if` is fixed to follow POSIX and a "Fixed" tag is added to the commit description (Commit description is different than the pr description...)

mrjoel: Updated for better POSIX idioms and added issue link to commit message.

Ansuel:
Anyway i think you can remove "on https://github.com/openwrt/packages"... Github has some problem detecting the issue if there is text after the id

mrjoel:
> Anyway i think you can remove "on https://github.com/openwrt/packages"... Github has some problem detecting the issue if there is text after the id

Hmm, I added it to remind myself if nothing else the issue (with the `openwrt` and `packages` repos using different issue systems it is an easy reminder to myself where the id maps). I did check that github correctly located and linked the issue, but I can remove it if needed.

Ansuel:
Sorry for test... Try to update the pr description with 
`Fixes: #13182`

Ansuel:
Ok thx as you can see now this pr is linked with the issue... and will be auto closed on merge

dbro:
Hello- this change broke DDNS updates for my setup after upgrading from 19.07 to 21.02RC1.

I am using curl and https to update duckdns according to a dynamic IP address assigned via PPPoE. With the code introduced by this commit, the curl command times out. If I comment out the line added in this commit, it works fine for me as it did in 19.07. The difference shows up as the mandatory addition of the curl command line option "--interface ethX.X" which leads to timeouts - if that option is removed then the curl command works fine. In my case, I want to have an empty value for bind_network.

@mrjoel I thinkj you might be able to accomplish what this commit does by setting the value of bind_interface to "wan6" in your example config. Can you please test that and report back if it works for you?

Assuming that solves the issue, I would like to request reversion of this commit.

I should add that I might be a bit confused about the purpose of the bind_network variable. In the wiki page https://openwrt.org/docs/guide-user/base-system/ddns it says "Network to use for communication when detecting IP and sending updates. :!: Needs Wget or cURL package to be installed ! Wget will bind to the IP and cURL to physical interface of given network." The part about it being used "when detecting IP" does not seem to match what happens in the code because it is only ever used in the function do_transfer() and is never mentioned in the function get_local_ip() If bind_network does not affect the detection of the IP, then I recommend removing or rephrasing this information in the wiki. Also it is not included in the list of variables documented in the comments starting on line 120 in this file packages/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh - should it be added?

Dan

dbro:
@mrjoel Have you been able to test the alternate fix proposed above?

mrjoel:
@dbro I don't see any reference to a `bind_interface` config value, either on the wiki documentation or in the code. I assume you meant to reference `bind_network` instead? If so, then yes that will work, but as mentioned in the original issue my intent was to make that the default instead of having to explicitly set it.

I'm doing a further scouring and writeup of cases and handling for each that I'll post shortly. In the meantime, can you confirm if you're using IPv6 or only IPv4 with DuckDNS? I also happen to be using DuckDNS as one of my ddns services, and as they note on their site it's a bit of an oddball for IPv6 support (all requests have to go over IPv4 with an IPv6 URL parameter value).

dbro:
Hi @mrjoel yes, sorry I should have written "bind_network" instead of "bind_interface".

In my setup I am using only an IPv4 address with DuckDNS. Hope that helps.

PR #13191, opened 2020-08-22
fakepop: add new package

Foxtrot:
**Maintainer**: me
**Compile tested**: ramips/mt76x8, unielec_u7628, r11063-85e04e9f46
**Run tested**: ramips/mt76x8, unielec_u7628, r11063-85e04e9f46

**Description**:
fakepop is a fake pop3 daemon. It returns always the same messages to all users, it does not care about usernames and passwords. All user/pass combinations are accepted.

**Tests Done**:
Running xinetd on device and monitoring logread to show if the fakepop service started correctly, and using telnet to connect to the device on the default port (110) to send basic POP commands and get a response.

### DIFF
--- a/net/fakepop/Makefile
+++ b/net/fakepop/Makefile
@@ -1,0 +1,42 @@
+#
+# This software is licensed under the Public Domain.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=fakepop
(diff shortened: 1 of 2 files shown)


ZeroChaos-:
thanks @Foxtrot !

PR #13220, opened 2020-08-27
openfortivpn: use more standard config option names

aaronjg:
@lucize - I found a few more inconsistencies with how other tunneling packages name variables.  6in4, 6rd, dslite, gre, vxlan, ipip, xfrm, and vti all use 'tunlink' for the interface that the tunnel is established through and 'peeraddr' for the remote peer.

Breaking changes like this are not ideal, but the netifd version of openfortivpn has never been in a stable release, so I'd like to make breaking changes to the config now before the netifd version of openfortivpn has made it to a stable release.

Maintainer: @lucize


lucize:
sounds good to me!

PR #13221, opened 2020-08-27
openconnect: allow disable dtls with bool option no_dtls

yousong:
Maintainer: @nmav 
Compile tested: x86/64, openwrt master
Run tested: x86/64 on qemu, openwrt master

Tests done: Observe openconnect command line arguments and syslog lines under the following condition combinations

 - On/off the `no_dtls`
 - Enable/disable dtls udp traffic with iptables rule

### DIFF
--- a/net/openconnect/Makefile
+++ b/net/openconnect/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openconnect
 PKG_VERSION:=8.10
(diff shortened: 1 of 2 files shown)


yousong:
In case of ECONNREFUSED from read call, openconnect of openssl variant will emit error log `DTLS handshake failed: 5` where 5 is likely for `SSL_ERROR_SYSCALL`

nmav:
Thank you

PR #13246, opened 2020-08-30
simple-adblock: add config auto-update feature

stangri:
Maintainer: me
Compile tested: mvebu, WRT3200ACM, 19.07.3
Run tested: mvebu, WRT3200ACM, 19.07.3, start/config auto-update

Description: 
This update contains following:
* Adds an option (disabled by default) to perform user's config auto-update before downloading allow/block-lists. Config auto-update is a simple sed-control file to update/remove obsolete and outdated URLs to block-lists in the config file.
* Updated README.
* Added `version` parameter to the init-script to simplify corresponding luci app ACL.

Please merge, as I don't have merge access to packages.

I also request a merge access to packages as I will need to update most of my other apps with the `version` parameter for init-script.

### DIFF
--- a/net/simple-adblock/Makefile
+++ b/net/simple-adblock/Makefile
@@ -5,8 +5,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=simple-adblock
-PKG_VERSION:=1.8.3
-PKG_RELEASE:=15
+PKG_VERSION:=1.8.4
+PKG_RELEASE:=1
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.net>
 PKG_LICENSE:=GPL-3.0-or-later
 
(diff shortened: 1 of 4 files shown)


hnyman:
I am not sure that it is wise to override the default enable/disable commands (which are part of the standard rc.common functionality (setting/removing the autostart symlink in /etc/rc.d):
```
	enable	Enable service autostart
	disable	Disable service autostart
```

stangri:
I'd really like to use the `luci.sys.init.enable()` and `luci.sys.init.disable()` in luci. Is there a way to inherit and supplement default action instead of overwriting it?

stangri:
Looping in @jow- for comment. I'd really like to lessen dependency of luci app on shell commands and combination of cursor:set(), cursor:save() and cursor:commit() to change the value of enabled in config file didn't work. Also I believe sys.init.enable/disable is more elegant. Is there a way to inherit the rc.common functionality and add to it in the init-script?

PR #13252, opened 2020-08-30
lcd4linux: drop dependencies on ar71xx/rb532

adschm:
The ar71xx target has finally been removed in master, so drop the
dependencies in this package.

Maintainer: @feckert @neheb
Compile tested: no
Run tested: no

### DIFF
--- a/utils/lcd4linux/Config.in
+++ b/utils/lcd4linux/Config.in
@@ -274,12 +274,6 @@ config LCD4LINUX_CUSTOM_DRIVER_TeakLCM
 	bool
 	prompt "TeakLCM"
 
-config LCD4LINUX_CUSTOM_DRIVER_TEW673GRU
-	bool
-	select LCD4LINUX_CUSTOM_NEEDS_libgd
-	depends on TARGET_ar71xx
-	default TARGET_ar71xx
-
 config LCD4LINUX_CUSTOM_DRIVER_Trefon
 	bool
 	prompt "Trefon"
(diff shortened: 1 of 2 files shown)


neheb:
Why remove the patch?

adschm:
> Why remove the patch?

@neheb We don't "need" it anymore as it has no users in the code?

However, I'm completely agnostic about this; I just found the ar71xx dependency and dealt with it. I'm fine with every other solution as well; would keeping the patch allow it to be used for ath79 if the device gets ported at some point?

neheb:
I assume it would. Up to @feckert

adschm:
Okay, so he may either pick and update this himself or tell me what to change.

feckert:
The way I see it, this is the [router]( https://www.amazon.de/Trendnet-TEW-673GRU-Concurrent-Wireless-Gigabit/dp/B003C2SZQE)
I do not have them but I think it is apparently an ar71xx target.
If we support it in the new target ath79, then we should rename it so that it is only displayed in ath79. Otherwise we can delete the stuff. Since it is only for this router.

adschm:
If migrated, the device will probably be named trendnet_tew-673gru (trendnet,tew-673gru for board name) . According to our Wiki it has 8M flash and 64M RAM, so would generally be acceptable.

However, I'm not aware of any effort to port this particular device to ath79 at the moment. It's still on my "missing devices list" here: https://forum.openwrt.org/t/missing-ar71xx-devices-in-ath79-before-20-xx/70687

So, depending on the spirit of how support is handled here, one could either remove it now as it won't be supported in the near future, and add it back when ath79 device support is there (which will most probably be forgotten then).
Or you just keep the dead link to the ar71xx version as a reminder.

adschm:
@feckert I updated the commit to keep the patch. Since I'm about to remove the 4.14-only targets from main repo, I also added a second patch that removes rb532 here already.

feckert:
I like this solution. :+1: from my side

adschm:
Just pushed the removal to main repo, so this can be merged (I'd prefer to not merge it myself in packages repo).

Issue #12629, opened 2020-06-27
ctop: Buildbot errors


jefferyto:
Maintainer: @jmarcet 
Environment: multiple

Description:

Buildbots have been failing to build ctop:

* https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a53/packages/ctop/compile.txt
* https://downloads.openwrt.org/snapshots/faillogs/arm_cortex-a15_neon-vfpv4/packages/ctop/compile.txt
* https://downloads.openwrt.org/snapshots/faillogs/i386_pentium4/packages/ctop/compile.txt

Quoting from the error log for aarch64_cortex-a53:

```
install: cannot stat '/builder/shared-workdir/build/sdk/build_dir/target-aarch64_cortex-a53_musl/ctop-0.7.3/.go_work/build/bin/ctop': No such file or directory
Makefile:52: recipe for target '/builder/shared-workdir/build/sdk/build_dir/target-aarch64_cortex-a53_musl/ctop-0.7.3/.pkgdir/ctop.installed' failed
make[3]: *** [/builder/shared-workdir/build/sdk/build_dir/target-aarch64_cortex-a53_musl/ctop-0.7.3/.pkgdir/ctop.installed] Error 1
```

It's unfortunate that the CircleCI check uses ath79, because it is mips the package (with `@!mips`) was probably not actually compiled in #11159 (~I didn't check the CircleCI log~ I checked it). If the check was arm or i386 (or any platform other than x86_64) it would have failed.

jmarcet:
How should we got about this then, with `DEPENDS:=$(GO_ARCH_DEPENDS) @(aarch64||arm||x86_64)` like [docker-ce](https://github.com/openwrt/packages/blob/e2fe9bda3d6a2755f36031d8d69c6ea8e08ba715/utils/docker-ce/Makefile#L46) ?

I don't have time right now to try building for different architectures, I'm the middle of a relocation to another country.

jefferyto:
aarch64 and arm are exactly the architectures where build failures are occurring.

If you are busy and don't have the time for this, just say so. If you want to fix this please at least read the issue description clearly.

I don't want to be mean, but a basic difference between creating a package for yourself and creating/maintaining a package for others is testing on architectures used by others.

jmarcet:
From the logs I see all the components seem to be built, but the final executable is not found where it should be. Without testing locally I do not know what the actual error is.

[...]

PR #11159, opened 2020-01-28
ctop: top-like interface for container metrics

jmarcet:
Maintainer: Javier Marcet <javier@marcet.info>
Compile tested: x86-64
Run tested: x86-64

Description:

new package ctop: top-like interface for container metrics

PR #13254, opened 2020-08-31
CI: use GitHub action for Pull Requests

aparcar:
This builds packages for multiple architectures and stores the created packages and logs. It's trivial to build even more targets/architectures.

In a second step I'd implement simple runtime tests via qemu, emulating all the archs possible.

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -1,0 +1,59 @@
+name: Test Build
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: ${{ matrix.arch }} build
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - x86_64
+          - mips_24kc
+          - arm_cortex-a9_neon
+          - arm_cortex-a9_vfpv3-d16
+          - arc_arc700
+          - aarch64_generic
+          - powerpc_464fp
+          - powerpc_8540
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Determine changed packages
+        run: |
+          PACKAGES="$(git diff --diff-filter=d --name-only origin/master \
+            | grep 'Makefile$' | grep -Ev '/files/|/src/' \
+            | awk -F/ '{ print $(NF-1) }' | tr '\n' ' ')"
+
+          echo "Building $PACKAGES"
+
+          echo "::set-env name=PACKAGES::$PACKAGES"
+
+      - name: Build
+        uses: openwrt/gh-action-sdk@v1
+        env:
+          ARCH: ${{ matrix.arch }}
+          BUILD_LOG: 1
+          IGNORE_ERRORS: ""
+          V: s
+
+      - name: Store packages
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.arch}}-packages
+          path: bin/packages/${{ matrix.arch }}/packages/*.ipk
+
+      - name: Store logs
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.arch}}-logs
+          path: logs/
+


aparcar:
The prometheus patch is just to show the CI running

aparcar:
I'll leave this open for a week and merge it then, I don't see why we should have some more CI build tests. It's not blocking anyway and doesn't cost us anything.

.github/workflows/multi-arch-test-build.yml
```diff
@@ -0,0 +1,57 @@
...
+          - arm_cortex-a9_neon
+          - arm_cortex-a9_vfpv3-d16
+          - arc_arc700
+          - aarch64_generic
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 20
+      - run: git fetch --depth=20
```
champtar, line 26: I remember doing a lot of tests for CircleCI and no numbers are enough to handle old PR, you need full checkout to be safe

.github/workflows/multi-arch-test-build.yml
```diff
@@ -0,0 +1,57 @@
...
+          PACKAGES="$(git diff --diff-filter=d --name-only origin/master \
+            | grep 'Makefile$' | grep -Ev '/files/|/src/' \
+            | awk -F/ '{ print $(NF-1) }' | tr '\n' ' ')"
+
+          echo "Building $PACKAGES"
+
+          echo "::set-env name=PACKAGES::$PACKAGES"
+
+      - name: Build
+        uses: aparcar/action-openwrt-sdk@composite
```
champtar, line 39: We should move that to OpenWrt repo / org

.github/workflows/multi-arch-test-build.yml
aparcar: Done

.github/workflows/multi-arch-test-build.yml
aparcar: What name would you prefer? `action-sdk` is okay?

champtar: I'm bad at naming things, but maybe `gh-action-sdk` (action is a bit generic)

aparcar: Done

neheb:
No powerpc?

.github/workflows/multi-arch-test-build.yml
```diff
@@ -0,0 +1,56 @@
...
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: ${{ matrix.arch }} build
+    runs-on: ubuntu-latest
```
neheb, line 11: This maps to Ubuntu 18.04. Would it make sense to switch to 16.04? I believe 16.04 is still supported.

macos-10.15 is also an option. Actually that would be really interesting as I've seen several patches recently that fix compilation under OS X.

macos would be interesting should we decide to keep CircleCI, which I believe runs on Debian.

aparcar:
> No powerpc?

I'm happy to add that

.github/workflows/multi-arch-test-build.yml
aparcar: This is something else. This pull requests uses our SDKs running within a Debian 10. The Ubuntu is just the host OS starting the Docker container.

It is additionally possible to test OS X, however at this point we do not offer any SDKs for that. I don't think it is feasible to rebuild the entire toolchain for every pull requests, what do you think?

neheb: Ah never mind then.

aparcar:
In this particular case golang isn't available for a architecture and therefore not build.

jefferyto:
Some random thoughts (which may or may not be useful):

* It would be nice to see a build failure case.

* It would also be nice to have some way of highlighting any packages that were not built for a particular architecture (e.g. prometheus on powerpc). Currently it is indirectly indicated by missing artifacts. (If a PR contains multiple packages and some packages succeeded, then there would not be this missing artifacts warning.)

  (This is also an issue with the current CircleCI build, e.g. #12629 / https://github.com/openwrt/packages/pull/11159#issuecomment-650632217.)

* Re: building on bare macOS. I think it may be possible to build the full toolchain once then cache it for subsequent PRs. I agree that it would nice to have this to help debug macOS build errors.

* I've been thinking of setting up a repo template for users that want to host their own feed, using GitHub Actions to build then uploading/pushing the packages to a GitHub Pages site. I think it would help to reduce the number of one-off packages/maintainers in this feed.

  I haven't had time to pursue this idea yet; if you want to give it a try, please be my guest :joy:

aparcar:
>  It would be nice to see a build failure case.

I added a broken build, let's see if the CI test builds complain

> It would also be nice to have some way of highlighting any packages that were not built for a particular architecture (e.g. prometheus on powerpc). Currently it is indirectly indicated by missing artifacts. (If a PR contains multiple packages and some packages succeeded, then there would not be this missing artifacts warning.)

This is somewhat outside my scope, if packages are explicitly not available for a architecture they just won't be there. I think it's misleading to show an error message in that case. If really desired we could run an extra script afterwards that checks if the package is upstream available and fail if locally missing. But that's optional CI candy to me.

> Re: building on bare macOS. I think it may be possible to build the full toolchain once then cache it for subsequent PRs. I agree that it would nice to have this to help debug macOS build errors.

I recently talked to @adschm and discussed CI for `openwrt.git`, I'll extend our existing docker setup with pre-compiled containers for CI tests, this could also include OS X.

> I've been thinking of setting up a repo template for users that want to host their own feed, using GitHub Actions to build then uploading/pushing the packages to a GitHub Pages site. I think it would help to reduce the number of one-off packages/maintainers in this feed.

Please see the template below, it builds all packages and uploads them to a S3 storage.

```yaml
name: Build packages

on: [push]

jobs:
  build:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v2

      - name: Build
        uses: aparcar/action-openwrt-sdk@master

      - name: Upload
        uses: jakejarvis/s3-sync-action@master
        with:
          args: --acl public-read --follow-symlinks --delete
        env:
          AWS_S3_BUCKET: libremesh
          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
          AWS_S3_ENDPOINT: https://images.aparcar.org
          SOURCE_DIR: bin/packages/x86_64/action/
          DEST_DIR: packages/
```

aparcar:
Okay, I tested the PR with a intentionally broken commit and the PR failed, tada. I'll merge this now and extend it step by step.

jefferyto:
> This is somewhat outside my scope, if packages are explicitly not available for a architecture they just won't be there. I think it's misleading to show an error message in that case. If really desired we could run an extra script afterwards that checks if the package is upstream available and fail if locally missing. But that's optional CI candy to me.

I understand if you feel it is out of your scope. I think it would also be misleading to show a test success in that case; I was hoping for some neutral way to convey that the test/arch was skipped.

> Please see the template below, it builds all packages and uploads them to a S3 storage.

Thanks for the template - this isn't exactly what I was looking for (to use this one would need to pay for S3 :joy:) but perhaps you can publish it somewhere so users who want to use it can find it easily.

aparcar:
> I understand if you feel it is out of your scope. I think it would also be misleading to show a test success in that case; I was hoping for some neutral way to convey that the test/arch was skipped.

There is a yellow annotation saying no artefacts where created, is that enough?

> Thanks for the template - this isn't exactly what I was looking for (to use this one would need to pay for S3 joy) but perhaps you can publish it somewhere so users who want to use it can find it easily.

Sorry here is a action step deploying it as a GH page.

```yaml
    - name: Deploy to GitHub Pages
      uses: peaceiris/actions-gh-pages@v3
      with:
        github_token: ${{ secrets.GITHUB_TOKEN }}
        publish_dir: ./output/
        allow_empty_commit: true
```

jefferyto:
> There is a yellow annotation saying no artefacts where created, is that enough?

If the PR contains multiple new/updated packages and only some are tested and others skipped, then there will still be artifacts created.

PR #13258, opened 2020-08-31
python-cryptography: fix compilation without deprecated OpenSSL APIs

neheb:

Maintainer: @jefferyto @commodo 
Compile tested: ath79

https://github.com/pyca/cryptography/pull/4920


jefferyto:
I'm not strictly against this, but this patchset doesn't look trivial... I would feel better if it was accepted in upstream first.

Can we wait until upstream [drops 1.0.2 support](https://github.com/pyca/cryptography/pull/4920#issuecomment-669489471) or do we have a strong reason to do this now?

neheb:
This is a patch I’ve kept locally for a long time. Throwing most of them on here to reduce the patches I carry.

neheb:
Looks like this is needed for this and next release. Going to merge.

PR #13279, opened 2020-09-02
xtables-addons: drop support for kernel 4.14

adschm:
Support for kernel 4.14 has been removed in main repo, so also remove
support here.

Maintainer: @dengqf6 @jow-
Compile tested: no
Run tested: no

Description:
I did _not_ check whether the subsequent patches need refresh.


neheb:
What's the point? Is it breaking 5.4 builds? This patch can be dropped on version update.

PR #13280, opened 2020-09-02
openvswitch: remove support for kernel 4.14 and 4.9

adschm:
Support for kernel 4.14 has been removed in main repo, so drop the
dependencies here as well (and those for even older 4.9).

Also drop a patch that is required only for 4.14 and lower.

Maintainer: @yousong


yousong:
Thank you!

PR #13281, opened 2020-09-02
libreswan: drop support for kernel 4.14

adschm:
Support for kernel 4.14 has been dropped in main repo, so remove it
here as well.

Maintainer: @lucize 
Compile tested: none
Run tested: none


lucize:
LGTM

adschm:
I prefer not to merge myself on packages, so just proceed if you think it good. :-)

PR #13297, opened 2020-09-03
gerbera: update to 1.6.1

neheb:
Removed upstream ffmpeg backport.

Updated npupnp patch with latest upstream version.

Enabled CURL since it's needed for libnpupnp anyway.

Fixed up GCC filesystem CMake variables to be more accurate.

Removed npupnp include hack since it was properly fixed now.

Maintainer: nobody
Compile tested: ath79


neheb:
```
Package more is missing dependencies for the following libraries:
libmagic.so.1
```

Depends on https://patchwork.ozlabs.org/project/openwrt/patch/20200901223703.785743-1-rosenp@gmail.com/

PR #13313, opened 2020-09-05
CI: remove leftover travis files

aparcar:
The current CI uses both CircleCI and GitHub Action CI, but not Travis.


neheb:
@BKPepe might have a problem with this.

aparcar:
If it's still in use please close this PR, but Travis does not appear under the PRs as test

BKPepe:
Even though check for Travis was replaced by Circle CI, it does not mean that somebody like me is using it. It does not harm that it is in repo, right? I am still not convinced by your reasoning. I am using it just in case that it does not proceed with Circle CI even my pull requests are compiled and run tested. Some packages fail with Travis, but still at least I feel much confident if package is compiled against three targets. Travis is free for open-source projects.

For example: https://travis-ci.org/github/BKPepe/packages/builds/711078742

Still, your Github Actions needs to be improved according to https://github.com/openwrt/packages/pull/13254#issuecomment-687509637

BKPepe:
And there are some issues with Github Actions as well for example in this PR: https://github.com/openwrt/packages/pull/13320

BKPepe (approved):
Github Actions is these days in really good shape, so we can remove Travis. :)

aparcar:
Please proceed

BKPepe:
Well, you seems that you done it, but it requires to rebase your branch, which would take less time than creating a new commit, creating new PR and such.

PR #13315, opened 2020-09-06
[RFC]CI: add runtime test support

aparcar:
This is a bit of a long shot but the idea is that binaries are tested inside a OpenWrt container with the specific architecture.

Currently we offer Docker contains in x86/64 and armvirt/{32,64}, however we could easily add more architectures.

The idea of `test.sh` is that it's a shell script that tests if the compiled tools has some minimal functionality.

The attached example runs `prometheus -version` and checks if it exists cleanly. These tests could be more sophisticated. There is obviously no hardware emulation and much more is missing for authentic tests, but ideally it simplifies the lives of maintainers.

It's not fully tested yet an more of a concept.

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -1,0 +1,6 @@
+ARG ARCH=x86-64
+FROM openwrtorg/rootfs:$ARCH
+
+ADD entrypoint.sh /entrypoint.sh
+
+CMD ["/entrypoint.sh"]
--- a/.github/workflows/entrypoint.sh
+++ b/.github/workflows/entrypoint.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+mkdir -p /var/lock/
+
+opkg update
+
+for PKG in /ci/*.ipk; do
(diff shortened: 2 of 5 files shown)


aparcar:
This PR fails because Go is not available on some architectures, however the new *runtime* tests work. The binaries for both x86/64 and armvirt/64 run, produce an output and exit with 0. The current Prometheus `test.sh` script is trivial, however it can be as complex as desired for each package.

aparcar:
@jefferyto should Go packages fail for not supported architectures or silently skip?

jefferyto:
This is labelled as "RFC" but I don't see any comments before this was merged.

Also, is there any documentation on how to add a test.sh or what is should do?

aparcar:
> This is labelled as "RFC" but I don't see any comments before this was merged.

I talked a bit to @neheb via IRC who liked the idea.

> Also, is there any documentation on how to add a test.sh or what is should do?

Currently only in the commit message, I wasn't sure where to add this. Maybe on the bottom of the README.md?

PR #13317, opened 2020-09-06
prometheus: set build info via x-flags

aparcar:
Similar to `syncthing` also `prometheus` build environment sets
variables later shown in the `--version` string. Set those varialbes.

### DIFF
--- a/utils/prometheus/Makefile
+++ b/utils/prometheus/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=prometheus
 PKG_VERSION:=2.20.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


aparcar:
@jefferyto This doesn't quite work yet, could you please have a look?

jefferyto:
Can you include a bit more information on how it is not working?

aparcar:
The variables in `info.go` are not set and therefore `prometheus --version` stays empty. I tried to do the same as for syncthing but it doesn't work.

jefferyto:
I think `github.com/prometheus/common/version/info.XXX` should be changed to `github.com/prometheus/common/version.XXX` ([source](https://github.com/prometheus/prometheus/blob/v2.20.1/.promu.yml#L16-L21)).

aparcar:
@jefferyto that looks very promising, updated!

aparcar:
Tested and works, amazing!

PR #13328, opened 2020-09-07
ninja: use for CMake

neheb:
CMake supports Ninja for faster compilation and less bugginess when it
comes to parallel compilation. That is, some CMake packages currently
have PKG_BUILD_PARALLEL set where it is not needed with ninja.

Maintainer: @dhewg 
Compile tested: ath79

I'm unsure how to properly implement this. I initially tried to get this in base by modifying cmake.mk but was rejected.

The idea currently is to include ninja.mk after cmake.mk.

### DIFF
--- a/devel/ninja/ninja-cmake.mk
+++ b/devel/ninja/ninja-cmake.mk
@@ -1,0 +1,25 @@
+include $(INCLUDE_DIR)/cmake.mk
+include ../../devel/ninja/ninja.mk
+
+CMAKE_HOST_OPTIONS += -DCMAKE_GENERATOR="Ninja"
+CMAKE_OPTIONS += -DCMAKE_GENERATOR="Ninja"
+
+define Host/Compile/Default
+	$(call Ninja,-C $(HOST_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR),)
+endef
+
+define Host/Install/Default
+	$(call Ninja,-C $(HOST_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR) install,DESTDIR="$(HOST_INSTALL_DIR)")
+endef
+
+define Host/Uninstall/Default
+	-$(call Ninja,-C $(HOST_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR) uninstall,)
+endef
+
+define Build/Compile/Default
+	$(call Ninja,-C $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR),)
+endef
+
+define Build/Install/Default
+	$(call Ninja,-C $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR) install,DESTDIR="$(PKG_INSTALL_DIR)")
+endef
--- a/devel/ninja/ninja.mk
+++ b/devel/ninja/ninja.mk
@@ -6,6 +6,9 @@
 #   $(call Ninja,-C $(MY_NINJA_BUILD_DIR),$(MY_NINJA_ENV_VARS))
 # endef
 
+HOST_BUILD_DEPENDS += ninka/host
+PKG_BUILD_DEPENDS += ninja/host
+
 NINJA_ARGS:=$(filter -j%,$(filter-out -j,$(MAKEFLAGS)))
 ifneq ($(findstring c,$(OPENWRT_VERBOSE)),)
   NINJA_ARGS+=-v
--- a/devel/ninja/Makefile
+++ b/devel/ninja/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ninja
 PKG_VERSION:=1.10.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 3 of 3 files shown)


diizzyy:
Hardsetting it is probably a bad idea, see: https://svnweb.freebsd.org/ports/head/Mk/Uses/cmake.mk?revision=488341&view=markup#l7 for instance

neheb:
Well, as is this is opt-in. ninja.mk and cmake.mk must be included.

dhewg:
How about creating `devel/ninja/cmake+ninja.mk`, including `ninja.mk` there and then defining your build commands?
As-is, it looks like it could interfere with non cmake ninja projects?

neheb:
@dhewg done.

devel/ninja/ninja-cmake.mk
```diff
@@ -0,0 +1,13 @@
+include $(INCLUDE_DIR)/cmake.mk
```
dhewg, line 1: If you include `ninja.mk` here too, projects can simply switch from `cmake.mk` to `cmake-ninja.mk` without including both

devel/ninja/ninja-cmake.mk
```diff
@@ -0,0 +1,13 @@
...
+CMAKE_OPTIONS += -DCMAKE_GENERATOR="Ninja"
+
+define Build/Compile/Default
+	mkdir -p $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR)
+	cd $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR) && $(STAGING_DIR_HOSTPKG)/bin/ninja
+endef
+
+define Build/Install/Default
+	DESTDIR=$(PKG_BUILD_DIR)/ipkg-install \
+	$(STAGING_DIR_HOSTPKG)/bin/ninja install -C $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR)
```
dhewg, line 12: `$(call Ninja,-C $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR),DESTDIR=$(PKG_BUILD_DIR)/ipkg-install)` ?

devel/ninja/ninja-cmake.mk
```diff
@@ -0,0 +1,13 @@
+include $(INCLUDE_DIR)/cmake.mk
+
+CMAKE_OPTIONS += -DCMAKE_GENERATOR="Ninja"
+
+define Build/Compile/Default
+	mkdir -p $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR)
+	cd $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR) && $(STAGING_DIR_HOSTPKG)/bin/ninja
```
dhewg, line 7: `$(call Ninja,-C $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR),)` ?

dhewg (commented):
Small comments, otherwise looks okay to me, but I can't really comment on the cmake interaction itself

devel/ninja/ninja-cmake.mk
neheb: Done.

devel/ninja/ninja-cmake.mk
neheb: Done.

devel/ninja/ninja-cmake.mk
neheb: I can't seem to get this working. I'm getting a too many open files error.

neheb: Never mind.

devel/ninja/ninja-cmake.mk
neheb: This doesn't seem to work. Specifically DESTDIR doesn't seem to get set to ipkg-install.

neheb: Fixed. Gave up and copied meson.mk.

dhewg:
Look neat now, thanks.
What's the plan now? Selectively moving packages over to using this? Bulk move?

neheb:
Selective. This still does not handle Host packages. I've also seen this fail with certain packages when I modified CMake.mk directly.

neheb:
@dhewg added host package support.

Btw I noticed a typo is meson.mk : https://github.com/openwrt/packages/blob/master/devel/meson/meson.mk#L110

Is the beginning - intentional?

dhewg:
The `-` prefix? Yes, that's intentional, that's for `make` to ignore errors in case that command fails (`uninstall` may fail because the build dir was been wiped, it's not properly implemented by the project, etc)

neheb:
OK. Made the change here as well.

neheb:
Added BUILD_DEPENDS for ninja/host. Might be better in ninja.mk?

dhewg:
It's been too long, but I think I tried that and something broke... Sorry, don't recall the details. Give it a try?

neheb:
Small test shows it as working.

devel/ninja/ninja.mk
```diff
@@ -6,6 +6,9 @@
 #   $(call Ninja,-C $(MY_NINJA_BUILD_DIR),$(MY_NINJA_ENV_VARS))
 # endef
 
+HOST_BUILD_DEPENDS += ninka/host
```
dhewg, line 9: typo ninka -> ninja

PR #13335, opened 2020-09-08
nextdns: Update to version 1.8.2

rs:
Update to version 1.8.2


neheb:
@aparcar CI doesn't seem to respect DEPENDS.

aparcar:
> @aparcar CI doesn't seem to respect DEPENDS.

What do you mean by that? It uses `feeds install <pkg>` which should handle DEPENDS

neheb:
arc and powerpc are failing since they are not part of $(GO_ARCH_DEPENDS)

PR #13340, opened 2020-09-08
safe-search: duckduckgo.com new ip address

farmergreg:
Maintainer: me 
Compile tested: x86
Run tested: x86 / 19.07.3

Description:
Update to 40.89.244.237 which is the new IP address that duckduckgo.com is using for safe-search.

### DIFF
--- a/net/safe-search/Makefile
+++ b/net/safe-search/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=safe-search
 PKG_VERSION:=1.0.2
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_LICENSE:=MIT
 PKG_MAINTAINER:=Gregory L. Dietsche <Gregory.Dietsche@cuw.edu>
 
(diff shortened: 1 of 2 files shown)


farmergreg:
Is there a way to get this to mainline more quickly than usual? Without this fix, duckduckgo is broken for users of this package. Thanks!

dibdot:
Could you link to official duckduckgo source for confirmation? I doubt that this is a valid fix, cause in Germany I'll get this IP in return ...
```
root@2go_ar750s:~# nslookup safe.duckduckgo.com
Server:		127.0.0.1
Address:	127.0.0.1#53

Name:      safe.duckduckgo.com
Address 1: 40.114.177.246
```
Most probably you'll need more matching ip addresses to cover this domain/cname globally.

aparcar:
You have to bump the PKG_RELEASE

farmergreg:
> root@2go_ar750s:~# nslookup safe.duckduckgo.com
> Server:		127.0.0.1
> Address:	127.0.0.1#53
> 
> Name:      safe.duckduckgo.com
> Address 1: 40.114.177.246
> ```
> 
> Most probably you'll need more matching ip addresses to cover this domain/cname globally.

Thanks for pointing that out. I think I should write some code that can automatically generate the hosts file based on where safe.duckduckgo.com points. cnames are hard to work with when using dnsmasq because of some dnsmasq limitations.

Issue #13211, opened 2020-08-25
mwan3: Should check_quality be enabled on the default WAN interface configurations?


jamesmacwhite:
Maintainer: @feckert 
Environment: N/A

Description:

The default mwan3 config has the four `check_quality` configuration options set on the `wan` and `wanb` interfaces.

https://github.com/openwrt/packages/blob/master/net/mwan3/files/etc/config/mwan3#L15-L18
https://github.com/openwrt/packages/blob/master/net/mwan3/files/etc/config/mwan3#L47-L50

However because `check_quality` is not set this would default to be off and therefore having these configuration options set in the config is redundant currently, as they aren't doing anything, as they'd never be used here, with the boolean flag being off.

https://github.com/openwrt/packages/blob/master/net/mwan3/files/usr/sbin/mwan3track#L220-L254

Should we set `option check_quality '1'` in the default configuration for these interfaces or remove them entirely?

A minor change but just opening any discussion on it being a default configuration.

aaronjg:
Makes sense to me to turn them off. While we are at it, why not remove everything after `track_ip` and let the defaults work. 
https://github.com/openwrt/packages/blob/master/net/mwan3/files/usr/sbin/mwan3track#L137-L152

The default values in luci should also be updated so that they match the defaults in mwan3track:
https://github.com/openwrt/luci/blob/master/applications/luci-app-mwan3/luasrc/model/cbi/mwan/interfaceconfig.lua

jamesmacwhite:
@aaronjg Funny you mention default values. I found the same thing:

https://github.com/openwrt/luci/issues/4389

Several configuration options for the interface configuration were missing in the OpenWrt docs, so I opened an issue on that luci-app-mwan3.

feckert:
I have fixed the default values in [luci-app-mwan3](https://github.com/openwrt/luci/commit/8e4edd20352d0d458c988e17a3dd6b0626c5ba29) for **failure_loss** and **recovery_loss**.

In the OpenWrt there is no clear line how we should proceed here.
Either we use only the minimal setting, or we list all options as comments on the options we have, because not everyone uses LuCI.
And if an option is set/changed via LuCI then all comments with a # are deleted in mwan3.

jamesmacwhite:
[...]

PR #13343, opened 2020-09-09
mwan3: Remove check_quality options from default config

jamesmacwhite:

Maintainer: @feckert 
Compile tested: N/A
Run tested: N/A

Description:

As per https://github.com/openwrt/packages/issues/13211. This removes the check_quality options from the default configuration shipping with mwan3. Because the default configuration doesn't have `check_quality` enabled, they aren't doing anything and can safely fallback to the defaults unless enabled by the user.


feckert:
I would like to have first https://github.com/openwrt/packages/pull/13169 merged. After that we could bump the PKG_RELEASE number and merge this.

jamesmacwhite:
@feckert No worries. I'll leave it open for now

jamesmacwhite:
@feckert Wondering if it's also worth removing some of the other options, keeping the default interface definitions basic, I don't know if there is any benefit from setting these outside of their default for most configurations?

* `reliability` - 2 (Default: 1)
* `count` - 1 (Default: 1)
* `timeout` - 2 (Default: 4)
* `interval` - 5 (Default: 10)
* `down` - 3 (Default: 5)
* `up` -  (Default: 5)

It seems `count` is redundant as it's the same, `reliability` is different for the the `wan` example, which I guess you could argue as it's often the primary interface, the rest do different but should we cut down the default options?

We could have a comment at the top of the example configuration file with?

```
# Full documentation of options available: 
# https://openwrt.org/docs/guide-user/network/wan/multiwan/mwan3#mwan3_configuration

jamesmacwhite:
I noticed there are some additional options that are mismatched between LuCI and mwan3track, so I opened a PR for that here: https://github.com/openwrt/luci/pull/4429

feckert:
This is already in [P/R](https://github.com/openwrt/packages/pull/13169/commits/6083251740bf89d7c6c53b06d7e3f6397743c7ed) staged so we could close them.

PR #13345, opened 2020-09-09
[WIP] acme: use micrond as default cron

feckert:
Maintainer: @tohojo
Compile tested: not neede only script changes
Run tested: x86_64, APU3, check if micrond file was added 

Description:
The advantage of micrond is that it does not pollute the crontab of the
user root. This cron job was not created by the user. It was created by
acme. This is a system cron job and therefore it should not appear in
the crontab of the user root.

### DIFF
--- a/net/acme/Makefile
+++ b/net/acme/Makefile
@@ -25,7 +25,7 @@ include $(INCLUDE_DIR)/package.mk
 define Package/acme
   SECTION:=net
   CATEGORY:=Network
-  DEPENDS:=+wget-ssl +ca-bundle +openssl-util +socat
+  DEPENDS:=+wget-ssl +ca-bundle +openssl-util +socat +micrond
   TITLE:=ACME (Letsencrypt) client
   URL:=https://acme.sh
   PKGARCH:=all
@@ -59,7 +59,12 @@ endef
 
 define Package/acme/prerm
 #!/bin/sh
-sed -i '/\/etc\/init\.d\/acme start/d' /etc/crontabs/root
+if [ -z "$${IPKG_INSTROOT}" ]; then
+	[ -f "/usr/lib/micron.d/acme" ] && {
+		rm -rf "/usr/lib/micron.d/acme
+		/etc/init.d/micrond restart
+	}
+fi
 endef
 
 define Package/acme-dnsapi
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -21,14 +21,15 @@ NGINX_WEBSERVER=0
 UPDATE_NGINX=0
 UPDATE_UHTTPD=0
 USER_CLEANUP=
+MICROND_FILE="/usr/lib/micron.d/acme"
 
 . /lib/functions.sh
 
 check_cron()
 {
-	[ -f "/etc/crontabs/root" ] && grep -q '/etc/init.d/acme' /etc/crontabs/root && return
-	echo "0 0 * * * /etc/init.d/acme start" >> /etc/crontabs/root
-	/etc/init.d/cron start
+	[ -f "$MICROND_FILE" ] && return
+	echo "0 0 * * * /etc/init.d/acme start" > "$MICROND_FILE"
+	/etc/init.d/micrond restart
 }
 
 log()


tohojo:
Hmm, I see your point, but I'm not sure it's a good idea to add a hard dependency on a particular cron daemon to acme. So how about we just look for the micrond directory and use that if available, and otherwise fall back to the regular crontab file?

feckert:
> Hmm, I see your point, but I'm not sure it's a good idea to add a hard dependency on a particular cron daemon to acme.

Now we just rely on using the crontab from the busybox. What other crons are there in OpenWrt besides [micrond](https://github.com/openwrt/packages/tree/master/utils/micrond)?

I think this is a basic discussion, if it wouldn't be better not to show the user the cron jobs the system creates. I think it would be better if the system jobs do not appear in the crontab of the user. Because when I look at the file and there is a job in it that I don't know I find it strange. Also there are probably users who delete this kind of entries because they do not create them.

> So how about we just look for the micrond directory and use that if available, and otherwise fall back to the regular crontab file?

That would of course be a solution. I don't like it, but at least it would be a compromise.
In OpenWrt there are other packages that write their tasks to the root busybox crontab e.g. [wifischedule](https://github.com/openwrt/packages/tree/master/net/wifischedule) or even worse have a service for tasks that a crontab could actually do [ddns-scripts](https://github.com/openwrt/packages/tree/master/net/ddns-scripts). I am looking at the ddns-scripts right now and want to refactor it because it is not very well maintainable. And there I would like to use the micrond as well.
That are only two packages that could also use the micrond. There are certainly more

tohojo:
Florian Eckert <notifications@github.com> writes:

>> Hmm, I see your point, but I'm not sure it's a good idea to add a hard dependency on a particular cron daemon to acme.
>
> Now we just rely on using the crontab from the busybox. What other
> crons are there in OpenWrt besides
> [micrond](https://github.com/openwrt/packages/tree/master/utils/micrond)?

Well, my Turris Omnia is running vixie-cron, it would appear :)

> I think this is a basic discussion, if it wouldn't be better not to
> show the user the cron jobs the system creates. I think it would be
> better if the system jobs do not appear in the crontab of the user.
> Because when I look at the file and there is a job in it that I don't
> know I find it strange. Also there are probably users who delete this
> kind of entries because they do not create them.

Most openwrt systems are single-purpose devices, in which case I think
your distinction between 'user cron' and 'system cron' is somewhat
theoretical. But wouldn't object to adding a comment when we install
things into the crontab so anyone who does look will know what it's
there for...

>> So how about we just look for the micrond directory and use that if available, and otherwise fall back to the regular crontab file?
>
> That would of course be a solution. I don't like it, but at least it would be a compromise.
> In OpenWrt there are other packages that write their tasks to the root busybox crontab e.g. [wifischedule](https://github.com/openwrt/packages/tree/master/net/wifischedule) or even worse have a service for tasks that a crontab could actually do [ddns-scripts](https://github.com/openwrt/packages/tree/master/net/ddns-scripts). I am looking at the ddns-scripts right now and want to refactor it because it is not very well maintainable. And there I would like to use the micrond as well.
> That are only two packages that could also use the micrond. There are certainly more

Sure, plenty of packages that *could* use micrond. And as I said I'm
fine with supporting it, I just don't think a package should make that
choice for the user.

Actually, I also think it's a bit weird that micrond has its own notion
of where the system cron files should reside. Any other system users
/etc/cron.d so why not just use that? Any package could check if the
directory exists and drop files in there if it does, otherwise write to
/etc/crontabs/root. No need to care about which cron daemon is installed
at all...

feckert:
> Well, my Turris Omnia is running vixie-cron, it would appear :)

This may be true but in the official packages of openwrt there is only the crontab for the user root from the busybox and the mircond. Otherwise I did not find a cron service in the sources. Thus the argument does not count.

> Most openwrt systems are single-purpose devices, in which case I think your distinction between 'user cron' and 'system cron' is somewhat theoretical. But wouldn't object to adding a comment when we install things into the crontab so anyone who does look will know what it's there for...

I agree with you. But not everyone knows so well that they know exactly what does each line in busybox crontabs root stands for and what it is for. I think it's better if the system services that want to have a cron jobs don't write this in the busybox crontabs of root. They should place there cron jobs into another place for example micrond. In debian it is the same way the user does not see from any jobs the system creates.

> Sure, plenty of packages that *could* use micrond. And as I said I'm fine with supporting it, I just don't think a package should make that choice for the user. Actually, I also think it's a bit weird that micrond has its own notion of where the system cron files should reside. Any other system users /etc/cron.d so why not just use that? Any package could check if the directory exists and drop files in there if it does, otherwise write to /etc/crontabs/root. No need to care about which cron daemon is installed at all...

That with the mircon.d include file that can be changed. We only need to change this [line](https://github.com/openwrt/packages/blob/65f63a0e9da889325c9497a06a9dd9d8ffcb6a7f/utils/micrond/files/etc/init.d/micrond#L6).

The suggestion that the service should decide where to place its cronjobs depending on which cron service is installed is not very good, because it would complicate the scripts. Maybe there is a way to work with package variable [VARIANT](https://github.com/openwrt/packages/search?q=VARIANT&type=)?.

tohojo:
Florian Eckert <notifications@github.com> writes:

> I agree with you. But not everyone knows so well that they know
> exactly what does each line in busybox crontabs root stands for and
> what it is for. I think it's better if the system services that want
> to have a cron jobs don't write this in the busybox crontabs of root.
> They should place there cron jobs into another place for example
> micrond. In debian it is the same way the user does not see from any
> jobs the system creates.

As I said, I'm fine with creating a file in /etc/cron.d/ if it exists,
and falling back to /etc/crontabs/root if it doesn't...

>> Sure, plenty of packages that *could* use micrond. And as I said I'm fine with supporting it, I just don't think a package should make that choice for the user. Actually, I also think it's a bit weird that micrond has its own notion of where the system cron files should reside. Any other system users /etc/cron.d so why not just use that? Any package could check if the directory exists and drop files in there if it does, otherwise write to /etc/crontabs/root. No need to care about which cron daemon is installed at all...
>
> That with the mircon.d include file that can be changed. We only need
> to change this
> [line](https://github.com/openwrt/packages/blob/65f63a0e9da889325c9497a06a9dd9d8ffcb6a7f/utils/micrond/files/etc/init.d/micrond#L6).
>
> The suggestion that the service should decide where to place its
> cronjobs depending on which cron service is installed is not very
> good, because it would complicate the scripts. Maybe there is a way to
> work with package variable
> [VARIANT](https://github.com/openwrt/packages/search?q=VARIANT&type=)?.

I'd much rather have this as a one-line conditional in the script, than
having to deal with variants of the package...

feckert:
@tohojo lets pull in @NeoRaider who is the author of [micrond](https://github.com/openwrt/packages/tree/master/utils/micrond).
@NeoRaider as @tohojo suggest we should change the cron directory for [micrond](https://github.com/openwrt/packages/blob/master/utils/micrond/files/etc/init.d/micrond#L6) to /etc/cron.d, so we have a central cron directory for cronjobs.
If another user wants to use another cron daemon and not mircond.
The advantage of this is that we could place the cron files into this dir and we do not have to add a dependency to an explicit cron daemon.
But for this to work we have to change the mircon.d cron directory.
@NeoRaider What is you opinion on that?

jow-:
I'd rather adjust busybox crond to handle something like /etc/cron.d/. Having two different cron daemons just to avoid a line in /etc/crontabs/root seems overkill and waste of resources to me.

tohojo:
Jo-Philipp Wich <notifications@github.com> writes:

> I'd rather adjust busybox crond to handle something like /etc/cron.d/.
> Having two different cron daemons just to avoid a line in
> /etc/crontabs/root seems overkill and waste of resources to me.

+1 - but as far as a consumer of cron like acme is concerned, it
shouldn't matter which daemon it is, there should just be an
/etc/cron.d/ to drop things into...

feckert:
>> I'd rather adjust busybox crond to handle something like /etc/cron.d/. Having two different cron daemons just to avoid a line in /etc/crontabs/root seems overkill and waste of resources to me.

> +1 - but as far as a consumer of cron like acme is concerned, it shouldn't matter which daemon it is, there should just be an /etc/cron.d/ to drop things into...

That would of course be the best solution. But as I see it, the busybox cron can currently only parse and use one file. This file **must** be called root.
I have already tried everything before to work with other files not named root, because I did also not lick it to have another cron in the system. 
But this does not worked.

I got the following output if a place a file with the name test into /tmp/spool/cron
`Tue Sep 29 13:26:00 2020 cron.info crond[6757]: ignoring file 'test' (no such user)`

I think this is because of the following [line](https://github.com/mirror/busybox/blob/b6946084296525f0e2aa3e72df7b896cc98197d7/miscutils/crond.c#L427) in the busybox source.

SvenRoederer:
based on #15667 I usually use micrond for system default jobs and the crontabs for user configured jobs. This way my packages can install the "sytem jobs" in  /usr/... (which will also not backed up  during sysupgrade)  and only the user-generated jobs are preserved in /etc/crontabs/.
So having both ways would be really great.
Just asking: "why not using the cron that debian is using?" it's supporting both ways.

feckert:
> based on #15667 I usually use micrond for system default jobs and the crontabs for user configured jobs. This way my packages can install the "sytem jobs" in /usr/... (which will also not backed up during sysupgrade) and only the user-generated jobs are preserved in /etc/crontabs/.

I do the same one my openWrt system. And this is much cleaner, to put the system task into its own cron task file. But this is for now not supported by the crontab of the busybox. So I also use an additional cron service for this. But not every one wants to install an additional crontab service. So we have to options.

1. Extend crontab to also source the files under /etc/cron.d for system tasks
2. Use another cron service to use the files under /etc/cron.d (micrond or vixie-cron)

The preferred solution would be the first one, so we do not need a second service. If this is done all packages should drop the cron task into a file of the /etc/cron.d directory and not in the crontab of the root user directly anymore.

Since we can't force anyone to install an additional cron in the meantime (these additional packages are only in the package feed and not in the openwrt) we could update the packages to be made intelligent enough to detect if there is an additional cron service. But this is difficult!

So I think we should extend crontab from the busybox and change micrond to use the files under /etc/cron.d. And all package should drop the  cron system jobs file into /etc/cron.d. Then every one is happy :-) only then i don't need an additional cron service but i can replace the one from the busybox because maybe the other one has more features.

feckert:
> based on #15667 I usually use micrond for system default jobs and the crontabs for user configured jobs. This way my packages can install the "sytem jobs" in /usr/... (which will also not backed up during sysupgrade) and only the user-generated jobs are preserved in /etc/crontabs/.

I do the same one my openWrt system. And this is much cleaner, to put the system task into its own cron task file. But this is for now not supported by the crontab of the busybox. So I also use an additional cron service for this. But not every one wants to install an additional crontab service. So we have to options.

1. Extend crontab to also source the files under /etc/cron.d for system tasks
2. Use another cron service to use the files under /etc/cron.d (micrond or vixie-cron)

The preferred solution would be the first one, so we do not need a second service. If this is done all packages should drop the cron task into a file of the /etc/cron.d directory and not in the crontab of the root user directly anymore.

Since we can't force anyone to install an additional cron in the meantime (these additional packages are only in the package feed and not in the openwrt) we could update the packages to be made intelligent enough to detect if there is an additional cron service. But this is difficult!

So I think we should extend crontab from the busybox and change micrond to use the files under /etc/cron.d. And all package should drop the  cron system jobs file into /etc/cron.d. Then every one is happy :-) only then i don't need an additional cron service but i can replace the one from the busybox because maybe the other one has more features.

SvenRoederer:
> I got the following output if a place a file with the name test into /tmp/spool/cron
> `Tue Sep 29 13:26:00 2020 cron.info crond[6757]: ignoring file 'test' (no such user)`
> 
> I think this is because of the following [line](https://github.com/mirror/busybox/blob/b6946084296525f0e2aa3e72df7b896cc98197d7/miscutils/crond.c#L427) in the busybox source.

isn't this the consequence of the sanity check? Usually the name of the crontab file matches to the username these cronjobs will run as.

GeorgeSapkin:
No updates in over four years. Closing as stale. Feel free to reopen if anything changes.

PR #13355, opened 2020-09-10
isc-dhcp: add support for RFC-3442 classless routes to hosts

pprindeville:
Maintainer: me
Compile tested: x86_64, generic, d75e753
Run tested: same, built package, copied over and installed on running production router

Description:

added support to "config host" sections


pprindeville:
Note: this is useful if you have a device, for instance, that you don't want to have access to the internet, like a thermostat or home security system or other IoT that you're worried about being hacked into.  Also useful if you have multiple subnets, possibly connected with different routers.

PR #10664, opened 2019-11-28
Packages to support the SELinux security mechanism

tpetazzoni:
This patch series integrates a number of additional user-space packages for the various components that are needed to have SELinux support in OpenWrt. It has been sent on the OpenWrt mailing list already, together with another patch series for the core OpenWrt, with other related SELinux changes. But those packages are what is needed first to bring SELinux support in OpenWrt.

PR #13367, opened 2020-09-12
audit: fix dependency error when libev also selected

flyn-org:

Maintainer: @tpetazzoni
Compile tested: x86_64 master

Description:
See also https://github.com/openwrt/packages/pull/10664#issuecomment-691456775.

### DIFF
--- a/utils/audit/Makefile
+++ b/utils/audit/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=audit
 PKG_VERSION:=2.8.5
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=http://people.redhat.com/sgrubb/audit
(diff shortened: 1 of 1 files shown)


neheb:
I don't like this. I've seen this fail with both MPD and gammu.

How about fording the bundled version?

As usual, PKG_RELEASE should be bumped.

flyn-org:
Sorry---what do you mean by "fording?" Is that a typo?

aparcar:
Please bump the PKG_RELEASE and also rebase, I added the breaking `arc_archs` to CI testing.

neheb:
> Sorry---what do you mean by "fording?" Is that a typo?

forcing. Was tryping on a phone.

flyn-org:
> forcing. Was tryping on a phone.

Got it. I stared at what you typed for some time, but for some reason "forcing" did not occur to me!

I decided to go the other direction---audit now unconditionally depends on libev. I am not convinced it is not an OpenWrt build idiosyncrasy that is causing auditd to link against the system libev. From what I can tell, audit still builds its libev parts even when linking against the system libev.

@aparcar, I do not think the CI here would have caught this. This was a case where the package build failed only when libev was also installed. This is the case on the build servers, but not this CI.

PR #13377, opened 2020-09-13
apache: use fakeroot instead of postinst script

micmac1:
fakeroot support was recently added to OpenWrt. Let's make use of it.

Maintainer: @heil 
Compile tested: master ath79
Run tested: ath79 master, installed package to device, checked that permissions are as expected

Description:
This drops postinst script in favor of using fakeroot.

Kind regards,
Seb

### DIFF
--- a/net/apache/Makefile
+++ b/net/apache/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=apache
 PKG_VERSION:=2.4.46
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_SOURCE_NAME:=httpd
 
 PKG_SOURCE:=$(PKG_SOURCE_NAME)-$(PKG_VERSION).tar.bz2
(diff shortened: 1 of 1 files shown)


aparcar:
Thanks for this patch. I somewhat dislike the raw usage of user numbers instead of names. This would however require a `passwd` like file containing a mapping, generated from the USERID in all available feeds. Would you be interested to build such feature?

micmac1:
Hi Paul, my time these days is rather limited. If I had a brilliant quick idea on how to do it I'd do it, though. But currently no such thing comes to mind ;-) Maybe some day in the future.

karlp:
those fixed uids looks absolutely horrific

micmac1:
It's just a number. And it's the same that is used elsewhere for the user creation. As such in my opinion there is nothing horrific about it. It is correct after all and works as expected.

aparcar:
@karlp thanks for your feedback. I'm already working on a easier readable version [here](https://github.com/openwrt/openwrt/pull/3415)

aparcar:
You can now use `apache` instead of a numeric value. Please update this PR.

micmac1:
> You can now use `apache` instead of a numeric value. Please update this PR.

Thanks! Updated.

aparcar:
Great thanks, I'm not the maintainer here so won't merge it. But I approve the change :)

aparcar (approved):

PR #13380, opened 2020-09-13
syslog-ng: tweak shell code of network_localhost little bit

Cynerd:
We can get rid of pipe with -n flag to sysctl.

Maintainer: @BKPepe 
Compile tested:
Run tested: Turris Omnia, OpenWrt 19.07

Description:
It is just simple shell code tweak. It simplifies code. It is not necessary but refactor is always good.

### DIFF
--- a/admin/syslog-ng/Makefile
+++ b/admin/syslog-ng/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=syslog-ng
 PKG_VERSION:=3.29.1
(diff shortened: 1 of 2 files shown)


BKPepe:
Tested.

```
root@omnia:~# sysctl net.ipv6.conf.lo.disable_ipv6
net.ipv6.conf.lo.disable_ipv6 = 0
root@omnia:~# sysctl -n net.ipv6.conf.lo.disable_ipv6
0
```

Thanks.

PR #13388, opened 2020-09-14
numpy,django,pymysql: version bumps

commodo:
Maintainer: me, @peter-stadler (for django)
Compile tested: x86 https://github.com/openwrt/openwrt/commit/a9790dff532e154a93b12abaeccd7889d9159791
Run tested: x86 https://github.com/openwrt/openwrt/commit/a9790dff532e154a93b12abaeccd7889d9159791

--------------------------------


commodo:
need to see about arc700

neheb:
I wouldn't bother. arc700 is an outdated platform. I believe the issue is fixed with GCC9.

There are no real devices that use this target.

PR #13366, opened 2020-09-12
gateway-go: add new package

IoTServ:
Maintainer: me
Compile tested: x86-64, OpenWrt master
Run tested:  x86-64, OpenWrt master

Description:
[OpenIoTHub](https://github.com/OpenIoTHub) contain [server](https://github.com/OpenIoTHub/server-go), [gateway](https://github.com/OpenIoTHub/gateway-go) and [app for mobile](https://github.com/OpenIoTHub/OpenIoTHub) ([iOS APP](https://apps.apple.com/cn/app/云易连/id1501554327) ), OpenIoTHub is best way access your remote network service and IoT Device.
This is OpenIoTHub [gateway](https://github.com/OpenIoTHub/gateway-go) :)

PR #13391, opened 2020-09-14
gateway-go: add new package

IoTServ:
Signed-off-by: Yu Fang <yu@iotserv.com>

Maintainer: me
Compile tested:  x86-64, OpenWrt master
Run tested:  x86-64, OpenWrt master

Description:
[OpenIoTHub](https://github.com/OpenIoTHub) contain [server](https://github.com/OpenIoTHub/server-go), [gateway](https://github.com/OpenIoTHub/gateway-go) and [app for mobile](https://github.com/OpenIoTHub/OpenIoTHub) ([iOS APP](https://apps.apple.com/cn/app/云易连/id1501554327) ), OpenIoTHub is best way access your remote network service and IoT Device.
This is OpenIoTHub [gateway](https://github.com/OpenIoTHub/gateway-go) :)
[refs](https://github.com/openwrt/packages/pull/13366)

### DIFF
--- a/net/gateway-go/Makefile
+++ b/net/gateway-go/Makefile
@@ -1,0 +1,44 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gateway-go
+PKG_VERSION:=0.1.92
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/OpenIoTHub/gateway-go/tar.gz/v${PKG_VERSION}?
+PKG_HASH:=dd8074d9312e00ff957ffd1f3be7ba118a9b8cc31f07aa1ed594ef07931dab16
+
+PKG_MAINTAINER:=Yu Fang <newfarry@126.com>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/gateway-go-$(PKG_VERSION)
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/OpenIoTHub/gateway-go
+
+GO_PKG_LDFLAGS_X:=\
+	main.version=v$(PKG_VERSION) \
+	main.commit=$(PKG_VERSION)  \
+	main.builtBy=openwrt \
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/gateway-go
+  SECTION:=net
+  CATEGORY:=Network
+  SUBMENU:=Web Servers/Proxies
+  TITLE:=gateway-go - gateway of OpenIoTHub
+  URL:=https://github.com/OpenIoTHub/gateway-go
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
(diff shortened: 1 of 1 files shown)


IoTServ:
@neheb
maybe golang compiler not support linux/arc arch
https://github.com/openwrt/packages/pull/13391/checks?check_run_id=1112952402#step:4:3251
https://github.com/openwrt/packages/pull/13391/checks?check_run_id=1112952350#step:4:3248
and do not support linux/powerpc
https://github.com/openwrt/packages/pull/13391/checks?check_run_id=1112952580#step:4:3187
https://github.com/openwrt/packages/pull/13391/checks?check_run_id=1112952599#step:4:3187
maybe need set as ”linux/ppc64“ or ”linux/ppc64le“

I run "go tool dist list" found go compiler support the follow arch pairs：
aix/ppc64
android/386
android/amd64
android/arm
android/arm64
darwin/amd64
darwin/arm64
dragonfly/amd64
freebsd/386
freebsd/amd64
freebsd/arm
freebsd/arm64
illumos/amd64
js/wasm
linux/386
linux/amd64
linux/arm
linux/arm64
linux/mips
linux/mips64
linux/mips64le
linux/mipsle
linux/ppc64
linux/ppc64le
linux/riscv64
linux/s390x
netbsd/386
netbsd/amd64
netbsd/arm
netbsd/arm64
openbsd/386
openbsd/amd64
openbsd/arm
openbsd/arm64
plan9/386
plan9/amd64
plan9/arm
solaris/amd64
windows/386
windows/amd64
windows/arm

aparcar:
Yes, that's the case. @jefferyto do you have a suggestion on how to handle those cases? The CI just skip those jobs, but I think there isn't a "orange" state for GitHub Actions. CircleCI just builds on a target that happens to support Golang.

neheb:
I am very curious why it is being attempted to be built on unsupported targets.

aparcar:
@neheb the CI just runs `make package/gopackage/compile` which fails I guess...

neheb:
The only way that can fail is if there's some InstallDev stuff...

jefferyto:
The CircleCI check also [checks if the package can be enabled](https://github.com/openwrt/packages/blob/832a4e8a4d88dd43ca889e1fe5d7d739df1b344d/.circleci/config.yml#L151-L160) before trying to build it.

I suggest adding similar "can-be-enabled" checking in the Actions check, collecting the packages/architectures that are skipped (while marking those build attempts as successful), and add a [conditional job](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif) that only runs when there are skipped builds (and perhaps have it output the skipped packages/architectures to the log).

Seeing the conditional job run would be the signal that there are builds that were skipped.

jefferyto:
Go program builds on unsupported platforms (arc, powerpc, etc.) will fail immediately (when doing `make package/gopackage/compile`). It has nothing to do with InstallDev.

aparcar:
@jefferyto I modified the `gh-action-sdk` and tagged a v2 which behaves like CircleCI. [Here](https://github.com/aparcar/packages/actions/runs/254692672) is a running example which adds warning annotations if a package is skipped due to problematic architectures.

image

jefferyto:
@aparcar Looks good - I didn't know warnings would be surfaced in the annotations section.

aparcar:
@IoTServ please rebase this PR.

IoTServ:
@aparcar ok

aparcar:
Looks good, unsupported archs are skipped.

IoTServ:
> Looks good, unsupported archs are skipped.

yes, all unsupported architecture (4) skipped

PR #13396, opened 2020-09-15
CI: shellcheck ignore OpenWrt unrelated warnings

aparcar:
@aaronjg found the following unrelated shellcheck warnings which do not
apply to OpenWrt:

* It is unable to find sourced scripts (SC1091), which causes unused
  variable warnigs (SC2039)

* It doesn't seem to understand how variables are assigned using
  config_get, causing many unused variable warnings (SC2039)

* It does not understand the capabilities of busybox ash causing many
  SC2039 warnings

Add a workflow env variable to ignore them.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -17,6 +17,7 @@ jobs:
       - name: Lint Code Base
         uses: github/super-linter@v3
         env:
+          SHELLCHECK_OPTS: "-e SC1091 -e SC2039"
           VALIDATE_ALL_CODEBASE: false
           DEFAULT_BRANCH: master
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


neheb:
It might make sense to use -s ash as well to avoid a bunch of the POSIX warnings.

aaronjg:
It would be better to source the files and not turn off SC2034 for the unused variable warnings. Can the Makefile be parsed to figure out where the scripts will be installed, and can this be run in an OpenWRT environment so that it can find the system scripts?

aaronjg:
>  It might make sense to use -s ash as well to avoid a bunch of the POSIX warnings.

Using -s ash handles a lot of POSIX warnings. Only false positive POSIX warning I am seeing on mwan3 is that it does not know that you can do "echo -e" to allow for escape characters. However, it looks like mwan3 did not need that "-e" anyway.

neheb:
@aaronjg can be worked around by using printf. printf is basically echo -e -n.

One other annoying warning is with tr '[lower]' '[upper'. That can be worked around with awk.

aaronjg:
Is it possible to install OpenWRT without busybox ash? How much portability do we want to encourage for writing shell scripts in packages?

Looks like there are a few other warnings, that I missed in my previous pass, that still come up with -s ash (it appears shellcheck thinks it is dash):
```
In dash, $'..' is not supported.
In dash, &> is not supported.
In dash, 'let' is not supported.
In dash, ++ is not supported.
```

neheb:
So the ash in shellcheck equates to dash, which is not quite the same as busybox ash. busybox ash is somewhere in the middle between dash and bash.

```
In dash, &> is not supported.
In dash, ++ is not supported.
```

These two have POSIX equivalents.

```
In dash, $'..' is not supported.
In dash, 'let' is not supported.
```

No idea about the first. For the second: https://github.com/koalaman/shellcheck/wiki/SC3039

aaronjg:
@neheb  Is there any reason to strive to have all the OpenWRT package code be dash compliant?  I see lots of SC2169 warnings in the OpenWRT source code, so I don't think it would be possible to swap in dash for busybox ash.

```
$ shellcheck -s ash $(find . -name '*.sh') 2>/dev/null|grep 'In dash'| sed 's/^ *//'|sed 's/\^-*^\?//' |sort|uniq |sed 's/SC2169: In dash, //'|sed 's/\(is\|are\)\? not supported.//' 
 $'..' 
 echo flags besides -n not supported.
 exponentials are not supported.
 export -n 
 FDs outside 0-9 are not supported.
 ^ in place of ! in glob bracket expressions 
 == in place of = 
 &> 
 ++ 
 'let' 
 read -n 
 string indexing 
 string replacement 
```

neheb:
I mean...we can't expect a bash script to work under ash. There's some standard that we should arrive to.

Not mandating bash is good for size reasons.

aaronjg:
Agreed, but those are all warnings from code that is in openwrt base already, so presumably they are working in busybox ash. It would be really nice not to have false positive shellcheck warnings for them.

neheb:
We have to be careful when we say false positive. It only seems that way as shellcheck targets dash, not BusyBox ash. In the former case, it is not false.

Having said that, it would be nice to fix up the code in base so that it throws fewer shellcheck warnings.

aaronjg:
Yes exactly. This is not a bug in shellcheck, but it is producing a false positive in the OpenWRT CI environment since BusyBox ash is more capable than dash. 

It would be great to have shellcheck tuned properly so that it raises errors when appropriate. For times when it is raising an issue that can safely be ignored, we can add `# shellcheck disable=SC####` directives.  However, we shouldn't need these around 'standard' OpenWRT practices, such as variables assigned using 'config_get', `network_get_...` etc, or sourcing other files from base OpenWRT, from the package and from the package's dependencies.

Just disabling SC2169, SC1091, SC2034, SC2154, SC2086 etc. will hide errors when shellcheck _should_ be going off.

I worry that as developers get into the habit of ignoring shellcheck errors, it will also train them to ignore build errors from the other CI tests.

neheb:
I assume this is no longer relevant.

PR #13402, opened 2020-09-16
ddns-scripts: move providers into json

feckert:
Maintainer: The way I see it, there is none unique. Many have changed it.
Run tested: x86_64, APU3, I currently have problems with my ddns provider. Therefore I could not make a complete function test. But when I add debug output I see that the information is loaded correctly from the file.

Description:
I think the current ddns-scripts provider handling is a bit cumbersome. To improve this I suggest to outsource the individual providers to JSON. Although outsourcing in JSON increases the size of the package but I see several advantages:
- The provider file does not need to be changed during installation (postinst) and remove (prerm) by ddns providers that require special handling. This simplifies the install and remove script and makes the Makefile clearer.
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L247
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L288
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L326
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L367
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L408
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L446
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L487
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L528
- Simplifies the reading of the ddns provider information. So we do not need the PIPE hack. Whatever it is used for, I didn't understand it and didn't get into it.
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/files/dynamic_dns_functions.sh#L321
- New providers can easily be added. Just create a JSON with the information and put it in the right directory
- It is also not always necessary to add all providers. If we want this we have to change this in the future. For now all providers get installed.

Note:
Since we install **all** providers for now. I have to note that the installation of the package requires more space on the system before we install two files for IPv4(service) and IPv6(service_ipv6). But we can change that so that we do not always install every provider. The big advantage of this change is that the code is maintainable.

From my point of view this is only the first step. There are still other changes that I would like to change. For example change to procd and change to a crond. I do not see the advantage why the script should run as a service. A cron job should be enough here to update the ddns.


Ansuel:
I really like the change...
Wonder if we can find a way to make the package download the needed service on demand... (something like we do with the adblock list... i mean after all if you need a ddns service YOU MUST HAVE AN INTERNET CONNECTION + DDNS right?)

About the cron... I know it would simplify things but i don't really like the idea that the luci of the script will need to parse the cron file and add a rule (or modify the one present)

I still think it's better to have a separate service that can be killed and restarted on demand... What do think about this?

feckert:
@ thanks for your feedback.

> Wonder if we can find a way to make the package download the needed service on demand... (something like we do with the adblock list...

I have already thought of that. Then we would have done away with the size thing right away. Not everyone needs all ddns providers. There should be the possibility to add all of them at build time if you have no space problem. Or you can give a URL to github.com/openwrt/packages/net/ddns-scripts/files/ .. where the json files are and user curl or wget to install them. But we have to think about that again. We could also add them in individual packages? That would be a lot new packages.

>  i mean after all if you need a ddns service YOU MUST HAVE AN INTERNET CONNECTION + DDNS right?)

I agree with you. DDNS is then the topping

> About the cron... I know it would simplify things but i don't really like the idea that luci or the script will need to parse the cron file and add a rule (or modify the one present)

I see that differently. On small systems there is always a service running which does nothing and is not necessary from my point of view. It only needs processor time. I have recently created a [P/R](https://github.com/openwrt/packages/pull/13345#issuecomment-690128290) for acme so that you can use the micrond. The conclution was that I have to update the P/R to use micrond only if this service is installed. This would certainly make it easier to handle the ddns stuff and would remove a lot of lines of code that are not needed anymore. Additional if we simplify the source we could remove a lot of comments that describe what the line of cods to.

> I still think it's better to have a separate service that can be killed and restarted on demand... What do think about this?

If you look at acme there is still a [service](https://github.com/openwrt/packages/blob/master/net/acme/files/acme.init) for procd that handles all that stuff.

Ansuel:
> I have already thought of that. Then we would have done away with the size thing right away. Not everyone needs all ddns providers. There should be the possibility to add all of them at build time if you have no space problem. Or you can give a URL to github.com/openwrt/packages/net/ddns-scripts/files/ .. where the json files are and user curl or wget to install them. But we have to think about that again. We could also add them in individual packages? That would be a lot new packages.

Think the easiest way would be add a config option for the ddns-script to include alle the script at compile time (or create a variant with the option enabled by default) and have the default package download the script by default...
The url to download the services would be the github url where the services are stored for the ""all-services"" variant.
The package will include just a list of the supported url (or we can also make this list to be downloaded by the system so if someone wants to add a new services to the package, it can just add the json and the url to the list and it will magically appear on the remote system without releasing a new version) (but if we need to have a "all-services" variant we will need to update the PKG_RELEASE anyway... so the remove list approach can be useless)

> I see that differently. On small systems there is always a service running which does nothing and is not necessary from my point of view. It only needs processor time. I have recently created a [P/R](https://github.com/openwrt/packages/pull/13345#issuecomment-690128290) for acme so that you can use the micrond. The conclution was that I have to update the P/R to use micrond only if this service is installed. This would certainly make it easier to handle the ddns stuff and would remove a lot of lines of code that are not needed anymore. Additional if we simplify the source we could remove a lot of comments that describe what the line of cods to.

I think it would be better to discuss this in a different PR... Baby steps to improve this. I agree that we should simplify this since the ddns-script in reality should just contact an url and parse the response... I like the idea of use micrond (fun thing the fact that by searching online the first result is a microwave ahahahha) to separate system cron from user cron...

Ansuel:
Some question about the micrond package... It would replace the default crond or just work with it?

feckert:
> Some question about the micrond package... It would replace the default crond or just work with it?

It works in parallel.
- busybox for the root user cron jobs
- micrond for the system cron jobs

The busybox cron has no cron files for system service only for the user root. You must add the cron jobs to the busybox root user cron file. So every user can see this. To hide system cron jobs from user root cron jobs I have suggested to use micrond for system cron jobs instead. The advantage are:
- micrond is that the service has a small footprint. 
- system cron jobs cannot be deleted by user.

> I think it would be better to discuss this in a different PR

I agree with you

dibdot:
@feckert sorry, I can't review/check your changes cause I don't use this package. In general the json approach looks good, personally I would try to merge the IPv4/IPv6 information in one json structure, e.g.:
```
{
  "name": "xxx",
  "urlv4": "yyy",
  "urlv6": "zzz"
}
```
Maybe you can add more useful meta information in the json file like protocol version etc. Said that, I'm not sure if this helps in ddns script logic ...

feckert:
> personally I would try to merge the IPv4/IPv6 information in one json structure, e.g.:
> 
> ```
> {
>   "name": "xxx",
>   "urlv4": "yyy",
>   "urlv6": "zzz"
> }
> ```
> 

@dibdot  I agree, that would be better. But not every ddns provider has ipv6 and ipv4 and therefore the same url or the same answer. I think it is better if we leave this in different directories. This is already handled like this

> > I have already thought of that. Then we would have done away with the size thing right away. Not everyone needs all ddns providers. There should be the possibility to add all of them at build time if you have no space problem. Or you can give a URL to github.com/openwrt/packages/net/ddns-scripts/files/ .. where the json files are and user curl or wget to install them. But we have to think about that again. We could also add them in individual packages? That would be a lot new packages.
> 
> Think the easiest way would be add a config option for the ddns-script to include all the script at compile time (or create a variant with the option enabled by default) and have the default package download the script by default...
> The url to download the services would be the github url where the services are stored for the ""all-services"" variant.
> The package will include just a list of the supported url (or we can also make this list to be downloaded by the system so if someone wants to add a new services to the package, it can just add the json and the url to the list and it will magically appear on the remote system without releasing a new version) (but if we need to have a "all-services" variant we will need to update the PKG_RELEASE anyway... so the remove list approach can be useless)

@Ansuel I am not yet sure how we do that. Do we want to deal with it in this pullrequest or can we live with it for now, that we install **all** providers. The unpleasant thing is that the package might be too big for small devices. Can you verify that? Is this a problem? On my systems I have enough memory, so that from my point of view this is not a problem. The old opkg has 16k and the new one I jused build has 20K. In unpacked condition this is of course still worser.

Ansuel:
@feckert since this is a WIP and it's related to the providers... i would do the changes here with multiple commits... That way we can totally skip the size increase problem...
To sum up my idea is:

- Create a new package that includes all the supported services (a variant that provide ddns-scripts)
- Gen a list of the supported services (to be decided if this will be provided locally or remote)
- Change the current ddns-script to download the requested service json from github (based on the luci configuration)

For other improvement to this logic
- Add a function to refresh the supported service list (if we decided to store it locally)

For security reason (if we really want to implement this)
- The download should be done using https (that shouldn't be a problem since now openwrt support ssl by defaults) and optionally with a config using plain http

feckert:
@Ansuel I have staged another [P/R](https://github.com/TDT-AG/packages/commits/pr/20200917-ddns-scripts) in my repo ontop of them todo additional cleanups. That I would like to see in this as well. So we have a clean working base.

>     * Create a new package that includes all the supported services (a variant that provide ddns-scripts)
I suggest to name the new package **ddns-scripts_common**. If this is selected then all services are installed it depends on **ddns-scripts**.

>     * Gen a list of the supported services (to be decided if this will be provided locally or remote)

Normaly we only install ddns-scripts. I suggest to place the file **/usr/share/ddns/services/list** on build time including all service that are known at this time. So we could genarte a download link for the providers include in this list file.

>     * Change the current ddns-script to download the requested service json from github (based on the luci configuration)

I remembered that we should give the branch where this is loaded at github. Now we have master, but when we have a release it will be called openwrt-20.xx. The scirpt API would be:

- `ddns service update` update the list file from github
- `ddns service list` show all ddns service in the list file
- `ddns service install` xxxx download the service from github
- `ddns service remove` xxxx remove service from 


>     * Add a function to refresh the supported service list (if we decided to store it locally)

See above

> For security reason (if we really want to implement this)

Here we should use wget and I would also suggest that maybe we make the url configurable. Maybe somebody wants to create his own url where he can find the list file and the services. So everybody can decide if he wants to do http or https. I would put these options in the global section of ddns. As described above we have to generate the url in the configuration at build time, because the branch can change if we download this from github.

Ansuel:
> I suggest to name the new package **ddns-scripts_common**. If this is selected then all services are installed it depends on **ddns-scripts**.

Nice idea... so have a general package that have the core files... a package that only contains all the services... much better.. (think we need a better name)



> I remembered that we should give the branch where this is loaded at github. Now we have master, but when we have a release it will be called openwrt-20.xx. The scirpt API would be:

Do we have a way to check the branch directly from openwrt? That way we can provide the general url to the package and make the system at runtime decide the needed url. 

> Here we should use wget and I would also suggest that maybe we make the url configurable. Maybe somebody wants to create his own url where he can find the list file and the services. So everybody can decide if he wants to do http or https. I would put these options in the global section of ddns. As described above we have to generate the url in the configuration at build time, because the branch can change if we download this from github.

Much better...

feckert:
> Nice idea... so have a general package that have the core files... a package that only contains all the services... much better.. (think we need a better name)

I think we should use the name ddns-scripts_services?

> Do we have a way to check the branch directly from openwrt? That way we can provide the general url to the package and make the system at runtime decide the needed url.

The command to get the branch is `git rev-parse --abbrev-ref HEAD`. But I have to look if we get this also on a fresh/new build in the Makefile. I have not seen that git is used in an openwrt Makefile.
If I clone the package feed and enter the command then I get the right branch.

@Ansual Please  also see my comments on https://github.com/TDT-AG/packages/commit/fd42dac6d82c0e3917714a69e5366a753f2666ee#commitcomment-42395758

Ansuel:
> The command to get the branch is `git rev-parse --abbrev-ref HEAD`. But I have to look if we get this also on a fresh/new build in the Makefile. I have not seen that git is used in an openwrt Makefile.
> If I clone the package feed and enter the command then I get the right branch.

I mean from a running system...  (set the general url and get the branch based on the image flashed on the system... but i don't know if the running system can get the running branch)

feckert:
@Ansuel Thanks for the review and the input. I have update this P/R with my whole work to use json ddns services selection and ddns packaging cleanup. The ddns services are now moved to a new package **ddns-scripts_service** which is not installed by default.

I think there is still work todo on the new `ddns` command to get the service list on demand.
Maby we should add two urls one for the service list and one for the service description directory on the server to get the services json. I have also added the `cacert` option für service update to verify the ssl cert.

I checked it but since there were a lot of entries it could have happened that by splitting the urls from the service files to the individual json files a copy and past error occurred. Can you or someone else please verify this?

Ansuel:
Again thx a lot for the great work...

- We still need to decide if we want to use master branch for everything or we want to parse the branch and change the url accordingly (both in the compile process or runtime...)

Will check the json conversion
Any other hint about this? I think we are good

Ansuel:
@feckert how you generated the list? Checking the json file and the new list these services lacks of a json file

- ddnss.de.old
- dyntdt.de

feckert:
> @feckert how you generated the list?

Semi automatic. Some of them must have slipped through!
Many thanks for the check :+1:

> - ddnss.de.old

Is comment out in the [service](https://github.com/openwrt/packages/blob/master/net/ddns-scripts/files/services_ipv6#L50) file. Therefore we do not need it any more. I named it old and forgot to delete them.

> - dyntdt.de

This is the one from my company. I'm deleting it because it is not public. My mistake.

> - noip.com

That is the follwing entry [](https://github.com/openwrt/packages/blob/master/net/ddns-scripts/files/services_ipv6#L38) so this is a special [script](https://github.com/openwrt/packages/blob/master/net/ddns-scripts/files/update_no-ip_com.sh). I have also removed this.

> - 3222.org is a typo (should be 3322.org, domain 3222.org is on sale LOL)

I should buy them ;-).
No kidding. I have fixed and renamed it.

> And you added new service?
> - cnkuai.cn
> - bind-nsupdate
> - digitalocean.com-v2

These are the services with the scripts that I have now deleted - we don't need them. These are the services from the special packages. Also removed the other special scripts servives which were also in this file.

Ansuel:
Think this is not WIP anymore?

feckert:
@Ansuel 
> Think this is not WIP anymore?

No I have to update the json files. As far as I remember I still have to escape something.
And I have to update my code for ddns. I am not yet satisfied with this and have to run the shellcheck on it.

feckert:
Also the build test [failed](https://github.com/openwrt/packages/pull/13402/checks?check_run_id=1134016557#step:4:6389)
Because I install ddns.sh into the wrong [dir](https://github.com/openwrt/packages/pull/13402/checks?check_run_id=1134016557#step:4:6389). I have to look on this on Monday

Ansuel:
> shellcheck

this will be a mess ahahah

feckert:
@Ansuel So I have now fixed my outstanding tasks. And the CI builds the package again.
I have also updated the PKG_VERSION from 2.7.8 to 2.8.0.
The bigger version jump should make clear that something important has changed.

Ansuel:
@feckert should be good now right?

feckert:
@Ansuel now we have to update luci-app-ddns to use not the service and service6 file

Ansuel:
And also the cacert global option

feckert:
@Ansuel  That would be great

Ansuel:
@feckert just one info... I wan to add also the button to refresh / download the services...
Any idea how to handle that without encounter a timeout?

Ansuel:
Anyway working on it 👍 
If you want you can start working on the cleanup process

Ansuel:
@feckert I almost finished the luci app work... I notice that in the list there is no cloudflare service... am i missing something?

feckert:
Is that what you are looking for https://github.com/openwrt/packages/pull/13402/files#diff-0fbdd382013e3dc994432db4af944d35?

That is a special script and so in a own package

Ansuel:
Mh ok I need your help then...

How it work now is that the luci app use a ubus command to get the service file (now that i think about it... why not use the standard luci way to download a file... anyway...)

The standard list lacks of this special package...  Any idea how to add this special json to the standard list? (a postinst script is not possible since the list can be updated and the added entry will be lost)
Wonder if we can create an external/custom list where user/these package can add json file and make the script check for both. What do you think?

feckert:
> Wonder if we can create an external/custom list where user/these package can add json file and make the ddns script check for both. What do you think? (or a separate directory where to put all the custom json)

I had not even thought of that. That we would now have problems if someone also installed services from github and from opkg. I think the best solution now is to create a custom directory with the services and a custom file with the installed services and put theme there. The new ddns tool should only work with the costum DDNS services.

> The standard list lacks of this special package... Any idea how to add this special json to the standard list? (a postinst script is not possible since the list can be updated and the added entry will be lost)

If we use opkg then the special services also get installed into the service directory.

> How it work now is that the luci app use a ubus command to get the service file list (now that i think about it... why not use the standard luci way to download a file... anyway...)

I am not sure what you mean? But I would use the this code https://github.com/openwrt/luci/blob/master/applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js#L159 to get all installed services. (custom and from opkg)

Ansuel:
> I had not even thought of that. That we would now have problems if someone also installed services from github and from opkg. I think the best solution now is to create a custom directory with the services and a custom file with the installed services and put theme there. The new ddns tool should only work with the costum DDNS services.

To not add too many code I think we can just add a custom list and keep all the custom script in the services list... That way we can still keep separate opkg services and not change the ddns script to try to load the json from another directory.
(the opkg service will just ad the entry in the list)

> If we use opkg then the special services also get installed into the service directory.

They get placed in the services dir but not added to the list. And from what I have done to the luci app the services list is based on that...

> I am not sure what you mean? But I would use the this code https://github.com/openwrt/luci/blob/master/applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js#L159 to get all installed services. (custom and from opkg)

Yep think I will drop the custom ubus call and use the standard way to get files...
Anyway I will post a WIP to the luci repo and tag you.

jow-:
Please avoid using underscores in package names, this may lead to undefined behaviour in buildroot and opkg.

feckert:
@jow-  I just followed the convention in the ddns-scripts package
There are all the [additional](https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L387) packages with an underscore in it. Should we change that?

jow-:
Yes, it should be changed

PR #13398, opened 2020-09-15
zmq: update to version 4.3.3 (security fix)

ja-pa:
maintainer: @srdgame 
Compile tested: Turris Omnia (TOS6), OpenWrt master
Run tested: Turris Omnia (TOS6), OpenWrt master

Description:
This PR updates zmq to version 4.3.3. It fixes [CVE-2020-15166](https://nvd.nist.gov/vuln/detail/CVE-2020-15166) and multiple other issues see [Changelog](https://github.com/zeromq/libzmq/releases/tag/v4.3.3)

Run tested with internal tests (https://github.com/zeromq/libzmq/tree/master/tests) 

This should be also backported to **OpenWrt 19.07**

PR #13404, opened 2020-09-16
zmq: enable libbsd by default

ja-pa:
maintainer: @srdgame 

Description:
This PR enables libbsd and adds it to dependency.  This suppresses libbsd autodetection.

Related to https://github.com/openwrt/packages/pull/13398


neheb:
NAK. libbsd is a pointless dependency. It's only useful for glibc. Both musl and uclibc-ng support the strl* functions.

Alternative: https://github.com/openwrt/packages/pull/13406

ja-pa:
Ok, thx for the fix. closing

PR #13407, opened 2020-09-17
mpd: switch to using libnpupnp

neheb:
Removed iconv const patch. After discussing with upstream. it turns out
that libiconv-full in OpenWrt is broken.

Maintainer: nobody
Compile tested: ath79
Run tested: TPLink Archer C7 v2

Runtime testing has revealed some unfortunate things.

1: UPnP is slow. I tried with both libupnp and libnpupnp. The issue might be with my copy of gerbera. Gerbera goes into 100% CPU usage. foobar2000 works perfectly though... ping @medoc92

2: mp4 playback is totally broken. That is, only ffmpeg can play it. The faad plugin can only play raw aac streams. It might make sense to completely remove it. Then again, I do remember a user that was using faad to play radio streams.


neheb:
@aparcar what are these errors?

aparcar:
@neheb I added runtime tests which try to install a package and if exists, install it. Seems like `mpd` isn't easily installable?

medoc92:
@neheb maybe try to start gerbera with debug logging enabled (I think that it's --debug but really not sure) to see what it is doing ?

neheb:
Meh that's unfortunate.

medoc92:
I tested the mpd upnp plugin with both minimserver and the upmpdcli media server (all on linux amd64), and I don't see any particular issues. I did not test with gerbera, because I can't build it on my Ubuntu bionic system.

neheb:
I think it's because I'm running Gerbera on a slow ARM based system with ~13k entries. MPD I guess times out.

PR #13410, opened 2020-09-17
json-glib: bump to 1.6.0

micmac1:
Version bump. New stable branch.

ABI_VERSION added. Tests can now be disabled directly.

Maintainer: me
Compile tested: ath79 master
Run tested: 19.07 ath79. Made phone calls through kamailio via asterisk between phones, proxying media through rtpengine linked against json-glib 1.6. No problems found.

Description:
Update to new stable branch.

Thanks!
Seb

### DIFF
--- a/libs/json-glib/Makefile
+++ b/libs/json-glib/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=json-glib
-PKG_VERSION:=1.4.4
+PKG_VERSION:=1.6.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_SOURCE_URL:=@GNOME/json-glib/1.4
-PKG_HASH:=720c5f4379513dc11fd97dc75336eb0c0d3338c53128044d9fabec4374f4bc47
+PKG_SOURCE_URL:=@GNOME/json-glib/1.6
+PKG_HASH:=0d7c67602c4161ea7070fab6c5823afd9bd7f7bc955f652a50d3753b08494e73
 
 PKG_INSTALL:=1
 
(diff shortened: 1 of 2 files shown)


neheb:
```
../../../../../../build/openwrt/build_dir/target-i386_pentium4_musl/json-glib-1.6.0/meson.build:153:0: ERROR: <ExternalProgram 'python3' -> ['/home/build/openwrt/staging_dir/host/bin/python']> is not a valid python or it is missing setuptools
```

micmac1:
> ```
> ../../../../../../build/openwrt/build_dir/target-i386_pentium4_musl/json-glib-1.6.0/meson.build:153:0: ERROR: <ExternalProgram 'python3' -> ['/home/build/openwrt/staging_dir/host/bin/python']> is not a valid python or it is missing setuptools
> ```

Thanks! Added the sed scripts back.

neheb:
Same error. The problem is here: https://github.com/GNOME/json-glib/blob/master/meson.build#L154

Replace with

```
python3 = 'python3'
```

or remove if not needed.

micmac1:
I don't really get why it works at home and apparently also in circlecli. It is the new tests that are failing, right? I don't really have time for this right now. Will check later.

neheb:
It's an issue with debian's packaging of python.

micmac1:
Removed sed scripts again. Verified that tests are indeed disabled via MESON_ARGS and that build_aux/gen-installed-test.py is indeed not run.

neheb:
You also need the change I mentioned.

micmac1:
Is that because of the Debian python issue? If so I'd want to look at this closer. I haven't looked at the new tests that were introduced into the packages repo at all.

My first instinct is if there's a problem with Debian python package it should be fixed there, so all can benefit. I mean lots of PRs have a red "x" these days.

Will look into it more when I get around to it. Have a good weekend!

Am September 19, 2020 10:48:22 AM UTC schrieb Rosen Penev <notifications@github.com>:
>You also need the change I mentioned.
>

PR #13413, opened 2020-09-17
[TEST] gnunet: add runtime test

dangowrt:
Test if gnunet-arm is executable for now.

### DIFF
--- a/net/gnunet/test.sh
+++ b/net/gnunet/test.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+case "$1" in
+	"gnunet")
+		gnunet-arm --version | grep "$2"
+		;;
+esac


aparcar:
```
2020-09-17T22:10:03.6795211Z Use package specific test.sh
2020-09-17T22:10:03.6836006Z gnunet-arm v0.13.3 release
2020-09-17T22:10:03.6839951Z Test successfull
```

Simple test but work.

net/gnunet/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+gnunet-arm --version
```
aparcar, line 3:
```suggestion
gnunet-arm --version | grep "$2"
```

This way it checks if the current version is actually printed, not just if something (e.g. Segfault) is printed.

aparcar (changes requested):

aparcar:
Not sure what happened here but this seems merged https://github.com/openwrt/packages/commit/25fb4794fa37956be58b3dffc05d5a211cac8cf8

PR #12222, opened 2020-05-19
Fix ttyd link error on 19.07 using glibc:/lib/libdl.so.2: error addin…

topeqin:
…g symbols: DSO missing from command line

Signed-off-by: Qin Guang <topeqin@gmail.com>

Maintainer: tsl0922@gmail.com
Compile tested: (arm bcm53xx and bcm2708, OpenWrt 19.07 master)
Run tested: (arm, bcm53xx, OpenWrt 19.07, tests done)

Description:
ttyd meet libdl.so.2: error adding symbols: DSO missing from command line
When compile bcm53xx ARM target no matter using openwrt default toolchain with glibc not musl or broadcom external toolchain both meet below link DSO error.
...... 
[ 16%] Building C object CMakeFiles/ttyd.dir/src/server.c.o
[ 33%] Building C object CMakeFiles/ttyd.dir/src/http.c.o
[ 50%] Building C object CMakeFiles/ttyd.dir/src/protocol.c.o
[ 66%] Building C object CMakeFiles/ttyd.dir/src/terminal.c.o
[ 83%] Building C object CMakeFiles/ttyd.dir/src/utils.c.o
[100%] Linking C executable ttyd
/home/eq936498/vanilla/openwrt/staging_dir/toolchain-arm_cortex-a9_gcc-8.4.0_glibc_eabi/lib/gcc/arm-openwrt-linux-gnueabi/8.4.0/../../../../arm-openwrt-linux-gnueabi/bin/ld: /home/eq936498/vanilla/openwrt/staging_dir/target-arm_cortex-a9_glibc_eabi/usr/lib/libwebsockets.a(unix-plugins.c.       o): undefined reference to symbol 'dlopen@@GLIBC_2.4'
/home/eq936498/vanilla/openwrt/staging_dir/toolchain-arm_cortex-a9_gcc-8.4.0_glibc_eabi/lib/gcc/arm-openwrt-linux-gnueabi/8.4.0/../../../../arm-openwrt-linux-gnueabi/bin/ld: /home/eq936498/vanilla/openwrt/staging_dir/toolchain-arm_cortex-a9_gcc-8.4.0_glibc_eabi/arm-openwrt-linux-gnueabi/bin/../../../toolchain-arm_cortex-a9_gcc-8.4.0_glibc_eabi/lib/libdl.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
CMakeFiles/ttyd.dir/build.make:170: recipe for target 'ttyd' failed
[...]

PR #13423, opened 2020-09-18
ttyd: fix compilation under glibc

neheb:

Maintainer: @tsl0922 

Continuation of : https://github.com/openwrt/packages/pull/12222

I'll try to compile this when I can.

@tsl0922 do you know why the original user tried building against libdl? I don't see any dl function usage in the codebase.


neheb:
Oh this is a libwebsockets problem. Never mind.

Issue #12805, opened 2020-07-13
ImageMagick: Compilation failed with glibc GCC 10


DazzyWalkman:
Maintainer: @val-kulkov
 
package version: 7.0.9-5, x86-64, snapshot r13828-cef9e5a49f

In fact I am not quite sure it's related to GCC 10, however, nothing changed except I tried switching to the lastest GCC 10 toolchain.

Seemed like there's one step that failed silently, making an empty part between "|"
in command `libtool: link: x86_64-openwrt-linux-gnu-gcc-nm  coders/.libs/aai_la-aai.o   |  | /opt/openwrt/staging_dir/host/bin/sed 's/.* //' | sort | uniq > coders/.libs/aai.exp`
, thus throwing out an error.

I haven't got hold of a previously successful copy of buildlog to see what actually is.

Bumping the IM version to 7.0.10-23 doesn't help, either.

Here is the related part of the faillog:
```
make  all-am

make[4]: Entering directory '/opt/openwrt/build_dir/target-x86_64_glibc/ImageMagick-7.0.9-5'
[...]

PR #13432, opened 2020-09-20
imagemagick: do not use flto under GCC10

neheb:
Causes compilation failure.

Maintainer: @val-kulkov 
Compile tested: ath79

Fixes: https://github.com/openwrt/packages/issues/12805


val-kulkov:
@DazzyWalkman posted a [comment](https://github.com/openwrt/packages/issues/12805#issuecomment-696565324) 17 hours ago that the issue was no longer present with gcc 10.2.0, glibc and binutil 2.35. Therefore, is this patch necessary?

neheb:
Nope. That's why I closed it.

PR #13442, opened 2020-09-20
gateway-go: add default config and gateway-go.init for background service

IoTServ:
Signed-off-by: Yu Fang <newfarry@126.com>

Maintainer: me
Compile tested: x86-64, openwrtorg/rootfs docker image container
Run tested: x86-64, openwrtorg/rootfs docker image container

Description:
gateway-go: add default config and gateway-go.init for background service


neheb:
Needs a PKG_RELEASE bump.

PR #13454, opened 2020-09-21
elektra: fix compilation when cargo is present

neheb:

Maintainer: @haraldg 
Compile tested: Fedora 32, CentOS 7


haraldg (approved):
Thanks

PR #13461, opened 2020-09-22
mpd: update to 0.21.26

neheb:

Maintainer: nobody
Compile tested: ath79


dhewg:
Not related to this bump, but I just ran into the situation that mpd failed to compile:
* I only selected mpd-mini
* libmicrohttpd was not selected

And the build failed, because libnpupnp complained because libmicrohttpd was missing. I think it tried to build the full variant too? But the dependencies were not built for that. Manually running `make package/libmicrohttpd/compile` made the complete build succeed afterwards, even though nothing of all of that was part of the image...

neheb:
That's strange. libnpupnp is only used for mpd-full. I'll look into this.

neheb:
Duplicated. I don't understand why it's failing. libnpupnp has a dependency on libmicrohttpd. libnpupnp shouldn't even be built. It's only used for mpd-full.

PR #13467, opened 2020-09-22
tor: update to version 0.4.4.5

ja-pa:
maintainer: @tripolar 

Description:
This PR updates tor to version 0.4.4.5  Series 0.4.x should be supported till June 2021

[Changes](https://blog.torproject.org/node/1921)

Run tested with torsocks

### DIFF
--- a/net/tor/Makefile
+++ b/net/tor/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tor
-PKG_VERSION:=0.4.3.6
+PKG_VERSION:=0.4.4.5
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://dist.torproject.org/ \
 	https://archive.torproject.org/tor-package-archive
-PKG_HASH:=6a2d0637d4e514be2ec574723a05065245cce51da78a21cec1dc831be5ccac62
+PKG_HASH:=a45ca00afe765e3baa839767c9dd6ac9a46dd01720a3a8ff4d86558c12359926
 PKG_MAINTAINER:=Hauke Mehrtens <hauke@hauke-m.de> \
 		Peter Wagner <tripolar@gmx.at>
 PKG_LICENSE_FILES:=LICENSE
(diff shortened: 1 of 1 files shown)


ja-pa:
For arc it's failing on operand out of range. Don't know why. @neheb Do you have any suggestions, please?

```
src/lib/meminfo/meminfo.c: In function 'tor_log_mallinfo':
src/lib/meminfo/meminfo.c:49:8: warning: function call has aggregate value [-Waggregate-return]
   mi = mallinfo();
        ^
{standard input}: Assembler messages:
{standard input}:14660: Error: operand out of range (-130 is not between -128 and 127)
```

neheb:
GCC bug. I'd ignore.

r-salvaterra:
As an aside, in the package makefile we have `--with-pic` in the `CONFIGURE_ARGS`, when the correct autoconf switch is `--enable-pic`, I believe this should be fixed too.

ja-pa:
> As an aside, in the package makefile we have `--with-pic` in the `CONFIGURE_ARGS`, when the correct autoconf switch is `--enable-pic`, I believe this should be fixed too.

Thanks, fixed.

ja-pa:
Arc_arc700 test is still failing because of GCC bug.
```
src/lib/meminfo/meminfo.c: In function 'tor_log_mallinfo':
src/lib/meminfo/meminfo.c:49:8: warning: function call has aggregate value [-Waggregate-return]
   mi = mallinfo();
        ^
{standard input}: Assembler messages:
{standard input}:14660: Error: operand out of range (-130 is not between -128 and 127)
```

neheb:
https://github.com/torproject/tor/blob/release-0.4.4/src/lib/meminfo/meminfo.c#L46

This is true under uClibc-ng/glibc. Not under musl. I'd try disabling it somehow.

neheb:
Just pass ac_cv_func_mallinfo=no to CONFIGURE_VARS.

ja-pa:
@neheb It looks, that it didn't help. Or Am I missing something? https://github.com/openwrt/packages/pull/13467/commits/db5f30cc412aed2594e6a14774c5b68e251bbeed#diff-c15961669a77858fb31cd6d73d77c3f4R114

neheb:
I looked into this again. It's a compiler problem. There's no good way to fix it without updating GCC to GCC9. Let's ignore it for now.

Please update your 19.07 backport.

ja-pa:
> I looked into this again. It's a compiler problem. There's no good way to fix it without updating GCC to GCC9. Let's ignore it for now.
> 
> Please update your 19.07 backport.

Ok, Thx. I disabled mallinfo also in https://github.com/openwrt/packages/pull/13468

PR #13478, opened 2020-09-23
netifyd: Updated to v3.05

dsokoloski:

Change log for v3.05:

[FIX] Fixed packet filtering for offline capture files.
[FIX] Removed redundant check for sink or JSON save flags.
[FIX] Fixed incorrect references to libpcap len to caplen.
[FIX] Fixed application detection order for TLS (SSL) flows.
[FIX] EdgeOS: Fixed installer return code.

[IMP] Added missing IPv4/6 private IP blocks.
[IMP] Added ability to suppress uploading NAT flows (defaults to not upload).
[IMP] Added some basic command options (--help).
[IMP] Added conntrack ID and mark to flow structure when enabled and available.
[IMP] Reset CURL handle after too many POST errors.
[IMP] Refactored application detection logic.
[IMP] Always perform a DNS Hint Cache lookup (when enabled).  Populate host_server_name if a match was found and host_server_name was not previously set.
[IMP] Added new JSON field (when available): dhc_hit (DNS Hint Cache Hit).
[IMP] OpenWrt: Integrate with uci to support 'enabled/disabled' startup option and override Netify's default auto-interface detection.
[IMP] OpenWrt: Preserve important Agent config files across sysupgrade: agent.uuid, serial.uuid, and site.uuid

Signed-off-by: Darryl Sokoloski <darryl@sokoloski.ca>


neheb:
```
WARNING: PKG_MIRROR_HASH does not match netifyd-2020-09-15-v3.05.tar.xz hash 2bcaa54f1660a30070f0c79a79259e96f169a9c7a0fe4a1195bb74f7de184204
```

commit title also looks wrong.

neheb:
If that hash is indeed correct for your end, it would be good to upload somewhere to see why it is different.

dsokoloski:
Oops!  Will fix!

PR #13486, opened 2020-09-24
coreutils: fix compilation for when selinux is present

elkablo:
According to https://bugs.gentoo.org/301782 coreutils does not respect
--without-selinux correctly when libselinux.so is present on the target
system (in the staging dir).

This solves a weird issue of some coreutils programs not building (for
example stdbuf), because the configure script uses -Werror to determine
whether __attribute__((constructor)) is respected, but -Werror causes
compilation to fail because of a warning about redefinition of
HAVE_SELINUX_SELINUX_H macro.

Signed-off-by: Marek Behún <kabel@blackhole.sk>

Compile tested: mvebu

### DIFF
--- a/utils/coreutils/Makefile
+++ b/utils/coreutils/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=coreutils
 PKG_VERSION:=8.32
-PKG_RELEASE:=2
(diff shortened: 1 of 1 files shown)


neheb:
How strange.

Please bump PKG_RELEASE.

aparcar (approved):

PR #13500, opened 2020-09-26
CI: test build with Full Lanaguage Support

aparcar:
The added SDK is build with `CONFIG_BUILD_NLS` which enabled full language support. According to @dangowrt this is a common problem in packages, so we could add this test as well.

Containers are created by another CI job and official and also the labelling is still wip (x86-64 instead of x86_64).

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   build:
-    name: Test ${{ matrix.arch }}
+    name: Test ${{ matrix.arch }} ${{ matrix.config }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


neheb:
I believe vim has a problem with full NLS. A bunch of others too.

aparcar:
Odd, this one doesn't seem to fail. To I need any additional options for this to happen?

neheb:
This CI doesn't work looks like.

BKPepe:
Oh, this one would be really handy to have. 🙌  We need to have additional Full Language Support check! 😇

BKPepe:
@aparcar Anything new?

aparcar:
Not really, first we'd need to create SDKs with NLS support which is its entire journey on its own. We could use CI for that and do it weekly instead of daily to save some computation. Would you be up to look into that @BKPepe?

openwrt-ai (commented):
Reviewed 2 new commits; no new issues found.

Both diffs are intentional CI scaffolding whose caveats are already acknowledged by the author: the `arch: x86-64` label mismatch is noted in the PR description as WIP, and the `gphoto2` `PKG_RELEASE` bump is described in its own commit message as a throwaway ("pointless bump to test CI") that should be dropped before merge. Commit messages match their diffs, so no commit-check discrepancies to report.

---

Issue #13495, opened 2020-09-25
ddns-scripts: DDNS Service provider is not available


peterwillcn:
LuCI Master git-20.267.53658-b17ec09
    ddns-scripts	2.8.0-24	
    luci-app-ddns	git-20.250.44589-809d085	

![Screen Shot 2020-09-26 at 7 31 37 AM](https://user-images.githubusercontent.com/323794/94324240-91ee9f00-ffcb-11ea-82e7-fb54a27ea4f3.png)

snakwu:
@feckert
me to！！！

Collected errors:                                                               
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for ddns-
scripts_nsupdate:                                                               
 *      bind-client                                                             
 * opkg_install_cmd: Cannot install package ddns-scripts_nsupdate.              
 * check_data_file_clashes: Package ddns-scripts_service wants to install file /
home/snakwu/openwrt-master/openwrt/build_dir/target-x86_64_musl/root-x86/usr/sha
re/ddns/services/cnkuai.cn.json                                                 
        But that file is already provided by package  * ddns-scripts_cnkuai_cn  
 * check_data_file_clashes: Package ddns-scripts_service wants to install file /
home/snakwu/openwrt-master/openwrt/build_dir/target-x86_64_musl/root-x86/usr/sha
re/ddns/services/freedns.42.pl.json                                             
        But that file is already provided by package  * ddns-scripts_freedns_42_
pl                                                                              
 * check_data_file_clashes: Package ddns-scripts_service wants to install file /
home/snakwu/openwrt-master/openwrt/build_dir/target-x86_64_musl/root-x86/usr/sha
re/ddns/services/route53-v1.json                                                
        But that file is already provided by package  * ddns-scripts_route53-v1 
 * opkg_install_cmd: Cannot install package ddns-scripts_service.               
make[2]: *** [package/Makefile:68: package/install] Error 255                   
make[2]: Leaving directory '/home/snakwu/openwrt-master/openwrt'                
make[1]: *** [package/Makefile:108: /home/snakwu/openwrt-master/openwrt/staging_
dir/target-x86_64_musl/stamp/.package_install] Error 2                          
make[1]: Leaving directory '/home/snakwu/openwrt-master/openwrt'                
make: *** [/home/snakwu/openwrt-master/openwrt/include/toplevel.mk:236: world] E
rror 2

Ansuel:
[...]

PR #13509, opened 2020-09-28
ddns-scripts: fix regression added by version 2.8.0

feckert:
Maintainer: me / @Ansuel 

Description:

- Rename package for services installation https://github.com/openwrt/packages/pull/13402#issuecomment-699025971
- Rename packages for special services installation https://github.com/openwrt/packages/pull/13402#issuecomment-699025971
- Remove installation list file for custom providers this file gets installed by the new ddns script
- Refactoring new ddns script to save additional providers into custom directory
- Fix installation json files for special providers

The following issues should get resolved with this P/R.
https://github.com/openwrt/packages/issues/13495
https://lists.openwrt.org/pipermail/openwrt-devel/2020-September/031507.html

@Ansuel I have not add the dependency ddns-scripts-services to ddns-scripts for now. When I do that the menuconfig does not look clean! Could you verify this.


Ansuel:
What problem do you have with the deps?
```
define Package/ddns-scripts
  $(call Package/ddns-scripts/Default)
  TITLE:=Dynamic DNS Client scripts (with IPv6 support)
  DEPENDS:=+ddns-scripts-services
endef

define Package/ddns-scripts/description
  Dynamic DNS Client scripts (with IPv6 support)
  A highly configurable set of scripts for doing dynamic dns updates.
  - IPv6 support
  - DNS server support
  - Glue Record support (require BIND host or KNOT host)
  - DNS requests via TCP
  - Proxy server support
  - log file support
  - support to run once
  Version: $(PKG_VERSION)-$(PKG_RELEASE)
  Info   : https://openwrt.org/docs/guide-user/services/ddns/client
endef

define Package/ddns-scripts/conffiles
/etc/config/ddns
endef


define Package/ddns-scripts-services
  $(call Package/ddns-scripts/Default)
  TITLE:=Common ddns providers
#   DEPENDS:=ddns-scripts
endef
```
Immagine 2020-09-15 000925

net/ddns-scripts/Makefile
```diff
@@ -224,14 +222,24 @@ exit 0
...
+	rm $(1)/usr/share/ddns/services/digitalocean.com-v2.json
+	rm $(1)/usr/share/ddns/services/no-ip.com.json
+	rm $(1)/usr/share/ddns/services/bind-nsupdate.json
+	rm $(1)/usr/share/ddns/services/route53-v1.json
+	rm $(1)/usr/share/ddns/services/cnkuai.cn.json
 endef
 
 
-define Package/ddns-scripts_cloudflare.com-v4/install
+define Package/ddns-scripts-cloudflare/install
```
Ansuel, line 242: I would keep the same name and just change the '_' to '-'. People would get confuse and AFAIK there could be various version of the same script.

Ansuel:
I was thinking of renaming the 'services' dir to 'default'. This would make it more clear that in that dir there are json installed by the opkg and should not be touched...
Aside from this. Can we keep the list file on the repo but not install it?

net/ddns-scripts/Makefile
feckert: > I would keep the same name and just change the '_' to '-'. People would get confuse and AFAIK there could be various version of the same script.

I do not understand that. We can add more scripts in the future. So I think it's legitimate if we keep the name more generic. The json file is still that way. So the uci parameter will not change.

feckert:
> I was thinking of renaming the 'services' dir to 'default'. This would make it more clear that in that dir there are json installed by the opkg and should not be touched...

Sounds good :+1:. I will change that.

> Aside from this. Can we keep the list file on the repo but not install it?

This should fit. I have only removed the installation call and the file is still in the repo

> I have not add the dependency ddns-scripts-services to ddns-scripts for now. When I do that the menuconfig does not look clean! Could you verify this.

forget it. That has been cleared up I have rebuilt and then it worked. But thanks

Ansuel:
> This should fit. I have only removed the installation call and the file is still in the repo

Didn't notice... Will update the luci app pr with the default dir so we can merge both... What do you think?

feckert:
@Ansuel I have to update this P/R with another commit to read also the custom json values.
For now the [read function](https://github.com/openwrt/packages/blob/master/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh#L332) does only read the files from the services directory (In the future this directory is changed in this P/R to default, as you suggested).

But we could not read the custom directory if we have installed a provider in this directory for now.
Any solution to fix this?

My suggestions are:

- Read also in another call the values from the custom directory if we do not find the json file in the default directory
- Check first the custom directory and then the default directory

An if we find no json then we gave up?

I would use the second suggest. So we have always use the newest json information if we try to read first the custom directory.

Ansuel:
> I would use the second suggest.

Yes the default json should be the fallback if custom is not provided. Maybe add a comment somewhere to make this clear (that any json have priority than default?)

net/ddns-scripts/Makefile
Ansuel: The old name directly told what service the pkg installs (without reading the description). But now that I think about it's better to have a more generic name (if more version of the same service will comes up, the generic special pkg will install all of them). Sorry for the mess.

snakwu:
@feckert @Ansuel   
什么时候可以合并？

neheb:
```
 In net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh line 48:
	printf %s\n "$MYPROG: $@" >&2
                               ^-- SC2145: Argument mixes string and array. Use * or separate argument.
```

feckert:
> ```
>  In net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh line 48:
> 	printf %s\n "$MYPROG: $@" >&2
>                                ^-- SC2145: Argument mixes string and array. Use * or separate argument.
> ```

@neheb Thanks for your advice. But I don't want to deal with the shellcheck things in this PR. Now I have to make sure that the JSONs work properly. I want to update some other things in the package. The restructuring of the package is not yet complete. There is still a lot to be done.

feckert:
@Ansuel I would merge the changes now. If it works for you, too.

Ansuel:
Sorry for the delay... I tested now and it seems to work all correctly

ivanich:
Not sure how but this bricks my edgerouterx, had to recovery via tftp. When I remove this commit from packages repo it works fine

feckert:
@ivanich That are only script changes. So what did you exactly do. This changes were recently committed so i do not think this has already picked up by the buildbot and that this package are available in the snapshot for installation.

ivanich:
@feckert I make my own builds for 3 routers including edgerouter for a quite long time. As I said before build with this commit reverted works fine. I even tried to disable all ddns staff in the config but it still wont boot.

ivanich:
@feckert I simply build openwrt from sources as I I always do. Basically build I did before your commit works fine, with your commit router won't boot i.e. no ethernet links on all lan ports except WAN.

PR #13512, opened 2020-09-28
https-dns-proxy: update binary to 2020-08-21

stangri:
Maintainer: me
Compile tested: mvebu, WRT3200ACM, 18.06.8/19.07.4
Run tested: mvebu, WRT3200ACM, 18.06.8, start/stop, query DoH servers

Description: this updates to the latest release binary and contains cosmetic fixes to README and init script.

PS. I'm also requesting merge rights to packages.


net/https-dns-proxy/files/README.md
```diff
@@ -1,5 +1,8 @@
+<!-- markdownlint-disable MD013 -->
 # DNS Over HTTPS Proxy (https-dns-proxy)
 
+[![HitCount](http://hits.dwyl.com/stangri/openwrt/https-dns-proxy.svg)](http://hits.dwyl.com/stangri/openwrt/https-dns-proxy)
```
BKPepe, line 4: No way. Sorry. NACK for this change.

BKPepe, line 4: No way. Sorry. NACK for this change.

Also the CDN which you are using is questionable as we don't have such things under control.

stangri: As far as CDN goes, are you talking about screenshots at jsdelivr? Please elaborate how is linking private repo on github better than using jsdelivr.

stangri: Yeah, please explain why?

dibdot: User tracking without user consent - not very trustworthy ...

stangri: > User tracking without user consent - not very trustworthy ...

By tracking you mean placing a cookie? My understanding was it's the same as counting hits on a screenshot, but I see that there is a hits.dwyl.com cooking in my browser, so I guess I was wrong on that.

Just to clarify, there's nothing against the cookie-less counter of visits, right? Like as if I count hits on the screenshot and display the counter without the cookie?

dibdot: In my opinion it's not a technical question regarding cookies, hidden pixels, analytics, tracking infrastructure etc., simply take the user privacy seriously and don't collect any data without consent. There is no valid reason for a piece of software within a opensource repo to ship any kind of data collector. 
In europe the GDPR (General Data Protection Regulation) has six legal bases outlined for processing personal data (https://gdpr.eu/article-6-how-to-process-personal-data-legally/) and none of them matched.

Bottom line: just don't do it - please.

stangri: > In my opinion it's not a technical question regarding cookies, hidden pixels, analytics, tracking infrastructure etc., simply take the user privacy seriously and don't collect any data without consent. There is no valid reason for a piece of software within a opensource repo to ship any kind of data collector. 

I totally agree. No personal data is collected.

> In europe the GDPR (General Data Protection Regulation) has six legal bases outlined for processing personal data (https://gdpr.eu/article-6-how-to-process-personal-data-legally/) and none of them matched.

Again, no personal data is collected.


> Bottom line: just don't do it - please.

I don't. Moreover this is the github-targeting counter, so almost all if not all its users are open source projects/developers. I've reached out to the author and they confirmed that even a cookie doesn't come from the counter but from the cloudlfare as with any cloudlfare-protected site.

So I'm yet again asking why the knee-jerk reaction to the counter which doesn't collect any personal information?

I'm not opposed to taking the readme to the personal repo if the hit counter is the reason it can't be accepted to packages repo, I just find it puzzling.

dibdot: Please do not close your eyes to the obvious: Look in the issue list of the referenced counter repo (marked as "enhancements"): "How to Track and Display Profile Views on GitHub", "Analytics for Startups", "Create a context for the user, useragent and repository database calls" and so on ... counters were popular in the 1990s, but were later replaced by other web traffic measures such as self-hosted scripts like Analog, and later on by remote systems that used JavaScript, like Google Analytics. Don't collect any data that you do not need!

You have two NACKs for this change.

stangri: > Don't collect any data that you do not need!

Just to be clear, you agree that no personal data is collected, you're just opposed to the counter itself?

stangri: README has been moved out of OpenWrt repo.

stangri: README has been moved out of OpenWrt repo.

stangri:
@aparcar @neheb can you help me out with the Makefile please? I'd like to incorporate the PKG_VERSION stamping into the init-file like [here](https://github.com/stangri/source.openwrt.melmac.net/blob/master/simple-adblock/Makefile#L34-L38) so I could create the test.sh as well, but if I define the `Build/Prepare` in this Makefile then the actual binary build fails.

Any suggestions?

neheb:
None.

neheb:
ping @hnyman . I don't know who can give merge access.

hnyman:
> README has been moved out of OpenWrt repo.

So the proposed solution to the tracker thing is to remove the whole README from here and move it to your own repo so that the tracker can be included there?  Sounds unfriendly. Why is the tracker so important to you?

> I'd like to incorporate the PKG_VERSION stamping into the init-file

Sorry. No real help for that. 
Manual editing of that file every time you update the version in Makefile, is possibly the easiest approach.

stangri:
> > README has been moved out of OpenWrt repo.
> 
> So the proposed solution to the tracker thing is to remove the whole README from here and move it to your own repo so that the tracker can be included there? Sounds unfriendly. Why is the tracker so important to you?

Hannu, there's no tracker. It's a dumb hits counter. Calling it a tracker or saying I'm tracking users without permission is insulting. I still have not heard a single solid argument against the hits counter btw. Just for comparison, below is a piece of code from OpenWrt home webpage:

```
<script language="javascript" type="text/javascript" src="/js/awstats_misc_tracker.js" ></script>
<noscript><img src="/js/awstats_misc_tracker.js?nojs=y" height="0" width="0" border="0" style="display: none"></noscript>
```

It's a bit hypocritical to accuse me of user tracking for using a highly visible dumb hits counter when there's a javascript code included in OpenWrt home webpage which is literally called tracker and is collecting stats. I understand BKPepe has a personal vendetta against me, having previously falsely accused me of stealing code, but I sincerely don't understand the reasons for pushback from anyone else.

Also, since I have to self-host screenshots I don't see a big deal about self-hosting README. Also, I certainly prefer the layout of the GH pages published README rather than linking to it within the repo. I would recommend something similar is set up by OpenWrt teams for maintainers who want to publish documentation.

> > I'd like to incorporate the PKG_VERSION stamping into the init-file
> 
> Sorry. No real help for that.
> Manual editing of that file every time you update the version in Makefile, is possibly the easiest approach.

Thanks, that option does make most sense, yet it completely escaped me. I'll update this PR sometime this week.

net/https-dns-proxy/files/https-dns-proxy.init
```diff
@@ -1,9 +1,17 @@
 #!/bin/sh /etc/rc.common
-# Copyright 2019 Stan Grishin (stangri@melmac.net)
+# Copyright 2019-2020 Stan Grishin (stangri@melmac.net)
 # shellcheck disable=SC2039
 
-export START=80
-export USE_PROCD=1
+readonly PKG_VERSION='2020-08-21-1'
```
aparcar, line 5: Please don't do this. If you forget to bump this here users go bananas. 

Please use something like `readonly PKG_VERSION='%PKG_VERSION%'` and add to `Package/https-dns-proxy/install` the following step `$(SED) "s;%PKG_VERSION%;$(PKG_VERSION);" $(1)/etc/init.d/https-dns-proxy`

net/https-dns-proxy/files/README.md
```diff
@@ -1,94 +1,3 @@
...
-|user|String|nobody|Local user to run instance under.|
-|group|String|nogroup|Local group to run instance under.|
-|use_http1|Boolean|0|If set to 1, use HTTP/1 on installations with broken/outdated ```curl``` package. Included for posterity reasons, you will most likely not ever need it on OpenWrt.|
-|verbosity|Integer|0|logging verbosity level. fatal = 0, error = 1, warning = 2, info = 3, debug = 4|
-|use_ipv6_resolvers_only|Boolean|0|If set to 1, Forces IPv6 DNS resolvers instead of IPv4|
-
-## Thanks
-
-This OpenWrt package wouldn't have been possible without [@aarond10](https://github.com/aarond10)'s [https-dns-proxy](https://github.com/aarond10/https_dns_proxy) and his active participation in the OpenWrt package itself. Special thanks to [@jow-](https://github.com/jow-) for general package/luci guidance.
+README has been moved to [https://docs.openwrt.melmac.net/https-dns-proxy/](https://docs.openwrt.melmac.net/https-dns-proxy/).
```
aparcar, line 3: I'd prefer you leave the README here (and format the table nicely via e.g. [remark](https://github.com/remarkjs/remark)), lose the ~tracker~ ~counter~ badge and use the [*official* stats](https://downloads.openwrt.org/stats/awstats.downloads.openwrt.org.allextra2.html).

I'll ask the download maintainers if they can beautify the stats a bit, ideally merging stats for single packages, instead of (just) showing it per architecture, which makes a *total downloads* count a bit tedious. Seems like you got multiple hundreds of installs :rocket:. 

I'll also ask about the JavaScript snipped you posted, from my understanding it is not really required to track which files are downloaded or Wiki pages are visited.

If you wanted the counter to increase *visibility* of your project, we could think of a better way like *featured LuCI apps* via blog posts or forum entries.

aparcar (changes requested):

net/https-dns-proxy/files/https-dns-proxy.init
stangri: Done. That was what I was doing in `build/prepare` before.

net/https-dns-proxy/files/README.md
stangri: > I'd prefer you leave the README here (and format the table nicely via e.g. [remark](https://github.com/remarkjs/remark)), lose the ~tracker~ ~counter~ badge and use the [_official_ stats](https://downloads.openwrt.org/stats/awstats.downloads.openwrt.org.allextra2.html).

Could you please elaborate on "formatting the table nicely"? 

For keeping README here -- what's the plan for screenshots? I would strongly prefer to keep the screenshots hosted together with the README (not an option for the source code repo as far as I understand) and, having experienced it now, I would strongly prefer an option to have the README published thru GitHub pages (or some such, after the move), so it opens full height/width, without GitHub header/side navigation. 

> I'll also ask about the JavaScript snipped you posted, from my understanding it is not really required to track which files are downloaded or Wiki pages are visited.

As far as I remember the awstats is a web-site log analyzer, but back when I used it, they didn't have the javascript helper, so I'm puzzled what's the reason to use one now. It's simply weird that if there's a crusade against hits counter, there's tracker javascript code on the home page which no one has objected to.

aparcar: > Could you please elaborate on "formatting the table nicely"?

If you scroll down a bit you see a "nicely formatted" table in the [contributing document](https://raw.githubusercontent.com/openwrt/packages/master/CONTRIBUTING.md)

> For keeping README here -- what's the plan for screenshots? I would strongly prefer to keep the screenshots hosted together with the README (not an option for the source code repo as far as I understand) and, having experienced it now, I would strongly prefer an option to have the README published thru GitHub pages (or some such, after the move), so it opens full height/width, without GitHub header/side navigation.

I think that's a fair point, screenshots within `packages.git` make no sense. The discussion about the counter seems a waste of time to me. Hosting anything on GitHub involves much more tracking than just a hit counter, so if anyone is concerned about luring people into being tracked, we shouldn't offer `packages.git` on GitHub at all. On the other hand what do you get from this counter? How many people read your documentation? You could host it yourself and check Nginx access logs. Or, as stated before, you look at the official stats and actually get a idea of how many people actually install the package (minus all the mirror pulls).

> I'll also ask about the JavaScript snipped you posted, from my understanding it is not really required to track which files are downloaded or Wiki pages are visited.

I think it's mostly legacy which where added in good intentions some long time ago. I pinged Thomas and kindly asked for removal.

stangri: Thanks, installed remark processor and reformatted all the READMEs.

Hits on READMEs approximate how many people install packages from my own repo (where I usually post test versions before submitting PRs). Before I switched to a docs repo there were thousands on README hits for vpn-policy-routing for example.

For the issue at hand -- until there's a solution to host and publish READMEs and the screenshots on OpenWrt infrastructure, what's the proposed solution for now?

aparcar: I think it's cleaner to have only a single README.md in your own repo. If you feel like adding a counter, that's your choice - I usually block/delete cookies from such sites, if other people happen to do the same your results aren't very accurate, but that's up to you.

@neheb @BKPepe @hnyman any objections?

stangri: @aparcar I feel that it's been enough time for anyone to voice any objections. If there's no separate OpenWrt repo for READMEs which can also host screenshots, I'd prefer to keep READMEs and screenshots together at mine. Please advise if there are any other issues preventing this (and https://github.com/openwrt/packages/pull/13513 and https://github.com/openwrt/packages/pull/13514) from merging. Thanks!

PR #13534, opened 2020-09-30
apk: add package

aparcar:
The Alpine Package Keeper is the default package manager for Alpine
Linux. They manage their own repositoriy with thousends of packages and
as musl is used for compilation, their binaries can directly run on
OpenWrt.


aparcar:
This has currently some limitations:

- [ ] The folder `/var/cache/apk` needs to be created during runtime
- [ ] Limit package installation to the following ARCHs:
    aarch64, armhf, armv7, mips64, ppc64le, s390x, x86, x86_64

aparcar:
@neheb 

Do you have an idea for the CI error on arc?

```
 ln -sf libapk.so.3.12.0 src/libapk.so
  CC      src/atom.o
In file included from src/apk_hash.h:15,
                 from src/apk_atom.h:13,
                 from src/atom.c:10:
src/apk_defines.h:125:2: error: #error APK_DEFAULT_ARCH not detected for this architecture
 #error APK_DEFAULT_ARCH not detected for this architecture
  ^~~~~
```

neheb:
I mean, musl doesn't even support ARC, let alone alpine linux.

aparcar:
It's not about using Alpine repos, that's just a proof of concept.

utils/apk/Makefile
```diff
@@ -0,0 +1,75 @@
...
+PKG_NAME:=apk
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://gitlab.alpinelinux.org/alpine/apk-tools.git
+PKG_SOURCE_DATE:=2020-08-25
+PKG_SOURCE_VERSION:=v2.12.0_rc2
+PKG_MIRROR_HASH:=4dbd95a0dca5d6e17b314af23a5c9cbe5828bfba2b6325138887665d06082585
+
+PKG_LICENSE:=GPL-2.0
```
neheb, line 12: -only or -or-later.

utils/apk/Makefile
```diff
@@ -0,0 +1,75 @@
...
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_MAINTAINER:=Paul Spooren <mail@aparcar.org>
+
+PKG_INSTALL:=1
+
```
neheb, line 18: I would add PKG_BUILD_PARALLEL.

BKPepe:
Ehhh... I am not sure if this belongs to the packages repository. People would be confused. We have opkg and we should stick to it. This adds just overhead and more work. Feel free to convince your colleagues OpenWrt main repo, but I don't think we should merge it.

aparcar:
@BKPepe If you don't mind I'll leave this open for now and eventually propose it to *my colleagues* over at openwrt.git. It's posted here just for Makefile feedback (thanks @neheb ) or people curious of testing it. 

In my opinion we should replace `opkg` long term with `apk` as it's a more stable and feature rich package manger. And different from `opkg`, it's actually still actively developed.

However this is a bigger topic which needs to be cooked slow. If an open draft PR bums you feel free to close it.

mayli:
@aparcar are you proposing an alpine based openwrt?

aparcar:
> @aparcar are you proposing an alpine based openwrt?

No it's just a coincidence that two distributions chose to package, maintain and compile some ~6000 packages which run on each others distribution. As I tried to formulate above, this is just a "proof of concept" so people can test APK. A package manager without any package feeds seems so boring.

Anyway, overall the APK manager is much more actively developed and I (and others) therefore like the idea of replacing OPKG in some distant future and have an *APK based OpenWrt*.

neheb:
I don't know if that's accurate. Alpine uses musl 1.2 while OpenWrt does not. Musl 1.2 has 64-bit time_t for 32-bit platforms.

APK OpenWrt sounds nice. Given the amount of NIH stuff in OpenWrt, that's where it stops I think.

aparcar:
> APK OpenWrt sounds nice. Given the amount of NIH stuff in OpenWrt, that's where it stops I think.

OPKG < APK. We continue to use our own compiled packages, we'd just [package them differently](https://github.com/aparcar/openwrt/commit/ba51f0b038731c7b69b2197450abe234ad9cada9).

aparcar:
@blogic ping

blogic:
LGTM

utils/apk/Makefile
```diff
@@ -0,0 +1,76 @@
...
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=apk-tools-v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://gitlab.alpinelinux.org/alpine/apk-tools/-/archive/v$(PKG_VERSION)
+PKG_HASH:=b191dbd5019e8933b78b66779412e901e9b88fb12c460a22995e342b3efe83eb
+
+PKG_LICENSE:=GPL-2.0-only
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_MAINTAINER:=Paul Spooren <mail@aparcar.org>
```
BKPepe, line 14:
```suggestion
PKG_MAINTAINER:=Paul Spooren <mail@aparcar.org>
PKG_LICENSE:=GPL-2.0-only
PKG_LICENSE_FILES:=LICENSE
```

aparcar: Added.

utils/apk/Makefile
```diff
@@ -0,0 +1,75 @@
...
+
+PKG_SOURCE:=apk-tools-v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://gitlab.alpinelinux.org/alpine/apk-tools/-/archive/v$(PKG_VERSION)
+PKG_HASH:=b191dbd5019e8933b78b66779412e901e9b88fb12c460a22995e342b3efe83eb
+
+PKG_MAINTAINER:=Paul Spooren <mail@aparcar.org>
+PKG_LICENSE:=GPL-2.0-only
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/apk-tools-v$(PKG_VERSION)
```
neheb, line 15: I would move this below PKG_HASH.

utils/apk/Makefile
```diff
@@ -0,0 +1,75 @@
...
+MAKE_FLAGS += LUA=no
+
+define Package/apk/install
+	$(INSTALL_DIR) $(1)/lib/apk/db
+
+	$(INSTALL_DIR) $(1)/bin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/sbin/apk $(1)/bin/apk
+
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/lib/* $(1)/usr/lib/
```
neheb, line 52: I assume this library is used by nothing else. Is it easy to statically link?

aparcar: Good call, I guess I could add a make var

https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/master/Makefile#L46

utils/apk/Makefile
aparcar: Done, thanks.

utils/apk/Makefile
aparcar: I looked into statically building it but it seem to pull in openssl as well. I'd avoid that for now and just ship it as an external lib.

neheb:
There's a new version.

Issue #11152, opened 2020-01-28
ddns-scripts_no-ip_com: add no-ip.com support for ipv6


EmbeddedG:
Maintainer: @\chris5560
Environment: OpenWrt master

Description:
no-ip.com doesn't appear in the "DDNS Service provider [IPv6]" dropdown. I opened this issue after #8095 was closed because of wrong package reference in the subject.
I tested no-ip.com ipv6 update with a separate custom configuration for v6 by manually providing the update URL and it works. no-ip.com supports updating both v4 and v6 addresses together, or updating separately by using `myipv6` parameter for v6 instead of `myip`. Sending together is the method preferred by no-ip.com and they say that `myipv6` is for compatibility with clients that doesn't support updating both v4 and v6 addresses together ([see here](https://www.noip.com/integrate/request)).

hnyman:
implemented by 33d6dc402 in late 2020...

PR #13561, opened 2020-10-02
ddns-scripts: add ipv6 capability for no-ip.com

disturbednny:
Maintainer: me / @feckert (find it by checking history of the package Makefile)
Compile tested: BCM27xx Raspberry Pi 4B openwrt - snapshot
Run tested: BCM27xx Raspberry Pi 4B openwrt - snapshot

Description: Add in the ability to select no-ip.com as an IPV6 ddns provider. Tested in my home with snapshot build on Raspberry Pi 4B.

### DIFF
--- a/net/ddns-scripts/files/usr/lib/ddns/update_no-ip_com.sh
+++ b/net/ddns-scripts/files/usr/lib/ddns/update_no-ip_com.sh
@@ -10,8 +10,8 @@
 # so we send a dummy (localhost) and a seconds later we send the correct IP addr
 #
 local __DUMMY
-local __UPDURL6="http://[USERNAME]:[PASSWORD]@dynupdate6.no-ip.com/nic/update?hostname=[DOMAIN]&myip=[IP]"
-local __UPDURL="http://[USERNAME]:[PASSWORD]@dynupdate.no-ip.com/nic/update?hostname=[DOMAIN]&myip=[IP]"
+local __UPDURL6="http://[USERNAME]:[PASSWORD]@dynupdate6.noip.com/nic/update?hostname=[DOMAIN]&myip=[IP]"
+local __UPDURL="http://[USERNAME]:[PASSWORD]@dynupdate.noip.com/nic/update?hostname=[DOMAIN]&myip=[IP]"
 # inside url we need username and password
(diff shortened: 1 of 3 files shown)


net/ddns-scripts/files/usr/lib/ddns/update_no-ip_com.sh
```diff
@@ -10,13 +10,15 @@
 # so we send a dummy (localhost) and a seconds later we send the correct IP addr
 #
 local __DUMMY
+local __UPDURL6="http://[USERNAME]:[PASSWORD]@dynupdate6.no-ip.com/nic/update?hostname=[DOMAIN]&myip=[IP]"
```
pkgadd, line 13: Using uclient-fetch (2020-06-17-c6609861-1)/ libustream-openssl(20200215 2020-03-13-5e1bc342-1) of current master with dynupdate6.no-ip.com fails for me with error 5 (Server hostname does not match SSL certificate).

Switching this dynupdate.no-ip.com seems to be successful.

disturbednny: dynupdate.noip.com is ipv4 only, and dynupdate6.noip.com is ipv6 only, just updated PR with the changes.

pkgadd: That seems to work, thanks.

```
Tested-by: Stefan Lippers-Hollmann <s.l-h@gmx.de>
```

pkgadd:
Btw. this pr would fix #11152 (which requested IPv6 support for no-ip).

feckert:
@disturbednny I have not yet looked at it closely. But why does it say **http**! Can't we use **https**?

disturbednny:
@feckert there's a sed further down in the script that replaces http with https if the use https option is selected in setting up the service

feckert:
Staged the commits and added the following Testing-by to the commit.
`Tested-by: Stefan Lippers-Hollmann <s.l-h@gmx.de>`

See https://github.com/openwrt/packages/commit/33d6dc402743f9598b9f50bbc3aa7495acd49c4a

PR #13562, opened 2020-10-03
rsync: disable option for bundled zlib, add option for zstd

antonlacon:
Maintainer: @mstorchak 
Compile tested: arm/mvebu, openwrt HEAD as of 2020-10-02
Run tested: arm/mvebu with openwrt acting as server transferring to local clients, and as client retrieving from remote servers.

Description:

This removes the option to use rsync's bundled zlib. I added it ~6 years ago to give compatibility with even older versions of rsync. Said old versions should hopefully no longer be in use, so remove the option.

This adds the option to use zstd for transfer compression.

### DIFF
--- a/net/rsync/Config.in
+++ b/net/rsync/Config.in
@@ -12,4 +12,9 @@ if PACKAGE_rsync
 		default y if USE_FS_ACL_ATTR
 		default n
 
+	config RSYNC_zstd
+		bool
+		prompt "Enable zstd stream compression"
+		default n
+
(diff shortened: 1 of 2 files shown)


neheb:
should bump PKG_RELEASE I think.

antonlacon:
Thanks for the reminder. It's added in the zstd commit.

mstorchak:
LGTM, thank you, @antonlacon

PR #13540, opened 2020-10-01
mariadb: Add basic test

miska:
Maintainer: me
Compile tested: NA
Run tested: NA

Description:
Try to add basic test for MariaDB, mainly to get a feel how CI works. Test to be enhanced later, locally script works, but want to ensure that I got how the whole system works.

PR #13564, opened 2020-10-03
CI: Add some helper functions

miska:
Maintainer: @aparcar 
Compile tested: NA
Run tested: NA

Description:
Add some minor helper functions that could be used to produce nicer output during CI runs. How it works can be seen in #13540

### DIFF
--- a/.github/workflows/ci_helpers.sh
+++ b/.github/workflows/ci_helpers.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+color_out() {
+	printf "\e[0;$1m$PKG_NAME: %s\e[0;0m\n" "$2"
+}
+
+success() {
+	color_out 32 "$1"
+}
+
+info() {
+	color_out 36 "$1"
+}
+
+err() {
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_helpers.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+# shellcheck disable=SC2039
+color_out() {
+	echo -en "\033[0;$1m"
```
neheb, line 5: Why not use printf?

miska: Out of habit. Replaced with printf.

neheb:
Looks good.

aparcar:
Any reason not to include the helper functions directly in the entrypoint?

miska:
Individual test are executed, not sourced, so the helper functions wouldn't be available. I could change it to sourcing, but then individual test might influence the environment for the following tests.

aparcar:
So individual tests source the CI variable? I guess I'm fine with it

PR #13571, opened 2020-10-05
spice: convert to meson

neheb:
spice: convert to meson

Faster compilation.

Removed autotools patches.

Backported meson patch to fix compilation and pkgconfig file.

Backported meson patch to avoid regenerating included files. Also made
another patch to that extent.

Added patch to avoid building tools and docs since they're not needed.

Before:

time make package/spice/compile -j 12
Executed in   62.85 secs   fish           external
   usr time   56.45 secs  276.00 micros   56.45 secs
   sys time    8.06 secs   46.00 micros    8.06 secs

Before + PKG_BUILD_PARALLEL:

Executed in   45.40 secs   fish           external
   usr time   63.08 secs  253.00 micros   63.08 secs
   sys time    8.57 secs   44.00 micros    8.57 secs

After:

time make package/spice/compile -j 12
Executed in   16.54 secs   fish           external
   usr time   41.29 secs  266.00 micros   41.29 secs
   sys time    4.76 secs   45.00 micros    4.76 secs

Maintainer: @yousong 
Compile tested: ath79

Those After numbers are now bogus. They should be even faster as I removed test, tool, and doc building.


libs/spice/patches/030-include-generated-code.patch
```diff
@@ -0,0 +1,22 @@
+--- a/subprojects/spice-common/common/meson.build
++++ b/subprojects/spice-common/common/meson.build
```
yousong, line 2: Please add patch like it's exported with git format-patch, with subject title and message body.  Like they are to be upstreamed (or serve as a indication even if they dont) and reviewed by others.

libs/spice/Makefile
```diff
@@ -32,6 +32,30 @@ define Package/libspice-server
   DEPENDS:=+glib2 +libjpeg +libopenssl +pixman +zlib
 endef
 
+MESON_ARGS += \
+	-Dgstreamer=no \
+	-Dlz4=false \
+	-Dsasl=false \
+	-Dcelt051=disabled \
+	-Dopus=disabled \
```
yousong, line 42: I'd like to at least keep the original comment singling out these two as related "audio codec"

libs/spice/Makefile
```diff
@@ -56,16 +58,18 @@ MESON_ARGS += \
...
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/pkgconfig/spice-server.pc $(1)/usr/lib/pkgconfig
 endef
 
-define Build/InstallDev
+define Package/libspice-server/install
 	$(INSTALL_DIR) $(1)/usr/lib
-	$(INSTALL_DIR) $(1)/usr/include
-	$(CP) $(PKG_INSTALL_DIR)/usr/include/* $(1)/usr/include
-	$(CP) $(PKG_INSTALL_DIR)/usr/lib/* $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libspice-server.so* $(1)/usr/lib
```
yousong, line 72: To be honest, I am not a fan of commit like this.  A bit messy and takes more time to check if there are unexpected changes.  Try not stuffing in another subject in the message body please.

yousong (changes requested):

libs/spice/patches/030-include-generated-code.patch
neheb: Done.

libs/spice/Makefile
neheb: Done.

libs/spice/Makefile
neheb: Split off to a separate commit.

libs/spice/Makefile
```diff
@@ -32,30 +34,45 @@ define Package/libspice-server
...
+	-Dgstreamer=no \
+	-Dlz4=false \
+	-Dsasl=false \
+	-Dsmartcard=disabled \
+	-Dalignment-checks=false \
+	-Dextra-checks=false \
+	-Dstatistics=false \
+	-Dmanual=false \
+	-Dinstrumentation=no \
+	\
```
karlp, line 52: Why this fragile form? Why not just MESON_ARGS += again? (personally, I'd do that for _every_ arg, so you don't have diffs based on the last \ being present or not when sorting/adding new options, but at least, putting in artificial line breaks seems pointless

neheb: This is to separate the main arguments from the subproject ones.

yousong: I second the action to make it another `MESON_ARGS +=` for the subproject.

As for the diff generation on presence of last backslash, usually I always keep a backslash at the end.  It's convenient when moving them around, adding or removing some of them.  `utils/qemu/Makefile` as an example.

neheb: Done.

yousong (approved):

PR #13572, opened 2020-10-05
ddns-scripts: re-add no-ip.com to list of supported providers

Borromini:
@feckert I'm not sure if this list is purely descriptive or if it's parsed by e.g. the LuCI frontend, but No-IP.com got dropped from that list in the overhaul. This adds it back.


feckert:
Special providers could not be installed via download.
They need a special script to work proper.
So they must not be included into the list file.

PR #13578, opened 2020-10-05
bind: update to version 9.16.7

BKPepe:
Maintainer: @nmeyerhans 
Run tested: Turris Omnia, mvebu (cortex-a9), OpenWrt master - _Dig works, DNS resolving via named works, too._

Description:
Changelog 9.16.7: https://downloads.isc.org/isc/bind9/9.16.7/doc/arm/html/notes.html#notes-for-bind-9-16-7

We should include this into OpenWrt 19.07 as well.

### DIFF
--- a/net/bind/Makefile
+++ b/net/bind/Makefile
@@ -9,7 +9,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bind
-PKG_VERSION:=9.16.6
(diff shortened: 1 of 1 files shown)


BKPepe:
@nmeyerhans hopefully, you don't mind I'm merging this. There are just two changes (version bump, checksum).

PR #13582, opened 2020-10-06
[lighttpd] Pull in mod-authn_file with mod-auth

jan-kardell:
If lighttpd loads mod-auth, it also automatically tries to load
mod-authn_file, and fails if it's not available. That is a compatibility
feature of lighttpd after the funtionality was split into modules.

Signed-off-by: Jan Kardell <jan.kardell@telliq.com>

Maintainer: @flyn-org
Compile tested: (omap, 19.07)
Run tested: (omap, 19.07)

Description:

### DIFF
--- a/net/lighttpd/Makefile
+++ b/net/lighttpd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lighttpd
 PKG_VERSION:=1.4.55
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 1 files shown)


neheb:
:). I don't know if this is related to my meson conversion.

Ping @BKPepe . I know you guys use lighttpd in TurrisOS.

jan-kardell:
@neheb This has nothing to do with the meson conversion. It's because of the "compatibility feature" that is in lighttpd, and it has been like this for quite a while. I guess no openwrt dev spotted this lighty feature before, as it's not an obvious thing hidden in the lighty C code.

neheb:
Hmm OK.

PR #13593, opened 2020-10-07
packr: New package

jefferyto:
Maintainer to be: me
Compile tested: x86-64/armvirt-32/armvirt-64/malta-be, 2020-10-01 snapshot sdk
Run tested: x86-64/armvirt-32/armvirt-64/malta-be, 2020-10-01 snapshot

Description:
From the README:

Packr is a simple solution for bundling static assets inside of Go binaries. Most importantly it does it in a way that is friendly to developers while they are developing.

### DIFF
--- a/devel/packr/Makefile
+++ b/devel/packr/Makefile
@@ -1,0 +1,55 @@
+#
+# Copyright (C) 2020 Jeffery To
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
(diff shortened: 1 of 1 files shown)


jefferyto:
This is mainly added to support #12408.

v2 of packr is available, but v2 uses a [backward-incompatible file format](https://github.com/gobuffalo/packr/blob/c6a5045204dacffbe11d9f6c68485ae7e7fe3d19/v2/README.md#new-file-format-faqs). packr2 can read/generate v1 files but this support may be removed in the future.

PR #13605, opened 2020-10-08
prometheus-node-exporter-lua: hostad ubus stats

PolynomialDivision:
There is already the hostapd_stations exporter, which uses hostapd-utils (more precisely hostapd-cli) to get client statistics. However, the ubus interface is permanently integrated under hostapd in OpenWrt. So this exporter needs one dependency less.

For now it exports mainly the rrm statistics. Many people are interested in what your device supports. The exporter provides information about the radio-resource-managment extensions.

Maintainer: @champtar
Run tested: r14549+2-036221ce5a89

### DIFF
--- a/mail/mutt/Makefile
+++ b/mail/mutt/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mutt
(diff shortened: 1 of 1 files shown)


PolynomialDivision:
Whups.

PR #13609, opened 2020-10-08
transmission: create openssl and mbedtls variant for transmission-web

ja-pa:
Maintainer: @neheb 
Run tested: N/A

Description:
The main point of this PR is to try to open discussion about transmission-web package dependency.
I think that it would be better to have 2 transmission-web package variants (mbedtls and openssl) even if transmission-web is a generic package.

After installing the transmission-web package user has to decide which daemon variant he/she wants to use and then install it.
With transmission-web (openssl or mbedtls variant) the decision is made in the first install and other steps are maintained by the package manager.

### DIFF
--- a/net/transmission/Makefile
+++ b/net/transmission/Makefile
@@ -47,6 +47,7 @@ define Package/transmission-daemon-openssl
   TITLE+= (with OpenSSL)
   DEPENDS+=+libopenssl
   VARIANT:=openssl
+  PROVIDES:=transmission-daemon
 endef
 
 define Package/transmission-daemon-mbedtls
@@ -54,6 +55,7 @@ define Package/transmission-daemon-mbedtls
   TITLE+= (with mbed TLS)
   DEPENDS+=+libmbedtls
   VARIANT:=mbedtls
+  PROVIDES:=transmission-daemon
 endef
 
 define Package/transmission-cli-openssl
@@ -87,10 +89,9 @@ endef
 define Package/transmission-web
   $(call Package/transmission/template)
   TITLE+= (webinterface)
-  DEPENDS:=@(PACKAGE_transmission-daemon-openssl||PACKAGE_transmission-daemon-mbedtls)
+  DEPENDS:=transmission-daemon
   PKGARCH:=all
 endef
-
 
 define Package/transmission-daemon/Default/description
  Transmission is a simple BitTorrent client.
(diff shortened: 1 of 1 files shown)


neheb:
I'm ambivalent on this.

Transmission itself depends on libcurl which itself depends on a TLS library. I've previously thought about getting rid of transmission variants and defaulting to whatever libcurl comes with. That way there's no need for two TLS libraries installed at the same time.

Cynerd:
I am sorry to my colleague but this is wrong.

It makes no sense to create variants for web just because we have variants of daemon. The correct approach is to make daemon variants provide common (virtual) package `transmission-daemon` and make `transmission-web` depend on that.

@neheb it is also I think not possible to remove transmission-daemon variants as if you look in to Makefile they are linking to SSL library directly (see configuration options there). That would be possible only if those libraries would be ABI compatible.

neheb:
It's possible to remove the variants by using what libcurl uses. See uacme for an example.

Cynerd:
> It's possible to remove the variants by using what libcurl uses. See uacme for an example.

I see the point is that we do not have curl variants so there is no need to have variants on anything else that uses libcurl as it pulls curl SSL library to system anyway. I missed that.

neheb:
ping @mstorchak would you be in favor of removing the SSL variants and using the one selected by curl?

Note that this would default to wolfssl. I don't know if transmission is crypto sensitive.

neheb:
I looked at that last commit. Seems incomplete. curl also supports GnuTLS and the lack of SSL. Those need to be handled as transmission does not support GnuTLS. See https://github.com/openwrt/packages/blob/master/net/uacme/Makefile as an example.

mstorchak:
Is transmission going to accept libcurl's choice of crypto library? If so, works for me.

As for the default wolfssl, I don't have any opinion - I'm using openssl variant atm.

neheb:
> 
> 
> Is transmission going to accept libcurl's choice of crypto library? If so, works for me.
> 
> As for the default wolfssl, I don't have any opinion - I'm using openssl variant atm.

Yeah. The only exception so far is GnuTLS. Maybe I can convince upstream to add support.

net/transmission/Makefile
```diff
@@ -42,20 +44,28 @@ define Package/transmission/template
...
-define Package/transmission-daemon
+
+define Package/transmission/template-depends
   $(call Package/transmission/template)
+  DEPENDS+=\
+   +LIBCURL_OPENSSL:libopenssl \
+   +LIBCURL_MBEDTLS:libmbedtls \
+   +LIBCURL_GNUTLS:libmbedtls \
+   +LIBCURL_NOSSL:libmbedtls \
+   +LIBCURL_WOLFSSL:libmbedtls
```
neheb, line 55: Eh no, Transmission supports wolfssl.

ja-pa:
Closing this PR, because https://github.com/openwrt/packages/pull/13723 fix it.

PR #11775, opened 2020-04-06
wget: make the ssl variant provides wget-ssl 

yousong:
Maintainer: wget @tripolar , acme @tohojo
Compile tested: n/a
Run tested: n/a

Description:

wget: make the ssl variant provides wget-ssl

This is to address the need for specifying dependency on a wget
implementation with ssl support.

Now we have a game of names for opkg

 1. uclient-fetch: minimal version by openwrt project
 2. wget-nossl: gnu wget w/o ssl support
 3. wget-ssl: for the moment since this commit, gnu wget w/ ssl support
 4. wget: uclient-fetch, wget-nossl, or wget-ssl
 5. gnu-wget: wget-nossl or wget-ssl

By the time we provide some dummy package like uclient-fetch-ssl and
make it also provide wget-ssl, I guess by then we will also need
gnu-wget-ssl...

Ref: https://github.com/openwrt/packages/issues/11534
Ref: https://github.com/openwrt/packages/pull/9941
Signed-off-by: Yousong Zhou <yszhou4tech@gmail.com>

PR #13613, opened 2020-10-09
wget: rename wget to wget-ssl

Cynerd:
Maintainer: @tripolar 
Compile tested:
Run tested:

Description:
The idea behind this is to prevent confusion between "virtual" package
wget and real one. Wget is provided by not just wget packages but also
by uclient-fetch so technically it is better to threat wget as virtual
package.

In reality I need to cover this case: https://forum.turris.cz/t/wget-busybox-vs-package-wget/2547. Turris Updater-ng minimizes amount of installed packages (compared to opkg that does nothing like that) and that makes confusion between users. `opkg instrall wget` installs real wget package. Updater has then request to install `wget` and `uclient-fetch` (because of `opkg` package). Because `uclient-fetch` provides `wget` it removes `wget` package as that is already satisfied by `uclient-fetch`. It is hard to explain users that they have to install virtual package `gnu-wget` to get real wget as "virtual" packages are not reported by opkg and thus by LuCI. This makes it so listing of existing packages lists `wget-ssl` that is unique name of that package that can be satisfied by only installing real wget.

And yeh the reason why it is this way right now is partially my fault thanks to commit 75f2be7d509e802f2d946d732e7a7539c2b27d2e.

### DIFF
--- a/net/wget/Makefile
+++ b/net/wget/Makefile
@@ -29,7 +29,7 @@ define Package/wget/Default
   SUBMENU:=File Transfer
   TITLE:=Non-interactive network downloader
   URL:=https://www.gnu.org/software/wget/index.html
-  PROVIDES:=gnu-wget
+  PROVIDES:=gnu-wget wget
 endef
 
 define Package/wget/Default/description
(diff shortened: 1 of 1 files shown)


neheb:
ping @yousong

yousong:
We will need a wiki or README.md for the names ;)  This can serve as a start, https://github.com/openwrt/packages/pull/11775#issue-399347101

neheb:
Can you rebase?

Cynerd:
Rebased...

neheb:
Meh OK.

PR #13615, opened 2020-10-09
chrony: update to 4.0

mlichvar:
Maintainer: me
Compile tested: mips, Archer C2 v3, OpenWrt 19.07
Run tested: mips, Archer C2 v3, OpenWrt 19.07, tested NTP client

Description: Update chrony to 4.0 and add a config option to enable NTS support, which depends on gnutls.

### DIFF
--- a/net/chrony/Makefile
+++ b/net/chrony/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=chrony
-PKG_VERSION:=3.5
+PKG_VERSION:=4.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://download.tuxfamily.org/chrony/
-PKG_HASH:=4e02795b1260a4ec51e6ace84149036305cc9fc340e65edb9f8452aa611339b5
+PKG_HASH:=be27ea14c55e7a4434b2fa51d53018c7051c42fa6a3198c9aa6a1658bae0c625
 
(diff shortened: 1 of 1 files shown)


net/chrony/Makefile
```diff
@@ -52,7 +63,7 @@ CONFIGURE_ARGS+= \
 	--sysconfdir=/etc/chrony \
 	--prefix=/usr \
 	--chronyrundir=/var/run/chrony \
-	--disable-nts \
+	$(if $(CONFIG_CHRONY_NTS),,--disable-nts) \
```
neheb, line 66: --$(if $(CONFIG_CHRONY_NTS),en,dis)able-nts \ is better I think.

mlichvar: In this case it's not an autoconf-generated configure script and it doesn't have an `--enable-nts` option. The detection of gnutls and nettle, which are required for NTS, can be just disabled.

neheb: OK.

BKPepe:
We should update it also in OpenWrt 19.07, right?

mlichvar:
As I understand it, this a major upgrade, which is not allowed in stable branches (per the contributing guidelines).

BKPepe:
In this case, I am in favor of doing an exception here as you tested this release on OpenWrt 19.07 and there is still version 3.5 of chrony, which is affected by security vulnerability (CVE-2020-14367), which was fixed in chrony 3.5.1. 

It is up to you what you prefer if you want to bump it there to this version than using the latest version.

mlichvar:
Ok, submitted PR #13624 to update chrony in 19.07 to 3.5.1.

PR #13617, opened 2020-10-09
sipcalc: add IPv4/IPv6 calculator

PolynomialDivision:
Sipcalc is an advanced ip calculator supporting both IPv4 and IPv6.

Maintainer: me
Compile tested: e5f7a9889cad5df86ff65689b1f0a68637e672bd
Run tested: IP40xx, r14507+2-14247c8

### DIFF
--- a/utils/sipcalc/Makefile
+++ b/utils/sipcalc/Makefile
@@ -1,0 +1,41 @@
+#
+# This software is licensed under the Public Domain.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=sipcalc
+PKG_SOURCE_DATE:=2014-10-24
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/sii/sipcalc
+PKG_SOURCE_VERSION:=9b97dbca1a0ccaaeb322b8f0a4100dc234cfcc8c
+PKG_MIRROR_HASH:=fac4a3e967055bac734bda86c34d318c9bf8dc17b0feb24f556b1d5af57d2896
+
+PKG_MAINTAINER:=Nick Hainke <vincent@systemli.org>
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=COPYING
+
+PKG_BUILD_PARALLEL:=1
+PKG_FIXUP:=autoreconf
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/sipcalc
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=IPv4/IPv6 Calculator
+  URL:=https://github.com/sii/sipcalc
+endef
+
+define Package/sipcalc/description
+   Sipcalc is an advanced ip calculator supporting both IPv4 and IPv6. 
+endef
+
+define Package/sipcalc/install
+	$(INSTALL_DIR) $(1)/usr/sbin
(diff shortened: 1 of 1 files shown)


BKPepe:
I'm just wondering why we need generate-ipv6-address package, which was added recently by you and now this package, which does the same for IPv6, but it is different that there is also support for IPv4.

PolynomialDivision:
> I'm just wondering why we need generate-ipv6-address package, which was added recently by you and now this package, which does the same for IPv6, but it is different that there is also support for IPv4.

I wrote a script based on this tool (still work in progress):
https://github.com/Freifunk-Spalter/auto-prefix-configuration-with-source-specifc-routing/blob/master/files/new-prefixes

The generate ipv6 address tool, was just a quick helper to generate in a quick way a valid ipv6 address. I can remove it again, if it is block sipcalc? I thought there would be no problem with a wide range of tools. I'm sorry that I did not have the time to fix your ipv6-generator bug. :/

BKPepe:
No, no. I was just asking, why there will two packages for almost the same. Both seems dead in upstream.

PolynomialDivision:
> No, no. I was just asking, why there will two packages for almost the same. Both seems dead in upstream.

I know. :/ I found this tool on a recent stackoverflow entry. xD Even my Manjaro-Linux Distro has this tool included. Maybe it is simply finished.

I searched for another tool, that allows easy expanding and compression of ipv6 prefixes and to give all subnets, but no tool was able to do that? So actually I would have to write some own script or just use this small c programm. ;) Actually, both tools have no dependencies and are just for parsing/generating ipv6 addresses. Maybe it is just enough.

Do you have an idea how to distribute prefixes in the layer 3 meshnetwork? The script simply takes the source specific routes and all other babel announced routes and randomly generates the prefix using free subnetworks and assigns itself at the end.

jow-:
Note that there also is https://github.com/openwrt/openwrt/blob/master/package/network/utils/owipcalc/src/owipcalc.c in the tree since quite a few years. It might or might not do what you need.

PolynomialDivision:
> Note that there also is https://github.com/openwrt/openwrt/blob/master/package/network/utils/owipcalc/src/owipcalc.c in the tree since quite a few years. It might or might not do what you need.

Maybe we can merge functionality of sipcalc and your package? So we have a working upstream version that is maintained?

jow-:
How are you using sipcalc right now?

PolynomialDivision:
> How are you using sipcalc right now?

https://github.com/Freifunk-Spalter/auto-prefix-configuration-with-source-specifc-routing/blob/master/files/new-prefixes

1. Give me prefix of all default gateway announces by babel
2. Calculate all subnetworks of length 64
3. Give all Prefixes `/64` that are announced by babel
4. Check if we have already a prefix assigned from the default route; then break
5. Iterate over the subprefixes list until we found a prefix that is not already announced by babel

We need more reandomizatino and collision detection... That is just a script I wrote in the night...^^

PolynomialDivision:
FIY: The script works! ISP changed IPv6 prefix and Fritz!Box renewed it. A quick call of that scripts give u a new prefix that u have to assign manually. I will automatize it.

PolynomialDivision:
Could we maybe merge the plugin? I would like to package my script (now with prefix network auto configuration) and add a cronjob. The last thing I add is to remove none gateway prefixes again.
https://github.com/Freifunk-Spalter/auto-prefix-configuration-with-source-specifc-routing/tree/master/files

So people of my freifunk community (local mesh networking) can easily test this, too.

PolynomialDivision:
Changed url to https://github.com/Freifunk-Spalter/dice-prefix

PolynomialDivision:
Can we merge? I would like to enable easy testing for other people. ;)

PR #13621, opened 2020-10-10
actions: remove arm_cortex-a9_neon

neheb:
This is the second ARM NEON target that is tested. It's unlikely that
one will fail and the other succeed.

Maintainer: @aparcar


aparcar:
You're the expert

PR #13629, opened 2020-10-11
cgi-io: move into out of tree project

ynezz:
Maintainer: @blogic @jow- 
Compile tested: x86_64
Run tested: none

Description:

No functional changes, just moved the sources into out of tree
project[1] so it's going to be easier to do CI with unit testing,
fuzzing etc.

1. https://git.openwrt.org/?p=project/cgi-io.git;a=shortlog


blogic:
ack, lgtm

PR #13630, opened 2020-10-11
static-neighbor-reports: add package

blocktrron:

Maintainer: me
Compile tested: ath79-generic
Run tested: ath79-generic / OCEDO Koala / OCEDO Ursus / UnifFi AC Lite

Description:

This package allows a user to configure a list of pre-defined neighbors hostapd will report to a client when requesting a 802.11k neighbor report. This enables clients to achieve a significantly reduced roaming time.

### DIFF
--- a/net/static-neighbor-reports/Makefile
+++ b/net/static-neighbor-reports/Makefile
@@ -1,0 +1,44 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=static-neighbor-reports
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=David Bauer <mail@david-bauer.net>
+PKG_LICENSE:=GPL-2.0-only
+
+include $(INCLUDE_DIR)/package.mk
+
+Build/Compile=
+
+define Package/static-neighbor-reports/Default
+endef
+
+define Package/static-neighbor-reports
+  SECTION:=net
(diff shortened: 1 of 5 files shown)


net/static-neighbor-reports/Makefile
```diff
@@ -0,0 +1,44 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=static-neighbor-reports
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=David Bauer <mail@david-bauer.net>
+PKG_LICENSE:=GPL-2.0
```
neheb, line 8: This should match SPDX, meaning add -only or -or-later at the end.

net/static-neighbor-reports/Makefile
```diff
@@ -0,0 +1,44 @@
...
+
+define Package/static-neighbor-reports/Default
+endef
+
+define Package/static-neighbor-reports
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Configure static 802.11k neighbor reports
+  PKGARCH:=all
+  DEPENDS:=+libuci-lua +lua +libubus-lua
```
neheb, line 22: No need for +lua

blocktrron:
@neheb addressed your comments. Can you have a second look?

neheb:
Seems fine.

PR #13632, opened 2020-10-11
golang: make golang compilable on aarch64 host

WiktorS:
Maintainer: @jefferyto 
Compile tested: host-aarch64 (RPi4)
Run tested: host-aarch64 (RPi4)

Description:
Initially I couldn't compile dnscrypt-proxy2 on RPi4. Little investigation and found out the issue is Go bootstrap (1.4) that does not support arm64. It supports arm(32) so I tried forcing it as it should be backwards compatible on arm64 and it worked.
Bootstrap (1.4) is the last version that it written in C++ and it used only to compile the result version of golang in the host arch (arm64 in my case).

OT: Later I found out that golang does not support powerpc, the arch of device I targeted :(

### DIFF
--- a/lang/golang/golang/Makefile
+++ b/lang/golang/golang/Makefile
@@ -76,7 +76,7 @@ BOOTSTRAP_GO_VALID_OS_ARCH:= \
   darwin_386     darwin_amd64 \
   dragonfly_386  dragonfly_amd64 \
   freebsd_386    freebsd_amd64    freebsd_arm \
-  linux_386      linux_amd64      linux_arm \
+  linux_386      linux_amd64      linux_arm    linux_arm64\
   netbsd_386     netbsd_amd64     netbsd_arm \
   openbsd_386    openbsd_amd64 \
   plan9_386      plan9_amd64 \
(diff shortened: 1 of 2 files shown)


jefferyto:
Forcing bootstrap Go to run on arm64 is **not** a good idea.

I suggest you install Go on your build system (either through your distro's package manager or from golang.org) then set [`CONFIG_GOLANG_EXTERNAL_BOOTSTRAP_ROOT`](https://github.com/openwrt/packages/blob/cb2b03a719265cb66f4c7b846e2e4528efbbd135/lang/golang/golang/Config.in#L3-L13).

WiktorS:
I understand that this is more like a dirty hack, I just like to have things working out of the box.

OT: Was it considered to download the target binary builds directly from golang.org without bootstrap? Since it is required to provide bootstrap (at least for some host archs), wouldn't it be better download binaries as it is needed anyways or maybe there is a host/arch combination not available on golang.org?

jefferyto:
I think building from source is safer (theoretically at least) than downloading binaries.

BKPepe:
So, this can be closed what I understand, right?

WiktorS:
If there's no go for this patch, please close it

jefferyto:
I'll think about adding an option to download pre-built bootstrap binaries, but I'm not inclined to accept any patches for bootstrap Go (unless it is to fix a serious issue and the full Go test suite has been run and passed).

WiktorS:
Closing. Looking forward to pre-build binaries ;)

PR #13640, opened 2020-10-12
mariadb: fix compilation when selinux is present

neheb:

Maintainer: @miska 
Compile tested: ath79

### DIFF
--- a/utils/mariadb/patches/200-no-selinux.patch
+++ b/utils/mariadb/patches/200-no-selinux.patch
@@ -1,0 +1,11 @@
+--- a/support-files/CMakeLists.txt
++++ b/support-files/CMakeLists.txt
+@@ -61,7 +61,7 @@ IF(UNIX)
+     INSTALL(FILES magic DESTINATION ${inst_location} COMPONENT SupportFiles)
+     INSTALL(DIRECTORY policy DESTINATION ${inst_location} COMPONENT SupportFiles)
+     FIND_PROGRAM(CHECKMODULE checkmodule)
+-    FIND_PROGRAM(SEMODULE_PACKAGE semodule_package)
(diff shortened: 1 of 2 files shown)


miska:
Fine by me.

utils/mariadb/patches/200-no-selinux.patch
```diff
@@ -0,0 +1,11 @@
+--- a/support-files/CMakeLists.txt
++++ b/support-files/CMakeLists.txt
+@@ -61,7 +61,7 @@ IF(UNIX)
+     INSTALL(FILES magic DESTINATION ${inst_location} COMPONENT SupportFiles)
+     INSTALL(DIRECTORY policy DESTINATION ${inst_location} COMPONENT SupportFiles)
+     FIND_PROGRAM(CHECKMODULE checkmodule)
+-    FIND_PROGRAM(SEMODULE_PACKAGE semodule_package)
++#    FIND_PROGRAM(SEMODULE_PACKAGE semodule_package)
```
dr-m, line 8: I didn’t find anything obvious in [FIND_PROGRAM](https://cmake.org/cmake/help/latest/command/find_program.html), but my colleague suggests that passing `-DSEMODULE_PACKAGE_EXECUTABLE=NOT-FOUND` to CMake should pretend that the program does not exist. I would suggest that you add that to `CMAKE_OPTIONS` in `Makefile` and reviewing the contents of `CMakeCache.txt` in the build directory. In that way, this patch could be dropped.

While reviewing the Makefile, I noticed a setting `-DSTACK_DIRECTION=-1`, which should have been made redundant by MariaDB/server@d0abbdf56e11ccc88447c1dc80caaf355c94be3b.

PR #13643, opened 2020-10-12
docker-ce: disable docker iptables changes

feckert:
Maintainer: @G-M0N3Y-2503 
Compile tested: x86_64, APU3, OpenWrt master)

Description:
Openwrt has a own firewall service called fw3, that supports firewall zones. Docker can bypass the handling of the zone rules in openwrt via custom tables. These are "always" processed before the openwrt firewall. Which is prone to errors!

Since not everyone is aware that the firewall of openwrt will not be passed. And this is a security problem because a mapped port is visible on all interfaces and so also on the WAN side. If the firewall handling in docker is switched off, then the port in fw3 must be explicitly released and it cannot happen that the port is accidentally exported to the outside world via the interfaces on the WAN zone.

So all rules for the containers should and so must be made in fw3.

Edit:
This describes the problem https://fralef.me/docker-and-iptables.html And since we are on a router that has several interfaces. Is that a security problem, the way docker does it

### DIFF
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -137,7 +137,7 @@ ucidel() {
 }
 
 process_config() {
-	local alt_config_file data_root log_level bip
+	local alt_config_file data_root log_level iptables bip
 
 	[ -f /etc/config/dockerd ] || {
 		# Use the daemon default configuration
@@ -150,9 +150,6 @@ process_config() {
 	mkdir -p "${DOCKER_CONF_DIR}"
 
 	config_load 'dockerd'
-
-	config_list_foreach firewall blocked_interfaces add_docker_firewall_rules
-
 	config_get alt_config_file globals alt_config_file
 	[ -n "${alt_config_file}" ] && [ -f "${alt_config_file}" ] && {
 		ln -s "${alt_config_file}" "${DOCKERD_CONF}"
@@ -161,6 +158,7 @@ process_config() {
 
 	config_get data_root globals data_root "/opt/docker/"
 	config_get log_level globals log_level "warn"
+	config_get_bool iptables globals iptables "1"
 	config_get bip globals bip ""
 
 	. /usr/share/libubox/jshn.sh
@@ -174,6 +172,9 @@ process_config() {
 	json_add_array "hosts"
 	config_list_foreach globals hosts json_add_array_string
 	json_close_array
+
+	json_add_boolean iptables "${iptables}"
+	[ "${iptables}" -ne "0" ] && config_foreach iptables_add_blocking_rule firewall
 
 	json_dump > "${DOCKERD_CONF}"
 
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -181,16 +181,39 @@ service_triggers() {
 	procd_add_reload_trigger 'dockerd'
 }
 
-add_docker_firewall_rules() {
-	. /lib/functions/network.sh
-	local device interface="${1}"
+iptables_add_blocking_rule() {
+	local cfg="$1"
 
-	# Ignore errors as it might already be present
-	iptables --table filter --new DOCKER-USER 2>/dev/null
-	network_get_physdev device "${interface}"
-	if ! iptables --table filter --check DOCKER-USER --in-interface "${device}" --out-interface docker0 --jump DROP 2>/dev/null; then
-		iptables --table filter --insert DOCKER-USER --in-interface "${device}" --out-interface docker0 --jump DROP
-	fi
+	local device=""
+
+	handle_iptables_rule() {
+		local interface="$1"
+		local outbound="$2"
+
+		local inbound=""
+
+		. /lib/functions/network.sh
+		network_get_physdev inbound "${interface}"
+
+		[ -z "$inbound" ] && {
+			logger -t "dockerd-init" -p notice "Unable to get physical device for interface ${interface}"
+			return
+		}
+
+		if ! iptables --table filter --check DOCKER-USER --in-interface "${inbound}" --out-interface "${outbound}" --jump DROP 2>/dev/null; then
+			logger -t "dockerd-init" -p notice "Drop traffic from ${inbound} to ${outbound}"
+			iptables --table filter --insert DOCKER-USER --in-interface "${inbound}" --out-interface "${outbound}" --jump DROP
+		fi
+	}
+
+	config_get device "$cfg" device
+
+	[ -z "$device" ] && {
+		logger -t "dockerd-init" -p notice "No device configured for ${cfg}"
+		return
+	}
+
+	config_list_foreach "$cfg" blocked_interfaces handle_iptables_rule "$device"
 }
 
 ip4tables_remove_nat() {
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -3,8 +3,8 @@
 USE_PROCD=1
 START=25
 
-extra_command "uciadd" "Add default bridge configuration to network and firewall uci config"
-extra_command "ucidel" "Delete default bridge configuration from network and firewall uci config"
+extra_command "uciadd" "<interface> <device> <zone> Add docker bridge configuration to network and firewall uci config"
+extra_command "ucidel" "<interface> <device> <zone> Delete docker bridge configuration from network and firewall uci config"
 
 DOCKER_CONF_DIR="/tmp/dockerd"
 DOCKERD_CONF="${DOCKER_CONF_DIR}/daemon.json"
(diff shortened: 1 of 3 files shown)


neheb:
```
2020-10-12 11:47:33 [WARN  ]   Error: File:[utils/docker-ce/files/dockerd.init] is not executable
```

Interesting...

G-M0N3Y-2503:
I'm curious about what errors you saw...
Because I'm wondering if we can [restrict connections to the Docker host](https://docs.docker.com/network/iptables/#restrict-connections-to-the-docker-host) so that open ports can only be accessed via localhost or explicitly opened. or perhaps [setting the default bind address for containers](https://docs.docker.com/network/iptables/#setting-the-default-bind-address-for-containers) to localhost by default. If these ways work then at least we can mention these documented ways that we set by default rather than leave everything up to the user. I'll see what I can figure out this weekend.

If that won't work, can we remove any of the `ip4tables_remove()` with this change?

feckert:
> I'm curious about what errors you saw...

That was not because I saw errors.
I wanted to express that when creating containers I need to know what I am doing.
It can easily happen that I take some compose file and start the containerstack and then all ports are mapped to the outside world what I don't want.
The router has several interfaces, which is not usually the case when using a standard docker setup.

> Because I'm wondering if we can [restrict connections to the Docker host](https://docs.docker.com/network/iptables/#restrict-connections-to-the-docker-host) so that open ports can only be accessed via localhost or explicitly opened. or perhaps [setting the default bind address for containers](https://docs.docker.com/network/iptables/#setting-the-default-bind-address-for-containers) to localhost by default. If these ways work then at least we can mention these documented ways that we set by default rather than leave everything up to the user. I'll see what I can figure out this weekend.

sound good

> If that won't work, can we remove any of the `ip4tables_remove()` with this change?

Of course.
If we disable the firewall change by docker on default, then we should also remove the delete, because these rules are no longer created.

feckert:
> ```
> 2020-10-12 11:47:33 [WARN  ]   Error: File:[utils/docker-ce/files/dockerd.init] is not executable
> ```

G-M0N3Y-2503:
I can see why you wanted to disable iptables now, docker does its own thing and OpenWrt does its own thing.
However, since we have a `DOCKER-USER` chain we could just add the `DROP` rules we need to it, like drop all from wan and all from the docker zone too.
I haven't figured out how to do that just yet, but do you have any thoughts on it?

feckert:
> However, since we have a `DOCKER-USER` chain we could just add the `DROP` rules we need to it, like drop all from wan and all from the docker zone too.

I don't know if there are any other things with the firewall in docker that is usefull (I haven't looked at it closely).
I can only say that if I deactivate the firewall handling in docker, then I have to configure everything with fw3.
That is what I wanted to achieve.
Because like you said before, fw3 doesn't know about the firewall handling in docker!

> I haven't figured out how to do that just yet, but do you have any thoughts on it?

The question now is what we want to do:
1. If we disable docker's firewall handling, then we have to do everything ourselves. The advantage would be that we have better control over the rules, because we do everything in fw3. It can't happen that someone accidentally sets up a container which is then implicitly accessible from the WAN.
2. We leave the firewall on through dockers. Then the system behaves the same as on the other systems. But the disadvantage is that we don't know what docker is actually doing unless you look at the rule directly. Can the user be expected to do that?


By the way I am working on to add podman for openWrt. And the way I see it, podman is not doing anything with the firewall.

utils/docker-ce/files/dockerd.init
```diff
@@ -145,6 +145,7 @@ process_config() {
 	json_add_array "hosts"
 	config_list_foreach globals hosts json_add_array_string
 	json_close_array
+	json_add_boolean iptables 0
```
G-M0N3Y-2503, line 148: I think for this to go in we'd have to make it configurable and only add it if it's `false` as per another comment

G-M0N3Y-2503 (changes requested):

G-M0N3Y-2503:
> I don't know if there are any other things with the firewall in docker that is usefull (I haven't looked at it closely).

I myself haven't used dockers networking extensively either.

> 1. If we disable docker's firewall handling, then we have to do everything ourselves. The advantage would be that we have better control over the rules, because we do everything in fw3. It can't happen that someone accidentally sets up a container which is then implicitly accessible from the WAN.

Given that "It is not possible to completely prevent Docker from creating iptables rules, and creating them after-the-fact is extremely involved". And that "Setting iptables to false will more than likely break container networking for the Docker engine." from [Prevent Docker from manipulating iptables](https://docs.docker.com/network/iptables/#prevent-docker-from-manipulating-iptables), And with iptables disabled "Docker containers will not be able to connect to external hosts or the internet when using network other than default bridge." from [Run multiple daemons](https://docs.docker.com/engine/reference/commandline/dockerd/#run-multiple-daemons). 

It leads me to believe that disabling iptables is only a workaround for some cases.

> 2. We leave the firewall on through dockers. Then the system behaves the same as on the other systems. But the disadvantage is that we don't know what docker is actually doing unless you look at the rule directly. Can the user be expected to do that?

I think that if people also choose to use docker, they can be expected to do [what the documentation explains](https://docs.docker.com/network/iptables/). That said, I think we can make it a bit nicer by adding another custom configuration option that blocks traffic from configured interfaces, with the `wan` interface there by default. I've had a look at it and should be possible, I can submit a PR for this.

That said, I tend to agree that disabling iptables can be a valid solution for simple cases, though don't think it should be the default.

feckert:
This is superseded by https://github.com/openwrt/packages/pull/13796

feckert:
**This is my firewall filter ruleset if docker firewall handling is disabled:**
Docker firewall handling is integrated in fw3 with additional uci configs set by `/etc/init.d/dockerd uciadd`.
iptables-filter


**This is my firewall filter ruleset if docker sets the firewall rules:**
Docker firewall handling is not integrated in fw3.
iptables-filter

champtar:
@feckert what tool are you using to display the iptables rules ? really nice

feckert:
> @feckert what tool are you using to display the iptables rules ? really nice

@champtar I use this tool https://metacpan.org/release/App-Iptables2Dot on my development machine.
Unfortunately it does not run on the router. That would realy a great benefit

feckert:
@G-M0N3Y-2503 I have now made a few adjustments which you can have a look at.
I have described the changes in the commit messages.

utils/docker-ce/files/dockerd.init
```diff
@@ -149,9 +150,14 @@ process_config() {
 	config_list_foreach globals hosts json_add_array_string
 	json_close_array
 
-	json_dump > "${DOCKERD_CONF}"
+	if [ "${iptables}" -eq '0' ]; then
+		json_add_boolean iptables "0"
+	else
+		config_foreach iptables_add_blocking_rule firewall
+		json_add_boolean iptables "1"
+	fi
```
G-M0N3Y-2503, line 158: we could reduce this to something like
```suggestion
	json_add_boolean iptables "${iptables}"

	[ "${iptables}" -ne "0" ] && config_foreach iptables_add_blocking_rule firewall
```

utils/docker-ce/files/dockerd.init
```diff
@@ -135,6 +135,7 @@ process_config() {
 
 	config_get data_root globals data_root "/opt/docker/"
 	config_get log_level globals log_level "warn"
+	config_get_bool iptables globals iptables "1"
```
G-M0N3Y-2503, line 138: Since we don't do all config options we should probably add this to the config file for documentation.
I'd prefer we had `'true'` in the config file instead of a number so it matches the JSON closer.

G-M0N3Y-2503 (changes requested):
Awesome work!
Just a couple of things to improve but otherwise looks good to me.

utils/docker-ce/files/dockerd.init
feckert: > Since we don't do all config options we should probably add this to the config file for documentation.

I added the option to the uci template configuration as you suggested.

> I'd prefer we had 'true' in the config file instead of a number so it matches the JSON closer.

I understand but in the uci I think it is better to use 0 or 1 for boolean.
The data become valid json bythe function json_add_boolean.
This function converts the 0 and 1 to a valid boolean.
If we take true or false, then i have to adjust the if state as well.

I think and my experience with openwrt is, that it is rather normal to work with 0 or 1 at bool values.

G-M0N3Y-2503: From the scripting perspective [It looks like we'd still get a 1/0](https://github.com/openwrt/openwrt/blob/master/package/base-files/files/lib/functions.sh#L122), so I thought it would be fine. That said, it was just a personal preference so whatever you think is best works for me.

G-M0N3Y-2503 (approved):

feckert:
@G-M0N3Y-2503 thanks for reviewing my changes

PR #13645, opened 2020-10-12
Update libqmi to 1.26.6 and make QMI-over-MBIM support optional

aleksander0m:
Maintainer: @nickberry17 

Description: Upgrade library version to 1.26.6 and also add a new CONFIG_LIBQMI_WITH_MBIM_QMUX that allows disabling the libmbim dependency altogether (i.e. for systems running with modems in QMI mode explicitly).

### DIFF
--- a/libs/libqmi/Makefile
+++ b/libs/libqmi/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libqmi
 PKG_VERSION:=1.26.6
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://www.freedesktop.org/software/libqmi
@@ -23,10 +23,16 @@ PKG_BUILD_PARALLEL:=1
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/nls.mk
 
+define Package/libqmi/config
+  source "$(SOURCE)/Config.in"
+endef
+
 define Package/libqmi
   SECTION:=libs
   CATEGORY:=Libraries
-  DEPENDS:=+libmbim
+  DEPENDS:= \
+    +glib2 \
+    +LIBQMI_WITH_MBIM_QMUX:libmbim
   TITLE:=Helper library to talk to QMI enabled modems
   URL:=https://www.freedesktop.org/wiki/Software/libqmi
   LICENSE:=LGPL-2.0-or-later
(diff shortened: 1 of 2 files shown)


aleksander0m:
I've ignored the optional QRTR support for now, as that is not fully stable yet.

libs/libqmi/Makefile
```diff
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libqmi
 PKG_VERSION:=1.26.6
-PKG_RELEASE:=1
+PKG_RELEASE:=2
```
neheb, line 12: If you're already updating the package with this change, why bump PKG_RELEASE?

libs/libqmi/Makefile
```diff
@@ -59,11 +65,16 @@ CONFIGURE_ARGS += \
 	--disable-gtk-doc-pdf \
 	--disable-silent-rules \
 	--enable-firmware-update \
-	--enable-mbim-qmux \
```
neheb, line 62: Could do

```
--$(if $(LIBQMI_WITH_MBIM_QMUX),en,dis)able-mbim-qmux \
```

libs/libqmi/Makefile
nickberry17: A very welcome change, @aleksander0m !  Thanks for your hard work.  Correct, if the PKG_VERSION increases, then PKG_RELEASE resets to 1.

aleksander0m: @neheb @nickberry17 the thing is that this PR has 3 commits, one doing `1.26.2-1 -> 1.26.4-1`; another one doing `1.26.4-1 -> 1.26.6-1`; and the last one adding the new config option, bumping PKG_RELEASE: `1.26.6-1 -> 1.26.6-2`. I didn't want to have the config change in the same commit as the version bump, as it was a completely orthogonal thing. Would it be better if I just send one single commit with everything? I personally prefer the 3 separate commits, but whatever you prefer.

libs/libqmi/Makefile
aleksander0m: oh, nice, will change that

libs/libqmi/Makefile
neheb: I mean up to you. Typically, a PR either bumps the PKG_RELEASE or resets it to 1 as anything intermediate is not present.

aleksander0m: If it's up to me, I'd prefer to leave it as 3 separate commits, each bumping only the version field that needs to be bumped. In my mind a PR is a collection of commits, but each commit needs to be self-contained by itself. Makes it easier e.g. to cherry-pick to the 19.07 branch if we want to cherry-pick the version bump only and not the new config change.

PR #13650, opened 2020-10-12
conserver: remove deprecated usleep

neheb:

Maintainer: @bmork 
Compile tested: ath79

### DIFF
--- a/net/conserver/Makefile
+++ b/net/conserver/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=conserver
 PKG_VERSION:=8.2.5
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


bmork:
Has this been submitted upstream?  Upstream is quite responsive.  I do not understand why this should be a distro specific patch.  We definitely do not want to keep it around forever.  So either it is accepted upstream or it is dropped.  And there is nothing here requiring immediate action, so this can wait for the next upstream relase.

neheb:
https://github.com/bstansell/conserver/pull/52

Upstream seems dead from what I see.

bmork:
Thanks.  Upstream tends to do batch processing, which is fine. And the release frequency matches the number of changes.  So you may have to wait for a while before a submisison is processed and pushed out.  I really do not see that as a problem wrt this pacth, which AFAIU does not solve any real issue for anyone.

neheb:
Seems like this was merged. This fixes compilation with a non-default configuration of uClibc-ng.

PR #13653, opened 2020-10-13
znc: Add support for many configurable listeners for znc

FinnixB:
Function add_listener called from /etc/init.d/znc, but not implemented

Signed-off-by: Pavel Demkovich <finn@finnix.servebeer.com>

Description:


neheb:
Maintainer is @KanjiMonster 

PR #13658, opened 2020-10-14
mariadb: procd adoption

miska:
Maintainer: me
Compile tested: mvebu, Turris Omnia, Turris OS 5.2 = OpenWrt 19.07
Run tested: mvebu, Turris Omnia, Turris OS 5.2 = OpenWrt 19.07, server started via init

Description:
Cleaning up init script a little, running as a user from start and relying on procd.


utils/mariadb/files/mysqld.init
```diff
@@ -116,48 +122,39 @@ start() {
...
+
+	# shellcheck disable=SC2086
+	procd_set_param command "$MYSQLD" $options
+	procd_set_param respawn "${respawn_threshold:-3600}" "${respawn_timeout:-5}" "${respawn_retry:-5}"
+	# run as user
+	procd_set_param user "$my_user"
+	# forward stderr to logd
+	procd_set_param stderr 1
+	# use HUP to reload
+	procd_set_param reload_signal HUP
```
neheb, line 143: this can be replaced with procd_send_signal I believe.

miska: From what I understood from https://openwrt.org/docs/guide-developer/procd-init-scripts it seemed that I can either define my own reload and use `procd_send_signal` in it or I can pass the signal to procd and used the already defined reload function. Did I misunderstood?

neheb: Hmm good point. I've always seen HUP implemented in a custom reload function.

```
~/d/packages (master)> git grep procd_send_signal | wc -l
16
~/d/packages (master)> git grep reload_signal | wc -l
2
```

The last two are

```
admin/backuppc/files/backuppc.init:    procd_set_param reload_signal 1
net/haproxy/files/haproxy.init: procd_set_param reload_signal USR2
```

miska: I don't mind changing that, just wondering if there is some difference I missed in functionality.

PR #13676, opened 2020-10-15
boost: improve package compilation

### LINKED ISSUE #13652: Boost: can't be linked with stdcpp automatically
Maintainer: @ClaymorePT @neheb 
Environment: x86_64

Description:

Find an issue that the package stdcpp in the `DEPENDS` can't be linked automatically. 
Check the linked libs with `ldd`, and get the errors as following:
```
/usr/lib/libboost_chrono.so.1.74.0
	ldd (0x7fd4a5872000)
	libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x7fd4a5854000)
	libc.so => ldd (0x7fd4a5872000)
Error relocating /usr/lib/libboost_chrono.so.1.74.0: _ZTVN10__cxxabiv117__class_type_infoE: symbol not found
Error relocating /usr/lib/libboost_chrono.so.1.74.0: _ZTVN10__cxxabiv117__class_type_infoE: symbol not found
Error relocating /usr/lib/libboost_chrono.so.1.74.0: _ZTVN10__cxxabiv117__class_type_infoE: symbol not found
Error relocating /usr/lib/libboost_chrono.so.1.74.0: _ZTVN10__cxxabiv120__si_class_type_infoE: symbol not found
Error relocating /usr/lib/libboost_chrono.so.1.74.0: _ZTVN10__cxxabiv120__si_class_type_infoE: symbol not found
Error relocating /usr/lib/libboost_chrono.so.1.74.0: _ZTVN10__cxxabiv120__si_class_type_infoE: symbol not found
Error relocating /usr/lib/libboost_chrono.so.1.74.0: _ZTVN10__cxxabiv120__si_class_type_infoE: symbol not found
```
If add the `-lstdc++` to `TARGET_LDFLAGS`, everything is OK. 
```
ldd /usr/lib/libboost_chrono.so.1.74.0
        ldd (0x7ff624929000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7ff6247e6000)
        libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x7ff6247d2000)
        libc.so => ldd (0x7ff624929000)
```

However, if some packages depend on boost libs and are linked directly or indirectly with `libstdc++.so.*`, it seems that they can also woks well. Therefore, is it necessary to link the boost library by adding `-lstdc++` to `TARGET_LDFLAGS` manually.

Test the boost on archlinux, and it seems that the libstdc++ is linked dynamically.
```
ldd /tmp/libboost_chrono.so.1.72.0
        ldd (0x7f056eefc000)
        librt.so.1 => ldd (0x7f056eefc000)
        libpthread.so.0 => ldd (0x7f056eefc000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7f056edb6000)
        libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x7f056eda2000)
        libc.so.6 => ldd (0x7f056eefc000)
```

brvphoenix:
Maintainer: @ClaymorePT @neheb
Environment: x86_64

Description:
* Add `-lstdc++` to `LDFLAGS`, and close #13652
* Add `-Wl,--gc-sections,--as-needed` to  `LDFLAGS` and `-ffunction-sections -fdata-sections -flto` to `CFLAGS` which will reduce the size by 10% ~ 13%
* Only a virtual package will be created if only static libs are built, which will avoid compiliation failures
* Add python3-base to boost-python3's dependences according to [python.jam](https://github.com/boostorg/build/blob/develop/src/tools/python.jam#L1010~#L1014)

Signed-off-by: Van Waholtz <vanwaholtz@gmail.com>

### DIFF
--- a/libs/boost/Makefile
+++ b/libs/boost/Makefile
@@ -13,7 +13,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=boost
 PKG_VERSION:=1.74.0
 PKG_SOURCE_VERSION:=1_74_0
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 PKG_SOURCE:=$(PKG_NAME)_$(PKG_SOURCE_VERSION).tar.bz2
 PKG_SOURCE_URL:=@SF/$(PKG_NAME)/$(PKG_NAME)/$(PKG_VERSION) https://dl.bintray.com/boostorg/release/$(PKG_VERSION)/source/
@@ -80,7 +80,7 @@ There are many more header-only libraries supported by Boost.
 See more at http://www.boost.org/doc/libs/1_74_0/
 endef
 
-PKG_BUILD_DEPENDS:=boost/host PACKAGE_boost-python3:python3
+PKG_BUILD_DEPENDS:=boost/host
 
 include ../../lang/python/python3-version.mk
 BOOST_PYTHON3_VER=$(PYTHON3_VERSION)
@@ -302,12 +302,17 @@ endef
 define Build/Configure
 endef
 
+define Package/boost/Default/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/lib/libboost_$(2)*.so* $(1)/usr/lib/
+endef
+
 # 1: short name
 # 2: dependencies on other boost libraries (short name)
 # 3: dependencies on other packages
 # 4: conditional/inward dependencies
+# 5: dependencies compiled only when this package has been selected
 define DefineBoostLibrary
-
   BOOST_DEPENDS+= +$(if $(4),$(4):boost-$(1),boost-$(1))
   PKG_CONFIG_DEPENDS+= CONFIG_PACKAGE_boost-$(1)
 
@@ -316,12 +321,16 @@ define DefineBoostLibrary
   define Package/boost-$(1)
     $(call Package/boost/Default)
     TITLE+= ($(1))
-    DEPENDS+= $$(foreach lib,$(2),+boost-$$(lib)) $(3) $(if $(4),@$(4),) $(patsubst %,+PACKAGE_boost-$(1):%,$(5))
+    DEPENDS+= $(foreach lib,$(2),+boost-$(lib)) $(3) $(if $(4),@$(4),) $(patsubst %,+PACKAGE_boost-$(1):%,$(5))
     HIDDEN:=1
   endef
 
   define Package/boost-$(1)/description
    This package contains the Boost $(1) library.
+  endef
+
+  define Package/boost-$(1)/install
+    $(if $(CONFIG_boost-static-libs),true,$(call Package/boost/Default/install,$$(1),$(1)))
   endef
 endef
 
@@ -343,7 +352,7 @@ $(eval $(call DefineBoostLibrary,math))
 #$(eval $(call DefineBoostLibrary,mpi,,)) # OpenMPI does no exist in OpenWRT at this time.
 $(eval $(call DefineBoostLibrary,nowide))
 $(eval $(call DefineBoostLibrary,program_options))
-$(eval $(call DefineBoostLibrary,python3))
+$(eval $(call DefineBoostLibrary,python3,,,,python3-base))
 $(eval $(call DefineBoostLibrary,random,system))
 $(eval $(call DefineBoostLibrary,regex))
 $(eval $(call DefineBoostLibrary,serialization))
@@ -368,10 +377,10 @@ define Host/Compile
 endef
 
 CONFIGURE_PREFIX:=$(PKG_INSTALL_DIR)
-TARGET_LDFLAGS += -pthread -lrt
+TARGET_LDFLAGS += -pthread -lrt -lstdc++ -Wl,--gc-sections,--as-needed,--print-gc-sections
 
 TARGET_CFLAGS += \
-	$(if $(CONFIG_SOFT_FLOAT),-DBOOST_NO_FENV_H) -fPIC
+	$(if $(CONFIG_SOFT_FLOAT),-DBOOST_NO_FENV_H) -fPIC -ffunction-sections -fdata-sections -flto
 
 EXTRA_CXXFLAGS += $(if $(CONFIG_GCC_USE_VERSION_10),-std=gnu++20,$(if $(CONFIG_GCC_USE_VERSION_5),-std=gnu++14,-std=gnu++17))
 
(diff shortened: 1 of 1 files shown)


libs/boost/Makefile
```diff
@@ -302,12 +302,17 @@ endef
 define Build/Configure
 endef
 
+define Package/boost/Default/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/lib/libboost_$(2)*.so* $(1)/usr/lib/
+endef
+
```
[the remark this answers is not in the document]
brvphoenix: I have moved `define Package/boost-$(1)/install` into `DefineBoostLibrary`, and the definition of `Package/boost/Default/install` should be placed before the `$(eval $(call DefineBoostLibrary,*))`.

libs/boost/Makefile
```diff
@@ -368,10 +377,10 @@ define Host/Compile
 endef
 
 CONFIGURE_PREFIX:=$(PKG_INSTALL_DIR)
-TARGET_LDFLAGS += -pthread -lrt
+TARGET_LDFLAGS += -pthread -lrt -lstdc++ -Wl,--gc-sections,--as-needed
 
 TARGET_CFLAGS += \
-	$(if $(CONFIG_SOFT_FLOAT),-DBOOST_NO_FENV_H) -fPIC
+	$(if $(CONFIG_SOFT_FLOAT),-DBOOST_NO_FENV_H) -fPIC -ffunction-sections -fdata-sections -flto
```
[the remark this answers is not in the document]
brvphoenix: All right.

ClaymorePT:
@brvphoenix These additions are pretty cool! Thanks for the effort.

libs/boost/Makefile
```diff
@@ -368,10 +377,10 @@ define Host/Compile
 endef
 
 CONFIGURE_PREFIX:=$(PKG_INSTALL_DIR)
-TARGET_LDFLAGS += -pthread -lrt
+TARGET_LDFLAGS += -pthread -lrt -lstdc++ -Wl,--gc-sections,--as-needed,--print-gc-sections
```
neheb, line 380: Do you have some insight as to why -lstdc++ is not added by the Make system?

Locally I'm having similar trouble when compiling pdns and friends with libcxx:

```
/home/mangix/devstuff/openwrt/staging_dir/toolchain-powerpc_8540_gcc-10.2.0_musl/lib/gcc/powerpc-openwrt-linux-musl/10.2.0/../../../../powerpc-openwrt-linux-musl/bin/ld: /home/mangix/devstuff/openwrt/staging_dir/target-powerpc_8540_musl/usr/lib/libboost_context.so: undefined reference to `__once_proxy'
/home/mangix/devstuff/openwrt/staging_dir/toolchain-powerpc_8540_gcc-10.2.0_musl/lib/gcc/powerpc-openwrt-linux-musl/10.2.0/../../../../powerpc-openwrt-linux-musl/bin/ld: /home/mangix/devstuff/openwrt/staging_dir/target-powerpc_8540_musl/usr/lib/libboost_context.so: undefined reference to `std::__once_callable'
/home/mangix/devstuff/openwrt/staging_dir/toolchain-powerpc_8540_gcc-10.2.0_musl/lib/gcc/powerpc-openwrt-linux-musl/10.2.0/../../../../powerpc-openwrt-linux-musl/bin/ld: /home/mangix/devstuff/openwrt/staging_dir/target-powerpc_8540_musl/usr/lib/libboost_context.so: undefined reference to `std::__throw_system_error(int)'
/home/mangix/devstuff/openwrt/staging_dir/toolchain-powerpc_8540_gcc-10.2.0_musl/lib/gcc/powerpc-openwrt-linux-musl/10.2.0/../../../../powerpc-openwrt-linux-musl/bin/ld: /home/mangix/devstuff/openwrt/staging_dir/target-powerpc_8540_musl/usr/lib/libboost_context.so: undefined reference to `std::__once_call'
```

brvphoenix: Have no idea. It seems that it has something to do with the `b2` and the build system based on musl-libc. The boost libs on alpine have the same issue.

neheb: Hmm OK.

neheb:
Interesting failure: https://downloads.openwrt.org/snapshots/faillogs/arc_arc700/packages/boost/compile.txt

brvphoenix:
@neheb 
The uClibc doesn't support unicode conversion and maybe need to enable the `icu`.

I'm curious whether the boost-regex can work well on arcs.

neheb:
In any case, it's picking up and using icu.

neheb:
To use iconv, nls.mk must be included.

ClaymorePT:
nls.mk is included and so is BUILD_NLS being used for boost.locale

`$(eval $(call DefineBoostLibrary,locale,system,$(ICONV_DEPENDS),BUILD_NLS))`

neheb:
Yeah I got confused with a different package that allows selection of either iconv or icu.

ClaymorePT:
Awesome would be to be able to run all boost tests for each specific target. But not hardware for that :) (maybe just for x86_64)

brvphoenix:
>Boost.Locale creates the natural glue between the C++ locales framework, iostreams, and the powerful ICU library.

>Boost.Locale provides non-ICU based localization support as well. It is based on the operating system native API or on the standard C++ library support. Sacrificing some less important features, Boost.Locale becomes less powerful but lighter and easier to deploy and use library.

From: https://www.boost.org/doc/libs/1_74_0/libs/locale/doc/html/index.html

The boost-locale can also be compiled with icu instead of iconv. If enable ICU for boost-regex, we can also use ICU rather than ICONV in boost-locale.


>Using Boost.Locale DLL's with MinGW gcc also requires dynamic linking with the runtime libraries libstdc++ and libgcc. Some gcc builds use static linking by default so make sure you use correct link options with your compiler when you build your own programs.

From: https://www.boost.org/doc/libs/1_74_0/libs/locale/doc/html/building_boost_locale.html

It seems that only when stdc++ has been linked, the uncicode conversion can be enabled. In the past, the boost-regex was compiled without ICU enabled. For consistency, may we just need to disable the ICU support in boost-regex by default?

ClaymorePT:
@brvphoenix I don't really remember the reason of why iconv is used instead of ICU, but I think it was because of ICU issues in openwrt. Can't really recall what they were... It was a very long time ago.

neheb:
ICU works fine with uClibc-ng currently. It was actually a bug in the toolchain that broke ICU: https://github.com/openwrt/openwrt/commit/1647790ae6202bd70b199e2aa44f8eb43e8d3af1

ICU should also work better with uClibc-ng as iconv support in the libc is totally missing. It's stubbed out with libiconv-stub.

brvphoenix:
@neheb @ClaymorePT 

Now the boost-locale can only use libiconv-full as the backend when `BUILD_NLS`. In fact, we can use ICU or libiconv-full as the only backend. It can be compiled without iconv support by the libc.

I have let boost use icu as the only backend in https://github.com/openwrt/packages/pull/13730.

PR #13682, opened 2020-10-16
fix readme

Qh3xg3m:
Maintainer: me / @\Qh3xg3m (find it by checking history of the package Makefile)

Description: fix readme


BKPepe:
Shouldn't be your pull request part of Spamtoberfest?

PR #13684, opened 2020-10-16
CI: fix grammar typo when test is succesful

BKPepe:
Maintainer: @aparcar

While adding test.sh to one of my packages, which I maintain, I noticed in the output, that there is typo. It's nitpick, but hopefully you don't mind that.

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -46,7 +46,7 @@ jobs:
           PACKAGES="${PACKAGES:-vim tmux bmon}"
 
           echo "Building $PACKAGES"
-          echo "::set-env name=PACKAGES::$PACKAGES"
+          echo "PACKAGES=$PACKAGES" >> $GITHUB_ENV
 
       - name: Build
         uses: openwrt/gh-action-sdk@v1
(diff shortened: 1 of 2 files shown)


BKPepe:
Ok, because it failed, I noticed there were some warnings:

https://github.com/openwrt/packages/actions/runs/310968253

so here goes another commit.

.github/workflows/multi-arch-test-build.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           PACKAGES="${PACKAGES:-vim tmux bmon}"
 
           echo "Building $PACKAGES"
-          echo "::set-env name=PACKAGES::$PACKAGES"
+          echo "name=PACKAGES::$PACKAGES" >> $GITHUB_ENV
```
aparcar, line 49:
```suggestion
          echo "PACKAGES=$PACKAGES" >> $GITHUB_ENV
```

Based on the [reference](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable) this fix seems wrong.

aparcar (changes requested):

.github/workflows/multi-arch-test-build.yml
BKPepe: Yes, you are correct. I noticed this as well when CI failed.

BKPepe:
Now, it is ok. The rest fails also in other PRs. There is something messed up with SDK.

aparcar:
Thanks

PR #13687, opened 2020-10-16
[openwrt-19.07] python3: Update to 3.7.9, refresh/remove backported patches

jefferyto:
Maintainer: me, @commodo 
Compile tested: armvirt-32, 19.07.4 sdk
Run tested: armvirt-32 (qemu), 19.07.4

Description:


BKPepe (approved):

commodo (approved):

commodo:
Are you working on updating master Python?
I started on it this week, but I have 2 patches I’ve been a bit stuck on.

One is a setuptools patch for setup.py that might be dropped.
The other is a patch that needs re-applying, but I got ambitious and wanted to try to drop it (the platform triplet remove patch); I might just re-apply it; every once in a while I try to get rid of it.

jefferyto:
> Are you working on updating master Python?

Just started looking at it now :joy: you're probably ahead of me in terms of work done.

> One is a setuptools patch for setup.py that might be dropped.

Which patch is this?

> The other is a patch that needs re-applying, but I got ambitious and wanted to try to drop it (the platform triplet remove patch); I might just re-apply it; every once in a while I try to get rid of it.

I would prefer to drop this patch as well, but IIRC (found when working on #12680) if host Python shared objects (.so files) have the platform triplet, then they will not be loaded when building target Python library packages (i.e. when the target setup.py has a host package dependency). I went into a little more detail in the commit message for #12754.

If you want to try dropping the patch, I would suggest also try compiling a package like python-cryptography to make sure it works. (Perhaps there is a way to force Python to load shared objects even when the platform triplet is different.)

jefferyto:
> > One is a setuptools patch for setup.py that might be dropped.

> Which patch is this?

Oh you mean 004-site-patch.patch. Looks like site.py was removed in pypa/setuptools#2166 so it should be safe to drop the patch. I suggest also undoing the change to [python3-package-setuptools.mk](https://github.com/openwrt/packages/pull/12245/files#diff-222f8ec97d7df3f87ecd7f9e54e703d0118139e9b18892c286a3748fee3b88d2).

jefferyto:
> > The other is a patch that needs re-applying, but I got ambitious and wanted to try to drop it (the platform triplet remove patch); I might just re-apply it; every once in a while I try to get rid of it.

> I would prefer to drop this patch as well, but &lt;snip&gt;

I see 016-adjust-config-paths.patch only deals with _sysconfigdata.py so everything I said can be ignored :sweat_smile:

commodo:
> > > One is a setuptools patch for setup.py that might be dropped.
> 
> > Which patch is this?
> 
> Oh you mean 004-site-patch.patch. Looks like site.py was removed in [pypa/setuptools#2166](https://github.com/pypa/setuptools/pull/2166) so it should be safe to drop the patch. I suggest also undoing the change to [python3-package-setuptools.mk](https://github.com/openwrt/packages/pull/12245/files#diff-222f8ec97d7df3f87ecd7f9e54e703d0118139e9b18892c286a3748fee3b88d2).

i did that change for python3-package-setuptools.mk; i found the patch that you added;



> > > The other is a patch that needs re-applying, but I got ambitious and wanted to try to drop it (the platform triplet remove patch); I might just re-apply it; every once in a while I try to get rid of it.
> 
> > I would prefer to drop this patch as well, but <snip>
> 
> I see 016-adjust-config-paths.patch only deals with _sysconfigdata.py so everything I said can be ignored sweat_smile

yes, that patch; i keep wanting to remove it, but it always proves too much work sometimes; maybe i'll just reduce it; it's a pretty big patch, and it's been around for a while

PR #13689, opened 2020-10-16
acme: Allow custom ACME server directory URL

jannispinter:
The underlying `acme.sh` allows custom ACME server URLs (using `--server`). Adding the necessary fields to specify a custom ACME server URL from UCI.

Maintainer: @tohojo 
Compile tested: ipq40xx
Run tested: ipq40xx (tested issuance of a certificate from a custom ACME compatible CA)

Signed-off-by: Jannis Pinter <jannis+openwrt@pinterjann.is>

### DIFF
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -210,6 +210,7 @@ issue_cert()
 	local user_cleanup
 	local ret
 	local domain_dir
+	local acme_server
 
 	config_get_bool enabled "$section" enabled 0
 	config_get_bool use_staging "$section" use_staging
@@ -223,6 +224,7 @@ issue_cert()
 	config_get dns "$section" dns
 	config_get user_setup "$section" user_setup
 	config_get user_cleanup "$section" user_cleanup
+	config_get acme_server "$section" acme_server
 
 	UPDATE_NGINX=$update_nginx
 	UPDATE_UHTTPD=$update_uhttpd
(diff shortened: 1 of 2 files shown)


net/acme/files/run.sh
```diff
@@ -210,11 +210,14 @@ issue_cert()
 	local user_cleanup
 	local ret
 	local domain_dir
+	local use_custom_acme_directory
+	local custom_acme_directory_url
```
tohojo, line 214: Why two variables? Also, please use shorter variable names...

tohojo (changes requested):
Fine with adding this but please fix the variable names. Also, you need to bump the package revision, and fix you S-o-B as the bot says :)

net/acme/files/run.sh
jannispinter: Thanks, I will update the PR as requested. 

I have changes for this feature ready to push for the corresponding LuCI application, and added `use_custom_acme_directory` for a checkbox that conditionally shows/hides the input field for the URL (I don't want to show the input field, when the user doesn't need/understand this option).

jannispinter:
I have added the Signed-off-by line, bumped `PKG_RELEASE` and changed the variable names. I decided to remove `use_custom_acme_directory` entirely, as it is indeed redundant. When the checkbox in LuCI is unchecked, the `directory_url` won't get written to UCI.

net/acme/files/run.sh
```diff
@@ -210,6 +210,7 @@ issue_cert()
 	local user_cleanup
 	local ret
 	local domain_dir
+	local directory_url
```
tohojo, line 213: What does "directory_url" mean? Why not just "acme_server" or "server_url"?

tohojo (changes requested):
Cool, that's better. Going to bikeshed the variable name a bit more, though ;)

net/acme/files/run.sh
jannispinter: The resource an ACME client connects to is called "Directory" (see [RFC 8555](https://tools.ietf.org/html/rfc8555#page-23)). Most implementations put that resource at `/directory` (such as Let's Encrypt, which uses the following directory URL: https://acme-v02.api.letsencrypt.org**/directory**)

I have implemented both client and server versions of the RFC and might have a bit of a tunnel vision on the naming here. I think changing it to `acme_server` actually makes more sense and is a very concise name.

tohojo (approved):
Great, that's better :)

Going to assume those test failures are unrelated, so let's just merge this!

Issue #12942, opened 2020-07-26
python3: Circular dependency between python3-email and python3-urllib


jefferyto:
Maintainer: me, @commodo 
Environment: N/A

Description:
In the Python 3 modules, urllib [imports](https://github.com/python/cpython/blob/v3.8.5/Lib/urllib/request.py#L86) email and email [imports](https://github.com/python/cpython/blob/v3.8.5/Lib/email/utils.py#L31) from urllib.

In our packages, python3-urllib depends on python3-email (#8307) but python3-email does not depend on python3-urllib.

If python3-email has a dependency on python3-urllib, then make will complain about the circular dependency, but if python3-email does not depend on python3-urllib, then installing python3-email only and importing the email module will result in a run-time error. (I found this when testing for #12941.)

Not sure how this should be addressed :frowning_face:

jefferyto:
I ran the stdlib dependency script on Python itself - I haven't gone through the results in detail yet but it looks like there may be a number of imports that are not captured as package dependencies (not sure if there are any other circular dependencies).

commodo:
> Not sure how this should be addressed

One idea would be to merge the 2 packages into a single one.
Maybe `python3-urllib` for the final merged package? Or some other name?


> I ran the stdlib dependency script on Python itself - I haven't gone through the results in detail yet but it looks like there may be a number of imports that are not captured as package dependencies (not sure if there are any other circular dependencies).

I'll admit that when I split the packages into small chunks, I did not take into account these imports much.
It was more empirical. If some package [usually a folder] has a few hundred of kilobytes, it would become it's own package.
I was trying to figure out a good-enough mechanism for reducing the Python/Python3 minimum-required-to-start size [python[3]-base, which was different for Python & Python3], then `python[3]-light` size, then something to make combine `python[3]-light` + all the splits back into a "full" Python/Python3 package [the `files/python[3]-package-xxxx.mk` auto-includes].
Size was more crucial than import sanity.

In the back of my head I always knew this is far from perfect, but if I was going to try to make it perfect, it would take too long.
After ~5 years, it doesn't look like people needed more than what's currently being done.
AFAICT,some people just use `python3` as a dependency, others try to go from `python3-base` / `python3-light`.
Both are equally fine from my side, I guess.

I still feel, that putting CPython with all libs in places with small flash-sizes is weird, but if people want to do that, and this package split helps, then sure.

jefferyto:
> One idea would be to merge the 2 packages into a single one.
[...]

PR #13691, opened 2020-10-17
python-babel: add new package

BKPepe:
Maintainer: me

Description:
- Add [Babel](http://babel.pocoo.org/en/latest/)

Dependency pytz is also required:
`AttributeError: module 'pytz' has no attribute 'UnknownTimeZoneError`
while trying to run tested it with:
`from babel.dates import format_date, format_datetime, format_time`

@jefferyto and @commodo can you please review it? :)

### DIFF
--- a/lang/python/python-babel/Makefile
+++ b/lang/python/python-babel/Makefile
@@ -1,0 +1,49 @@
+#
+# Copyright (C) 2019-2020 CZ.NIC, z. s. p. o. (https://www.nic.cz/)
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=python-babel
+PKG_VERSION:=2.8.0
+PKG_RELEASE:=1
+
+PYPI_NAME:=Babel
+PKG_HASH:=1aac2ae2d0d8ea368fa90906567f5c08463d98ade155c0c4bfedd6a0f7160e38
+
+PKG_MAINTAINER:=Josef Schlehofer <josef.schlehofer@nic.cz>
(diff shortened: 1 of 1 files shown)


lang/python/python-babel/Makefile
```diff
@@ -0,0 +1,50 @@
...
+  URL:=https://babel.pocoo.org/en/latest/
+  DEPENDS:= \
+    +python3-cgi \
+    +python3-decimal \
+    +python3-distutils \
+    +python3-email \
+    +python3-light \
+    +python3-pytz \
+    +python3-urllib \
+    +python3-xml
```
jefferyto, line 39: Looks like the xml module is only used in [scripts/import_cldr.py](https://github.com/python-babel/babel/blob/v2.8.0/scripts/import_cldr.py#L21-L24) which appears to be used during compile-time and not run-time, so I don't think `+python3-xml` is necessary here.

(urllib is only explicitly imported in [scripts/download_import_cldr.py](https://github.com/python-babel/babel/blob/v2.8.0/scripts/download_import_cldr.py#L10-L13), but because email depends on urllib (#12942) I think it is okay to keep this dependency.)

jefferyto (commented):
The [installation instructions](http://babel.pocoo.org/en/latest/installation.html#living-on-the-edge) mention running `python setup.py import_cldr` (this is where the download_import_cldr.py and import_cldr.py scripts are used), are you sure this isn't necessary here?

lang/python/python-babel/Makefile
BKPepe: Thanks @jefferyto. You are right. `python3-xml` is not necessary.

commodo (approved):

PR #13693, opened 2020-10-17
Add Python package - Tornado

BKPepe:
Maintainer: me
Run tested: with example Hello, word - https://pypi.org/project/tornado/

Description:
- Adds new package Tornado (could be in the future dependency for Salt)

### DIFF
--- a/lang/python/python-tornado/Makefile
+++ b/lang/python/python-tornado/Makefile
@@ -1,0 +1,53 @@
+#
+# Copyright (C) 2020 CZ.NIC z.s.p.o. (https://www.nic.cz/)
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=python-tornado
+PKG_VERSION:=6.0.4
+PKG_RELEASE:=1
+
+PYPI_NAME:=tornado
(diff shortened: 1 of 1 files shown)


BKPepe:
cc: @jefferyto and @commodo

lang/python/python-tornado/Makefile
```diff
@@ -0,0 +1,54 @@
...
+define Package/python3-tornado
+  SECTION:=lang
+  CATEGORY:=Languages
+  SUBMENU:=Python
+  TITLE:=Web framework and asynchronous networking library
+  URL:=https://tornadoweb.org
+  DEPENDS:= \
+    +python3-asyncio \
+    +python3-codecs \
+    +python3-ctypes \
```
jefferyto, line 34: ctypes appears to only be imported in [tornado/platform/windows.py](https://github.com/tornadoweb/tornado/blob/v6.0.4/tornado/platform/windows.py#L4-L5), which I believe is only used when the OS is Windows, so I don't think `+python3-ctypes` is necessary.

BKPepe: 🤦 You are correct.

jefferyto (approved):
Thanks! (Don't forget to squash the commits before merging :joy:)

commodo (approved):

PR #13706, opened 2020-10-18
acme: fix the 'Unknown parameter' problem caused by acme_server

Meano:
Maintainer: @tohojo
Compile tested: ramips, mt7621-xiaomi_mir3g, Openwrt master
Run tested: ramips, mt7621-xiaomi_mir3g, Openwrt master, test done

Description: Recently commit for 'acme_server' shell script change will caused acme.sh 'Unknown parameter' problem when $acme_server is void value, this commit fix it.

### DIFF
--- a/net/acme/Makefile
+++ b/net/acme/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=acme
 PKG_VERSION:=2.8.7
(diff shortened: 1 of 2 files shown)


tohojo:
Ah, damn, should have spotted that! Thanks for the fix.

Please fix your DCO signoff, and also bump the package revision.

PR #13709, opened 2020-10-19
python3: bump to version 3.9.0

commodo:
Maintainer: me, @jefferyto 
Compile tested: x86 https://github.com/openwrt/openwrt/commit/0830ae3a2fdf444380153d311a7de167b92a68af
Run tested: x86  https://github.com/openwrt/openwrt/commit/0830ae3a2fdf444380153d311a7de167b92a68af

----------------------------------------------

Refreshed patches.

Dropped 'patches-setuptools/004-site-patch.patch'
Does not apply anymore. Setuptools has removed site.py support:
   https://github.com/pypa/setuptools/issues/2165
If this is still needed, we may need to re-think it's implementation.


lang/python/python3/patches/016-adjust-config-paths.patch
```diff
@@ -20,16 +20,16 @@
...
  			esac; \
  		done; \
  	done
 -	$(INSTALL_DATA) `cat pybuilddir.txt`/_sysconfigdata_$(ABIFLAGS)_$(MACHDEP)_$(MULTIARCH).py \
 +	$(INSTALL_DATA) `cat pybuilddir.txt`/_sysconfigdata*.py \
  		$(DESTDIR)$(LIBDEST); \
  	$(INSTALL_DATA) $(srcdir)/LICENSE $(DESTDIR)$(LIBDEST)/LICENSE.txt
- ifeq (@COMPILE_ALL_TESTS@,yes)
-@@ -1618,7 +1618,7 @@ sharedinstall: sharedmods
+ 	if test -d $(DESTDIR)$(LIBDEST)/distutils/tests; then \
```
jefferyto, line 31: Was this change intentional?

commodo: umm, let me re-apply it;
the patch context may have changed;

commodo: ok; so it's semi-intentional; i'll update this;
this shows up because i applied this patch on a Python-3.9 folder without applying the previous patches; so the context did not update;

commodo: thanks for catching this :)

commodo:
Changelog v1 -> v2:
* refreshed patch `016-adjust-config-paths.patch` with `make package/python3/refresh` ; it was manually generated on a Python-3.9 tree and not refreshed/re-applied in the series; so the context differs a bit from the original

jefferyto (approved):
Thanks for handling this :joy:

commodo:
> Thanks for handling this joy

I was going to get through all of them, but you beat me to the others.
And some days are slower than others.

neheb:
Breaks https://downloads.openwrt.org/snapshots/faillogs/mips64_octeonplus/packages/python-intelhex/compile.txt

jefferyto:
A fix has been merged upstream (https://github.com/python-intelhex/intelhex/pull/46), unsure when there will be a new release (https://github.com/python-intelhex/intelhex/issues/45#issuecomment-712410569) - ping @Cynerd

PR #13719, opened 2020-10-19
zoneinfo: Updated to the latest release

Wedmer:
Maintainer: me
Compile tested: TI OMAP3/4/AM33xx, default, OpenWRT/LEDE master
Run tested: n/a

Notes:

    I've updated build parameters to build in TZif "fat" format. 
    If musl will implement support for TZif "slim" format we should switch to default compile options. 
    "Slim" format significantly reduces size of zoneinfo packages.

Description:

   Briefly:

     Fiji starts DST later than usual, on 2020-12-20.

   Changes to future timestamps

     Fiji will start DST on 2020-12-20, instead of 2020-11-08 as
     previously predicted.  DST will still end on 2021-01-17.
     (Thanks to Raymond Kumar and Alan Mintz.)  Assume for now that
     the later-than-usual start date is a one-time departure from the
     recent pattern.

   Changes to build procedure

     Rearguard tarballs now contain an empty file pacificnew.
     Some older downstream software expects this file to exist.
     (Problem reported by Mike Cullinan.)

### DIFF
--- a/utils/zoneinfo/Makefile
+++ b/utils/zoneinfo/Makefile
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2007-2019 OpenWrt.org
+# Copyright (C) 2007-2020 OpenWrt.org
 #
 # This is free software, licensed under the GNU General Public License v2.
 # See /LICENSE for more information.
@@ -9,7 +9,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=zoneinfo
-PKG_VERSION:=2020a
+PKG_VERSION:=2020c
 PKG_RELEASE:=1
 
 #As i couldn't find real license used "Public Domain"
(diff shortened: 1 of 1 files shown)


neheb:
ping @hnyman

hnyman:
I will test this tonight.

(There is already one positive review in the forum discussion: https://forum.openwrt.org/t/timezone-etc-tz-not-updated-correctly/77002/14 )

hnyman:
@neheb 
Would it be possible for you to test 2020b (or 2020c without "fat" option) with musl 1.2.1 ?
It is possible that the slim/fat issue has already been fixed in musl 1.2 series, as 64bit time handling has been improved in musl. (However, I did not noticed anything in musl changelog mentioning slim/fat zoneinfo issues)

hnyman:
Seems to work. Thanks.

PR #13723, opened 2020-10-19
transmission: remove variants and use libcurl's TLS library

neheb:
Allows the Makefile to be cleaned up and to have fewer dependencies.
There's no need for multiple TLS libraries to be installed.

Maintainer: me
Compile tested: powerpc

Alternative to https://github.com/openwrt/packages/pull/13609

@ja-pa please review.

### DIFF
--- a/net/transmission/Makefile
+++ b/net/transmission/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=transmission
 PKG_VERSION:=3.00
-PKG_RELEASE:=6
+PKG_RELEASE:=7
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@GITHUB/transmission/transmission-releases/master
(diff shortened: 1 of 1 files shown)


net/transmission/Makefile
```diff
@@ -34,93 +40,52 @@ define Package/transmission/template
...
-  TITLE+= (with mbed TLS)
-  DEPENDS+=+libmbedtls
-  VARIANT:=mbedtls
+  TITLE+= (remote)
 endef
 
 define Package/transmission-web
   $(call Package/transmission/template)
   TITLE+= (webinterface)
-  DEPENDS:=@(PACKAGE_transmission-daemon-openssl||PACKAGE_transmission-daemon-mbedtls)
```
ja-pa, line 90: Would it be possible to transmission-daemon as transmission-web dependency ?

net/transmission/Makefile
```diff
@@ -23,6 +23,12 @@ PKG_CPE_ID:=cpe:/a:transmissionbt:transmission
 PKG_FIXUP:=autoreconf
 PKG_INSTALL:=1
 PKG_BUILD_PARALLEL:=1
+PKG_CONFIG_DEPENDS:= \
+	CONFIG_LIBCURL_GNUTLS \
+	CONFIG_LIBCURL_MBEDTLS \
+	CONFIG_LIBCURL_OPENSSL \
+	CONFIG_LIBCURL_WOLFSSL \
+	CONFIG_LIBCURL_NOSSL
```
BKPepe, line 31: Not sure if we want to have possibility to have transmission variant without SSL...

ja-pa: libmbedtls should be used in case of LIBCURL_NOSSL see https://github.com/openwrt/packages/pull/13723/files#diff-3dc735e7e3d28d09b2f110d6d92050af27a448236d24f10ab3984074498f1fd9R43

net/transmission/Makefile
neheb: Sure.

net/transmission/Makefile
neheb: Those are just configuration depends. In the case of nossl and gnutls, libmbedtls is used.

neheb:
ping @mstorchak what do you think?

net/transmission/Makefile
ja-pa: Something like this would be great DEPENDS:=+PACKAGE_transmission-daemon

PR #13727, opened 2020-10-20
modemmanager: include all init commands in the procd instance

aleksander0m:
If procd relaunches the ModemManager daemon after e.g. a crash, we
also want it to notify all cached hotplug events, or otherwise we
would end up leaving the daemon running without the full initial
processing done.

This change modifies the init script to include all the required init
commands as part of the procd instance command, so that procd launches
all of them on every respawn.

Maintainer: @nickberry17

### DIFF
--- a/net/modemmanager/Makefile
+++ b/net/modemmanager/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=modemmanager
 PKG_VERSION:=1.14.6
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=ModemManager-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://www.freedesktop.org/software/ModemManager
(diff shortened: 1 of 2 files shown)


aleksander0m:
I had mixed feelings about how to do this. I initially thought of creating a whole new "launcher" script, e.g. in  `/usr/share/ModemManager/modemmanager.run`, that would take care of running all required commands. But the solution with `sh -c` seems to work as well, and avoids needing an extra file around.

aleksander0m:
I'm seeing issues with these changes, so they're not good to go I'm afraid.
For some reason there are cases where the initial port events are now missing, I don't understand yet how or why...

aleksander0m:
> I'm seeing issues with these changes, so they're not good to go I'm afraid.
> For some reason there are cases where the initial port events are now missing, I don't understand yet how or why...

Ok, nevermind, the issue I was seeing was due to a different problem: https://github.com/openwrt/packages/pull/13813

aleksander0m:
Rebased these changes on top of latest git master now

aleksander0m:
@neheb @nickberry17 any issue with these changes?

PR #13729, opened 2020-10-20
knot: update to version 3.0.1

Payne-X6:
Signed-off-by: Jan Hak <jan.hak@nic.cz>

Compile tested: ARM Cortex-A9 vfpv3, Turris Omnia, TurrisOS 5.2 (hbl)
Run tested: ARM Cortex-A9 vfpv3, Turris Omnia, TurrisOS 5.1.2 (hbk), all tests successful (1 skipped)

### DIFF
--- a/libs/boost/Makefile
+++ b/libs/boost/Makefile
@@ -13,7 +13,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=boost
 PKG_VERSION:=1.74.0
 PKG_SOURCE_VERSION:=1_74_0
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
 PKG_SOURCE:=$(PKG_NAME)_$(PKG_SOURCE_VERSION).tar.bz2
 PKG_SOURCE_URL:=@SF/$(PKG_NAME)/$(PKG_NAME)/$(PKG_VERSION) https://dl.bintray.com/boostorg/release/$(PKG_VERSION)/source/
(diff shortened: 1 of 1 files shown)


hnyman:
You forgot to explain your other changes beside the version bump:

* why libnghttp2 dependency is now needed?  Could it be patched away instead?
* is -DPSELECT_COMPAT config option deprecated, or why is it removed?

salzmdan:
- `PSELECT_COMPAT` was removed many years ago https://gitlab.nic.cz/knot/knot-dns/-/commit/c03c51cf02d745993f03d997bc99768fab87676b
- `libnghttp2` is needed for DoH support in utilities (kdig) https://www.knot-dns.cz/docs/3.0/html/migration.html#building-notes

ja-pa:
I think that it should be enough to update the commit message with notes from @dsalzman.

<personal thoughts>
I think that it's a non-written rule to describe more briefly changes in package updates especially in case of dependency change because it can help maintainers with debugging build errors. 

(maybe it should be added to https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#commits-in-your-pull-requests-should) 
</personal thoughts>

salzmdan:
I don't understand why `Package knot is missing dependencies for the following libraries: libnghttp2.so.14` . Because `knot` shouldn't be linked against this library. But don't like the idea adding this dependency for knot server. So I propose removing DoH support from `kdig` at all. It's rather a question to @ja-pa what is the preferred variant?

PR #13730, opened 2020-10-20
boost: fix boost-regex compiling failure on targets of uClibc

brvphoenix:
Maintainer: @ClaymorePT @neheb
Environment: compiled using the SDK of archs38

Description:
Boost-regex requires the ICU to support unicode conversion which can be disabled by `--disable-icu`. Enable the icu support when `BUILD_NLS` is set.

Refer to: https://www.boost.org/doc/libs/1_61_0/libs/regex/doc/html/boost_regex/install.html

Signed-off-by: Van Waholtz <vanwaholtz@gmail.com>

### DIFF
--- a/libs/boost/Makefile
+++ b/libs/boost/Makefile
@@ -13,7 +13,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=boost
 PKG_VERSION:=1.74.0
 PKG_SOURCE_VERSION:=1_74_0
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
 PKG_SOURCE:=$(PKG_NAME)_$(PKG_SOURCE_VERSION).tar.bz2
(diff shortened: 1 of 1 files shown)


neheb:
Hmmm that’s interesting. So BUILD_NLS is normally used for switching between libiconv-stub and libiconv-full. I guess this is a fine usecase.

brvphoenix:
> Hmmm that’s interesting. So BUILD_NLS is normally used for switching between libiconv-stub and libiconv-full. I guess this is a fine usecase.

Should I add a configurable option to enable the ICU for boost-regex? It seems to be more appropriate in this case. Or only need to enable icu by default which is redundant when no unicode support is required.

I think, whether the ICU should be enabled depends on the packages that using boost-regex. I just can't find a variable which can enable the unicode support globally.

neheb:
I'd unconditionally enable it. Size is not an issue when it comes to boost I don't think.

neheb:
Needs a PKG_RELEASE bump.

ClaymorePT:
Thank both of you for the contributions!

PR #13732, opened 2020-10-20
[RFC] package gnupg2, libgpgme{,pp} and podman

dangowrt:
Maintainer: me
Compile tested: x86/64
Run tested: -

Description:
Make libgpgme{,pp} available.

### DIFF
--- a/libs/gpgme/Makefile
+++ b/libs/gpgme/Makefile
@@ -1,0 +1,125 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gpgme
+PKG_VERSION:=1.14.0
+PKG_RELEASE:=1
+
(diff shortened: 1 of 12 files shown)


aparcar:
A containers.conf should be added which automatically uses the correct `engine` and `cni` plugin path:

```ini
[containers]
netns="private"

[network]
cni_plugin_dir="/usr/lib/cni"
network_config_dir="/etc/cni/net.d/"
default_network="podman"

[engine]
runtime="/sbin/uxc"
```

neheb:
@dangowrt podman needs PKG_USE_MIPS16:=0

https://downloads.openwrt.org/snapshots/faillogs/mips64_octeonplus/packages/podman/compile.txt

Issue #13195, opened 2020-08-23
mwan3: VPN tunnels over PPTP/L2TP/PPPoE interfaces are broken


jamesmacwhite:
Environment: OpenWrt 19.07.3 Linksys WRT3200ACM (mwan3 2.9.0)
FAO: @aaronjg 

**It has been found this isn't directly a mwan3 problem, but mwan3 triggers it due to the use of fwmark. Further details and explanation here: https://github.com/openwrt/packages/issues/13195#issuecomment-680295448**

**Original issue description:**

@aaronjg Continuing the discussion from the comment here: https://github.com/openwrt/packages/pull/13169#issuecomment-678830151.

After @openwrt-diy reported the issues on L2TP, I decided to test out L2TP and mwan3 myself. An ISP in the UK offers a L2TP transit service here: https://www.aa.net.uk/broadband/l2tp-service/. It has full dual stack. A fully routed static IPv4 and delegated IPv6 prefix. They also publish specific information on configuring with OpenWrt: https://support.aa.net.uk/L2TP_Client:_OpenWRT

I ended up testing it out. Some of the issues encountered with mwan3 enabled are known and likely related to https://github.com/openwrt/packages/issues/10712 and https://github.com/openwrt/packages/issues/13145, which have a pending PR, but one area I did encounter was mwan3 appears to be interfering with being able to configure IPv6 on the L2TP interface.

Andrews & Arnold use PPP and DHCPv6 to establish the L2TP tunnel for IPv4 and IPv6. It's worth noting this doesn't use IPSec, it's L2TPv2 over UDP.

The setup is two interfaces `aaisp` and `aaisp6`. Like the example information, the IPv4 and IPv6 interfaces are split, but the underlying network interface `l2tp-aaisp` is dual stack. You can also make `aaisp6` an alias interface (which I have done). With mwan3 enabled the DHCPv6 client doesn't pick up the IPv6 configuration, if I disable mwan3 and restart the `aaisp6` interface IPv6 will then be configured.

With mwan3 enabled it seems it can only establish the IPv4 interface not IPv6. It looks like DHCPv6 related communication is either not going over the right interface or the packets are being blocked as the IPv6 side is not established.

As noted, there are rules preventing icmp6-type 133-137 from being interfered with, but I don't know if there's anything else around L2TP or DHCPv6 that might need to be looked at in this regard, as currently with mwan3 enabled the tunnel cannot even be fully established.

aaronjg:
[...]

PR #13748, opened 2020-10-22
mwan3: fix mwan3 'use'

aaronjg:

Maintainer: @feckert

Description:

### DIFF
--- a/net/mwan3/files/usr/sbin/mwan3
+++ b/net/mwan3/files/usr/sbin/mwan3
@@ -123,7 +123,7 @@ restart() {
 	/etc/init.d/mwan3 start
 }
 
-wrap() {
+use() {
 	# Run a command with the device, src_ip and fwmark set to avoid processing by mwan3
 	# firewall rules
 
@@ -132,7 +132,7 @@ wrap() {
 	config_load mwan3
 
 	interface=$1 ; shift
-	[ -z "$*" ] && echo "no command specified for mwan3 wrap" && return
+	[ -z "$*" ] && echo "no command specified for mwan3 use" && return
 	network_get_device device $interface
 	[ -z "$device" ] && echo "could not find device for $interface" && return
 
--- a/net/mwan3/files/usr/sbin/mwan3rtmon
+++ b/net/mwan3/files/usr/sbin/mwan3rtmon
@@ -74,7 +74,7 @@ mwan3_rtmon_route_handle()
 	route_line=${1##"Deleted "}
 	route_family=$2
 
-	config_get_boolean source_routing globals source_routing 0
+	config_get_bool source_routing globals source_routing 0
 	[ $source_routing -eq 0 ] && unset source_routing
 
 	if [ "$route_line" = "$1" ]; then
@@ -107,22 +107,22 @@ mwan3_rtmon_route_handle()
 		local iface=$1
 		tbl=$($IP route list table $tid 2>/dev/null)$'\n'
 
-		if [ "$(cat /var/run/mwan3track/$iface/STATUS)" != "online" ]; then
-			LOG debug "interface $iface is offline - skipping $route_line";
+		if [ -n "$iface" ] && [ "$(mwan3_get_mwan3track_status $iface)" != "active" ]; then
+			LOG debug "interface $iface is disabled - skipping '$route_line'";
 			return
 		fi
 
 		# check that action needs to be performed. May not need to take action if we
 		# got a delete event, but table was already flushed
 		if [ $action = "del" ] && [ -n "${tbl##*$route_line$'\n'*}" ]; then
-			LOG debug "skipping already deleted route table $tid - skipping $route_line"
+			LOG debug "skipping already deleted route table $tid - skipping '$route_line'"
 			return
 		fi
 
 		network_get_device device "$iface"
-		LOG debug "adjusting route $device: $IP route $action table $tid $route_line"
+		LOG debug "adjusting route $device: '$IP route $action table $tid $route_line'"
 		$IP route "$action" table $tid $route_line ||
-			LOG warn "failed: $IP route $action table $tid $route_line"
+			LOG warn "failed: '$IP route $action table $tid $route_line'"
 	}
 	handle_route_cb(){
 		local iface=$1
@@ -163,9 +163,9 @@ main()
 	sh -c "echo \$\$; exec $IP monitor route" | {
 		read -r monitor_pid
 		trap_with_arg func_trap "$monitor_pid" SIGINT SIGTERM SIGKILL
-		while read -r line; do
+		while IFS='' read -r line; do
 			[ -z "${line##*table*}" ] && continue
-			LOG debug "handling route update $family $line"
+			LOG debug "handling route update $family '$line'"
 			mwan3_lock "service" "mwan3rtmon"
 			mwan3_rtmon_route_handle "$line" "$family"
 			mwan3_unlock "service" "mwan3rtmon"
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.10.0
+PKG_VERSION:=2.10.1
 PKG_RELEASE:=1
 PKG_MAINTAINER:=Florian Eckert <fe@dev.tdt.de>
 PKG_LICENSE:=GPL-2.0
--- a/net/mwan3/files/etc/init.d/mwan3
+++ b/net/mwan3/files/etc/init.d/mwan3
@@ -111,6 +111,11 @@ stop_service() {
 	mwan3_unlock "command" "mwan3"
 }
 
+reload_service() {
+	stop
+	start
+}
+
 service_triggers() {
 	procd_add_reload_trigger 'mwan3'
 }
--- a/net/mwan3/files/lib/mwan3/common.sh
+++ b/net/mwan3/files/lib/mwan3/common.sh
@@ -23,6 +23,9 @@ MMX_UNREACHABLE=""
 MM_UNREACHABLE=""
 MAX_SLEEP=$(((1<<31)-1))
 
+command -v ip6tables > /dev/null
+NO_IPV6=$?
+
 LOG()
 {
 	local facility=$1; shift
--- a/net/mwan3/files/lib/mwan3/mwan3.sh
+++ b/net/mwan3/files/lib/mwan3/mwan3.sh
@@ -24,9 +24,6 @@ IPv4_REGEX="((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[0
 
 DEFAULT_LOWEST_METRIC=256
 
-command -v ip6tables > /dev/null
-NO_IPV6=$?
-
 mwan3_push_update()
 {
 	# helper function to build an update string to pass on to
--- a/net/mwan3/files/usr/libexec/rpcd/mwan3
+++ b/net/mwan3/files/usr/libexec/rpcd/mwan3
@@ -5,8 +5,6 @@
 . /usr/share/libubox/jshn.sh
 . /lib/mwan3/common.sh
 
-MWAN3TRACK_STATUS_DIR="/var/run/mwan3track"
-
 IPS="ipset"
 IPT4="iptables -t mangle -w"
 IPT6="ip6tables -t mangle -w"
@@ -22,6 +20,7 @@ report_connected_v4() {
 }
 
 report_connected_v6() {
+	[ $NO_IPV6 -ne 0 ] && return
 	local address
 
 	if [ -n "$($IPT6 -S mwan3_connected 2> /dev/null)" ]; then
@@ -60,6 +59,7 @@ report_policies_v4() {
 }
 
 report_policies_v6() {
+	[ $NO_IPV6 -ne 0 ] && return
 	local policy
 
 	for policy in $($IPT6 -S | awk '{print $2}' | grep mwan3_policy_ | sort -u); do
--- a/net/mwan3/files/usr/sbin/mwan3track
+++ b/net/mwan3/files/usr/sbin/mwan3track
@@ -345,8 +345,12 @@ main() {
 		get_uptime > $MWAN3TRACK_STATUS_DIR/$INTERFACE/TIME
 
 		host_up_count=0
-		sleep "${sleep_time}" &
-		wait
+		if [ "${IFDOWN_EVENT}" -ne 0 ] && [ "${IFUP_EVENT}" -ne 0 ]; then
+			sleep "${sleep_time}" &
+			SLEEP_PID=$!
+			wait
+			unset SLEEP_PID
+		fi
 
 		if [ "${IFDOWN_EVENT}" -eq 1 ]; then
 			LOG debug "Register ifdown event on interface ${INTERFACE} (${DEVICE})"


feckert:
@aaronjg we must have overlooked that!

I have seen that we should also change the line:
`files/usr/sbin/mwan3:135:	[ -z "$*" ] && echo "no command specified for mwan3 wrap" && return`

to

`files/usr/sbin/mwan3:135:	[ -z "$*" ] && echo "no command specified for mwan3 use" && return`


@ptpt52 thanks for reporting
The name was changed from wrap to use shortly before the merge.

net/mwan3/files/usr/sbin/mwan3
```diff
@@ -132,7 +132,7 @@ wrap() {
 	config_load mwan3
 
 	interface=$1 ; shift
-	[ -z "$*" ] && echo "no command specified for mwan3 wrap" && return
+	[ -z "$*" ] && echo "no command specified for mwan3 wuse" && return
```
feckert, line 135: I think you mean `use` ;-)

aaronjg: Yes 🤓

feckert:
@aaronjg i have added the reload handling to your P/R.
I think this is now the fastest solution to get the mwan3 ready for the next release.
If there is still time we can think of a smarter solution.
@ptpt52 could you please verify if this commits fixes your issues?

ptpt52:
see on the code, reload should work, I will test it today

aaronjg:
> i have added the reload handling to your P/R.
I think this is now the fastest solution to get the mwan3 ready for the next release.
If there is still time we can think of a smarter solution.

👍 Sounds good to me

ptpt52:
tested and mwan3 use work now
and change config via luci web then it works

but I found a new bug
route table not copy from main table to other tables

ip route add 1.2.3.4 via 192.168.101.1
and then
ip route show table 1
ip route show table 2

no route for 1.2.3.4 found on table 1/2

ptpt52:
```
root@X-WRT:~# ip r
default via 192.168.18.1 dev pppoe-wan proto static metric 40 
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
1.2.3.4 via 192.168.18.1 dev pppoe-xwan01 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.22 
192.168.18.1 dev pppoe-wan proto kernel scope link src 192.168.18.24 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.23 

root@X-WRT:~# ip r delete 1.2.3.4

root@X-WRT:~# ip r s t 1
default via 192.168.18.1 dev pppoe-wan proto static metric 40 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-wan proto kernel scope link src 192.168.18.24 

root@X-WRT:~# ip r s t 2
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
1.2.3.4 via 192.168.18.1 dev pppoe-xwan01 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.22 

root@X-WRT:~# ip r s t 3
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.23 
```

ptpt52:
```
root@X-WRT:~# ip r
default via 192.168.18.1 dev pppoe-wan proto static metric 40 
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.22 
192.168.18.1 dev pppoe-wan proto kernel scope link src 192.168.18.24 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.23 
root@X-WRT:~# ip route add 1.2.3.4 via 192.168.18.1
root@X-WRT:~# ip r
default via 192.168.18.1 dev pppoe-wan proto static metric 40 
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
1.2.3.4 via 192.168.18.1 dev pppoe-xwan01 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.22 
192.168.18.1 dev pppoe-wan proto kernel scope link src 192.168.18.24 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.23 
root@X-WRT:~# ip r s t 1
default via 192.168.18.1 dev pppoe-wan proto static metric 40 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-wan proto kernel scope link src 192.168.18.24 
root@X-WRT:~# ip r s t 2
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
1.2.3.4 via 192.168.18.1 dev pppoe-xwan01 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.22 
root@X-WRT:~# ip r s t 3
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.23 
```

ptpt52:
I add a fix to this  https://github.com/openwrt/packages/pull/13756

ptpt52:
and the correct test resault:
```
root@X-WRT:~# ip r
default via 192.168.18.1 dev pppoe-wan proto static metric 40 
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
1.2.3.4 via 192.168.18.1 dev pppoe-xwan01 
4.3.2.1 via 192.168.18.1 dev pppoe-wan 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.22 
192.168.18.1 dev pppoe-wan proto kernel scope link src 192.168.18.24 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.23 
root@X-WRT:~# ip r s t 1
default via 192.168.18.1 dev pppoe-wan proto static metric 40 
1.2.3.4 via 192.168.18.1 dev pppoe-xwan01 
4.3.2.1 via 192.168.18.1 dev pppoe-wan 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-wan proto kernel scope link src 192.168.18.24 
root@X-WRT:~# ip r s t 2
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
1.2.3.4 via 192.168.18.1 dev pppoe-xwan01 
4.3.2.1 via 192.168.18.1 dev pppoe-wan 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.22 
root@X-WRT:~# ip r s t 3
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
1.2.3.4 via 192.168.18.1 dev pppoe-xwan01 
4.3.2.1 via 192.168.18.1 dev pppoe-wan 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.23 
```

ptpt52:
I found more bugs

when I change network interface `wan` from pppoe to static ip
mwan3 not work, status go wrong

ptpt52:
> I found more bugs
> 
> when I change network interface `wan` from pppoe to static ip
> mwan3 not work, status go wrong

I may found the reason in `mwan3track`

when change the network interface `wan` from pppoe to static
DEVICE would change name
but the WRAP use old DEVICE so it ping failed

maybe the `mwan3track` should restart/reload config when `wan` change network settings

ptpt52:
Maybe there is no strong reason to use procd script
I prefer the old scripts in version 2.9.0
where mwan3track would restart process when any related iface up/down.

aaronjg:
> when change the network interface wan from pppoe to static
DEVICE would change name
but the WRAP use old DEVICE so it ping failed

It is not surprising that when you change some configurations, some services may need to be restarted. Nevertheless, it would be better if this is handled automatically. I believe we can also add some triggers to procd so that this happens. I added a commit for this. Can you please test?


> Maybe there is no strong reason to use procd script
I prefer the old scripts in version 2.9.0

We would have to go back  than 2.9.0 to remove procd. It was changed in version 2.8.6 here: a6dc75428c1e3a47700b5c164a16385b5c640b48.


Does your routing issue still appear after you have restarted mwan3 after the network change? If so con you please provide more information, since I can not reproduce this. The commands you describe do not appear to create the routes you are showing in the table. What are the actual commands you are running?

> ip route add 1.2.3.4 via 192.168.101.1

Does not produce the route line:
`1.2.3.4 via 192.168.18.1 dev pppoe-xwan01 ` Maybe you have some extra arguments?

Can you also share the debug logs (enable debug logging in common.sh) and your complete mwan3 configuration?

Also, a small suggestion: can you include the complete "ip route" commands rather than the one-letter shorthand when posting here, it makes it easier to follow what you are doing.

ptpt52:
Just think about these situations
1. if the main table is bellow, what should table 1,2 be?
```
main:
default via 192.168.18.1 dev pppoe-wan proto static metric 40 
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.22 
192.168.18.1 dev pppoe-wan proto kernel scope link src 192.168.18.24 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.23 
```

2. if the main table is bellow, what should table 1,2 be?
```
default via 192.168.18.1 dev pppoe-wan proto static metric 40 
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
192.168.18.0/24 dev pppoe-wan proto kernel scope link src 192.168.18.24 metric 40
192.168.18.0/24 dev pppoe-xwan01 proto kernel scope link src 192.168.18.22  metric 41
192.168.18.0/24 dev pppoe-xwan02 proto kernel scope link src 192.168.18.23 metric 42
```

ptpt52:
@aaronjg  I could give you ssh access to do debug my router if you need?
and you will see what the problem is.

ptpt52:
>  I added a commit for this. Can you please test?


I think the reload_trigger on config `network` is not a good idea, it looks like a hack fix
we have to do the things in hotplug event framwork

that is:
when catch ifup/ifdown event, mwan3track on that iface should reload config and restart

ptpt52:
The previous version is quite perfect, at least logically complete, we should keep the design consistent with the previous version

what `mwan3track` do is watching the specific interface, if the iface down, the mwan3track could stop/pause or exit
and if the iface up again, we call mwan3track backup via hotplug event.

aaronjg:
> 2. if the main table is bellow, what should table 1,2 be?
> ...
> 192.168.18.0/24 dev pppoe-xwan01 proto kernel scope link src 192.168.18.22  metric 41

That line looks suspicious. If there are two spaces between src and metric, I see how this could cause a problem. However, I'm not sure how that line appeared. Nevertheless, I think the issue is that newer iproute2 versions insert a trailing space at the end of the line, so I added a change to fix this.

> @aaronjg I could give you ssh access to do debug my router if you need?
and you will see what the problem is.

If more debugging info does not help, we can go that route, however I'd prefer to go over the logs here. That way, it will be clear to other developers what the issue is, so that we can avoid regressions in the future.

> I think the reload_trigger on config network is not a good idea, it looks like a hack fix
we have to do the things in hotplug event framwork

The reload trigger is just hit when there is a configuration change in `network`, and not when a network comes up or down. Since most network changes could impact the entire mwan3 configuration, it makes sense to reload `mwan3` on network changes. This does something different from the hotplug framework to detect ifup and ifdown devices.

> The previous version is quite perfect,

The previous procd implementation was rather incomplete, which caused issues with locking, and then there was a hack to get around this, so either mwan3 should be moved entirely off of procd, or should fully use the procd protocol. The mixed solution had its own issues.

> what mwan3track do is watching the specific interface, if the iface down, the mwan3track could stop/pause or exit
and if the iface up again, we call mwan3track back up via hotplug event.

This is what goes on - mwan3track is paused on ifdown and resumed on ifup. It appears that you have found some UCI changes are not propagated through, and I agree that they should be propagated, but if there is a UCI change, the change should be done through procd or ubus, and not via the hotplug scripts.

aaronjg:
Overall, it would be much better to handle everything that goes through `iproute2` as json with the `-j` command, and that will save a lot of headaches with trying to parse this stuff. I think that should be doable since mwan3 already depends on the `iproute2` rather than the busybox implementation. But that is a larger change and for a separate P/R.

net/mwan3/files/usr/sbin/mwan3rtmon
```diff
@@ -107,14 +107,20 @@ mwan3_rtmon_route_handle()
 		local iface=$1
 		tbl=$($IP route list table $tid 2>/dev/null)$'\n'
 
-		if [ "$(cat /var/run/mwan3track/$iface/STATUS)" != "online" ]; then
+		if [ -n "$iface" ] && [ "$(cat /var/run/mwan3track/$iface/STATUS)" != "online" ]; then
```
ptpt52, line 110: we should not check iface status here, even it is offline we should continue set the route, so remove the check is better.

net/mwan3/files/usr/sbin/mwan3rtmon
```diff
@@ -107,14 +107,20 @@ mwan3_rtmon_route_handle()
...
 		# check that action needs to be performed. May not need to take action if we
 		# got a delete event, but table was already flushed
-		if [ $action = "del" ] && [ -n "${tbl##*$route_line$'\n'*}" ]; then
+
+		# need to test with and without trailing space because some `iproute2` versions
+		# output a trailing space
+
+		if [ $action = "del" ] &&
+			   [ -n "${tbl##*$route_line$'\n'*}" ] &&
+			   [ -n "${tbl##*$route_line$' \n'*}" ]; then
```
ptpt52, line 123: You don't seems to get my point in my P/R for this bug, where in https://github.com/openwrt/packages/pull/13756/commits/ef853528dec0a4b89a4ead0039c252558d3ff663

what the problem is
`"${tbl##*$route_line$'\n'*}"` is diff from `${tbl##*$route_line$'\n'*}`
that is why you get wrong test resault on this line

and more changes on https://github.com/openwrt/packages/pull/13756/commits/ef853528dec0a4b89a4ead0039c252558d3ff663

you could see the detail, and I am sure my changes resolve the real problem.

ptpt52 (commented):
and
network config change is not the only reason to cause ifup/ifdown
and even, network config change not necessarily lead to ifup/ifdown

at these points , no reason to trigger reload mwan3 when network config changes

ptpt52:
@aaronjg @feckert 
good news
I found a good way to handle route table change
so that we can avoid the complicated route table monitor

we could use the `mwan3_connected` ipset
when any changes in main table
we sync all non-default route dst to ipset `mwan3_connected`
so it would work as expected

and we don't have to copy these route to table 1/2/3...

aaronjg:
> we should not check iface status here, even it is offline we should continue set the route, so remove the check is better.

Yes, that is correct. Thank you for the testing and pointing that out.


> You don't seems to get my point in my P/R for this bug, where in ef85352
> you could see the detail, and I am sure my changes resolve the real problem.

Yes, I don't totally understand your issue, that's why I was asking for more debug information. Your commit may solve your problem, but it introduces new bugs.

> we could use the mwan3_connected ipset

Again, I don't follow the change you are suggesting. The routes should already be added to the connected ipset:
https://github.com/openwrt/packages/blob/master/net/mwan3/files/usr/sbin/mwan3rtmon#L83

In your set up are the directly connected hosts not added to that ipset?

In principle, I see the point that these routes should not need to be added to the tables, since they should get handled by the directly connected rule. However there are so many different configurations out there and different idiosyncrasies with various setups that I am somewhat conservative about changing the routing tables. Let's first get the route add/delete working again as a baseline, then we can explore ways to simplify this further.

I have done some testing with this, and some of the routes are definitely necessary so that traffic gets routed back to the correct host on the LAN. Routes out the WAN may not need to be added, but we need more testing to make sure.

feckert:
@aaronjg @ptpt52 I think we should first merge this P/R and then we can concentrate on the [routing](https://github.com/openwrt/packages/pull/13756) problem. I always find it difficult when there are several P/Rs open that deal with the same problem.

@aaronjg The question is why we have to match after the [new line](https://github.com/openwrt/packages/pull/13748/files#diff-8a37c6770c4fcad6a416392044a5e3bc4a9624592ca319cb9a4036c60a49e7f5R121-R123). Wouldn't the @ptpt52 proposal be better than his [change](https://github.com/openwrt/packages/pull/13756/commits/ef853528dec0a4b89a4ead0039c252558d3ff663#diff-a79526f0a000ba74d1bdaee37e1c9544d0e215cf231a42ab37fa6d29d13c0f64R452)? Then we do not need to distinguish if we use the old or the new iproute2.

ptpt52:
I am not intend to get my P/R merge, and hope that @aaronjg can fix the issues and get this merge

my P/R https://github.com/openwrt/packages/pull/13756 just demonstrate the issue and how to fix it.

ptpt52:
for the ping -I device/IP issue, it is bug in wireguard, so I try to fix it in https://github.com/x-wrt/x-wrt/blob/master/package/network/services/wireguard/patches/100-scrub-skb-sk-before-send-out.patch

with this patch,  `ping -I device` work again in wireguard vpn

so we could roll back the mwan3track to use `ping -I device`
I am sure that `ping -I ip` is not correct and not works as expected

so the wrap is no needed at this scope, we could drop this hack.

ptpt52:
@aaronjg 
for the test case checklist:
```
1. ifconfig down/up some interface and see all route table work as expected?

2. change network interface in web luci from pppoe-wan to static ip, and see what happens to mwan3, works as expected?

3. in web ui network-interface page, restart one interface and see what happen to mwan3?


```

ptpt52:
test environment
```
PC---router---pppoe-server----network
```
pppoe-server(Ubuntu):
```
pppoe-server -I {DEVICE} -L 192.168.18.1 -R 192.168.18.10 -N 200
```

router:
```
wan1-static
wan2-pppoe
wan3-pppoe

ip route:
default via 192.168.17.1 dev wan proto static metric 40 
default via 192.168.18.1 dev pppoe-xwan01 proto static metric 41 
default via 192.168.18.1 dev pppoe-xwan02 proto static metric 42 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
192.168.18.1 dev pppoe-xwan01 proto kernel scope link src 192.168.18.80 
192.168.18.1 dev pppoe-xwan02 proto kernel scope link src 192.168.18.79 
```

aaronjg:
> @aaronjg The question is why we have to match after the [new line](https://github.com/openwrt/packages/pull/13748/files#diff-8a37c6770c4fcad6a416392044a5e3bc4a9624592ca319cb9a4036c60a49e7f5R121-R123). Wouldn't the @ptpt52 proposal be better than his [change](https://github.com/openwrt/packages/pull/13756/commits/ef853528dec0a4b89a4ead0039c252558d3ff663#diff-a79526f0a000ba74d1bdaee37e1c9544d0e215cf231a42ab37fa6d29d13c0f64R452)? Then we do not need to distinguish if we use the old or the new iproute2.

The new line match is required so that we match exactly the line in question, and not a prefix of that line. For example, without it, a route line like:
```
<ip> dev <dev> proto static scope link metric 5
```
Will match against
```
<ip> dev <dev> proto static scope link metric 50
```

I was hoping that we could use `ip -j` to do everything in json, but unfortunately ip monitor route does not output json so this will not be feasible.

ptpt52:
@aaronjg  I know that the new line match is needed
but this line `[ -n "${tbl##*$route_line$'\n'*}" ]` will match nothing
you know that it is different bw `"${tbl##*$route_line$'\n'*}"` and `${tbl##*$route_line$'\n'*}`

ptpt52:
maybe you could rewrite in this way, so that it works
```
route_line_re=$route_line$'\n'
[ -n "${tbl##*$route_line_re*}" ]
```

ptpt52:
> maybe you could rewrite in this way, so that it works
> 
> ```
> route_line_re=$route_line$'\n'
> [ -n "${tbl##*$route_line_re*}" ]
> ```

I try this but it still match anything, seems not work.

ptpt52:
if it is one space at the end of line, so we do not need to pad '\n', just pad a space and match the line
```
[ -n "${tbl##*$route_line *}" ]
```

ptpt52:
and if you need to pad '\n'
then must write the code in this way:
```
route_line_re="$route_line "$'\n'
[ -n "${tbl##*$route_line_re*}" ]
```
this cannot archive in one line like this:
```
[ -n "${tbl##*$route_line $'\n'*}" ]  #this not works as expected
```

net/mwan3/files/usr/sbin/mwan3rtmon
ptpt52: reload trigger on network cannot be accepted, I think.

ptpt52 (changes requested):

aaronjg:
I was able to reproduce the "mwan3rtmon" delete events not matching, and have an alternative fix, which is a bit cleaner than my prior fix.

I was not able to reproduce your issue with changing from DHCP to static IP. The only thing I could think of is perhaps it was a bit slower than before because it was waiting "$sleep_time" before registering the events. I have removed the network reload trigger, since it is not needed.

If you still have issues, please run the following script:
https://gist.github.com/aaronjg/5981991c1046e642358b0dff48fb1444

and upload the contents of /tmp/log_mwan3_diagnose.txt to a new gist.

feckert:
@ptpt52 could you be a little more precise in the error description.
It is difficult for us to understand the problem if we don't know what your setup and use case is like.
Since it is a community project and we have never seen ourselves in real life, communication is very important!
For the communication and the translation I can recommend this [tool](https://www.deepl.com/translator).

I would also appreciate it if you didn't just say **if I do it this way then it works**.
I think we should already use a common code base to fix the problem.
And not just revert the last version and then say with other added changes that now it works.

I wouldn't like it if we now have in the future 2 mwan3 versions.
Simply reversing the last hole version is not practical.
Nor do I want to, because we have now properly integrated mwna3 into the procd.

If then we can discuss the C-wrapper again.
@aaronjg maybe you can make clear why we chose the C-wrapper approach to @ptpt52.

ptpt52:
@feckert  for now I don't have to much time to do the test, and you think all test cases pass, you could get this merge.

I will re-do all test case and report again when I have time.

Besides, please make sure all test case I mention before is being tested and passed.

ptpt52:
for the `ping -I device` issue, it is going to fix in openwrt, you could track it here: https://github.com/openwrt/openwrt/pull/3541

and I think we could also drop all recent changes on mwan3
and start with the `ipset mwan3_connected` idea.

zx2c4:
> for the ping -I device issue, it is going to fix in openwrt, you could track it here: openwrt/openwrt#3541

What you're waiting for is https://lore.kernel.org/netdev/20201029025606.3523771-3-Jason@zx2c4.com/

aaronjg:
The issue with the 5.x kernel is that when a packet has a firewall mark, for another device, it is unable to find the gateway. EG if you have wwan as a secondary interface for eth0.2, then `ping -I wwan 1.1.1.1` will get the firewall mark for eth0.2, and will not be able to find wwan's gateway. I am able to reproduce this behavior with a router that has an ethernet connection and my wireless hotspot.

It seems that whatever changed in the 5.x kernel caused the issues that wackejohn had, but it is also caused issues with non tunnelling interfaces, so this is a larger problem than just the wireguard interfaces. From the description of  @zx2c4's patch, it sounds like it will fix the wireguard issue, and possibly #13195, but will not fix the gateway issue. 

@zx2c4, do you know why the route_me_harder issue did not come up in the 4.x kernels?

There are also issues with other tracking methods like `nping` and `httping` which do not bind to a device using SO_BINDTODEVICE, and set the source based on source ip address, which has caused issues before eg (#8139). Using the wrapper library allows us to handle the interface binding consistently across all tracking methods.

zx2c4:
> @zx2c4, do you know why the route_me_harder issue did not come up in the 4.x kernels?

As far as I can tell, my patch fixes code that goes back to v2.3.15 (that's not a typo), which was never adjusted to handle tunnels. The test case in the wireguard test suite shows that both 4.19 is broken in addition to 5.4. So I'm not sure which 5.x regression you're referring to; it's not the bug that I fixed and linked to in my last message.

ptpt52:
> if you have wwan as a secondary interface for eth0.2, then ping -I wwan 1.1.1.1 will get the firewall mark for eth0.2, and will not be able to find wwan's gateway. 

not true, if `ping -I device`, packet would goes out via the `device`, no metter what mark is set
I test these cases, and notice that the mark could be 0x100, 0x200, 0x300 but all go out via `device` as `ping -I device` request.

we could take a look at the kernel code
```
 259 static int fib_rule_match(struct fib_rule *rule, struct fib_rules_ops *ops,
 260               struct flowi *fl, int flags,
 261               struct fib_lookup_arg *arg)
 262 {
 263     int ret = 0;
 264 
 265     if (rule->iifindex && (rule->iifindex != fl->flowi_iif))
 266         goto out;
 267 
 268     if (rule->oifindex && (rule->oifindex != fl->flowi_oif))
 269         goto out;
 270 
 271     if ((rule->mark ^ fl->flowi_mark) & rule->mark_mask)
 272         goto out;
 273 
 274     if (rule->tun_id && (rule->tun_id != fl->flowi_tun_key.tun_id))
 275         goto out;
 276 
 277     if (rule->l3mdev && !l3mdev_fib_rule_match(rule->fr_net, fl, arg))
 278         goto out;
 279 
 280     if (uid_lt(fl->flowi_uid, rule->uid_range.start) ||
 281         uid_gt(fl->flowi_uid, rule->uid_range.end))
 282         goto out;
 283 
 284     ret = INDIRECT_CALL_MT(ops->match,
 285                    fib6_rule_match,
 286                    fib4_rule_match,

```

it shows that, even with mark set, the ip rule would not match if `rule->oifindex` not match
so if `ping -I device`, it will lookup the main table and find the correct gateway

ptpt52:
Sorry, misunderstanding, it's not a problem with this code

ptpt52:
take a lookup at ip_route_output_key_hash_rcu()
it seems that it get the route result force dev_out to by oif and not care match about the fib rules

aaronjg:
> > @zx2c4, do you know why the route_me_harder issue did not come up in the 4.x kernels?
> 
> As far as I can tell, my patch fixes code that goes back to v2.3.15 (that's not a typo), which was never adjusted to handle tunnels. The [test case in the wireguard test suite](https://git.zx2c4.com/wireguard-linux/commit/?id=909d3f944288f0e46dc021a9f1251587a8d1a616) shows that both 4.19 is broken in addition to 5.4. So I'm not sure which 5.x regression you're referring to; it's not the bug that I fixed and linked to in my last message.

Interesting - the issues didn't start showing up in mwan3 until later. One user reports it was working on 4.14 https://github.com/openwrt/packages/issues/10712#issuecomment-657357292

---

re @ptpt52 
> not true, if `ping -I device`, packet would goes out via the `device`, no metter what mark is set

Please be mindful of your tone. I have observed that the packets do not always leave the device, though it is possible that there is another root cause. However I can assure you the behavior is reproducible and real.

In my testing, if the packet has a firewall mark set so that the route table does not have a gateway for the default route for the device specified with `-I`, then linux will send out arp requests to try to route the packet, rather than routing it through the gateway. Because of this, the ICMP packet will never leave the device.

You can see the test script and the sample output here:
https://gist.github.com/aaronjg/13057c9fdfbb16f45cd71e687aed1dc0

When the gateway is present, the ping succeeds, when the gateway is absent the ping fails. If the gateway is present in the main routing table, but absent from the one in the auxiliary table that is used due to the firewall mark, the ping still fails.

I understand you are using several tunneling interfaces in your setup. If the tunneling interface does not require a gateway to be set, it is likely that you would not see this issue. However, many mwan3 users use it to load balance or failover between physical WANs so this is a very real issue.

zx2c4:
> > > @zx2c4, do you know why the route_me_harder issue did not come up in the 4.x kernels?
> > 
> > As far as I can tell, my patch fixes code that goes back to v2.3.15 (that's not a typo), which was never adjusted to handle tunnels. The [test case in the wireguard test suite](https://git.zx2c4.com/wireguard-linux/commit/?id=909d3f944288f0e46dc021a9f1251587a8d1a616) shows that both 4.19 is broken in addition to 5.4. So I'm not sure which 5.x regression you're referring to; it's not the bug that I fixed and linked to in my last message.
> 
> Interesting - the issues didn't start showing up in mwan3 until later. One user reports it was working on 4.14 https://github.com/openwrt/packages/issues/10712#issuecomment-657357292
>

Okay. If you have an issue that affects code I maintain, please send me a reproducer or detailed description (as ptpt52 did before). Otherwise I'll assume you're discussing something different that doesn't need my involvement, as seems to be the case here with mwan3.

ptpt52:
@aaronjg 

Please don't mind the tone. It's machine translation.
It's just a matter of fact.


> You can see the test script and the sample output here:
https://gist.github.com/aaronjg/13057c9fdfbb16f45cd71e687aed1dc0

my test result:
```
root@X-WRT:~# ls
test_route.sh
root@X-WRT:~# cat test_route.sh 
#!/bin/sh

DEV=wan
PING="/usr/bin/ping -I $DEV -c1 -W1 8.8.8.8"
PING_MARK="/usr/bin/ping -m99 -I $DEV -c1 -W1 8.8.8.8"

default_line=$(ip route list |grep "default.*dev $DEV")
no_gateway=$(echo $default_line|sed 's/via [^ ]* //')
echo "route table:"
ip route list
$PING

echo "replacing route"
ip route replace $no_gateway
ip route list
$PING

ip route replace $default_line

ip route list| while read -r line; do
        ip route add $line table 99
done

ip rule add pref 99 fwmark 99 lookup 99

echo "route rules"
ip rule list

echo "route table 99:"
ip route list table 99
$PING_MARK

echo "replacing route table 99"
ip route replace $no_gateway table 99
ip route list table 99
$PING_MARK

ip route flush table 99
ip rule del pref 99
root@X-WRT:~# sh test_route.sh 
route table:
default via 192.168.17.1 dev wan proto static metric 40 
default dev wg proto static scope link metric 43 
10.8.11.0/24 dev wg proto static scope link metric 43 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
PING 8.8.8.8 (8.8.8.8) from 192.168.17.254 wan: 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=250 time=6.50 ms

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 6.501/6.501/6.501/0.000 ms
replacing route
default dev wan proto static scope link metric 40 
default dev wg proto static scope link metric 43 
10.8.11.0/24 dev wg proto static scope link metric 43 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
PING 8.8.8.8 (8.8.8.8) from 192.168.17.254 wan: 56(84) bytes of data.

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 0 received, 100% packet loss, time 0ms

route rules
0:	from all lookup local
99:	from all fwmark 0x63 lookup 99
1001:	from all iif wan lookup 1
1004:	from all iif wg lookup 4
2001:	from all fwmark 0x100/0x3f00 lookup 1
2004:	from all fwmark 0x400/0x3f00 lookup 4
32766:	from all lookup main
32767:	from all lookup default
route table 99:
default via 192.168.17.1 dev wan proto static metric 40 
default dev wg proto static scope link metric 43 
10.8.11.0/24 dev wg proto static scope link metric 43 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
PING 8.8.8.8 (8.8.8.8) from 192.168.17.254 wan: 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=250 time=6.23 ms

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 6.229/6.229/6.229/0.000 ms
replacing route table 99
default dev wan proto static scope link metric 40 
default dev wg proto static scope link metric 43 
10.8.11.0/24 dev wg proto static scope link metric 43 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
PING 8.8.8.8 (8.8.8.8) from 192.168.17.254 wan: 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=250 time=9.02 ms

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 9.023/9.023/9.023/0.000 ms
```

I see the test you post on gist. 
That doesn't tell us anything, it doesn't show that the packet couldn't be sent because of the mark.

ptpt52:
see, I cannot reproduce your test result using your test_route.sh
the kernel version also is
```
root@X-WRT:~# uname -r
5.4.72
```

aaronjg:
Please turn off mwan3 and the wireguard interface and run the script again and see if you get the same result. I believe it will not change things, but let's make sure.

Just because you can not reproduce the result does not mean that it is not an issue for other users. I am also on `5.4.72`

ptpt52:
@aaronjg 
test with wg del and mwan3 stop
```
root@X-WRT:~# mwan3 stop
root@X-WRT:~# ip link del wg
root@X-WRT:~# ip r
default via 192.168.17.1 dev wan proto static metric 40 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
root@X-WRT:~# sh test_route.sh 
route table:
default via 192.168.17.1 dev wan proto static metric 40 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
PING 8.8.8.8 (8.8.8.8) from 192.168.17.254 wan: 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=250 time=9.07 ms

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 9.065/9.065/9.065/0.000 ms
replacing route
default dev wan proto static scope link metric 40 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
PING 8.8.8.8 (8.8.8.8) from 192.168.17.254 wan: 56(84) bytes of data.

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 0 received, 100% packet loss, time 0ms

route rules
0:	from all lookup local
99:	from all fwmark 0x63 lookup 99
32766:	from all lookup main
32767:	from all lookup default
route table 99:
default via 192.168.17.1 dev wan proto static metric 40 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
PING 8.8.8.8 (8.8.8.8) from 192.168.17.254 wan: 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=250 time=11.6 ms

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 11.602/11.602/11.602/0.000 ms
replacing route table 99
default dev wan proto static scope link metric 40 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
PING 8.8.8.8 (8.8.8.8) from 192.168.17.254 wan: 56(84) bytes of data.

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 0 received, 100% packet loss, time 0ms


```
it seems the same as your result now.

ptpt52:
```
root@X-WRT:~# ip rule
0:	from all lookup local
99:	from all fwmark 0x63 lookup 99
32766:	from all lookup main
32767:	from all lookup default
root@X-WRT:~# ip route show table 99
default dev wan proto static scope link metric 40 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
root@X-WRT:~# ip route del table 99 default dev wan proto static scope link metric 40 
root@X-WRT:~# /usr/bin/ping -m99 -I wan 8.8.8.8
PING 8.8.8.8 (8.8.8.8) from 192.168.17.254 wan: 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=250 time=6.94 ms
64 bytes from 8.8.8.8: icmp_seq=2 ttl=250 time=6.28 ms
^C
--- 8.8.8.8 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1004ms
rtt min/avg/max/mdev = 6.279/6.610/6.942/0.331 ms
root@X-WRT:~# 

```

but if I remove `wrong default troute` from table 99, the ping success.

ptpt52:
```
root@X-WRT:~# ip route add default via 192.168.15.254 dev br-lan table 99
root@X-WRT:~# ip route show table 99
default via 192.168.15.254 dev br-lan 
169.254.0.0/16 dev br-lan proto kernel scope link src 169.254.254.254 
192.168.15.0/24 dev br-lan proto kernel scope link src 192.168.15.1 
192.168.16.1 via 192.168.17.1 dev wan proto static metric 40 
192.168.17.0/24 dev wan proto static scope link metric 40 
root@X-WRT:~# /usr/bin/ping -m99 -I wan 8.8.8.8
PING 8.8.8.8 (8.8.8.8) from 192.168.17.254 wan: 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=250 time=12.2 ms
64 bytes from 8.8.8.8: icmp_seq=2 ttl=250 time=10.8 ms
^C
--- 8.8.8.8 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1002ms
rtt min/avg/max/mdev = 10.755/11.502/12.249/0.747 ms
```

and even more, if I add somewhat default route to table 99, it work also.

aaronjg:
Right, when you remove the default route, it falls through to the main routing table.  That test script was from memory of the issue, and was a little different from the issue I first encountered. I went back over my notes and found the actual use case that originally raised this issue:

It is if you have a a policy that uses wana, then wanb, and if both wana and wanb are down then unreachable.  With the 4.x kernel, it will work, but with the 5.x kernel it won't.

Here is the test case to illustrate the difference between the kernel versions:

 https://gist.github.com/aaronjg/fa0592ee58b415933284296895cb6047

The new kernel version is the correct behavior, but mwan3 had been relying on the broken behavior `unreachable` rule not being applied when `ping -I` was set.

ptpt52:
You are right.

ptpt52:
I notice that a big change in the kernel https://github.com/torvalds/linux/commit/2fbc6e89b2f1403189e624cabaf73e189c5e50c6

maybe this is related

ptpt52:
could we use the `iputils-ping` for ping with `ping -m$((0x3f00))`?
to archive the ping check in mwan3track, instead of the WRAP?

ptpt52:
and even we could add `ping -m` support to busybox ping

aaronjg:
> I notice that a big change in the kernel [torvalds/linux@2fbc6e8](https://github.com/torvalds/linux/commit/2fbc6e89b2f1403189e624cabaf73e189c5e50c6)
> 
> maybe this is related

That change was after 5.4.x.


> could we use the `iputils-ping` for ping with `ping -m$((0x3f00))`?
> to archive the ping check in mwan3track, instead of the WRAP?

@feckert and I discussed that alternative. Unfortunately, it adds the dependency on iputils-ping and won't work with httping, nping, etc

What is the issue you are running into with the C wrapper? Can you create a test script to demonstrate the problem? If we can understand this better, it will help come up with a fix and to explore alternative solutions.

ptpt52:
nothing issue about wrap, but I am just think the perfect solution.
but so far, as you answered, the `wrap method` is the best solution.

feckert:
LGTM

aaronjg:
> LGTM thanks

Great. Thanks for the review!

PR #13751, opened 2020-10-22
nginx: adapt to changed ubus.sock path

dangowrt:

Maintainer: @heil @Ansuel 
Compile tested: --
Run tested: --

Description:
ubus socket was moved from /var/run/ubus.sock to /var/run/ubus/ubus.sock upstream.
Reflect that change in nginx and migrate existing configurations.

### DIFF
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nginx
 PKG_VERSION:=1.19.2
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=nginx-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://nginx.org/download/
--- a/net/nginx/files-luci-support/60_nginx-luci-support
+++ b/net/nginx/files-luci-support/60_nginx-luci-support
@@ -6,12 +6,15 @@ if nginx -V 2>&1 | grep -q ubus; then
 
 location /ubus {
         ubus_interpreter;
-        ubus_socket_path /var/run/ubus.sock;
+        ubus_socket_path /var/run/ubus/ubus.sock;
         ubus_parallel_req 2;
 }
 EOT
 	fi
 fi
+
+grep -q /var/run/ubus.sock /etc/nginx/conf.d/luci.locations &&
+	sed -i 's#/var/run/ubus.sock#/var/run/ubus/ubus.sock#' /etc/nginx/conf.d/luci.locations
 
 if [ -x /etc/init.d/uhttpd ]; then
         /etc/init.d/uhttpd disable


Ansuel:
thx do you know when ubus will be migrated to non - root?

dangowrt:
Already happened with the recent procd update https://github.com/openwrt/openwrt/commit/a2def3663a0feff12550906c33f9ecc3e5fb2a5d

Ansuel:
Didn't notice... Will do more test and do a dedicated pr for this (remove root user from nginx)... Anyway to not have this broken, we can merge this.

dangowrt:
> Didn't notice... Will do more test and do a dedicated pr for this (remove root user from nginx)...
I don't think that's related. Access _to_ ubus still requires either being root or having rules setup in `/usr/share/acl.d`.
What did change now is just the user the ubus daemon itself is running as and for that the location of the socket had to be moved. Apart from that changed path nothing changes when it comes to using ubus.

Ansuel:
Tought that that change permit other process to use ubus feature... 
With the current implementation the path is
luci -> nginx -> ubus 
And the dedicated module directly use ubus api as uhttpd does. If I don't run nginx as root, the module can't comunicate with ubus.
Can you link me or if you want explain me if there is a way to use ubus as another user?  Should I add nginx as an user that can access all the ubus request. (i quickly looked at luci-base in the acl.d directory)

dangowrt:
Currently uhttpd runs as root for the same reasons. You would need to write acl allowing more or less everything (?) for the users the web-server is running at in /usr/share/acl.d. It's worth trying as the same could then be applied for uhttpd as well.

Ansuel:
Will do and report here

Ansuel:
@dangowrt Some info about this... 
We have 2 way to have webserver running non root (and use acl.d)
1. Add every permission that use luci. This is not possibile since this would create problem if some new app creates new ubus rule using rpcd (they will have to both update the luci acl.d and the webserver acl.d)
2. We need a patch to the ubusd source to readd (don't know what happen but wildecard support has been dropped) support for wildcard in methods... this way we can just put `[ "*" ]` in the array and support all. (the first problem doesn't occur since any app that exports service with rpc should add `luci.` as a prefix (wildcard in the object is supported so we don't need a patch for that)

I'm still testing this (The patch to add wildcard support is straightforward) but it seems to work... I still missing some permission but most of the feature works.

(this is the change... tell me if you want to send a RFC o a PR in the mailing list or on openwrt github)

```
		case UBUS_ACL_ACCESS:
			if (acl->methods) {
				struct blob_attr *cur;
				size_t rem;

				blobmsg_for_each_attr(cur, acl->methods, rem)
					if (blobmsg_type(cur) == BLOBMSG_TYPE_STRING) {
						char * func = blobmsg_get_string(cur);

						if (!strcmp(method, func) || !strcmp("*", func))
							return 0;
					}
			}
			break;
		}
```

Ansuel:
@dangowrt Friendly ping... Did you missed this or it makes no sense to add the patch and all this permission to address this problem? (It's 3 days that i'm testing this and I can't really see any problem, and if any it will be easily fixable by just adding the permission with the wildcards)

anomeome:
Does this address the current _No related RPC reply_ error currently experienced with nginx

Ansuel:
Yes why this is still not merged? @hnyman

hnyman:
I will merged, as you seem to indicate that this is needed to fix things.

(I usually avoid merging PRs where the author could merge it by himself)

PR #13756, opened 2020-10-24
mwan3: fix route table copy not handle correctly

ptpt52:

Description:

this add one more commit after P/R https://github.com/openwrt/packages/pull/13748

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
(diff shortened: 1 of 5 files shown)


feckert:
I will merge the other [PR](https://github.com/openwrt/packages/pull/13748) first and then we can look at this one.
Could you please explain more precisely why you want to make this [changes](
https://github.com/openwrt/packages/pull/13756/commits/ef853528dec0a4b89a4ead0039c252558d3ff663) I can't follow you there!

ptpt52:
drop this P/R since it is only used to report and demonstrate issues

PR #13763, opened 2020-10-25
openconnect: allow specify --protocol from config

mayli:
Maintainer: @nmav
Run tested: x86_64, master, connect to globalprotect vpn with OTP

Description:

openconnect v8.10 supports 4 VPN protocols

  --protocol=anyconnect  Compatible with Cisco AnyConnect SSL VPN, as well as ocserv (default)
  --protocol=nc          Compatible with Juniper Network Connect
  --protocol=gp          Compatible with Palo Alto Networks (PAN) GlobalProtect SSL VPN
  --protocol=pulse       Compatible with Pulse Connect Secure SSL VPN

This patch allows user to specify protocol use the new "vpn_protocol"
option and deprecate the old option "juniper" which seems to be missing in
the current openconnect client.

Signed-off-by: Mengyang Li <mayli.he@gmail.com>


nmav:
LGTM, the failures seem unrelated with the change. You'll also need to bump the the release version in Makefile

nmav:
I'd also recommend an equivalent change in luci openconnect

mayli:
Got it, will update luci once this patch get merged.

nmav:
Merged manually. Thank you!

PR #13767, opened 2020-10-25
nss: update to 3.58

lucize:

Maintainer: me
Tested: x86

Sorry for not jumping to the latest, I somehow missed the latest version

### DIFF
--- a/libs/nss/Makefile
+++ b/libs/nss/Makefile
@@ -7,14 +7,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nss
-PKG_VERSION:=3.57
+PKG_VERSION:=3.58
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


neheb:
I would like this to be rebased in a day so that the SDK gets updated and the buildbots fixed.

ja-pa:
@lucize Do you know, if OpenWrt nss is affected by [CVE-2020-25648](https://nvd.nist.gov/vuln/detail/CVE-2020-25648) ? If so, would it be possible to update commit message with CVE id ?

lucize:
prior versions of 3.58 are affected

ja-pa:
> prior versions of 3.58 are affected

Yes,  so your updated fixes CVE-2020-25648 .  For example this commit message has CVE id, users than know that it's security relevant package update without looking to package changelog. https://github.com/openwrt/packages/commit/fc7cf379e381997f0c31e69f808911d756d613e3
Your commit message is missing that info so the user has to check the package changelog or use cvechecker (which is not very reliable).

lucize:
it's already committed

PR #13779, opened 2020-10-26
docker-compose: move to Utilities in menuconfig

bobafetthotmail:
it makes no sense to show docker-compose sit in Languages -> Python
submenu in menuconfig, it is a tool and not a library.
Move it to Utilities section like docker-ce also is.

Signed-off-by: Alberto Bursi <bobafetthotmaiil@gmail.com>

Maintainer: @jmarcet


neheb:
This makes sense.

PR #13780, opened 2020-10-26
gunicorn: move to webserver category in menuconfig

bobafetthotmail:
this is a webserver and not a python library, so it makes
sense to have it in the same category other webservers
also are placed in the menuconfig interface

Signed-off-by: Alberto Bursi <bobafetthotmail@gmail.com>

Maintainer: @commodo

### DIFF
--- a/lang/python/gunicorn/Makefile
+++ b/lang/python/gunicorn/Makefile
@@ -23,9 +23,9 @@ include $(INCLUDE_DIR)/package.mk
 include ../python3-package.mk
 
 define Package/gunicorn/Default
-  SUBMENU:=Python
-  SECTION:=lang
-  CATEGORY:=Languages
+  SUBMENU:=Web Servers/Proxies
+  SECTION:=net
+  CATEGORY:=Network
   TITLE:=WSGI HTTP Server for UNIX
   URL:=https://gunicorn.org
 endef


jefferyto:
Perhaps it would be good to move the package directory to net/gunicorn as well.

commodo:
> Perhaps it would be good to move the package directory to net/gunicorn as well.

Possibly.
I have no strong opinion about where to put this.
It was put by the person who added  this package (not me).
Since it’s a pure python package, it can also stay here.
There’s also the python-twisted package that would have the same discussion.
python-flup maybe as well.

bobafetthotmail:
I think packages were (and should be) categorized by what they actually do, not by what language they are in. Most stuff in C or shell or perl isn't in the appropriate language folders for example.

by googling, "gunicorn" is an actual webserver that is supposed to be used on its own and has its own text-based configuration https://docs.gunicorn.org/en/latest/configure.html

by googling, "python-twisted" is "an event-driven network programming framework" so it's some kind of library or component and should stay in Python libraries folder.

by googling, "python-flup" is again a component/library that allows you to embed webserver functionality in your python application, it's not a standalone application https://www.saddi.com/software/flup/ and again this should stay in Python libraries folder.

Still, this was mostly a drive-by, I noticed gunicorn only because it was under "docker-compose" which was also in the wrong folder (and has another PR from me to move it to Utilities like docker-ce). I don't use/need gunicorn.

neheb:
Seems fine to me. There are other Python packages such as seafile that do not live under the Python section.

Issue #13589, opened 2020-10-06
Runtime test ideas/enhancements


jefferyto:
Maintainer: @aparcar (also ping @neheb)
Environment: N/A

Description:
(These are ideas that occurred to me from reading #13449.)

Instead of having every package include a `test.sh`, I'd like to suggest having a default runtime test that:

* Checks to see if the package contains at least one executable
* Tries running the executable with various versions of the `--version` command line option, e.g. `foo --version`, `foo -version`, `foo version`
* If any attempt exits with a success and the version number is found in the output, then the test is considered successful

It would be great if there is a variable that can be added to the package Makefile to disable this default runtime test.

If a custom runtime test is very simple, it would also be great if the test code can be written directly in the package Makefile, i.e. assigned to a Makefile variable (perhaps the same as the previous variable).

BKPepe:
@aparcar Any response would be highly appreciated. This is here for 10 days. It does not look good without a reply.

aparcar:
Hi, thanks for pinging me, this issues got lost in my notifications.

> Checks to see if the package contains at least one executable

That sounds great, could you share you implementations ideas? We can unpack the package in a temporary folder and scan for executables. How do we deal with a package like `vim-docs` and how with libraries? Are package like these rare enough to require maintainers to adopt a Makefile?

> Tries running the executable with various versions of the --version command line option, e.g. foo --version, foo -version, foo version

I had a similar idea but chose the `test.sh` way because it's more conscious to enable tests, rather than having them enabled by default. However, as people seem to like runtime tests, let's enable this version checking command.

> It would be great if there is a variable that can be added to the package Makefile to disable this default runtime test.

Sure let's do this. As written before I meant to have it optional but we can also make it an opt-out.

> If a custom runtime test is very simple, it would also be great if the test code can be written directly in the package Makefile, i.e. assigned to a Makefile variable (perhaps the same as the previous variable).

We could introduce a variable `PKG_TEST` which can also be added on a *subpackage* base:
* Empty or `default`: Run automated version checks
* `no` (or `0` or `none` or ...?) disables tests
* `script` run a `test.sh` script imitating the current behaviour
* Anything else is run as a script?

I'm a bit afraid of allowing actual scripts in the Makefiles, people may get fancy and bloat Makefiles?

aparcar:
> It does not look good without a reply.

jefferyto:
> > It does not look good without a reply.

Is this is directed at me?

[...]

PR #13785, opened 2020-10-27
CI: generic runtime tests

aparcar:
Requested via https://github.com/openwrt/packages/issues/13589


aparcar:
@jefferyto Here we face the first problem. The `vim` package contains a tool called `xxd` which is an entirely different version:

```
Testing package xxd in version 8.2 from vim
Installing xxd (8.2-2) to root...
Configuring xxd.
Use generic test
Test executable /usr/bin/xxd
```

jefferyto:
I don't want to comment too much on the immediate issue with `xxd` because I'm not crazy about this PR as it stands. I think doing everything inside the container is too inflexible and you have no access to the package Makefile.

I suggest something like this (it doesn't need to be all done in one PR), for each package/subpackage:

* Checking the package Makefile, if the package has opted into tests (as mentioned in https://github.com/openwrt/packages/issues/13589#issuecomment-717119578)
  * If the package does not have a custom test script, output the default test script to something like `custom_test.sh` or `custom_test_<subpackage-name>.sh`
  * (optional) If the package has a customized command line argument (instead of `--version -version ...`), output the default test script with the custom command line argument into `custom_test.sh` / `custom_test_<subpackage-name>.sh` (perhaps there are other variables in the default test script that can be customized in a similar manner)
  * If the package has a custom test script, output it into `custom_test.sh` / `custom_test_<subpackage-name>.sh`
* Inside the container
  * If there is a `custom_test.sh` / `custom_test_<subpackage-name>.sh`, run it
  * Otherwise if there is a `test.sh`, run that

This way the content of `test.sh` can even be completely defined inside the Makefile (similar to config options, they can be defined inside the Makefile or `source`d from a separate file).

champtar:
We expect contributor to be able to fix CI failure, so we need to be able to replicate what the CI does as easily as possible
=> `make package/aaa/test` would be really nice
And also CI should be green for working packages, it's nice to test more things but if we end up ignoring 90% of the times is it really better

aparcar:
Sorry this slipped through my notifications.

> everything inside the container is too inflexible

Please elaborate what parts are too inflexible.

> you have no access to the package Makefile

The `test.sh` script is placed literally next to the Makefile and therefore accessible from the container. In fact the entire `packages.git` is mounted within the Docker container.

We could use Makefile `define` to *define* tests rather than having separate files, not sure if that's a big advantage.

>  => make package/aaa/test would be really nice

How do you envision this to be implemented? Your host system is unlikely to support all these architectures, likely not even x86-musl. Then you need a working `opkg` installing dependencies. You can make it work, but that's plenty of work. One could automatically spin up a dozen Docker containers, but then again that's what the CI is for. My lazy approach would be that people just draft PRs and push until the CI is happy. I don't expect anyone to test their PRs for 10 archs.

.github/workflows/entrypoint.sh
```diff
@@ -32,7 +32,34 @@ for PKG in /ci/*.ipk; do
...
+				done
+
+			fi
+		done
+		if [ "$FOUND_EXE" -eq 1 ]; then
+			if [ "$SUCCESS" -eq 0 ]; then
+				echo "Generic test failed"
+				exit 1
+			else
+				echo "Test succesful or"
```
BKPepe, line 58: Shouldn't this be without or?

```suggestion
				echo "Test successful"
```

.github/workflows/entrypoint.sh
```diff
@@ -32,7 +32,34 @@ for PKG in /ci/*.ipk; do
...
+			fi
+		else
+			echo "No executeable found in package $PKG_NAME"
```
BKPepe, line 61:
```suggestion
			echo "No executable found in package $PKG_NAME"
```

BKPepe:
Tested with lighttpd. Seems OK to me!  More tested packages will follow soon.

aparcar:
I'd like to get more input from @champtar and @jefferyto before continuing.

BKPepe:
> * (optional) If the package has a customized command line argument (instead of `--version -version ...`), output the default test script with the custom command line argument into `custom_test.sh` / `custom_test_<subpackage-name>.sh` (perhaps there are other variables in the default test script that can be customized in a similar manner)

I agree on this one. We should have something like this. 
There might be some packages, which does not have e.g. `--version`, `-version` and it that case check fails and in the logs, there is: generic test failed. In that case, there could be fallback to custom test or make it even complicated: 

1. try to check for `version`
2. try to show `help`
3. try to run app without any parameters to check if it runs and that it does not segfault or does not show any errors
4. have custom script if method 1.,2. fails (they are not there).

Because I doubt that anyone wants to create his/her `test.sh` or  ship the generic `test.sh` with each package folder, so having something by default :+1: 

TL;DR: I agree with @jefferyto's suggestions, but one difference that all packages will be using tests even though could be complicated for libraries.

BKPepe:
It looks like several members are really depend on run testing packages inside CI/CD, but most of those packages does not have test.sh file at all and it is not run tested in the end. We should proceed with this PR.

Any thoughts, @1715173329 , @neheb, @robimarko ?

.github/workflows/entrypoint.sh
```diff
@@ -32,7 +32,34 @@ for PKG in /ci/*.ipk; do
 			exit 1
 		fi
 	else
-		echo "No test.sh script available"
+		echo "Use generic test"
+	        PKG_FILES=$(opkg files "$PKG_NAME" | grep -E "/bin|/sbin|/usr/bin|/usr/sbin")
```
GeorgeSapkin, line 36:
```suggestion
			PKG_FILES=$(opkg files "$PKG_NAME" | grep -E "/bin|/sbin|/usr/bin|/usr/sbin")
```

.github/workflows/entrypoint.sh
```diff
@@ -32,7 +32,34 @@ for PKG in /ci/*.ipk; do
...
+			if [ -x "$FILE" ] && [ -f "$FILE" ]; then
+				FOUND_EXE=1
+				echo "Test executable $FILE"
+				for V in --version -version version -v -V; do
+					echo "Trying $FILE $V"
+					if "$FILE" "$V" 2>&1 | grep "$PKG_VERSION"; then
+						SUCCESS=1
+						break
+					fi
+				done
```
GeorgeSapkin, line 49: I think ensuring the main binary can display help should be enough for a success already, as a last resort after the version checks. This at least checks that binaries are valid and have the linked dependencies installed.

Also I think it makes sense to do `grep -F` if that's not the default.

wehagy: But how we can test this? Because not all binaries return `exit 0`, maybe `grep` the binary name?

wehagy (commented):
Also is better move to https://github.com/openwrt/actions-shared-workflows/blob/main/.github/dockerfiles_feeds/entrypoint.sh?

.github/workflows/entrypoint.sh
wehagy: Need `apk info -L` support

.github/workflows/entrypoint.sh
wehagy: > I think ensuring the main binary can display help should be enough for a success already, as a last resort after the version checks. This at least checks that binaries are valid and have the linked dependencies installed

The problem is how? Because not all packages use `exit 0` as a return. Maybe we can use `grep` to check the binary name itself?

GeorgeSapkin: I meant on top of `--version -version version -v -V` do something like `--help`, etc. Worst case scenario somebody's PR fails and they add a proper test.

Might be a good idea to also do an `ldd` (if it's a binary) and check if any linked dependencies are missing. And do this even if packages have their own tests.

GeorgeSapkin:
I ported this PR to https://github.com/openwrt/actions-shared-workflows/pull/60 since this logic no-longer lives here. Can this be closed?

PR #13796, opened 2020-10-28
docker-ce: Added blocked_interfaces config option

G-M0N3Y-2503:
Maintainer: @G-M0N3Y-2503

Description:
* blocked_interfaces blocks all packets to docker0 from the given
  interface since the iptables commands docker adds all operate before
  any of the fw3 generated rules.
* Made shellcheck recommendations
* Fixed issue where `/tmp/dockerd/` didn't exist for `alt_config_file`

### DIFF
--- a/utils/docker-ce/files/etc/config/dockerd
+++ b/utils/docker-ce/files/etc/config/dockerd
@@ -1,11 +1,18 @@
+# The following settings require a restart to take full effect, A reload will
+# only have partial or no effect:
+# option bip
+# list blocked_interfaces
 
 config globals 'globals'
 #	option alt_config_file "/etc/docker/daemon.json"
 	option data_root "/opt/docker/"
 	option log_level "warn"
 	list hosts "unix:///var/run/docker.sock"
-	# If the bip option is changed, dockerd must be restarted.
-	# A service reload is not enough.
 	option bip "172.18.0.1/24"
 #	list registry_mirrors "https://<my-docker-mirror-host>"
 #	list registry_mirrors "https://hub.docker.com"
+
+# Docker ignores fw3 rules and by default all external source IPs are allowed
+# to connect to the Docker host. See https://docs.docker.com/network/iptables/
+config firewall 'firewall'
+	list blocked_interfaces 'wan'
--- a/utils/docker-ce/Makefile
+++ b/utils/docker-ce/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=docker-ce
 PKG_VERSION:=19.03.13
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=components/cli/LICENSE components/engine/LICENSE
 
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -25,7 +25,7 @@ boot() {
 uciupdate() {
 	local net="${1}"
 
-	uci -q get network.docker >/dev/null || {
+	uci_quiet get network.docker || {
 		logger -t "dockerd-init" -p warn "No network uci config section for docker default bridge (docker0) found"
 		return
 	}
@@ -52,7 +52,7 @@ uciadd() {
 	}
 
 	# Add network interface
-	if ! uci -q get network.docker >/dev/null; then
+	if ! uci_quiet get network.docker; then
 		logger -t "dockerd-init" -p notice "Adding docker default interface to network uci config (docker)"
 		uci_quiet add network interface
 		uci_quiet rename network.@interface[-1]="docker"
@@ -63,7 +63,7 @@ uciadd() {
 	fi
 
 	# Add docker bridge device
-	if ! uci -q get network.docker0 >/dev/null; then
+	if ! uci_quiet get network.docker0; then
 		logger -t "dockerd-init" -p notice "Adding docker default bridge device to network uci config (docker0)"
 		uci_quiet add network device
 		uci_quiet rename network.@device[-1]="docker0"
@@ -74,7 +74,7 @@ uciadd() {
 	fi
 
 	# Add firewall zone
-	if ! uci -q get firewall.docker >/dev/null; then
+	if ! uci_quiet get firewall.docker; then
 		logger -t "dockerd-init" -p notice "Adding docker default firewall zone to firewall uci config (docker)"
 		uci_quiet add firewall zone
 		uci_quiet rename firewall.@zone[-1]="docker"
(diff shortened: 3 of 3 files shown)


G-M0N3Y-2503:
@feckert does this address the main concerns from https://github.com/openwrt/packages/pull/13643?

feckert:
@G-M0N3Y-2503 thanks for taking care of it.
I still have to do a runtime test.
But a quick source check seems to solve the problem I described in #13643.

Could you please split the commits in smaller parts?

- Fix shellcheck warning
- extend iptables command with long commands
- Add new option to prohibit docker access from selected interfaces
- Fixed issue where /tmp/dockerd/ didn't exist for alt_config_file

G-M0N3Y-2503:
I've split it up into more manageable commits. 
If you wanted to test it yourself to be sure, I'm happy to wait.
Otherwise, I'll remove the draft status of this PR.

G-M0N3Y-2503:
made a slight amendment allowing `blocked_interfaces` to be used with `alt_config_file` because it's not really related to the config anyway.

utils/docker-ce/files/etc/config/dockerd
```diff
@@ -1,11 +1,18 @@
...
-	# If the bip option is changed, dockerd must be restarted.
-	# A service reload is not enough.
 	option bip "172.18.0.1/24"
 #	list registry_mirrors "https://<my-docker-mirror-host>"
 #	list registry_mirrors "https://hub.docker.com"
+
+# Docker ignores fw3 rules and by default all external source IPs are allowed
+# to connect to the Docker host. See https://docs.docker.com/network/iptables/
+config firewall 'firewall'
+	list blocked_interfaces 'wan'
```
feckert, line 18: @G-M0N3Y-2503 Would it not be better if we move the list blocked_interfaces to the globals section?
And from a security point of view would it not be better if we did not block everything and only allow those who are on the list. Then it is not possible to make all containers public by accident when a new interface are added

feckert:
@G-M0N3Y-2503 I have now had another discussion with a colleague and mine.
And I wanted to make sure that I didn't make a mistake in my thinking.

We have a complete zone firewall concept in the openwrt, so I still remain that we switch off the firewall handling in docker 
https://github.com/openwrt/packages/pull/13643 and then rely on the fw3. I would also make this optional https://github.com/openwrt/packages/pull/13643/files#r509242182

From my point of view the firewall handling in the docker only makes sense if I don't have any other firewall in use and I don't want to deal with it and I want to isolate the containers form each others.

For the default bridge interface docker0 i can create everything (which already happens) in fw3 and so uci config.

If I add a new bridge, because I want to put my containers into another bridge, I have to redo this process again.
We can also script this and put it into the hotplug.d/iface so If docker creates a new interface we could also add this to uci as with the docker0.

utils/docker-ce/files/etc/config/dockerd
G-M0N3Y-2503: I think this option isn't documented well enough right now, the initial intention was to have it relate to [the docker documentation on this subject](https://docs.docker.com/network/iptables/#restrict-connections-to-the-docker-host).

>  Would it not be better if we move the list blocked_interfaces to the globals section?

I thought it might be nice to have options that don't correlate directly with a [dockerd config entry](https://docs.docker.com/engine/reference/commandline/dockerd/) in a separate section.
I chose `firewall` because these settings modify the firewall for `dockerd`.

> would it not be better if we did not block everything and only allow those who are on the list.

I didn't consider this option thinking it should relate to the docs, but I think it would be better. We'd just have to enumerate all interfaces and add drops for all not mentioned, I'll look into this.

Do you think only the `lan` should be allowed by default? maybe not the OpenWRT interfaces and just the loopback interface and people would have to look up what `ethX` they wanted to allow?

G-M0N3Y-2503:
> We have a complete zone firewall concept in the openwrt, so I still remain that we switch off the firewall handling in docker
> #13643 and then rely on the fw3. I would also make this optional https://github.com/openwrt/packages/pull/13643/files#r509242182
> 
> From my point of view the firewall handling in the docker only makes sense if I don't have any other firewall in use and I don't want to deal with it and I want to isolate the containers form each others.
> 
> For the default bridge interface docker0 i can create everything (which already happens) in fw3 and so uci config.
> 
> If I add a new bridge, because I want to put my containers into another bridge, I have to redo this process again.
> We can also script this and put it into the hotplug.d/iface so If docker creates a new interface we could also add this to uci as with the docker0.

I'm not sure if I follow what you're proposing exactly. 

But In general, I agree with exposing the `iptables` option to be in the config file, I think it will be fine for some simple cases.

The concern is where people are trying to do more complex stuff with networking, perhaps unknowingly if they are using something like `docker-compose`. The warnings about `--iptables=false` [here](https://docs.docker.com/network/iptables/#prevent-docker-from-manipulating-iptables) and [here](https://docs.docker.com/engine/reference/commandline/dockerd/#run-multiple-daemons) lead me to believe that docker would break with `fw3` or break the integrity of `fw3` if people tried to do complex networking.

The part I prefer with the angle of this PR for the default case, is that it does what `docker` expects, so as long as [the docs](https://docs.docker.com/network/iptables/#restrict-connections-to-the-docker-host) are correct we should be able to have `docker` handle its traffic itself and have `fw3` handle everything else.

feckert:
@G-M0N3Y-2503 I would like to have the possibility that we have both the firewall handling by docker and handling with uci with to get a proper integration into openwrt network and firewall managment. If this P/R works I would merge this and suggest another P/R with my changes to handle everything with uci and fw3 that docker does.

G-M0N3Y-2503:
> another P/R with my changes to handle everything with uci and fw3 that docker does.
@feckert, I'm interested to see what you mean, But if you mean another P/R that would supersede this one, I might just separate out the refactoring from this one and then we can leave it until we're happy.

feckert:
> @feckert, I'm interested to see what you mean, But if you mean another P/R that would supersede this one, I might just separate out the refactoring from this one and then we can leave it until we're happy.

No I do not mean to replace this one.
It solves the problem that the containers are also accessible over the WAN. When I am ready I will create a new P/R which will improve the integration into fw3. From my point of view it can be merged.
Adding the the **allow_interfaces** config option as suggested by me should also go into a separate P/R.

G-M0N3Y-2503:
Oh, sorry I misunderstood, This is ok to go in for me.

> Adding the the **allow_interfaces** config option as suggested by me should also go into a separate P/R.

I'll leave this to you also then.

feckert:
@G-M0N3Y-2503 could you also update PKG_RELEASE in a separate commit.

G-M0N3Y-2503:
Oh, forgot something went in, Fixed.

feckert:
@G-M0N3Y-2503 
This looks good for docker0.
I will merge this now.

But I already have a new P/R in the pipeline, where the device (docker0) can be entered as an option.
```
# Docker ignores fw3 rules and by default all external source IPs are allowed
# to connect to the Docker host. See https://docs.docker.com/network/iptables/
config firewall 'firewall'
        option device 'docker0'
        list blocked_interfaces 'wan'
```

The current change only handles docker0.
If you add another bridge over docker and then run containers in it, then we have the same problem again.

feckert:
@G-M0N3Y-2503 Thanks for your work

PR #13800, opened 2020-10-28
lighttpd: revamp Makefile for lighttpd 1.4.56

gstrauss:
Maintainer: W. Michael Petullo <mike@flyn.org>
Compile tested: x86-64 OpenWrt master branch (v19.07.4)

Description:
revamp Makefile for lighttpd 1.4.56

Note: lighttpd-1.4.56 has not yet been released.  lighttpd-1.4.56\~rc2 is available.  lighttpd-1.4.56\~rc3 (not yet released) will contain meson.build updates for OpenWrt which have been compile tested with this PR.

Submitting PR for review in tandem with lighttpd-1.4.56 release candidates.
**Please do not merge until after lighttpd-1.4.56 is released**


BKPepe:
Marked it as Draft to ensure that this is not getting merged until version 1.4.56 is released.

net/lighttpd/Makefile
```diff
@@ -52,41 +52,64 @@ config LIGHTTPD_SSL
...
+  endif
+endif
+
+ifneq ($(CONFIG_PACKAGE_lighttpd-mod-auth),)
+  ifeq ($(CONFIG_PACKAGE_lighttpd-mod-authn_file),)
+    CONFIG_PACKAGE_lighttpd-mod-authn_file=m
+  endif
+endif
+
+# (lighttpd supports -Dwith_brotli if openwrt provided libbrotli as a package)
```
neheb, line 81: Just add -Dwith_brotli=false below.

net/lighttpd/Makefile
```diff
@@ -52,41 +52,64 @@ config LIGHTTPD_SSL
...
+
 MESON_ARGS += \
 	-Dwith_bzip=false \
-	-Dwith_dbi=$(if $(CONFIG_PACKAGE_lighttpd-mod-vhostdb_dbi),true,false) \
+	-Dwith_dbi=$(if $(CONFIG_PACKAGE_lighttpd-mod-authn_dbi)$(CONFIG_PACKAGE_lighttpd-mod-vhostdb_dbi),true,false) \
 	-Dwith_fam=false \
 	-Dwith_gdbm=$(if $(CONFIG_PACKAGE_lighttpd-mod-trigger_b4_dl),true,false) \
-	-Dwith_geoip=$(if $(CONFIG_PACKAGE_lighttpd-mod-geoip),true,false) \
```
neheb, line 64: Why remove this? Just set it to false if it's no longer used.

net/lighttpd/Makefile
```diff
@@ -159,25 +183,26 @@ $(eval $(call BuildPlugin,access,Access restrictions,,30))
 $(eval $(call BuildPlugin,accesslog,Access logging,,30))
 $(eval $(call BuildPlugin,alias,Directory alias,,30))
 $(eval $(call BuildPlugin,cgi,CGI,,30))
-#$(eval $(call BuildPlugin,cml,Cache Meta Language,+PACKAGE_lighttpd-mod-cml:liblua +PACKAGE_lighttpd-mod-cml:libmemcached,30))
-$(eval $(call BuildPlugin,cml,Cache Meta Language,+PACKAGE_lighttpd-mod-cml:liblua,30))
-$(eval $(call BuildPlugin,compress,Compress output,+PACKAGE_lighttpd-mod-compress:zlib,30))
+#$(eval $(call BuildPlugin,cml,Cache Meta Language,+PACKAGE_lighttpd-mod-cml:liblua5.3 +PACKAGE_lighttpd-mod-cml:libnettle +PACKAGE_lighttpd-mod-cml:libmemcached,30))
```
neheb, line 186: I suggest getting rid of this comment.

net/lighttpd/Makefile
```diff
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lighttpd
-PKG_VERSION:=1.4.55
-PKG_RELEASE:=3
+PKG_VERSION:=1.4.56
```
neheb, line 11: for testing purposes you can set this to 1.4.56~rc2

net/lighttpd/Makefile
```diff
@@ -8,12 +8,12 @@
...
+PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://download.lighttpd.net/lighttpd/releases-1.4.x
-PKG_HASH:=6a0b50e9c9d5cc3d9e48592315c25a2d645858f863e1ccd120507a30ce21e927
+PKG_HASH:=46d54858c252d361d4e0ebac61bd54b951318b187e3c972f7edb93f5a59487d4
```
neheb, line 17: for testing purposes you can add

```
PKG_BUILD_DIR=$(PKG_NAME)-1.4.56
```

net/lighttpd/Makefile
gstrauss: GeoIP version 1 database from MaxMind are deprecated and will no longer be supported as of May 2022.
GeoIP2 (version 2) databases from MaxMind are supported and used by mod_maxminddb.

Since mod_geoip has been disabled (forever?) in OpenWrt and the GeoIP version 1 databases are not available, I figured that I would remove this for cleanliness.  I'll change this to set it to `false`

net/lighttpd/Makefile
gstrauss: ok.

(I had added a comment to document the choice.  If brotli were available as a standard package in OpenWrt, then I would add an option for lighttpd to use it)

net/lighttpd/Makefile
gstrauss: Some years ago, the maintainers had tried to build lighttpd with memcached, but gave up and move on so that other changes would proceed.

Should I try to get it working with memcached?

net/lighttpd/Makefile
gstrauss: For testing purposes, I am using git-src symlink. :)

net/lighttpd/Makefile
gstrauss: I have removed the comment for mod_cml and for mod_trigger_b4_dl

gstrauss: OpenWrt does not currently provide a libmemcached package.  I have removed those commented-out lines.

net/lighttpd/Makefile
neheb: Right but the CI cannot build this.

gstrauss: ackn.  will update with lighttpd-1.4.56~rc3 after I push rc3 in the next day or so.  There are some meson.build fixes that have been committed, but are not part of rc2.

neheb:
nls.mk might no longer be needed.

gstrauss:
lighttpd packages for mod_authn_mysql and mod_vhostdb_mysql build fine on x86_64 without including nls.mk.  I'll remove nls.mk in my next push.

neheb:
```
 FAILED: src/mod_cml.so.p/mod_cml_lua.c.o 
powerpc-openwrt-linux-musl-gcc -Isrc/mod_cml.so.p -Isrc -I../src -I/home/build/openwrt/staging_dir/target-powerpc_464fp_musl/usr/include -I/home/build/openwrt/staging_dir/toolchain-powerpc_464fp_gcc-8.4.0_musl/usr/include -I/home/build/openwrt/staging_dir/toolchain-powerpc_464fp_gcc-8.4.0_musl/include/fortify -I/home/build/openwrt/staging_dir/toolchain-powerpc_464fp_gcc-8.4.0_musl/include -fdiagnostics-color=always -pipe -std=c11 -Os -mcpu=464fp -fno-caller-saves -fno-plt -fhonour-copts -Wno-error=unused-but-set-variable -Wno-error=unused-result -ffile-prefix-map=/home/build/openwrt/build_dir/target-powerpc_464fp_musl/lighttpd-1.4.56=lighttpd-1.4.56 -Wformat -Werror=format-security -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro -fPIC -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_LARGE_FILES -D_GNU_SOURCE -DHAVE_CONFIG_H -Wall -g -Wshadow -W -pedantic -MD -MQ src/mod_cml.so.p/mod_cml_lua.c.o -MF src/mod_cml.so.p/mod_cml_lua.c.o.d -o src/mod_cml.so.p/mod_cml_lua.c.o -c ../src/mod_cml_lua.c
../src/mod_cml_lua.c:3:10: fatal error: lua.h: No such file or directory
 #include <lua.h>
          ^~~~~~~
compilation terminated.
```

That's no good.

neheb:
If I had to guess it's because lua5.3 gets installed to /usr/include/lua5.3

gstrauss:
I'll revert the dependency change for mod_cml and mod_magnet to depend on `lua` instead of `lua5.3` so that the rest of this can get tested.  There's not dire need to change the dependency to lua5.3, but just general "why not make available modern stable versions rather than ancient versions?"

neheb:
> I'll revert the dependency change for mod_cml and mod_magnet to depend on `lua` instead of `lua5.3` so that the rest of this can get tested. There's not dire need to change the dependency to lua5.3, but just general "why not make available modern stable versions rather than ancient versions?"

Size is one reason. API compatibility is the bigger one.

gstrauss:
Yes, agreed.  The ABI compatibility is the bigger one.  Such a change occur during a major OpenWrt release, not an incremental upgrade.

neheb:
A different failure on musl 1.2.1 locally

```
/home/mangix/devstuff/openwrt/staging_dir/target-powerpc_8540_musl/usr/include/wolfssl/wolfcrypt/types.h:811:10: error: #error "bad math long / long long settings"
  811 |         #error "bad math long / long long settings"
      |          ^~~~~
/home/mangix/devstuff/openwrt/staging_dir/target-powerpc_8540_musl/usr/include/wolfssl/wolfcrypt/types.h:813:5: error: empty enum is invalid
  813 |     };
      |     ^

```

gstrauss:
Likely a local build environment issue if the wolfssl package is available on that platform.

I will look into the currently running CI build results tomorrow.

gstrauss:
Something early in preprocessor madness is not happy with lighttpd mod_wolfssl and the 32-bit builds.  The error is emanating very early, near the top of mod_wolfssl.c.  I'll set up an environment to reproduce this.

neheb:
Sounds like wolfssl has issues.

I assume setting the following to 1 fixes it: https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/wolfcrypt/types.h#L88

gstrauss:
The meson.build is probably contributing.

build_dir/target-x86_64_musl/lighttpd-1.4.56/openwrt-build/src/config.h defines SIZEOF_LONG and SIZEOF_OFF_T
but these are not defined in my autoconf build outside of the OpenWRT build environment.  We'll see if the mismatch is there.

I am guessing: if meson.build does not know it is building 32-bit, it may #define SIZEOF_LONG 8, and result in the fun we are seeing, since that would be incorrect for the 32-bit build.

gstrauss:
Yeah, it's that #define SIZEOF_LONG in config.h generated by meson.

It trips up an unfortunate and fragile bit of code in wolfssl types.h, which fails generating the enum value, which is employed in a macro a few lines down, but that macro `#define CheckCtcSettings()` is not used by lighttpd.  The wolfssl types.h also presumes that HAVE_LIMITS_H will be set if limits.h is available.  The code also does not first check default preprocessor defines, as clang and gcc provide size and endian information, and other compilers generally do, too.
```
clang -dM -E - </dev/null | sort | less
gcc -dM -E - </dev/null | sort | less
```

neheb:
ping @dhewg

gstrauss:
Workaround committed to lighttpd: https://github.com/lighttpd/lighttpd1.4/commit/e9590277ca85f58129b1c349bfc5e3edcd931438 (and should fix the issue in the build here, even though I have to adjust the patch a bit further)

gstrauss:
5th time is a charm right?  Let's hope so.  I'll be back later.

neheb:
If I had to guess, you forgot to add the patch to the commit.

gstrauss:
The patch tried to be more generic, in case lighttpd was built differently (e.g. with a different set of TLS libraries available or not), but there was a (previous) preprocessor directive which hid the patch when also built with openssl.

I have refactored and verified the new patch.  I'll push it later today.

gstrauss:
grrr.  forgot to do a (local) `clean` before compile test.  missed adding meson option `--with-xxhash`.  This build will fail again

gstrauss:
Finally.  All checks passed.  My last patches further isolate wolfssl.

neheb:
Cool upstream broke the CI again :).

gstrauss:
I have pushed updates to the default lighttpd.init and lighttpd.conf.  Please review.

lighttpd.init supports a graceful restart, and also an extra command `relog` to reopen logs, similar to nginx.
lighttpd.conf adds some commented out flags to enable HTTP/2 (a new feature in lighttpd-1.4.56), as well as setting a 5 second timeout for graceful restart.

Do you think that HTTP/2 should be enabled by default in the config?  For the initial release of lighttpd-1.4.56, I have (so far) left it disabled by default.  However, depending on feedback from release candidate testers, I may change it to be enabled by default.  Of course, there will still be the ability to disable it.

For ease of use, I might enable it by default.  For such a major new feature, I might disable it by default.  Do you have a preference for OpenWrt default lighttpd.conf?

Also, the `Lint Pull Request / Lint Code Base` test has failed.  What's up with the lint errors for a standard procd init script?

neheb:
shellcheck has no idea about busybox ash or how these standard variables are defined.

gstrauss:
The arc_arc700 platform core dumped, probably while the CI was broken.  The other jobs appear to have been resubmitted and succeeded.

Any suggestions for the `Lint Pull Request / Lint Code Base` CI test?
Feedback on lighttpd.init and lighttpd.conf changes?
Thanks.

neheb:
I'd ignore.

neheb:
If HTTP/2 support is new, I’d enable it. It’s a pretty important feature.

gstrauss:
lighttpd 1.4.56 has been released.  This PR is ready for final review.  Thanks.

neheb:
Needs https://github.com/openwrt/packages/pull/14081 to be merged first.

gstrauss:
Also, I need to update the hash to the official release hash.  Will rebase shortly.

gstrauss:
> Needs #14081 to be merged first.

Yeah, the lighttpd build check is failing on error with pam.

neheb:
Merged. Please rebase.

gstrauss:
rebased and all github checks have passed

gstrauss:
Thank you @neheb !

neheb:
That lua thing seems to be an issue: https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a72/packages/lighttpd/compile.txt

gstrauss:
Thanks for the pointer.  I'll try to reproduce and will submit a workaround as a new PR.  I also plan to add a patch for the warnings in the log for mod_webdav.c

gstrauss:
> Package lighttpd-mod-cml is missing dependencies for the following libraries:
> liblua5.3.so.0.0.0

This seems to be a limitation of the dependency listing in the Makefile.  Since lighttpd/Makefile declares a dependency on liblua (and not on a specific version), I think the proper solution may be to have a patch for openwrt which disables the lua version search in upstream lighttpd `src/meson.build` so that it checks only for 'lua' and not for specific lua versions.

Aside: FYI: the build log you linked above contains: "NOTICE: You are using Python 3.5 which is EOL. Starting with v0.57, Meson will require Python 3.6 or newer"

neheb:
That's a meson warning which should be ignored. Actually I think the buildbots are using Python 3.7. Or whatever comes with Buster.

gstrauss:
I filed PR #14098 to address the lighttpd build issue with lua dependency

PR #13801, opened 2020-10-28
uw-imap: install c-client/linkage.c to staging

micmac1:
linkage.c can be included by other programs as described in
docs/internal.txt (quote: "By using linkage.c instead of explicit
mail_link() calls, you are guaranteed that you will have a consistant
linkage among all software built on this system").

This commit adds linkage.c to staging.

Maintainer: @lucize 
Compile tested: ath79 master
Run tested: N/A, no runtime change

Description:
Hello Lucian,

Would be good if this file can be added. Asterisk 18 is using it, so then we could use it, too.

Kind regards,
Seb

### DIFF
--- a/libs/uw-imap/Makefile
+++ b/libs/uw-imap/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=uw-imap
 PKG_VERSION:=2007f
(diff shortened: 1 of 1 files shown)


lucize:
LGTM

micmac1:
Thanks!

PR #13762, opened 2020-10-24
domoticz: fix compilation with python 3.9

neheb:
Added Fedora patch to fix compilation.

Added python3 dependency as it seems it's needed now.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @dwmw2 
Compile tested: ath79

Fixes: https://github.com/openwrt/packages/issues/13757

PR #13802, opened 2020-10-29
setools: fix compilation with python 3.9

neheb:
Don't warn on deprecated declarations.

Remove format patch. Better to fix the warnings.

Maintainer: @flyn-org 
Compile tested: ath79


flyn-org:
CI pipelines are still running, but this looks good to me. Thank you!

utils/setools/patches/020-no-deprecated.patch
```diff
@@ -0,0 +1,10 @@
+--- a/setup.py
++++ b/setup.py
+@@ -126,6 +126,7 @@ ext_py_mods = [Extension('setools.policyrep', ['setools/policyrep.pyx'],
+                                              '-Wno-unreachable-code',
+                                              '-Wno-implicit-fallthrough',
+                                              '-Wno-cast-function-type',
++                                             '-Wno-deprecated-declarations',
```
BKPepe, line 7: Why this was not submitted to upstream? There is no commit message in this patch.

neheb: This is not a proper fix. The issue is that setools uses APIs deprecated by Python3. The only reason the build fails is because it passes -Werror.

BKPepe: Which one? This is a missing here. You are doing some changes, which are not documented. Neither reported to upstream to issue.

neheb: No idea which APIs are used. I don't care much for python.

Upstream will find out eventually. Python 3.9 only recently got released.

BKPepe: This is not a good approach.

jefferyto: See also #13762 - I'm not a fan of just getting things to compile without making sure it works / fixing the underlying issue.

BKPepe: I would be rather if I don't know about that. As this is such bad.

PR #13804, opened 2020-10-29
DDNS cloudflare: Separate __HOST and __DOMAIN variables

Qrbaker:
Maintainer: @feckert (find it by checking history of the package Makefile)
Run tested: 
* Arch: ARMv7 Processor rev 1 (v7l)
* model: Linksys WRT3200ACM 
* version: OpenWrt 19.07.0-rc1 r10649-c4fdb377a2

## Description
This small update allows the user to more effectively use the LuCI GUI to create Cloudflare DDNS entries with subdomains. Previously, the user needed to ignore the "lookup hostname" field and instead write into the domain field `subdomain@example.com`.

Now the user may put `subdomain.example.com` in the lookup hostname field, and `example.com` in the domain field.

If configuring over CLI, the config will look similar to the following:

```
config service 'my_ddns'
        option service_name 'cloudflare.com-v4'
        option enabled '1'
        option username '<CloudFlare email>'
        option password '<CloudFlare API Key>'
        option use_https '1'
        option lookup_host 'subdomain.example.com'
        option domain 'example.com'
```
## Example of GUI Configuration
image


neheb:
Needs a PKG_RELEASE bump.

Also, please rebase against master.

Issue #13735, opened 2020-10-21
umurmur: config file location is wrong


fatbob313:
Maintainer: @micmac1
Environment: trunk

Description:
Since some time back, maybe a year, the location of the configuration file is searched for at:
```/usr/etc/umurmur.conf``` 
instead of the correct place:
```/etc/umurmur.conf```
But sometimes when I update to a new release it uses the correct path, which is confusing, is it a race condition in the build system?

I tried decoding how the path is set. In uMurmur I can find where it is set:
https://github.com/umurmur/umurmur/blob/55689f8918dde299c47c06eee4b8eda4a1c9889e/src/config.h.in#L14

I'm CMake illiterate. Is CMAKE_INSTALL_PREFIX wrong var to use here? Should the issue be directed to uMurmur instead?

micmac1:
Hi there,

You already reported this upstream [here](https://github.com/umurmur/umurmur/issues/146). I think upstream is the right place to discuss this. In OpenWrt this could probably be worked around by patching the init script to call umurmurd with the "-c /etc/umurmur.conf".

Anyway, I'm not the maintainer here. Removing myself from this issue.

Regards,
Seb

PR #13812, opened 2020-10-30
umurmur: Specify config file location at startup

fatbob313:
...to workaround issue where default location is set to '/usr/etc/...'

Maintainer: me 
Compile tested: On Ubuntu WSL. Target arm-cortex-a15 (Netgear Nighthawk R7800), OpenWrt r14728-21cf17bb4c
Run tested: On same as above. Verified startup OK)

Refer to issue https://github.com/openwrt/packages/issues/13735#issue-726463731

### DIFF
--- a/net/umurmur/patches/020-specify-conffile-path.patch
+++ b/net/umurmur/patches/020-specify-conffile-path.patch
@@ -1,0 +1,13 @@
+diff --git a/openwrt/files/umurmur.init b/openwrt/files/umurmur.init
+index 392cd69..7223d03 100644
+--- a/openwrt/files/umurmur.init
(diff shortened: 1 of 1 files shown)


BKPepe:
Hello,

Thank you for your contribution. Would you please send this PR first to upstream and then backport the patch here?

Also, in the Makefile of urmurmur you need to increase the version of PKG_RELEASE. This is for case when someone has already installed umurmur and the system will not be aware that there is a new version. Right now, it will just affect new installations.

fatbob313:
Thanks!
I 'll fix this upstream instead + a bunch of other issues and get back with a new pull req. Closing this now.

PR #13819, opened 2020-11-01
docker-ce: Move creation of /tmp/docker above first use

voltagex:
Signed-off-by: Adam Baxter <github@voltagex.org>
Maintainer: @feckert (find it by checking history of the package Makefile)
Compile tested: N/A
Run tested: N/A

Description: If you've got alt_config_file defined, the init script will try to create a symlink as /tmp/dockerd/daemon.json before /tmp/dockerd exists. Moving the mkdir call prevents the init script from failing here.

### DIFF
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -124,6 +124,7 @@ process_config() {
 
 	config_load 'dockerd'
 
+	mkdir -p /tmp/dockerd
 	config_get alt_config_file globals alt_config_file
 	[ -n "$alt_config_file" ] && [ -f "$alt_config_file" ] && {
 		ln -s "$alt_config_file" "$DOCKERD_CONF"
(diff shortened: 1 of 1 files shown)


feckert:
Could you please update the commit header and add the body from you pullrequest description to the commit?

G-M0N3Y-2503:
hmmm, I had https://github.com/openwrt/packages/pull/13796/commits/5d610c81cd0cca75ecc70cf3a48495d4a72246bf that will fix/change this, but if you want this fix in sooner than that one I guess that works too.

Do we also need to update the `PKG_RELEASE` with this too?

voltagex:
Updated.

voltagex:
@G-M0N3Y-2503 your fix looks better, I can close this if you want.

G-M0N3Y-2503:
I'm happy either way, It depends on your timeline whether you're willing to wait for the other PR or not.

feckert:
Lets close this and take commit from this P/R https://github.com/openwrt/packages/pull/13796

PR #13827, opened 2020-11-02
cgroupfs-mount: fix symbolic link name

cobaltchang:
Fix the symbolic link name to /etc/rc.d/S01cgroupfs-mount.

Description:
The created symbolic link S1cgroupfs-mount will be executed after S10fstab.
Fix the START setting to ensure that cgroupfs-mount is executed at S01.

### DIFF
--- a/utils/cgroupfs-mount/Makefile
+++ b/utils/cgroupfs-mount/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cgroupfs-mount
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/tianon/$(PKG_NAME)
--- a/utils/cgroupfs-mount/files/cgroupfs-mount.init
+++ b/utils/cgroupfs-mount/files/cgroupfs-mount.init
@@ -1,6 +1,6 @@
 #!/bin/sh /etc/rc.common
 
-START=1
+START=01
 
 boot() {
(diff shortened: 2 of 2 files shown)


neheb:
ping @G-M0N3Y-2503

G-M0N3Y-2503 (changes requested):
LGTM, just need to increment the [`PKG_RELEASE`](https://github.com/openwrt/packages/blob/master/utils/cgroupfs-mount/Makefile#L4) I believe.

Also, I'm curious as to what issues you had before this change?

feckert:
@G-M0N3Y-2503 

> Also, I'm curious as to what issues you had before this change?

as mentioned by @cobaltchang:
The created symbolic link S1cgroupfs-mount will be executed after S10fstab.
Fix the START setting to ensure that cgroupfs-mount is executed at S01.

This i also done in https://github.com/openwrt/packages/blob/master/net/sshtunnel/files/sshtunnel.init#L8

cobaltchang:
@G-M0N3Y-2503 
Thank you very much, I updated PKG_RELEASE and force-pushed.

feckert:
Thanks

G-M0N3Y-2503:
> as mentioned by @cobaltchang:
> The created symbolic link S1cgroupfs-mount will be executed after S10fstab.
> Fix the START setting to ensure that cgroupfs-mount is executed at S01.

Oh, I wasn't disputing whether it was wrong before, I was just wondering if anything was broken.  
I couldn't find a `S10fstab`, just a [`S40fstab`](https://github.com/openwrt/openwrt/blob/master/package/system/fstools/files/fstab.init#L4) so I couldn't reproduce it myself.
Also, I thought the only thing that had cgroup v1 mounts in it was `procd`?

PR #13836, opened 2020-11-03
python-markdown: update to version 3.3.3

BKPepe:
Maintainer: me

Description:
- Update to version 3.3.3
Changelog:  https://python-markdown.github.io/change_log/


jefferyto:
It looks like the setuptools run-time dependency was removed in 3.2.2 - can you confirm if python-setuptools is still needed in `DEPENDS`?

BKPepe:
Thanks for feedback and letting me know about this. I will check it.

BKPepe:
Indeed, python3-setuptools dependency is no longer needed. Removing it.

PR #13841, opened 2020-11-03
unbound: update to 1.1.12

### LINKED ISSUE #13178: unbound: allow forwarding to localhost without setting dns_assist
Maintainer: @EricLuehrsen 
Environment: OpenWrt master

Description:
With the introduction of `dns_assist` from https://github.com/openwrt/packages/commit/e81529596df88eae272c00bf8ac69b15e19a5bd5, unbound init script "helpfully" translated this:

```
config zone
	option enabled '1'
	option fallback '1'
	option zone_type 'forward_zone'
	list server '127.0.0.1@5053'
	list server '127.0.0.1@5054'
	list server '127.0.0.1@5055'
	list server '127.0.0.1@5056'
	list server '127.0.0.1@5057'
	list zone_name '.'
```

into:

```
forward-zone:
  name: .
  forward-host: 127.0.0.1@5053
  forward-host: 127.0.0.1@5054
  forward-host: 127.0.0.1@5055
  forward-host: 127.0.0.1@5056
  forward-host: 127.0.0.1@5057
  forward-first: yes
  forward-tls-upstream: no
```

When `dns_assist` is undefined, which is completely wrong (`forward-host` should be `forward-addr`) and breaks forwarding to `https-dns-proxy`, trashing my router DNS. An option to override this "smart" feature is very much needed.

The walkaround right now is to override the fallback `dns_ast=0` to `dns_ast=1` here:

https://github.com/openwrt/packages/blob/95362efd29cef6433c820fd15bf33e73637a336c/net/unbound/files/unbound.sh#L479-L482

### LINKED ISSUE #13623: [unbound] update to source release 1.12.0
Maintainer: @EricLuehrsen 
____
Description:
source release version 1.12.0 https://github.com/NLnetLabs/unbound/releases/tag/release-1.12.0 introduces DoH for downstream clients. It would be appreciated for the package to be updated.

### LINKED ISSUE #13750: Unbound: remove 'interface: 0.0.0.0 / ::0' from generated config - it is redundant, limiting and plain insecure
Maintainer: @EricLuehrsen
Environment: 19.07

Description:
Please remove ```interface: 0.0.0.0``` (and similar ip6 wildcard) directives added by UCI's ```option protocol``` handler
1. this prevents one to configure unbound to listen only on ```lo``` (which is the default upstream behavior and doesn't require any cfg options) or some particular ifaces via ```unbound_srv.conf```
2. this is redundant - ```interface-automatic``` (UCI ```interface_auto```, enabled by-default) achieves similar outcome for those who want/need it
3. It is just nonsense to have non-overridable listen-on-wildcard and thus force users to patch security with acl and/or firewall rules or doing ```manual_conf 1```

I also see no reason for ```outgoing-interface: 0.0.0.0```. Default upstream behavior with no ```outgoing-interface``` specified is to use all available ifaces. So it is also redundant and limiting in terms of configuration flexibility/freedom.

EricLuehrsen:
Maintainer: me
Tested: Linksys WRT3200ACM
Description: resolves #13178, resolves #13623, resolves #13750, resolves #13852
If DoH is enabled, then a few glitches in the upstream tuning options that are being worked out. Default compile option is disabled, but it is available if someone wants to try it.


net/unbound/Makefile.keep
```diff
@@ -0,0 +1,269 @@
+#
```
neheb, line 1: What is this file?

EricLuehrsen: fix-up artifact. removed.

PR #13798, opened 2020-10-28
treewide: use new extra_command function for init

feckert:
Maintainer: The follwing maintainer should get pinged to verify the changes I have made to call the new extra_command function.

- [ ] @Shulyaka for postfix
- [x] @dibdot for adblock
- [ ] @liudf0716 apfree-wifidog
- [x] @dibdot banip
- [ ] @dangowrt gnunet
- [x] @gladiac1337 haproxy
- [x] @Ansuel nginx
- [x] @yousong openvswitch
- [ ] @chris5560 radicale
- [x] @jefferyto seafile-server
- [x] @stangri simple-adblock
- [ ] @mstorchak tgt
- [x] @zioproto tinc
- [x] @dibdot tavelmate
- [x] @stangri vpn-policy-routing
- [x] @stangri vpnbypass
- [x] @cotequeiroz wifidog
- [x] @G-M0N3Y-2503 docker-ce

Compile tested: not needed only script changes
Run tested: todo if package maintainer verified my changes

Description:
This is a draft P/R if the following openwrt [P/R](https://github.com/openwrt/openwrt/pull/3246) is merged in the openwert master. For consistency the init scripts in the package feed must also use the new function extra_command, to ensure that the help is displayed consistently.

If every think works as expected I will also update the PKG_RELEASE number of each package.

PR #13849, opened 2020-11-03
adblock: adapt openwrt rc.common changes

dibdot:
Maintainer: me / @dibdot
Compile tested: -
Run tested: OpenWrt SNAPSHOT, r14851-08d90a75f9

Description:
* since openwrt master has merged the depending P/R, the old
extra_help/extra_commands syntax is no longer working, see #13798 for
reference
* removed test.sh script from package


BKPepe:
This was not merged yet https://github.com/openwrt/packages/pull/13785 so I would keep the test file

PR #13853, opened 2020-11-04
mwan3: code cleanup and fixes

feckert:
Maintainer: me
Compile tested: x86_64, APU3, OpenWrt master
Run tested: x86_64, APU3, OpenWrt master

Description:
This is a maintenance P/R with the following goals.
- Clean up the source code
- Fix pending errors through community tests which are found through different test scenarios

Fixes #13823 
Fixes #13865

PR #13857, opened 2020-11-05
mwan3: reload unreachable/blackhole rules on hotplug, fix mwan3track bugs, resolve deadlock

aaronjg:
Maintainer: @feckert 
Compile tested: Scripts Only
Run tested: Snapshot

Description:
when the network procd service restarts, it flushes the ip rules. We
need to add these rules back. Since hotplug events are triggered when
the networks come back online, adding this call to the hotplug script
is the most convenient place to refresh the rules.

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.10.1
+PKG_VERSION:=2.10.2
 PKG_RELEASE:=1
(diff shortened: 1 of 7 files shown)


feckert:
@aaronjg I have added your commits to my pending [P/R](https://github.com/openwrt/packages/pull/13853).
So that we can work together on the matter.

Could you please add your Signed-of-by to the following commit https://github.com/openwrt/packages/pull/13853/commits/59d12598b50edbc5a7a1f178c3c291524bd85b13 in my P/R?

Therefore close this P/R.

aaronjg:
Thanks. I can't push to your repo, but I added the sign-off to my commit here. Feel free to amend the commit in your P/R to include my sign-off.

PR #13358, opened 2020-09-10
CI: add super-linter for pull requests

aparcar:
This commit adds a linter which checks most common filetypes, including
Shell and Lua. Ideally this improves the quality of especially `init`
scripts written in Shell.

PR #13861, opened 2020-11-06
CI: Revert "add super-linter for pull requests"

aaronjg:
Initial motivation for this commit was 'Just merge and see how people like it?'
https://github.com/openwrt/packages/pull/13358#issuecomment-692309158

It seems that many contributors don't like it.
https://github.com/openwrt/packages/issues/13395

So this should be reverted until issues with the linter can be
resolved.

This reverts commit 2b5de22202dafe849bfbce3adf1488d542ba8b24.


champtar:
Except if we can make it optional / ignore failure, +1 for revert

BKPepe:
ping @aparcar

Issue #13759, opened 2020-10-24
mwan3: $DEVICE environment variable not populated on ifdown event


jamesmacwhite:
Maintainer: @feckert
Environment: OpenWrt 19.07.3 r11063-85e04e9f46 / LuCI openwrt-19.07 branch git-20.234.16293-a82f67e

mwan3 2.8.13-1

Description:

The $DEVICE environment variable does not seem to be populated on an ifdown event. I implemented some custom email notifications functionality within `/etc/mwan3.user` recently, to test I restarted a network interface being tracked by mwan3. This triggered four email notifications, each for the disconnected, ifdown, ifup and connected actions. On the ifdown action, the $DEVICE value was blank, for all others it was populated.

ptpt52:
ping @aaronjg

aaronjg:
@jamesmacwhite I believe this is an issue with OpenWRT. Can you add a hotplug script outside of mwan3, disable mwan3 and check if the DEVICE is set on ifdown?

jamesmacwhite:
@aaronjg You seem to be correct here, it's not a mwan3 specific problem

This is the custom hotplug I made for testing:

```
[ "${ACTION}" = "ifup" ] && {
    logger -t hotplug "Device: ${DEVICE} / Action: ${ACTION}"
} 
```

The output was the following, when reloading the network interface

```
Sat Oct 24 17:47:57 2020 user.notice hotplug: CUSTOM HOTPLUG: Device:  / Action: ifdown
```

`$DEVICE` was indeed blank.

aaronjg:
I'm not sure why it is not set on ifdown. It could be that in some cases, such as tunnel devices, the device will have been destroyed by the time the ifdown hotplug event happens. Maybe file a bug report/feature request with a use case for having the DEVICE set in the ifdown event call.

jamesmacwhite:
@aaronjg Interesting theory. Although I don't think it is specifically related to tunnel related interfaces. I tested it on an independent WAN connection through Ethernet, the only unique element would be using a LAN port as a WAN port via a VLAN config, but otherwise it's pretty much a standard network interface. It isn't really an issue, just maybe worth highlighting in the docs for `/etc/mwan3.user` that $DEVICE isn't available on ifdown.

aaronjg:
> It isn't really an issue, just maybe worth highlighting in the docs for `/etc/mwan3.user` that $DEVICE isn't available on ifdown.

[...]

PR #13867, opened 2020-11-06
mwan3: Document hotplug events and $DEVICE not being available on ifdown

jamesmacwhite:
Maintainer: @feckert 
Compile tested:  N/A
Run tested: N/A

Description:

Relates to https://github.com/openwrt/packages/issues/13759, $DEVICE is not available on ifdown, so it cannot be used in any custom logic in `/etc/mwan3.user` workaround would be to use $INTERFACE instead which is available on ifdown. $DEVICE is usable on any other event.

While it has been discussed that ifupdate is not mwan3 related, it is possible to hook into this action event with the mwan3.user file, therefore I think it can be mentioned, but clarified it is not a mwan3 event.


feckert:
Could we please remove the line https://github.com/openwrt/packages/pull/13867/files#diff-3886166364a8812230202269afb1b464f007ad9d9b186815632c175bdd9e60b9R12 Because if we add this then we have to add all other actions as well. But mwan3 does only use for now ifup, ifdown, connected and disconnected. In the documentation we have already mentioned this https://openwrt.org/docs/guide-user/network/wan/multiwan/mwan3?s[]=mwan3#alertsnotifications

The hint for $DEVICE is sufficient

jamesmacwhite:
@feckert Thanks! I hadn't updated the documentation at the time of this PR but have since rewrote and expanded the section around the `/etc/mwan3.user` file so I'll amend accordingly.

feckert:
If we merge this [P/R](https://github.com/openwrt/packages/pull/13889#pullrequestreview-528169654). We react also on ifupdate events. So I think the ifupdate info add for mwan3.user should go this P/R. Then it fits together

jamesmacwhite:
@feckert I've dropped ifupdate for now, I've added a reference to the URL path of the alerts/notifications documentation which covers the purpose and details of the `/etc/mwan3.user` file.

feckert:
@jamesmacwhite Sorry if I'm annoying but can you please also update the PKG_RELEASE number.

jamesmacwhite:
@feckert No problem! I've just resynced the latest changes in master, as this PR is behind, so the Makefile is old. I've just bumped the PKG_RELEASE to `2` as this is a minor change.

feckert:
Thanks

PR #13875, opened 2020-11-08
udpspeeder: Add new package udpspeeder to implement FEC for VPN over lossy links

codemarauder:
udpspeeder: Add new package udpspeeder to implement Forward Error Correction (FEC) for UDP VPNs like OpenVPN
Maintainer: @codemarauder
Compile tested: Yes, 19.07.4, 18.06.2
Run tested: x86_64 PCEngines APU, 19.07.4, 18.06.2
Description:
A Tunnel which Improves your Network Quality on a High-latency Lossy Link by using Forward Error Correction,for All Traffics(TCP/UDP/ICMP)
It does it by sending redundant packets and re-arranging them to account for packet loss over the link. It uses Reed_Solomon code. By setting appropriate FEC parameters, a packet loss of 20% can be reduced to less than 1% at the cost of increased latency and 1.5 to 1.8 times bandwidth utilisation.

Signed-off-by: Nishant Sharma <codemarauder@gmail.com>


net/udpspeeder/Makefile
```diff
@@ -0,0 +1,63 @@
...
+# Copyright (c) 2017 Yu Wang <wangyucn@gmail.com>
+# Copyright (c) 2020 Nishant Sharma <nishant@hopbox.in>
+#
+# This is free software, licensed under MIT
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=udpspeeder
+PKG_VERSION:=20200818.0
```
neheb, line 11: There's a newer version available.

net/udpspeeder/Makefile
```diff
@@ -0,0 +1,63 @@
...
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
```
neheb, line 14: Upstream has tagged releases. Please use those instead of git.

git grep the repo for "codeload" to see examples.

net/udpspeeder/Makefile
```diff
@@ -0,0 +1,63 @@
...
+PKG_SOURCE_URL:=https://github.com/wangyu-/UDPspeeder.git
+PKG_SOURCE_VERSION:=13d5e40473f354939dd196980a4e3fcc6affade8
+PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=Nishant Sharma <nishant@hopbox.in>
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)
```
neheb: Not needed at all.

net/udpspeeder/Makefile
codemarauder: Thanks for your review. Updated to the newer version.

net/udpspeeder/Makefile
codemarauder: Removed proto git and instead using codeload. Thanks again.

net/udpspeeder/Makefile
codemarauder: This is also fixed.

net/udpspeeder/Makefile
```diff
@@ -46,7 +47,7 @@ define Build/Prepare
 	$(Build/Patch)
 endef
 
-define Package/udpspeeder/install
+define Package/UDPspeeder/install
 	$(INSTALL_DIR) $(1)/usr/bin
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/speederv2_cross $(1)/usr/bin/udpspeeder
```
neheb, line 52: I'd use PKG_INSTALL with PKG_INSTALL_DIR.

codemarauder: This software doesn't have any configure script and builds the single binary in the build directory itself. That's why I have to use PKG_BUILD_DIR here.

neheb: Then why do you have autoreconf above?

codemarauder: I had picked the Makefile from some other source and modified it. This is my first time sending pull request for adding a  completely new package.

Should I convert it to a draft?

codemarauder: Removed autoreconf. Please review when you have time for further changes required.

One thing that I am unable to do is run multiple named instances as it happens with OpenVPN. As of now, we can start / stop / restart all the instances at once but not selectively. I will work on the init script to be able to do that soon.

neheb:
I'd squash all of the commits.

codemarauder:
Squashed all the commits to a single one. Thanks @neheb  for review and handholding through the process.

PR #13877, opened 2020-11-08
transmission-web-control: remove Transmission SSL variants

BKPepe:
Maintainer: @ysc3839 
Compile tested: N/A
Run tested: N/A

Description:
In recent commits, there were removed Transmission SSL variants and
there is just used one variant of transmission-daemon. (https://github.com/openwrt/packages/pull/13723).

Let's adjust it here as well.


ysc3839:
LGTM

PR #13880, opened 2020-11-08
banip: removed logd check since logd dep has been dropped

### LINKED ISSUE #13820: banip: dependency on logd breaks syslog-ng installation
Hi,

Building an image fails, when syslog-ng and banip is selected, because banip depends on logd.

logd and syslog-ng both want to provide /sbin/logread

### Solution
depend on logd OR syslog-ng

jmarcet:
Maintainer: Dirk Brenken <dev@brenken.org>
Compile tested: x86_64 master
Run tested: x86_64 master

Description:
#13848 dropped the logd dependency in commit 14a69715899accff66009d45a4a13ab7cb5a38b7 trying to fix #13820 

I say trying because **banip.sh** checks whether **/etc/init.d/log** is running at runtime, failing when syslog-ng is used.

### DIFF
--- a/net/banip/Makefile
+++ b/net/banip/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=banip
 PKG_VERSION:=0.3.12
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 PKG_LICENSE:=GPL-3.0-or-later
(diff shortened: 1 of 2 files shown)


neheb:
ping @dibdot

dibdot:
Thanks! Please bump the "PKG_RELEASE" in the package Makefile as well and squash your changes.

PR #13881, opened 2020-11-08
mwan3: fixup ipset and install errors

aaronjg:
Maintainer: @feckert 
Compile tested: 19.07.4 SDK, Snapshot
Run tested: 19.07.4

- fixup ipset errors on startup
- restore +x flags on mwan3 executables
- add IPKG_INSTALLROOT to resolve image builder errors


feckert:
@aaronjg 

https://github.com/openwrt/packages/pull/13881/commits/7837061aaee51076351d0d3f2b0b19e5ab1f7b6c
If we use the makefile targets $(INSTALL_*) then we don't have to set the files with the right bits in git.
This will be done automatically with the makefile targets. I have changed.

net/mwan3/files/etc/init.d/mwan3
```diff
@@ -1,9 +1,8 @@
 #!/bin/sh /etc/rc.common
 
-. /lib/functions.sh
-. /lib/mwan3/common.sh
-. /lib/functions/network.sh
-. /lib/mwan3/mwan3.sh
+. $IPKG_INSTROOT/lib/mwan3/common.sh
```
feckert, line 3: Could you please add quotation marks around this?

aaronjg: It seems that the practice in openwrt is to not quote those (eg `/etc/rc.common`).

feckert: I would say this is not uniform.

package feed:
https://github.com/openwrt/packages/blob/ffdbc7a637df8e3f39dd6916405f78840edccfb4/net/ocserv/files/ocserv.init#L6
https://github.com/openwrt/packages/blob/0038d6e971bf98ef440e15cc3a71e643d4fa65ba/net/nut/files/nut-monitor.init#L174
https://github.com/openwrt/packages/blob/c05e92a58d84c20750f2974da6a89e872d1a1041/net/fwknop/files/fwknopd.init#L11

openwrt:
https://github.com/openwrt/openwrt/blob/5792f6a1043d686d4ee3c8b8f7ce070ac1bb3e5d/package/system/procd/files/procd.sh#L43
https://github.com/openwrt/openwrt/blob/0232bf601d6c8c174faa4c5bac1f6c86b20853c9/package/base-files/files/etc/rc.common#L4

aaronjg:
I'm also now getting some errors with `mwan3 interfaces`

```
# mwan3 interfaces
Interface status:
Command failed: Not found
Failed to parse message data
WARNING: Variable 'interfaces' does not exist or is not an array/object
WARNING: Variable 'wan' does not exist or is not an array/object
 interface wan is online 00h:00m:00s, uptime 00h:00m:00s and tracking is active
```

But it seems to have happened before this P/R. Can you reproduce the issue?

net/mwan3/files/etc/init.d/mwan3
feckert: Most init scripts move the include to the function so that they don't get an error on install during image build.
That would be one possibility, but I don't like that.
If i need it in different functions in init i always have to include it.

For example https://github.com/openwrt/packages/blob/be6312296c6132d494505f5c6f0b7bb4bb3aec6d/utils/collectd/files/collectd.init#L191
https://github.com/openwrt/packages/blob/10da8b0b8e45136cb5b10c1fd105ea0844bd998e/net/banip/files/banip.init#L94

aaronjg: Okay, looks like it can go either way. I'll add the quotation marks around it to accommodate shellcheck.

feckert:
I can't reproduce this right away, but when I make the `ubus call mwan3 status`, the interface is online
 But when I execute the command `mwan3 interfaces` then the interface is on error!

```
root@G3400-LWV / # mwan3 interfaces
Interface status:
 interface wan is error and tracking is active
 interface wwan is offline and tracking is paused
```

I think the evaluation is broken on this section
https://github.com/openwrt/packages/blob/16938c84f4b7ea72baa99825539fa13b1319783a/net/mwan3/files/lib/mwan3/mwan3.sh#L1034

When I look at the output of the command in the if state ment then I get the following error:
```
1001:   from all iif eth0 lookup 1
2001:   from all fwmark 0x100/0x3f00 lookup 1
3001:   from all fwmark 0x100/0x3f00 unreachable
iptables: No chain/target/match by that name.
default via 172.16.2.254 proto static src 172.16.2.171 metric 100
```
This line is suspicious!
`iptables: No chain/target/match by that name.`

feckert:
This is my mangle table
iptables-mangle

aaronjg:
Sorry, previous error that I was having was because the rpcd script did not have the +x set when I copied it over. I deleted that comment.

The "error" state actually looks like it is from the fact that there is no "mwan3_iface_in_wan" chain, but your mangle diagram clearly shows the wan chain (also, that's a very cool visualization, how do you generate those?)

Why do you think it is broken on that line? It should match with 
`1001:   from all iif eth0 lookup 1`

aaronjg:
I added some error codes to the interface status, which should make debugging these sorts of errors easier.

feckert:
> The "error" state actually looks like it is from the fact that there is no "mwan3_iface_in_wan" chain, but your mangle diagram clearly shows the wan chain (also, that's a very cool visualization, how do you generate those?)

@aaronjg I use this tool https://metacpan.org/release/App-Iptables2Dot on my development machine.
Unfortunately it does not run on the router. That would realy a great benefit

> Why do you think it is broken on that line? It should match with
> `1001: from all iif eth0 lookup 1`

Forget what I said in the comments above, the back and forth in development breaks my image and that was just noise!

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -1036,11 +1036,23 @@ mwan3_report_iface_status()
 
 	if [ -z "$id" ] || [ -z "$device" ]; then
 		result="offline"
-	elif [ -n "$($IP rule | awk '$1 == "'$((id+1000)):'"')" ] && \
-		     [ -n "$($IP rule | awk '$1 == "'$((id+2000)):'"')" ] && \
-		     [ -n "$($IP rule | awk '$1 == "'$((id+3000)):'"')" ] && \
-		     [ -n "$($IPT -S mwan3_iface_in_$1 2> /dev/null)" ] && \
-		     [ -n "$($IP route list table $id default dev $device 2> /dev/null)" ]; then
+	else
+		error=0
```
feckert, line 1040: Could we please move the error check out of the else branch? So we do a check on every run.
So we do not need the `:` in the if statement on line https://github.com/openwrt/packages/pull/13881/commits/8bdbf259f894a9a06df5bc0384e17fc165046dd7#diff-a79526f0a000ba74d1bdaee37e1c9544d0e215cf231a42ab37fa6d29d13c0f64R1054

net/mwan3/files/usr/libexec/rpcd/mwan3
```diff
@@ -77,70 +77,70 @@ get_mwan3_status() {
 	local online=0
 	local offline=0
 	local up="0"
-	local enabled device time_p time_n time_u time_d status track_status
```
feckert, line 80: Could you also please move this two separate commits?
- One for remove unnecessary network_get_device_call from rpcd
- And another for fixup indenting on status call

Then this is enforceable.

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -1017,7 +1017,7 @@ mwan3_get_iface_hotplug_state() {
```
feckert, line 1017: You also mentioned in your commit message `don't send track command output in ubus status call`
but i can not find this in this commit?

feckert (changes requested):
Could we also split this commit in three commits?
- One for `connected ipset handled on mwan3rtmon, so not needed in hotplug
` as the commit message already describe.
- One for the fix that a missed on my last P/R change `$MWAN3_STARTUP" != "init"`
- One for adding LOG message on ipset creation fail

This makes it comprehensible to others.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: I changed how the files are processed in this loop. `for file in $MWAN3TRACK_STATUS_DIR/${iface}/*; do`  Before, it was taking everything that started with "TRACK_*" and putting it into the ubus status message. This included "TRACK_OUTPUT" which did not need to go there.

net/mwan3/files/lib/mwan3/mwan3.sh
aaronjg: I don't follow what you are proposing.

In master, right now if the `[ -z "$id" ] || [ -z "$device" ]` check passes, then the four checks are not done. Same as in this commit.

In master, it will do the four checks, and if they all pass, it goes on to get more status information. If any fail, it reruns the checks and if they didn't all fail, and then outputs "error", otherwise it outputs offline.

Now, it will  `[ -z "$id" ] || [ -z "$device" ]`, and if it passes, then skip the further checks. If it fails, it will do the checks and encode the results in the error code. This way, if any are failing, it does not need to run the check again. 

We can't do all the checks if we don't have `$id` or `$device`

net/mwan3/files/lib/mwan3/mwan3.sh
feckert: A ok I understand.
So could you also please do this in another commit so someone else could also follow.

aaronjg:
@feckert - I sliced up the commits a bit more. Take a look now.  I also found another issue in `15-mwan3`. The "cmd" variable is not set when the hotplug is called by hotplug daemon, so we need to check that MWAN3_STARTUP != init, rather than MWAN3_STARTUP = cmd.

net/mwan3/files/usr/libexec/rpcd/mwan3
```diff
@@ -79,67 +79,68 @@ get_mwan3_status() {
 	local up="0"
 	local enabled time_p time_n time_u time_d status track_status
 
+	if [ "${iface}" != "${iface_select}" ] && [ "${iface_select}" != "" ]; then
+                return
+        fi
```
feckert, line 84: \<space\> vs. \<tab\>

feckert:
On my system it runs so far.
Ready to merge?
I would suggest to do the rpcd [dependency](https://github.com/openwrt/packages/issues/13896) for master in a separate P/R, because as you suggest we have to change more than in openwrt-19.07.

I want to fix the bugs and stabilize the version as soon as possible.

aaronjg:
Yep. Thanks.

PR #13882, opened 2020-11-08
NetStinky IDS for detecting IoCs on your network

null-cipher:
Signed-off-by: Andrew Mackintosh <amackint@waikato.ac.nz>

Maintainer: me / @null-cipher
Compile tested: Raspberry Pi 3 / brcm2708-bcm2710, OpenWrt 19.07.4
                Hyper-V VM / x86_64, OpenWrt 19.07.4
Run tested: Raspberry Pi 3 / brcm2708-bcm2710, OpenWrt 19.07.4
            Hyper-V VM / x86_64, OpenWrt 19.07.4

Description:
The NetStinky IDS is a component of the NetStinky suite of tools. It
monitors the traffic on the LAN interfaces of your router for
Indications of Compromise (IoCs), drawn from an auto-updating list of
definitions.  IoCs are subsequently reported to the NetStinky smartphone
applications.

### DIFF
--- a/net/netstinky/Makefile
+++ b/net/netstinky/Makefile
@@ -1,0 +1,62 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=netstinky
+PKG_VERSION:=1.0.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=nsids-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/wanduow/netstinky-ids/releases/download/v$(PKG_VERSION)/
+PKG_HASH:=143e8d7a9ee7f58893d92b065aed7eca35798f5807b07d9a78f404ea8614e216
+PKG_BUILD_DIR:=$(BUILD_DIR)/nsids-$(PKG_VERSION)
+
+PKG_MAINTAINER:=Andrew Mackintosh <amackint@waikato.ac.nz>
+PKG_LICENSE:=BSD-2-Clause
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/netstinky
+  SECTION:=net
+  CATEGORY:=Network
+  DEPENDS:=+libpcap +libuv +libopenssl +umdns
+  TITLE:=NetStinky IDS for detecting IoCs on an active network
+  URL:=https://netstinky.wand.net.nz/
+endef
+
+define Package/netstinky/description
+  The NetStinky IDS is a component of the NetStinky suite of tools. It
+  monitors the traffic on the LAN interfaces of your router for Indications
+  of Compromise (IoCs), drawn from an auto-updating list of definitions.
+  IoCs are subsequently reported to the NetStinky smartphone applications.
+endef
+
+CONFIGURE_ARGS += \
+	--enable-mdns=no \
+	--enable-updates
+
+define Package/netstinky/conffiles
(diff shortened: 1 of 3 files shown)


net/netstinky/files/nsids.init
```diff
@@ -0,0 +1,65 @@
...
+    fi
+    procd_append_param command -p "${port}"
+    [ -n "${updatehost}" ] && procd_append_param command --update-host "${updatehost}"
+    [ -n "${updateport}" ] && procd_append_param command --update-port "${updateport}"
+    [ "${sslnoverify}" -eq 1 ] && procd_append_param command --ssl-no-verify
+
+    procd_set_param stderr 1
+    procd_set_param stdout 1
+
+    procd_set_param respawn ${respawn_threshold:-3600} ${respawn_timeout:-5} ${respawn_retry:-5}
```
karlp, line 43: if those are the defaults, just leave them off, otherwise it's just cargo culting that will be harder to determine the true cause for down the road.

null-cipher: I've removed line 43 as it was in fact the defaults.

The `nsids` program logs to stdout and stderr (depending on severity) as it is intended to run on a few different platforms and doing it that way prevents us from having to depend on a particular logging solution.

As I understand it, procd expects a service to be running in the foreground anyway, so logging via stdout/stderr seemed appropriate.

net/netstinky/Makefile
```diff
@@ -0,0 +1,63 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=netstinky
+PKG_VERSION:=1.0.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=nsids-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/wanduow/netstinky-ids/releases/download/v1.0.0/
```
neheb, line 8: change that last part to $(PKG_VERSION)

net/netstinky/Makefile
```diff
@@ -0,0 +1,63 @@
...
+PKG_HASH:=143e8d7a9ee7f58893d92b065aed7eca35798f5807b07d9a78f404ea8614e216
+PKG_BUILD_DIR:=$(BUILD_DIR)/nsids-$(PKG_VERSION)
```
neheb, line 10: Is this necessary?

net/netstinky/Makefile
```diff
@@ -0,0 +1,63 @@
...
+
+PKG_MAINTAINER:=Andrew Mackintosh <amackint@waikato.ac.nz>
+PKG_LICENSE:=BSD-2-Clause
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_PARALLEL:=1
+PKG_FIXUP:=autoreconf
```
neheb, line 17: Is this needed?

net/netstinky/Makefile
null-cipher: Done.

net/netstinky/Makefile
null-cipher: It appears to be. The package is named `netstinky` but the tarball is `nsids` (NetStinky IDS) when I remove the PKG_BUILD_DIR it will untar into nsids-1.0.0 but then tries to build in netstinky-1.0.0.

net/netstinky/Makefile
null-cipher: It doesn't appear to be! The project does use autotools, but in the testing I just did it seems to work OK with cross compilation. I added this in because the [documentation](https://openwrt.org/docs/guide-developer/packages#pkg_fixup) seemed to suggest it was necessary.

PR #13883, opened 2020-11-09
icu: update to 68.1

nxhack:
Maintainer: me 
Compile tested: head r14870-93adba4, aarch64, arm, i386, mips, mips64, mips64el, mipsel, x86_64
 Run tested: (qemu-5.1.0) aarch64, x86_64

Description: 
It updates to CLDR 38. New features including locale-dependent smart unit preferences (road distance, temperature, etc.) and locale ID canonicalization conformant with CLDR.

Signed-off-by: Hirokazu MORIKAWA <morikw2@gmail.com>

PR #13893, opened 2020-11-10
php7: fix for icu 68.1

nxhack:
Maintainer: @mhei 
Compile tested: aarch64
 Run tested: none

Description:
fix for icu 68.1

https://github.com/openwrt/packages/pull/13883
https://github.com/php/php-src/commit/975735c02751d9409c4a4757e7b70d217f0f54fe

### DIFF
--- a/lang/php7/Makefile
+++ b/lang/php7/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=php
 PKG_VERSION:=7.4.12
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_MAINTAINER:=Michael Heimpold <mhei@heimpold.de>
 PKG_LICENSE:=PHP-3.01
--- a/lang/php7/patches/1005-support_for_icu4c_68_1.patch
+++ b/lang/php7/patches/1005-support_for_icu4c_68_1.patch
@@ -1,0 +1,13 @@
+--- a/build/php.m4	2020-10-28 00:01:52.000000000 +0900
++++ b/build/php.m4	2020-11-13 08:57:27.939303216 +0900
+@@ -1907,6 +1907,10 @@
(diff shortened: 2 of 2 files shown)


neheb:
Upstream’s solution for older versions of PHP is https://github.com/php/php-src/commit/8eaaabdd58a395853d3658418622ee24635cb871

This commit is a backport from 8.0. I don’t think it’s a good idea.

PKG_RELEASE bump is missing.

Also, can you run

‘’’
make package/php7/refresh
‘’’

The other patches need to be cleaned up.

mhei:
I did not have a chance to look into this deeper, please be patient until next weekend.

> The other patches need to be cleaned up.

No, please don't do this. As long as the patches applies, I don't see a big gain in doing this.

neheb:
@ldir-EDB0 would disagree with you :).

I remember when I was working on fixing PHP compilation without deprecated OpenSSL APIs, one of the patches was seriously broken. I was wondering how it even applied. One of the sections was applied upstream but GNU patch was not smart enough to figure it out, so that section was applied to a different part of the code. I don't quite remember the details.

nxhack:
I checked with aarch64.
```
root@OpenWrt:~# php-cli --version
PHP 7.4.12 (cli) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies

root@OpenWrt:~# cat test.php 
<?php
            $slug ='こんにちは';
            $transliterator = Transliterator::create('Any-Latin;Latin-ASCII');
            $slug           = $transliterator->transliterate($slug);
	    echo $slug;
?>

root@OpenWrt:~# php-cli test.php 
kon'nichiha
```
I wait for the maintainer to make a decision.

mhei:
I had a look and I agree with @neheb : please rework this PR and use php/php-src@8eaaabd . Thanks.

nxhack:
The upstream patch didn't apply well, so I made a change.

Run test OK (aarch64)

PR #13903, opened 2020-11-11
net/iodine: add uci configurable interface options

Tabiskabis:
This is my very first pull request and I have no idea how this works.

Compile tested: nope
Run tested: (x86-64, OpenWRT 19.07.4, tests done)

Description:
I just wanted to set the interface name and MTU. This is what i came up with.
Also the followup patch that i'm unable to put into one single pull request.

### DIFF
--- a/net/iodine/files/iodined.init
+++ b/net/iodine/files/iodined.init
@@ -10,11 +10,28 @@ start_instance () {
 	config_get tunnelip "$section" 'tunnelip'
 	config_get tld      "$section" 'tld'
 	config_get port     "$section" 'port'
+	config_get ifname   "$section" 'ifname'
+	config_get mtu      "$section" 'mtu'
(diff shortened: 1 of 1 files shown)


neheb:
maintainer is @ukleinek

```
Author name (Tabiskabis) need to be your real name 'firstname lastname'
Commit subject line seems ok (net/iodine: add uci configurable interface options)
Signed-off-by is missing or doesn't match author (should be 'Signed-off-by: Tabiskabis <Tabiskabis@users.noreply.github.com>')
```

This needs to be signed off properly.

BKPepe:
This also requires to bump ``PKG_RELEASE`` in Makefile. Because Signed-off-by was not added, I'm closing this for now. If you do requested changes, feel free to re-open or create a new PR.

Issue #13896, opened 2020-11-10
mwan3 - Missing dependency


reinerotto:
On trunk, and on 19.07.x I get following error messages 
mwan3 status
Interface status:
interface wan is offline and tracking is paused
Command failed: Not found
Failed to parse message data
WARNING: Variable 'interfaces' does not exist or is not an array/object
WARNING: Variable 'wwan' does not exist or is not an array/object
interface wwan is online 00h:00m:00s, uptime 00h:00m:00s and tracking is active
Command failed: Not found
Failed to parse message data
WARNING: Variable 'interfaces' does not exist or is not an array/object
WARNING: Variable 'wifiwan' does not exist or is not an array/object
interface wifiwan is online 00h:00m:00s, uptime 00h:00m:00s and tracking is active

Current ipv4 policies:
standard:
wifiwan (100%)

The reason for this is a missing dependency to rpcd when building own/custom image, like I did.
_Explicitly_ including package rpcd (without options) gives

mwan3 status
Interface status:
interface wan is offline and tracking is down
interface wwan is online 03h:42m:10s, uptime 03h:42m:12s and tracking is active
interface wifiwan is online 03h:42m:43s, uptime 03h:42m:45s and tracking is active

Current ipv4 policies:
standard:
wifiwan (100%)

It looks like having LuCi in the image, it causes silently (also) to include package rpcd. 
However, custom built images do _not_ include LuCi by default, and will be hit byt this issue. Like me. 
For more details, refere to https://forum.openwrt.org/t/mwan3-2-10-1-cpu-hog/78443/15

aaronjg:
It looks like the dependency was introduced here: b0acbf057e and is just needed to get "online" and "uptime" statistics. Maybe these functions could be moved into common.sh and called directly, so we don't need to add the `rpcd` dependency.

If the rpcd package is not built, we can alter the Makefile so the mwan3 rpcd module is not installed.

feckert:
> It looks like the dependency was introduced here: [b0acbf0](https://github.com/openwrt/packages/commit/b0acbf057e05bbfb165fd83a181c0d1607da5558) and is just needed to get "online" and "uptime" statistics. Maybe these functions could be moved into common.sh and called directly, so we don't need to add the `rpcd` dependency.

Thats right this commit adds the dependency to rpcd. I thought that this rpcd is always include in the software.
And that there is no openwrt system where this rpcd is not included.

[...]

PR #13908, opened 2020-11-13
mwan3: remove dependency on rpcd & fix iputils-ping check

aaronjg:
Maintainer @feckert 

Allow `mwan3 interfaces` to get uptime via an internal function and
thus remove the dependency on rpcd for `mwan3 interface` calls.

Do not install the rpcd script if an image is built without rpcd. rpcd
is included in the standard images, so this will only impact custom images.

Fixes: #13896

### DIFF
--- a/net/mwan3/files/lib/mwan3/common.sh
+++ b/net/mwan3/files/lib/mwan3/common.sh
@@ -1,10 +1,5 @@
 #!/bin/sh
 
-get_uptime() {
-	local uptime=$(cat /proc/uptime)
-	echo "${uptime%%.*}"
-}
-
 IP4="ip -4"
 IP6="ip -6"
 SCRIPTNAME="$(basename "$0")"
@@ -179,4 +174,19 @@ mwan3_count_one_bits()
 		count=$((count+1))
 	done
 	echo $count
+}
+
+get_uptime() {
+	local uptime=$(cat /proc/uptime)
+	echo "${uptime%%.*}"
+}
+
+get_online_time() {
+	local time_n time_u iface
+	iface="$1"
+	time_u="$(cat "$MWAN3TRACK_STATUS_DIR/${iface}/ONLINE" 2>/dev/null)"
+	[ -z "${time_u}" ] || [ "${time_u}" = "0" ] || {
+		time_n="$(get_uptime)"
+		echo $((time_n-time_u))
+	}
 }
--- a/net/mwan3/files/usr/libexec/rpcd/mwan3
+++ b/net/mwan3/files/usr/libexec/rpcd/mwan3
@@ -69,6 +69,26 @@ report_policies_v6() {
 	done
 }
 
+get_age() {
+	local time_p time_u
+	iface="$1"
+	time_p="$(cat "$MWAN3TRACK_STATUS_DIR/${iface}/TIME")"
+	[ -z "${time_p}" ] || {
+		time_n="$(get_uptime)"
+		echo $((time_n-time_p))
+	}
+}
+
+get_offline_time() {
+	local time_n time_d iface
+	iface="$1"
+	time_d="$(cat "$MWAN3TRACK_STATUS_DIR/${iface}/OFFLINE")"
+	[ -z "${time_d}" ] || [ "${time_d}" = "0" ] || {
+		time_n="$(get_uptime)"
+		echo $((time_n-time_d))
+	}
+}
+
 get_mwan3_status() {
 	local iface="${1}"
 	local iface_select="${2}"
@@ -85,23 +105,9 @@ get_mwan3_status() {
 
 	track_status="$(mwan3_get_mwan3track_status "$1")"
 	[ "$track_status" = "active" ] && running="1"
-	time_p="$(cat "$MWAN3TRACK_STATUS_DIR/${iface}/TIME")"
-	[ -z "${time_p}" ] || {
-		time_n="$(get_uptime)"
-		let age=time_n-time_p
-	}
-
-	time_u="$(cat "$MWAN3TRACK_STATUS_DIR/${iface}/ONLINE")"
-	[ -z "${time_u}" ] || [ "${time_u}" = "0" ] || {
-		time_n="$(get_uptime)"
-		let online=time_n-time_u
-	}
-
-	time_d="$(cat "$MWAN3TRACK_STATUS_DIR/${iface}/OFFLINE")"
-	[ -z "${time_d}" ] || [ "${time_d}" = "0" ] || {
-		time_n="$(get_uptime)"
-		let offline=time_n-time_d
-	}
+	age=$(get_age "$iface")
+	online=$(get_online_time "$iface")
+	offline=$(get_offline_time "$iface")
 
 	local uptime="0"
 
--- a/net/mwan3/files/usr/sbin/mwan3track
+++ b/net/mwan3/files/usr/sbin/mwan3track
@@ -40,19 +40,35 @@ if_up() {
 	stop_subprocs
 }
 
+ping_test_host() {
+	if [ "$FAMILY" = "ipv6" ]; then
+		echo "::1"
+	else
+		echo "127.0.0.1"
+	fi
+}
+
+get_ping_command() {
+	if [ -x "/usr/bin/ping" ] && /usr/bin/ping -${FAMILY#ipv} -c1 -q $(ping_test_host) &>/dev/null; then
+		# -4 option added in iputils c3e68ac6 so need to check if we can use it
+		# or if we must use ping and ping6
+		echo "/usr/bin/ping -${FAMILY#ipv}"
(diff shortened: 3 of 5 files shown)


net/mwan3/files/lib/mwan3/common.sh
```diff
@@ -180,3 +175,18 @@ mwan3_count_one_bits()
...
+get_uptime() {
+	local uptime=$(cat /proc/uptime)
+	echo "${uptime%%.*}"
+}
+
+get_age() {
+	local time_p time_u
+	iface="$1"
+	time_p="$(cat "$MWAN3TRACK_STATUS_DIR/${iface}/TIME")"
+	[ -z "${time_p}" ] || {
```
feckert, line 188: Do we not have also to the same check as in https://github.com/openwrt/packages/pull/13908/files#diff-512330b586e317486d8128ca7fffafbd5ec1463d9e8de07bdcd2ebb8f770d445R86

feckert:
Everything else LGTM

net/mwan3/files/lib/mwan3/common.sh
aaronjg: OFFLINE/ONLINE gets set to zero when the interface is connected/disconnected, but TIME never gets set to zero, so the check does not appear to be needed.

I just took the existing code and moved them into functions, so there should be no change in behavior here.

ptpt52:
good

ptpt52:
it is better to create a standalone package `mwan3-rpcd` to install the rpcd scripts

net/mwan3/files/usr/libexec/rpcd/mwan3
```diff
@@ -69,6 +69,26 @@ report_policies_v6() {
...
+	iface="$1"
+	time_u="$(cat "$MWAN3TRACK_STATUS_DIR/${iface}/ONLINE")"
+	[ -z "${time_u}" ] || [ "${time_u}" = "0" ] || {
+		time_n="$(get_uptime)"
+		echo $((time_n-time_u))
+	}
+}
+
+get_offline_time() {
+	local time_n time_d iface
```
ptpt52, line 83: it it better to move these functions to `lib/mwan3/common.sh`

aaronjg: I disagree. These are not used by anything else, and we should keep common.sh small to reduce memory footprint of other scripts.

aaronjg:
@ptpt52 These comments are not helpful. If you believe an alternative layout would be better, please describe why so that we can have an informed discussion about it. Unless what you are suggesting was overlooked by the author (in this case it was not overlooked), then simply saying an alternative is better, does not add anything to the conversation.

aaronjg:
@feckert converting to draft for now until I fix syntax of PKG_CONFIG_DEPENDS and test build scripts.

ptpt52:
@aaronjg 
hi,

if someone install mwan3.ipk to system-with-rpcd
what should the mwan3.ipk contains? does it contain the mwan3-rpcd scripts?

and what about system-without-rpcd
what version mwan3 should be installed? the mwan3-with-rpcd-script or the mwan3-without-rpcd-script?

aaronjg:
As @feckert mentioned, rpcd is installed by default, so it will only be custom images that do not have this option. For the custom images without rpcd, mwan3 will be installed (or package will be built) without the rpcd module. If the user builds a custom image, and for some reason installs mwan3 from one of the official feeds, the extra file will do no harm. 

The alternative would be to have a second package which adds additional complexity in users maintaining a separate package and risks versions becoming mismatched.

After this P/R, I would not be opposed to moving the rpcd module to the mwan3-luci package, which, as far as I know, is now the sole consumer of the mwan3 rpcd service.

ptpt52:
> After this P/R, I would not be opposed to moving the rpcd module to the mwan3-luci package, which, as far as I know, is now the sole consumer of the mwan3 rpcd service.

ye! it is more reasonable to move the rpcd module to luci-app-mwan3

ptpt52:
got some issue when do this:
```
opkg remove mwan3 --force-depends
Removing package mwan3 from root...
sh: missing ]
sh: missing ]
sh: missing ]
Not deleting modified conffile /etc/config/mwan3.
```
I cannot locate the issue so I post here to see if there's a problem.

aaronjg:
> I cannot locate the issue so I post here to see if there's a problem.

Can not reproduce:

```
# opkg --force-depends remove mwan3
Removing package mwan3 from root...
Not deleting modified conffile /etc/config/mwan3.
# opkg install mwan3_2.10.3-2_all.ipk 
Installing mwan3 (2.10.3-2) to root...
Configuring mwan3.
Collected errors:
 * resolve_conffiles: Existing conffile /etc/config/mwan3 is different from the conffile in the new package. The new conffile will be placed at /etc/config/mwan3-opkg.
# opkg --force-depends remove mwan3
Removing package mwan3 from root...
Not deleting modified conffile /etc/config/mwan3.
# opkg install mwan3_2.10.3-2_all.ipk
Installing mwan3 (2.10.3-2) to root...
Configuring mwan3.
Collected errors:
 * resolve_conffiles: Existing conffile /etc/config/mwan3 is different from the conffile in the new package. The new conffile will be placed at /etc/config/mwan3-opkg.
```

Likely an issue with whatever version of mwan3 you had installed before.

aaronjg:
@feckert: I also noticed the posrm/postinst hooks can be cleaned up for install without rpcd. Will fix soon.

aaronjg:
Something is off with the iputils-ping version string in s20200821:
https://github.com/openwrt/packages/issues/13920

I added a workaround for this.

net/mwan3/files/usr/sbin/mwan3track
```diff
@@ -41,10 +41,15 @@ if_up() {
...
+	local ping_version_string
 	case "$1" in
 		ping)
-			if [ -x "/usr/bin/ping" ] && [ "$(/usr/bin/ping -V | grep -o '[0-9]*$')" -gt 20150519 ]; then
+			if [ -x "/usr/bin/ping" ]; then
+				ping_version_string=$(/usr/bin/ping -V)
+			fi
+			if [ -n "$ping_version_string" ] && { [ -z "${ping_version_string##*iputils v*}" ] || [ "$(echo ${ping_version_string} | sed 's/.*iputils s\([0-9]\+\).*/\1/')" -gt 20150519 ] ; }; then
 				# -4 option added in iputils c3e68ac6
+				# change in build between s20200821 and s20190709 causes the binary in OpenWRT package to report a version tag rather than date tag
```
ptpt52, line 52: why do this ping version check?
we may check it by using it
do ping test before we use it
```
ping -c1 -q -6 ::1 || somewhat
ping -c1 -q -4 127.0.0.1 || somewhat
```
could we use this method the determinate `what ping` ? before the loop.

aaronjg: Yes, that would be an alternative solution. I don't see a strong reason to use one way or the other. This was the smaller change that kept with the existing version check logic and works around the iputils package issue. Why do you prefer the test ping method over the version check?

ptpt52: emm
The version has countless possibilities including future possibilities, so that the check needs to be updated again
The ping test is more direct and will not cause problems due to subsequent version updates

aaronjg: Sure. iputils pings should stably increment its version number, but as the latest iputils-ping.ipk shows, that isn't always translated into the OpenWRT. 

I just added it. It is a bit faster than the `sed` based version I had, but a bit slower than a purely `sh` based implementation of processing the version string. It only happens once on startup and is very fast regardless of which way it is done, so performance isn't a big issue here.

feckert:
I do not fully agree with [this](https://github.com/openwrt/packages/pull/13908/commits/66fd0d39100352b30b0be059c1bb4c79e5b98b97) amendment.

I would not complicate matters too much and take a more pragmatic approach. We simply install the rpcd file, in the directory provided for this purpose. If the rpcd is installed then the rpcd will mount it and if not then not.
Since we do not use rpcd internally from mwan3 we do not have to add dependency. If someone installs the rpcd afterwards rpcd of mwan3 is then available.
I have seen other packages do the same https://github.com/openwrt/packages/search?q=rpcd to install rpcd but do not have a dependency because they do not use it internal.

aaronjg:
>  If someone installs the rpcd afterwards rpcd of mwan3 is then available.

Just to be clear, it would be a bit more narrower situation that could cause the problem. This flag is checked at build time rather than install time. Someone would have to build the mwan3 from an image builder with rpcd turned off, and then turn on rpcd and install that. 

I was hoping to save users without rpcd 5 kb by excluding the file, but it's a pretty minor concern, so we can just always install it like you say. I'd rather have it always installed than create a separate .ipk for the one file.

feckert:
@aaronjg thanks :+1:

PR #13913, opened 2020-11-13
libudev-zero: fix installation of so files.

sch-m:
Maintainer: @neheb 
Compile tested: x86_64, latest master
Run tested: x86_64, latest master

Description:
This fixes the installation of the file `/usr/lib/libudev.so`.

### DIFF
--- a/libs/libudev-zero/Makefile
+++ b/libs/libudev-zero/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libudev-zero
 PKG_VERSION:=0.4.5
-PKG_RELEASE:=1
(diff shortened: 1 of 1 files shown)


neheb:
There’s no need. You’re installing an extra symlink for no benefit.

sch-m:
> 
> 
> There’s no need. You’re installing an extra symlink for no benefit.

`/usr/lib/libudev.so` is not a symlink, it's the "real" file.

neheb:
I have now verified this.

Please bump the package version.

sch-m:
> I have now verified this.
> 
> Please bump the package version.

Done.

sch-m:
Oh, there was a version update in the meantime. I will update my patch again.

PR #13919, opened 2020-11-15
gitlab-runner: add new package

ja-pa:
Maintainer: me

Description:
This PR adds new package gitlab-runner. It is useful in case you want to connect GitLab CI/CD to the router.

More info https://docs.gitlab.com/runner/

### DIFF
--- a/devel/gitlab-runner/Makefile
+++ b/devel/gitlab-runner/Makefile
@@ -1,0 +1,47 @@
+#
+# Copyright (C) 2020 CZ.NIC, z. s. p. o. (https://www.nic.cz/)
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gitlab-runner
+PKG_VERSION:=13.5.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://gitlab.com/gitlab-org/gitlab-runner/-/archive/v$(PKG_VERSION)
+PKG_HASH:=765c1556ed9dd4c1b36f9946224c62f068b171e2c1581eeb8f71055327d8a274
+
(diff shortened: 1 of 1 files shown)


ja-pa:
Test mips_24kc fails on following error:
```
../../../../../dl/go-mod-cache/github.com/docker/engine@v1.4.2-0.20190822180741-9552f2b2fdde/pkg/system/stat_linux.go:11:3: cannot use s.Rdev (type uint32) as type uint64 in field value
```
It is probably related to https://github.com/containers/storage/issues/444

BKPepe:
In any case, this should be reviewed by @jefferyto

jefferyto:
Looks like there are [variables](https://gitlab.com/gitlab-org/gitlab-runner/-/blob/v13.5.0/Makefile#L43-45) that can be set with `GO_PKG_LDFLAGS_X`, but LGTM otherwise.

> Test mips_24kc fails on following error:

This was fixed in https://github.com/docker/engine/commit/5f0231bca193320e1a3d785a3ade0e64241fe580, not sure why gitlab-runner is using such an old version of that package.

ja-pa:
> Looks like there are [variables](https://gitlab.com/gitlab-org/gitlab-runner/-/blob/v13.5.0/Makefile#L43-45) that can be set with `GO_PKG_LDFLAGS_X`, but LGTM otherwise.
> 
> > Test mips_24kc fails on following error:
> 
> This was fixed in [docker/engine@5f0231b](https://github.com/docker/engine/commit/5f0231bca193320e1a3d785a3ade0e64241fe580), not sure why gitlab-runner is using such an old version of that package.

I have forbidden build for MIPS and mipsel architecture. (Upstream issue https://gitlab.com/gitlab-org/gitlab-runner/-/issues/27234) 
When it comes to variables, I left it unfilled (or do you think it's blocker? @jefferyto )

jefferyto:
> When it comes to variables, I left it unfilled (or do you think it's blocker? @jefferyto )

I haven't looked into where they are used, but usually variables like these are only for the version string. On the one hand, the version string isn't critical (though it might confuse users); on the other hand, it's probably not hard to set some/all of the variables. I think it's nice to have but not a strict blocker.

PR #13923, opened 2020-11-16
knot: fix compilation with musl 1.2.0

neheb:
Missing header for strlcpy.

Moved linux defines down as they conflict with netinet/in.h , which
must be include first.

Refreshed patches.

Maintainer: @salzmdan 
Compile tested: powerpc

This does not compile with full NLS. I tried adding nls.mk to fix, but then I get this error:

```
powerpc-openwrt-linux-musl-gcc: error: none: No such file or directory
powerpc-openwrt-linux-musl-gcc: error: required: No such file or **directory**
```

No idea how to fix.

### DIFF
--- a/net/knot/Makefile
+++ b/net/knot/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=knot
 PKG_VERSION:=3.0.1
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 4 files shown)


BKPepe:
Did you send your patches to Knot DNS project?

salzmdan:
We are analyzing the patches but we don't understand why xdp.c is compiled? It should be disabled on owrt.

neheb:
I have no idea.

salzmdan:
Okay, we will prepare another patch...

neheb:
Superseded

PR #13930, opened 2020-11-16
mariadb: Update to 10.4.17

miska:
Update to the latest version from 10.4 series

Signed-off-by: Michal Hrusecky <michal.hrusecky@turris.com>

Maintainer: me
Compile tested: mvebu, Turris Omnia, Turris OS 5.2 ~ OpenWrt 19.07
Run tested: mvebu, Turris Omnia, Turris OS 5.2 ~ OpenWrt 19.07, runs

### DIFF
--- a/utils/mariadb/Makefile
+++ b/utils/mariadb/Makefile
@@ -8,8 +8,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mariadb
-PKG_VERSION:=10.4.14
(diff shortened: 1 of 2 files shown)


neheb:
Instead of manually refreshing the patches, you could just run make package/mariadb/refresh

In other news, the SDK is broken again...

miska:
Thanks for the tip, will try it next time, but with quilt it was pretty straight forward.

PR #13931, opened 2020-11-16
mwan3: add mwan3 reload if firewall gets restarted

feckert:
Signed-off-by: Florian Eckert <fe@dev.tdt.de>

Maintainer: me / @aaronjg 

Description:

If firewall is restarted, then the mwan3 rules are purged and not
reinstalled. To fix this add a hotplug script to restart mwan3 on
firewall restart.

Depends on procd change:
This is still pending https://patchwork.ozlabs.org/project/openwrt/list/?series=218099

### DIFF
--- a/net/mwan3/files/etc/init.d/mwan3
+++ b/net/mwan3/files/etc/init.d/mwan3
@@ -108,6 +108,30 @@ reload_service() {
 	start
 }
 
+extra_command "firewall" "Reload all mwan3 firewall rules"
+firewall() {
+	handle_reload() {
+		local iface="$1"
+		local device
+
+		[ "$(mwan3_get_iface_hotplug_state "$iface")" = "online" ] && {
+			network_get_device device $iface
+			mwan3_create_iface_iptables $iface $device
+		}
+	}
+
+	mwan3_init
+	mwan3_set_general_iptables
+
+	config_load mwan3
+	config_foreach handle_reload interface
+
+	mwan3_set_policies_iptables
+
+	LOG notice "Re-apply firewall rules"
+}
+
 service_triggers() {
+	procd_add_restart_service_trigger "firewall" /etc/init.d/mwan3 firewall
 	procd_add_reload_trigger 'mwan3'
 }


aaronjg:
Is it possible to do this in the procd script? I know that procd has triggers for this sort of thing, but I don't know if we can do it on just restart and not trigger on reload, when it would not be needed.

feckert:
Unfortunately I don't know exactly what you mean. But when you restart the firewall the whole firewall will be deleted and we have to re-read the rules for the mwan3 because they are no longer there. When the firewall is reloaded nothing happens, the changes are not touched that mwan3 does, because this only refers to the fw3 and it reads the uci again and adapts the rules.

aaronjg:
I was wondering if we can do it with a [service_trigger](https://openwrt.org/docs/guide-developer/procd-init-scripts#specifying_triggers) in the procd script:
> Most system important changes result in generating events that service_triggers() can use for triggering various actions. 

However, I'm not sure if it is possible to set up a trigger just on the firewall reload. It looks like none of the [helpers](https://github.com/openwrt/openwrt/blob/master/package/system/procd/files/procd.sh#L560-L566) do exactly what we want, but I think it should still be possible through the `add_raw_trigger`, but I can't find much documentation around that.

I don't think we need to do a complete restart of mwan3 here. mwan3track and mwan3rtmon processes can keep running. I believe we just need to do the following pseudocode:

```
mwan3_set_general_iptables
for each interface in mwan3_interfaces
   if interface is connected:
      network_get_device device INTERFACE
      mwan3_create_iface_iptables $INTERFACE $DEVICE
mwan3_set_policies_iptables

**Comment by feckert** [comment]
> However, I'm not sure if it is possible to set up a trigger just on the firewall reload. It looks like none of the [helpers](https://github.com/openwrt/openwrt/blob/master/package/system/procd/files/procd.sh#L560-L566) do exactly what we want, but I think it should still be possible through the `add_raw_trigger`, but I can't find much documentation around that.

So I don't see any support from the procd either.
We have to hook into the /etc/hotplug.d/firwall framwork.
Just like I did it. 

> ```
> mwan3_set_general_iptables
> for each interface in mwan3_interfaces
>    if interface is connected:
>       network_get_device device INTERFACE
>       mwan3_create_iface_iptables $INTERFACE $DEVICE
> mwan3_set_policies_iptables
> ```

But I agree with you that my change to restart the mwan3 'only' because the firewall is restarted is pretty rough.
I would also prefer if we could solve this in a more charming way. 
But that was the fastest solution I could think of.
Your suggestion sounds not bad.

We can create a custom command in /etc/init.d/mwan3 that does just that, adding the firewall rules again. We could than call this custom initd target from /etc/hotplug.d/firewall

**Comment by ptpt52** [comment]
I never seem the hotplug firewall usage
we may use something like miniupnpd did
https://github.com/openwrt/packages/blob/master/net/miniupnpd/files/miniupnpd.defaults

**Comment by aaronjg** [comment]
> we may use something like miniupnpd did

That script is installed here:
/etc/uci-defaults/99-miniupnpd

The [documentation](https://openwrt.org/docs/guide-developer/uci-defaults) states that files in that directory are run the first time the device boots. So it doesn't seem like the correct place to put it. I'm not sure how minupnpd ensures that the reload rule is installed when it is installed from an ipk. 

> I never seem the hotplug firewall usage

vpnbypass, banip, and vpn-policy-routing all use the firewall firewall hotplug scripts.

gnunet, yggdrasil, apfree-wifidog, minupnpd, bcp38, family-dns-update, and adblock modify the firewall uci config.

So it seems it can go either way. Since we have the option of using the firewall hotplug, my preference would be to not change the uci config of firewall. I'm a bit surprised that procd doesn't have the functionality to add a trigger for this, since it is such a common need, but I went through the procd source and looked at the ubus messages, and it seems it is impossible to trigger this just with procd.

> We can create a custom command in /etc/init.d/mwan3 that does just that, adding the firewall rules again. We could than call this custom initd target from /etc/hotplug.d/firewall

I think that would be a great way to handle it. Let's keep the meat of the script in /etc/init.d/mwan3 and just have a minimal firewall hotplug script.

**Comment by feckert** [comment]
> I never seem the hotplug firewall usage
> we may use something like miniupnpd did
> https://github.com/openwrt/packages/blob/master/net/miniupnpd/files/miniupnpd.defaults

@ptpt52 I have also thought about it and tried it out. But as described by @aaronjg there are some problems. Also it is not a configuration. The approach with the hotplug.d is the right way from my point of view. This is more generic. Every system that wants to be informed about a reload of the firewall or other firewall action puts its script there and is called and can execute to perform the task.

File: net/mwan3/files/etc/hotplug.d/firewall/mwan3-restart [hunk 1]
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+if [ "$ACTION" = "restart" ]; then
```
ptpt52, line 3: have you test this?
I try to /etc/init.d/firewall restart
but not ACTION=restart called
just somewhat `add` `remove` ACTION

feckert: Good hint. I completely forgot that this is a patch set of mine that is not upstream.t was self-evident to me that this would work! 
https://github.com/lede-project/source/pull/1311
For some reason, which I can not restore in my mind, I closed the PR and no attention was paid to it anymore.
Maybe I should try this again.

It was probably due to the lede and openwrt remerge!

ptpt52:
>>I'm not sure how minupnpd ensures that the reload rule is installed when it is installed from an ipk.

`opkg install xxx` would invoke the uci-defaults scripts, so that would works even install from ipk

aaronjg:
> opkg install xxx would invoke the uci-defaults scripts,

Thanks. If that's true, it sounds like the documentation for uci-defaults needs to be updated.

net/mwan3/files/etc/hotplug.d/firewall/mwan3-restart
feckert: @stangri As I see this your code does not work as well https://github.com/openwrt/packages/blob/master/net/vpnbypass/files/vpnbypass.hotplug? Or do I miss something.

aaronjg: Sounds like the hotplug solution used in vpnbypass and vpn-policy-routing would not be working either.

If this is going to require changing other packages, it would be great to add triggers for this sort of thing in procd. Otherwise we can modify the firewall config using uci-defaults like @ptp52 suggests.

feckert: I would prefer the hotplug.d, as it does not require any configuration changes. But that means that we have to get the change into fw3 first. As I said before, I tried that but I didn't finish it. 
Also, nobody has tried my patches, which are still applicable by the way.
In the source of the fw3 nothing has changed in this area so far.

@ptpt52 and @aaronjg Can you also test this [P/R](https://github.com/lede-project/source/pull/1311/) if it con work like this on your own builds?

aaronjg: Yeah, I too wouldd prefer not to change the firewall config. But if we are going to change an upstream package, why not change procd so that one init script can have triggers on restart of other init scripts? This seems like the more general solution.

feckert: procd is also an upstream package.
But that would also be a possibility for us to adapt the procd or /etc/rc.common
We would have to extend the procd so that it calls the scripts in /etc/hotplug.d/procd/* with the parameters SERVICE and ACTION when start/stop/restart/reload from a service is executed.

aaronjg: `procd` already has facilities to handle triggers on config changes and on interface changes, so rather than creating a new hotplug directory, we could hook into these existing facilities.

We could add a `procd` trigger that would be fired on procd events, and then in `/etc/init.d/mwan3` `service_triggers()` add something like:

`procd_add_procd_trigger firewall restart|start reload_firewall_rules`

This would register the trigger with procd, and then when an event happens on the associated procd module, it would fire the trigger for mwan3.

stangri: > @stangri As I see this your code does not work as well https://github.com/openwrt/packages/blob/master/net/vpnbypass/files/vpnbypass.hotplug? Or do I miss something.

Thanks for pinging me, I just tested this myself and I don't see the `/etc/hotplug.d/firewall/99-vpn-policy-routing` called at all, with any parameters on neither fw reload/restart. I may have used incorrect syntax for the `procd_add_procd_trigger` as well, because I don't see VPR being reloaded after firewall restart either. I haven't tested this part for a while, not sure when it broke.

aaronjg: > I may have used incorrect syntax for the procd_add_procd_trigger as well, 

Just to be clear, in my previous comment, I was suggesting an extension to procd to handle this. No such feature exists yet as far as I know

stangri: > > I may have used incorrect syntax for the procd_add_procd_trigger as well,
> 
> Just to be clear, in my previous comment, I was suggesting an extension to procd to handle this. No such feature exists yet as far as I know

I was under impression that things like `procd_add_procd_trigger firewall` in `service_triggers()` are supposed to work in current releases/snapshots. Are they not?

aaronjg: I don’t see any such code for that kind of trigger:

https://github.com/openwrt/openwrt/blob/master/package/system/procd/files/procd.sh#L556

feckert:
@aaronjg Please have a look at the two patches as proof of conecpt. 
This patch adds a ubus service call and generates a event a service gets started stopped or reloaded.
The event is generated for all services that use the prcod
[0001-base-files-extend-rc.common-to-generate-start-stop-a.patch.txt](https://github.com/openwrt/packages/files/5572719/0001-base-files-extend-rc.common-to-generate-start-stop-a.patch.txt)

This patch adds an lisener to mwan3 so that if the firwall gets reloaded we call /etc/init.d/mwan3 firewall. There we only call a logger to show that this event is delivered.
[0001-mwan3-add-procd-lisener-to-firewall-reload-events.patch.txt](https://github.com/openwrt/packages/files/5572720/0001-mwan3-add-procd-lisener-to-firewall-reload-events.patch.txt)

As I said, this is only a proof of concept. And it was relatively easy to script it with shell. The question is now, is this acceptable or do we have to implement this in the procd in native c. I have oriented myself on the following package [br2684ctl](https://github.com/openwrt/openwrt/commit/1d1265b40bbb15e8bb64d0424f2f1b9d2bfce8f2).

net/mwan3/files/etc/hotplug.d/firewall/mwan3-restart
stangri: Sorry, confused `procd_add_procd_trigger` and `procd_add_reload_trigger`, of which only the latter works. 

If the PROCD fix is merged, how likely that it will be backported to current releases?

aaronjg: According to the documentation, the reload_trigger is only fired on a config change, so that is not what we want here.

aaronjg:
@feckert 

On the mwan3 side, the event should trigger on firewall start, since the start trigger is hit on a restart, and we don't need to reload the firewall rules on a firewall reload.

I'm not sure what the best way to handle this is on the services side. On the one hand `ubus` is primarily used for interprocess communication and since mwan3 and firewall are procd scripts, with the appropriate APIs, this could be handled entirely within procd. On the other hand, if we are looking for a general solution, other scripts may want to trigger on non-procd restarts, so having it in accessible outside of procd might be better.

What you proposed is certainly a simple solution, so if we can get this into openwrt-core, I'm fine with it.

ptpt52:
You guys made things complicated,
It is easy and simple to use a firewall script to do the trick.

```
config include 'mwan3'
	option type 'script'
	option path '/usr/share/mwan3/mwan3-reload.sh'
	option reload '0'
```

aaronjg:
> You guys made things complicated,
> It is easy and simple to use a firewall script to do the trick.

Yes, we are aware that is an option. Please read the above discussion again to understand the pros and cons of that approach.

aaronjg:
So it looks like in general, there are ubus messages for what we want to do:

```
# ubus monitor|grep 'method":"service.' &
# etc/init.d/firewall stop 2>/dev/null
<- 18b5721b #e24b1713         notify: {"objid":-498395373,"method":"service.stop","data":{"service":"firewall"},"no_reply":true}
# /etc/init.d/firewall start 2>/dev/null
<- 18b5721b #e24b1713         notify: {"objid":-498395373,"method":"service.start","data":{"service":"firewall"},"no_reply":true}
```

The problem is that `/etc/init.d/firewall restart` bypasses procd and calls `fw3 restart` directly. This seems incorrect to me, since in doing so, it bypasses the procd lock functions. 

Shorewall has a [restart_service](https://github.com/openwrt/packages/blob/ffdbc7a637df8e3f39dd6916405f78840edccfb4/net/shorewall/files/shorewall.init#L22-L26) method, which appears to be the right way to do things. However, I don't understand how that would be called, because /etc/rc.common does not appear to ever call the restart_service.

feckert:
I agree with you that this is all a bit difficult to see through!

> So it looks like in general, there are ubus messages for what we want to do:

That would of course be the best.
I also see a start and stop event but I cannot trigger a reload or a restart.
How this is connected I do not understand myself.

If we want to have a real solution that should have existed longer we must dig through this confusion!
We must be careful. There are also packages that do not use procd!

aaronjg:
I think it would be useful enough to add some functionality so that packages can trigger on other package's restart, even if it is just procd packages for now. Most of the main system components have been migrated to procd already.

Here is a patch to generate the restart trigger events. We could also do something similar for reload events. Can mwan3 trigger the firewall reload off of this?
[procd_restart.txt](https://github.com/openwrt/packages/files/5586619/procd_restart.txt)

This is what it I see on ubus when I do a firewall stop, start, restart:
```
# ubus monitor|grep notify|grep firewall

<- bca783a3 #59eb21cd         notify: {"objid":1508581837,"method":"service.stop","data":{"service":"firewall"},"no_reply":true}
<- bca783a3 #59eb21cd         notify: {"objid":1508581837,"method":"service.start","data":{"service":"firewall"},"no_reply":true}
<- bca783a3 #59eb21cd         notify: {"objid":1508581837,"method":"service.restart","data":{"service":"firewall"},"no_reply":true}

```

feckert:
@aaronjg thanks for your patch!
I have staged a [P/R](https://github.com/TDT-AG/openwrt/tree/pr/20201124-procd-restart) for this in my repository, but did not public them to openwrt.
I will do some more testing and add also the related changes to mwan3.

aaronjg:
Thanks for cleaning that up. If this all works, we should add a similar function to procd for reload and then upstream to procd rather than patching in openwrt, but for testing it is convenient to have the patch in procd.

If you have updated code for mwan3 and push it here, I can also help test

feckert:
@aaronjg So now I've played around with it and tried to get into it.
Here are my remarks.
I now also get a restart notify on my system :tada: .
> ubus monitor | grep service

`notify: {"objid":-1100491970,"method":"service.restart","data":"service":"firewall"},"no_reply":true}`

But now I am failing to use this notify to trigger the reload for mwan3.

I have add to the procd api the function `_procd_add_service_trigger`
```

_procd_add_service_trigger() {
        json_add_array
        _procd_add_array_data "$1"
        shift

        json_add_array
        _procd_add_array_data "if"

        json_add_array
        _procd_add_array_data "eq" "service" "$1"
        shift
        json_close_array

        json_add_array
        _procd_add_array_data "run_script" "$@"
        json_close_array

        json_close_array
        _procd_add_timeout
        json_close_array
}
```
So we could call the function in the init script and this function will install the notification we are interested in.
In our use case I have update the mwan3 init script with the following changes.
```
extra_command "firewall" "Reload all mwan3 firewall rules into iptables"
firewall() {
        logger "mwan3: Firewall reload execute"
}

service_triggers() {
        procd_open_trigger
        procd_add_service_trigger "service.restart" "firewall" /etc/init.d/mwan3 firewall
        procd_add_config_trigger "config.change" "mwan3" /etc/init.d/mwan3 reload
        procd_close_trigger
}
```
But this does not work.
If I call this command `ubus call service event '{"type":"service.restart", "data":{"service":"firewall"}}` on the system in a shell it **works** as expected. That would be my original [implementation in this comment](https://github.com/openwrt/packages/pull/13931#issuecomment-731071995).
So that I only make the ubus call when the service is started, stopped and restart.
The change in the procd would then not have been necessary.

I don't think I really understand the whole thing yet how this all works together.
I think I need to rethink how it all fits together

nulh:
e21dd267bb7cbb37a0120e9defa278dc25c31556

feckert:
@aaronjg The problem here is that we have no service that listens to this notification.
We would have to install a notify handler in every service who is interested in this notification.
But for this purpose the system service daemon must have a ubus interface.
Which in most cases is not the case if they are not developed for openwrt.

With the change we could now extend a service that performs an action on a service restart
 For this purpose, the service must use the ubus notify interface.

The way I see it, these are two different things notify from ubus and event over ubus from procd.
I think we have to use the event API of of the procd service.

net/mwan3/files/etc/hotplug.d/firewall/mwan3-restart
feckert: @aaronjg @stangri  I have added to your function the following code.
```
        blob_buf_init(&b, 0);
        blobmsg_add_string(&b, "service", s->name);
        trigger_event("service.restart", b.head);
```
I have update the staged [P/R](https://github.com/TDT-AG/openwrt/tree/pr/20201124-procd-restart) with my changes in my forked openwrt repository.
The following changes must also be done in mwan3 so the callback is execute on an firewall restart.

```
extra_command "test" "This is a test callback"
test() {
  logger "mwan3: Firewall reload execute test"
}

service_triggers() {
  procd_open_trigger
  procd_add_service_trigger "service.restart" "firewall" /etc/init.d/mwan3 test
  procd_add_config_trigger "config.change" "mwan3" /etc/init.d/mwan3 reload
  procd_close_trigger
}
```

And now I get an event which triggers my mwan3 callback on a firewall restart.
Everything is wrapped in the new restart method for the service ubus object.
I also get the notify.
Whatever I need it for.
I am not a ubus specialist so maybe someone could explain me for what this notify is for.

GeorgeSapkin:
@aaronjg @feckert still relevant?

feckert:
I have that on my patch set. I will update it when I get the chance.

PR #13932, opened 2020-11-16
adblock: modify StevenBlack blocklist sources

cyberflamingo:
Maintainer: @dibdot
Compile tested: -
Run tested: TurrisOS 5.1.3

Description:
* replace fakenews-gambling-porn-social with fakenews-gambling-porn
* modify title accordingly

Comment:
  The current title for this blocklist is stevenblack_porn although
  it also contains other extensions. Particularly, it contains the social
  extension which block legitimate websites the like ofTwitter, Reddit etc.
  I replaced the list with the fakenew-gambling-porn list which I believe is
  more adequat.

Signed-off-by: Alexandre Mercier <alex@cyberflamingo.net>

### DIFF
--- a/net/adblock/files/adblock.sources
+++ b/net/adblock/files/adblock.sources
@@ -292,8 +292,8 @@
 		"focus": "compilation",
 		"descurl": "https://github.com/StevenBlack/hosts"
 	},
-	"stevenblack_porn": {
-		"url": "https://raw.githubusercontent.com/StevenBlack/hosts/master/alternates/fakenews-gambling-porn-social/hosts",
+	"stevenblack_fakenews-gambling-porn": {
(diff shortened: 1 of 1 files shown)


dibdot:
Thanks, but I like to keep the current list - many users asked for so called "social media" filtering. We already provide two StevenBlack variants, which is enough IMHO.

cyberflamingo:
Thank you for the fast feedback!

I understand. In that case would it be possible to modify only the title? The current one is confusing; using LuCI there is no way (as far as I know) to check the URL and I believe people would not expect `stevenblack_porn` to have every extensions, including social.

What do you think?

dibdot:
Well it's a list compilation with many list flavours. I doubt that we can handle that with name "optimizations" (the names will be quite long and break other things like the limited select box). Today you can look in the readme and follow the list sources to get more information about a list.

cyberflamingo:
Got it! Thanks for taking the time to respond.

PR #13935, opened 2020-11-17
travelmate: remove wrong f_net call in source

feckert:
Maintainer: @dibdot 

Description:
The call to f_net on this point in the source makes no sense.

### DIFF
--- a/net/travelmate/files/travelmate.sh
+++ b/net/travelmate/files/travelmate.sh
@@ -610,7 +610,6 @@ f_check()
 				ifname="$(printf "%s" "${dev_status}" | jsonfilter -l1 -e '@.*.interfaces[@.config.mode="sta"].ifname')"
 				if [ -n "${ifname}" ] && [ "${enabled}" = "1" ]
 				then
-					result="$(f_net)"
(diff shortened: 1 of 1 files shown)


dibdot:
NAK for this change. The result will be later used if user has disabled the captive portal detection (l. 649, 657) or in error cases (l. 670).

feckert:
My problem here is that I don't want to connect to a captive portal, but a curl is started anyway.
This makes no sense.
I understand now that you need this.
But wouldn't it be better if we then moved it to line this code section.
https://github.com/openwrt/packages/blob/316f87022847e3cde1d0ef34cf4e9322a70e287a/net/travelmate/files/travelmate.sh#L619-L690

where it is really needed.

dibdot:
Nope, that doesn't handle the error case. With travelmate 2.x curl is a package dependency and not only used for cp handling - and I do not plan to change that ... ;-)

dibdot:
BTW, you've created a branch with your PR in this repo. Most probably not intended, isn't it?

feckert:
@dibdot Oops, I must have missed that. I will delete it immediately
THANKS

PR #13946, opened 2020-11-19
runc: enable seccomp support by default

dangowrt:
It's nice to have seccomp support which is enabled in OpenWrt on
supported platforms on targets which are not marked as SMALL_FLASH.
(and it's kinda obvious that you wouldn't want to install runc on a
SMALL_FLASH target to begin with)
So let's enable seccomp by default.

Maintainer: @G-M0N3Y-2503

### DIFF
--- a/utils/runc/Makefile
+++ b/utils/runc/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=runc
 PKG_VERSION:=1.0.0-rc10
-PKG_RELEASE:=2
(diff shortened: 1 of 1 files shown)


G-M0N3Y-2503:
Sorry I missed this, From what I can tell [This PR](https://github.com/openwrt/packages/pull/13874/files#diff-5917365273ff1a968ced689ba8b5e6dd0ba9bfbf2083e0e98a167c87db63736a) will also have the same effect just refactored slightly.
If I've misunderstood this patch, feedback on that PR would be appreciated.

PR #13968, opened 2020-11-20
Provide both libdaq 2.2.2 and 3.0.0-beta1 to satisfy Snort 2 and 3.

flyn-org:
Maintainer: me
Compile tested: mips_24kc master

Description:
Earlier bump of libdaq left Snort 2 unable to compile. This provides both libdaq and libdaq3 packages so that both Snort 3 and 3 compile.

### DIFF
--- a/libs/libdaq3/Makefile
+++ b/libs/libdaq3/Makefile
@@ -1,0 +1,67 @@
+#
+# Copyright (C) 2012-2015 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libdaq3
+PKG_VERSION:=3.0.0-beta1
+PKG_RELEASE:=1
+
+PKG_SOURCE_URL:=https://www.snort.org/downloads/snortplus/
+PKG_SOURCE:=daq-$(PKG_VERSION).tar.gz
+PKG_HASH:=ef74aa1c30a6ee93eacbe7967d1c85d7df3214cf3783d4eabbb6b64305fd273e
+PKG_BUILD_DIR:=$(BUILD_DIR)/daq-$(PKG_VERSION)
+
+PKG_LICENSE:=GPL-2.0-only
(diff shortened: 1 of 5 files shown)


libs/libdaq3/Makefile
```diff
@@ -0,0 +1,67 @@
...
+PKG_NAME:=libdaq3
+PKG_VERSION:=3.0.0-beta1
+PKG_RELEASE:=1
+
+PKG_SOURCE_URL:=https://www.snort.org/downloads/snortplus/
+PKG_SOURCE:=daq-$(PKG_VERSION).tar.gz
+PKG_HASH:=ef74aa1c30a6ee93eacbe7967d1c85d7df3214cf3783d4eabbb6b64305fd273e
+PKG_BUILD_DIR:=$(BUILD_DIR)/daq-$(PKG_VERSION)
+
+PKG_LICENSE:=GPL-2.0
```
neheb, line 19: should be -only or -or-later .

libs/libdaq3/Makefile
```diff
@@ -0,0 +1,67 @@
...
+PKG_MAINTAINER:=W. Michael Petullo <mike@flyn.org>
+
+PKG_INSTALL:=1
```
neheb, line 22: Should add PKG_BUILD_PARALLEL:=1

libs/libdaq3/Makefile
```diff
@@ -0,0 +1,67 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/nls.mk
```
neheb, line 25: Is NLS actually used here?

libs/libdaq3/Makefile
flyn-org: Fixed.

libs/libdaq3/Makefile
flyn-org: Fixed.

libs/libdaq3/Makefile
flyn-org: Fixed.

PR #14005, opened 2020-11-24
avahi: incorrectly requires IFF_MULTICAST on point-to-point (tunnel) …

pprindeville:
Maintainer: @thess 
Compile tested: x86_64, generic, 359a4b46b
Run tested: ditto, built and installed on production routers

Description:

The notion of "multicast" is meaningless on a point-to-point (P2P) connection because there's only ever going to be a single recipient at the other end.

This issue has been submitted upstream as: https://github.com/lathiat/avahi/issues/305 and a fix as PR: https://github.com/lathiat/avahi/pull/306.

### DIFF
--- a/libs/avahi/Makefile
+++ b/libs/avahi/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=avahi
 PKG_VERSION:=0.8
-PKG_RELEASE:=6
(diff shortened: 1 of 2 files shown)


skleeschulte:
Maybe I'm wrong, but at least someone might have a look at my comment here: https://github.com/lathiat/avahi/pull/306#issuecomment-986150582

PR #14019, opened 2020-11-25
lsof: update to 4.94.0

neheb:
Upstream switched to GitHub and changed their tarball. Adjusted.

Refreshed patches.

Maintainer: @mstorchak 
Compile tested: ath79


mstorchak:
@neheb , why did you decide to drop the reproducible build patch?

nulh:
2ecbc0442388a71340c5e8e43617459aae45a176

nulh (changes requested):
2ecbc0442388a71340c5e8e43617459aae45a176

mstorchak:
Most of the packages use codeload.github.com. maybe lsof should follow?
`https://codeload.github.com/lsof-org/lsof/tar.gz/$(PKG_VERSION)?`

neheb:
The date patch no longer applies. A quick look indicates upstream fixed this in their own way.

I don't use codeload as they provide a tarball.

mstorchak:
LGTM
We'll see if any extra work is required to restore reproducible builds here https://tests.reproducible-builds.org/openwrt/openwrt_ath79.html

PR #14028, opened 2020-11-25
strongswan: add x.509 certificate based authentication

pprindeville:
Maintainer: me, @Thermi
Compile tested: x86_64, generic, 359a4b4
Run tested: x86_64, generic, installed on production router

Description:

The main change is adding support for X.509 certificates, which many prefer over PSK's.  A secondary change is adding support for the `setup` section.

I'm trying to separate cleanup (which anyone can comment on) from new functionality.  If there are no further review comments, I can squash the commits or leave them separate.


net/strongswan/files/ipsec.init
```diff
@@ -166,9 +166,9 @@ config_conn() {
 	config_get reqid                    "$1"           reqid
 	config_get packet_marker            "$1"           packet_marker
 
-	[ -n "$local_nat" ] && local_subnet=$local_nat
+	[ -n "$local_nat" ] && local_subnet="$local_nat"
 
-	ipsec_xappend "conn $config_name-$1"
+	ipsec_xappend "conn $1"
```
pprindeville, line 171: I don't see the point to prefixing the section with 'cfgNNNN-' so let's just keep the same name.

net/strongswan/files/ipsec.init
```diff
@@ -229,11 +229,11 @@ config_transport() {
 config_remote() {
 	local enabled
 	local gateway
+	local local_gateway
+	local remote_gateway
 	local pre_shared_key
 	local auth_method
 
-	config_name=$1
```
pprindeville, line 235: Unused below (and above).

net/strongswan/files/ipsec.init
```diff
@@ -245,7 +245,7 @@ config_remote() {
 
 	[ "$gateway" = "any" ] && remote_gateway="%any" || remote_gateway="$gateway"
 
-	[ -z "$local_identifier" ] && {
```
pprindeville, line 248: Unclear what this has to do with the local_identifier.  Seems more related to the local_gateway.

stintel:
I believe this is all about the uci support, which I don't know and don't use.
@dedeckeh can you or the authors review?

net/strongswan/files/ipsec.init
```diff
@@ -187,29 +201,31 @@ config_conn() {
...
 	[ -n "$inactivity" ] && ipsec_xappend "  inactivity=$inactivity"
 	[ -n "$reqid" ] && ipsec_xappend "  reqid=$reqid"
+	[ $reauth -eq 0 ] && ipsec_xappend "  reauth=no"
+	[ $mobike -eq 0 ] && ipsec_xappend "  mobike=no"
 
 	if [ "$auth_method" = "psk" ]; then
 		ipsec_xappend "  leftauth=psk"
 		ipsec_xappend "  rightauth=psk"
 
-		[ "$remote_sourceip" != "" ] && ipsec_xappend "  rightsourceip=$remote_sourceip"
```
pprindeville, line 195: Use a consistent notation for "not empty string".

net/strongswan/files/ipsec.init
```diff
@@ -94,30 +94,26 @@ add_crypto_proposal() {
 	config_get hash_algorithm        "$1" hash_algorithm
 	config_get dh_group              "$1" dh_group
 
-	[ -n "${encryption_algorithm}" ] && \
```
pprindeville, line 97: Strip unnecessary braces here and  below.

net/strongswan/files/ipsec.init
```diff
@@ -94,30 +94,26 @@ add_crypto_proposal() {
...
+	[ -n "$encryption_algorithm" ] && \
 		crypto="${crypto:+${crypto},}${encryption_algorithm}${hash_algorithm:+-${hash_algorithm}}${dh_group:+-${dh_group}}"
 }
 
 set_crypto_proposal() {
 	local conf="$1"
 	local proposal
+	local crypto=""
```
pprindeville, line 104: Make locally scoped here and elsewhere.

net/strongswan/files/ipsec.init
```diff
@@ -94,30 +94,26 @@ add_crypto_proposal() {
...
 
-	crypto=""
+	config_list_foreach "$conf" crypto_proposal add_crypto_proposal
```
pprindeville, line 106: It's a `list`, not an `option`.

net/strongswan/files/ipsec.init
```diff
@@ -94,30 +94,26 @@ add_crypto_proposal() {
...
-	for proposal in $crypto_proposal; do
-		add_crypto_proposal "$proposal"
-	done
-
-	[ -n "${crypto}" ] && {
+	[ -n "$crypto" ] && {
 		local force_crypto_proposal
 
-		config_get_bool force_crypto_proposal "$conf" force_crypto_proposal
+		config_get_bool force_crypto_proposal "$conf" force_crypto_proposal 0
```
pprindeville, line 111: Default is not to force.

net/strongswan/files/ipsec.init
```diff
@@ -187,29 +201,31 @@ config_conn() {
...
+		[ -n "$remote_subnet" ] && ipsec_xappend "  rightsubnet=$remote_subnet"
 	else
 		warning "AuthenticationMethod $auth_method not supported"
 	fi
 
-	[ -n "$local_identifier" ] && ipsec_xappend "  leftid=$local_identifier"
-	[ -n "$remote_identifier" ] && ipsec_xappend "  rightid=$remote_identifier"
+	ipsec_xappend "  auto=$mode"
+
+	[ -n "$local_identifier" ] && ipsec_xappend "  leftid=\"$local_identifier\""
```
pprindeville, line 219: DN's frequently have spaces...

net/strongswan/files/ipsec.init
```diff
@@ -245,17 +274,17 @@ config_remote() {
...
 		[ "$remote_gateway" = "%any" ] && ipdest="1.1.1.1" || ipdest="$remote_gateway"
-		local_gateway=`ip route get $ipdest | awk -F"src" '/src/{gsub(/ /,"");print $2}'`
+		local_gateway=`ip -o route get $ipdest | awk '/ src / { gsub(/^.* src /,""); gsub(/ .*$/, ""); print $0}'`
 	}
-
-	[ -n "$local_identifier" ] && secret_xappend -n "$local_identifier " || secret_xappend -n "$local_gateway "
-	[ -n "$remote_identifier" ] && secret_xappend -n "$remote_identifier " || secret_xappend -n "$remote_gateway "
-
-	secret_xappend ": PSK \"$pre_shared_key\""
+	if [ "$auth_method" = "psk" ]; then
```
pprindeville, line 283: DN-based keys don't require a match, the cert proves its own binding to the private key associated with it.

net/strongswan/files/ipsec.init
```diff
@@ -187,29 +205,40 @@ config_conn() {
...
+	[ -n "$remote_sourceip" ] && ipsec_xappend "  rightsourceip=$remote_sourceip"
+	[ -n "$remote_subnet" ] && ipsec_xappend "  rightsubnet=$remote_subnet"
 
 	if [ "$auth_method" = "psk" ]; then
 		ipsec_xappend "  leftauth=psk"
 		ipsec_xappend "  rightauth=psk"
 
-		[ "$remote_sourceip" != "" ] && ipsec_xappend "  rightsourceip=$remote_sourceip"
-		[ "$remote_subnet" != "" ] && ipsec_xappend "  rightsubnet=$remote_subnet"
+	elif [ "$auth_method" = "pubkey" ]; then
```
pprindeville, line 218: X.509 requires a private key and certificate that was generated with that key.

net/strongswan/files/ipsec.init
```diff
@@ -267,24 +296,59 @@ config_remote() {
...
 
+do_preamble() {
 	ipsec_xappend "# generated by /etc/init.d/ipsec"
 	ipsec_xappend "version 2"
 	ipsec_xappend ""
 
 	secret_xappend "# generated by /etc/init.d/ipsec"
+}
+
+config_setup() {
```
pprindeville, line 313: Add support for a `config setup` section.

net/strongswan/files/ipsec.init
```diff
@@ -267,24 +296,59 @@ config_remote() {
...
+	local cachecrls
+	local charondebug
+	local strictcrlpolicy
+	local uniqueids
+
+	[ -n "$seen_setup" ] && return
```
pprindeville, line 319: Doesn't make sense to have multiple such sections but there's no built-in notion of singular sections in UCI.

net/strongswan/files/ipsec.init
```diff
@@ -267,24 +296,59 @@ config_remote() {
...
+
+	config_get_bool cachecrls setup cachecrls 0
+	config_get charondebug setup charondebug ""
+	config_get strictcrlpolicy setup strictcrlpolicy "no"
+	config_get uniqueids setup uniqueids "yes"
+
+	ipsec_xappend "config setup"
+	[ $cachecrls -eq 1 ] && ipsec_xappend "  cachecrls=yes"
+	[ -n "$charondebug" ] && ipsec_xappend "  charondebug \"$charondebug\""
```
pprindeville, line 328: Use defaults if nothing is explicitly set.

net/strongswan/files/ipsec.init
```diff
@@ -178,6 +186,16 @@ config_conn() {
...
 	[ -n "$remote_firewall" ] && ipsec_xappend "  rightfirewall=$remote_firewall"
 
+	case "$fragmentation" in
+	yes|accept|force|no)
+		ipsec_xappend "  fragmentation=$fragmentation" ;;
+	esac
+
+	case "$closeaction" in
+	none|clear|hold|restart)
+		ipsec_xappend "  closeaction=$closeaction" ;;
```
karlp, line 196: is there any method or benefit in having a warning printed for the *) cases for these two? Are these hand edited uci config files, or written via tooling/LuCI? Just a thought, might not be useful.

net/strongswan/files/ipsec.init
```diff
@@ -334,9 +398,14 @@ config_ipsec() {
 }
 
 prepare_env() {
+	local seen_setup=
```
karlp, line 401: Don't think you need the =.  In fact, if you're unlucky, or didn't notice, local is actually a function, and has a return code, so these sorts of things are quite different.  (not relevant here, but just something to ponder with local...

```
local output=$(something that might fail)
[ $? -eq 0 ] && {
    printf "hoho, this will always run, because local will 'never' fail\n"
}
```
vs
```
local output
output=$(something that might fail)
[ $? -eq 0 ] && {
    printf "this will do what you expected...\n"
}
```

net/strongswan/files/ipsec.init
pprindeville: > is there any method or benefit in having a warning printed for the *) cases for these two? Are these hand edited uci config files, or written via tooling/LuCI? Just a thought, might not be useful.

I hand-edit my own configs... I don't know how common that is.  Was thinking that adding sanity checking and error reporting might be useful at some point, but it doesn't seem to be very common in the UCI scripts so maybe that's a dead-end.

net/strongswan/files/ipsec.init
pprindeville: > Don't think you need the =. In fact, if you're unlucky, or didn't notice, local is actually a function, and has a return code, so these sorts of things are quite different. (not relevant here, but just something to ponder with local...

I figured this was the same as:

```
local seen_setup=""
```

I suppose I could also set it to `0` and then test for `[ $seen_setup -eq 1 ] && return` ... any preferences?

pprindeville:
I think this is ready to merge if no-one has an issue with it at this point.

net/strongswan/files/ipsec.init
pprindeville: Looked some more and indeed there's a function called `warning()` that burps a message to `stderr`, but it's not used very frequently.

net/strongswan/files/ipsec.init
pprindeville: Ended up explicitly setting it to `""`.

pprindeville:
@karlp Have your issues been addressed?  Anyone else care to review this?

pprindeville:
@feckert Can you take this PR for a spin and tell us if you have any problems with it?

feckert:
@pprindeville Everything you have done looks plausible (but I have not done a run test).

The question is only is this necessary, if we want to switch to swanctl anyway?
Isn't it better to leave it as it is and use the new swanctl config interface in a new init script?

As I mentioned before.
Can we note write a new startup script `/etc/init.d/strongswan`  and parse the uci config from `/etc/config/strongswan`?
Then we could start on a greenfield site and have no legacy to worry about!
We can add a note to the old starup script `/etc/init.d/ipsec` to manually switch to the new configuration, as we do not offer an upgrade.

pprindeville:
Needs to be rebased to PR #14535 after merge.

CodeFetch:
```
Can we note write a new startup script /etc/init.d/strongswan and parse the uci config from /etc/config/strongswan?
Then we could start on a greenfield site and have no legacy to worry about!
```

Please...

pprindeville:
> @pprindeville Everything you have done looks plausible (but I have not done a run test).
> 
> The question is only is this necessary, if we want to switch to swanctl anyway?
> Isn't it better to leave it as it is and use the new swanctl config interface in a new init script?
> 
> As I mentioned before.
> Can we note write a new startup script `/etc/init.d/strongswan` and parse the uci config from `/etc/config/strongswan`?
> Then we could start on a greenfield site and have no legacy to worry about!
> We can add a note to the old starup script `/etc/init.d/ipsec` to manually switch to the new configuration, as we do not offer an upgrade.

Done.  See PR #14708 instead.

net/strongswan/files/ipsec.init
```diff
@@ -245,17 +270,17 @@ config_remote() {
 
 	[ "$gateway" = "any" ] && remote_gateway="%any" || remote_gateway="$gateway"
 
-	[ -z "$local_identifier" ] && {
+	[ -z "$local_gateway" ] && {
 		local ipdest
 
 		[ "$remote_gateway" = "%any" ] && ipdest="1.1.1.1" || ipdest="$remote_gateway"
-		local_gateway=`ip route get $ipdest | awk -F"src" '/src/{gsub(/ /,"");print $2}'`
+		local_gateway=`ip -o route get $ipdest | awk '/ src / { gsub(/^.* src /,""); gsub(/ .*$/, ""); print $0}'`
```
luizluca, line 277: How about `awk '/ src / { print gensub(/.* src ([^ ]+).*/,"\\1","1") }'`?

pprindeville: Just saw this comment, but ended up doing something very similar in PR #14758.

net/strongswan/files/ipsec.init
```diff
@@ -62,13 +62,15 @@ add_crypto_proposal() {
...
+	local prf_algorithm
 
 	config_get encryption_algorithm  "$1" encryption_algorithm
 	config_get hash_algorithm        "$1" hash_algorithm
 	config_get dh_group              "$1" dh_group
+	config_get prf_algorithm         "$1" prf_algorithm
 
 	[ -n "$encryption_algorithm" ] && \
-		crypto="${crypto:+${crypto},}${encryption_algorithm}${hash_algorithm:+-${hash_algorithm}}${dh_group:+-${dh_group}}"
+		crypto="${crypto:+${crypto},}${encryption_algorithm}${hash_algorithm:+-${hash_algorithm}}${dh_group:+-${dh_group}}${prf_algorithm:+-${prf_algorithm}}"
```
Thermi, line 73: prf needs to go after hash-algorithm

Also, make sure the hash algorithm is not defined if an AEAD Algorithm is used, but then you have to define the prf manually. By default it's the same algorithm as the hmac (but used as prf instead). But if no hmac is defined, then the prf has to be specified manually.

net/strongswan/files/ipsec.init
```diff
@@ -238,11 +248,11 @@ config_remote() {
...
 		local_gateway=$(ip -o route get "$ipdest" | awk '/ src / { print gensub(/^.* src ([^ ]*) .*$/, "\\1", "g"); }')
 	}
-
-	secret_xappend -n "$local_gateway "
-	[ -n "$remote_identifier" ] && secret_xappend -n "$remote_identifier " || secret_xappend -n "$remote_gateway "
-
-	secret_xappend ": PSK \"$pre_shared_key\""
+	if [ "$auth_method" = "psk" ]; then
+		secret_xappend -n "$local_gateway "
+		[ -n "$remote_identifier" ] && secret_xappend -n "$remote_identifier " || secret_xappend -n "$remote_gateway "
```
Thermi, line 253: That doesn't look right. Aren't there the [] missing around the last two conditions?
Also, a local and remote identifier aren't strictly required for PSKs. They are just necessary for finding the right secret if multiple can be candidates.

A good solution would be the following:
Remote and local identifiers/IPs are optional, but if more than one PSK is detected without an ID or IP, then issue a warning

pprindeville: > That doesn't look right. Aren't there the [] missing around the last two conditions?

No, `[ test ] && then-stmt || else-stmt` is a shorthand notation.

> Also, a local and remote identifier aren't strictly required for PSKs. They are just necessary for finding the right secret if multiple can be candidates.

Knowing if there can be multiple matches might require sophisticated logic to detect that in this script...

net/strongswan/files/ipsec.init
```diff
@@ -58,7 +60,24 @@ warning() {
...
+fatal() {
+	echo "ERROR: $@" >&2
+
+	CONFIG_FAIL=1
+}
+
+is_aead() {
+	local cipher="$1"
+
+	case "$cipher" in
```
Thermi, line 72: You could just generalize that to *gcm*|*ccm*|*gmac*, because those are all AEAD cipher modes. Meaning, if potentially, some algorithm xez was standardized, then xez*cbc, xez*gcm*, xez*ccm*, xez*gmac* could occur and the current expression would not catch these parts of a proposal.

net/strongswan/files/ipsec.init
Thermi: OK.

net/strongswan/files/ipsec.init
pprindeville: Understood, but this is the check we already have in `swanctl.init`

pprindeville: Will handle `chacha20poly1305` as a follow-up PR.

pprindeville:
Closing this as it's not worth sinking effort into `ipsec`-related support.

PR #14013, opened 2020-11-25
curl: update to 7.73.0

neheb:
Remove upstream patch.

Refresh other one.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @kaloz is listed
Compile tested: ath79

PR #14030, opened 2020-11-26
curl: fix build against OpenSSL

dengqf6:
Maintainer: @kaloz @aparcar 
Compile tested: ramips, mvebu
Run tested: ramips, mvebu

Description:
Ref: https://curl.se/mail/lib-2020-10/0037.html


hnyman:
I suddenly stumbled onto this error today.

I suspect #14013 as the reason

stintel:
> I suspect #14013 as the reason

Correct, reverting that solved it for me.

stintel:
@dengqf6 thanks!

PR #14031, opened 2020-11-26
net/miniupnpd: ext_ip_reserved_ignore support

ptpt52:
This add an option to ignore ext_ip reserved
Enable this option in config by default.

Fixes: #13773 
Patche is upstream pending https://github.com/miniupnp/miniupnp/pull/511

### DIFF
--- a/net/miniupnpd/Makefile
+++ b/net/miniupnpd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=miniupnpd
 PKG_VERSION:=2.2.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_URL:=https://miniupnp.tuxfamily.org/files
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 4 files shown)


stangri:
@neheb @ptpt52 can we roll this one back? it's creating many issues: https://github.com/openwrt/packages/issues/15258 https://github.com/miniupnp/miniupnp/pull/511.

Thanks!

neheb:
this creates a new option: ext_ip_reserved_ignore , which is enabled by default.

Why not just disable?

stangri:
> this creates a new option: ext_ip_reserved_ignore , which is enabled by default.
> 
> Why not just disable?

Oh, right, thanks for the work-around, I did ask @ptpt52 [here](https://github.com/openwrt/packages/issues/15258#issuecomment-828662674) on why is it needed and why is it turned on by default about a month ago and yet to hear back.

As far as I understand, this setting being enabled by default is the root of the problem, should I maybe send PR to just disable it by default?

The message in the log is grammatically incorrect and isn't informative, I'd like to fix it (hence asking for feedback from @ptpt52 ), but I do not understand the background.

neheb:
sure. this patch is for a specific use case anyway.

Neustradamus:
There is a new patch from @ptpt52:
- https://github.com/openwrt/packages/pull/16011#issuecomment-920538698

It replaces the old 301-ext_ip_reserved_ignore.patch.

Tested by @openips!

PR #14045, opened 2020-11-28
frr: update to 7.5

neheb:
Remove upstream patches.

Refresh the other ones.

Use upstream tarball instead of codeload.

Maintainer: @lucize 
Compile tested: ath79


neheb:
Note that the current version of frr has some problems with newer OSes: https://downloads.openwrt.org/snapshots/faillogs/aarch64_generic/packages/frr/host-compile.txt

This version works though.

lucize:
I'm reluctant about this version because I can't start multiple ospf instances and need to bisect where the problem started
also I would like to use latest stable 7.5 git code based on date because of how the updates are released on stable versions (none)

neheb:
Do you know which commit fixes the issue I linked to?

lucize:
odd, python was a very old problem, I guess you could look in the history of configure.ac for something meaningful in the commit message

neheb:
I can't see anything useful.

In any case, would be great if you could look. frr currently does not build on the buildbots.

lucize:
I would like to use the latest 7.5 commit and revert the commits that fails to start ospf multiple instances, should I propose a new PR ? I addressed the commit upstream

neheb:
Sure.

PR #14046, opened 2020-11-28
libusb-compat: fix libusb-config

neheb:
Turns out, some packages like serdisplib actually use libusb-config.
Fix the paths in it so that they can work properly. Method taken from
libpng.

Maintainer: @nbd168 
Compile tested: ath79


karlp:
meh, packages still using libusb-compat after all these years should be seriously reviewed for relevance :|

PR #14053, opened 2020-11-28
netdata: fix log spam caused by procfile

tiagogaspar8:
Maintainer: Josef Schlehofer <pepe.schlehofer@gmail.com>, Daniel Engberg <daniel.engberg.lists@pyret.net>
Compile tested: WRT3200ACM on latest trunk
Run tested: WRT3200ACM on latest trunk

Description:

Fix log spam:
daemon.err netdata[2090]: PROCFILE: Cannot open file '/proc/sysvipc/shm'
That's caused by a non existant /proc/sysvipc/shm

### DIFF
--- a/admin/netdata/Makefile
+++ b/admin/netdata/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=netdata
 PKG_VERSION:=1.29.3
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_MAINTAINER:=Josef Schlehofer <pepe.schlehofer@gmail.com>, Daniel Engberg <daniel.engberg.lists@pyret.net>
 PKG_LICENSE:=GPL-3.0-or-later
--- a/admin/netdata/files/netdata.conf
+++ b/admin/netdata/files/netdata.conf
@@ -27,3 +27,6 @@
 
 [health]
 	enabled = no
+
+[plugin:proc:ipc]
+	shared memory totals = no


hnyman:
cc maintainers  @BKPepe  @diizzyy

diizzyy:
LGTM (not tested)

neheb:
Needs a PKG_RELEASE bump.

tiagogaspar8:
You're absolutely right,
it's now fixed.

BKPepe:
Before merging this, could you tell us which platform (I mean router) are you using?

On my router there is such file::
```
root@omnia:~# cat /proc/sysvipc/shm
       key      shmid perms       size  cpid  lpid nattch   uid   gid  cuid  cgid      atime      dtime      ctime        rss       swap
```

I am not saying this is wrong, but thinking if ipc should not be disabled as whole plugin. Because netdata config is dynamic and users should change their config depending on their usage.

tiagogaspar8:
You're absolutely right, let's clear this out.
I'm using a WRT3200ACM router with a image I built myself.
How did you enable the `/proc/sysvipc/shm` file? I was searching arround in the build options and found nothing.

Strykar:
`shm` does not exist on a PC Engines APU2 x86/64 running Openwrt too:
```
# uname -a
Linux apu 4.14.195 #0 SMP Sun Sep 6 16:19:39 2020 x86_64 GNU/Linux

# cat /proc/sysvipc/shm
cat: can't open '/proc/sysvipc/shm': No such file or directory
```
Could this be because there are no shm segments in use when Netdata checks?
I couldn't check because [shmctl](https://www.systutorials.com/docs/linux/man/2-shmctl/) isn't available. I know for a fact, gpsd uses shm and it works on Openwrt.

neheb:
Needs a rebase.

tiagogaspar8:
Rebase done!

neheb:
Where's the PKG_RELEASE bump?

tiagogaspar8:
Sorry for the delay. Done.

neheb:
@BKPepe opinion?

BKPepe:
Well, as I said earlier, netdata config is very dynamic. It is complicated to have a configuration file for every router as each of it can be different.However, I see that many people have this harmless spam, so it can be merged, but before merging, these things need to be done:

1. If it would be up to me, I will remove ellipsis from the commit subject because they should not be there and instead of it, it should say what is going to be disabled.
Procfile? Which one? There could be any proc file or proc module.
See: https://github.com/netdata/netdata/tree/master/collectors/proc.plugin#procplugin

 It should be clear and exact something like this:
- ``netdata: disable shared memory totals by default``
- ``netdata: disable plugin:proc:ipc for shared memory totals``
- ``netdata: disable ipc in proc.plugin``
You are doing this for everyone who wants to have better Git history and more detailed. 

2. Which version of OpenWrt are both of you using? Based on the kernel 4.14, I'd say OpenWrt 19.07.

3. Commit message should be improved.  It says what it is going to fix, which is good, but we don't know why there isn't such a file (in some configurations) and what the ipc is about.  Most likely, this is caused that I don't have set ``CONFIG_PROC_STRIPPED``.

I'd say this is a culprit:

- OpenWrt 19.07: https://github.com/openwrt/openwrt/blob/openwrt-19.07/target/linux/generic/hack-4.14/902-debloat_proc.patch#L162

- OpenWrt master: https://github.com/openwrt/openwrt/blob/master/target/linux/generic/hack-5.4/902-debloat_proc.patch#L165

tiagogaspar8:
Hi everyone!

Sorry for the long delay.
I totally agree with what you say, it is now fixed I believe, if you can take a look.
Any suggestion is welcome!

BKPepe (approved):

BKPepe:
Added to OpenWrt 21.02: https://github.com/openwrt/packages/commit/f5d172aedeb2b24cb58212cddbac0b51e46a893b and OpenWrt 19.07:  https://github.com/openwrt/packages/commit/807e088bbc4a993c28a3f0ad9df3a439fc57c081

Thank you for your contribution!

PR #14060, opened 2020-11-29
uwsgi: recreate uwsgi-python3-plugin after upgrading python

peter-stadler:
Maintainer: @Ansuel 
Run tested: x86_64, x86_64 qemu, master, run etesync-server

Description: Create new version of uwsgi-python3-plugin that is linked to libpython3.9 instead of 3.8.


Ansuel:
Any idea how to prevent this in the future? Wonder if the python module suffer of the same problem.

peter-stadler:
Hmm, not really; the only solution I have in mind, is to version it (uwsgi-python38-plugin and uwsgi-python39-plugin), but I am not sure if this is a good thing. I will look if I find another solution ...

Ansuel:
A hint should be how the python module handle the version.

neheb:
Interesting buildbot failure...

I don't experience it locally but @pprindeville does.

ping @jefferyto

neheb:
Please rebase to remove those red checkmarks.

peter-stadler:
Rebased, thank you for the fix :-)

I am not sure if we need this change: in the package system `uwsgi` is already rebuild using python3.9. It was just not updated on my system because it had the same version number, but I successfully reinstalled the other python3-packages (maybe their version numbers were increased unrelated to the direct upgrade of python3) ...

jefferyto:
I think this issue still exists for most/all Python packages, although it would only affect master and not release branches, since release branches will stay on the same 3.x version of Python.

Python-dependent packages may be rebuilt by buildbots, but opkg on users' systems will not know to reinstall them after a Python upgrade like 3.8 to 3.9. In this case it is a change from libpython3.8 to libpython3.9, but even for "plain" Python packages, their install path changes from `/usr/lib/python3.8/site-packages` to `/usr/lib/python3.9/site-packages`.

I think this may be a good case for adding `ABI_VERSION` to the python3 package (not sure how to have other "plain" Python packages hook into this value though). @commodo any thoughts?

commodo:
ABI_VERSION in Python3 sounds like a good idea
it can be parametrized from `PYTHON3_VERSION_MAJOR` & `PYTHON3_VERSION_MINOR`

neheb:
Sounds like this can be closed?

commodo:
if ABI_VERSION will fix this, then sure;
let me open a PR on that; i forgot about this

commodo:
created PR;
no idea how to proceed from here;
merge the other PR and check back here?

i can see that i would forget about this PR [again] :p

pprindeville:
> I don't experience it locally but @pprindeville does.

@neheb I do?

neheb:
AFAIK it was some issue that is now fixed. I don't remember the details.

peter-stadler:
I am closing this PR as the other one will solve it in the right and broader way (it is not merged, so technically the problem is still open, but I think it is enough to have one PR for it) ...

jefferyto:
Now that #14237 is merged, perhaps it would help to have uwsgi-python3-plugin depend on libpython3?

PR #14061, opened 2020-11-29
python-pynacl: include only minimal version

peter-stadler:
Maintainer: @jmarcet 
Run tested: x86_64, x86_64 qemu, master, run etebase

Description: Make it compatible with libsodium that is compiled per default with --enable-minimal.


jefferyto:
I believe (but haven't confirmed) that pynacl should already detect if it is working with a minimal build of libsodium. Is it not working?

neheb:
Please rebase to fix the red checkmarks.

peter-stadler:
Rebased and rechecked: The error occurs only if I use the package that I compile locally not the pre-built one.
Then for `python3 -c 'import nacl.signing'` I get the error:
```
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.9/site-packages/nacl/signing.py", line 17, in <module>
  File "/usr/lib/python3.9/site-packages/nacl/bindings/__init__.py", line 17, in <module>
  File "/usr/lib/python3.9/site-packages/nacl/bindings/crypto_aead.py", line 18, in <module>
ImportError: Error relocating /usr/lib/python3.9/site-packages/nacl/_sodium.abi3.so: crypto_core_ed25519_scalar_invert: symbol not found
```
Maybe it is because pynacl is using libsodium of the build host system. 

I don´t know if we want to change this at all and how: Should we apply my proposed patch depending on the symbol `CONFIG_LIBSODIUM_MINIMAL` (Is it good to use the config variable of the other package `libsodium` here)?

jefferyto:
> Maybe it is because pynacl is using libsodium of the build host system.

It could be - would need to check the pynacl build log to be sure.

I wonder... did this issue appear recently? Could this be related to https://github.com/openwrt/openwrt/commit/e52d0487e88c3c8c57e1310d1a02b18eae0d142e?

> Should we apply my proposed patch depending on the symbol `CONFIG_LIBSODIUM_MINIMAL`

Setting `CONFIG_LIBSODIUM_MINIMAL` should build a minimal version of libsodium, which should have `SODIUM_LIBRARY_MINIMAL` defined in it's header file. So I am more curious why it isn't working as it should.

peter-stadler:
> I wonder... did this issue appear recently? 

I don´t know, I did not install it before, but:

* the `openwrt/staging_dir/target-x86_64_musl/usr/include/sodium/version.h` would contain `#define SODIUM_LIBRARY_MINIMAL 1` ...

> Could this be related to [openwrt/openwrt@e52d048](https://github.com/openwrt/openwrt/commit/e52d0487e88c3c8c57e1310d1a02b18eae0d142e)?

* Adding the following two lines to pynacl`s Makefile does not help ...
```
TARGET_CPPFLAGS:=-I$(STAGING_DIR)/usr/include
TARGET_LDFLAGS:=-L$(STAGING_DIR)/usr/lib -L$(STAGING_DIR)/lib
```

How is it used?

peter-stadler:
Sorry for the inconvenience, it is working (-:

PR #14064, opened 2020-11-29
docker-ce: Added firewall.extra_iptables_args

G-M0N3Y-2503:
Maintainer: @G-M0N3Y-2503

Description:
Added dockerd.firewall.extra_iptables_args

This is a convenience argument to primarily facilitate outbound wan
connections from a docker container. However, all docker containers still can't
bidirectionally communicate with the internet by default.

### DIFF
--- a/utils/docker-ce/files/dockerd.init
+++ b/utils/docker-ce/files/dockerd.init
@@ -185,10 +185,12 @@ iptables_add_blocking_rule() {
 	local cfg="${1}"
 
 	local device=""
+	local extra_iptables_args=""
 
 	handle_iptables_rule() {
 		local interface="${1}"
 		local outbound="${2}"
+		local extra_iptables_args="${3}"
 
 		local inbound=""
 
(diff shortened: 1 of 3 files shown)


feckert:
I know I'm annoying.
But can you please split your commit?
Your shell check fixes have nothing to do with the commit you describe.

utils/docker-ce/files/dockerd.init
```diff
@@ -200,9 +202,11 @@ iptables_add_blocking_rule() {
 			return
 		}
 
-		if ! iptables --table filter --check DOCKER-USER --in-interface "${inbound}" --out-interface "${outbound}" --jump DROP 2>/dev/null; then
+		# Ignore errors as it might already be present
+		iptables --table filter --new DOCKER-USER 2>/dev/null
+		if ! iptables --table filter --check DOCKER-USER --in-interface "${inbound}" --out-interface "${outbound}" ${extra_iptables_args} --jump DROP 2>/dev/null; then
```
feckert, line 207: I'm not sure if that works.
Let's assume someone changes **extra_iptables_args** then the old one remains because the check failed and the new one is added additional to the old one.

G-M0N3Y-2503: That's right, I added [thease two lines to the config](https://github.com/openwrt/packages/pull/14064/files#diff-58eb4aa24d76fbb36d87fd7b82d84026397c40c01914dec58c1c777cc48e8868R19-R20) to help explain this behaviour.
I don't think the effort or complexity for tighter integration with the firewall via this method would be worth it. 
I assume people could set `iptables` to false and try setting their rules via the UCI or at that point, I'd think they'd be savvy enough to do their own thing with iptables for anything more complex.

PR #14072, opened 2020-11-29
hplip: use libusb-compat

neheb:
Some change to base broke compilation. From looking at the CFLAGS,
hplip tries to use the host libusb.

libusb-compat seems to work properly. So use that.

Maintainer: @luizluca 
Compile tested: powerpc

### DIFF
--- a/utils/hplip/Makefile
+++ b/utils/hplip/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=hplip
 PKG_VERSION:=3.20.9
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


neheb:
Uhh I accidentally merged this. I'll revert when I can.

luizluca:
You were too quick. I would say that libusb-compat was not the way to go. I saw you reverted and already fixed the libcups include with https://github.com/openwrt/packages/pull/14077

hplip uses some "shortcuts" that are not recommended. I reported the libusb hardcoded include path upstream but they are not the fastest project to track bug reports (and they do not have an open development process).

https://bugs.launchpad.net/hplip/+bug/1906257

Anyway, thanks, you saved me some time to fix this.

PR #14087, opened 2020-11-30
libwebsocket: add new version 3.2.0

zokl:
Signed-off-by: Zbyněk Kocur <zbynek.kocur@fel.cvut.cz>

Maintainer: @karlp 
Compile tested: (x86_64, x86_64, OpenWrt Master)
Run tested: (x86_64, x86_64, OpenWrt Master, OK)


libs/libwebsockets/Makefile
```diff
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libwebsockets
-PKG_VERSION:=3.1.0
+PKG_VERSION:=3.2.0
 PKG_RELEASE:=2
```
neheb, line 12: Decrement to 1

karlp:
no?  we already had 3.2 and it was reverted, at some stage, we'll move to 4.1.x, certainly not another round of 3.2.x.  see d5bace89cf or work on testing and verifying all downstream apps with lws 4.1.x

PR #14090, opened 2020-12-01
zerotier: source package hash changed

mwarning:
The contents remained identical.


mwarning:
It would be interesting to know if the latest openwrt builds of this package fail. If not, then this needs further investigation.

mwarning:
@neheb

neheb:
There's a new version. Just update to that. Also get rid of that valloc hack.

PR #14097, opened 2020-12-01
openvpn: import from base

neheb:
Maintainer: none?
Compile tested: ath79


aparcar:
I think @nbd168 is currently the maintainer.

neheb:
ping @mkrkn

mkrkn:
Are you asking for my input on the PR in general, or about maintainership? I think it is appropriate to move OpenVPN to feeds/packages at this point, I see no reason it needs to remain in base. 

I can take maintainership of the openvpn package itself. 

I'm more wary of openvpn-easy-rsa. First of all I don't use it, and if I did it wouldn't be on an OpenWrt device. I wouldn't want a CA to live on an always-on networked device, and for a CA host I see no reason in particular for it to run OpenWrt. Secondly embedded devices tend to perform such tasks slowly. IMO packaging openvpn-easy-rsa in OpenWrt encourages bad habits.

jow-:
Please keep `openvpn-easy-rsa` - it is useful for people where the OpenWrt system is the only Linux system at hand.

luizluca:
I can take care of easyrsa. Do I need to create a PR or this one can be updated?

PKG_MAINTAINER:=Luiz Angelo Daros de Luca <luizluca@gmail.com>

neheb:
@jow- not sure what you mean. Keep it in base and move openvpn here?

neheb:
@mkrkn I put your commits in this PR, hence the ping.

It would be great if you would take maintainership of OpenVPN. I see that many of the version bumps come from you.

@luizluca I'll update.

aparcar:
I think jow just wants to keep the packet in general and not drop it

neheb:
It's in this PR.

aparcar:
Let's do this.

luizluca:
The easy-rsa package was committed without maintainers. Should I PR?

neheb:
Sure.

PR #14098, opened 2020-12-01
lighttpd: build and bugfix patches

gstrauss:
Maintainer: W. Michael Petullo mike@flyn.org
Compile tested: x86-64 OpenWrt master branch (v19.07.4)

Description:
lighttpd: build and bugfix patches

Signed-off-by: Glenn Strauss <gstrauss@gluelogic.com>

### DIFF
--- a/net/lighttpd/Makefile
+++ b/net/lighttpd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lighttpd
 PKG_VERSION:=1.4.56
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 # release candidate ~rcX testing; remove for release
 #PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-1.4.56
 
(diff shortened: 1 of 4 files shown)


neheb:
I don't know that this will fix it. There are two luas in the tree. I don't know which version meson will pick up like this.

gstrauss:
> I don't know that this will fix it. There are two luas in the tree. I don't know which version meson will pick up like this.

With this patch, lighttpd meson build will look for liblua.so and lua/lua.h.
With this patch, lighttpd meson build does not look for lua w/ version, e.g. liblua-5.3.so and lua-5.3/lua.h

lighttpd does not have requirements on a specific lua version and should work with any of them, so the openwrt packaging should end up detecting that the lighttpd build is satisfied by the declared dependeny on `liblua` package in the lighttpd/Makefile.

gstrauss:
BTW, in my test build for `target-i386_pentium-mmx_musl`, on which I have installed the `lua` package but not the `lua5.3` package, the build result had a dependency on `liblua.so.5.1.5` (from the `lua` package)

neheb:
Hmm OK.

PR #14117, opened 2020-12-02
strongswan: migrate to swanctl configurations

pprindeville:
Maintainer: @stintel 
Compile tested: x86_64, generic, HEAD (b0cb3052)
Run tested: same, built Strongswan ipk's, copied over to production systems, created IPsec profiles, and brought up tunnels

Description:

In the interests of having better Strongswan+XFRM integration, I'm rewriting the UCI support to use the `swanctl` backend instead of `ipsec`.

Note, this requires merge of PR #14028 to proceed.


pprindeville:
Should the current `ipsec.conf`-based `/etc/init.d/ipsec` be moved from `strongswan` to `strongswan-ipsec`, and the revamped version be put into `strongswan-swanctl`?

feckert:
@pprindeville Do you plan to add a conversion script from the current uci to the new uci also?
Since I have not looked into this the question arises whether this is possible at all.

pprindeville:
> Do you plan to add a conversion script from the current uci to the new uci also?
> Since I have not looked into this the question arises whether this is possible at all.

@feckert I've tried to make the new UCI as similar to the old UCI as possible.  I did find some mistakes in the old UCI which I've tried to fix.  I've also noticed that some things were misplaced in the old UCI, regarding `remote` vs. `tunnel`/`transport` sections.  That *will* require some hand-editing to make work, because there's no way to have a mechanical rewrite of that work without making some possibly incorrect assumptions.

Also, some parameters have fallen by the wayside, like `force_crypto_proposal`...  I'm wondering if I should silently ignore these, or (for the next few releases) complain when they're detected, but eventually retire that logic after a reasonable amount of time.

As @jow- pointed out in an email thread (sorry, offline conversation), what we really need is a schema for automated validation of the UCI including complaining about options that aren't valid (obsolete, misspelled, in the wrong section, etc.).

pprindeville:
A few things to note about these changes: the settings in `ipsec.conf` don't map 100% to those in `swanctl.conf`, and a certain number of hand-rewriting will be necessary when you make the transition.

Part of that has to do with the fact that some settings were per-peer in `ipsec.conf` but have become per-profile in `swanctl.conf`, that is a peer might have multiple profiles associated with it, and this corresponds to the `children { }` section within a particular `connection { }`.

There's no easy way to mechanically migrate that or I'd have looked into writing a migration script.

pprindeville:
@feckert One other observation: since this has to do with security and privacy, I wouldn't trust a conversion script blindly anyway, without doing a lot of manual double-checking of the results.  It's too sensitive to entrust purely to a mechanical process...

feckert:
> @feckert One other observation: since this has to do with security and privacy, I wouldn't trust a conversion script blindly anyway, without doing a lot of manual double-checking of the results. It's too sensitive to entrust purely to a mechanical process...

@pprindeville With this argument, it sounds plausible.
That was just a question.
Then I'll have to take a closer look in my setup when I switch to swanctl.

stintel:
> A few things to note about these changes: the settings in `ipsec.conf` don't map 100% to those in `swanctl.conf`, and a certain number of hand-rewriting will be necessary when you make the transition.
> 
> Part of that has to do with the fact that some settings were per-peer in `ipsec.conf` but have become per-profile in `swanctl.conf`, that is a peer might have multiple profiles associated with it, and this corresponds to the `children { }` section within a particular `connection { }`.
> 
> There's no easy way to mechanically migrate that or I'd have looked into writing a migration script.

I'm not happy with this. This can potentially result in people losing access to devices in the field, especially with corona virus related travel restrictions this is 100% unacceptable. There should be 100% compatibility or I will not accept the PR.

feckert:
How about we go two ways.

We can keep the old but also offer the new? I am not a strongswan expert, but from my point of view it would not be bad if we switch to the new configuration handling in openwrt strongswan upstream has already deprecated the old one. People could then switch to swanctl and we could then remove the old support in the next  openwrt release in 2022?

pprindeville:
> I'm not happy with this. This can potentially result in people losing access to devices in the field, especially with corona virus related travel restrictions this is 100% unacceptable. There should be 100% compatibility or I will not accept the PR.

@stintel Yeah, this is why I was suggesting putting the UCI handling into `strongswan-ipsec` and `strongswan-swanctl`... But to your point: this isn't a deficiency in my PR; it's a change in Strongswan about what options are supported (and where) in `ipsec` vs `swanctl`.

pprindeville:
Okay, how about we install the modified script as `ipsec-sc` and start to deprecate the original `ipsec` script?

pprindeville:
@stintel Can we close on PR #14028 which doesn't change the parameters (though it does add a couple of new ones)?

feckert:
@pprindeville @stintel As I said I am not an expert, but can we not to it this way?

Just add a check part to the ipsec.init script.
The check checks if there is already a configuration (old ipsec), if so give a warning that this is deprecated and start it, because we sill support this kind of configuration backend. If there is no old one, then use the new swanctl.

The advantage would be, that we give a release to the people time to change their uci configuration to new swanctl backend.
The people can then change the routers in the field, since you already have the new swanctl available and could do a smooth migration and there would be no hard cut.

pprindeville:
> The check checks if there is already a configuration (old ipsec), if so give a warning that this is deprecated and start it, because we sill support this kind of configuration backend. If there is no old one, then use the new swanctl.

@feckert Not sure I understand.  A `/etc/config/ipsec` file could go either way, i.e. it could be legacy and require the old interpretation of settings, or it could be new and added on a brand-new box with the new post-swanctl based packaging.  You can't tell which... unless we add a flag that's post-swanctl and that signifies that we created a specifically swanctl-based config.

vschagen:
There are already python scripts out there to convert ipsec.conf and ipsec.secrets into swanctl.conf

See: https://gitlab.com/Thermi/ipsec2swanctl/-/blob/master/ipsec2swanctl.py

Problem obviously is that it needs python installed on the router. I'm not familiar enough with the edge cases (like mentioned above eg multiple peer configurations) to know if the mentioned script works 100%.

pprindeville:
See PR #14708 instead.  The consensus is that we'll let people using `ipsec` continue to do so, while also offering more powerful `swanctl` support for those willing to migrate over (or set up from scratch).

PR #14127, opened 2020-12-03
ksmbd and ksmbd-tools 3.3.1

xdarklight:
Maintainer: @Andy2244
Compile tested: mips, lantiq
Run tested: lantiq BT Home Hub 5A

Description:
Updates ksmbd and ksmbd-tools to the latest upstream version, which is 3.3.1 (for both packages)
Upstream changelogs are included in the commit for each package

### DIFF
--- a/kernel/ksmbd/Makefile
+++ b/kernel/ksmbd/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ksmbd
-PKG_VERSION:=3.2.5
+PKG_VERSION:=3.3.1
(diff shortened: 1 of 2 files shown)


neheb:
I wonder how much of these kmods are actually needed.

xdarklight:
> I wonder how much of these kmods are actually needed.

`kmod-lib-crc32c` is a new dependency so I added it. I don't know about the existing ones (so far I assumed that they are fine because they are there, so they must be there for a reason)

neheb:
Right they're there for compatibility.

I wonder if it's possible to remove DES, ECB, and whatnot.

PR #14134, opened 2020-12-04
frr: update to 7.5

lucize:
also switch to codeload for latest patches applied to branch

Maintainer: me 
Tested x86/mips


lucize:
@neheb this should be ok also for ospf regression

Issue #14113, opened 2020-12-02
Darkstat multiple local networks


simm42:
Maintainer: @padre-lacroix (find it by checking history of the package Makefile)
Environment: (19.7, OpenWrt version)

Description:

I recently started using darkstat on my openwrt router and noted that you are the listed maintainer (assuming I have the right Jean-Michel Lacroix)

I wanted to use multiple local networks for the host diplay and graphing to include both the ipv4 and ipv6 traffic along with having my internal and DMZ traffic in the same graph.

Here is the patch if you want to include it in the maintained package source - I have also sent this to the package developer.

```

diff --git a/acct.c b/acct.c
index 540bf35..35460b8 100644
--- a/acct.c
+++ b/acct.c
@@ -37,8 +37,9 @@

 uint64_t acct_total_packets = 0, acct_total_bytes = 0;

+#define LOCAL_NET_MAX 10
 static int using_localnet4 = 0, using_localnet6 = 0;
-static struct addr localnet4, localmask4, localnet6, localmask6;
+static struct addr localnet4[LOCAL_NET_MAX], localmask4[LOCAL_NET_MAX], localnet6[LOCAL_NET_MAX], localmask6[LOCAL_NET_MAX];

 /* Parse the net/mask specification into two IPs or die trying. */
 void
@@ -120,13 +121,19 @@ acct_init_localnet(const char *spec)
    /* Register the correct netmask and calculate the correct net.  */
    addr_mask(&localnet, &localmask);
    if (localnet.family == IPv6) {
-      using_localnet6 = 1;
-      localnet6 = localnet;
-      localmask6 = localmask;
+      if(using_localnet6 >= LOCAL_NET_MAX){
+       errx(1, "Exceeded maximum IPv6 local networks");
+      }
+      localnet6[using_localnet6] = localnet;
+      localmask6[using_localnet6] = localmask;
+      using_localnet6++;
    } else {
-      using_localnet4 = 1;
-      localnet4 = localnet;
-      localmask4 = localmask;
+      if(using_localnet4 >= LOCAL_NET_MAX){
+       errx(1, "Exceeded maximum IPv4 local networks");
+      }
+      localnet4[using_localnet4] = localnet;
+      localmask4[using_localnet4] = localmask;
+      using_localnet4++;
    }

    verbosef("local network address: %s", addr_to_str(&localnet));
@@ -138,11 +145,15 @@ static int addr_is_local(const struct addr * const a,
    if (is_localip(a, local_ips))
[...]

PR #14135, opened 2020-12-04
Add support to darkstat to allow multiple local networks to be specified for montioring

simm42:
specified for montioring

eg allows ipv4 and ipv6 forwarded traffic to be monitored from both main network and dmz in single graph

Maintainer: @padre-lacroix  (find it by checking history of the package Makefile)

Description:
Add support to darkstat to allow multiple local networks to be specified for montioring

eg allows ipv4 and ipv6 forwarded traffic to be monitored from both main network and dmz in single graph

patched provided to package developer and acknowledged, but no schedule give for when they might be able to to add it to the base package

addresses: https://github.com/openwrt/packages/issues/14113

### DIFF
--- a/net/darkstat/Makefile
+++ b/net/darkstat/Makefile
@@ -1,7 +1,7 @@
 #
 # This is free software, licensed under the GNU General Public License v2.
 # See /LICENSE for more information.
-#
+# 
 
 include $(TOPDIR)/rules.mk
 
--- a/net/darkstat/Makefile
+++ b/net/darkstat/Makefile
@@ -1,13 +1,13 @@
 #
 # This is free software, licensed under the GNU General Public License v2.
 # See /LICENSE for more information.
-# 
+#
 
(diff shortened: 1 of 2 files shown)


simm42:
sucessfully installed and tested build artefact darkstat_3.0.719-5_mips_24kc.ipk on openwrt 19.07-4 build on lantiq/xrx200/bt_homehub_v5a install

net/darkstat/Makefile
```diff
@@ -1,7 +1,7 @@
 #
 # This is free software, licensed under the GNU General Public License v2.
 # See /LICENSE for more information.
-#
+#
```
neheb, line 4: ?

simm42: added

padre-lacroix:
That seems ok now.  I approve the change and this can be merged.
Sorry for the delay!

PR #14136, opened 2020-12-04
unbound: update to 1.13.0

EricLuehrsen:
Maintainer: me
Tested: Linksys WRT3200ACM
Description: Update to Unbound 1.13.0

### DIFF
--- a/net/unbound/Makefile
+++ b/net/unbound/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=unbound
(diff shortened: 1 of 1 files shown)


vcunat:
Do you get lots of lines like this?
> error: udp connect failed: Permission denied

I suspect that cut-down kernels may not support connected UDP sockets.  Those cases might want to add `udp-connect: no` to avoid Unbound trying it (and logging about it).

Note that upstream decided to use these to make some class of spoofing attacks harder, though I'd personally expect that `no` will be OK.

PR #14137, opened 2020-12-04
mosquitto: bump to v2.0 (plus new dependent package)

karlp:
Description: Bumps mosquitto to the new major v2 release.
Includes a new dependent package that I will (yay! joy!) take on maintainership for.

Without comments, I'll merge this myself in a few days.

### DIFF
--- a/libs/cjson/Makefile
+++ b/libs/cjson/Makefile
@@ -1,0 +1,53 @@
+# This Makefile is free software, SPDX codes: GPL-2.0-or-later OR MIT
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=cJSON
+PKG_VERSION:=1.7.14
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/DaveGamble/cJSON/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=fb50a663eefdc76bafa80c82bc045af13b1363e8f45cec8b442007aef6a41343
+
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+PKG_CPE_ID:=cpe:/a:cjson_project:cjson
+
+CMAKE_INSTALL:=1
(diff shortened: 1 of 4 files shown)


karlp:
mosquitto_ctrl shouldn't be attempting to be installed on nonssl builds, and worked locally, I'll try again with a "different" clean?

karlp:
Who can help me setup how these tests are run locally? I can't for the life of me reproduce it here.  make clean, select _only_ mosquitto-nossl, make package/mosquitto/{clean,compile} ==> works just fine.  What am I missing, and why is it different in these test builds?

karlp:
Ok, think I figured it out...
```
22:35 <karlp> the "## Continuous Integration
22:35 <karlp> chapter in Contributing has at least some docs on the test.sh magic, but nothing on setting up how the sdk builds work
22:42 <karlp> I have no idea how to turn https://github.com/openwrt/packages/blob/master/.github/workflows/multi-arch-test-build.yml#L51 into something I can do locally
22:43 <karlp> ah, brainwave, I think it references https://github.com/openwrt/gh-action-sdk/blob/master/entrypoint.sh
22:50 <karlp> not sure how it builds variants though
22:53 <karlp> would https://github.com/openwrt/gh-action-sdk/blob/master/entrypoint.sh#L85 not be better to just enable log files and cat the probelmatic packagfe's file, rather than rebuilding it?
22:53 <karlp> guess I shouldn't talk when I a) can't actualyl reproduce it and b) my packages don't pass
23:00 <karlp> ok, I can reproduce it now.
23:01 <karlp> it happens for building both variants in the same tree, and a package config that has a DEPENDS on one variant 
23:01 <karlp> but DEPENDS only blocks whether you can set it
23:01 <karlp> once it's set, it's set...
23:02 <karlp> for the bored on saturday night crowd: https://paste.jvnv.net/view/oXEeC
23:05 <karlp> who builds both options at the same time.... :)
```

karlp:
ok, now that it's all "green" I'll still merge this, maybe monday if no other comments

neheb:
I'd suggest CMake for mosquitto but I'm sure you'll NACK it :P 

I still keep https://github.com/openwrt/packages/pull/12782 locally.

karlp:
merged

PR #14147, opened 2020-12-06
squid: fix compilation with ccache

neheb:
Added a patch to remove BUILDCXXFLAGS. For some reason, TARGET_CXXFLAGS
are leaking.

Removed custom Build/Compile section. There's already PKG_INSTALL.

Maintainer: @ratkaj 
Compile tested: ath79

### DIFF
--- a/net/squid/patches/010-no-buildbxxflags.patch
+++ b/net/squid/patches/010-no-buildbxxflags.patch
@@ -1,0 +1,11 @@
+--- a/src/Makefile.am
++++ b/src/Makefile.am
+@@ -765,7 +765,7 @@ cache_cf.o: cf_parser.cci
+ 
+ # cf_gen builds the configuration files.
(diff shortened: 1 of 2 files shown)


net/squid/patches/010-no-buildbxxflags.patch
```diff
@@ -0,0 +1,11 @@
+--- a/src/Makefile.am
```
BKPepe, line 1: Was this patch sent to upstream? There isn't a commit author, commit description.

neheb: I don't think it's appropriate for upstream as this is specific to OpenWrt.

BKPepe: Hm, but this should be noted or written in this file, what it does, why it is needed. We need to have a clean and proper history.

PR #14155, opened 2020-12-07
zerotier: add patch to avoid including sys/auxv.h

neheb:
It seems it's not needed and not supported under uClibc-ng.

Maintainer: @mwarning 
Compile tested: arc700

Refreshed all patches.


mwarning:
Look good!

wongsyrone:
@neheb @mwarning 

I think this patch can be removed as uclibc-ng is removed from the openwrt repo.

mwarning:
https://github.com/openwrt/packages/pull/17279

PR #14162, opened 2020-12-07
rpcd-mod-lxc: add postinst to reload rpcd on update/installation

Cynerd:
Maintainer: @lperkov 
Compile tested: 
Run tested: 

Description:
This is dependency of luci-app-lxc and when users install that package
it is no way clear that they have to reload rpcd to get it working
correctly. Without it container listing does not work.
In general this reload should be in this package simply because other
rpcd-mod-* packages reload rpcd as well.

### DIFF
--- a/utils/rpcd-mod-lxc/Makefile
+++ b/utils/rpcd-mod-lxc/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=rpcd-mod-lxc
-PKG_RELEASE=20171206
+PKG_RELEASE=20201208
 
 PKG_LICENSE:=ISC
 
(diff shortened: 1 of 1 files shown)


neheb:
I think this needs a PKG_RELEASE bump.

Cynerd:
> I think this needs a PKG_RELEASE bump.

I think as well. I forgot to address it here. This package uses `PKG_RELEASE` as it seems instead of `PKG_VERSION`. I am not sure if I should just bump it to `PKG_RELEASE=20201208` or if I should rather move current `PKG_RELEASE` to `PKG_VERSION` and set `PKG_RELEASE=2`. What is the correct approach?

neheb:
I think update PKG_RELEASE. This is a local package.

PR #14163, opened 2020-12-08
krb5: install krb5-config to host

neheb:
Helps old packages that don't use pkgconfig.

Fix prefix paths.

Maintainer: @flyn-org 
Compile tested: ath79


flyn-org:
Looks good. Recommend merge.

PR #14166, opened 2020-12-08
neon: install neon-config to host path

neheb:
Helps old packages that do not use pkgconfig.

Fixed paths.

Maintainer: @fededim 
Compile tested: ath79

Interestingly enough with this change, davfs2 does not need to specify neon manually.


fededim:
Changes are fine for me.

PR #14169, opened 2020-12-08
libdaq: fix config file

neheb:
Install to host to fix compilation with depending packages.

Fix paths to point to OpenWrt and not OS.

Don't install to device. This is strictly an InstallDev thing.

Maintainer: @flyn-org 
Compile tested: ath79

Might fix https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a72/packages/snort/compile.txt no idea.


flyn-org:
Looks good. Did you test that snort still builds?

neheb:
Not quite. It requires a 64-bit target platform because of LuaJIT.

PR #14181, opened 2020-12-09
uci2: add library package

jpetrina:
Maintainer: @jpetrina
Compile tested: arm, ipq4018, master@r15169-36e0268aa6
Run tested: arm, ipq4018, master@r15169-36e0268aa6

Description:

UCI2 is a C library that provides an alternative UCI parser with an Abstract
Syntax Tree (AST) representation of configuration files.

### DIFF
--- a/libs/uci2/Makefile
+++ b/libs/uci2/Makefile
@@ -1,0 +1,52 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=uci2
+PKG_VERSION:=1.0
+PKG_RELEASE:=1
+
+PKG_SOURCE_URL:=https://github.com/sartura/uci2.git
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_VERSION:=2d082b587d32ce019fd94317026911e9b51f90dd
+PKG_MIRROR_HASH:=965339e85ff60724fc25ef812c117d203df63055821aa4d0006cc7f59c421b5b
+
+PKG_MAINTAINER:=Jakov Petrina <jakov.petrina@sartura.hr>
+PKG_LICENSE:=BSD-3-Clause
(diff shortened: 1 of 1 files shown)


libs/uci2/Makefile
```diff
@@ -0,0 +1,54 @@
...
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_VERSION:=2d082b587d32ce019fd94317026911e9b51f90dd
+PKG_MIRROR_HASH:=965339e85ff60724fc25ef812c117d203df63055821aa4d0006cc7f59c421b5b
+
+PKG_MAINTAINER:=Jakov Petrina <jakov.petrina@sartura.hr>
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_PARALLEL:=1
+CMAKE_INSTALL:=1
```
neheb, line 17: Not needed since there's an explicit InstallDev section.

libs/uci2/Makefile
```diff
@@ -0,0 +1,54 @@
...
+endef
+
+define Package/libuci2/description
+UCI2 is a C library that provides an alternative UCI parser with an Abstract
+Syntax Tree (AST) representation of configuration files.
+endef
+
+CMAKE_OPTIONS += \
+	-DENABLE_TESTS=OFF \
+	-DCMAKE_BUILD_TYPE=Release
```
neheb, line 40: This is already default.

libs/uci2/Makefile
jpetrina: Removed.

libs/uci2/Makefile
jpetrina: Removed, thanks.

PR #14187, opened 2020-12-09
augeas: disable selinux

BKPepe:
Maintainer: @ja-pa 
Run tested: N/A

Description:

Fixes:
Package augeas is missing dependencies for the following libraries:
libselinux.so.1

### DIFF
--- a/utils/augeas/Makefile
+++ b/utils/augeas/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=augeas
 PKG_VERSION:=1.12.0
-PKG_RELEASE=1
+PKG_RELEASE=2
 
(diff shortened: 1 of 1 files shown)


neheb (approved):

utils/augeas/Makefile
```diff
@@ -50,6 +50,9 @@ define Package/augeas-lense/description
   Set of Augeas lenses.
 endef
 
+CONFIGURE_ARGS+= \
+        --without-selinux \
```
ja-pa, line 54: I would remove this backslash since it's the only argument.

ja-pa:
Run tested on Omnia with OpenWrt master 

LGTM

utils/augeas/Makefile
BKPepe: Removed, but I didn't do compile neither run test.

PR #14191, opened 2020-12-10
openvpn-easy-rsa: bump to 3.0.8

luizluca:
New easyrsa will look for missing vars and x509-types where easyrsa
is located (following symlink). /usr/bin/easyrsa is now a link
to /usr/lib/easyrsa/easyrsa and /usr/lib/easyrsa/{vars,x509-types} a
link to /etc/easyrsa/{vars,x509-types}. This keeps the same previous
OpenWrt easyrsa behavior which tries to use $PWD/pki and
/etc/easyrsa/{vars,x509-types}, but without patching it.

Easyrsa can also use env vars to set pki root path (instead of
/usr/lib/easyrsa), pki path (instead of $PWD/pki) and vars path.
Those variables are commented in /etc/profile.d/50-openvpn-easy-rsa.sh
as an example of how to make easyrsa run independent of $PWD. That
scriptlet also sets $EASYRSA_TEMP_DIR from $EASYRSA_PKI/tmp to /tmp
in order to avoid writing to persistent media (normally flash). However,
as a profile scriptlet, it will only be used after session is restarted.

The "build" tgz was replaced by the "source" tar. "build" version has a
different file structure, making any patch backports too complex.

I'm also putting myself as maintainer.

Closes openwrt/openwrt#2926, since it moved to openwrt/packages.

Maintainer: me (since this PR)
Compile tested: arc_generic,armvirt_64,ath79_generic,powerpc,ramips_mt7620,x86_64,x86_generic
Run tested: x86_64 creating a new ca

Description:


luizluca:
easy-rsa is a bit messy to package as it is built to be executed in a standalone directory. I'm open for suggestions, even if it requires some upstream changes.

I really does not like the idea of keeping openssl-easyrsa.cnf -> openssl-1.0.cnf forever. Should I do some kind of post-install migration script instead of a symlink?

PR #14194, opened 2020-12-11
github: test i386_pentium-mmx instead of pentium4

neheb:
It seems the former causes more compilation failures due to not having
SSE.

Maintainer: @aparcar 

See: https://downloads.openwrt.org/snapshots/faillogs/i386_pentium-mmx/packages/i2pd/compile.txt


aparcar:
Sure go ahead

PR #14196, opened 2020-12-11
isc-dhcp: seeing crashes when attempting to update dynamic dns

pprindeville:
Maintainer: me
Compile tested: x86_64, master, HEAD (b0cb305236)
Run tested: same... Built and installed dhcpd, then tested with hand-generated dhcpd.conf file using rndc.key file and dynamic updates to locally running Bind server.

Description:


BKPepe:
I have one nitpick. Would it be possible to include in the commit message the same or similar to what is said in the issue?

pprindeville:
This will not be fixed upstream:

https://gitlab.isc.org/isc-projects/dhcp/-/issues/156#note_184076

PR #14198, opened 2020-12-11
python3: add python3-webbrowser package from python3-light

BKPepe:
Maintainers: @jefferyto , @commodo 
Compile and run tested on Turris Omnia, mvebu (cortex-a9), OpenWrt 19.07.5
I didn't test it so far on OpenWrt master, but I can do it based on request.

According to the documentation of webbrowser [1], it should allow displaying
Web-based documents to users.

Most likely, there is no much usage as there is not any browser
available for OpenWrt AFAIK.

However, somebody is using it because of Microsoft Authentication Library
which uses webbrowser module. This was requested on the Turris forum [2]
and it was confirmed that it works.

[1] https://docs.python.org/3/library/webbrowser.html
[2] https://forum.turris.cz/t/missing-webbrowser-module-in-python3-standard-distribution-on-turris-5-x/14407?u=pepe

### DIFF
--- a/lang/python/python3/Makefile
+++ b/lang/python/python3/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 include ../python3-version.mk
 
 PKG_NAME:=python3
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 PKG_VERSION:=$(PYTHON3_VERSION).$(PYTHON3_VERSION_MICRO)
 
 PKG_SOURCE:=Python-$(PKG_VERSION).tar.xz
@@ -262,7 +262,6 @@ define Py3Package/python3-light/filespec
 -|/usr/lib/python$(PYTHON3_VERSION)/lib-dynload/readline*.so
 -|/usr/lib/python$(PYTHON3_VERSION)/pdb.doc
 -|/usr/lib/python$(PYTHON3_VERSION)/test
--|/usr/lib/python$(PYTHON3_VERSION)/webbrowser.py
 -|/usr/lib/python$(PYTHON3_VERSION)/*/test
 -|/usr/lib/python$(PYTHON3_VERSION)/*/tests
 -|/usr/lib/python$(PYTHON3_VERSION)/_osx_support.py
--- a/lang/python/python3/files/python3-package-webbrowser.mk
+++ b/lang/python/python3/files/python3-package-webbrowser.mk
@@ -1,0 +1,16 @@
+#
+# Copyright (C) 2020 Josef Schlehofer <pepe.schlehofer@gmail.com>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+define Package/python3-webbrowser
+$(call Package/python3/Default)
+  TITLE:=Python $(PYTHON3_VERSION) Web-browser controller
+  DEPENDS:=+python3-light
+endef
+
+$(eval $(call Py3BasePackage,python3-webbrowser, \
+	/usr/lib/python$(PYTHON3_VERSION)/webbrowser.py \
+))


commodo:
The original split was done empirically, based on largest folders in the Python Lib.
Looking at that report it wasn't clear to me if installing `python3-light` was an option for the user that raised the report.

However, I don't mind having this separately.
Let's see if @jefferyto has any objections.

BKPepe:
> Looking at that report it wasn't clear to me if installing python3-light was an option for the user that raised the report.

It was not as it was removed in filespec.
https://github.com/openwrt/packages/blob/master/lang/python/python3/Makefile#L265

Maybe I should write it more clear.

commodo:
> > Looking at that report it wasn't clear to me if installing python3-light was an option for the user that raised the report.
> 
> It was not as it was removed in filespec.
> https://github.com/openwrt/packages/blob/master/lang/python/python3/Makefile#L265
> 
> Maybe I should write it more clear.

well, at ~24k unzipped, it would also be fine to leave it in python3-light
either works for me;

BKPepe:
I would not include it in python3-light let's have it slim as possible. I am still not convinced that there is going to be high usage of webbrowser. I'd have it rather in a separate package, but let's wait for Jeffery's thoughts.

commodo:
> I would not include it in python3-light let's have it slim as possible. I am still not convinced that there is going to be high usage of webbrowser. I'd have it rather in a separate package, but let's wait for Jeffery's thoughts.

i'm also a bit skeptical about it's usage; but if it works, it's one of those weird things with python on OpenWrt that I never imagined;
FWIW: I don't remember if I excluded webbrowser.py or whether that is how I inherited it.

jefferyto:
I'm okay with adding a separate package for webbrowser but until we ship some kind of browser package, this only changes the error from "cannot import `webbrowser`" to "cannot find/launch browser".

I would be interested to see if the testing Turris package worked for the original reporter and how they managed to work around the fact that there is no browser on their router.

jjcale:
I am the original reporter and the Turris package `python3-webbrowser_3.7.9-3.7-2_arm_cortex-a9_vfpv3-d16.ipk`  worked for me well. The reason I requested webbrowser module is as follows:

1. I need to download/upload files/to from OneDrive for Business through a Python script
2. OneDrive for Business is exposed through MS Graph API
3. MS Graph API requires MSAL (MS Authentication Library): `import msal`
4. And MSAL uses webbrowser module, that is standard in a full-blown Python3 distribution

root@turris:\~# cat /etc/*release
DISTRIB_ID='TurrisOS'
DISTRIB_RELEASE='5.1.4'
DISTRIB_REVISION='589c3cf'
DISTRIB_TARGET='mvebu/cortexa9'
DISTRIB_ARCH='arm_cortex-a9_vfpv3-d16'
DISTRIB_DESCRIPTION='TurrisOS 5.1.4 589c3cf4e08731e4971d2db6c83e16b5ab959878'
DISTRIB_TAINTS='busybox'
NAME="TurrisOS"
VERSION="5.1.4"
ID="turrisos"
ID_LIKE="lede openwrt"
PRETTY_NAME="TurrisOS 5.1.4"
VERSION_ID="5.1.4"
HOME_URL="https://www.turris.cz/"
BUG_URL="https://gitlab.labs.nic.cz/groups/turris/-/issues/"
SUPPORT_URL="https://www.turris.cz/support/"
BUILD_ID="589c3cf"
OPENWRT_BOARD="mvebu/cortexa9"
OPENWRT_ARCH="arm_cortex-a9_vfpv3-d16"
OPENWRT_TAINTS="busybox"
OPENWRT_DEVICE_MANUFACTURER="CZ.NIC"
OPENWRT_DEVICE_MANUFACTURER_URL="https://www.turris.cz/"
OPENWRT_DEVICE_PRODUCT="Turris Omnia"
OPENWRT_DEVICE_REVISION="v0"
OPENWRT_RELEASE="TurrisOS 5.1.4 589c3cf4e08731e4971d2db6c83e16b5ab959878"

sha256sum python3-webbrowser_3.7.9-3.7-2_arm_cortex-a9_vfpv3-d16.ipk 
0b65a5eb3399f4c9f9c1713052887c935c0465d4f88d72709484f448451a1dcb  python3-webbrowser_3.7.9-3.7-2_arm_cortex-a9_vfpv3-d16.ipk`

root@turris:\~# opkg list | grep webbrowser
python3-webbrowser - 3.7.9-3.7-2
root@turris:~# python3
Python 3.7.9 (default, Nov 03 2020, 15:03:41) 
[GCC 7.5.0] on linux
>>> import webbrowser
>>>

jefferyto:
@jjcale Thanks for replying - it's clear from your original Turris forum post why you wanted the `webbrowser` module.

But what happens when you run your script? Does MSAL not throw an error saying it could not launch a web browser?

If it doesn't actually try to open a browser then there is no point to include the `webbrowser` module; perhaps it would be better to patch the library to load `webbrowser` only when it is necessary.

jjcale:
@jefferyto here is the script that fetches the data from my OneDrive for Business:

1)

import sys
import json
import logging
import requests
import msal

\# enable optional logging
\# logging.basicConfig(level=logging.DEBUG)

config = {
  "authority": "https://login.microsoftonline.com/my_authority",
  "client_id": "my_client_id",
  "scope": [ "https://graph.microsoft.com/.default" ],
  "secret": "my_1y_secret",
  "endpoint": "https://graph.microsoft.com/v1.0/users/my_user_id/drive/root:/my_path/screentime.xlsx:/content"
}

\# create a preferably long-lived app instance which maintains a token cache.
app = msal.ConfidentialClientApplication(
    config["client_id"], authority=config["authority"],
    client_credential=config["secret"],
    )

result = None
result = app.acquire_token_silent(config["scope"], account=None)

if not result:
    logging.info("No suitable token exists in cache. Let's get a new one from AAD.")
    result = app.acquire_token_for_client(scopes=config["scope"])

if "access_token" in result:
    \# call graph using the access token  
    graph_content = requests.get(config["endpoint"], headers={'Authorization': 'Bearer ' + result['access_token']}, )._content

    f = open("/tmp/screentime.xlsx", "wb")
    f.write(graph_content) 
    f.close() 

else:
    \# report error
    print(result.get("error"))
    print(result.get("error_description"))
    print(result.get("correlation_id"))
`

2)
before I had installed the webrowser module, it threw 7 errors and complained that webbrowser was not imported.

3)
if you need more debugging and/or inputs, please let me know what commands I should run and I will give it a shot.

Tx

jefferyto:
@jjcale The next release of MSAL should have a fix for this issue (AzureAD/microsoft-authentication-library-for-python#288).

jjcale:
@jefferyto @BKPepe @commodo Great. Thank you everyone.

BKPepe:
@jefferyto thank you!

So, this is no longer needed as there are no browsers or packages which require this package as a dependency.

Closing. Maybe in the future, if there is going to be a browser, I will open this.

dynasticorpheus:
@BKPepe  Apologies for reviving this (very) old pull request  but wanted to check if we could still consider going ahead with adding this optional package to openwrt? Running home assistant in a virtual environment on openwrt and it needs webbrowser.py. Would be nice to have this made available through regular package install.

PR #14204, opened 2020-12-11
collectd: add dhcpleases plugin

PolynomialDivision:
Count the current dhcpleases. Currently, we use a bash script that does the same job (Freifunk Berlin). We want to use native collectd plugin for that.

Maintainer: @hnyman 
Compile tested: trunk
Run tested: 19.07.5

### DIFF
--- a/utils/collectd/Makefile
+++ b/utils/collectd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=collectd
 PKG_VERSION:=5.12.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=https://collectd.org/files/ \
@@ -144,6 +144,7 @@ COLLECTD_PLUGINS_SELECTED:= \
 	csv \
 	curl \
 	df \
+	dhcpleases \
 	disk \
 	dns \
 	email \
@@ -408,6 +409,7 @@ $(eval $(call BuildPlugin,csv,CSV output,csv,))
 $(eval $(call BuildPlugin,curl,cURL input,curl,+PACKAGE_collectd-mod-curl:libcurl))
 #$(eval $(call BuildPlugin,dbi,relational database input,dbi,+PACKAGE_collectd-mod-dbi:libdbi))
 $(eval $(call BuildPlugin,df,disk space input,df,))
+$(eval $(call BuildPlugin,dhcpleases,show dhcpleases,dhcpleases,))
 $(eval $(call BuildPlugin,disk,disk usage/timing input,disk,))
 $(eval $(call BuildPlugin,dns,DNS traffic input,dns,+PACKAGE_collectd-mod-dns:libpcap))
 $(eval $(call BuildPlugin,email,email output,email,))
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
My plugin is correct. The build error is:
```

Run docker build -t test-container --build-arg ARCH .github/workflows/
Sending build context to Docker daemon  8.192kB

Step 1/4 : ARG ARCH=x86-64
Step 2/4 : FROM openwrtorg/rootfs:$ARCH
manifest for openwrtorg/rootfs:i386_pentium-mmx not found: manifest unknown: manifest unknown
```

neheb:
Yeah I have no idea. Maybe @aparcar knows.

hnyman:
Have you already proposed this at upstream collectd?

PolynomialDivision:
> Have you already proposed this at upstream collectd?

Yes.
- https://github.com/collectd/collectd/pull/3790
- https://github.com/collectd/collectd/pull/3789

We need this for freifunk. I will remove the patches again, if they get accepted.

PolynomialDivision:
@BKPepe changed that now to the collectd upstream patch.

BKPepe:
Thanks! Even though, I don't know why my comment is gone.

PolynomialDivision:
You made a comment here ;)
https://github.com/collectd/collectd/pull/3790#issuecomment-743722614

BKPepe:
Ehh... my bad... Sorry!

PolynomialDivision:
No problem. ;)

I made the same PR for 19.07. Can we then merge everything? :)
19.07:
https://github.com/openwrt/packages/pull/14206

And that here. ;)
Luci-Statistics:
https://github.com/openwrt/luci/pull/4654

PolynomialDivision:
I have so many open PRs. Can we merge this, please? xD
I'm slowly losing the overview. ^^

hnyman:
> Can we merge this, please? xD

Would be nice hear first some feedback/advice from upstream.

> I have so many open PRs. 

No need to open 19.07 yet.
Let's get master first in.

PolynomialDivision:
> Would be nice hear first some feedback/advice from upstream.

Okay. But there are so many open PRs in collectd repro. :/ And the last commit was on 15 September.

utils/collectd/Makefile
```diff
@@ -146,6 +146,7 @@ COLLECTD_PLUGINS_SELECTED:= \
 	df \
 	disk \
 	dns \
+	dhcpleases \
```
hnyman, line 149: This line is not sorted alphabetically.   
df < dhcpleases < disk < dns

Same goes also for the eval BuildPlugin list below list .

And seems to be a problem also in the upstream patch. You have generally placed the dhcp... line/section as the "last of d" instead of sorted properly into middle of the d items.

PolynomialDivision: changed everywhere, even upstream plugin. ;)

PolynomialDivision:
> You opened PRs at collectd 4 days ago. You should wait a little bit. Christmas is coming.

It seems i have to be more patient. ;)
I just would be very happy to see the commits merged and be finally done with the migration of our statistics. ;)

hnyman:
I compiled this and also the LuCI part for ipq806x/R7800. Seem to work ok.
I will merge this, and will also merge the LuCI part (with slight tweaks to the graph).


For 19.07, I think that it would be possible to backport the collectd part, but the LuCI part would be more difficult, as there is the lua/js difference, so the backport would require more effort.

PolynomialDivision:
> For 19.07, I think that it would be possible to backport the collectd part, but the LuCI part would be more difficult, as there is the lua/js difference, so the backport would require more effort.

Okay. I will wait for 20.X ;)

PR #14206, opened 2020-12-11
collectd: add dhcpleases plugin

PolynomialDivision:
Count the current dhcpleases. Currently, we use a bash script
that does the same job (Freifunk Berlin). We want to use native
collectd plugin for that.

Maintainer:@hnyman

### DIFF
--- a/utils/collectd/Makefile
+++ b/utils/collectd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=collectd
 PKG_VERSION:=5.9.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


BKPepe:
Are we sure that we want to merge and have new features in OpenWrt 19.07? :thinking:

PolynomialDivision:
> Are we sure that we want to merge and have new features in OpenWrt 19.07? thinking

Hmmm. I would be happy to have that statistics in 19.07 since we build our community wireless mesh network on it.
But if 20.X will be soon released, I'm also fine. ;)

hnyman:
We can later backport minor features, if there is no negative feedback from master once it is there.

PolynomialDivision:
Okay, then I will close this and reopen later. ;)

PR #14201, opened 2020-12-11
curl: update to version 7.74.0 (security fix)

ja-pa:
Maintainer: @kaloz 
Compile tested: Turris Omnia (TOS6), OpenWrt master
Run tested: Turris Omnia (TOS6) OpenWrt master

Description:
This PR updates curl to version 7.74.0 [Changelog](https://curl.se/changes.html#7_74_0)

This release fixes multiple security issues CVE-2020-8286, CVE-2020-8285, CVE-2020-8284 and removes SRP patch which is part of the release.

PR #14212, opened 2020-12-12
simple-adblock: bugfix - config update

stangri:
Maintainer: me
Compile tested: ramips, ER-X, 19.07.5
Run tested: ramips, ER-X, 19.07.5, start/download blocklists

Description:
This is a bugfix to update the default config and config update files, otherwise new installs will be producing errors for no longer active https://www.malwaredomainlist.com/hostslist/hosts.txt list.
This also includes bugfix to use `extra_command` only on supported systems and small Makefile improvement.

### DIFF
--- a/net/simple-adblock/Makefile
+++ b/net/simple-adblock/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=simple-adblock
 PKG_VERSION:=1.8.4
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.net>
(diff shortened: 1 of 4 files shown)


stangri:
Regarding the failed check: `manifest for openwrtorg/rootfs:i386_pentium-mmx not found: manifest unknown: manifest unknown` -- is there something I can do about it?

hnyman:
> Regarding the failed check: manifest for openwrtorg/rootfs:i386_pentium-mmx not found
>
>  is there something I can do about it?

No. That MMX error message has materialized in other PRs too.
Likely due to #14194,  and based on discussion in https://github.com/openwrt/packages/pull/14201#issuecomment-743196822 the error message will be there until @aparcar (or somebody) adds a suitable CI test image to the system.

PR #14217, opened 2020-12-13
boost: Updates package to version 1.75

ClaymorePT:
Maintainer: @ClaymorePT 
Compile tested: BCM2711 Boards - Raspberry Pi 4 (64bits)
Run tested: N/A

Description:
This commit updates boost to version 1.75

This release brings three new packages
* [JSON](https://www.boost.org/libs/json/) parsing, serialization, and DOM in C++11, from Vinnie Falco and  Krystian Stasiowski.
* [LEAF](https://www.boost.org/libs/leaf/): A lightweight error-handling library for C++11, from Emil  Dotchevski.
* [PFR](https://www.boost.org/libs/pfr/): Basic reflection without macro or boilerplate code for user  defined types, from Antony Polukhin.

[More info about Boost 1.75.0 can be found at the usual place](https://www.boost.org/users/history/version_1_75_0.html).


libs/boost/patches/001-uclibc-asio.patch
```diff
@@ -29,40 +29,40 @@ index 38d4b2a61..e16cc8b00 100644
  #include <sys/stat.h>
  #include <sys/types.h>
  #include <fcntl.h>
--#if __GLIBC__ == 2 && __GLIBC_MINOR__ < 8
-+#if __GLIBC__ == 2 && __GLIBC_MINOR__ < 2
+-#if __GLIBC__ == 2 && __GLIBC_MINOR__ < 8 && !defined(__UCLIBC__)
++#if __GLIBC__ == 2 && __GLIBC_MINOR__ < 2 && !defined(__UCLIBC__)
```
neheb, line 33: Why this change?

ClaymorePT:
@neheb The patch needed to be updated otherwise it would fail to be applied. Only that extra macro verification was added to the source. Creating a new patch was more troublesome. As so, I just updated directly the patch.

libs/boost/patches/001-uclibc-asio.patch
ClaymorePT: @neheb The patch needed to be updated otherwise it would fail to be applied. Only that extra macro verification was added to the source. Creating a new patch was more troublesome. As so, I just updated directly the patch.

neheb: That makes no sense. __GLIBC_MINOR__ is still 2 under uClibc-ng. The change from 8 to 2 is what should fix it.

Out of curiosity, did you run make package/boost/refresh ?

ClaymorePT: No. My usual procedure is to `wget` the master, script update & install the packages, update the makefile, go to menu config and enable all packages from boost (+ python3) and build all. Check builded packages and if all goes well without errors, I push the changes. If not, check errors, make necessary adjustments, then clean and compile again.

It has served me well in the past.
Btw, `&& !defined(__UCLIBC__)` was added in the boost sources. Is OpenWRT even supporting uclibc? I don't recall seeing it in the advance toolchain options for the Raspberry Pi 4 target.

neheb: I see now. This patch can be removed. Upstream fixed the issue differently it seems.

make package/x/refresh will soon be added as a check. Although it seems that without this patch, boost is patch free.

ClaymorePT: Yep. I just also concluded that... Sorry about the mess. I was in automatic mode and didn't event checked the versions properly.

neheb: I have that patch as a PR. Seems the ASIO maintainers don't look at them.

ClaymorePT: Just removed the patch. Since there are no more patches, removed directory as well.

brvphoenix:
@ClaymorePT 

The v1.75.0 version introduces some new libs which should also be added.

ref: https://www.boost.org/users/history/version_1_75_0.html

ClaymorePT:
@brvphoenix  Yes I was aware. Boost.JSON can still work as a header-only.
I've added it in https://github.com/openwrt/packages/pull/14217/commits/6b79c46bb4cda0c492b9b09defeeed3aaa4f9c68

The others (LEAF and PFR) seem to be header-only libraries. I did not find a B2 target for them.

brvphoenix:
> @brvphoenix Yes I was aware. Boost.JSON can still work as a header-only.
> I've added it in [6b79c46](https://github.com/openwrt/packages/commit/6b79c46bb4cda0c492b9b09defeeed3aaa4f9c68)
> 
> The others (LEAF and PFR) seem to be header-only libraries. I did not find a B2 target for them.

👍

PR #14225, opened 2020-12-14
watchcat: add restart_iface and ping size features + refactor

nickberry17:
Maintainer: @roger- 
Compile tested: ath79/Telco T1
Run tested: ath79/Telco T1, git master
Description:
Adds features and refactors code with aim to be more maintainable.  The package version bumps multiple times because this is split across multiple commits.

Features
- ability to restart a specific interface
- ability to specify ping packet size
- backwards compatibility with older mode names

LuCI
luci-app-watchcat has been updated to support these features: https://github.com/openwrt/luci/pull/4658

### DIFF
--- a/utils/watchcat/files/initd_watchcat
+++ b/utils/watchcat/files/initd_watchcat
@@ -5,10 +5,10 @@ START=97
 PIDFILE="/tmp/run/watchcat"
 
 append_string() {
-	local varname="$1"
-	local add="$2"
-	local separator="${3:- }"
-	local actual
+	varname="$1"
+	add="$2"
+	separator="${3:- }"
+	actual
 	eval "actual=\$$varname"
 
 	new="${actual:+$actual$separator}$add"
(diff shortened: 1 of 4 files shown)


utils/watchcat/files/watchcat.sh
```diff
@@ -45,29 +46,98 @@ reboot_now() {
...
+			if ping -I "$iface" -s "$ping_size" -c 1 "$host" &>/dev/null; then
+				time_lastcheck_withinternet="$time_now"
+			else
+				logger -p daemon.info -t "watchcat[$$]" "no internet connectivity on "$iface" for $((time_now - time_lastcheck_withinternet)). Restarting "$iface" after reaching $failure_period"
+			fi
+		done
+
+		[ "$((time_now - time_lastcheck_withinternet))" -ge "$failure_period" ] && {
+
+			if [ "$mm_iface_name" -ne "" ]; then
```
wulfy23, line 114: inverted logic?

utils/watchcat/files/initd_watchcat
```diff
@@ -17,91 +17,114 @@ append_string() {
...
+	config_get mode "$1" mode
+	config_get pinghosts "$1" pinghosts
+	config_get pingperiod "$1" pingperiod
+	config_get forcedelay "$1" forcedelay
+	config_get pingsize "$1" pingsize
+	config_get interface "$1" interface
+	config_get mmifacename "$1" mmifacename
+	config_get unlockbands "$1" unlockbands
+
+	local forcedelay_default="0"
```
wulfy23, line 43: positive you want 0 here?

utils/watchcat/files/watchcat.sh
nickberry17: Thanks for catching that, fixed. Wrong comparison operator for strings.

utils/watchcat/files/initd_watchcat
nickberry17: I can put it back to 60s if that makes more sense for most users.  I just thought I would disable the sysreq reboot by default since it has the potential to lead to data loss.

nickberry17:
Does anyone know why the build is failing on arc_arc700?

utils/watchcat/files/initd_watchcat
wulfy23: its kind of a weird place to static out such a hardcoded setting... one solution might be to use ":-N" here... and have a global option in uci for this...

or something equivalent... which can be ignored if... ' ! -z 'custom_reboot_script'... blah blah blah... etc. etc. i.e. i'd expect a few pre-conditions or 'alt-action-methods'... or even 'per-interface/watch' action method logic to proceed this eventually...

neheb:
Don't worry too much about it. It should go away soon.

utils/watchcat/files/initd_watchcat
nickberry17: I refactored using `config_get`s  parameter for a default value, which let me remove a lot of the conditional checks later on in the file.

nickberry17:
Are we ready to merge this?  I think the main benefit is that it's now easy to extend watchcat to support more actions.

utils/watchcat/files/watchcat.sh
```diff
@@ -1,65 +1,223 @@
...
+ping_reboot)
+    watchcat_ping "$2" "$3" "$4" "$5" "$6"
+    ;;
+restart_iface)
+    watchcat_monitor_network "$2" "$3" "$4" "$5" "$6" "$7"
+    ;;
+*)
+    echo "Error: invalid mode selected: $mode"
+    ;;
+esac
```
feckert, line 223: The whole file now has spaces instead of tabs.
This increases the file size unnecessarily.
Can we avoid this next time?
With one file this is not a problem, but if there are many in the system this becomes important.

PR #14226, opened 2020-12-14
keyutils: update to 1.6.3 and improve packaging

dangowrt:
Maintainer: @flyn-org (?)
Compile tested: x86/64
Run tested: x86/64

Description:
 * switch to kernel.org upstream (most recent versions seem to be available only there)
 * package request-key, key.dns_resolver and their config file into new package 'keyutils'
 * refresh patches


neheb:
Looks fine.

PR #14227, opened 2020-12-14
CI: test aarch64_cortex-a53 instead of _generic

aparcar:
The aarch64_cortex-a53 architecture is used by more targets and should
therefore be tested rather than the relatively rare _generic one.


aparcar:
This may fails for now as armvirt nor Docker wasn't rebuild just yte.

BKPepe (approved):
Thanks!

aparcar:
I'll merge it once the entire chain is updated

PR #14228, opened 2020-12-15
vips: do not install libvips-cpp

neheb:
Nothing uses it. Allows to slim the package down.

Also made the InstallDev section more explicit.

Maintainer: @flyn-org 
Compile tested: ath79

### DIFF
--- a/libs/vips/Makefile
+++ b/libs/vips/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=vips
 PKG_VERSION:=8.10.2
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


flyn-org:
Looks good. Recommend merge.

The back story is that I used to build vips without C++ support, removing the need for libstdc++ during the build. However, the patch requirements eventually became difficult enough that this was not worth the time required. I tried to work with upstream to fix the configure flag that would deactivate C++ support, but autotools was stubborn in deducing the need C++ compiler even when we did not explicitly request it. (John Cupitt, the vips maintainer, has been great, by the way.) At least that is what I remember; I cannot find an archive of VIPSIP@JISCMAIL.AC.UK.

neheb:
Right. libvips itself depends on a libc++. libvips-cpp is not used by anything.

Issue #14123, opened 2020-12-03
mwan3: missing routing table 1 on wan interface after restart


thomaslunze:
I am running mwan3 with 2 wan interfaces (wan and wan1). With the latest build firmware I am having a routing issue.

see my troubleshooting output here:
Software-Version
-------------------------------------------------
OpenWrt - OpenWrt SNAPSHOT r14966-7330348f2d
LuCI - git-20.311.85590-5c5b134

Output of "ip a show"
-------------------------------------------------
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1508 qdisc mq state UP group default qlen 1024
    link/ether 14:91:82:c2:82:ca brd ff:ff:ff:ff:ff:ff
    inet6 fe80::1691:82ff:fec2:82ca/64 scope link 
       valid_lft forever preferred_lft forever
3: lan4@eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master br-lan state UP group default qlen 1000
    link/ether 14:91:82:c2:82:ca brd ff:ff:ff:ff:ff:ff
4: lan3@eth0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue master br-lan state LOWERLAYERDOWN group default qlen 1000
    link/ether 14:91:82:c2:82:ca brd ff:ff:ff:ff:ff:ff
5: lan2@eth0: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 14:91:82:c2:82:ca brd ff:ff:ff:ff:ff:ff
6: lan1@eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 14:91:82:c2:82:ca brd ff:ff:ff:ff:ff:ff
    inet 192.168.3.2/24 brd 192.168.3.255 scope global lan1
       valid_lft forever preferred_lft forever
[...]

Issue #14003, opened 2020-11-24
mwan3: error on ifup event


pgaufillet:
Maintainer: @feckert @aaronjg 
Environment: TP Link Archer C7 v2, OpenWRT r15024-bc64b9c32e

Description: when an interface goes up, mwan3 fails at updating the corresponding routing table:

```
Sat Nov 28 13:03:18 2020 user.notice mwan3-hotplug[3753]: Execute ifup event on interface wan (eth0.2)
Sat Nov 28 13:03:18 2020 user.warn mwan3-hotplug[3753]: failed to add default via 192.168.70.1 dev eth0.2 proto static src 192.168.70.3 to table 1
Sat Nov 28 13:03:18 2020 user.warn mwan3-hotplug[3753]: failed to add 192.168.70.0/24 dev eth0.2 proto kernel scope link src 192.168.70.3 to table 1
Sat Nov 28 13:03:20 2020 user.info mwan3track[1803]: Detect ifup event on interface wan ()
Sat Nov 28 13:03:21 2020 user.notice mwan3track[1803]: Interface wan (eth0.2) is online
Sat Nov 28 13:03:27 2020 user.notice mwan3-hotplug[4542]: Execute ifup event on interface vpn (tun0)
Sat Nov 28 13:03:27 2020 user.warn mwan3-hotplug[4542]: failed to add 10.45.3.0/24 via 10.45.3.13 dev tun0 to table 2
Sat Nov 28 13:03:27 2020 user.warn mwan3-hotplug[4542]: failed to add 10.45.3.1 via 10.45.3.13 dev tun0 to table 2
Sat Nov 28 13:03:27 2020 user.warn mwan3-hotplug[4542]: failed to add 10.45.3.13 dev tun0 proto kernel scope link src 10.45.3.14 to table 2
Sat Nov 28 13:03:27 2020 user.warn mwan3-hotplug[4542]: failed to add 192.168.0.0/24 via 10.45.3.13 dev tun0 to table 2
Sat Nov 28 13:03:27 2020 user.warn mwan3-hotplug[4542]: failed to add 192.168.2.0/24 via 10.45.3.13 dev tun0 to table 2
Sat Nov 28 13:03:27 2020 user.warn mwan3-hotplug[4542]: failed to add 192.168.50.0/24 via 10.45.3.13 dev tun0 to table 2

```

aaronjg:
[...]

PR #14231, opened 2020-12-15
mwan3: linkdown fix

aaronjg:
Maintainer: @feckert 

fix issue on router startup where routes are marked 'linkdown' before interfaces come on line. These routes must be added to the mwan3 routing tables. I have been unable to reproduce, but @feckert has tested and confirmed that it fixes the bug.

fixes: #14003
fixes: #14123

### DIFF
--- a/net/mwan3/files/usr/sbin/mwan3rtmon
+++ b/net/mwan3/files/usr/sbin/mwan3rtmon
@@ -22,7 +22,7 @@ func_trap()
 
 mwan3_add_all_routes()
 {
-	local tid IP IPT route_line family active_tbls tid initial_state
+	local tid IP IPT route_line family active_tbls tid initial_state error
 	local ipv=$1
 
 	add_active_tbls()
@@ -40,8 +40,8 @@ mwan3_add_all_routes()
 	{
 		let tid++
 		[ -n "${active_tbls##* $tid *}" ] && return
-		$IP route add table $tid $route_line ||
-			LOG warn "failed to add $route_line to table $tid"
+		error=$($IP route add table $tid $route_line 2>&1) ||
+			LOG warn "failed to add $route_line to table $tid - error: $error"
 	}
 
 	mwan3_update_dev_to_table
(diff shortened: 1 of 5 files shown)


net/mwan3/files/usr/sbin/mwan3rtmon
```diff
@@ -155,7 +155,7 @@ main()
 		IP="$IP4"
 	fi
 	mwan3_init
-	sh -c "echo \$\$; exec $IP monitor route" | {
+	sh -c "KILL -SIGSTOP $$; echo \$\$; exec $IP monitor route" | {
```
feckert, line 159: @aaronjg Can you explain to me the problem you are fixing here?

feckert:
@aaronjg could you please update the PKG_VERSION

net/mwan3/files/usr/sbin/mwan3rtmon
aaronjg: We must handle the case when new routes are being added while we are processing the initial routing table. I had initially solved this by suspending the ip monitor reading process shortly after it was started, but that still allowed for a race condition.

This was designed to suspend the process a bit sooner. On closer inspection, it appears it was actually suspending the monitor process, rather than the reading process, so I fixed this with 34b58f40.

This is a pretty edge case that is being handled, so if you want to keep this P/R cleaner I can drop that commit and the subsequent one from this P/R.

aaronjg:
> @aaronjg could you please update the PKG_VERSION

Done

feckert:
@aaronjg Sorry if I annoy you again.
But can you please update also the PKG_VERSION and change the PKG_RELEASE?
So that we have the following version
```
PKG_VERSION 2.10.4
PKG_RELEASE 1
```
PKG_VERSION is for source code changes
PKG_RELEASE is for makefile changes.

Sources that are maintained in openwrt repository, this does not make much sense I know!
There was already a discussion in the community.
But as long as this is not clarified I handle it this why in mwan3.

feckert:
@aaronjg I changed the version string to https://github.com/openwrt/packages/commit/85dbb482017faec4d69c21b57a5a0afb8095f48a and pushed to master without merge commit.
Thanks :+1: for the fix

aaronjg:
> @aaronjg I changed the version string to [85dbb48](https://github.com/openwrt/packages/commit/85dbb482017faec4d69c21b57a5a0afb8095f48a) and pushed to master without merge commit.
> Thanks 👍 for the fix

Thanks for taking care of that!

PR #14235, opened 2020-12-15
[openwrt-19.07] utils/lcd4linux: fix package source

sch-m:
Maintainer: @feckert 
Compile tested: lantiq xrx200, latest openwrt-19.07
Run tested: no

Description:
The former svn source is not available any more.
To keep the same version in openwrt-19.07, only change package source to openwrt sources server.


feckert:
Because of this [change](https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=605adb1023efda0c29d160839fccc20801b717cd) in the branch openwrt-19.07and because this package has no download hash (I have heard that this is not possible with svn, because there is always a different hash for the same commit.) and because the upstream repo is not available anymore -> own builds should fail now (In any case, this is also the case on my system).

@neheb People who maintain their own build should now run into an error on openwrt-19.07?
The question now is why is this built not failing on the openwrt buildbots?

My colleague has seen that the package is on the openwrt download server, so he wants to use it.
Is this a problem?

In the master branch I have already set this to my [repository](https://github.com/openwrt/packages/commit/d1b6c0c7976a9841acceda8c5f65c151134c1782#diff-3c777ab4bd147847a72048939ab97bdc2fbcbcc5e30c43e98c0c077715213f12).

neheb:
Because the package is available there.

feckert:
@neheb Excuse me for asking, but if the package is present but no checksum is included in the Makefile, then the check should fail and the buildsystem wants to download the package. This is becaus of this [change](https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=605adb1023efda0c29d160839fccc20801b717cd).

On my system the package is also in the dl folder but since this change my builds fail because the buildsystem wants to try to load the package from the svn. I just want to understand why the build system of openwrt behaves differently.

neheb:
Ummm this commit adds a new URL and HASH. git or svn is not used. It's a simple tarball.

PR #14233, opened 2020-12-15
bind: start named early enough to be present when isc-dhcp starts

pprindeville:
Maintainer: @nmeyerhans 
Compile tested: x86_64, generic, HEAD (b0cb305236)
Run tested: same, built ipk's and reinstalled on production dhcp/bind server in network

Description:

Start named before dhcpd so that dhcpd can prime the local zones at startup.

Restore the empty domain zone for rfc1918 addresses that previously existed.

Create an additional subsidiary named.conf.local file (initially empty) in /tmp/bind/ that can be seeded with dynamic zones and primed with rndc reload, and add it to the watched list of config files for procd.

PR #14240, opened 2020-12-15
isc-dhcp: add dynamic dns integration for local bind server

pprindeville:
Maintainer: me
Compile tested: x86_64, generic, HEAD (b0cb305236)
Run tested: x86_64, generic, HEAD (b0cb305236), built .ipk's and installed on production dhcpd/dns server

Description:

If `dynamicdns` is true for the `isc_dhcpd` section, then the following happens:

1. an empty "seed" zone (`db.empty`) is copied into `/tmp/bind/db.$domain if absent;
2. same for `/tmp/bind/db.168.192.in-addr.arpa`, if absent;
3. `named` is configured with two new zones via `/tmp/bind/named.conf.local` and told to reload;
4. additional configuration is provisioned for `dhcpd.conf` as:

```
ddns-domainname "$domain.";
ddns-update-style interim;
ddns-updates on;
ignore client-updates;

update-static-leases on;
use-host-decl-names on;
option domain-name "$domain.";
update-conflict-detection off;
update-optimization off;

include "/tmp/run/dhcpd-rndc.key";

zone $domain. {
       primary 127.0.0.1;
       key rndc-key;
}

zone 168.192.in-addr.arpa. {
       primary 127.0.0.1;
       key rndc-key;
}
```

and then `dhcpd` is started normally (but now after Bind is running).

Requires PR #14233.


neheb:
You need to rebase this properly.

PR #14242, opened 2020-12-16
nss: update to 3.60

neheb:
Added patch to fix compilation with musl 1.2.x.

Refreshed others.

Maintainer: @lucize 
Compile tested: ath79

@lucize what's with the first patch? It's creating files that were apparently present before but no longer are.

edit: NVM I see now.


lucize:
I've taken it from Linux from scratch project

neheb:
OK

PR #14253, opened 2020-12-17
Import acsccid

vjardin:
First version based on 1.1.8

It is based on the ccid build framework
from Daniel Golle <daniel@makrotopia.org>

Description:

### DIFF
--- a/utils/acsccid/Makefile
+++ b/utils/acsccid/Makefile
@@ -1,0 +1,54 @@
+#
+# Copyright (C) 2009-2012 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=acsccid
+PKG_VERSION:=1.1.8
+PKG_RELEASE:=1
+
+PKG_SOURCE:=v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/acshk/acsccid/archive/
+PKG_HASH:=68d15eb20e7f52153509f1dc300cf0c68b388c59d7d124ba494fd96c61a6e7c6
+PKG_MAINTAINER:=Vincent JARDIN <vjardin@free.fr>
+PKG_LICENSE:=LGPL-2.1-or-later
+PKG_LICENSE_FILES:=COPYING
+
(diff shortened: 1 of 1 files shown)


dangowrt:
Looks very good now. arc700 is failing because you probably need to add  `$(INTL_DEPENDS)` to the package dependencies to make builds based on uclibc happy.

vjardin:
> Looks very good now. arc700 is failing 

I keep trying to add ICONV or your advices regarding arc700, but still, it keeps failing:
ifdhandler.c:60:10: fatal error: iconv.h: No such file or directory
 #include <iconv.h>
          ^~~~~~~~~

I am quite new with openwrt, I do not understand how to pull the dependencies and the include path in order to fix it. Moreover, I do not understand why it fails only on this target knowing that the same include is required for the other target.

dangowrt:
> > Looks very good now. arc700 is failing
> 
> I keep trying to add ICONV or your advices regarding arc700, but still, it keeps failing:
> ifdhandler.c:60:10: fatal error: iconv.h: No such file or directory
> #include <iconv.h>
> ^~~~~~~~~
> 

You need to also add
```
include $(INCLUDE_DIR)/nls.mk
```

I don't think `$(INTL_DEPENDS)` is needed as it fails for `<iconv.h>` which is provided by `$(ICONV_DEPENDS)` once you include `nls.mk`.

> I am quite new with openwrt, I do not understand how to pull the dependencies and the include path in order to fix it. Moreover, I do not understand why it fails only on this target knowing that the same include is required for the other targets.

The main difference here is that arc700 is using uClibc instead of musl (which doesn't support that architecture) as it's standard C library.

vjardin:
> ```
> include $(INCLUDE_DIR)/nls.mk
> ```
> 
> I don't think `$(INTL_DEPENDS)` is needed as it fails for `<iconv.h>` which is provided by `$(ICONV_DEPENDS)` once you include `nls.mk`.
> 

Thanks, right, `$(ICONV_DEPENDS)` along with `nls.mk` were the solution.

It builds for all the platform including arc700 core architectures.

dangowrt:
Nice. So final step is to fold those two commits into a single one to not pollute our git history with a commit adding a package directly followed by a commit fixing it (ie. use `git rebase -i ...` and `git push -f ...`)

dangowrt:
Thank you for your contribution!

PR #14257, opened 2020-12-18
treewide: move binaries of alternatives to /usr/libexec

zhanhb:
Maintainer: 
    coreutils: @neheb 
    binutils: @neheb @yousong 
    bridge-utils: @neheb 
    openssh: @tripolar @neheb 
    wget: @tripolar @yousong 
    findutils: @neheb 
    xz: @dangowrt @neheb 
Compile tested: ath79/nand
Run tested: ath79/nand

Description:
Move binaries of alternatives to /usr/libexec, rename as suggested in https://github.com/openwrt/packages/pull/12281#issuecomment-634015071.

### DIFF
--- a/devel/diffutils/Makefile
+++ b/devel/diffutils/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=diffutils
 PKG_VERSION:=3.7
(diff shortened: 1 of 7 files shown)


jow-:
Thanks a lot @zhanhb

PR #14275, opened 2020-12-19
elektra: explicitly disable gpgme plugin

tofurky:
Maintainer: @haraldg 
Compile tested: arc, master https://github.com/openwrt/openwrt/commit/13581a5fa7ed09aa04393d496f44d0bf068e3024
Run tested: N/A

Description:
With the addition of gpgme-config per 91a6168, elektra has started
detecting the presence of gpgme and is automatically building the
plugin. Even though it's not being installed, it has broken the build on
CONFIG_ARCH="arc" due to some LFS weirdness with uClibc.

Explicitly disable it since the package in its current form doesn't
expect it to be built.

Signed-off-by: Matt Merhar <mattmerhar@protonmail.com>


aparcar (approved):

haraldg (approved):
Seems reasonable. Thanks.

PR #14284, opened 2020-12-21
ttyd: temporarily req login to mitigate non-luci lan access

wulfy23:
ttyd via luci-app is authenticated. The underlying server is not, which means anyone with a port scanner on the lan has unauthenticated root level access to the device.

Require authentication in the interim ( within luci also )... until a more elegant fix can be implemented by package maintainers.

### DIFF
--- a/utils/ttyd/files/ttyd.config
+++ b/utils/ttyd/files/ttyd.config
@@ -1,5 +1,5 @@
 
 config ttyd
 	option interface '@lan'
-	option command '/usr/libexec/login.sh'
+	option command '/bin/login'
 


tsl0922 (approved):
I'm OK with the change, but the CI build failed.

feckert:
Please update your commit message. Your lines are much too long.
Also did you sign off your work correctly?

neheb:
```
Author name (wulfy23) need to be your real name 'firstname lastname'
Commit subject line seems ok (ttyd: temp require login to mitigate no-auth to port)
Signed-off-by is missing or doesn't match author (should be 'Signed-off-by: wulfy23 <37495210+wulfy23@users.noreply.github.com>')
```

wulfy23:
thanks @neheb for your clear and helpful pr suggestions... for a single line... will leave it in the hands of maintainers...

PR #14292, opened 2020-12-21
mwan3: minor fixes

feckert:
Maintainer: me / @aaronjg 
Compile tested: no

Description:
Fix some error and add code cleanup.

### DIFF
--- a/net/mwan3/files/etc/hotplug.d/iface/16-mwan3-user
+++ b/net/mwan3/files/etc/hotplug.d/iface/16-mwan3-user
@@ -3,7 +3,7 @@
 [ -f "/etc/mwan3.user" ] && {
 	. /lib/functions.sh
 	. /lib/mwan3/mwan3.sh
-        initscript=/etc/init.d/mwan3
+	initscript=/etc/init.d/mwan3
 	. /lib/functions/procd.sh
 
 	[ "$MWAN3_SHUTDOWN" != 1 ] && procd_lock
--- a/net/mwan3/files/etc/hotplug.d/iface/16-mwan3-user
+++ b/net/mwan3/files/etc/hotplug.d/iface/16-mwan3-user
@@ -16,7 +16,6 @@
 
 	config_get_bool enabled "$INTERFACE" enabled 0
 	[ "${enabled}" -eq 1 ] || {
-		[ "$MWAN3_SHUTDOWN" != 1 ] && mwan3_unlock "$ACTION" "$DEVICE-user"
 		exit 0
 	}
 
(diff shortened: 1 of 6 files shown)


net/mwan3/files/lib/mwan3/common.sh
```diff
@@ -89,16 +89,12 @@ mwan3_get_mwan3track_status()
 	config_list_foreach "$1" track_ip mwan3_list_track_ips
 
 	if [ -n "$track_ips" ]; then
-		pid="$(pgrep -f "mwan3track $1$")"
-		if [ -n "$pid" ]; then
-			if [ "$(cat /proc/"$(pgrep -P $pid)"/cmdline)" = "sleep${MAX_SLEEP}" ]; then
-				tracking="paused"
```
aaronjg, line 95: I'd like to keep the "paused" status. The tracker enters the "paused" state when the tracker is running but the interface is down, and it is helpful to know when it is in this state.

net/mwan3/files/etc/hotplug.d/iface/16-mwan3-user
```diff
@@ -16,7 +16,6 @@
 
 	config_get_bool enabled "$INTERFACE" enabled 0
 	[ "${enabled}" -eq 1 ] || {
```
aaronjg, line 18: Can also now remove braces here

aaronjg:
small typo in commit msg for 5073817 (logger tag**e**)

net/mwan3/files/lib/mwan3/common.sh
feckert: I am referring to this line [here](https://github.com/openwrt/packages/blob/master/net/mwan3/files/usr/sbin/mwan3rtmon#L107). Should we not then also update the tables when a  tracking is `paused`?

net/mwan3/files/etc/hotplug.d/iface/16-mwan3-user
feckert: I will change this also

feckert: I looked this up again in the source of other packages.
There is no uniformity.
There are different possibilities.
I do not know what is right or wrong.
What does the shellcheck say?

net/mwan3/files/lib/mwan3/common.sh
aaronjg: Tracking should only be paused when the interface is “down” so it will have no routing table, so no need to update.

net/mwan3/files/etc/hotplug.d/iface/16-mwan3-user
aaronjg: I don't think shellcheck has an opinion on this. In general if it is one line, I prefer no braces, if it is a few lines, I like to use braces, and if it is many lines or needs an 'else', I prefer to use an 'if ; then; fi' statement. 

I don't have a strong preference, so it is fine with me to leave as is if you prefer.

PR #14299, opened 2020-12-22
nft-qos: silence buildsystem errors

wulfy23:
Maintainer: @rosysong 

add IPKG_INSTROOT to init script to
silence build system errors

### DIFF
--- a/net/nft-qos/files/nft-qos.init
+++ b/net/nft-qos/files/nft-qos.init
@@ -3,12 +3,12 @@
 # Copyright (C) 2018 rosysong@rosinson.com
 #
 
-. /lib/nft-qos/core.sh
-. /lib/nft-qos/monitor.sh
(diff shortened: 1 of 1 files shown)


feckert:
Please update PKG_RELEASE https://github.com/openwrt/packages/blob/f88485f572ec1ff9082106ccf0ccb20fc7af5801/net/nft-qos/Makefile#L12

neheb:
@feckert I don't know if it's needed. This seems like a cosmetic issue having to do with InstallDev.

Wouldn't hurt to add though.

feckert:
@neheb just meant that any change in the packages requires a bump of the PKG_RELEASE no matter how small it is. This has been pointed out to me several times, especially in the openwrt core. Unfortunately, there is not a clear announcement.

its up to you :-)

PR #14316, opened 2020-12-23
gnutls: update to 3.7.0

neheb:
Added autoreconf to fix bad linking under Arch Linux.

Added patch to fix compilation with autoreconf.

Maintainer: @nmav 
Compile tested: ath79

### DIFF
--- a/libs/gnutls/Makefile
+++ b/libs/gnutls/Makefile
@@ -8,14 +8,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gnutls
-PKG_VERSION:=3.6.15
+PKG_VERSION:=3.7.0
 PKG_RELEASE:=1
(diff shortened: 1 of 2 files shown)


neheb:
ping @NoTengoBattery

NoTengoBattery:
Thank you, I will test tomorrow in my environment and report to you as soon as I can.

nmav (approved):

nmav:
Thank you for taking the time. The change in the m4 macros is ugly but I could not see any better solution either.

NoTengoBattery:
> Thank you for taking the time.

Thanks to you. This seems to fix the libnettle issue, but the issue with other packages still present. Anyway, applying a small patch to them may solve the issue locally, I would like to see it fixed upstream. This is beyond your scope, thought, but it will help the other maintainers to pinpoint and fix the issue.

PR #14320, opened 2020-12-23
adblock: fix when PROCD_RELOAD_DELAY is set

tarruda:
Maintainer: @dibdot 
Run tested: (ar71xx, WZR-600DHP, OpenWrt 19.07.5)

Description:

PROCD_RELOAD_DELAY was being set before "delay", so the "adb_triggedelay" option was having no effect.


dibdot:
Thanks, clearly a bug. 
Please bump the PKG_RELEASE in the Makefile as well.

dibdot:
Could you please squash the two commits? Then we're are good to go ... ;-)

Many thanks for your contribution.

PR #14329, opened 2020-12-24
host-pip-requirements: add Babel

BKPepe:
Python maintainers: @jefferyto and @commodo 
Compile tested: Turris MOX, mvebu/cortexa53 & Turris Omnia, arm_cortex-a9_vfpv3-d16 
Description:

One package (reforis-distutils [1]), which is not in this repo and it is required for Turris routers
because of reForis (simplified user interface), which is being developed
depends on Babel host package [2].

It fails with:
`ERROR: Could not open requirements file: [Errno 2] No such file or directory: '/foo-omnia/build/feeds/packages/lang/python/host-pip-requirements/Babel.txt'
`
I tried to remove HOST_PYTHON3_PACKAGE_BUILD_DEPENDS (as Babel is
dependency for the package), then
reforis-distutils is compiled and package is created, but thing is that
if I want to try compile package reForis [3] then it fails with
following output:

```
ERROR: Could not find a version that satisfies the requirement reforis_distutils (from versions: none)
ERROR: No matching distribution found for reforis_distutils
  File "/foo/mox/build/staging_dir/hostpkg/lib/python3.8/site-packages/setuptools/installer.py", line 128, in fetch_build_egg
    subprocess.check_call(cmd)
  File "/foo/mox/build/staging_dir/target-aarch64_cortex-a53_musl/usr/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/foo/mox/build/staging_dir/hostpkg/bin/python3.8', '-m', 'pip', '--disable-pip-version-check', 'wheel', '--no-deps', '-w', '/foo/mox/build/tmp/tmp3s9rxjym', '--quiet', '--find-links', 'git+https://gitlab.nic.cz/turris/reforis/reforis-distutils.git#egg=reforis-distutils', 'reforis_distutils']' returned non-zero exit status 1.
```

No matter if I add python3-babel into reForis dependency.
This was working on OpenWrt 19.07, but there is regression in OpenWrt
master, which I fixed as packages reForis-distutils and reForis are
compiled successfully with this commit.

[1] https://gitlab.nic.cz/turris/turris-os-packages/-/blob/v5.1.4/web/reforis/reforis-distutils/Makefile#L23
[2] https://gitlab.nic.cz/turris/reforis/reforis-distutils/-/blob/v0.1.0/setup.py#L13
[3] https://gitlab.nic.cz/turris/turris-os-packages/-/blob/v5.1.4/web/reforis/reforis/Makefile

_____

I hope that I described my issue clear.

OT: Guys, Merry Christmas!


BKPepe:
Simplified commit message as I removed useless sentences. They were not related.

jefferyto:
Sorry but I don't see why this file should be in this repo when there are no packages here that need it.

BKPepe:
It is just a text file, which should not break anything, but I got your point. There can be additional packages, which I am not aware right now, which can use this.

BKPepe:
For me, it's better to see it here instead of forking this repository or even having patch forever against this repository like I did with reforis-distutils ref: https://gitlab.nic.cz/turris/turris-build/-/commit/3c402c93e2a74760707999bc029ca0eab3cb1dcb

On the other hand, I can add some package, which depends on Babel, but the question would be if I have any usage for it. I thought that Babel should be OK instead of submitting some Turris specific package.

jefferyto:
If there are packages here that require Babel to be installed for host, then including this file would be fine. But currently there are none, so there is no benefit (for the packages in this repo) to include it.

I realize it is only one file, but when we include something here, users who have custom packages or downstream projects (like Turris) will have a (somewhat reasonable) expectation that it will work and be supported/updated/etc. There is a maintenance cost, however small it is. (Not to mention that currently these requirements files are in a no-man's land of maintenance responsibility.) This is why I am not in favour of accepting this PR (no benefit, non-zero cost).

However, I have an alternate suggestion for you: I didn't mention this in the readme or the commit message for the change, but [if the file name in `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` contains a `/`](https://github.com/openwrt/packages/blob/3416108451741d8067dc6de45a46dde6c6c295d7/lang/python/python3-package.mk#L200), then instead of looking up that file in the host-pip-requirements directory, the string is interpreted as a file path.

So you should be able to do:

```make
HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=./Babel.txt
```

and have Babel.txt in the package directory. (I'm reasonable sure relative paths are relative to the package directory, though I'm not 100% sure I tested this or what the results were. Let me know if it isn't relative to the package directory.)

BKPepe:
That's sounds like a solution. I will test it during upcoming days and will get back to you. Thank you. 👍

PR #14335, opened 2020-12-25
collectd: move own plugins to seperate files

PolynomialDivision:
Patchfiles can be hard to read and maintain. Some plugins seem not
to get upstreamed. Move the patchfiles into code + patchfile.

Plugins
- 900-add-iwinfo-plugin.patch
	Move iwinfo.c into src/src/iwinfo.c
- 930-dhcpleases-add-dhcpleases-plugin.patch
	Move dhcpleases.c into src/src/dhcpleases.c

Maintainer: @hnyman
Compile tested: trunk
Run tested: tbd


hnyman:
@jow-   
Any objections?

This PR proposes to ease maintenance by changing the two OpenWrt-specific collectd plugins (iwinfo, dhcpleases) to be provided via normal source files in `src/` instead of being created with the current "patch for collectd" approach.  

The end outcome is identical, as Build/Prepare/Default first unpacks upstream sources, then copies the possible src/ contents and finally applies the possible patches.
https://github.com/openwrt/openwrt/blob/6f96a4d043a9367c6c0d166299d808df764e88e6/include/package-defaults.mk#L67-L70

I do not remember seeing this kind of mixed upstream source + local src/ approach in any package(?), so I am asking for style confirmation...

A possible negative effect from this change might come if upstreaming of one such plugin is unnoticed, and we continue to overwrite it with our local plugin. Otherwise, defining the plugins as normal source files would be more straightforward

PolynomialDivision:
Very good points. The background was that I wanted to add the channel utilization. Then I thought, it can be better understood in the commit history if we had separate files.

I am also not sure now, anymore. xD
If u are against it, feel free to close the PR.

hnyman:
@jow- 
Any views to this idea?

Simplifying the add-on plugin maintenance by converting them from "create them via patch" to "provide them as add-on files to upstream sources" ?

I do not remember seeing this kind of mixed upstream source + local src/ approach in any package(?), so this would be the first, I think.  (collectd is a rare item, as it natively offers extendability via side-by-side plugins defined in source files.) 

Somehow the old patch approach seems more logical in the long run:
* "upstream sources + our patches" vs.
* "upstream sources + our add-on sources + our patches"

PolynomialDivision:
Let just close this PR. ;)

PR #14336, opened 2020-12-25
add SpamAssassin packages

dangowrt:
Maintainer: me / @Naoir
Compile tested: x86_64
Run tested: x86_64 (spamc only)

Description:
Add SpamAssassin packages.
Mostly in order to have spamc available, but also as a first step towards actually running fully-featured SpamAssassin on OpenWrt one day (for that still many more Perl packages would need to be added).

### DIFF
--- a/lang/perl-mail-spamassassin/Makefile
+++ b/lang/perl-mail-spamassassin/Makefile
@@ -1,0 +1,101 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=perl-mail-spamassassin
+PKG_RELEASE:=1
+PKG_VERSION:=3.4.4
+PKG_HASH:=8ea27a165b81e3ce8c84ae85c3ecba1f2edfa04ef4a86f07fe28ab612fc8ff60
+
+PKG_SOURCE_NAME:=Mail-SpamAssassin
+PKG_SOURCE_URL:=@APACHE/spamassassin/source
+PKG_SOURCE:=$(PKG_SOURCE_NAME)-$(PKG_VERSION).tar.gz
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(BUILD_VARIANT)/$(PKG_SOURCE_NAME)-$(PKG_VERSION)
+PKG_MAINTAINER:=Daniel Golle <daniel@makrotopia.org>
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE
+PKG_CPE_ID:=cpe:/a:apache:spamassassin
+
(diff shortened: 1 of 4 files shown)


neheb:
Ping @pprindeville

lang/perl-netaddr-ip/Makefile
```diff
@@ -0,0 +1,62 @@
...
+define Package/perl-netaddr-ip
+  SUBMENU:=Perl
+  SECTION:=lang
+  CATEGORY:=Languages
+  TITLE:=NetAddr::IP - Manages IPv4 and IPv6 addresses and subnets
+  URL:=http://search.cpan.org/dist/NetAddr::IP/
+  DEPENDS:=perl +perlbase-essential +perlbase-test
+endef
+
+define Host/Configure
```
pprindeville, line 33: Wondering if we need a short-hand for this, or even make it a default that one needs to explicitly disable with `DISABLE_PERL_HOSTBUILD:=1` (or some such)?

lang/perl-mail-spamassassin/Makefile
```diff
@@ -0,0 +1,101 @@
...
+endif
+
+define Package/spamassassin/conffiles
+/etc/mail/spamassassin
+endef
+
+define Build/Configure
+	$(call perlmod/Configure,,)
+	$(call Build/Configure/Default)
+	( cd "$(PKG_BUILD_DIR)/$(CONFIGURE_PATH)" ; ./version.h.pl --with-version=$(PKG_SOURCE_VERSION) )
```
pprindeville, line 70: `;` or `&&`?

pprindeville (changes requested):
Couple of minor comments/questions...

lang/perl-netaddr-ip/Makefile
dangowrt: I'm not really into that at this point, but yes, that'd be nice (making host-build the default for Perl libraries).

lang/perl-mail-spamassassin/Makefile
dangowrt: Good catch. Should be `&&`, I'll fix that.

lang/perl-netaddr-ip/Makefile
pprindeville: Got most of a tentative PR done, I'm just trying to figure out a clean way to have the `$(eval $(call HostBuild))` happen if `PERL_HOSTBUILD` hasn't been explicitly set to `0` before including `../perl/perlmod.mk`.  Any ideas?

dangowrt: I think it'd be ok to require that `PERL_HOSTBUILD:=0` needs to be set before `include ../perl/perlmod.mk`.

PR #14338, opened 2020-12-26
exim: add package

dangowrt:
Add Exim MTA variants:
 * `exim`
   plain variant without any TLS library which hence comes without TLS, DANE and DKIM.
 * `exim-openssl`
   linked against libopenssl
 * `exim-gnutls`
   linked against libgnutls
 * `exim-ldap`
   linked against libopenssl, libopenldap and libsasl2

Provide packages for lookup modules
 * `cdb`
 * `dbmdb`
 * `dnsdb`
 * `json` (depends on `jansson`)
 * `mysql` (depends on `libmariadb`)
 * `passwd`
 * `pgsql` (depends on `libpq`)
 * `redis` (depends on `libhiredis`)
 * `sqlite` (depends on `libsqlite3`)

Note:

As uClibc doesn't provide RES_USE_DNSSEC and hence lacks support for using DNSSec, DANE is disabled when building against uClibc.

As gnutls requires libunbound which depends on libopenssl to provide libgnutls-dane, also disable DANE by default when building with gnutls.

--

Cross-compiling Exim is a hell.

### DIFF
--- a/libs/gnutls/Makefile
+++ b/libs/gnutls/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gnutls
 PKG_VERSION:=3.7.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_USE_MIPS16:=0
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
@@ -37,6 +37,7 @@ PKG_CONFIG_DEPENDS:= \
 	CONFIG_GNUTLS_SRP \
 	CONFIG_GNUTLS_TPM \
 	CONFIG_LIBNETTLE_MINI \
+	CONFIG_PACKAGE_libgnutls-dane \
 
 include $(INCLUDE_DIR)/package.mk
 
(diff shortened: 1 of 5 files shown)


neheb:
uClibc-ng is gone in master.

libs/gnutls/Makefile
```diff
@@ -178,16 +183,20 @@ ifeq ($(CONFIG_GNUTLS_CRYPTODEV),y)
 CONFIGURE_ARGS += --enable-cryptodev
 endif
```
neheb, line 185: You can make this a one liner. $( if (x,,). This can keep it part of the original CONFIGURE_ARGS.

dangowrt:
Apparently SDK used for arc700 in the Github CI still uses uClibc. That's the whole reason why I went out to fix it, so that the CI shows "all green"...

neheb:
That SDK should switch to building with glibc once http://patchwork.ozlabs.org/project/openwrt/patch/20201225003616.11895-1-rosenp@gmail.com/ is merged.

dangowrt:
@neheb I've removed all remains of uClibc distinction, ie. arc CI build should succeed once we either switch it to archs or https://patchwork.ozlabs.org/project/openwrt/patch/20201225003616.11895-1-rosenp@gmail.com/ gets merged.

PR #14342, opened 2020-12-26
transmission: do not use wolfSSL

neheb:
wolfSSL is broken at runtime. It seems to be some error with the random
number generator. It's not currently known what the reason is.

Upstream issue: transmission/transmission#1537

Maintainer: me

### DIFF
--- a/net/transmission/Makefile
+++ b/net/transmission/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=transmission
 PKG_VERSION:=3.00
-PKG_RELEASE:=9
+PKG_RELEASE:=10
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@GITHUB/transmission/transmission-releases/master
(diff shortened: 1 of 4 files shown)


BKPepe:
Can you please be more specific in the commit message? It would be helpful to know what is going on and after approx week you won't know why you do such changes like this.

neheb:
@BKPepe I assume this is better?

neheb:
I think I know what the issue is. The code uses some global RNG context. I don't feel confident that I can fix it but we'll see.

neheb:
@dangowrt can you test this again? I'm getting some kind of failure but I can't pinpoint what it is.

```
(gdb) bt
#0  memset (dest=0x77ff4ca8, c=0, n=1056) at src/string/memset.c:16
#1  0x77c77670 in mp_init_multi () from /usr/lib/libwolfssl.so.24
Backtrace stopped: frame did not save the PC
```

dangowrt:
> @dangowrt can you test this again? I'm getting some kind of failure but I can't pinpoint what it is.

@neheb Doesn't look good... Still crashes and yes, it's weird.
I tried:
```
transmission-remote localhost --add http://cdimage.debian.org/debian-cd/current/armel/bt-cd/debian-10.7.0-armel-netinst.iso.torrent
```
with `option encryption '2'` that leads to instant crash.

> 
> ```
> (gdb) bt
> #0  memset (dest=0x77ff4ca8, c=0, n=1056) at src/string/memset.c:16
> #1  0x77c77670 in mp_init_multi () from /usr/lib/libwolfssl.so.24
> Backtrace stopped: frame did not save the PC
> ```

PR #14346, opened 2020-12-27
CI: test x86/64 19.07.5 SDK

aparcar:
Not only use snapshot SDKs for building but also stable releases.


champtar:
NAK, master test against master, 19.07 against 19.07, this will just get us more failures for no benefits

aparcar:
@champtar good thinking

PR #14359, opened 2020-12-28
CI: use SDK depending on destination branch

aparcar:
Not only use snapshot SDKs for master test but whatever destination
branch is used.

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -2,8 +2,6 @@ name: Test Build
 
 on:
   pull_request:
-    branches:
-      - master
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


champtar:
Thanks for this @apacar. Only problem I see is if/when arch are renamed, or new arch are introduced in master by not present in older branches, this will cause breakage.

BKPepe:
Also, this could be a little bit problematic if you are not using Github Actions just for pull request, but for push as well. Then it fails with: `manifest for openwrtorg/sdk:mips_24kc- not found: manifest unknown: manifest unknown`. I get it working while using hardcoded version of OpenWrt, I know hardcoding is bad, but for my use case, it suits well.  On the other hand, someone with commit access is still comitting it directly to this repo to avoid hassle.

aparcar:
@champtar I hoped that while target names may change, architectures would stay the same - maybe I assumed that wrong. We can extend the current logic to use something like "master only", similar to what's done with "runtime tests".

@BKPepe do you plan to run these CI tests on every push? It may makes sense for other projects that maintain their own package repositories but as we use the buildbot that doesn't seem of use for us. If we want a *staging* repository that builds after each run, we can just define another "workflow" that does only that.

BKPepe:
@aparcar My idea was rather that in the `packages` repository we would know that every package is compile tested (and maybe run tested, too). This would prevent that there is no package that was not tested by CI. Thus prevent anybody avoids that by pushing changes directly.

aparcar:
We can just disable direct pushes which we could discuss in an external issue.

aparcar:
I think we can just backport the current CI script and modify whatever targets/archs we want to change there. I initially thought we could define the CI on the master branch and all other branches inherit the CI script, but that doesn't seem to work (at least on my fork).

PR #14366, opened 2020-12-30
smartsnmpd: remove dysfunctional package

bmork:
This package has been completely broken on several levels since
commit eadd5abe40a0 ("smartsnmpd: Update to 2015-02-22 version")

The update changed the configuration file syntax without fixing
the scripts generating this file.  The OpenWrt package would
therefore fail to run.

Worse than that, the last upstream version is completely broken
is unable to generate proper snmp replies even if the
configuration file is fixed, This has been tested and verified
on both OpenWrt/realtek and Debian/amd64.

I considered fixing the package, but dropped it for the following
reasons
 1) upstream has abandoned the project
 2) the upstream fork "smithsnmp" continues to make aribitrary
    changes to configuration files and other packaging challenges
 3) the package is tied to the lua5.1 C API, and further upgrades
    will be non-trivial
 4) there are several other snmp daemons available in OpenWrt,
    without any of these issues
 5) no one is interested in this package

The last point is proven by the lack of testing and feedback.
The last update was not even tested by the person preparing and
commiting the update.

Signed-off-by: Bjørn Mork <bjorn@mork.no>


neheb:
Hilarious.

Please also open a PR at packages-abandoned.

karlp:
agreed.  I've looked at it once or twice, as it's a really interesting project, but it's clearly dead upstream, and has been for years.

bmork:
> Please also open a PR at packages-abandoned.

really?  What's the point? There is no working package at this point, and has not been for some time.  We could rewrite history by pushing the last functional version to  packages-abandoned.  But that won't built with the current toolchain, so it will not be a currently working snapshot either.

Just to be clear: I won't put any effort into fixing or recreating this package.  I noted that it was dead, and recommend that you delete it instead of pushing untested updates.  But if you won't delete it because you failed to move it to packages-abandoned when that still was an option, then that's fine with me

PR #14370, opened 2020-12-30
libpqxx: adding Makefile, Config.in and patch

garikello3d:
Signed-off-by: Igor Bezzubchenko <garikello@gmail.com>

Maintainer: me / @garikello3d

Compile tested - based on OpenWRT revision 7ba2f5c96f2697e8 (28 Nov 2020)
- QEMU x86_64
- QEMU ARM Cortex-a15 (_-M virt_)
- QEMU MIPS (_-M malta_)

Run tested: the same as compile

Description: I'm adding the official C++ Postgres Client library (official [web](http://pqxx.org/development/libpqxx/l)).

My build and functional test includes successful run of full test suite available in the project against a host-based PostgreSQL server.


libs/libpqxx/Makefile
```diff
@@ -0,0 +1,53 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libpqxx
+PKG_VERSION:=7.3.0
+PKG_RELEASE:=1
+PKG_INSTALL:=1
```
neheb, line 6: cmake.mk makes this default. Remove.

libs/libpqxx/Makefile
```diff
@@ -0,0 +1,53 @@
...
+	SUBMENU:=Database
+endef
+
+define Package/libpqxx/config
+	source "$(SOURCE)/Config.in"
+endef
+
+CMAKE_OPTIONS += \
+	-DBUILD_DOC=off \
+	$(if $(CONFIG_LIBPQXX_STATIC), -DBUILD_STATIC_LIBS=on) \
```
neheb, line 35: I'd rather see something like

-DBUILD_STATIC_LIBS=O$(if $(CONFIG_LIBPQXX_STATIC),N, FF)

same as below.

libs/libpqxx/Makefile
```diff
@@ -0,0 +1,53 @@
...
+	URL:=http://pqxx.org/development/libpqxx
+	SUBMENU:=Database
+endef
+
+define Package/libpqxx/config
+	source "$(SOURCE)/Config.in"
+endef
+
+CMAKE_OPTIONS += \
+	-DBUILD_DOC=off \
```
neheb, line 34: Please capitalize options. That's how it's done elsewhere.

libs/libpqxx/Makefile
```diff
@@ -0,0 +1,53 @@
...
+	$(if $(CONFIG_LIBPQXX_STATIC), -DBUILD_STATIC_LIBS=on) \
+	$(if $(CONFIG_LIBPQXX_SHARED), -DBUILD_SHARED_LIBS=on)
+
+ifeq ($(CONFIG_LIBPQXX_INSTALL_TEST),y)
```
neheb, line 38: There's no reason for this. Just make them a part of the main CMAKE_OPTIONS.

libs/libpqxx/Makefile
```diff
@@ -0,0 +1,53 @@
...
+PKG_VERSION:=7.3.0
+PKG_RELEASE:=1
+PKG_INSTALL:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/jtv/libpqxx
+PKG_SOURCE_VERSION:=$(PKG_VERSION)
+PKG_MIRROR_HASH:=c176771dbbf02e38b84a8f9bf986a83587e2d86d9e0d2238e2a17a036e865951
+
+CMAKE_INSTALL:=1
```
neheb, line 13: One issue I've had with CMAKE_INSTALL in the past is that the pkgconfig paths are totally wrong. See https://github.com/openwrt/packages/blob/master/libs/pugixml/Makefile#L45 for an example.

Please check that libdir and includedir are correct.

libs/libpqxx/Makefile
garikello3d: Fixed

libs/libpqxx/Makefile
garikello3d: Fixed

libs/libpqxx/Makefile
garikello3d: Fixed

libs/libpqxx/Makefile
garikello3d: Fixed

libs/libpqxx/Makefile
garikello3d: I checked the **libpqxx.pc** file generated.

> prefix=/usr
> exec_prefix=${prefix}
> libdir=${prefix}/lib
> includedir=${prefix}/include

So libdir and include dir are parameterized by _prefix_, nothing bad that I see...

PR #14372, opened 2020-12-30
CI: switch back to ARCHS

neheb:
Upstream wants to remove ARC700. That and currently, ARC700 is not
compiling with glibc.

Maintainer: @aparcar


aparcar:
I'll wait for the official announcement for removing arc700

neheb:
I mean...the current situation is that arc700 still gets built with uClibc-ng despite it not being in the tree any longer.

PR #14374, opened 2020-12-31
ksmbd-tools: fix compilation with glibc

neheb:
Needs to be linked to lpthread.

Maintainer: @Andy2244 
Compile tested: arc700


neheb:
This seems to be a consequence of linking glib2 statically. ksmbd-tools itself does not use pthreads.

PR #14377, opened 2020-12-31
bfdd: fix compilation with glibc

neheb:
Remove uClibc-ng patch as it's not in the tree anymore.

Also remove the _GNU_SOURCE CFLAG for the same reason.

Refreshed patches.

Maintainer: @lucize 
Compile tested: arc700


lucize:
LGTM

PR #14384, opened 2020-12-31
postfix: update to 3.5.8

neheb:
Pass TARGET_[C/LD]FLAGS to reduce filesize.

Fix glibc compilation by adding lresolv.

Remove uClibc-ng patch as that's now gone.

Refreshed other patches.

Maintainer: @Shulyaka 
Compile tested: arc700


Shulyaka (approved):
Thank you!

PR #14386, opened 2020-12-31
opendkim: fix compilation with glibc

neheb:
Add libbsd dependency.

Maintainer: @val-kulkov 
Compile tested: arc700


val-kulkov:
LGTM.

PR #14394, opened 2020-12-31
freeradius3: add missing libbsd dependency to glibc

neheb:
Fixes packages generation.

Maintainer: nobody
Compile tested: arc700


neheb:
```
The distutils module is unusable: install "python-devel"?
```

Right...

PR #14410, opened 2021-01-02
Keepalived won't start without enable_script_security option

tkteun:

Description:

### DIFF
--- a/net/keepalived/files/keepalived.init
+++ b/net/keepalived/files/keepalived.init
@@ -111,7 +111,8 @@ globals() {
 		router_id \
 		vrrp_mcast_group4 \
 		vrrp_mcast_group6 \
-		vrrp_startup_delay
+		vrrp_startup_delay \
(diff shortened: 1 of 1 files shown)


feckert:
We already add `enable_script_security` on every service start to the keepalived [configuration](https://github.com/openwrt/packages/blob/master/net/keepalived/files/keepalived.init#L102). So this is not needed.

tkteun:
OK, I can confirm this does not work on 19.07.5

feckert:
Now I understand. You are referring here to the last stable branch openwrt-19.07.
We should then backport the following commit to openwrt-19.07
https://github.com/openwrt/packages/commit/1b1ba710632f8f4850f02c22e83734a6c8b5c41a#diff-96375c71e80146e6de3ce6a9adf3c8cdadcd21daf4a834ca6df4659ea6d23c4f

PR #14413, opened 2021-01-03
mwan3: fix regression in ipv6 routing tables

aaronjg:

Maintainer: me, @feckert 
Compile tested: 19.07.5/ openwrt/openwrt@55e23f2c
Run tested: 19.07.5/ openwrt/openwrt@55e23f2c

Description: 
Introduced a bug when consolidating the route handling SED line, where it did not check config variable and broke default routes for IPv6 setups. This commit fixes that bug.

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.10.5
+PKG_VERSION:=2.10.6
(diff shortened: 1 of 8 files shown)


champtar:
I see this PR is against master, but tested against 19.07, is this correct ?

aaronjg:
> I see this PR is against master, but tested against 19.07, is this correct ?

This didn't touch any of the code that was sensitive to the 5.4.x kernel, so I did the testing on 19.07. I have since tested against the master branch and it works as expected.

PR #14415, opened 2021-01-03
openssh: Add FIDO2 hardware token support

linosgian:
Maintainer: me
Compile tested: x86-64/generic/snapshot(c5d033a34d)
Run tested: x86-64/generic/snapshot(c5d033a34d) - Tested this end-to-end with a hardware token for the `ecdsa-sk` type

Description:

Version [8.2](https://lwn.net/Articles/812537/)  added support for two new key types: "ecdsa-sk" and
"ed25519-sk". These two types enable the usage of hardware tokens that
implement the FIDO (or FIDO2) standard, as an authentication method for
SSH.

Since we're already on version 8.4 all we need to do is to explicitly enable
the support for hardware keys when compiling OpenSSH and add all the
missing dependencies OpenSSH requires.

OpenSSH depends on [libfido2](https://github.com/Yubico/libfido2), to communicate with the FIDO devices
over USB. In turn, libfido2 depends on libcbor, a C implementation of
the [CBOR protocol](tools.ietf.org/html/rfc7049) and OpenSSL.

I've also compiled openssh-server for 19.07.04 (ramips/mt7621) and I've been using the new `ecdsa-sk` type for a while now.

### DIFF
--- a/libs/libcbor/Makefile
+++ b/libs/libcbor/Makefile
@@ -1,0 +1,47 @@
+#
+# Copyright (C) 2020 Linos Giannopoulos
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libcbor
+PKG_VERSION:=0.8.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/PJK/libcbor/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=618097166ea4a54499646998ccaa949a5816e6a665cf1d6df383690895217c8b
+
+PKG_LICENSE:=GPL-3.0-or-later
+PKG_LICENSE_FILES:=COPYING
+PKG_MAINTAINER:=Linos Giannopoulos <linosgian00+openwrt@gmail.com>
+
+CMAKE_OPTIONS += \
+        -DBUILD_SHARED_LIBS=ON
+CMAKE_INSTALL:=1
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+define Package/libcbor
+  SECTION:=libs
+  CATEGORY:=Libraries
+  TITLE:=libcbor
+  URL:=https://github.com/PJK/libcbor
+  ABI_VERSION:=0
+endef
+
+define Package/libcbor/description
+ libcbor is a C library for parsing and generating CBOR, the general-purpose schema-less binary data format.
+endef
+
+
+define Package/libcbor/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libcbor.so.* $(1)/usr/lib/
+endef
+
+$(eval $(call BuildPackage,libcbor))
--- a/libs/libfido2/Makefile
+++ b/libs/libfido2/Makefile
@@ -1,0 +1,51 @@
+#
+# Copyright (C) 2020 Linos Giannopoulos
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libfido2
+PKG_VERSION:=1.6.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/Yubico/libfido2/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=6aed47aafd22be49c38f9281fb88ccd08c98678d9b8c39cdc87d1bb3ea2c63e4
+
+PKG_FORTIFY_SOURCE:=0
+CMAKE_INSTALL:=1
+
+TARGET_CFLAGS += -Wno-error=overflow -Wno-error=sign-conversion
+
+PKG_MAINTAINER:=Linos Giannopoulos <linosgian00+openwrt@gmail.com>
+PKG_LICENSE:=GPL-3.0-or-later
+PKG_LICENSE_FILES:=COPYING
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+define Package/libfido2
+  SECTION:=libs
+  CATEGORY:=Libraries
+  TITLE:=FIDO2 Library
+  URL:=https://github.com/Yubico/libfido2
+  ABI_VERSION:=1
+  DEPENDS += +libcbor +libopenssl +libudev
+endef
+
+define Package/libfido2/description
+ libfido2 provides library functionality and command-line tools to communicate with a FIDO device over USB, and to verify attestation and assertion signatures.
+
+ libfido2 supports the FIDO U2F (CTAP 1) and FIDO 2.0 (CTAP 2) protocols.
+endef
+
+
+define Package/libfido2/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libfido2.so.* $(1)/usr/lib/
+endef
+
+$(eval $(call BuildPackage,libfido2))
--- a/net/openssh/Makefile
+++ b/net/openssh/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openssh
 PKG_VERSION:=8.4p1
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/ \
@@ -21,6 +21,10 @@ PKG_LICENSE_FILES:=LICENCE
 PKG_CPE_ID:=cpe:/a:openssh:openssh
 
 PKG_REMOVE_FILES:=
+PKG_CONFIG_DEPENDS := \
+	CONFIG_OPENSSH_LIBFIDO2
+
+PKG_BUILD_DEPENDS += OPENSSH_LIBFIDO2:libfido2
 
 include $(INCLUDE_DIR)/package.mk
 
@@ -82,9 +86,13 @@ endef
 
 define Package/openssh-server
 	$(call Package/openssh/Default)
-	DEPENDS+= +openssh-keygen
+	DEPENDS+= +openssh-keygen +OPENSSH_LIBFIDO2:libfido2
 	TITLE+= server
 	USERID:=sshd=22:sshd=22
+endef
+
+define Package/openssh-server/config
+	source "$(SOURCE)/Config.in"
 endef
 
 define Package/openssh-server/description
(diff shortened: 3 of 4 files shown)


linosgian:
cc'ing @tripolar as he's the maintainer of the openssh package

linosgian:
It seems like the failing builds are due to `musl` being used instead of glibc.

libfido2 includes `#include <linux/hidraw.h>` which defines the [constants](https://github.com/Yubico/libfido2/blob/7032a63c1e0eaa76e57d1eb7a5c85cfb65d446f3/src/hid_linux.c#L33) that are passed to `ioctl` as `unsigned int`s.
Glibc defines `ioctl` as `ioctl (int fd, unsigned long int request, ...)` whereas `musl` defines it as: [`ioctl(int fd, int request)`](https://git.musl-libc.org/cgit/musl/tree/include/sys/ioctl.h?h=v1.1.21&id=1691b23955590d1eb66a11158fdd91c86337e886#n9), hence the implicit conversion.

I am not sure what would be the "correct" solution to this issue.. Apart from `-Wno-sign-conversion`

neheb:
Commit subject line MUST start with '<package name>: ' (fixup! openssh: Add FIDO2 hardware token support)

Remove fix up!

neheb:
Alright...

Please split up the commits for each package added/edited.

The OpenSSH change requires a version bump of PKG_RELEASE.

net/openssh/Makefile
```diff
@@ -82,7 +82,7 @@ endef
 
 define Package/openssh-server
 	$(call Package/openssh/Default)
-	DEPENDS+= +openssh-keygen
+	DEPENDS+= +openssh-keygen +libfido2
```
neheb, line 85: I would change this to be en/disabled at compile time. Look at packages that use Config.in

libs/libfido2/Makefile
```diff
@@ -0,0 +1,52 @@
...
+  DEPENDS += +libcbor +libopenssl +libudev
+endef
+
+define Package/libfido2/description
+ libfido2 provides library functionality and command-line tools to communicate with a FIDO device over USB, and to verify attestation and assertion signatures.
+
+ libfido2 supports the FIDO U2F (CTAP 1) and FIDO 2.0 (CTAP 2) protocols.
+endef
+
+define Build/InstallDev
```
neheb, line 43: Pointless. Remove and replace with CMAKE_INSTALL:=1

libs/libfido2/Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+PKG_NAME:=libfido2
+PKG_VERSION:=1.6.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/Yubico/libfido2/archive/
+PKG_HASH:=6aed47aafd22be49c38f9281fb88ccd08c98678d9b8c39cdc87d1bb3ea2c63e4
+
+PKG_FORTIFY_SOURCE:=0
```
neheb, line 18: Don't bother with this.

libs/libfido2/Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+TARGET_CFLAGS += -Wno-error=overflow -Wno-error=sign-conversion
```
neheb, line 20: Just pass -Wno-error

libs/libcbor/Makefile
```diff
@@ -0,0 +1,48 @@
...
+endef
+
+define Package/libcbor/description
+ libcbor is a C library for parsing and generating CBOR, the general-purpose schema-less binary data format.
+endef
+
+CMAKE_OPTIONS += \
+        -DBUILD_SHARED_LIBS=ON
+
+define Build/InstallDev
```
neheb, line 39: Remove and replace with CMAKE_INSTALL:=1

libs/libfido2/Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+PKG_MAINTAINER:=Linos Giannopoulos <linosgian00+openwrt@gmail.com>
+PKG_LICENSE:=GPL-3.0-or-later
```
neheb, line 23: Add PKG_LICENSE_FILES

libs/libcbor/Makefile
```diff
@@ -0,0 +1,48 @@
...
+PKG_NAME:=libcbor
+PKG_VERSION:=0.8.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/PJK/libcbor/archive/v$(PKG_VERSION)
+PKG_HASH:=618097166ea4a54499646998ccaa949a5816e6a665cf1d6df383690895217c8b
+
+PKG_LICENSE:=GPL-3.0-or-later
+PKG_MAINTAINER:=Linos Giannopoulos <linosgian00+openwrt@gmail.com>
```
neheb, line 19: Add PKG_LICENSE_FILES

libs/libcbor/Makefile
```diff
@@ -0,0 +1,48 @@
...
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libcbor
+PKG_VERSION:=0.8.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/PJK/libcbor/archive/v$(PKG_VERSION)
```
neheb, line 15: I would use codeload instead.

libs/libfido2/Makefile
```diff
@@ -0,0 +1,52 @@
...
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libfido2
+PKG_VERSION:=1.6.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/Yubico/libfido2/archive/
```
neheb, line 15: I would replace with codeload.

libs/libfido2/Makefile
linosgian: I've ran into issues with this one, because [libfido2](https://github.com/Yubico/libfido2/blob/7032a63c1e0eaa76e57d1eb7a5c85cfb65d446f3/CMakeLists.txt#L245) itself sets the -DFORTIFY_SOURCE=2, then OpenWRT does as well, if I understand correctly. If you have any suggestions on how to handle this, please let me know

linosgian: The aforementioned error thrown is about redefining the same macro: `Error: 'warning: "_FORTIFY_SOURCE" redefined'`

neheb: sounds like -Wno-error should fix that?

net/openssh/Makefile
```diff
@@ -165,6 +173,12 @@ CONFIGURE_ARGS += \
 	--without-kerberos5 \
 	--with-stackprotect \
 	--with$(if $(CONFIG_OPENSSL_ENGINE),,out)-ssl-engine
+	
+
+ifdef CONFIG_OPENSSH_LIBFIDO2
```
neheb, line 178: A similar trick can be applied as with CONFIG_OPENSSL_ENGINE above to reduce this to one line.

libs/libfido2/Makefile
linosgian: Unfortunately it doesn't. I've just tried it and it still fails when I remove `PKG_FORTIFY_SOURCE:=0`

net/openssh/Makefile
linosgian: Pushed a fix for this. I guess I should I add this to the `openssh-server-pam` variant as well?

neheb: If you want.

libs/libfido2/Makefile
linosgian: This seems to get overridden by explicit `-Werror=foo` flags (e.g. -Werror=overflow).
It currently breaks all the PR checks. 

An excerpt from a failing build:
```
cd /home/build/openwrt/build_dir/target-aarch64_cortex-a53_musl/libfido2-1.6.0/src && /home/build/openwrt/staging_dir/toolchain-aarch64_cortex-a53_gcc-8.4.0_musl/bin/aarch64-openwrt-linux-musl-gcc -DHAVE_CLOCK_GETTIME -DHAVE_DEV_URANDOM -DHAVE_ENDIAN_H -DHAVE_ERR_H -DHAVE_EXPLICIT_BZERO -DHAVE_GETLINE -DHAVE_GETOPT -DHAVE_GETPAGESIZE -DHAVE_GETRANDOM -DHAVE_SIGNAL_H -DHAVE_STRLCAT -DHAVE_STRLCPY -DHAVE_SYSCONF -DHAVE_SYS_RANDOM_H -DHAVE_UNISTD_H -DSIGNAL_EXAMPLE -DTLS=__thread -D_DEFAULT_SOURCE -D_FIDO_INTERNAL -D_FIDO_MAJOR=1 -D_FIDO_MINOR=6 -D_FIDO_PATCH=0 -D_GNU_SOURCE -I/home/build/openwrt/build_dir/target-aarch64_cortex-a53_musl/libfido2-1.6.0/src -Os -pipe -mcpu=cortex-a53 -fno-caller-saves -fno-plt -fhonour-copts -Wno-error=unused-but-set-variable -Wno-error=unused-result -Wno-error -ffile-prefix-map=/home/build/openwrt/build_dir/target-aarch64_cortex-a53_musl/libfido2-1.6.0=libfido2-1.6.0 -Wformat -Werror=format-security -fstack-protector -Wl,-z,now -Wl,-z,relro -Wall -Wextra -Werror -Wshadow -Wwrite-strings -Wmissing-prototypes -Wbad-function-cast -pedantic -pedantic-errors -fstack-protector-all -std=c99 -Wno-unused-result -Wcast-qual -DNDEBUG -D_FORTIFY_SOURCE=2 -fPIC -Wconversion -Wsign-conversion -o CMakeFiles/fido2.dir/hid_linux.c.o -c /home/build/openwrt/build_dir/target-aarch64_cortex-a53_musl/libfido2-1.6.0/src/hid_linux.c
/home/build/openwrt/build_dir/target-aarch64_cortex-a53_musl/libfido2-1.6.0/src/hid_linux.c: In function 'get_report_descriptor':
/home/build/openwrt/build_dir/target-aarch64_cortex-a53_musl/libfido2-1.6.0/src/hid_linux.c:30:16: error: overflow in conversion from 'long unsigned int' to 'int' changes value from '2147764225' to '-2147203071' [-Werror=overflow]
  if (ioctl(fd, HIDIOCGRDESCSIZE, &s) < 0 || s < 0 ||
                ^~~~~~~~~~~~~~~~
/home/build/openwrt/build_dir/target-aarch64_cortex-a53_musl/libfido2-1.6.0/src/hid_linux.c:38:16: error: overflow in conversion from 'long unsigned int' to 'int' changes value from '2416199682' to '-1878767614' [-Werror=overflow]
  if (ioctl(fd, HIDIOCGRDESC, hrd) < 0) {
                ^~~~~~~~~~~~
cc1: all warnings being treated as errors
```

neheb: I looked into this briefly. This is a bug and should be fixed by not including sys/ioctl.h

neheb: I reported upstream https://github.com/Yubico/libfido2/issues/259

Patch is included if you want to port it here.

net/openssh/Makefile
```diff
@@ -164,8 +172,9 @@ CONFIGURE_ARGS += \
 	--without-bsd-auth \
 	--without-kerberos5 \
 	--with-stackprotect \
-	--with$(if $(CONFIG_OPENSSL_ENGINE),,out)-ssl-engine
-
+	--with$(if $(CONFIG_OPENSSL_ENGINE),,out)-ssl-engine \
```
neheb, line 175: I find it interesting that CONFIG_OPENSSL_ENGINE is not in PKG_CONFIG_DEPENDS. But that's beside the point.

libs/libfido2/Makefile
linosgian: Should we address this in another PR, once it's merged upstream?

linosgian:
@neheb This should be okay. Anything else that needs to be addressed before merging?

net/openssh/Makefile
linosgian: It seems like an omission indeed, but I think it escapes the scope of this PR

PR #14418, opened 2021-01-03
gnuplot: Updated gnuplot to 5.4.1

datafl4sh:
Signed-off-by: Matteo Cicuttin <datafl4sh@toxicnet.eu>

Maintainer: @datafl4sh 
Compile tested: x86, sunxi, bcm27xx
Run tested: x86, sunxi, bcm27xx

Description: Updated gnuplot from 5.4.0 to 5.4.1


neheb:
libdg needs a PKG_RELEASE bump then. I don't see any reason for the change though.

Also, run make package/x/refresh instead of manually refreshing patches.

datafl4sh:
> libdg needs a PKG_RELEASE bump then. 

Forgot it, sorry.

> I don't see any reason for the change though.

Uhm on my machine `libgd` build breaks specifically because of `libjpeg`
not being there anymore. Let me check, there is certainly an issue between
my keyboard and my chair then.

> Also, run make package/x/refresh instead of manually refreshing patches.

Will do.

utils/gnuplot/Makefile
```diff
@@ -1,14 +1,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gnuplot
-PKG_VERSION:=5.4.0
-PKG_RELEASE:=2
+PKG_VERSION:=5.4.1
+PKG_RELEASE:=3
```
neheb, line 5: reset to 1.

PR #14422, opened 2021-01-04
usbutils: add usbreset

kuanyili:
Add `usbreset` here to reflect the removal of `usbreset` package.

Maintainer: @nbd168
CC: @aparcar
Compile tested: aarch64, Raspberry Pi 4 Model B Rev 1.1, OpenWrt SNAPSHOT r15418-acdf07cd3e, `usbutils` installed
Run tested: as above

Description:
Package `usbreset` is removed in commit openwrt/openwrt@6cda95431984312d15032f6466b3b7238936185b. Add `usbreset` here so we can still use it.

### DIFF
--- a/utils/usbutils/Makefile
+++ b/utils/usbutils/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=usbutils
 PKG_VERSION:=013
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


BKPepe:
@kuanyili I wonder, why it was not possible to use ``usbutils`` package, which offers the same functionality as described in the referenced commit?

kuanyili:
@BKPepe It's been more than a year and there is no existing package that depends on `usbreset`, this may not be strictly necessary but could be done for sure. Should I submit another PR?

Issue #14414, opened 2021-01-03
utils/collectd: WriteThreads setting


f00b4r0:
Maintainer: @hnyman 
Environment: ramips 19.07

*Feature request.*

Description:
By default collectd uses 5 read and 5 write threads (see [here](https://collectd.org/documentation/manpages/collectd.conf.5.shtml#readthreads_num)). Setting `ReadThreads` to 2 as is currently done in the default uci config helps reduce memory/cpu pressure, but it makes little sense to still have 5 write threads active, especially when the only write output plugin is typically the local rrd database.

I'd suggest either matching the number of write threads with the read threads settings, or even better: making it a separable uci tunable.

I was able to lower collectd's memory usage by about 0.7MB by setting `WriteThreads` to 1.

HTH

mstorchak:
Fixed by #14424 in 8ab9fd7fff6f7ba432b6cf9bcaa0c58519aa399a

f00b4r0:
Thanks!

PR #14424, opened 2021-01-05
collectd: improvements

mstorchak:
Maintainer: @jow- , @hnyman 
Compile ath79, WNDR3800, r15339+5-bc99b56d7e
Run tested: ath79, WNDR3800, r15339+6-bc99b56d7e. ps shows the expected number of threads according to `WriteThreads`, `CollectInternalStats` works, `network/Forward` can be set to `false`, swap plugin _almost_ works: because ath79 has `CONFIG_VM_EVENT_COUNTERS` disabled, it doesn't report swap_io

Description: 
- fix the wrong var name in the initscript: should be `Forward`, not `value`
- cosmetic: remove extra newlines from the generated config
- expose `WriteThreads`, `WriteQueueLimitHigh`, `WriteQueueLimitLow` and `CollectInternalStats` to uci. The last one is useful for optimizing the number of threads and queue limits (addresses #14414)
- enable swap plugin
- bump PKG_RELEASE

Since almost any of the changes would require bumping PKG_RELEASE and thus would conflict with each other, I decided to post this as a single PR, but of course, each change can be approved or not individually.


hnyman:
Thanks. Looks ok to me.
Nice that you spotted the error in the network plugin Forward parameter.

Could you please rebase the PR: as the snmp6 plugin was merged, the plugin list change related to swap causes conflict (and also PKG_RELEASE should be bumped to .4)

mstorchak:
Done.
thanks for the review!

PR #14425, opened 2021-01-06
[WIP] modemmanager: add experimental support for SIM swtiching

nickberry17:
Signed-off-by: Nicholas Smith <nicholas@nbembedded.com>

WIP/RFC, not ready to merge yet. 

Maintainer: me
Compile tested: ipq40xx/Telco X1 Pro, git master
Run tested: ipq40xx/Telco X1 Pro

Description:
Adds ability to specify SIM slot in UCI.  Modem will switch to that slot before starting the connection, during the setup phase.

Caveat:
Requires qmi-utils installed to work.  Checking with @aleksander0m to see if it can be done without `qmicli`.

Example (add to a modemmanager interface)
`	option simslot '2'`

### DIFF
--- a/net/modemmanager/Makefile
+++ b/net/modemmanager/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=modemmanager
 PKG_VERSION:=1.14.8
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=ModemManager-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://www.freedesktop.org/software/ModemManager
--- a/net/modemmanager/files/modemmanager.proto
+++ b/net/modemmanager/files/modemmanager.proto
@@ -344,6 +344,7 @@ proto_modemmanager_init_config() {
 	proto_config_add_string	 pincode
 	proto_config_add_string	 iptype
 	proto_config_add_int	 signalrate
+	proto_config_add_int	 simslot
 	proto_config_add_boolean lowpower
 	proto_config_add_defaults
 }
@@ -355,11 +356,11 @@ proto_modemmanager_setup() {
 	local bearermethod_ipv4 bearermethod_ipv6 auth cliauth
 	local operatorname operatorid registration accesstech signalquality
 
-	local device apn allowedauth username password pincode iptype metric signalrate
+	local device apn allowedauth username password pincode iptype metric signalrate simslot
 
 	local address prefix gateway mtu dns1 dns2
 
-	json_get_vars device apn allowedauth username password pincode iptype metric signalrate
+	json_get_vars device apn allowedauth username password pincode iptype metric signalrate simslot
 
 	# validate sysfs path given in config
 	[ -n "${device}" ] || {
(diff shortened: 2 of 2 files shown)


aleksander0m:
@nickberry17 ModemManager git master already supports multiple SIMs and SIM switching; e.g. with `mmcli --set-primary-sim-slot=[SLOT NUMBER]`. I would definitely suggest to use that instead of plain qmicli, as that avoids *touching* the modem out of MM's view

aleksander0m:
Also, you're assuming that the modem is Qualcomm based and that the QMI support is enabled.... too many assumptions I think. Ideally, the protocol handler would use the MM API first to check whether multiple SIMs are reported, and if they are, allow switching between them.

The current implementation in MM works with "UIM Switch Slot" as the command you tried with qmicli, which gives you support for Dual SIM Single Standby, but it's not limited to that really. E.g. I have in my TODO list the task to add support for Dual SIM Dual Standby using other UIM commands (changing primary provisioning sessions and such), e.g. as required in the postmarketos integration for several phones out there.

nickberry17:
> @nickberry17 ModemManager git master already supports multiple SIMs and SIM switching; e.g. with `mmcli --set-primary-sim-slot=[SLOT NUMBER]`. I would definitely suggest to use that instead of plain qmicli, as that avoids _touching_ the modem out of MM's view

Thanks Aleksander, can you give any indication as to when this would be included in a release?
Regarding `--set-primary-sim-slot`  After reading the [commit message](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/merge_requests/330/diffs?commit_id=e7409b689876488594409e945687486adb1d9514) it sounds like the user is the one responsible for defining conditions that would trigger a SIM switch event, upon which we would `--set-primary-sim-slot=n` as the new primary SIM slot; did I understand that correctly?


>The 'PrimarySimSlot' property exposes which of the SIM slots available
in the system is the one configured as being primary. In a Multi-SIM
Single-Standby setup, the primary slot will be the one corresponding
to the single active SIM in the system. In a Multi-SIM Multi-Standby
setup, the primary slot will be the one configured to act as primary
(e.g. the one that will be used for the data connection) among all the
active SIM cards found. 



>too many assumptions I think.

Agreed, I just wanted to get a proof of concept running and then check with you to see if there was a better way that supports more devices.

aleksander0m:
> >     @nickberry17 ModemManager git master already supports multiple SIMs and SIM switching; e.g. with mmcli --set-primary-sim-slot=[SLOT NUMBER]. I would definitely suggest to use that instead of plain qmicli, as that avoids touching the modem out of MM's view
> 
> Thanks @aleksander0m, can you give any indication as to when this would be included in a release?

Next stable release; I don't have a clear date yet, but I'd like it to be soon, maybe in the next month or so.

> Regarding --set-primary-sim-slot After reading the commit message it sounds like the user is the one responsible for defining conditions that would trigger a SIM switch event, upon which we would --set-primary-sim-slot=n to change the active SIM; did I understand that correctly?

Yes, that is right. The MM API will tell you know many SIM slots were found, which is the currently in use, and if the slots had a valid SIM in them or not. Then it's up to the user to switch to a different SIM slot. The logic to select the slot can definitely be added to the protocol handler in openwrt, even e.g. allowing the user to select "the first available SIM" or "the slot which has SIM with ICCID xxxx" or "especifically SIM slot 2" and such.

nickberry17:
closing

PR #14429, opened 2021-01-07
libxerces-c: set iconv support to gnuiconv

neheb:
Fixes compilation under glibc.

Maintainer: @Noltari 
Compile tested: malta


neheb:
Probably some local problem...

PR #14432, opened 2021-01-07
libnetwork: update to 2020-12-15

neheb:
Remove some depends as MIPS is supported now.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @G-M0N3Y-2503 
Compile tested: mips64

PR #14435, opened 2021-01-08
docker: update to 20.10.2

neheb:
Remove non MIPS depends as MIPS is supported now.

Maintainer: @G-M0N3Y-2503 
Compile tested: mips64

### DIFF
--- a/utils/docker/Makefile
+++ b/utils/docker/Makefile
@@ -1,20 +1,21 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=docker
-PKG_VERSION:=20.10.1
+PKG_VERSION:=20.10.2
 PKG_RELEASE:=1
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=LICENSE
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/docker/cli/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=4ee7cc9c3f6287ca834166aaa1a945790c54d9a8345a1b449a193d9c739f2a7d
-PKG_SOURCE_VERSION:=562ea3c09c # SHA1 used within the docker executable
+PKG_HASH:=a663f54a158c6b2b23b253b14bf0de56ff035750098e760319de1edb7f4ae76d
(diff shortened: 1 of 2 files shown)


neheb:
I have no idea what this PKG_SOURCE_VERSION is.

utils/docker/Makefile
```diff
@@ -1,20 +1,21 @@
...
+PKG_VERSION:=20.10.2
 PKG_RELEASE:=1
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=LICENSE
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/docker/cli/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=4ee7cc9c3f6287ca834166aaa1a945790c54d9a8345a1b449a193d9c739f2a7d
+PKG_HASH:=a663f54a158c6b2b23b253b14bf0de56ff035750098e760319de1edb7f4ae76d
 PKG_SOURCE_VERSION:=562ea3c09c # SHA1 used within the docker executable
```
G-M0N3Y-2503, line 12: This is just [the short hash of the `v20.10.2` tag](https://github.com/docker/cli/releases/tag/v20.10.2), so, in this case.
```suggestion
PKG_SOURCE_VERSION:=2291f61 # SHA1 used within the docker executable
```

G-M0N3Y-2503 (changes requested):
Could we also make `/utils/dockerd/Makefile` the same version? The daemon verifies that all the dependencies are the expected release versions including this one.

Due to this verification, it can be easier just doing all version updates in the one PR, That said https://github.com/openwrt/packages/pull/14432 looks good to me so these ones can go in shortly after that one if it's easier.

G-M0N3Y-2503 (approved):
Looks good to me, Thanks!

neheb:
This will probably fail on the glibc buildbot because of https://downloads.openwrt.org/snapshots/faillogs/arc_archs/packages/tini/compile.txt

I don't think that should be fixed in this PR.

G-M0N3Y-2503:
> This will probably fail on the glibc buildbot because of https://downloads.openwrt.org/snapshots/faillogs/arc_archs/packages/tini/compile.txt
> 
> I don't think that should be fixed in this PR.

Agreed, I don't mind if these previously unsupported arches are left in a partially working state.

neheb:
I forgot about GO_ARCH_DEPENDS :).

PR #14441, opened 2021-01-08
clamav: allow configuring clamd to use TCP socket

flyn-org:

Maintainer: @lucize 
Compile tested: in progress
Run tested: in progress

Description:
I have been running Postfix on OpenWrt for a long time, and I would like to add virus detection. I set things up, but it seems something having to do with passing a file descriptor over spamd's Unix socket is causing the milter to fail. The milter says "ERROR: Unknown reply from clamd" and running strace on spamd reveals "No file descriptor received. ERROR."

Some work by others can be found on the Internet that suggests using a TCP socket for the communication between the milter and spamd fixes this. Lucian Cristian confirmed this on OpenWrt.

I am not sure why the Unix socket does not work. I suspect it is something related to musl, but I have not yet found evidence of this.

This merge request adds the option to configure spamd to use a TCP socket. I am presently testing this, but I wanted to start the merge request to elicit comments.


lucize:
Maybe also add the default port in the config file (commented) but LGTM

flyn-org:
@lucize, do you think we should add an init script for clamav-milter? During testing, I run the clamav-milter daemon by hand. I would be happy to add this too.

lucize:
yes, it doesn't do any harm

flyn-org:
@lucize, the latest force push contains an init script for clamav-milter. It also shifts the default clamd socket in clamd's configuration to TCP.

lucize:
Looks good

PR #14446, opened 2021-01-09
stress-ng: update to 0.12.01

neheb:
Simplified bsd patch to be compatible with glibc. It's easier to just
use libbsd with glibc. Size shouldn't be an issue.

Maintainer: @commodo 
Compile tested: arc700


neheb:
Hmm didn't work. Easier to just add libbsd to dependencies.

commodo (approved):

PR #14458, opened 2021-01-11
[openwrt-19.07] haproxy: Update HAProxy to v2.0.20

gladiac1337:
Maintainer: Thomas Heil <heil@terminal-consulting.de>, Christian Lachner <gladiac@gmail.com>
Run tested: ipq806x (r7800)

Description: Update HAProxy to v2.0.20

### DIFF
--- a/net/haproxy/Makefile
+++ b/net/haproxy/Makefile
@@ -10,12 +10,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=haproxy
-PKG_VERSION:=2.0.19
+PKG_VERSION:=2.0.20
 PKG_RELEASE:=1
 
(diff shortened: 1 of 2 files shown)


hnyman:
The change in the commit is simple, but where is "Update haproxy download URL and hash" ?
Similarly the commit message mentions "Fix uclibc patch offset" which I do not see in the commit.

Copy-paste error from #13878?

gladiac1337:
> The change in the commit is simple, but where is "Update haproxy download URL and hash" ?
> Similarly the commit message mentions "Fix uclibc patch offset" which I do not see in the commit.
> 
> Copy-paste error from #13878?

Yes, it is a copy-pase error, sorry.

PR #14468, opened 2021-01-11
madplay: remove OSS variant

neheb:
with the latest version, there's no OSS support. At least --without-oss
does not work.

Remove autoconf patches as they don't seem needed.

Maintainer: @probonopd 
Compile tested: powerpc


probonopd:
Thank you very much @neheb

PR #14470, opened 2021-01-11
gnunet: fix compilation with libdane

neheb:

Maintainer: @dangowrt 
Compile tested: powerpc

Fails everywhere currently: https://downloads.openwrt.org/snapshots/faillogs/arc_archs/packages/gnunet/compile.txt


dangowrt:
I'd kinda say it's better to mark libgnutls-dane as @BROKEN. I meant to find if there is a way to make it work in a meaningful way without indirectly depending on libopenssl, and I didn't find such way.

neheb:
ping @nmav

dangowrt (approved):

PR #14489, opened 2021-01-16
isc-dhcp: support dynamic dns for 10/8 and 172.16/12

pprindeville:
Maintainer: me
Compile tested: x86_64, generic, HEAD (b0cb305)
Run tested: same, built and installed on production router

Description:

We previously only supported 192.168/16 subordinate networks.  This adds support for 10/8 and 172.16/12 networks from RFC-1918 as well.


net/isc-dhcp/files/dhcpd.init
```diff
@@ -494,9 +528,18 @@ start_service() {
 
 		config_load dhcp
 
+		local rfc1918_nets=""
+
+		# alas we have to make 2 passes...
+		config_foreach dhcpd_add dhcp 0
+
+		rfc1918_nets="$(echo "$rfc1918_nets" | tr ' ' $'\n' | sort | uniq | tr $'\n' ' ')"
```
pprindeville, line 536: If I had associative arrays in Ash I would have done this "uniquify" by putting using the values as keys into an array and then retrieving the keys.

PR #14519, opened 2021-01-20
docker-compose: update to v1.28.0

jmarcet:
Maintainer: me
Compile tested: master x86_64
Run tested: master x86_64

Description:
New upstream version released this morning. Seems to work all right.

Features:
 - Support for Nvidia GPUs via device requests
 - Support for service profiles
 - Change the SSH connection approach to the Docker CLI's via shellout to the local SSH client (old behaviour enabled by setting COMPOSE_PARAMIKO_SSH environment variable)
 - Add flag to disable log prefix
 - Add flag for ansi output control

Bugs:
 - Make parallel_pull=True by default
 - Bring back warning for configs in non-swarm mode
 - Take --file in account when defining project_dir
 - On compose up, attach only to services we read logs from

### DIFF
--- a/utils/docker-compose/Makefile
+++ b/utils/docker-compose/Makefile
@@ -1,11 +1,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=docker-compose
-PKG_VERSION:=1.27.4
-PKG_RELEASE:=2
+PKG_VERSION:=1.28.0
+PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


jmarcet:
When executed it now prints this message:

`Building with native build. Learn about native build in Compose here: https://docs.docker.com/go/compose-native-build/`

The link is broken ATM though.

Also, it will recreate all the existing instances with a simple `docker-compose up -d`.

PR #14521, opened 2021-01-21
gitlab-runner: update to 13.9.0

neheb:
Backport patches that fix compilation under MIPS.

Maintainer: @ja-pa 
Compile tested: mips64

The backported patch seems like it will not be included under 13.8.0 final. Unfortunate if true.

### DIFF
--- a/devel/gitlab-runner/Makefile
+++ b/devel/gitlab-runner/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gitlab-runner
-PKG_VERSION:=13.5.0
+PKG_VERSION:=13.9.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-v$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://gitlab.com/gitlab-org/gitlab-runner/-/archive/v$(PKG_VERSION)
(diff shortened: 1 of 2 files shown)


aparcar:
Is it really worth to add a 25k line patch instead of waiting for their next release?

neheb:
Up to @ja-pa

BKPepe:
What its the need to use RC version instead of stable version?

neheb:
Easier to backport that patch. Note that this is a draft.

1715173329:
IMHO, it's really meaningless to add 25k line patch to support mips arch,
as I don't think there will be someone who wants to run it on such devices.

neheb:
Updated to 3.8.0.

BKPepe:
As others, I don't the patch, which is added, but that's up to you @ja-pa

neheb:
13.9.0 was released, so no need for massive patch.

aparcar:
Checks are failing?

neheb:
```
../../common/buildtest/masking.go:42:29: constant 9223372036854775807 overflows int
```

Now that is strange. Seems upstream only tests on 64-bit OS.

neheb:
Untested last minute change... https://github.com/gitlabhq/gitlab-runner/commit/6caa4c9f4054ac590fa245c8f36e4c46cfd9316d

neheb:
ping @ja-pa

PR #14525, opened 2021-01-21
dnscrypt-proxy2: don't enable blacklist by default

brndd:
Maintainer: @BKPepe 
Compile tested: no
Run tested: no

Description: the default blacklist.txt contains many very questionable entries that break sites, and is enabled by default. Some examples of questionable entries: `*.local`, `*.invalid`, `*.lan`, `stats.*`, `tracker.*`, `*.10.in-addr.arpa`.

It's better to have the blacklist disabled by default. The contents of the file should probably be looked at too (IMO just don't include it at all or include the dnscrypt-proxy2 example file instead, titled as such) but I don't want to do that.

### DIFF
--- a/net/dnscrypt-proxy2/Makefile
+++ b/net/dnscrypt-proxy2/Makefile
@@ -52,7 +52,6 @@ define Package/dnscrypt-proxy2/install
 	$(INSTALL_BIN) ./files/dnscrypt-proxy.init $(1)/etc/init.d/dnscrypt-proxy
 
 	sed -i "s/^listen_addresses = .*/listen_addresses = ['127.0.0.53:53']/" $(1)/etc/dnscrypt-proxy2/dnscrypt-proxy.toml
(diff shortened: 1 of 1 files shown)


BKPepe:
There were some changes in blocked names and I synced it for now with upstream one in this commit https://github.com/openwrt/packages/commit/d53d2df2832c392b8426cda4c99efeda17039ca7 which removes most of the issues, which you were having or noticed.

Your commit misses `Signed-off-by` and because of that, it could not be merged. Also, it required to bump `PKG_RELEASE`. 

However, I will take a look at why I added it earlier and why it was necessary. I think that I somewhere read that it was recommended to turn it in by default in a guide for dnscrypt-proxy, but I couldn't find it just now.

brndd:
> Your commit misses Signed-off-by and because of that, it could not be merged. Also, it required to bump PKG_RELEASE.

No worries, as long as the issue gets fixed!

As for the DNS filtering, I don't really see a purpose for it in dnscrypt-proxy especially on OpenWRT, because on OpenWRT this is going to be chained after dnsmasq which supports dns filtering and is where it should be done in my opinion.

PR #14530, opened 2021-01-22
iperf: import iperf3 and iperf from base

feckert:
Maintainer: @aparcar  @pprindeville 
Compile tested: x87_64, APU3, OpenWrt master
Run tested: no

Description:
Import iperf and iperf3 package from openwrt base.
https://github.com/openwrt/openwrt/pull/3807

### DIFF
--- a/net/iperf/Makefile
+++ b/net/iperf/Makefile
@@ -1,0 +1,68 @@
+#
+# Copyright (C) 2007-2010 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=iperf
+PKG_VERSION:=2.0.13
+PKG_RELEASE:=2
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_HASH:=c88adec966096a81136dda91b4bd19c27aae06df4d45a7f547a8e50d723778ad
+PKG_SOURCE_URL:=@SF/iperf2
+
+PKG_MAINTAINER:=Felix Fietkau <nbd@nbd.name>
+PKG_LICENSE:=BSD-3-Clause
+
+PKG_BUILD_PARALLEL:=1
+
+PKG_CONFIG_DEPENDS:=CONFIG_IPERF_ENABLE_MULTICAST
+
+include $(INCLUDE_DIR)/uclibc++.mk
+include $(INCLUDE_DIR)/package.mk
+
+define Package/iperf
+  SECTION:=net
+  CATEGORY:=Network
+  DEPENDS:= $(CXX_DEPENDS) +libpthread
+  TITLE:=Internet Protocol bandwidth measuring tool
+  URL:=http://sourceforge.net/projects/iperf2/
+endef
+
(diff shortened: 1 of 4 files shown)


net/iperf/Makefile
```diff
@@ -0,0 +1,68 @@
...
+PKG_SOURCE_URL:=@SF/iperf2
+
+PKG_MAINTAINER:=Felix Fietkau <nbd@nbd.name>
+PKG_LICENSE:=BSD-3-Clause
+
+PKG_BUILD_PARALLEL:=1
+
+PKG_CONFIG_DEPENDS:=CONFIG_IPERF_ENABLE_MULTICAST
+
+include $(INCLUDE_DIR)/uclibc++.mk
```
neheb, line 25: Curious if this should be removed.

uClibc++ is relatively unmaintained upstream and also has trouble being built in Arch Linux. Not to mention bugs.

OTOH, there's a big difference between 60KB and 450KB for libstdcpp (320 with a pending patch). Maybe static linking with flto could be a solution.

net/iperf/Makefile
```diff
@@ -0,0 +1,68 @@
...
+define Package/iperf/config
+	config IPERF_ENABLE_MULTICAST
+		depends on PACKAGE_iperf
+		bool "Enable multicast support"
+endef
+
+
+TARGET_CFLAGS += -D_GNU_SOURCE
+ifeq ($(CONFIG_IPERF_ENABLE_MULTICAST),y)
+CONFIGURE_ARGS += --enable-multicast
```
neheb, line 51: This could be rewritten as CONFIGURE_ARGS += $(call autoconf_bool,CONFIG_IPERF_ENABLE_MULTICAST,multicast)

net/iperf/Makefile
```diff
@@ -0,0 +1,68 @@
...
+else
+CONFIGURE_ARGS += --disable-multicast
+endif
+
+ifeq ($(CONFIG_IPV6),)
+	CONFIGURE_ARGS += --disable-ipv6
```
neheb, line 57: This could be rewritten as CONFIGURE_ARGS += $(call autoconf_bool,CONFIG_IPV6,ipv6)

net/iperf/Makefile
```diff
@@ -0,0 +1,68 @@
...
+endif
+
+CONFIGURE_VARS += CXXFLAGS="$$$$CXXFLAGS -fno-rtti"
+CONFIGURE_VARS += LIBS="-lpthread -lm"
```
neheb, line 61: This should probably be done only for GLIBC.

pprindeville:
@jow- Should we do any magic to move the history over from one repo to another?  I had to do it with `xtables-addons` when that moved, and it's not trivial... but probably worth the trouble.

aparcar:
> @jow- Should we do any magic to move the history over from one repo to another?  I had to do it with `xtables-addons` when that moved, and it's not trivial... but probably worth the trouble.

That would add many commits at once and wasn't done for the other packages. Who really needs to see the last 7 package bumps can always check openwrt.git

neheb:
I don't think history is worthwhile. There's already git history in the main repo.

aparcar:
I'd prefer to merge this now and then do any modifications. This way we had some sort of direct move of the packge.

PR #14536, opened 2021-01-23
oci-runtime-tools: add package

dangowrt:
Package oci-runtime-tool itself as well as oci-runtime-tests containing runtime validation test executables as well as rootfs tarballs used by the tests.
`oci-runtime-tool` can be used to generate OCI spec files or validate OCI bundles.
To validate the OCI runtime (runc, crun, uxc, maybe others) itself, install the `oci-runtime-tests` package as well as `node-npm`, use `npm` to install `node-tap` and symlink the `tap` executable to `/usr/bin`.
Then
```
cd /usr/libexec/oci-runtime-tests
tap *.t
```

### DIFF
--- a/utils/oci-runtime-tools/Makefile
+++ b/utils/oci-runtime-tools/Makefile
@@ -1,0 +1,83 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=oci-runtime-tools
+PKG_VERSION:=1.0.0-pre20210122
+PKG_RELEASE:=1
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_NAME:=runtime-tools
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/opencontainers/runtime-tools.git
+PKG_SOURCE_VERSION:=453547deb724af135587e654f52d86b8619f21b9
+PKG_MIRROR_HASH:=34f3aefc4f6f3bb93330ec7cbeb2eb987b82184da53d453e8b04b5750faa8728
+
+PKG_MAINTAINER:=Daniel Golle <daniel@makrotopia.org>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/opencontainers/runtime-tools/
+GO_PKG_LDFLAGS_X:=main.gitCommit=$(PKG_SOURCE_VERSION) main.version=$(PKG_SOURCE_VERSION)
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/oci-runtime-tool
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=oci-runtime-tool
+  URL:=https://github.com/opencontainers/runtime-tools
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
+define Package/oci-runtime-tool/description
+  OCI (Open Container Initiative) runtime tools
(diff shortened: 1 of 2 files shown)


utils/oci-runtime-tools/Makefile
```diff
@@ -0,0 +1,72 @@
...
+	( cd $(PKG_BUILD_DIR) ; $(FIND) ./validation/ -name *.go | grep -v util | \
+	xargs -I'%' -n 1 basename % .go | while read testbin; do \
+		$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/$$$${testbin} \
+			$(1)/usr/libexec/oci-runtime-validation-executables/$$$${testbin}.t ; \
+	done )
+endef
+
+
+$(eval $(call GoBinPackage,oci-runtime-tools))
+$(eval $(call BuildPackage,oci-runtime-tool))
```
aparcar, line 71: Wouldn't it be better if you use plural for both?

dangowrt: While the source package is called `runtime-tools` (plural) the packaged executable is called `oci-runtime-tool` (singular).
The other binaries, `runtimetest` as well as the validation executables are packaged in `oci-runtimetest`.

utils/oci-runtime-tools/Makefile
```diff
@@ -0,0 +1,72 @@
...
+PKG_MAINTAINER:=Daniel Golle <daniel@makrotopia.org>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/opencontainers/runtime-tools/
+GO_PKG_LDFLAGS_X:=main.gitCommit=$(PKG_SOURCE_VERSION) main.version=$(PKG_SOURCE_VERSION)
+
+
```
BKPepe, line 25:
```suggestion
```

utils/oci-runtime-tools/Makefile
```diff
@@ -0,0 +1,72 @@
...
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_SOURCE_NAME:=runtime-tools
+PKG_SOURCE:=$(PKG_SOURCE_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/opencontainers/runtime-tools/archive/v$(PKG_VERSION)
+PKG_SOURCE_VERSION:=ee63cfa6f66491e78f8e63c37806d7a905eb836e
+PKG_HASH:=4c2978e9097a3c93648c8ecf4f785f728044cddf717686e0b05a9244a1ad273c
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_SOURCE_NAME)-$(PKG_VERSION)
+
+PKG_MAINTAINER:=Daniel Golle <daniel@makrotopia.org>
```
BKPepe, line 16: What about something like this?  It's compile tested, and it's simpler and cleaner.

```suggestion
PKG_NAME:=oci-runtime-tools
PKG_VERSION:=0.9.0
PKG_RELEASE:=1

PKG_SOURCE:=v$(PKG_VERSION).tar.gz
PKG_SOURCE_URL:=https://github.com/opencontainers/runtime-tools/archive/v$(PKG_VERSION)
PKG_HASH:=4c2978e9097a3c93648c8ecf4f785f728044cddf717686e0b05a9244a1ad273c
PKG_BUILD_DIR:=$(BUILD_DIR)/runtime-tools-$(PKG_VERSION)

PKG_MAINTAINER:=Daniel Golle <daniel@makrotopia.org>
PKG_LICENSE:=Apache-2.0
PKG_LICENSE_FILES:=LICENSE
```

BKPepe (changes requested):
Can you please add in the commit description why you would like to add this package to this repository or its usage/purpose, please?

utils/oci-runtime-tools/Makefile
neheb: That PKG_SOURCE is no good. It potentially conflicts with other packages sharing the same version. Codeload would be better as it redirects to it anyway.

PR #14539, opened 2021-01-23
owipcalc: fix contains not respect default route

PolynomialDivision:
In IPv4 the default route can be written as
 0.0.0.0/0

In IPv6 the default route can be written as
 ::/0

If u try
  owipcalc 0.0.0.0/0 contains 1.1.1.1
or
  owipcalc ::/0 contains ::1
owipcalc will respond with 0 meaning that the "default prefixes" do not
contain the routes.

That is why we check now for 0 prefix.

Maintainer: @jow- 
Compile tested: Trunk
Run tested: TP-Link TL-WDR4300 v1 (Trunk)

Moved from: https://github.com/openwrt/openwrt/pull/3793/commits

### DIFF
--- a/net/owipcalc/Makefile
+++ b/net/owipcalc/Makefile
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=owipcalc
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 PKG_LICENSE:=Apache-2.0
 
 include $(INCLUDE_DIR)/package.mk
--- a/net/owipcalc/src/owipcalc.c
+++ b/net/owipcalc/src/owipcalc.c
@@ -227,7 +227,7 @@ static bool cidr_contains4(struct cidr *a, struct cidr *b)
 	if (printed)
 		qprintf(" ");
 
-	if ((b->prefix >= a->prefix) && (net1 == net2))
+	if ((a->prefix == 0) || ((b->prefix >= a->prefix) && (net1 == net2)))
 	{
 		qprintf("1");
 		return true;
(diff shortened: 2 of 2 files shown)


adschm:
Maintainer is "jow-", not "jow" ...

adschm:
I still wonder whether this is a fix or a workaround ...

PolynomialDivision:
> I still wonder whether this is a fix or a workaround ...

I will debug this again.

PolynomialDivision:
> I still wonder whether this is a fix or a workaround ...

I think there is some bug with the calculation. I will update.

PolynomialDivision:
@adschm There was a negative array access. If prefix is `0`, i is `16` and we will access `[-1]`. Now I know why my XDP script was not working. xD I think the code is now fine.

neheb:
autorelease would be nice.

jow-:
LGTM, thanks!

PolynomialDivision:
> autorelease would be nice.

But now we are at PKR_RELEASE 6 and AUTORELEASE would it set to 2? Bu I think it is a good idea.

adschm:
> > autorelease would be nice.
> 
> But now we are at PKR_RELEASE 6 and AUTORELEASE would it set to 2? Bu I think it is a good idea.

Interesting corner case. @aparcar FYI, with a moved package COMMITCOUNT will actually lower the PKG_RELEASE in specific cases.

aparcar:
The variable should only be introduced if at least one "bump" or "update" happened in the commit log. Moving packages between repositories is a bit of a special case. If we move something with autorelease to a different repo, we should temporarily move back to traditional numbers.

PR #14548, opened 2021-01-24
nebula: initial commit

stangri:
Maintainer: me
Compile tested: ramips, er-x, 19.07.6
Run tested: ramips, er-x, 19.07.6, start/connect to nebula IP from LAN/stop

Description:
This PR brings [SlackHQ's Nebula](https://github.com/slackhq/nebula) package to OpenWrt. The init script searches for `*.yml` files in `/etc/nebula/` and starts a nebula instance for each discovered `.yml`-file. Given the size of the binaries and the fact that `nebula-cert` is very unlikely to be used on the router, I've split it into variants.

THIS IS READY TO BE MERGED.

PS. SlackHQ confirmed I need to include the LICENSE file: https://github.com/slackhq/nebula/issues/371.

### DIFF
--- a/net/nebula/Makefile
+++ b/net/nebula/Makefile
@@ -1,0 +1,74 @@
+# Copyright 2021 Stan Grishin (stangri@melmac.net)
+# This is free software, licensed under the MIT License.
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=nebula
+PKG_VERSION:=1.3.0
+PKG_RELEASE:=1
+PKG_MAINTAINER:=Stan Grishin <stangri@melmac.net>
+
+PKG_SOURCE:=v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/slackhq/nebula/archive/
+PKG_HASH:=b94fba0251a4a436e25b127d0b9bc0181b991631f1dc8e344b1c8e895b55375d
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+GO_PKG:=github.com/slackhq/nebula
+GO_PKG_LDFLAGS_X:=main.Build=$(PKG_VERSION)
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/nebula
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=nebula
+  URL:=https://github.com/slackhq/nebula
+  DEPENDS:=$(GO_ARCH_DEPENDS) +kmod-tun
+endef
+
+define Package/nebula-cert
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=nebula-cert
+  URL:=https://github.com/slackhq/nebula
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
+define Package/nebula/description
+  Nebula is a scalable overlay networking tool with a focus on performance, simplicity
+  and security. It lets you seamlessly connect computers anywhere in the world.
+endef
+
+define Package/nebula-cert/description
+$(call Package/nebula/description)
+  This package contains only nebula-cert binary.
+endef
+
+define Package/nebula/install
+	$(call GoPackage/Package/Install/Bin,$(PKG_INSTALL_DIR))
+	$(INSTALL_DIR) $(1)/etc/init.d $(1)/usr/sbin $(1)/usr/share/doc/nebula $(1)/lib/upgrade/keep.d 
+	$(INSTALL_BIN) ./files/nebula.init $(1)/etc/init.d/nebula
+	$(SED) "s|^\(PKG_VERSION\).*|\1='$(PKG_VERSION)-$(PKG_RELEASE)'|" $(1)/etc/init.d/nebula
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/nebula $(1)/usr/sbin/nebula
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/LICENSE $(1)/usr/share/doc/nebula/LICENSE
+	$(INSTALL_DATA) ./files/nebula.upgrade $(1)/lib/upgrade/keep.d/nebula
+endef
+
+define Package/nebula-cert/install
+	$(call GoPackage/Package/Install/Bin,$(PKG_INSTALL_DIR))
+	$(INSTALL_DIR) $(1)/usr/sbin $(1)/usr/share/doc/nebula-cert $(1)/lib/upgrade/keep.d
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/nebula-cert $(1)/usr/sbin/nebula-cert
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/LICENSE $(1)/usr/share/doc/nebula-cert/LICENSE
+	$(INSTALL_DATA) ./files/nebula.upgrade $(1)/lib/upgrade/keep.d/nebula-cert
+endef
+
+$(eval $(call GoBinPackage,nebula))
+$(eval $(call BuildPackage,nebula))
+
+$(eval $(call GoBinPackage,nebula-cert))
+$(eval $(call BuildPackage,nebula-cert))
--- a/net/nebula/files/nebula.init
+++ b/net/nebula/files/nebula.init
@@ -1,0 +1,53 @@
+#!/bin/sh /etc/rc.common
+# Copyright 2021 Stan Grishin (stangri@melmac.net)
+# shellcheck disable=SC2039
+PKG_VERSION='dev-test'
+
+# shellcheck disable=SC2034
+START=80
+# shellcheck disable=SC2034
+USE_PROCD=1
+
+if type extra_command 1>/dev/null 2>&1; then
+	extra_command 'version' 'Show version information'
+else
+# shellcheck disable=SC2034
+	EXTRA_COMMANDS='version'
+fi
+
+readonly PROG=/usr/sbin/nebula
+
+version() { echo "Version: $PKG_VERSION"; }
+
+start_instance() {
+	local cfg="$1" port name="${1##*/}"
+	port="$(grep -A2 "^listen:" "$cfg" | grep "port: " | awk '{print $2}')"
+	procd_open_instance
+	procd_set_param command ${PROG} -config "${cfg}"
+	procd_set_param stderr 1
+	procd_set_param stdout 1
+	procd_set_param respawn
+	procd_open_data
+	json_add_array firewall
+		json_add_object ''
+		json_add_string type 'rule'
+		json_add_string name "Allow-$name"
+		json_add_string src 'wan'
+		json_add_string dest_port "$port"
+		json_add_string proto 'udp'
+		json_add_string target 'ACCEPT'
+		json_close_object
+	json_close_array
+	procd_close_data
+	procd_close_instance
+}
+
+start_service() {
+	local f
+	for f in /etc/nebula/*.yml; do
+		[ -s "$f" ] && start_instance "$f"
+	done
+}
+
+service_started() { procd_set_config_changed firewall; }
+service_stopped() { procd_set_config_changed firewall; }
--- a/net/nebula/files/nebula.upgrade
+++ b/net/nebula/files/nebula.upgrade
@@ -1,0 +1,1 @@
+/etc/nebula/
--- a/net/nebula/test.sh
+++ b/net/nebula/test.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+# shellcheck disable=SC2039
+
+"/usr/sbin/${1//-full}" -version 2>&1 | grep "$2"


nebula/files/nebula.init
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh /etc/rc.common
+# Copyright 2021 Stan Grishin (stangri@melmac.net)
+# shellcheck disable=SC2039
+PKG_VERSION='dev-test'
```
BKPepe, line 4: I think this is a leftover from testing and I am not sure if you want to have it here, because this will just add maintenance overhead and it is possible that you will forget to update it sometimes.

nebula/Makefile
```diff
@@ -0,0 +1,105 @@
...
+PKG_RELEASE:=1
+PKG_MAINTAINER:=Stan Grishin <stangri@melmac.net>
+
+PKG_SOURCE:=v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/slackhq/nebula/archive/
+PKG_HASH:=b94fba0251a4a436e25b127d0b9bc0181b991631f1dc8e344b1c8e895b55375d
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/nebula-$(PKG_VERSION)
```
BKPepe, line 17: Isn't this already default?

nebula/Makefile
```diff
@@ -0,0 +1,105 @@
...
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+GO_PKG:=github.com/slackhq/nebula
+GO_PKG_LDFLAGS_X:=main.Build=$(PKG_VERSION)
+
+include $(INCLUDE_DIR)/package.mk
+include $(TOPDIR)/feeds/packages/lang/golang/golang-package.mk
```
BKPepe, line 25:
```suggestion
include ../../lang/golang/golang-package.mk
```

nebula/Makefile
```diff
@@ -0,0 +1,105 @@
...
+
+define Package/nebula/default
+	SECTION:=net
+	CATEGORY:=Network
+	TITLE:=nebula
+	URL:=https://github.com/slackhq/nebula
+	DEPENDS:=$(GO_ARCH_DEPENDS)
```
BKPepe, line 32:
```suggestion
  SECTION:=net
  CATEGORY:=Network
  TITLE:=nebula
  URL:=https://github.com/slackhq/nebula
  DEPENDS:=$(GO_ARCH_DEPENDS)
```

nebula/Makefile
```diff
@@ -0,0 +1,105 @@
...
+endef
+
+define Package/nebula
+$(call Package/nebula/default)
+	DEPENDS+= +kmod-tun
```
BKPepe, line 37:
```suggestion
  DEPENDS+= +kmod-tun
```

nebula/Makefile
```diff
@@ -0,0 +1,105 @@
...
+endef
+
+define Package/nebula-cert
+$(call Package/nebula/default)
+	VARIANT:=cert
+	TITLE+=-cert
```
BKPepe, line 43:
```suggestion
  VARIANT:=cert
  TITLE+=-cert
```

nebula/Makefile
```diff
@@ -0,0 +1,105 @@
...
+endef
+
+define Package/nebula-full
+$(call Package/nebula/default)
+	DEPENDS+= +kmod-tun
+	VARIANT:=full
+	TITLE+=-full
```
BKPepe, line 50:
```suggestion
  DEPENDS+= +kmod-tun
  VARIANT:=full
  TITLE+=-full
```

nebula/Makefile
```diff
@@ -0,0 +1,105 @@
...
+endef
+
+define Package/nebula/description
+	Nebula is a scalable overlay networking tool with a focus on performance, simplicity
+	and security. It lets you seamlessly connect computers anywhere in the world.
```
BKPepe, line 55:
```suggestion
  Nebula is a scalable overlay networking tool with a focus on performance, simplicity
  and security. It lets you seamlessly connect computers anywhere in the world.
```

nebula/Makefile
```diff
@@ -0,0 +1,105 @@
...
+endef
+
+define Package/nebula-cert/description
+	$(call Package/nebula/description)
+	This package only contains nebula-cert binary.
```
BKPepe, line 60:
```suggestion
$(call Package/nebula/description)
  This package only contains nebula-cert binary.
```

nebula/Makefile
```diff
@@ -0,0 +1,105 @@
...
+endef
+
+define Package/nebula-full/description
+	$(call Package/nebula/description)
+	This package also contains nebula-cert binary.
```
BKPepe, line 65:
```suggestion
$(call Package/nebula/description)
  This package also contains nebula-cert binary.
```

nebula/Makefile
```diff
@@ -0,0 +1,105 @@
...
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/nebula-cert $(1)/usr/sbin/nebula-cert
+endef
+
+define Package/nebula-full/install
+	$(call Package/nebula/install,$(1))
+	$(call Package/nebula-cert/install,$(1))
+endef
+
+define Package/nebula/postinst
+	#!/bin/sh
```
BKPepe, line 89:
```suggestion
#!/bin/sh
```

nebula/Makefile
stangri: Is there an issue explicitly including this?

nebula/Makefile
stangri: This works for both buildbots and when the package compiled outside of the feeds. Your suggestion fails in the latter case.

nebula/Makefile
stangri: All the other Makefiles I've contributed to have tabs and not spaces. Has there been a change in this recently?

nebula/Makefile
BKPepe: The right question is why you would do that in the first place?

What are you doing is exactly this:
`PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)`

That's default and I don't see a reason, why it should be explicitly mentioned.

nebula/Makefile
BKPepe: But your solution is not correct if you want to have included your package here. Because you count that packages repo in feeds.conf is named as packages, but it could be named/mounted differently. Also, downstream distro might modify the build system architecture.

nebula/Makefile
BKPepe: I don't think there was a change looking in https://openwrt.org/docs/guide-developer/packages and https://openwrt.org/docs/guide-developer/helloworld/chapter3

nebula/files/nebula.init
stangri: This gets replaced automatically in the Makefile like in my other packages.

BKPepe: Ok.

nebula/Makefile
stangri: Thanks, resolved in the version I force-pushed.

nebula/Makefile
stangri: Thanks, I've updated formatting and force-pushed updates.

neheb:
ping @jefferyto as this is a go package.

net/nebula/Makefile
```diff
@@ -0,0 +1,104 @@
...
+  This package also contains nebula-cert binary.
+endef
+
+define Package/nebula/install
+	$(call GoPackage/Package/Install/Bin,$(PKG_INSTALL_DIR))
+	$(INSTALL_DIR) $(1)/etc/init.d $(1)/usr/sbin $(1)/usr/share/doc/nebula
+	$(INSTALL_BIN) ./files/nebula.init $(1)/etc/init.d/nebula
+	sed -i "s|^\(PKG_VERSION\).*|\1='$(PKG_VERSION)-$(PKG_RELEASE)'|" $(1)/etc/init.d/nebula
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/nebula $(1)/usr/sbin/nebula
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/LICENSE $(1)/usr/share/doc/nebula/LICENSE
```
jefferyto, line 73: Packages here don't usually include the license file as part of the .ipk file. Are there specific terms in upstream that require distribution of this file?

net/nebula/Makefile
```diff
@@ -0,0 +1,104 @@
...
+define Package/nebula-cert/install
+	$(call GoPackage/Package/Install/Bin,$(PKG_INSTALL_DIR))
+	$(INSTALL_DIR) $(1)/usr/sbin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/nebula-cert $(1)/usr/sbin/nebula-cert
+endef
+
+define Package/nebula-full/install
+	$(call Package/nebula/install,$(1))
+	$(call Package/nebula-cert/install,$(1))
+endef
```
jefferyto, line 85: If nebula-full is just nebula and nebula-cert combined, perhaps it would be better to have nebula-full be an empty package that depends on nebula and nebula-cert?

net/nebula/Makefile
```diff
@@ -0,0 +1,104 @@
...
+
+define Package/nebula/postinst
+#!/bin/sh
+if [ -z "$${IPKG_INSTROOT}" ]; then
+	grep -q '^/etc/nebula/' /etc/sysupgrade.conf || echo '/etc/nebula/' >> /etc/sysupgrade.conf
+fi
+endef
```
jefferyto, line 92: I think it would be cleaner to install a file to `/lib/upgrade/keep.d` instead of modifying `/etc/sysupgrade.conf`, e.g. [openvpn](https://github.com/openwrt/packages/blob/a64044339b4a3aa597cc58154de60ae93edfc546/net/openvpn/Makefile#L138-L140).

jefferyto:
A side question: Is there software other than Slack that uses this package? I'm not even aware that this is necessary to use Slack. (I'm not a real Slack user so I don't know :joy:)

I don't believe everything here has to have widespread usage, just wondering how/when this package is useful.

stangri:
> A side question: Is there software other than Slack that uses this package? I'm not even aware that this is necessary to use Slack. (I'm not a real Slack user so I don't know 😂)
> 
> I don't believe everything here has to have widespread usage, just wondering how/when this package is useful.

I'm not a Slack user either, but this has nothing to do with their principal product, it's an equivalent of zerotier/neorouter and hamachi to some extent. Helps you quickly create an almost zero-setup VPN between different machines.

net/nebula/Makefile
stangri: If we keep nebula and nebula-cert packages, I don't think nebula-full is needed. I've had three options in the Makefile because of the LICENSE file collision between nebula and nebula-cert and didn't know which way to go.

net/nebula/Makefile
stangri: https://github.com/slackhq/nebula/blob/master/LICENSE -- did I misunderstand it? An IPK file is a published variant of the upstream software and it says: The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

net/nebula/Makefile
stangri: Is there a flag for `$(INSTALL_DATA)` to not complain if file is already there?

stangri:
@jow- maybe you can help me figure this out. Is it possible to have two packages, `nebula` and `nebula-cert`, both of them having LICENSE and the `/lib/upgrade/keep.d/nebula` files, but without having a file collision on these? Like a conditional `$(INSTALL_DATA)` which wouldn't complain if the file is already there?

If this is impossible, would it be better to have `nebula` and `nebula-full` packages instead with `nebula-full` requiring `nebula` package to be removed first?

stangri:
@jow-  @aparcar  maybe you can suggest a solution. Is there a conditional/option for `$(INSTALL_DATA)` so that if the file is already there it doesn't complain during installation (so I could keep only `nebula` and `nebula-cert` non-exclusive packages)?

Or is it better to keep exclusive packages `nebula` and `nebula-full`?

Other than this issue, this is ready to be merged.

stangri:
Unless there are objections from reviews, this is ready to be merged.

net/nebula/Makefile
```diff
@@ -0,0 +1,109 @@
...
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/nebula $(1)/usr/sbin/nebula
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/LICENSE $(1)/usr/share/doc/nebula/LICENSE
+	$(INSTALL_DATA) ./files/nebula.upgrade $(1)/lib/upgrade/keep.d/nebula
+endef
+
+define Package/nebula-cert/install
+	$(call GoPackage/Package/Install/Bin,$(PKG_INSTALL_DIR))
+	$(INSTALL_DIR) $(1)/usr/sbin $(1)/tmp
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/nebula-cert $(1)/usr/sbin/nebula-cert
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/LICENSE $(1)/tmp/nebula.LICENSE
```
jefferyto, line 69: Both packages can have their own license file if you install this one to `/usr/share/doc/nebula-cert/LICENSE`.

net/nebula/Makefile
```diff
@@ -0,0 +1,109 @@
...
+	$(INSTALL_DATA) ./files/nebula.upgrade $(1)/tmp/nebula.upgrade
```
jefferyto, line 70: I think only the main nebula package needs this upgrade file.

jefferyto (commented):
Apologies, I haven't had time to follow this PR - I think the postinst/prerm scripts can be removed with these changes.

net/nebula/Makefile
stangri: Oh, brilliant, the license file is not that big, it's a great idea to just have two copies in a rare case both packages are installed! I'll update the Makefile.

net/nebula/Makefile
stangri: > I think only the main nebula package needs this upgrade file.

Actually, if anything, it's more important to keep the nebula files thru upgrade if the `nebula-cert` is installed, because then the `/etc/nebula` may contain the ca key.

I think the best option would be to deal with it just like with LICENSE and keep `nebula` and `nebula-cert` files in `/lib/upgrade/keep.d/`.

net/nebula/Makefile
```diff
@@ -0,0 +1,77 @@
...
+endef
+
+define Package/nebula
+$(call Package/nebula/default)
+  DEPENDS+= +kmod-tun
+endef
+
+define Package/nebula-cert
+$(call Package/nebula/default)
+  VARIANT:=cert
```
jefferyto, line 41: Is `VARIANT` here still necessary?

jefferyto (commented):
Sorry, noticed one more thing

net/nebula/Makefile
stangri: Is it better to define two different packages?

jefferyto: They already are two different packages. `VARIANT` is usually used when compilation is different for each variant, e.g. [openvpn](https://github.com/openwrt/packages/blob/6b9fdf2e77cdccb74902b84938d1744d8b8aafea/net/openvpn/Makefile#L53-L58), [less](https://github.com/openwrt/packages/blob/6b9fdf2e77cdccb74902b84938d1744d8b8aafea/utils/less/Makefile#L62-L70). AFAICT nebula and nebula-cert are packaging different binaries from the same compilation process.

stangri: > They already are two different packages. `VARIANT` is usually used when compilation is different for each variant, e.g. [openvpn](https://github.com/openwrt/packages/blob/6b9fdf2e77cdccb74902b84938d1744d8b8aafea/net/openvpn/Makefile#L53-L58), [less](https://github.com/openwrt/packages/blob/6b9fdf2e77cdccb74902b84938d1744d8b8aafea/utils/less/Makefile#L62-L70). AFAICT nebula and nebula-cert are packaging different binaries from the same compilation process.

Yeah, makes sense, was the leftover from nebula/nebula-full I guess. I pushed an update without variants.

jefferyto (approved):
Thanks!

Issue #14537, opened 2021-01-23
Please use "<package>: bump to <version>" for version changes


aparcar:
With the recent commit https://github.com/openwrt/openwrt/commit/9ae3c6f94c616cfbf854d3ec749c7fafc9893942 a new variable `$(AUTORELEASE)` is added to search the commit subject log of a package for the ": [bB]ump to " and ": [uU]pdate to ". The variable contains the number of commits since the latest bump, making manual changes of `PKG_RELEASE` obsolete.

Please use `<package>: bump to <version>` as commit message subject and set `PKG_RELEASE` to `$(AUTORELEASE)` to use this feature.

karlp:
Why would you not just look for commits that changed the PKG_VERSION instead of this string matching?  Especially if you have to opt in via AUTORELEASE anyway?

aparcar:
Some packages don't use PKG_VERSION, e.g. `odhcpd`.

PolynomialDivision:
@aparcar If source code is checked in here (like net/owipcalc), I can also use this feature, or? But there will never be a bump or update commit.

PolynomialDivision:
I changed all the packages that I maintain officially, but the corresponding packages do not have the release identifier in their name? Maybe I did a mistake.

aparcar:
> I changed all the packages that I maintain officially, but the corresponding packages do not have the release identifier in their name? Maybe I did a mistake.

I tried it locally and the release is attached:

```
Package: generate-ipv6-address
Version: 0.1-2
Depends: libc
License: MIT
```

```
generate-ipv6-address_0.1-2_x86_64.ipk
```

PolynomialDivision:
> I tried it locally and the release is attached:

I only looked at the output of github builds. Is there a difference?

aparcar:
Maybe the SDKs did not yet pickup that change.

aparcar:
@openwrt/package-maintainers I'd bin this issue if nobody objects.

cotequeiroz:
OK with pinning this. Additionally, we could perhaps add this information to the PR template text, where it is even more visible.

jefferyto:
I have only briefly skimmed through the main repo commit, but it seems to me that:

* If I make a change that does not affect the output package, e.g. fix a spelling mistake in the Makefile, `$(AUTORELEASE)` would still be incremented
* If I separate a change into multiple commits in one PR, because it makes logical sense to do so, `$(AUTORELEASE)` would advance by the number commits in the PR

[...]

PR #14551, opened 2021-01-24
Fix ddns provider

geraldhansen:
Maintainer: none
Compile tested: x86 openwrt-19.07.5
Run tested: VirtualBox

Description:
* Change parameter for myonlineportal
    As the default uclient-fetch doesn't support authentication header
    and the ddns provider myonlineportal.net support also username and
    passwort as url parameter this can be changed.
* Remove not working ddns provider
    I checked the hostname for existing DNS A and AAAA entries and these
    ones didn't have an entry.

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
 PKG_LICENSE:=GPL-2.0
 
--- a/net/ddns-scripts/files/usr/share/ddns/default/ddns.com.br.json
+++ b/net/ddns-scripts/files/usr/share/ddns/default/ddns.com.br.json
@@ -1,6 +1,0 @@
-{
-	"name": "ddns.com.br",
-	"ipv4": {
-		"url": "http://[DOMAIN]:[PASSWORD]@members.ddns.com.br/nic/update?hostname=[DOMAIN]&myip=[IP]"
-	}
-}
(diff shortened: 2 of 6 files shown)


hnyman:
This PR currently included in addition to your 2 commits also quite 15 unrelated commits.
You should likely rebase fix-ddns-provider branch on top on the master branch

geraldhansen:
Rebased to openwrt/packages:master

neheb:
change PKG_RELEASE to $(AUTORELEASE) . See https://github.com/openwrt/packages/issues/14537

geraldhansen:
> change PKG_RELEASE to $(AUTORELEASE) . See #14537

is this something I have to do?

neheb:
Scratch that. Just bump the PKG_RELEASE variable. Autorelease is not appropriate here.

geraldhansen:
I removed the PKG_RELEASE variable from the Makefile

feckert:
Sorry  but @neheb meant to increase it by one, and not to remove the number.

```
- PKG_RELEASE:=5
+ PKG_RELEASE:=6
```

You could also change this to
```
- PKG_RELEASE:=5
+ PKG_RELEASE:=$(AUTORELEASE) 
```
Then the buildsystem takes care of it and we don't have to count up the number all the time.

However, I would prefer the first method first, which now has a smaller impact, because the second method is quit new.

geraldhansen:
Ok, I got it and changed it.

feckert:
Sorry but the Pullrequest needs another rebase to current master

geraldhansen:
Rebase done

feckert:
I don't use these providers, so I can't say if the change works and if we can remove the others.
Opensource is based on trust, so thank you.
Merged.

geraldhansen:
Thanks, I tested them of course and also the one I removed.
If anyone complains, blame me and I will re-add them.

PR #14568, opened 2021-01-25
collectd: refactor snmp6

PolynomialDivision:
Getifaddrs() is not working all the time (e.g. wireguard interfaces).
Instead we use if_nameindex() syscall as suggested in:
 https://stackoverflow.com/a/45796495/8474618

Maintainer: @hnyman 
Compile tested: trunk
Run tested: wdr4300 (trunk)

### DIFF
--- a/utils/collectd/Makefile
+++ b/utils/collectd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=collectd
 PKG_VERSION:=5.12.0
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=https://collectd.org/files/ \
--- a/utils/collectd/patches/931-snmp6-add-ipv6-statistics.patch
+++ b/utils/collectd/patches/931-snmp6-add-ipv6-statistics.patch
@@ -155,7 +155,7 @@ Signed-off-by: Nick Hainke <vincent@systemli.org>
 +int snmp_read(char *ifname) {
 +  FILE *fh;
 +  char buffer[1024];
-+  char *fields[16];
++  char *fields[2];
 +  int numfields;
 +  int currline = 0;
 +  derive_t data[76];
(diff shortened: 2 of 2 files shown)


neheb (approved):

PolynomialDivision:
I added another changes. We saw now in production some weird stuff happening.

If u compare that:
http://monitor.berlin.freifunk.net/host.php?h=ohlauer-gw&p=snmp6
to
http://monitor.berlin.freifunk.net/host.php?h=ohlauer-gw&p=interface
You see that the is a difference with factor ~10?

I double check that now that the correct values are read. Not sure what is happening.

hnyman:
Any more changes coming?

PolynomialDivision:
> Any more changes coming?

Not now.
The changes are running in production now for 1 day. I definitely read the correct values.

PolynomialDivision:
Please merge. ;)

PolynomialDivision:
@hnyman Tested again with iperf and it is correct. Maybe there are issues with with a wireguard tunnel that has an ip4 endpoint? Further, it is weird that the values are in Byte/s and noch Mbit/s. Not sure if I can convert them somehow in the luci statistics app?

PolynomialDivision:
Okay, I tested again with wireguard interface and local graphs and the plugin is definitely correct! :) The difference is from the web interface plugin... xD

hnyman:
> the values are in Byte/s and noch Mbit/s. Not sure if I can convert them somehow in the luci statistics app?

Sure you can. 
data operations can be applied with RPN (reverse polish notation) format.
(See rrdgraph rrdtool docs upstream)

Just look examples from some other graphs. Ping, nut, chrony, sqm

PolynomialDivision:
> data operations can be applied with RPN (reverse polish notation) format.

Thanks! But now the total value is also in Bit? :/

hnyman:
> the total value is also in Bit?

It is just a number without any unit. It is upto you to give it a sensible unit as label in the graph.

PolynomialDivision:
> > the total value is also in Bit?
> 
> It is just a number without any unit. It is upto you to give it a sensible unit as label in the graph.

I mean it is now multiplied also by 8.

hnyman:
> I mean it is now multiplied also by the factor I give.

The data item that you tell rrdgraph to multiply, gets multiplied according to your instructions both in the graph and in the totals/last labels data table.

PR #14569, opened 2021-01-26
watchcat: update to support procd

nickberry17:
Signed-off-by: Nicholas Smith <nicholas@nbembedded.com>

Maintainer: rogerdammit@gmail.com 
Compile tested: ath79 Telco T1, master
Run tested: ath79 Telco T1 master, added and deleted WatchCat rules via LuCI with https://github.com/openwrt/luci/pull/4658 applied. 

Description:  Update to move watchcat into its own config file instead of in system and support procd


utils/watchcat/Makefile
```diff
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=watchcat
 PKG_VERSION:=1
-PKG_RELEASE:=11
+PKG_RELEASE:=13
```
neheb, line 12: should be 12, no?

neheb:
ping @roger-

utils/watchcat/Makefile
```diff
@@ -41,6 +41,8 @@ define Package/watchcat/install
 	$(INSTALL_BIN) ./files/watchcat.sh $(1)/usr/bin/watchcat.sh
 	$(INSTALL_DIR) $(1)/etc/uci-defaults
 	$(INSTALL_BIN) ./files/uci_defaults_watchcat $(1)/etc/uci-defaults/50-watchcat
+	$(INSTALL_DIR) $(1)/etc/config
+	$(INSTALL_DATA) ./files/watchcat.config $(1)/etc/config/watchcat
```
feckert, line 45: why do we need a uci default script for the configuration? If we have our own file then we can write it directly.

feckert (commented):
I would then also suggest that we extend the init script to use the procd?

utils/watchcat/Makefile
nickberry17: done that

utils/watchcat/Makefile
nickberry17: Fixed, my mistake. I had already bumped the version in my testing and had forgotten about that.

nickberry17:
> I would then also suggest that we extend the init script to use the procd?

I would like to do this in the future, but right now I am only just starting to get familiar with procd init scripts, so it would be some time before I can do that.

feckert:
> > I would then also suggest that we extend the init script to use the procd?
> 
> I would like to do this in the future, but right now I am only just starting to get familiar with procd init scripts, so it would be some time before I can do that.

ok fine with me.

@nickberry17 @hnyman Now that we have our own config file. What happens to those who still have it stored in System? Don't we need a uci_default upgrade script? To also make the people happy who restore he configuration at the firmware upgrade?

nickberry17:
Hey, 

I am working on the procd migration (Gist: https://gist.github.com/nickberry17/61d2431214f71c5083df934af9354415) but am having a problem. It sort of works, but it is only loading and running the first watchcat entry defined in the conf file.  I know that `config_foreach` is supposed to run the supplied function for each entry of the defined type (in this case: watchcat), however when I run with shell debugging (set -x) I can see that config_load is seeing triplicate entries of the ones in /etc/config/watchcat.  Can anyone spot any obvious error in the script?

feckert:
@nickberry17 Please have a look at https://github.com/openwrt/packages/blob/master/net/mwan3/files/etc/init.d/mwan3#L21
You have to start an instance for every service in the for each loop.
You could set a name for the instance as in mwan3, but this is not necessary.
If you do not set an argument to the function `procd_open_instance`. then the instances are named **instance1** **instance2** ...

nickberry17:
> @nickberry17 Please have a look at https://github.com/openwrt/packages/blob/master/net/mwan3/files/etc/init.d/mwan3#L21
> You have to start an instance for every service in the for each loop.
> You could set a name for the instance as in mwan3, but this is not necessary.
> If you do not set an argument to the function `procd_open_instance`. then the instances are named **instance1** **instance2** ...

Thanks, I got it working now

utils/watchcat/files/watchcat.init
```diff
@@ -90,53 +90,42 @@ load_watchcat() {
...
-		logger -p user.info -t "watchcat" "all tasks started"
-	else
-		logger -p user.info -t "watchcat" "no tasks defined"
-	fi
+service_triggers()
+{
+	procd_add_reload_trigger "watchcat"
 }
+
+reload_service()
```
neheb, line 127: This is already default.

utils/watchcat/files/watchcat.init
```diff
@@ -90,53 +90,42 @@ load_watchcat() {
...
+{
+	stop
+	start
+}
```
neheb, line 131: No newline?

utils/watchcat/files/watchcat.init
nickberry17: removed

utils/watchcat/files/watchcat.init
nickberry17: I can add it, but I cannot find a consistent pattern of examples or style guide to follow.  For instance, this one does not have a newline. https://github.com/openwrt/packages/blob/master/mail/emailrelay/files/emailrelay.init

Maybe I can start a style guide page on the wiki?

feckert: That would not be bad if you can do that.

utils/watchcat/files/watchcat.init
```diff
@@ -90,53 +90,36 @@ load_watchcat() {
 		logger -p user.err -t "watchcat" "reboot program $1 not started - $error"
 		return
 	}
+	
+	procd_open_instance
```
feckert, line 94: I wouldn't do it that way. If the switch case is not [defined](https://github.com/openwrt/packages/pull/14569/files#diff-34a04232b2d4ace6c3a9fc39702b0c51cc79a7a00825329430c10674d1faf700L108), then the object was opened anyway. I don't think that's clean.
I think it would be better to save the command for procd_set_param and then pass it to a complete procd_* block. And if the mode in switch case is not found then to a return to proceed to the next mode

utils/watchcat/files/watchcat.init
```diff
@@ -90,53 +90,36 @@ load_watchcat() {
...
-		rm "${PIDFILE}.pids"
-
-		logger -p user.info -t "watchcat" "all tasks stopped"
-	else
-		logger -p user.info -t "watchcat" "no tasks running"
-	fi
+start_service() {
+	config_load watchcat
+	config_foreach config_watchcat watchcat
+	procd_set_param file /etc/config/watchcat
```
feckert, line 119: Does this work? I think it belongs into the foreach loop to the procd object.

utils/watchcat/files/watchcat.init
nickberry17: > I think it would be better to save the command for procd_set_param and then pass it to a complete procd_* block. And if the mode in switch case is not found then to a return to proceed to the next mode

Agreed, there is nothing stopping it from opening an instance when the input is invalid, so it needs to be done that way.

utils/watchcat/files/watchcat.init
nickberry17: I've started it here: https://openwrt.org/code_style_guide. Let's continue discussion on the mailing list, wiki and forum.

utils/watchcat/files/watchcat.init
nickberry17: Yes, it works

```
 1864 root      1264 S    /bin/sh /usr/bin/watchcat.sh restart_iface 21600 8.8.8.8 30 standard wan
 1865 root      1264 S    /bin/sh /usr/bin/watchcat.sh periodic_reboot 21600 60
```

utils/watchcat/files/watchcat.init
nickberry17: I have added the new line at the end, as that should be the standard.

utils/watchcat/files/watchcat.init
```diff
@@ -90,53 +87,36 @@ load_watchcat() {
...
+			procd_set_param command /usr/bin/watchcat.sh "restart_iface" "$period" "$pinghosts" "$pingperiod" "$pingsize" "$interface" "$mmifacename"
+			;;
+		*)
+			echo "Error starting Watchcat service. Invalid mode selection: $mode"
+			;;
 	esac
 
-	echo $! >>"${PIDFILE}.pids"
+	procd_set_param respawn ${respawn_threshold:-3600} ${respawn_timeout:-5} ${respawn_retry:-5}
+	procd_close_instance
```
feckert, line 111: From my point of view, this is still not correct. What if we end up in the `*`? Then the functions`procd_set_param_respawn` and `procd_close_instance` are still called.

utils/watchcat/files/watchcat.init
```diff
@@ -90,53 +87,36 @@ load_watchcat() {
...
-	config_load system
-	if [ -n "$(uci show system.@watchcat[0])" ]; then # at least one watchcat section exists
-		logger -p user.info -t "watchcat" "starting all tasks"
-		config_foreach load_watchcat watchcat
-		logger -p user.info -t "watchcat" "all tasks started"
-	else
-		logger -p user.info -t "watchcat" "no tasks defined"
-	fi
+service_triggers()
+{
```
feckert, line 120: This bracket should not be in a new line. Better at the end of the upper line. If you want to have it diffrent, then it should be however uniform in this file.

feckert (changes requested):
What about the upgrade script? Shouldn't we port the existing configuration from system to watchat? If not how do we make it known to the user that their current configuration will no longer work when upgrading from 19.07 to 21.xx?

utils/watchcat/files/watchcat.init
nickberry17: You're right, and my preferred style is on the same line too.  I just missed that.

feckert:
@nickberry17 From my point of view now everything fits except for the handling during sysupgrade to the upcoming stable release?
What do we want to do there?
Can we transfer the configuration one to one from /etc/config/system to /etc/config/watchcat?
Or has something else changed.
As a base we can take the following uci-defaults scripts.
* https://github.com/openwrt/packages/blob/e00afb6dc90db404b00d74be6380f01576801d90/net/keepalived/files/etc/uci-defaults/keepalived
* https://github.com/openwrt/packages/blob/9ef2e619b9bba054d6848714f77b462f1f359c2c/net/mwan3/files/etc/uci-defaults/mwan3-migrate-flush_conntrack

This is how I always do it when the configuration has changed in my maintained packages.
I think in openwrt base there are also some. I have taken them as a basis see https://github.com/openwrt/openwrt/blob/e02a41f67db52726d3c2c9e4e50e51a5510e0a27/package/system/rpcd/files/50-migrate-rpcd-ubus-sock.sh
The advantage is that the user does not notice any difference on upgrade. Otherwise we have destroyed the configuration which the user did not expect and that is not good.

nickberry17:
> @nickberry17 From my point of view now everything fits except for the handling during sysupgrade to the upcoming stable release?
> What do we want to do there?
> Can we transfer the configuration one to one from /etc/config/system to /etc/config/watchcat?
> Or has something else changed.
> As a base we can take the following uci-defaults scripts.
> 
> * https://github.com/openwrt/packages/blob/e00afb6dc90db404b00d74be6380f01576801d90/net/keepalived/files/etc/uci-defaults/keepalived
> * https://github.com/openwrt/packages/blob/9ef2e619b9bba054d6848714f77b462f1f359c2c/net/mwan3/files/etc/uci-defaults/mwan3-migrate-flush_conntrack
> 
> This is how I always do it when the configuration has changed in my maintained packages.
> I think in openwrt base there are also some. I have taken them as a basis see https://github.com/openwrt/openwrt/blob/e02a41f67db52726d3c2c9e4e50e51a5510e0a27/package/system/rpcd/files/50-migrate-rpcd-ubus-sock.sh
> The advantage is that the user does not notice any difference on upgrade. Otherwise we have destroyed the configuration which the user did not expect and that is not good.

Hey, we should be able to one-to-one transfer from system to watchcat.  How about this one liner to do the migration?

`cat /etc/config/system | grep -s "config watchcat" -A 5 | tr -d "\-\-" > /etc/config/watchcat`

feckert:
@nickberry17 This would be my suggestion, do convert the script
```
#!/bin/sh

. /lib/functions.sh

upgrade_watchcat() {
        local cfg="$1"

        config_get period "$cfg" period
        config_get mode "$cfg" mode
        config_get pinghosts "$cfg" pinghosts
        config_get forcedelay "$cfg" forcedelay

        [ -f "/etc/config/watchcat" ] || touch /etc/config/watchcat
        uci_add watchcat watchcat
        uci_set watchcat @watchcat[-1] period "$period"
        uci_set watchcat @watchcat[-1] mode "$mode"
        uci_set watchcat @watchcat[-1] pinghosts "$pinghosts"
        uci_set watchcat @watchcat[-1] forcedelay "$forcedelay"

        uci_remove system "$cfg"
}

config_load system
config_foreach upgrade_watchcat watchcat

uci_commit watchcat
uci commit system
```

nickberry17:
ping @feckert

feckert:
thanks for the ping.

PR #14584, opened 2021-01-27
ddns-scripts: make WGET_SSL executable

EdiTurn:
Maintainer: @feckert
Compile tested: x86, master
Run tested: x86, master

Description: 
In 9eab8cc, WGET_SSL has been changed to something like `-cares +digest -gpgme +https +ipv6 -iri +large-file -metalink -nls`, and it breaks some thirdparty ddns scripts. So I change it back to a valid path.


feckert:
Thanks, I totally missed that! This should not happen!

PR #8000, opened 2019-01-20
Golang: Use upx to compress files by default

lrinQVQ:
Golang generated files are usually larger,
this is very unfriendly for small Flash devices
Signed-off-by: lrinQVQ <cqlrin@gmail.com>

Maintainer: @jefferyto
Compile tested: (ipq40xx, Asus RT-AC58U, OpenWrt master)
Run tested: (ipq40xx, Asus RT-AC58U, OpenWrt SNAPSHOT r9060+19-4029788ff3, tests done)

Description:
See details: https://github.com/openwrt/openwrt/pull/1751

PR #14598, opened 2021-01-28
xray-core: add new package

1715173329:
Maintainer: me
Compile tested: i386, ipq40xx, ipq806x, ramips, rockchip, x86_64
Run tested: ipq40xx, rockchip, x86_64

Description:
v2ray-core is a platform for building proxies to bypass network restrictions,
and it secures your network connections and thus protects your privacy.

It supports most popular proxies protocols like Shadowsocks, Trojan, VMESS and VLESS,
as well as port forwarding, NAT traversal, DNS resolving etc.
More information can be found here: https://www.v2fly.org/en_US / https://xtls.github.io/config/

Xray is the superset of V2Ray, with extra XTLS support and higher performance, and fully
compatible with configuration of V2Ray, so it's fine to use Xray to provide V2Ray.

### DIFF
--- a/net/xray-core/Makefile
+++ b/net/xray-core/Makefile
@@ -1,0 +1,107 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=xray-core
+PKG_VERSION:=1.2.4
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/XTLS/xray-core/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=25f6c9edec0ac1f98328943cd2bb760ac7b69107582f9d27e43559da39dc01ed
+
+PKG_MAINTAINER:=Tianling Shen <cnsztl@project-openwrt.eu.org>
+PKG_LICENSE:=MPL-2.0
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/Xray-core-$(PKG_VERSION)
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/xtls/xray-core
+GO_PKG_BUILD_PKG:=github.com/xtls/xray-core/main
+GO_PKG_LDFLAGS:=-s -w
+GO_PKG_LDFLAGS_X:= \
+	$(GO_PKG)/core.version=$(PKG_VERSION) \
+	$(GO_PKG)/core.codename=OpenWrt
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/xray/template
+  TITLE:=A platform for building proxies to bypass network restrictions
+  SECTION:=net
+  CATEGORY:=Network
+  URL:=https://xtls.github.io
+endef
+
+define Package/xray-core
+  $(call Package/xray/template)
+  DEPENDS:=$(GO_ARCH_DEPENDS) +ca-bundle
+  PROVIDES:=v2ray-core
+endef
+
+define Package/xray-geodata
+  $(call Package/xray/template)
+  TITLE+= (geodata files)
+  DEPENDS:=xray-core
+endef
+
+define Package/xray/description
+  Xray, Penetrates Everything. Also the best v2ray-core, with XTLS support. Fully compatible configuration.
+  It secures your network connections and thus protects your privacy.
+endef
+
+define Package/xray-core/description
+  $(call Package/xray/description)
+endef
+
+define Package/xray-geodata/description
+  $(call Package/xray/description)
+
+  This includes GEO datas used for xray-core.
+endef
+
+GEOIP_VER:=202101280019
+GEOIP_FILE:=geoip.dat.$(GEOIP_VER)
+
+define Download/geoip
+  URL:=https://github.com/v2fly/geoip/releases/download/$(GEOIP_VER)/
+  URL_FILE:=geoip.dat
+  FILE:=$(GEOIP_FILE)
+  HASH:=69bb1f820f416e4591a7b76bfabf9fde9cce6550cddcc1a99b2ccafeb2a8ebd3
+endef
+
+GEOSITE_VER:=20210130061540
+GEOSITE_FILE:=dlc.dat.$(GEOSITE_VER)
+
+define Download/geosite
+  URL:=https://github.com/v2fly/domain-list-community/releases/download/$(GEOSITE_VER)/
+  URL_FILE:=dlc.dat
+  FILE:=$(GEOSITE_FILE)
+  HASH:=bc9748c95b0545376356174b005fccfbc21d647df46293d5197853ee06a9c0a2
+endef
+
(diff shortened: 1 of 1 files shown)


1715173329:
Similar PR here: #10154.

1715173329:
BTW, v2ray official provides the geoip datas for route controlling, but the datas are
updated in every 2 days.
In my option, it's improssible to trace it mannually, so is there any way to keep it always
up-to-date automatically?

https://github.com/v2fly/domain-list-community/releases
https://github.com/v2fly/geoip/releases

yichya:
I would recommend building Xray with CGO_ENABLED=0 for better compatibility across different openwrt distributions and versions (especially when built with glibc instead of musl)

1715173329:
> I would recommend building Xray with CGO_ENABLED=0 for better compatibility across different openwrt distributions and versions (especially when built with glibc instead of musl)

I don't see any point of doing this.
And as far as I know there's no C code included in both V2Ray and XRay.

yichya:
> > I would recommend building Xray with CGO_ENABLED=0 for better compatibility across different openwrt distributions and versions (especially when built with glibc instead of musl)
> 
> I don't see any point or benifit of doing this.
> And as far as I know there's no C code included in both V2Ray and XRay.

By default OpenWrt toolchain build go executables with CGO enabled so the binary is dynamically linked to the provided libc (musl in most cases). However that can be troublesome when the binary is executed under some other environment (like a normal linux distribution which uses glibc). Even official OpenWrt provides both glibc and musl as optional libc implementation and there are cases (for example, better iconv support) someone may build own image with glibc in which situation this package won't be able to simply "opkg install" ed.

see below:

```
yichya@yichya-nas /m/Public> ls xray* -al
-rwxr-xr-x 1 yichya yichya 19205592 Jan 29 02:31 xray*
-rwxr-xr-x 1 yichya yichya 19174352 Jan 29 02:15 xray_cgo*
yichya@yichya-nas /m/Public> ldd xray
        not a dynamic executable
yichya@yichya-nas /m/Public [1]> ldd xray_cgo
./xray_cgo: error while loading shared libraries: /usr/lib/libc.so: invalid ELF header
yichya@yichya-nas /m/Public [1]> ./xray -version
Xray 1.2.3 (Xray, Penetrates Everything.) Custom (go1.15.7 linux/amd64)
A unified platform for anti-censorship.
yichya@yichya-nas /m/Public> ./xray_cgo -version
Failed to execute process './xray_cgo'. Reason:
The file './xray_cgo' does not exist or could not be executed.
yichya@yichya-nas /m/Public [125]>
```

1715173329:
> However that can be troublesome when the binary is executed under some other environment

This is designed for OpenWrt.

> some may build their own images with glibc in which situation this package won't be able to simply "opkg install" ed.

That's not in our consideration, as opkg feeds is used for official image.

yichya:
> That's not in our consideration, as opkg feeds is used for official image.

My point is, one can build his or her own OpenWrt image with any libc implementation supported while still use official OpenWrt package feeds and install Xray with just a simple `opkg install xray` command.

And at the same time, official released xray (https://github.com/XTLS/Xray-core/releases/tag/v1.2.3) is also built with CGO turned off.

And actually turning off CGO is fairly easy, just 3 lines of code to override Build/Compile target:

```
define Build/Compile
	cd $(PKG_BUILD_DIR)/main; $(GO_PKG_VARS) CGO_ENABLED=0 go build -o $(PKG_INSTALL_DIR)/bin/xray .; 
endef
```

1715173329:
Again, that's not in our consideration.

1715173329:
@jefferyto Could you please make a review?

1715173329:
Updated to latest v1.2.4 version, and added geodata package.

1715173329:
Any idea? @neheb

jefferyto:
Apologies, I haven't had time to follow this PR. I think it is not necessary to set `GO_PKG_LDFLAGS:=-s -w` since the build system will strip binaries as necessary, and perhaps setting `$(GO_PKG)/core.build` instead of `$(GO_PKG)/core.codename` would be more appropriate, but otherwise I don't see any major issues.

1715173329:
@jefferyto That's great, do thanks for your precious advice, and I'll put these changes in my next PR (maybe a week later, with program updating.)

1715173329:
> I think it is not necessary to set `GO_PKG_LDFLAGS:=-s -w` since the build system will strip binaries as necessary.

However I have tested it, and the result shows we still need it to work, as stirping from build system seems not to be the best solution.

jefferyto:
> However I have tested it, and the result shows we still need it to work, as stirping from build system seems not to be the best solution.

Did you unselect `DEBUG` and select either `USE_STRIP` or `USE_SSTRIP` (in .config / make menuconfig)?

1715173329:
> Did you unselect `DEBUG` and select either `USE_STRIP` or `USE_SSTRIP` (in .config / make menuconfig)?

Yes, I unselected `CONFIG_DEBUG` and selected `CONFIG_USE_SSTRIP` with ARGS `-z`

jefferyto:
I did two test compiles using the x86-64 sdk, one with `GO_PKG_LDFLAGS:=-s -w` and one without, with `CONFIG_DEBUG` unselected and `CONFIG_USE_SSTRIP` selected, and the resulting `xray` binaries in the .ipk files were identical.

When you say "seems not to be the best solution", what do you mean exactly?

1715173329:
However, in my cases, with LDFLAGS, the size of binary was ~19 MB, while with SSTRIP it was ~26.8 MB.
After being compressed with UPX, the size of former was ~5 MB while the latter was 11 MB.

And also, I tested with upx disabled.
Yes, if they were packed into .ipk, the size seemed to be identical (~6.8 MB),
but actually the binaries were not, which meant more memory was required when running it (in SquashFS we need decompress the binaries to memory)
or more space it used (when installing manually), and this resulting being unfriendly to those small memory / flash devices.

Even though these differences seems not to be problems on modern devices, but on ath79 or ramips liked targets, the size is really important, and I just wanna make the binary smaller.

jefferyto:
When I say the binaries are identical, I mean I used `diff`/`cmp` to do a bit-by-bit comparison. This has nothing to do with UPX. I have also compiled using armvirt-32 and malta-be sdks with the same result (the malta-be binaries differed by 1 byte).

Can you try doing a `make dirclean` or starting from a clean sdk, before compiling? I would really like to know why stripping isn't working for you.

1715173329:
Okay! I finally understood why.
The build system will do STRIP only when the binary is about to be packed into .ipk,
so when I compared with the original binaries in `GO_PKG_BUILD_BIN_DIR` it caused different results.

However it's fine to keep it as-is, as I'm planning on introduing upx, this would be fair.

jefferyto:
Someone has already looked at adding upx: #8000

I believe the conclusion was that because squashfs already uses compression, compressing again using upx causes file sizes to increase. (I haven't actually tested this myself though.)

1715173329:
> I believe the conclusion was that because squashfs already uses compression, compressing again using upx causes file sizes to increase.

@jefferyto 

I'm not sure if the size would be larger, but according to #8000, at least it should be the same.
However, squashfs only works for users who build with this by theirselves.
If they install or upgrade it online, then a 19MB binary will be put into their flash, it's really not okay.

jefferyto:
JFFS2 also has compression: https://openwrt.org/docs/techref/filesystems#jffs2

You are welcome to try to package upx, but it will need give actual storage savings for it to be merged.

PR #14603, opened 2021-01-29
collectd: add plugin to compare ipv4,ipv6 stats

PolynomialDivision:
Add "ipstatistics"-plugin. This plugin parses "/proc/net/netstat" and "/proc/net/snmp6" to get the overall ipv4 and ipv6 usage.

Maintainer: tbf
Compile tested: WDR4300
Run tested: WDR4300

### DIFF
--- a/utils/collectd/Makefile
+++ b/utils/collectd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=collectd
 PKG_VERSION:=5.12.0
-PKG_RELEASE:=6
+PKG_RELEASE:=7
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=https://collectd.org/files/ \
@@ -153,6 +153,7 @@ COLLECTD_PLUGINS_SELECTED:= \
 	filecount \
 	fscache \
 	interface \
+	ipstatistics \
 	iptables \
 	irq \
 	iwinfo \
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
https://github.com/openwrt/luci/pull/4753

hnyman:
I am a bit hesitant of us adding even more OpenWrt-specific generic plugins to collectd. It would be great to get your plugins into the upstream collectd.

We also already have e.g. nlbwmon (with LuCI support), which already collectds and shows ipv4/ipv6 traffic data.

PolynomialDivision:
> We also already have e.g. nlbwmon (with LuCI support), which already collectds and shows ipv4/ipv6 traffic data.

Interesting! :)
I did not know of it berfore. Thanks for pointing me to the daemon.

PolynomialDivision:
> I am a bit hesitant of us adding even more OpenWrt-specific generic plugins to collectd. It would be great to get your plugins into the upstream collectd.

If I try up-streaming that, could we merge it? For us this statistics are very interesting since we introduced IPv6 recently and now I would like to have an easy view for anyone that IPv6 out-roles IPv4. ;) I can not put it into the snmp6 interface, since I'm not able to have different IP stats for different interface.

Actually, I planned to make a babel statistics plugin. xD

PolynomialDivision:
I saw that we have to update the plugins if collectd switches to 6.0.

neheb:
Is this ready/should be merged?

PolynomialDivision:
> Is this ready/should be merged?

There is a problem that I and @hnyman see this differently. @hnyman is not that excited about having more openwrt specific plugins and prefer to upstream everything. I can understand that, as we will soon have to rewrite all plugins again when it comes to the new collectd. But I take care of my packages and would of course do that for my plugins.

I'm a bit frustrated with upstreaming to collectd since all my PRs are still opened, but I think this is just because they have this big new version change.

hnyman:
We can merge this, but in the long run it really would be better to have the stuff upstream.

(But upstream collectd has been in some turmoil for the last 2-3 years, and there has been e.g. arguments about branching strategy etc )

PolynomialDivision:
> (But upstream collectd has been in some turmoil for the last 2-3 years, and there has been e.g. arguments about branching strategy etc )

Interesting! What exactly happened?

PR #14608, opened 2021-01-30
libgpg-error: fix build error on macOS

httpstorm:
libgpg-error 1.39 does not compile on macOS, because src/gen-lock-obj.sh runs in /bin/sh, which does not support echo -n, and generates garbage. Since /bin/sh is an absolute path on a read-only file-system, it is not possible to replace the binary or adjust the PATH. Resolved by switching to /bin/dash.

Maintainer: @flyn-org
Compile tested: macOS 11.1, Ubuntu 20.04 LTS
Run tested: arm9 mvebu, WRT3200ACM, OpenWrt SNAPSHOT r14965+644-3f1109bf2a

### DIFF
--- a/libs/libgpg-error/Makefile
+++ b/libs/libgpg-error/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libgpg-error
 PKG_VERSION:=1.39
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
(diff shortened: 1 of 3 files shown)


flyn-org:
I think it would be better to patch the script itself so that it is POSIX compliant. The upstream libgpg-error would benefit from this. The script should use printf rather than "echo -n," as the latter is not required by POSIX.

httpstorm:
Thank you for the advice, @flyn-org !
I updated the patch to use printf.

neheb:
Please bump PKG_RELEASE. also run make package/libgpg-error/refresh

httpstorm:
@neheb 
Thank you Rosen, for helping me learn!
I made the changes you requested and pushed an update.

flyn-org:
Looks good to me.

PR #14609, opened 2021-01-30
hwdata 0.343: fix build error on macOS

httpstorm:
The install command on macOS does not support the -T flag.
Resolved by omitting the -T flag  when building on macOS.

Maintainer: @vcrhonek, @zpc0
Compile tested: macOS 11.1, Ubuntu 20.04 LTS
Run tested: arm9 mvebu, WRT3200ACM, OpenWrt SNAPSHOT r14965+644-3f1109bf2a

### DIFF
--- a/utils/hwdata/Makefile
+++ b/utils/hwdata/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=hwdata
 PKG_VERSION:=0.343
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/vcrhonek/hwdata/tar.gz/v$(PKG_VERSION)?
@@ -16,9 +16,6 @@ PKG_HASH:=ccb4d21337b3773cc02654b360e91feb44c9258728d2f027b72013bd5628113b
 PKG_MAINTAINER:=
 PKG_LICENSE:=GPL-2.0-or-later  XFree86-1.0
 PKG_LICENSE_FILES:=LICENSE
-
-PKG_INSTALL:=1
-PKG_BUILD_PARALLEL:=1
 
 include $(INCLUDE_DIR)/package.mk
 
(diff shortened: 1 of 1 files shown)


neheb:
What does make install do for hwdata? I would think the actual files can just be copied over.

httpstorm:
@neheb 
Hello Rosen!
Here is the install target. Effectively the patch removes the -T flag ([--no-target-directory](https://linux.die.net/man/1/install) = treat DEST as a normal file). It should be safe to remove it. I added the check for macOS to preserve the original behaviour.

```
install: Makefile.inc
	mkdir -p -m 755 $(DESTDIR)$(datadir)/$(NAME)
	for foo in $(IDFILES) ; do \
		install -m 644 $$foo $(DESTDIR)$(datadir)/$(NAME) ;\
	done
	mkdir -p -m 755 $(DESTDIR)$(libdir)/modprobe.d
-	install -m 644 -T blacklist.conf $(DESTDIR)$(libdir)/modprobe.d/dist-blacklist.conf
+	install -m 644 blacklist.conf $(DESTDIR)$(libdir)/modprobe.d/dist-blacklist.conf

```

neheb:
No I mean something like this:

```
--- a/utils/hwdata/Makefile
+++ b/utils/hwdata/Makefile
@@ -17,9 +17,6 @@ PKG_MAINTAINER:=
 PKG_LICENSE:=GPL-2.0-or-later  XFree86-1.0
 PKG_LICENSE_FILES:=LICENSE
 
-PKG_INSTALL:=1
-PKG_BUILD_PARALLEL:=1
-
 
 define Package/pciids
@@ -38,12 +35,12 @@ endef
 
 define Package/pciids/install
        $(INSTALL_DIR) $(1)/usr/share/hwdata
-       $(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/share/hwdata/pci.ids $(1)/usr/share/hwdata
+       $(INSTALL_DATA) $(PKG_BUILD_DIR)/pci.ids $(1)/usr/share/hwdata
 endef
 
 define Package/usbids/install
        $(INSTALL_DIR) $(1)/usr/share/hwdata
-       $(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/share/hwdata/usb.ids $(1)/usr/share/hwdata
+       $(INSTALL_DATA) $(PKG_BUILD_DIR)/usb.ids $(1)/usr/share/hwdata
 endef
 
 $(eval $(call BuildPackage,pciids))
```

There's no need to compile anything or use PKG_INSTALL.

httpstorm:
@neheb 
Thank you, Rosen!
I implemented your patch. Everything works correctly. I also changed PKG_RELEASE.

PR #14616, opened 2021-01-31
snort3: fix alert_syslog

flyn-org:
Maintainer: me
Compile tested: x86_64
Run tested: x86_64

Description:
Snort 3.1.0.0 contains a bug that causes snort to crash when loading the syslog alert module. This adds a patch that fixes this bug. The package maintainer submitted the patch to the upstream snort project.

### DIFF
--- a/libs/libdaq3/Makefile
+++ b/libs/libdaq3/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libdaq3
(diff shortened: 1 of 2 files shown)


neheb:
Needs a PKG_RELEASE bump and make package/snort3/refresh

snort and snort3 need to be cleaned up. snort currently does not install because libdaq and libdaq3 install to the same location causing a race condition between what gets compiled when.

flyn-org:
Closing this because I have packaged a new version.

PR #14619, opened 2021-02-01
antfs: patch for linux >= 5.8.0

lipnitsk:

Maintainer: @neheb @dengqf6 
Compile tested: mt7621, linux 5.10

### DIFF
--- a/kernel/antfs/Makefile
+++ b/kernel/antfs/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=antfs
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_URL:=https://github.com/klukonin/antfs.git
 PKG_SOURCE_PROTO:=git
--- a/kernel/antfs/patches/0001-file.c-Patch-for-Linux-v5.8.patch
+++ b/kernel/antfs/patches/0001-file.c-Patch-for-Linux-v5.8.patch
@@ -1,0 +1,47 @@
+diff --git a/file.c b/file.c
+index 105e41c..98d5204 100644
+--- a/file.c
++++ b/file.c
+@@ -625,6 +625,9 @@ static int antfs_readpages(struct file *file, struct address_space *mapping,
+ 		unsigned page_idx = nr_pages;
+ 		pgoff_t page_idx_to_init;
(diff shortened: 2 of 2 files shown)


neheb:
Interesting. I wonder if paragon will make an out of tree module for their driver.

neheb:
ping @klukonin There's a version 7.21.

lipnitsk:
Yeah I tried 7.21 but it didn't bring any new changes... But yeah, looks like the Paragon driver is [on its way](https://lore.kernel.org/lkml/20210128090455.3576502-1-almaz.alexandrovich@paragon-software.com/#r), probably for 5.12.

neheb:
I wonder how hard it would be to backport.

lipnitsk:
Probably easier to port to 5.10 than to 5.4 :) So maybe when we switch away from 5.4 completely?

dengqf6:
https://github.com/LGA1150/ntfs3-oot

dengqf6:
I'd like to abandon ANTFS and use NTFS3 instead.

lipnitsk:
We could merge this to fix antfs until it is removed from the modules completely, how does that sound?

neheb:
lol kernel 5.10 isn't even in the tree.

@dengqf6 that port is...newer than when you posted. strange.

lipnitsk:
Strange indeed, maybe https://aur.archlinux.org/packages/ntfs3-dkms/ is a better source - it is based directly on the patchset. Although `readahead` patching (https://github.com/LGA1150/ntfs3-oot/commit/3fbb0e8e01145982515798055ee360eb7fce36a5) will be needed for older kernels.

neheb:
@dengqf6 just looked. antfs is in master only and not in 19.07. Might as well abandon it before release.

1715173329:
> @dengqf6 that port is...newer than when you posted. strange.

The owner used git rebase so...

klukonin:
@neheb I can update my fork.

1715173329:
Maybe it's more reasonable to backport the patches directly than introduce a 3rd-party fork?

klukonin:
Well. There are no any changes between 07.19 and 07.22 as I can see.

klukonin:
So if it's a semi-stable code, why not?

neheb:
@1715173329 there's a difference between the packages feed and base. base will almost certainly not accept such a thing. Which leaves an out of tree module for the packages feed.

PR #14633, opened 2021-02-03
dockerd: change dockerd start level to 99 to avoid unknown conflicts

qinyuhang:
Maintainer: @G-M0N3Y-2503
Compile tested: (rockchip, rk3399, snapshot)
Run tested: (rockchip, rk3399, snapshot)

Description: I found in my device, then dockerd started, but some of the containers are not started, though I put `--restart=always` param. But after the device boot, I manually restart dockerd in `luci/admin/system/startup`, all the containers started. So I test to change the Start level of dockerd, it resolve the problem.


1715173329:
1. You need bump PKG_RELEASE.
2. The commit title should describe your changes,
     e.g. 'dockerd: adjust startup level'

BTW, I have no idea if this would break anything.

qinyuhang:
> 1. You need bump PKG_RELEASE.
> 2. The commit title should describe your changes,
>    e.g. 'dockerd: adjust startup level'
> 
> BTW, I have no idea if this would break anything.

Thank you, I'll change the PKG_RELEASE and add change intension.

G-M0N3Y-2503:
I don't recall why it was 25 initially or any reason for it to stay that way, So if it fixes this for you I'm happy for this to go in.

Thanks for the work!

qinyuhang:
> I don't recall why it was 25 initially or any reason for it to stay that way, So if it fixes this for you I'm happy for this to go in.
> 
> Thanks for the work!

I found some container not started up with 25 level, I guess it may because the network config is interrupted by some other start up process. After I change the level to 99, it works, all containers started successfully.

PR #14634, opened 2021-02-03
alpine: try to make reproducible

neheb:

Maintainer: @pesintta

### DIFF
--- a/mail/alpine/patches/020-reproducible.patch
+++ b/mail/alpine/patches/020-reproducible.patch
@@ -1,0 +1,18 @@
+--- a/pico/blddate.c
++++ b/pico/blddate.c
+@@ -28,6 +28,7 @@ main(argc, argv)
+ 	exit(1);
+     }
+ 
++/*
+     time(&ltime);
+     t = localtime(&ltime);
+     fprintf(outfile,"char datestamp[]=\"%02d-%s-%d\";\n", t->tm_mday, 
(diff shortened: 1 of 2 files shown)


mail/alpine/patches/020-reproducible.patch
```diff
@@ -0,0 +1,18 @@
...
+     }
+ 
++/*
+     time(&ltime);
+     t = localtime(&ltime);
+     fprintf(outfile,"char datestamp[]=\"%02d-%s-%d\";\n", t->tm_mday, 
+@@ -46,6 +47,7 @@ main(argc, argv)
+ 	    1900 + t->tm_year);
+ 
+     fprintf(outfile, "char hoststamp[]=\"random-pc\";\n");
```
pesintta, line 14: Does this really work? Grepping alpine sources it seems to refer to `extern char datestamp[], hoststamp[];` and then snprintf() the contents.

I wonder what the contents are if the actual definition here is commented out and would this cause potential crashes in some unexpected places.

Would it be safer and make more sense to hardcode the values here to some fixed values rather than leave them undefined?

neheb: No idea.

neheb:
https://tests.reproducible-builds.org/openwrt/dbd/packages/arm_cortex-a8_vfpv3/packages/alpine_2.24-1_arm_cortex-a8_vfpv3.ipk.html

PR #14636, opened 2021-02-03
netifyd: add reproducibility patch

neheb:

Maintainer: @dsokoloski 
Compile tested: ath79

### DIFF
--- a/net/netifyd/Makefile
+++ b/net/netifyd/Makefile
@@ -6,7 +6,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=netifyd
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_MAINTAINER:=Darryl Sokoloski <darryl@egloo.ca>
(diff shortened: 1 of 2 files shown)


dsokoloski:
@neheb Hi Rosen!

Thanks for the patch.  I'm not sure what you mean by "reproducibility".  Could you explain what the motivation is here?

Maybe we should just remove that line of output altogether?

1715173329:
> Thanks for the patch. I'm not sure what you mean by "reproducibility". Could you explain what the motivation is here?

@dsokoloski This means users could build the same binaries from source as what OpenWrt buildbot built.
Yes, they even share the same HASH.

dsokoloski:
@1715173329 I understand...  I have removed the git date/time text from the source.

neheb:
https://tests.reproducible-builds.org/openwrt/dbd/packages/arm_cortex-a8_vfpv3/packages/netifyd_2020-09-15-v3.05-1_arm_cortex-a8_vfpv3.ipk.html

PR #14639, opened 2021-02-03
xray-core: add example configs and init script

1715173329:
Maintainer: me

Description:
1. Add example conffiles from v2fly/v2ray-core wikis.
2. Add init script so the users could run it with their configs quickly.

### DIFF
--- a/net/xray-core/Makefile
+++ b/net/xray-core/Makefile
@@ -40,10 +40,18 @@ define Package/xray-core
   PROVIDES:=v2ray-core
 endef
 
+define Package/xray-example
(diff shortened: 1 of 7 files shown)


neheb:
Instead of bumping PKG_RELEASE, just set it to $(AUTORELEASE)

1715173329:
> Instead of bumping PKG_RELEASE, just set it to $(AUTORELEASE)

Thanks for reminding me of this useful function which I almost forgot.
Updated now.

zxlhhyccc:
Suggest to add Luci Adapter for easier operation!

Issue #14625, opened 2021-02-01
prometheus_node_exporter_lua not sending wifi stats


matthewhelmke:
Maintainer: @champtar 
Environment: mvebu, Linksys WRT-1900ac v1, OpenWrt 19.07.6 r11278-8055e38794

Description:
Metrics for the prometheus_node_exporter_lua_wifi and _wifi_stations packages do not exist in `IP:9100/metrics` while all other expected metrics continue to be updated as expected.

The change occurred starting with the upgrade to OpenWrt 19.07.6.

Here is the current content of `IP:9100/metrics` with the following packages from the set installed.
Packages:
- prometheus-node-exporter-lua
- prometheus-node-exporter-lua-nat_traffic
- prometheus-node-exporter-lua-netstat
- prometheus-node-exporter-lua-openwrt
- prometheus-node-exporter-lua-wifi
- prometheus-node-exporter-lua-wifi_stations

```
# TYPE node_scrape_collector_duration_seconds gauge
# TYPE node_scrape_collector_success gauge
# TYPE node_nf_conntrack_entries gauge
node_nf_conntrack_entries 1259
# TYPE node_nf_conntrack_entries_limit gauge
node_nf_conntrack_entries_limit 16384
node_scrape_collector_duration_seconds{collector="conntrack"} 0.00044703483581543
node_scrape_collector_success{collector="conntrack"} 1
# TYPE node_boot_time_seconds gauge
node_boot_time_seconds 1611234660
# TYPE node_context_switches_total counter
node_context_switches_total 554732730
# TYPE node_cpu_seconds_total counter
node_cpu_seconds_total{cpu="cpu0",mode="user"} 2058
node_cpu_seconds_total{cpu="cpu0",mode="nice"} 0
node_cpu_seconds_total{cpu="cpu0",mode="system"} 4699.78
node_cpu_seconds_total{cpu="cpu0",mode="idle"} 943681.61
node_cpu_seconds_total{cpu="cpu0",mode="iowait"} 0.25
node_cpu_seconds_total{cpu="cpu0",mode="irq"} 0
node_cpu_seconds_total{cpu="cpu0",mode="softirq"} 10015.62
node_cpu_seconds_total{cpu="cpu0",mode="steal"} 0
node_cpu_seconds_total{cpu="cpu0",mode="guest"} 0
node_cpu_seconds_total{cpu="cpu0",mode="guest_nice"} 0
[...]

PR #14641, opened 2021-02-03
prometheus-node-exporter-lua: fix wifi

PolynomialDivision:
The "get_wifi_interfaces" function is not returning the wifi interface names. This causes the bug #14625.

Fix the "get_wifi_interfaces" function.

### DIFF
--- a/utils/prometheus-node-exporter-lua/Makefile
+++ b/utils/prometheus-node-exporter-lua/Makefile
@@ -4,7 +4,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=prometheus-node-exporter-lua
-PKG_VERSION:=2021.01.27
+PKG_VERSION:=2021.02.15
 PKG_RELEASE:=1
 
 PKG_MAINTAINER:=Etienne CHAMPETIER <champetier.etienne@gmail.com>
--- a/utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/hostapd_stations.lua
+++ b/utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/hostapd_stations.lua
@@ -6,9 +6,9 @@ local function get_wifi_interfaces()
   local status = u:call("network.wireless", "status", {})
   local interfaces = {}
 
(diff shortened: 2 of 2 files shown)


PolynomialDivision:
Okay, I tested now the commit and it makes the situation better, but actually needed hostapd_cli does not work?

Can someone try?

    hostapd_cli -i wlan0 all_sta

PolynomialDivision:
Bug is reported:
https://bugs.openwrt.org/index.php?do=details&task_id=3608

champtar:
I'm pretty sure it was working in the past, curious if there was breaking change in ubus response

PolynomialDivision:
> I'm pretty sure it was working in the past, curious if there was breaking change in ubus response

If u print the names from get_wifi, they will be "null". No idea. But more problematic is the hostapd-utils. xD

champtar:
This was introduced in https://github.com/openwrt/packages/commit/af8f8f80469fe595171d9dd4895a7dc0fbe0e29b
```
... however they require a full build of hostapd / wpad.
```

champtar:
On a 19.07 build, wpad-basic-wolfssl - 2019-08-08-ca8c2bd2-4, I get
```
~# ubus call network.wireless status
{
	"radio0": {
...
		"interfaces": [
			{
				"section": "default_radio0",
				"ifname": "wlan0",
				"config": {
					"mode": "ap",
...
```
(ifname at the same level as config, not in config)

Anyone with a full build can run the same ubus command ?

neheb:
Needs rebasing.

PolynomialDivision:
Actually, there is more that should be somehow addressed. Why bit32 and not `local bit = require "bit" `.

PolynomialDivision:
@champtar I would merge, since u also tested this fix, or?

champtar:
> Actually, there is more that should be somehow addressed. Why bit32 and not `local bit = require "bit" `.

I'm on my phone but that bit vs bit32 rings a bell, I think it was discussed in the past, so look at the git log / PR before changing

champtar:
I had someone else test in the issue ;)
still need better dependency but this is a needed fix so let's not delay

PR #14644, opened 2021-02-04
libs/nss: Don't run nsinstall with QUILT

lipnitsk:
Allows targets such as prepare, refresh, or update to be run without
building dependencies for easier patch maintenance.

Maintainer: @lucize @neheb


lucize:
LGTM

PR #14646, opened 2021-02-04
treewide: Run refresh on all packages

lipnitsk:
The crude loop I wrote to come up with this changeset:
```bash
  find -L package/feeds/packages/ -name patches | \
  sed 's/patches$/refresh/' | sort | xargs make
```

Maintainer: @neheb


neheb:
Love it. Sometimes refreshing these patches like this catches already applied patches that are hidden by fuzz.

I would love to see a solution for GitHub Actions so that massive commits like these would no longer be necessary.

mhei:
I don't like this. Please exclude at least the php stuff - most of the chances seems to be just noise. The reason why I prefer to accept some fuzz is that it makes maintaining easier when you use imported patches (e.g. from Debian). It is then easier to compare upstream changed patches with the local ones.

neheb:
Note that debian also uses quilt.

Also note that historically, there was a php patch that reapplied some code already present in the codebase because of fuzz.

karlp:
While the idea is sound, this needs to be reviewed package by package.  Many of these simply end up stripping all patch header context, with no other change, which is "correctly refreshed" but just removes metadata useful to package maintainers for the low low price of churning commits.

neheb:
Headers are not affected. In some cases where there are two patches in one file, this is true.

I'm mostly seeing it get rid of git fuzz.

karlp:
"git fuzz"? it's mostly changing the context functionheader line lenght, lookat  admin/backuppc/patches/001-debian-fixes.patch , there's not a single "refresh" it's purely diff options  the ipmi ones are clearly removing contextual header that isuseful for manual patch tracking and has no impact, and there's no actual line refresh, just truncating context to a different line width.  I'm actually seeing very few that are real at all, gkrellem, schroot actually _changes_ the patch, erlang looks valid perhaps.

honestly, given the masive amoutn of noise in this patch, I actualyl think it's rather inconsiderate filing it.  it's churn of the highest degree

lipnitsk:
> honestly, given the masive amoutn of noise in this patch, I actualyl think it's rather inconsiderate filing it. it's churn of the highest degree
I think if we come up with a way to enforce that patches must be refreshed going forward this churn might be justified. Also it was just a matter of running a simple script to come up with it, although it took 2-3 hours to run due to so many make targets. Whatever the consensus, I think it is worth discussing dirty patches, especially those that result in offsets and fuzz in the `patch` tool sense.

lang/python/python-requests/patches/0001-idna-dependency-bump.patch
```diff
@@ -12,43 +12,15 @@ https://github.com/psf/requests/pull/5711
...
+ 
  requires = [
      'chardet>=3.0.2,<5',
 -    'idna>=2.5,<3',
-+    'idna>=2.5,<4',
-     'urllib3>=1.21.1,<1.27',
-     'certifi>=2017.4.17'
-
-
-From d3e00a4958af046879f24de365d5589d861ea6ef Mon Sep 17 00:00:00 2001
```
neheb, line 29: This needs to be split to two patches. Notice how quilt is merging the patches here.

karlp:
How about we grep the build logs for "offset" or whatever comes up when it doesn't quite match, and just call refresh on those?  Those ones would _benefit_ from refreshing, and we would instantly avoid all the noise of "still applies perfectly, but has different diff options for formatting and headers"

neheb:
All previous problems seem to have been fixed. I can't see any bad patch anymore.

I see some patches that should not be here but that can be dealt with separately.

karlp:
oh, so you made changes, and then there was no chance for review again? ok, whatever. fine.

karlp:
And no, lots of these were still bad patches, jsut scrolling the first 10 or so, the gkrellem one is correct, the others are all just quilt options on line lengths changing, they're completely trash noise commits.  great job.  Hope the next person to do this uses the same settings you did. :(

neheb:
@stintel complained at https://github.com/openwrt/packages/pull/14808

It's better to not let new contributors get discouraged.

jefferyto:
It would be nice if this could be cherry-picked to the openwrt-21.02 branch so that future cherry-picks from master to openwrt-21.02 do not lead to merge conflicts.

jefferyto:
> It would be nice if this could be cherry-picked to the openwrt-21.02 branch so that future cherry-picks from master to openwrt-21.02 do not lead to merge conflicts.

For example, when I try to cherry pick 6d9005d5ac63d5a3092577fdd4f085f00d1f5fdd I get a merge conflict for lang/python/python3/patches/016-adjust-config-paths.patch because the patch was modified by this PR on master but not on openwrt-21.02.

Yes I can manually resolve the conflict, but potentially dozens of maintainers will waste their time manually resolving merge conflicts that could be avoided by cherry-picking this change into openwrt-21.02.

PR #14647, opened 2021-02-04
[19.07] ksmbd(-tools): update to 3.3.4

neheb:
Maintainer: @Andy2244 

ping @ynezz @hauke as this is a kernel module, it need to be merged right before a new release is cut. The current version has several bugs including some big endian ones.

### DIFF
--- a/kernel/ksmbd/Makefile
+++ b/kernel/ksmbd/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ksmbd
-PKG_VERSION:=3.1.3
+PKG_VERSION:=3.3.4
 PKG_RELEASE:=1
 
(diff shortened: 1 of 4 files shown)


Andy2244:
yeah would be nice to get this finally in stable, so we have the stable ABI on next release.

zorun:
Just to be clear, once this is merged, it will break all previous 19.07.X versions, right?

neheb:
I believe so.

Andy2244:
> Just to be clear, once this is merged, it will break all previous 19.07.X versions, right?

Kinda, if a user updates a older 19.07.x version of the ksmbd-tools to the new version, while still having the old kernel module.
This means the current kmod in 19.07.x is not ABI compatible to this 3.3.x+ userspace tools package.

neheb:
Set to ready. I assume .7 will be released soon.

hauke:
Thank you for the change. I agreed with @neheb  to merge this before tagging 19.07.7

PR #14655, opened 2021-02-04
sqlite3: add option for batch-atomic write support

janh:
This allows to build SQLite with support for the batch-atomic write
optimization. It makes use of atomic write support provided by the
underlying file system. Currently, this is only supported on F2FS. It
it does not work with overlayfs.

Signed-off-by: Jan Hoffmann <jan@3e8.eu>

Compile tested: x86_64
Run tested: x86_64


neheb:
this needs a PKG_RELEASE bump.

janh:
Fixed.

champtar:
https://www.sqlite.org/compile.html#enable_batch_atomic_write
what is the size difference with/without ?

janh:
On x86_64, the size of the libsqlite binary is identical. The package is 447 bytes larger with the option.

champtar:
I don't know if we can default to yes if F2FS is compiled in the kernel, else I'm ok to merge like that

janh:
> I don't know if we can default to yes if F2FS is compiled in the kernel

I couldn't find a way make the default depend on the actual kernel config. It seems to be possible only for those kernel options that are configured in the main configuration (`config/Config-kernel.in`).

I have to say that I am unsure about the best default here. This seems to be enabled on recent Android, so it should be well-tested. I just kept it off by default to stay with the upstream default. And due to overlayfs it is unlikely to be useful on a typical OpenWrt installation anyway.

PR #14656, opened 2021-02-04
miniupnpd: introduce IGDv1 variant

stintel:
Microsoft Windows, Xbox and possibly other operating systems do not
support IGDv2. With IGDv2 enabled, they send a HTTP GET request for
rootDesc.xml and WANIPCn.xml, and then nothing happens. The Microsoft
implementation probably doesn't like the WANIPCn.xml response and
decides UPnP is not available. When miniupnpd is built without IGDv2
support, after the 2 HTTP GET requests, there is a HTTP POST request to
/ctl/IPConn, and miniupnpd configures the port forward as expected.

The runtime option force_igd_desc_v1=yes (UCI: igvd1) does not solve
this problem. It's possible this was enough in earlier miniupnpd
versions, but it does not fix the problem the current version.

Since we are a modern distro, we want to support the latest and
greatest, so we should default to IGDv2 enabled. Introducing a
menuconfig option to disable IGDv2 would only help people who build
their own images, so offer a separate package variant for IGDv1.

Signed-off-by: Stijn Tintel <stijn@linux-ipv6.be>

Maintainer: NA
Compile tested: x86/64, apu2, master r15679-43ff6e641e
Run tested: x86/64, apu2, master r15679-43ff6e641e, tested UPnP with xbox series x

### DIFF
--- a/net/miniupnpd/Makefile
+++ b/net/miniupnpd/Makefile
@@ -26,7 +26,7 @@ PKG_BUILD_PARALLEL:=1
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/version.mk
 
-define Package/miniupnpd
+define Package/miniupnpd/Default
   SECTION:=net
   CATEGORY:=Network
   DEPENDS:=+iptables +libip4tc +IPV6:libip6tc +IPV6:ip6tables +libuuid +libcap-ng
@@ -35,9 +35,25 @@ define Package/miniupnpd
   URL:=https://miniupnp.tuxfamily.org/
 endef
 
+define Package/miniupnpd
+  $(call Package/miniupnpd/Default)
+  TITLE+= (IGDv2)
+  CONFLICTS:=miniupnpd-igdv1
+  VARIANT:=igdv2
+endef
+
+define Package/miniupnpd-igdv1
+  $(call Package/miniupnpd/Default)
+  TITLE+= (IGDv1)
+  PROVIDES:=miniupnpd
+  VARIANT:=igdv1
+endef
+
 define Package/miniupnpd/conffiles
 /etc/config/upnpd
 endef
+
+Package/miniupnpd-igdv1/conffiles = $(Package/miniupnpd/conffiles)
 
 define Build/Prepare
 	$(call Build/Prepare/Default)
@@ -46,11 +62,14 @@ endef
 
 CONFIGURE_ARGS = \
 	$(if $(CONFIG_IPV6),--ipv6) \
-	--igd2 \
 	--leasefile \
 	--portinuse \
 	--firewall=iptables \
 	--disable-fork
+
+ifeq ($(BUILD_VARIANT),igdv2)
+	CONFIGURE_ARGS += --igd2
+endif
 
 TARGET_CFLAGS += $(FPIC) -flto
 TARGET_LDFLAGS += -Wl,--gc-sections,--as-needed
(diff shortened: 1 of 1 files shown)


neheb:
Wonder why that runtime option is broken.

tievolu:
I applied this patch and successfully built the new `miniupnpd-igdv1` package, and I can confirm that it allows Windows 10 and Xbox to open Teredo ports.

1715173329:
However it seems not so fair.
If the runtime option doesn't work, should we just find out the reason and get it fixed instead of disable anything?

stintel:
Sure, anyone can take it up with upstream. My change will fix it immediately for everyone affected. When upstream is fixed we can remove the 2nd variant again.

1715173329:
yes you're right, but i'm not sure if it is a good idea to set PROVIDES in igdv1 variant...

stintel:
> yes you're right, but i'm not sure if it is a good idea to set PROVIDES in igdv1 variant...

I would say it's required, otherwise it cannot be used with packages that depend on miniupnpd. E.g. installing luci-app-upnp will always require miniupnpd, so people wanting to use the IGDv1 variant can no longer use luci-app-upnp, or any other package that depends on miniupnpd.

1715173329:
Oh I see, thanks for explanation.

pprindeville:
Sorry, I’m not following: why can’t a single binary be built that supports both v1 and v2 with runtime determination, at least at startup via configuration?

stintel:
> Sorry, I’m not following: why can’t a single binary be built that supports both v1 and v2 with runtime determination, at least at startup via configuration?

That would be the ideal solution, but probably non-trivial to implement. Feel free to take it up with upstream. I'm keeping my igdv1 variant locally and have no interest in spending more time on this.

YamazakiRobert:
hello misters, there is any way i can test miniupnpd-igdv1 on x86_64, i have many issues with the current version with games like CoD BO2, WaW and Warzone with multiple pcs on the same network

tievolu:
@YamazakiRobert 

Here's the package I built myself. It's working great for me:

[miniupnpd-igdv1_2.2.1-1_x86_64.ipk.zip](https://github.com/openwrt/packages/files/6326062/miniupnpd-igdv1_2.2.1-1_x86_64.ipk.zip)

(Github wouldn't let me attach the ipk so I had to zip it up.)

YamazakiRobert:
@tievolu 
requires libcap-ng, do yiou have this ipk? thanks in advance

tievolu:
Sorry, I forgot about that dependency. Here you go:

[libcap-ng_0.8.1-1_x86_64.ipk.zip](https://github.com/openwrt/packages/files/6326237/libcap-ng_0.8.1-1_x86_64.ipk.zip)

YamazakiRobert:
edit4: installed on last snapshot working great, tested with Warzone, Destiny 2, World at War and Black Ops 2 with two computers simultaneous 

thanks @tievolu worked great and resolved all my problems

neheb:
Why not.

stintel:
> @stintel is it possible to create a variant with the luci-app-upnp as well?

What do you mean? Both miniupnpd and miniupnpd-igdv1 satisfy luci-app-upnp's dependency on miniupnpd so you can use luci-app-upnp with miniupnpd-igdv1 already.

neilsan1366:
hello doesn't work for me  i have a rt3200 belkin thanks

Neustradamus:
Please read the ticket here: https://github.com/miniupnp/miniupnp/issues/539

PR #14664, opened 2021-02-05
acme: adopt to new behavior of nginx

peter-stadler:
Maintainer: @tohojo 
Run tested: x86_64, x86_64 qemu, master, run it with luci-app-acme

Description: Use nginx-util to add the certificates to the corresponding domain(s) instead of the main config file, which does not contain the ssl* directives anymore.

### DIFF
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -335,7 +335,18 @@ issue_cert()
 		# commit and reload is in post_checks
 	fi
 
-	if [ -e /etc/init.d/nginx ] && [ "$update_nginx" -eq "1" ]; then
+	local nginx_updated
+	nginx_updated=0
+	if command -v nginx-util 2>/dev/null && [ "$update_nginx" -eq "1" ]; then
+		nginx_updated=1
+		for domain in $domains; do
+			nginx-util add_ssl "${domain}" acme "${domain_dir}/fullchain.cer" \
+				"${domain_dir}/${main_domain}.key" || nginx_updated=0
+		done
+		# reload is in post_checks
+	fi
+
+	if [ "$nginx_updated" -eq "0" ] && [ -w /etc/nginx/nginx.conf ] && [ "$update_nginx" -eq "1" ]; then
 		sed -i "s#ssl_certificate\ .*#ssl_certificate ${domain_dir}/fullchain.cer;#g" /etc/nginx/nginx.conf
 		sed -i "s#ssl_certificate_key\ .*#ssl_certificate_key ${domain_dir}/${main_domain}.key;#g" /etc/nginx/nginx.conf
 		# commit and reload is in post_checks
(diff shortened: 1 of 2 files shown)


tohojo:
You're removing the old code completely. Is nginx-util always available if nginx is installed?

peter-stadler:
Yes (it is a dependency of the nginx packages). So if it is not there, the user has likely setup a special `/etc/nginx/nginx.conf` file; do you think we should use the original code then?

tohojo:
peter-stadler <notifications@github.com> writes:

> Yes (it is a dependency of the nginx packages). So if it is not there,
> the user has likely setup a special `/etc/nginx/nginx.conf` file; do
> you think we should use the original code then?

Yeah, I think it would be better to fall back to the old code if
nginx-util is not available...

peter-stadler:
Reinserted the code with a modified condition: If the file `/etc/nginx/nginx.conf` exists (it does not by default) and is writable, change it ...

net/acme/files/run.sh
```diff
@@ -335,7 +335,15 @@ issue_cert()
...
-	if [ -e /etc/init.d/nginx ] && [ "$update_nginx" -eq "1" ]; then
+	if command -v nginx-util 2>/dev/null && [ "$update_nginx" -eq "1" ]; then
+		for domain in $domains; do
+			nginx-util add_ssl "${domain}" acme "${domain_dir}/fullchain.cer" \
+				"${domain_dir}/${main_domain}.key"
+		done
+		# reload is in post_checks
+	fi
+
+	if [ -w /etc/nginx/nginx.conf ] && [ "$update_nginx" -eq "1" ]; then
```
tohojo, line 346: Hmm, this will potentially end up doing both updates; is that a problem? Will nginx-util always succeed, or can it tell if it's not being used? If so, maybe it's better to fall back and only do the config file update if nginx-util is not installed or if it can't update things?

peter-stadler: Although it should not be a problem, the nginx-util only returns 0 if successfully; I changed the code to try editing the `nginx.conf` only when `nginx-util` fails or is not there at all ...

net/acme/files/run.sh
```diff
@@ -335,7 +335,17 @@ issue_cert()
 		# commit and reload is in post_checks
 	fi
 
-	if [ -e /etc/init.d/nginx ] && [ "$update_nginx" -eq "1" ]; then
+	nginx_updated=0
```
tohojo, line 338: This should be declared `local` somewhere :)

peter-stadler: done :-)

tohojo: Declaring a variable local and assigning to it on the same line is a bashism, unfortunately... doesn't work on the busybox shell :(

peter-stadler: Oh, shellcheck.net was not complaining (for #!/bin/ash) ... Changed it now, too. Thanks.

tohojo (approved):
Right, LGTM. Let's wait for the bots to finish their thing before merging...

PR #14665, opened 2021-02-05
uacme: adopt to new behavior of nginx

peter-stadler:
Maintainer: @lucize 
Run NOT tested: I had problems setting it up at all, but the similar changes worked for the acme package, so I suppose the problem is somewhere else on my side and it should work here, too.

Description: Use nginx-util to add the certificates to the corresponding domain(s) instead of the main config file, which does not contain the ssl* directives anymore.

**Update:** Follow the change for acme (PR #14664): reinserted the code with a modified condition: If the file `/etc/nginx/nginx.conf` exists (it does not by default) and is writable, change it as before.

### DIFF
--- a/net/uacme/files/run.sh
+++ b/net/uacme/files/run.sh
@@ -355,7 +355,23 @@ issue_cert()
 	# commit and reload is in post_checks
     fi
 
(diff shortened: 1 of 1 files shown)


lucize:
LGTM

PR #14667, opened 2021-02-06
strongswan: avoid duplicate logging

pprindeville:
Maintainer: @stintel 
Run tested: same, been running this on a production VPN concentrator for months

Description:

Seeing the same logging twice when creating new sessions or authenticating them is confusing.  Authentication is already logged by default, it doesn't need duplication.


karlp (approved):
sure, this seems like the sort of thing a package maintainer should be able to just do... :)

stintel (approved):

PR #14669, opened 2021-02-06
strongswan: since swanctl enables xfrm tunnels, add that dependency

pprindeville:
Maintainer: @stintel 
Run tested: same, been running this on a production VPN concentrator for months

Description:

### DIFF
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=strongswan
 PKG_VERSION:=5.9.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


karlp:
does it need to be a hard dep?

stintel:
We also don't depend on the vti package or the related kmod packages, so I'm inclined to say NACK here. If you're going to use xfrm you should install the requirements.

pprindeville:
> We also don't depend on the vti package or the related kmod packages, so I'm inclined to say NACK here. If you're going to use xfrm you should install the requirements.

The `xfrm` package in turn depends on `+kmod-xfrm-interface`.  Are there other kmod packages you're referring to?

CodeFetch:
xfrm is what people want to use for Strongswan in most cases I guess. I was also very frustrated that it was not available, but I don't remember whether it just didn't work or whether it showed an error.

@pprindeville Is it a hard dependency?

pprindeville:
Nixed.

PR #14676, opened 2021-02-06
perl: fix build errors on macOS

httpstorm:
- The current version of perl 5.28.1 cannot be compiled on macOS 11, because hints/darwin.sh checks only for version 10, and fails on 11:

`*** Unexpected product version 11.2.`
Resolved by adding a check for macOS 11 and newer.

- When multithreaded build is used on macOS 11, a race condition results in failing to locate the current directory (cwd), and producers one of the following errors:

```
Can't locate Errno.pm in @INC
/miniperl not found
```

Resolved by adding Internals::getcwd, and then fallback to the built-in getcwd.

Maintainer: @pprindeville, @Naoir, @neheb
Compile tested: macOS 11.2
Run tested: (mvebu, Linksys WRT3200ACM, OpenWrt 17fa01bb79be12a08f8f7a0beb7f0d73618a6193)

### DIFF
--- a/lang/perl/Makefile
+++ b/lang/perl/Makefile
@@ -11,7 +11,7 @@ include perlver.mk
 
 PKG_NAME:=perl
 PKG_VERSION:=$(PERL_VERSION)
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 PKG_SOURCE_URL:=\
 		https://cpan.metacpan.org/src/5.0 \
--- a/lang/perl/patches/001-macos_11_support.patch
+++ b/lang/perl/patches/001-macos_11_support.patch
@@ -1,0 +1,29 @@
+--- a/hints/darwin.sh
++++ b/hints/darwin.sh
+@@ -301,7 +301,7 @@ case "$osvers" in  # Note: osvers is the
+    # We now use MACOSX_DEPLOYMENT_TARGET, if set, as an override by
+    # capturing its value and adding it to the flags.
+     case "$MACOSX_DEPLOYMENT_TARGET" in
+-    10.*)
++    [1-9][0-9].*)
+       add_macosx_version_min ccflags $MACOSX_DEPLOYMENT_TARGET
+       add_macosx_version_min ldflags $MACOSX_DEPLOYMENT_TARGET
+       ;;
+@@ -313,7 +313,7 @@ case "$osvers" in  # Note: osvers is the
+ 
+ *** Unexpected MACOSX_DEPLOYMENT_TARGET=$MACOSX_DEPLOYMENT_TARGET
+ ***
+-*** Please either set it to 10.something, or to empty.
++*** Please either set it to a valid macOS version number (e.g., 10.15) or to empty.
+ 
+ EOM
+       exit 1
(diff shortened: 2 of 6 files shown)


httpstorm:
@jow- @pprindeville @Naoir @neheb @nbd168

I am afraid the failed checks are not caused by my patch. I was able to reproduce the errors without the patch on a fresh x64 build performed on Ubuntu 20.04. Starting with `make defconfig`, I added everything under Languages, perl as modules. Next I installed them:
`opkg install perl*`

And then as I tried to uninstall perl and all packages which are dependent on it, I got the same error:
```
opkg remove perl --force-removal-of-dependent-packages
…
Collected errors:
 * pkg_run_script: Internal error: perl-www has a NULL tmp_unpack_dir.
 * opkg_remove_pkg: not removing package "perl-www", prerm script failed
echo $?
255
```

Here is the complete log:
https://pastebin.com/0L6Get8F

This seems to be a known issue with opkg:
_opkg fails with "NULL tmp_unpack_dir" when removing multiple packages_
https://dev.archive.openwrt.org/ticket/18320.html

neheb:
Yeah known problem.

httpstorm:
@neheb 
Any chance to integrate my PR, and disregard the fall-positive errors? Without this fix building on macOS is not possible.

neheb:
Up to @pprindeville

pprindeville:
A couple of questions:

- how much of this is tracking upstream changes, 
- why are we seeing so many CI/CD failures with this PR

httpstorm:
> A couple of questions:
> 
> * how much of this is tracking upstream changes,
> * why are we seeing so many CI/CD failures with this PR

@pprindeville 
Hi Philip! 

All patches track upstream changes:
patch 1: `54d41b60822734cf8df543053f7bd270dfded3ed` _Future proof version check_
patch 2: `15f67d146cf1f32504e8a11de3faa2abc0f467cd` _(perl #133951) add Internals::getcwd_
patch 3: `0a4d1779402e8824dd675302ad6a1914ea6ccf23` _(perl #133951) fallback to the built-in getcwd if we can_

Patches 2 and 3 are from upstream commits. Patch 1 only affects building on macOS. I implemented it on my own, and now that I found the actual commit, I noticed it is missing the change on line 304. I will update it to be 100% consistent with the upstream:
https://github.com/Perl/perl5/commit/54d41b60822734cf8df543053f7bd270dfded3ed

The CI/DC failures are caused by a bug in opkg, as described in comment 1 here:
> The problem is just opkg tries to remove a dep ipkg twice if it there is multiple ipkg that depends on it.
https://dev.archive.openwrt.org/ticket/18320.html#comment:1

I compiled the unmodified OpenWRT master for target x86_64 on Ubuntu 20.04, selecting all Perl options to build as modules. Next I installed and then uninstalled them. During the remove phase I reproduced the CI/CD errors.
https://github.com/openwrt/packages/pull/14676#issuecomment-774638496

httpstorm:
@pprindeville 
Updated patch 1 to fully match upstream as described in my previous comment https://github.com/openwrt/packages/pull/14676#issuecomment-775752333.

PR #14683, opened 2021-02-07
ttyd: force enable authentication for login

1715173329:
Maintainer: @tsl0922
Compile tested: ipq40xx, rockchip, x86_64
Run tested: ipq40xx, rockchip, x86_64

Description:
```
Currently, we called `/usr/libexec/login.sh` as login command, but unfortunately the auth
is disabled by default in it[1], and this is really serious as it could be a free "backdoor"
for any spoiler who has conntectd to the router via LAN or wireless.

In my option, it shouldn't be exposed to anyone without auth, so I set the default login
command to `/bin/login`. And for those who really want that, they can do it themselves.

1. `login.sh` adjusts whether use authentication or not from system config named ttylogin,
which is set to disabled by default. See package/base-files/files/bin/config_generate#L243.
```

### DIFF
--- a/utils/ttyd/Makefile
+++ b/utils/ttyd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ttyd
 PKG_VERSION:=1.6.3
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


karlp:
not even going to consider a response from the package maintainer on best path forward?

PR #14686, opened 2021-02-07
wg-installer: add babeld hotplug.d script

PolynomialDivision:
Add a hotplug.d-extension that automatically configures babeld for meshing via wireguard interfaces.

It checks for "add" and "remove" of a wireguard interface with name "wg_*". Depending on the action, it removes it from the babeld config or adds the interface and reloads babeld.

Maintainer: me
Compile tested: trunk
Run tested: trunk

### DIFF
--- a/net/wg-installer/Makefile
+++ b/net/wg-installer/Makefile
@@ -21,6 +21,7 @@ endef
 define Package/wg-installer-server
 	$(call Package/wg-installer/Default)
 	TITLE+= (server)
+	MENU:=1
 	DEPENDS:=+rpcd +uhttpd +uhttpd-mod-ubus +owipcalc
 endef
 
@@ -47,6 +48,16 @@ define Package/wg-installer-server/postinst
 	fi
 endef
 
+define Package/wg-installer-server-hotplug-babeld
+	$(call Package/wg-installer-server)
+	DEPENDS:=wg-installer-server
+endef
+
+define Package/wg-installer-server-hotplug-babeld/install
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/net/
+	$(INSTALL_BIN) ./wg-server/hotplug.d/99-mesh-babeld $(1)/etc/hotplug.d/net/99-mesh-babeld
+endef
+
 define Package/wg-installer-client
 	$(call Package/wg-installer/Default)
 	TITLE+= (client)
@@ -66,4 +77,5 @@ define Package/wg-installer-client/install
 endef
 
 $(eval $(call BuildPackage,wg-installer-server))
+$(eval $(call BuildPackage,wg-installer-server-hotplug-babeld))
 $(eval $(call BuildPackage,wg-installer-client))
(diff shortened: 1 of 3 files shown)


net/wg-installer/Makefile
```diff
@@ -40,6 +41,16 @@ define Package/wg-installer-server/install
...
 endef
 
+define Package/wg-installer-server-hotplug-babeld
+	$(call Package/wg-installer-server)
+	DEPENDS:=wg-installer-server
+endef
+
+define Package/wg-installer-server-hotplug-babeld/install
+	$(INSTALL_DIR) $(1)/etc/config/hotplug.d/
+	$(INSTALL_BIN) ./wg-server/hotplug.d/99-mesh-babeld $(1)/etc/config/hotplug.d/99-mesh-babeld
```
dangowrt, line 51: This path looks wrong

PolynomialDivision:
Maybe in the future we can just do: https://github.com/jech/babeld/issues/76

net/wg-installer/Makefile
PolynomialDivision: Yep.  Thanks. I'm still in WIP-Phase. ;)

spolack:
Wouldnt it be a way more pretty and universal approach to add regex support to babel uci configuration - And then reload babel once the interfaces have changed?

PolynomialDivision:
> Wouldnt it be a way more pretty and universal approach to add regex support to babel uci configuration - And then reload babel once the interfaces have changed?
> 
> Babel is imho not a strict dependency of the wireguard tunnel setup. You could also use ospf3 or other IGPs

That is why I added this as extra plugin, to make it optional. So people can start adding hotplug-scripts for IGPs. ;)

PolynomialDivision:
image

spolack:
> As u ca read in my post above, I try to find a solution in babeld also. 
I'm not talking exactly about babeld but about babel UCI, which could then handle the reload etc.

PolynomialDivision:
> > As u ca read in my post above, I try to find a solution in babeld also.
> 
> I'm not talking exactly about babeld but about babel UCI, which could then handle the reload etc.

You mean probably here?
https://github.com/openwrt-routing/packages/blob/master/babeld/files/babeld.init
We could probably add [here](https://github.com/openwrt-routing/packages/blob/master/babeld/files/babeld.init#L115-L130) some extra bool for regex matching.
@spolack Any pointers to code or scripts that u prefer?

Maybe @kerneis , @jech and have seomthing to say to this.

PolynomialDivision:
@spolack I tested everything and it works. Since it is a just a plugin, and maybe more plugins will follow, I will merge this. :)
Meanwhile I will look again at regex support in babeld, too.

kerneis:
@spolack 
> Wouldnt it be a way more pretty and universal approach to add regex support to babel uci configuration

That could be done (assuming there is a way to list all interfaces in UCI).

> And then reload babel once the interfaces have changed?

But how do you do that in UCI? Is there a hotplug mechanism to register some events?

PolynomialDivision:
You could go through `/sys/class/net`. But there is also some `network.sh` library which brings a lot of functionality.

PR #14699, opened 2021-02-09
vpnc-scripts: bump to use latest version

lipnitsk:
Switch to a version of the script maintained by OpenConnect.

Maintainer: @nmav 
Compile tested: ramips, mt7621


lipnitsk:
@nmav I see that you have made some fixes to the script in the past - do you know if they were submitted upstream or do we need to patch it still?

nmav:
I do not think it is a good idea to have the generic vpnc script in openwrt. The script that is present contains only the relevant items for openwrt; there is no need in my opinion to include macos or systemd support to openwrt.

lipnitsk:
No known bugs, but it's always good practice to stay up to date with maintained packages to get generic fixes and improvements, and the upstream script mentions OpenWrt, implying support. The only downside is that it is larger and may be missing some changes you have made in the past. If we want to just have our own fork for good reasons - that is fine with me - just wanted to get your opinion on this.

nmav:
> No known bugs, but it's always good practice to stay up to date with maintained packages to get generic fixes and improvements, and the upstream script mentions OpenWrt, implying support. The only downside is that it is larger and may be missing some changes you have made in the past. If we want to just have our own fork for good reasons - that is fine with me - just wanted to get your opinion on this.

I'm not sure if all the changes done are upstream. The reason I used a separate script was that it was too hard to debug an issue with the upstream script as it was too complicated (and the autodetection could take the wrong path). I agree that we should stick to the upstream, but we may want to separate the system part (i.e., openwrt-specific) from the rest first so we have something simple to use and debug. I would like to do that upstream, but it is unlikely to do it any time soon, is that something you'd like to try?

neheb:
What's the status here?

lipnitsk:
Eh, let's close it for now - I don't have time or ability to test any changes or work with upstream on this at this point in time.

PR #14709, opened 2021-02-10
https-dns-proxy: support for force DNS/DNS hijacking

stangri:
Maintainer: me
Compile tested: ramips, er-x, 19.07.6
Run tested:  ramips, er-x, 19.07.6, start/stop/dns hijack

Description: This introduces support for DNS Hijacking. 

PS. Please grant merge rights.

### DIFF
--- a/net/https-dns-proxy/Makefile
+++ b/net/https-dns-proxy/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=https-dns-proxy
 PKG_VERSION:=2021-01-17
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/aarond10/https_dns_proxy
@@ -41,7 +41,7 @@ define Package/https-dns-proxy/install
 	$(INSTALL_DIR) $(1)/usr/sbin $(1)/etc/init.d ${1}/etc/config
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/https_dns_proxy $(1)/usr/sbin/https-dns-proxy
 	$(INSTALL_BIN) ./files/https-dns-proxy.init $(1)/etc/init.d/https-dns-proxy
-	sed -i "s|^\(PKG_VERSION\).*|\1='$(PKG_VERSION)-$(PKG_RELEASE)'|" $(1)/etc/init.d/https-dns-proxy
+	$(SED) "s|^\(PKG_VERSION\).*|\1='$(PKG_VERSION)-$(PKG_RELEASE)'|" $(1)/etc/init.d/https-dns-proxy
 	$(INSTALL_CONF) ./files/https-dns-proxy.config $(1)/etc/config/https-dns-proxy
 endef
 
--- a/net/https-dns-proxy/files/https-dns-proxy.config
+++ b/net/https-dns-proxy/files/https-dns-proxy.config
@@ -1,18 +1,19 @@
 config main 'config'
 	option update_dnsmasq_config '*'
-
-config https-dns-proxy
-	option bootstrap_dns '8.8.8.8,8.8.4.4'
-	option resolver_url 'https://dns.google/dns-query'
-	option listen_addr '127.0.0.1'
-	option listen_port '5053'
-	option user 'nobody'
-	option group 'nogroup'
+	option force_dns '1'
 
 config https-dns-proxy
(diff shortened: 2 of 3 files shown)


1715173329:
However, in my option, both hijacking and force hijacking shouldn't be enabled by default.

stangri:
> However, in my option, both hijacking and force hijacking shouldn't be enabled by default.

Could you please clarify what do you mean by "both hijacking and force hijacking"? 

I saw a post on the forum recently where someone had to manually configure DNS hijacking after installing the https-dns-proxy and the luci app and realized that this package lacks the force dns feature. My thinking is that if people are running DoH proxy on their router, they would probably want to route all DNS requests thru it, hence force DNS being enabled by default.

1715173329:
> Could you please clarify what do you mean by "both hijacking and force hijacking"?

Apologize, I didn't look at them carefully.
Hijacking I meant, is updating dnsmasq config (upstream servers).
And, in my option, could it be better if we move the firewall settings to dnsmasq? This would be useful.

> I saw a post on the forum recently where someone had to manually configure DNS hijacking

Okay, but not all users wanna this. Could we just make it optional, and disable it by default? Then if someone needs this, they could enable it themself.

Also, in the init:
```bash
config_get dnsmasqConfig	'config' 'update_dnsmasq_config' '*'
```
This means hijacking cannot be disabled, though someone make it as none but it still means all.

stangri:
> Hijacking I meant, is updating dnsmasq config (upstream servers).
It's been like that for quite some time, I don't think changing the default behaviour now would be practical. 

> And, in my option, could it be better if we move the firewall settings to dnsmasq? This would be useful.
Please elaborate. Do you suggest making DNS hijacking PROCD firewall setup a part of the dnsmasq init-script and exposing this setting in luci?

> Okay, but not all users wanna this. Could we just make it optional, and disable it by default? Then if someone needs this, they could enable it themself.
If users largly have no problem with dnsmasq auto-update, what's the argument for making dns hijacking disabled by default? I strongly believe that people setting up DoH proxy on their router wish to have DNS requests from all their devices processed by DoH proxy, hence the defaults I'm suggesting. Can you quantify "not all users wanna this" to change my mind?

> Also, in the init:
> 
> ```shell
> config_get dnsmasqConfig	'config' 'update_dnsmasq_config' '*'
> ```
> 
> This means hijacking cannot be disabled, though someone make it as none but it still means all.

It's covered in README and the luci-app handles it properly, to disable it you need to set value to `'-'`.

1715173329:
> Do you suggest making DNS hijacking PROCD firewall setup a part of the dnsmasq init-script and exposing this setting in luci?

Yes, but PROCD is not forced.

stangri:
> > Do you suggest making DNS hijacking PROCD firewall setup a part of the dnsmasq init-script and exposing this setting in luci?
> 
> Yes, but PROCD is not forced.

Before I subimtted this PR, I've tested two different packages having this same PROCD firewall data and there was no issue with that. If at some point, dnsmasq has the same implemented, it will not affect this package. I do agree it's a great idea to have dns hijacking implemented in openwrt's dnsmasq packages tho. I also agree that if it's implemented in dnsmasq itself, it should be disabled by default. In this package, I believe it should be enabled by default.

1715173329:
> If at some point, dnsmasq has the same DNS hijacking functionality implemented, it should not affect this package. I do agree it's a great idea to have DNS hijacking implemented in openwrt's dnsmasq packages tho. I also agree that if it's implemented in dnsmasq itself, it should be disabled by default. In this package, I believe it should be enabled by default.

However, it's up to you, but it would be great if we have this function in DNSMasq.

stangri:
> > If at some point, dnsmasq has the same DNS hijacking functionality implemented, it should not affect this package. I do agree it's a great idea to have DNS hijacking implemented in openwrt's dnsmasq packages tho. I also agree that if it's implemented in dnsmasq itself, it should be disabled by default. In this package, I believe it should be enabled by default.
> 
> However, it's up to you, but it would be great if we have this function in DNSMasq.

Sure, I can also have a look at implementing the same in DNSMASQ. This can be merged then?

1715173329:
> Sure, I can also have a look at implementing the same in DNSMASQ.

Great, thanks so much!

> This can be merged then?

Ping @neheb.

stangri:
@neheb, could you please cherry-pick this for 21.02? Or should I send a separate PR?

neheb:
Send a PR. I can't spend time on the packages feed until I rebuild my NAS...

stangri:
> Now all requests are intercepted. What for?

To force router's (encrypted) resolution on LAN.

PR #14711, opened 2021-02-10
strongswan: make the include's in the .conf files persistent

pprindeville:
Maintainer: @stintel 
Run tested: same, running on test VM

Description:

Having scripts diddle user written config files seems potentially
dangerous.  Plus there's really no downside to including some
empty files.  Best to just make the includes be permanent.


feckert:
@pprindeville Can you please write in the commit wha we need this? So that I can understand later why we have done this.

pprindeville:
> Can you please write in the commit wha we need this? So that I can understand later why we have done this.

@feckert Done.

luizluca:
> an you please write in the commit why we need this?

If init.d script changes a file in /etc, it will always be considered "user modified" when you update/reinstall the same package.
The result is that, when you update strongswan, you'll get the same old config (without the includes) in a foo-opkg file. You have multiple issues with that behavior. 1) Useless -opkg files will always exists after the strongswan is reinstalled or installed after a flash update with backup 2) an strongwan update will never be able to replace that config file without a migration script, even if the user never changed it. And never is a long time. I imagine that, in a matter of years, an update will require some change in config files.

As I user, I think that a service start should not change /etc files whenever possible.

pprindeville:
> As I user, I think that a service start should not change /etc files whenever possible.

Well, other than uci-defaults, yeah... agreed.

luizluca (commented):
It will avoid replacing the same /etc files on every service start.
It will avoid generating some useless -opkg files.

I'm not sure but I guess the second time it is installed, opkg will consider an existing /etc/ file as untouched if it matches the config file hash from the package. It would be nice to have a cleanup post install script to remove -opkg file if the config file hash matches. Something like:

( opkg list-changed-conffiles | grep -qx /etc/ipsec.conf ) || ( [ -e /etc/ipsec.conf-opkg ] && rm /etc/ipsec.conf-opkg )

I'll not mark as approved as I'm not involved in strongswan dev. Anyway, I'll send my :+1:

pprindeville:
@neheb Can you have a look?

pprindeville:
There are other commits I'm waiting to submit that require this first.

pprindeville:
@luizluca Would the following work?

```
[ -n "$(opkg list-changed-conffiles /etc/ipsec.conf)" ] || rm -f /etc/ipsec.conf-opkg
```

luizluca:
> opkg list-changed-conffiles /etc/ipsec.conf

It will not work for 19.07:

```
# opkg list-changed-conffiles | grep -x /etc/ipsec.conf; echo $?
/etc/ipsec.conf
0
# opkg list-changed-conffiles /etc/ipsec.conf; echo $?
0
# opkg list-changed-conffiles /etc/ipsec.conf.xxx; echo $?
0
```

But I didn't test if snapshot opkg works differently. If it works as expected, just keep in mind that it will not work if backported.

Regards,

pprindeville:
> > opkg list-changed-conffiles /etc/ipsec.conf
> 
> It will not work for 19.07:
> 
> ```
> # opkg list-changed-conffiles | grep -x /etc/ipsec.conf; echo $?
> /etc/ipsec.conf
> 0
> # opkg list-changed-conffiles /etc/ipsec.conf; echo $?
> 0
> # opkg list-changed-conffiles /etc/ipsec.conf.xxx; echo $?
> 0
> ```
> 
> But I didn't test if snapshot opkg works differently. If it works as expected, just keep in mind that it will not work if backported.
> 
> Regards,

Okay, maybe I was reading the `help` output wrong.

pprindeville:
@luizluca Amended.  Please review.

pprindeville:
@stintel If you're back from vacation and have some cycles, can you please try this trivial PR?

Neustradamus:
@ all : Any news on this PR?

pprindeville:
> @ all : Any news on this PR?

Pinged @stintel above.  Waiting to hear back.

It's blocking for other commits.

I've also volunteered a VM for testing if that makes things easier for Stijn to reproduce the issue he was seeing.

net/strongswan/Makefile
```diff
@@ -502,9 +504,20 @@ endef
 define Package/strongswan-ipsec/install
 	$(INSTALL_DIR) $(1)/etc/ $(1)/usr/sbin
 	$(INSTALL_CONF) $(PKG_INSTALL_DIR)/etc/ipsec.conf $(1)/etc/
+	echo -e "\ninclude /var/ipsec/ipsec.conf" >> $(1)/etc/ipsec.conf
 	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/sbin/ipsec $(1)/usr/sbin/
 endef
 
+define Package/strongswan-ipsec/postinst
```
stintel, line 511: This is not mentioned at all in the commit message. It's also completely unrelated to the changes in this commit, so this must go in a different commit, with the commit message explaining what/why.

stintel (changes requested):
Aside from the other comment, I'm not sure I like this change. It feels ugly/hacky to me. Did someone actually run into a problem that this change aims to fix?

stintel:
> Should be able to spend some time on that, this weekend.

I apologize. Chores and social life got in the way, again.

stintel:
Here's a thought: both the current situation and the proposed changes in this PR seem suboptimal and/or hacky. The reason we are in this situation is because I demanded that UCI and configuring /etc/ipsec.conf and friends manually should always work, also the combination of the two, allowing people to migrate connections one by one. I still believe we should support both, and try to avoid breaking existing setups as much as we can. That said, maybe there's a better solution to achieve this.

luizluca:
> Here's a thought: both the current situation and the proposed changes in this PR seem suboptimal and/or hacky. The reason we are in this situation is because I demanded that UCI and configuring /etc/ipsec.conf and friends manually should always work, also the combination of the two, allowing people to migrate connections one by one. I still believe we should support both, and try to avoid breaking existing setups as much as we can. That said, maybe there's a better solution to achieve this.

Sorry but I can't see in what way this PR prevents what you ask. Indeed, it makes it even easier to use /etc files without uci as include is added during build and not during service start. If you want to remove uci confs, comment or remove the include and, after this PR, init script will not add it back. Before this PR, init script could not tell if you removed the include or if it was still not added.

The only temporary hack I see is the -opkg deletion. If the user didn't touched ipsec.conf, -opkg is useless as it is a copy of an old untouched conf. It only exists because init script was changing that file.

net/strongswan/Makefile
Thermi: I'm with @stintel on that one.

Thermi:
Rest looks fine.

net/strongswan/files/ipsec.init
```diff
@@ -267,6 +240,14 @@ config_remote() {
 	ipsec_xappend ""
 }
 
+do_preamble() {
```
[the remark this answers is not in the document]
pprindeville: The code is as it was before, it's just being moved around.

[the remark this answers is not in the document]
Thermi: Yes, but does it make a valid file?

net/strongswan/Makefile
pprindeville: Okay, hopefully @stintel and @luizluca are okay with the added description in the new commit message.

net/strongswan/files/ipsec.init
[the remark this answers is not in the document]
pprindeville: It seems to.  Starting/restarting ipsec service I don't see any noise in the logs about rejected config files.

net/strongswan/Makefile
Thermi: Looks okay. It's legacy anyway with that.

Thermi (approved):

PR #14713, opened 2021-02-10
yq: Update to 4.5.1

1715173329:
Maintainer: me
Compile tested: rockchip, x86
Run tested: rockchip nanopi-r2s

Description:
Changelog: https://github.com/mikefarah/yq/releases/tag/v4.5.1


BKPepe:
Maybe it would be interesting to see added test script in a separate commit, but this is fine as it is. LGTM & merging.

PR #14716, opened 2021-02-10
libgpiod: update to 1.6.3

mhei:
Maintainer: me
Compile tested: mxs
Run tested: mxs

### DIFF
--- a/libs/libgpiod/Makefile
+++ b/libs/libgpiod/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libgpiod
-PKG_VERSION:=1.4.4
+PKG_VERSION:=1.6.3
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@KERNEL/software/libs/libgpiod/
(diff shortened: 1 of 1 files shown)


neheb:
Was this the project that needed kernel 5.5 and above or was that something else?

mhei:
Yes, you're right. At least, this was my understanding, too. I'm just wondering, that it actually compiles without warnings and also runs on the target with kernel v5.4.94.
I'd mark this PR as draft for now, I'd study the sources once again just be sure that this is expected behavior...

neheb:
There's also version 1.4.5. Could update to that

john-tho:
> it actually compiles without warnings and also runs

I saw that libgpiod 1.6.2 includes a commit that allows compiling with older kernel headers:
https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/commit/?h=v1.6.x&id=325873cf2e9b85550e296a00d44c5a56aaf475bd
```
+ * These are symbols first available in linux v5.5. In order to allow to build
+ * libgpiod with kernel headers as early as v4.8 we redefine them here for pre
+ * v5.5 build environments. Certain features will not work on older kernels.
```
1.6.3 and 1.4.5 are the current versions now

neheb:
Makes sense.

mhei:
Will have a look.

mhei:
Compile and run tested on mxs platform with both kernel 5.4 and 5.10.

neheb (approved):

PR #14717, opened 2021-02-11
adguardhome: bump to 0.105.1

dobo90:
Maintainer: me
Compile tested: aarch64_generic, NanoPi R2S, OpenWrt SNAPSHOT r15730
Run tested: aarch64_generic, NanoPi R2S, OpenWrt SNAPSHOT r15730

* Full changelog available at:
  * https://github.com/AdguardTeam/AdGuardHome/releases/tag/v0.105.0
  * https://github.com/AdguardTeam/AdGuardHome/releases/tag/v0.105.1
* Add node-yarn/host dependency as it is needed since [1].
* Adjust LDFLAGS to the new ones introduced in [2].
* Invoke targets from make instead of manually running npm and yarn.
* Replace GO_PKG_EXCLUDES with GO_PKG_BUILD_PKG as our intention is to
  build only one specific package (a cosmetic change).

[1]: https://github.com/AdguardTeam/AdGuardHome/commit/5e20ac7ed5de861ea5c945d1775f75312d62b69f#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R140
[2]: https://github.com/AdguardTeam/AdGuardHome/commit/0d67aa251d18f8ab47cfd90072e9d0387dff4224#diff-82ef468ec5547f1ed424776755a7f87dfec4eba9838d2c2ac02c9881bb67d737R60

### DIFF
--- a/net/adguardhome/Makefile
+++ b/net/adguardhome/Makefile
@@ -6,25 +6,30 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=adguardhome
-PKG_VERSION:=0.104.3
+PKG_VERSION:=0.105.1
 PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_VERSION:=v$(PKG_VERSION)
 PKG_SOURCE_URL:=https://github.com/AdguardTeam/AdGuardHome
-PKG_MIRROR_HASH:=9051c08ebefccd918cad9b487d2d3b2c4b276ac71f16706c2ae8ee2a37ba9d03
+PKG_MIRROR_HASH:=745024632c9bd462f962abf6828b971fec6fc9f6154a7a05f6ed1eaa81882c66
 
 PKG_LICENSE:=GPL-3.0-only
 PKG_LICENSE_FILES:=LICENSE.txt
 PKG_MAINTAINER:=Dobroslaw Kijowski <dobo90@gmail.com>
 
-PKG_BUILD_DEPENDS:=golang/host node/host packr/host
+PKG_BUILD_DEPENDS:=golang/host node/host node-yarn/host packr/host
 PKG_BUILD_PARALLEL:=1
 PKG_USE_MIPS16:=0
 
 GO_PKG:=github.com/AdguardTeam/AdGuardHome
-GO_PKG_EXCLUDES:=dhcpd/standalone
-GO_PKG_LDFLAGS_X:=main.version=$(PKG_VERSION) main.channel=release main.goarm=$(GO_ARM)
+GO_PKG_BUILD_PKG:=github.com/AdguardTeam/AdGuardHome
+
+AGH_VERSION_PKG:=github.com/AdguardTeam/AdGuardHome/internal/version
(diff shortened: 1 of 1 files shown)


dobo90:
Please don't merge it yet. It seems to be running fine but I've noticed that Makefile uses also yarn. Let me run this software for more time and try to dig if yarn is really needed for compilation.

dobo90:
I did some digging and AdGuardHome states that yarn is needed for compilation. It serves [redesigned incomplete](https://github.com/AdguardTeam/AdGuardHome/issues/2554) interface on 3001 port.

To make this work with yarn, there are two things missing:
1) Make host version of node compile with icu
2) Add yarn package

For now I'll push those missing functionality into this pull request. When it passes CI I'll create one pull request (for node fix) and one issue for including node-yarn package (in order to keep original authorship).

dobo90:
Ok, node-yarn fails to build for arm_cortex-a9_vfpv3-d16, arm_cortex-a15_neon-vfpv4 and i386_pentium-mmx targets. I'll try to figure it out. If not I'll reach for @nxhack for help.

nxhack:
@dobo90 
To build node.js 32bit targets, multilib is required in the host development environment.
However, multilib is not installed in the current OpenWrt CI and buildbot environments.
Therefore, building the 32-bit target will fail.

nxhack:
@dobo90 
I agree that icu should be enabled for node.js host build. I'm going to change it after testing it upstream.

dobo90:
Hmm, I don't understand one thing: why node is compiled for a target architecture when I specify only node/host and node-yarn/host dependency.
The last revision I've pushed contains host-only Makefile of node-yarn package and CI is fine with it (as apparently node is compiled only for host architecture).

@nxhack,
Are you fine with that approach, that for now we can ship node-yarn host only package recipe? If not, please give me some clues how the problem can be solved. Thanks!

nxhack:
The above limitation (lack of multilib) can be avoided by restricting yarn to host build only. However, due to the above limitation, the node module required by adguardhome will not run because the 32bit version of node.js itself is not built.

Of course, if you install multilib and then build OpenWrt by yourself, there is no problem.

nxhack:
@dobo90 

I tried to build it in my environment.

Am I correct in my understanding that the node.js environment is only needed to build adguardhome itself, and not on the target machien?

If so, I think it's a good idea to limit yarn to host builds.

If that's okay, I'll send a pull request for each package related to node.js.

dobo90:
> @dobo90
> 
> I tried to build it in my environment.
> 
> Am I correct in my understanding that the node.js environment is only needed to build adguardhome itself, and not on the target machien?
> 
> If so, I think it's a good idea to limit yarn to host builds.
> 
> If that's okay, I'll send a pull request for each package related to node.js.

That's correct. Adguardhome needs nodejs and yarn only to build some frontend files (only on the host machine, not the target).
Thanks for your assistance. Please let me know when pull requests are merged so I can rebase this on top of them. Have a great day!

nxhack:
@dobo90 

I've sent the necessary pull requests.

dobo90:
It's ready for review. Friendly poke: @neheb, @jefferyto.

1715173329 (approved):

PR #14729, opened 2021-02-13
openpgm: Add Pragmatic General Multicast library

ye-holmes:
OpenPGM is a library implementing the PGM reliable multicast
network protocol; The famous messaging library ZMQ has an
optional dependency on OpenPGM, with OpenPGM enabled, we can
foster the development of multicast network applications.

Signed-off-by: Ye Holmes <yeholmes@outlook.com>

Description:
The PR is intended to add new package under libs directory, OpenPGM.
The downloaded source tarball is named as openpgm-release-5-3-128.tar.gz.
Two more patches since the release tarball are also included.

### DIFF
--- a/libs/openpgm/Makefile
+++ b/libs/openpgm/Makefile
@@ -1,0 +1,65 @@
+#
+# Copyright (C) 2021 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=openpgm
+PKG_VERSION:=5.3.128
+PKG_RELEASE:=1
+
+PKG_SOURCE_URL_FILE:=release-5-3-128
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_SOURCE_URL_FILE).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/steve-o/openpgm/tar.gz
+PKG_HASH:=8d707ef8dda45f4a7bc91016d7f2fed6a418637185d76c7ab30b306499c6d393
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_SOURCE_URL_FILE)
+
+PKG_FIXUP:=autoreconf
+MAKE_PATH:=openpgm/pgm
+PKG_AUTOMAKE_PATHS:=$(MAKE_PATH)
+
+PKG_MAINTAINER:=Ye Holmes <yeholmes@outlook.com>
+PKG_LICENSE:=LGPL-2.1-or-later
+PKG_LICENSE_FILES:=$(MAKE_PATH)/LICENSE
+PKG_CPE_ID:=cpe:/a:openpgm:openpgm
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/openpgm
+  TITLE:=OpenPGM, an implementation of the PGM protocol
+  URL:=http://openpgm.googlecode.com/
+  SECTION:=libs
+  CATEGORY:=Libraries
+endef
+
+define Package/openpgm/description
+  OpenPGM is a library implementing the PGM reliable multicast
+  network protocol. For more information about OpenPGM, see:
+  http://openpgm.googlecode.com/
+endef
+
+CONFIGURE_VARS += ac_cv_file__proc_cpuinfo=yes \
+    ac_cv_file__dev_rtc=no ac_cv_file__dev_hpet=no
+CONFIGURE_ARGS += --enable-static=no --enable-shared=yes
+
+define Build/Configure
+	$(call Build/Configure/Default,,,$(MAKE_PATH))
+endef
+
+define Build/InstallDev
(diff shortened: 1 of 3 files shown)


libs/openpgm/Makefile
```diff
@@ -0,0 +1,62 @@
...
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=openpgm
+PKG_VERSION:=5.3.128
+PKG_RELEASE:=1
+
+PKG_SOURCE:=release-5-3-128.tar.gz
+PKG_SOURCE_URL:=https://github.com/steve-o/openpgm/archive
```
neheb, line 15: Use codeload

libs/openpgm/Makefile
```diff
@@ -0,0 +1,62 @@
...
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=openpgm
+PKG_VERSION:=5.3.128
+PKG_RELEASE:=1
+
+PKG_SOURCE:=release-5-3-128.tar.gz
```
neheb, line 14: this tarball name can conflict with others.

libs/openpgm/Makefile
```diff
@@ -0,0 +1,62 @@
...
+PKG_SOURCE_URL:=https://github.com/steve-o/openpgm/archive
+PKG_HASH:=8d707ef8dda45f4a7bc91016d7f2fed6a418637185d76c7ab30b306499c6d393
+
+PKG_MAINTAINER:=Ye Holmes <yeholmes@outlook.com>
+PKG_LICENSE:=LGPL-2.1-or-later
+PKG_LICENSE_FILES:=LICENSE
+PKG_CPE_ID:=cpe:/a:openpgm:openpgm
+
+PKG_FIXUP:=autoreconf
+PKG_BUILD_DIR:=$(BUILD_DIR)/openpgm-release-5-3-128
```
neheb, line 24: Place below PKG_HASH.

libs/openpgm/Makefile
```diff
@@ -0,0 +1,62 @@
...
+PKG_SOURCE:=release-5-3-128.tar.gz
+PKG_SOURCE_URL:=https://github.com/steve-o/openpgm/archive
+PKG_HASH:=8d707ef8dda45f4a7bc91016d7f2fed6a418637185d76c7ab30b306499c6d393
+
+PKG_MAINTAINER:=Ye Holmes <yeholmes@outlook.com>
+PKG_LICENSE:=LGPL-2.1-or-later
+PKG_LICENSE_FILES:=LICENSE
+PKG_CPE_ID:=cpe:/a:openpgm:openpgm
+
+PKG_FIXUP:=autoreconf
```
neheb, line 23: PKG_INSTALL and PKG_BUILD_PARALLEL look good here. That way you can get rid of those nasty paths below.

neheb:
Run through make package/openpgm/refresh

libs/openpgm/Makefile
ye-holmes: I tried to download source tarball from codeload, but the downloaded file names can only be `master` or `commit-ID`, without any `openpgm` prefix. For example, the latest release tarball has a corresponding codeload [URL](https://codeload.github.com/steve-o/openpgm/tar.gz/cd3456a5c437bb542d9af2a37778ec7941dc05a5), the saved file name is the `commit-ID`. I think setting PKG_SOURCE_PROTO to git, might be able to fetch a source with openpgm prefix.

libs/openpgm/Makefile
ye-holmes: Yes, I don't like the tarball name either.

libs/openpgm/Makefile
neheb: I see the issue. The URL is https://codeload.github.com/steve-o/openpgm/tar.gz/release-5-3-128  but that does not match the version.

libs/openpgm/Makefile
ye-holmes: The post-move after tarball extraction is not good, however, by looking into include/package*.mk, I found another method to avoid the post-move, and handling the subdirectory nicely, by defining `MAKE_PATH` variable. I will post the patch shortly.

libs/openpgm/Makefile
ye-holmes: I've updated the PR again, using the codeload URL, the tarball is named as openpgm-release-5-3-128.tar.gz by download.pl, via setting variable `PKG_SOURCE_URL_FILE`. Is it necessary to change PKG_VERSION from `5.3.128` to `5-3-128`?

ye-holmes:
I've refreshed the two additional patches. The official release source tarball is fetched from this [URL](https://codeload.github.com/steve-o/openpgm/tar.gz/release-5-3-128), with variable `PKG_SOURCE_URL_FILE`, the downloaded tarball is named as openpgm-release-5-3-128.tar.gz. To compile openpgm under subdirectory **openpgm/pgm**, `Build/Configure` has been redefined, adding `MAKE_PATH` variable; _package-defaults.mk_ makes use of `MAKE_PATH` to determine the correct build subdirectory.

PR #14737, opened 2021-02-14
ddns-scripts: add * to DNS_CHARSET list

psych0d0g:
Compile tested: none - just edited the SH in my local FS
Run tested:
Model：PC Engines apu2
Architecture：AMD GX-412TC SOC
Firmware Version：OpenWrt SNAPSHOT r14528-7190fb2da4

Description:
add wildcard domain support to ddns scripts.
nothing more is needed to get this to work with * in domain names

### DIFF
--- a/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh
+++ b/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh
@@ -71,7 +71,7 @@ IPV6_REGEX="\(\([0-9A-Fa-f]\{1,4\}:\)\{1,\}\)\(\([0-9A-Fa-f]\{1,4\}\)\{0,1\}\)\(
 SHELL_ESCAPE="[\"\'\`\$\!();><{}?|\[\]\*\\\\]"
 
 # dns character set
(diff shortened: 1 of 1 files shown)


neheb:
ping @feckert

looi:
Can this be merged? It is still a problem

feckert:
My problem here is that the check is also used for other things.
And there is definitely no '*' allowed in dns.
https://github.com/openwrt/packages/blob/816c41b6fcabfefea36f98568477cd06e74be386/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh#L257-L261

If so, this only applies to the domain?

Maybe it would be better to work with a boolean uci option?
If we set this boolean option, then the domain will be prefixed with '*'.
I have not verified whether this is possible.

looi:
How about adding another `DNS_CHARSET_DOMAIN` that only applies to the domain?

looi:
I have created https://github.com/openwrt/packages/pull/15657

feckert:
Superseded by https://github.com/openwrt/packages/commit/93344aafa065ebfe63330e60f7d02ecdcf39c9b8

PR #14742, opened 2021-02-14
net/dnscrypt-proxy: update the CSV list of resolvers

jedisct1:
Maintainer: @thess
Compile tested: -
Run tested: -

Description:

Update the CSV file with the list of resolvers for dnscrypt-proxy v1.

The previous list was very out of date.

An always up-to-date v1-compatible list is available at: https://download.dnscrypt.info/dnscrypt-resolvers/v1/

Also use different default resolvers since the previous ones don't exist any longer.


neheb:
Needs a PKG_RELEASE bump.

neheb:
Just squash it. New commit is making stuff fail.

PR #14744, opened 2021-02-14
python-cryptography: Update to 3.3.2

jefferyto:
Maintainer: me, @commodo 
Compile tested: armvirt-32, 2021-02-14 snapshot sdk
Run tested: none

Description:
This fixes CVE-2020-36242 (buffer overflows caused by integer overflow in OpenSSL).

### DIFF
--- a/lang/python/python-cryptography/Makefile
+++ b/lang/python/python-cryptography/Makefile
@@ -8,11 +8,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-cryptography
-PKG_VERSION:=3.3.1
+PKG_VERSION:=3.3.2
 PKG_RELEASE:=1
 
 PYPI_NAME:=cryptography
-PKG_HASH:=7e177e4bea2de937a584b13645cab32f25e3d96fc0bc4a4cf99c27dc77682be6
+PKG_HASH:=5a60d3780149e13b7a6ff7ad6526b38846354d11a15e21068e57073e29e19bed
 
 PKG_LICENSE:=Apache-2.0 BSD-3-Clause
 PKG_LICENSE_FILES:=LICENSE.APACHE LICENSE.BSD
@@ -20,7 +20,7 @@ PKG_MAINTAINER:=Jeffery To <jeffery.to@gmail.com>, Alexandru Ardelean <ardeleana
 
 PKG_BUILD_DEPENDS:=libffi/host
 
-HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=cffi  # cffi>=1.8,!=1.11.3
+HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=cffi  # cffi>=1.12
 
 include ../pypi.mk
 include $(INCLUDE_DIR)/package.mk


jefferyto:
Just wanted to add some notes for posterity (I'm sure @commodo is aware of these): Starting with 3.4, cryptography [requires a Rust toolchain](https://cryptography.io/en/latest/changelog.html#v3-4) to build. There is a way to [bypass this requirement](https://cryptography.io/en/latest/faq.html#installing-cryptography-fails-with-error-can-not-find-rust-compiler) for 3.4 but this will be removed in the next major release so I don't think it is worth updating to 3.4 (with this bypass) for now.

There is some discussion in upstream of a 3.3 LTS release (pyca/cryptography#5799) but this hasn't been decided yet, and even if it happens it would only be supported until the end of 2021.

I suppose the future of this package depends on the success of #13916 (or some other effort to bring Rust into the packages feed).

commodo:
I started update on 3.4.1 last week, but I stumbled over this Rust thing.
I patched it so that the CRYPTOGRAPHY_DONT_BUILD_RUST env-var works as intended.

But I didn't manage to finish this, or test it real-quick.

If you want I can take this and finish it. Version 3.4.5 should be out now.
The biggest blocker for me, was re-applying patches from @neheb ; I think I managed to apply them.

neheb:
Hmm? Do my patches need rebasing?

commodo:
> Hmm? Do my patches need rebasing?
> 
> note that upstream fixed many of the issues. My PR is down to 4 patches.

so, they did not apply the CUSTOMIZATIONS additions;
which i'm puzzled if they're needed:
this seems to build: 
https://github.com/commodo/packages/commit/c179c129a11ec5cadad0420a1e02b9a4685ea1d3

neheb:
Yeah. Use the latest ones from: https://github.com/pyca/cryptography/pull/4920

Upstream actually fixed some of the issues by killing 1.0.2 compatibility. I adjusted my PR as well.

jefferyto:
> If you want I can take this and finish it. Version 3.4.5 should be out now.

Without knowing what will happen in the future, I don't think it is worth updating past 3.3, but if you want to update to 3.4 it's fine by me.

If we update to 3.4, upstream moves on to 35 (new versioning scheme) and adds a 3.3 LTS branch, and we still don't have Rust available yet, then we may need to downgrade from 3.4 to 3.3 LTS.

commodo:
> If we update to 3.4, upstream moves on to 35 (new versioning scheme) and adds a 3.3 LTS branch, and we still don't have Rust available yet, then we may need to downgrade from 3.4 to 3.3 LTS.

what usually happens with these packages, is that people will ask for newer versions, rarely for older;
i admit, we don't often get requests for package upgrades;

so, updating now may be a bit more annoying;
but, it's more annoying if we may have to do it later;

so, i'll spin-up a PR for 3.4.5; and let's see where this takes us;

commodo:
created https://github.com/openwrt/packages/pull/14753

jefferyto:
If people are annoyed that we don’t update to the newer version (and this is a big if), they will be even more annoyed if we have to downgrade because the newer version isn’t getting security updates.

commodo:
> If people are annoyed that we don’t update to the newer version (and this is a big if), they will be even more annoyed if we have to downgrade because the newer version isn’t getting security updates.

ok, so then 3.3.2 or 3.4.5?
your preference

jefferyto:
My preference would be 3.3.2 for now and we can keep your branch for later.

commodo (approved):

commodo:
> My preference would be 3.3.2 for now and we can keep your branch for later.

ack
dropped python-cryptography from my PR

jefferyto:
So upstream has decided not to do an LTS branch (but they are going to make ["backports easier"](https://github.com/pyca/cryptography/issues/5814#issuecomment-778639932) for CVE fixes).

@commodo do you still have your 3.4 branch available? It would be nice to have it available for 21.02 if possible.

(Or I'm also inclined to drop this package entirely. Apparently all the companies that are funding this project "[are unwilling to appropriately invest in](https://github.com/pyca/cryptography/issues/5814#issuecomment-778639932)" having Rust support and thus we are unworthy of upstream's attention.)

commodo:
> 
> @commodo do you still have your 3.4 branch available? It would be nice to have it available for 21.02 if possible.

sorry, i don't have it around anymore;
i'm somewhat of an anti-hoarder; whenever i can, i throw away stuff :)
it shouldn't be too hard to re-build it;

> 
> (Or I'm also inclined to drop this package entirely. Apparently all the companies that are funding this project "[are unwilling to appropriately invest in](https://github.com/pyca/cryptography/issues/5814#issuecomment-778639932)" having Rust support and thus we are unworthy of upstream's attention.)

dropping it would also mean getting rid of a few more packages that depend on this;
i.e. `python-service-identity`, `python-pyopenssl`, `python-paramiko` 
`pymysql` can work without `python3-cryptography` 

i can re-do it in the next few days;
i also have to plow through some other emails that piled up;
i will be slow this week;
my daughter brought home some weird flu from kinder-garden; 
it's a bit ironic, that we were all locked-down to our homes for covid, and now that things are opening up, we're all getting hit hard with all the normal stuff;

jefferyto:
> dropping it would also mean getting rid of a few more packages that depend on this;

I'm only half serious (I think) - while I understand that desire to switch to Rust, I'm annoyed that a critical piece of Python infrastructure can disregard all concern for downstream projects. If we dropped it perhaps more users will complain to upstream lol

> i can re-do it in the next few days;

Thanks for looking into it (no rush necessary). I don't think I will have time in the foreseeable future.

commodo:
> > dropping it would also mean getting rid of a few more packages that depend on this;
> 
> I'm only half serious (I think) - while I understand that desire to switch to Rust, I'm annoyed that a critical piece of Python infrastructure can disregard all concern for downstream projects. If we dropped it perhaps more users will complain to upstream lol
> 

python-cryptography isn't too bad [yet];
there are other projects that do worse with downstream

PR #14745, opened 2021-02-15
unbound: automatically setup odhcpd link

dangowrt:
Maintainer: @EricLuehrsen 
Compile tested: mediatek/mt7622 (snapshot)
Run tested: mediatek/mt7622 (snapshot)

Description:
In case odhcpd is selected as main dhcp server and no dhcplink is still set to 'none', automatically setup the leasetrigger and adapt unbound default settings.


EricLuehrsen:
I think too many forced defaults are set. If you have a file "well configued" in `[build_root]/files/etc/config`, then that will be overwritten on first boot. Many users of optional packages also build from source and this would be a negative.

tiagogaspar8:
Will it overwrite files? Are you sure? I don't remember that being the case back in the day when I did my own configuration with that method

net/unbound/files/unbound-odhcpd.defaults
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+[ "$(uci get dhcp.odhcpd.maindhcp)" = "1" ] || exit 0
+[ "$(uci get unbound.ub_main.dhcp_link)" = "none" ] || exit 0
+
+[ -x /usr/sbin/unbound-control ] && uci set unbound.ub_main.unbound_control='1'
+
+uci batch <<EOF
+set unbound.ub_main.add_local_fqdn='1'
+set unbound.ub_main.add_wan_fqdn='1'
```
EricLuehrsen, line 10: This one is certainly optional. It overrides the WAN DNS which may or may not have a public definition by your ISP.

net/unbound/files/unbound-odhcpd.defaults
```diff
@@ -0,0 +1,20 @@
...
+set unbound.ub_main.dhcp_link='odhcpd'
+set unbound.ub_main.dhcp4_slaac6='1'
```
EricLuehrsen, line 12: You want to be sure IPv6 is used locally before setting this.

net/unbound/files/unbound-odhcpd.defaults
```diff
@@ -0,0 +1,20 @@
...
+set unbound.ub_main.domain='lan'
```
EricLuehrsen, line 13: You want to be sure to set this to the same in `/ect/config/dhcp` or check that they match or check that this is default "lan" while DHCP UCI is not default "lan". Maybe check before doing any of this. It would be a good defense against unintended consequences of overriding defaults.

net/unbound/files/unbound-odhcpd.defaults
```diff
@@ -0,0 +1,20 @@
...
+set unbound.ub_main.domain_type='static'
+set unbound.ub_main.listen_port='53'
```
EricLuehrsen, line 15: This is the default shipped in the package. If it is different, then you are messing with a configuration that is already customized. Again maybe a defensive check.

net/unbound/files/unbound-odhcpd.defaults
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+[ "$(uci get dhcp.odhcpd.maindhcp)" = "1" ] || exit 0
```
EricLuehrsen, line 3: This needs to be sure the installed package is not `odhcpd+ipv6_only`. It must be `odhcpd` plain. What if dnsmasq is still enabled? Unbound and dnsmasq will contend for privileged port 53 and not everyone will want the same solution. The package README.md provides two rather easy options, so those at least must be considered.

EricLuehrsen (changes requested):
Again the concern is too many defaults are overwritten based on trivial preconditions. This can have negative effects on package install or first boot of well configured prebuilt ROM.

net/unbound/files/unbound-odhcpd.defaults
dangowrt: I understand that the way this package was built is to provide flexibility to power-users. However, in this way it is unsuitable for inclusion by default in images eg. generated on `chef.libremesh.org`. It's just be great if things would work just by installing packages without having to manually intervene and configure stuff (I know this can be done by including stuff via `FILES=...` and in a lot of ways, I still consider it a very high barrier compared to selecting a package in a browser-UI)

Another option would be to have dedicated packages only containing UCI defaults, so users may choose `unbound-dhcp-odhcpd` (or `unbound-dhcp-dnsmasq`, in case anyone really wants that).

net/unbound/files/unbound-odhcpd.defaults
dangowrt: Ok, the best would be then to move ALL default configuration into an `uci-defaults` script which exists right away if `/etc/config/unbound` already exists. In that way, we would be sure to never overwrite user configuration.

net/unbound/files/unbound-odhcpd.defaults
dangowrt: Good point.

net/unbound/files/unbound-odhcpd.defaults
EricLuehrsen: Buy you can NOT have that. OpenWrt default is to have dnsmasq installed using port#53. Unbound wants to use port#53. So which one wins? How do you know that this user wants overall? Maybe they just disable dnsmasq and use Unbound as is? Maybe the serialize the two? Maybe the use odhcpd? Maybe DHCP and static resources are handled by NSD with proper authoritative definition?

tiagogaspar8:
Hi,
Is this pull request dead or it might still be considered?

dangowrt:
@tiagogaspar8 I understood the approach of having the `unbound` package itself ship with `uci-defaults`-style dynamic config generation was not well received as it may be difficult to exclude causes of disruption for existing setups, or rather, the approach conflicts with shipping a static `/etc/config/unbound` file for that reason.

I guess having additional config meta-packages to trigger configuration for `odhcpd` or `dnsmasq` integration is still an option, as that would not mess with anyone's existing configuration unless explicitly selected by the user and also doesn't require any structural changes to the existing package shipping a static default `/etc/config/unbound` configuration file.

Ie. I could introduce two packages (for the beginning): `unbound-link-odhcpd` and `unbound-link-dnsmasq`.
Both would only ship a uci-defaults scripts and depend on the corresponding DHCP server. This would allow users to select either meta-package to have unbound with dhcp integration ready out-of-the-box when generating images in the ImageBuilder or using [chef.libremesh.org](https://chef.libremesh.org/).

@EricLuehrsen do you thing this could be an acceptable solution?

EricLuehrsen:
That seems it would work with each package having some short name for "Unbound with Canned Configuration for ..."

tiagogaspar8:
@dangowrt are you still around this? If you want I can give it a shot and rework it as needed!

dangowrt:
@tiagogaspar8 Please go ahead, I'd be happy to have it off my shoulders (but will of course review and test).

tiagogaspar8:
Rest assure, as soon as I have a pull request I'll mention you so you can do a review 😆
I'll be taking your work and continuing, is there any problem?

dangowrt:
Apart from the concerns voiced in this PR, summarized by https://github.com/openwrt/packages/pull/14745#issuecomment-972910765 the approach worked fine for me.
So if you want to give it a re-spin we should have that in separate packages (`unbound-config-odhcpd`, different options to co-exist with dnsmasq each in one package).

OR converting *all* of `/etc/config/unbound` to be generated by uci-defaults if it doesn't yet exist, depends on what @EricLuehrsen is willing to deal with :smile_cat:

EricLuehrsen:
You could make /etc/config/unbound itself a "package" that is required by Unbound, and make variants of the package that "PROVIDE" the original package. "Unbound Base Config" or "Unbound Odhcp Config" and so on. Then scripting UCI-defaults isn't even a problem. Also regardless of package selected, a prebuild ROM config in .../files/etc/config/unbound will supersede it.

GeorgeSapkin:
No updates in over four years and PR has merge conflicts. Closing as stale. Feel free to reopen if anything changes.

PR #14750, opened 2021-02-15
python-requests: split up patches into two files

neheb:
quilt cannot handle two patches in one file. It ends up merging them
and removing the description from the second. To avoid this, split into
two.

Maintainer: @BKPepe 

No PKG_RELEASE bump as this is mostly cosmetic.

### DIFF
--- a/lang/python/python-requests/patches/0001-idna-dependency-bump.patch
+++ b/lang/python/python-requests/patches/0001-idna-dependency-bump.patch
@@ -12,43 +12,14 @@ https://github.com/psf/requests/pull/5711
  setup.py | 2 +-
  1 file changed, 1 insertion(+), 1 deletion(-)
 
-diff --git a/setup.py b/setup.py
-index 7ba4b2a25f..f265384236 100755
 --- a/setup.py
 +++ b/setup.py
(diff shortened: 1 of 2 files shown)


neheb:
```
 * pkg_hash_fetch_best_installation_candidate: Packages for tmux found, but incompatible with the architectures configured
```

Strange...

BKPepe:
Hmm... 

I am thinking if this was really necessary as it is cosmetic and not sure if I am a big fan of doing this because this comes from pull request (in requests repository) and by adding .patch to the end of the URL.

See: https://github.com/psf/requests/pull/5711.patch

There are two commits in one file. It is not in two files and we should not do it just to make `quilt` satisfied. But this could wait until requests release a new version, right?

neheb:
This comes as a separate change from https://github.com/openwrt/packages/pull/14646 . Note what happening there. It's merging the patches and removing the description.

As you mentioned, that PR has two commits. It should be one patch per commit, not PR.

BKPepe:
Still I think that the issue is elsewhere, but I don't mind having this merged though. It's a questinable if it patch series should be in one file as it is or in multiple ffiles...

PR #14758, opened 2021-02-15
strongswan: fix local_gateway discovery

pprindeville:
Maintainer: @stintel
Run tested: same, running on a production router

Description:

This has been observed by myself and @luizluca: `ip route get` is appending `uid0` to the output, as seen from:

```
root@OpenWrt2:~# ip route get 1.1.1.1
1.1.1.1 via 174.27.160.1 dev eth3 src 174.27.182.184 uid 0 
    cache 
root@OpenWrt2:~#
```

so the fix is an anchored match, discarding all else.  Also, using `ip -o` means never having to do multiline matches...

### DIFF
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=strongswan
 PKG_VERSION:=5.9.1
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=https://download.strongswan.org/ https://download2.strongswan.org/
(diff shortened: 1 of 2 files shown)


pprindeville:
@luizluca Can you verify this as well?

pprindeville:
@dedeckeh /@stintel This one is a little urgent, and I'd like to get it into 19.07. Thanks.

luizluca:
> @luizluca Can you verify this as well?

The patch is good.

```
:~# ipdest=8.8.8.8
:~# ip route get $ipdest | awk -F"src" '/src/{gsub(/ /,"");print $2}'
10.48.200.198uid0
:~# ip -o route get "$ipdest" | awk '/ src / { print gensub(/^.* src ([^ ]*) .*$/, "\\1", "g"); }'
10.48.200.198
```

dedeckeh:
Fix looks fine to me but the git commit description should contain a word about the issue and how it is fixed (as done in the PR)

pprindeville:
> Fix looks fine to me but the git commit description should contain a word about the issue and how it is fixed (as done in the PR)

Thanks for that suggestion.  Done!

dedeckeh (approved):

pprindeville:
Something squirrelly is going on with x86_64 on CI.  Proceeding anyway.

PR #14760, opened 2021-02-16
libmariadb: add dependency on libcurl

pprindeville:
Maintainer: @miska 
Run tested: same, tested build only

Description:

Seeing build failure that it can't find `<curl/curl.h>` header.

### DIFF
--- a/libs/libmariadb/Makefile
+++ b/libs/libmariadb/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libmariadb
 PKG_VERSION:=3.1.10
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=mariadb-connector-c-$(PKG_VERSION)-src.tar.gz
 PKG_SOURCE_URL := \
@@ -38,6 +38,7 @@ MARIADB_CLIENT_PLUGINS := \
 	auth_gssapi_client \
 	remote_io
 
+PKG_BUILD_DEPENDS:=curl
 PKG_CONFIG_DEPENDS := \
 	$(patsubst %,CONFIG_PACKAGE_$(PKG_NAME)-plugin-%,$(subst _,-,$(MARIADB_CLIENT_PLUGINS)))
 


neheb:
I remember hitting this.

So the issue is not that it is needed as a dependency. Just a build dependency.

pprindeville:
> I remember hitting this.
> 
> So the issue is not that it is needed as a dependency. Just a build dependency.

Adjusted then.  How does it look?

neheb:
Fine.,

pprindeville:
Can't figure out how to add @miska as a reviewer...

neheb:
AFAIK only people under https://github.com/orgs/openwrt/teams/package-maintainers can be marked as reviewers.

BKPepe (approved):

micmac1:
Isn't it the case that PKG_BUILD_DEPENDS only works when it holds the PKG_NAME of the package? Like, "curl" here?

pprindeville:
> Isn't it the case that PKG_BUILD_DEPENDS only works when it holds the PKG_NAME of the package? Like, "curl" here?

Can someone confirm that?

pprindeville:
> > Isn't it the case that PKG_BUILD_DEPENDS only works when it holds the PKG_NAME of the package? Like, "curl" here?
> 
> Can someone confirm that?

Okay, changed.

hnyman:
> > Isn't it the case that PKG_BUILD_DEPENDS only works when it holds the PKG_NAME of the package? Like, "curl" here?

> Can someone confirm that?

E.g. wiki
https://openwrt.org/docs/guide-developer/packages

> PKG_BUILD_DEPENDS - Packages that need to be built before this package. Use this option if you need to make sure that your package has access to includes and/or libraries of another package at build time. **Specify the directory name (i.e. openssl) rather than the binary package name (i.e. libopenssl).** This build variable only establishes the build time dependency. 

And a zillion earlier issue discussions here...
https://github.com/openwrt/packages/search?q=PKG_BUILD_DEPENDS&type=issues

nmeyerhans (approved):

PR #14761, opened 2021-02-16
perl-try-tiny: use relative include of perlmod.mk

tofurky:
The $(TOPDIR) seemed to cause Makefile parsing issues with CI.

Signed-off-by: Matt Merhar <mattmerhar@protonmail.com>

Maintainer: me
Compile tested: x86_64
Run tested: x86_64

Description:

There are intermittent CI failures seen when parsing the Makefile. Oddly it doesn't seem to happen all the time.

This is the only thing I can think of being the cause without seeing the dump.txt referenced in the CI logs.

### DIFF
--- a/lang/perl-try-tiny/Makefile
+++ b/lang/perl-try-tiny/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=perl-try-tiny
 PKG_VERSION:=0.30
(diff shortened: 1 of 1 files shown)


pprindeville (approved):
LGTM

neheb:
That is...accurate. I've fixed issues like this as well.

neheb:
Note that GitHub Actions uses the path packages_ci instead of packages.

PR #14765, opened 2021-02-16
openssl-gostengine: Add GOST2012 Engine for OpenSSL

6ec123321:
Engine with GOST2012 support for OpenSSL.

dir src is "git clone https://github.com/gost-engine/engine.git --branch openssl_1_1_0_release2" with my CMakeLists.txt for Openwrt/Entware libopenssl 1.1.1i.

For compile need enable in (make) menuconfig>Libraries>SSL>libopenssl:
[+]Enable engine support
[+] Prepare library for GOST engine

or we get error: unknown type name 'ENGINE_CMD_DEFN'


libs/openssl-gostengine/Makefile
```diff
@@ -0,0 +1,44 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=openssl-gostengine
+PKG_RELEASE:=1.1.0r2
+CMAKE_INSTALL:=1
+ENGINES_DIR:=engines-1.1
+OPENSSL_INCLUDE_DIR:=/opt/include/openssl
```
neheb, line 7: Not valid here.

neheb:
Why are you duplicating the source here?

ping @cotequeiroz

cotequeiroz:
I think you're trying to do too much here, and should consider starting it over.
I can point to the [afalg_engine](https://github.com/openwrt/packages/tree/master/libs/afalg_engine) as an example of how to add an openssl engine package.  I've run a test-build with the following Makefile.  You may use it as a template, or almost as is, to write the package Makefile:
``` Makefile
include $(TOPDIR)/rules.mk

PKG_NAME:=openssl-gostengine
# I'm not sure which versions you have tested, and or want this.  This is the one I've test-built
PKG_VERSION:=1.1.0.3  
PKG_RELEASE:=$(AUTORELEASE)

PKG_SOURCE_URL:=https://github.com/gost-engine/engine/archive/v$(PKG_VERSION)
# Use 'skip' here, and let the build system fill this for you with `make package/openssl-gostengine/{download,check} FIXUP=1`
PKG_HASH:=skip

# Please consider adding yourself as the maintainer
PKG_MAINTAINER:=Artur Petrov <github@phpchain.ru>
PKG_LICENSE:=OpenSSL
# the current master & openssl_1_1_1 branches have an actual license file, missing in current releases:
#PKG_LICENSE:=Apache-2.0
#PKG_LICENSE_FILES:=LICENSE

# The package tarball extracts to `engine-1.1.0.3`, which may cause confusion.  Use the full name with this:
PKG_UNPACK:=$(HOST_TAR) -C "$(PKG_BUILD_DIR)" --strip-components=1 -xzf "$(DL_DIR)/$(PKG_SOURCE)"
#The alternative, to use the generic engine-v.v.v.v dir is to define the unusual build dir: 
#PKG_BUILD_DIR:=$(BUILD_DIR)/engine-$(PKG_VERSION)
# the package does not understand `make install`
PKG_INSTALL:=

ENGINES_DIR:=engines-1.1

define Package/libopenssl-gostengine
    SECTION:=libs
    CATEGORY:=Libraries
    SUBMENU:=SSL
# the TITLE must be short enough to fit the menu
    TITLE:=Russian GOST crypto algorithms for OpenSSL
    URL:=https://github.com/gost-engine/engine/
    DEPENDS:=libopenssl @OPENSSL_ENGINE @OPENSSL_WITH_GOST +libopenssl-conf
endef

define Package/libopenssl-gostengine/description
    A reference implementation of the Russian GOST crypto algorithms for OpenSSL
endef

CMAKE_OPTIONS += -DOPENSSL_ENGINES_DIR=/usr/lib/$(ENGINES_DIR)

define Package/libopenssl-gostengine/install
        $(INSTALL_DIR) $(1)/usr/lib/$(ENGINES_DIR)
        $(INSTALL_BIN) $(PKG_BUILD_DIR)/bin/gost.so $(1)/usr/lib/$(ENGINES_DIR)/
# I would strongly recommend creating a separate `gostsum` package for these:
        $(INSTALL_DIR) $(1)/usr/bin
        $(INSTALL_BIN) $(PKG_BUILD_DIR)/bin/{gost12sum,gostsum} \
                        $(1)/usr/bin/
endef

$(eval $(call BuildPackage,libopenssl-gostengine))
```

Note that, at the moment, the package will NOT be built by the bots, and won't be available for install from the official feeds.  I'll see about changing the default openssl build options to support it (set `OPENSSL_WITH_GOST=y`).  Let's get at least close to merging this first, then I'll propose the patch.

cotequeiroz:
I've gone ahead and [proposed to always build OpenSSL with GOST engine support](https://patchwork.ozlabs.org/project/openwrt/patch/20210218005008.19974-1-cotequeiroz@gmail.com/).  If it gets accepted before this is merged, then `@OPENSSL_WITH_GOST` must be removed from the package `DEPENDS` definition.

6ec123321:
> I've gone ahead and [proposed to always build OpenSSL with GOST engine support](https://patchwork.ozlabs.org/project/openwrt/patch/20210218005008.19974-1-cotequeiroz@gmail.com/). If it gets accepted before this is merged, then `@OPENSSL_WITH_GOST` must be removed from the package `DEPENDS` definition.



> I've gone ahead and [proposed to always build OpenSSL with GOST engine support](https://patchwork.ozlabs.org/project/openwrt/patch/20210218005008.19974-1-cotequeiroz@gmail.com/). If it gets accepted before this is merged, then `@OPENSSL_WITH_GOST` must be removed from the package `DEPENDS` definition.

Many thanks. A patch for OpenSSL to enable GOST ENGINE will help a lot. I'll start from the beginning. I use the Makefile you suggested. You can close a pull request.

6ec123321:
> I've gone ahead and [proposed to always build OpenSSL with GOST engine support](https://patchwork.ozlabs.org/project/openwrt/patch/20210218005008.19974-1-cotequeiroz@gmail.com/). If it gets accepted before this is merged, then `@OPENSSL_WITH_GOST` must be removed from the package `DEPENDS` definition.

The patch for OpenSSL 1.1.1j needs to be slightly changed for the [gost] section:
This settings work perfectly for Russian. In russian we are always writing CRYPT_PARAMS = id-Gost28147-89-CryptoPro-A-ParamSet, like this:
```
# gost section
[gost]
engine_id = gost
dynamic_path = /opt/lib/engines-1.1/gost.so
default_algorithms = ALL
CRYPT_PARAMS = id-Gost28147-89-CryptoPro-A-ParamSet
```

cotequeiroz:
The `engine_id` and `dynamic_path` are not needed, since the id is not changed, and it is installed to the default engines dir.  The `CRYPT_PARAMS` I used as example is just the default defined in [gost_crypt.c#L326](https://github.com/gost-engine/engine/blob/bad240e1016c8cb2554994eda9f67ea0ad8ee89d/gost_crypt.c#L326).  The string itself is defined by openssl in [openssl/openssl:crypto/objects/obj_dat.h#L2124](https://github.com/openssl/openssl/blob/458d168cd48ab57ffd8e6c8322073e4a77d03d26/crypto/objects/obj_dat.h#L2124).  I may be reading the code wrong, though.  Please explain why I should propose an example that deviates from the default.  I can also just leave it blank--i fact, I should have.  I just don't want to lead people to use some arbitrary or random value.

Issue #14652, opened 2021-02-04
libgpiod: does not compile on macOS 11.2-x86_64 host


k-ronny:
Maintainer: @mhei
Environment: all?

Description: cross compiling of _libgpiod_ on a macOS 11.2-x86_64 host is failing

```
...
gpiodmodule.c:8:10: fatal error: Python.h: No such file or directory
 #include <Python.h>
          ^~~~~~~~~~
compilation terminated.
...
```
I think it is related to this definition in the Makefile:
```
CONFIGURE_VARS += \
        PYTHON="$(STAGING_DIR_HOSTPKG)/bin/$(PYTHON3)" \
        PYTHON_CPPFLAGS="$(shell $(STAGING_DIR)/host/bin/$(PYTHON3)-config --includes)" \
        PYTHON_LIBS="$(shell $(STAGING_DIR)/host/bin/$(PYTHON3)-config --libs)"
endif
```
If I apply this patch it compiles on macOS and Debian Buster and it runs on a Onion Omega2+:
```
diff --git a/libs/libgpiod/Makefile b/libs/libgpiod/Makefile
index 68996c831..b944dbcfa 100644
--- a/libs/libgpiod/Makefile
+++ b/libs/libgpiod/Makefile
@@ -33,9 +33,8 @@ endif
 ifneq ($(CONFIG_PACKAGE_python3-gpiod),)
 CONFIGURE_ARGS += --enable-bindings-python
 CONFIGURE_VARS += \
-       PYTHON="$(STAGING_DIR_HOSTPKG)/bin/$(PYTHON3)" \
-       PYTHON_CPPFLAGS="$(shell $(STAGING_DIR)/host/bin/$(PYTHON3)-config --includes)" \
-       PYTHON_LIBS="$(shell $(STAGING_DIR)/host/bin/$(PYTHON3)-config --libs)"
+       PYTHON_CPPFLAGS="-I$(PYTHON3_INC_DIR)" \
+       PYTHON_LIBS="-lpython$(PYTHON3_VERSION)"
 endif
 
 define Package/libgpiod
```
But I do not know if this is the right approach.

neheb:
ping @jefferyto

neheb:
@k-ronny what does $(STAGING_DIR)/host/bin/$(PYTHON3)-config --includes actually return?

k-ronny:
@neheb

```
[...]

PR #14767, opened 2021-02-16
python3: Fix target python-config when building on macOS

jefferyto:
Maintainer: me, @commodo 
Compile tested: armvirt-32, 2020-02-14 snapshot sdk (Ubuntu 20.10); ramips (mt76x8), OpenWrt master (a45953c) (macOS Catalina 10.15.7)
Run tested: N/A

Description:
The Python version of python-config is installed when building on macOS, rather than the shell script version when building on Linux. When run on macOS, the Python version of python-config will return values with Mac-specific customizations.

This patches the python-config install recipe so that which version is installed can be controlled by the package makefile. When building on macOS, this installs the Python version for host Python and the shell script version for target Python.

This also updates `Host/Compile` and `Host/Install` to use the default host build recipes, so that the various `HOST_*` variables are taken into account automatically.

Fixes https://github.com/openwrt/packages/issues/14652

### DIFF
--- a/lang/python/python3/Makefile
+++ b/lang/python/python3/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 include ../python3-version.mk
 
 PKG_NAME:=python3
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_VERSION:=$(PYTHON3_VERSION).$(PYTHON3_VERSION_MICRO)
 
 PKG_SOURCE:=Python-$(PKG_VERSION).tar.xz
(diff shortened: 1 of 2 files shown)


httpstorm:
@jefferyto 
Both host and target compile correctly. Solves https://github.com/openwrt/packages/issues/14652
I recommend running `make package/feeds/packages/python3/host/refresh`, and updating the pull-request, because some of the previous patches need to have their offsets adjusted. The person who created them forgot to do this.
```
make package/feeds/packages/python3/host/{clean,prepare} V=sc
Applying ./patches/003-do-not-run-distutils-tests.patch using plaintext: 
patching file Makefile.pre.in
Hunk #1 succeeded at 1525 (offset 1 line).
…
```

neheb:
The refresh thing will become mandatory for CI soon. Might as well do it now.

commodo (approved):

jefferyto:
Refreshing patches is unrelated to this change, and Python 3.9.2 is expected to be released on [Mar 1](https://www.python.org/dev/peps/pep-0596/#bugfix-releases). I would prefer to refresh patches with that update.

PR #14772, opened 2021-02-17
owipcalc: change maintainer

PolynomialDivision:
As suggested by others, I would like to take care of this tool. I am developing certain tools that rely on the library and also owipcalc.

Maintainer: @jow-


jow-:
ACK from me.

PR #14784, opened 2021-02-18
gost_engine: Add GOST2012 Engine for OpenSSL

6ec123321:
Continuation of the previous pull request:
https://github.com/openwrt/packages/pull/14765


libs/gost_engine/Makefile
```diff
@@ -0,0 +1,56 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gost_engine
+PKG_VERSION:=1.1.0.3
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/gost-engine/engine/archive/v$(PKG_VERSION)
+# Use 'skip' here, and let the build system fill this for you with `make package/openssl-gostengine/{download,check} FIXUP=1`
```
cotequeiroz, line 9: All of the comments I wrote in the example Makefile are not supposed to be left in the final package Makefile.  Please remove them.

libs/gost_engine/Makefile
```diff
@@ -0,0 +1,56 @@
...
+endef
+
+define Package/libopenssl-gost_engine/description
+A reference implementation of the Russian GOST crypto algorithms for OpenSSL
+endef
+
+CMAKE_OPTIONS += -DOPENSSL_ENGINES_DIR=/usr/lib/$(ENGINES_DIR)
+
+define Package/libopenssl-gost_engine/install
+	$(INSTALL_DIR) $(1)/opt/lib/$(ENGINES_DIR)
```
cotequeiroz, line 48: You should not use the `/opt` directory in openwrt.  Use the default openssl `/usr/lib/$(ENGINES_DIR)` instead.

libs/gost_engine/Makefile
```diff
@@ -0,0 +1,56 @@
...
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/bin/gost.so \
+			$(1)/opt/lib/$(ENGINES_DIR)/
+	$(INSTALL_DIR) $(1)/opt/bin
```
cotequeiroz, line 51: Same here.  Use `/usr/bin`.

libs/gost_engine/patches/001-CMakeLists.patch
```diff
@@ -0,0 +1,31 @@
...
+ 
+-#add_executable(gostsum12 ${GOST_SUM_12_SOURCE_FILES})
+-#
+-#target_link_libraries(gostsum12 gost)
++INSTALL(
++  TARGETS gost
++  ARCHIVE DESTINATION lib
++  LIBRARY DESTINATION lib
++  COMPONENT library
++)
```
cotequeiroz, line 31: This patch is unnecessary.  OpenSSL is installed in the default search directories (`$(STAGING_DIR)/usr/include`, `$(STAGING_DIR)/usr/lib`), and we are not calling `make install`.  We can just copy the binaries from `$(PKG_BUILD_DIR)/bin`.

cotequeiroz (changes requested):
Let's try to keep things simpler.  Use the default openwrt directories, where openssl will find the engine without having to tell it where they are.  We don't need to maintain a patch either.  Please test the package with my proposed changes.  Don't forget to NOT use the `dynamic_path` line in `openssl.cnf` then.

6ec123321:
> Let's try to keep things simpler. Use the default openwrt directories, where openssl will find the engine without having to tell it where they are. We don't need to maintain a patch either. Please test the package with my proposed changes. Don't forget to NOT use the `dynamic_path` line in `openssl.cnf` then.

Yes, it work without dynamic_path.

cotequeiroz:
LGTM so far.  I'll ping @neheb to see about the preferred Makefile style.  

Since it is your first contribution, as important as getting the package right, in my opinion, is for you to understand how the package Makefile works, so that you can maintain the package.  If you have any doubt about how the package is built, or installed, please don't hesitate to ask.

A good exercise, and I strongly recommend doing that anyway, is to separate the `gostsum` and `gost12sum` utilities from the engine package.  It would usually be mandatory, but since the engine is over 300K uncompressed, 50K utilities will probably not be such a big deal, **iff they are almost always installed together**.  You can do that by defining a new package (inside the same Makefile), and moving the binary installation to the new package.

The `DEPENDS` definition will be tricky.  IMO, it's best to keep the utils hidden until you select the engine; in that case, both packages (engine and utils) will have the same constraints--i.e. will depend on libssl, engines, and gost_engine preparation.  Additionally, it would force the selection of the gost engine.  Constraints are `DEPENDS` tokens without a `+` sign, while force-selections have the `+` in front.  The other option would be to force-select everything.

See the openssl-utils package--for package placement inside menuconfig--inside [package/libs/openssl/Makefile](https://github.com/openwrt/openwrt/blob/master/package/libs/openssl/Makefile) for inspiration.  Note that `openssl-utils` is force-installing openssl, and I'm suggesting you to not do that.

You'll need to squash the commits before we can merge this, but for now, you may keep piling them on so you can backtrack easily.

Cheers

6ec123321:
Moved the binary installation to the utils package gost_engine-util like in openssl package. Keep the utils hidden until you select the engine.

6ec123321:
Closed https://github.com/openwrt/packages/pull/14784

PR #14795, opened 2021-02-19
libnetfilter-queue: re-add PKG_FIXUP to fix build

apple-etan:
Currently when trying to build this package a build error is produced:

WARNING: 'automake-1.16' is missing on your system.
         You should only need it if you modified 'Makefile.am' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'automake' program is part of the GNU Automake package:
         <https://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <https://www.gnu.org/software/autoconf>
         <https://www.gnu.org/software/m4/>
         <https://www.perl.org/>

This error is due to an attempt to use the native host tools instead of
the OpenWrt build system generated ones. By re-adding PKG_FIXUP it is
ensured that the correct version of the host tools are used.

Signed-off-by: Etan Kissling <etan_kissling@apple.com>
(cherry picked from commit cab73ac82606068ac081df7cd6a2100acc620c77)
Signed-off-by: Etan Kissling <etan_kissling@apple.com>

### DIFF
--- a/libs/libnetfilter-queue/Makefile
+++ b/libs/libnetfilter-queue/Makefile
@@ -9,12 +9,13 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libnetfilter_queue
 PKG_VERSION:=1.0.5
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


apple-etan:
Backport from https://github.com/openwrt/packages/pull/14786

PR #14801, opened 2021-02-19
rng-tools: start as early as possible

Noltari:
Maintainer: @hnyman @nwf @pprindeville 
Compile tested: bcm27xx/bcm2708
Run tested: bcm27xx/bcm2708 (Raspberry Pi B+)

Description:
Raspberry Pi devices (bcm2708) need this to get from:
[  102.310494] random: crng init done
to:
[   12.539744] random: crng init done

### DIFF
--- a/utils/rng-tools/files/rngd.init
+++ b/utils/rng-tools/files/rngd.init
@@ -1,7 +1,7 @@
 #!/bin/sh /etc/rc.common
 # Copyright (C) 2011-2014 OpenWrt.org
 
-START=25
+START=0
 
 USE_PROCD=1
 PROG=/sbin/rngd
--- a/utils/rng-tools/Makefile
+++ b/utils/rng-tools/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=rng-tools
 PKG_VERSION:=6.10
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/nhorman/rng-tools


neheb:
How about haveged?

pprindeville:
> How about haveged?

Eh?

utils/rng-tools/files/rngd.init
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh /etc/rc.common
 # Copyright (C) 2011-2014 OpenWrt.org
 
-START=25
+START=0
```
pprindeville, line 4: Zero seems a bit early.  I don't like the idea of not being able to put something before this...  What about `5`?

Noltari:
> How about haveged?

@neheb haveged works because it has START=01:
https://github.com/openwrt/packages/blob/4b6948bbf01989c2c299a64748485b7b65edcbf2/utils/haveged/files/haveged.init#L3

@pprindeville do you think START=1 is OK?
BTW, urngd is using START=0

pprindeville:
> BTW, urngd is using START=0

Okay, let's give that a try then.

pprindeville (approved):

neheb (approved):

hnyman:
> BTW, urngd is using START=0

Actually, it is using 00

All priorities so far have been done with two digits (00-99). I think that rng-tools should be so, too.
The run order is alphabetical, so it would now be:  S00sysfixtime, S00urngd, S0rng-tools

I also think that all relevant code, e.g. in rc.common expects a two-digit priority:

https://github.com/openwrt/openwrt/blob/3f1109bf2af2cf4977689a33ea481afaaac8dc98/package/base-files/files/etc/rc.common#L48

`		ln -sf "../init.d/$name" "$IPKG_INSTROOT/etc/rc.d/S${START}${name##S[0-9][0-9]}" && \`

hnyman:
> merged commit f5a615b

Hmm. Just missed with my comment :-(

Noltari:
> > merged commit [f5a615b](https://github.com/openwrt/packages/commit/f5a615b32c67646df21e14686912c1650d4860a1)
> 
> Hmm. Just missed with my comment :-(

Ah shit, you're right...
Should I open a new PR to fix this?

hnyman:
Just commit a fix change from 0 to 00.
No need to have a PR

tofurky:
starting rngd before kmods are loaded prevents the use of any rng that requires a kmod to be loaded before its device appears, e.g. ttyACM0.

moving it to START=11 allows it to be used with devices like this. would that be a good compromise between starting early enough and remaining compatible for hwrng drivers not built into kernel??

pprindeville:
> starting rngd before kmods are loaded prevents the use of any rng that requires a kmod to be loaded before its device appears, e.g. ttyACM0.
> 
> moving it to START=11 allows it to be used with devices like this. would that be a good compromise between starting early enough and remaining compatible for hwrng drivers not built into kernel?

Do serial-to-USB devices generate enough events quickly enough to be meaningful sources of entropy?

tofurky:
https://www.gniibe.org/memo/development/gnuk/rng/neug.html
https://ubld.it/products/truerng-hardware-random-number-generator/

etc.

pprindeville:
> https://www.gniibe.org/memo/development/gnuk/rng/neug.html
> https://ubld.it/products/truerng-hardware-random-number-generator/
> 
> etc.

Hmm.

> "The TrueRNG is a low-cost USB Hardware Random Number Generator that provides more than 350 kilobits/second of true random numbers."

PR #14811, opened 2021-02-20
ksmbd and ksmbd-tools 3.3.5

xdarklight:
Maintainer: @Andy2244
Compile tested: mips, lantiq
Run tested: lantiq BT Home Hub 5A

Description:
Updates ksmbd and ksmbd-tools to the latest upstream version, which is 3.3.5 (for both packages)
Upstream changelogs are included in the commit for each package

The main change for me is that it fixes a crash, possibly the one fixed in cifsd-team/cifsd@d877f11fbd5aab09e73a09fdb93d1bae4efae46e
Possibly this should be backported to 21.02 for that reason

### DIFF
--- a/kernel/ksmbd/Makefile
+++ b/kernel/ksmbd/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ksmbd
-PKG_VERSION:=3.3.4
+PKG_VERSION:=3.3.5
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/cifsd-team/cifsd/tar.gz/$(PKG_VERSION)?
-PKG_HASH:=4f8b7610ba084f6813cbb85bb6c07af50ba542e928c370e79022039fa027bc9a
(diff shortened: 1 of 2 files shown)


Andy2244:
Looks fine, you guys are fast 👍 

Yes, since we have the 3.3 ABI now on stable, we should try to always backport to 19.07+ and 20.+ too.

neheb:
Those buildbot errors are worrying.

xdarklight:
> Those buildbot errors are worrying.

to be honest, I have no idea where these errors are coming from.
From the ``arm_cortex-a15_neon-vfpv4`` target for example:
```
ERROR: "des_expand_key" [/home/build/openwrt/build_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/linux-armvirt_32/ksmbd-3.3.5/ksmbd.ko] undefined!
ERROR: "des_encrypt" [/home/build/openwrt/build_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/linux-armvirt_32/ksmbd-3.3.5/ksmbd.ko] undefined!
```

The ksmbd package does have a kmod-crypto-des dependency.
From 3.3.4 to 3.3.5 nothing related to "des_" functions has changed:
```
$ git diff 3.3.4 3.3.5 | grep des_
$
```

so I have to chicken out and wait for others to comment with ideas as to what's going wrong here.

neheb:
ping @aparcar

neheb:
Hmm strange. These errors might be false positives. kmod-crypto-des is selected. It might be erroring because DES symbols are not in the main kernel config.

PR #14816, opened 2021-02-21
dockerd: misc maintainence

G-M0N3Y-2503:
Maintainer: @G-M0N3Y-2503
Compile tested: x86_x64, VirtualBox, OpenWrt Master
Run tested: x86_x64, VirtualBox, OpenWrt Master

Description:
* `wan` blocking rule now jumps to `REJECT` instead of `DROP` to be consistent with other firewall rules  
* Refactored `/etc/config/dockerd` 
  * Default config only sets defaults dictated by OpenWrt, Otherwise, docker defaults can be assumed.
  * Normalized quotation marks
 * Mitigated case where `wan` blocking rule wasn't added if previous `iptables` command was taking too long
 * Updated patch file format


feckert:
looks good. We should then also use this in the Openwrt 21.02 branch.

G-M0N3Y-2503:
> looks good. We should then also use this in the Openwrt 21.02 branch.

Agreed. 
see https://github.com/openwrt/packages/pull/14839

feckert (approved):

PR #14817, opened 2021-02-21
sane-backends: remove outdated C++ patch

neheb:
std::round was unavailable because of a wrong configure var from the
uClibc days. Fixed upstream with 1647790ae6202bd70b199e2aa44f8eb43e8d3af1

Maintainer: @luizluca


neheb:
https://github.com/openwrt/openwrt/commit/1647790ae6202bd70b199e2aa44f8eb43e8d3af1

neheb:
Fails on libv4l... Wonder why.

neheb:
Cool.

utils/sane-backends/Makefile
```diff
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=sane-backends
 PKG_VERSION:=1.0.31
-PKG_RELEASE:=1
+PKG_RELEASE:=2
```
luizluca, line 13: I don't think this is needed. The patch added a define that was always false without uclibc. In the end, it will get the same binary. Anyway, it is not a blocker, specially for master.

luizluca:
It looks like uclibc is gone since dec/20. Without uclibc, sane-backends is good without that patch.
Ack from me.

As you touched libv4l, @thess should also take a look.

PR #14820, opened 2021-02-21
readsb: update to 3.9.0

neheb:
Remove upstreamed patches.

Use autorelease to avoid bumping the release manually.

Fix bad linker variable on glibc.

Maintainer: @stintel 
Compile tested: arc770


stintel:
Deprecated upstream. Better replace it with https://github.com/Mictronics/readsb-protobuf.

neheb:
Requires new dependencies including librrd, which is not in the tree.

neheb:
Let's merge for now. readsb-protobuf can be an exercise for someone else.

PR #14826, opened 2021-02-21
libs/icu:  fix host compile in macOS

jjm2473:

Maintainer: Hirokazu MORIKAWA <morikw2@gmail.com>
Compile tested: macOS 10.15.7
Run tested: skiped

Description:
just a litter fix


nxhack:
I don't have a MacOS environment for verification, so I'll just comment on quick look.
It seems that there is no harm in adapting it, so I approve it.

jjm2473:
Bumped the PKG_RELEASE

Issue #14573, opened 2021-01-26
unbound: does not compile on macOS 11.1-x86_64 host


k-ronny:
Maintainer: @EricLuehrsen
Environment: all?
Description: cross compiling of unbound on a macOS 11.1-x86_64 host is failing in the configure phase:
```
...
checking whether CONNECT_RESUME_ON_READ_WRITE is declared... no
configure: error: TCP Fast Open is not available for client mode: please rerun without --enable-tfo-client
...
```
I think `configure` is using the wrong target because of the use of `uname` in configure.ac:
```
AC_ARG_ENABLE(tfo-client, AS_HELP_STRING([--enable-tfo-client],[Enable TCP Fast Open for client mode]))
case "$enable_tfo_client" in
	yes)
		case `uname` in
			Linux) AC_CHECK_DECL([MSG_FASTOPEN], [AC_MSG_WARN([Check the platform specific TFO kernel parameters are correctly configured to support client mode TFO])],
			                     [AC_MSG_ERROR([TCP Fast Open is not available for client mode: please rerun without --enable-tfo-client])], 
			                     [AC_INCLUDES_DEFAULT 
#include <netinet/tcp.h>
])
					AC_DEFINE_UNQUOTED([USE_MSG_FASTOPEN], [1], [Define this to enable client TCP Fast Open.])
			  ;;
			Darwin) AC_CHECK_DECL([CONNECT_RESUME_ON_READ_WRITE], [AC_MSG_WARN([Check the platform specific TFO kernel parameters are correctly configured to support client mode TFO])], 
			                      [AC_MSG_ERROR([TCP Fast Open is not available for client mode: please rerun without --enable-tfo-client])], 
			                      [AC_INCLUDES_DEFAULT
#include <sys/socket.h>
])
					AC_DEFINE_UNQUOTED([USE_OSX_MSG_FASTOPEN], [1], [Define this to enable client TCP Fast Open.])
			  ;;
		esac
	;;
	no|*)
		;;
esac
```

with this patch it compiles on macOS and Debian Buster:
```
--- a/configure.ac	2020-12-03 09:31:18.000000000 +0100
+++ b/configure.ac	2021-01-26 01:16:58.000000000 +0100
@@ -120,6 +120,9 @@
 AC_SUBST(LIBUNBOUND_REVISION)
 AC_SUBST(LIBUNBOUND_AGE)
 
+# AC_CANONICAL_HOST is needed to access the 'host_os' variable
[...]

PR #14828, opened 2021-02-21
unbound: fix cross compile in macOS

jjm2473:
don't use `uname` when cross compile,
uname should always be 'linux' in openwrt target.

Maintainer: Eric Luehrsen <ericluehrsen@gmail.com>
Compile tested: macOS 10.15.7
Run tested: Not tested

Description:

### DIFF
--- a/net/unbound/patches/001-macos-host.patch
+++ b/net/unbound/patches/001-macos-host.patch
@@ -1,0 +1,28 @@
+--- a/configure.ac
++++ b/configure.ac
+@@ -1209,23 +1209,13 @@ esac
+ 
+ AC_ARG_ENABLE(tfo-client, AS_HELP_STRING([--enable-tfo-client],[Enable TCP Fast Open for client mode]))
+ case "$enable_tfo_client" in
+-	yes)
+-		case `uname` in
+-			Linux) AC_CHECK_DECL([MSG_FASTOPEN], [AC_MSG_WARN([Check the platform specific TFO kernel parameters are correctly configured to support client mode TFO])],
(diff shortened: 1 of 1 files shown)


pprindeville:
If we’re modifying configure.ac and we run autoreconf then we don’t need to patch configure...

jjm2473:
The 'configure' comes from source, so maybe we should patch it

pprindeville:
> The 'configure' comes from source, so maybe we should patch it

Look at:

https://github.com/openwrt/packages/blob/master/net/unbound/Makefile#L24

this guarantees that any patch to `configure` will be clobbered when `autoreconf` gets run, as per my original comment.

jjm2473:
OK, I will test it

jjm2473:
OK, tested. Now only patch configure.ac

neheb:
Needs a rebase looks like. ping @EricLuehrsen

jjm2473:
@neheb Rebased. `PKG_RELEASE:=$(AUTORELEASE)`, is that right?

neheb:
Yeah.

jjm2473:
If Makefile not touched, `Test build` may have bug in step `Determine changed packages`, 
https://github.com/openwrt/packages/pull/14828/checks?check_run_id=1958652137

Detected changed packages were `vim tmux bmon`, not `unbound`

neheb:
ping @aparcar

EricLuehrsen:
The patch should probably be named `number-original_file_name-why_fix.patch` but otherwise the content seems reasonable.

EricLuehrsen:
The patch should probably be named `number-original_file_name-why_fix.patch` but otherwise the content seems reasonable. Resolves #14573 locally to openwrt/packages.

jjm2473:
The patch #14573 is better than mine. My patch only suitable for openwrt (TARGET always linux).

PR #14829, opened 2021-02-21
node: fix compile on macOS

jjm2473:
when build target package on macOS
1.v8_libbase.host.mk will link to platform-linux.o, should be platform-macos.o
2.clang does not support '-Wl,--start-group'

Maintainer: @nxhack Hirokazu MORIKAWA <morikw2@gmail.com>, Adrian Panella <ianchi74@outlook.com>
Compile tested: macOS 10.15.7
Run tested: macOS 10.15.7 (Host only)

Description:

### DIFF
--- a/lang/node/Makefile
+++ b/lang/node/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=node
 PKG_VERSION:=v14.15.5
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://nodejs.org/dist/$(PKG_VERSION)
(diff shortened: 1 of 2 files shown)


nxhack:
I don't have a MacOS environment for verification, so I'll just comment on quick look.

This is a big fix. Is it being discussed in the upstream node.js repository? If so, I'll wait for the upstream fix.

jjm2473:
Yes, it's a bug of upstream node.js, but this patch only suit the openwrt, because in openwrt, the `target` always linux.

I have another question, why node.js compile `host` files when cross compiling?
In openwrt, we have a splited `host` build.

If we don't need `host` files when cross compiling, is there a way to avoid compile `host` files? (only target)

jjm2473:
I reported a bug to node.js

nxhack:
Some packages require node/host. For example, the node module and adguardhome.

https://github.com/openwrt/packages/search?q=%22node%2Fhost%22

To work around the Darwin host build problem, install node.js yourself using brew as the Darwin build environment. Then modify the OpenWrt node package to ignore the Darwin case.

Can this workaround fix it?

jjm2473:
This PR not affact host build, can you try apply this patch in linux host?

neheb:
This needs a rebase.

PR #14854, opened 2021-02-22
hplip: bump to 3.21.2

luizluca:
Patches refreshed:
* 010-libusb_fix.patch and
* 020-remove_cups_dep_on_scan.patch.
Patches dropped:
* 060-fix-uclibc.patch as uclibc support is gone in OpenWrt

Maintainer: me
Compile tested: armvirt_64,ath79_generic,mvebu_cortexa9,powerpc,ramips_mt7620,x86_64,x86_generic
Run tested: scanned using a TP-Link archer C7v5

Description:

### DIFF
--- a/utils/hplip/Makefile
+++ b/utils/hplip/Makefile
@@ -1,7 +1,4 @@
 #
-# Copyright (C) 2006-2011 OpenWrt.org
-# Copyright (C) 2017-2018 Luiz Angelo Daros de Luca <luizluca@gmail.com>
-#
 # This is free software, licensed under the GNU General Public License v2.
 # See /LICENSE for more information.
 #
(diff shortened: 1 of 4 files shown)


neheb:
Please run make package/hplip/refresh

luizluca:
> Please run make package/hplip/refresh

Thanks! Now I see the errors. CI was showing "error:" on every section when I clicked on them while I was using an HTTP proxy. Now, without the proxy, I can see the problem.

neheb:
```
 scan/sane/OrbliteScan/LinuxCommon.h:22:9: error: unknown type name 'uint64_t'
 typedef uint64_t   UInt64;
```

That patch you removed seems to be needed for glibc.

luizluca:
> That patch you removed seems to be needed for glibc.

Yes, thanks! My glibc test build was broken because of libv4l (which you fixed just now, thanks again!).
I pushed reverting the patch but renaming uclibc reference to glibc.

I'm just curious how this same build does not happen with a normal desktop build as it normally uses glibc.
I was expecting glibc targets to require no extra patches.

neheb:
I assume it's because a more recent glibc is used in OpenWrt.

PR #14856, opened 2021-02-22
mariadb: update to 10.4.18

neheb:
Remove uClibc depends. It's gone now.

Remove CMAKE_INSTALL to avoid installing to InstallDev.

Maintainer: @miska 
Compile tested: ath79

### DIFF
--- a/utils/mariadb/Makefile
+++ b/utils/mariadb/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mariadb
-PKG_VERSION:=10.4.17
+PKG_VERSION:=10.4.18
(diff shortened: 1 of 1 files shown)


neheb:
ping @micmac1 I think you dealt with the InstallDev stuff.

micmac1:
Hello Rosen,

I think this is fine. I'm pretty sure I put the CMAKE_INSTALL:=1 in
there at some point because I thought CMAKE_INSTALL for cmake has the
same effect as PKG_INSTALL for make. But checking the cmake.mk include I
see that I was mistaken. So getting rid of CMAKE_INSTALL and the
InstallDev define seems like the way to go.

Kind regards,
Seb

PR #14863, opened 2021-02-23
minizip: update to 3.0.0

neheb:
Switch to new upstream and switch package name.

Further disable features to avoid extra dependencies.

Build with PIC to avoid build failures with mips16.

Remove no longer needed patch. Upstream fixed it differently.

Use AUTORELEASE for simplicity.

Maintainer: @dwmw2 
Compile tested: ath79


hnyman:
This seems like a duplicate of your own #14822

PR #14882, opened 2021-02-24
CI: enhance package detection logic

lipnitsk:
Maintainer: @aparcar

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -34,10 +34,18 @@ jobs:
 
       - name: Determine changed packages
         run: |
-          # only detect packages with changed Makefiles
-          PACKAGES="$(git diff --diff-filter=d --name-only origin/master \
-            | grep -E 'Makefile$|test.sh$' | grep -Ev '/files/|/src/' \
-            | awk -F/ '{ print $(NF-1) }' | tr '\n' ' ')"
+          # only detect packages with changes
+          PKG_ROOTS=$(find . -name Makefile | grep -v ".*/src/Makefile" | sed -e 's@./\(.*\)/Makefile@\1/@')
+          CHANGES=$(git diff --diff-filter=d --name-only origin/master)
+
+          for ROOT in $PKG_ROOTS; do
+            for CHANGE in $CHANGES; do
+              if [[ "$CHANGE" == "$ROOT"* ]]; then
+                PACKAGES+=$(echo "$ROOT" | sed -e 's@.*/\(.*\)/@\1 @')
+                break
+              fi
+            done
+          done
 
           # fallback to test packages if nothing explicitly changes this is
           # should run if other mechanics in packages.git changed


.github/workflows/multi-arch-test-build.yml
```diff
@@ -35,9 +35,17 @@ jobs:
...
         run: |
           # only detect packages with changed Makefiles
-          PACKAGES="$(git diff --diff-filter=d --name-only origin/master \
-            | grep -E 'Makefile$|test.sh$' | grep -Ev '/files/|/src/' \
-            | awk -F/ '{ print $(NF-1) }' | tr '\n' ' ')"
+          PKG_ROOTS=$(find -name Makefile | sed -e 's@./\(.*\)/Makefile@\1/@')
+          CHANGES=$(git diff --diff-filter=d --name-only origin/master)
+
+          for ROOT in $PKG_ROOTS; do
+            for CHANGE in $CHANGES; do
```
aparcar, line 42: Shouldn't we swap these as it's likely always only a single package that changes?

.github/workflows/multi-arch-test-build.yml
```diff
@@ -35,9 +35,17 @@ jobs:
...
+              if [[ "$CHANGE" == "$ROOT"* ]]; then
+                PACKAGES+=$(echo $ROOT | sed -e 's@.*/\(.*\)/@\1 @')
```
aparcar, line 44: Doesn't this adds duplicates if two changes happened in a single package.

.github/workflows/multi-arch-test-build.yml
```diff
@@ -35,9 +35,17 @@ jobs:
       - name: Determine changed packages
         run: |
           # only detect packages with changed Makefiles
-          PACKAGES="$(git diff --diff-filter=d --name-only origin/master \
-            | grep -E 'Makefile$|test.sh$' | grep -Ev '/files/|/src/' \
-            | awk -F/ '{ print $(NF-1) }' | tr '\n' ' ')"
+          PKG_ROOTS=$(find -name Makefile | sed -e 's@./\(.*\)/Makefile@\1/@')
```
aparcar, line 38: I think we have to make sure that there is no `/src/` in the path as some src folders contain a Makefile.

.github/workflows/multi-arch-test-build.yml
lipnitsk: yeah, the `break` below kind of fixes that, by moving onto the next ROOT as soon as a match is found, so duplicates should not match again.

.github/workflows/multi-arch-test-build.yml
lipnitsk: Let me test that.

.github/workflows/multi-arch-test-build.yml
lipnitsk: Yeah, then the de-dup logic will need some work - I'll take a look

.github/workflows/multi-arch-test-build.yml
aparcar: Did you run a benchmark on how long this loop actually takes?

lipnitsk: I think it's negligible - I saw CI kick off and like 10 seconds later it was already past that step. On my machine it takes about 200ms to run the script, but CI is likely slower. Let me add some debug in my temp commit.

aparcar: Okay nevermind then, thanks for checking

.github/workflows/multi-arch-test-build.yml
lipnitsk: I don't think it's really worth optimizing this - even in CI this whole thing is done in 1 second or less. Take a look at the CI logs - temp commit prints out `date` before and after this check.

.github/workflows/multi-arch-test-build.yml
lipnitsk: good catch, added a 'grep -v "src/"', seems to fix that.

aparcar:
Excellent thank you very much.

PR #14887, opened 2021-02-24
vpn-policy-routing: custom user scripts improvements

stangri:
Maintainer: me
Compile tested: ramips, er-x, 19.07.7
Run tested: ramips, er-x, 19.07.7, start/stop

Description: Based on [user request](https://github.com/stangri/source.openwrt.melmac.net/issues/128) I've made minor changes to only redownload the IP ranges within the custom user scripts if the target file doesn't exist or empty, making it a bit easier for users to control the expiry/redownload options for these.

I have also modified the AWS custom user file to send proper return code.

PS. Please grant the merge rights for packages.

### DIFF
--- a/net/vpn-policy-routing/Makefile
+++ b/net/vpn-policy-routing/Makefile
@@ -5,7 +5,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=vpn-policy-routing
 PKG_VERSION:=0.3.2
-PKG_RELEASE:=12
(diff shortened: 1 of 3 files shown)


dibdot:
@stangri any reason for not using builtin jsonfilter for bgview downloads? `jsonfilter -e '@.data.ipv4_prefixes[*].prefix'` should work as well and you don't need sed workarounds or an additional jq package.

PR #14890, opened 2021-02-24
strongswan: force PIC on all builds

pprindeville:
Maintainer: me, @Thermi
Compile tested: x86_64, generic, master (2ff9686040)
Run tested: for now, just seeing if CI/CD builds... will merge into my tree and rebuild/retest if it does

Description:

### DIFF
--- a/net/strongswan/patches/305-minimal_dh_plugin.patch
+++ b/net/strongswan/patches/305-minimal_dh_plugin.patch
@@ -68,7 +68,7 @@
 +	../gmp/gmp_diffie_hellman.c ../gmp/gmp_diffie_hellman.h 
 +
 +	
-+libstrongswan_gmpdh_la_LDFLAGS = -module -avoid-version -Wl,-Bstatic -Wl,-lgmp -Wl,-Bdynamic -Wl,--as-needed
++libstrongswan_gmpdh_la_LDFLAGS = -module -avoid-version -Wl,-Bstatic -Wl,-lgmp -Wl,-Bdynamic -Wl,--as-needed $(FPIC)
 +libstrongswan_gmpdh_la_LIBADD  =
 --- /dev/null
 +++ b/src/libstrongswan/plugins/gmpdh/gmpdh_plugin.c
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=strongswan
 PKG_VERSION:=5.9.1
-PKG_RELEASE:=8
+PKG_RELEASE:=9
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=https://download.strongswan.org/ https://download2.strongswan.org/


stintel:
Commit fails to mention why.

pprindeville:
> Commit fails to mention why.

Doesn't matter.  I can't reproduce the failure on my desktop, but I wanted to see if this fixed it in CI/CD.  It doesn't.  There's some divergence that's not well understood.

micmac1:
This only happens with the gmpdh plugin, which is an OpenWrt addition. Must be something related to the linker flags supplied by the patch. When I change them like below it compiles fine. But then one'd have to add a dep on libgmp, which the author of the patch apparently wanted to avoid (the plugins comment says "no libgmp dep").

```
diff --git a/net/strongswan/patches/305-minimal_dh_plugin.patch b/net/strongswan/patches/305-minimal_dh_plugin.patch
index dcd2448..b317508 100644
--- a/net/strongswan/patches/305-minimal_dh_plugin.patch
+++ b/net/strongswan/patches/305-minimal_dh_plugin.patch
@@ -68,8 +68,8 @@
 +      ../gmp/gmp_diffie_hellman.c ../gmp/gmp_diffie_hellman.h 
 +
 +      
-+libstrongswan_gmpdh_la_LDFLAGS = -module -avoid-version -Wl,-Bstatic -Wl,-lgmp -Wl,-Bdynamic -Wl,--as-needed
-+libstrongswan_gmpdh_la_LIBADD  =
++libstrongswan_gmpdh_la_LDFLAGS = -module -avoid-version
++libstrongswan_gmpdh_la_LIBADD  = -lgmp
 --- /dev/null
 +++ b/src/libstrongswan/plugins/gmpdh/gmpdh_plugin.c
 @@ -0,0 +1,101 @@
```

micmac1:
Here is the commit that introduced this patch in 2015: https://github.com/openwrt/packages/commit/f705b3c0bf15d4f231f83ef59753c4496bc0cf22

pprindeville:
> Here is the commit that introduced this patch in 2015: [f705b3c](https://github.com/openwrt/packages/commit/f705b3c0bf15d4f231f83ef59753c4496bc0cf22)

Interesting.  What happens if, instead, we explicitly passing `-fPIC` to `libstrongswan_gmpdh_la_LDFLAGS`?

net/strongswan/patches/305-minimal_dh_plugin.patch
```diff
@@ -68,7 +68,7 @@
 +	../gmp/gmp_diffie_hellman.c ../gmp/gmp_diffie_hellman.h 
 +
 +	
-+libstrongswan_gmpdh_la_LDFLAGS = -module -avoid-version -Wl,-Bstatic -Wl,-lgmp -Wl,-Bdynamic -Wl,--as-needed
++libstrongswan_gmpdh_la_LDFLAGS = -module -avoid-version -Wl,-Bstatic -Wl,-lgmp -Wl,-Bdynamic -Wl,--as-needed $(FPIC)
```
aparcar, line 71:
```suggestion
+libstrongswan_gmpdh_la_LDFLAGS = -module -avoid-version -Wl,-Bstatic -Wl,-lgmp -Wl,-Bdynamic -Wl,--as-needed "$(FPIC)"
```

Isn't that saver?

pprindeville: No, because spaces between args need to be maintained.

pprindeville: I'd really rather have all of $(TARGET_CFLAGS) and $(TARGET_LDFLAGS) get interpolated...

aparcar: Ack

aparcar:
Since you're the new maintainer and suggested this, I'm happy to merge this as is. Proceed?

pprindeville:
> Since you're the new maintainer and suggested this, I'm happy to merge this as is. Proceed?

Sure.

pprindeville:
Still would like to see `$(TARGET_CFLAGS)` and $(TARGET_LDFLAGS)` used additionally, but that's a separate issue.

pprindeville:
Thanks!

PR #14892, opened 2021-02-25
libupm: enforce disabling python2

cotequeiroz:
Maintainer: @blogic, @nxhack 
Compile tested: arm_cortex-a9+vfpv3-d16, gentoo linux with python 3.8 and 3.9 installed
Run tested: none

Description:
If there are more than two python3 interpreters, and no python2,
libupm thinks one of them is python2 and fails.  Avoid that by
telling it that python2 was not found.


nxhack:
LGTM

PR #14902, opened 2021-02-26
vpn-policy-routing: update user netflix file

stangri:
Maintainer: me
Compile tested: ramips, er-x, 19.07.7
Run tested: ramips, er-x, 19.07.7, none

Description: This implements use of jsonfilter instead of jq/sed as suggested by @dibdot: https://github.com/openwrt/packages/pull/14887#issuecomment-785826254

### DIFF
--- a/net/vpn-policy-routing/files/vpn-policy-routing.netflix.user
+++ b/net/vpn-policy-routing/files/vpn-policy-routing.netflix.user
@@ -25,11 +25,7 @@ if [ ! -s "$TARGET_FNAME" ]; then
 
 	if [ "$DB_SOURCE" = "api.bgpview.io" ]; then
 		TARGET_URL="https://api.bgpview.io/asn/${TARGET_ASN}/prefixes"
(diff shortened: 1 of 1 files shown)


BKPepe:
This does not require to bump `PKG_RELEASE`?

dibdot:
@stangri Most probably you need another update anyway. If I understand the logic correctly your last commit missed the redirect to `> "$TARGET_FNAME"`

stangri:
> This does not require to bump `PKG_RELEASE`?

Thanks for the suggestion, originally I thought not, because it's a feature which is used by a fraction of the users and I've sent the update pretty quickly after the previous merge, but now that there's an actual bugfix, I believe bump is required.

> @stangri Most probably you need another update anyway. If I understand the logic correctly your last commit missed the redirect to `> "$TARGET_FNAME"`

Good catch, I've only tested the output to the console and was in a hurry to send a PR. Lesson learnt.

PR #14905, opened 2021-02-26
vim: fix compile with BUILD_NLS

neheb:
Avoids having to use msgfmt, which happens to be broken with gettext-full.

Maintainer: @ratkaj 
Compile tested: ath79

ping @BKPepe

### DIFF
--- a/utils/vim/Makefile
+++ b/utils/vim/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=vim
 PKG_VERSION:=8.2
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 VIMVER:=82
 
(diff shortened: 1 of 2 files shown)


BKPepe:
Hi @neheb,

Thank you for creating the PR. I was aware of that patch in the issue, however, you were faster than me. I want to give it a try with a full build to see if it helps or not. I will get back to you today or tomorrow.

In any case, the commit message needs to be improved a little bit like what is msgfmt and why we don't need it for vim.
You can use `Suggested-by` or add author and commit description to the patch like it fixes issue and link, so we know why we need it and remove it in the future.

BKPepe (approved):
`vim` is compiled while running full build. Thank you! :+1: 

While merging this, don't forget to change a commit description a bit or add commit message to the patch.

PR #14907, opened 2021-02-26
docker,dockerd: Updated to 20.10.3

G-M0N3Y-2503:
Maintainer: @G-M0N3Y-2503
Compile tested: x86_x64, VirtualBox, OpenWrt Master
Run tested: x86_x64, VirtualBox, OpenWrt Master

Description:
* Added script to verify/obtain the correct short SHA for a given version
* Minor style update
* Updated docker CLI and the daemon to 20.10.3

### DIFF
--- a/utils/dockerd/git-short-commit.sh
+++ b/utils/dockerd/git-short-commit.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+#
+#	USAGE: git-short-commit.sh <GIT_URL> <GIT_REF> <GIT_DIR>
+#
+
+set -e
+
+error() {
+	echo "ERROR: ${*}" >&2
+	exit 1
+}
+
+GIT_URL="${1}"
+if [ -z "${GIT_URL}" ]; then
+	error "Git URL not specified"
+fi
+
+GIT_REF="${2}"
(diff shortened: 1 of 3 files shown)


utils/dockerd/git-short-commit.sh
```diff
@@ -0,0 +1,48 @@
...
+	exit 1
+}
+
+GIT_URL="${1}"
+if [ ! -n "${GIT_URL}" ]; then
+	error "Git URL not specified"
+fi
+
+GIT_REF="${2}"
+if [ ! -n "${GIT_REF}" ]; then
```
neheb, line 19: ! -n should be -z . Run through shellcheck.

G-M0N3Y-2503: Nice catch.
I guess that specific check isn't in version 0.5.0 of shellcheck.  
Is there a recommended version I should use?

neheb: latest one you can I guess.

neheb: if you're running debian 10, run

sudo apt -t buster-backports install shellcheck

PR #14916, opened 2021-02-27
utils/findutils: update to 4.8.0

bernhard-voelker:
* utils/findutils/Makefile: Update PKG_VERSION and PKG_RELEASE.
While at it, update the Copyright year number range.


utils/findutils/Makefile
```diff
@@ -1,18 +1,18 @@
 #
-# Copyright (C) 2006-2016 OpenWrt.org
+# Copyright (C) 2006-2021 OpenWrt.org
```
neheb, line 2: Please no.

PR #14929, opened 2021-02-27
libdrm: add option to build the intel driver

bkuhls:
libpciaccess is not needed for libdrm itself, only the intel driver
depends on it:
https://cgit.freedesktop.org/mesa/drm/tree/meson.build#n240

Signed-off-by: Bernd Kuhls <bernd.kuhls@t-online.de>

Maintainer: @lucize 
Compile tested: x86_64


lucize:
LGTM

PR #14942, opened 2021-02-28
aria2: fix linking with full language support enabled

mhei:
Maintainer: @kuoruan
Compile tested: mxs
Run tested: -

Description:

After d18692c, packages which links to libxml2 need to include
the dependency to libiconv too when built with full language
support enabled.


neheb:
Won't it just get included as a transient dependency?

mhei:
Hm, you're right. I checked the binary again, and it does not link directly to libiconv. However, without including nls.mk the build fails. I'll update the PR.

PR #14980, opened 2021-03-01
rsync: fix compilation for undefined reference ot makedev

feckert:
Maintainer: @mstorchak
Compile tested: x86_64, APU3, OpenWrt [master](https://github.com/openwrt/openwrt/commit/ebcb4f1d0a091d1fa87d6aa04ecae355ef23ef22)
Run tested: no

Description:

This change fixes the following compilation error for rsync
`
generator.c:(.text+0x4872): undefined reference to `makedev'
`
Glibc removes sys/sysmacros.h which defines makedev from sys/types.h
since v2.28.

### DIFF
--- a/net/rsync/patches/001-fix-makedev-issue.patch
+++ b/net/rsync/patches/001-fix-makedev-issue.patch
@@ -1,0 +1,10 @@
+--- a/rsync.h
++++ b/rsync.h
+@@ -307,6 +307,7 @@ enum delret {
+ #include <stdio.h>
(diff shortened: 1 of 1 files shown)


neheb:
This is not the issue. The issue is with autoconf-lean and its config. I will send a patch to the mailing list once my testing is done.

feckert:
@arparcar thanks for the hint.
This has happened to me for the second time now!
That annoys me now.
Sorry.
Is there not the possibility to prevent this in github?
In gitlab I think there are the protected branches.

PR #14996, opened 2021-03-02
mwan3: add connecting and disconnecting event to mwan3track

feckert:
Maintainer: me / @aaronjg 
Run tested: x86_64, APU3, OpenWrt master, mwan3track check if event is generated

Description:
If the interface goes into failure state (is disconnecting)
then with this change one hotplug.d event is generated.

The same is true for the recovery state (is connecting), when the interface
comes back from a failure state.

In both cases, a hotplug.d event for the iface is triggered. Once
with the $ACTION=disconnecting and once for the $ACTION=connecting.

### DIFF
--- a/net/mwan3/files/usr/sbin/mwan3track
+++ b/net/mwan3/files/usr/sbin/mwan3track
@@ -128,6 +128,24 @@ connected() {
 	env -i FIRSTCONNECT=$1 ACTION="connected" INTERFACE="$INTERFACE" DEVICE="$DEVICE" /sbin/hotplug-call iface
 }
 
+disconnecting() {
+	if [ "$STATUS" != "disconnecting" ] ; then
+		STATUS="disconnecting"
+		echo "disconnecting" > $MWAN3TRACK_STATUS_DIR/$INTERFACE/STATUS
+		LOG notice "Interface $INTERFACE ($DEVICE) is disconnecting"
+		env -i ACTION="disconnecting" INTERFACE="$INTERFACE" DEVICE="$DEVICE" /sbin/hotplug-call iface
+	fi
+}
+
+connecting() {
+	if [ "$STATUS" != "connecting" ] ; then
+		STATUS="connecting"
+		echo "connecting" > $MWAN3TRACK_STATUS_DIR/$INTERFACE/STATUS
+		LOG notice "Interface $INTERFACE ($DEVICE) is connecting"
+		env -i ACTION="connecting" INTERFACE="$INTERFACE" DEVICE="$DEVICE" /sbin/hotplug-call iface
+	fi
+}
+
 disabled() {
 	STATUS='disabled'
 	echo "disabled" > $MWAN3TRACK_STATUS_DIR/$INTERFACE/STATUS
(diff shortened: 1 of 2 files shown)


aaronjg:
If I understand correctly, this will cause a hotplug event to be triggered on each failing track request until fail threshold is reached, and then on each successful track request until up threshold is reached.

Not opposed to such a change, but curious about what the use case is.

feckert:
@aaronjg
Yes, an event is triggered when it detects a connection error (Connection is disconnecting on failure).
And if the connection is going to recover then an another event is triggered again (Connection is connecting on recover).
The event and therefore the ACTION is only called once when it goes into this state. 

I need this for my LED trigger.
I set the LED timer trigger (LED flashing) on connecting or disconnecting event (ACTION).
This indicates that the connection is going up (recovering) or down (failure).

feckert:
@aaronjg I would merge this tomorrow, if there are no objections from your side
 I have already described my reasons why I need this.

aaronjg:
Sorry, I’d like to look this over a bit more. I think the principle is fine, but I need to review exactly when the connecting and disconnecting events are triggered. The logic in mwan3track is a bit convoluted at the moment and that should probably be replaced with a proper state machine. However that shouldn’t precluded merging this patch. Give me another couple days to review.

net/mwan3/files/usr/sbin/mwan3track
```diff
@@ -345,6 +365,7 @@ main() {
 				echo "online" > $MWAN3TRACK_STATUS_DIR/$INTERFACE/STATUS
 				score=$((down+up))
 			elif [ $score -le $up ]; then
+				connecting
```
aaronjg, line 368: It looks like this will cause it to trigger a "connecting" event with each successful ping when it is in the down state, no?

feckert: The connecting function is called on every check if we are in tear up (get connecting).
That`s right.
But if the connection is setting up, then the hotplug event is only called once.
The function [set the STATUS to connecting](https://github.com/openwrt/packages/pull/14996/files#diff-df4fb4fa188ff19ed301dde4878253e6b8931bed61bcc92c7c5de32beec9840eR143) and so does not do another hotplug event because we [check STATUS](https://github.com/openwrt/packages/pull/14996/files#diff-df4fb4fa188ff19ed301dde4878253e6b8931bed61bcc92c7c5de32beec9840eR141) if we have already entered this code section.

So I think this does not happen.

aaronjg: I see. `mwan3track` should really be refactored to use a state machine architecture, to make this sort of thing more clear,  but that is a larger project. This P/R looks fine. Approved.

aaronjg (approved):

net/mwan3/files/usr/sbin/mwan3track
feckert: I agree with you completely.
The question is also whether we should do it in native shell again or perhaps use lua.
I've been toying with the idea for a while.

aaronjg: Native shell is nice in that is accessible to more developers, but it would be great to do it in a language that allows multithreading. Having separate mwan3track processes for each device is a large part of mwan3's memory footprint.

If in Lua, we could hook in directly to netifd ubus notifications and get rid of the separate hotplug script too.

PR #15000, opened 2021-03-02
hs20: fix linking with full language support enabled

mhei:
Maintainer: @dangowrt 
Compile tested: mxs
Run tested: -

Description:
After d18692c, we need to include nls.mk to setup correct
environment variables so that linking succeeds.

Reported-by: Josef Schlehofer <pepe.schlehofer@gmail.com>


dangowrt:
Thank you!

Issue #14956, opened 2021-02-28
nft-qos: unable to install properly using image builder


storm1er:
Maintainer: @rosysong 
Environment: target `rpi-4`, OpenWrt latest snapshot **but may apply to every target**

## Description

When adding the nft-qos package to image builder script, we get some configuration error.  
This is due to the fact that install scripts are refering to host root instead of builder folder (*virtual system root folder?*).

Command is something like `opkg install $OPTIONS $PACKAGES` which I can't give as it's build automatically by the image builder, here's the output

```
Enabling ...
Enabling ...
Enabling ...
/home/user/openwrt-imagebuilder-bcm27xx-bcm2711.Linux-x86_64/build_dir/target-aarch64_cortex-a72_musl/root-bcm27xx/lib/nft-qos/core.sh: line 7: /lib/functions/procd.sh: No such file or directory
/home/user/openwrt-imagebuilder-bcm27xx-bcm2711.Linux-x86_64/build_dir/target-aarch64_cortex-a72_musl/root-bcm27xx/lib/nft-qos/monitor.sh: line 6: /lib/nft-qos/core.sh: No such file or directory
/home/user/openwrt-imagebuilder-bcm27xx-bcm2711.Linux-x86_64/build_dir/target-aarch64_cortex-a72_musl/root-bcm27xx/lib/nft-qos/dynamic.sh: line 6: /lib/nft-qos/core.sh: No such file or directory
/home/user/openwrt-imagebuilder-bcm27xx-bcm2711.Linux-x86_64/build_dir/target-aarch64_cortex-a72_musl/root-bcm27xx/lib/nft-qos/static.sh: line 6: /lib/nft-qos/core.sh: No such file or directory
/home/user/openwrt-imagebuilder-bcm27xx-bcm2711.Linux-x86_64/build_dir/target-aarch64_cortex-a72_musl/root-bcm27xx/lib/nft-qos/mac.sh: line 6: /lib/nft-qos/core.sh: No such file or directory
/home/user/openwrt-imagebuilder-bcm27xx-bcm2711.Linux-x86_64/build_dir/target-aarch64_cortex-a72_musl/root-bcm27xx/lib/nft-qos/priority.sh: line 6: /lib/functions/network.sh: No such file or directory
[...]

PR #15002, opened 2021-03-03
I2pd: fix imagebuilder and startup

plntyk:
Maintainer:  @yangfl 
Compile tested: x86-64 VM
Run tested: x86-64 VM

Description:
coding style: remove the unneeded include; fixes an issue similar to #14956 
https://openwrt.org/docs/guide-developer/config-scripting

further testing uncovered the rights issue with i2pd.conf file


yangfl:
Hi, could you update to the latest release? Thanks

plntyk:
i added the third patch to do an update as requested
 : AVX patch was removed as upstream closed all issues and seems to have integrated all AVX upstream

tested an image in qemu  x86-64  and i386 - start/stop of i2pd seems to have no issues

PR #15007, opened 2021-03-04
pigeonhole: bump to 0.5.14

cotequeiroz:
Maintainer: @flyn-org CC: @neheb 
Compile tested: arm-cortex-a15-neon-vfp4, openwrt-21.02
Run tested: none

Description:
ChangeLog:
 - IMAP FILTER command: cmd-filter-sieve - Do not allow NIL as script
   name argument.

---
There are 5 changed files from 0.5.13, one change in actual code, and the rest is rather cosmetic:
 - A one-liner change in code from 0.5.13, really, to check for the NIL script.  
- A version correction in configure.ac because it was not updated in 0.5.13, having version 0.6-devel and dovecot 2.4 instead. 
- The rest are all formatting changes and typos.

Therefore I opted to review the code instead of run-testing this (I would need to do a lot of setup to actually run it).

### DIFF
--- a/mail/pigeonhole/Makefile
+++ b/mail/pigeonhole/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dovecot-pigeonhole
-PKG_VERSION_PLUGIN:=0.5.13
+PKG_VERSION_PLUGIN:=0.5.14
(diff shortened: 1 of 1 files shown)


flyn-org:
I will test this tonight since there is talk of porting to 21.02.

flyn-org:
@cotequeiroz I just installed dovecot-pigeonhole_2.3.13-0.5.14-1 on master, and it seems to work find.

PR #15008, opened 2021-03-04
gcc: add cc symlink

### LINKED ISSUE #14755: gcc: missing link from gcc to cc
Maintainer: @noblepepper 
Environment: x86_64 master

Description:
The gcc package does not create a symbolic link from /usr/bin/gcc to /usr/bin/cc. This breaks some Makefiles that use $(CC), as GNU make sets CC to cc by default. I think POSIX required the presence of cc as some point.

plntyk:
fixes #14755

Signed-off-by: Dirk Neukirchen <plntyk.lede@plntyk.name>

Maintainer: @noblepepper
Compile tested: x86-64 VM
Run tested: x86-64 VM

fixes #14755

### DIFF
--- a/devel/gcc/Makefile
+++ b/devel/gcc/Makefile
@@ -24,7 +24,7 @@ endef
 PKG_NAME:=gcc
 # PKG_VERSION=7.3.0
 PKG_VERSION=7.4.0
-PKG_RELEASE:=5
+PKG_RELEASE:=6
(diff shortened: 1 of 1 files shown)


neheb:
Needs a PKG_RELEASE bump.

plntyk:
thx. forgot that

PR #15010, opened 2021-03-04
php8: update to 8.0.3

mhei:
Maintainer: me
Compile tested: mxs
Run tested: mxs

### DIFF
--- a/lang/php8/Makefile
+++ b/lang/php8/Makefile
@@ -6,7 +6,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=php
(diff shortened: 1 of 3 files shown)


neheb:
```
2021-03-05T00:12:44.4646132Z checking for mmap() using shm_open() shared memory support... no
2021-03-05T00:12:44.5599046Z configure: error: No supported shared memory caching support was found when configuring opcache. Check config.log for any errors or missing dependencies.
```

mhei:
Odd, I cannot reproduce this error in my local build environment, but I pushed an updated patch. Let's see whether this solves this issue...

PR #15012, opened 2021-03-04
treewide: build CMake projects with Ninja

neheb:
faster to compile.

### DIFF
--- a/devel/ninja/ninja-cmake.mk
+++ b/devel/ninja/ninja-cmake.mk
@@ -5,7 +5,7 @@ CMAKE_HOST_OPTIONS += -DCMAKE_GENERATOR="Ninja"
 CMAKE_OPTIONS += -DCMAKE_GENERATOR="Ninja"
 
 define Host/Compile/Default
-	$(call Ninja,-C $(HOST_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR),)
+	$(call Ninja,-C $(HOST_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR) $(1),)
 endef
 
 define Host/Install/Default
@@ -17,7 +17,7 @@ define Host/Uninstall/Default
 endef
 
 define Build/Compile/Default
-	$(call Ninja,-C $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR),)
+	$(call Ninja,-C $(PKG_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR) $(1),)
 endef
 
 define Build/Install/Default
--- a/lang/lua-cjson/Makefile
+++ b/lang/lua-cjson/Makefile
@@ -22,7 +22,7 @@ HOST_BUILD_DEPENDS:=lua/host
 
 include $(INCLUDE_DIR)/host-build.mk
 include $(INCLUDE_DIR)/package.mk
-include $(INCLUDE_DIR)/cmake.mk
+include ../../devel/ninja/ninja-cmake.mk
 
 define Package/lua-cjson
   SUBMENU:=Lua
--- a/lang/lzmq/Makefile
+++ b/lang/lzmq/Makefile
@@ -18,7 +18,7 @@ PKG_SOURCE_URL:=https://codeload.github.com/zeromq/lzmq/tar.gz/v$(PKG_VERSION)?
 PKG_HASH:=cf70200045b8bcb0e929c338ad421b6a291cf1038053532888dc201af3224d8b
 
 include $(INCLUDE_DIR)/package.mk
-include $(INCLUDE_DIR)/cmake.mk
+include ../../devel/ninja/ninja-cmake.mk
 
 define Package/lzmq
   SUBMENU:=Lua
--- a/libs/afalg_engine/Makefile
+++ b/libs/afalg_engine/Makefile
@@ -24,7 +24,7 @@ PKG_CONFIG_DEPENDS:= \
 	CONFIG_AFALG_ZERO_COPY
 
 include $(INCLUDE_DIR)/package.mk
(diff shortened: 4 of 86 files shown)


champtar:
@neheb numbers ?
Also do we depend on host Ninja version ? if yes this might cause issues depending on the distro

neheb:
Ninja already has a host package here. That will not be an issue.

I'll get numbers at a later date. This is a draft after all :).

champtar:
Perfect, waiting for the numbers then :)

neheb:
It's a little unfair honestly. Some of these packages did not have PKG_BUILD_PARALLEL set. Others that had PKG_BUILD_PARALLEL set to 0 end up compiling because...Ninja.

champtar:
Please read it as "excited to see some faster compile time soon", with Ninja host build I'm 100% for this PR, I'm just curious at how much we can gain by switching to Ninja

neheb:
@stangri I added a workaround for a CMake bug with Ninja. Please review.

stangri:
@neheb -- Thanks! Would it be possible to backport the relevant ninja updates to 19.07 so that the Makefile for https-dns-proxy can stay the same for 19.07 and newer (21.02/snapshot) builds?

neheb:
Done.

neheb:
Was able to compile all locally.

neheb:
With https://github.com/openwrt/packages/pull/15070 this will be ready. Just need to get the numbers.

neheb:
```
A small selection of packages was tested going from:

Executed in  696.30 secs   fish           external
   usr time   82.98 mins  395.00 micros   82.98 mins
   sys time    9.02 mins    0.00 micros    9.02 mins

to:

Executed in  592.20 secs   fish           external
   usr time   84.84 mins  361.00 micros   84.84 mins
   sys time    8.85 mins   57.00 micros    8.85 mins

Tested by running make -j 12 and wiping staging/build_dir/target_x
```

The full suite would show a bigger speedup I think. ping @champtar

karlp:
Can we not just pull this inside the "$(INCLUDE cmake)" bit? is there ever going to be a reason to want this disabled?  if we do, could it just be controlled by the existing PKG_PARALLEL?

neheb:
The issue is that ninja is not in base. They also do not want it for "bloat" reasons. It'll be done eventually when meson becomes a requirement for base.

karlp:
so if it's not in base, what pulls in ninja host build at all then?  It would seem that making all the packages simply "require" ninja like this would be doing an end run around the "don't want ninja in base" requirement and forcing their hand?

neheb:
> so if it's not in base, what pulls in ninja host build at all then? It would seem that making all the packages simply "require" ninja like this would be doing an end run around the "don't want ninja in base" requirement and forcing their hand?

ninja-cmake includes ninja.mk which has

https://github.com/openwrt/packages/blob/master/devel/ninja/ninja.mk#L9

Ninja is just a build system. There's no requirement at all. CMake can work with GNU Make although slower.

Their hand will be forced when util-linux switches to meson which requires ninja.

aparcar:
So `openobex` seem to fail? Can't you just remove the change for now on that specific package?

aparcar:
Well the individual package seem to require individual fixes.

aparcar:
How about you split this PR into single PRs which can then be approved individually by both maintainers and CI?

neheb:
Too many honestly. This PR is just a compilation change. There's no change in compiled output.
There are around 110 packages right now. I'll rebase and push when there are around 80.

neheb:
@micmac1 since you've dealt with mariadb, any idea why it's not working properly? It seems plugins are not being disabled.

micmac1:
> @micmac1 since you've dealt with mariadb, any idea why it's not working properly? It seems plugins are not being disabled.

Hello Rosen,

The host compile of mariadb fails. Without the host compile the cross-compile will fail.

The host compile wants to build the special target "import_executables". Hence there is this in the mariadb Makefile:

```
define Host/Compile
        $(call Host/Compile/Default,import_executables)
endef
```

But ninja-cmake.mk seems to not set up the defines properly. The result is that it kicks off the host compile of mariadb like this:

`/openwrt/staging_dir/hostpkg/bin/ninja  -C /home/sk/tmp/openwrt/build_dir/hostpkg/mariadb-10.4.18/`

It should be something like this:

`/openwrt/staging_dir/hostpkg/bin/ninja  -C /home/sk/tmp/openwrt/build_dir/hostpkg/mariadb-10.4.18/ import_executables`

The below seems to make it work, although I suspect that may not be the best solution:

```
diff --git a/devel/ninja/ninja-cmake.mk b/devel/ninja/ninja-cmake.mk
index 1ba079203..290e86ed9 100644
--- a/devel/ninja/ninja-cmake.mk
+++ b/devel/ninja/ninja-cmake.mk
@@ -5,7 +5,7 @@ CMAKE_HOST_OPTIONS += -DCMAKE_GENERATOR="Ninja"
 CMAKE_OPTIONS += -DCMAKE_GENERATOR="Ninja"
 
 define Host/Compile/Default
-       $(call Ninja,-C $(HOST_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR),)
+       $(call Ninja,-C $(HOST_BUILD_DIR)/$(CMAKE_BINARY_SUBDIR) $(1),)
 endef
 
 define Host/Install/Default
diff --git a/utils/mariadb/Makefile b/utils/mariadb/Makefile
index 03bdd51c3..40919abe6 100644
```

I think the other defines in ninja-cmake.mk might also need some reworking. The question would be if users can call them with extra arguments or not, for instance $(call Build/Compile/Default,extra arguments).

neheb:
Interesting...

Something like that could also get rid of the openobex patch.

neheb:
Let's see if anything blows up...

PR #15017, opened 2021-03-05
meson: update to 0.56.2

neheb:
This is the last version supporting Python 3.5, which is still in wide
use. Going forward, python3/host would need to be used. Not sure if
worth it.

Maintainer: @dhewg 
Compile tested: mips64


neheb:
@dhewg a few things bother me.

```
DEPRECATION: The [paths] section is deprecated, use the [built-in options] section instead.
DEPRECATION: The [paths] section is deprecated, use the [built-in options] section instead.
WARNING: Unknown options: "build.cpp_args, build.cpp_link_args"
```

It also bothers me that all packages need to have an explicit PKG_BUILD_DEPENDS on meson/host . This should be done in meson.mk I think.

dhewg:
LGTM, thanks!

For the record:
I tried to bump the python3 requirements, but the patch was rejected because the buildbots run debian oldstable, which uses python3.5 [0].
Debian oldstable is the current LTS though, and I think upstream missed that detail [1].

I've seen the deprecated warnings too, resolving those is probably easy. But I won't get to checking if that won't affect any ports using meson soon though.

[0] http://patchwork.ozlabs.org/project/openwrt/patch/20210216072108.26107-1-a.heider@gmail.com/
[1] https://github.com/mesonbuild/meson/pull/7681#discussion_r577658270

neheb:
That doesn't bother me. We have python3/host.

```
WARNING: Unknown options: "build.cpp_args, build.cpp_link_args"
```

bothers me. It comes from the txt.in files and shows up whenever a package does not use any host binaries.

dhewg:
> That doesn't bother me. We have python3/host.

It would bother me, since it would drag in that dependency for just meson in my case, so hopefully we won't need to add it.

> WARNING: Unknown options: "build.cpp_args, build.cpp_link_args"

I didn't test it yet, but is that new due to this PR? I can still see both options used in upsteam git master, so at first glance they don't seem unknown?

neheb:
It started warning with 0.56

dhewg:
That warning pops up since you moved `c_args` and friends to `[built-in options]` with 242afe6d3e1be57a4b33fec8cf510f386e33d235. Which was correct, since without it you get `DEPRECATION: c_args in the [properties] section of the machine file is deprecated, use the [built-in options] section.`

I even get `WARNING: Unknown options: "build.c_args, build.c_link_args, build.cpp_args, build.cpp_link_args"`, but that's obviously nonsense, as it uses those args for building just fine. I guess this can be just ignored, looks like another case of cross compilation being not a meson first class citizen.

PR #15029, opened 2021-03-06
wg-installer: add link-local to client interface

PolynomialDivision:
Wireguard has no link-local address on an interface automatically. Add a link-local to the interface. The server has fe80::1/64 and the client fe80::2/64.


champtar:
@PolynomialDivision I curious why link local IPs are needed on wireguard tunnel, my understanding is that wireguard is L3 so there is nothing that really requires a link local IP

PolynomialDivision:
> @PolynomialDivision I curious why link local IPs are needed on wireguard tunnel, my understanding is that wireguard is L3 so there is nothing that really requires a link local IP

Ah okay. I talked to another person who said that we need link-local. And babeld meshing was not working. I will remove it and put the logic somewhere else.

PR #15031, opened 2021-03-06
haproxy: Update HAProxy to v2.2.10

gladiac1337:
Maintainer: Thomas Heil <heil@terminal-consulting.de>, Christian Lachner <gladiac@gmail.com>
Run tested: ipq806x (r7800)

Description: Update HAProxy to v2.2.10

### DIFF
--- a/net/haproxy/Makefile
+++ b/net/haproxy/Makefile
@@ -10,12 +10,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=haproxy
-PKG_VERSION:=2.2.9
+PKG_VERSION:=2.2.10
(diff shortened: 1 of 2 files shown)


gladiac1337:
Hmm, not sure what "Dirty patches detected, please refresh and review the diff" actually means. The patch (this commit only has one) applies cleanly without any wrong offsets. Not sure why those test-builds are failing.

neheb:
Run make package/haproxy/refresh

gladiac1337:
I got rid of the one patch as it was not all that important and will be in the next release anyways. I currently do not have enough time to investigate further into why those patches cause the test-builds to fail. I will do that in a future PR.

PR #15035, opened 2021-03-06
nginx: fix compilation with BUILD_NLS

neheb:
Requires nls.mk because of libxml2.

Maintainer: @Ansuel @peter-stadler 
Compile tested: ath79


mhei:
The PKG_RELEASE bump is not strictly necessary since no meta-data change or binary change in the resulting package. Also the commit message is...short :-) But otherwise LGTM, let's get the fallout fixed...

mhei (approved):

PR #15040, opened 2021-03-07
libstrophe: fix build with full language support enabled

mhei:
Maintainer: @changeway 
Compile tested: mxs
Run tested: -

Description:

After d18692c, we need to include nls.mk to setup correct
environment variables so that linking succeeds.


changeway:
Thanks for update.

PR #15046, opened 2021-03-08
openocd: update to 0.11.0

paulfertser:
Signed-off-by: Paul Fertser <fercerpav@gmail.com>

Maintainer: me


utils/openocd/Makefile
```diff
@@ -8,14 +8,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openocd
-PKG_SOURCE_VERSION:=v0.11.0-rc1
+PKG_SOURCE_VERSION:=v0.11.0
 PKG_VERSION:=$(PKG_SOURCE_VERSION)
 PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
```
neheb, line 16: Not needed.

utils/openocd/Makefile
```diff
@@ -8,14 +8,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openocd
-PKG_SOURCE_VERSION:=v0.11.0-rc1
+PKG_SOURCE_VERSION:=v0.11.0
 PKG_VERSION:=$(PKG_SOURCE_VERSION)
```
neheb, line 12: Not needed. PKG_VERSION gets set from PKG_SOURCE_VERSION

utils/openocd/Makefile
```diff
@@ -8,14 +8,14 @@
...
 PKG_RELEASE:=1
```
neheb, line 13: Can be switched to AUTORELEASE for simplicity.

utils/openocd/Makefile
```diff
@@ -8,14 +8,14 @@
...
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
 PKG_SOURCE_URL:=git://git.code.sf.net/p/openocd/code
-PKG_MIRROR_HASH:=c4a0a6cad821946608c843f557b993e7cfeb0f193e4c0f149ebfbfbfd57054a6
+PKG_MIRROR_HASH:=ec31ce908d951c6c7e173c7474269f3f105d7a7b3991afbe8f54f8d96646c725
 
 PKG_LICENSE:=GPL-2.0
```
neheb, line 20: Should be -only or -or-later.

paulfertser: Noted

utils/openocd/Makefile
```diff
@@ -8,14 +8,14 @@
...
 
 PKG_NAME:=openocd
-PKG_SOURCE_VERSION:=v0.11.0-rc1
+PKG_SOURCE_VERSION:=v0.11.0
 PKG_VERSION:=$(PKG_SOURCE_VERSION)
 PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
 PKG_SOURCE_URL:=git://git.code.sf.net/p/openocd/code
```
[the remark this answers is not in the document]
paulfertser: Do you insist? I would prefer git so that it would be easy to package a reasonably tested but not officially released version in the future. What is the advantage of using https here?

utils/openocd/Makefile
paulfertser: Noted

utils/openocd/Makefile
paulfertser: Noted

utils/openocd/Makefile
paulfertser: Noted

utils/openocd/Makefile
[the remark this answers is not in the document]
neheb: Just checked. Looks like you can just do https://git.code.sf.net/p/openocd/code . It's used elsewhere in the tree.

The benefit is that https goes through some corporate firewalls. Note that I'm not asking to use a release tarball.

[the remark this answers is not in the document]
paulfertser: Got it, thank you!

PR #15055, opened 2021-03-08
atheepmgr: bugfix missing libpciaccess dependency

davidrapan:
Maintainer: @rsa9000  
Compile tested: x86-64 VM 19.07.7-SNAPSHOT
Run tested: x86-64 VM 19.07.7-SNAPSHOT

### DIFF
--- a/utils/atheepmgr/Makefile
+++ b/utils/atheepmgr/Makefile
@@ -25,6 +25,7 @@ include $(INCLUDE_DIR)/package.mk
 define Package/atheepmgr
   SECTION:=utils
   CATEGORY:=Utilities
+  DEPENDS:=+libpciaccess
   TITLE:=EEPROM/boarddata management utility for Atheros WLAN chips
   MENU:=1
 endef


rsa9000:
atheepmgr in fact **can** use libpciaccess to simplify access to a PCI card/chip. But this feature is optional. As you can see in the package build configuration, the libpciaccess support is explicitly disabled  -https://github.com/openwrt/packages/blob/7bc897eb94c09c8fe2472091709fab7446b37976/utils/atheepmgr/Makefile#L63 (see reasons below). That is why atheepmgr package passes build tests on the initial submission. So package does not miss any dependencies since it simply have no dependencies :)

Even without *libpciaccess* you have a couple of options to access hardware, e.g. you could use *memory* "connector" or driver (ath9k/ath10k) debugfs interface. But. But OpenWrt by default disables */dev/mem* in the kernel, so direct I/O mem access (and *libpciaccess* too) is useless in a common case. That is why I implement debugfs access method in the utility. To be able to use *libpciaccess* you need to build a custom OpenWrt firmware with enabled */dev/mem* (see KERNEL_DEVMEM option). That is why I choose to not add the *libpciaccess* support on the initial package submission. It is useless for most users :(

@davidrapan, if you would like to build and use the utility with libpciaccess and building a custom OpenWrt firmware do not scare you, then I suggest to add a package configuration option. This option will enable libpciaccess dependency and activate the library support in the utility.

davidrapan:
@rsa9000 Ahh I see, thanks for this exhausting clarification. I mean, fact that it end up as pull req into master was the matter more of a too quick fingers than intention. But anyway valuable, your point of view gave me some new ideas, or maybe I'll do as u suggested that opt-in, thx. :)

neheb:
Alternative patch was merged..

PR #15066, opened 2021-03-08
nextdns: Update to version 1.11.0

rs:
Update to version 1.11.0


neheb:
Note that there is also 21.02.

rs:
Thx for the heads up.

PR #15069, opened 2021-03-09
selinux updates

neheb:
Maintainer: @tpetazzoni @flyn-org 
Compile tested: ath79

PR #15072, opened 2021-03-09
selinux: update to upstream

doverride:
Maintainer: Thomas Petazzoni <thomas.petazzoni@bootlin.com>
Compile tested: arm_cortex-a9_vfpv3-d16, mvebu/cortexa9, r16152-bff84f3e8e
Run tested: arm_cortex-a9_vfpv3-d16, mvebu/cortexa9, r16152-bff84f3e8e ran various applicable commands on a selinux enforcing system with linux 5.10 and selinux 3.2

I would argue that most of these should probably be dropped because most of it is not actively used and it presents a significant maintenance burden.


doverride:
CI keeps complaining about dirty patches telling me to refresh but patches apply fine locally. Ran out of patience.

see: https://github.com/openwrt/packages/pull/15069

neheb:
That's done with make package/x/refresh

neheb:
Hmm I am very curious why there are multiple selinux packages coming from the same tarball.

PR #15075, opened 2021-03-09
stubby: Update README.md

lgladdy:
Update to correct dnssec configuration value

Description:
Hey team, just a little update to the readme for configuring stubby with DNSSEC. I tried with `1` as per the original readme, but it appears to actually need to be `GETDNS_EXTENSION_TRUE` as per the stubby documentation: https://dnsprivacy.org/wiki/display/DP/Configuring+Stubby#ConfiguringStubby-DNSSEC


aparcar:
Please see the CI output on what you have to change.

PR #15076, opened 2021-03-10
safe-search: check for changed IP addresses weekly

farmergreg:
Maintainer: me
Compile tested: 19.07.6
Run tested: x86, archer-c7

Description:
The current default of hourly is too fast. Some services such as
DuckDuckGo return IPs from a pool based on the user's location instead
of a fixed IP address. This change prevents unnecessary writes to the
flash memory by only updating once per week.


farmergreg:
Thanks @neheb sorry for that oversight. I've pushed a new version of the patch.

PR #15078, opened 2021-03-10
git: update to version 2.30.2 (security fix)

ja-pa:
Maintainer: @tripolar 
Compile testd: Turris Omnia(TOS6), OpenWrt master
Run tesed: Turris Omnia (TOS6), OpenWrt master

Description:
This PR updates git to version 2.30.2 [Changelog](https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.30.2.txt) It fixes [CVE-2021-21300](https://nvd.nist.gov/vuln/detail/CVE-2021-21300)


ja-pa:
I think that severity for OpenWrt is rather low but it should be probably cherry-picked to 19.07 and 21.02.

tripolar:
thanks, can you provide patches for  19.07 and 21.02 too?

But the bugs doesn't affect linux, it's a windows/macos only bug

>  A specially crafted repository that contains symbolic links as well as files using a clean/smudge filter such as Git LFS, may cause just-checked out script to be executed while cloning onto a case-insensitive file system such as NTFS, HFS+ or APFS (i.e. the default file systems on Windows and macOS).

See: https://github.com/git/git/security/advisories/GHSA-8prw-h3cq-mghm

tripolar:
the bugs doesn't affect linux, it's a windows/macos only bug

>  A specially crafted repository that contains symbolic links as well as files using a clean/smudge filter such as Git LFS, may cause just-checked out script to be executed while cloning onto a case-insensitive file system such as NTFS, HFS+ or APFS (i.e. the default file systems on Windows and macOS).

See: https://github.com/git/git/security/advisories/GHSA-8prw-h3cq-mghm

tripolar:
thanks, can you provide patches for 19.07 and 21.02 too?

PR #15107, opened 2021-03-12
ksmbd: update to 3.3.6

neheb:
Maintainer: @Andy2244
Compile tested: ath79


Andy2244:
Fine as always

Btw maybe you want to take over ksmbd as maintainer, since you update it way faster than me. I still use samba4 primarily so my ksmbd updates only get a quick test anyway? So i don't mind if you feel like it.

neheb:
bogus CI errors yet again.

PR #15108, opened 2021-03-12
ksmbd updates for 19.07

neheb:
Maintainer: @Andy2244 

keeping this here until a new release is to be cut. A really nasty file bug crept in the 3.3.4 release: https://github.com/cifsd-team/cifsd/issues/498

ping @hauke

### DIFF
--- a/kernel/ksmbd/Makefile
+++ b/kernel/ksmbd/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ksmbd
-PKG_VERSION:=3.3.4
+PKG_VERSION:=3.3.5
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/cifsd-team/cifsd/tar.gz/$(PKG_VERSION)?
-PKG_HASH:=4f8b7610ba084f6813cbb85bb6c07af50ba542e928c370e79022039fa027bc9a
+PKG_HASH:=d34c73d1fe1d6d2136472e37cf54bb819a59fc7e79b4f95772ce9da1eff7fff9
 
 PKG_MAINTAINER:=Andy Walsh <andy.walsh44+github@gmail.com>
 PKG_LICENSE:=GPL-2.0-or-later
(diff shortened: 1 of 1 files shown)


erdoukki:
Needed also for master !

thanks for the PR

neheb:
https://github.com/openwrt/packages/commit/851b9b252fc5f5fdd541fd3a18ab393bdac9a6e8 https://github.com/openwrt/packages/commit/7a1a8f3659cf79237fba6394fbea42755af38a52

erdoukki:
Thanks !

neheb:
@Andy2244 what do you think about using ksmbd.control to handle the termination instead of the init script?

neheb:
@Andy2244 bad news. latest version is 5.4 and above only. Can you add the fix from https://github.com/cifsd-team/cifsd/issues/507 in the init script for master so it can be backported here?

Andy2244:
> 
> 
> @Andy2244 bad news. latest version is 5.4 and above only. Can you add the fix from [cifsd-team/cifsd#507](https://github.com/cifsd-team/cifsd/issues/507) in the init script for master so it can be backported here?

Sure if i'm back from Easter-holidays.

Neustradamus:
@neheb: Thanks for the update!

neheb:
Forgot this can't be updated past 3.3.7. kernel minimum is 5.4 now.

BKPepe:
Based on a recent conversation on IRC channel ``#openwrt-adm``, there should be released OpenWrt 19.07.8 somehow soon. They were talking about next week, so we should merge it a few days early so the situation with [the ksmbd and it's dependency](https://openwrt.org/releases/19.07/notes-19.07.7#regressions) will not happen anymore.

Neustradamus:
@neheb: Can you look for a 3.4.0 update?

neheb:
Not for 19.07. newer versions of ksmbd require kernel 5.4.

Neustradamus:
Yes for 21.02...

hauke:
@neheb @BKPepe  I would like to do an 19.07.8 today, if this should go in could you merge it and also test it please. I can also shift the tagging to tomorrow.

hauke:
Update merged for the next release.

Neustradamus:
Thanks to @neheb and @hauke.

PR #15112, opened 2021-03-13
freetype: meson.build: Fix .pc file generation

bkuhls:
Fixes error when building fontconfig after switch to meson (https://github.com/openwrt/packages/pull/15018):
```
checking for FREETYPE... no
configure: error: Package requirements (freetype2 >= 21.0.15) were not met:

Package dependency requirement 'freetype2 >= 21.0.15' could not be satisfied.
Package 'freetype2' has version '6.17.4', required version is '>= 21.0.15'
```
Signed-off-by: Bernd Kuhls <bernd.kuhls@t-online.de>

Maintainer: @val-kulkov 
Compile tested: x86_64


bkuhls:
Forced-pushed with fixed patch after running make target "package/feeds/packages/freetype/refresh"

neheb:
Interesting... I wonder why I never hit this.

PR #15133, opened 2021-03-15
tesseract updates

neheb:
Maintainer: @vk496
Compile tested: mips64el


neheb:
```
make[1]: *** No rule to make target 'package/tesseract-data-mri/download'.  Stop.
``

@aparcar more issues.

neheb:
Locally built with x86. Works now.

PR #15136, opened 2021-03-15
mwan3: remove mwan3 ubus call on mwan3 iface hotplug ACTION

feckert:
Maintainer: me / @aaronjg 
Run tested: x86_64, APU3, OpenWrt openwrt-19.07, tests done)

Description:
With this change, the interface status is no longer read from the mwan3 ubus.
The status of the interface is read directly from the status directory.

I had a strange behavior on my system. It turned out that for reasons unknown to me, the rpcd could no longer find the mwan3 ubus object. But since the status is checked via the mwan3 ubus object, the router was in a loop and always restarted the interface. To prevent this from happening again, remove tha call do ubus and so read the status information from status directory.
This was already implemented in the master with other change in commit c07f5230be128669f7b6731415de26f8176fbf5b.

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.8.15
+PKG_VERSION:=2.8.16
 PKG_RELEASE:=1
 PKG_MAINTAINER:=Florian Eckert <fe@dev.tdt.de>
 PKG_LICENSE:=GPL-2.0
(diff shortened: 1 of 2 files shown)


feckert:
> It just cruft from before procd

@aaronjg It's been a long time since I did that.
But when I [look at](https://github.com/openwrt/packages/commit/815e83d46178f7f57044d0a6a8efb292d9106c93) it, I would say that `running` is not necessary.


If the `inital_check` is offline we start the tracker in offline state on boot and after that we read the STATUS value.
If hte `inital_check` is online we always start the tracker in online state.

From my point of view, that should fit.

aaronjg:
Thanks for explaining. Looks good to me.

PR #15061, opened 2021-03-08
igmpproxy: move package from openwrt repo

ja-pa:
Maintainer: Felix Fietkau <nbd@nbd.name>
Compile tested: Turris Omnia (TOS6), OpenWrt master
Run tested: Turris Omnia (TOS6), OpenWrt master

Description:
This PR moves igmpproxy package from openwrt repo to packages repo.
The proposal for this change was raised in https://github.com/openwrt/openwrt/pull/3819

Related MR for openwrt repo https://github.com/openwrt/openwrt/pull/3976

cc  @bobafetthotmail @rsalvaterra

PR #15137, opened 2021-03-15
igmpproxy: drop package from feed; it's in core now

commodo:
Maintainer: @nbd168 
Compile tested: n/a
Run tested: n/a

-------------------------------------------------------

In core it's present here:
  https://github.com/openwrt/openwrt/tree/master/package/network/services/igmpproxy

Having it in the feed in master causes a warning
```
WARNING: Not overriding core package 'igmpproxy'; use -f to force
```

While this isn't a major issue, we should clean this and not have the
package in 2 places.


champtar:
igmpproxy was just moved here by https://github.com/openwrt/packages/pull/15061

commodo:
Oh
I had it backwards
Thanks for the heads up

PR #15145, opened 2021-03-17
mwan3: reload interface parameters on ifupdate event

aaronjg:
Maintainer: me, @feckert 
Compile tested: none needed
Run tested: @wackejohn , openwrt 20.x

mwan3track was failing if an interface changed IP address on DHCP renew. I was not able to recreate it, but a user was able to test this fix and confirmed that it resolved the issue.

### DIFF
--- a/net/mwan3/files/etc/hotplug.d/iface/15-mwan3
+++ b/net/mwan3/files/etc/hotplug.d/iface/15-mwan3
@@ -10,11 +10,11 @@ initscript=/etc/init.d/mwan3
 
 
 SCRIPTNAME="mwan3-hotplug"
-[ "$ACTION" = "ifup" ] || [ "$ACTION" = "ifdown" ] || [ "$ACTION" = "connected" ] || [ "$ACTION" = "disconnected" ] || exit 1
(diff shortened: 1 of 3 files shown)


net/mwan3/files/etc/hotplug.d/iface/15-mwan3
```diff
@@ -66,11 +66,14 @@ case "$ACTION" in
...
 		if [ "$MWAN3_STARTUP" != "init" ]; then
 			mwan3_create_iface_route $INTERFACE $DEVICE
-	                mwan3_set_general_rules
+			mwan3_set_general_rules
 			[ "$status" = "online" ] && mwan3_set_policies_iptables
 		fi
 		[ "$ACTION" = ifup ] && procd_running mwan3 "track_$INTERFACE" && procd_send_signal mwan3 "track_$INTERFACE" USR2
 		;;
+	ifupdate)
+		[ "$ACTION" = ifup ] && procd_running mwan3 "track_$INTERFACE" && procd_send_signal mwan3 "track_$INTERFACE" USR2
```
feckert, line 75: Does this really work? 
The ACTION is ifupdate but we check `[ "$ACTION" = ifup ]`

aaronjg: Yes, looks like it would not really work. This is a real issue, but I haven't been able to recreate it or test. I updated the P/R, but perhaps we should close this for now until another user runs into it and can test the fix.

PR #15156, opened 2021-03-18
tmate: add new package

1715173329:
Maintainer: me
Compile tested: rockchip, x86_64
Run tested: rockchip nanopi-r2s

Description:
Tmate is a fork of tmux. It provides an instant pairing solution.
For more details, see https://tmate.io.

### DIFF
--- a/libs/msgpack-c/Makefile
+++ b/libs/msgpack-c/Makefile
@@ -1,0 +1,48 @@
+# SPDX-License-Identifier: GPL-3.0-only
+#
+# Copyright (C) 2021 ImmortalWrt.org
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=msgpack-c
+PKG_VERSION:=3.3.0
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/msgpack/msgpack-c/tar.gz/cpp-$(PKG_VERSION)?
+PKG_HASH:=754c3ace499a63e45b77ef4bcab4ee602c2c414f58403bce826b76ffc2f77d0b
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-cpp-$(PKG_VERSION)
+
+PKG_LICENSE:=BSL-1.0
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=Tianling Shen <cnsztl@immortalwrt.org>
+
+PKG_BUILD_PARALLEL:=1
+CMAKE_INSTALL:=1
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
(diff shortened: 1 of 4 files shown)


1715173329:
Ping @aparcar.

aparcar:
The CI fails are due to outdated Docker containers and not due to this package. I'm on it to provide a fix.

Thank you @1715173329 for the quick migration of the package.

1715173329:
> Thank you @1715173329 for the quick migration of the package.

It's my pleasure.😊

libs/msgpack-c/Makefile
```diff
@@ -0,0 +1,51 @@
+#
+# Copyright (C) 2021 ImmortalWrt.org
+#
+# This is free software, licensed under the GNU General Public License v3.
```
aparcar, line 4:
```suggestion
# SPDX-License-Identifier: GPL-2.0
```
Instead of the old style please use a SPDX tag and put it as very first line.

net/tmate/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+tmate -V | grep "${2#*v}"
```
aparcar, line 3:
```suggestion
tmate -V | grep "$PKG_VERSION"
```
Instead of grepping via this expression you could also use an exported variable.

aparcar (changes requested):

libs/msgpack-c/Makefile
1715173329: Updated.

aparcar:
Unsure why the CI fails, but tested and works great. Thanks again I'll merge this.

1715173329:
> Unsure why the CI fails

The CI will uninstall the package once it has been tested,
so when installing `tmate`, opkg couldn't find `msgpack-c` and threw the error.

BTW, I found a interesting bug that opkg shows `incompatible architecture found` instead of `missing the following dependencies`.

aparcar:
That's indeed curious, one more reason to switch to APK :)

libs/msgpack-c/Makefile
```diff
@@ -0,0 +1,48 @@
...
+PKG_VERSION:=3.3.0
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/msgpack/msgpack-c/tar.gz/cpp-$(PKG_VERSION)?
+PKG_HASH:=754c3ace499a63e45b77ef4bcab4ee602c2c414f58403bce826b76ffc2f77d0b
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-cpp-$(PKG_VERSION)
+
+PKG_LICENSE:=BSL-1.0
+PKG_LICENSE_FILES:=LICENSE
```
BKPepe, line 17: Well, ``LICENSE`` file does not exist, but ``LICENSE_1_0.txt`` does.

aparcar: Good catch, I'll update this.

PR #15161, opened 2021-03-19
isc-dhcp: add dynamic DNS as meta package

pprindeville:
Maintainer: me
Compile tested: x86_64, generic, HEAD (ebcb4f1d0a)
Run tested: same, installed on VM, `opkg info` show correct dependencies

Description:

Have moved dependencies of `bind-client` and `bind-server` into meta-package instead.

### DIFF
--- a/net/isc-dhcp/Makefile
+++ b/net/isc-dhcp/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=isc-dhcp
 UPSTREAM_NAME:=dhcp
 PKG_VERSION:=4.4.1
-PKG_RELEASE:=16
+PKG_RELEASE:=17
 
 PKG_LICENSE:=BSD-3-Clause
 PKG_LICENSE_FILES:=LICENSE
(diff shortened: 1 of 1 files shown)


net/isc-dhcp/Makefile
```diff
@@ -224,6 +251,10 @@ define Package/isc-dhcp-server-ipv6/conffiles
 /etc/dhcpd6.conf
 endef
 
+define Package/isc-dhcp-dyndns-$(BUILD_VARIANT)/install
+	:
```
aparcar, line 255: What's this?

pprindeville: > What's this?

Yeah, I was pulling out my hair trying to get a meta-package to generate anything... turns out it won't unless this macro is defined. Here's why:

https://github.com/openwrt/openwrt/blob/master/include/package-bin.mk#L8

Look at `libnl` for example:

https://github.com/openwrt/openwrt/blob/master/package/libs/libnl/Makefile#L55
https://github.com/openwrt/openwrt/blob/master/package/libs/libnl/Makefile#L120

aparcar: I guess i was confused about the `:`

pprindeville: > I guess i was confused about the `:`

Yeah, it ends up being a recipe with a single step of `:` in shell which does nothing...

PR #15171, opened 2021-03-20
miniupnpd: add patience to firewall include

ldir-EDB0:
Occasionally, mostly at startup, miniupnpd reports "Another app is
currently holding the xtables lock. Perhaps you want to use the -w
option?"

Take iptables' advice and wait up to 1 second before giving up.

Signed-off-by: Kevin Darbyshire-Bryant <ldir@darbyshire-bryant.me.uk>


pprindeville (approved):
LGTM

PR #15194, opened 2021-03-21
mandoc: fix compilation with GCC10

neheb:
This is probably fixed in CVS. Unfortunately, I have no idea how to
import such a patch.

Maintainer: @flyn-org 
Compile tested: mips64el


flyn-org:
I am familiar with the change in GCC 10 that is related to this patch, and this does indeed appear required. At first I was confused that I did not notice this myself, but I think other OpenWrt configuration parameters will determine if GCC actually objects to the code this patch fixes due to #ifdefs in the code. Is this your experience? Does this only manifest if the OpenWrt configuration leads to activating getline features?

Recommend merge.

PR #15195, opened 2021-03-21
CI: test build dependend packages

aparcar:
By using OPKGs `whatdepends` all packages dependend on a library are
printed. Use that feature to optain packages which a version change may
break and build them as well.

While add it, split a long line.

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -77,10 +77,20 @@ jobs:
 
           # fallback to test packages if nothing explicitly changes this is
           # should run if other mechanics in packages.git changed
-          PACKAGES="${PACKAGES:-vim tmux bmon}"
+          PACKAGES="${PACKAGES:-vim attendedsysupgrade-common bmon}"
 
           echo "Building $PACKAGES"
           echo "PACKAGES=$PACKAGES" >> $GITHUB_ENV
+
+      - name: Determine depended packages
+        run: |
+          DEPENDS=$(docker run --rm \
+            "openwrt/imagebuilder:${{ matrix.target }}-$BRANCH" \
+            make whatdepends PACKAGE="$PACKAGES" | grep $'\t' | \
+            grep -v luci-i18n | awk '{ print $1 }' | tr '\n' ' ')
+
+          echo "Building $DEPENDS"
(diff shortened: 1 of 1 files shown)


aparcar:
@BKPepe please review this, do you think it makes sense or get out of hand because so much will be recompiled?

neheb:
Seems fine.

BKPepe:
Hey @aparcar,

Thank you for such changes.

If I could add something, I'd split your changes in several commits to make clear Git history.

Those arch/target changes can be in a separate commit. 

The same applies for replacing `tmux` with `attended-sysupgrade-common` is not mentioned anywhere. It would be good to have in the commit message why you are replacing it.

We definitely need something like this regarding determine packages, but I am not sure how long it will take as these tests should be quick and you don't wait to wait for a long time to see if the check is green or not. This is a lot of simplified. 
In any case, I mean why not and we will see.

aparcar:
In case anyone ever changes e.g. `libncurses` the CI will go nuts, since so much depends on it. However, I'd say we try it.

@BKPepe I think I took care of the mentioned issues, please see again.

neheb:
Makes the most sense to get https://github.com/openwrt/packages/pull/15012 sorted out first I think. I would like to get it built without the CI timing out.

aparcar:
> Makes the most sense to get #15012 sorted out first I think. I would like to get it built without the CI timing out.

I don't think they are related, but sure let's get the Ninja thing going.

neheb:
The current problem with that PR is that the CI takes too long. I'm trying to handle these changes in separate PRs to reduce the workload.

BKPepe (approved):

pprindeville:
@aparcar: question about this...  Looking at the CircleCI build logs, I noticed that a commit to `strongswan` would involve rebuilding `pigeonhole`.  WTAH.

There is not dependency graph I can find that `strongswan` depends on `pigeonhole` or depends on anything in turn directly or indirectly dependent on `pigeonhole`.

This seems like an enormous waste of time and cycles.  What am I missing?

aparcar:
What logs? I disabled CircleCI and what you see here is GitHub Actions. Whatever packages are changed cause a build test, if there is no relation to `pigeonhole` and no changes to it's contents, there shouldn't be a build test, else it's a bug of the CI.

PR #15222, opened 2021-03-22
nss: quote $(PIC) usage and fix compilation with QUILT

cotequeiroz:
Maintainer: @lucize 
Compile tested: x86_64, openwrt-21.2
Run tested: none

Description:
openwrt/openwrt#4006 intends to add `-DPIC` to the global definition of `FPIC`.  Then it becomes mandatory to quote it when assigning it to a variable or a parameter in a shell context.

While looking at the previous issue, I found out that the package was not compiling with `QUILT=1` because of 657574f45 (#14644).  To allow building with quilt, and still keep #14644's goal, I've moved it from `Build/Prepare` to `Build/Configure`.  Since the nss package does not have a `configure` step, I've skipped calling `Build/Configure/Default`.

### DIFF
--- a/libs/nss/Makefile
+++ b/libs/nss/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nss
 PKG_VERSION:=3.61
-PKG_RELEASE:=1
+PKG_RELEASE:=2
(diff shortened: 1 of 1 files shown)


lucize:
LGTM

PR #15190, opened 2021-03-21
umurmur: fix compilation without deprecated OpenSSL APIs

neheb:
Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @fatbob313 
Compile tested: mips64el

PR #15228, opened 2021-03-22
uMurmur: Update to 0.2.20. Fix build against OpenSSL without deprecat…

fatbob313:
…ed API enabled.

Signed-off-by: Martin Johansson <martin@fatbob.nu>

Maintainer: me 
Compile tested: (Ubuntu WSL, OpenWrt 21.02)
Run tested: (arm_cortex-a15_neon-vfpv4 (Netgear R7800),  OpenWrt 21.02, working as expected)

Description:

### DIFF
--- a/net/umurmur/Makefile
+++ b/net/umurmur/Makefile
@@ -5,12 +5,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=umurmur
-PKG_VERSION:=0.2.19
+PKG_VERSION:=0.2.20
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


fatbob313:
I believe this can replace #15190

neheb:
I must have done this wrong...

Can you replace include $(INCLUDE_DIR)/cmake.mk with include ../../devel/ninja/ninja-cmake.mk as in https://github.com/openwrt/packages/pull/15012 ?

The goal is to fix many of the packages separately so that GitHub Actions does not timeout when building all the packages in that PR.

fatbob313:
Now it's added but the commit message doesn't reflect it though...

PR #15247, opened 2021-03-24
boost: fixup mips64 fiber patch

neheb:
It turns out that this is also broken on mips64el. Further testing
reveals that

-mips32r2 -mtune=mips32r2 -mabi=32

compiles the PAUSE instruction just fine whereas

-mips64r2 -mtune=mips64r2 -mabi=64

does not. The PAUSE instruction was introduced in version 2.6 of the
MIPS ISA and GCC for some reason does not allow usage of it with MIPS64.
Modify the macro to fix the situation instead of just matching on
octeon, which is not quite correct.

Maintainer: @ClaymorePT 
Compile tested: mips64el

### DIFF
--- a/libs/boost/Makefile
+++ b/libs/boost/Makefile
@@ -13,7 +13,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=boost
 PKG_VERSION:=1.75.0
 PKG_SOURCE_VERSION:=1_75_0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)_$(PKG_SOURCE_VERSION).tar.bz2
 PKG_SOURCE_URL:=@SF/$(PKG_NAME)/$(PKG_NAME)/$(PKG_VERSION) https://dl.bintray.com/boostorg/release/$(PKG_VERSION)/source/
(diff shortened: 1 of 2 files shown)


neheb:
```
make[1]: *** No rule to make target 'package/kea-libs/download'.  Stop.
```

Unrelated to this PR. ping @aparcar

aparcar:
What is kea-libs?

aparcar:
Ugh okay, so problem is that the packages is compiled via a folder while the packages have different names themselfs. Maybe I need to rewrite the whole package dependency thing, I hope there is a mapping between package folder (used by make) and package names (as shown by opkg).

ClaymorePT:
@neheb Thanks for looking at this. Out of curiosity, what version of the GCC compiler is not allowing the usage of the right instruction? Could this be a matter of using a more recent compiler ?

neheb:
I don't think so. This is an issue with version 8 and 10.

ClaymorePT:
So, what other action could be taken here, other than just disabling Boost.Fiber for this target ?

neheb:
What do you mean? This patch works fine. The CI is failing for an unrelated reason. It will pass after a rebase.

ClaymorePT:
Oh. Okay. I though the issues were related.

Thanks for the update.

neheb:
Passes after a rebase.

PR #15250, opened 2021-03-25
travelmate: added login script for mikrotik captive portal

ChristianKuehnel:
Maintainer: @dibdot
Compile tested: xRX200 rev 1.2, AVM FRITZ!Box 7362 SL, OpenWrt 19.07.7 r11306-c4a6851c72
Run tested: same as compile tested

Description:
Added handling of Mikrotik captive portals

### DIFF
--- a/net/travelmate/files/mikrotik.login
+++ b/net/travelmate/files/mikrotik.login
@@ -4,11 +4,40 @@
 #    username - user name for login
 #    password - password for login
 
-set -eu
+# check arguments
+if [ "$#" -ne 2 ]; then
+    echo "Error: Invalid number of arguments."
+    echo "usage: mikrotik.login <username> <password>"
+    exit 1
+fi
+
+cmd="$(command -v curl)"
+# curl check
+if [ ! -x "${cmd}" ]
+then
+	echo "Error: this script depends on curl. Please install this first."
+    exit 2
+fi
+
 username=$1
 password=$2
+login_success="You are logged in"
+
+# Get login domain from redirection information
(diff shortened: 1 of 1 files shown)


dibdot:
@ChristianKuehnel Thanks for that! Just a few remarks:

- please remove this "set -eu" stuff, this subscript may error out in the middle of nowhere ... - you should return/error out if user/password parameter are not filled
- travelmate 1.5.x has no dependency to curl, therefore you should check for that, e.g.
```
cmd="$(command -v curl)"

# curl check
#
if [ ! -x "${cmd}" ]
then
	exit 1
fi
```

ChristianKuehnel:
`set -eu` does exactly what you want: It fails if the arguments are missing or if curl is not installed. There is no need for additional checks there. Undefined variables and failing commands are errors, so the script should exit. This make debugging much easier.

Also this PR is for the master branch, I guess there we do have a curl dependency there, right?

dibdot:
No, that doesn't help cause it always breaks with a unspecific "parameter not set" error. I want to receive individual return/error codes which will be evaluated/logged by the calling travelmate.

ChristianKuehnel:
You're the maintainer, so I updated the script to match your requests. Please take a look.

dibdot:
Sorry, still something todo:
- please remove the "echo" statements, only the return codes are visible in the log
- a final return is currently missing at the end of your file
- you need to squash your commits

Thanks!

dibdot:
now superseded by/included in #15254

PR #15253, opened 2021-03-25
Fixed mosquitto cannot be started when system boot

srdgame:
I have included this package in my own SUNXI (H3) board build, but it failed to be started when system booted.  Either change the start seq to 81 or set the correct respawn flag cloud fix this.

So I have changed the start sequence and added respawn flag, to make sure the mosquitto service could be started when system booting

Compile tested: SUNXI A7 Openwrt SNAPSHOT
Run tested:  SUNXI A7 Openwrt SNAPSHOT

Description:

### DIFF
--- a/net/mosquitto/files/etc/init.d/mosquitto
+++ b/net/mosquitto/files/etc/init.d/mosquitto
@@ -3,7 +3,7 @@
 # April 2012, OpenWrt.org
 # Provides support for the luci-app-mosquitto package, if installed
 
-START=80
+START=81
 USE_PROCD=1
 TCONF=/tmp/mosquitto.generated.conf
 CONF_WATCH=/etc/config/mosquitto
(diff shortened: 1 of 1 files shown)


srdgame:
will correct commit comment

karlp:
We also cna't really merge this as it is anyway, why doesn't it start? what's it waiting for? do you want to add network interface triggers?

srdgame:
I think it is because the relayd service reloading the network, That is why I have increased the boot sequence to 81.  I will try to get the logs to know what exactly happen.

@karlp

srdgame:
@karlp I have fired an new pull request.  The root reason is because network interfaces are not ready when mosquitto boot (only happen in FIRST boot of new flashed device)

srdgame:
#15272

PR #15256, opened 2021-03-25
[19.07] libftdi1: Improve build binary reproducibility

thg2k:
The library embeds the result of "git describe" inside the source code, making the binary result dependent of the particular commit being used in the build root when building inside a git working copy.

As this is unnecessary information, remove this option and fallback to the default "unknown", which is also the value compiled by tools that do not clone but export the openwrt base tree.

### DIFF
--- a/libs/libftdi1/Makefile
+++ b/libs/libftdi1/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libftdi1
 PKG_VERSION:=1.4
-PKG_RELEASE:=6
+PKG_RELEASE:=7
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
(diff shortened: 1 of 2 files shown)


neheb:
Required a PKG_RELEASE bump. Please make this PR against master.

thg2k:
Good point for the PKG_RELEASE bump.

So should I open a PR against master and then open 2 more to backport it to 21 and 19 or someone will cherry-pick it for me?

neheb:
I can cherry pick. Please remind me in your PR.

thg2k:
Sorry but that won't work, as master and 21 run with package version 1.5-2 while 19 has 1.4-6, so I just bumped the PKG_RELEASE in the 19 to 1.4-7.

Could we just merge this in 19? I would need to check whether for 1.5 this change is still relevant and generate a refreshed patch for it, if it is. I currently haven't time for doing this check and I wouldn't submit an untested patch blindly.

Thx

PR #15259, opened 2021-03-25
uacme: allow including run-acme

lemoer:
Maintainer: @lucize
Compile tested: x86-64, openwrt master (bc2538a460b7409f690d1bab6d188cb3fca747ad)
Run tested: x86-64, openwrt master (bc2538a460b7409f690d1bab6d188cb3fca747ad)

Description: The following changes allows including the run-acme script into other scripts. This does not change any behavior, when the script is invoked normally.


lucize:
LGTM

Issue #15236, opened 2021-03-23
prometheus: 2.22.0 failing on RPI4


sashasimkin:
Maintainer: @aparcar 
Environment: `linux/arm64, RaspberryPI4 model B, snapshot`

Description:

Prometheus version 2.22.0 is failing on RPI4 with a segfault. The same version built by `go1.15.3` for arm64 arch has worked on the same RPI.

See more details in https://github.com/prometheus/prometheus/issues/8589

I see that golang version is set to 1.16 in here: https://github.com/openwrt/packages/blob/master/lang/golang/golang/Makefile, but I suppose that many other packages depend on that version, and setting it to 1.15.3 most likely not something that will be accepted :(

Does it make sense trying to specify golang version to 1.16.2 for the whole build as suggested in https://github.com/prometheus/prometheus/issues/8589#issuecomment-797330164?

aparcar:
I'll just update to the newest release which seem to fix the issue with later Golang versions:
https://github.com/openwrt/packages/pull/15237

aparcar:
I updated Prometheus, could you please test 2.25.2?

sashasimkin:
The binary is definitely working, thank you! But I'm getting some weird behavior :(

When I run it myself, using the command line and all parameters supplied by hand - it works well and I can see metrics in grafana, yet if I do start it as `/etc/init.d/prometheus start` - I see only flatline and no data after a while, like this:
![image](https://user-images.githubusercontent.com/1117019/112556262-83ded700-8dd2-11eb-8d62-e594c3bd5894.png)

while in logs it says:
```
Fri Mar 26 01:21:33 2021 daemon.info procd: Instance prometheus::instance1 s in a crash loop 6 crashes, 0 seconds since last crash
```

Later I figured out that it was most likely due to the permissions and that prometheus was not able to access `/data` directory under user `prometheus` because I've created it with prometheus running as root.

This is how I figured out it's starting with user `prometheus`:
```
root@home /53# PROCD_DEBUG=1 /etc/init.d/prometheus start
[...]

PR #15265, opened 2021-03-26
prometheus: Update default tsdb path

sashasimkin:
Use /var/prometheus-data instead of /data, because user `prometheus` doesn't have permissions to create directory in root filesystem, while it's totally fine to be in /var

Maintainer: @aparcar
Compile tested: -
Run tested: linux/arm64, RaspberryPI4 model B, snapshot - by editing /etc/config/prometheus manually

Description: Use /var/prometheus-data instead of /data, because user prometheus doesn't have permissions to create directory in root filesystem, while it's totally fine to be in /var

related: #15236

### DIFF
--- a/utils/prometheus/files/etc/init.d/prometheus
+++ b/utils/prometheus/files/etc/init.d/prometheus
@@ -12,8 +12,14 @@ start_service() {
 	local web_listen_address
 	config_load "prometheus"
 	config_get config_file prometheus config_file "$CONFFILE"
-	config_get storage_tsdb_path prometheus storage_tsdb_path "/data"
+	config_get storage_tsdb_path prometheus storage_tsdb_path "/srv/prometheus"
 	config_get web_listen_address prometheus web_listen_address "127.0.0.1:9090"
+	
+	# Create tsdb dir & permissions if needed
+	if [ ! -d "$storage_tsdb_path" ]; then
+		mkdir "$storage_tsdb_path"
+		chown prometheus:prometheus "$storage_tsdb_path"
+	fi;
 
 	procd_open_instance
 	procd_set_param command "$PROG"
--- a/utils/prometheus/Makefile
+++ b/utils/prometheus/Makefile
@@ -47,6 +47,9 @@ Prometheus, a Cloud Native Computing Foundation project, is a systems and
 service monitoring system. It collects metrics from configured targets at given
 intervals, evaluates rule expressions, displays the results, and can trigger
 alerts if some condition is observed to be true.
+
+Default tsdb path is /srv/prometheus, you might want to edit /etc/config/prometheus
+in order to place it on USB storage or external SD card.
 endef
 
 define Package/prometheus/install
--- a/utils/prometheus/files/etc/uci-defaults/prometheus-defaults
+++ b/utils/prometheus/files/etc/uci-defaults/prometheus-defaults
@@ -6,7 +6,7 @@ uci -q get prometheus.prometheus || {
 	uci -q batch <<EOF
 	set prometheus.prometheus=prometheus
 	set prometheus.prometheus.config_file='/etc/prometheus.yml'
-	set prometheus.prometheus.storage_tsdb_path='/data'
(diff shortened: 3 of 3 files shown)


aparcar:
So /var is mounted in RAM, after every reboot your data would be lost. Having it in /data forces the user to think at least for a second on where to store things. Ideally on an external drive or USB stick, not so ideally in memory or some RPi flash chip.

What do you think?

sashasimkin:
Oh, I didn't know that `/var` is mounted in RAM!

The problem with `/data`, is that it doesn't give any feedback on why prometheus is crushing :(

Is it maybe possible to add a remark about that on the OpenWRT packages page, or name the dir something like `/pick-proper-location-for-data` :D

_Unless_, `/data` is available on other devices, or some common mount-point for more durable than sd storage?

sashasimkin:
Is it possible to document this quirk somehow, or am I missing some fundamental piece about OpenWRT?

erdoukki:
Is it not a better place like /srv/data ?

aparcar:
So /var is mounted in /tmp and wiped every time you reboot. /srv sounds better to me, however I don't know how it's mounted at this point.

erdoukki:
/srv is default in rootfs, and may be switched by user in extroot or in any mountpoint...

sashasimkin:
thanks @erdoukki , that makes sense, and I think I've done `/srv/data`(or other, more specific mountpoint) for myself in the end.

I've updated the path in PR to be in `/srv/data/prometheus`, will test on my installation soon.

aparcar:
Please update me and I'll merge this.

sashasimkin:
Just had time to boot up my installation and check:

OpenWRT(+wulfy's) image: `OpenWrt SNAPSHOT r16357-f8669c174e`

`/srv/data` doesn't exist, and when pointing prometheus tsdb data to `/srv/data/prometheus` - the service will appear running, but no data can be seen in grafana and not even `/srv/data` is created.

So I attempted to configure it with `/srv/prometheus-tsdb` path, and no luck either :(

----

I just realized that first message doesn't mention why it's fine for data to be in `/var`.  
It boils down to sticky bit set on `/var/` directory, so when prometheus creates it - it has his user's ownership.

There's no such thing is `/srv`.

----
So I'm thinking that instead of trying to figure out best place for user's data - maybe it's better to just document that by default installation is broken and users need to thing and configure the package as they please?

erdoukki:
Why not doing these stuff in the install script ?

sashasimkin:
The only con I see is that it'll be burning PRI's flash chip by default while being in /srv, unless users have exroot or external USB mounted over that path.

Sounds good to me though, let me update the PR.

sashasimkin:
I've pushed latest changes as per our discussion, but decided to create the directory if it doesn't exist on each prometheus service start.

This way it'll be also properly handled for cases when you move data to a USB partition.

(In fact, this is what prometheus tries to do currently, but fails because not having enough permissions. Here we create the dir as root and grant permissions to `prometheus` user.)

utils/prometheus/files/etc/init.d/prometheus
```diff
@@ -12,8 +12,14 @@ start_service() {
...
 	config_load "prometheus"
 	config_get config_file prometheus config_file "$CONFFILE"
-	config_get storage_tsdb_path prometheus storage_tsdb_path "/data"
+	config_get storage_tsdb_path prometheus storage_tsdb_path "/srv/prometheus"
 	config_get web_listen_address prometheus web_listen_address "127.0.0.1:9090"
+	
+	# Create tsdb dir & permissions if needed
+	if [ ! -d "$storage_tsdb_path" ]; then
+		mkdir $storage_tsdb_path
+		chown prometheus:prometheus $storage_tsdb_path
```
aparcar, line 21: Please add quotes around the variable

aparcar (changes requested):

sashasimkin:
I've fixed requested changes - can you merge it now 🙏 ?

PR #15272, opened 2021-03-26
mosquitto: update boot sequence and add respawn

srdgame:
Maintainer:  Karl Palsson <karlp@etactica.com>
Compile tested: sunxi h3 board
Run tested: SNAPSHOT

Description:
Mosquitto cannot be booted at the *FIRST* boot of device.

By added system("ifconfig -a > /tmp/mosq.log"), it says there is no any IP in all interfaces in *FIRST* boot

I think this is caused by relayd service is not started, thus there is not IP been set on lo? and br-lan not created?

So by change the service boot sequence to 81, this will fixed the *FIRST* boot issue.

And added respawn flag in case it crashed and cloud be rebooted.

### DIFF
--- a/net/mosquitto/files/etc/init.d/mosquitto
+++ b/net/mosquitto/files/etc/init.d/mosquitto
@@ -3,7 +3,7 @@
 # April 2012, OpenWrt.org
 # Provides support for the luci-app-mosquitto package, if installed
 
-START=80
+START=81
 USE_PROCD=1
 TCONF=/tmp/mosquitto.generated.conf
 CONF_WATCH=/etc/config/mosquitto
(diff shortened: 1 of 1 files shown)


karlp:
Yeah, I've already got this on my plate.  changing from 80 to 81 is just moving the goalposts, it will work for you maybe, but won't be enough for others.  (and for instance, it's not a problem for others at all) At the same time, trigger a sighup or a restart when interfaces change is unlikely to be the correct behaviour either.  This may be best addressed by changing things just for your own system.  The respawn flag is independent, and seems fine.

feckert:
@srdgame Please add a `Signed-off-by: ` with your realname and  a valid email address.

karlp:
Don't worry abotu that. I'll add the respawn, but like I said, changing the particular run number will just move goal posts around, it's no solution.

karlp:
respawn fixed in 0a33b49ff  some other solution needs to be worked out for network interface availability, and if it's appropriate for everyone.

PR #15273, opened 2021-03-26
CI: revert adding of *depending* packages

aparcar:
While the idea may make sense the current implementation is faulty.
Problem is that OpenWrt uses the folder name of packages within the
build system while `opkg` spits out the actual packages names.

An example, compiling the packages of folder `vim` (`make
package/vim/compile`) creates a package called `xxd`, where `make
package/xxd/compile` would fail.

The current implementation uses `opkg` to figure out dependent packages,
but the resulting names do not match the above mentioned folders.

Revert this for now until we come up with a better implementation to
avoid false positive CI failures.

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -82,16 +82,6 @@ jobs:
           echo "Building $PACKAGES"
           echo "PACKAGES=$PACKAGES" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


aparcar:
@BKPepe @neheb

neheb:
ACK

BKPepe:
Ok.

aparcar:
sadface.jpg

PR #15230, opened 2021-03-22
 forked-daapd: update to 27.4

ejurgensen:
Maintainer: me
Compile tested: Atheros ATH79, trunk
Run tested: no

Description:
Update to version 27.4. Incl remove "--enable-itunes" compile option (no longer exists).

Signed-off-by: Espen Jürgensen <espenjurgensen+openwrt@gmail.com>

PR #15274, opened 2021-03-26
forked-daapd: add missing dependency (libuuid)

ejurgensen:
Maintainer: me
Compile tested: Atheros ATH79, trunk
Run tested: No

Description:
Fix build error after PR #15230 due to missing libuuid dependency, see: https://downloads.openwrt.org/snapshots/faillogs/i386_pentium4/packages/forked-daapd/compile.txt

Signed-off-by: Espen Jürgensen <espenjurgensen+openwrt@gmail.com>


BKPepe:
Nitpick: Can you add the same email address, which you used for commit to the GitHub account? It will show your username instead of your full name so anyone can find you or at least it will be a lot of easier in the future.

ejurgensen:
Not sure exactly what I should change - can you advise? My git setting is `git config user.email espenjurgensen+openwrt@gmail.com`, and if I change that I think the checks start complaining that there is a mismatch with signed-of-by? I would like to keep the "+openwrt" for email filtering purposes.

ejurgensen:
Forget the provious message, I found out now. Yes, I have added the address.

BKPepe:
Thanks! :) Looks better.

PR #15275, opened 2021-03-26
acme: Get listener name from /proc/PID/exe instead of netstat output

### LINKED ISSUE #15071: acme, nginex: acme can't stop nginex instance
Maintainer: acme: @tohojo  nginx: @neheb, @peter-stadler 
Environment: OpenWrt SNAPSHOT r16142-12dbad1a86 / LuCI Master git-21.060.51374-cd06e70

Description:
---------------

Acme is not able to restart nginex server, because it is not getting properly detected.

```
port80 listens: 2868/uci.conf 2868/uci.conf
Cannot run in standalone mode; another daemon is listening on port 80.
```

In `netstat -nptl` command (used by acme to determinate what service is listening) command it is decribed as:

```
tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN      2868/uci.conf -g da
```

in `ps` it figgures as:

```
nginx: master process /usr/sbin/nginx -c /etc/nginx/uci.conf -g daemon off;
```

tohojo:
It seems the command name output from netstat can be truncated in weird
ways, so let's get the binary name from /proc instead and use that for
matching which listener we have.

Fixes #15071.

Signed-off-by: Toke Høiland-Jørgensen <toke@toke.dk>

Maintainer: me

Posting this as a draft so @majkrzak can test it :)

### DIFF
--- a/net/acme/Makefile
+++ b/net/acme/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=acme
 PKG_VERSION:=2.8.7
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/acmesh-official/acme.sh/tar.gz/$(PKG_VERSION)?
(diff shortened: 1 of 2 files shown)


majkrzak (approved):
Works fine for me :+1:

tohojo:
Great! Thanks for testing :)

PR #15284, opened 2021-03-27
openvpn: fix up down scripts execution in hotplug 01-user

ezplanet:
Signed-off-by: Mauro Mozzarelli <ezplanet@users.noreply.github.com>

Maintainer: me / @nbd168 
Compile tested: (arch: lantiq, BT Home Hub 5, OpenWrt 21.02)
  Run tested: (lantiq, BT Home Hub 5, OpenWrt 21.02, tests done)

Description: openvpn hotplug script 01-user tries to get up/down scripts parameters from the translated ovpn configuration file, but these were stripped and are no longer present.  This patch amends hotplug 01-user to get up/down scripts parameters from the uci openvpn instance configuration instead

Issue #15121, opened 2021-03-14
openvpn ignores up/down scripts, the replacement /usr/libexec/openvpn-hotplug ignores fundamental parameters


ezplanet:
Maintainer: Felix Fietkau  @ndb168 
Environment: Any - OpenWRT 21.02

Description: openvpn in OpenWRT 21.02 up/down scripts configuration is ignored

It appears that in OpenWRT 21.02 openvpn was changed to ignore up down scripts configuration.
It appears that undocumented script /usr/libexec/openvpn-hotplug replaced this configuration.
This script in turn picks only one parameter of the 7 passed by openvpn (the instance name) and ignores all the other parameters that are fundamental to setting up, for example, ip rules and ip routes on vpn connection.

Please see the following forum thread for a detailed explanation:
https://forum.openwrt.org/t/openwrt-21-02-openvpn-client-up-down-scripts-configuration-ignored/91126

ezplanet:
Please see my proposed fix in the following PR:
https://github.com/openwrt/packages/pull/15124

ezplanet:
Initially I followed the the hotplug scripts flow starting with /usr/libexec/openvpn-hotplug. Following a more in depth analysis I realized that it is sufficient to amend /etc/hotplug.d/openvpn/01-user that tries and fails to get up/down scripts configuration from the translated ovpn instance configuration. Those parameters are ignored in the translation from the uci openvpn configuration and thus are missing
This proposed change simply takes the parameters from the uci configuration instead.

https://github.com/openwrt/packages/pull/15125

ezplanet:
New PR created because previous ones did not pass formalities:
https://github.com/openwrt/packages/pull/15284

PR #15285, opened 2021-03-27
openvpn: fix invoking user up & down commands from hotplug wrapper

jow-:
Maintainer: @mkrkn 
Compile tested: x86/64
Run tested: x86/64 - tried OpenVPN client config using /etc/openvpn/test.conf, using /etc/config/openvpn referencing /etc/openvpn/test.conf and translating /etc/openvpn/test.conf to /etc/config/openvpn directly.

Description:

This commit adds a number of fixes to the OpenVPN up/down hotplug command
wrapper which currently fails to actually invoke user defined up and down
commands for uci configurations not using external native configurations.

 - Use the `--setenv` to pass the user configured `up` and `down` commands
   as `user_up` and `user_down` environment variables respectively

 - Instead of attempting to scrape the `up` and `down` settings from the
   (possibly generated) native OpenVPN configuration in
   `/etc/hotplug.d/openvpn/01-user`, read them from the respective
   environment variables instead

 - Fix parsing of native configuration values in `get_openvpn_option()`;
   first try to parse a given setting as single quoted value, then as
   double quoted and finally as non-quoted, potentially white-space
   escaped one. This ensures that `up '/bin/foo'` is interpreted as
   `/bin/foo` and not `'/bin/foo'`

Ref: https://forum.openwrt.org/t/openvpn-up-down-configuration-ignored/91126
Supersedes: #15121, #15284
Signed-off-by: Jo-Philipp Wich <jo@mein.io>

### DIFF
--- a/net/openvpn/Makefile
+++ b/net/openvpn/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=openvpn
 
 PKG_VERSION:=2.5.1
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_URL:=\
 	https://build.openvpn.net/downloads/releases/ \
--- a/net/openvpn/files/etc/hotplug.d/openvpn/01-user
+++ b/net/openvpn/files/etc/hotplug.d/openvpn/01-user
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-. /lib/functions/openvpn.sh
-
 [ -e "/etc/openvpn.user" ] && {
 	env -i ACTION="$ACTION" INSTANCE="$INSTANCE" \
 		/bin/sh \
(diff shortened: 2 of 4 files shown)


jow-:
Cc: @feckert

mkrkn:
Acked-by: Magnus Kroken <mkroken@gmail.com>

Thank you for the PR, I see my own fix-in-progress was incomplete.

feckert:
@jow- Thank you for taking on this poblem.
I only use the up down hotplug handling and have no custom up down scripts in my configuration. So I never had this problem
 
If I see it correctly, the configured custom scripts paths are passed as environment variables and then processed in hotplug handling. **Nice**

Sorry I don't have a running system to test right now. But I don't see any mistake there now.

On the amendment `Fix parsing of native configuration values in get_openvpn_option(); `I didn't take a closer look because I'm not an expert with **sed**. If you say so, then it will fit.

mkrkn:
Pinging some contributors with commit access to have this merged.

@aparcar @commodo @BKPepe @neheb @dangowrt

PR #13924, opened 2020-11-16
Suricata6: initial commit

Grommish:
Signed-off-by: Donald Hoskins <grommish@gmail.com>

Maintainer: me
Compile tested: mips64 mipsel / main branch 0f7a3288e10fb244a33d80230e7b04b323668821

Description:
Requires Rust/Cargo to compile (https://github.com/openwrt/packages/pull/13916)

PR #15287, opened 2021-03-28
libhtp: Initial commit OISF LibHTP 0.5.37

Grommish:
Maintainer: me 
Compile tested: mipsel_24kc, OpenWrt from mai

Description:
LibHTP is a security-aware parser for the HTTP protocol and the related bits and pieces.   It is used by the OISF Suricata IDS/IPS/NSM from https://github.com/openwrt/packages/pull/13924

### DIFF
--- a/libs/libhtp/Makefile
+++ b/libs/libhtp/Makefile
@@ -1,0 +1,54 @@
+#
+# Copyright (C) 2019-2020 CZ.NIC z.s.p.o. (http://www.nic.cz/)
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libhtp
+PKG_VERSION:=0.5.38
+PKG_RELEASE=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/OISF/libhtp/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=8365f338fdf5ef0f3b37f3cb4e2670d218c843eccc42aa87119819c0f018ef39
+
+PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=1
+PKG_FIXUP:=autoreconf
+
+PKG_MAINTAINER:=CZ.NIC <packaging@turris.cz>
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE
+PKG_CPE_ID:=cpe:/a:oisf:libhtp
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/libhtp
+  SECTION:=libs
+  CATEGORY:=Libraries
+  TITLE:=HTTP normalizer and parser library
+  URL:=https://github.com/OISF/libhtp
+  DEPENDS:=+zlib
+endef
+
+define Package/libhtp/description
+  LibHTP is a security-aware parser for the HTTP protocol and the related bits and pieces.
+endef
+
+CONFIGURE_ARGS += --disable-static
+
+define Build/InstallDev
+	sed -i 's|^Libs:\(.*\)*|Libs:\1 -liconv '"-L$(STAGING_DIR)/usr/lib/libiconv-full/lib"'|' $(PKG_INSTALL_DIR)/usr/lib/pkgconfig/htp.pc
+	sed -i 's|^prefix=\(.*\)*|prefix='"$(STAGING_DIR)"'\1|' $(PKG_INSTALL_DIR)/usr/lib/pkgconfig/htp.pc
+	$(CP) $(PKG_INSTALL_DIR)/* $(1)/
+endef
+
+define Package/libhtp/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libhtp*.so* $(1)/usr/lib/
+endef
+
+$(eval $(call BuildPackage,libhtp))


libs/libhtp/Makefile
```diff
@@ -0,0 +1,54 @@
...
+PKG_RELEASE=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/OISF/libhtp/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=0f8296ce9db815bbedc1c8a0b738dd25146fdaf910995a8c5f8e147799902b7e
+
+PKG_INSTALL:=1
+PKG_FIXUP:=autoreconf
+
+PKG_MAINTAINER:=CZ.NIC <packaging@turris.cz>
```
neheb, line 21: Was this copied from the turris repo?

libs/libhtp/Makefile
```diff
@@ -0,0 +1,54 @@
...
+
+PKG_NAME:=libhtp
+PKG_VERSION:=0.5.37
+PKG_RELEASE=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/OISF/libhtp/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=0f8296ce9db815bbedc1c8a0b738dd25146fdaf910995a8c5f8e147799902b7e
+
+PKG_INSTALL:=1
```
neheb, line 18: Should have PKG_BUILD_PARALLEL

libs/libhtp/Makefile
Grommish: It has the Copyright Block still, but yes, I believe so.  Although the version I pulled was outdated and had to be changed.

BKPepe: It depends, which branch you used, though. I updated it a week ago:
https://gitlab.nic.cz/turris/turris-os-packages/-/commit/90a75aae93f39f65de26edb4a5a074fbb413305c

Grommish: @BKPepe Was the one you linked to in the Suricata6 PR.  Was for version 0.5.33 I believe.  I dropped the -fno-stack-protector and repetitious Build/Compile from that version, and the ICONV stuff simply for expediency because I was confused over the iconv-full vs iconv/host and was getting errors.

libs/libhtp/Makefile
```diff
@@ -0,0 +1,55 @@
...
+	$(CP) $(PKG_INSTALL_DIR)/* $(1)/
+endef
+
+define Package/libhtp/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libhtp*.so* $(1)/usr/lib/
+endef
+
+$(eval $(call BuildPackage,libhtp))
+
```
BKPepe, line 55: Why you added here a newline? We can remove it, right? This is not in the Makefile.

libs/libhtp/Makefile
BKPepe: You should assign to yourself if you want to be maintainer, because currently I don't want to take maintainership of this package as there is no Suricata present in this repository. Also, I didn't use try it on OpenWrt master.

Which errors did you get?

libs/libhtp/Makefile
```diff
@@ -0,0 +1,55 @@
...
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/OISF/libhtp/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=0f8296ce9db815bbedc1c8a0b738dd25146fdaf910995a8c5f8e147799902b7e
+
+PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=1
+PKG_FIXUP:=autoreconf
+
+PKG_MAINTAINER:=CZ.NIC <packaging@turris.cz>
+PKG_LICENSE:=BSD 3-Clause
```
BKPepe, line 23:
```suggestion
PKG_LICENSE:=BSD-3-Clause
```

BKPepe (changes requested):
@Grommish:

1. If you are using things from somewhere, it is important to say where they are from.
You could use **commit description** (it's better than PR) or PR description. Something like this:
I used Makefile from here: https://gitlab.nic.cz/turris/os/packages/-/blob/191a014cb78d904001da81646ba49851ce47b129/libs/libhtp/Makefile
or
https://gitlab.nic.cz/turris/os/packages/-/blob/e2971b2109dabe98c0081951cda4a0f5857ee537/libs/libhtp/Makefile

Because right now, it looks like you are the author and does all the changes, which is not alright. 

1a) What you could do to looks better is left the original author (anyone from CZ.NIC would be OK, you could also use `Co-authored-by`) and then add you as committer with your Signed-off-by and in square brackets said something like: [Added foo, removed foo because ...]

2. Then it is important which changes you do and why.

**If you are adding something on behalf of someone, you should get ACK in that PR that he wants to be maintainer or not. If not, you should assign yourself.**, because you don't want to add any package when there is someone else as maintainer and he does not even know about that.

On the other hand, as you have all the testing environments, go ahead and finish it. :+1: 
Make sure that this is get compiled when you use `libiconv-full`.

Grommish:
> 
> Because right now, it looks like you are the author and does all the changes, which is not alright.
> 

Not only is the copyright block still intact, but the PKG_MAINTAINER is still listed as it was.  If anything, that will probably need to be changed.  However, if/when the package is accepted, I WILL be the maintainer and responsible for the changes in the OpenWrt Tree.  I'm not taking credit for anything, and I'd rather not deal with it at all, if I could.  But `libHTP` is something I need accessible.  Either thru this PR, or one you/others want to put up - as long as it's there, I'm good.  I'll pull and test within minutes of getting a link for it.

libs/libhtp/Makefile
Grommish: Maybe a `nano` artifact?  I'll check.

BKPepe:
Closing, OP is inactive.

PR #15291, opened 2021-03-28
mpd-full: enable lame and vorbis encoders

egorenar:
Enable Vorbis encoder for OGG format and lame for MP3.
This enables mpd to stream OGG and MP3 to e.g. Icecast2.

Maintainer: @neheb
Compile tested: Netgear R9000, OpenWrt SNAPSHOT r16368-9f31c417a2
Run tested: Netgear R9000, OpenWrt SNAPSHOT r16368-9f31c417a2

```
root@OpenWrt:~# mpd -V
Music Player Daemon 0.22.6 (reboot-16368-g9f31c417a2)

Encoder plugins:
 null vorbis opus lame wave flac
```

Example config:
```
audio_output {
        type            "shout"
        encoder         "lame"
        name            "my music"
        host            "192.168.1.1"
        port            "8000"
        mount           "/mpd"
        user            "source"
        password        "secret"
        bitrate         "64000"
        format          "48000:16:2"
}

# Need this so that mpd still works if icecast is not running
audio_output {
        type    "null"
        name    "fake out"
}
```

### DIFF
--- a/sound/mpd/Makefile
+++ b/sound/mpd/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mpd
 PKG_VERSION:=0.22.6
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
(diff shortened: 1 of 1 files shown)


neheb:
Why not.

PR #15299, opened 2021-03-29
mpd: run mpd as a non-root user

egorenar:

Maintainer: @neheb
Compile tested: Netgear R9000, OpenWrt SNAPSHOT r16368-9f31c417a2 
Run tested: Netgear R9000, OpenWrt SNAPSHOT r16368-9f31c417a2 

```
root@OpenWrt:~# ps aux | grep mpd
mpd      17800  4.7  1.3  36296 13436 ?        S<l  06:50   0:23 /usr/bin/mpd --no-daemon /etc/mpd.conf
```

```
root@OpenWrt:~# ls -l /mnt/sda1/openwrt/mpd/
-rw-r--r--    1 mpd      mpd          76892 Mar 29 06:50 database
drwxr-xr-x    1 mpd      mpd              0 Mar 29 06:50 playlists
```

Example config:
```
root@OpenWrt:~# cat /etc/mpd.conf
bind_to_address "192.168.1.1"

music_directory "/mnt/sda1/music/"
playlist_directory "/mnt/sda1/openwrt/mpd/playlists"
db_file "/mnt/sda1/openwrt/mpd/database"

audio_output {
        type            "shout"
        encoder         "lame"
        name            "my music"
        host            "192.168.1.1"
        port            "8000"
        mount           "/mpd"
        user            "source"
        password        "secret"
        bitrate         "64000"
        format          "48000:16:2"
}

# Need this so that mpd still works if icecast is not running
audio_output {
        type    "null"
        name    "fake out"
}
```

### DIFF
--- a/sound/mpd/files/mpd.init
+++ b/sound/mpd/files/mpd.init
@@ -9,6 +9,9 @@ PROG=/usr/bin/mpd
 CONFIGFILE=/etc/mpd.conf
 NICEPRIO=-10
 
+USER="mpd"
+GROUP="mpd"
+
 #TODO: Add uci config - nice, config
 
 start_service() {
@@ -16,7 +19,10 @@ start_service() {
 
 	#create mpd directories from config
 	pld=$(grep ^playlist_directory "$CONFIGFILE" | cut -d "\"" -f 2 | sed "s/~/\/root/g")
-	[ -d "$pld" ] || mkdir -m 0755 -p "$pld"
+	if [ ! -d "$pld" ]; then
+		mkdir -m 0755 -p "$pld"
+		chown $USER:$GROUP $pld
+	fi
 
 	lport=$(grep ^port "$CONFIGFILE" | cut -d "\"" -f 2)
 	[ -z "$lport" ] && lport=6600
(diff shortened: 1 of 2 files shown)


egorenar:
User has to make sure that `mpd.conf` is readable by `mpd` user, i had to fix permissions because `mpd.conf` hasd permissions `0600`.
Not sure how to handle it better. Any ideas ?

sound/mpd/files/mpd.init
```diff
@@ -9,21 +9,32 @@ PROG=/usr/bin/mpd
...
 
+USER="mpd"
+USERID="991"
+
 #TODO: Add uci config - nice, config
 
 start_service() {
 	local pld lport
 
+	user_exists $USER $USERID || user_add $USER $USERID
```
neheb, line 20: NAK. This can be handled in the Makefile.

neheb:
I believe mpd.conf is set to 0600 by INSTALL_CONF. INSTALL_DATA sets it to 0644.

I still think mpd needs a UCI config to avoid editing mpd.conf.

sound/mpd/files/mpd.init
egorenar: Ah thanks, fixed. I've seen `user_exists` in many other packages as example :)

neheb:
Needs a PKG_RELEASE bump.

sound/mpd/files/mpd.init
egorenar: One more question: do i have to hardcoded a specific user id like 991 or can it be automated. Some packages do not assign any user id.

egorenar:
> I believe mpd.conf is set to 0600 by INSTALL_CONF. INSTALL_DATA sets it to 0644.
> 
> I still think mpd needs a UCI config to avoid editing mpd.conf.

I agree wrt uci, that's on my todo list :)

sound/mpd/files/mpd.init
neheb: If you don’t include one it gets generated.

PR #15304, opened 2021-03-29
Fix null socket connections list

oskarirauta:
Maintainer: Daniel Golle / @dangowrt
Compile tested: x86_64, xeon powered server, OpenWRT snapshot (recent)
Run tested: x86_64, xeon powered server, OpenWRT snapshot (recent), tests: issue fixed, no more segfault

Description:
When podman stops a container, conmon segfaults.
This is output of podman when stopping a container:
`
error stopping container 0c1488f14bf30b16ad4f70a752400038d50fde4a63305d24d8543e3ce310ecab: timed out waiting for file /run/libpod/exits/0c1488f14bf30b16ad4f70a752400038d50fde4a63305d24d8543e3ce310ecab: internal libpod error
`

And this is output of system log:
`
Mon Mar 29 12:37:14 2021 kern.info kernel: [44090.269813] conmon[32066]: segfault at 8 ip 00007fa70ca8d549 sp 00007fff6bcaac50 error 4 in libglib-2.0.so.0.6600.7[7fa70ca23000+6f000]
Mon Mar 29 12:37:14 2021 kern.info kernel: [44090.271638] Code: eb d2 31 c0 5b 5d 41 5c 41 5d 41 5e c3 48 89 d1 31 d2 e9 9b ff ff ff 90 41 55 49 89 d5 41 54 49 89 f4 55 48 89 fd 53 31 db 51 <39> 5d 08 76 14 48 8b 45 00 89 da 4c 89 ee ff c3 48 8b 3c d0 41 ff
`

When conmon stops a container, it closes all currently active connections to container's socket, but segfault occurs when list of sockets is NULL, I also added a second check for NULL data just to be sure. These are minor changes that won't hurt anyone and this patch fixes the issue.

I also reported and provided a patch to conmon's maintainers, but it's not that sure if it ever gets there since this seems to be openwrt only issue, same issue for example does not occur in Alpine Linux which is very much similar to OpenWRT as it uses musl and does not need gigabytes......

### DIFF
--- a/utils/conmon/Makefile
+++ b/utils/conmon/Makefile
@@ -28,6 +28,7 @@ endef
 
 define Package/conmon/description
   Podman: A tool for managing OCI containers and pods
+  Conmon is necessary part of podman.
 endef
 
 define Package/conmon/install
--- a/utils/conmon/Makefile
+++ b/utils/conmon/Makefile
@@ -21,7 +21,7 @@ include ../../devel/meson/meson.mk
 define Package/conmon
   SECTION:=utils
   CATEGORY:=Utilities
-  TITLE:=Podmon conmon
+  TITLE:=Podman conmon
   URL:=https://podman.io
   DEPENDS:=+glib2 $(INTL_DEPENDS) $(ICONV_DEPENDS)
 endef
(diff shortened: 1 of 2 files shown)


oskarirauta:
Maybe I'll try this again with new pr..

dangowrt:
Thank you for helping getting podman run nicely on OpenWrt!
I've seen that weird segfault on container stop when I bumped podman to 3.0.1 and did superficial test, getting back to it... Did you check if Alpine is using the same version of glib2 or maybe they have more patches for glib2 on top which we don't?
In any case, having this fixed in which ever way is great :)
If you need help with this PR, let me know.

oskarirauta:
I do need help. Now everything should be set just right, except issue is that my patch file is marked as a “dirty patch” and I cannot find any mention on developer guide on how to fix that...

PR #15333, opened 2021-04-02
mosh: Fix for building mosh by adding libprotobuf-c to the dependency list

db4rne:
Signed-off-by: Arne Zachlod <arne@nerdkeller.org>

Maintainer:@neheb 
Compile tested: ramips, erx, master

Description:
mosh doesn't build and fails with libprotobuf not found without this patch.

### DIFF
--- a/net/mosh/Makefile
+++ b/net/mosh/Makefile
@@ -29,7 +29,7 @@ define Package/mosh/Default
   SECTION:=utils
   CATEGORY:=Utilities
   TITLE:=Mosh mobile shell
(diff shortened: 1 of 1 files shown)


neheb:
libprotobuf-c is not the same as protobuf. There is no reference to libprotobuf-c in the mosh code.

neheb:
Yeah. I can't reproduce your failure.

db4rne:
wait... you are saying it builds for you in ramips without the patch? maybe I should hava slept a bit more yesterday, but that is weird. I'll test again and resubmit if necessary, sorry for the noise.

PR #15338, opened 2021-04-04
nut: fix typo in nutshutdown script

SvenRoederer:
Maintainer: @neheb, @micmac1 

Description: Even it's only cosmetic and should not affect the function of regular system, fix the name of the IPKG_INSTROOT variable.


SvenRoederer:
@neheb can you also cherry-pick to 21.02? probably 19.07 is also affected

neheb:
Done for both.

PR #15341, opened 2021-04-04
strongswan: bump to 5.9.2

pprindeville:
Maintainer: me, @Thermi
Compile tested: x86_64, generic, head (85fa8ad8af)

Description:

Bump version to 5.9.2.

Retire weak algorithms like MD5 and 3DES.

### DIFF
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=strongswan
-PKG_VERSION:=5.9.1
-PKG_RELEASE:=9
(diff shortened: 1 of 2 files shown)


pprindeville:
BTW, how long before @Thermi can be added to the reviewers list?

Thermi (approved):
Very simple changes, look good to me from the strongSwan side.

stintel:
Probably a nit, but "Retire weak algorithms like MD5 and 3DES." in the commit message sounds like something you did. I don't see this in the changes, so it's probably just an upstream change. In this case, I would write the message indicating this was an upstream change. Now I'm just reading the commit message and I am thinking "why is this not reflected in the change".

PR #15366, opened 2021-04-08
dbus - Fix build

robcore:
Maintainer:Various
Compile tested: x86_64, generic, master
Run tested: x86_64, generic, master - tested and built without error

Description:
The recent move to cmake broke dbus build on my Ubuntu 20.04 machine due to an unnecessary backslash at the end of CMAKE_OPTIONS, and attempting to install qch files that were never built.  This solves both those problems (the latter may be unnecessary but it worked for me).


neheb:
Related: https://github.com/openwrt/packages/commit/0fb5d3ed2cb31a0a6076d36fb7a668cfe5328c92#commitcomment-49147445

The CMake update broke more than that.

BKPepe:
Test for formalities fails with following output:

```
Author name (Rob) need to be your real name 'firstname lastname'
Commit subject line MUST start with '<package name>: ' (Fix dbus build)
Signed-off-by is missing or doesn't match author (should be 'Signed-off-by: Rob <robpatershuk@gmail.com>')
```

robcore:
Yeah the test is flawed regarding the sign off.  It clearly states that first and last name are required yet its advice is to only put my first name.

The commit itself I would need to resubmit to rename but with the sign off warning it will still fail the formalities test.

The actual contents of the commit have been seen by @neheb so he's aware.  Besides, this only fixes the build for unused modules, but there are reports that functionality is still broken for those who actually use it so I'm letting this remain as reference material on the build side of the issues.  I just care that openwrt works, not getting credit for it.

BKPepe:
> Yeah the test is flawed regarding the sign off. It clearly states that first and last name are required yet its advice is to only put my first name.

No, it is not.
You can take a look e.g. here: https://github.com/openwrt/packages/commit/67efbcf5c4d7796aa5652805707fea548ea0d8aa.patch

You commit it with just your first name.

robcore:
Oh I see what happened.  I didn't have my last night name set in my github profile and I made the pull request using github's interface (thought I'd give it a shot).  I've updated the profile name so I can resubmit this with the formalities taken care of, so long as @neheb finds it useful enough for me to do so.  At the very least, the changes I've made here allow dbus to be built as a module, at least (I only encountered an issue because it must be a dependency for something my config builds a module for).

He may have something in mind that renders this moot, though, so I don't want to waste anyone's time.

egorenar:
I have this problem too on Void Linux as a build machine.

robcore:
Problem is fixed thanks to @neheb

BKPepe:
Which commits or PR fixes it? You mean the one which was not merged, yet? E.g. https://github.com/openwrt/packages/pull/15442

robcore:
Yup my bad meant to reference it.  That said, his PR solves both build and runtime fixes, so mine is moot in favour of his approach.

PR #15367, opened 2021-04-08
suricata-update: Initial commit

Grommish:
Provide the framework for Suricata-update

Maintainer: me 

Description:

Suricata-update framework, uci config and init.d service entry.

Requires [Suricata6](https://github.com/openwrt/packages/pull/13924)

### DIFF
--- a/net/suricata-update/Makefile
+++ b/net/suricata-update/Makefile
@@ -1,0 +1,48 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=suricata-update
+PKG_VERSION:=1.1.2
+PKG_RELEASE:=1
+PKG_LICENSE:=GPL-2.0
+
+include $(INCLUDE_DIR)/package.mk
+
+define Build/Prepare
+	true
+endef
+
+define Build/Configure
+	true
+endef
+
+define Build/Compile
+	true
+endef
+
+define Package/suricata-update
+    SUBMENU:=Firewall
+    SECTION:=net
+    CATEGORY:=Network
+    DEPENDS:=+suricata6 +python3-pip +python3-yaml
+    TITLE:=OISF Suricata IDS Update Utility
+    URL:=https://www.openinfosecfoundation.org/
+endef
+
+define Package/suricata-update/description
+ Suricata-Update provides updated ruleset lists for the Suricata IDS/IPS/NMS
+endef
+
+define Package/suricata-update/conffiles
+/etc/config/suricata-update
+endef
+
+define Package/suricata-update/install
+	$(INSTALL_DIR) $(1)/etc/init.d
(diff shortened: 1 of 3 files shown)


net/suricata-update/files/install_suricata_update
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh /etc/rc.common
+/usr/bin/python3 -m pip install --upgrade pip
```
neheb, line 2: uhhhh what?

Grommish:
The whole thing is ludicrous, but pip provides suricata-update, but I need
to get the config and init.d files to the device.  I am still working on
trying to get Package/postinst to work.

So...  Suggestions?

net/suricata-update/files/install_suricata_update
Grommish: It ensures pip3 is up to date, since it's currently outdated in master.  Not a great idea?

neheb:
ping @jefferyto

jefferyto:
There are lots of Python program packages in the repo, e.g. [youtube-dl](https://github.com/openwrt/packages/blob/master/multimedia/youtube-dl/Makefile), [docker-compose](https://github.com/openwrt/packages/blob/master/utils/docker-compose/Makefile), etc. If you search for "python3-package.mk" in the repo you will find many other examples. I suggest you take one of these makefiles as a template.

The Python packages folder readme also has a section on [building a Python package](https://github.com/openwrt/packages/blob/master/lang/python/README.md#building-a-python-package) which explains most (if not all) of the various Python package variables.

If you run into any problems writing a Python package makefile, there are lots of folks here who can help, but I'm afraid your `install_suricata_update` is definitely not the way to do this.

Grommish:
> 
> 
> There are lots of Python program packages in the repo, e.g. [youtube-dl](https://github.com/openwrt/packages/blob/master/multimedia/youtube-dl/Makefile), [docker-compose](https://github.com/openwrt/packages/blob/master/utils/docker-compose/Makefile), etc. If you search for "python3-package.mk" in the repo you will find many other examples. I suggest you take one of these makefiles as a template.
> 
> The Python packages folder readme also has a section on [building a Python package](https://github.com/openwrt/packages/blob/master/lang/python/README.md#building-a-python-package) which explains most (if not all) of the various Python package variables.
> 
> If you run into any problems writing a Python package makefile, there are lots of folks here who can help, but I'm afraid your `install_suricata_update` is definitely not the way to do this.

I absolutely agree.  What I have been playing with is simply a file check on the suricata-update command as part of the init.d service and calling `pip` to install it if it isn't there.  This allows me to include the service and have it self-installing without a scripted mess.  I've been busy with other things, but yes, this was totally a cluster 👍 

I'll get a push up eventually and see if that isn't more acceptable.  Until I get rust in, it's pretty much a moot point though.  It's still all a WIP.

BKPepe:
Closing, OP is inactive.

PR #15375, opened 2021-04-09
safe-search: prevent duplicate cron job installation

farmergreg:
Maintainer: me
Compile tested: ath79, x86
Run tested: latest openwrt snapshot ( r16450-d92a9c97bf  )

Description:
This patch prevents multiple cron jobs from being created to run the
safe-search-maintenance script.

To reproduce this bug, perform the following:
  - Install safe-search
  - Perform an OpenWRT firmware upgrade (choose to preserve user settings so that the crontab stays intact)
  - Install safe-search again

### DIFF
--- a/net/safe-search/Makefile
+++ b/net/safe-search/Makefile
@@ -6,8 +6,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=safe-search
-PKG_VERSION:=2.0.0
-PKG_RELEASE:=3
+PKG_VERSION:=2.0.1
+PKG_RELEASE:=$(AUTORELEASE)
 PKG_LICENSE:=MIT
 PKG_MAINTAINER:=Gregory L. Dietsche <Gregory.Dietsche@cuw.edu>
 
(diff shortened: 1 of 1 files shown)


net/safe-search/Makefile
```diff
@@ -6,8 +6,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=safe-search
-PKG_VERSION:=2.0.0
-PKG_RELEASE:=3
+PKG_VERSION:=2.0.1
+PKG_RELEASE:=1
```
neheb, line 10: could change this to $(AUTORELEASE)

BKPepe: Which is not mandatory, it is up to you.

farmergreg: I love it!
I've forgotten so many times to update PKG_RELEASE...

BKPepe:
Should I cherry-pick it also to OpenWrt 21.02?

farmergreg:
@BKPepe yes, please cherry-pick it into 21.02. Thanks!!

BKPepe:
Done in 49535edffdd44e1db109f687a5f6e87b7fe0ea3c.

farmergreg:
thank you!

Issue #15376, opened 2021-04-09
StrongSwan missing dependency under master branch!!


zxlhhyccc:
![image](https://user-images.githubusercontent.com/45259624/114189754-1fe51280-997d-11eb-9158-722cc584a209.png)

zxlhhyccc:
define Package/strongswan-libnttfft
$(call Package/strongswan/Default)
  TITLE+= nttfft library
  DEPENDS:= +strongswan
endef
Should be modified to：
define Package/strongswan-libnttfft
$(call Package/strongswan/Default)
  TITLE+= nttfft library
  DEPENDS:= +strongswan
endef

pprindeville:
Fixed with PR #15382.

PR #15382, opened 2021-04-10
strongswan: libnttft must not select strongswan

cotequeiroz:
Maintainer: @pprindeville, @thermi
Compile tested: none - should not be needed, menuconfig-tested only
Run tested: none

Description:

The strongswan-libnttfft package should not select the strongswan
package, but should depend on it instead.  Otherwise a circular
dependency is created.


BKPepe:
Fixes: https://github.com/openwrt/packages/issues/15376

pprindeville (approved):
LGTM

PR #14862, opened 2021-02-23
boinc: update to 7.16.16

neheb:
Remove upstreamed patch.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @N30dG
Compile tested: ath79

Issue #12248, opened 2020-05-21
boinc: did not transition from snapshot to 19.07.3


smoe:
Maintainer: @smoe  @N30dG
Environment: all, all, 19.07.3

Description: boinc is not installable with opkg, missing from 19.07 branch. Is there something we can do to help its adoption, say for 19.07.4 and/or 20.01?

hnyman:
For 20.0x it will get automatically, as the 20.0x branch will be branched off the HEAD of master (with boinc).

For 19.07 branch, somebody should author and test the PR for backporting.

(note that 19.07 is mainly June 2019 code plus fixes since then. But nothing gets automatically backported, and new packages are rarely backported. But if a package does not interfere with others, it is possible.)

Also the 19.07 packages are constantly built by the buildbot, so once the package is in 19.07 sources, it will be automatically available by opkg after a few days.

smoe:
Thank you for your kind reply, hnyman. My access to OpenWrt is mostly via Gluon. Initially I had hoped that Gluon would jump on 20.0x with its experimental release but the Gluon firmware just proudly adopted 19.07 (for their 2020.1.x and 2020.2.x trees) and from what I understand it will stick with 19.07 for a while. So, I would very much like to see boinc backported to 19.07. Can I do something to help initiate/accelerate the backport?

hnyman:
> Can I do something to help initiate/accelerate the backport?

You can build and test it on 19.07 and then author a backport pull request, like I already said above:

> > For 19.07 branch, somebody should author and test the PR for backporting.

BKPepe:
Closing. It was explained how it gets to OpenWrt 19.07. It does not need to be open here and stalled until somebody test it on 19.07 and send a pull request with cherry-picked related commits from master.

PR #15384, opened 2021-04-10
Transfer of boinc 7.16.16 from 21.02 to 19.07

smoe:
Intentionally unchanged from 43d21e650d4409b45ccc2c70fe507a29f783dda3,
i.e. the pull request #14862 from neheb/boi

Maintainer: @smoe 
Compile tested:
 - openwrt-sdk-19.07.7-ipq806x-generic_gcc-7.5.0_musl_eabi.Linux-x86_64
 - openwrt-sdk-19.07.7-mvebu-cortexa9_gcc-7.5.0_musl_eabi.Linux-x86_64

Run tested: ipq806x/arm_cortex-a15_neon-vfpv4 19.07 as shipping with Gluon 2020.2.2 on 10-Apr-2021 17:17:04 .

Description:

This is my attempt to follow the kind instructions of https://github.com/openwrt/packages/issues/12248 to bring BOINC to the masses of OpenWrt. The package itself is just fine, left everything unchanged from what already works on the snapshot (which is how Debian requests it for its backports). I found that for my Freifunk Community a few packages were not installable (like kmod-usb-storage). I prepared https://github.com/freifunk-gluon/gluon/issues/2198, since uncertain about where this problem should be fixed.

Signed-off-by: Steffen Moeller <moeller@debian.org>

### DIFF
--- a/net/boinc/Makefile
+++ b/net/boinc/Makefile
@@ -1,0 +1,83 @@
+# Copyright 2020 by Christian Dreihsig and Steffen Möller
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=boinc
+PKG_VERSION:=7.16.16
+PKG_VERSION_SHORT:=$(shell echo $(PKG_VERSION)| cut -f1,2 -d.)
+PKG_RELEASE:=1
+
+PKG_SOURCE_DATE:=2020-02-25
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/BOINC/boinc/tar.gz/client_release/$(PKG_VERSION_SHORT)/$(PKG_VERSION)?
+PKG_HASH:=0d5656a9f8ed1048936a5764270848b892d63f27bdb863d0ace447f1eaae6002
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-client_release-$(PKG_VERSION_SHORT)-$(PKG_VERSION)
+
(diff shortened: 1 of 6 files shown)


neheb:
@BKPepe opinion?

BKPepe:
Well, to be honest, I don't have a strong opinion on this one, and I don't mind both variants - if it gets it or not. I also know the feeling if you are working on it and that you want to get it to a large audience that is using stable release. Also, this is for a good thing, right? :-) I am pretty sure that this does not break anything, but if it is a question if we should add it to OpenWrt 19.07.

smoe:
I need to apologize for not having proposed this package any earlier to transition to stable. @N30dG already prepared a LuCI interface for BOINC, so the overall experience of OpenWrt with BOINC will improve further.

I know though that for TN-GRID and WorldCommunityGrid this works just fine even with small memory machines, and traffice does not seem affected. I can see BOINC change a bit with an increased acceptance of this OpenWrt BOINC package since these routers are basically comming with no extra overhead over what users are already running 24/7 and I myself want to have USB-controlled FPGA accepted as accelerators for BOINC in a not-too-far-future. To me, this BOINC package is a first tick on the checklist on my way towards that. To have this in stable is important for me mostly for the Freifunk community that I emerged from and respective feedback from a larger set of users. Nothing pressing. Would be nice, though.

While this package at the very periphery of OpenWrt certainly does not disturb, I am at the same time hoping for unforseen problems to surface that help the overall stability - just like the kmod-fs-ext4 not being available on some targets that have USB ports. The gluon folks have closed that bug report that I referenced in my PR, suggesting to just extend respective make variables for OpenWrt when building the image. Should you prefer me to transform that (still incomplete list of extra dependencies for USB storage) into a patch for OpenWrt then please kindly direct me.

neheb:
Enjoy.

PR #15399, opened 2021-04-11
uacme: add use_auto_staging

lemoer:
Maintainer: @lucize
Compile tested: tested: x86-64, openwrt master (https://github.com/openwrt/openwrt/commit/9397b22df1473f315552578b58322db7f7750361)
Run tested: tested: x86-64, openwrt master (https://github.com/openwrt/openwrt/commit/9397b22df1473f315552578b58322db7f7750361)

Description:

Staging certificates have the advantage that their retry limits are loose.
Therefore they can be obtained quickly when automatic retries are used.
Unfortunately they can not be used for deployments because their CA is not
accepted by clients. Production certificates do not have this limitation, but
their retry limits are strict. For production certificates, automatic retries
can only be performed a few times per hour. This makes automatic obtainment of
certificates tenacious.

With use_auto_staging=1, the advantages of the two certificate types are
combined. Uacme will first obtain a staging certificate. When the staging
certificate is successfully obtained, uacme will switch and obtain a production
certificate. Since the staging certificate has already been successfully
obtained, we can ensure that the production certificate is successfully
obtained in the first attempt. This means that "retries" are performed on the
staging certificate and the production certificate is obtained in the first
attempt.

In summary, this feature enables fast obtaining of production certificates when
automatic retries are used.

By default, this feature is set to use_auto_staging=0, which means that
uacme will behave as before by default.

### DIFF
--- a/net/uacme/files/run.sh
+++ b/net/uacme/files/run.sh
@@ -28,7 +28,7 @@ export CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
 export NO_TIMESTAMP=1
 
 UHTTPD_LISTEN_HTTP=
-STATE_DIR='/etc/acme'
+PRODUCTION_STATE_DIR='/etc/acme'
 STAGING_STATE_DIR='/etc/acme/staging'
 
 ACCOUNT_EMAIL=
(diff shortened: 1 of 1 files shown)


lucize:
LGTM

Issue #15370, opened 2021-04-08
python-greenlet not compiling on 32-bit MIPS


neheb:
Maintainer: @ja-pa 
Environment: 32-bit mips

https://downloads.openwrt.org/snapshots/faillogs/mipsel_24kc_24kf/packages/python-greenlet/compile.txt

@commodo @jefferyto I tried adding PKG_USE_MIPS16:=0 but I still see -mips16 being applied to CFLAGS. Any idea what's going on?

commodo:
hmm;
will take a look

it's likely that some python setup magic is getting in the way;

commodo:
so, the problem is that Python3 is not being built with `PKG_USE_MIPS16:=0`
the CFLAGS for Python modules is taken from `./staging_dir/target-mips_24kc_musl/usr/lib/python3.9/_sysconfigdata.py`

kind of like a `pkg-config` mechanism;

if i rebuild Python with `PKG_USE_MIPS16:=0` greenlet builds fine;

this is the first package that exposes this issue like this with MIPS;
or, maybe @jefferyto knows of other cases, and i wasn't paying attention;

when @ja-pa packaged it, he disabled it for `mips` and `arc` but not for `mipsel`

[1] so, we could add `mipsel`  to https://github.com/openwrt/packages/blob/master/lang/python/python-greenlet/Makefile#L34 

[2] or, we could just add  `PKG_USE_MIPS16:=0` to Python3 altogether;
i see that other interpreters do it already (node, php, golang)

Python is also in the class of these too-big-interpreters-for-small-OpenWrt-devices anyway

between [1] & [2] I am leaning towards [2];
but I don't have a really strong opinion; we can also have [1]

neheb:
The annoying thing is that this compiles fine under GCC10.

commodo:
> The annoying thing is that this compiles fine under GCC10.

Well, annoying or not, it happens with compilers. They get better over time, and we need to wait until old ones get phased out.

If no objections, I will convert Python3 to  `PKG_USE_MIPS16:=0` next week

BKPepe:
That's up to you as this is the first package, which exposes this issue, I'd be in favor to disable ``mipsel`` in `python-greenlet` package.

commodo:
Well, there were some users of greenlet that used to package it outside of OpenWrt feeds for MIPS.
[...]

PR #15404, opened 2021-04-12
python-greenlet: support MIPS architecture

commodo:
Maintainer: @ja-pa 
Compile tested: mips_24kc https://github.com/openwrt/openwrt/commit/d92a9c97bf3700e90af1d3c9157502af660365c0
Run tested: n/a

Fixes: https://github.com/openwrt/packages/issues/15370

This is inspired from:
   https://github.com/wlanslovenija/firmware-packages-opkg/blob/330bc94dccd16a3e92ac2fdde08c81a598e12f94/lang/python-greenlet/Makefile

The `PKG_USE_MIPS16:=0` is not taken into consideration when building
Python modules. That's because the sysconfig is used.

This is only an issue with greenlet (on MIPS) so far.

One option is to do `PKG_USE_MIPS16:=0` in the core Python package.
But, since we know that the `wlanslovenija` group has successfully used
greenlet on MIPS with this construct, we might as well adopt it until GCC10
becomes the main compiler.

As noted here:
  https://github.com/openwrt/packages/issues/15370#issuecomment-817015484
GCC10 doesn't have this problem.

### DIFF
--- a/lang/python/python-greenlet/Makefile
+++ b/lang/python/python-greenlet/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-greenlet
 PKG_VERSION:=1.0.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PYPI_NAME:=greenlet
 PKG_HASH:=719e169c79255816cdcf6dccd9ed2d089a72a9f6c42273aae12d55e8d35bdcf8
@@ -17,6 +17,8 @@ PKG_HASH:=719e169c79255816cdcf6dccd9ed2d089a72a9f6c42273aae12d55e8d35bdcf8
 PKG_MAINTAINER:=Jan Pavlinec <jan.pavlinec@nic.cz>
 PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=LICENSE
+# FIXME: remove when GCC10 is the oldest supported compiler, or the issue goes away
+PKG_USE_MIPS16:=0
 
 include ../pypi.mk
 include $(INCLUDE_DIR)/package.mk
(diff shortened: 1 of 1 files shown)


neheb:
Do note that you also have CONFIG_GCC_USE_VERSION_8 as part of the config file. Could make it conditional on that.

Based on my reading of what's happening in base, it seems version 9 will be skipped going straight to 10.

commodo:
> CONFIG_GCC_USE_VERSION_8

ok;
will use that

commodo:
> CONFIG_GCC_USE_VERSION_8

are you sure this gets defined in the `.config` file?
i can't seem to find it there

neheb:
All the versions are in there yes.

neheb:
See https://github.com/openwrt/packages/blob/master/libs/boost/Makefile#L385

commodo:
> See https://github.com/openwrt/packages/blob/master/libs/boost/Makefile#L385

I'm afraid there is some confusion here.
The only way I get `CONFIG_GCC_USE_VERSION_X` symbols, is if I manually enable `[*] Advanced configuration options (for developers)  --->`
I guess another options is to build with all CONFIG_ symbols enabled to some defaults.

If I disable advanced config options, all `CONFIG_GCC_USE_VERSION_X` symbols go away.

Now, I could do `ifndef CONFIG_GCC_USE_VERSION_10`
This makes things tricky, because [say] if someone wants to default to GCC10 [now], they will get the `TARGET_CFLAGS += -mno-mips16 -mno-interlink-mips16`  option anyway.
I wouldn't be bothered if this happens. Not sure about others.

Anyway, this is getting maybe a little more complicated than [I feel] it needs to be.
I'd prefer to set these CFLAGS for MIPS as done here; if anyone using greenlets needs some more fine-tuning, they can submit a PR or something.

Right now, greenlets seems a bit more used on Turis, which is [I think] ARM.

neheb:
Hrm didn't know that. This is good enough then.

PR #15406, opened 2021-04-13
strongswan: handle chacha20poly1305 as AEAD

pprindeville:
Maintainer: me, @Thermi
Compile tested: x86_64, generic, head (880c8b4422)
Run tested: same, built and installed on test VM

Description:

Having an explicit hash algorithm with AEAD ciphers is not permitted.

### DIFF
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -141,6 +141,8 @@ is_aead() {
 	case "$cipher" in
 	aes*gcm*|aes*ccm*|aes*gmac*)
 		return 0 ;;
+	chacha20poly1305)
+		return 0 ;;
 	esac
 
 	return 1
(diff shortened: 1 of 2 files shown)


net/strongswan/files/swanctl.init
```diff
@@ -134,6 +134,8 @@ is_aead() {
 	case "$cipher" in
 	aes*gcm*|aes*ccm*|aes*gmac*)
 		return 0 ;;
+	chachapoly1305)
```
Thermi, line 137: chacha20poly1305

pprindeville: Fixed

Thermi (approved):

pprindeville:
Per @jow-, added logic to avoid starting the service if the configuration is too severely broken.

Thermi (approved):

Issue #14870, opened 2021-02-23
CI: repo packages have higher priority than built ones during build test


luizluca:
Description:

Tests for #14696 are installing dependencies packages from repo instead of using the ones available during the build.
When a makefile generates both pkgA-boo and pkgA-foo and pkgA-boo requires pkgA-foo, it looks like the CI scripts gets the pkgA-foo from repository instead of using the just compiled one.

There are many cases but I'll focus on ruby-singleton:

```
Testing package ruby-erb in version 3.0.0 from ruby
...
Installing ruby-singleton (2.7.2-1) to root...
Downloading https://downloads.openwrt.org/snapshots/packages/aarch64_cortex-a53/packages/ruby-singleton_2.7.2-1_aarch64_cortex-a53.ipk
...
Testing package ruby-singleton in version 3.0.0 from ruby
Upgrading ruby-singleton on root from 2.7.2-1 to 3.0.0-1...
Removing obsolete file /usr/lib/ruby/2.7/singleton.rb.
Removing obsolete file /usr/lib/ruby/2.7/singleton/version.rb.
Removing obsolete file /usr/lib/ruby/gems/2.7/specifications/default/singleton-0.1.0.gemspec.
Configuring ruby-singleton.
No test.sh script available
Removing package ruby-drb from root...
Removing package ruby-rake from root...
Removing package ruby-singleton from root...
...
```

Maybe I could use EXTRA_DEPENDS for locking the version requirements to force the selection of a specific version but OpenWrt does not seem to heavily use version dependencies.

luizluca:
I submitted new ruby released for 18.07 through master.

It looks like 19.07 CI is trying to use snapshot packages (ruby-3.0.0) to solve 19.07 dependencies (ruby-2.6.7) https://github.com/openwrt/packages/pull/15413/checks?check_run_id=2330493500

Shouldn't it use 19.07-snapshot?

PR #15413, opened 2021-04-13
[19.07] ruby: update to 2.6.7

luizluca:
Fixes two CVEs:

CVE-2020-25613: Potential HTTP Request Smuggling Vulnerability in WEBrick
CVE-2021-28965: XML round-trip vulnerability in REXML

After this release, ruby 2.6 is now in security maintenance phase.

Maintainer: me
Compile tested: arc_generic, armvirt_64, ath79_generic, mediatek_mt7622, mvebu_cortexa9, powerpc, ramips_mt7620, x86_64, x86_generic
Run tested: x86_64 running "gem update"

Description:

### DIFF
--- a/lang/ruby/Makefile
+++ b/lang/ruby/Makefile
@@ -11,7 +11,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ruby
-PKG_VERSION:=2.6.6
+PKG_VERSION:=2.6.7
 PKG_RELEASE:=1
 
 # First two numbes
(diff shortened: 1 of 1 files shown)


BKPepe (approved):
It's interesting that it fails for some targets, but I'm in favor to merge this in.

luizluca:
> It's interesting that it fails for some targets, but I'm in favor to merge this in.

It is using snapshot (ruby-3.0.0) to solve dependencies for this PR, which targets 19.07 (19.07-snapshot). I already reported #14870 because CI prefer snapshot packages instead of preferring the ones generated by the PR, even if older. If CI was using 19.07-snapshot, it will probably still use ruby-2.6.6 (19.07-snapshot) instead of ruby-2.6.7 (this PR). In that case, it might work as dependencies where not changed.

BKPepe:
Okay, thank you for your clarification!

PR #15431, opened 2021-04-15
strongswan: add certificate generation utility

pprindeville:
Maintainer: me, @Thermi
Compile tested: x86_64, generic, head (cb3fb45ed1)

Description:

Generate X.509 certificates backed by a common (self-signed) CA, also generated by this tool.

### DIFF
--- a/net/strongswan/files/gencerts.sh
+++ b/net/strongswan/files/gencerts.sh
@@ -1,0 +1,155 @@
+#!/bin/sh
+
+#
+# see:
+#	https://www.howtoforge.com/tutorial/strongswan-based-ipsec-vpn-using-certificates-and-pre-shared-key-on-ubuntu-16-04/
+#
+
+PROG=$(basename "$0")
+
+[ -z "$EUID" ] && EUID=$(id -u)
+
+if [ $# -lt 5 ]; then
+	echo "Usage: $PROG { -s | -c | -u } country domain organization identities [ ... ]" >&2
+	exit 1
+fi
+
+case "$1" in
+-s)
+	S_OPT=1 ;;
+-c)
+	C_OPT=1 ;;
+-u)
+	U_OPT=1 ;;
+*)
+	echo "$PROG: require an option specifying server/client/user credential type" >&2
+	exit 1
+	;;
+esac
+shift
+
+C="$1"; shift
+DOMAIN="$1"; shift
+SHORT_DOMAIN="${DOMAIN%%.*}"
+ORG="$1"; shift
+
+# invariants...
+STRONGSWANDIR=/etc
+SWANCTL_DIR=$STRONGSWANDIR/swanctl
+: ${KEYINFO:="rsa:4096"}
+: ${CADAYS:=3650}
+: ${CRTDAYS:=730}
+
+makeDN()
+{
+	printf "C=%s, O=%s, CN=%s" "$1" "$2" "$3"
+}
+
+field()
+{
+	local arg="$1"
+	local nth="$2"
+
+	echo "$arg" | cut -d ':' -f "$nth"
+}
+
+genmasterkey()
+{
+	local keytype keybits
+
+	keytype=$(field "$KEYINFO" 1)
+	keybits=$(field "$KEYINFO" 2)
+
+	pki --gen --type "$keytype" --size "$keybits" --outform pem > "$SWANCTL_DIR/private/$SHORT_DOMAIN.key"
+	chmod 0400 "$SWANCTL_DIR/private/$SHORT_DOMAIN.key"
+}
+
+genca()
+{
+	local keytype
+
+	keytype=$(field "$KEYINFO" 1)
+
+	pki --self --ca --lifetime "$CADAYS" --in "$SWANCTL_DIR/private/$SHORT_DOMAIN.key" --type "$keytype" \
+		--dn "$ROOTDN" --outform pem > "$SWANCTL_DIR/x509ca/$SHORT_DOMAIN.crt"
+	chmod 0444 "$SWANCTL_DIR/cacerts/$SHORT_DOMAIN.crt"
+}
+
+genclientkey()
+{
+	local name="$1" keytype keybits
+
+	keytype=$(field "$KEYINFO" 1)
+	keybits=$(field "$KEYINFO" 2)
+
+	pki --gen --type "$keytype" --size "$keybits" --outform pem > "$SWANCTL_DIR/private/$name.key"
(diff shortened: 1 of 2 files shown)


net/strongswan/files/gencerts.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/bash
```
neheb, line 1: What?

neheb:
Please run through shellcheck.

net/strongswan/files/gencerts.sh
pprindeville: It requires bash.  It won't work with ash.

pprindeville:
> Please run through shellcheck.

Done

stintel:
I'm not a fan. It's generally recommended not to generate private keys and certificates on OpenWrt due to many devices lacking entropy. If you decide to add this anyway, the script should give a big fat warning about this imo.

champtar:
You could use getrandom to make sure the kernel csprng is properly seeded

net/strongswan/files/gencerts.sh
```diff
@@ -0,0 +1,155 @@
...
+DOMAIN="$1"; shift
+SHORT_DOMAIN="${DOMAIN%%.*}"
+ORG="$1"; shift
+
+# invariants...
+STRONGSWANDIR=/etc
+SWANCTL_DIR=$STRONGSWANDIR/swanctl
+KEYINFO="rsa:4096"
+CADAYS=3650
+CRTDAYS=730
```
Thermi, line 41: Please use ${parameter:-word} to set these values, so users can overwrite them using environmental variables.

net/strongswan/files/gencerts.sh
```diff
@@ -0,0 +1,155 @@
...
+[ -f "$SWANCTL_DIR/x509ca/$SHORT_DOMAIN.crt" ] || genca
+
+PARENT="$STRONGSWANDIR"
+BASEDIR="${SWANCTL_DIR##$PARENT/}"
+
+for name in "$@"; do
+	setparams "$name"
+	gendev
+
+	tar -zcf "$name-certs.tar.gz" -C "$PARENT" "$BASEDIR/x509ca/$SHORT_DOMAIN.crt" "$BASEDIR/x509/$name.crt" "$BASEDIR/private/$name.key"
```
Thermi, line 150: Please make sure the file permissions are 600 before creating the tar file. E.g. Check if you can touch and chmod the file first, then overwrite it with tar. On my system (Arch Linux) that preserves the previous ACLs. So they're the ones you set via chmod before. Otherwise, if you just create the tar, the file will be accessible to world (0644 ACLs).

Thermi:
> You could use getrandom to make sure the kernel csprng is properly seeded

The random bytes for the keys are generated/retrieved by the crypto library providing the code for generating the key. That is dependent on the plugins loaded at runtime. So that's a problem that can only be solved by making sure the lib (openssl, wolfssl, botan, ...) does the right thing.

pprindeville:
> You could use getrandom to make sure the kernel csprng is properly seeded

Not following... it's a script, not a C utility.

net/strongswan/files/gencerts.sh
pprindeville: Well, a few lines above I do:

```
umask 077
```

so this is just extra paranoia.

net/strongswan/files/gencerts.sh
pprindeville: Used:

```
: ${VAR:=default}
```

instead.

pprindeville:
Should the CA key be password protected by default?

It's only ever needed when the script is run.

champtar:
> > You could use getrandom to make sure the kernel csprng is properly seeded
> 
> Not following... it's a script, not a C utility.

We have (or had) in OpenWrt a little getrandom utility that just call getrandom() so when it returns you should be good to go in theory

pprindeville:
> We have (or had) in OpenWrt a little getrandom utility that just call getrandom() so when it returns you should be good to go in theory

Isn't that potentially a race condition?  What if two different scripts are both blocked on it?

The only thing that's truly foolproof is a blocking read from `/dev/random`.

pprindeville:
> You could use getrandom to make sure the kernel csprng is properly seeded

You're talking about `ubox`'s `getrandom`?

champtar:
> > We have (or had) in OpenWrt a little getrandom utility that just call getrandom() so when it returns you should be good to go in theory
> 
> Isn't that potentially a race condition? What if two different scripts are both blocked on it?
> 
> The only thing that's truly foolproof is a blocking read from `/dev/random`.

Race condition with 1 process only ??
/dev/random might block again so this is a bad idea, best thing is to have the key generation use getrandom() directly but it seems there are many possible (buggy ?) implementations for the key generation.

champtar:
> > You could use getrandom to make sure the kernel csprng is properly seeded
> 
> You're talking about `ubox`'s `getrandom`?

Yup

pprindeville:
> Race condition with 1 process only ??
> /dev/random might block again so this is a bad idea, best thing is to have the key generation use getrandom() directly but it seems there are many possible (buggy ?) implementations for the key generation.

I'm saying if there's another script also using `getrandom` at the same time.

Why is `/dev/random` a bad idea?  It blocks until there's sufficient entropy.

If `pki` is proceeding when it doesn't have sufficient entropy, that would seem to be a separate bug that should be prosecuted independently.

net/strongswan/files/gencerts.sh
```diff
@@ -0,0 +1,155 @@
...
+shift
+
+C="$1"; shift
+DOMAIN="$1"; shift
+SHORT_DOMAIN="${DOMAIN%%.*}"
+ORG="$1"; shift
+
+# invariants...
+STRONGSWANDIR=/etc
+SWANCTL_DIR=$STRONGSWANDIR/swanctl
```
Thermi, line 38: Please also make these defaults.

pprindeville: Uh... why?

Thermi: So people can test run the script in other dirs or relative to a different directory. E.g. if they want to prepare a CA in a directory with binaries of a different, architecture incompatible, device.

Thermi (approved):

net/strongswan/files/gencerts.sh
pprindeville: Changed `STRONGSWANDIR` to `SYSCONFDIR` as per your out-of-band suggestion.

I'm not convinced this change isn't in the "give them enough rope" category.

PR #15432, opened 2021-04-15
clamav: add libiconv dependencies when build with NLS

BKPepe:
Maintainer: @lucize 
Compile tested: Turris Omnia, mvebu (cortex-a9), pre-OpenWrt 21.02
Run tested: N/A

**Description:**

NLS means Native Language Support and when you have it enabled (it is
not default), clamav can not be compiled as it shows following error:

Package clamav is missing dependencies for the following libraries:
libiconv.so.2

Also, it is required that package libiconv-full is compiled first/before
than clamav and then try to compile clamav.


lucize:
LGTM

PR #15436, opened 2021-04-15
strongswan: change name of config base directory

pprindeville:
Maintainer: me, @Thermi
Compile tested: x86_64, generic, head (cb3fb45ed1)

Description:

Style changes that got left out of previous commit.

### DIFF
--- a/net/strongswan/files/gencerts.sh
+++ b/net/strongswan/files/gencerts.sh
@@ -34,8 +34,8 @@ SHORT_DOMAIN="${DOMAIN%%.*}"
 ORG="$1"; shift
 
 # invariants...
-STRONGSWANDIR=/etc
-SWANCTL_DIR=$STRONGSWANDIR/swanctl
+SYSCONFDIR=/etc
+SWANCTL_DIR="$SYSCONFDIR/swanctl"
 : ${KEYINFO:="rsa:4096"}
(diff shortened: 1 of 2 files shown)


net/strongswan/files/gencerts.sh
```diff
@@ -34,8 +34,8 @@ SHORT_DOMAIN="${DOMAIN%%.*}"
 ORG="$1"; shift
 
 # invariants...
-STRONGSWANDIR=/etc
-SWANCTL_DIR=$STRONGSWANDIR/swanctl
+SYSCONFDIR=/etc
```
Thermi, line 37: Forgot to make it default

Thermi (changes requested):

net/strongswan/files/gencerts.sh
pprindeville: As I said, I'm not convinced that this isn't in the "give them enough rope" category.  But if more people ask me, then I might change it.

What would make more sense is to have variables for CACERTDIR, CERTDIR, KEYDIR instead, since `ipsec` and `swanctl` use different directory names.

Thermi: Alright.

Thermi (approved):

PR #15442, opened 2021-04-18
dbus: fix new cmake build

neheb:
CMake was using the wrong paths. Fix them up.

Maintainer: nobody
Compile tested: ath79

ping @DazzyWalkman

### DIFF
--- a/utils/dbus/Makefile
+++ b/utils/dbus/Makefile
@@ -72,6 +72,9 @@ $(call Package/dbus/Default/description)
 endef
 
 CMAKE_OPTIONS += \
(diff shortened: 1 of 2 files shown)


DazzyWalkman:
The new cmake build compile & start-up fine on x86-64. It's usable.
However, when I tried to start/stop dbus during test, the new build will not quit using
`/etc/init.d/dbus stop`
No error messages given, the process was still running.
I had to kill the dbus process otherwise dbus would refuse to start again as there's already a dbus-daemon running.
This does not happen on the autoconf build.

wicadmin:
When will this get fixed?

neheb:
This is not in OpenWrt 21.02 nor do I plan on backporting it.

PR #15443, opened 2021-04-18
libtalloc: add Python3/host dependency

ryzhovau:
Maintainer: @thess, @jefferyto.
Compile tested: Clean Debian 10 host, Entware repo ([commit](https://github.com/Entware/entware-packages/commit/99c05d4192787cf4824e6f873a1c99fb476dc543)).
Run tested: Entware, mipsel feed

Description: bring back `Python3/host` dependency to prevent following error:
```
make package/libtalloc/compile V=s
…
The distutils module is unusable: install "python-devel"?
(complete log in /build/me/E.mipsel/build_dir/target-mipsel_mips32r2_glibc-2.27/talloc-2.3.1/bin/config.log)
make[2]: *** [Makefile:93: /build/me/E.mipsel/build_dir/target-mipsel_mips32r2_glibc-2.27/talloc-2.3.1/.configured_68b329da9893e34099c7d8ad5cb9c940] Error 1
make[2]: Leaving directory '/build/me/E.mipsel/feeds/packages/libs/libtalloc'
```

### DIFF
--- a/libs/libtalloc/Makefile
+++ b/libs/libtalloc/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=talloc
 PKG_VERSION:=2.3.1
 MAJOR_VERSION:=2
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://www.samba.org/ftp/talloc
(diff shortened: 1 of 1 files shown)


BKPepe:
I can see that this is also happening on OpenWrt 21.02 branch. Here is my full log: https://gist.github.com/BKPepe/9bbf83fec7a8f5756a7af57ba7253985

It would be nice if you can describe in the commit description why it is necessary because right now, I see that you are adding dependency without any explanation. However, it is in PR, but while looking in the git history in CLI, you will not most likely look for PR, where it was added.

ryzhovau:
Done.

neheb:
Quite interesting.

As meson in OpenWrt currently uses the host python3 binary instead of python3/host, I've patched out distutils stuff where it's needed.

Probably fine to depend on python3/host here.

PR #15459, opened 2021-04-19
[19.07] nextdns: Update to version 1.31.3

rs:
Update to version 1.31.3


neheb:
Wonder why this fails CI.

neheb:
Ah. Could be an older incompatible version of go.

rs:
Please don't merge, this tag is a typo

rs:
The v1.12.3 is coming to replace this. Sorry about that.

PR #15461, opened 2021-04-19
[19.07] nextdns: Update to version 1.12.3

rs:
Update to version 1.12.3

### DIFF
--- a/net/nextdns/Makefile
+++ b/net/nextdns/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nextdns
-PKG_VERSION:=1.11.0
+PKG_VERSION:=1.12.3
 PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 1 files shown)


neheb:
19.07 has golang 1.13.15. I take it this doesn't work with that?

neheb:
Yeah looks like it:

```
2021-04-19T23:30:02.4007865Z go build github.com/lucas-clemente/quic-go/internal/qtls: build constraints exclude all Go files in /home/build/openwrt/build_dir/target-x86_64_musl/nextdns-1.12.3/.go_work/build/pkg/mod/github.com/lucas-clemente/quic-go@v0.20.0/internal/qtls
```

rs:
That's probably the issue. I will submit a version without quic/http3 support for 19.07 so it does compile.

rs:
@neheb seems like GO_PKG_TAGS is not supported in 19.07. Do you know another way to pass a Go build tag?

rs:
A 1.12.5 is coming with a workaround

PR #15474, opened 2021-04-20
unbound: add cache-max-negative-ttl config option

ja-pa:
Maintainer: @EricLuehrsen

Description:
This PR adds a new UCI option for cache-max-negative-ttl and lowers default ttl from 3600 to 1000.

The use case for changing this value isn't too frequent and is related to weird behavior observed on some networks  (see unbound upstream issue NLnetLabs/unbound#465).


EricLuehrsen:
Looks good. Ready to merge if you change it out of draft.

EricLuehrsen (approved):

ja-pa:
Thanks. Changed from draft to ready for merge

BKPepe:
Well, I wanted to merged it, but I dont see any reference in the commit description either why you are doing such changes in the commit itself.

EricLuehrsen:
Ready to merge: I don't see a need for a "why" in this commit message. A **simple** upstream option from Unbound itself was exposed through UCI. "Negative Cache" and its TTL have a standard DNS definition. Tweaking this TTL to resolve upstream brokenness is a delicate issue with local organizations having their own reasons, real and philosophical. Much of that discussion is opinionated and should not be part of the git history.

PR #15478, opened 2021-04-20
[21.02] nextdns: Update to version 1.32.0

rs:
Update to version 1.32.0


BKPepe:
Within 18 hours, there are 4 updates from you. Once it was updated to 1.31.3 then it was downgraded to 1.23.3 and then to 1.25.5 in OpenWrt 21.02 branch and now there is update to 1.32.0. What is different? Are there any release notes? Would it be possible to include it in the PR description or in the commit?

rs:
1.31 was a fat fingered tag that created bad version. We fixed it later but some deb/rpm base client picked it already and were unable to "downgrade" to the fixed tag. After several tries to fix it we decided to just jump the version to 1.32 to fix all those clients. It is the exact same code as 1.12.5.

Sorry about the confusion.

PR #15480, opened 2021-04-20
podman: update to 3.1.1

oskarirauta:

Compile tested: x86_64, server, recent SnapShot
Run tested: x86_64, server, recent snapshot, tested

Description:
Update podman to version 3.1.1

**Other changes**
 - Updated init script: stop pods and containers on shutdown.
 - Enable AppArmor support

### DIFF
--- a/utils/podman/Makefile
+++ b/utils/podman/Makefile
@@ -1,9 +1,9 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=podman
-PKG_VERSION:=3.0.1
+PKG_VERSION:=3.1.1
 PKG_RELEASE:=$(AUTORELEASE)
-PKG_HASH:=259e682d6e90595573fe8880e0252cc8b08c813e19408b911c43383a6edd6852
+PKG_HASH:=4e6fb106c6363566b6edc4ac6caee0bdf6b788e01255c3b3bfcb64f4b6842229
 
 PKG_SOURCE_URL:=https://github.com/containers/podman/archive/v$(PKG_VERSION)
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
@@ -17,7 +17,7 @@ PKG_BUILD_PARALLEL:=1
 PKG_USE_MIPS16:=0
 
 GO_PKG:=github.com/containers/podman/
-GO_PKG_BUILD_PKG:=github.com/containers/podman/v2/cmd/podman/
+GO_PKG_BUILD_PKG:=github.com/containers/podman/v3/cmd/podman/
 
 include $(INCLUDE_DIR)/package.mk
 include ../../lang/golang/golang-package.mk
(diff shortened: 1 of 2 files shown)


BKPepe:
Please document your changes in the commit description because this is not a simple bump as stated in the commit subject "update to version 3.1.1" as you are doing many other changes.

utils/podman/Makefile
```diff
@@ -40,21 +40,21 @@ define Package/podman/Default
...
   VARIANT:=default
   PROVIDES:=podman
 endef
 
 define Package/podman-selinux
   $(call Package/podman/Default)
-  TITLE+= with SELinux support
-  DEPENDS+= libselinux
+  TITLE += with SELinux support
+  DEPENDS:=$(GO_ARCH_DEPENDS) +conmon +cni +cni-plugins +btrfs-progs +glib2 +gnupg2 +iptables +libgpg-error +libseccomp +libgpgme +nsenter +zoneinfo-simple +kmod-veth +libselinux
```
BKPepe, line 57: Why the previous issue was not okay for you?  While using `DEPENDS+=` there can be multiple dependencies.

oskarirauta: Sorry. My bad. My "private" version wasn't updated yet to that change which I made to version that I updated here earlier. Committing soon..

oskarirauta: @BKPepe fixed.

oskarirauta:
> Please document your changes in the commit description because this is not a simple bump as stated in the commit subject "update to version 3.1.1" as you are doing many other changes.

Not that many.  Below is complete expanded list of changes and reasons to them.

I found out that If you change location of containers to persistent storage instead of tmpfs, starting them will fail unless they have been stopped. If this is the case that reboot has occurred before pods and containers have been stopped, they cannot be started, they have to be removed and re-created. Change in initscript tries to avoid that. Even if containers are running at tmpfs, this won't hurt. Still, if something happens and system hangs/reboots/etc, script won't save you from that. It's just a attempt to make things easier. Luckily, there's no real damage when containers "fail" as you are supposed to _mount_ dynamic content anyways..

I provided a pull request for AppArmor. Before provided python3-readline is required by AppArmor. If AppArmor is accepted, applications supporting it natively, should have it enabled- this also won't hurt or grow codebase too much either.

BKPepe:
Now, it is better, thanks! The explanation could be added to the commit description as you already wrote it, but if you want to leave it as it is, I am fine with that as well.

However, we can not merge this first before apparmor is not merged.

BKPepe (approved):
Now, it is better, thanks! The explanation could be added to the commit description as you already wrote it, but if you want to leave it as it is, I am fine with that as well.

Otherwise it looks good to me, thanks! Once CI pass, I will merge it.

oskarirauta:
> Now, it is better, thanks! The explanation could be added to the commit description as you already wrote it, but if you want to leave it as it is, I am fine with that as well.
> 
> Otherwise it looks good to me, thanks! Once CI pass, I will merge it.

If you insist.. But I think it's good that way, there's a good reasons why I made these minor changes and I am not sure at all if they require that much of explaining. Well, I'll change it for you..

oskarirauta:
> > Now, it is better, thanks! The explanation could be added to the commit description as you already wrote it, but if you want to leave it as it is, I am fine with that as well.
> > Otherwise it looks good to me, thanks! Once CI pass, I will merge it.
> 
> If you insist.. But I think it's good that way, there's a good reasons why I made these minor changes and I am not sure at all if they require that much of explaining. Well, I'll change it for you..

Done

BKPepe (approved):

PR #15369, opened 2021-04-08
python3-readline: add package

oskarirauta:
Signed-off-by: Oskari Rauta <oskari.rauta@gmail.com>

Maintainer: Oskari Rauta / @oskarirauta
Compile tested: x86_64, xeon powered server, recent OpenWRT snapshot
Run tested: x86_64, xeon powered server, recent OpenWRT snapshot

Description:
Python wrapper to libreadline

Includes patch that allows it to compile without issues and instead of static link, links dynamically against system's libreadline- although, author never said or claimed that there would be any static linking anywhere..

PR #15481, opened 2021-04-20
apparmor: add new package

oskarirauta:

Maintainer: Oskari Rauta / @oskarirauta (find it by checking history of the package Makefile)
Run tested: x86_64, server, recent snapshot, tested, works

Description:
Contains AppArmor userland libraries and utilities.

Requires python3-readline #15369


oskarirauta:
@dangowrt - this is what we talked about earlier.

One item has been left out from this, apparmor pam helper plugin. I had it in here, but it required apparmor first to be built without pam helper plugin, so that libapparmor is installed on staging dir, until pam helper could be compiled. Probably this can be worked out, but at the moment, I think this is pretty nice initial commit for apparmor package.

oskarirauta:
Haven't tested it without NLS.. Let's see if it works out without..

oskarirauta:
Patience.. Seems that it's still work in progress, let's see if it now compiles..

feckert:
> AppArmor does not work independently - It requires auditd as it's companion.

@oskarirauta I read in your description that you need [auditd](https://github.com/linux-audit/audit-userspace) for AppArmor.
I'm only asking because I'm working on another task where I need auditd to find out who changed which file in the system.
For this I thought auditd could help me further. Or could AppArmor does this task also?

The only problem is that, as far as I know, this service (auditd-userspace) is not supported for all architectures.
https://github.com/linux-audit/audit-userspace/pull/133
https://github.com/linux-audit/audit-userspace/pull/115
https://github.com/linux-audit/audit-userspace/pull/73

If that is the case, then we need to limit its use in the DEPENDS field of AppArmor.
I haven't looked into it more closely now if you use audit-userspace or only the audit subsystem of the kernel (this works for the most/all architectures as I can see)

With which architectures did you test it?

oskarirauta:
> > AppArmor does not work independently - It requires auditd as it's companion.
> 
> @oskarirauta I read in your description that you need [auditd](https://github.com/linux-audit/audit-userspace) for AppArmor.
> I'm only asking because I'm working on another task where I need auditd to find out who changed which file in the system.
> For this I thought auditd could help me further. Or could AppArmor does this task also?
> 
> The only problem is that, as far as I know, this service (auditd-userspace) is not supported for all architectures.
> [linux-audit/audit-userspace#133](https://github.com/linux-audit/audit-userspace/pull/133)
> [linux-audit/audit-userspace#115](https://github.com/linux-audit/audit-userspace/pull/115)
> [linux-audit/audit-userspace#73](https://github.com/linux-audit/audit-userspace/pull/73)
> 
> If that is the case, then we need to limit its use in the DEPENDS field of AppArmor.
> I haven't looked into it more closely now if you use audit-userspace or only the audit subsystem of the kernel (this works for the most/all architectures as I can see)
> 
> With which architectures did you test it?

I only tested it on x86_64, but this is interesting point- I will need to look this up. Thanks.

oskarirauta:
@feckert 

audit's [Makefile](https://github.com/openwrt/packages/blob/master/utils/audit/Makefile) does not mention anything about not working on some archs.

I have tested this approach yet, but there are 2 kernel config flags that I haven't yet enabled or tried:
```CONFIG_SECURITY_PATH=y
CONFIG_SECURITY_NETWORK=y
```

Most likely, what you describe, is possible with AppArmor and **CONFIG_SECURITY_PATH** flag. As default AppArmor is set to unconfined state, which would mean that it will log changes to files to auditd log with information on username that did or attempted this. Then with proper configuration you can opt-out parts of system's paths to not get logged, and also you can set parts to not allowed for read/write/etc or combination of these, without using acl's.

But I haven't so far tested this. Let's see when I get these tests to run, it worked out at my setup just fine, but I have enabled full libintl through nls enabling, so it's struggling a bit here on tests where nls is disabled, but soon it's done..

oskarirauta:
Currently there seems to be problem with installing python parts that I didn't see when I compiled locally, most likely due to my build tree comes with quite a lot of things, so this was missed as dep..

```2021-04-21T08:04:22.9025249Z /home/build/openwrt/staging_dir/host/bin/python3 python-tools-setup.py install --prefix=/usr --root=/home/build/openwrt/build_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/apparmor-3.0.0/ipkg-install-utils --version=3.0.0
2021-04-21T08:04:22.9129613Z Traceback (most recent call last):
2021-04-21T08:04:22.9130515Z   File "python-tools-setup.py", line 23, in <module>
2021-04-21T08:04:22.9131270Z     from distutils.command.install import install as _install
2021-04-21T08:04:22.9132212Z ModuleNotFoundError: No module named 'distutils.command'
2021-04-21T08:04:22.9142569Z make[3]: *** [Makefile:61: install] Error 1
```

Strange. I though this came from python3-setuptools but I am obviously mistaken..

oskarirauta:
I hate python.. ;)

oskarirauta:
Finally..

oskarirauta:
It seems that I got all necessary kernel options added to config files as well - so doing a pr on openwrt.git next..

oskarirauta:
> AppArmor is an effective and easy-to-use Linux application security system. AppArmor proactively protects the operating system and applications from external or internal threats, even zero-day attacks, by enforcing good behavior and preventing both known and unknown application flaws from being exploited.
> 
> AppArmor supplements the traditional Unix discretionary access control (DAC) model by providing mandatory access control (MAC). It has been included in the mainline Linux kernel since version 2.6.36 and its development has been supported by Canonical since 2009.
>
> [https://apparmor.net](https://apparmor.net)

**Requirements**
 - Requires package python3-readline #15369
 - Various kernel flags

Since management of selinux isn't working at the moment and not anytime soon, I decided to investigate on other similar solutions and came up with AppArmor. Whilst AppArmor and SELinux have differences, they have very similar aim - so I started developing on how to integrate AppArmor on OpenWRT and could that be managed. Turns out I succeeded on this task. I have been testing it now for a month or so, and it works and can be managed.

My build is patched. Most patches are changes to python and shell scripts to avoid too many depencies and other minor changes. I also rebuilt apparmor functions script basing it on existing script, it's now less complicated, more informative (smarter log messages) and also a lot shorter than originally provided script. Also there is init script included.

AppArmor does not work independently - It requires auditd as it's companion and kernel support.

Required kernel flags can be enabled in kernel configuration of menuconf after applying patch from this [PR](https://github.com/openwrt/openwrt/pull/4101)

oskarirauta:
> The only problem is that, as far as I know, this service (auditd-userspace) is not supported for all architectures.
> [linux-audit/audit-userspace#133](https://github.com/linux-audit/audit-userspace/pull/133)
> [linux-audit/audit-userspace#115](https://github.com/linux-audit/audit-userspace/pull/115)
> [linux-audit/audit-userspace#73](https://github.com/linux-audit/audit-userspace/pull/73)
> 
> If that is the case, then we need to limit its use in the DEPENDS field of AppArmor.
> I haven't looked into it more closely now if you use audit-userspace or only the audit subsystem of the kernel (this works for the most/all architectures as I can see)

I don't think we need to change DEPENDS field of AppArmor, as AppArmor already depends on audit, if audit's depends are updated and one is attempting to build AppArmor on arch that does not support audit, AppArmor cannot be built. Recursive depends.

oskarirauta:
I broke this commit/pr and cannot fix it. Re-creating it, sorry..

oskarirauta:
Replaced by #15643

PR #15484, opened 2021-04-20
parted: add new package

oskarirauta:

Description:
GNU Parted manipulates partition tables.
This is useful for creating space for new operating systems,
reorganizing disk usage, copying data on hard disks and disk imaging.

### DIFF
--- a/utils/parted/Makefile
+++ b/utils/parted/Makefile
@@ -1,0 +1,53 @@
+# SPDX-License-Identifier: GPL-2.0-only
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=parted
+PKG_VERSION:=3.4
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=http://ftp.gnu.org/gnu/parted/
+PKG_HASH:=e1298022472da5589b7f2be1d5ee3c1b66ec3d96dfbad03dc642afd009da5342
+
+PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=1
+
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/parted
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=GNU Parted
+  SUBMENU:=Disc
+  DEPENDS:=+libblkid +libuuid +PARTED_READLINE:libreadline +CONFIG_PARTED_READLINE:libncurses +PARTED_LVM2:libdevmapper
+endef
+
+define Package/parted/description
+  GNU Parted manipulates partition tables. This is useful for
+  creating space for new operating systems, reorganizing
+  disk usage, copying data on hard disks and disk imaging.
+endef
+
+define Package/parted/config
+  source "$(SOURCE)/Config.in"
+endef
+
+ifneq ($(CONFIG_PARTED_READLINE),y)
+  CONFIGURE_ARGS += --without-readline
+endif
+
+ifneq ($(CONFIG_PARTED_LVM2),y)
+  CONFIGURE_ARGS += --disable-device-mapper
+endif
+
+define Package/parted/install
+	$(INSTALL_DIR) $(1)/usr/lib $(1)/sbin
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/*.so** $(1)/usr/lib/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/sbin/parted $(1)/sbin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/sbin/partprobe $(1)/sbin/
+endef
+
+$(eval $(call BuildPackage,parted))
--- a/utils/parted/Config.in
+++ b/utils/parted/Config.in
@@ -1,0 +1,14 @@
+menu "Configuration"
+	depends on PACKAGE_parted
+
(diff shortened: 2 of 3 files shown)


utils/parted/Makefile
```diff
@@ -0,0 +1,39 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=parted
+PKG_VERSION:=3.4
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=http://ftp.gnu.org/gnu/parted/
```
BKPepe, line 8:
```suggestion
PKG_SOURCE_URL:=https://ftp.gnu.org/gnu/parted/
```

utils/parted/Makefile
```diff
@@ -0,0 +1,39 @@
...
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+PKG_LICENSE_FILES:=COPYING
+PKG_LICENSE:=GPL-3.0-only
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/parted
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=GNU Parted
```
BKPepe, line 23:
```suggestion
  TITLE:=GNU Parted
  URL:=https://www.gnu.org/software/parted/
```

BKPepe (commented):
Would be good if you can create test.sh (like others do sometimes in this repository), so our CI which we are using can check the version in the Docker image at least some kind of run testing.  Otherwise, this looks good!

utils/parted/Makefile
oskarirauta: Done

utils/parted/Makefile
oskarirauta: Done

oskarirauta:
> Would be good if you can create test.sh (like others do sometimes in this repository), so our CI which we are using can check the version in the Docker image at least some kind of run testing. Otherwise, this looks good!

Maybe. I needed partx command; utility to refresh changed partition table and accidentally built parted, but let's not put that effort to waste.. Now I just am thinking, that where is my partx utility's package now, have I accidentally removed it? Or was it named something else than partx, can't remember...

Currently OpenWRT has a package that should do this, but I wasn't able to get it update /dev nodes, it did read new partition table, but that wasn't quite enough as I had format them and didn't want to boot at that moment.. So that's why I went this way..

oskarirauta:
> Would be good if you can create test.sh (like others do sometimes in this repository), so our CI which we are using can check the version in the Docker image at least some kind of run testing. Otherwise, this looks good!

Sorry, haven't seen one so far.. I made run tests in a production environment ;) box is still running so I guess it worked out ;)

BKPepe:
> Sorry, haven't seen one so far...

Take a look at patckages e.g. vim, apk, netdata, whois, tmux and such.

oskarirauta:
> > Sorry, haven't seen one so far...
> 
> Take a look at patckages e.g. vim, apk, netdata, whois, tmux and such.

I could add one- but it won't work in docker. I would need to know for e.g. root hard disks node, and still it wouldn't work, as is this is a partition manager + tool to refresh partitions to kernel. So only refreshing would had possibility to work, but would need full privileges cap..

oskarirauta:
`partprobe /dev/vda`

PolynomialDivision:
@oskarirauta You can just do a quick version check in test.sh.

oskarirauta:
Sure.. One moment..

oskarirauta:
I added some configuration options. LVM2 support is now optional; this reduces depencies and size of package. Usually LVM2 after all is required when using RAID configuration.

Oops- something is wrong. Tests are failing..

PolynomialDivision:
@oskarirauta Here is a good example: https://github.com/openwrt/packages/blob/master/net/gnunet/test.sh

The line returns with != 0 if the version string is not found.
More information here: https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md

oskarirauta:
@PolynomialDivision - Requested change has been pushed.

Hopefully it works, I did not test it, due to $2 being unknown to me.

oskarirauta:
Yes. Well it's just that it's been quite a while since @BKPepe visited this pr..

This package contains tool that is also provided by busybox, which is main reason that I created a package for this, since partprobe from busybox failed. It did detect updated partition table, but was not able to update dev nodes. This one did. I only needed partprobe, and I actually never really have use parted, since I really like gdisk, but it's included there as well since it belongs to this software..

Anyway, sometimes it's unnecessary to re-invent the wheel, whether it was adding another version of partprobe, even though a version of it is available- or a package that someone might need. Should I close this? Ignorance is not a bliss, if you are waiting..

wulfy23 (approved):
all good aarch64

PolynomialDivision:
> This package contains tool that is also provided by busybox, which is main reason that I created a package for this, since partprobe from busybox failed. It did detect updated partition table, but was not able to update dev nodes.

Did you report the issue somewhere?

Could you maybe change the tests like @wulfy23 described? I think it is fine to merge this.

oskarirauta:
No, I did not report it. Sure, I'll change test.sh soon..

wulfy23:
apologies... i'm unfamiliar with docker... so the example
tests were written to explicitly run the full variant in
the case of conflicts and to more specifically
test real disk interaction...

it appears my suggestions are unsuitable for this test
suite / format...

you can revert the test commands ( keep the new case
for partprobe ) to your original version check with perhaps
something more suitable to be implemented at a later
date

oskarirauta:
@wulfy23, now test.sh works as it should..
@PolynomialDivision not it works..

wulfy23 (approved):
thanks...

PR #15485, opened 2021-04-20
gnu-efi: add new package

oskarirauta:

Description:
gnu-efi binaries, required by some applications (like gummiboot #15486)


Yang-Wei-Ting:
Was there a specific reason for using the fork?

oskarirauta:
can't remember, but likely there was if it was chosen..

PR #15486, opened 2021-04-20
gummiboot: add new package

oskarirauta:

Description:
Simple UEFI boot manager

Requires #15485: gnu-efi

### DIFF
--- a/utils/gummiboot/Makefile
+++ b/utils/gummiboot/Makefile
@@ -1,0 +1,66 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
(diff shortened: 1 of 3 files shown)


oskarirauta:
I have a system that boots to Syslinux/extlinux on bios systems and gummiboot handles boot on UEFI systems and starts Syslinux/extlinux so for both systems, I have same configuration.

Unfortunately, I wasn't able to cross-compile syslinux, so that one I cannot publish since it's natively compiled and my build tree just re-packs pre-compiled code without stripping it (to avoid binaries from breaking).

But if someone is able to cross-compile syslinux, I'd be grateful, I have a pretty OpenWRT background image for it too ;)

PR #15487, opened 2021-04-20
conntrack-tools: add a patch to fix endianness issue

gongtao0607:
Maintainer: me / jo@mein.io

Compile tested: 
x86_64 kvm, QEMU, OpenWrt 19.07.3, and
Atheros AR7161 rev 2, Netgear WNDR3800, OpenWrt 19.07.5

Run tested: 
Mixed endianness firewall failover successful

Description:
There's a endianness bug in conntrack-tools causing SNAT/DNAT IP address being inverted in a mixed-endianness failover. (In my case, a little endian firewall fails-over to a big endian firewall). This patch is a backport of https://marc.info/?l=netfilter&m=161891034013669&w=2 to solve this issue.

### DIFF
--- a/net/conntrack-tools/Makefile
+++ b/net/conntrack-tools/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=conntrack-tools
 PKG_VERSION:=1.4.6
-PKG_RELEASE:=1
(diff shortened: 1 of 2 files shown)


neheb:
I think PKG_RELEASE needs to be bumped.

neheb:
ping @jow-

PR #15488, opened 2021-04-21
atlas-sw-probe: improve key creation

Ansuel:
- Exit start if a probe_key is not present
- Add create_key command to generate a private_key based on the provided username in the atlas config.

@ja-pa 

Signed-off-by: Ansuel Smith <ansuelsmth@gmail.com>


Ansuel:
Some context...

I tested the package and it works good...

Problem is that i notice that if the atlas-probe script is run without a key, by adding one and restarting the probe won't register.
I need to remove the data dir in /tmp and restart to actually start the probe registration. I fixed this by simply not start when there isn't a key.

I added a create_key function to facilitate all the things...
I still need to test if the dropbearkey function generate a correct private key. It does generate a binary private key so i don't know if it does work with atlas-probe scripts.

Anyway i can confirm that this package works in 2 different router. (both r7800 running openwrt)

net/atlas-sw-probe/files/atlas.init
```diff
@@ -57,6 +58,40 @@ create_backup() {
...
+
+		public_key="$(dropbearkey -t rsa -f "$PRIV_KEY_FILE" -s 2048 | sed -n 2p)"
+		public_key="$(echo "$public_key" | sed "s/ \S*$/ "$username"/")"
+		echo $public_key > "$PUB_KEY_FILE"
+	else
+		print_msg "Can't find a way to generate key."
+		exit 1
+	fi
+
+	print_msg "Key generate successfully. Uset the get_key command to show the public key and get instruction on how to register your probe."
```
ja-pa, line 92:
```suggestion
	print_msg "Key generate successfully. Use the get_key command to show the public key and get instruction on how to register your probe."
```

BKPepe:
Guys, don't forget to bump `PKG_RELEASE`.

Ansuel:
I notice that the key is lost on sysupgrade... I'm testing my solution and I will add it to this pr

Ansuel:
Ok I tested the changes... Now should be all correct. @ja-pa can you checkout the changes? What do you think about variable names? Any suggestions?

Ansuel:
I notice that dropbear private key is not compatible... so I had to add the package dropbearconvert if dropbear is present (with openssl server, ssh-keygen should be available right?)

Anyway I tested now and the converted private key works correctly and the probe correctly registers.

net/atlas-sw-probe/files/atlas.init
```diff
@@ -57,6 +58,58 @@ create_backup() {
...
+		echo $public_key > $probe_pub_key
+		dropbearconvert dropbear openssh /etc/atlas/probe_key_dropbear $probe_key
+		rm /etc/atlas/probe_key_dropbear
+	else
+		print_msg "Can't find a way to generate key."
+		exit 1
+	fi
+
+	#Link priv/pub key
+	[ ! -f $PRIV_KEY_FILE ] && ln -s $probe_key $PRIV_KEY_FILE
```
neheb, line 102: could replace with -f ||

net/atlas-sw-probe/files/atlas.readme
```diff
@@ -0,0 +1,12 @@
...
+The atlas probe software requires a rsa 2048-4096 key for registration.
+
+Follow these steps to register your probe on the ripe-atlas systems.
+1. Insert your username in the atlas config file (/etc/config/atlas)
+2. Use the command '/etc/init.d/atlas create_key' to create a priv/pub key.
+3. The priv/pub key will be stored on the directory /etc/atlas/
+4. Use the command '/etc/init.d/atlas get_key' to get the public key used for probe registration.
+   Make sure to copy the entire key and that the last value is the correct username
+5. Follow the instruction from the past command or go to 'https://atlas.ripe.net/apply/swprobe/'
+   and register your probe.
```
neheb, line 12: Missing newline

Ansuel:
@neheb i read some warning for the dropbearconvert bin... Do you think that can cause some security risk ?

neheb:
What error?

Ansuel:
```
All arguments must be specified
Usage: dropbearconvert <inputtype> <outputtype> <inputfile> <outputfile>

CAUTION: This program is for convenience only, and is not secure if used on
untrusted input files, ie it could allow arbitrary code execution.
All parameters must be specified in order.

The input and output types are one of:
```

ja-pa:
LGTM. Thanks !

net/atlas-sw-probe/files/atlas.readme
```diff
@@ -0,0 +1,12 @@
+# Atlas probe setup instruction
```
BKPepe, line 1: I am wondering if we want to ship this file together with the file as it seems we are supplying documentation. Well, there should be, I know, but in my view, I can copy&paste [the documentation](https://wiki.turris.cz/doc/en/howto/atlas-probe) which I wrote together with @ja-pa and move it to OpenWrt's documentation.

BKPepe: However, we can solve it later. Let's dont delay merging this. :-)

PR #15505, opened 2021-04-24
lighttpd: patches from upstream

gstrauss:
lighttpd: patches from upstream

Maintainer: @MikePetullo
Compile tested: ar71xx

Description:
- update files/lighttpd.conf as HTTP/2 is now enabled by default since lighttpd-1.4.59
- ignore Content-Length from backend if 101 Switching Protocols
- close HTTP/2 connection after bad password
- skip cert chain build for self-issued certs
- meson zstd fix
- ls-hpack upstream update
- discard some HTTP/2 DATA frames received after response

Signed-off-by: Glenn Strauss <gstrauss@gluelogic.com>

### DIFF
--- a/net/lighttpd/Makefile
+++ b/net/lighttpd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lighttpd
 PKG_VERSION:=1.4.59
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 # release candidate ~rcX testing; remove for release
 #PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-1.4.59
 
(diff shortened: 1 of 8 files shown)


gstrauss:
...It is very annoying that openwrt build process does not cleanly accept patches created with `git format-patch` and that whatever quilt errors have occurred are not visible in the logs other than "Refresh patch ..."

gstrauss:
...so the build fails because quilt feels the need to truncate lines?  That's unkind
```
-@@ -2347,16 +2370,25 @@ h2_send_end_stream_data (request_st * const r, connection * const con)
+@@ -2347,16 +2370,25 @@ h2_send_end_stream_data (request_st * co
```

gstrauss:
Is there anything additional that needs to be done to have these patches included in OpenWrt 21.02.0 ?  Thanks.

neheb:
@gstrauss the alternative to quilt doing that is dirty patches with no real way to mandate quality. make package/x/refresh is a standardized way to clean them.

neheb:
Backported to 21.02.

gstrauss:
Thanks for the hint, @neheb  I hadn't known about the `refresh` target

PR #15510, opened 2021-04-25
h2o: remove useless ruby dependency

Habbie:
Maintainer: @neheb @James-TR 
Compile tested: x86_64, OpenWRT master
Run tested: in the Docker hub image `openwrt/rootfs`, I installed dnsdist and this h2o build, and it could still serve DoH fine.

Description: h2o is the library dnsdist uses to offer DNS over HTTPS to clients. dnsdist is the only user of h2o in this tree.

While h2o can depend on Ruby (to build mruby support), this is disabled in the OpenWRT build of h2o. Hence, the Ruby dependency is unnecessary, and removing it saves a few megabytes of disk space.

Additionally, I noticed that the package contained 3 identical copies of the lib:
```
root@52170cbc2408:/# ls -ali /usr/lib/libh2o*
 162653 -rwxr-xr-x    1 root     root        348857 Apr 25 11:50 /usr/lib/libh2o-evloop.so
 162660 -rwxr-xr-x    1 root     root        348857 Apr 25 11:50 /usr/lib/libh2o-evloop.so.0.13
 162661 -rwxr-xr-x    1 root     root        348857 Apr 25 11:50 /usr/lib/libh2o-evloop.so.0.13.6
```

so this PR fixes that too:
```
root@472ad3a8404e:/# ls -ali /usr/lib/libh2o*
 289858 lrwxrwxrwx    1 root     root            21 Apr 25 12:43 /usr/lib/libh2o-evloop.so -> libh2o-evloop.so.0.13
 289859 lrwxrwxrwx    1 root     root            23 Apr 25 12:43 /usr/lib/libh2o-evloop.so.0.13 -> libh2o-evloop.so.0.13.6
 289860 -rw-r--r--    1 root     root        348857 Apr 25 12:41 /usr/lib/libh2o-evloop.so.0.13.6
```

### DIFF
--- a/libs/h2o/Makefile
+++ b/libs/h2o/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=h2o
 PKG_VERSION:=2.2.6
-PKG_RELEASE:=8
+PKG_RELEASE:=9
 
 PKG_SOURCE_URL:=https://codeload.github.com/h2o/h2o/tar.gz/v${PKG_VERSION}?
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


stintel:
Removing the Ruby dependency and fixing the duplicate so files are two unrelated changes, so should go in separate commits.

stintel (approved):

PR #15527, opened 2021-04-27
nginx: make nginx-rtmp module BigEndian friendly

jbubik:
Maintainer: Ansuel Smith <ansuelsmth@gmail.com>
Compile tested: ath79, TP-Link Archer C7 v5, 19.07
Run tested: ath79, TP-Link Archer C7 v5, 19.07, 10 months production use (OBS stream to NGINX to Youtube)
Signed-off-by: Jan Bubík <jbubik@centrum.cz>

Description:
The original nginx-rtmp-module was coded for LittleEndian machines. Changes for LE/BE compatibility were started by original author but never finished. I did some work a year ago and made it usable on both LE (cygwin x64) and BE (mips). PR against original repo still open - seems the repos are abandoned. Therefore I offer the fix to OpenWrt directly.  See https://github.com/ut0mt8/nginx-rtmp-module/pull/24

### DIFF
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -450,6 +450,7 @@ ifeq ($(CONFIG_NGINX_RTMP_MODULE),y)
   define  Prepare/nginx-rtmp
 	$(eval $(Download/nginx-rtmp))
 	xzcat $(DL_DIR)/$(FILE) | tar -C $(PKG_BUILD_DIR) $(TAR_OPTIONS)
+	$(call PatchDir,$(PKG_BUILD_DIR)/nginx-rtmp,./patches-rtmp-nginx)
   endef
 endif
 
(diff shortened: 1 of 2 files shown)


neheb:
ping @Ansuel

neheb:
Needs a PKG_RELEASE bump.;

Ansuel:
@neheb look good only problem is... Do we have some way to make this more friendly to the make refresh command?
It's not related to this pr specially but since we now have 4 directory with patches for submodule this is starting to became a problem.

neheb:
Hmm doesn't look like it. the x in make package/x/refresh is based on the directory of the Makefile, not necessarily the package.

@lipnitsk any input?

Ansuel:
If i remember correctly, one problem that i had every time i had to refresh patches of nginx is that refreshed patches all goes in the 
'patches' dir and not in the dedicated one. Don't know if this has been fixed.

neheb:
I think a rebase against current master will fix the CI.

jbubik:
I am completely new to OpenWrt build tools. The CI builds fail (for some platforms) in package "node" compiling dependency "libmraa". As far as I understand, this is not related to PR #15527. It must be a problem in code already committed to master branch.

Anyhow the patch is not applied at all (for all platforms). I have originally added this line into Makefile (works on 19.07 branch):
`$(call PatchDir,$(PKG_BUILD_DIR)/nginx-rtmp,./patches-rtmp-nginx)`
Has anything changed in Makefile processing between 19.07 and master?

I am not able to help with this further, sorry. Ideas?

jbubik:
FYI: I am trying to push the fix upstream. That should make it easier for OpenWrt package.
@neheb: let's keep this open, I will update my branch once upstream is ready.

neheb:
Yeah I have no idea what's going on with node.

PR #15529, opened 2021-04-28
seafile: drop packages and related quirks

commodo:
Maintainer: me, @jefferyto 
Compile tested: n/a
Run tested: n/a

This package got added around 2014, but maintaining it was mostly a pain.
The seafile suite keeps around some packages that are only needed to
support this package suite, like
* vala - Vala Compiler
* libevhtp - some static library only
* django1 - Django 1.11 (last LTS is 3.0)
    
The amount of quirks and effort to keep this package around becomes nearly
insane. The maintainer that brought it initially is no longer around, I am
co-maintaining it with @jefferyto , but at least I am getting tired of it.
    
At this point, I am not sure how much of this is working.
The current released version (upstream) of this is 8.0.4
The choice is now to either upgrade or drop.
Migrating it to Python3 was a big effort.
Thanks to @jefferyto for the work here. I know I didn't make things easy,
because I wanted to drop it (instead of migrating to Python3).
    
Trying to push some changes upstream is met with silence.
One example:
    https://github.com/haiwen/seahub/pull/3607
But even looking at the commit history, it seems that most changes aren't
from outside contributors.
    

commodo:
@cotequeiroz 
pinging you, since you had some interest in Seafile as well;

so, let's start the discussion here with removing this beast of a package;
the PR for the abandoned move is:  https://github.com/openwrt/packages-abandoned/pull/17

jefferyto:
Apologies for not responding to the earlier discussions - I haven't had time in the past few months for any open source activity.

I agree that maintaining the Seafile packages is a pain, and I haven't done a good job in updating it to the current version. However, AFAIK the current in-tree version (7.1.5) still works and I am reluctant to remove something that still works (unless Seafile clients are locked to the latest version of the server component, I don't know if this is the case).

Also, I'm not certain there are no users interested in using Seafile. The version in 19.07 may have compiled but I am 99% sure it doesn't actually run (or requires a lot of effort to get it running). 21.02 will be the first public release with a working Seafile and the first chance in a while for general users to really use it. (I was hoping to update Seafile to 8.0 in time for 21.02, seeing how 21.02-rc1 was tagged a few days ago I'm sure I have a few more months left to do so hahaha)

Alex, it is **completely okay** if you want to remove yourself as a maintainer - people change, circumstances change, there is nothing wrong with this. I will still be a maintainer (until such time when I remove myself :joy:) and @cotequeiroz may join as well.

There are packages in the repo that have no maintainer. For me, I think it is okay to keep an unmaintained package in the repo if it continues to work and/or requires only little maintenance by others. If it is broken and requires major work to get it working, then I am for removing it / moving to the abandoned repo. I don't think Seafile has reached this latter stage yet.

So I am not convinced that Seafile needs to be removed at this time, but I fully support a PR to remove @commodo as a maintainer. Thank you Alex for all your effort with this troublesome group of packages :bow: :pray:

neheb:
Since 21.02 has docker, I wonder if it's better to just use a seafile container.

jefferyto:
> Since 21.02 has docker, I wonder if it's better to just use a seafile container.

This could be a solution if we really can't get Seafile working, though Docker isn't available for all targets.

commodo:
> So I am not convinced that Seafile needs to be removed at this time, but I fully support a PR to remove @commodo as a maintainer. Thank you Alex for all your effort with this troublesome group of packages bow pray

Works for me.
I will close this and remove myself from the Seafile stuff.

Thanks :)
Alex

PR #15542, opened 2021-04-29
bind: bump to 9.17.12

nmeyerhans:
Fixes the following security issues:

* CVE-2021-25215 - named crashed when a DNAME record placed in the ANSWER
                   section during DNAME chasing turned out to be the final
                   answer to a client query.
* CVE-2021-25214 - Insufficient IXFR checks could result in named serving a
                   zone without an SOA record at the apex, leading to a
                   RUNTIME_CHECK assertion failure when the zone was
                   subsequently refreshed. This has been fixed by adding an
                   owner name check for all SOA records which are included
                   in a zone transfer.

Signed-off-by: Noah Meyerhans <frodo@morgul.net>

Maintainer: me 
Compile tested: master, x86_64, mipsel (qemu malta). tested with optional features enabled and disabled
Run tested: master, x86_64, mipsel (qemu malta). tested with optional features enabled and disabled


BKPepe (approved):
LGTM

PR #15553, opened 2021-04-30
samplicator: add samplicator

PolynomialDivision:
Samplicator receives UDP datagrams on a given port and resends those datagrams to a specified set of receivers.

Use Cases:
- replicate Flow Samples to multiple receivers
- use with conntrackd to synchronize via unicast to multiple targets

Maintainer: me 
Compile tested: trunk (ipq40xx)
Run tested: Mips WDR4300

### DIFF
--- a/net/samplicator/Makefile
+++ b/net/samplicator/Makefile
@@ -1,0 +1,42 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=samplicator
+PKG_VERSION:=1.3.6
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/sleinen/samplicator/releases/download/v$(PKG_VERSION)
+PKG_HASH:=3c4358b4b0992a77251f2b9e2221d4ae945781160732c73504eb126e69d72d40
+
+PKG_MAINTAINER:=Nick Hainke <vincent@systemli.org>
+PKG_LICENSE:=GPL-2.0-only
+PKG_LICENSE_FILES:=COPYING
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/samplicator
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=UDP Samplicator
+  URL:=https://github.com/sleinen/samplicator
+endef
+
+define Package/samplicator/description
+  Send copies of (UDP) datagrams to multiple receivers,
+  with optional sampling and spoofing.
+endef
+
+define Package/samplicator/conffiles
+/etc/samplicator.conf
+endef
+
+define Package/samplicator/install
+	$(INSTALL_DIR) $(1)/usr/sbin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/samplicate $(1)/usr/sbin/
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_BIN) ./files/samplicator.init $(1)/etc/init.d/samplicator
+	$(INSTALL_DIR) $(1)/etc
+	$(INSTALL_DATA) ./files/samplicator.conf $(1)/etc/
+endef
+
+$(eval $(call BuildPackage,samplicator))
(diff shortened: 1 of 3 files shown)


PolynomialDivision:
Sadly, the PID file stuff is not in the stable release.

net/samplicator/files/samplicator.init
```diff
@@ -0,0 +1,22 @@
...
+    procd_set_param command $SAMPLICATOR_BIN -f -c $SAMPLICATOR_CONF
+    procd_set_param file "$SAMPLICATOR_CONF"
+    procd_set_param stdout 1
+    procd_set_param stderr 1
+    procd_set_param respawn
+    procd_close_instance
+}
+
+reload_service() {
+    procd_send_signal bird
```
spolack, line 21: bird?

PolynomialDivision: Thanks! Fixed. Copied the relevant stuff from bird init file.

PolynomialDivision: I removed reloading completely.

net/samplicator/Makefile
```diff
@@ -0,0 +1,42 @@
...
+PKG_NAME:=samplicator
+PKG_VERSION:=1.3.6
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/sleinen/samplicator/releases/download/v$(PKG_VERSION)
+PKG_HASH:=3c4358b4b0992a77251f2b9e2221d4ae945781160732c73504eb126e69d72d40
+
+PKG_MAINTAINER:=Nick Hainke <vincent@systemli.org>
+PKG_LICENSE:=GPL-2.0
```
BKPepe, line 12: This is outdated [SPDX License Identifier](https://spdx.org/licenses/).

net/samplicator/Makefile
```diff
@@ -0,0 +1,42 @@
...
+PKG_LICENSE_FILES:=COPYING
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/samplicator
+	SECTION:=net
+	CATEGORY:=Network
+	TITLE:=UDP Samplicator
+	URL:=https://github.com/sleinen/samplicator
```
BKPepe, line 21:
```suggestion
  SECTION:=net
  CATEGORY:=Network
  TITLE:=UDP Samplicator
  URL:=https://github.com/sleinen/samplicator
```

net/samplicator/Makefile
```diff
@@ -0,0 +1,42 @@
...
+endef
+
+define Package/samplicator/description
+	Send copies of (UDP) datagrams to multiple receivers,
+	with optional sampling and spoofing.
+endef
```
BKPepe, line 27:
```suggestion
define Package/samplicator/description
  Send copies of (UDP) datagrams to multiple receivers,
  with optional sampling and spoofing.
endef
```

net/samplicator/Makefile
PolynomialDivision: changed to GPL-2.0-only. Sry, I do always just searching for the identifier. Did not see that it is in the deprecated section. xD

PolynomialDivision:
BTW.: https://github.com/sleinen/samplicator/issues/73

PolynomialDivision:
@BKPepe Still something left? Otherwise I would like to merge.

BKPepe:
Why are you shipping an empty config file? And then the second question, why do we need to have it merged, why this package it useful/what is your usage? Can you also add test.sh file to have it run-tested?

PolynomialDivision:
> Why are you shipping an empty config file?

I can put there an example config.

> And then the second question, why do we need to have it merged, why this package it useful/what is your usage?

We want to use conntrackd to synchronize state-full firewall rules across multiple gateways. To my knowledge, conntracd only has the ability to do this via unicast to one target. That is why we want to use samplicator to samplicate those "control messages".

> Can you also add test.sh file to have it run-tested?

Never did this. I have to check how this works.

PolynomialDivision:
> Can you also add test.sh file to have it run-tested?

@BKPepe After looking at the test.sh files, it looks like everyone is just checking for the correct version. Sadly, samplicator has no cmd to check the version.

PolynomialDivision:
@BKPepe is it fine without test.sh?

PolynomialDivision:
Here is what we do:
https://github.com/Freifunk-Spalter/bbb-configs/commit/c615c35b8b5e12396140eccfb3c4cfdf0586a7a6

PR #15569, opened 2021-05-04
system-notification: initial checkin

feckert:
Maintainer: me
Compile tested: no
Run tested: yes

Description:
@stangri @aparcar @dibdot 
This is a proof of concept backend implementation for https://github.com/openwrt/luci/issues/5018

### DIFF
--- a/utils/system-notification/Makefile
+++ b/utils/system-notification/Makefile
@@ -1,0 +1,81 @@
+#
+# Copyright (C) 2021 TDT AG <development@tdt.de>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See https://www.gnu.org/licenses/gpl-2.0.txt for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=system-notification
+PKG_VERSION:=0.1
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=Florian Eckert <fe@dev.tdt.de>
+PKG_LICENSE:=GPL-2.0
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/system-notification
+  SECTION:=base
+  CATEGORY:=Base system
+  TITLE:=System notification
+endef
+
+define Package/system-notification/description
+  This package provides a ubus backend that ca be used to save system
+  notification. The notifications are not persistent. They are lost after
+  every reboot.
+
+  The following ubus methods are available:
+
+  method <add>:
+  The <add> command adds a notification to the ubus backend.
+  ubus call system.notification add '{"title":"<string>","description":"<description>"}'
+  In addition, the following json string could be optional set.
+  * url: <"url": "string"> Additional information are available at this extrenal URL.
+  * opkg: <"opkg": "string"> Package could be upgrade via opkg.
+
+  method <remove>:
+  The <remove> command removes a notification from the ubus backend.
+  ubus call system.notification remove '{"uuid": "<string>"}'
+
+  method <list>:
+  The <list> command dumps all staged notification from the ubus backend.
+  ubus call system.notification list
+
+  method <flush>:
+  The <flush> command removes all stage notification from the ubus backend.
+  ubus call system.notification flush
+endef
+
+define Build/Compile
+endef
+
+define Package/system-notification/postinst
+#!/bin/sh
+if [ -z "$${IPKG_INSTROOT}" ] && [ -x /etc/init.d/rpcd ]; then
+	/etc/init.d/rpcd restart
+fi
+exit 0
+endef
+
+define Package/system-notification/postrm
+#!/bin/sh
+if [ -z "$${IPKG_INSTROOT}" ] && [ -x /etc/init.d/rpcd ]; then
+	/etc/init.d/rpcd restart
+fi
+exit 0
+endef
+
+define Package/system-notification/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) ./files/usr/bin/system-notification \
+		$(1)/usr/bin/system-notification
+
+	$(INSTALL_DIR) $(1)/usr/libexec/rpcd
+	$(INSTALL_BIN) ./files/usr/libexec/rpcd/system.notification \
+		$(1)/usr/libexec/rpcd/
+endef
+
+$(eval $(call BuildPackage,system-notification))
--- a/utils/system-notification/files/usr/libexec/rpcd/system.notification
+++ b/utils/system-notification/files/usr/libexec/rpcd/system.notification
@@ -1,0 +1,220 @@
+#!/bin/sh
+#
+# Copyright (C) 2021 TDT AG <development@tdt.de>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See https://www.gnu.org/licenses/gpl-2.0.txt for more information.
+#
+
+. /usr/share/libubox/jshn.sh
+
+MESSAGE_DIR="/tmp/notification/messages.d"
+MESSAGE_MAX=20
+
+get_uuid() {
+	local file="$1"
+	local uuid
+
+	uuid="${file##*/}"
+
+	echo "${uuid%*.json}"
+}
+
+gen_uuid() {
+	cat /proc/sys/kernel/random/uuid
+}
+
+list() {
+	json_init
+
+	json_add_object "add"
+	json_add_string "title"
+	json_add_string "description"
+	json_add_string "url"
+	json_add_string "package"
+	json_close_object
+
+	json_add_object "remove"
+	json_add_string "uuid"
+	json_close_object
+
+	json_add_object "list"
+	json_close_object
+
+	json_add_object "flush"
+	json_close_object
+
+	json_dump
+}
+
+notification_add() {
+	local title="$1"
+	local description="$2"
+	local url="$3"
+	locla package="$4"
+
+	local timestamp uuid
+
+	local code=-1
+	local msg="Unable to add notification"
+
+	timestamp="$(date +%s)"
+
+	uuid="$(gen_uuid)"
+	if [ -z "$uuid" ]; then
+		json_init
+		json_add_int "code" "-2"
+		json_add_string "message" "Unable to get uuid"
+		json_dump
+		return
+	fi
+
+	json_init
+	json_add_string title "$title"
+	json_add_string description "$description"
+	json_add_int timestamp "$timestamp"
+	[ -z "$url" ] || json_add_string "url" "$url"
+	[ -z "$package" ] || json_add_string "package" "$package"
+	json_dump > "${MESSAGE_DIR}/${uuid}.json"
+
+	if [ -f "${MESSAGE_DIR}/${uuid}.json" ]; then
+		code=0
+		msg="Notification added"
+	fi
+
+	local count=0
+	for file in "${MESSAGE_DIR}"/*.json; do
+		count=$((count + 1))
+		if [ "$count" -gt "$MESSAGE_MAX" ]; then
+			rm "${MESSAGE_DIR}/${file}"
+		fi
+	done
+
+	json_init
+	json_add_int "code" "$code"
+	json_add_string "message" "$msg"
+	json_dump
+}
+
+notification_remove() {
+	local uuid="$1"
+
+	local code=-1
+	local msg="Notification not found"
+
+
+	if [ -f "${MESSAGE_DIR}/${uuid}.json" ]; then
+		rm "${MESSAGE_DIR}/${uuid}.json"
+		code=0
+		msg="Notication removed"
+	fi
+
+	json_init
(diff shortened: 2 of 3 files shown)


utils/system-notification/files/usr/libexec/rpcd/system.notification
```diff
@@ -0,0 +1,190 @@
...
+
+	timestamp="$(date +%s)"
+	uptime="$(get_uptime)"
+
+	json_init
+	json_add_string service "$service"
+	json_add_string title "$title"
+	json_add_string message "$message"
+	json_add_int timestamp "$timestamp"
+	json_dump > "${MESSAGE_DIR}/${uptime}.json"
```
karlp, line 69: what's the purpose of the uptime here? is it just to get a "unique" filename to use?  what was wrong with the timestamp you already have? your uptime call throws away fractional seconds anyway, so it seems to be just same same, but different.  You  and you can use $$ to add the current process pid as well, or even just use mktemp?  it seems to be available in the default busybox configs.

utils/system-notification/files/usr/libexec/rpcd/system.notification
```diff
@@ -0,0 +1,190 @@
...
+
+notification_remove() {
+	local index="$1"
+	local code=-1
+	local msg="Notification not found"
+
+	local id
+
+	for entry in "$MESSAGE_DIR"/*.json; do
+		id="$(get_id "$entry")"
```
karlp, line 90: is there anythign wrong with just using "basename" ?

utils/system-notification/files/usr/libexec/rpcd/system.notification
```diff
@@ -0,0 +1,190 @@
...
+			json_close_object
+		done
+	fi
+	json_close_array
+	json_dump
+}
+
+notification_flush() {
+	local code=0
+	local msg="Notification flushed"
```
karlp, line 135: you're flushing _all_ notifications here right?

utils/system-notification/Makefile
```diff
@@ -0,0 +1,61 @@
...
+include $(INCLUDE_DIR)/package.mk
+
+define Package/system-notification
+  SECTION:=base
+  CATEGORY:=Base system
+  TITLE:=System notification
+endef
+
+define Package/system-notification/description
+  This package provid a ubus backend with which system notification
```
karlp, line 26: typo

utils/system-notification/Makefile
```diff
@@ -0,0 +1,61 @@
...
+  can be stored.
+
+  The following ubus methods are available:
+  * add
+  * remove
+  * list
+  * flush
```
karlp, line 33: there's no url listed, is this the only documentation?

utils/system-notification/Makefile
feckert: thanks will fix this

utils/system-notification/Makefile
feckert: So far there is no other documentation. I think I will expand the information here.

utils/system-notification/files/usr/libexec/rpcd/system.notification
feckert: > what's the purpose of the uptime here? is it just to get a "unique" filename to use? what was wrong with the timestamp you already have? your uptime call throws away fractional seconds anyway, so it seems to be just same same, but different.

I have a target without RTC so when there is an ntp connection the time is set during the runtime.
If I use the uptime it continues to be counted forward and does not jump during an ntp synchronization.

But I think your suggestion with the **mktemp** is not bad, I think I will rewrite this :+1:

utils/system-notification/files/usr/libexec/rpcd/system.notification
feckert: thank you I will change this

utils/system-notification/files/usr/libexec/rpcd/system.notification
feckert: yes?

utils/system-notification/files/usr/libexec/rpcd/system.notification
feckert: @karlp I just realised, or rather I remembered, that mktemp generates random alphanumeric file names
But I wanted only use a number for the id.
It should be an uint and not a string.
I guess mktemp doesn't work for this case.

I would use the timestamp this has the advantage that I could write more then one message per second and it is a unique id that increments.

utils/system-notification/files/usr/libexec/rpcd/system.notification
feckert: @karlp I have now taken a closer look.
I need not only the name of the file here but also the ID.
I have to return this via ubus so user can use the ID to delete a message with the delete command

utils/system-notification/files/usr/libexec/rpcd/system.notification
karlp: the msg could perhaps be clarified then...

utils/system-notification/files/usr/libexec/rpcd/system.notification
karlp: it doesn't matter if it jumps, you're just this is a fairly poor unique id :)  you should perhaps just use a uuid or something anyway.  perhaps, cat /proc/sys/kernel/random/uuid and never have to consider this anymore.... ?

karlp: (and why can't it be a string, there's no significance to the number, and I don't really buy that this is a json optimization for  ints vs strings, the "benefit" is far too small for something that is intended as notifications.  you're not generating thousands of these per minute, let alone per second....

feckert: Thanks for the good advice I have now selected the id of the message as the uuid string.

utils/system-notification/Makefile
```diff
@@ -0,0 +1,74 @@
...
+endef
+
+define Package/system-notification/description
+  This package provides a ubus backend that ca be used to save system
+  notification. The notifications are not persistent. They are lost after
+  every reboot.
+
+  The following ubus methods are available:
+
+  methode <add>:
```
karlp, line 32: this should just be "method" not "methode".

looks nice otherwise!

utils/system-notification/files/usr/libexec/rpcd/system.notification
```diff
@@ -0,0 +1,199 @@
...
+		msg="Notification added"
+	fi
+
+	json_init
+	json_add_int "code" "$code"
+	json_add_string "message" "$msg"
+	json_dump
+}
+
+notification_remove() {
```
aparcar, line 90: Isn't it easier to check if a file exists via `[ -f "$MESSAGE_DIR/$delete_uuid" ]` instead of iterating over all?

utils/system-notification/files/usr/libexec/rpcd/system.notification
```diff
@@ -0,0 +1,199 @@
...
+			json_close_object
+		done
+	fi
+	json_close_array
+	json_dump
+}
+
+notification_flush() {
+	local code=0
+	local msg="All notification flushed"
```
karlp, line 144: this still should be plural.

utils/system-notification/files/usr/libexec/rpcd/system.notification
feckert: Good point.
That is definitely smarter.

feckert:
I have now added a change in the list method on my lost push force.
The messages are now sorted by [timestamp](https://github.com/openwrt/packages/pull/15569/files#diff-f3c277a91b6aa2ec6964861b40fcfb2b5a6632a0c566fea41c385132a7c62c2dR115).

If I don't do this, I have to sort them in the LuCI, as I have just discovered.
I am also in the process of writing a luci-mod-notification.

Thanks for reviewing this :+1:

aparcar:
What about a `prio` field which is a numeric value which allows to alter the order of listing?

feckert:
> What about a `prio` field which is a numeric value which allows to alter the order of listing?

@aparcar I can add to that.

Independently of this, I have added the possibility to add a message ID.
This should be unique and should only be used once for this message kind of message type.
With this ID, the LuCI can then handle these messages differently, if this is known by LuCI.

In the associated [LuCI pullrequest](https://github.com/openwrt/luci/pull/5041/commits/3f8835e1929e288ec55c07e09e6d29d10a910a35) I have added a commit that clarifies this.

utils/system-notification/files/usr/libexec/rpcd/system.notification
```diff
@@ -0,0 +1,201 @@
...
+
+	echo "${uuid%*.json}"
+}
+
+gen_uuid() {
+	local uuid
+
+	uuid="$(cat /proc/sys/kernel/random/uuid)"
+	echo "$uuid"
+}
```
aparcar, line 27:
```suggestion
gen_uuid() {
	cat /proc/sys/kernel/random/uuid
}
```

aparcar:
@feckert could you apply the change suggestions from https://github.com/openwrt/luci/pull/5041 so we have a more generic framework and then move on adding this?

feckert:
@aparcar So I have now extended this and incorporated the suggestions for the actions array from @jow-.
I have also added a script that does output the notifications on the console.
The only thing that is still not clear to me is how we do this with the translation in the LuCI.

aparcar:
Thank you, I'll try to test and comment on this within the next two weeks

Ansuel:
Any news on this?

Ansuel:
@feckert @aparcar It's sad that this stalled so much since this was a very good feature and I could see may usecase for this once it does get some traction. Example we currently have some popup in luci when we don't have a a password set or we are running in initramfs... so we already have a form of notification, just not implemented globally.

And I can see may other case that should be used to hint the user that the router still has to be configured. (reg domain not set, using vulnerable option... even alerting the user that a security update is present or even warn the user of regression or configuration change) (all of these thing would be json to install in rootfs with a system similar to uci-defaults) (even some kind of integration with some remote notification system and suggest the user to update)

Even Ubuntu have a similar system recently that add message on first login to a system (the thing is configurable and can be disabled)


Sooo did we really lose the opportunity to implement this good feature for the only reason of translation?
Something totally optional? (lets be real for a minute, we expect advanced user installing openwrt, we can assume for these message to not be translated at least from the console view... For luci, we can think of implement some kind of dedicated .po and the ui will check if the translation is available)

Also I feel this thing should be handled in core and not as an optional package.

feckert:
> So did we really lose the opportunity to implement this good feature for the only reason of translation?

From my point of view, this is the case, because I do not know how to treat the translation generically in LuCI.
On the command we do not need a translation. 

> Something totally optional? (lets be real for a minute, we expect advanced user installing openwrt, we can assume for these message to not be translated at least from the console view... For luci, we can think of implement some kind of dedicated .po and the ui will check if the translation is available)

The problem I have here is that we have to maintain the translations all the time and it must match with the string of the message. I think it is better to work with IDs. Then the translation can be maintained centrally in the LuCI and if the ID does not exist then it is displayed by the command line without translation. But this is going too far now 

> Also I feel this thing should be handled in core and not as an optional package.

Yes, from my point of view, this should also be part of the core system. However, I have only implemented it as a proof of concept in the shell.

Ansuel:
> The problem I have here is that we have to maintain the translations all the time and it must match with the string of the message. I think it is better to work with IDs. Then the translation can be maintained centrally in the LuCI and if the ID does not exist then it is displayed by the command line without translation. But this is going too far now

Do we expect messages to be that different? We can try to make them generic enough. For the core message we have control of them (and we can totally ignore messaged from luci modules as they will have their own po file with translations)

Having an ID is just another way to handle the main problem where luci po needs to be synced with master... But this is doable. In luci there is already a script to sync po with all the luci modules.
We can reimplement something like that and then connect this to weblate to make translation easier for the community.
(even a CI action to sync and create a pr when it does detect changes)

But still the problem is still there ID or matching the string, it's always something to sync so I think this is why some pointed out how an ID was a bit redundant.

The idea of the ID was to update a message if we ever need it? (I can only see this as an improvement than matching the entire string)
But even this can be solved by regenerating the message po from scratch and remove extra entry

feckert:
@Ansuel So what do you suggest to move forward on this issue? Should I open a pull request against the Openwrt core?

Ansuel:
@feckert IMHO yep let move to core and maybe write something to the mailing list?

feckert:
@Ansuel as suggest I send this to [OpenWrt-Mailing-List](https://lists.openwrt.org/pipermail/openwrt-devel/2023-July/041287.html).

Ansuel:
I would also post it on GitHub to try to receive feedback by both source

Il Mer 19 Lug 2023, 13:22 Florian Eckert ***@***.***> ha
scritto:

systemcrash:
Not many comments on the mailing list. Is proceeding here a better way forward?

PR #15583, opened 2021-05-09
apache-utils: decouple from apache and add all missing tools.

jmarcet:
Maintainer: Thomas Heil <heil@terminal-consulting.de>
Compile tested: master x86_64
Run tested: master x86_64

Description:
This package is not really dependant on apache and its tools are
useful even without apache, for instance htpasswd can generate
passwords which can then be read by nginx's as auth_basic_user_file.


feckert:
The build bot complains that by removing the dependency from **apache**, the runtime dependencies are not fulfilled. To fix this, you need to add the following libraries.

Buildlog.
```
021-05-09T22:12:43.5067178Z libcrypto.so.1.1
2021-05-09T22:12:43.5067612Z libssl.so.1.1
```

feckert:
I did not express myself precisely.
You have to look in the system with a grep which install package provides these libraries.
```
cd openwrt/packages
grep -R libcrypto
----
libs/openssl/Makefile:  $(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/lib/libcrypto.so.* $(1)/usr/lib/
```
```
cd openwrt/packages
grep -R libssl
----
libs/openssl/Makefile:  $(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/lib/libssl.so.* $(1)/usr/lib/
```

To I would say the runtime dependency ist `libopenssl`.

GeorgeSapkin:
If this is still relevant, it should have a `PKG_RELEASE` bump. And the commit message needs to follow the 50/75 rule as per the [submissions guidelines](https://openwrt.org/submitting-patches#submission_guidelines).

GeorgeSapkin:
No updates in over four years and not following the [submission guidelines](https://openwrt.org/submitting-patches#submission_guidelines). Closing as stale. Feel free to reopen if anything changes.

PR #15586, opened 2021-05-10
python3: bump version to 3.9.5

commodo:
Maintainer: me, @jefferyto 
Compile tested: x86 https://github.com/openwrt/openwrt/commit/97e820c6d61d174da1091a692546e979a5a45e42
Run tested: x86 https://github.com/openwrt/openwrt/commit/97e820c6d61d174da1091a692546e979a5a45e42

Refreshed Python patches.
Updated pip & setuptools version.

For pip, patch '001-pep517-pyc-fix.patch' was reworked.

Also, the current version of the bundled pip (21.1.1) no longer supports
Python2, so the 'py2.py3' suffix gets replaced with just py3.

For setuptools, there is no longer a script/module:
   https://github.com/pypa/setuptools/pull/2544


commodo:
hmm, i'll need to run from a clean checkout

neheb:
Don't forget make package/x/refresh

commodo:
> Don't forget make package/x/refresh

i did that; turns out that the `pip` suffix is no longer `pip-$(PYTHON3_PIP_VERSION)-py2.py3-none-any.whl` ; 
the `py2` part went away;

commodo:
Changelog v1 -> v2:
* fixed pip installation ; removed `py2` from suffix of whl file
* fix setuptools installation; easy_install was removed; we need to also remove it from our installation

PR #15593, opened 2021-05-10
Update snort3 and libdaq3 packages

flyn-org:
Maintainer: me
Compile tested: x86_64 master

Description:

### DIFF
--- a/net/snort3/Makefile
+++ b/net/snort3/Makefile
@@ -6,13 +6,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=snort3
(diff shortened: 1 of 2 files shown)


flyn-org:
Note the use of `/../` in `-DFLEX_INCLUDES:PATH=$(STAGING_DIR)/../host/include`. I would appreciate any suggestions on the right way to do this, as my solution feels kludgy. Without something like this, the build fails because it cannot find `FlexLexer.h`.

flyn-org:
@neheb, `STAGING_DIR_HOST` is what I was looking for. I just pushed again. Thank you!

PR #15613, opened 2021-05-12
rsync: Add rrsync script

mattreeve:
Signed-off-by: Matt Reeve <matt@mreeve.com>

Maintainer: Maxim Storchak <m.storchak@gmail.com>
Compile tested: Qualcomm Atheros IPQ806X, NETGEAR Nighthawk X4S R7800
Run tested: Qualcomm Atheros IPQ806X, NETGEAR Nighthawk X4S R7800)

mkdir /tmp/xxx
echo hello >> /tmp/xxx/abc

Add to ~/.ssh/authorized_keys:

command="/usr/bin/rrsync -ro /tmp/xxx" ssh-rsa public key here

From another machine:

rsync -av root@router:/ /tmp/yyy

File "abc" transfers successfully.

touch /tmp/yyy/def

rsync -av /tmp/yyy root@router:/

Transfer fails as expected (-ro in authorized_keys)

On router, change ~/.ssh/authorized_keys:

command="/usr/bin/rrsync /tmp/xxx" ssh-rsa public key here

On other machine:

rsync -av /tmp/yyy root@router:/

File def now transfers successfully.

Description:

Rrsync allows selective access to subdirectories in either read-only, write-only or read-write, depending on settings in authorized_keys. This allows for safe, restrictive access. It's particularly useful for automated backup purposes.

### DIFF
--- a/net/rsync/Makefile
+++ b/net/rsync/Makefile
@@ -69,6 +69,15 @@ define Package/rsyncd
   URL:=https://rsync.samba.org/
 endef
 
+define Package/rrsync
+  SECTION:=net
+  CATEGORY:=Network
+  SUBMENU:=File Transfer
+  TITLE:=Restricted rsync script
+  DEPENDS:=+rsync +perlbase-file +perl @(PACKAGE_openssh-server||PACKAGE_openssh-server-pam)
+  URL:=https://www.samba.org/ftp/unpacked/rsync/support/rrsync
+endef
+
 define Package/rsync/description
  rsync is a program that allows files to be copied to and from remote machines
  in much the same way as rcp. It has many more options than rcp, and uses the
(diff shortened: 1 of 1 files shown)


net/rsync/Makefile
```diff
@@ -69,6 +69,15 @@ define Package/rsyncd
   URL:=https://rsync.samba.org/
 endef
 
+define Package/rrsync
+  SECTION:=net
+  CATEGORY:=Network
+  SUBMENU:=File Transfer
+  TITLE:=Restricted rsync script
+  DEPENDS:=+rsync +perlbase-file +perl
```
mstorchak, line 77: As to me, it should also depend on openssh-server since dropbear doesn't support ~/.authorized_keys, but I don't have a strong opinion on this.

mattreeve: That makes sense, though it could depend on either openssh-server or openssh-server-pam. So should it be this?

```suggestion
  DEPENDS:=+rsync +perlbase-file +perl @(PACKAGE_openssh-server||PACKAGE_openssh-server-pam)
```

mstorchak: Depending on either one makes sense, but I'm not an expert in dependencies like this, and my build host is down atm so I can't check it myself.
If this expression does the job, I'm good with it.

mattreeve: It works - if I don't select either openssh-server or openssh-server-pam then rrsync does not appear in menuconfig. If I select one or other, or both, then it appears.

mstorchak:
LGTM

PolynomialDivision:
Could you maybe add a more meaningful commit message describing this script?

mattreeve:
> Could you maybe add a more meaningful commit message describing this script?

Done

net/rsync/Makefile
PolynomialDivision: I was also a bit unsure about this line. However, similar things are done i pppossh and prometheus-node-exporter-lua.

PolynomialDivision:
Since @mstorchak is fine, I will merge. Thanks for your contribution.

PR #15615, opened 2021-05-13
libgpg-error: patch to fix cross-compile

flyn-org:
Paul Blazejowski discovered that the recent update to this package broke
cross-compiling for some OpenWrt platforms. This backports commit 33593864
(March 26, 2021) from the upstream libgpg-error project to fix this 
problem.

Maintainer: me
Compile tested: x86_64 master

Description:


egorenar:
Thanks. This fixed my problem on Void Linux x86 (host) & ARM cortex-a15(target).

PR #15637, opened 2021-05-16
tinc: Add possibility to populate more config variables via uci

ErwanMAS:
Signed-off-by: Erwan MAS <erwan@mas.nom.fr>

Maintainer: @ErwanMAS 
Compile tested: sdk-ramips-mt7621 , snapshot  
Run tested: ramips/ mt7621 / MT1300  snapshot

Description:
  - add Name,Ed25519PublicKey parameter to UCI configuration
  - alter logic to allow full configuration of a Host in UCI 

Can you grant access to openwrt/packages repository ?

### DIFF
--- a/net/tinc/Makefile
+++ b/net/tinc/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tinc
 PKG_VERSION:=1.1-git
-PKG_RELEASE=$(PKG_SOURCE_VERSION)-1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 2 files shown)


net/tinc/Makefile
```diff
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tinc
 PKG_VERSION:=1.1-git
-PKG_RELEASE=$(PKG_SOURCE_VERSION)-1
+PKG_RELEASE=$(PKG_SOURCE_VERSION)-2
```
PolynomialDivision, line 12: Why? Never saw something like this.

ErwanMAS: you are right , this now fixed

PR #15658, opened 2021-05-19
[21.02] rsync: Add rrsync script

mattreeve:
Maintainer: Maxim Storchak <m.storchak@gmail.com>
Compile tested: Qualcomm Atheros IPQ806X, NETGEAR Nighthawk X4S R7800
Run tested: Qualcomm Atheros IPQ806X, NETGEAR Nighthawk X4S R7800

Description:

Rrsync is a perl script that is supplied as an extra with the rsync program.
It must be used in conjunction with openssh-server or openssh-server-pam
as it requires ~/.ssh/authorized_keys which is not supported by dropbear.

Rrsync allows selective access to subdirectories in either read-only, write-only or read-write mode,
depending on settings in authorized_keys. This allows for safer, restrictive access.
It's particularly useful for automated backup purposes.

An example usage would be this entry:

command="/usr/bin/rrsync -ro /home" <public key here>

This would allow a system connecting with this public key to be able to rsync FROM the
/home directory tree only. It could not write to this directory, nor read from any other directory.

Signed-off-by: Matt Reeve <matt@mreeve.com>
(cherry picked from commit 081229aa09e28435d3d8802ce053a5d4eee8978a)

### DIFF
--- a/net/rsync/Makefile
+++ b/net/rsync/Makefile
@@ -69,6 +69,15 @@ define Package/rsyncd
   URL:=https://rsync.samba.org/
 endef
 
+define Package/rrsync
+  SECTION:=net
+  CATEGORY:=Network
+  SUBMENU:=File Transfer
+  TITLE:=Restricted rsync script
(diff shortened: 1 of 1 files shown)


mstorchak:
LGTM

Issue #15653, opened 2021-05-18
nginx-util: build broken


anomeome:
Maintainer: @Ansuel @peter-stadler

Description: Last build on 21/05/14 worked fine, pull latest code today broken, cleaned out likely culprits to no avail:

```
/home/kc/wrtpac/source/staging_dir/host/bin/ccache /home/kc/wrtpac/source/staging_dir/toolchain-arm_cortex-a9+vfpv3_gcc-10.3.0_musl_eabi/bin/arm-openwrt-linux-muslgnueabi-g++ -DVERSION=1.5  -O2 -pipe -fno-caller-saves -fno-plt -fhonour-copts -Wno-error=unused-but-set-variable -Wno-error=unused-result -mfloat-abi=hard -ffile-prefix-map=/home/kc/wrtpac/source/build_dir/target-arm_cortex-a9+vfpv3_musl_eabi/nginx-util-1.5=nginx-util-1.5 -Wformat -Werror=format-security -DPIC -fpic -fstack-protector-strong -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro -DNDEBUG   -Os -Wall -Werror -Wextra -g3 -Wno-unused-parameter -Wmissing-declarations -Wshadow -std=gnu++17 -MD -MT CMakeFiles/nginx-ssl-util.dir/nginx-util.cpp.o -MF CMakeFiles/nginx-ssl-util.dir/nginx-util.cpp.o.d -o CMakeFiles/nginx-ssl-util.dir/nginx-util.cpp.o -c nginx-util.cpp
In file included from nginx-util.cpp:4:
nginx-ssl-util.hpp: In lambda function:
nginx-ssl-util.hpp:169:40: error: attributes at the beginning of statement are ignored [-Werror=attributes]
  169 |                 case '^': ret += '\\'; [[fallthrough]];
      |                                        ^~~~~~~~~~~~~~~
nginx-ssl-util.hpp:170:27: error: attributes at the beginning of statement are ignored [-Werror=attributes]
  170 |                 case '_': [[fallthrough]];
      |                           ^~~~~~~~~~~~~~~
nginx-ssl-util.hpp: In lambda function:
nginx-ssl-util.hpp:169:31: error: this statement may fall through [-Werror=implicit-fallthrough=]
  169 |                 case '^': ret += '\\'; [[fallthrough]];
      |                           ~~~~^~~~~~~
nginx-ssl-util.hpp:170:17: note: here
[...]

PR #15661, opened 2021-05-19
nginx-util: do not use fallthrough attribute

peter-stadler:
Maintainer: me
Run tested: x86_64, x86_64 qemu, master, run tests

Description: fix compile warning/error (issue #15653)

### DIFF
--- a/net/nginx-util/Makefile
+++ b/net/nginx-util/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nginx-util
-PKG_VERSION:=1.5
+PKG_VERSION:=1.6
 PKG_RELEASE:=1
 PKG_MAINTAINER:=Peter Stadler <peter.stadler@student.uibk.ac.at>
 
--- a/net/nginx-util/src/nginx-ssl-util.hpp
+++ b/net/nginx-util/src/nginx-ssl-util.hpp
@@ -166,9 +166,14 @@ static constexpr auto _escape = _Line{
         std::string ret{};
         for (char c : strptr) {
             switch (c) {
-                case '^': ret += '\\'; [[fallthrough]];
-                case '_': [[fallthrough]];
-                case '-': ret += c; break;
+                case '^':
+                    ret += '\\';
+                    ret += c;
+                    break;
+                case '_':
+                case '-':
+                    ret += c;
+                    break;
                 default:
                     if ((isalpha(c) != 0) || (isdigit(c) != 0)) {
                         ret += c;


ptpt52:
looks like a wrong fix

peter-stadler:
Do you know a better method? Else: It should work, duplicating the code instead of falling through ...

ptpt52:
Please try this:
````
diff --git a/net/nginx-util/src/nginx-ssl-util.hpp b/net/nginx-util/src/nginx-ssl-util.hpp
index 4f12925db..fac8a3030 100644
--- a/net/nginx-util/src/nginx-ssl-util.hpp
+++ b/net/nginx-util/src/nginx-ssl-util.hpp
@@ -166,8 +166,10 @@ static constexpr auto _escape = _Line{
         std::string ret{};
         for (char c : strptr) {
             switch (c) {
-                case '^': ret += '\\'; [[fallthrough]];
-                case '_': [[fallthrough]];
+                case '^': ret += '\\';
+                         [[fallthrough]];
+                case '_':
+                         [[fallthrough]];
                 case '-': ret += c; break;
                 default:
                     if ((isalpha(c) != 0) || (isdigit(c) != 0)) {
````

**Comment by ptpt52** [comment]
try this:
````
diff --git a/net/nginx-util/src/nginx-ssl-util.hpp b/net/nginx-util/src/nginx-ssl-util.hpp
index 4f12925db..b52346b7f 100644
--- a/net/nginx-util/src/nginx-ssl-util.hpp
+++ b/net/nginx-util/src/nginx-ssl-util.hpp
@@ -166,8 +166,10 @@ static constexpr auto _escape = _Line{
         std::string ret{};
         for (char c : strptr) {
             switch (c) {
-                case '^': ret += '\\'; [[fallthrough]];
-                case '_': [[fallthrough]];
+                case '^': ret += '\\';
+                         /* fallthrough */
+                case '_':
+                         /* fallthrough */
                 case '-': ret += c; break;
                 default:
                     if ((isalpha(c) != 0) || (isdigit(c) != 0)) {
````

do not use ``[[fallthrough]]``

**Comment by ptpt52** [comment]
or more simple:
```
diff --git a/net/nginx-util/src/nginx-ssl-util.hpp b/net/nginx-util/src/nginx-ssl-util.hpp
index 4f12925db..4484b3550 100644
--- a/net/nginx-util/src/nginx-ssl-util.hpp
+++ b/net/nginx-util/src/nginx-ssl-util.hpp
@@ -166,8 +166,8 @@ static constexpr auto _escape = _Line{
         std::string ret{};
         for (char c : strptr) {
             switch (c) {
-                case '^': ret += '\\'; [[fallthrough]];
-                case '_': [[fallthrough]];
+                case '^': ret += '\\'; /* fallthrough */
+                case '_': /* fallthrough */
                 case '-': ret += c; break;
                 default:
                     if ((isalpha(c) != 0) || (isdigit(c) != 0)) {
```

**Comment by peter-stadler** [comment]
I changed it, thank you :-)
Do you happen to know why the `[[fallthrough]];` attribute is not working anymore?

**Comment by ptpt52** [comment]
it is likely
```
fallthrough;
```
work?

https://gcc.gnu.org/onlinedocs/gcc/Statement-Attributes.html

https://stackoverflow.com/questions/45349079/how-to-use-attribute-fallthrough-correctly-in-gcc

peter-stadler:
I thought about the [c++17 standard attribute](https://en.cppreference.com/w/cpp/language/attributes/fallthrough) not the legacy one, but it is not that important ...

peter-stadler:
I am not sure if the comment `/* fallthrough */` would work with gcc 10, so I copy the `ret += c; break;` instead of falling through.

(The command `fallthrough;` would work, too, but AFAIK it is not standard c++ and clang-tidy is not recognizing it. So I hesitate using it here.)

PR #15667, opened 2021-05-20
micrond: store cronjobs in /usr/share/micron.d

SvenRoederer:
Maintainer: @NeoRaider @feckert @adschm 

Description:
Use the /usr/share/micron.d folder to store the cronjob files. The /usr/lib
folder is commonly used for libraries and internal binaries only. The job
file do not belong to this category.

### DIFF
--- a/utils/micrond/Makefile
+++ b/utils/micrond/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=micrond
 PKG_VERSION:=1
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 PKG_LICENSE:=BSD-2-clause
 
 include $(INCLUDE_DIR)/package.mk
--- a/utils/micrond/files/etc/init.d/micrond
+++ b/utils/micrond/files/etc/init.d/micrond
@@ -3,7 +3,7 @@
 START=50
 USE_PROCD=1
 
-CRONDIR=/usr/lib/micron.d
+CRONDIR=/usr/share/micron.d
 
 start_service() {
 	procd_open_instance


feckert:
We have [discussed]( https://github.com/openwrt/packages/pull/13345) already if it would be better to put the micron.d files under `/etc/cron.d`. Have a look at this P/R also.

Other cron service implementation use this directory.
It would also nice if we could extentd the crontab of the busybox, so that we do not have two cron services.
One for the users and the other for the system.

Due to lack of time this has not been done yet.

SvenRoederer:
So when the mid-term plan of #13345 is to make busybox also honor ```/etc/cron.d``` it might a a short term solution to have micrond using this folder already. This might end up in using 2 cron-services until busybox is ready, but this is what I'm doing now anyway.

adschm:
I have no specific opinion on the subject, but I'm aware that changing stuff like this creates pain for all the people having to adapt downstream.

So, please make sure that
1. the new directory is _substantially better_ from an _objective_ point of view, and not just somebody's taste or an arbitrary choice
2. we don't change it twice (with respect to the /etc/cron.d comment)

SvenRoederer:
Sure , I think /etc/cron.d is a well accepted location that we can already start to use.

feckert:
I am torn about which is better now.
The note from @adschm is valid.
If we update the crontab from the busybox in the near future (hopefully), then we should leave the micrond cronjobs in its directory and not change it. Because we do not need the `micrond` anymore.

If the busybox has been extended, then the services that need a cronjob can be changed. I think it is really more purposeful if we concentrate on extending the crontab of the busybox, so that it also uses the /etc/cron.d directory.

SvenRoederer:
This all depend on the "if the cron.d feature will be implemented quickly, soon, this year". So it was the idea to use micrond for this as interim solution till someone finds the resources to extend busybox. when this feature is there, that might be the time, that micrond can be removed completely.
Its simply, that packages can use cron.d files already (via micrond) but work with the later busybox-cron.d out of the box.

feckert:
@SvenRoederer @adschm The only thing that comes to my mind that I would like to put up for discussion with you is:

Since openwrt is manly for embedded system, so I don't really want to write to the flash every time I change the system cron task files. I can only say that I have been using micrond for my system tasks for some time and I also have tasks that the user can change and are therefore saved in a uci file. So I do not need a service for task which are fine with the minute intervall

To make the handling easy, I write the cron file for micrond during bootup and then tell the mircond to read in the new directory.
What is your opinion on this? If something is changed I am removing the related cron file and recreate it. So I do not have to parse.

GeorgeSapkin:
No updates in over four years and PR has merge conflicts. Closing as stale. Feel free to reopen if anything changes.

PR #15669, opened 2021-05-20
openvpn: added tun_ipv6 config option

lweller:
Signed-off-by: Lucien Weller <lucien@wellernet.ch>

Maintainer: @mkrkn 
Compile tested: Netgear WNDR3700v4, 19.07.07
Run tested: Netgear WNDR3700v4, 19.07.07

Description: Added config option tun_ipv6 required for advanced IPv6 routing setup's (see https://community.openvpn.net/openvpn/wiki/IPv6)

### DIFF
--- a/net/openvpn/files/openvpn.options
+++ b/net/openvpn/files/openvpn.options
@@ -190,6 +190,7 @@ test_crypto
 tls_client
 tls_exit
 tls_server
+tun_ipv6
 up_delay
 up_restart
 username_as_common_name


mkrkn:
Can you provide a specific example where you use this option in a configuration and removing it causes a change in behavior? This option is [deprecated *and* removed](https://community.openvpn.net/openvpn/wiki/DeprecatedOptions#Option:--tun-ipv6). Looking at OpenVPN source, this option is commented out in the 2.4 through master branches. Adding or removing this option *should not* affect behavior other than additional log messages - if it does, it is a bug.

lweller:
Sorry, you're absolutely right. I oversaw that tun-ipv6 option is deprecated and not required at all any more, even together with configuration option like ifconfig-ipv6-pool in our case. @mkrkn excuse this eronous pull request.

PR #15673, opened 2021-05-21
podman: change permissions on /etc/containers

flyn-org:
Running podman as users other than root seems to require that those
users can read the configuration files in /etc/containers. This change
sets the permissions of /etc/containers and its contents to match those
used on Fedora.

Maintainer: @oskarirauta
Compile tested: x86_64 master
Run tested: x86_64 master

Description:

### DIFF
--- a/utils/podman/Makefile
+++ b/utils/podman/Makefile
@@ -93,10 +93,10 @@ endef
 define Package/podman/install
 	$(call GoPackage/Package/Install/Bin,$(1))
 	$(INSTALL_DIR) $(1)/etc/containers
-	$(INSTALL_CONF) $(DL_DIR)/default-policy.json-362f70b056 $(1)/etc/containers/policy.json
-	$(INSTALL_CONF) $(DL_DIR)/registries.fedora-da9a9c8778 $(1)/etc/containers/registries.conf
-	$(INSTALL_CONF) $(PKG_BUILD_DIR)/vendor/github.com/containers/storage/storage.conf $(1)/etc/containers/storage.conf
-	$(INSTALL_CONF) ./files/containers.conf $(1)/etc/containers/containers.conf
+	$(INSTALL_DATA) $(DL_DIR)/default-policy.json-362f70b056 $(1)/etc/containers/policy.json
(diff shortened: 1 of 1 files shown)


utils/podman/Makefile
```diff
@@ -93,10 +93,12 @@ endef
 define Package/podman/install
 	$(call GoPackage/Package/Install/Bin,$(1))
 	$(INSTALL_DIR) $(1)/etc/containers
+	chmod 0755 $(1)/etc/containers
```
dangowrt, line 96: This is unneeded, as `$(INSTALL_DIR)` already implies `0755`, see
https://github.com/openwrt/openwrt/blob/master/rules.mk#L284

utils/podman/Makefile
```diff
@@ -93,10 +93,12 @@ endef
...
 	$(INSTALL_CONF) $(DL_DIR)/default-policy.json-362f70b056 $(1)/etc/containers/policy.json
 	$(INSTALL_CONF) $(DL_DIR)/registries.fedora-da9a9c8778 $(1)/etc/containers/registries.conf
 	$(INSTALL_CONF) $(PKG_BUILD_DIR)/vendor/github.com/containers/storage/storage.conf $(1)/etc/containers/storage.conf
 	$(INSTALL_CONF) ./files/containers.conf $(1)/etc/containers/containers.conf
+	chmod 0644 $(1)/etc/containers/*
```
dangowrt, line 101: Please you use `$(INSTALL_DATA)` instead then, that will set `0644` permissions, see
https://github.com/openwrt/openwrt/blob/master/rules.mk#L285

dangowrt (changes requested):
Please use our macros instead of calling `chmod` directly.

utils/podman/Makefile
flyn-org: Fixed. Thank you.

utils/podman/Makefile
flyn-org: Fixed. Thank you.

dangowrt (approved):

dangowrt:
Thanks!

PR #15677, opened 2021-05-21
netifyd: update to 3.07

dsokoloski:
Signed-off-by: Darryl Sokoloski <darryl@sokoloski.ca>

Description:

[FIX] Process queued packets on termination instead of discarding.
[FIX] Force detection of all remaining undetected flows before terminating.
[FIX] Fixed packet queue memory leak.
[FIX] Increased TLS SNI buffer size from 64 to RFC 4366 size of 256 bytes.

[IMP] Sanitize MDNS answers of non-printable characters.
[IMP] Added support for plugin reloading on SIGHUP.
[IMP] Added support for new detection plugin types.
[IMP] Removed Git hash and build date/time from usage text.
[IMP] Reordered application detection logic.
[IMP] Added fully configurable multi-core DPI processing.
[IMP] Added GTP tunnel decapsulation support.
[IMP] Added new flow message type for sockets: flow_purge
[IMP] Completely refactored flow purge logic (moved to main thread).
[IMP] Removed lower/upper MACs from lower hash calculation. To be added back as an option.
[IMP] Added TCP seq order checking.
[IMP] Skype: always set application to skype when protocol was detected as Skype.
[IMP] FreeBSD: Various compilation fixes.
[IMP] OpenWrt: Added support for multiple manual interface configuration options.
[IMP] OpenWrt: Extended shutdown time from 5 to 20 seconds.


net/netifyd/files/netifyd.init
```diff
@@ -11,39 +11,45 @@ USE_PROCD=1
...
+        for ifn in $internal_if; do
+            interfaces="${interfaces} -I ${ifn}"
+        done
+        config_get external_if "$instance" external_if ""
+        for ifn in $external_if; do
+            interfaces="${interfaces} -E ${ifn}"
+        done
+    fi
+
+    config_get_bool options "$instance" options ""
```
ldir-EDB0, line 42: I suspect you mean config_get since options is a string

net/netifyd/files/netifyd.init
```diff
@@ -11,39 +11,45 @@ USE_PROCD=1
...
+
+    procd_open_instance
+    procd_set_param command $PROG -R $options $interfaces
```
ldir-EDB0, line 45: I can't tell you why but I've found making $options the last parameter to work in the case of trying to pass an option ' -F not (ip and udp and dst 239.255.255.250 and dst port 1900 and src 192.168.219.5)'

ldir-EDB0 (changes requested):

net/netifyd/files/netifyd.init
dsokoloski: Good catch!

Fixed.

Also, updated to use lists for internal/external_if entries, and options.

Anything obviously wrong with these changes?

ldir-EDB0: Nothing screams, in fact the updated options have explained the '-F' syntax to me and why I was getting such confusing results.  I'll update my test build and let you know if anything leaps out at me.

At the risk of stylistic war, tabs preferred to spaces and a tab=8.  Tab = 1 byte vs 4/8 spaces = 4/8 bytes.  The config & init script could do with that change basically to match everything else.

ldir-EDB0: Also L77 of Makefile - I don't think you need the CONFIGURE_ARGS parameter since it is included by default https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob;f=include/package-defaults.mk#l120

$(call Build/Configure/Default) should be sufficient, or it seems to work for me

dsokoloski: Thanks for the feedback, changes committed.

net/netifyd/files/netifyd.init
```diff
@@ -11,39 +11,52 @@ USE_PROCD=1
...
-		config_get external_if "$instance" external_if "eth1"
-		NETIFYD_OPTS="-E $external_if -I $internal_if"
+
+	if [ "$autoconfig" -gt 0 ]; then
+	    NETIFYD_AUTODETECT=yes
+	    procd_append_param command "$(auto_detect_options)"
 	fi
 
+	function append_internal_if() {
+	    procd_append_param command -I "$1"
```
ldir-EDB0, line 37: Here I hit a problem I've not idea how to solve.  The example config:
list internal_if 'eth1 -F "not (udp and dst 239.255.255.250 and dst port 1900)"'
will be passed to the -I parameter as a complete field and thus interface 'eth1 -F "not (udp and dst 239.255.255.250 and dst port 1900)"' doesn't exist.

It needs to be passed as 'eth1' '-F' 'not (foo foo foo)'

ldir-EDB0: Also I'd say the in-line functions don't enhance readability - all other scripts I've seen (which isn't all of them!) define all the functions at the top of file and then everything just uses them.

ldir-EDB0 (changes requested):

net/netifyd/files/netifyd.init
dsokoloski: Ok...
- Removed inline functions...
- Added common function to handle (massage) interface list entries with special handling for filter expressions (seems overly complex, but...)

Can't say I'm 100% happy -- there are possible edge cases where the "parser" could go awry, but it will do the right thing for the vast majority.  Let me know if there are bugs or a better way to implement.

ldir-EDB0: Looks better, let me give it a try :-)

net/netifyd/files/netifyd.init
```diff
@@ -10,40 +10,84 @@ STOP=50
...
+	    -*)
+	        if [ $filter -gt 0 ]; then
+	            procd_append_param command "${filter_expr#\ }"
+	            filter=0; filter_expr=
+	        fi
+	        procd_append_param command $a
+	        ;;
+	    *)
+	        if [ $filter -gt 0 ]; then
+	            a=${a#\"}; a=${a%\"}: a=${a#\'}; a=${a%\'}
```
ldir-EDB0, line 36: Rogue colon at the end of the 2nd assignment, should be semi-colon - always something :-)

dsokoloski: So embarrassed!  Of course I folded that file **after** testing :(

Fixed!  ...and thanks a lot of testing!  Much appreciated.

ldir-EDB0 (approved):

Issue #15513, opened 2021-04-25
https-dns-proxy: Memory leak


rhester72:
Maintainer: @stangri
Environment: aarch64, r16357-f8669c174e (Raspberry Pi 4 snapshot as of 2021-03-29)

Description:

There is a persistent and gradual memory leak in https-dns-proxy, at least using default settings:

```
config main 'config'
        option update_dnsmasq_config '*'
        list force_dns_port '53'
        list force_dns_port '853'
        option force_dns '1'

config https-dns-proxy
        option listen_addr '127.0.0.1'
        option user 'nobody'
        option group 'nogroup'
        option listen_port '5053'
        option bootstrap_dns '8.8.8.8,8.8.4.4'
        option resolver_url 'https://dns.google/dns-query'

config https-dns-proxy
        option bootstrap_dns '1.1.1.1,1.0.0.1,2606:4700:4700::1111,2606:4700:4700::1001'
        option resolver_url 'https://cloudflare-dns.com/dns-query'
        option listen_port '5054'
```

This can be pretty clearly demonstrated with a memory profile:

![image](https://user-images.githubusercontent.com/7516483/116002018-957bf000-a5c5-11eb-9db9-4f1e0d274627.png)

Usage patterns are typical of a 5-person household (2 working from home, 2 high school age children in remote learning) with standard usage patterns (video streaming, web browsing, Zoom sessions, etc.).

stangri:
@rhester72 -- which version of https-dns-proxy are you using?

@aarond10 -- Aaron, any other telemetry you'd need to look into this?

rhester72:
Version: 2021-01-17-4

aarond10:
That does seem odd. It could be fragmentation but the fact it's a straight line suggests otherwise. I'd love to know what version of libcurl you're running via something like this:

```
$ ./https_dns_proxy -vvv
[I] 1619485933.784025 main.c:194 Built Apr 16 2021 20:02:29.
[I] 1619485933.784052 main.c:195 System c-ares: 1.17.1
[I] 1619485933.784086 main.c:196 System libcurl: libcurl/7.76.0 OpenSSL/1.1.1k zlib/1.2.11 zstd/1.4.9 libidn2/2.3.0 libpsl/0.21.1 (+libidn2/2.3.0) libssh2/1.9.0 nghttp2/1.43.0
...
```

[...]

PR #15681, opened 2021-05-23
https-dns-proxy: support specifying section names and have no servers

oldium:
Maintainer: @stangri
Compile tested: mips_24kc/ath79, TP-Link Archer C7 v4, OpenWrt Git master (week old)
Run tested: mips_24kc/ath79, TP-Link Archer C7 v4, OpenWrt Git master (week old)

Description:

1. Allow specifying section names in `update_dnsmasq_config` option.
2. Correctly backup empty server list, restore it as so.

### DIFF
--- a/net/https-dns-proxy/Makefile
+++ b/net/https-dns-proxy/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=https-dns-proxy
 PKG_VERSION:=2021-01-17
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/aarond10/https_dns_proxy
--- a/net/https-dns-proxy/files/https-dns-proxy.init
+++ b/net/https-dns-proxy/files/https-dns-proxy.init
@@ -45,7 +45,7 @@ append_parm() {
 }
 
 start_instance() {
-	local cfg="$1" param listen_addr listen_port i
+	local cfg="$1" param listen_addr listen_port i section
 	append_parm "$cfg" 'resolver_url' '-r'
 	append_parm "$cfg" 'polling_interval' '-i'
 	append_parm "$cfg" 'listen_addr' '-a' '127.0.0.1'
@@ -84,8 +84,8 @@ start_instance() {
 		config_load 'dhcp'
 		config_foreach dnsmasq_add_doh_server 'dnsmasq' "${listen_addr}" "${listen_port}"
 	elif [ -n "$dnsmasqConfig" ]; then
-		for i in $dnsmasqConfig; do
-			dnsmasq_add_doh_server "@dnsmasq[${i}]" "${listen_addr}" "${listen_port}"
+		for section in $dnsmasqConfig; do
+			dnsmasq_add_doh_server "${section}" "${listen_addr}" "${listen_port}"
 		done
 	fi
 	p="$((p+1))"
@@ -96,7 +96,7 @@ is_force_dns_active() { iptables-save | grep -q -w -- '--dport 53'; }
 start_service() {
 	local p=5053 c
 	config_load 'https-dns-proxy'
-	config_get dnsmasqConfig	'config' 'update_dnsmasq_config' '*'
+	get_dnsmasq_sections
 	config_get_bool forceDNS	'config' 'force_dns' '1'
 	config_get forceDNSPorts	'config' 'force_dns_port' '53 853'
 	dhcp_backup 'create'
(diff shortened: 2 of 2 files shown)


stangri:
> 1. Allow specifying section names in `update_dnsmasq_config` option.

The existing code should already allow that ([README](https://docs.openwrt.melmac.net/https-dns-proxy/#configuration-settings)). If it doesn't, could you please post the config you've had issues with?

> 2. Correctly backup empty server list, restore it as so.

Good catch! Let me ponder if there isn't a cleaner looking solution than creating another option.

oldium:
> > 1. Allow specifying section names in `update_dnsmasq_config` option.
> 
> The existing code should already allow that ([README](https://docs.openwrt.melmac.net/https-dns-proxy/#configuration-settings)). If it doesn't, could you please post the config you've had issues with?

The README states, that you can use instances (more precisely section indexes), not section names:

> The update_dnsmasq_config option can be set to dash (set to '-' to not change DNSMASQ server settings on start/stop), can be set to '*' to affect all DNSMASQ instance server settings or have a space-separated list of DNSMASQ instances to affect (like '0 4 5'). If this option is omitted, the default setting is '*'.

Example:

```
config dnsmasq 'upstream_dns'
        option rebind_protection '1'
        ...
```

```
config main 'config'
        option update_dnsmasq_config 'upstream_dns'
        ...
```

The following code translates this into `dhcp.@dnsmasq[upstream_dns].server`, which does not work:

https://github.com/openwrt/packages/blob/d886127370fa849b13a69070dab3307db053403d/net/https-dns-proxy/files/https-dns-proxy.init#L83-L90

I found two more issues:

1. There is no check for `-`. It is expected that invalid section reference `dhcp.@dnsmasq[-].server` does all the work for not changing anything.

1. README mentions that entries with # are ignored, but from the reading of the code this is not true:

> Starting with https-dns-proxy version 2019-12-03-3 and higher, when the service is set to update the DNSMASQ servers setting on start/stop, it does not override entries which contain either # or /, so the entries like listed below will be kept in use:

https://github.com/openwrt/packages/blob/d886127370fa849b13a69070dab3307db053403d/net/https-dns-proxy/files/https-dns-proxy.init#L187

This tries to remove `127.0.0.1` from the string to check if it contains `127.0.0.1` and then it tries to remove `/` in order to check if it is there (same can be done with the similar `case` trick used for checking the non-number content). I think (not tested) that this does not check for `#` in any way.

stangri:
Thanks for the code review. I believe [this](https://github.com/stangri/source.openwrt.melmac.net/commit/a073092a0edba6c4e5164e65299ef937798dca09) fixes all the issues you've reported. This should allow using `upstream_dns` as one of the options in `update_dnsmasq_config`.

oldium:
Looks good, I will test it later today.

oldium:
I have tested empty list and it does not work. After starting:

```
config dnsmasq 'upstream_dns'                                                                                                                                                             
        ...
        list doh_backup_server ''\'''\'''
```

After stopping:

```
config dnsmasq 'upstream_dns'                                                                                                                                                             
        ...
        list server ''\'''\'''
```

stangri:
Thanks for a quick test. 

Right, I know why it's happening, I'll be back on Wednesday and I'll fix it then.

stangri:
> I have tested empty list and it does not work. After starting:

Should be fixed in this commit: https://github.com/stangri/source.openwrt.melmac.net/commit/4e1ffff0a436978bfb38b1a88d8a9e53a463362e 

Thanks again for testing!

oldium:
It works, but there is one issue, empty `doh_backup_server` is always created. There are actually two typos

https://github.com/stangri/source.openwrt.melmac.net/blob/b5937c0ebde37ce70c8557225d14bf3979038211/https-dns-proxy/files/https-dns-proxy.init#L185

It should be `server`, not `servers`. And it should use `${cfg}`, not `@dnsmasq[0]`.

stangri:
> It works, but there is one issue, empty `doh_backup_server` is always created. There are actually two typos
> 
> https://github.com/stangri/source.openwrt.melmac.net/blob/b5937c0ebde37ce70c8557225d14bf3979038211/https-dns-proxy/files/https-dns-proxy.init#L185
> 
> It should be `server`, not `servers`. And it should use `${cfg}`, not `@dnsmasq[0]`.

Oh, this is embarrassing, here's a new commit: https://github.com/stangri/source.openwrt.melmac.net/commit/e9ed96ebc8d45b633cde83eeb3e216c99aa33640

oldium:
Works good, thanks :-)

stangri:
Thank you for catching the issues and quick testing. I'm going to send PR for the new version as posted by @curtdept [here](https://github.com/openwrt/packages/issues/15513#issuecomment-850864035) and your issue fixes today or tomorrow.

oldium:
Ok, I will close this PR, then. Thanks.

PR #15694, opened 2021-05-27
dnsdist: update to 1.6.0 and make some features optional

Habbie:
Maintainer: @James-TR but I'm happy to take ownership
Compile tested: compiled on amd64 inside an openwrt git master tree, for the target below
Run tested:  OpenWrt SNAPSHOT, r16595-f4473baf6e, on TP-Link Archer C7/AC1750. Tested forwarding of UDP, DoT and DoH DNS queries.

Description:
this PR updates dnsdist 1.6.0 and updates dependencies accordingly. The additional commits make some dnsdist features optional at build time, which can make the package smaller.

dnsdist (and other PowerDNS products) recently got a configure-time check for the size of `time_t`, because this simplifies reasoning about timestamps. This PR patches that check out because many OpenWRT targets are still 32 bits.

### DIFF
--- a/net/dnsdist/patches/010-time_t-check.patch
+++ b/net/dnsdist/patches/010-time_t-check.patch
@@ -1,0 +1,12 @@
+--- a/configure
++++ b/configure
+@@ -5044,9 +5044,6 @@ cat >>confdefs.h <<_ACEOF
+ _ACEOF
+ 
+ 
+-if test $ac_size -lt 8; then :
+-  as_fn_error $? "size of time_t is $ac_size, which is not large enough to fix the y2k38 bug" "$LINENO" 5
+-fi
+ 
+  typename=`echo time_t | sed "s/[^a-zA-Z0-9_]/_/g"`
+  { $as_echo "$as_me:${as_lineno-$LINENO}: checking whether time_t is signed" >&5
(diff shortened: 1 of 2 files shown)


Habbie:
Need to add signoffs, and fix the quilt patch.

Habbie:
> Need to add signoffs, and fix the quilt patch.

Done, I believe.

net/dnsdist/patches/010-time_t-check.patch
```diff
@@ -0,0 +1,11 @@
+--- a/configure
++++ b/configure
+@@ -5045,7 +5045,7 @@ _ACEOF
+ 
+ 
+ if test $ac_size -lt 8; then :
+-  as_fn_error $? "size of time_t is $ac_size, which is not large enough to fix the y2k38 bug" "$LINENO" 5
++  # as_fn_error $? "size of time_t is $ac_size, which is not large enough to fix the y2k38 bug" "$LINENO" 5
+ fi
```
PolynomialDivision, line 9: Couldn't you just delete the whole if?

Habbie: Sure - I went for the least amount of typing. Would you prefer I remove the whole if?

Habbie: I removed the whole if.

PR #15702, opened 2021-05-28
kplex: add support for Sierra Wireless Gobi GPS

dangowrt:
Maintainer: @nunojpg 
Compile tested: aarch64-cortexa53, mediatek/mt7622
Run tested: BananaPi BPi-R64 with Sierra Wireless EM7455, OpenWrt SNAPSHOT r16799 openwrt/openwrt@7131f5a2fb

Description:
Sierra Wireless modems need the string `$GPS_START` to be sent to the GPS tty device as only then the modem firmware starts emitting NMEA-0183 sentences.
Add an option 'sierragpsstart' to kplex' serial driver to support that quirk as kplex can be very useful to spread GPS data over the network while also supplying 'ugps' using a PTY, allowing for correct system time to be set automatically on boot up from GPS.

This patch is also PR'ed at the upstream project:
https://github.com/stripydog/kplex/pull/54

### DIFF
--- a/net/kplex/Makefile
+++ b/net/kplex/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=kplex
 PKG_VERSION:=1.4
-PKG_RELEASE=2
+PKG_RELEASE=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tgz
 PKG_SOURCE_URL:=http://www.stripydog.com/download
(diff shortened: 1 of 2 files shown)


nunojpg:
We should give upstream some time to consider your pull request...

dangowrt:
> We should give upstream some time to consider your pull request...

Sure, but that shouldn't hinder us from reviewing and testing it for inclusion in OpenWrt packages.git in the meantime. In the end, using the built-in GNSS of LTE modems with kplex is mostly relevant when using routers which contains those modems and run OpenWrt. Hence I was hoping to back the upstream PR by illustrating its usefulness downstream and also make it available for users, as upstream hasn't been active for more than 6 months it can take some time...

PS: I'm also planning to suggest a hotplug-script for kplex taking care of restarting the service in case of a relevant tty device being added.

nunojpg (approved):

PR #15705, opened 2021-05-29
acme: change shell comparisons from numeric to string

### LINKED ISSUE #15704: acme: Missing optional value yields shell error
Maintainer: @tohojo
Environment: multiple arch, snapshot testing

Description:
If some optional config values are entirely missing from UCI then shell comparison used results in a error. This is due to the run script using `-eq` numeric comparisons when string comparison using `=` would seem to be preferred.

PR in work

mrjoel:
Maintainer: @tohojo
Compile tested: N/A, shell only
Run tested: mvebu, recent snapshot

Description:
Correct multiple actual and potential shell comparison issues with comparing input strings as numbers.

Only the first commit is immediately needed for my usage, but I haven't yet tested with full cert renewal or restart. The other commits incrementally change more comparisons to be generally more robust.

Fixes #15704

### DIFF
--- a/net/acme/Makefile
+++ b/net/acme/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=acme
 PKG_VERSION:=2.8.7
-PKG_RELEASE:=6
+PKG_RELEASE:=7
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/acmesh-official/acme.sh/tar.gz/$(PKG_VERSION)?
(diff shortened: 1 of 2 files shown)


tohojo:
Comparing them as numbers is deliberate: they are numbers. The second commit adding the defaults is the bugfix here, please drop the others...

mrjoel:
I'll test again shortly, but I believe that I found having the defaults added to be insufficient. Perhaps for the forced boolean config values it will suffice.

I intentionally made the commits and in higher to lower priority to adjust and rebate as needed, so will reduce to necessary changes.

neheb:
Needs rebasing.

BKPepe:
Ping @mrjoel

PR #15707, opened 2021-05-29
acme: don't log regular messages as error

### LINKED ISSUE #15706: acme: info log duplicates with stderr message as well
Maintainer: @tohojo
Environment: mvebu, snapshot testing

Description:
The usage of the `-s` flag in /usr/lib/acme/run-acme for standard log messages (via `log()`) results in double logging and misleading error category logs when indirectly run via `/etc/init.d/acme start`. This occurs because the `run-acme` script outputs the logger message to stderr, which in turn is captured and logged as an error message by the init system.

The result is that logread messages are generated at the `daemon.err` level which is misleading since no actual error resulted.

mrjoel:
Maintainer: @tohojo
Compile tested: N/A, shell only
Run tested: mvebu, recent snapshot

Description:
Don't duplicate regular log messages to stderr via logger, since when
run via init script the stderr gets captured and erroneously logged as
error messages, duplicating the info message.

Fixes #15706 15706

### DIFF
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -33,7 +33,7 @@ check_cron()
 
 log()
 {
-	logger -t acme -s -p daemon.info -- "$@"
+	logger -t acme -p daemon.info -- "$@"
 }
 
 err()


tohojo:
Doesn't this mean that there will be no output at all if the script is run manually?

mrjoel:
> Doesn't this mean that there will be no output at all if the script is run manually?

Yes that is correct that is a side effect result. I didn't think that direct execution was an expected use case since it's in /usr/lib. If direct execution is expected, then perhaps conditionally detecting whether running interactively and only in such cases adding the flag would be suitable. With such an approach though, I'd actually expect interactive execution to only print messages to stdout/stderr and not use the logger at all.

tohojo:
Joel Johnson ***@***.***> writes:

>> Doesn't this mean that there will be no output at all if the script is run manually?
>
> Yes that is correct that is a side effect result. I didn't think that
> direct execution was an expected use case since it's in /usr/lib. If
> direct execution is expected, then perhaps conditionally detecting
> whether running interactively and only in such cases adding the flag
> would be suitable. With such an approach though, I'd actually expect
> interactive execution to only print messages to stdout/stderr and not
> use the logger at all.

Well, is the init log capture picking up stdout as well for info
messages? In that case getting rid of calls to logger entirely and just
using stdout/stderr for info/error messages could be another way to
avoid the double-logging?

neheb:
Needs rebasing.

BKPepe:
Ping @mrjoel

GeorgeSapkin:
@mrjoel is this still relevant?

GeorgeSapkin:
No updates in over four years and has merge conflicts. Closing as stale. Feel free to reopen as necessary.

Issue #14385, opened 2020-12-31
hwinfo not compiling under glibc


neheb:
Maintainer: @bobafetthotmail 
Environment: glibc

https://downloads.openwrt.org/snapshots/faillogs/arc_archs/packages/hwinfo/compile.txt

bobafetthotmail:
cool, that architecture isn't in the define list in this upstream file so of course it blows up later saying it's undefined https://github.com/openSUSE/hwinfo/blob/master/src/hd/hd.c#L103

Is this the ARC architecture right?  (so it's just "arc" name)

I'll prepare another patch to add it and see if there are other compile problems. But of course I can't run test it as I have no device with arc architecture

PR #15708, opened 2021-05-29
hwinfo: fix compilation with ARC

neheb:
Added a small patch to fix compilation.

Renamed patch files to have a .patch suffix. Makes them easier to read
with vim and probably IDEs.

Remove UCLIBC depend since it's no longer present.

Maintainer: @bobafetthotmail 
Compile tested: arc700


bobafetthotmail:
thanks for taking the time to fix the issue about that https://github.com/openwrt/packages/issues/14385

LGTM

BKPepe:
Is the patch included in the upstream already or it was send to upstream?

bobafetthotmail:
I don't think they care, they don't support that arch either with the tool or the distro (SUSE/OpenSUSE).

PR #15710, opened 2021-05-29
luajit: disable compilation with powerpc

neheb:
Support was removed with version 2.1

Maintainer: @milani 
Compile tested: ath79


neheb:
works with floating point targets looks like.

PR #15726, opened 2021-06-01
https-dns-proxy: 2021-05-14 bugfix: fallback to HTTP/1 by default

stangri:
Maintainer: me
Compile tested: x86_64, Sophos SG-105, 21.02.0-rc2
Run tested:  x86_64, Sophos SG-105, 21.02.0-rc2, start, resolve some addresses, check log, stop/restart

Description: I haven't thoroughly tested the previous commit; it was resulting in error messages in the system log. Until a better solution is found, this is an emergency patch which will use HTTP/1 as a default to ensure proper operation without errors in the log.

PS. Alternatively, I can send PR to roll back to 2021-01-17-7 version instead, please advise.


neheb:
is the ninja patch still needed?

stangri:
> is the ninja patch still needed?

No, thank you, I force-pushed an amended commit.

PR #15730, opened 2021-06-02
strongswan: start charon directly from swanctl

pprindeville:
Maintainer: me, @Thermi
Compile tested: x86_64, generic, HEAD (cb3fb45ed1)
Run tested: same, installed on production satellite router

Description:

Don't use `/usr/lib/ipsec/starter` to start `charon` when we can start it directly more easily.

### DIFF
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=strongswan
 PKG_VERSION:=5.9.2
-PKG_RELEASE:=7
+PKG_RELEASE:=$(AUTORELEASE)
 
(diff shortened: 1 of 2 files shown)


net/strongswan/Makefile
```diff
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=strongswan
 PKG_VERSION:=5.9.2
-PKG_RELEASE:=7
+PKG_RELEASE:=$(AUTORELEASE)
```
Thermi, line 12: Please don't do that in this commit. It's not related to starting charon directly from the swanctl init script.

Thermi (changes requested):

net/strongswan/Makefile
pprindeville: Yup, backed it out. It was generating the wrong value anyway (12).

Issue #15725, opened 2021-05-31
boost 1.76 does not compile on older ARM


neheb:
Maintainer: @ClaymorePT 
Environment: arm

https://downloads.openwrt.org/snapshots/faillogs/arm_fa526/packages/boost/compile.txt
https://downloads.openwrt.org/snapshots/faillogs/arm_xscale/packages/boost/compile.txt
https://downloads.openwrt.org/snapshots/faillogs/arm_arm926ej-s/packages/boost/compile.txt

ClaymorePT:
@neheb It seems that regex library is getting built, something that is no longer necessary. Regex is now a header-only library and building it should be disabled, since it was deprecated for C++11 and forward. The regex library will only be supported for C++03, which is no longer OpenWRT case, right?

neheb:
I mean... OpenWrt supports C+(11 yes.

ClaymorePT:
Well, it seems that the library became a header-only, but for source `boost/regex/v5/perl_matcher_non_recursive.hpp`, someone forgot to update the mutex type along with it's initialization. 
I've opened an issue upstream https://github.com/boostorg/regex/issues/131 and  I'll devise a patch to fix this for OpenWRT.

PR #15771, opened 2021-06-03
boost: Fixes Boost.regex header-only (v5) include

ClaymorePT:
**Maintainer:** @ClaymorePT 
**Compile tested:** Target `Freescale i.MX23/i.MX28` - Profile `I2SE Duckbill boards`
**Run tested:** N/A

# Description:
In v1.76.0 Boost.Regex became a header-only library.
With this update, there are now two different versions:
- v4 for C++03 (deprecated)
- v5 header-only

This commit fixes an issue which was preventing Boost.Regex
from being built for old ArmV5 targets

### DIFF
--- a/libs/boost/patches/020-regex.patch
+++ b/libs/boost/patches/020-regex.patch
@@ -1,0 +1,29 @@
+--- a/boost/regex/v5/mem_block_cache.hpp
++++ b/boost/regex/v5/mem_block_cache.hpp
+@@ -85,10 +85,10 @@ struct mem_block_node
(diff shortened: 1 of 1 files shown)


ClaymorePT:
@neheb Fix for https://github.com/openwrt/packages/issues/15725

ClaymorePT:
btw @neheb The w[iki regarding how to create patches](https://openwrt.org/docs/guide-developer/build-system/use-patches-with-buildsystem) does not contain the instruction for the `make package/boost/refresh` step.

PR #15788, opened 2021-06-06
fix network luci error

JSZMonkey:

Description:

### DIFF
--- a/utils/dockerd/files/dockerd.init
+++ b/utils/dockerd/files/dockerd.init
@@ -43,7 +43,7 @@ uciadd() {
 		logger -t "dockerd-init" -p notice "Adding docker default interface to network uci config (${iface})"
 		uci_quiet add network interface
 		uci_quiet rename network.@interface[-1]="${iface}"
-		uci_quiet set network.@interface[-1].ifname="${device}"
+		uci_quiet set network.@interface[-1].device="${device}"
 		uci_quiet set network.@interface[-1].proto="none"
 		uci_quiet set network.@interface[-1].auto="0"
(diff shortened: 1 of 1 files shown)


G-M0N3Y-2503:
Oh it also looks like the commit format needs fixing, check out the CI results.

feckert:
@G-M0N3Y-2503 If I look at this again now, I would say that we no longer need to set the ports value.
Since we don't have a default container.
All containers that are started and do not create an explicit network are then placed in the bridge docker0.

Therefore I would take out this line completely https://github.com/openwrt/packages/pull/15788/files#diff-332c6bb384fe7efee3adbb8af5f3153a0933f414bb908d5ac0690ade2958e8bfL59

@JSZMonkey Thanks for your contribution but the follwing should be done by you to get this in
* Fill out the git description
* Update your commit header with `dockerd: <your message>`
* Add a commit description why do you have changed this
* Add a Signed-off-by with your real name to your commit
* Change your git config so we could see your real name on git diff

G-M0N3Y-2503:
> @G-M0N3Y-2503 If I look at this again now, I would say that we no longer need to set the ports value.
> Since we don't have a default container running.
> All containers that are started and do not create an explicit network are then placed in the bridge docker0.
> 
> Therefore I would take out this line completely https://github.com/openwrt/packages/pull/15788/files#diff-332c6bb384fe7efee3adbb8af5f3153a0933f414bb908d5ac0690ade2958e8bfL59

Right, I'm not sure what docker does with bridge ports exactly. but I think you're right in that there are none by default.
So I agree, we should just remove that line.

PR #15798, opened 2021-06-07
luaossl: add luaossl(rel-20200709) to feeds

siger-yeung:
Signed-off-by: Siger Yang <sigeryeung@gmail.com>

Maintainer: me
Compile tested: mipsel_24kc, 19.07.3
Run tested: YOUKU YK1

Description:

### DIFF
--- a/lang/luaossl/Makefile
+++ b/lang/luaossl/Makefile
@@ -1,0 +1,59 @@
+#
+# Copyright (C) 2021 Siger Yang <sigeryeung@gmail.com>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=luaossl
+PKG_VERSION:=20200709
+PKG_RELEASE:=1
+PKG_MAINTAINER:=Siger Yang <sigeryeung@gmail.com>
+
+PKG_MIRROR_HASH:=6dbca3cdc50ed7e3b0821783da2407accfb6d25addc3edf1d8e17b00530f5a25
+PKG_SOURCE_URL:=https://github.com/wahern/luaossl.git
+PKG_SOURCE_VERSION:=rel-$(PKG_VERSION)
+PKG_SOURCE_PROTO:=git
+
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_INSTALL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/luaossl
+  SUBMENU:=Lua
+  SECTION:=lang
+  CATEGORY:=Languages
+  TITLE:=Comprehensive binding to OpenSSL for Lua 5.1, 5.2, and later
+  URL:=http://25thandclement.com/~william/projects/luaossl.html
+  DEPENDS:=+liblua +libopenssl
+endef
+
+define Package/luaossl/description
+ luaossl is a comprehensive binding to OpenSSL for Lua 5.1, 5.2, and
+ later. It includes support for certificate and key management, key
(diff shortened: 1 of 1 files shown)


lang/luaossl/Makefile
```diff
@@ -0,0 +1,90 @@
...
+	LUA_APIS="5.1" \
+	lua51cpath="/usr/lib/lua" \
+	lua51path="/usr/lib/lua"
+
+define Package/luaossl/install
+	$(INSTALL_DIR) $(1)/usr/lib/lua
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/lib/lua/_openssl.so $(1)/usr/lib/lua/
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/lib/lua/openssl.lua $(1)/usr/lib/lua/
+
+	$(INSTALL_DIR) $(1)/usr/lib/lua/openssl
```
karlp, line 57: are there any files you're explicitly leaving out in the below long lists?
could this be just?
```
$(CP) $(PKG_INSTALL_DIR)/usr/lib/lua/openssl $(1)/usr/lib/lua
```

siger-young: Copying directly should be OK. I supposed I could only use $(INSTALL_XX) so I did such a stupid thing.

lang/luaossl/Makefile
```diff
@@ -0,0 +1,59 @@
...
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=luaossl
+PKG_VERSION:=20200709
+PKG_RELEASE:=1
+PKG_MAINTAINER:=William Ahern <william@25thandClement.com>, Daurnimator <quae@daurnimator.com>, Siger Yang <sigeryeung@gmail.com>
```
feckert, line 13: You have to be the PKG_MAINTAINER. Since you are then the maintainer of the package in openwrt not the upstream maintainer of luaossl

lang/luaossl/Makefile
```diff
@@ -0,0 +1,59 @@
...
+ distinguished name, alternative name, and X.509v3 extension interfaces. 
+endef
+
+TARGET_CFLAGS += $(FPIC)
+TARGET_LDFLAGS += $(FPIC)
+
+MAKE_FLAGS += \
+	LUA_APIS="5.1" \
+	lua51cpath="/usr/lib/lua" \
+	lua51path="/usr/lib/lua"
```
feckert, line 49: This path points to `/usr/lib/lua` and is so a host compile system dir.
Must this not point to the OpenWrt CrossCompile `STAGING_DIR`?
I do not know exactly how this make of the packages luaossl works.
Its just be a hint to thought about this

lang/luaossl/Makefile
```diff
@@ -0,0 +1,59 @@
...
+
+define Package/luaossl/install
+	$(INSTALL_DIR) $(1)/usr/lib/lua
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/lib/lua/_openssl.so $(1)/usr/lib/lua/
```
feckert, line 53: Is this `_` in `_openssl.so` valid?

siger-young: It's exactly "_openssl.so"

lang/luaossl/Makefile
siger-young: The variable `lua51cpath` is actually a option for "make install", indicating where Lua 5.1 libraries should install. I wrote it in `MAKE_FLAGS` instead of `MAKE_INSTALL_FLAGS` just for convenience. It will be prefixed by ${PKG_INSTALL_DIR} automatically.

PR #15815, opened 2021-06-10
vpn-policy-routing: individual interface reload; support for 21.02.0-rc2

stangri:
Maintainer: me
Compile tested: x86_64, Sophos SG-105, OpenWrt 21.02.0-rc2
Run tested: x86_64, Sophos SG-105, OpenWrt 21.02.0-rc2, start/ensure OpenVPN interfaces are supported

Description: This introduces support for reloading a single interface and also support for 21.02.0-rc2 and up (where the ifname has been renamed to device in network config).

PS. Please grant merge rights for packages.


net/vpn-policy-routing/Makefile
```diff
@@ -4,8 +4,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=vpn-policy-routing
-PKG_VERSION:=0.3.2
-PKG_RELEASE:=20
+PKG_VERSION:=0.3.4
+PKG_RELEASE:=8
```
BKPepe, line 8:
```suggestion
PKG_RELEASE:=1
```

net/vpn-policy-routing/Makefile
```diff
@@ -39,8 +39,8 @@ define Package/vpn-policy-routing/install
 	$(INSTALL_DIR) $(1)/etc/init.d $(1)/etc/config $(1)/etc/hotplug.d/firewall $(1)/etc/
 	$(INSTALL_BIN) ./files/vpn-policy-routing.init $(1)/etc/init.d/vpn-policy-routing
 	$(SED) "s|^\(PKG_VERSION\).*|\1='$(PKG_VERSION)-$(PKG_RELEASE)'|" $(1)/etc/init.d/vpn-policy-routing
-	$(INSTALL_CONF) ./files/vpn-policy-routing.conf $(1)/etc/config/vpn-policy-routing
```
BKPepe, line 42: Do you need the this file? If not, then remove it from this repository as well.

BKPepe (changes requested):

net/vpn-policy-routing/Makefile
stangri: Thanks for the comment, because the binaries are also available from my package repo and not to downgrade the version, it can either be 0.3.4-8 or I can bring it further up to 0.3.5.-1 if it makes any difference.

net/vpn-policy-routing/Makefile
stangri: Oh, thank you, good catch, I'll push the fix!

BKPepe:
As I said in some other PR, which you created:

1. Your commit subject is too long. Make it shorter
2. Consider to write commit description as it seems that you are doing more changes than what's written in the commit message.
For example, renaming the config file. 
3. Still my thoughts about PKG_RELEASE is still valid.

stangri:
> As I said in some other PR, which you created:
> 
> 1. Your commit subject is too long. Make it shorter
> 2. Consider to write commit description as it seems that you are doing more changes than what's written in the commit message.
>    For example, renaming the config file.
> 3. Still my thoughts about PKG_RELEASE is still valid.

1. Commit message has been made short.
2. I've addressed the versioning concern already.

BKPepe:
1. Commit message now says that the `vpn-policy-routing` is updated to _some_ version. Even though, in your commit, you do more changes than simply update the version, right? You should use a commit description and describe the changes which you are making. I told you this in https://github.com/openwrt/packages/pull/15817#pullrequestreview-680594069

You should really take a look at [Submitting patches](https://openwrt.org/submitting-patches)

2. You addressed it how you want not according to [CONTRIBUTING](https://github.com/openwrt/packages/blame/master/CONTRIBUTING.md#L60) :shrug: If you will notice that, you are the only one who is doing the magic with `PKG_RELEASE` even though it is not deserved in this upstream repository.  It is not hard to follow guidelines, right? Otherwise we will not need to have rules for any contribution and it will be mess.

stangri:
Hey, thanks for your feedback!

> 1. Commit message now says that the `vpn-policy-routing` is updated to _some_ version. Even though, in your commit, you do more changes than simply update the version, right? You should use a commit description and describe the changes which you are making. I told you this in [#15817 (review)](https://github.com/openwrt/packages/pull/15817#pullrequestreview-680594069)

Let's put all the conversations in one PR! Previously I would just put the description in the PR body, not the commit message and it seems to have been OK. Would you be OK if I just put a more descriptive text in the PR body?

> 2. You addressed it how you want not according to [CONTRIBUTING](https://github.com/openwrt/packages/blame/master/CONTRIBUTING.md#L60) 🤷 If you will notice that, you are the only one who is doing the magic with `PKG_RELEASE` even though it is not deserved in this upstream repository.  It is not hard to follow guidelines, right? Otherwise we will not need to have rules for any contribution and it will be mess.

I can't recall if it was this PR or 21.02 PR, but I suggested artificially bumping the version to the next major release if there's a problem with the current version number. It was never a problem in the past tho, so I've never had to do it, let me know if you see this as a workaround.

Let me know if these two suggestions work.

PR #15817, opened 2021-06-10
[19.07] vpn-policy-routing: individual interface reload

stangri:
Maintainer: me
Compile tested: x86_64, Sophos SG-105, OpenWrt 21.02.0-rc2
Run tested: x86_64, Sophos SG-105, OpenWrt 21.02.0-rc2, start/ensure OpenVPN interfaces are supported

Description: This introduces support for reloading a single interface and also support for 21.02.0-rc2 and up (where the ifname has been renamed to device in network config).

PS. Please grant merge rights for packages.


BKPepe (changes requested):
NACK from my side. Here are my reasonings:

1. Too many irrelevant changes in the stable branch
Like why should there be added support for changes done in network config in 21.02 in 19.07?

2. Stable branch should contain only security fixes, bug fixes—no new features.
This would prevent the situation when you need to do a new release, where you do fallback to HTTP/1

3.  Still don't understand why you are doing such things like
``PKG_RELEASE:=20`` to ``PKG_RELEASE:=8``once you bump the version. If you want to do it right, then you should do:
```
PKG_VERSION:=0.3.4
PKG_RELEASE:=1
```

4. You added completely new file ``vpn-policy-routing.config``, but you didn't removed the previous one (vpn-policy-routing.conf). It's leftover. It is not said anywhere.

5. Your commit subject is too long. 

6. Use commit description, if it is possible.

PR #15819, opened 2021-06-10
[21.02] samplicator: add samplicator

PolynomialDivision:
Samplicator receives UDP datagrams on a given port and resends those
datagrams to a specified set of receivers.

Use Cases:
- replicate Flow Samples to multiple receivers
- use with conntrackd to synchronize via unicast to multiple targets

(cherry picked from commit 41534e5a19dc8350c1aa27ec926b34f779914ac4)

Maintainer: me

### DIFF
--- a/net/samplicator/Makefile
+++ b/net/samplicator/Makefile
@@ -1,0 +1,42 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=samplicator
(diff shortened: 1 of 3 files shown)


BKPepe:
If new devices can be added in OpenWrt 21.02 RC3, then I don't see a reason why new packages can not be added. ;-)

PR #15822, opened 2021-06-10
libboringssl: Add boringssl library

martinschneider:
Migrated from https://github.com/openwrt/openwrt/pull/4243.

Maintainer: @martinschneider, @codewiz
Compile tested: bcm2711, 19.07.7
Run tested: work in progress (I'll have another PR for nginx coming up which will be a good test for BoringSSL. Meanwhile, I'll keep this PR as a draft.)

Description:
BoringSSL is Google's fork of OpenSSL. Amongst other features, it adds support for QUIC.


libs/boringssl/Makefile
```diff
@@ -0,0 +1,53 @@
...
+  TITLE:=An implementation of the TLS protocol
+  URL:=https://boringssl.googlesource.com/boringssl/
+endef
+
+define Package/boringssl/description
+ An implementation of the TLS protocol
+endef
+
+TARGET_CFLAGS += $(FPIC)
+CMAKE_OPTIONS +=
```
neheb, line 40: ?

libs/boringssl/Makefile
```diff
@@ -0,0 +1,53 @@
...
+  CATEGORY:=Libraries
+  TITLE:=An implementation of the TLS protocol
+  URL:=https://boringssl.googlesource.com/boringssl/
+endef
+
+define Package/boringssl/description
+ An implementation of the TLS protocol
+endef
+
+TARGET_CFLAGS += $(FPIC)
```
neheb, line 39: why not CMAKE_POSITION_INDEPENDENT_CODE ?

libs/boringssl/Makefile
```diff
@@ -0,0 +1,53 @@
...
+PKG_SOURCE_VERSION:=1f54fd9864c054dc33e15b1144e2a6a19fa0a52e
+PKG_MIRROR_HASH:=a60c79ca40d1a73639d4c1bba3431ca630ab1d670ab06307e8442b69cfaf0cc7
+
+PKG_MAINTAINER:=Martin Schneider <martschneider@google.com>\
+	Bernie Innocenti <codewiz@google.com>
+PKG_LICENSE:=OpenSSL,ISC
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=0
```
neheb, line 23: surprising...

libs/boringssl/Makefile
```diff
@@ -0,0 +1,53 @@
...
+PKG_SOURCE_DATE:=2021-06-08
+PKG_SOURCE_VERSION:=1f54fd9864c054dc33e15b1144e2a6a19fa0a52e
+PKG_MIRROR_HASH:=a60c79ca40d1a73639d4c1bba3431ca630ab1d670ab06307e8442b69cfaf0cc7
+
+PKG_MAINTAINER:=Martin Schneider <martschneider@google.com>\
+	Bernie Innocenti <codewiz@google.com>
+PKG_LICENSE:=OpenSSL,ISC
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_PARALLEL:=1
```
neheb, line 22: won't be needed soon.

libs/boringssl/Makefile
```diff
@@ -0,0 +1,53 @@
...
+CMAKE_OPTIONS +=
+
+define Package/boringssl/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/bin/libcrypto.a $(1)/usr/lib/
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/bin/libssl.a $(1)/usr/lib/
+endef
+
+define Package/boringssl-dev/install
+        $(INSTALL_DIR) $(1)/usr/lib
```
neheb, line 49: whitespace issue?

libs/boringssl/Makefile
martinschneider: This line was added by mistake, we don't want FPIC at all. I removed this.

martinschneider:
Hi @neheb, What are the next steps here? Do you have any feedback on this PR?

libs/boringssl/Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+define Package/boringssl
+	SECTION:=libs
+	CATEGORY:=Libraries
+	TITLE:=An implementation of the TLS protocol
+	URL:=https://boringssl.googlesource.com/boringssl/
+	DEPENDS:=+golang \
```
neheb, line 30: Replace with $(GO_ARCH_DEPENDS)

neheb: Actually why is this using go?

libs/boringssl/Makefile
```diff
@@ -0,0 +1,52 @@
...
+		 @(i386||x86_64||arm||armeb||aarch64)
+endef
+
+define Package/boringssl/description
+	An implementation of the TLS protocol
+endef
+
+define Package/boringssl/install
+	$(INSTALL_DIR) $(1)/usr/lib
```
neheb, line 39: You’re installing static libraries… This is wrong. You want to pass -DBUILD_SHARED_LIBS=ON CMAKE_OPTIONS

libs/boringssl/patches/901-remove-werror.patch
```diff
@@ -0,0 +1,28 @@
...
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 99e8cf455..82667b3e3 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -117,7 +117,7 @@ endif()
+ if(CMAKE_COMPILER_IS_GNUCXX OR CLANG)
+   # Note clang-cl is odd and sets both CLANG and MSVC. We base our configuration
+   # primarily on our normal Clang one.
+-  set(C_CXX_FLAGS "-Werror -Wformat=2 -Wsign-compare -Wmissing-field-initializers -Wwrite-strings -Wvla")
++  set(C_CXX_FLAGS "-Wformat=2 -Wsign-compare -Wmissing-field-initializers -Wwrite-strings -Wvla")
```
neheb, line 22: Curious what warnings these are.

libs/boringssl/Makefile
martinschneider: BoringSSL requires Go: https://boringssl.googlesource.com/boringssl/+/HEAD/BUILDING.md

jefferyto: If Go is only a build dependency (and not a runtime dependency), then you only need `PKG_BUILD_DEPENDS:=golang/host`, not `DEPENDS:=golang` or `DEPENDS:=$(GO_ARCH_DEPENDS)`.

martinschneider: I have updated it. Thanks!

libs/boringssl/Makefile
```diff
@@ -0,0 +1,49 @@
...
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+define Package/boringssl
+	SECTION:=libs
+	CATEGORY:=Libraries
+	TITLE:=An implementation of the TLS protocol
+	URL:=https://boringssl.googlesource.com/boringssl/
+	DEPENDS:=+libstdcpp
+	PKG_BUILD_DEPENDS:=golang/host
```
jefferyto, line 30: Apologies for not being specific enough: `PKG_BUILD_DEPENDS` needs to be set in the top level of the makefile (usually between `PKG_LICENSE_FILES` and `include` lines).

libs/boringssl/Makefile
```diff
@@ -0,0 +1,49 @@
...
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://boringssl.googlesource.com/boringssl
+PKG_SOURCE_DATE:=2021-06-08
+PKG_SOURCE_VERSION:=1f54fd9864c054dc33e15b1144e2a6a19fa0a52e
+PKG_MIRROR_HASH:=a60c79ca40d1a73639d4c1bba3431ca630ab1d670ab06307e8442b69cfaf0cc7
+
+PKG_MAINTAINER:=Martin Schneider <martschneider@google.com>
+PKG_LICENSE:=OpenSSL,ISC
```
jefferyto, line 18:
```suggestion
PKG_LICENSE:=OpenSSL ISC
```

Not sure if this is an absolute rule, but `PKG_LICENSE` and `PKG_LICENSE_FILES` are usually space-separated.

martinschneider: Fixed.

libs/boringssl/patches/901-remove-werror.patch
martinschneider: I can no longer reproduce them. I have removed this patch.

jefferyto:
Are the CI failures because those are unsupported architectures? If so, you should include/exclude architectures in `DEPENDS`.

If this supports the same architectures as those of Go, you can use [`$(GO_ARCH_DEPENDS)`](https://github.com/openwrt/packages/blob/e24a969571b0d79e052e467162c0cb59dc2a6d3a/lang/golang/golang-values.mk#L191) (you will also need `include ../../lang/golang/golang-values.mk`). Though from [this](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/include/openssl/base.h#102) it looks like mipsel is supported but not mips (big endian).

libs/boringssl/Makefile
```diff
@@ -0,0 +1,52 @@
+# SPDX-License-Identifier: GPL-2.0-only
```
martinschneider, line 1: @jefferyto Could I clarify what the guidelines for the License/Copyright are here? Is this specific to the package or the stuff that gets packaged? I saw different formats being followed across packages, is there are preferred way?

martinschneider:
> Are the CI failures because those are unsupported architectures? If so, you should include/exclude architectures in `DEPENDS`.
> 
> If this supports the same architectures as those of Go, you can use [`$(GO_ARCH_DEPENDS)`](https://github.com/openwrt/packages/blob/e24a969571b0d79e052e467162c0cb59dc2a6d3a/lang/golang/golang-values.mk#L191) (you will also need `include ../../lang/golang/golang-values.mk`). Though from [this](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/include/openssl/base.h#102) it looks like mipsel is supported but not mips (big endian).

I modified it to depend on $(GO_ARCH_DEPENDS). Could you help me re-trigger the checks?

martinschneider:
> > Are the CI failures because those are unsupported architectures? If so, you should include/exclude architectures in `DEPENDS`.
> > If this supports the same architectures as those of Go, you can use [`$(GO_ARCH_DEPENDS)`](https://github.com/openwrt/packages/blob/e24a969571b0d79e052e467162c0cb59dc2a6d3a/lang/golang/golang-values.mk#L191) (you will also need `include ../../lang/golang/golang-values.mk`). Though from [this](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/include/openssl/base.h#102) it looks like mipsel is supported but not mips (big endian).
> 
> I modified it to depend on $(GO_ARCH_DEPENDS). Could you help me re-trigger the checks?

Yes, only [big endian is not supported](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/include/openssl/base.h#122).

Is there an elegant way to remove mips and mips64 from the dependency string or should I just hard-code it to `@(aarch64||arm||i386||i686||mips64el||mipsel||powerpc64||x86_64)` instead?

neheb:
@!(mips||mips64)

libs/boringssl/Makefile
jefferyto: I'm not really sure if there are guidelines on specifying the license of the makefile/package itself (or what those guidelines may be). I had assumed all contributions followed the same license as the repo.

Maybe @neheb or @BKPepe would know more about this?

martinschneider:
Is the latest error related to my change? I'm not sure where this is coming from and I cannot reproduce it either.

```make[5]: *** No rule to make target '/home/build/openwrt/build_dir/target-x86_64_musl/linux-x86_64/cryptodev-linux-cryptodev-linux-1.12/ioctl.o', needed by '/home/build/openwrt/build_dir/target-x86_64_musl/linux-x86_64/cryptodev-linux-cryptodev-linux-1.12/cryptodev.o'.  Stop.```

champtar:
Nop, cryptodev is known broken in the CI

jefferyto:
> Nop, cryptodev is known broken in the CI

It's not the CI, cryptodev is broken when building with the x86_64 sdk.

champtar:
> > Nop, cryptodev is known broken in the CI
> 
> It's not the CI, cryptodev is broken when building with the x86_64 sdk.

Thanks for the extra precision, I didn't meant to imply that the CI is to blame, the failure is real but still not related to boringssl

PR #15825, opened 2021-06-11
dockerd: refactoring uciadd and ucidel handling

feckert:
Maintainer: @G-M0N3Y-2503 
Run tested: x86_64, APU3, OpenWrt master, Addin and deleting config options with uciadd and ucidel

Description:
This change makes the handling of adding and deleting interface, device-bridges and firewall more robust for dockerd. This change is also necessary to be configuration compliant with the uci option `device` again. With this change, I have also adjusted the log message.

Should replace and close #15788


G-M0N3Y-2503 (approved):
Otherwise, looks good to me.
just bump the release number.

Thanks for the work @feckert

feckert:
@G-M0N3Y-2503 I have added your suggestion for improvement. Can you verify this again?

G-M0N3Y-2503:
> @G-M0N3Y-2503 I have added your suggestion for improvement. Can you verify this again?

It looks much cleaner now, great work and thanks!
I'm just about to test it myself too, but if it works for you, that's good enough for me.

G-M0N3Y-2503:
just quickly tested it and seems good

feckert:
@G-M0N3Y-2503 thanks for testing an verification my changes.
I would also like to bring this into the new stable branch openwrt-21.02. Is there anything against it from your point of view?

G-M0N3Y-2503:
> @G-M0N3Y-2503 thanks for testing an verification my changes.
> I would also like to bring this into the new stable branch openwrt-21.02. Is there anything against it from your point of view?

From what I can see there shouldn't be anything different. So it should be fine.
If you'd like to do it for me, that would be great!

PR #15682, opened 2021-05-23
Add email address

lyda:
Add my email address to my credit line.

Maintainer: @PolynomialDivision
Compile tested: none
Run tested: none

Description: Changed a comment line to add my email address.

PR #15838, opened 2021-06-12
prometheus-node-export-lua: add my email

lyda:
Add my email address to the copyright for the file.

Signed-off-by: Kevin Lyda <kevin@lyda.ie>

Maintainer: @champtar
Compile tested: N/A
Run tested: N/A

Description: New PR based on feedback in #15682. Not totally clear who the maintainer is. I also think some other names should be added there because some folks have added really significant features - ones I'll definitely be using!

### DIFF
--- a/utils/prometheus-node-exporter-lua/files/usr/bin/prometheus-node-exporter-lua
+++ b/utils/prometheus-node-exporter-lua/files/usr/bin/prometheus-node-exporter-lua
@@ -3,7 +3,7 @@
 -- Metrics web server
 
 -- Copyright (c) 2016 Jeff Schornick <jeff@schornick.org>
--- Copyright (c) 2015 Kevin Lyda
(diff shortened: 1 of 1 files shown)


BKPepe (approved):

PolynomialDivision:
Since it is a change in the code, you have to bump `PKG_VERSION:=2021.02.15`? ;)

BKPepe:
That's a question. I would say that I would not bump it just because of added email address in the copyright. See it in the mentioned PR (https://github.com/openwrt/packages/pull/15682#issuecomment-851872233) Otherwise, you will force it to all users who has it installed and worn out their flash memory on their routers because of the missing copyright. I think we can bump it once there are some other changes as well.

PolynomialDivision:
Fine with me.

lyda:
No need to increment the version.  I might have some future fixes/additions that might merit it, but no, not for this.

PR #15847, opened 2021-06-13
autossh: fix compilation without ssh installed

neheb:
configure script looks for host ssh. Just pass the configure variable
directly. --with-ssh doesn't work.

Also get rid of custom Compile section. It's not needed.

Maintainer: @bk138 
Compile tested: alpine linux container

### DIFF
--- a/net/autossh/Makefile
+++ b/net/autossh/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=autossh
 PKG_VERSION:=1.4g
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tgz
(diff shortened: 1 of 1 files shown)


bk138:
Is `ac_cv_path_ssh=/usr/bin/ssh` the replacement for `-D\"SSH_PATH=\\\"\$$$$(SSH)\\\"\"`? If yes, should it maybe use the `SSH` var as well?

neheb:
according to config.log, path_ssh='/usr/bin/ssh' and ssh='/usr/bin/ssh' , so it looks like yes.

bk138:
As a general note, maybe you'd want to use the variable instead of hardcoding `/usr/bin/ssh`. Other than that, I can't really review this as I'm lacking the build-system-fu here, sorry :-/

neheb:
Hardcoding here is fine. Both dropbear and openssh on OpenWrt are installed at /usr/bin/ssh . Well, actually symlinked.

PR #15856, opened 2021-06-13
python3: Add readline to dependency finder script

jefferyto:
Maintainer: me, @commodo 
Compile tested: armvirt-32, 2021-06-07 snapshot sdk (see note below)
Run tested: N/A

Description:
I run-tested this on a regular Python package (that does not use readline) and there was no error. The only package I know that uses readline is apparmor but the package has an [invalid include](https://github.com/openwrt/packages/blob/d6b10350df2331ddd750131543af16d997b27637/utils/apparmor/Makefile#L24) which prevents me from compiling (or configuring) it.

### DIFF
--- a/lang/python/python3-find-stdlib-depends.sh
+++ b/lang/python/python3-find-stdlib-depends.sh
@@ -25,6 +25,7 @@ python3-multiprocessing: multiprocessing
 python3-ncurses: ncurses
 python3-openssl: ssl
 python3-pydoc: doctest pydoc
+python3-readline: readline
 python3-sqlite3: sqlite3
 python3-unittest: unittest
 python3-urllib: urllib


commodo (approved):

commodo:
so, just a quick question here as well;
does this change require a `PKG_RELEASE` bump?

there's also the `PKG_RELEASE:=$(AUTORELEASE)` mechanism; which may work
i'm a little reserved to use it yet;

jefferyto:
> does this change require a `PKG_RELEASE` bump?

This script isn't included in any package so I don't think bumping `PKG_RELEASE` is necessary.

> there's also the `PKG_RELEASE:=$(AUTORELEASE)` mechanism; which may work
> i'm a little reserved to use it yet;

I don't think `$(AUTORELEASE)` is ready - see https://github.com/openwrt/openwrt/commit/9ae3c6f94c616cfbf854d3ec749c7fafc9893942#commitcomment-50277872.

PR #15860, opened 2021-06-14
meson: use ninja from tools

neheb:
Maintainer: @dhewg 
Compile tested: ath79


dhewg:
LGTM, thanks!

dhewg:
The targets don't configure/build though

neheb:
Something wrong with the first commit.

Anyway, I've tested this patchset with all packages. Everything compiles.

cotequeiroz:
I'll give this a spin right now.  It may take me a while to report back, though.

neheb:
looks like the issue was an old SDK lacking ninja.

neheb:
This is missing all the BUILD_PARALLEL stuff I think.

PR #15863, opened 2021-06-14
python3: use gdbm for Berkley dbm

commodo:
Maintainer: me & @jefferyto for Python, @Naoir for gdbm
Compile tested: x86  https://github.com/openwrt/openwrt/commit/2cd1a108290f48fd35373f91056c05277c289687
Run tested: x86 https://github.com/openwrt/openwrt/commit/2cd1a108290f48fd35373f91056c05277c289687

-------------------------------------------------

Python works with GNU DBM and with Berkley DBM.
Berkley DBM has been under Oracle for some time.

And it's not clear how many Python users actually use DBM.

In the packages feed, we have both libdb47 (which is now under Oracle) and
GNU DBM. The GNU DBM has a compatibility layer for Berkley DBM.
There are newer versions than libdb47, but it's probably not worth having
them yet. The libbd47 tarball is ~40+ MB. Odds are newer versions will be
bigger and more bloated.


neheb (approved):

jefferyto (approved):
Apparently Python 2.7 had built-in support for Berkeley DB but this was removed in Python 3 (there is a [third-party interface](https://www.jcea.es/programacion/pybsddb.htm) available). So the dbm module probably hasn't worked all this time :joy:

jefferyto:
Would be great to have this for 21.02 as well :pray:

jefferyto:
This change allows for both [dbm.gnu](https://docs.python.org/3/library/dbm.html#module-dbm.gnu) and [dbm.ndbm](https://docs.python.org/3/library/dbm.html#module-dbm.ndbm) to be available, is this correct?

jefferyto:
Apologies for the multiple comments, I've been thinking about this a bit more (and reading [these notes](https://unixpapa.com/incnote/dbm.html)): Since the dbm module is only linked against gdbm, both dbm.gnu and dbm.ndbm modules actually communicate with the same underlying library (gdbm)... So is it necessary to enable both interfaces?

From my very limited understanding, the native gnu interface is preferable over the ndbm interface, and I believe applications just open and work with databases, they don't specifically select which interface module to use.

In other words, is it necessary to build the libgdbm compatibility library (with `--enable-libgdbm-compat`) just to have it work with Python, since Python already has support for gdbm (with the dbm.gnu module)?

We may need to warn users when upgrading though, since I believe the underlying file formats for db47 and gdbm are incompatible. They will need to find some way to migrate their data.

commodo:
> Apologies for the multiple comments, I've been thinking about this a bit more (and reading [these notes](https://unixpapa.com/incnote/dbm.html)): Since the dbm module is only linked against gdbm, both dbm.gnu and dbm.ndbm modules actually communicate with the same underlying library (gdbm)... So is it necessary to enable both interfaces?
> 

[1]
no; it's not necessary to enable both interfaces;
i don't even know who uses dbm.gnu/ndbm
i only want to get rid of libdb47 [for Python mostly];
mostly to reduce build-time;
and this also reduces the size of a Python installation;
libdb47 has been around for ages (even before my time), so it feels like we're carrying a huge lib/load that may be unused or dead;

that being said, supporting both interfaces is more about enabling/disabling some features of CPython;
CPython supports both GDBM and NDBM, and both of them have been enabled [up until now];
i'm tending to prefer to keep them both under libgdbm;
if people want to use Python to access some old NDBM, because Python let's them, I wouldn't want to get in their way;
i also don't think there are really active users of NDBM with Python on OpenWrt [though I could be surprised];

> From my very limited understanding, the native gnu interface is preferable over the ndbm interface, and I believe applications just open and work with databases, they don't specifically select which interface module to use.
> 
> In other words, is it necessary to build the libgdbm compatibility library (with `--enable-libgdbm-compat`) just to have it work with Python, since Python already has support for gdbm (with the dbm.gnu module)?
>

same reply as [1]
 
> We may need to warn users when upgrading though, since I believe the underlying file formats for db47 and gdbm are incompatible. They will need to find some way to migrate their data.

we could try to warn users;
though i have a good suspicion that if we try to warn users, a lot of them won't be reached [for whatever reason], and we will just get reports of  "hey, this is broken, what gives?"
this seems to be the way things work;
though i'm not against warning users;

it's very hard to tell what the scope of usage is for this (dbm) module;
[ this also relates to usage statistics of OpenWrt packages in general ]
also, the referenced article seems to be going as far back as 2000-2004 ; so, i wouldn't worry too much about NDBM/GDBM [until someone opens a bug report or Github issue]

commodo:
> This change allows for both [dbm.gnu](https://docs.python.org/3/library/dbm.html#module-dbm.gnu) and [dbm.ndbm](https://docs.python.org/3/library/dbm.html#module-dbm.ndbm) to be available, is this correct?

yes; through libgdbm

jefferyto:
> if people want to use Python to access some old NDBM, because Python let's them

Is it possible for someone to install db47 and use it with Python, even though Python was compiled with gdbm? I'm under the impression that because it is linked with gdbm it cannot be used with any other dbm library. (I'll try to find time to do some experiments on this today.)

> though i have a good suspicion that if we try to warn users, a lot of them won't be reached [for whatever reason], and we will just get reports of "hey, this is broken, what gives?"

I just think we should document this change somewhere, like in the Python readme and perhaps a forum post. It would be nice if there was a 21.02 changelog for packages, like the 21.02 release notes for the main repo.

> it's very hard to tell what the scope of usage is for this (dbm) module;

Because we don't know what the actual usage is, I think the safer choice is to assume it is being used and this change will break things for users.

commodo:
> > if people want to use Python to access some old NDBM, because Python let's them
> 
> Is it possible for someone to install db47 and use it with Python, even though Python was compiled with gdbm? I'm under the impression that because it is linked with gdbm it cannot be used with any other dbm library. (I'll try to find time to do some experiments on this today.)

maybe with some symlinking;
but that sounds a bit more tricky/hacky/more-work-to-maintain than my preference;
sounds like creating conflicting libs [libgdmb conflicting with libdb47];

> 
> > though i have a good suspicion that if we try to warn users, a lot of them won't be reached [for whatever reason], and we will just get reports of "hey, this is broken, what gives?"
> 
> I just think we should document this change somewhere, like in the Python readme and perhaps a forum post. It would be nice if there was a 21.02 changelog for packages, like the 21.02 release notes for the main repo.

sounds like a role for a release manager :)
as i remember there are recurring discussions about such a role;
and we get some people doing this for a few releases, but then they need to re-prioritize;

> 
> > it's very hard to tell what the scope of usage is for this (dbm) module;
> 
> Because we don't know what the actual usage is, I think the safer choice is to assume it is being used and this change will break things for users.

[ this is not a serious comment, but ]
we all need a bit of adrenaline in our lives :p
and unfortunately i have not tried extreme sports

> 
> Edit: Just wanted to add, I think this is a good change, just want to do it in the best way possible.

so do i;
"best way possible" is a bit relative
i am happy with just this PR as-is, and seeing how it performs in the wild, versus trying to think too much ahead :)

jefferyto:
> maybe with some symlinking;
> but that sounds a bit more tricky/hacky/more-work-to-maintain than my preference;
> sounds like creating conflicting libs [libgdmb conflicting with libdb47];

I guess what I'm trying to ask/say is, I'd like to see if it is possible to switch to gdbm without changing the gdbm package (i.e. without enabling libgdbm compatibility), unless it is really necessary or if it allows gdbm to work with db47 database files (I highly doubt this is possible).

> sounds like a role for a release manager :)
> as i remember there are recurring discussions about such a role;
> and we get some people doing this for a few releases, but then they need to re-prioritize;

I don't remember any previous release manager gathering release notes for packages in this repo.

I have more opinions on this and other project-wide issues but this is not the place to discuss them.

commodo:
> > maybe with some symlinking;
> > but that sounds a bit more tricky/hacky/more-work-to-maintain than my preference;
> > sounds like creating conflicting libs [libgdmb conflicting with libdb47];
> 
> I guess what I'm trying to ask/say is, I'd like to see if it is possible to switch to gdbm without changing the gdbm package (i.e. without enabling libgdbm compatibility), unless it is really necessary or if it allows gdbm to work with db47 database files (I highly doubt this is possible).
> 

`--enable-libgdbm-compat` is required to enable ndb/ dbm support in libgdbm;
their docs: https://www.gnu.org.ua/software/gdbm/manual/Compatibility.html

it's disabled by default;
https://git.gnu.org.ua/gdbm.git/tree/configure.ac#n49
```
AC_ARG_ENABLE([libgdbm-compat],
  AC_HELP_STRING(
    [--enable-libgdbm-compat],
    [Build and install libgdbm_compat. (Default is NO.)]),
  [want_compat=$enableval],
  [want_compat=no])
```

[setup.py](https://github.com/python/cpython/blob/main/setup.py#L1445)  in CPython seems to detect the settings to enable support for this;

we could maybe change the libgdbm Makefile to go around their `--enable-libgdbm-compat`  flag
but that sounds a bit hacky;
not sure if `libgdbm_compat` gets generated without that flag
what i remember is that the `ndbm.h / dbm.h ` headers don't get installed in the installation directory;
because: https://git.gnu.org.ua/gdbm.git/tree/Makefile.am#n19
```
if COMPAT_OPT
  MAYBE_COMPAT = compat
endif
SUBDIRS = po src doc $(MAYBE_COMPAT) tests
```

jefferyto:
> `--enable-libgdbm-compat` is required to enable ndb/ dbm support in libgdbm;

I'm aware this flag causes the gdbm_compat library to be built, required for ndbm support. My question is can we get away with not having ndbm support.

I tried compiling Python with only the `libdb47` to `libgdbm` change. The compile process errors out because `usr/lib/python3.9/lib-dynload/_dbm.cpython-39.so` is missing and our python3-dbm makefile is [expecting](https://github.com/openwrt/packages/blob/07dbb82e95dc72ddc87d48f97891de5dcb43d999/lang/python/python3/files/python3-package-dbm.mk#L16) this file. It makes sense for _dbm.cpython-39.so to be missing because this is the C module that implements the dbm.ndbm module.

While _dbm.cpython-39.so is left out, the compile process built `usr/lib/python3.9/lib-dynload/_gdbm.cpython-39.so`. This is the C module that implements the dbm.gdbm module.

If we changed the python3-dbm makefile to install _gdbm.cpython-39.so instead of _dbm.cpython-39.so, then it would work.

**However**, when opening a database file, the dbm module uses the [`whichdb()`](https://docs.python.org/3/library/dbm.html#dbm.whichdb) function to guess which database module to use. So I think including the dbm.ndbm module can be useful, if the user has ndbm-looking, gdbm-generated database files :joy:

(gdbm should still not be able to open ndbm-looking, db47-generated database files though, so there may still be breakage for some users.)

So I'm okay with enabling the gdbm_compat library / building _dbm.cpython-39.so (apologies for the trouble), but I think the python3-dbm makefile should also install `usr/lib/python3.9/lib-dynload/_gdbm.cpython-39.so`.

commodo:
> So I'm okay with enabling the gdbm_compat library / building _dbm.cpython-39.so (apologies for the trouble), but I think the python3-dbm makefile should also install `usr/lib/python3.9/lib-dynload/_gdbm.cpython-39.so`.

umm, that sounds a bit complicated to install the same file from 2 packages;
the build would find the collision and complain;
we could make `python3-gdbm` a dependency of `python3-dbm`

commodo:
> I'm against splitting dbm.gnu into a separate package. Users will not know to install this package because installing python3-dbm will most likely "just work", and they lose out on the advantages of the native gdbm interface.

a, we may have mismatch in expressing stuff;
merging `python3-gdbm` into `python3-dbm` is fine; if that's what you're saying

commodo:
Changelog v2 -> v3:
* merge `python3-gdbm` into `python3-dbm`

jefferyto:
Oh I'm sorry, I forgot python3-gdbm was already a package :facepalm: Now that libgdbm is the only dependency I think merging it into python3-dbm makes sense.

> for `gdbm` : disabling the static library

Are we sure no other package uses the static library?

commodo:
> Are we sure no other package uses the static library?

no; but I'm fine to cause temporary breakage to notify people to ping back if it's needed;
but to be fair, i do not own the package, so I'd let @Naoir decide on this one

neheb:
What's the status here?

jefferyto:
Waiting for comment from the gdbm maintainer (@Naoir) about the changes to that package - if the changes are okay @neheb please feel free to merge :pray:

neheb:
He hasn't been active for several years. I'll wait a while anyway.

commodo:
Changelog v3 -> v4:
* rebased on newer master; there was a Python3 update recently; PKG_RELEASE was conflicting with master

jefferyto:
The Python changes are not dependent on disabling static library build for gdbm, is this correct? Perhaps this PR can be merged if we split that change into a separate PR?

commodo:
> The Python changes are not dependent on disabling static library build for gdbm, is this correct? Perhaps this PR can be merged if we split that change into a separate PR?

yes, we can remove the static-library disabling from this PR;
only the `--enable-libgdbm-compat` flag is needed

commodo:
oh;
so this went in with the disabling of static libs;
let's see how this performs

PR #15853, opened 2021-06-13
[stubby,getdns]: Remove maintainer from stubby and getdns packages

jonathanunderwood:
Maintainer: me
Compile tested: N/A
Run tested: N/A

Description:
@neheb persists in merging PRs without package maintainer notification and approval. I have raised this previously, but this behaviour persists. A recent PR was merged which lacks the relevant documentation updates, and with problems that should have been addressed during review, and I am not prepared to pick up the mess and continue to maintain packages when a key maintainer does not follow process.

PR #15643, opened 2021-05-17
apparmor: add package

oskarirauta:
Signed-off-by: Oskari Rauta oskari.rauta@gmail.com

Maintainer: Oskari Rauta / @oskarirauta (find it by checking history of the package Makefile)
Compile tested: x86_64, server, recent snapshot
Run tested: x86_64, server, recent snapshot, tested, works

Description:
AppArmor userland libraries and utilities.

Requires python3-readline #15369
Replaces #15481

PR #15875, opened 2021-06-15
apparmor: fix packaging

CodeFetch:
Removes the BUILDONLY flag as otherwise no packages will be created.

@oskarirauta


neheb:
```
Commit sha: 6e5880b, Author: Vincent Wiemann, Committer: GitHub; Expected "Vincent Wiemann vincent.wiemann@ironai.com", but got "Vincent Wiemann me@bibbl.com".
```

CodeFetch:
@neheb Thanks for the hint! One more reason to not use Github's web interface for committing.

BKPepe:
Where is approve from maintaner? :eyes:

CodeFetch:
@BKPepe This was a point of discussion right from the beginning. https://github.com/openwrt/packages/pull/15643#discussion_r644097179

We have tested it and no packages will be created with the BUILDONLY flag set and the files are missing in rootfs. @oskarirauta said that something did not work, but I can't confirm that. The BUILDONLY flag is used for packages which are only used as build dependencies. It's wrong to use it here.

BKPepe:
Well, don't take it wrong, but what's the point of being maintainer if he didn't approve it and it is merged no matter what? 1 or 2 days from my point of view are not enough to review the things! 

Based on your comment, I could see that we don't need to have maintainers at all and that's why they are leaving like https://github.com/openwrt/packages/pull/15853

PR #15882, opened 2021-06-16
libesmtp: update to 1.1.0

neheb:
New upstream. Uses meson now. Removed outdated patches.

Maintainer: @tru7 
Compile tested: ath79


tru7:
thank you for taking care of.
I postponed working on it because the new version will break building packages depending on it because they rely on the now dropped libesmtp_config utility. I'm already working on linknx to deal with it.

Why don't you install the plugins into the created esmtp-plugins-6.2.0 folder?

neheb:
Mistake. Good catch.

tru7 (approved):
Well done. Thank you

PR #15886, opened 2021-06-16
python-cryptography: bump to version 3.4.7

commodo:
Maintainer: me, @jefferyto 
Compile tested:  x86 https://github.com/openwrt/openwrt/commit/fbb9b1f8ed0d8a76dd989cc6c16a4e0fda2b6e74
Run tested:  x86  https://github.com/openwrt/openwrt/commit/fbb9b1f8ed0d8a76dd989cc6c16a4e0fda2b6e74

--------------------------------------------------------------------------

Dropped patches:
  0004-Replace-EVP_CIPHER_CTX_cleanup-with-EVP_CIPHER_CTX_r.patch
  0005-Switch-get_-Update-APIs-to-get0.patch

Reworked patches:
  0001-Add-new-ASN1_STRING_get0_data-API.patch
  0006-Add-X509_STORE_CTX_trusted_stack-compatibility-macro.patch
These 2 require that we keep only the CUSTOMIZATIONS stuff for now. Maybe
later we can drop this.

Ran 'make package/python-cryptography/refresh'.

Added patch:
  0004-make-rust-completely-optional-during-build.patch
upstream did a sloppy job with the CRYPTOGRAPHY_DONT_BUILD_RUST logic; we
need to patch it, to make sure the setuptools-rust isn't installed.
We may need to carry this patch in our tree for a bit longer than upstream,
because in newer versions, CRYPTOGRAPHY_DONT_BUILD_RUST logic gets removed.


commodo:
Changelog v1 -> v2:
* re-ran `make package/python-cryptography/refresh` ; patch `0004-make-rust-completely-optional-during-build.patch` differs a bit after going through quilt

lang/python/python-cryptography/patches/0004-make-rust-completely-optional-during-build.patch
```diff
@@ -0,0 +1,109 @@
...
+@@ -41,13 +44,48 @@ with open(os.path.join(src_dir, "cryptog
+ 
+ # `install_requirements` and `setup_requirements` must be kept in sync with
+ # `pyproject.toml`
+-setuptools_rust = "setuptools-rust>=0.11.4"
+ install_requirements = ["cffi>=1.12"]
+-setup_requirements = install_requirements + [setuptools_rust]
++setup_requirements = install_requirements
++
++extra_requirements = {
```
jefferyto, line 25: This patch looks like it is correct to make Rust completely optional, but I think it could be a lot smaller if we instead just disable Rust completely (skip the `RustExtension` import test, remove `setuptools_rust` from `setup_requirements`). (I don't think upstream will ever accept this patch anyway.)

In particular, `extras_require` can be left alone as these dependencies are not installed by default. (To install these pip would need to be invoked like `pip install cryptography[test]` or `pip install cryptography[sdist]`.)

commodo: yeah; you're right;
this patch was initially made to be sent upstream, then i saw that they removed the `CRYPTOGRAPHY_DONT_BUILD_RUST`  stuff in master [after i made this];

commodo:
Changelog v2 -> v3:
* reworked patch `0004-make-rust-completely-optional-during-build.patch` to a minimal form that completely disables the Rust build without the `CRYPTOGRAPHY_DONT_BUILD_RUST` env var

lang/python/python-cryptography/patches/0004-make-rust-completely-optional-during-build.patch
```diff
@@ -0,0 +1,33 @@
...
+     rust_extensions = []
+ else:
+     rust_extensions = [
+@@ -129,9 +129,6 @@ try:
+                 "twine >= 1.12.0",
+                 "sphinxcontrib-spelling >= 4.0.1",
+             ],
+-            "sdist": [
+-                setuptools_rust,
+-            ],
```
jefferyto, line 30: I think these lines can be left in `extras_require` as `setuptools_rust` is just a string (`"setuptools-rust>=0.11.4"`) and `sdist` is (should?) never specified as an install option.

jefferyto (commented):
I think it would also be good to rename the patch :pray:

lang/python/python-cryptography/patches/0004-make-rust-completely-optional-during-build.patch
commodo: ack

commodo:
> I think it would also be good to rename the patch pray

ummm; sure: suggestion? :)
this sounds like a preference,
so i'll take suggestions for the rename, as it's likely that i missed any hints about what to rename this to

jefferyto:
> ummm; sure: suggestion? :)

0004-disable-rust.patch ?

The original name (0004-make-rust-completely-optional-during-build.patch) seems to indicate that it makes Rust completely optional but still available.

commodo:
Changelog v3 -> v4:
* renamed `0004-make-rust-completely-optional-during-build.patch` -> `0004-disable-rust.patch`
* simplified  `0004-disable-rust.patch` patch; the patch is changing only 2 lines in `setup.py` ; one for `from setuptools_rust import RustExtension` and `if os.environ.get("CRYPTOGRAPHY_DONT_BUILD_RUST"):` -> `if True:`

jefferyto:
I think you still need to remove `setuptools_rust` from `setup_requirements`, otherwise setup.py will try to install setuptools_rust.

hnyman:
ok to merge? discussion finished?

jefferyto:
@hnyman okay to merge - thanks again @commodo

PR #15888, opened 2021-06-16
linknx: adopt new libesmtp version

tru7:
Signed-off-by: Othmar Truniger <github@truniger.ch>

Maintainer: me
Compile tested: mpc85xx, tl-wdr4900-v1, trunk
Compile tested: ath79, tplink_archer-c7-v4, trunk
Compile tested: ramips, mt7620a, trunk
Run tested: ath79, tp-link Archer C7 v4, trunk

Description:
add patch to adopt new libesmtp version


BKPepe:
Isn't there missing bump of `PKG_RELEASE`?

tru7:
Yes, I forgot. But on the other hand there is no package upgrade needed as it only applies to building the package - the outcome is the same.

PR #15889, opened 2021-06-17
podman: change permissions on /usr/share/containers/seccomp.json

flyn-org:
Running podman as users other than root seems to require that those
users can read /usr/share/containers/seccomp.json. This change sets the
permissions on that file to match those used on Fedora.

Maintainer: @oskarirauta
Compile tested: x86_64 master
Run tested: x86_64 master

Description:


flyn-org:
@dangowrt, this is similar to https://github.com/openwrt/packages/pull/15673.

PR #15910, opened 2021-06-19
nextdns: Update to version 1.33.0

rs:
Update to version 1.33.0


neheb:
```
WARNING: PKG_MIRROR_HASH does not match nextdns-1.33.0.tar.xz hash 4e7f2ae6d40810a62036a736a7e733c4b394dc33a619fbb893100d75f869838a
```

rs:
I know, it's been several version like that, nothing changed on our side but the hash is now failing.

neheb:
One of these generated tarballs are needed for comparison.

PR #15922, opened 2021-06-20
strongswan: add support for missing options and fixup/cleanup

CodeFetch:
Maintainer: @pprindeville @thermi
Compile tested: ath79
Run tested: ath79

Adds a little fixup and support for configuring public keys for authentication without certificates (mod-pubkey).

### DIFF
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -70,6 +70,10 @@ xappend() {
 	echo "$@" >> "$file"
 }
 
+append_quoted_comma() {
+	append quoted_comma_list "\"$1\"" ","
+}
+
 swan_reset() {
 	file_reset "$STRONGSWAN_VAR_CONF_FILE"
 }
@@ -146,6 +150,66 @@ is_aead() {
 	esac
 
 	return 1
+}
+
+authority_one_exclusive() {
+	fatal "Configure either cacert, file, or handle per authority section!"
+}
+
+config_authority() {
+	local config_name="$1"
+
+	local enabled
+
+	local cacert
+	local cert_uri_base
+	local crl_uris
+	local file
+	local handle
+	local module
+	local oscp_uris
+	local slot
+
+	config_get_bool enabled "$1" enabled 0
+	[ $enabled -eq 0 ] && return
+
+	config_get cacert "$1" cacert ""
+	config_get cert_uri_base "$1" cert_uri_base ""
+	config_get crl_uris "$1" crl_uris ""
+	config_get file "$1" file ""
+	config_get handle "$1" handle ""
+	config_get module "$1" module ""
+	config_get oscp_uris "$1" oscp_uris ""
+	config_get slot "$1" slot ""
+
+	[ -z "$cacert" ] && [ -z "$file" ] && [ -z "$handle" ] && authority_one_exclusive
+
+	[ -n "$cacert" ] && [ -n "$file" ] && authority_one_exclusive
+	[ -n "$cacert" ] && [ -n "$handle" ] && authority_one_exclusive
+	[ -n "$file" ] && [ -n "$handle" ] && authority_one_exclusive
+
+	swanctl_xappend0 "authorities {"
+	swanctl_xappend1 "$config_name {"
+	[ -n "$cacert" ] && swanctl_xappend2 "cacert = $cacert"
+	[ -n "$cert_uri_base" ] && swanctl_xappend2 "cert_uri_base = $cert_uri_base"
+
+	local quoted_comma_list=""
+	config_list_foreach "$1" crl_uris append_quoted_comma_list
+	[ -n "$quoted_comma_list" ] && swanctl_xappend3 "crl_uris = $quoted_comma_list"
+
+	[ -n "$file" ] && swanctl_xappend2 "file = $file"
+	[ -n "$handle" ] && swanctl_xappend2 "handle = $handle"
+	[ -n "$module" ] && swanctl_xappend2 "module = $module"
+
+	local quoted_comma_list=""
+	config_list_foreach "$1" oscp_uris append_quoted_comma_list
+	[ -n "$quoted_comma_list" ] && swanctl_xappend3 "oscp_uris = $quoted_comma_list"
+
+	[ -n "$slot" ] && swanctl_xappend2 "slot = $slot"
+	swanctl_xappend1 "}"
+	swanctl_xappend0 "}"
+
+	swanctl_xappend0 ""
 }
 
 add_esp_proposal() {
@@ -329,10 +393,6 @@ config_transport() {
 	config_child "$1" "transport"
 }
 
-append_pubkey() {
-	append pubkey_list "\"$1\"" ","
-}
-
 config_connection() {
 	local config_name="$1"
 
@@ -351,14 +411,12 @@ config_connection() {
 	local mobike
 	local local_cert
 	local local_key
-	local ca_cert
 	local rekeytime
 
 	config_get_bool enabled "$1" enabled 0
 	[ $enabled -eq 0 ] && return
 
 	config_get auth_method "$1" authentication_method
-	config_get ca_cert "$1" ca_cert ""
 	config_get dpddelay "$1" dpddelay "30s"
 	config_get fragmentation "$1" fragmentation "yes"
 	config_get gateway "$1" gateway
@@ -441,9 +499,9 @@ config_connection() {
 	if [ "$auth_method" = pubkey ]; then
 		[ -n "$local_cert" ] && swanctl_xappend3 "certs = $local_cert"
 
-		local pubkey_list=""
-		config_list_foreach "$1" local_pubkeys append_pubkey
-		[ -n "$pubkey_list" ] && swanctl_xappend3 "pubkeys = $pubkey_list"
+		local quoted_comma_list=""
+		config_list_foreach "$1" local_pubkeys append_quoted_comma_list
+		[ -n "$quoted_comma_list" ] && swanctl_xappend3 "pubkeys = $quoted_comma_list"
 	fi
 	swanctl_xappend2 "}"
 
@@ -451,9 +509,9 @@ config_connection() {
 	swanctl_xappend3 "auth = $auth_method"
 	[ -n "$remote_identifier" ] && swanctl_xappend3 "id = \"$remote_identifier\""
 	if [ "$auth_method" = pubkey ]; then
-		local pubkey_list=""
-		config_list_foreach "$1" remote_pubkeys append_pubkey
-		[ -n "$pubkey_list" ] && swanctl_xappend3 "pubkeys = $pubkey_list"
+		local quoted_comma_list=""
+		config_list_foreach "$1" remote_pubkeys append_quoted_comma_list
+		[ -n "$quoted_comma_list" ] && swanctl_xappend3 "pubkeys = $quoted_comma_list"
 	fi
 	swanctl_xappend2 "}"
 
@@ -472,17 +530,6 @@ config_connection() {
 		swanctl_xappend2 "filename = $local_key"
 		swanctl_xappend1 "}"
 		swanctl_xappend0 "}"
-
-		swanctl_xappend0 ""
-
-		if [ -n "$ca_cert" ]; then
-			swanctl_xappend0 "authorities {"
-			swanctl_xappend1 "$config_name {"
-			swanctl_xappend2 "cacert = $ca_cert"
-			swanctl_xappend1 "}"
-			swanctl_xappend0 "}"
-		fi
-
 	elif [ "$auth_method" = psk ]; then
 		swanctl_xappend0 ""
 
@@ -577,6 +624,7 @@ prepare_env() {
 
 	config_load ipsec
 	config_foreach config_ipsec ipsec
+	config_foreach config_authority authority
 	config_foreach config_connection remote
 
 	do_postamble
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -303,10 +303,8 @@ config_child() {
 
 	local ah_proposal
 	local esp_proposal
-	local local_ts
 	local updown
 	local firewall
-	local remote_ts
 	local lifetime
 	local dpdaction
 	local closeaction
@@ -315,10 +313,8 @@ config_child() {
 	local rekeytime
 
 	config_get startaction "$1" startaction "route"
-	config_get local_ts "$1" local_ts ""
 	config_get updown "$1" updown ""
 	config_get firewall "$1" firewall ""
-	config_get remote_ts "$1" remote_ts ""
 	config_get lifetime "$1" lifetime ""
 	config_get dpdaction "$1" dpdaction "none"
 	config_get closeaction "$1" closeaction "none"
@@ -383,8 +379,14 @@ config_child() {
 
 	swanctl_xappend3 "$config_name {"
 
-	[ -n "$local_ts" ] && swanctl_xappend4 "local_ts = $local_ts"
-	[ -n "$remote_ts" ] && swanctl_xappend4 "remote_ts = $remote_ts"
+	local quoted_comma_list=""
+	config_list_foreach "$1" local_ts append_quoted_comma_list
+	[ -n "$quoted_comma_list" ] && swanctl_xappend4 "local_ts = $quoted_comma_list"
+
+	local quoted_comma_list=""
+	config_list_foreach "$1" remote_ts append_quoted_comma_list
+	[ -n "$quoted_comma_list" ] && swanctl_xappend4 "remote_ts = $quoted_comma_list"
+
 	[ -n "$if_id" ] && { swanctl_xappend4 "if_id_in = $if_id" ; swanctl_xappend4 "if_id_out = $if_id" ; }
 	[ -n "$startaction" -a "$startaction" != "none" ] && swanctl_xappend4 "start_action = $startaction"
 	[ -n "$closeaction" -a "$closeaction" != "none" ] && swanctl_xappend4 "close_action = $closeaction"
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -449,7 +449,6 @@ config_connection() {
 
 	local enabled
 	local gateway
-	local local_sourceip
 	local local_leftip
 	local remote_gateway
 	local pre_shared_key
@@ -478,7 +477,6 @@ config_connection() {
 	config_get local_identifier "$1" local_identifier ""
 	config_get local_key "$1" local_key ""
 	config_get local_leftip "$1" local_leftip "%any"
-	config_get local_sourceip "$1" local_sourceip ""
 	config_get_bool mobike "$1" mobike 1
 	config_get over_time "$1" over_time
 	config_get pre_shared_key "$1" pre_shared_key
@@ -537,7 +535,9 @@ config_connection() {
 	[ -n "$dpd_delay" ] && swanctl_xappend2 "dpd_delay = $dpd_delay"
 	[ "$keyingtries" = "%forever" ] && swanctl_xappend2 "keyingtries = 0" || swanctl_xappend2 "keyingtries = $keyingtries"
 
-	[ -n "$local_sourceip" ] && swanctl_xappend2 "vips = $local_sourceip"
+	local quoted_comma_list=""
+	config_list_foreach "$1" vips append_quoted_comma_list
+	[ -n "$quoted_comma_list" ] && swanctl_xappend2 "vips = $quoted_comma_list"
 
 	[ "$gateway" = "any" ] && remote_gateway="%any" || remote_gateway="$gateway"
 
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -301,6 +301,7 @@ config_child() {
 	local config_name="$1"
 	local mode="$2"
 
+	local hw_offload
 	local interface
 	local ipcomp
 	local priority
@@ -315,6 +316,7 @@ config_child() {
 	local if_id
 	local rekey_time
 
+	config_get hw_offload "$1" hw_offload ""
 	config_get interface "$1" interface ""
 	config_get_bool ipcomp "$1" ipcomp 0
 	config_get priority "$1" priority ""
@@ -393,6 +395,7 @@ config_child() {
 	config_list_foreach "$1" remote_ts append_quoted_comma_list
 	[ -n "$quoted_comma_list" ] && swanctl_xappend4 "remote_ts = $quoted_comma_list"
 
+	[ -n "$hw_offload" ] && swanctl_xappend4 "hw_offload = $hw_offload"
 	[ -n "$interface" ] && swanctl_xappend4 "interface = $interface"
 	[ $ipcomp -eq 1 ] && swanctl_xappend4 "ipcomp = 1"
 	[ -n "$priority" ] && swanctl_xappend4 "priority = $priority"
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -301,6 +301,7 @@ config_child() {
 	local config_name="$1"
 	local mode="$2"
 
+	local ipcomp
 	local priority
 	local ah_proposal
 	local esp_proposal
@@ -313,6 +314,7 @@ config_child() {
 	local if_id
 	local rekey_time
 
+	config_get_bool ipcomp "$1" ipcomp 0
 	config_get priority "$1" priority ""
 	config_get startaction "$1" startaction "route"
 	config_get updown "$1" updown ""
@@ -326,6 +328,7 @@ config_child() {
 	parse_ah_proposal "$1" ah_proposal
 	parse_esp_proposal "$1" esp_proposal
 
+	[ $ipcomp -eq 1 ] && swanctl_xappend4 "ipcomp = 1"
 	[ -n "$priority" ] && swanctl_xappend4 "priority = $priority"
 
 	# translate from ipsec to swanctl
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -328,9 +328,6 @@ config_child() {
 	parse_ah_proposal "$1" ah_proposal
 	parse_esp_proposal "$1" esp_proposal
 
-	[ $ipcomp -eq 1 ] && swanctl_xappend4 "ipcomp = 1"
-	[ -n "$priority" ] && swanctl_xappend4 "priority = $priority"
-
 	# translate from ipsec to swanctl
 	case "$startaction" in
 	add)
@@ -394,6 +391,8 @@ config_child() {
 	config_list_foreach "$1" remote_ts append_quoted_comma_list
 	[ -n "$quoted_comma_list" ] && swanctl_xappend4 "remote_ts = $quoted_comma_list"
 
+	[ $ipcomp -eq 1 ] && swanctl_xappend4 "ipcomp = 1"
+	[ -n "$priority" ] && swanctl_xappend4 "priority = $priority"
 	[ -n "$if_id" ] && { swanctl_xappend4 "if_id_in = $if_id" ; swanctl_xappend4 "if_id_out = $if_id" ; }
 	[ -n "$startaction" -a "$startaction" != "none" ] && swanctl_xappend4 "start_action = $startaction"
 	[ -n "$closeaction" -a "$closeaction" != "none" ] && swanctl_xappend4 "close_action = $closeaction"
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -210,7 +210,7 @@ parse_ike_proposal() {
 	export -n "$var=$crypto"
 }
 
-config_conn() {
+config_child() {
 	# Generic swanctl.conf children section shared by tunnel and transport
 	local config_name="$1"
 	local mode="$2"
@@ -322,18 +322,18 @@ config_conn() {
 }
 
 config_tunnel() {
-	config_conn "$1" "tunnel"
+	config_child "$1" "tunnel"
 }
 
 config_transport() {
-	config_conn "$1" "transport"
+	config_child "$1" "transport"
 }
 
 append_pubkey() {
 	append pubkey_list "\"$1\"" ","
 }
 
-config_remote() {
+config_connection() {
 	local config_name="$1"
 
 	local enabled
@@ -577,7 +577,7 @@ prepare_env() {
 
 	config_load ipsec
 	config_foreach config_ipsec ipsec
-	config_foreach config_remote remote
+	config_foreach config_connection remote
 
 	do_postamble
 }
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -280,7 +280,6 @@ config_child() {
 	local mode="$2"
 
 	local local_ts
-	local local_nat
 	local updown
 	local firewall
 	local remote_ts
@@ -293,7 +292,6 @@ config_child() {
 
 	config_get startaction "$1" startaction "route"
 	config_get local_ts "$1" local_ts ""
-	config_get local_nat "$1" local_nat ""
 	config_get updown "$1" updown ""
 	config_get firewall "$1" firewall ""
 	config_get remote_ts "$1" remote_ts ""
@@ -358,8 +356,6 @@ config_child() {
 		dpdaction=
 		;;
 	esac
-
-	[ -n "$local_nat" ] && local_ts="$local_nat"
 
 	swanctl_xappend3 "$config_name {"
 
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -279,11 +279,11 @@ config_child() {
 	local config_name="$1"
 	local mode="$2"
 
-	local local_subnet
+	local local_ts
 	local local_nat
 	local updown
 	local firewall
-	local remote_subnet
+	local remote_ts
 	local lifetime
 	local dpdaction
 	local closeaction
@@ -292,11 +292,11 @@ config_child() {
 	local rekeytime
 
 	config_get startaction "$1" startaction "route"
-	config_get local_subnet "$1" local_subnet ""
+	config_get local_ts "$1" local_ts ""
 	config_get local_nat "$1" local_nat ""
 	config_get updown "$1" updown ""
 	config_get firewall "$1" firewall ""
-	config_get remote_subnet "$1" remote_subnet ""
+	config_get remote_ts "$1" remote_ts ""
 	config_get lifetime "$1" lifetime ""
 	config_get dpdaction "$1" dpdaction "none"
 	config_get closeaction "$1" closeaction "none"
@@ -359,12 +359,12 @@ config_child() {
 		;;
 	esac
 
-	[ -n "$local_nat" ] && local_subnet="$local_nat"
+	[ -n "$local_nat" ] && local_ts="$local_nat"
 
 	swanctl_xappend3 "$config_name {"
 
-	[ -n "$local_subnet" ] && swanctl_xappend4 "local_ts = $local_subnet"
-	[ -n "$remote_subnet" ] && swanctl_xappend4 "remote_ts = $remote_subnet"
+	[ -n "$local_ts" ] && swanctl_xappend4 "local_ts = $local_ts"
+	[ -n "$remote_ts" ] && swanctl_xappend4 "remote_ts = $remote_ts"
 	[ -n "$if_id" ] && { swanctl_xappend4 "if_id_in = $if_id" ; swanctl_xappend4 "if_id_out = $if_id" ; }
 	[ -n "$startaction" -a "$startaction" != "none" ] && swanctl_xappend4 "start_action = $startaction"
 	[ -n "$closeaction" -a "$closeaction" != "none" ] && swanctl_xappend4 "close_action = $closeaction"
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -569,7 +569,7 @@ config_connection() {
 	swanctl_xappend2 "children {"
 	config_list_foreach "$1" beet config_beet
 	config_list_foreach "$1" drop config_drop
-	config_list_foreach "$1" drop config_pass
+	config_list_foreach "$1" pass config_pass
 	config_list_foreach "$1" transport config_transport
 	config_list_foreach "$1" transport_proxy config_transport_proxy
 	config_list_foreach "$1" tunnel config_tunnel
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -211,7 +211,7 @@ parse_ike_proposal() {
 }
 
 config_conn() {
-	# Generic ipsec conn section shared by tunnel and transport
+	# Generic swanctl.conf children section shared by tunnel and transport
 	local config_name="$1"
 	local mode="$2"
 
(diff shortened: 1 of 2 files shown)


CodeFetch:
Actually I'm not sure whether a list for local.pubkeys makes sense:
```
Comma separated list of raw public key candidates to use for authentication. The public keys may use a relative path from the swanctl pubkey directory or an absolute path.
Even though multiple local public keys could be defined in principle, only the first public key in the list is used for authentication.
```

CodeFetch:
Referring to https://github.com/openwrt/packages/issues/15929:

I think the local.pubkeys option can be used to implicitly define a corresponding private key to use... But I haven't figured it out, yet.

CodeFetch:
I think the UCI `remote` type should be renamed to `connection`. The old strongswan configuration used by OpenWrt is outdated. I think it's better to do it right now before rewriting the Wiki entries.

net/strongswan/files/swanctl.init
```diff
@@ -148,6 +152,88 @@ is_aead() {
...
+	[ -n "$quoted_comma_list" ] && swanctl_xappend3 "oscp_uris = $quoted_comma_list"
+
+	[ -n "$slot" ] && swanctl_xappend2 "slot = $slot"
+	swanctl_xappend1 "}"
+	swanctl_xappend0 "}"
+
+	swanctl_xappend0 ""
+}
+
+add_ah_proposal() {
```
pprindeville, line 215: Who uses AH (besides Microsoft) and **why**?  ESP provides you authenticity, integrity, *and* confidentiality.

net/strongswan/files/swanctl.init
```diff
@@ -210,38 +296,40 @@ parse_ike_proposal() {
...
-	local local_nat
+	local priority
+	local ah_proposal
+	local esp_proposal
 	local updown
 	local firewall
-	local remote_subnet
-	local lifetime
-	local dpdaction
+	local life_time
```
pprindeville, line 309: I don't want to change the parameter names and break existing configurations.  If that were to happen at all, it would be via its own commit.

net/strongswan/files/swanctl.init
```diff
@@ -275,103 +363,121 @@ config_conn() {
...
-	config_get pre_shared_key "$1" pre_shared_key
 	config_get auth_method "$1" authentication_method
-	config_get local_identifier "$1" local_identifier ""
-	config_get remote_identifier "$1" remote_identifier ""
-	config_get local_sourceip "$1" local_sourceip ""
-	config_get local_leftip "$1" local_leftip "%any"
-	config_get keyingtries "$1" keyingtries "3"
-	config_get dpddelay "$1" dpddelay "30s"
+	config_get dpd_delay "$1" dpd_delay "30s"
+	config_get fragmentation "$1" fragmentation "yes"
```
pprindeville, line 466: Reordering names creates additional churn... Let's preserve order for now.

net/strongswan/files/swanctl.init
```diff
@@ -443,21 +512,62 @@ config_remote() {
 		;;
 	esac
 
+	local ike_proposal
+	parse_ike_proposal "$1" ike_proposal
```
pprindeville, line 516: I don't want to move it here.  That doesn't reflect the fact that `swanctl.conf` can have one IKE definition but multiple child-SA's each with its own ESP parameters.

pprindeville (changes requested):
If you want to reorganize code, do that as one PR.

If you want to introduce new functionality (or change existing functionality, which is problematic in most cases because it breaks compatibility), then that should be another PR.

net/strongswan/files/swanctl.init
```diff
@@ -412,14 +416,24 @@ config_remote() {
 
 	swanctl_xappend2 "local {"
 	swanctl_xappend3 "auth = $auth_method"
-
 	[ -n "$local_identifier" ] && swanctl_xappend3 "id = \"$local_identifier\""
-	[ "$auth_method" = pubkey ] && [ -n "$local_cert" ] && swanctl_xappend3 "certs = $local_cert"
+	if [ "$auth_method" = pubkey ]; then
+		[ -n "$local_cert" ] && swanctl_xappend3 "certs = $local_cert"
+
+		local pubkey_list=""
```
pprindeville: I would just move `pubkey_list` to the top-level scope to avoid duplication.

net/strongswan/files/swanctl.init
```diff
@@ -433,13 +432,6 @@ config_remote() {
 
 	[ "$gateway" = "any" ] && remote_gateway="%any" || remote_gateway="$gateway"
 
-	[ -z "$local_gateway" ] && {
```
pprindeville: Actually, if `local_gateway` doesn't get set anywhere, then this is *always* true so this is far from "dead".

net/strongswan/files/swanctl.init
```diff
@@ -292,11 +292,11 @@ config_child() {
 	local rekeytime
 
 	config_get startaction "$1" startaction "route"
-	config_get local_subnet "$1" local_subnet ""
```
pprindeville, line 231: I'm not gratuitously breaking compatibility by changing names.

net/strongswan/files/swanctl.init
```diff
@@ -359,8 +357,6 @@ config_child() {
 		;;
 	esac
 
-	[ -n "$local_nat" ] && local_ts="$local_nat"
```
pprindeville: People sometimes have to deal with NATting RFC-1918 collisions in their address space for reasons beyond their control, or for legacy reasons, etc. They don't always have the option to do otherwise.

net/strongswan/files/swanctl.init
```diff
@@ -364,7 +388,8 @@ config_child() {
 	[ -n "$if_id" ] && { swanctl_xappend4 "if_id_in = $if_id" ; swanctl_xappend4 "if_id_out = $if_id" ; }
 	[ -n "$startaction" -a "$startaction" != "none" ] && swanctl_xappend4 "start_action = $startaction"
 	[ -n "$closeaction" -a "$closeaction" != "none" ] && swanctl_xappend4 "close_action = $closeaction"
-	swanctl_xappend4 "esp_proposals = $esp_proposal"
+	[ -n "$ah_proposal" ] && swanctl_xappend4 "ah_proposals = $ah_proposal"
```
pprindeville, line 399: Don't you need a check to make sure that one, or the other, is set but not both?

pprindeville (changes requested):

net/strongswan/files/swanctl.init
```diff
@@ -383,8 +379,14 @@ config_child() {
 
 	swanctl_xappend3 "$config_name {"
 
-	[ -n "$local_ts" ] && swanctl_xappend4 "local_ts = $local_ts"
-	[ -n "$remote_ts" ] && swanctl_xappend4 "remote_ts = $remote_ts"
+	local quoted_comma_list=""
+	config_list_foreach "$1" local_ts append_quoted_comma_list
```
pprindeville, line 389: This will break things, but **yes**, we should have supported multiple values.  Although maybe people already were with:

```
     option local_ts     "192.168.1.0/24, 192.168.2.0/24, 192.168.3.0/24"
```

Are quotes strictly needed?

net/strongswan/files/swanctl.init
```diff
@@ -324,6 +326,8 @@ config_child() {
 	parse_ah_proposal "$1" ah_proposal
 	parse_esp_proposal "$1" esp_proposal
 
+	[ -n "$priority" ] && swanctl_xappend4 "priority = $priority"
```
pprindeville, line 331: I would do priority as its own commit (maybe even PR along with other new functionality).

pprindeville (changes requested):

CodeFetch:
@pprindeville I see... I'm doing too many things at once.
What about https://github.com/openwrt/packages/pull/15935 ?
This gives an idea of how the code could be simplified. It could be simplified even more if the OpenWrt config names would reflect the names of the options in strongSwan.

What about introducing a /etc/config/strongswan config which uses the same names as the swanctl.conf? I'm using strongSwan for transport encryption. That's why I'm not happy with the current naming as they are for tunnel interfaces and it looks all wrong...

net/strongswan/files/swanctl.init
CodeFetch: I've tested the pubkeys option with quoting everything like "puba, pubb, pubc", but swanctl didn't accept it. It seems to expect "puba, "pubb", "pubc" instead. I think it is the same for local/remote_ts.

net/strongswan/files/swanctl.init
CodeFetch: From what I have read it is possible to use AH and ESP in parallel with different algorithms. I don't know why people should do this, but I think it's better to reflect the swanctl.conf rather than having users complain in the future and having to add options.

CodeFetch:
Thank you for the review. I'll close this now and do things step-by-step. I'd really appreciate if we could somehow stick more to the swanctl.conf by e.g. adding an additional strongswan.init file which uses /etc/config/strongswan and introduce breaking changes there.

net/strongswan/files/swanctl.init
Thermi: strongSwan does not support AH+ESP bundles. The daemon detects whether ESP or AH should be negotiated for the CHILD_SA by the presence of ah_proposals (if that is defined, it's AH, otherwise it's ESP).

net/strongswan/files/swanctl.init
Thermi: Is the presence/definition of local_nat in the UCI config having any influence on anything but the swanctl.init script? What happens if we just remove it?

net/strongswan/files/swanctl.init
Thermi: Having AH support is just extra sugar now. In the last 8 years I have only seen three configs with AH.

net/strongswan/files/swanctl.init
pprindeville: If it really is the latter case, then we might as well use `list`.

pprindeville:
> @pprindeville I see... I'm doing too many things at once.
> What about #15935 ?
> This gives an idea of how the code could be simplified. It could be simplified even more if the OpenWrt config names would reflect the names of the options in strongSwan.

Keep in mind that the naming and nesting of many variables changed between `ipsec` and `swanctl`. So in some cases I was just trying to keep the naming consistent for forward compatibility. If we chase after the underlying names then we're going to be at the mercy of every upstream change.

> What about introducing a /etc/config/strongswan config which uses the same names as the swanctl.conf? I'm using strongSwan for transport encryption. That's why I'm not happy with the current naming as they are for tunnel interfaces and it looks all wrong...

In that case, we might as well call it `/etc/config/swanctl` if it were that closely tied, but honestly I don't see the point.  In a perfect world, we'd have an abstract `IPsec` configuration that gets mapped to the appropriate Strongswan, or Racoon/KAME, or Wireguard, etc. configuration.

This is what YANG, etc. attempts to do.  Decouple the configuration from the underlying implementation...

CodeFetch:
> This is what YANG, etc. attempts to do. Decouple the configuration from the underlying implementation...

I think this is a wrong approach from usability perspective. I started this whole fiddling, because I didn't expect OpenWrt's strongSwan configuration to be so far away from strongSwan's. It took me an hour to migrate my swanctl config to OpenWrt. That shouldn't be so hard. Furthermore half of the options I need for my transport setup to work are missing.

> In that case, we might as well call it /etc/config/swanctl if it were that closely tied, but honestly I don't see the point.

@pprindeville @Thermi If you give me an ACK for this I'll put work into supporting the strongSwan options 1 to 1 and documenting it in the Wiki. I think strongSwan is clearly the successor of traditional IPSec broker implementations. strongSwan had good reasons to break with traditional configuration and I'd like to make it easier for "future generations" to build IPSec setups in which I see more value than migrating old IPSec configs. To me, who started with strongSwan, it is like first having to learn what old IPSec configs did to see through this, which is a rather frustrating and confusing task as I will never do any legacy IPSec configs.

Thermi:
The question then is, if we actually need the mutability of the UCI config if we just start putting in all the strongSwan settings. At that point we might just as well add sections for including certain files in the swanctl config and enabling/disabling these includes like the tunnels right now.
That's a lot less work than adding all the settings. You might as well just have the user write verbatim swanctl configs.
What's the use of the UCI config then anyway? Programmatically queryable/settable values via bash? That could be done via a helper script in a generalized case. We wouldn't need to write a script to translate UCI settings to a valid swanctl config.

CodeFetch:
@Thermi

With my changes in https://github.com/openwrt/packages/pull/15935 it is quite simple.
Reasons for UCI:
- It's easy to create a LuCi prototype
- Conformant with OpenWrt configuration policy
- We can add checks for options and do additional work like setting up firewall rules automatically

By the way... To get it working at all I needed to add XFRM kernel options:
kmod-xfrm-interface

I haven't tested tunnel mode, but doesn't it need following options?:
IPV6:kmod-ip6-vti
kmod-ip-vti

For AEAD kmod-crypto-aead is needed.
For mod-gcm kmod-crypto-gcm is needed.
For mod-random kmod-crypto-rng is needed.
For mod-chacha kmod-crypto-lib-chacha20poly1305 is needed.
For mod-curve25519 kmod-crypto-lib-curve25519 is needed.

While some options needed for plugins are currently default in OpenWrt, they should also be selected by the corresponding strongswan mods.
Therefore I think the mod-configuration must be refactored...

And for transport I also needed (I don't know why, but charon complained):
kmod-udptunnel4 
IPV6:kmod-udptunnel6

Furthermore there is a problem with mod-socket-dynamic. I needed to manually erase the socket-default library to get it working when selecting mod-socket-dynamic. I think this needs a fix upstream, but for now maybe mod-socket-dynamic should be the default.

Thermi:
@CodeFetch 

* Well, yes, but it's still a lot of work
* Conformity for the sake of conformity is not a good argument TBH
* If we added explicite settings for the relevant details in UCI (or maybe a runtime parser for the loaded/configured VICI config? It's "just" bison and lex.)

From time to time the list of necessary kernel modules on the [wiki](https://wiki.strongswan.org/projects/strongswan/wiki/KernelModules) is updated. But the XFRM interface bit is missing.
So evidently that needs to be updated, too.

>While some options needed for plugins are currently default in OpenWrt, they should also be selected by the corresponding strongswan mods.

OK, true.

> Furthermore there is a problem with mod-socket-dynamic. I needed to manually erase the socket-default library to get it working when selecting mod-socket-dynamic. I think this needs a fix upstream, but for now maybe mod-socket-dynamic should be the default.

How so? You need to disable loading of the -default plugin for the -dynamic one to be loaded. That is not a bug. It's because of plugin priorities.

There's right now no good reason to use -dynamic unless you need to use a dynamic local port (per IKE config, not configured per strongswan.conf).

Thank you for taking the time to figure this out for us. Your work is greatly appreciated.

CodeFetch:
> How so? You need to disable loading of the -default plugin for the -dynamic one to be loaded. That is not a bug. It's because of plugin priorities.

Right... I figured it out, too.

Hm... I still have a more concerning issue. It seems ESP offload doesn't work with OpenWrt for some years now due to:
https://github.com/openwrt/openwrt/commit/b0d7fcdc4951aea64f93600776d8a5a535a2e9cc

I get the following error even though I have included all other needed modules like chacha20-poly1305 and AEAD etc.:
```
daemon.info ipsec: 08[KNL] adding SAD entry with SPI c6ca7a62 and reqid {3}
daemon.info ipsec: 08[KNL]   using encryption algorithm CHACHA20_POLY1305 with key size 288
daemon.info ipsec: 08[KNL]   using replay window of 32 packets
daemon.info ipsec: 08[KNL]   HW offload: no
daemon.info ipsec: 08[KNL] sending XFRM_MSG_UPDSA 210: => 352 bytes @ 0x778a24c4
daemon.info ipsec: 08[KNL]    0: 00 00 01 60 00 1A 00 05 00 00 00 D2 00 00 06 86  ...`............
daemon.info ipsec: 08[KNL]   16: C0 A8 00 C6 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]   32: C0 A8 00 12 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]   48: 00 00 00 00 00 00 00 00 00 02 20 20 00 00 00 00  ..........  ....
daemon.info ipsec: 08[KNL]   64: 00 00 00 00 00 00 00 00 C0 A8 00 C6 00 00 00 00  ................
daemon.info ipsec: 08[KNL]   80: 00 00 00 00 00 00 00 00 C6 CA 7A 62 32 00 00 00  ..........zb2...
daemon.info ipsec: 08[KNL]   96: C0 A8 00 12 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]  112: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
daemon.info ipsec: 08[KNL]  128: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
daemon.info ipsec: 08[KNL]  144: 00 00 00 00 00 00 0D 4A 00 00 00 00 00 00 0F 78  .......J.......x
daemon.info ipsec: 08[KNL]  160: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]  176: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]  192: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]  208: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]  224: 00 00 00 03 00 02 00 20 00 00 00 00 00 00 00 00  ....... ........
daemon.info ipsec: 08[KNL]  240: 00 70 00 12 72 66 63 37 35 33 39 65 73 70 28 63  .p..rfc7539esp(c
daemon.info ipsec: 08[KNL]  256: 68 61 63 68 61 32 30 2C 70 6F 6C 79 31 33 30 35  hacha20,poly1305
daemon.info ipsec: 08[KNL]  272: 29 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  )...............
daemon.info ipsec: 08[KNL]  288: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]  304: 00 00 00 00 00 00 01 20 00 00 00 80 AE A7 0B 14  ....... ........
daemon.info ipsec: 08[KNL]  320: 55 F4 23 9A 76 24 6C 64 C1 D1 A9 22 13 F7 4E 53  U.#.v$ld..."..NS
daemon.info ipsec: 08[KNL]  336: 3D 31 FE 29 E1 5E A3 D6 3E 87 E3 7D 39 60 52 13  =1.).^..>..}9`R.
daemon.info ipsec: 08[KNL] received (2) 210: => 372 bytes @ 0x023e48a0
daemon.info ipsec: 08[KNL]    0: 00 00 01 74 00 02 00 00 00 00 00 D2 00 00 06 86  ...t............
daemon.info ipsec: 08[KNL]   16: FF FF FF A7 00 00 01 60 00 1A 00 05 00 00 00 D2  .......`........
daemon.info ipsec: 08[KNL]   32: 00 00 06 86 C0 A8 00 C6 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]   48: 00 00 00 00 C0 A8 00 12 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]   64: 00 00 00 00 00 00 00 00 00 00 00 00 00 02 20 20  ..............
daemon.info ipsec: 08[KNL]   80: 00 00 00 00 00 00 00 00 00 00 00 00 C0 A8 00 C6  ................
daemon.info ipsec: 08[KNL]   96: 00 00 00 00 00 00 00 00 00 00 00 00 C6 CA 7A 62  ..............zb
daemon.info ipsec: 08[KNL]  112: 32 00 00 00 C0 A8 00 12 00 00 00 00 00 00 00 00  2...............
daemon.info ipsec: 08[KNL]  128: 00 00 00 00 FF FF FF FF FF FF FF FF FF FF FF FF  ................
daemon.info ipsec: 08[KNL]  144: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
daemon.info ipsec: 08[KNL]  160: FF FF FF FF 00 00 00 00 00 00 0D 4A 00 00 00 00  ...........J....
daemon.info ipsec: 08[KNL]  176: 00 00 0F 78 00 00 00 00 00 00 00 00 00 00 00 00  ...x............
daemon.info ipsec: 08[KNL]  192: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]  208: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]  224: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]  240: 00 00 00 00 00 00 00 03 00 02 00 20 00 00 00 00  ........... ....
daemon.info ipsec: 08[KNL]  256: 00 00 00 00 00 70 00 12 72 66 63 37 35 33 39 65  .....p..rfc7539e
daemon.info ipsec: 08[KNL]  272: 73 70 28 63 68 61 63 68 61 32 30 2C 70 6F 6C 79  sp(chacha20,poly
daemon.info ipsec: 08[KNL]  288: 31 33 30 35 29 00 00 00 00 00 00 00 00 00 00 00  1305)...........
daemon.info ipsec: 08[KNL]  304: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
daemon.info ipsec: 08[KNL]  320: 00 00 00 00 00 00 00 00 00 00 01 20 00 00 00 80  ........... ....
daemon.info ipsec: 08[KNL]  336: AE A7 0B 14 55 F4 23 9A 76 24 6C 64 C1 D1 A9 22  ....U.#.v$ld..."
daemon.info ipsec: 08[KNL]  352: 13 F7 4E 53 3D 31 FE 29 E1 5E A3 D6 3E 87 E3 7D  ..NS=1.).^..>..}
daemon.info ipsec: 08[KNL]  368: 39 60 52 13                                      9`R.
daemon.info ipsec: 08[KNL] received netlink error: Function not implemented (89)
daemon.info ipsec: 08[KNL] unable to add SAD entry with SPI c6ca7a62 (FAILED)
```

But I'm tired now. Maybe hauke had a reason. So next step for me is looking through required kernel options to get ESP offload running. Otherwise IPSec is pretty lame compared to WireGuard.
https://wiki.strongswan.org/projects/strongswan/wiki/KernelModules

Thermi:
Well, do you have a crypto driver for that proposal in `/proc/crypto`?

CodeFetch:
@Thermi 
Yes:
```
name         : poly1305
driver       : poly1305-mips
module       : poly1305_mips
priority     : 200
refcnt       : 1
selftest     : passed
internal     : no
type         : shash
blocksize    : 16
digestsize   : 16

name         : xchacha12
driver       : xchacha12-mips
module       : chacha_mips
priority     : 200
refcnt       : 1
selftest     : passed
internal     : no
type         : skcipher
async        : no
blocksize    : 1
min keysize  : 32
max keysize  : 32
ivsize       : 32
chunksize    : 64
walksize     : 64

name         : xchacha20
driver       : xchacha20-mips
module       : chacha_mips
priority     : 200
refcnt       : 1
selftest     : passed
internal     : no
type         : skcipher
async        : no
blocksize    : 1
min keysize  : 32
max keysize  : 32
ivsize       : 32
chunksize    : 64
walksize     : 64

name         : chacha20
driver       : chacha20-mips
module       : chacha_mips
priority     : 200
refcnt       : 1
selftest     : passed
internal     : no
type         : skcipher
async        : no
blocksize    : 1
min keysize  : 32
max keysize  : 32
ivsize       : 16
chunksize    : 64
walksize     : 64

```

From my understanding when ESP offloading doesn't work, there is a fallback to userland encryption. But this gives a huge performance penalty due to context-switching. What I expect is ~40 Mbit/s with chacha20 on ath79 devices and ~300 Mbit/s on ipq40xx (quad core ARM v8).

CodeFetch:
Hm... I was wrong. XFRM_OFFLOAD is only for HW offloading.

pprindeville:
> To me, who started with strongSwan, it is like first having to learn what old IPSec configs did to see through this, which is a rather frustrating and confusing task as I will never do any legacy IPSec configs.

If that's the case, then why not hand-edit `/etc/swanctl/swanctl.ctl` directly and use that?

net/strongswan/files/swanctl.init
pprindeville: Why remove it?  I thought we were trying to expose as many underlying swanctl controls as possible for completeness...

net/strongswan/files/swanctl.init
```diff
@@ -314,6 +315,7 @@ config_child() {
 	local if_id
 	local rekey_time
 
+	config_get interface "$1" interface ""
```
pprindeville, line 326: Can this be one interface or many?  If many, should it be a list?

Thermi:
> Hm... I was wrong. XFRM_OFFLOAD is only for HW offloading.

Yep.

In /proc/crypto, there need to be specifics for ESP or the RFC for the encapsulation with the algorithms you want to use.
That can look like that (example from a working machine): `rfc7539esp(chacha20,poly1305)`

CodeFetch:
@Thermi Can you please post an output of lsmod on the working machine?
I've added all somehow related kernel modules, but the entry is still missing:
```
aead                    4057 10 mac80211,esp6,esp4,algif_aead,seqiv,gcm,echainiv,chacha20poly1305,ccm,authenc
af_alg                 11663  4 algif_skcipher,algif_rng,algif_hash,algif_aead
af_key                 27554  0 
ah4                     5010  0 
ah6                     5074  0 
algif_aead              3730  0 
algif_hash              4306  0 
algif_rng               1202  0 
algif_skcipher          2738  0 
ath                    19750  4 ath9k,ath9k_common,ath9k_hw,ath10k_core
ath10k_core           451169  1 ath10k_pci
ath10k_pci             38066  0 
ath9k                 106171  0 
ath9k_common           12995  1 ath9k
ath9k_hw              355716  2 ath9k,ath9k_common
authenc                 3806  0 
cbc                     1618  0 
ccm                     8274  0 
cfg80211              287311  5 ath9k,ath9k_common,ath10k_core,ath,mac80211
chacha_generic          2962  0 
chacha_mips             5883  1 libchacha20poly1305
chacha20poly1305        5778  0 
cmac                    2354  0 
compat                  1346  4 ath9k,ath9k_common,mac80211,cfg80211
crc_ccitt               1774  1 ppp_async
crypto_acompress        4155  1 deflate
crypto_hash            10424 19 mac80211,poly1305_mips,ah6,ah4,xfrm_algo,algif_hash,sha256_generic,sha1_generic,drbg,poly1305_generic,md5,hmac,ghash_generic,gcm,cmac,chacha20poly1305,ccm,authenc,crypto_null
crypto_null             3238  4 algif_aead,gcm,authenc,aead
crypto_user             4468  0 
cryptomgr               2140  0 
ctr                     2770  0 
deflate                 3474  0 
des_generic             1682  0 
drbg                   17842  0 
echainiv                1554  0 
esp4                    8245  0 
esp6                    7992  0 
gcm                    10114  0 
gf128mul                6409  1 ghash_generic
ghash_generic           1938  0 
gpio_button_hotplug     7058  0 
hmac                    2642  0 
ip_tables              10307  4 iptable_nat,iptable_mangle,iptable_filter
ip6_tables             10215 24 ip6table_mangle,ip6table_filter
ip6_udp_tunnel          1850  1 wireguard
ip6t_REJECT             1298  2 
ip6table_filter          978  1 
ip6table_mangle         1458  1 
ipcomp                  1714  0 
ipcomp6                 1938  0 
ipt_REJECT              1234  2 
ipt_ah                   786  0 
iptable_filter           946  1 
iptable_mangle          1138  1 
iptable_nat             1138  1 
jitterentropy_rng       6121  0 
libblake2s              1938  1 wireguard
libblake2s_generic      5980  1 libblake2s
libchacha                760  1 chacha_generic
libchacha20poly1305     4614  1 wireguard
libcurve25519_generic   10486  1 wireguard
libdes                 18389  1 des_generic
libpoly1305             2810  1 poly1305_generic
libsha256               8428  1 sha256_generic
mac80211              539883  2 ath9k,ath10k_core
md5                     5010  0 
nf_conntrack           64807  7 xt_state,xt_nat,xt_conntrack,xt_REDIRECT,xt_MASQUERADE,nf_nat,nf_flow_table
nf_defrag_ipv4          1209  1 nf_conntrack
nf_defrag_ipv6          5524  1 nf_conntrack
nf_flow_table          15763  2 xt_FLOWOFFLOAD,nf_flow_table_hw
nf_flow_table_hw        2322  1 
nf_log_common           2936  2 nf_log_ipv4,nf_log_ipv6
nf_log_ipv4             3602  0 
nf_log_ipv6             4306  0 
nf_nat                 21277  4 iptable_nat,xt_nat,xt_REDIRECT,xt_MASQUERADE
nf_reject_ipv4          2313  1 ipt_REJECT
nf_reject_ipv6          2670  1 ip6t_REJECT
poly1305_generic        1810  0 
poly1305_mips           3695  1 libchacha20poly1305
ppp_async               7122  0 
ppp_generic            24335  3 pppoe,ppp_async,pppox
pppoe                   8690  0 
pppox                   1488  1 pppoe
seqiv                   1746  0 
sha1_generic            2393  0 
sha256_generic          2397  0 
slhc                    5378  1 ppp_generic
tunnel4                 2082  1 xfrm4_tunnel
tunnel6                 2178  1 xfrm6_tunnel
udp_tunnel              2969  1 wireguard
wireguard              52455  0 
x_tables               16679 26 xt_state,xt_nat,xt_conntrack,xt_REDIRECT,xt_MASQUERADE,xt_FLOWOFFLOAD,ipt_REJECT,xt_time,xt_tcpudp,xt_policy,xt_multiport,xt_mark,xt_mac,xt_limit,xt_esp,xt_comment,xt_TCPMSS,xt_LOG,iptable_mangle,iptable_filter,ipt_ah,ip_tables,ip6table_mangle,ip6table_filter,ip6_tables,ip6t_REJECT
xfrm_algo               4034  7 esp6,ah6,esp4,ah4,xfrm_user,xfrm_ipcomp,af_key
xfrm_interface          6962  0 
xfrm_ipcomp             2978  2 ipcomp6,ipcomp
xfrm_user              25042  2 
xfrm4_tunnel            1106  0 
xfrm6_tunnel            3011  1 ipcomp6
xt_FLOWOFFLOAD          3282  0 
xt_LOG                   818  0 
xt_MASQUERADE            882  1 
xt_REDIRECT              754  0 
xt_TCPMSS               3218  4 
xt_comment               562133 
xt_conntrack            2482 14 
xt_esp                   882  0 
xt_limit                1394 20 
xt_mac                   690  0 
xt_mark                  722  0 
xt_multiport            1362  0 
xt_nat                  2482  0 
xt_policy               2418  0 
xt_state                 882  0 
xt_tcpudp               1842 18 
xt_time                 1970  0 
zlib_deflate           19876  1 deflate
zlib_inflate           14060  1 deflate
```

CodeFetch:
It works now... Seems it needed to be triggered by strongSwan.

```
root@OpenWrt:~# cat /proc/crypto 
name         : hmac(sha256)
driver       : hmac(sha256-generic)
module       : hmac
priority     : 100
refcnt       : 2
selftest     : passed
internal     : no
type         : shash
blocksize    : 64
digestsize   : 32

name         : seqiv(rfc7539esp(chacha20,poly1305))
driver       : seqiv(rfc7539esp(chacha20-mips,poly1305-mips))
module       : seqiv
priority     : 200
refcnt       : 5
selftest     : passed
internal     : no
type         : aead
async        : no
blocksize    : 1
ivsize       : 8
maxauthsize  : 16
geniv        : <none>

name         : rfc7539esp(chacha20,poly1305)
driver       : rfc7539esp(chacha20-mips,poly1305-mips)
module       : chacha20poly1305
priority     : 200
refcnt       : 5
selftest     : passed
internal     : no
type         : aead
async        : no
blocksize    : 1
ivsize       : 8
maxauthsize  : 16
geniv        : <none>

name         : poly1305
driver       : poly1305-mips
module       : poly1305_mips
priority     : 200
refcnt       : 9
selftest     : passed
internal     : no
type         : shash
blocksize    : 16
digestsize   : 16

name         : xchacha12
driver       : xchacha12-mips
module       : chacha_mips
priority     : 200
refcnt       : 1
selftest     : passed
internal     : no
type         : skcipher
async        : no
blocksize    : 1
min keysize  : 32
max keysize  : 32
ivsize       : 32
chunksize    : 64
walksize     : 64

name         : xchacha20
driver       : xchacha20-mips
module       : chacha_mips
priority     : 200
refcnt       : 1
selftest     : passed
internal     : no
type         : skcipher
async        : no
blocksize    : 1
min keysize  : 32
max keysize  : 32
ivsize       : 32
chunksize    : 64
walksize     : 64

name         : chacha20
driver       : chacha20-mips
module       : chacha_mips
priority     : 200
refcnt       : 5
selftest     : passed
internal     : no
type         : skcipher
async        : no
blocksize    : 1
min keysize  : 32
max keysize  : 32
ivsize       : 16
chunksize    : 64
walksize     : 64

name         : sha224
driver       : sha224-generic
module       : sha256_generic
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : shash
blocksize    : 64
digestsize   : 28

name         : sha256
driver       : sha256-generic
module       : sha256_generic
priority     : 100
refcnt       : 2
selftest     : passed
internal     : no
type         : shash
blocksize    : 64
digestsize   : 32

name         : sha1
driver       : sha1-generic
module       : sha1_generic
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : shash
blocksize    : 64
digestsize   : 20

name         : jitterentropy_rng
driver       : jitterentropy_rng
module       : jitterentropy_rng
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : rng
seedsize     : 0

name         : stdrng
driver       : drbg_nopr_hmac_sha256
module       : drbg
priority     : 207
refcnt       : 2
selftest     : passed
internal     : no
type         : rng
seedsize     : 0

name         : stdrng
driver       : drbg_nopr_hmac_sha512
module       : drbg
priority     : 206
refcnt       : 1
selftest     : passed
internal     : no
type         : rng
seedsize     : 0

name         : stdrng
driver       : drbg_nopr_hmac_sha384
module       : drbg
priority     : 205
refcnt       : 1
selftest     : passed
internal     : no
type         : rng
seedsize     : 0

name         : stdrng
driver       : drbg_nopr_hmac_sha1
module       : drbg
priority     : 204
refcnt       : 1
selftest     : passed
internal     : no
type         : rng
seedsize     : 0

name         : stdrng
driver       : drbg_pr_hmac_sha256
module       : drbg
priority     : 203
refcnt       : 1
selftest     : passed
internal     : no
type         : rng
seedsize     : 0

name         : stdrng
driver       : drbg_pr_hmac_sha512
module       : drbg
priority     : 202
refcnt       : 1
selftest     : passed
internal     : no
type         : rng
seedsize     : 0

name         : stdrng
driver       : drbg_pr_hmac_sha384
module       : drbg
priority     : 201
refcnt       : 1
selftest     : passed
internal     : no
type         : rng
seedsize     : 0

name         : stdrng
driver       : drbg_pr_hmac_sha1
module       : drbg
priority     : 200
refcnt       : 1
selftest     : passed
internal     : no
type         : rng
seedsize     : 0

name         : poly1305
driver       : poly1305-generic
module       : poly1305_generic
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : shash
blocksize    : 16
digestsize   : 16

name         : md5
driver       : md5-generic
module       : md5
priority     : 0
refcnt       : 1
selftest     : passed
internal     : no
type         : shash
blocksize    : 64
digestsize   : 16

name         : ghash
driver       : ghash-generic
module       : ghash_generic
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : shash
blocksize    : 16
digestsize   : 16

name         : des3_ede
driver       : des3_ede-generic
module       : des_generic
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : cipher
blocksize    : 8
min keysize  : 24
max keysize  : 24

name         : des
driver       : des-generic
module       : des_generic
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : cipher
blocksize    : 8
min keysize  : 8
max keysize  : 8

name         : zlib-deflate
driver       : zlib-deflate-scomp
module       : deflate
priority     : 0
refcnt       : 1
selftest     : passed
internal     : no
type         : scomp

name         : deflate
driver       : deflate-scomp
module       : deflate
priority     : 0
refcnt       : 1
selftest     : passed
internal     : no
type         : scomp

name         : deflate
driver       : deflate-generic
module       : deflate
priority     : 0
refcnt       : 1
selftest     : passed
internal     : no
type         : compression

name         : xchacha12
driver       : xchacha12-generic
module       : chacha_generic
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : skcipher
async        : no
blocksize    : 1
min keysize  : 32
max keysize  : 32
ivsize       : 32
chunksize    : 64
walksize     : 64

name         : xchacha20
driver       : xchacha20-generic
module       : chacha_generic
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : skcipher
async        : no
blocksize    : 1
min keysize  : 32
max keysize  : 32
ivsize       : 32
chunksize    : 64
walksize     : 64

name         : chacha20
driver       : chacha20-generic
module       : chacha_generic
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : skcipher
async        : no
blocksize    : 1
min keysize  : 32
max keysize  : 32
ivsize       : 16
chunksize    : 64
walksize     : 64

name         : ecb(cipher_null)
driver       : ecb-cipher_null
module       : crypto_null
priority     : 100
refcnt       : 2
selftest     : passed
internal     : no
type         : skcipher
async        : no
blocksize    : 1
min keysize  : 0
max keysize  : 0
ivsize       : 0
chunksize    : 1
walksize     : 1

name         : digest_null
driver       : digest_null-generic
module       : crypto_null
priority     : 0
refcnt       : 1
selftest     : passed
internal     : no
type         : shash
blocksize    : 1
digestsize   : 0

name         : compress_null
driver       : compress_null-generic
module       : crypto_null
priority     : 0
refcnt       : 1
selftest     : passed
internal     : no
type         : compression

name         : cipher_null
driver       : cipher_null-generic
module       : crypto_null
priority     : 0
refcnt       : 1
selftest     : passed
internal     : no
type         : cipher
blocksize    : 1
min keysize  : 0
max keysize  : 0

name         : aes
driver       : aes-generic
module       : kernel
priority     : 100
refcnt       : 1
selftest     : passed
internal     : no
type         : cipher
blocksize    : 16
min keysize  : 16
max keysize  : 32


```

pprindeville:
> It works now... Seems it needed to be triggered by strongSwan.

Triggered how?

Thermi:
Probably done automatically by XFRM when the XFRM states are added.

net/strongswan/files/swanctl.init
```diff
@@ -326,6 +328,7 @@ config_child() {
 	parse_ah_proposal "$1" ah_proposal
 	parse_esp_proposal "$1" esp_proposal
 
+	[ $ipcomp -eq 1 ] && swanctl_xappend4 "ipcomp = 1"
```
pprindeville: Shouldn't this be `3`, and the next line as well?

CodeFetch:
Oh it seems I was right in the beginning with ESP offload. It's not only for HW crypto, but enables GRO :+1:...

Thermi:
How did you come to _that_ conclusion?

CodeFetch:
@Thermi Page 28
https://libreswan.org/wiki/images/e/e0/Netdev-0x12-ipsec-flow.pdf

https://github.com/torvalds/linux/commit/7785bba299a8dc8fe8390a0183dad3cafb3f1d80

Thermi:
XFRM_OFFLOAD is _not_ INET_ESP_OFFLOAD.

XFRM_OFFLOAD is the generic support for offloading of any XFRM operation (even GRO (Generic Receive Offload in software here) /and detecting if that was done/. That's why INET_ESP_OFFLOAD hard enables XFRM_OFFLOAD in Kconfig. It needs to detect if any of the IPsec stuff was already done in hardware.

The reverse is not true. If you do not have any hardware capable of offloading off XFRM operations (ESP/AH transformation in the NIC for example), then enabling XFRM_OFFLOAD does exactly nothing for you because it does not enable INET_ESP_OFFLOAD (which you require for GRO in software).

CodeFetch:
@Thermi
I was first referring to ESP offload

> It seems ESP offload doesn't work with OpenWrt for some years now due to:

and then I was mixing it up with XFRM offloading, because of strongSwan's documentation which only mentions ESP offloading for use in XFRM offloading...

> Hm... I was wrong. XFRM_OFFLOAD is only for HW offloading.

But from what I see the most distributions ship the ESP_OFFLOAD module, but don't load it by default. So I wonder if there are any disadvantages.

Thermi:
I see.

That's the kernel behaviour. Distros fixing that would either need to patch the kernel to load the module automatically for XFRM, try to build it in statically (didn't check if possible), or add a record for it in modules-load.d.
Pick your poison. :)

I am not aware of any drawbacks as of this point in time.

Kind regards
Noel

PR #15925, opened 2021-06-21
liburcu: update to version 0.13.0

Payne-X6:

Description:

- update of the library on which `knot` depends on


Payne-X6:
Problems during the build

PR #15928, opened 2021-06-21
samba4: remove domain if fqdn was set

fda77:
My fqdn `openwrt.testarea.separeatednet.internallan.local` looks ugly in windows

### DIFF
--- a/net/samba4/files/samba.init
+++ b/net/samba4/files/samba.init
@@ -25,7 +25,7 @@ smb_header() {
 	)
 
 	# we dont use netbios anymore as default and wsd/avahi is dns based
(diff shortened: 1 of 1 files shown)


Andy2244:
Looks trivial enough, but is the machine listed/reachable via wsdd2 in explorer?

fda77:
I have not much free ram and thus removed wsdd2 package (sadly it is not optional and cant be disabled in kconfig)
But the FQDN which is visible in the windows workgroup (with domain) was not accessible by click on it. Manually opening "\\openwrt" worked
(i know i could edit the samba template with a custom hostname as a workaround)

OT: Syslog shows also the complete FQDN in every single line. All non-openwrt devices shows use only without domain part

neheb:
Needs a rebase

PR #15915, opened 2021-06-19
[RFC] mariadb: update to 10.6.1

neheb:
Maintainer: @miska 
Compile tested: ath79

A bunch of things changed with the package. I'm not sure that the changes here are completely correct.

PR #15937, opened 2021-06-23
libmariadb: fix compilation cmake issue

feckert:
Maintainer: @miska @neheb @pprindeville
Compile tested: mips_24kc, lantiq_xrx200, latest OpenWrt master)
Run tested: only compilation fix

Description:
This patch fixes the compilation issue of the upstream Cmake file.
If this patch is not applied we get the following compilation error output for this package.
```
CMake Error at cmake/ConnectorName.cmake:30 (ENDMACRO):
  Flow control statements are not properly nested.
Call Stack (most recent call first):
  CMakeLists.txt:423 (INCLUDE)
```

The blamed cmake/ConnectorName.cmake file gets fixed with this patch.

### DIFF
--- a/libs/libmariadb/patches/0001-libmariadb-fix-cmake.patch
+++ b/libs/libmariadb/patches/0001-libmariadb-fix-cmake.patch
@@ -1,0 +1,11 @@
+--- a/cmake/ConnectorName.cmake
++++ b/cmake/ConnectorName.cmake
+@@ -22,7 +22,7 @@ IF(CMAKE_SYSTEM_NAME MATCHES "Windows")
+     SET(MACHINE_NAME "x64")
+   ELSE()
+     SET(MACHINE_NAME "32")
+-  END()
++  ENDIF()
+ ENDIF()
+ 
+ SET(product_name "mysql-connector-c-${CPACK_PACKAGE_VERSION}-${PLATFORM_NAME}${CONCAT_SIGN}${MACHINE_NAME}")


neheb:
What an issue...

feckert:
> What an issue...

@neheb If I do not add this patch, I can not compile this package.
As mention in my summary I get the following error during compilation on my target.

```
CMake Error at cmake/ConnectorName.cmake:30 (ENDMACRO):
  Flow control statements are not properly nested.
Call Stack (most recent call first):
  CMakeLists.txt:423 (INCLUDE)
```
That´s because the CMAKE statement is wrong.

neheb:
I know. It's just interesting that came allowed this behavior in the first place.

feckert:
@neheb can this be because we have compiled this package with [ninja](https://github.com/openwrt/packages/commit/2e654b19f38f1f941c70a5adcdac6311234362c3#diff-ee330474014bc1944faa85418c85d0d15ed4977eae770232e7dade4e9f66b3b5) in the meantime? Maybe it also has something to do with the Cmake version. Anyway, if we don't patch this, it won't work for me with a clean new build.

I'm just wondering why the openwrt buildbot didn't fail there too.

neheb:
It's because of the version update to cmake, not Ninja. See: https://github.com/openwrt/openwrt/pull/4285

Also see https://github.com/openwrt/packages/pull/15915

buildbot is failing on this package: https://downloads.openwrt.org/snapshots/faillogs/arc_arc700/packages/libmariadb/compile.txt

feckert:
@neheb Should we merge this? Or should we wait till the #15915 gets merged?

neheb:
ping @BKPepe

micmac1:
Hi all,

This is the correct fix. The new versions introduced in #15915 are beta versions (both lib and server). I think merging this fix may be preferable. The same fix should be done in mariadb because the included lib there will have the same issue.

Kind regards,
Seb

neheb:
Merging. Hopefully I don’t get pushback…

PR #15945, opened 2021-06-24
tvheadend: fix compilation with GCC 9 and 10

neheb:
Without this, compilation fails because of a fortify-source header.
There's something defining and undefining _GNU_SOURCE somewhere.

Maintainer: @sairon 
Compile tested: ath79


neheb:
The actual issue is GNU libiconv defining and undefining _GNU_SOURCE internally, causing havoc with fortify-source headers.

PR #15973, opened 2021-06-26
strongswan: miscellaneous cleanup of config file, etc

pprindeville:
Maintainer: me, @Thermi
Compile tested: x86_64, generic, HEAD (db9784bedd)

Description:

Forbid absolute/relative paths for certificate and key files.  Get rid of last remnants of left/right naming.  Factor invariant code paths.  Drop redundant `secrets.rsa.filename` section.

Also add a couple of additional new settings.


pprindeville:
@CodeFetch Have a look.

pprindeville:
Contains some of the fixes from PR #15922.

net/strongswan/files/swanctl.init
```diff
@@ -299,12 +301,23 @@ config_child() {
 		;;
 	esac
 
+	case "$hw_offload" in
+	yes|auto|"")
```
Thermi, line 307: Add "no".

libs/libzip/Makefile
```diff
@@ -6,12 +6,12 @@
 include $(TOPDIR)/rules.mk
```
Thermi, line 6: This shouldn't be in this PR at all.

net/strongswan/files/swanctl.init
```diff
@@ -219,6 +219,7 @@ config_child() {
 	local mode="$2"
```
Thermi, line 219: The description of the commit is not correct.

The "connections.<conn>.children.<child>.interface" key in swanctl sets the interface of the policy. It does not influence routing.

This means that the policy will have the additional criteria of the outgoing interface being the value of this setting.
Means, you can potentially have different policies (and thus states) for different outgoing interfaces. That can enable routing of traffic going out of different interfaces to different peers.

Thermi (changes requested):

net/strongswan/files/swanctl.init
pprindeville: Fixed.

libs/libzip/Makefile
pprindeville: Doh. Not sure how that got there.  Rebased.

net/strongswan/files/swanctl.init
pprindeville: Fixed.

PR #15977, opened 2021-06-26
qtbase: add package

yangfl:
Signed-off-by: David Yang <mmyangfl@gmail.com>

Maintainer: me
Compile tested: 19.07 arm

### DIFF
--- a/libs/qtbase/Config.in
+++ b/libs/qtbase/Config.in
@@ -1,0 +1,37 @@
+choice
+	prompt "Build Variant"
+	default QT_BUILD_RELEASE
+	help
+		Choose which Qt variant should be built:
+		-> Release: Build Qt with debugging turned off.
+		-> Debug: Build Qt with debugging turned on.
+
+	config QT_BUILD_RELEASE
+		bool "Release"
+
+	config QT_BUILD_DEBUG
+		bool "Debug"
+endchoice
+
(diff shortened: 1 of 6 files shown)


BKPepe:
Hello @yangfl,

I see that qt{4,5} including your package qt5base, which you are trying to add are present in the [video feed](https://github.com/openwrt/video/tree/master/frameworks) under openwrt organization. 

Because of that, I am closing it.

yangfl:
@BKPepe Thanks for mentioning. But this package of qt5base contains only non-gui libraries and is meant for other programs (qbittorrent). Is that OK to add other packages depending on qt into packages feed without adding qt?

BKPepe:
Then you can try to improve the package there or discuss with the maintainer what's the proper solution for it. We can not add packages here, which will depend on external feeds. However, I can see that in the video feed is [something](https://github.com/openwrt/video/blob/1f37050afa6e8a60c490bfc470f866070ada4aee/frameworks/qt5base/Makefile#L108) related to GUI, but I didn't check if it suits your case because it does not make sense to have two similar packages that do the same and will be named differently.

mirko:
I'm not willing to have qt5base in the packages feed *and* qt5base in the video feed.
I'm willing to discuss moving the qt5base package into the packages feed but let's please agree on one ticket to discuss that.

PR #15983, opened 2021-06-27
ieee8021x: add package

yangfl:
Signed-off-by: David Yang <mmyangfl@gmail.com>

Maintainer: me 
Run tested: 21.02 x86


net/ieee8021x/files/ieee8021x.sh
```diff
@@ -0,0 +1,182 @@
...
+	json_get_vars eap_server eap_user_file ca_cert server_cert private_key private_key_passwd server_id
+
+	local eapserver_cfg
+	[ "${eap_server}" = "1" ] && eapserver_cfg="eap_server=1
+${eap_user_file:+eap_user_file=${eap_user_file}}
+${ca_cert:+ca_cert=${ca_cert}}
+${server_cert:+server_cert=${server_cert}}
+${private_key:+private_key=${private_key}}
+${private_key_passwd:+private_key_passwd=${private_key_passwd}}
+${server_id:+server_id=${server_id}}"
```
feckert, line 51: Observe indentation

net/ieee8021x/files/ieee8021x.sh
```diff
@@ -0,0 +1,182 @@
...
+auth_server_addr=${auth_server}
+auth_server_port=${auth_port:1812}
+auth_server_shared_secret=${auth_secret}"
+		[ -n "${acct_server}" ] && radius_cfg="${radius_cfg}
+acct_server_addr=${acct_server}
+acct_server_port=${acct_port:1812}
+acct_server_shared_secret=${acct_secret}"
+		[ -n "$dae_client" -a -n "$dae_secret" ] && radius_cfg="${radius_cfg}
+radius_das_port=${dae_port:3799}
+radius_das_client=${dae_client} ${dae_secret}"
```
feckert, line 75: Observe indentation

net/ieee8021x/files/ieee8021x.sh
```diff
@@ -0,0 +1,182 @@
...
+	[ "${dynamic_vlan}" = "0" ] && dynamic_vlan=
+	[ "${per_sta_vif}" = "0" ] && per_sta_vif=
+	[ -n "${vlan_tagged_interface}" -a -z "${vlan_naming}" ] && vlan_naming=0
+	radius_cfg="${radius_cfg}
+${dynamic_vlan:+dynamic_vlan=1}
+${per_sta_vif:+per_sta_vif=1}
+${vlan_file:+vlan_file=${vlan_file}}
+${vlan_tagged_interface:+vlan_tagged_interface=${vlan_tagged_interface}}
+${vlan_bridge:+vlan_bridge=${vlan_bridge}}
+${vlan_naming:+vlan_naming=${vlan_naming}}"
```
feckert, line 87: Observe indentation

net/ieee8021x/files/ieee8021x.sh
```diff
@@ -0,0 +1,182 @@
...
+acct_server_port=${acct_port:1812}
+acct_server_shared_secret=${acct_secret}"
+		[ -n "$dae_client" -a -n "$dae_secret" ] && radius_cfg="${radius_cfg}
+radius_das_port=${dae_port:3799}
+radius_das_client=${dae_client} ${dae_secret}"
+	fi
+
+	[ "${dynamic_vlan}" = "0" ] && dynamic_vlan=
+	[ "${per_sta_vif}" = "0" ] && per_sta_vif=
+	[ -n "${vlan_tagged_interface}" -a -z "${vlan_naming}" ] && vlan_naming=0
```
feckert, line 80: please use `&&` instead of `-a`

net/ieee8021x/files/ieee8021x.sh
```diff
@@ -0,0 +1,182 @@
...
+	[ -z "${use_pae_group_addr}" ] && use_pae_group_addr=1
+	[ "${erp_send_reauth_start}" = "0" ] && erp_send_reauth_start=
+	local ieee8021x_cfg="ieee8021x=1
+${eapol_version:+eapol_version=${eapol_version}}
+${eap_message:+eap_message=${eap_message}}
+${eapol_key_index_workaround:+eapol_key_index_workaround=1}
+${eap_reauth_period:+eap_reauth_period=${eap_reauth_period}}
+${use_pae_group_addr:+use_pae_group_addr=1}
+${erp_send_reauth_start:+erp_send_reauth_start=1}
+${erp_domain:+erp_domain=${erp_domain}}"
```
feckert, line 106: Observe indentation

net/ieee8021x/files/ieee8021xclient.sh
```diff
@@ -0,0 +1,118 @@
...
+	local phase1 phase2 ca_cert2 client_cert2 private_key2 private_key_passwd2 dh_file2 subject_match2
+	local eap_workaround
+	json_get_vars eapol_version
+	json_get_vars identity anonymous_identity password
+	json_get_vars ca_cert client_cert private_key private_key_passwd dh_file subject_match
+	json_get_vars phase1 phase2 ca_cert2 client_cert2 private_key2 private_key_passwd2 dh_file2 subject_match2
+	json_get_vars eap_workaround
+
+	# launch
+	cat > "${_config}" << EOF
```
feckert, line 51: What do we do if the interface is not there. And therefore no configuration can be written. 

We should return an error if [this file](https://github.com/openwrt/packages/pull/15983/files#diff-96f9a3e11e1c3382cb7bf3844ee1d26e86577c2a080bae547c403a1bf4313b9eR29) is not there

net/ieee8021x/files/ieee8021x.sh
```diff
@@ -0,0 +1,182 @@
...
+		5) echo "USER_REQUEST" ;;
+		*) echo "UNKNOWN_ERROR" ;;
+	esac
+}
+
+_hostapd_common() {
+	local ifname="$1"
+
+	_config="/var/run/hostapd-$ifname.conf"
+	_pid="/var/run/wired-$ifname.pid"
```
feckert, line 30: What do we do if the ifname is not there. And therefore no configuration can be written.

feckert (changes requested):
The indentation looks strange, because you can't work with EOL here. How about usind append linke in https://github.com/openwrt/packages/blob/master/net/knxd/files/knxd.init ? This would make the code more readable

net/ieee8021x/files/ieee8021xclient.sh
yangfl: What do you mean by 'no configuration can be written'?

The _config file is not supposed to be here before we create it.

yangfl: ppp proto does not check for present of interface too? https://github.com/openwrt/openwrt/blob/master/package/network/services/ppp/files/ppp.sh#L221

feckert: Just because it's not done there doesn't mean it's right.
But it was just a suggestion.
I can live without the check.
Just thought that it might simplify the debugging task.

yangfl: I can't image why proto_X_setup will not be given a valid interface.

Nor I can find an example on how to do the check...

net/ieee8021x/files/ieee8021xclient.sh
```diff
@@ -0,0 +1,122 @@
...
+	${ca_cert2:+ca_cert2=${ca_cert2}}
+	${client_cert2:+client_cert2=${client_cert2}}
+	${private_key2:+private_key2=${private_key2}}
+	${private_key_passwd2:+private_key_passwd2=${private_key_passwd2}}
+	${dh_file2:+dh_file2=${dh_file2}}
+	${subject_match2:+subject_match2=${subject_match2}}
+	${eap_workaround:+eap_workaround=1}
+}
+EOF
+	proto_run_command "$cfg" /usr/sbin/wpa_supplicant -c "$_pid" -D wired -i "$iface"
```
dangowrt, line 75: Please use `ubus call wpa_supplicant config_add "{ \"driver\":\"wired\", \"iface\": \"$iface\", \"config\": \"$_pid\" }"` or something like that. There is no need to have an extra instance of wpa_supplicant, you can use the system wide instance.

net/ieee8021x/files/ieee8021xclient.sh
```diff
@@ -0,0 +1,122 @@
...
+		2)
+			proto_notify_error "$interface" "$errorstring"
+			proto_block_restart "$interface"
+		;;
+		*)
+			proto_notify_error "$interface" "$errorstring"
+		;;
+	esac
+
+	proto_kill_command "$interface"
```
dangowrt, line 94: Also here, use `ubus call wpa_supplcant config_del ...` instead

dangowrt (changes requested):

blogic:
the problem here is that on the server side the ports are not excluded from the bridge, unless I am missing something
I would suggest that we use the client proto from this package and combine it with the ieee8021x daemon I recently wrote which has full netifd integration via ubus.
--> https://github.com/blogic/ieee8021x
and thanks to OP for building the client proto, been meaning to do so for a while now

net/ieee8021xclient/Makefile
```diff
@@ -0,0 +1,42 @@
...
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=ieee8021xclient
+PKG_RELEASE:=1
+PKG_LICENSE:=GPL-2.0+
```
neheb, line 14: Should be -or-later instead of +

PR #15990, opened 2021-06-28
uci2: update revision

jpetrina:
Pull in new API function from upstream.

Signed-off-by: Jakov Petrina <jakov.petrina@sartura.hr>


BKPepe:
I am reverting this commit for the following reasons:

1. You are pulling different sources, but you didn't update the `PKG_VERSION` or even `PKG_RELEASE.` This ends up in a situation when existing users who have installed your library will have a different version than users who installed it after this PR.

2. Please, don't use cryptic commit messages like "update revision". These are not helping to know others to know something more while going through history. If you are using the same commit subject, then you will be lost. If possible specific to which version you do the bump.

jpetrina:
@BKPepe Thanks, I agree with necessity of either `PKG_VERSION` or `PKG_RELEASE`. I suppose `PKG_RELEASE` would be appropriate here until upstream decides to release v1.1

Regarding cryptic commit messages, would using Git commit hashes be appropriate e.g. `uci2: bump to revision ccaf677` or perhaps something like `uci2: bump to v1.0-ccaf677`?

BKPepe:
Fine by me.

pprindeville:
> @BKPepe Thanks, I agree with necessity of either `PKG_VERSION` or `PKG_RELEASE`. I suppose `PKG_RELEASE` would be appropriate here until upstream decides to release v1.1
> 
> Regarding cryptic commit messages, would using Git commit hashes be appropriate e.g. `uci2: bump to revision ccaf677` or perhaps something like `uci2: bump to v1.0-ccaf677`?

Don't forget that dates are always an option too, like `uci2: bump to head as of 2021-12-15` or whatever...

Issue #15988, opened 2021-06-28
python-dateutil: Compilation Fails on Debian 11 


DeathCamel58:
Maintainer: @commodo (find it by checking history of the package Makefile)
Environment: 
```
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye
```

Description:
`python-dateutil` fails to compile with setuptools version 6.0.1 (which the build system uses). It looks like support was dropped for Python 2. 
It also looks like the `19.07` branch buildbot failed to compile this package as well.

Did some research and found these:
[Very similar setuptools error](https://github.com/pypa/setuptools_scm/issues/541)

[Failing build log from buildbot](https://downloads.openwrt.org/releases/faillogs-19.07/aarch64_generic/packages/python-dateutil/python/compile.txt)

commodo:
> Maintainer: @commodo (find it by checking history of the package Makefile)
> Environment:
> 
> ```
> No LSB modules are available.
> Distributor ID:	Debian
> Description:	Debian GNU/Linux 11 (bullseye)
> Release:	11
> Codename:	bullseye
> ```
> 
> Description:
> `python-dateutil` fails to compile with setuptools version 6.0.1 (which the build system uses). It looks like support was dropped for Python 2.
> It also looks like the `19.07` branch buildbot failed to compile this package as well.
> 
> Did some research and found these:
> [Very similar setuptools error](https://github.com/pypa/setuptools_scm/issues/541)
> 
> [Failing build log from buildbot](https://downloads.openwrt.org/releases/faillogs-19.07/aarch64_generic/packages/python-dateutil/python/compile.txt)

so, this looks like an issue on 19.07 only
will check

commodo:
this seems to fix it  https://github.com/openwrt/packages/pull/15995

DeathCamel58:
Yep, thanks!

PR #15995, opened 2021-06-29
[19.07] python-dateutil: disable setuptools-scm for build

commodo:
Maintainer: me
Compile tested: ath79  openwrt-19.07  https://github.com/openwrt/openwrt/commit/71fa524e1bef2b266d9ed3ee94c2b7ff9d8e59b0
Run tested: n/a

Fixes https://github.com/openwrt/packages/issues/15988

It seems that the newer setuptools-scm package (6.0.1) has some
Python3-only syntax.
For the 19.07 release, where Python2 is still around this causes the
python-dateutil package to fail to build.

See https://github.com/pypa/setuptools_scm/issues/541

However, removing 'setuptools-scm' from the build also works.
This change does that.

Description:

### DIFF
--- a/lang/python/python-dateutil/Makefile
+++ b/lang/python/python-dateutil/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-dateutil
 PKG_VERSION:=2.8.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_LICENSE:=BSD-2-Clause
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
--- a/lang/python/python-dateutil/patches/0001-disable-setuptools-scm.patch
+++ b/lang/python/python-dateutil/patches/0001-disable-setuptools-scm.patch
@@ -1,0 +1,32 @@
+--- a/pyproject.toml
++++ b/pyproject.toml
+@@ -2,8 +2,7 @@
+ requires = [
+     "setuptools; python_version != '3.3'",
+     "setuptools<40.0; python_version == '3.3'",
+-    "wheel",
+-    "setuptools_scm"
++    "wheel"
+ ]
+ build-backend = "setuptools.build_meta"
+ 
+--- a/setup.py
++++ b/setup.py
+@@ -49,9 +49,6 @@ README = README()  # NOQA
+ 
+ 
+ setup(name="python-dateutil",
+-      use_scm_version={
+-          'write_to': 'dateutil/_version.py',
+-      },
+       description="Extensions to the standard Python datetime module",
+       author="Gustavo Niemeyer",
+       author_email="gustavo@niemeyer.net",
+@@ -65,7 +62,6 @@ setup(name="python-dateutil",
+       python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*",
+       package_data={"dateutil.zoneinfo": ["dateutil-zoneinfo.tar.gz"]},
+       zip_safe=True,
+-      setup_requires=['setuptools_scm'],
+       install_requires=["six >=1.5"],
+       classifiers=[
+           'Development Status :: 5 - Production/Stable',


DeathCamel58:
`python-dateutil` compiles successfully on my system now. I truly appreciate your help commodo!

commodo:
> `python-dateutil` compiles successfully on my system now. I truly appreciate your help commodo!

you're welcome :)

jefferyto:
A few questions:

* This removes the use of setuptools-scm but does not pass `PKG_VERSION` to `setup()` - does the compiled package have the right version number (e.g. in the egg-info directory)?
* Does this mean python-dateutil in the master and openwrt-21.02 branches need `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm`
* Are there other packages in the openwrt-19.07 branch that use setuptools-scm and are affected by the Python 3-only syntax in 6.0.1?

commodo:
> A few questions:
> 
>     * This removes the use of setuptools-scm but does not pass `PKG_VERSION` to `setup()` - does the compiled package have the right version number (e.g. in the egg-info directory)?
>       Edit: It appears that setuptools-scm is told to write the version number to a file and is loaded at runtime (and available as [`dateutil.__version__`](https://dateutil.readthedocs.io/en/stable/changelog.html#version-1-3).

so, in the tar-ball, there is a `dateutil/_version.py`  [contents below]:
```
# coding: utf-8
# file generated by setuptools_scm
# don't change, don't track in version control
version = '2.8.0'
```

but i am puzzled as to why the version is 0.0.0; will check

> 
>     * Does this mean python-dateutil in the master and openwrt-21.02 branches need `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm`?

yes;
though that makes me wonder how things worked up until now;
i'm still not sure who installs `setuptools-scm` in the python-dateutil case


> 
>     * Are there other packages in the openwrt-19.07 branch that use setuptools-scm and are affected by the Python 3-only syntax in 6.0.1?

there are a few other packages in the 19.07 branch that define `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm`
this seems to be the only Python2 package affected;

commodo:
> but i am puzzled as to why the version is 0.0.0; will check

looks like setuptools-scm puts in something like:
```
# coding: utf-8
# file generated by setuptools_scm
# don't change, don't track in version control
version = u'2.8.0'
version_tuple = (2, 8, 0)
```

so another fix [for 19.07] would be to pin setuptools-scm to version 5.0.2
that looks like somehting that still works with Python2

jefferyto:
> there are a few other packages in the 19.07 branch that define `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm`
> this seems to be the only Python2 package affected;

That's good - thanks for checking :smile: 

> yes;
> though that makes me wonder how things worked up until now;
> i'm still not sure who installs setuptools-scm in the python-dateutil case

I think setuptools probably just installed setuptools-scm as part of the setup process for python-dateutil.

Would you mind also adding `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm` for master and openwrt-21.02? This would make the dependency on setuptools-scm more visible in the future.

commodo:
> Would you mind also adding `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm` for master and openwrt-21.02? This would make the dependency on setuptools-scm more visible in the future.

I started this right after this patch.
But got side-tracked by something else.

I have the branch, just need to test it on master.

commodo:
> > Would you mind also adding `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm` for master and openwrt-21.02? This would make the dependency on setuptools-scm more visible in the future.
> 
> I started this right after this patch.
> But got side-tracked by something else.
> 
> I have the branch, just need to test it on master.

created  https://github.com/openwrt/packages/pull/16013  and  https://github.com/openwrt/packages/pull/16012

PR #15998, opened 2021-06-29
udhcpsnoop: import package from TIP

mweinelt:
```
Import the udhcpsnoop package from the TIP feed.

The original Makefile in TIP repo spcifies GPL-2.0 as the PKG_LICENSE.
However this is a mistake, as the SPDX headers of the repos source files
all specify BSD-3-Clause as their license.

See https://github.com/Telecominfraproject/wlan-ap/blob/uCentral-trunk/feeds/ucentral/udhcpsnoop/Makefile

```

Maintainer: @blogic (?)
Compile tested: ipq40xx-generic
Run tested: gl-b1300

Description: dhcp snooping daemon that relays dhcp packets as ubus events

### DIFF
--- a/net/udhcpsnoop/Makefile
+++ b/net/udhcpsnoop/Makefile
@@ -1,0 +1,31 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=udhcpsnoop
+PKG_RELEASE:=1
+
+PKG_LICENSE:=BSD-3-Clause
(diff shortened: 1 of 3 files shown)


blogic:
ah nice, yes the Makefile is incorrect, the license is bsd3

blocktrron (approved):

blogic (approved):

PR #16000, opened 2021-06-29
adblock: update 4.1.3-2

dibdot:
Maintainer: me / @dibdot
Compile tested: -
Run tested: TP-Link RE650 v1, OpenWrt SNAPSHOT r17073-8bb4437c01

Description:
* add a tcpdump option to resolve IPs in adblock reporting,
  set 'adb_represolve' accordingly (disabled by default). If enabled
  tcpdump will perform a reverse DNS (PTR) lookup for each IP address
* add 'stalkerware' source (provided by @astryzia/@rahlquist)
* update readme

### DIFF
--- a/net/adblock/Makefile
+++ b/net/adblock/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=adblock
 PKG_VERSION:=4.1.3
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_LICENSE:=GPL-3.0-or-later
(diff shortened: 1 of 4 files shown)


neheb:
Controversial suggestion: https://github.com/openwrt/openwrt/pull/4300

It changes the script here quite a bit.

dibdot:
Many of the stuff is taste (e.g. indent changes, wrapping lines etc.). From maintainer perspective it's most important that I can read & understand the script code (and not to satisfy a linting utility or the "common practise"). I test all my scripts locally with shellcheck to find obvious bugs beforehand, but that's enough.

PR #16010, opened 2021-07-01
strongswan: trivial improvement in 'stroke' packaging

pprindeville:
Maintainer: me, @Thermi
Compile tested: x86_64, generic, HEAD (db9784bedd)
Run tested: same, verified in build_dir

Description:

Try to make builds go faster by reducing the number of shell commands each recipe calls.

### DIFF
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -614,14 +614,7 @@ define Plugin/attr-sql/install
 endef
 
 define Plugin/stroke/install
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/aacerts
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/acerts
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/cacerts
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/certs
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/crls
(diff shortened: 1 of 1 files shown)


net/strongswan/Makefile
```diff
@@ -614,14 +614,7 @@ define Plugin/attr-sql/install
...
 define Plugin/stroke/install
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/aacerts
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/acerts
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/cacerts
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/certs
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/crls
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/ocspcerts
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/private
-	$(INSTALL_DIR) $(1)/etc/ipsec.d/reqs
+	$(INSTALL_DIR) $(1)/etc/ipsec.d/{aacerts,acerts,cacerts,certs,crls,ocspcerts,private,reqs}
```
Thermi, line 617: That's a bashism. Posix shell doesn't have taht AFAIK, so it's actually not strictly posix compliant. But neither is that in line 620/627. So it probably doesn't even matter.

PR #16027, opened 2021-07-02
qemu: update to 6.1.0, add usb passthrough option

vooon:
Signed-off-by: Vladimir Ermakov <vooon341@gmail.com>

Maintainer: @yousong
Compile tested: x86_64, OpenWrt master (https://github.com/openwrt/openwrt/commit/90e167abaae80a59c8780d3305f8e1260d9f1339), GCC 10.3
Run tested: x86_64 host, KVM for qemu-x86_64-softmmu + UEFI VM (homeassistant ova).

Description:
- Version updated to 6.1.0
- I had have compile problems with gcc 10 on 5.x
- Added option to enable USB pass-through (used it before with 5.x)
- Some changes in build system do not allow to use `make x86_64-softmmu` or `make qemu-img`, current workaround - build all

### DIFF
--- a/utils/qemu/Makefile
+++ b/utils/qemu/Makefile
@@ -9,12 +9,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=qemu
-PKG_VERSION:=5.0.0
-PKG_RELEASE:=5
+PKG_VERSION:=6.1.0
+PKG_RELEASE:=$(AUTORELEASE)
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_HASH:=2f13a92a0fa5c8b69ff0796b59b86b080bbb92ebad5d301a7724dd06b5e78cb6
+PKG_HASH:=eebc089db3414bbeedf1e464beda0a7515aad30f73261abc246c9b27503a3c96
 PKG_SOURCE_URL:=http://download.qemu.org/
-PKG_LICENSE:=GPL-2.0
+PKG_LICENSE:=GPL-2.0-only
 PKG_LICENSE_FILES:=LICENSE tcg/LICENSE
 PKG_MAINTAINER:=Yousong Zhou <yszhou4tech@gmail.com>
 PKG_CPE_ID:=cpe:/a:qemu:qemu
@@ -23,7 +23,7 @@ PKG_BUILD_PARALLEL:=1
 PKG_INSTALL:=1
 PKG_USE_MIPS16:=0
 
-PKG_BUILD_DEPENDS+=spice-protocol
+PKG_BUILD_DEPENDS+=spice-protocol meson/host
 
 include $(INCLUDE_DIR)/nls.mk
 include $(INCLUDE_DIR)/package.mk
@@ -96,7 +96,7 @@ endef
 
 define Package/qemu-img/install
 	$(INSTALL_DIR) $(1)/usr/bin/
-	$(INSTALL_BIN) $(PKG_BUILD_DIR)/qemu-img $(1)/usr/bin/qemu-img
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/build/qemu-img $(1)/usr/bin/qemu-img
 endef
 
 
@@ -111,7 +111,7 @@ endef
 
 define Package/qemu-nbd/install
 	$(INSTALL_DIR) $(1)/usr/sbin/
-	$(INSTALL_BIN) $(PKG_BUILD_DIR)/qemu-nbd $(1)/usr/sbin/qemu-nbd
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/build/qemu-nbd $(1)/usr/sbin/qemu-nbd
 endef
 
 
@@ -190,12 +190,13 @@ define qemu-target
     SUBMENU:=Virtualization
     TITLE:=QEMU target $(1)
     URL:=http://www.qemu.org
-    DEPENDS:= +glib2 +libpthread +zlib $(QEMU_DEPS_IN_HOST) $(qemu-target-$(1)-deps) \
+    DEPENDS:= +glib2 +libpthread +zlib +libbpf $(QEMU_DEPS_IN_HOST) $(qemu-target-$(1)-deps) \
 	+QEMU_UI_VNC:qemu-keymaps \
 	+QEMU_UI_VNC_JPEG:libjpeg \
 	+QEMU_UI_VNC_PNG:libpng \
 	+QEMU_UI_VNC_SASL:libsasl2 \
 	+QEMU_UI_SPICE:libspice-server \
+	+QEMU_DEV_USB:libusb-1.0 \
 	$(if $(filter %-softmmu,$(1)),+libncurses +libfdt +pixman +qemu-firmware-efi $(ICONV_DEPENDS))
   endef
 
@@ -257,6 +258,9 @@ config QEMU_UI_VNC_SASL
 config QEMU_UI_SPICE
 	bool "QEMU SPICE ui support"
 
+config QEMU_DEV_USB
+	bool "QEMU USB passthrough support"
+
 endif
 endef
 
@@ -266,6 +270,7 @@ PKG_CONFIG_DEPENDS += \
 	CONFIG_QEMU_UI_VNC_PNG \
 	CONFIG_QEMU_UI_VNC_SASL \
 	CONFIG_QEMU_UI_SPICE \
+	CONFIG_QEMU_DEV_USB \
 
 
 # QEMU configure script does not recognize these options
@@ -299,7 +304,9 @@ CONFIGURE_ARGS +=			\
 	--enable-vhost-scsi		\
 	--enable-vhost-user		\
 	--enable-vhost-user-fs		\
+	--enable-vhost-user-blk-server	\
 	--enable-vhost-vsock		\
+	--enable-vhost-vdpa		\
 
 # Image formats support
 CONFIGURE_ARGS +=			\
@@ -351,6 +358,8 @@ CONFIGURE_ARGS +=			\
 	--enable-live-block-migration	\
 	--enable-membarrier		\
 	--enable-replication		\
+	--enable-lto			\
+	--enable-tools			\
 
 # Review configure options not explicitly specified here
 #
@@ -375,7 +384,7 @@ CONFIGURE_ARGS +=			\
 	--disable-debug-tcg		\
 	--disable-docs			\
 	--disable-gcrypt		\
-	--disable-git-update		\
+	--with-git-submodules=ignore	\
 	--disable-glusterfs		\
 	--disable-gnutls		\
 	--disable-guest-agent-msi	\
@@ -385,7 +394,7 @@ CONFIGURE_ARGS +=			\
 	--disable-libpmem		\
 	--disable-libssh		\
 	--disable-libudev		\
-	--disable-libusb		\
+	--$(if $(CONFIG_QEMU_DEV_USB),enable,disable)-libusb		\
 	--disable-libxml2		\
 	--disable-linux-aio		\
 	--disable-linux-io-uring	\
@@ -405,19 +414,16 @@ CONFIGURE_ARGS +=			\
 	--disable-rdma			\
 	--disable-sanitizers		\
 	--disable-seccomp		\
-	--disable-sheepdog		\
 	--disable-smartcard		\
 	--disable-snappy		\
 	--disable-sparse		\
 	--disable-strip			\
 	--disable-tcg-interpreter	\
 	--disable-tcmalloc		\
-	--disable-tools			\
 	--disable-tpm			\
 	--disable-usb-redir		\
 	--disable-vde			\
 	--disable-virtfs		\
-	--disable-vxhs			\
 	--disable-werror		\
 	--disable-xen-pci-passthrough	\
 	--disable-xkbcommon		\
@@ -434,17 +440,6 @@ MAKE_VARS += V=1
 MAKE_FLAGS:=$(filter-out	\
 	ARCH=%			\
 	,$(MAKE_FLAGS))
-
-QEMU_MAKE_TARGETS := \
-	$(if $(CONFIG_PACKAGE_qemu-ga),qemu-ga) \
-	$(if $(CONFIG_PACKAGE_qemu-bridge-helper),qemu-bridge-helper) \
-	$(if $(CONFIG_PACKAGE_qemu-img),qemu-img) \
-	$(if $(CONFIG_PACKAGE_qemu-nbd),qemu-nbd) \
-	$(foreach target,$(qemu-target-list),$(if $(CONFIG_PACKAGE_qemu-$(target)),$(target)/all)) \
-
-define Build/Compile
-	$(if $(strip $(QEMU_MAKE_TARGETS)),$(call Build/Compile/Default,$(QEMU_MAKE_TARGETS)))
-endef
 
 $(eval $(call BuildPackage,virtio-console-helper))
 $(eval $(call BuildPackage,qemu-ga))
--- a/utils/qemu/patches/0001-configure-allow-disable-fortify_source.patch
+++ b/utils/qemu/patches/0001-configure-allow-disable-fortify_source.patch
@@ -11,9 +11,9 @@ OpenWrt base build system decide flavor of fortify_source to use
 
 --- a/configure
 +++ b/configure
-@@ -1601,6 +1601,8 @@ for opt do
+@@ -1581,6 +1581,8 @@ for opt do
    ;;
-   --gdb=*) gdb_bin="$optarg"
+   --disable-slirp-smbd) slirp_smbd=no
    ;;
 +  --disable-fortify-source) fortify_source="no"
 +  ;;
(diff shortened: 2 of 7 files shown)


vooon:
Works perfectly fine for my home assistant ova.
```
# cat /etc/init.d/qemu-homeassistant 
#!/bin/sh /etc/rc.common

#set -x

START=99
STOP=1
 
qemu_pidfile="/var/run/qemu-homeassistant.pid"
qemu_logfile="/srv/log/qemu-homeassistant.log"
qemu_uuid="88cad999-3483-48cb-b324-ea62dc641ed9"
qemu_qmp_port=4445

start() {
qemu-system-x86_64 -enable-kvm -cpu host -smp 2 -m 2G \
	-uuid $qemu_uuid \
	-drive if=pflash,file=/srv/vm/OVMF/OVMF_CODE.fd,format=raw,readonly=on \
	-drive if=pflash,file=/srv/vm/OVMF/homeassistant_vars.fd,format=raw \
	-drive file=/srv/vm/homeassistant.qcow2,if=virtio \
	-drive file=/mnt/share/vm/ha-media.qcow2,if=virtio \
	-device virtio-net-pci,mac=E2:F2:6A:01:9D:D0,netdev=br0 \
	-netdev bridge,br=br-lan,id=br0 \
	-serial pty \
	-serial pty \
	-device qemu-xhci \
	-device usb-host,vendorid=0x0a12,productid=0x0001 \
	-vnc :0 \
	-chardev socket,path=/tmp/ha-qga.sock,server,nowait,id=qga0 \
	-device virtio-serial \
	-device virtserialport,chardev=qga0,name=org.qemu.guest_agent.0 \
	-qmp tcp:127.0.0.1:$qemu_qmp_port,server,nowait \
	-daemonize &> $qemu_logfile

	#-serial /dev/tty-senseair-s8-co2 \
	#-serial /dev/tty-zzh \
	#-device usb-host,vendorid=0x0a12,productid=0x0001 \

#/usr/bin/pgrep qemu-system-x86_64 > $qemu_pidfile
/usr/bin/pgrep -f "qemu-system-x86_64.*$qemu_uuid" > $qemu_pidfile
echo "QEMU: Started VM with PID $(cat $qemu_pidfile)."
}
 
stop() {
echo "QEMU: Sending 'system_powerdown' to VM with PID $(cat $qemu_pidfile)."
nc localhost $qemu_qmp_port <<QMP 
{ "execute": "qmp_capabilities" } 
{ "execute": "system_powerdown" } 
QMP
 
if [ -e $qemu_pidfile ]; then
	if [ -e /proc/$(cat $qemu_pidfile) ]; then
		echo "QEMU: Waiting for VM shutdown."
		while [ -e /proc/$(cat $qemu_pidfile) ]; do sleep 1s; done
		echo "QEMU: VM Process $(cat $qemu_pidfile) finished."
	else
		echo "QEMU: Error: No VM with PID $(cat $qemu_pidfile) running."
	fi
 
	rm -f $qemu_pidfile
else
	echo "QEMU: Error: $qemu_pidfile doesn't exist."
fi
}
```

neheb:
Need to run make package/qemu/refresh

vooon:
@neheb done.

vooon:
@neheb do you know what flag is needed to have python3 setuptools on host?
```
../meson.build:56:0: ERROR: <ExternalProgram 'python3' -> ['/home/build/openwrt/staging_dir/host/bin/python3']> is not a valid python or it is missing setuptools
```

vooon:
That probably bad idea to add python3/host to build depends, but i don't know much about meson and qemu's use of it.
And since qemu does a lot of code generation it is likely that python was being imported in many build scripts.

neheb:
Yes I do. See https://github.com/openwrt/packages/blob/master/multimedia/gstreamer1/patches/020-distutils.patch

vooon:
I patched meson.build, let's see would it work. For previous one it's not clear why x86 fail.

neheb:
Need to refresh patches.

vooon:
Ugh, no, i cannot make it build with patch 0008. reverted back to python3/host.

neheb:
ping @jefferyto @dhewg 

Not sure about the python3 host dependency...

jefferyto:
@vooon perhaps you can post the build error you get after patching meson.build

vooon:
@jefferyto , @neheb i wanted to reproduce build error to post here, but instead i got it build.
But I updated openwrt and rebased packages repos. To re-check i'm started clean rebuild again.

vooon:
@neheb friendly ping.

neheb:
python3/host is no go. It increases build time too much.

neheb:
While this is converted to meson, I still see old Makefile type patches. Intentional?

neheb:
Doesn't compile on mips locally

```
../tests/fp/fp-bench.c: In function 'set_host_precision':
../tests/fp/fp-bench.c:491:17: error: 'FE_TOWARDZERO' undeclared (first use in this function)
  491 |         rhost = FE_TOWARDZERO;
      |                 ^~~~~~~~~~~~~
../tests/fp/fp-bench.c:491:17: note: each undeclared identifier is reported only once for each function it appears in
../tests/fp/fp-bench.c:494:17: error: 'FE_DOWNWARD' undeclared (first use in this function)
  494 |         rhost = FE_DOWNWARD;
      |                 ^~~~~~~~~~~
../tests/fp/fp-bench.c:497:17: error: 'FE_UPWARD' undeclared (first use in this function)
  497 |         rhost = FE_UPWARD;

```

Interesting that tests are being run...

neheb:
Another one...

```
../block/export/fuse.c: In function 'fuse_fallocate':
../block/export/fuse.c:563:23: error: 'FALLOC_FL_ZERO_RANGE' undeclared (first use in this function)
  563 |     } else if (mode & FALLOC_FL_ZERO_RANGE) {
      |                       ^~~~~~~~~~~~~~~~~~~~
../block/export/fuse.c:563:23: note: each undeclared identifier is reported only once for each function it appears in
../block/export/fuse.c: In function 'fuse_lseek':
../block/export/fuse.c:639:19: error: 'SEEK_HOLE' undeclared (first use in this function)
  639 |     if (whence != SEEK_HOLE && whence != SEEK_DATA) {
      |                   ^~~~~~~~~
../block/export/fuse.c:639:42: error: 'SEEK_DATA' undeclared (first use in this function); did you mean 'SEEK_SET'?
  639 |     if (whence != SEEK_HOLE && whence != SEEK_DATA) {
      |                                          ^~~~~~~~~
      |                                          SEEK_SET
ninja: build stopped: subcommand failed.
```

vooon:
> While this is converted to meson, I still see old Makefile type patches. Intentional?

I planned to rebase them out after CI pass.

> Doesn't compile on mips locally

I didn't try to build there, only amd64 and armv7.
Could you please provide config file?

vooon:
@neheb note, i updated PR to 6.1.0, which i'm testing now.

utils/qemu/Makefile
```diff
@@ -435,15 +441,16 @@ MAKE_FLAGS:=$(filter-out	\
 	ARCH=%			\
 	,$(MAKE_FLAGS))
 
-QEMU_MAKE_TARGETS := \
-	$(if $(CONFIG_PACKAGE_qemu-ga),qemu-ga) \
-	$(if $(CONFIG_PACKAGE_qemu-bridge-helper),qemu-bridge-helper) \
-	$(if $(CONFIG_PACKAGE_qemu-img),qemu-img) \
-	$(if $(CONFIG_PACKAGE_qemu-nbd),qemu-nbd) \
-	$(foreach target,$(qemu-target-list),$(if $(CONFIG_PACKAGE_qemu-$(target)),$(target)/all)) \
+# QEMU_MAKE_TARGETS := \
```
neheb, line 444: Sounds like this should be removed.

utils/qemu/Makefile
```diff
@@ -435,15 +441,16 @@ MAKE_FLAGS:=$(filter-out	\
...
+# 	$(if $(CONFIG_PACKAGE_qemu-ga),qemu-ga) \
+# 	$(if $(CONFIG_PACKAGE_qemu-bridge-helper),qemu-bridge-helper) \
+# 	$(if $(CONFIG_PACKAGE_qemu-img),qemu-img) \
+# 	$(if $(CONFIG_PACKAGE_qemu-nbd),qemu-nbd) \
+# 	$(foreach target,$(qemu-target-list),$(if $(CONFIG_PACKAGE_qemu-$(target)),$(target)/all)) \
 
 define Build/Compile
```
neheb, line 451: Sounds like this should be removed.

utils/qemu/Makefile
```diff
@@ -9,10 +9,10 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=qemu
-PKG_VERSION:=5.0.0
-PKG_RELEASE:=5
+PKG_VERSION:=6.0.0
+PKG_RELEASE:=1
```
neheb, line 13: Could replace with $(AUTORELEASE)

utils/qemu/Makefile
```diff
@@ -9,10 +9,10 @@
...
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_HASH:=2f13a92a0fa5c8b69ff0796b59b86b080bbb92ebad5d301a7724dd06b5e78cb6
+PKG_HASH:=87bc1a471ca24b97e7005711066007d443423d19aacda3d442558ae032fa30b9
 PKG_SOURCE_URL:=http://download.qemu.org/
 PKG_LICENSE:=GPL-2.0
```
neheb, line 17: This should be either -only or -or-later, as per SPDX.

utils/qemu/Makefile
```diff
@@ -266,6 +270,7 @@ PKG_CONFIG_DEPENDS += \
 	CONFIG_QEMU_UI_VNC_PNG \
 	CONFIG_QEMU_UI_VNC_SASL \
 	CONFIG_QEMU_UI_SPICE \
+	CONFIG_QEMU_DEV_USB \
 
 
 # QEMU configure script does not recognize these options
```
neheb, line 276: Interesting that CONFIGURE_ARGS is used for a meson package...

utils/qemu/patches/0008-meson-python-fix.patch
```diff
@@ -0,0 +1,51 @@
+--- a/meson.build
```
neheb, line 1: This patch is no longer needed.

utils/qemu/Makefile
vooon: I'm not quite sure what should be used: https://github.com/qemu/qemu/blob/master/LICENSE

It sounds to me, that most of the code is GPL-2.0-only, but some parts could be GPL-2.0-or-later.

utils/qemu/Makefile
vooon: Will do.

utils/qemu/patches/0008-meson-python-fix.patch
vooon: Yes, i removed that patch and it is building. Note i'm not yet got anything working. Seems at lease x86 need libbpf now.

utils/qemu/Makefile
neheb: Use -only then.

utils/qemu/Makefile
vooon: Done.

utils/qemu/Makefile
vooon: Done.

utils/qemu/Makefile
vooon: Done.

utils/qemu/Makefile
vooon: It still uses `./configure`. Do you think that it should be changed?

utils/qemu/patches/0008-meson-python-fix.patch
vooon: Qemu compilled. Will update my router soon...

utils/qemu/Makefile
neheb: I mean the entire Build/Compile section

vooon:
@neheb tested on x86_64 - all my VMs are working. USB and serial pass-trough works.

neheb (approved):

neheb:
ping @yousong

utils/qemu/Makefile
```diff
@@ -298,8 +303,9 @@ CONFIGURE_ARGS +=			\
 	--enable-vhost-net		\
 	--enable-vhost-scsi		\
 	--enable-vhost-user		\
-	--enable-vhost-user-fs		\
```
yousong, line 301: Please add a note about why this was removed

yousong (approved):
Maybe we should squash the two commits together for final merge.  I am sure this version bump requires many diggings and tryings to have such shape.  Please considering adding your thoughts and findings in the commit message for other devs to learn the details.  It will be useful for future maintenance.

vooon:
@yousong done rebase. Added some notes to the commit.
But actually there are not much findings. Perhaps primary one - how to fix meson to use correct python.
Other changes mostly result of build trials...

utils/qemu/Makefile
vooon: That option was removed from qemu:
```
qemu-6.1.0$ ./configure --help | grep vhost
  vhost-net       vhost-net kernel acceleration support
  vhost-vsock     virtio sockets device support
  vhost-scsi      vhost-scsi kernel target support
  vhost-crypto    vhost-user-crypto backend support
  vhost-kernel    vhost kernel backend support
  vhost-user      vhost-user backend support
  vhost-user-blk-server    vhost-user-blk server support
  vhost-vdpa      vhost-vdpa kernel backend support
```

yousong: It's still there, just not in the show_help output：https://github.com/qemu/qemu/blob/f9baca549e44791be0dd98de15add3d8452a8af0/configure#L1283-L1286

Maybe try again with the commands in the Makefile comment:

https://github.com/openwrt/packages/blob/f2280ae82e3ec04649639b020e4676f1d2125d13/utils/qemu/Makefile#L355-L363

vooon:
Frinedly ping @neheb .

neheb:
@vooon the meson and python thing was fixed. distutils is a requirement now.

As far as merging, up to @yousong

yousong:
Hi, @vooon , the conversation about --enable-vhost-user-fs is still ongoing ;)

vooon:
@yousong ah, right. I do remember, that i had to remove some of args cause configure failed. Checking.

utils/qemu/Makefile
vooon:
```
user@6b03bfc66c94:~/ws$ qemu_configure=build_dir/target-x86_64_musl/qemu-6.1.0/configure 
user@6b03bfc66c94:~/ws$ openwrt_makefile=feeds/packages/utils/qemu/Makefile 
user@6b03bfc66c94:~/ws$ for arg in $(grep -E '^\s*--disable-[^)]+\)' "$qemu_configure" | cut -f1 -d')'); do grep -qE "(--enable|--disable)${arg#--disable}" "$openwrt_makefile" || echo "$arg"; done
--disable-virtiofsd
--disable-vnc
--disable-gettext
--disable-block-drv-whitelist-in-tools
--disable-tsan
--disable-vnc-sasl
--disable-vnc-jpeg
--disable-vnc-png
--disable-slirp
--disable-nvmm
--disable-malloc-trim
--disable-spice
--disable-spice-protocol
--disable-pie
--disable-safe-stack
--disable-cfi
--disable-cfi-debug
--disable-fdt
--disable-bpf
--disable-blobs
--disable-vhost-user-fs
--disable-u2f
--disable-libusb
--disable-zlib-test
--disable-avx2
--disable-avx512f
--disable-virtio-blk-data-plane|--enable-virtio-blk-data-plane
--disable-git-update
--disable-fuzzing
--disable-rng-none
--disable-keyring
--disable-libdaxctl
--disable-fuse
--disable-fuse-lseek
--disable-multiprocess
--disable-gio
--disable-slirp-smbd
```

vooon: Checked with vhost-user-fs - all ok.

vooon:
@yousong reverted back vhost-user-fs - build successful. What do you suppose to do with other options?

yousong:
> @yousong reverted back vhost-user-fs - build successful. What do you suppose to do with other options?

I do not have much experience with those other options, e.g. `--disable-sheepdog`, `--disable-vxhs`.   As such and as current maintainer I prefer there is a note in the commit message sharing with us why the changes were made.

vooon:
@yousong added a note to the commit message. Now i recall why i got unknown options passed to configure.

yousong:
Merged.  Thank you all for the time and efforts!

neheb:
https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a53/packages/qemu/compile.txt

vooon:
@neheb seems we need `--disable-fuse-lseek` for that targets, even though it's support detected.
```
                          libudev: NO
                       FUSE lseek: YES
```

neheb:
Nah it's just a missing header.

vooon:
As for x86_64 i do not have that error on local builds.

vooon:
Ah. Ok.

PR #16029, opened 2021-07-03
mwan3: use default routes from additional tables

dangowrt:
Maintainer: @feckert @aaronjg
Compile tested: x86/64
Run tested: x86/64 (PCEngines APU3)

Description:
Until now the additional tables listed in gobal `rt_table_lookup` were not considered for interfaces.
In order to be able to also use interface-defined routes from tables other than main, consider also tables listed in `rt_table_lookup`.

### DIFF
--- a/net/mwan3/files/lib/mwan3/mwan3.sh
+++ b/net/mwan3/files/lib/mwan3/mwan3.sh
@@ -408,9 +408,17 @@ mwan3_delete_iface_iptables()
 
 }
 
+mwan3_extra_tables_routes()
+{
+	$IP route list table "$1"
+}
+
 mwan3_get_routes()
 {
-	$IP route list table main | sed -ne "$MWAN3_ROUTE_LINE_EXP" | uniq
+	{
+		$IP route list table main
+		config_list_foreach "globals" "rt_table_lookup" mwan3_extra_tables_routes
+	} | sed -ne "$MWAN3_ROUTE_LINE_EXP" | sort -u
 }
 
 mwan3_create_iface_route()
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.10.9
+PKG_VERSION:=2.10.10
 PKG_RELEASE:=1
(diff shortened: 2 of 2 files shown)


net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -408,9 +408,16 @@ mwan3_delete_iface_iptables()
...
 
+mwan3_extra_tables_routes()
+{
+	$IP route list table "$1"
+}
+
 mwan3_get_routes()
 {
-	$IP route list table main | sed -ne "$MWAN3_ROUTE_LINE_EXP" | uniq
+	( $IP route list table main ;
```
feckert, line 418: Can we make the source part a little different? I think it is cryptic

```
$IP route list table main | sed -ne "$MWAN3_ROUTE_LINE_EXP" | uniq
config_list_foreach "globals" "rt_table_lookup" mwan3_extra_tables_routes | sed -ne "$MWAN3_ROUTE_LINE_EXP" | uniq
		
```

feckert (commented):
Please pump also the [PKG_VERSION](https://github.com/openwrt/packages/blob/master/net/mwan3/Makefile#L11) to 2.10.10

net/mwan3/files/lib/mwan3/mwan3.sh
dangowrt: In the way you suggest the result is likely to contain duplicates, hence I wrapped the call to `grep ... | uniq` around all of it.

aaronjg:
Can you use brackets rather parens to avoid spawning a sub shell? Also, likely need to sort before uniq

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -408,9 +408,17 @@ mwan3_delete_iface_iptables()
...
+	$IP route list table "$1"
+}
+
 mwan3_get_routes()
 {
-	$IP route list table main | sed -ne "$MWAN3_ROUTE_LINE_EXP" | uniq
+	{
+		$IP route list table main
+		config_list_foreach "globals" "rt_table_lookup" mwan3_extra_tables_routes
+	} | sed -ne "$MWAN3_ROUTE_LINE_EXP" | sort -u
```
dangowrt, line 421: @aaronjg I cleaned it up a bit...

PR #16030, opened 2021-07-03
Revert "openvpn: disable LZO support by default"

jow-:
Maintainer: @mkrkn 

Description:

This reverts commit e4376793b4e093089543cb1bad64eef34ed25eca.

Many OpenVPN users are not in the position to disable LZO support on the
server side. Official OpenVPN packages should still support it to
retain broad compatibility with existing deployments.

Just because upstream does not recommend using compression, we should
not remove the ability to negotiate compressed connections, otherwise
we break existing deployments.

Signed-off-by: Jo-Philipp Wich <jo@mein.io>

### DIFF
--- a/net/openvpn/Config-mbedtls.in
+++ b/net/openvpn/Config-mbedtls.in
@@ -2,7 +2,7 @@ if PACKAGE_openvpn-mbedtls
 
 config OPENVPN_mbedtls_ENABLE_LZO
 	bool "Enable LZO compression support"
-	default n
+	default y
 
 config OPENVPN_mbedtls_ENABLE_LZ4
 	bool "Enable LZ4 compression support"
--- a/net/openvpn/Config-openssl.in
+++ b/net/openvpn/Config-openssl.in
@@ -2,7 +2,7 @@ if PACKAGE_openvpn-openssl
 
 config OPENVPN_openssl_ENABLE_LZO
 	bool "Enable LZO compression support"
-	default n
+	default y
 
 config OPENVPN_openssl_ENABLE_LZ4
 	bool "Enable LZ4 compression support"
--- a/net/openvpn/files/openvpn.config
+++ b/net/openvpn/files/openvpn.config
@@ -300,7 +300,9 @@ config openvpn sample_server
 	#
 	# LZ4 requires OpenVPN 2.4+ client and server
 #	option compress lz4
-	
+	# LZO is compatible with most OpenVPN versions
+	# (set "compress lzo" on 2.4+ clients, and "comp-lzo yes" on older clients)
+#	option compress lzo
 	# Control how OpenVPN handles peers using compression
 	#
 	# Do not allow any connections using compression
@@ -492,6 +494,8 @@ config openvpn sample_client
 	#
 	# LZ4 requires OpenVPN 2.4+ on server and client
 #	option compress lz4
+	# LZO is compatible with most OpenVPN versions
+#	option compress lzo
 
 	# Set log file verbosity.
 	option verb 3


champtar:
@mkrkn as we already have some people going back to 19.07 (likely) because of this, it would be really nice if you can ack this asap so we include it in 21.02 release

neheb:
> 
> 
> @mkrkn as we already have some people going back to 19.07 (likely) because of this, it would be really nice if you can ack this asap so we include it in 21.02 release

I don't think the packages feed works like that. opkg update still works.

jow-:
> > @mkrkn as we already have some people going back to 19.07 (likely) because of this, it would be really nice if you can ack this asap so we include it in 21.02 release
> 
> I don't think the packages feed works like that. opkg update still works.

What do you mean with this comment? There is no lzo enabled ooenvpn package to update to.

neheb:
I mean post release.

champtar:
Once it's merged in master cherry-pick in 21.02, but if we can avoid breakage, hour of troubleshooting and some frustration that's good. RC are here to identify issues before release ;)

mkrkn:
> 
> 
> @mkrkn as we already have some people going back to 19.07 (likely) because of this, it would be really nice if you can ack this asap so we include it in 21.02 release

If this change is so obvious that I should ack this ASAP, why did no one say so when I first sent a patch proposing dropping LZO support 11 months ago, or when I sent revised patches keeping this change? I'd be interested in discussing approaching this differently, I liked the idea of making -openssl full-featured and -mbedtls size optimized, as was suggested in an issue earlier this week. But I can't be bothered to have a rushed (or no) discussion in between rc3 and release, so I'm stepping away (#16033 ). This may seem overly dramatic, but it's just a "straw that broke the camel's back" situation, I don't get the things going on and I can't put the energy into figuring it out.

I think it is unfair that all OpenWrt users with a sane OpenVPN setup must install liblzo (or compile from source if they don't want to) because users with *objectively bad* setups are sticking to their footguns.

champtar:
> > @mkrkn as we already have some people going back to 19.07 (likely) because of this, it would be really nice if you can ack this asap so we include it in 21.02 release
> 
> If this change is so obvious that I should ack this ASAP, why did no one say so when I first sent a patch proposing dropping LZO support 11 months ago, or when I sent revised patches keeping this change? I'd be interested in discussing approaching this differently, I liked the idea of making -openssl full-featured and -mbedtls size optimized, as was suggested in an issue earlier this week. But I can't be bothered to have a rushed (or no) discussion in between rc3 and release, so I'm stepping away (#16033 ). This may seem overly dramatic, but it's just a "straw that broke the camel's back" situation, I don't get the things going on and I can't put the energy into figuring it out.

Hi @mkrkn,
11 months ago maybe it was not clear it would break some people usage.
Sorry if I rushed you, I just wanted to have an easy solution for all users as soon as possible, but this doesn't mean this weekend. I was the one to propose having openssl full featured and mbedtls/wolfssl stay the same, and I should have written please ack OR nack. If we didn't wanted discussion both @jow- and I can merge, we didn't because we do want the discussion to happen.

> I think it is unfair that all OpenWrt users with a sane OpenVPN setup must install liblzo (or compile from source if they don't want to) because users with _objectively bad_ setups are sticking to their footguns.

If you don't control the OpenVPN server then you can't upgrade from 19.07 :(

champtar:
See #16035 for the OpenSSL full featured version
This PR is missing `PKG_RELEASE` bump

mkrkn:
> Hi @mkrkn,
> 11 months ago maybe it was not clear it would break some people usage.
> Sorry if I rushed you, I just wanted to have an easy solution for all users as soon as possible, but this doesn't mean this weekend. 

Sorry, this timeline was not that easy to grasp. In the moment I had a very large pile of stuff suddenly drop on my figurative plate, and I've lost my own real-life use case for OpenVPN so my test setup is entirely virtual and a bit cumbersome. I thought I was quite clear in the commit message that this will break certain setups although I couldn't predict a number of affected users, and I thought this was settled as accepted back then.

> > I think it is unfair that all OpenWrt users with a sane OpenVPN setup must install liblzo (or compile from source if they don't want to) because users with _objectively bad_ setups are sticking to their footguns.
> 
> If you don't control the OpenVPN server then you can't upgrade from 19.07 :(

Yes you can, but you will need to build your own OpenVPN package. Which I don't find such a high bar to set for a deprecated and not-recommended feature that costs approx. 10% more in "total install size" (package plus all dependencies with mbed TLS), as I think of OpenWrt as a tool for tinkering and tweaking more than a full-fledged flashy all-in-one package. That said, OpenSSL is huge and liblzo doesn't matter in that case.

champtar:
Merged #16035, so closing this one

PR #16035, opened 2021-07-04
openvpn: enable LZO support by default for OpenSSL variant

champtar:
Maintainer: @mkrkn or none
Compile tested: no
Run tested: no
Description:
User that don't control both OpenVPN client and server
might still need LZO support, so keep it enable by default for at least
OpenSSL variant.


mkrkn:
Acked by me.

PR #16046, opened 2021-07-05
openssh: add HPN patch from https://sourceforge.net/projects/hpnssh/

jmarcet:
Maintainer: Peter Wagner <tripolar@gmx.at>
Compile tested: master x86_64
Run tested: master x86_64

Description:
I have an x86 server which acts as router and NAS. I just upgraded the storage to improve throughput while doing btrfs backups with [btrbk](https://github.com/digint/btrbk).

While the storage is now fast enough to saturate a 4gigabit link, doing a btrfs backup through ssh I barely reached 75mb/s. Testing with netcat instead of ssh saturated my 866.7 wifi link, in practice around 650mb/s, which reduced my backup/restoration times in more than 1/10th.

After searching about it I came across [HPN-SSH](https://sourceforge.net/projects/hpnssh/) which offers the option to disable both encryption and HMAC verification in `openssh-server`. More important than that, though, is that it fixes openssh's buffer handling, yielding a huge improvement in transfer times, so much so that it nearly matches transfers with netcat. This is in its default configuration, which has both the optional encryption and hmac verification disabled, i.e., it improves hugely just by fixing the buffers.

It is not a low quality patch either, Cisco paid for its development. All in all, I think this is a good fit for our openssh, after all it is not even a default package, since we have dropbear.

I also have the following in uci-defaults, although I'm not sure whether to include it in the PR:

```
$ cat etc/uci-defaults/sshd.sh
#!/bin/sh

sed -e 's:^#PasswordAuthentication yes:PasswordAuthentication no:' \
    -i /etc/ssh/sshd_config

chmod 700 /etc/ssh
chmod 644 /etc/ssh/*_config /etc/ssh*.pub
chmod 600 /etc/ssh/*key

exit 0
```

### DIFF
--- a/net/openssh/patches/openssh-8_6_P1-hpn-15.2.diff
+++ b/net/openssh/patches/openssh-8_6_P1-hpn-15.2.diff
@@ -1,0 +1,2678 @@
+--- /dev/null
++++ b/HPN-README
+@@ -0,0 +1,153 @@
++Notes:
++
++MULTI-THREADED CIPHER:
++The AES cipher in CTR mode has been multithreaded (MTR-AES-CTR). This will allow ssh installations
++on hosts with multiple cores to use more than one processing core during encryption.
++Tests have show significant throughput performance increases when using MTR-AES-CTR up
++to and including a full gigabit per second on quad core systems. It should be possible to
++achieve full line rate on dual core systems but OS and data management overhead makes this
++more difficult to achieve. The cipher stream from MTR-AES-CTR is entirely compatible with single
++thread AES-CTR (ST-AES-CTR) implementations and should be 100% backward compatible. Optimal
++performance requires the MTR-AES-CTR mode be enabled on both ends of the connection.
++The MTR-AES-CTR replaces ST-AES-CTR and is used in exactly the same way with the same
++nomenclature.
++Use examples:
++		ssh -caes128-ctr you@host.com
++		scp -oCipher=aes256-ctr file you@host.com:~/file
++
++NONE CIPHER:
++To use the NONE option you must have the NoneEnabled switch set on the server and
++you *must* have *both* NoneEnabled and NoneSwitch set to yes on the client. The NONE
++feature works with ALL ssh subsystems (as far as we can tell) *AS LONG AS* a tty is not
++spawned. If a user uses the -T switch to prevent a tty being created the NONE cipher will
++be disabled.
++
++The performance increase will only be as good as the network and TCP stack tuning
++on the reciever side of the connection allows. As a rule of thumb a user will need
++at least 10Mb/s connection with a 100ms RTT to see a doubling of performance. The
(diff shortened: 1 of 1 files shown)


jmarcet:
I do not understand the CI error.

What does it mean by saying `Dirty patches detected, please refresh and review the diff` ?

neheb:
make package/openssh/refresh

champtar:
Hi @jmarcet, when people install openssh they expect openssh, not some patched up version IMO.
I don't want to speak for @tripolar but I'm pretty sure this will be a nack.
Best way forward is to make another package named something like openssh-hpn, so you can also keep the openssh source version needed for the patch in sync (I see your patch is named 8.5 but current openssh version is 8.6)

jmarcet:
I refreshed the patch and renamed it accordingly.

Shall I make it a separate package? My rationale to patch the normal openssh-sever is:

1. it is an optional package, we only include dropbear by default
2. t is not a random patch, Cisco paid for its development. IMHO, I don't see any downside to the patch itself, it seems to be of high quality
3. The risky features, like disabling encryption must be manually enabled. By simply reworking the buffer handling I see impressive throughput improvements.

In brief, I would patch the normal package but you tell me.

cotequeiroz:
There's an official git repo for HPN openssh at https://github.com/rapier1/openssh-portable.  Consider creating a package based on that instead of keeping the patch in the openwrt tree.

champtar:
Thanks @cotequeiroz 
@jmarcet can you make a new package named openssh-*hpn based on this git repo ?

jmarcet:
@CodeFetch @champtar
Thanks for the pointer. I'll make the new package then.

stintel:
There's a reason the HPN patch is not in OpenSSH upstream though. I recall having issues connecting to a Huawei switch with OpenSSH+HPN, even with HPNDisabled=yes. You should probably try changing the order of encryption algorithms instead. Almost anything has accelerated AES these days, yet OpenSSH prefers chacha20-poly1305 over AES. Before spending more time on HPN, I suggest you try `ssh -o Ciphers aes128-gcm@openssh.com` first. In my testing, HPN did not improve speeds at all.

CodeFetch:
> After searching about it I came across [HPN-SSH](https://sourceforge.net/projects/hpnssh/) which offers the option to disable both encryption and HMAC verification in `openssh-server`.

That something like this exists is scary. I think SSH should never be able to turn off encryption.



> Cisco paid for its development

To me sounds like NSA has paid it.
I do not support the idea of adding this patch. Use FTP if you want flawed security, but don't break SSH. This patch is out-of-tree and both questionable security-wise from its source and from its "function".

jmarcet:
Well, after doing some more tests I really see no benefit from it at all, all the improvement came from openssh-server itself, with respect to dropbear, none from these patches.

So I close the PR and drop the patch from by builds.

PR #16052, opened 2021-07-06
[WIP] Support nginx with QUIC

martinschneider:

Maintainer: me
Compile tested: aarch64_cortex-a53, aarch_cortex-a72, OpenWrt 19.07.7
Run tested: WIP

Description: This is still highly experimental. While it works to build nginx with
QUIC support based on https://hg.nginx.org/nginx-quic/shortlog/quic, the
current draft simply replaces the stable nginx release with an
experimental one and offers no way to disable HTTP3/QUIC support. This
is, obviously, not the desired outcome.

There are two ways to continue (and I'm exploring both options at the
moment).

1. Make QUIC support configurable through a flag
2. Move nginx-quic to a separate package until there is support for QUIC
in an official nginx release

The challenge with the first approach is that the differences would be
more than passing an additional --with-http_v3_module option. It would
require pulling the sources from a different branch (or applying the
necessary changes as patches) and some minor tweaks in the build itself
(for example, nginx-quic, for some reason, has its configure file in a
sub-directory). I wonder whether we want to introduce this complexity in
the package.

Lastly, nginx-quic depends on BoringSSL while for most other use-cases
we probably want to stick with OpenSSL.

This depends on: https://github.com/openwrt/packages/pull/15822. See
also https://github.com/openwrt/openwrt/pull/4243 for further
background. The open PR for BoringSSL has been merged into this PR.


CodeFetch:
It's better to wait for an official nginx stable release with QUIC support as it is experimental and that's not what package repository users expect.

martinschneider:
> It's better to wait for an official nginx stable release with QUIC support as it is experimental and that's not what package repository users expect.

I tend to agree. Now, there is also a roadmap for this: https://www.nginx.com/blog/our-roadmap-quic-http-3-support-nginx. However, this will still depend on BoringSSL or https://github.com/quictls/openssl so we could probably prepare for it by driving https://github.com/openwrt/packages/pull/15822.

TheB1gG:
> > It's better to wait for an official nginx stable release with QUIC support as it is experimental and that's not what package repository users expect.
> 
> I tend to agree. Now, there is also a roadmap for this: https://www.nginx.com/blog/our-roadmap-quic-http-3-support-nginx. However, this will still depend on BoringSSL or https://github.com/quictls/openssl so we could probably prepare for it by driving #15822.

Since it got merged in nginx 1.25.0 it would be possible to build nginx with boringssl on stable packages now.
I think this change would be very desirable because OpenSSLs roadmaps says it will take them still years until they fully support http/3.

martinschneider:
@CodeFetch What do you think? I could spend some time to try with 1.25.0 and update the PR.

TheB1gG:
not sure if @CodeFetch is still active, so tagging @Ansuel because is one of the nginx maintainers so i guess can give advise here how to proceed

Ansuel:
Can you link patch needed for boringssl support? 

I would create a variant for it instead of replacing openssl. (I assume dynamic module should work with both variant?)

SopaDeMacaco-UmaDelicia:
Bump

martinschneider:
@Ansuel, BoringSSL support was merged in #15822

Ansuel:
@martinschneider what is the state of this? afaik nginx now have support for httpv3

martinschneider:
I'm not currently using this setup. I will look into it with the latest version of nginx when I find some time.

GeorgeSapkin:
@martinschneider any plans to pick this up?

GeorgeSapkin:
No updates in over four years, merge conflicts and still marked as draft. Closing as stale. Feel free to reopen as necessary.

PR #16069, opened 2021-07-07
libelektra: fix compilation with GCC11

neheb:
Missing header.

Maintainer: @haraldg
Compile tested: ath79


haraldg (approved):
Thanks for looking into this.

I already have a package for 0.9.6 locally, but it is blocked by an upstream bug. So let's include this fix for now.

Issue #14736, opened 2021-02-14
erlang: does not compile with hardening


BKPepe:
Maintainer: @cretingame
Compile tested: Turris Omnia, mvebu (cortex-a9), OpenWrt master and enabled hardening

Description:

```
arm-openwrt-linux-muslgnueabi-gcc: error: /foo/build/include/hardened-ld.specs: No such file or directory
make[7]: *** [arm-openwrt-linux-gnu/Makefile:574: //foo/build/build_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/otp_src_23.0/lib/erl_interface/bin/arm-openwrt-linux-gnu/erl_call] Error 1
make[7]: Leaving directory '/foo/build/build_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/otp_src_23.0/lib/erl_interface/src'
make[6]: *** [Makefile:31: opt] Error 2
```

I tried it locally with version 23.2, but the issue is same:
```
arm-openwrt-linux-muslgnueabi-gcc-ranlib /foo/build/build_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/otp_src_23.2/lib/erl_interface/obj/arm-openwrt-linux-gnu/libei.a
[...]

PR #16098, opened 2021-07-11
erlang: disable PIE

neheb:
Fails to compile with it on.

Maintainer: @cretingame
Compile tested: ath79


BKPepe:
This should fix: https://github.com/openwrt/packages/issues/14736

I can try it in my environment to see.

BKPepe (approved):
Yes, it fixes my issue, which I reported. Tested for Turris Omnia, OpenWrt 21.02-rc3. Thus approved.

PR #16099, opened 2021-07-11
flac: don't remove automake files

neheb:
Fixes compilation with glibc.

Maintainer: @thess 
Compile tested: ath79


neheb:
Huh. This seems to be needed with musl but not glibc... Very strange.

neheb:
OTOH switching to the CMake build would avoid this mess...

neheb:
CMake build has other problems fixed in master. Might as well leave it broken until a new release.

PR #16101, opened 2021-07-11
dawn: update to 2021-07-11

PolynomialDivision:
ec9a3a9 fix GCC11 compilation

Thanks to @neheb and @cotequeiroz.


BKPepe:
You should change your commit message as it is not good to see in the git history still the same as: "bump to the latest version".

PolynomialDivision:
> You should change your commit message as it is not good to see in the git history still the same as: "bump to the latest version".

done ;)

BKPepe:
👍

PR #16107, opened 2021-07-11
strongswan: update to version 5.9.3

Thermi:
Maintainer: me, @pprindeville
Compile tested: Not yet. Let's see.
Run tested: same

Description:

Update to latest release 5.9.3.


neheb:
Need to run mak package/strongswan/{clean,refresh}

Thermi:
Commands fail because there is no such target.

neheb:
make package/strongswan/{clean,refresh} works fine locally.

Here's the commit with that applied: https://github.com/neheb/packages/commit/8124e01b1026205daf2ad21c5c8408465fa3ccb2

Thermi:
Great. Here, running the commands eventually succeeds after feeds were updated, but there is no difference in the source tree.

neheb:
Right. but the CI demands cleaned patches.

Thermi:
What are cleaned patches exactly? Patches that apply without fuzz?

neheb:
The CI check basically runs all patches through quilt. If they differ, it errors.

Most "issues" here deal with diff entries.

Thermi:
Thank you. I figured out the problem. "packages" was in the wrong directory. Duh.

PR #16118, opened 2021-07-14
python-{simplejson,cffi,pillow}: update versions

commodo:
Maintainer: me, @jefferyto for cffi
Compile tested: x86 https://github.com/openwrt/openwrt/commit/1236cbe30cec8e3e8246237005140596f8611ce9
Run tested: x86 https://github.com/openwrt/openwrt/commit/1236cbe30cec8e3e8246237005140596f8611ce9

--------------------------------------------------------------


commodo:
I didn't seem to find any CVEs for these.
cffi and simplejson could also go into 21.02 though
pillow is 8.1.2 in 21.02 and in master it's 8.3.x ; has a few extra features;

jefferyto:
Thanks @commodo - would you mind also updating cffi in host-pip-requirements? :pray:

commodo:
> Thanks @commodo - would you mind also updating cffi in host-pip-requirements? pray

oh;
good point;
we need a CI/build mechanism to catch this, or keep it in sync;

commodo:
Changelog v1 -> v2:
* update cffi in host-pip-requirements

jefferyto (approved):
Thanks!

jefferyto:
> we need a CI/build mechanism to catch this, or keep it in sync;

A build/CI check would be great - I don't have any good ideas on how to do this in a general way though :joy:

commodo:
> > we need a CI/build mechanism to catch this, or keep it in sync;
> 
> A build/CI check would be great - I don't have any good ideas on how to do this in a general way though joy

the only quick way i can think of, is to centralize the versions in a file;
which sounds like adding a bit more fragmentation;

maybe the best solution would be to add a build check for updating files in `lang/python/host-pip-requirements/` during a package build;
i.e. something like "if (grep cffi) then (grep cffi==version)"

i can make a note of this, but i don't think i can do it this week;

jefferyto:
> maybe the best solution would be to add a build check for updating files in lang/python/host-pip-requirements/ during a package build;
> i.e. something like "if (grep cffi) then (grep cffi==version)"

Sounds good to me :smile: 

> i can make a note of this, but i don't think i can do it this week;

No need to rush on this :joy: thanks again!

commodo:
I wonder if the mips build failures is due to long build time

jefferyto:
From the log:

```
2021-07-14T08:55:14.1943033Z     ERROR: package/feeds/packages_ci/gdbm failed to build.
2021-07-14T08:55:58.6932640Z make: *** [/home/build/openwrt/include/toplevel.mk:223: package/python-simplejson/compile] Error 1
2021-07-14T08:55:58.6934237Z make -r package/python-simplejson/compile: build failed. Please re-run make with -j1 V=s or V=sc for a higher verbosity level to see what's going on
```

Further down the log where I think the CI does `make -j1` to get an error log, gdbm compiled successfully. Not sure what is going on.

neheb:
Probably a parallel build problem. Anyway, not relevant here.

PR #16123, opened 2021-07-15
lxc: update to 4.0.10

neheb:
Remove getline patch. It seems to be for uClibc, which is no longer in
the tree.

Remove commands patch. Issue was fixed upstream. Same with the tests
patch.

Refreshed others.

Maintainer: @ratkaj 
Compile tested: ath79


neheb:
Updated to .10. One more patch removed.

PR #16125, opened 2021-07-15
[openwrt-21.02] addrwatch: Various fixes

jefferyto:
Maintainer: @oskar456 
Compile tested: none (cherry picked from #15899)
Run tested: none

Description:
Makefile changes include:

* Remove `USE_UCLIBC`, as uclibc is no longer supported

* Package output modules

* Move main binary (back) to /usr/sbin, as it is system administration related and requires superuser privileges

New patches:

* 003-add-space-for-null-byte.patch - from https://github.com/fln/addrwatch/commit/374cfd2cabe4db9882d8a210adff430cc579f859

* 004-more-specific-library-linking.patch - from https://github.com/fln/addrwatch/commit/27b57d9da322fc16c6904d8e35aae4557a3e517b

* 005-use-c99-format-macro-constants.patch - from https://github.com/fln/addrwatch/pull/28

Init script changes include:

* Change from explicit disable to explicit enable, so that the service is disabled by default and on first install

* Set config option default values to default values of the main binary

* Fix command-line option names and format (from https://forum.openwrt.org/t/cant-start-addrwatch-service/60499/3)

* Always use the `--quiet` command-line option, as the procd instance is not configured to capture stdout/stderr

* Change the syslog config option to start the syslog output module

(cherry picked from commit 31ae85bca963ce240c9e1b8df55294587b079161)

### DIFF
--- a/net/addrwatch/Makefile
+++ b/net/addrwatch/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=addrwatch
 PKG_VERSION:=1.0.2
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://github.com/fln/addrwatch/releases/download/v$(PKG_VERSION)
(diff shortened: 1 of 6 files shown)


oskar456:
Thank you for your effort!

PR #16126, opened 2021-07-15
openvpn: add OpenVPN option push-peer-info

minhng99:
This will allow the server to know more info about the client like
HWADDR, very useful for managing IoT devices.

See: https://www.mankier.com/8/openvpn#--push-peer-info

Signed-off-by: Nguyen Quang Minh <minhnq31@fpt.com.vn>

Maintainer: Magnus Kroken <mkroken@gmail.com>
Compile tested: EA8300, ARM, master branch
Run tested: EA8300, ARM, master branch, server now receive HWADDR containing the client's gateway MAC address when connected

Description:
This option could be backported to OpenWrt-19.07, which seems to be supported at least way back in 2014.

### DIFF
--- a/net/openvpn/Makefile
+++ b/net/openvpn/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=openvpn
 
 PKG_VERSION:=2.5.3
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_URL:=\
 	https://build.openvpn.net/downloads/releases/ \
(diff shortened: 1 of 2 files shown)


mkrkn:
Feature-ack from me, but you also need to bump PKG_RELEASE in the Makefile, due to the files installed on-device changes.

The change is small and quite safe to backport to 21.02. 19.07 is about to reach end-of-life with the release of 21.02, so while it would work I don't see a strong case for backporting it to 19.07.

minhng99:
Ah, make sense, first time contributor, sorry :)

PR #16140, opened 2021-07-16
travelmate: fix for meta-refresh redirect on captive portal 

BennoGAP:
Maintainer: @dibdot 
Run tested: 21.02.0-rc3

Description:
Current captive portal detection works for url redirect. This does not work on captive portals who work with "<meta http-equiv=refresh..." redirections. When using the Google-Service for connectivity check which return 204 on success we can check if we only get a 200-response and filter out the url from there.

Signed-off-by: BennoGAP <15093499+BennoGAP@users.noreply.github.com>

### DIFF
--- a/net/travelmate/files/travelmate.sh
+++ b/net/travelmate/files/travelmate.sh
@@ -523,9 +523,10 @@ f_prepif()
 #
 f_net()
 {
-	local IFS err err_rc err_domain json_raw json_cp json_rc cp_domain result="net nok"
+	local IFS err err_rc err_domain json_raw json_tmp json_cp json_rc cp_domain result="net nok"
 
 	json_raw="$(${trm_fetch} --user-agent "${trm_useragent}" --referer "http://www.example.com" --write-out "%{json}" --silent --show-error --connect-timeout $((trm_maxwait/10)) "${trm_captiveurl}" 2>/tmp/trm_fetch.err)"
+	json_tmp="${json_raw%%\{*}"
 	json_raw="${json_raw#*\{}"
 	if [ -s "/tmp/trm_fetch.err" ]
 	then
@@ -548,6 +549,15 @@ f_net()
 			cp_domain="${json_cp#http*://*}"
 			cp_domain="${cp_domain%%/*}"
 			result="net cp '${cp_domain}'"
+		elif [ "${trm_captiveurl}" = "http://connectivitycheck.android.com/generate_204" ] && [ "${json_rc}" = "200" ]
+                then
+                        cp_domain="${json_tmp#*url=}"
+                        cp_domain="${cp_domain#http*://*}"
+                        cp_domain="${cp_domain%%/*}"
+                        if [ -n "${cp_domain}" ]
+                        then
+                                result="net cp '${cp_domain}'"
+                        fi
 		else
 			if [ "${json_rc}" = "200" ] || [ "${json_rc}" = "204" ]
 			then


dibdot:
Please provide an example captive portal and the "$json_raw" output of such portal - thanks.

BennoGAP:
The page would look like that
`<html>
<head>
<meta http-equiv="refresh" content="0; url=https://gwhs.logostech.it/login?dst=http%3A%2F%2Fconnectivitycheck.android.com%2Fgenerate%5F204"/>
</head>
</html>`

The output from json_raw would be
`<html>
<head>
<meta http-equiv="refresh" content="0; url=https://gwhs.logostech.it/login?dst=http%3A%2F%2Fconnectivitycheck.android.com%2Fgenerate%5F204"/>
</head>
</html>{"content_type":"text/html","errormsg":null,"exitcode":0,"filename_effective":null,"ftp_entry_path":null,"http_code":200,"http_connect":0,"http_version":"1.1","local_ip":"10.80.0.78","local_port":59504,"method":"GET","num_connects":1,"num_headers":8,"num_redirects":0,"proxy_ssl_verify_result":0,"redirect_url":null,"referer":"http://www.example.com","remote_ip":"10.80.0.1","remote_port":80,"response_code":200,"scheme":"HTTP","size_download":171,"size_header":242,"size_request":121,"size_upload":0,"speed_download":833,"speed_upload":0,"ssl_verify_result":0,"time_appconnect":0.000000,"time_connect":0.114974,"time_namelookup":0.065601,"time_pretransfer":0.115638,"time_redirect":0.000000,"time_starttransfer":0.203377,"time_total":0.205267,"url":"http://connectivitycheck.android.com/generate_204","url_effective":"http://connectivitycheck.android.com/generate_204","urlnum":0,"curl_version":"libcurl/7.77.0 wolfSSL/4.7.0"}`

As there is no redirect_url set, the json_cp will be null and as the http_code is 200 - so the result will be "net ok".

dibdot:
Thanks, could you please repeat this test with the default captive provider (captive.apple.com) and post the results as well?

BennoGAP:
Unfortunately i am not at that location anymore - but the result had looked completely the same (expect the URL of course).
Every external site had a 200 reply and the html code like above.
The reason i decided here for the google check is the fact, that a 204 is returned if connected to the internet (unlike the apple page, which would also response with a 200).

BennoGAP:
To get more independent from the used connectivity-check-provider, there could also be a check, if some "refresh" code is on the page itself (which would never happen on any of the connectivity-check sites)

```
elif [[ "${json_tmp}" =~ "<meta http-equiv=\"refresh\"" ]]
```

dibdot:
Hi, currently I'm working on travelmate 2.0.4 and I've implemented your requirement a little bit differently:

- not limited to a special captive portal provider
- responds only to the html meta refresh tag
- more fault tolerant in parsing the html stuff
- distinguish between json_cp and html_cp (mainly for debugging/support reasons)

For reference the enhanced f_net function from travelmate 2.0.4. Maybe you could test this next time when you arrive at the special cp again - thanks.

```
# check net status
#
f_net()
{
	local IFS err err_rc err_domain raw html_raw html_cp json_raw json_cp json_rc result="net nok"

	raw="$(${trm_fetch} --user-agent "${trm_useragent}" --referer "http://www.example.com" --write-out "%{json}" --silent --show-error --connect-timeout $((trm_maxwait / 10)) "${trm_captiveurl}" 2>/tmp/trm_fetch.err)"
	json_raw="${raw#*\{}"
	html_raw="${raw%%\{*}"
	if [ -s "/tmp/trm_fetch.err" ]; then
		err="$(awk 'BEGIN{FS="[()'\'' ]"}{printf "%s %s",$3,$(NF-1)}' "/tmp/trm_fetch.err")"
		err_rc="${err% *}"
		err_domain="${err#* }"
		if [ "${err_rc}" = "6" ]; then
			if [ -n "${err_domain}" ] && [ "${err_domain}" != "timed" ] && [ "${err_domain}" != "${trm_captiveurl#http*://*}" ]; then
				result="net cp '${err_domain}'"
			fi
		fi
	elif [ -n "${json_raw}" ]; then
		json_cp="$(printf "%s" "{${json_raw}" | jsonfilter -l1 -e '@.redirect_url' 2>/dev/null | awk 'BEGIN{FS="/"}{printf "%s",$3}')"
		json_rc="$(printf "%s" "{${json_raw}" | jsonfilter -l1 -e '@.response_code' 2>/dev/null)"
		if [ -n "${json_cp}" ]; then
			result="net cp '${json_cp}'"
		else
			if [ "${json_rc}" = "200" ] || [ "${json_rc}" = "204" ]; then
				html_cp="$(printf "%s" "${html_raw}" | awk 'match(tolower($0),/^<meta[ ]*http-equiv=["]*refresh.* url=/){print substr(tolower($0),RLENGTH+1)}' | awk 'BEGIN{FS="/"}{printf "%s",$3;exit}')"
				if [ -n "${html_cp}" ]; then
					result="net cp '${html_cp}'"
				else
					result="net ok"
				fi
			fi
		fi
	fi
	rm -f "/tmp/trm_fetch.err"
	printf "%s" "${result}"
	f_log "debug" "f_net     ::: fetch: ${trm_fetch}, timeout: $((trm_maxwait / 6)), url: ${trm_captiveurl}, user_agent: ${trm_useragent}, cp (json/html): ${json_cp:-"-"}/${html_cp:-"-"}, result: ${result}, error: ${err:-"-"}"
}
```

BennoGAP:
Hi, thanks for implementing that so fast.
As i am not sure when i'll be at that cp again and in some quick tests this worked as intended i consider that as closed for now and will report back if neccessary.

BennoGAP:
Hi,
i found another portal which is using the same method, but the fixed version will not work there, as there is no newline before the <meta tag. Also there is now a port part of the cp-url, which could lead to a problem to the "dhcp rebind whitelist".

```
<HTML><HEAD><TITLE> Web Authentication Redirect</TITLE><META http-equiv="Cache-control" content="no-cache"><META http-equiv="Pragma" content="no-cache"><META http-equiv="Expires" content="-1"><META http-equiv="refresh" content="1; URL=https://portal.foobar.de:8443/portal/gateway?sessionId=123456789012345678901234567890123&portal=12345678-1234-1234-1234-123456789012&action=cwa&token=12345678901234567890123456789012&redirect=google.de/"></HEAD></HTML>
{"content_type":"text/html","errormsg":null,"exitcode":0,"filename_effective":null,"ftp_entry_path":null,"http_code":200,"http_connect":0,"http_version":"1.1","local_ip":"100.68.240.18","local_port":40338,"method":"GET","num_connects":1,"num_headers":3,"num_redirects":0,"proxy_ssl_verify_result":0,"redirect_url":null,"referer":null,"remote_ip":"200.50.200.131","remote_port":80,"response_code":200,"scheme":"HTTP","size_download":451,"size_header":275,"size_request":73,"size_upload":0,"speed_download":177,"speed_upload":0,"ssl_verify_result":0,"time_appconnect":0.000000,"time_connect":2.538305,"time_namelookup":2.534619,"time_pretransfer":2.538905,"time_redirect":0.000000,"time_starttransfer":2.542009,"time_total":2.543835,"url":"http://google.de","url_effective":"http://google.de/","urlnum":0,"curl_version":"libcurl/7.77.0 wolfSSL/4.7.0"}
```

dibdot:
Sigh, what a mess! :smile: 
Please try fixed regex (simply replace that line in your script):
```
html_cp="$(printf "%s" "${html_raw}" | awk 'match(tolower($0),/^.*<meta[ \t]+http-equiv=["]*refresh.*[ \t;]url=/){print substr(tolower($0),RLENGTH+1)}' | awk 'BEGIN{FS="[:/]"}{printf "%s",$4;exit}')"
```

PR #16172, opened 2021-07-21
irssi: update to 1.2.3

neheb:
Switch to AUTORELEASE for simplicity.

Maintainer: @tripolar 
Compile tested: ath79


BKPepe:
We should add `test.sh` to have it run tested. But your PR looks good as it is. I can do it in the next PR.

BKPepe (approved):

PR #16198, opened 2021-07-22
dnsproxy: Update to 0.39.0 and add init script

1715173329:
Maintainer: me
Compile tested: ipq807x, rockchip
Run tested: rk3328 nanopi-r2s

Description:
- Updated to latest 0.39.0 release
- Added basic init script

### DIFF
--- a/net/dnsproxy/files/dnsproxy.config
+++ b/net/dnsproxy/files/dnsproxy.config
@@ -1,0 +1,41 @@
+
+# For documents, please see https://github.com/AdguardTeam/dnsproxy#usage
+
+config dnsproxy 'global'
+	option enabled '0'
+	option listen_addr '127.0.0.1'
+	option listen_port '5353'
+	option log_file ''
+	option all_servers '0'
+	option fastest_addr '0'
+	option insecure '0'
+	option ipv6_disabled '0'
+	option max_go_routines ''
+	option rate_limit ''
+	option refuse_any '0'
+	option udp_buf_size ''
+	option verbose '0'
+
+config dnsproxy 'bogus_nxdomain'
+	list ip_addr ''
+
+config dnsproxy 'cache'
+	option enabled '0'
+	option cache_optimistic '0'
+	option size '65535'
+	option min_ttl ''
+	option max_ttl ''
+
+config dnsproxy 'dns64'
+	option enabled '0'
+	option dns64_prefix '64:ff9b::'
+
+config dnsproxy 'edns'
+	option enabled '0'
+	option edns_addr ''
+
+config dnsproxy 'servers'
+	list bootstrap 'tls://8.8.8.8'
+	list fallback 'tls://9.9.9.9'
+	list upstream 'tls://1.1.1.1'
+
--- a/net/dnsproxy/files/dnsproxy.init
+++ b/net/dnsproxy/files/dnsproxy.init
@@ -1,0 +1,115 @@
+#!/bin/sh /etc/rc.common
+# Copyright (C) 2021 Tianling Shen <cnsztl@immortalwrt.org>
+
+USE_PROCD=1
+START=90
+
+CONF="dnsproxy"
+PROG="/usr/bin/dnsproxy"
+
+is_enabled() {
+	local enabled
+	config_get enabled "$1" "$2" "0"
+	if [ "$enabled" -eq "1" ]; then
+		return 0
+	else
+		return 1
+	fi
+}
+
+is_empty() {
+	local empty
+	config_get empty "$1" "$2" $3
+	if [ -z "$empty" ]; then
+		return 0
+	else
+		return 1
+	fi
+}
+
+append_param() {
+	procd_append_param command "$1" $2
+}
+
+append_param_arg() {
+	local value
+	config_get value "$1" "$2" $4
+	[ -n "$value" ] && append_param "$3" "$value"
+}
+
+append_param_bool() {
+	is_enabled "$1" "$2" && append_param "--${2//_/-}"
+}
+
+load_config_arg() {
+	append_param_bool "$1" "all_servers"
+	append_param_bool "$1" "fastest_addr"
+	append_param_bool "$1" "insecure"
+	append_param_bool "$1" "ipv6_disabled"
+	append_param_bool "$1" "refuse_any"
+	append_param_bool "$1" "verbose"
+}
+
+load_config_list() {
+	is_empty "bogus_nxdomain" "ip_addr" || config_list_foreach "bogus_nxdomain" "ip_addr" "append_param '--bogus-nxdomain'"
+
+	for i in "bootstrap" "fallback" "upstream"; do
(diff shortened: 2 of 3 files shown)


1715173329:
Related discussion: AdguardTeam/dnsproxy#111

Cc: @dhewg

net/dnsproxy/files/dnsproxy.config
```diff
@@ -0,0 +1,42 @@
+
+# For documents, please see https://github.com/AdguardTeam/dnsproxy#usage
+
+config dnsproxy 'enabled'
+	option enabled '0'
```
dhewg, line 5: splitting this seems excessive, maybe rename `config` to `global` and move this there?

net/dnsproxy/files/dnsproxy.config
```diff
@@ -0,0 +1,42 @@
...
+	option insecure '0'
+	option ipv6_disabled '0'
+	option max_go_routines ''
+	option rate_limit ''
+	option refuse_any '0'
+	option udp_buf_size ''
+	option verbose '0'
+
+config dnsproxy 'bogus_nxdomain'
+	list ip_addr ''
```
dhewg, line 22: This could be `list bogus_nxdomain` under `config`/`global` since there's just this one knob

net/dnsproxy/files/dnsproxy.init
```diff
@@ -0,0 +1,114 @@
...
+
+	is_enabled "enabled" "enabled" || exit 1
+
+	procd_open_instance "$CONF"
+	procd_set_param command "$PROG"
+
+	load_config_arg "config"
+	load_config_list
+	load_config_param
+
```
dhewg, line 98: This is a network daemon with an attack surface, let's run it under it's own user account (like stubby)?

`procd_set_param user dnsproxy`

net/dnsproxy/files/dnsproxy.init
```diff
@@ -0,0 +1,114 @@
...
+	procd_set_param stderr 1
+	procd_set_param respawn
+
+	procd_close_instance
+}
+
+reload_service() {
+	stop
+	sleep 2s
+	start
```
dhewg, line 109: Is that really required?

net/dnsproxy/files/dnsproxy.init
```diff
@@ -0,0 +1,114 @@
...
+	append_param_arg "config" "log_file" "--output"
+	append_param_arg "config" "max_go_routines" "--max-go-routines"
+	append_param_arg "config" "rate_limit" "--ratelimit"
+	append_param_arg "config" "udp_buf_size" "--udp-buf-size"
+
+	is_enabled "cache" "enabled" && {
+		append_param "--cache"
+		append_param_arg "cache" "size" "--cache-size"
+		append_param_arg "cache" "min_ttl" "--cache-min-ttl"
+		append_param_arg "cache" "max_ttl" "--cache-max-ttl"
```
dhewg, line 73: Not using this cache, but what about `--cache-optimistic`?

net/dnsproxy/files/dnsproxy.init
```diff
@@ -0,0 +1,114 @@
...
+	append_param_bool "$1" "ipv6_disabled"
+	append_param_bool "$1" "refuse_any"
+	append_param_bool "$1" "verbose"
+}
+
+load_config_list() {
+	is_empty "bogus_nxdomain" "ip_addr" || config_list_foreach "bogus_nxdomain" "ip_addr" "append_param '--bogus-nxdomain'"
+
+	for i in "bootstrap" "fallback" "upstream"; do
+		is_empty "servers" "$i" || config_list_foreach "servers" "$i" "append_param '--$i'"
```
dhewg, line 57: Not using dnscrypt myself, but is `--dnscrypt-config` required for using that protocol here?
Maybe ship upstream's `testdata/dnscrypt-cert.*.txt` in `/etc/dnscrypt`?

dhewg (changes requested):
Nice, thanks for this!
Left some comments ;)

net/dnsproxy/files/dnsproxy.init
1715173329: > Not using this cache, but what about --cache-optimistic?

It's a lack of documentation.. I will add this later.

net/dnsproxy/files/dnsproxy.init
1715173329: > Is that really required?

dnsproxy seems not to support reload, so use legacy restart.

net/dnsproxy/files/dnsproxy.init
1715173329: > Not using dnscrypt myself, but is --dnscrypt-config required for using that protocol here?

It is required only when run as a server.

net/dnsproxy/files/dnsproxy.init
dhewg: Maybe I misremember, but removing `reload_service` should make it do exactly that?
But if not, do you need that delay?

net/dnsproxy/files/dnsproxy.init
dhewg: Alright, thanks for checking

net/dnsproxy/files/dnsproxy.init
1715173329: > Maybe I misremember, but removing reload_service should make it do exactly that?

Removing reload() in legacy init will do that, but in procd, removing reload_service() seems not.
Referred from: https://openwrt.org/docs/guide-developer/procd-init-scripts#forcing_service_restart.

> But if not, do you need that delay?

The delay seems not to be necessary, I will remove it.
Thanks ;)

net/dnsproxy/files/dnsproxy.init
1715173329: Great idea.

net/dnsproxy/files/dnsproxy.config
1715173329: I think splitting this could be more intuitive and clean (list vs. option).

net/dnsproxy/files/dnsproxy.config
1715173329: Done.

net/dnsproxy/files/dnsproxy.config
dhewg: You could of course have multiple 'list nxdomain' entries under 'global', which would also get rid of `ip_addr`, but this works too ;)

dhewg:
Looks pretty good to me now, but you're not creating the user/group it's now using.
The Makefile misses a `USERID:=dnsproxy=411:dnsproxy=411` or similar

1715173329:
> Looks pretty good to me now, but you're not creating the user/group it's now using.
> The Makefile misses a `USERID:=dnsproxy=411:dnsproxy=411` or similar

Thank you so much! It's new to me.

dhewg:
LGTM and it even works fine for me, thanks!

dhewg (approved):

PR #16210, opened 2021-07-23
minidlna: add patch to be compatible with Android DLNA Channels

jmarcet:
Maintainer: No one
Compile tested: master x86_64
Run tested: master x86_64

Description:
As is, minidlna does not work with [Android DLNA Channels](https://play.google.com/store/apps/details?id=at.nullptr.dlnachannels), a nice little app for Android TV which shows DLNA folders as content in the Android TV Home screen. Seemingly the error should only be a warning but in practice they cannot be used together.

With the attached patch, they not just work together, minidlna is by far the best backend I've tried it with.

### DIFF
--- a/multimedia/minidlna/patches/999-force-compatibility-android_channels.patch
+++ b/multimedia/minidlna/patches/999-force-compatibility-android_channels.patch
@@ -1,0 +1,19 @@
+--- ./minissdp.c-orig	2020-11-24 19:53:50.000000000 +0100
++++ ./minissdp.c	2021-07-06 11:05:37.422070969 +0200
+@@ -650,11 +650,11 @@ ProcessSSDPRequest(struct event *ev)
+ 			DPRINTF(E_INFO, L_SSDP, "WARNING: Ignoring invalid SSDP M-SEARCH from %s [bad source port %d]\n",
+ 				inet_ntoa(sendername.sin_addr), ntohs(sendername.sin_port));
+ 		}
+-		else if (!man || (strncmp(man, "\"ssdp:discover\"", 15) != 0))
+-		{
+-			DPRINTF(E_INFO, L_SSDP, "WARNING: Ignoring invalid SSDP M-SEARCH from %s [bad %s header '%.*s']\n",
+-				inet_ntoa(sendername.sin_addr), "MAN", man_len, man);
+-		}
++//		else if (!man || (strncmp(man, "\"ssdp:discover\"", 15) != 0))
++//		{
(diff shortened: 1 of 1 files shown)


multimedia/minidlna/patches/999-force-compatibility-android_channels.patch
```diff
@@ -0,0 +1,19 @@
+--- ./minissdp.c-orig	2020-11-24 19:53:50.000000000 +0100
++++ ./minissdp.c	2021-07-06 11:05:37.422070969 +0200
```
BKPepe, line 2: Here I am missing details:
- Git commit author
- Git commit description

And if this patch was sent upstream or not. Because from your changes it is not clear why this is necessary at all.

jmarcet: It does what I said, makes it compatible with the closed source `DLNA Channels`, it simply fails to get content from `minidlnad`. The warning was the only thing I could see in the logs, I tried commenting the conditional out and indeed it fixed the issue.

I haven't tried pushing it upstream, it is definitely worth the shot, that would be the really right place.

BKPepe: You said it well in the pull request description. In the commit subject, there is just to be compatible with Android DLNA Channel, but the description in the patch itself is missing.

multimedia/minidlna/patches/999-force-compatibility-android_channels.patch
```diff
@@ -0,0 +1,19 @@
...
+@@ -650,11 +650,11 @@ ProcessSSDPRequest(struct event *ev)
+ 			DPRINTF(E_INFO, L_SSDP, "WARNING: Ignoring invalid SSDP M-SEARCH from %s [bad source port %d]\n",
+ 				inet_ntoa(sendername.sin_addr), ntohs(sendername.sin_port));
+ 		}
+-		else if (!man || (strncmp(man, "\"ssdp:discover\"", 15) != 0))
+-		{
+-			DPRINTF(E_INFO, L_SSDP, "WARNING: Ignoring invalid SSDP M-SEARCH from %s [bad %s header '%.*s']\n",
+-				inet_ntoa(sendername.sin_addr), "MAN", man_len, man);
+-		}
++//		else if (!man || (strncmp(man, "\"ssdp:discover\"", 15) != 0))
```
neheb, line 12: Instead of commenting out, why not else if (0) ?

neheb:
An alternative to minidlna is gerbera, also available in packages.

BKPepe:
Closing due to inactivity.

PR #16227, opened 2021-07-26
pdns: Update to 4.5.1

WBNet-Wout:
Maintainer: @Habbie 
Compile tested: ramips (MIPS), HLK-7621 evaluation board, OpenWrt commit 8c2509dc5ff72ca47bd0787da407f5adc051a2b9
Run tested: ramips (MIPS), HLK-7621 evaluation board, OpenWrt commit 8c2509dc5ff72ca47bd0787da407f5adc051a2b9, ran `pdns_server --launch=bind --local-port=5353` and checked that it responded with `REFUSED` when `dig`ging random domains

Description:
This PR updates PowerDNS Authoritative Server to its latest version, v4.5.1.


Habbie (approved):

Habbie:
Thanks! LGTM

BKPepe (changes requested):
I am so sorry! Even maintainer approved it, you are doing so many changes, which are not documented anywhere, and because of that, I can not merge it.

You said in the PR subject and in the commit subject that it is updated to update 4.5.1, which is not correct.
Why there was it necessary to add `TARGET_CXX` in Makefile? You added a new patch **010-time_t_check.patch**.
**Is it backport?**
- If yes, then commit hash, commit message, commit author, commit description are missing!
- If no, why is it necessary at all? Are you going to upstream it?

Habbie:
> Why there was it necessary to add `TARGET_CXX` in Makefile?

OpenWrt overrides CXX, breaking C++17 compilation. The same override is present in the dnsdist package (which shares some code with pdns).

> You added a new patch **010-time_t_check.patch**.

This is also present in dnsdist and pdns-recursor. Upstream (which I represent) requires a 64 bit time_t. Until OpenWrt upgrades musl, we patch that check out so that the software can run on various 32 bit platforms.

WBNet-Wout:
I had already discussed these changes with @Habbie on IRC.

- `TARGET_CXX` is needed since PowerDNS now uses some C++17 features.
- That new patch is also there for the other PowerDNS packages and it disabled the check for a 64-bit `time_t`, which is needed because OpenWrt still runs on quite some 32-bit devices
- The other patch was just refreshed and fixed for the new PowerDNS version
- `pdns.conf-dist` was updated to reflect changes in the upstream one

neheb:
his point is to add those comments to the commit message.

neheb:
@Habbie that's curious that upstream doesn't mandate C++17 even though they use such features. I do remember they use boost variants of std functionality.

Habbie:
> @Habbie that's curious that upstream doesn't mandate C++17 even though they use such features. I do remember they use boost variants of std functionality.

We (upstream) do mandate it, check for it during configure, and abort configure if it's not present. However, something that OpenWrt does to override CXX or CXXFLAGS destroys the output of that check (which is adding `-std=c++17` to `CXX`). This might be a bug on our end, but as it's only happening on OpenWrt, and we see that we are not the only affected upstream, we have not investigated yet.

neheb:
Strange. Probably some autotools quirk. Then again, autotools should not be used :)

WBNet-Wout:
I have updated the commit message.

BKPepe (approved):

PR #16228, opened 2021-07-26
openvswitch: support setting OpenFlow datapath ID

stintel:
Maintainer: me / @yousong 
Compile tested: NA
Run tested: OpenWrt master r17170-86a61e716e on x86/64

Description:
By default, Open vSwitch will generate the OpenFlow datapath ID of a
bridge based on the MAC address of one of its ports. Due to this, it's
possible that the datapath ID changes when new ports are added. When the
datapath ID changes, Open vSwitch disconnects from the controller, as
there is no way to notify the controller that the datapath ID has
changed.

Add an option to set the datapath ID so that the above situation can be
avoided. The option either takes 16 hex characters, or when prefixed
with @, the name of a UCI interface section. In the latter case, it will
lookup the MAC address of the interface, and prefix it with 6f77, which
is hex for "ow".

### DIFF
--- a/net/openvswitch/files/openvswitch.init
+++ b/net/openvswitch/files/openvswitch.init
@@ -171,18 +171,39 @@ ovs_bridge_port_cleanup() {
 	done
 }
 
+ovs_bridge_validate_datapath_id() {
+	local dpid="$1"
+
+	if expr "$dpid" : '[[:xdigit:]]\{16\}$' > /dev/null; then
+		return 0
+	elif expr "$dpid" : '0x[[:xdigit:]]\{1,16\}$' > /dev/null; then
+		return 0
+	else
+		logger -t openvswitch "invalid datapath_id: $dpid"
+		return 1
+	fi
+}
+
 ovs_bridge_init() {
 	local cfg="$1"
 
 	local disabled
 	local name
 	local controller
+	local datapath_id
 
 	config_get_bool disabled "$cfg" disabled 0
 	[ "$disabled" == "0" ] || return
 
 	config_get name "$cfg" name $cfg
 	ovs-vsctl --may-exist add-br "$name"
+
+	config_get datapath_id "$cfg" datapath_id
+	[ -n "$datapath_id" ] && {
+		ovs_bridge_validate_datapath_id "$datapath_id" && {
+			ovs-vsctl --if-exists set bridge "$name" other-config:datapath-id="$datapath_id"
+		}
+	}
 
 	config_list_foreach "$cfg" "ports" ovs_bridge_port_add
 	config_foreach ovs_bridge_port_add_complex ovs_port "$name"
--- a/net/openvswitch/Makefile
+++ b/net/openvswitch/Makefile
@@ -17,7 +17,7 @@ include ./openvswitch.mk
 #
 PKG_NAME:=openvswitch
 PKG_VERSION:=$(ovs_version)
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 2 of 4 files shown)


net/openvswitch/files/openvswitch.init
```diff
@@ -95,6 +95,29 @@ ovs_xx() {
...
+	local dpid dpif dpnet
+
+	case "$1" in
+		@*)
+			dpnet="${1##@}"
+			dpif="$(uci -q get network."${dpnet}".device)"
+			dpid="6f77$(cat /sys/class/net/${dpif}/address 2>/dev/null | sed 's/://g')"
+			;;
+		*)
+			dpif="$1"
```
yousong, line 108: typo? `s/dpif/dpid/`

stintel: Good catch. I made the error in my draft, then fixed it, then reworked everything again and made the same mistake. 🤦🏻

net/openvswitch/files/openvswitch.init
```diff
@@ -95,6 +95,29 @@ ovs_xx() {
...
 }
 
+ovs_bridge_parse_datapath_id() {
+	local dpid dpif dpnet
+
+	case "$1" in
+		@*)
+			dpnet="${1##@}"
+			dpif="$(uci -q get network."${dpnet}".device)"
+			dpid="6f77$(cat /sys/class/net/${dpif}/address 2>/dev/null | sed 's/://g')"
```
yousong, line 105: Why `6f77`? I am being curious ;)

stintel: It's explained at the end of the commit message - hex for "ow".

stintel: Maybe better to change it to 4f 57 (OW), as we write OpenWrt ;-)

yousong: Or use mac address as is by prefixing with `0x`.  All okay with me.

> The value must either have a 0x prefix or be exactly 16 hex digits long.

The other thing is that maybe we could directly specify linux net device name after `@`, instead of netifd interface name.  It's more straightforward.

stintel: Didn't know that value could have a 0x prefix and not be 16 digits long. I'll have to think about this again, if this is the case, we might report a valid value as invalid.

As for the linux net device name after @, I used UCI interface config section name, as we use this already in netifd (e.g. for wan6 we use @wan by default). Or did this change in the recent netifd changes?

yousong: > As for the linux net device name after @, I used UCI interface config section name, as we use this already in netifd (e.g. for wan6 we use @wan by default). Or did this change in the recent netifd changes?

Likely `@wan` works just like before.  It's just that if we want value from a linux net device, use it directly.  No need for a redirection though network uci config.

It may worth mentioning that mac address of ovs bridge itself may change before and after the first netdevice with non-local mac address is added as port.  In this case mac address of the bridge can be somewhat "volatile".  On the other hand, member net devices of a bridge usually do not appear as value of `device` option of a netifd network interface section.

stintel: In that case I'd rather leave this part out completely, and only offer the user the possibility to enter a valid datapath ID. This will avoid confusing by @ doing one thing here, and another thing in /etc/config/network. This also saves us the trouble of having to warn the user about using interfaces with possibly volatile MAC addresses. The whole idea of adding this feature is to avoid the datapath ID changing at all. So maybe let's keep it simple. What do you think?

yousong: Yeah, start small and add when needed and ready

yousong (approved):
LGTM

blogic (approved):

PR #16230, opened 2021-07-27
transmission: apply DSCP to UDP sockets too

rsalvaterra:
Maintainer: @dangowrt (?)
Compile tested: mvebu/cortexa9 (Turris Omnia)

Backport a pending patch in order to DSCP-mark UDP traffic. This allows for
correct binning of traffic in diffserv-capable routers.

### DIFF
--- a/net/transmission/Makefile
+++ b/net/transmission/Makefile
@@ -9,13 +9,12 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=transmission
 PKG_VERSION:=3.00
-PKG_RELEASE:=12
+PKG_RELEASE:=13
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@GITHUB/transmission/transmission-releases/master
 PKG_HASH:=9144652fe742f7f7dd6657716e378da60b751aaeda8bef8344b3eefc4db255f2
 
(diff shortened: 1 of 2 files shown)


neheb:
Can you take me off the maintainer list? I don't really use transmission at this point. Also, @dangowrt is doing most of the work on this package.

As an aside, I have this commit locally: https://github.com/neheb/packages/commit/2c95f52db18d005adb17eb8d410871c49b8b6dcc to try and fix transmission with wolfSSL but have not been successful.

rsalvaterra:
> Can you take me off the maintainer list? I don't really use transmission at this point.

Done!

> Also, @dangowrt is doing most of the work on this package.

I also noticed from the git history, but since you are listed as the maintainer and had most of the latest commits…
However, this morning I chatted (ranted, actually, on how such an IMHO useful [upstream pull request](https://github.com/transmission/transmission/pull/1043) can be left pending for so long) with Daniel about this patch, and he said it makes sense. 🙂

> As an aside, I have this commit locally: [neheb@2c95f52](https://github.com/neheb/packages/commit/2c95f52db18d005adb17eb8d410871c49b8b6dcc) to try and fix transmission with wolfSSL but have not been successful.

Yeah, that does look more complicated.

neheb:
I meant removing me from PKG_MAINTAINER ;)

I remember another issue actually. The next version of transmission will only support CMake compilation. This is a problem because the CMake build produces much larger binaries. I think because of static linking. It's a rabbit hole that I'd rather not deal with.

PR #16231, opened 2021-07-27
openvswitch: add SSL support

stintel:
Maintainer: @yousong
Compile tested: NA
Run tested: OpenWrt master r17184-3cb22b277a on x86/64

Description:
Open vSwitch supports SSL to connect to an OpenFlow controller. This is
recommended for security. Expand the UCI ovs config section to allow
configuring SSL CA, certificate and private key.

### DIFF
--- a/net/openvswitch/README.md
+++ b/net/openvswitch/README.md
@@ -69,6 +69,19 @@ ovs ovn_northd, ovn_controller & ovs_bridge.
 Each of these supports a disabled option, which should be 
 set to 0 to launch the respective daemons.
 
+The ovs section section also supports the options below, to configure a set of
+SSL CA, certificate and private key. After adding these to Open vSwitch, you
+may specify ssl: connection methods for e.g. the OpenFlow controller. Note that
+Open vSwitch only reads these files during startup, so it needs to be restarted
+after adding or changing these options.
+
+| Name     | Type    | Required | Default | Description                       |
+|----------|---------|----------|---------|-----------------------------------|
+| disabled | boolean | no       | 0       | If set to 1, do not configure SSL |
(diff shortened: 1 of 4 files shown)


stintel:
Will bump PKG_RELEASE after merging #16228

net/openvswitch/README.md
```diff
@@ -69,6 +69,17 @@ ovs ovn_northd, ovn_controller & ovs_bridge.
 Each of these supports a disabled option, which should be 
 set to 0 to launch the respective daemons.
 
+The ovs section section also supports the options below, to configure a set of
+SSL CA, certificate and private key. After adding these to Open vSwitch, you
+may specify ssl: connection methods for e.g. the OpenFlow controller.
```
yousong, line 74: Quote manpage for ovs-vsctl

> These files are read only once, at ovs−vswitchd startup time. If their contents change, ovs−vswitchd must
be killed and restarted.

Is this still the case ?  maybe we also need to notify user about this

stintel: From reading through the documentation, this does seem the case. Do you want to notify the user on stdout when running the init script, or should I just mention it in the README.md?

I've also been looking into adding file triggers for these files, but since the init script does not make use of procd instances, this isn't possible.

yousong: > From reading through the documentation, this does seem the case. Do you want to notify the user on stdout when running the init script, or should I just mention it in the README.md?

I think mentioning it in the README.md should be okay.  It's a one-time thing.  No need to notify after first set and for each user there is no need to notify them again after the quirk is known.

yousong (approved):
LGTM

blogic (approved):

PR #16237, opened 2021-07-27
openvswitch: check correct config symbol

stintel:
Maintainer: @yousong 
Compile tested: OpenWrt master r17184-3cb22b277a on x86/64
Run tested: OpenWrt master r17184-3cb22b277a on x86/64

Description:
The config symbol is named CONFIG_OPENVSWITCH_WITH_LIBUNBOUND, so check
for that instead of the non-existent CONFIG_OPENVSWITCH_WITH_UNBOUND.

Reported by @fda77


yousong (approved):
LGTM

PR #16238, opened 2021-07-28
telegraf: Add package for telegraf

JonnyTischbein:
Signed-off-by: Jonathan Pagel <j.pagel@riotcat.org>

**Maintainer**: me 
**Compile tested**: on amd64 for aarch64 on https://git.openwrt.org/openwrt/openwrt.git commit 39f81b0bf687017b4d086255c94166e662ac177e
**Run tested:** aarch64 (b2710 Raspberry Pi 3 Model B,  OpenWrt 21.02.0-rc3, r16172-2aba3e9784), package installed, configured, starting via `/etc/init.d/`, configuring telegraf config and successfull working of input plugin prometheus and output plugin influxdb tested.

**Description**: Telegraf is a plugin-driven agent for collecting and sending metrics and events. It supports various inputs (including prometheus endpoints) and is able to send data into InfluxDB. https://www.influxdata.com/time-series-platform/telegraf/


JonnyTischbein (commented):
`files/etc/config/telegraf` dummy file, so that in the package the directory `config` is created. The default config is copy from the source code in [Makefile:67](https://github.com/openwrt/packages/pull/16238/files#diff-a57b57b3fa6826e90325587c850feaf1886e503c8f78a36b8f8edc219798e8f0R67) 

Sorry for reviewing... I wanted to comment on the files...

JonnyTischbein:
The CI failed for powerpc because `lang/golang/goland-values.mk` did not "translated" `powerpc` correctly.
According to the [go docu](https://golang.org/doc/install/source#environment) it should be `ppc64`

jefferyto:
I suggest you add `DEPENDS:=$(GO_ARCH_DEPENDS)`. I also suggest you take a look at makefiles of other Go packages in the repo.

utils/telegraf/Makefile
```diff
@@ -0,0 +1,41 @@
...
+include $(TOPDIR)/feeds/packages/lang/golang/golang-values.mk
+
+MAKE_VARS = \
+	GOARCH=$(GO_ARCH)
+
+define Package/telegraf
+	SECTION:=utils
+	CATEGORY:=Utilities
+	TITLE:=Telegraf
+	PKGARCH:=all
```
PolynomialDivision, line 26: here you wrote all. maybe that is a mistake? and also the DEPENDS should exclude targets with no go support (maybe, I'm not sure)... ;)

JonnyTischbein: Should I remove `PKGARCH` and only use `DEPENDS` or both with `$(GO_ARCH_DEPENDS)` ?

PolynomialDivision: I think I would do it like that (but not sure if it fixes your issue).

JonnyTischbein: I did fix it and the CI test was successfull. Should I resolve this ?

JonnyTischbein:
I made the default package smaller (but still 47M) and add a full package (147M). I squashed the commits.

JonnyTischbein:
With the help of the [developers](https://github.com/influxdata/telegraf/issues/9556) I managed to reduce the binary size to 34M (small) and 108M (full) by deactivating debug symbols.

PolynomialDivision:
Maybe you want to turn from draft into "ready for review"? ;)

JonnyTischbein:
I wanted to discuss the possiblity of using upx first. With this tool I was able to shrink the small version to 6M, but I found other PRs that this does not work with the squashfs. I started a discussion in the [openwrt forum](https://forum.openwrt.org/t/package-makefile-compress-binary-with-upx/102595/6).

JonnyTischbein:
I found the dependencies in the mentioned repository and already tested them successfully. Now they need to be merged into this repository before I can "undraft" my PR.
I'll wait if the author want's to be PKG maintenaner or if I fork it and create a new PR.

hnyman:
> the file consumed 10860−436 = 10424 bytes

Are you mixing units here? 
Do you mean 10424 x 1 kB blocks, in total a bit over 10 MB ???? 
(instead of 10 kB = 10424 bytes)

JonnyTischbein:
Yes your totally right! I mixed them up. 10MB ist what I meant! Sorry

jefferyto:
> Can you please a bit more precise ? I do not see that much difference to other go packages. My tests (and the CI excluded not supported architecture) were successfull. Is there anything problematic with my script ?

Most Go packages here use the default package build recipes in golang-package.mk, this package uses telegraf's own Makefile. While there is no requirement to use the default Go package build recipes, those recipes enable integration with OpenWrt build options (stripping binaries, ASLR PIE compilation, any other relevant build options added in the future).

utils/telegraf/Makefile
```diff
@@ -0,0 +1,89 @@
...
+endef
+
+define Package/telegraf-full/description
+	Telegraf is a plugin-driven agent for collecting and sending metrics and events.
+	It supports various inputs (including prometheus endpoints) and is able to send data into InfluxDB.
+	(Full build. including all plugins)
+endef
+
+define Build/Compile/telegraf
+	$(CP) ./variant-small/* $(PKG_BUILD_DIR)/
```
jefferyto, line 56: I suggest copying these files during `Build/Prepare` (after calling `Build/Prepare/Default`). (You should be able to check `$(BUILD_VARIANT)` to see which variant is being built.)

Goorzhel:
This tested well on a Linksys WRT1900ACS (`armv7l`/`mvebu`) running 21.02, but — and this is most likely because I built the package on my server and copied it to the router — the binary took on the `HEAD` of `openwrt/openwrt` for its version string:
```
root@router:~# telegraf --version
Telegraf 21.02.0 (git: HEAD b2ae4233)
```
According to the [upstream Makefile](https://github.com/influxdata/telegraf/blob/v1.19.1/Makefile#L14-L20), the version string _should_ be coming from [`build_version.txt`](https://github.com/influxdata/telegraf/blob/v1.19.1/build_version.txt). I tried `make -d --trace package/telegraf/compile` to find out more, but `--trace` doesn't apply to child `make`s.

PolynomialDivision:
Great! @jefferyto could you ack?

llaumgui:
Hi,

I have an issue with the last version:

```
root@farpoint:~# cat /etc/openwrt_release
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='21.02.1'
DISTRIB_REVISION='r16325-88151b8303'
DISTRIB_TARGET='ramips/mt7621'
DISTRIB_ARCH='mipsel_24kc'
DISTRIB_DESCRIPTION='OpenWrt 21.02.1 r16325-88151b8303'
DISTRIB_TAINTS=''
```
```
root@farpoint:~# opkg install telegraf_1.21.2-1_mipsel_24kc.ipk

Package telegraf (1.21.2-1) installed in root is up to date.
```
```
root@farpoint:~# telegraf  --test
Error relocating /usr/bin/telegraf: __nanosleep_time64: symbol not found
```

I have downloaded the good architecture ?

jefferyto:
@llaumgui This package hasn't been backported to the openwrt-21.02 branch. You can't install a snapshot package in 21.02 and expect it to work.

I suggest you try backporting this package to the openwrt-21.02 branch (or compiling the package locally using the 21.02 SDK).

llaumgui:
@jefferyto Thx. I haven't any space or memory for that. I'm waiting an official port. Perhaps the 21.03 ?

PolynomialDivision:
> Perhaps the 21.03 ?

Do you mean the 21.02.3, or the next stable branch? Since noone backported telegraf to 21.02, it will also not appear in 21.02.3, but in the next stable release. However, I have no idea when the next stable branch is planned.

JonnyTischbein:
I tried to backport it to 21.02 without problems on x86-64. I don't have a `mipsel_24kc` here to test it :/

https://github.com/openwrt/packages/pull/17737

Issue #16239, opened 2021-07-28
https-dns-proxy: uses 100% of cpu when no internet


Azq2:
Openwrt from master, TL-WR842N

Problem:
![image](https://user-images.githubusercontent.com/5139482/127311345-93d4e0b6-bb9c-4fa9-ad93-2077dd05cfee.png)

Strace:
```
...a lot of logs like this...
epoll_pwait(3, [], 64, 0, NULL, 16)     = 0
open("/etc/hosts", O_RDONLY|O_LARGEFILE) = 8
read(8, "127.0.0.1 localhost\n\n::1     loc"..., 1024) = 110
read(8, "", 1024)                       = 0
close(8)                                = 0
socket(AF_INET, SOCK_DGRAM, IPPROTO_IP) = 8
fcntl64(8, F_GETFL)                     = 0x2 (flags O_RDWR)
fcntl64(8, F_SETFL, O_RDWR|O_NONBLOCK|O_LARGEFILE) = 0
fcntl64(8, F_SETFD, FD_CLOEXEC)         = 0
connect(8, {sa_family=AF_INET, sin_port=htons(53), sin_addr=inet_addr("1.1.1.1")}, 16) = -1 ENETUNREACH (Network unreachable)
close(8)                                = 0
socket(AF_INET, SOCK_DGRAM, IPPROTO_IP) = 8
fcntl64(8, F_GETFL)                     = 0x2 (flags O_RDWR)
fcntl64(8, F_SETFL, O_RDWR|O_NONBLOCK|O_LARGEFILE) = 0
fcntl64(8, F_SETFD, FD_CLOEXEC)         = 0
connect(8, {sa_family=AF_INET, sin_port=htons(53), sin_addr=inet_addr("1.0.0.1")}, 16) = -1 ENETUNREACH (Network unreachable)
close(8)                                = 0
epoll_pwait(3, [], 64, 0, NULL, 16)     = 0
open("/etc/hosts", O_RDONLY|O_LARGEFILE) = 8
read(8, "127.0.0.1 localhost\n\n::1     loc"..., 1024) = 110
read(8, "", 1024)                       = 0
close(8)                                = 0
socket(AF_INET, SOCK_DGRAM, IPPROTO_IP) = 8
fcntl64(8, F_GETFL)                     = 0x2 (flags O_RDWR)
fcntl64(8, F_SETFL, O_RDWR|O_NONBLOCK|O_LARGEFILE) = 0
[...]

Issue #12968, opened 2020-07-29
List of "problematic" architectures


aparcar:
Hi, I'm currently building a CI which tests changed packages for multiple architectures. I'm not familiar with the lower level implementation of these architectures, but understand that a package may compile fine on one but break badly on another architecture. Could someone please help with with a short list of top 5 popular and top 5 breaking architectures to test within the CI?

neheb:
ARC is a constant source of breakage since it uses a different libc (uClibc-ng).
PowerPC occasionally breaks, either because of no FPU or because of -lssp_nonshared
ARM64 has issues with PIC.
x86_64 is used by @stintel , who constantly complains of his builds being broken.
some of the older ARM platforms break sometimes.

aparcar:
Amazing that's what I wanted to know :)!

neheb:
ARM64 fails occasionally with PKG_ASLR_PIE. Not sure if that can be turned on.

It would be great to also compile without deprecated OpenSSL APIs if possible.

aparcar:
> It would be great to also compile without deprecated OpenSSL APIs if possible.

Not sure how to do that. I'm currently working on a CI for openwrt.git, however it could be used for packages.git, too.

neheb:
It’s a Kconfig option. OPENSSL_WITH_DEPRECATED I believe.

neheb:
Not valid anymore I believe.

PR #16252, opened 2021-07-30
https-dns-proxy: update to 2021-07-29-1

### LINKED ISSUE #16222: HTTPS-DNS-PROXY doesn't let router works if WAN is disconnected
Maintainer: @stangri

Environment: Qualcomm Atheros IPQ806X, Archer C2600 v1, "21.02, Master

Description:

 I found the trouble was https-dns-proxy and not of others packages. I had to uninstall it to have my routers working well back and after that remove the package from firmware images.

 The problem shows itself when there isn't connection at the WAN port.  The router becomes inacessible from LAN.
 I tried just disabling the service but anyway it is enabled at boot by default. So the only way I had to solve this trouble was to forget this package. In the paste, because I couldn't understand if it was due to any kind of incompatibility issue with nginx service I opened tickets for all two in the past. Now I can say that the matter is https-dns-proxy.

stangri:
Maintainer: me
Compile tested: x86_64, Sophos SG-105, OpenWrt 21.02.0-rc3
Run tested: x86_64, Sophos SG-105, OpenWrt 21.02.0-rc3, start/stop/name resolution/log output

Description:
* update binary to the latest commit (2021-07-29) to fix #16222 and #16239
* add hotplug.d/iface file and update Makefile to install it
* use Cloudflare's and Google's bootstrap DNS if bootstrap DNS is missing
* minor improvements in append_bool function
* add append_counter function for verbosity setting
* add append_bootstrap function (and supporting functions) to parse/sanitize bootstrap setting
* move firewall array from 'main' instance to the first proxy instance
* delete useless 'main' instace

### DIFF
--- a/net/https-dns-proxy/Makefile
+++ b/net/https-dns-proxy/Makefile
@@ -1,14 +1,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=https-dns-proxy
-PKG_VERSION:=2021-06-03
+PKG_VERSION:=2021-07-29
 PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
-PKG_SOURCE_URL:=https://github.com/aarond10/https_dns_proxy
-PKG_SOURCE_DATE:=2021-06-03
-PKG_SOURCE_VERSION:=5651b984f770a8bcecb14aeffc224703f8f82586
-PKG_MIRROR_HASH:=b65161936269aa3117debad0fcfce157024726b78d7e7da77c226f7aa8da5b4d
+PKG_SOURCE_URL:=https://github.com/aarond10/https_dns_proxy/
+PKG_SOURCE_DATE:=2021-07-29
+PKG_SOURCE_VERSION:=750c4dbdfe3d3c65edcd57349df181ffd1f6ceeb
+PKG_MIRROR_HASH:=88f487d80f8f20b83d1f9940c30ad1d5b6a4eda6be48fbaea196c69e90c5429e
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.net>
 PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=LICENSE
@@ -38,11 +38,12 @@ define Package/https-dns-proxy/conffiles
 endef
 
 define Package/https-dns-proxy/install
-	$(INSTALL_DIR) $(1)/usr/sbin $(1)/etc/init.d ${1}/etc/config
+	$(INSTALL_DIR) $(1)/usr/sbin $(1)/etc/init.d ${1}/etc/config $(1)/etc/hotplug.d/iface
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/https_dns_proxy $(1)/usr/sbin/https-dns-proxy
 	$(INSTALL_BIN) ./files/https-dns-proxy.init $(1)/etc/init.d/https-dns-proxy
 	$(SED) "s|^\(PKG_VERSION\).*|\1='$(PKG_VERSION)-$(PKG_RELEASE)'|" $(1)/etc/init.d/https-dns-proxy
 	$(INSTALL_CONF) ./files/https-dns-proxy.config $(1)/etc/config/https-dns-proxy
+	$(INSTALL_DATA) ./files/https-dns-proxy.hotplug.iface $(1)/etc/hotplug.d/iface/90-https-dns-proxy
 endef
 
 $(eval $(call BuildPackage,https-dns-proxy))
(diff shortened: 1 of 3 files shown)


gregarios:
Why do I keep getting this as of today:

`# pkgupdate --batch
INFO:Target Turris OS: 5.2.5
line not found
ERROR:
inconsistent: Package luci-app-https-dns-proxy requires package https-dns-proxy that is not available.`

stangri:
> Why do I keep getting this as of today:
> 
> _# pkgupdate --batch INFO:Target Turris OS: 5.2.5 line not found ERROR: inconsistent: Package luci-app-https-dns-proxy requires package https-dns-proxy that is not available._

Probably because TurrisOS is far enough from OpenWrt to cause issues like [this](https://github.com/openwrt/packages/issues/16393).

TurrisOS maintainers are welcome to contribute a patch/PR.

BKPepe:
> TurrisOS is far enough from OpenWrt

I would not say that. Why was the issue caused in the first place? Oh, right. It was not a security patch, and it got into the stable branch.  If I am not mistaken, it happens with OpenWrt as well, as I was able to reproduce it and described steps that need to be done. So, no. Turris OS maintainers were not responsible for this issue. ;-) They can not ensure that every package in OpenWrt builds, so that's why Turris OS 5.2.5 was released today.

There is a reason why we should use more hardenings than defaults, but yeah, let's blame Turris for that in the first place.

But I would be curious what means "far enough from OpenWrt" in your words, but that's going to be OT.

___

@gregarios I am so sorry that you encountered this issue. I will narrow it down for you once we release Turris OS 5.2.6. Right now, ETA next week if it is OK for you. In the meantime, you can use localrepo.

champtar:
For anyone from Turris team, why not add Turris SDK to this repo CI ? Just one arch could catch a number of issues and would prevent this kind of discussion.

For updating packages in stable branches not everyone wants Debian style policy, and as long as the users of said packages are happy and the maintainer test / fix breakages (with OpenWrt) I don't see a problem.

Not sure all packages maintainers have the skills or wants to backport just security fixes Debian style when just bumping to latest version is better 99% of the time.

BKPepe:
Most of the packages which had issues with ASLR PIE were fixed or reported. Isn't in OpenWrt 21.02 highlighted and featured ASLR activated by default? 

This is not about Turris SDK or Turris issue at all. In the mentioned issue, I provided steps on how to reproduce it. 

You need to enable or set these kernel options for ARM64:

```
CONFIG_PKG_FORTIFY_SOURCE_2=y
CONFIG_PKG_ASLR_PIE_ALL=y
CONFIG_PKG_CC_STACKPROTECTOR_STRONG=y
CONFIG_PKG_RELRO_FULL=y
```

The one architecture, which I tested it to be broken, is already included in GitHub Actions, but it does not use the stronger ASLR settings to reproduce it. It was mentioned as problematic in this issue:

https://github.com/openwrt/packages/issues/12968 

Yes, there are currently two missing builds in GitHub Actions for NLS (Full language support) and ASLR PIE.

But back to the track, then, in that case, package maintainers need to count with issues that can be reproduced with OpenWrt (and this is) as they can be using different settings and fix it as it was introduced recently. No one uses a single configuration. Saying that it won't be fixed is not a good approach.

champtar:
> Most of the packages which had issues with ASLR PIE were fixed or reported. Isn't in OpenWrt 21.02 highlighted and featured ASLR activated by default?

It builds fine in master so I would say no :( (https://downloads.openwrt.org/snapshots/packages/aarch64_cortex-a53/packages/https-dns-proxy_2021-07-29-1_aarch64_cortex-a53.ipk)

I think it's opt in by default for now
`PKG_ASLR_PIE_REGULAR:=1` to opt in
`PKG_ASLR_PIE:=0` to disable ASLR PIE
```
include/hardening.mk:PKG_ASLR_PIE ?= 1
include/hardening.mk:PKG_ASLR_PIE_REGULAR ?= 0

include/hardening.mk:ifdef CONFIG_PKG_ASLR_PIE_ALL
include/hardening.mk:  ifeq ($(strip $(PKG_ASLR_PIE)),1)
include/hardening.mk:ifdef CONFIG_PKG_ASLR_PIE_REGULAR
include/hardening.mk:  ifeq ($(strip $(PKG_ASLR_PIE_REGULAR)),1)

package/libs/libpcap/Makefile:PKG_ASLR_PIE_REGULAR:=1
package/network/services/dnsmasq/Makefile:PKG_ASLR_PIE_REGULAR:=1
package/network/services/dropbear/Makefile:PKG_ASLR_PIE_REGULAR:=1
package/network/services/hostapd/Makefile:PKG_ASLR_PIE_REGULAR:=1
package/network/services/odhcpd/Makefile:PKG_ASLR_PIE_REGULAR:=1
package/network/services/ppp/Makefile:PKG_ASLR_PIE_REGULAR:=1
package/network/services/uhttpd/Makefile:PKG_ASLR_PIE_REGULAR:=1
package/system/procd/Makefile:PKG_ASLR_PIE_REGULAR:=1
package/system/rpcd/Makefile:PKG_ASLR_PIE_REGULAR:=1
package/system/ubus/Makefile:PKG_ASLR_PIE_REGULAR:=1
```

PR #16255, opened 2021-07-30
curl: enable HTTP/2 support by default

stangri:
Maintainer: @kaloz 

Description: Lack of support of HTTP/2 by default starts to hurt, for example with `https-dns-proxy` package, some DoH resolvers (like mullvad) no longer support HTTP/1 and are not usable.

Can we just enable HTTP/2 support by default (which would bring ~68Kb libnghttp) or would it require a variant? If a variant would be preferred way forward, should the default libcurl include HTTP/2 support (so the variant would be libcurl-nonghttp) or should the default stay as is (and the variant be libcurl-nghttp)?

### DIFF
--- a/net/curl/Config.in
+++ b/net/curl/Config.in
@@ -103,7 +103,7 @@ config LIBCURL_TFTP
 
 config LIBCURL_NGHTTP2
 	bool "HTTP2 protocol"
-	default n
+	default y
 
 comment "Miscellaneous"
 


BKPepe:
Copy&paste your PR description to the commit message and then it sounds good to me!

stangri:
> Copy&paste your PR description to the commit message and then it sounds good to me!

Sorry, forgot about that, done now.

Would a similar PR against 21.02 RC be acceptable?

BKPepe (approved):

aparcar:
I'd wait for the maintainer to respond

neheb:
@kaloz is not really active

kaloz:
> @kaloz is not really active

...or he has been told that maintainership no longer matters? ;)

my $0.02 is that anything adding over 50k should be configurable and disabled by default until that standard (in this case http/2) becomes the more common one

neheb:
> 
> 
> > @kaloz is not really active
> 
> ...or he has been told that maintainership no longer matters? ;)
:)
> 
> my $0.02 is that anything adding over 50k should be configurable and disabled by default until that standard (in this case http/2) becomes the more common one

Top websites including google, microsoft, twitter, etc... already do HTTP/2. That includes github.

Off-topic: it's also amusing how broken buster-backports are regarding this:

sshot-37

champtar:
> > @kaloz is not really active
> 
> ...or he has been told that maintainership no longer matters? ;)
> 
> my $0.02 is that anything adding over 50k should be configurable and disabled by default until that standard (in this case http/2) becomes the more common one

Following your reasoning we should disable IPv6 :)

Seriously we can create a libcurl variant but http/2 is already super common and required by some use cases as explained by @stangri, so it would be nice to have a ready to install package.

dangowrt:
If making variants for libcurl was so easy, we'd already have them for different SSL libraries. Unfortunately you'd have to make yet another [gnurl](https://git.taler.net/gnurl.git)-like fork for to avoid namespace collisions...

champtar:
Thanks @dangowrt, forgot the pains of lib variants :(

dangowrt:
That being said, I'm all for enabling HTTP/2 support by default for curl. Who ever needs something really tiny and doesn't need more than classic HTTP(S) should be using libuclient and uclient-fetch anyway.

stangri:
> my $0.02 is that anything adding over 50k should be configurable and disabled by default until that standard (in this case http/2) becomes the more common one

I'm totally on board with that! However http/2 while not being totally unavoidable is already widely in use and we're being restricted without its support and as @dangowrt pointed out it's not an easy feat to make variants.

With all that said, @kaloz could you please approve this so it can be merged?

dangowrt (approved):

aparcar:
@kaloz do you want to give up your maintainership and pass it one to someone else? In case, does somebody else would want to maintain it? It seems to be a rather important tool.

mhei (approved):

neheb (approved):

kaloz:
> @kaloz do you want to give up your maintainership and pass it one to someone else? In case, does somebody else would want to maintain it? It seems to be a rather important tool.

@aparcar that wording sounds weird, as neither I or others "wanted to give up" maintainership of the platforms either, we were forced to let go. GH insights clearly show how "positively" did developments like that affect our contributions compared to the 10+ years before ;)

tl;dr: if someone wants to take it over, he/she is free to.

dangowrt:
@kaloz @aparcar I think there is a misunderstanding regarding what maintainership means:
One thing is to oblige yourself to maintain a package or a target, do reviews and keep things up to date.
Another thing is to expect that doing so (or even just having your name written there) would result in some kind of special privileges as in being the only person holding the golden key to make changes to that package or target.

So as a result of switching to a democratic and way more transparent governance model (the LEDE fork in 2016, created by a vast majority of active OpenWrt contributors), we did indeed abolish that territorial aspect, and maintainers had to give up their special privileges (we switched to git. Previously there were SVN ACLs on certain folders which allowed only the maintainer to do commits). Nobody, however, was ever forced to stop maintaining a package or a platform.

Regarding curl, the last time @kaloz had touched it was in 2014, after a total of 5 commits out of which 3 were version bumps.
https://git.openwrt.org/?p=openwrt/svn-archive/packages.git;a=commit;h=5cb60609b36b0258fd7fb470a468c39b3ade3817

Ever since March 2014, @hauke, @nbd168, @nicolas-thill and @blogic have done most of the work **and apparently just committed to the tree themselves** without any review nor approval from @kaloz (please correct me if I'm wrong, I've checked in openwrt.git as well as the current packages.git, but maybe I'm missing something here?)

All this being said, I don't believe @kaloz would have had any objection if someone else would have stepped up and offered to maintain the package. It just never happened. Neither in the pre-2016 OpenWrt, nor during LEDE, nor in the remerged OpenWrt project. Looking at the recent git history suggests that @ja-pa or @hauke would be good candidates -- but both are already maintaining a huge amount of packages and might not be up for it.

kaloz:
@dangowrt I could also look through git to see how many times it turned out almost none of you did runtime testing (which was on required by the oh-so-not-democratic-enough old commit policy), but you're too trigger happy to care about facts when it comes to pushing your ideologies. Choosing politics as a profession might have suited you better seeing the amount of energy you spend on reinterpreting past and present events using half-truths. What you did with LEDE is yet another example for "revolutions devour their children".

neheb:
That escalated quickly.

bobafetthotmail:
Since @kaloz isn't an active on this package anymore since 2014, why is he still listed as maintainer?
And why people just commit regardless of maintainer, what's the point of listing someone as a maintainer?
Leaving the old maintainer for years past his de-facto step down and letting everyone commit random stuff in his absence makes no sense.

My opinion on this matter (I have already said the last time I saw similar discussions for packages in package feeds), is that a maintainer should actually do the job, and if a package maintainer goes MIA or steps down the package should be dropped from snapshot. 
Anyone that actually needs it will notice the breakage and will be motivated to become a maintainer or find a replacement.

None of this bs where people remain chained to a package for life just because nobody else steps up when asking nicely if someone else wants to take over. It does not work. We have already seen what happens, the "maintainer" eventually goes MIA anyway and the package becomes de-facto unmaintained, then random people with commit acces try to at least keep it afloat by merging things as they see fit but can't do the job properly for hundreds of packages in the same situation and quality goes down.

Also, and more to the point for this, **does this PR still need the approval of kaloz if he has not touched the package or reviewed any contribution for 5 years**?

> neither I or others "wanted to give up" maintainership of the platforms either, we were forced to let go. 

Your last commit on this package was before Barrier Breaker 14.07 release, are you sure you can blame LEDE split and new OpenWrt rules for your lack of interest/activity?
It seems to me that you just stopped maintaining this package well before those events transpired.

bobafetthotmail:
> If making variants for libcurl was so easy, we'd already have them for different SSL libraries. Unfortunately you'd have to make yet another [gnurl](https://git.taler.net/gnurl.git)-like fork for to avoid namespace collisions...

Wait, we do have variants for different SSL libraries. They are simply not compatible with each other so you can't have them all installed at the same time.

This package's two variant can be the same. Since one is the superset of the other anyway, there is no need to have two packages of the same lib installed. 

If you want http2 you remove the non-http2 version and install the other, and stuff that needs HTTP1 still works fine.

champtar:
The problem is that it's a lib, so if compile options produce different headers, what headers should other programs compile against ?

kaloz:
@bobafetthotmail TBH the whole maintainer role became empty both here and on the platforms when people started merging patches without ever involving said maintainers, wreaking havoc on quality. There is little use in blaming, the above is part of the agenda @dangowrt and others were pushing for.

Anyways, this is not the right place to discuss how cuckoos' eggs landed in OpenWrt. As I've said, the "role" is free to be taken.

dangowrt:
@kaloz Please stay with the facts. I was invited to join the project only at the moment of the LEDE fork in 2016, so whatever happened before that was certainly not due to my "political agenda" what-so-ever, simply because I didn't have commit access nor access to (or rather didn't even know about the existence of) non-public channels like `#openwrt-hackers` etc...
Anyway, if all this is really the result of my influence, then yes, maybe you are right and I should consider a career in politics 🙃

kaloz:
> @kaloz Please stay with the facts. I was invited to join the project only at the moment of the LEDE fork in 2016, so whatever happened before that was certainly not due to my "political agenda" what-so-ever, simply because I didn't have commit access nor access to (or rather didn't even know about the existence of) non-public channels like `#openwrt-hackers` etc...
> 
> Anyway, if all this is really the result of my influence, then yes, maybe you are right and I should consider a career in politics 🙃 

@dangowrt It would be useless to ask you to stay in your lane, given you were part of the charlie foxtrot going on ever since, even now repainting things you state you have no idea about - it dorsn't matter if you do that based on hearsay or self-interest. Anyways, as I've stated earlier, tickets are ****not the right medium to discuss this****, and even if they would be, you've run out this week's energy credit to deal with people of political extremes. That's the end of the matter.

@aparcar : Paul, as above, please feel free to give the role to someone who takes the plunge.

bobafetthotmail:
> The problem is that it's a lib, so if compile options produce different headers, what headers should other programs compile against ?

Since the change is just adding HTTP2 support I thought the header difference is limited to that, so programs compiled against this library without HTTP2 should work with the "full fat" version of this library too.

Is my theory unrealistic?

bobafetthotmail:
> @bobafetthotmail TBH the whole maintainer role became empty both here and on the platforms when people started merging patches without ever involving said maintainers, wreaking havoc on quality.

You already said that multiple times, it's still not an answer to my question.

>Anyways, this is not the right place to discuss how cuckoos' eggs landed in OpenWrt. 

Yeah that's what I've been saying, this package left your hands well before that

aparcar:
> @aparcar : Paul, as above, please feel free to give the role to someone who takes the plunge.

All right, anyone up for the mission? If so please create a PR to add yourself as a maintainer.

In the meantime I'd merge this PR to make OpenWrt a bit more future proof?

aparcar (approved):

bobafetthotmail:
> All right, anyone up for the mission? If so please create a PR to add yourself as a maintainer.

What if nobody shows up (since most people will never read this PR thread thus never figure out a new maintainer is needed), we keep kaloz name and email there forever against his own will?

aparcar:
@stangri since you recently got commit access and heavily use this curl for your packages, would you mind to step in?

stangri:
> @stangri since you recently got commit access and heavily use this curl for your packages, would you mind to step in?

I wish I could, but I'm in no way qualified enough to do reviews on curl. :(

aparcar:
> > @stangri since you recently got commit access and heavily use this curl for your packages, would you mind to step in?
> 
> I wish I could, but I'm in no way qualified enough to do reviews on curl. :(

It doesn't mean you actually code around with curl but to check for updates and runtime test them.

stangri:
> > > @stangri since you recently got commit access and heavily use this curl for your packages, would you mind to step in?
> > 
> > 
> > I wish I could, but I'm in no way qualified enough to do reviews on curl. :(
> 
> It doesn't mean you actually code around with curl but to check for updates and runtime test them.

Unless someone else steps up in the next few days, let me have a crack at updating to 7.78, I'll send a PR with all the updates then.

PR #16274, opened 2021-08-02
simple-adblock: update to 1.8.7-6

stangri:
Maintainer: me
Compile tested: x86_64, Sophos SG-105, 21.02.0-rc4
Run tested: x86_64, Sophos SG-105, 21.02.0-rc4, start, lists download, check blocking, check status

Description:
* supports newer shellcheck
* restore EXTRA_COMMANDS compatibility with 19.07
* move status display from various functions to status_service
* bugfix: status_service line break after output
* minor arythmetic fix in status_service

### DIFF
--- a/net/simple-adblock/Makefile
+++ b/net/simple-adblock/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=simple-adblock
 PKG_VERSION:=1.8.7
-PKG_RELEASE:=3
+PKG_RELEASE:=6
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.net>
 PKG_LICENSE:=GPL-3.0-or-later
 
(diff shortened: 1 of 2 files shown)


BKPepe:
> restore EXTRA_COMMANDS compatibility with 19.07

Why? You pushed this to the `master` branch. This is not necessary at all.

stangri:
> > restore EXTRA_COMMANDS compatibility with 19.07
> 
> Why? You pushed this to the `master` branch. This is not necessary at all.

I want to maintain a single code base across branches.

BKPepe:
I would say that this is not possible as you are bloating it with some compatibility layers between releases, and this is not good. To be honest, this does not belong to OpenWrt master branch. Since it is merged, you don't count on the low-memory devices running OpenWrt. Every byte counts there. Also, you are still playing magic with **PKG_RELEASE**.

I don't see necessary why this should be included in `OpenWrt 19.07`. Your reasoning about maintaining a single code base across branches, I don't take. Sorry.

PR #16282, opened 2021-08-03
iperf3: add uci server handling

feckert:
Maintainer: @pprindeville
Run tested: x86_64, APU3, latest openwrt master (iperf3 with uci client/server)

Server iperf3 uci config
```
config server lan
        option enabled '1'
        option port '5202'
        option iface 'lan'
        option redirect '1'
```

Client iperf3 uci config:
```
config client lan
        option server '192.168.0.53'
        option port '5202'
        option proto 'ipv4'
        option redirect '1'
```

Execute the command `/etc/init.d/perf3 client lan` to start a scan against the server und the client device

Description:

As discussed an suggest in the [P/R](https://github.com/openwrt/luci/pull/4651) I have extended the iperf3.
With the following improvements.

- Add uci hanndling
- Redirect output to `ubus listen luci.iperf3.notify.data` As for the path, I'm not sure whether we should use luci or something else. Suggestions are welcome ;-)

This commits adds the possibility that iperf3 can be configured via
the uci. The functionality has also been extended so that the output of
iperf3 (server/client) sends a notification event via the ubus to the  `luci.iperf3.notify.data` endpoint.

@stangri @wjowsa Could you also look at my proposal? I think this is a generally valid proposal and would simplify the LuCI. You just have to figure out how to subscribe to the ubus notification in the LuCI and show them.

Then, from my point of view, we would no longer need this [patch](https://patchwork.ozlabs.org/project/openwrt/patch/20201110080209.6368-1-wojciech.jowsa@gmail.com/) and could move on. As discussed with @jow- a while a go this has some security issues.

### DIFF
--- a/net/iperf3/files/etc/config/iperf3
+++ b/net/iperf3/files/etc/config/iperf3
@@ -1,0 +1,16 @@
+
+#config server wan
+#	option enabled '1'
+#	option port '5202'
+#	option iface 'wan'
+#	option redirect '1'
+#	option interval '1'
+
+config client internet
+	option server 'speedtest.wtnet.de'
+	option port '5201'
+	option duration '10'
+	option proto 'any'
+	option reverse '0'
+	option udp '1'
+	option redirect '1'
--- a/net/iperf3/files/etc/init.d/iperf3
+++ b/net/iperf3/files/etc/init.d/iperf3
@@ -1,0 +1,140 @@
+#!/bin/sh /etc/rc.common
+
+USE_PROCD=1
+
+START=50
+PROG_DEFAULT="/usr/bin/iperf3"
+PROG_REDIRECT="/usr/libexec/iperf3-ubus-redirect"
+
+DIR="/var/run/iperf3"
+PIDFILE="${DIR}/pidfile"
+LOGFILE="${DIR}/logfile"
+
+. $IPKG_INSTROOT/lib/functions/network.sh
+
+start_server() {
+	local cfg="$1"
+
+	local enabled port iface interval proto ip redirect prog
+
+	config_get_bool enabled "$cfg" 'enabled' 1
+	[ $enabled -gt 0 ] || return
+
+	rm -f "${LOGFILE}-${cfg}.log"
+
+	config_get port "$cfg" port "5201"
+	if [ -z "$port" ]; then
+		echo "No server port configured" >&2
+		exit 1
+	fi
+
+	config_get proto "$cfg" proto 'any'
+	config_get iface "$cfg" iface ''
+
+	if [ "$proto" = "ipv4" ] && [ "$iface" != '' ]; then
+		network_get_ipaddr ip "$iface"
+	elif [ "$proto" = "ipv6" ] && [ "$iface" != '' ]; then
+		network_get_ipaddr6 ip "$iface"
+	fi
+
+	config_get interval "$cfg" interval '1'
+	config_get_bool redirect "$cfg" 'redirect' 0
+
+	if [ "$redirect" -eq 1 ]; then
+		prog="$PROG_REDIRECT"
+	else
+		prog="$PROG_DEFAULT"
+	fi
+
+	procd_open_instance "$cfg"
+	procd_set_param command "$prog" --server
+	[ -n "$ip" ] && procd_append_param command --bind "$ip"
+	procd_append_param command --port "$port"
+	procd_append_param command --interval "$interval"
+	[ "$redirect" -eq 0 ] && procd_append_param command --logfile "${LOGFILE}-${cfg}.log"
+	procd_append_param command --forceflush
+	procd_set_param pidfile "${PIDFILE}-${cfg}.pid"
+	procd_set_param respawn
+	procd_set_param stderr 1
+	procd_close_instance
+}
+
+service_triggers() {
+	procd_add_reload_trigger "iperf3"
+}
+
+start_service() {
+	mkdir -p "$DIR"
+	config_load iperf3
+	config_foreach start_server server
+}
+
+start_client() {
+	local cfg="$1"
+
+	local server port duration reverse proto udp
+	local version redirect json prog
+
+	config_load iperf3
+	config_get server "$cfg" server
+	config_get port "$cfg" port '5201'
+	config_get duration "$cfg" duration '10'
+	config_get_bool reverse "$cfg" reverse '0'
+	config_get_bool udp "$cfg" udp '0'
+
+	config_get proto "$cfg" proto "any"
+	if [ "$proto" = "ipv6" ]; then
+		version="--version6"
+	elif [ "$proto" = "ipv4" ]; then
+		version="--version4"
+	fi
+
+	if [ -z "$server" ]; then
+		echo "No server configured" >&2
+		exit 1
+	fi
+
+	rm -f "${LOGFILE}-${cfg}.log"
+
+	config_get_bool redirect "$cfg" 'redirect' 0
+	config_get_bool json "$cfg" 'json' 0
+
+	if [ "$redirect" -eq 1 ]; then
+		prog="$PROG_REDIRECT"
+	else
+		prog="$PROG_DEFAULT"
+	fi
+
+	procd_open_instance "$cfg"
+	procd_set_param command "$prog"
+	procd_append_param command --client "$server"
+	procd_append_param command --port "$port"
+	procd_append_param command --time "$duration"
+	[ "$proto" != "any" ] && procd_append_param command "$version"
+	[ "$redirect" -eq 0 ] && {
+		procd_append_param command --logfile "${LOGFILE}-${cfg}.log"
+		[ "$json" -eq 1 ] && procd_append_param command --json
+	}
+	[ "$reverse" -eq 1 ] && procd_append_param command --reverse
+	[ "$udp" -eq 1 ] && procd_append_param command --udp
+	procd_append_param command --forceflush
+	procd_close_instance
+}
+
+extra_command "client" "<cfg> Start a client"
+
+client() {
+	local cfg="$1"
+
+	local section_type
+
+	section_type="$(uci_get iperf3 "$cfg")"
+
+	if [ "${section_type}" != "client" ]; then
+		echo "Client config \"${cfg}\" not found."
+		exit 1
+	fi
+
+	mkdir -p "$DIR"
+	rc_procd start_client "$cfg"
+}
--- a/net/iperf3/Makefile
+++ b/net/iperf3/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=iperf
 PKG_VERSION:=3.11
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://downloads.es.net/pub/iperf
@@ -63,6 +63,12 @@ define Package/iperf3/description
  characteristics.
 endef
 
+define Package/iperf3/conffiles
+/etc/config/iperf3
+endef
+
+Package/iperf3-ssl/conffiles = $(Package/iperf3/conffiles)
+
 # autoreconf fails if the README file isn't present
 define Build/Prepare
 	$(call Build/Prepare/Default)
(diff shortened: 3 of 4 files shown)


wjowsa:
@fecker your proposal looks good and I agree that the frontend might be simpler with the one exception. 
This approach eliminates using of uhttpd ubus API and server-side events. It means that Luci would have to somehow poll for the output of the iperf3 that might worse UX.  

If possible, please send comments to the [patch](https://patchwork.ozlabs.org/project/openwrt/patch/20201110080209.6368-1-wojciech.jowsa@gmail.com/) with the description of security issues so I could address them.

net/iperf3/files/etc/init.d/iperf3
```diff
@@ -0,0 +1,121 @@
...
+	[ "$reverse" -eq 1 ] && procd_append_param command --reverse
+	[ "$udp" -eq 1 ] && procd_append_param command --udp
+	procd_append_param command --forceflush
+	procd_close_instance
+}
+
+extra_command "client" "<cfg> Start a client"
+
+client() {
+	local cfg="$1"
```
stangri, line 117: Would a check to make sure something was passed from CLI make sense?

Not sure if you'd want to first parse the config to even verify there's a client setting and provide information/help on that first.

net/iperf3/files/etc/init.d/iperf3
```diff
@@ -0,0 +1,121 @@
...
+
+	local enabled port iface interval proto ip redirect prog
+
+	config_get_bool enabled "$cfg" 'enabled' 1
+	[ $enabled -gt 0 ] || return
+
+	rm -f "${LOGFILE}-${cfg}.log"
+
+	config_get port "$cfg" port '5201'
```
stangri, line 25: If there's a chance people would be defining multiple servers, would it maybe be better to implement lazy port specification, where you set the default value in [start_service](https://github.com/openwrt/packages/blob/master/net/https-dns-proxy/files/https-dns-proxy.init#L155) and then [use it](https://github.com/openwrt/packages/blob/master/net/https-dns-proxy/files/https-dns-proxy.init#L91) and [increment it](https://github.com/openwrt/packages/blob/master/net/https-dns-proxy/files/https-dns-proxy.init#L149) in the instance start.

stangri:
@feckert I've had a few comments, maybe @wjowsa would want to implement these after merging your proposal.

net/iperf3/files/etc/init.d/iperf3
feckert: good advice. I have added it.
https://github.com/openwrt/packages/pull/16282/commits/c42bade1bbaef53700d540ce5127551d4473cbdf

Will stash the commits if it works

net/iperf3/files/etc/init.d/iperf3
feckert: good advice. I have added it.
https://github.com/openwrt/packages/pull/16282/commits/cda55baf05f5fa575b1ff1d18fa50e13a6e09834

Will stash the commits if it works

feckert:
> @feckert your proposal looks good and I agree that the frontend might be simpler with the one exception.
> This approach eliminates using of uhttpd ubus API and server-side events.

I have not taken a closer look. But I can also subscribe to the event `luci.iperf3.notify.data`, can't I? I do the same as you in the LuCI. Only I do the whole thing in the iperf3 init script.

>It means that Luci would have to somehow poll for the output of the iperf3 that might worse UX.

That's what we do in the Luci on different places. We add a [poll.add](https://github.com/openwrt/luci/search?q=poll.add) function.

If that doesn't work with the ubus event, then we could always poll the log output from iperf3 and display that.
That would be even easier, of course. There's not that much data, if that's your concern.

> If possible, please send comments to the [patch](https://patchwork.ozlabs.org/project/openwrt/patch/20201110080209.6368-1-wojciech.jowsa@gmail.com/) with the description of security issues so I could address them.

Unfortunately, I don't remember. It's been a while since I chatted with him about this topic on IRC. But I could remember that he said that the patch cannot be accepted in this way. I think you have to ask him. Sorry

wjowsa:
>>@feckert your proposal looks good and I agree that the frontend might be simpler with the one exception.
This approach eliminates using of uhttpd ubus API and server-side events.

>I have not taken a closer look. But I can also subscribe to the event luci.iperf3.notify.data, can't I? I do the same as you in the LuCI. Only I do the whole thing in the iperf3 init script.

It looks the same but it is not the same. You listen for events but I subscribe to an object. More, subscribing to an ubus object is supported by uhttpd so data is [easier](https://git.openwrt.org/?p=project/uhttpd.git;a=commit;h=11723570af9cb7bd87842e79c85ee99530be9902) and in a [more elegant way](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) accessible from the browser application.  

>>It means that Luci would have to somehow poll for the output of the iperf3 that might worse UX.

>That's what we do in the Luci on different places. We add a poll.add function.

Yes but I think we could start using a new uhttpd ubus API and utilize server-side events instead of polling. This could be a starting point for the unification of applications that read data from the backend - you just create an ubus object on the backend side, subscribe to the object on the frontend side.  

>If that doesn't work with the ubus event, then we could always poll the log output from iperf3 and display that.
That would be even easier, of course. There's not that much data, if that's your concern.

Yes, that is what I would like to avoid :)

feckert:
@wjowsa Thank you for your explanation, now I understand.
We have two possibilities

1. publish/subscribe: 
publisher lua reference implementation:
https://git.openwrt.org/?p=project/ubus.git;a=blob_plain;f=lua/publisher.lua;hb=refs/heads/master
subcriber lua reference implementation:
https://git.openwrt.org/?p=project/ubus.git;a=blob_plain;f=lua/subscriber.lua;hb=refs/heads/master

2. listen/send
send event via ubus:
`ubus send luci.iperf3.status '{"message":"<message>"}'`
listen on events via ubus:
`ubus listen luci.iperf3.status`

As far as I can see, the ubus API only offers the publish/subcribe model currently.
Events cannot be output yet over the API, can they?
But would be quite useful here in my opinion

For security reasons, I found the following article on the internet.
https://exceptionshub.com/http-authorization-header-in-eventsource-server-sent-events.html
As far as I can remember, this was also the concern of @jow- .

wjowsa:
@feckert 
Yes, ubus API offers the publish/subcribe model currently. Even if events were part of the API then I think we will end up with the same issue with passing the authentication token / ubus session id. 

I know that passing a token through the URL might not look secure enough. What I would like to know is how it is less secure from the way the sid is passed currently from Lucy to the uhttpd.

feckert:
> I know that passing a token through the URL might not look secure enough. What I would like to know is how it is less secure from the way the sid is passed currently from Lucy to the uhttpd.

Your suggested [implementation](https://patchwork.ozlabs.org/project/openwrt/patch/20201110080209.6368-1-wojciech.jowsa@gmail.com/) uses the ubus_rpcd session id, which does not change during a session. The timeout is also restarted when a new request is made. The suggestion from the [link](https://exceptionshub.com/http-authorization-header-in-eventsource-server-sent-events.html) would generate a token that is valid once per request.

In summary, I would say that the difference is that there is no timeout for the token, but the token is only valid once. This is not the case for the ubus session id. If you have connected once, the session id remains valid until you have logged out or the timeout strikes.

I am not an expert in this field, I can also get this wrong. But these are the things I have found on the subject.

stangri:
@feckert @wjowsa I know the conversation digressed towards WebUI use of the package, this PR does simplify the CLI use quite a bit tho, you should consider merging it.

feckert:
I have add my latest changes to this pullrequest.
@stangri could you please review?

net/iperf3/files/etc/config/iperf3
```diff
@@ -0,0 +1,16 @@
+
+config server wan
+	option enabled '1'
```
stangri, line 3: I'm not sure I fully understand the users' expectations from this package, but I'd probably not expect the server to start on wan by default.

net/iperf3/files/etc/config/iperf3
```diff
@@ -0,0 +1,16 @@
...
+	option port '5202'
+	option iface 'wan'
+	option redirect '1'
+	option interval '1'
+
```
stangri, line 8: Maybe in a future version you could add an option for logging, given that it's already possible to capture the output thru redirect.

net/iperf3/files/etc/config/iperf3
```diff
@@ -0,0 +1,16 @@
...
+#config client internet
```
stangri, line 9: Do you think it would be a good idea to have a config for one of the public iperf3 servers in the disabled state rather than commented out config?

net/iperf3/files/etc/config/iperf3
feckert: good point will fix with the next push and comment out this section

net/iperf3/files/etc/config/iperf3
feckert: Ok I will do that in the next iteration

net/iperf3/files/etc/config/iperf3
feckert: Sounds good, I will look for a server on the internet and use this as client config.

feckert:
@stangri Would you be so kind again and do a review or an integration test?

stangri:
Sorry @feckert I will not be able to provide a meaningful response until late next week, if you want to merge it before then, go ahead, even my previous comments weren't about substantial issues.

net/iperf3/files/etc/init.d/iperf3
stangri: If you could help me understand why you want back to static port number from the lazy/dynamic port number, I'd be grateful.

stangri:
GitHub WebUI is confusing, I'm still seeing the old code when browsing Files Changed of the PR and the new code when  clicking on the hash (https://github.com/openwrt/packages/commit/4f8155e5eaed5f334d5dcfc4578284f85dd265e8). 

Great work on that last commit!

net/iperf3/files/etc/init.d/iperf3
feckert: I have now decided against the lazy port binding. When I have a server, I want to specify exactly which port it should listen to.

feckert:
@nbd168  Do you have any comments on my change? If not, I would like to merge it.

feckert:
No this was not merged.

raenye:
Hmm, so why close?

feckert:
I did not pursue this any further, as no reviewer was found. I then moved it to my own package feed, as it doesn't necessarily have to be integrated in the iperf3 package. If it is useful, we can integrate it, but I need more feedback.

raenye:
It is useful for me (instead of having a line in `/etc/rc.local`).

I cherry-picked this (and changed v3.11 to v3.15).
Server mode works well fine (including manual reload when I change the config file);
I'm frankly not sure what's the point of client mode, but it doesn't work for me.
```
root@OpenWrt:~# cat /etc/config/iperf3
config server lan
        option enabled '1'
        option iface 'lan'
        #option interval '1'
        #option port '5201'
        option redirect '1'

config client door
        option server 'door.lan'
        option port '5201'
        option duration '10'
        option proto 'any'
        option reverse '0'
        option udp '1'
        option redirect '1'
root@OpenWrt:~# service iperf3 client door
root@OpenWrt:~# logread -l 1
Wed Nov 29 08:32:56 2023 daemon.err iperf3-ubus-redirect[8121]: iperf3: the client has terminated
root@OpenWrt:~# iperf3 -u -t1 -c door.lan
Connecting to host door.lan, port 5201
[  5] local 192.168.1.2 port 55522 connected to 192.168.1.25 port 5201
[ ID] Interval           Transfer     Bitrate         Total Datagrams
[  5]   0.00-1.00   sec   129 KBytes  1.05 Mbits/sec  91
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[  5]   0.00-1.00   sec   129 KBytes  1.05 Mbits/sec  0.000 ms  0/91 (0%)  sender
[  5]   0.00-1.00   sec   129 KBytes  1.05 Mbits/sec  0.125 ms  0/91 (0%)  receiver

iperf Done.

raenye:
So client mode works fine too when using `redirect '0'`. The results of the most recent run are in `/var/run/iperf3/logfile-${cfg}.log`.

It still seems to me that client mode would be better as a LuCI application rather than a service.

- For server mode. I think the examples in the default config file should be changed. Options that are the default (e.g. port) should be pre-commented out. 
- For client mode (if it stays) I guess it makes sense to support all three directions: outbound (no flags), inbound (`--reverse`) and both (`--bidir`) as well as parallel streams.

raenye:
@feckert 
I came to a conclusion that the `init.d` script should be moved to a separate package (cf. `rsync` and `rsyncd`), which is probably what you meant above. 

Would it be OK with you if I submitted this modification as a new PR (without client mode).
Should I write you as Authored-by, Co-authored-by, or Signed-off-by?

Also, what problem does the ubus redirect script solve? I'm not sure why it is needed.

Thanks!

Issue #15723, opened 2021-05-31
stubby: trigger option and multiple WAN interfaces


jamesmacwhite:
Maintainer: @jonathanunderwood 
Environment: OpenWrt 19.07.7 mvebu Linksys WRT3200ACM

Description:

For environments that have multiple WAN interfaces, it might be advantageous to have the trigger option support multiple WAN interfaces, specified as each interface name such as `wan wanb wanc`

https://forum.openwrt.org/t/stubby-option-trigger-help/56694

The original poster in the OpenWrt forum thread highlights that the trigger option doesn't seem to work with multiple WANs however the following modification to the init script, would allow this:

https://github.com/openwrt/packages/blob/master/net/stubby/files/stubby.init#L271

Replaces with a for loop for each wan interface referenced:

```sh
for trigger_item in $trigger
do
    procd_add_interface_trigger "interface.*.up" "$trigger_item" "$stubby_init" start
done
```

There are situations that with multiple WAN interfaces and having something like mwan3 enabled, stubby may not be reloaded when a WAN interface goes down and hence this will cause DNS lookups to time out.

jamesmacwhite:
All credit to @deaddy64 for the suggested code change, as per the original thread, I thought it was worth highlighting here for potential inclusion into the main package.

jonathanunderwood:
Thanks @jamesmacwhite . I am no longer maintaining this package, but I am sure that if you prepare a PR, maintainers will merge it.

jamesmacwhite:
@jonathanunderwood Thanks for letting me know. I'll close this and open a PR.

PR #16285, opened 2021-08-03
stubby: Add multi WAN support for procd trigger

jamesmacwhite:
Maintainer: No one, previous maintainer no longer wishes to be a maintainer see https://github.com/openwrt/packages/issues/15723#issuecomment-860279906

Compile tested: N/A
Run tested: Linksys WRT3200ACM OpenWrt 19.07.7

Description:

This adds multi WAN trigger support for stubby, in the stubby config you can now supply multiple WAN interfaces separated with a space, currently you cannot do this. For multi WAN environments you would most likely want to have Stubby be able to have triggers on multiple WAN interfaces to not break DNS resolution if a WAN interface goes down.

Original credit/author: https://forum.openwrt.org/t/stubby-option-trigger-help/56694

As this is minor change to the init script, can this also be cherry picked across 21.02 and 19.07?

### DIFF
--- a/net/stubby/Makefile
+++ b/net/stubby/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=stubby
 PKG_VERSION:=0.4.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/getdnsapi/$(PKG_NAME)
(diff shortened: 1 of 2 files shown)


dibdot:
Hi, thanks for the contribution. A few remarks:

- This seems to be copied 1:1 from ancient adblock init script in 18.06 branch (see https://github.com/openwrt/packages/blob/openwrt-18.06/net/adblock/files/adblock.init for reference). The cases 'none' or 'timed' are not handled by stubby - therefore please just remove the outer if clause.
- Please bump the PKG_RELEASE in the Makefile as well

Thanks!

jamesmacwhite:
Hi @dibdot, thanks for your feedback and guidance! I have updated the PR as requested with those changes.

Please me know if this is OK and ready for merging.

dibdot:
LGTM - many thanks!

PR #16287, opened 2021-08-03
[19.07] stubby: Add multi WAN support for procd trigger

jamesmacwhite:
Maintainer: No one, previous maintainer no longer wishes to be a maintainer see https://github.com/openwrt/packages/issues/15723#issuecomment-860279906

Compile tested: N/A
Run tested: Linksys WRT3200ACM OpenWrt 19.07.7

Description:

This adds multi WAN trigger support for stubby, in the stubby config you can now supply multiple WAN interfaces separated with a space, currently you cannot do this. For multi WAN environments you would most likely want to have Stubby be able to have triggers on multiple WAN interfaces to not break DNS resolution if a WAN interface goes down.

Original credit/author: https://forum.openwrt.org/t/stubby-option-trigger-help/56694

### DIFF
--- a/net/stubby/files/stubby.init
+++ b/net/stubby/files/stubby.init
@@ -268,7 +268,10 @@ service_triggers()
 
     if [ "$trigger" != "none" ] && [ "$trigger" != "timed" ]; then
         PROCD_RELOAD_DELAY=$((${delay:-2} * 1000))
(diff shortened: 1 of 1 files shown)


dibdot:
This can be closed as we could cherry pick proposed changes to other branches.

jamesmacwhite:
Thank you I will close this and note to cherry-pick for other branches.

PR #16288, opened 2021-08-03
[21.02] libxslt: fix compilation because of wrong libxml2 check in configure script

Linaro1985:
Compilation failed after libxml2 update. More details: https://github.com/GNOME/libxslt/commit/90c34c8bb90e095a8a8fe8b2ce368bd9ff1837cc

### DIFF
--- a/libs/libxslt/patches/010-fix-xml2-config-check.patch
+++ b/libs/libxslt/patches/010-fix-xml2-config-check.patch
@@ -1,0 +1,11 @@
+--- a/configure.ac
++++ b/configure.ac
+@@ -546,7 +546,7 @@ dnl make sure xml2-config is executable,
+ dnl test version and init our variables
+ dnl
+ 
+-if test "x$LIBXML_LIBS" = "x" && ${XML_CONFIG} --libs print > /dev/null 2>&1
++if test "x$LIBXML_LIBS" = "x" && ${XML_CONFIG} --libs > /dev/null 2>&1
+ then
+     AC_MSG_CHECKING(for libxml libraries >= $LIBXML_REQUIRED_VERSION)
+     XMLVERS=`$XML_CONFIG --version`


neheb:
Why is this targeting 21.02?

Linaro1985:
Because I'm currently working on the 21.02 branch (v21.02-rc4). I can't say anything about the master branch.

neheb:
That's no go. Generally speaking, commits like these need to go in master first and then backported.

ping @micmac1

micmac1:
Hi al,

Of course I agree with @neheb on this.

I'd be interested to see how it fails. Maybe tonight or tomorrow I can find some time to check.

Maintainer is @jslachta :-)

Kind regards,
Seb

micmac1:
The code in libxslt's configure script is this:

```
if test "x$LIBXML_LIBS" = "x" && ${XML_CONFIG} --libs print > /dev/null 2>&1
then
  ...
fi
```

I checked and we're not reaching the contents of this "if" block during a target package build, because $LIBXML_LIBS is already set after

```
if test "x$LIBXML_CONFIG_PREFIX" = "x" -a "x$LIBXML_LIBS" = "x"; then
    if test "$build_static_libs" = "no"; then
        PKG_CHECK_MODULES([LIBXML], [libxml-2.0 >= $LIBXML_REQUIRED_VERSION], [
            WITH_MODULES="`$PKG_CONFIG --variable=modules libxml-2.0`"
        ], [ ])
    else
        PKG_CHECK_MODULES_STATIC([LIBXML], [libxml-2.0 >= $LIBXML_REQUIRED_VERSION],
        [ ], [ ])
    fi
fi
```

But for the libxslt host build we add "--with-libxml-prefix=$(STAGING_DIR_HOSTPKG)" to the configure args, so we do not end up in the "if" block above and then we reach the first snippet, which then bombs because xml2-config doesn't like the extra "print".

```
checking for x86_64-pc-linux-gnu-xml2-config... /home/sk/tmp/openwrt/staging_dir/hostpkg/bin/x86_64-pc-linux-gnu-xml2-config
configure: error: Could not find libxml2 anywhere, check ftp://xmlsoft.org/.
```

The thing is, I see no package depending on libxslt/host :-) So I guess you have an extra package that does and ran into this issue.

Anyway, this fix should go in master first. Maybe you can just raise a separate pull request for master and do the exact same thing there. The fix itself is obviously fine.

Kind regards,
Seb

Linaro1985:
@micmac1 , @neheb thanks! I created pull request for master branch https://github.com/openwrt/packages/pull/16293
Commit for 21.02 was cherry-picked from master branch.

PR #16291, opened 2021-08-04
curl: update to 7.78.0

stangri:
Maintainer: @kaloz -> @stangri 
Compile tested: x86_64, Sophos SG-105, OpenWrt 21.02.0-rc4
Run tested: x86_64, Sophos SG-105, OpenWrt 21.02.0-rc4, install, use with `https-dns-proxy`, download via HTTPS.

Description:
* [update to 7.78.0](https://daniel.haxx.se/blog/2021/07/21/curl-7-78-0-five-in-one/)
* change maintainer
* remove obsolete mirror (curl.mirror.anstey.ca)
* update main curl URLs

### DIFF
--- a/net/curl/Makefile
+++ b/net/curl/Makefile
@@ -8,15 +8,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=curl
-PKG_VERSION:=7.77.0
+PKG_VERSION:=7.78.0
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://dl.uxnr.de/mirror/curl/ \
-	https://curl.mirror.anstey.ca/ \
 	https://curl.askapache.com/download/ \
-	https://curl.haxx.se/download/
-PKG_HASH:=0f64582c54282f31c0de9f0a1a596b182776bd4df9a4c4a2a41bbeb54f62594b
+	https://curl.se/download/
+PKG_HASH:=be42766d5664a739c3974ee3dfbbcbe978a4ccb1fe628bb1d9b59ac79e445fb5
 
 PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=COPYING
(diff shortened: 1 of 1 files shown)


champtar (approved):
LGTM

net/curl/Makefile
```diff
@@ -115,7 +114,6 @@ CONFIGURE_ARGS += \
 	--enable-static \
 	--disable-manual \
 	--without-nss \
-	--without-libmetalink \
```
aparcar, line 118: What about this line?

stangri: [Metalink support was removed from last version of curl](https://daniel.haxx.se/blog/2021/06/07/bye-bye-metalink-in-curl/) -- if left in the Makefile it resulted in the error.

BKPepe: Well, it should be described in the commit message, why you are removing it. Otherwise you are doing more changes than you should.

stangri: @BKPepe Looking at https://github.com/openwrt/packages/pull/16290 which you just merged (which lacked any sort of commit message at all), I'll add the link to the curl changelog to the PR description.

BKPepe: Oh, you mentioned what I merged. May I know how you would do that? Because there are 3 options.

1. Include link for release notes for relevant release in the commit message.
Sometimes the link can be changed in the future, so it might not be helpful.

2. Include release notes in the commit message

3. You dont need to include anything.

All these options are valid since the commit message itself is updated to a specific version because there were two changes - **bump version and change commit hash**. 

In your case, I see that you removed one configure option, which you didn't mention anywhere, and that's the point, which I and @aparcar mentioned. You are not doing it just for me. You are doing it for others. Why should anyone have a crystal ball to know in a sec why you removed it? My point is that this change should be mentioned somewhere, ideally in the commit message itself. Why? There is a simple reason for it. If you will use ``git log``/``git show``/``git blame``, etc. to take a look at past changes, then it is good to know why such changes were done in the first place, so you don't need search the Internet and spend time to realize: "Oh, this is why he did it, well it could be mentioned in the commit message."

PR #16299, opened 2021-08-07
dawn: update to 2021-08-05

PolynomialDivision:
* c70773a - datastorage: use signal strength as a metric
* 14e0f83 - Don't display debugging output with DAWN_NO_OUTPUT
* 97e5de1 - uci: add neighbor list priority options
* 2b1a53c - dawn_uci: set default values
* 6eb747b - Use separate configs for 802.11g & 802.11a bands
* 1e34357 - Verify compatibility before parsing config message
* a7a8309 - List all neighbors with same score when kicking
* 3ba0fa4 - Change beacon request fields to appropriate values
* 009aab9 - Change mode config parameter from int to string


net/dawn/files/dawn.config
```diff
@@ -25,38 +25,49 @@ config times
...
+
+config metric 802_11a
+    option rssi_weight          '2'
+    option rssi_center          '-70'
+    option initial_score        '125'
+
+config metric 802_11g
+    option rssi_weight          '2'
+    option rssi_center          '-70'
+    option initial_score        '100'
```
neheb, line 73: no newline? What editor is this?

cotequeiroz:
I'll take a look at this when able.  There's also `luci-app-dawn` that needs to be adjusted.  BTW, if I can learn lua, I have a long-term 👴  plan of adding a graphic simulator--like I've done in a spreadsheet--so one can test the parameters before committing them.  It'll need a ubus procedure so dawn can calculate a score, then we can use that to create a graph.  I'm short on time at the moment, so don't hold your breath for it.

cotequeiroz:
Well, in that case `sed s/lua/javascript/` 😄

PolynomialDivision:
> Well, in that case `sed s/lua/javascript/` smile

Exam period is over. :) Finally, I can also code again. I will test that now and if it works I will merge the PR. ;)

PolynomialDivision:
But yeah, you are right. I have to adapt the config options in the lua app.

PolynomialDivision:
I think I also have adapt some other config exchange options.

champtar:
> > BTW, if I can learn lua
> 
> I think the general openwrt plan is to drop lua and completely switch to javascript. ;)

champtar:
> > BTW, if I can learn lua
> 
> I think the general openwrt plan is to drop lua and completely switch to javascript. ;)

I think server side will stay Lua forever, but putting most WebUI logic on the client side and just talk to a as small as possible API (ubus) make sense.

PR #16300, opened 2021-08-07
modemmanager: support setting plmn

maxrd2:
Maintainer: @nickberry17 
Compile tested: no (changed bash script only)
Run tested: ramips, DWR-921, openwrt-19.07 branch (git-21.044.30835-34e0d65) / OpenWrt 19.07.7 r11306-c4a6851c72

Tested connection using netifd with and without plmn setting

Description: 
This change allows to set network operator MCC:MNC and connect to different network with better signal quality and speed.

### DIFF
--- a/net/modemmanager/files/modemmanager.proto
+++ b/net/modemmanager/files/modemmanager.proto
@@ -343,6 +343,7 @@ proto_modemmanager_init_config() {
 	proto_config_add_string	 password
 	proto_config_add_string	 pincode
 	proto_config_add_string	 iptype
+	proto_config_add_string	 plmn
 	proto_config_add_int	 signalrate
 	proto_config_add_boolean lowpower
 	proto_config_add_defaults
@@ -355,11 +356,11 @@ proto_modemmanager_setup() {
 	local bearermethod_ipv4 bearermethod_ipv6 auth cliauth
 	local operatorname operatorid registration accesstech signalquality
 
-	local device apn allowedauth username password pincode iptype metric signalrate
+	local device apn allowedauth username password pincode iptype plmn metric signalrate
 
 	local address prefix gateway mtu dns1 dns2
 
-	json_get_vars device apn allowedauth username password pincode iptype metric signalrate
+	json_get_vars device apn allowedauth username password pincode iptype plmn metric signalrate
 
 	# validate sysfs path given in config
 	[ -n "${device}" ] || {
@@ -392,7 +393,7 @@ proto_modemmanager_setup() {
 	echo "starting connection with apn '${apn}'..."
 	proto_notify_error "${interface}" MM_CONNECT_IN_PROGRESS
 
-	connectargs="apn=${apn}${iptype:+,ip-type=${iptype}}${cliauth:+,allowed-auth=${cliauth}}${username:+,user=${username}}${password:+,password=${password}}${pincode:+,pin=${pincode}}"
+	connectargs="apn=${apn}${iptype:+,ip-type=${iptype}}${plmn:+,operator-id=${plmn}}${cliauth:+,allowed-auth=${cliauth}}${username:+,user=${username}}${password:+,password=${password}}${pincode:+,pin=${pincode}}"
 	mmcli --modem="${device}" --timeout 120 --simple-connect="${connectargs}" || {
 		proto_notify_error "${interface}" MM_CONNECT_FAILED
 		proto_block_restart "${interface}"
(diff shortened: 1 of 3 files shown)


feckert:
@nickberry17 ping

nickberry17:
ping @aleksander0m

net/modemmanager/files/modemmanager.proto
```diff
@@ -393,11 +394,16 @@ proto_modemmanager_setup() {
 		cliauth="${cliauth}${cliauth:+|}$auth"
 	done
 
+	if [ -n "$plmn" ]; then
+		echo "registering to network '${plmn}'..."
+		mmcli --modem="${device}" --3gpp-register-in-operator="$plmn"
```
aleksander0m, line 399: Just using the `operator-id` in the list of settings sent to `--simple-connect` should be enough; this explicit `--3gpp-register-in-operator` action should not be required. Plus, we're not checking the return status of this command, and that would mean that if it fails to register in that operator, it would still go on with the connection attempt.

The other option would be to add the error checking in the `--3gpp-register-in-operator` command, plus maybe add `--timeout=300` as the process may take longer than the default 30s timeout of a DBus request, and in that case, if the network registration to the explicit PLMN is done out of the connection attempt, then we could remove the `operator-id` setting in the `--simple-connect`.

In the very specific case of openwrt, where we configure the modem for data connectivity, I think using the `operator-id` only in the `simple-connect` should be more than enough.

maxrd2: > Just using the operator-id in the list of settings sent to --simple-connect should be enough; this explicit --3gpp-register-in-operator action should not be required. Plus, we're not checking the return status of this command, and that would mean that if it fails to register in that operator, it would still go on with the connection attempt.

Have tried that but it doesn't register to requested network immediately - it still registers to home network  with bad signal.  
Afterwards if i do ifdown/up it registers to requested network.
Doing `--3gpp-register-in-operator` before connect works always.

Will check for error and unset `$plmn` if it fails

maxrd2: Updated the patch.. Have added following in the commit description:

    The mmcli --3gpp-register-in-operator cmd always fails with
    "error: modem not enabled yet".
    
    This seems useless as "operator-id=MCC:MNC" is already passed in
    mmcli --simple-connect command.
    
    In my case not doing 3gpp-register-in-operator sometimes connects to
    wrong (home) MCC:MNC first time after reboot.

aleksander0m: > The mmcli --3gpp-register-in-operator cmd always fails with "error: modem not enabled yet".

This is telling you the command is fully ignored by ModemManager. It's like not sending the command at all.

> This seems useless as "operator-id=MCC:MNC" is already passed in mmcli --simple-connect command.

So, this is what is making the operator-id selection happen.

> In my case not doing 3gpp-register-in-operator sometimes connects to wrong (home) MCC:MNC first time after reboot.

That looks like a bug somewhere. Can you gather ModemManager debug logs when this happens, and report the issue in fd.o gitlab? https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/issues

So, after those comments, I'm now fully convinced we should remove the `--3gpp-register-in-operator` command from this patch.

maxrd2: great... let me know if you'd like me to squash two commits.
I've tried rebooting the modem few times now and was unable to make it happen, it happened once before that.
As soon as it happens will open an upstream issue.

maxrd2: It's been awhile, since this happened. If it happens it happens very rarely. :-/

maxrd2:
@aleksander0m have rebased to master

aleksander0m:
This looks fine to me

aleksander0m (approved):

feckert:
This PR is quite old! And seems to be forgotten!
But this change is still valid.
@maxrd2  I am willing to merge this change, but you have to update the PKG_RELEASE number. Could you please do that?

maxrd2:
> This PR is quite old! And seems to be forgotten! But this change is still valid. @maxrd2 I am willing to merge this change, but you have to update the PKG_RELEASE number. Could you please do that?

Sure... how do I do that exactly? Should I rebase to current master too?

maxrd2:
@feckert have rebased to master and incremented PKG_RELEASE by 1... is that correct?

feckert:
Thanks!

PR #16312, opened 2021-08-10
c-ares: update to version 1.17.2

ynezz:
Maintainer: @karlp 
Compile tested: None
Run tested: None

Missing input validation of host names returned by Domain Name Servers
in the c-ares library can lead to output of wrong hostnames (leading to
Domain Hijacking).

The Common Vulnerabilities and Exposures (CVE) project has assigned the
name CVE-2021-3672 to this issue.

References: https://c-ares.haxx.se/adv_20210810.html


neheb (approved):

BKPepe:
Would you please prepare pull requests against `OpenWrt 21.02` and `OpenWrt 19.07` branches as well, please?

BKPepe (approved):

karlp:
we've already done 8 service releases on 19.07, and it has a version of c-ares that doesn't have a patch available for it.  I'm happy with updating master and 21.02, but I've got zero time in figuring out and testing a patch for the older version in 19.07, nor would I feel comfortable making the major version bumps there after such a long time.  I propose just ignoring 19.07.

ynezz:
>  I propose just ignoring 19.07

There are only two options. Either we declare 19.07 EOL or fix this issue.

I would simply bump the c-ares version 1.15.0 in the openwrt-19.07 to the latest 1.17.2 release, because otherwise we would need to backport other [security related fixes](https://c-ares.haxx.se/changelog.html#1_15_0) as well:
  * Prevent possible use-after-free and double-free in ares_getaddrinfo() if ares_destroy() is called prior to ares_getaddrinfo() completing. Reported by Jann Horn at Google Project Zero.  (1.16.1)
  * avoid read-heap-buffer-overflow in ares_parse_soa_reply found during fuzzing (1.17.0)
  * Avoid theoretical buffer overflow in RC4 loop comparison (1.17.0)
  * Empty hquery->name could lead to invalid memory access (1.17.0)
  * ares_parse_{a,aaaa}_reply() could return a larger *naddrttls than was passed in (1.17.0)
  * NodeJS passes NULL for addr and 0 for addrlen to ares_parse_ptr_reply() on systems where malloc(0) returns NULL. This would cause a crash. (1.17.2)
  * When building c-ares with CMake, the RANDOM_FILE would not be set and therefore downgrade to the less secure random number generator (1.17.2)
  * If ares_getaddrinfo() was terminated by an ares_destroy(), it would cause a crash (1.17.2)
  * Crash in sortaddrinfo() if the list size equals 0 due to an unexpected DNS response (1.17.2)
  * Expand number of escaped characters in DNS replies as per RFC1035 5.1 to prevent spoofing follow-up (1.17.2)
  * Perform validation on hostnames to prevent possible XSS due to applications not performing valiation themselves (1.17.2)
  
It seems like they're doing good work with [ABI compatibility](https://abi-laboratory.pro/index.php?view=timeline&l=c-ares) and not much packages depends on c-ares anyway, I've found just `shadowsocks-libev`.

Cc: @badger @bradh352

neheb:
Lot of work for just one dependent package :)

karlp:
well, mosquitto uses it, which is how I ended up maintainer.  Bumping everything to latest in old branches is a garbage plan IMO, there's _always_ security fixes and security breakages, and there's _always_ things missing, the only way to make 19.07 "secure"is to replace it with master, which has "all" the fixes.  But, whatever, I know people have mixed and fantastic opinions on that sort of thing.

Regardless, I've been put in isolation for a week, and won't be able to do anything about this until (earliest) wednesday, august 18, I don't have any of my test hardware here to do any run testing on this.

PR #16315, opened 2021-08-11
collectd: enable cpufreq plugin for mediatek

selanf:
tested on e8450 ubi+non-ubi

### DIFF
--- a/utils/collectd/Makefile
+++ b/utils/collectd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=collectd
 PKG_VERSION:=5.12.0
-PKG_RELEASE:=9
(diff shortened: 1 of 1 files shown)


feckert:
Can you please bump the PKG_RELEASE number in the Makefile.

selanf:
Screenshot: 
image

feckert:
Thanks Merged

hnyman:
Interesting. I didn't even test it on my E8450/RT3200 as I though the scaling to be disabled.

selanf:
@hnyman: You need to set another governor, by menuconfig or during runtime. Without any chnages the device runs at full speed and full temperature.
I use it this way: https://github.com/openwrt/openwrt/pull/4440 but dangowrt want to do it in another way

PR #16317, opened 2021-08-11
samba4: allow disabling wsdd2 in menuconfig

selanf:


Andy2244:
sure this works too, np

PR #16318, opened 2021-08-11
xinetd: support for anonym configs

selanf:
this is needed for pr in luci with xinetd ipv6 support commits

### DIFF
--- a/net/xinetd/files/xinetd.init
+++ b/net/xinetd/files/xinetd.init
@@ -71,7 +71,7 @@ config_cb() {
 			local value="$2"
 
 			# write out last list option if new list starts
-			if [ "$ListName" != "" -a "$ListName" != "$name" ]; then
+			if [ -n "$ListName" ] && [ "$ListName" != "$name" ]; then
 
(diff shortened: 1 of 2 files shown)


net/xinetd/files/xinetd.init
```diff
@@ -62,7 +62,7 @@ config_cb() {
 			local option="$1"
 			local value="$2"
 
-			[ -n "$value" ] && echo -e "\t$option = $value" >> $GENERATED_CONF_FILE
+			[ -n "$value" -a "$option" != "name" ] && echo -e "\t$option = $value" >> $GENERATED_CONF_FILE
```
feckert, line 65: Please use `&&` instead of `-a`

feckert (changes requested):

net/xinetd/files/xinetd.init
selanf: Whats the advance of `-a`?  `value` should be in most cases not empty

feckert (approved):

feckert:
Thanks merged

PR #16319, opened 2021-08-11
collectd: select interface by default

selanf:
this is needed for commit in luci which allows disabling interface to keep default config as it is


selanf:
Related: https://github.com/openwrt/luci/pull/5246

hnyman:
Let's not change/complicate the collectd modules selection here.

PR #16325, opened 2021-08-11
stress-ng: add missing libkmod dependency

egorenar:
This fixes build problem introduced in commit 9a1bb4baf55f ("stress-ng: bump to version 0.13.00"):
Package stress-ng is missing dependencies for the following libraries:
libkmod.so.2

Fixes: 9a1bb4baf55f ("stress-ng: bump to version 0.13.00")

Maintainer: @commodo
Compile tested: Netgear XR700  OpenWrt SNAPSHOT r17335-5337e42a8be7
Run tested: Netgear XR700  OpenWrt SNAPSHOT r17335-5337e42a8be7

### DIFF
--- a/utils/stress-ng/Makefile
+++ b/utils/stress-ng/Makefile
@@ -27,7 +27,7 @@ define Package/stress-ng
   CATEGORY:=Utilities
   TITLE:=stress-ng is a stress test utility
   URL:=https://kernel.ubuntu.com/~cking/stress-ng/
-  DEPENDS:=+zlib +libbsd +libaio +libsctp
+  DEPENDS:=+zlib +libbsd +libaio +libsctp +libkmod
 endef
 
 define Package/stress-ng/description


cotequeiroz:
As an alternative, if libkmod functionality is not wanted, one can add
```
MAKE_FLAGS += HAVE_LIB_KMOD=0
```
This way it will not need that dependency.

commodo (approved):

commodo:
> Package stress-ng is missing dependencies for the following libraries:
> libkmod.so.2

hmm; not sure how i missed this on x86


> As an alternative, if libkmod functionality is not wanted, one can add
> 
> ```
> MAKE_FLAGS += HAVE_LIB_KMOD=0
> ```
> 
> This way it will not need that dependency.

thank you for the suggestion;
i think libkmod is ok to have pulled in [by default];
as a continuation, we could add a config symbol to remove it if people want, but i am not sure it's worth adding another symbol for this;

commodo:
> Sorry to butt in on this, but I think the dependency should be reviewed.
> 

thank you for stepping in;
this helps :)

> 
> On mvebu, `libkmod` adds only 32k to the installation of a package of 395k, but it has absolutely no use.
> 
> I'd add `HAVE_LIB_KMOD=0` for every arch but x86, adding the dependency for x86 only as well.

sounds like the best idea;
i will make some time to check more in-depth on this;
i am a bit swamped these days

commodo:
> > Sorry to butt in on this, but I think the dependency should be reviewed.
> 
> thank you for stepping in;
> this helps :)
> 
> > On mvebu, `libkmod` adds only 32k to the installation of a package of 395k, but it has absolutely no use.
> > I'd add `HAVE_LIB_KMOD=0` for every arch but x86, adding the dependency for x86 only as well.
> 
> sounds like the best idea;
> i will make some time to check more in-depth on this;
> i am a bit swamped these days

@cotequeiroz 
i took some time to check the repo locally;
i found this: https://github.com/ColinIanKing/stress-ng/blob/7741d6a373565021b1f809ecbe1370f23073fd91/core-module.c#L27

looks like `core-module.c` also uses libkmod to do kmod load/unload stress testing

my sentiment is that allowing libkmod as a default dependency would make sense;
any thoughts/objections?

PR #16330, opened 2021-08-11
ksmbd stuff: update to 3.4.1

neheb:
Maintainer: @Andy2244 
Compile tested: ath79

ping @Vladdrako @egorenar  @fda77 @ptpt52

### DIFF
--- a/kernel/ksmbd/Makefile
+++ b/kernel/ksmbd/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ksmbd
-PKG_VERSION:=3.3.9
-PKG_RELEASE:=1
+PKG_VERSION:=3.4.1
+PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/cifsd-team/cifsd/tar.gz/$(PKG_VERSION)?
-PKG_HASH:=c196d1773b9f89221133780fd189b550acbc56ac93c2e79260a70eab9853b3e1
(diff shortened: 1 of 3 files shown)


ptpt52:
but it is not reasonable to make depends on nfsd
we could patch the kernel to select the CONFIG_OID_REGISTRY independently

https://github.com/x-wrt/x-wrt/blob/master/target/linux/generic/hack-5.4/999-tristate-hack.patch

neheb:
In the buildbot case, all the kmods are selected as M and thus the kernel config enables all that's needed.

Vladdrako:
@neheb Compiled succssesfully.

egorenar:
Compiled for Netgear XR700 OpenWrt SNAPSHOT r17341-94e252da3f4d.

egorenar:
I did some quick tests on my router, seems to work just fine :+1:

ptpt52:
> In the buildbot case, all the kmods are selected as M and thus the kernel config enables all that's needed.

but it is not acceptable to make it depends on nfsd.

egorenar:
> > In the buildbot case, all the kmods are selected as M and thus the kernel config enables all that's needed.
> 
> but it is not acceptable to make it depends on nfsd.
good point, weird dependency for samba

neheb:
Right. I can't see a way to get it to compile without various hacks.

ptpt52:
ether put the ksmbd to kernel source tree or patch the kernel Kconfig to export symbol like CONFIG_OID_REGISTRY

neheb:
ping @hauke

Neustradamus:
@neheb: Can you look for 3.4.2?
- 19.07.x/21.02.x/master?

Thanks in advance

neheb:
no.

Neustradamus:
@neheb: Why? :/

neheb:
Because kernel modules cannot be freely updated for released versions.

PR #16331, opened 2021-08-12
ksmbd: set stoplevel

selanf:
to allow graceful stop of the daemon

### DIFF
--- a/net/ksmbd-tools/files/ksmbd.init
+++ b/net/ksmbd-tools/files/ksmbd.init
@@ -1,6 +1,7 @@
 #!/bin/sh /etc/rc.common
 
 START=98
+STOP=05
 USE_PROCD=1
 
 SMBD_IFACE=""


hnyman:
cc maintainer @Andy2244

Andy2244:
@hnyman looks fine

BKPepe:
This does not require to bump `PKG_RELEASE`? 🤔

hnyman:
Well,
* stop level does not change the launch or generated runtime config,
* to my knowledge stop level is rarely actually used in the home router context (as many routers do not have any proper shoutdown, and sysupgrade just kill processes),
* there is soon coming the change from #16332 where I hope to see the PKG_RELEASE bump.

So I didn't think that PKG_RELEASE would be that important here.

champtar:
@hnyman I agree it's no a huge issue, but to make it simple if the package hash changes so should the version ;)

selanf:
Sry about the PKG_RELEASE...
stoplevel is the order in which the services are stopped on halt (not really used here) and reboot by terminal or luci.
eg "umount" is STOP=90. If you have a file transfer to usb device on reboot, unmounting could fail and in worst case filesystem problems (not only samba/ksmbd).
So imho ALL services should be stopped gracefull on reboot! No stoplevel is set? -> A very high default value is used, eg99. But this value should be smaller than umount! I dont know openwrt good enough to say this is a good idea for all services, maybe some should not be stopped
I think the startlevel in openwrt are not so well designed, see https://github.com/openwrt/openwrt/pull/4446 and https://github.com/openwrt/openwrt/pull/4445 (maybe a security problem).
If you want to take a look at freetz: https://freetz-ng.github.io/freetz-ng/wiki/60_Development/startlevel_of_packages.en.html
The vaiables in the init script are a little slow to parese with many services 
See also https://github.com/openwrt/luci/pull/5271

This all does not count for sysupgrade! It seems to ignores all /etc/rc.d/K* files and just kills the system! So i have put localy a call to a /etc/rc.reboot file into sysupgrade itself so at least statistic backups from /tmp/ are saved on sysupgrade

PR #16332, opened 2021-08-12
ksmbd: use hostname only as name

selanf:
even if a fqdn (with domain) is set by user


hnyman:
cc maintainer @Andy2244

Andy2244:
@hnyman  looks fine

hnyman:
@selanf 
The package needs also a bump of PKG_RELEASE (as it doesn't use the AUTORELEASE mechanism).

PR #16339, opened 2021-08-12
netdata: update to version 1.33.1

paper42:
Signed-off-by: Michal Vasilek <michal.vasilek@nic.cz>

Maintainer: @BKPepe, @diizzyy
Compile tested: armv7, Turris Omnia, 21.02
Run tested: armv7, Turris Omnia, 21.02

Description: Without patching the version, the OS version is reported instead [[1]](https://github.com/netdata/netdata/issues/11214) [[2]](https://github.com/netdata/netdata/issues/11416).
They also changed the UI, so the old patches don't apply, but they might not be necessary anymore.

old UI:
old UI

new UI:
new UI

Should I patch out the automatic update notification from the right of the top panel? (yellow icon on the left) The lightbulb icon next to that is Netdata news, I don't think it needs to be patched out.


diizzyy:
Thanks for looking into this!

That should be patched imho, discussed breifly here:
https://github.com/openwrt/packages/commit/1d532fa545eef1ebd3ebef6ab41dfd709ad991e1#commitcomment-52088675
This is what we ended up doing in FreeBSD
https://cgit.freebsd.org/ports/tree/net-mgmt/netdata/Makefile#n9 (fd15080f65fde1c484a06cbc5a13a2853639a960.patch)

What I would be a bit concerned about is the total package size due to this change of WebUI

paper42:
> That should be patched imho, discussed breifly here:
> [1d532fa#commitcomment-52088675](https://github.com/openwrt/packages/commit/1d532fa545eef1ebd3ebef6ab41dfd709ad991e1#commitcomment-52088675)

The two commits from your second comment sound fine, but let's ask @BKPepe 
https://github.com/forkbomb9/dashboard/commit/d4c1d0b2306a154b94426d0b26c33e3b2a8394f1 https://github.com/forkbomb9/dashboard/commit/dc300d1c84118f9b1c7f6b618aac35aa5a75aac4

> What I would be a bit concerned about is the total package size due to this change of WebUI

I will measure it tomorrow, but I don't think there is much we can do with it.

diizzyy:
We should also move back to using release archives by upstream

paper42:
> That should be patched imho, discussed breifly here:
> [1d532fa#commitcomment-52088675](https://github.com/openwrt/packages/commit/1d532fa545eef1ebd3ebef6ab41dfd709ad991e1#commitcomment-52088675)
> This is what we ended up doing in FreeBSD
> [cgit.freebsd.org/ports/tree/net-mgmt/netdata/Makefile#n9](https://cgit.freebsd.org/ports/tree/net-mgmt/netdata/Makefile#n9) (fd15080f65fde1c484a06cbc5a13a2853639a960.patch)

So if I understand this well, FreeBSD uses a fork of netdata/dashboard that applies the small patch mentioned before and pre-builds it, so it can be used for netdata/netdata, but that also makes the patch quite big. So I could import a big patch (probably not a good idea) or switch dashboard to the mentioned fork (which may break on the next update). I am not sure either option is ideal.

> What I would be a bit concerned about is the total package size due to this change of WebUI

It's a big size increase:
```
2261321 netdata_1.30.1
7539985 netdata_1.31.0
```

> We should also move back to using release archives by upstream

I am not sure what you mean by that, could you elaborate?

diizzyy:
The dashboard is built using npm and on top of that it's minified so it's hard to generate a patch for the included dashboard which is why it's rebuilt. Irregardless it'll break during each update but thankfully netdata isn't released very frequently. I don't see any obvious way of switching dashboard looking at https://github.com/forkbomb9/netdata/commit/fd15080f65fde1c484a06cbc5a13a2853639a960 . Unfortunately due the move of a new dashboard netdata grew a lot in size (for embedded hardware at least) which is what I suspected, you could probably slim it down bit by removing a few variants of the fonts but that's about it. It would be great if these changes were toggles but I doubt upstream have interest in "debranding" their  product.

Regarding the release archive I meant that we should use https://github.com/netdata/netdata/releases/download/v1.31.0/netdata-v1.31.0.tar.gz instead of a GitHub generated one.

Beginner-Go:
@paper42 Hi, Upstream released a new version
from: https://github.com/netdata/netdata/releases/tag/1.32.1

BKPepe:
> It's a big size increase 2261321 vs 7539985

That's not acceptable here, though. It's huge! 

> We should also move back to using release archives by upstream

I don't mind it.

___

I will look into updating netdata to the version 1.32.1 soon.

diizzyy:
Looking at development I guess there's no way around it unfortunately...

BKPepe:
@diizzyy, I agree with you. However, I think there is no good solution for everyone. In my case, I don't mind it, because most devices which I have are running on mvebu, and they have enough storage. 

I looked at recently released version 1.33.0, and they require OpenSSL and a few additional dependencies. It may be required in older releases, but the size increase is incredible.

Based on @paper42 calculations, I'm posting size for netdata 1.33.0: ``8276112``.

diizzyy:
The WebUI uses a lot and I don't mind OpenSSL dependency as it is by far the fastest SSL library.

BKPepe:
Netdata released a new version 1.33.1, where with explicitly disabled HTTPS, you don't pull OpenSSL dependency (https://github.com/netdata/netdata/issues/12056#issuecomment-1042697538)

But the size of this release is ``8301799`` with all removed patches. :-(

BKPepe (approved):

diizzyy:
Use upstream tarball and I would prefer if we could import https://github.com/forkbomb9/netdata/commit/eaaddaa3b1f4b748e2468ffd4d7afbc7cb0efe78 to avoid clutter in UI

BKPepe:
Switched to upstream releases tarball as suggested. I will take a look at the patch if it is applies and let you know about the package size.

BKPepe:
Netdata 1.33.1 size:
- Before the FreeBSD patch: 8407758
- After the FreeBSD patch: 8097939

Compile tested for Turris 1.1, powerpc8540, OpenWrt 21.02.2. I will do run testing shortly.

BKPepe:
Run tested.

diizzyy:
Go for it :)

neheb:
```
Package netdata is missing dependencies for the following libraries:
liblz4.so.1
```

BKPepe:
Thanks, I will take a look at it shortly.

paper42:
Hmm, why didn't the CI catch this?

neheb:
because liblz4 was not built with CI.

diizzyy:
"Dirty" build environment

PR #16342, opened 2021-08-12
audit: move audit and libev to openwrt core

feckert:
Maintainer: @aparcar @karlp 
Run tested: x86_64 , APU3, only package move

Description:
Move audit to openwrt core to unify the compilation packege libaudit and audit
The package libev is moved as well because audit depends on this lib.


feckert:
The pull request is rotten, so I'm closing it.

PR #16346, opened 2021-08-12
auc: backport version 0.1.9

aparcar:
Maintainer: @dangowrt

### DIFF
--- a/utils/auc/Makefile
+++ b/utils/auc/Makefile
@@ -5,7 +5,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=auc
-PKG_VERSION:=0.1.4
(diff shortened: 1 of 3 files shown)


dangowrt:
I don't think backport to 19.07 is feasible due to lacking ABIVersion-stripping support, we would need to backport commits
https://github.com/openwrt/openwrt/commit/67166fabda53f17beac304db5bc44e4be6a3ba13
and
https://github.com/openwrt/openwrt/commit/69b77dc3b4480f90f02878e09e924a8e3f0116b6
to openwrt-19.07

aparcar:
ABI stripping is now also done by the server, clients don't need to to it.

dangowrt:
superseded

PR #16349, opened 2021-08-12
isc-dhcp: use uci_get_state

champtar:
Maintainer: @pprindeville 
Compile tested: NO
Run tested: NO

Description:

Since https://git.openwrt.org/?p=openwrt/openwrt.git;a=commitdiff;h=4b9a67362d70c544b85078b8d5c661f43f7472d9,
uci network config interface sections use 'device' instead of 'ifname'
Also when wan is a bridge the uci config ifname is a list,
but the uci state ifname is the actual interface

Leaving as draft as this really needs testing to be sure the uci state is populated when using isc-dhcp


pprindeville:
If it avoids additional processes at startup, and leads to a faster boot, then I'm all for it... just need it tested...

jow-:
The `uci_get_state` is an ancient deprecated api and only maintained for backwards compatibility with very old scripts. Consider using `/lib/functions/network.sh` to determine the l3 devices of logical interfaces, e.g. 

```bash
. /lib/functions/network.sh 

if network_get_device devname wan; then
    echo "Device of wan is $devname"
else
   echo "Cannot determine device of wan"
fi
```

pprindeville:
> The `uci_get_state` is an ancient deprecated api and only maintained for backwards compatibility with very old scripts. Consider using `/lib/functions/network.sh` to determine the l3 devices of logical interfaces, e.g.

Seems reasonable.  @champtar?

pprindeville:
Superseded by #21666

PR #16361, opened 2021-08-14
v4l-utils: add libudev-zero dependency

stintel:
Maintainer: @thess
Run tested: NA

Description:
When libudev-zero is enabled, v4l-utils links against it, and the
package build fails due to a missing dependency:

Package v4l-utils is missing dependencies for the following libraries:
libudev.so.1

As disabling udev doesn't seem to be supported in v4l-utils, simply add
the dependency to fix the build.


neheb:
Sounds wrong. There's a patch to disable it: https://github.com/openwrt/packages/blob/59771bde0ceff2a3b84b1ee7d119948d18da6e65/libs/libv4l/patches/010-remove-libudev-check.patch

Strange that it's not working.

stintel:
Honestly I rather add the dependency and drop a patch that obviously isn't working and probably wasn't submitted upstream either.

neheb:
The patch was to deal with host libudev leaking into the build. Now that there's libudev-zero, might as well add the dependency and remove that patch.

neheb:
Actually that patch was introduced when OpenWrt was still using pkg-config. I don't think it's needed now that OpenWrt uses pkgconf.

thess:
I think that adding the dependency and dropping the `010-remove-libudev-check.patch` is the right direction. If we are to include `libudev` here, we could also add back the **dvbv5** library and utils to the package as well as the associated "dev" components - maybe as an option?

stintel:
> I think that adding the dependency and dropping the `010-remove-libudev-check.patch` is the right direction. If we are to include `libudev` here, we could also add back the **dvbv5** library and utils to the package as well as the associated "dev" components - maybe as an option?

I don't think **dvbv5** and utils are related to this PR. This PR is just about fixing a build problem.

neheb:
Sure. Can you also remove the udev patch?

thess:
LGTM - Thanks.

PR #16363, opened 2021-08-14
tcpreplay: avoid host lib leakage

stintel:
Maintainer: @commodo
Run tested: NA

Description:
On hosts that have pcapnav-config installed, there is host lib leakage.
From config.log:

LNAVLIB='-L/usr/lib64 -lpcapnav -lpcap'
LNAV_CFLAGS='-I/usr/include'

Fix this by disabling pcapnav-config, which isn't available anyway.


neheb (approved):

commodo (approved):

commodo:
just one note;
maybe bump PKG_RELEASE to 2
or set to AUTORELEASE

rest looks good; feel free to merge when ready;

thanks

stintel:
This shouldn't be needed, as the package itself doesn't change?

commodo:
> This shouldn't be needed, as the package itself doesn't change?

some people may argue otherwise;
but it's fine from my side;

PR #16364, opened 2021-08-14
mtools: fix host lib leakage + version bump

stintel:
Maintainer: @oskarirauta 
Run tested: not relevant for the host lib leakage fix, bumped to test if it would fix the host leakage which it didn't


stintel:
@oskarirauta ping

PR #16366, opened 2021-08-14
rtl_433: disable SoapySDR

stintel:
Maintainer: @TimelessNL
Run tested: NA

Description:
When building on a host with SoapySDR installed, the package fails to
build as CMake picks up the host SoapySDR CMake module. As SoapySDR is
not available in OpenWrt, simply disable SoapySDR to fix build.

```
CMake Error at /usr/share/cmake/SoapySDR/SoapySDRConfig.cmake:153 (message):
  cannot find SoapySDR library in /usr/lib
Call Stack (most recent call first):
  CMakeLists.txt:113 (find_package)
```

### DIFF
--- a/utils/rtl_433/Makefile
+++ b/utils/rtl_433/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=rtl_433
 PKG_VERSION:=20.11
-PKG_RELEASE:=1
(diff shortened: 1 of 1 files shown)


stintel:
@TimelessNL ping

stintel:
Ok. Let's merge as-is and if SoapySDR lands in OpenWrt someday, we can look for another way to avoid host lib leakage.

PR #16378, opened 2021-08-16
hplip: update to 3.21.6

neheb:
Refreshed patches.

Maintainer: @luizluca 
Compile tested: ath79


luizluca:
LGTM

@neheb , do you have a real HP printer to test hplip? I lost my old printer and the new one uses proprietary scanner driver. Would you mind to take the role as maintainer?

neheb:
I do not.

I also do not use this.

luizluca:
So, let's keep updating it with only build test until something break...

PR #16389, opened 2021-08-17
apparmor: cleanup

paper42:
Maintainer: @oskarirauta 
Run tested: TODO

Description:
* replace 020-fix-ss-path.patch with 020-remove-which-dep.patch which fixes the problem that fix-ss-path was fixing and removes a runtime dependency on which (https://gitlab.com/apparmor/apparmor/-/merge_requests/784)
* remove the aa-notify patch and do not install aa-notify, because it doesn't make much sense for OpenWrt @oskarirauta do you agree with this?


oskarirauta:
I agree.

paper42:
@oskarirauta what's the purpose of the 190 line patch called `060-openwrt-dnsmasq-profile.patch`? As far as I can tell, it just adds 2 lines and removes a few comments, otherwise it's just rearranging the lines. Can we remove most of the changes?
```patch
+  /tmp/** r,
+  include <abstractions/user-tmp>
-  # Neighbor Discovery protocol (RFC 2461)
-  # NetworkManager integration
-  # Site-specific additions and overrides. See local/README for details.
-  # access to iface mtu needed for Router Advertisement messages in IPv6
-  # for the read-only TFTP server
-  # libvirt config and hosts file for dnsmasq
-  # libvirt lease helper
-  # libvirt pid files for dnsmasq
-  # lxc-net pid and lease files
-  # lxd-bridge pid and lease files
-    # libvirt lease and status files for dnsmasq
```

paper42:
* simplify 060-openwrt-dnsmasq-profile.patch                                                                                                                                              
* add two upstream patches for Python 3.10 compatibility

oskarirauta:
Yes, comment section removal from patch does not affect to build. This was some time ago; so I am not sure why I did patch it that way; but could be that patch is based on a patch for some other distro and I maybe didn't clean it up- difficult to say when so much time has passed since that..

jefferyto:
I'd like to suggest also incorporating these changes:

* Download the tagged release (tar.bz2) from https://gitlab.com/apparmor/apparmor/-/tags, instead of checking out a commit from the master branch
* Include [this patch](https://gitlab.com/apparmor/apparmor/-/commit/ab4cfb5e84a95decd06487dc4e59827ddb8ce79d) (from the `apparmor-3.0` release branch) to fix the Python 3.10 distutils error (I will remove the distutils error patch from the python3 package soon)
* For the other two Python 3.10 patches in this PR, use the versions from the `apparmor-3.0` release branch (I believe they are [this commit](https://gitlab.com/apparmor/apparmor/-/commit/6cfc6eeec22d3c9ffae413819654359b76680150) and [this commit](https://gitlab.com/apparmor/apparmor/-/commit/06dc32a9f3ede351c46167afa5002a0fe062c0f7))

@oskarirauta would these changes be okay with you?

In the future (not necessary for this PR), I believe it would be better to follow [Debian/Ubuntu's packaging](https://packages.debian.org/source/sid/apparmor) more closely, i.e.:

* Rename python3-apparmor to python3-libapparmor
* Move core binaries/scripts from apparmor-utils into an "apparmor" package

paper42:
> I'd like to suggest also incorporating these changes:
> 
>     * Download the tagged release (tar.bz2) from [gitlab.com/apparmor/apparmor/-/tags](https://gitlab.com/apparmor/apparmor/-/tags), instead of checking out a commit from the master branch

done

>     * Include [this patch](https://gitlab.com/apparmor/apparmor/-/commit/ab4cfb5e84a95decd06487dc4e59827ddb8ce79d) (from the `apparmor-3.0` release branch) to fix the Python 3.10 distutils error (I will remove the distutils error patch from the python3 package soon)

I imported a different tiny patch which just ignores the warning, is that ok or would you prefer the patch you linked?

>     * For the other two Python 3.10 patches in this PR, use the versions from the `apparmor-3.0` release branch (I believe they are [this commit](https://gitlab.com/apparmor/apparmor/-/commit/6cfc6eeec22d3c9ffae413819654359b76680150) and [this commit](https://gitlab.com/apparmor/apparmor/-/commit/06dc32a9f3ede351c46167afa5002a0fe062c0f7))

done, thank you for the review

jefferyto:
> I imported a different tiny patch which just ignores the warning, is that ok or would you prefer the patch you linked?

I have a slight preference for the upstream patch because it is upstreamed but I'm okay with the current patch as well, since it is temporary (the next release will have the upstream patch included).

Apologies for missing this earlier: for 020-remove-which-dep.patch I would also prefer to keep the original commit message (I believe it is [this commit](https://gitlab.com/apparmor/apparmor/-/commit/c31d49234f0c7a81b69d525e249b154e11c486a9)).

jefferyto:
The upstream distutils patch replaces the use of distutils with setuptools. distutils installs package metadata as *.egg-info files, whereas setuptools installs package metadata as *.egg-info directories. CI is failing because the Makefile install rules haven't been adjusted to account for this change.

paper42:
> The upstream distutils patch replaces the use of distutils with setuptools. distutils installs package metadata as *.egg-info files, whereas setuptools installs package metadata as *.egg-info directories. CI is failing because the Makefile install rules haven't been adjusted to account for this change.

Thanks, I fixed it by not installing egg-info files as discussed previously in another PR.

jefferyto:
> Thanks, I fixed it by not installing egg-info files as discussed previously in another PR.

Those files are required for pip to recognize that the Python packages are installed.

paper42:
> Those files are required for pip to recognize that the Python packages are installed.

Fixed, I am sorry that I previously misunderstood the discussion.

jefferyto:
> Fixed, I am sorry that I previously misunderstood the discussion.

No problem - thanks for your work on this PR :smile:

jefferyto (approved):

PR #16396, opened 2021-08-18
git: update to 2.33.0

paper42:
Maintainer: @tripolar
Compile tested: aarch64, Turris MOX, OpenWrt 21.04
Run tested: aarch64, Turris MOX, OpenWrt 21.04

Description: I cloned a few repositories and tested log, commit and pull. All patches seem to still be necessary.


BKPepe (approved):

paper42:
The build seems to fail only in CI.

PR #16399, opened 2021-08-18
libwebsockets: avoid a compilation error due to a missing uloop

k-ronny:
With a clean build environment and at least two parallel jobs
a compilation error may occur:
```
...
-- Looking for uv/version.h - found
libubox include dir: ULOOP_INCLUDE_DIRS-NOTFOUND
libubox libraries: ULOOP_LIBRARIES-NOTFOUND
...
CMake Error in lib/event-libs/uloop/CMakeLists.txt:
  Found relative path while evaluating include directories of
  "websockets-evlib_uloop":

    "ULOOP_INCLUDE_DIRS-NOTFOUND"
...
```

Signed-off-by: Ronny Kotzschmar <ro.ok@me.com>


neheb:
ping @karlp

feckert:
@k-ronny Do we need a DEPENDS here and not just a builld dependency?

karlp:
We could also just turn off uloop.  I enabled it as it was a new feature upstream, but no-one's using it.  It's optional though, so it shouldn't be a hard depends, this goes into the soup of "what is the default package"  You should absolutely be able to install lws without uloop.

k-ronny:
@feckert It is a `PKG_BUILD_DEPENDS`. The way I understand the documentation [OpenWrt: BuildPackage variables](https://openwrt.org/docs/guide-developer/packages#buildpackage_variables): 
> Packages that need to be built before this package

karlp:
this has merged to master, thanks!

Issue #16353, opened 2021-08-13
jq: Current version 1.6 is awfully slow


Fanfwe:
Maintainer: @ratkaj
Environment: any

Description:
jq in its version 1.6 has got a change which gives it a very bad performance. This is especially visible in the OpenWRT world as the target platforms often feature low end CPUs.

As an example, running this basic command:
`
echo '{}' | jq
`
On my ATH79 based device takes over a second (!) when it literally was instant with the previous 1.5 version.

jq upstream is aware of the problem, has acknowledged it and even fixed it. See https://github.com/stedolan/jq/issues/1826

Problem is this is only fixed in master but there hasn't been any new release, although the fix has been there for over two years now.

So I'm writing this ticket because I'm willing to submit a pull request in order to get this fixed in OpenWRT, but I'd like to know what approach would be preferred.

I can think of two ways:

- Changing the Makefile so that it pulls a version from a more recent commit instead of the 1.6 tag. This means we will no longer have a true 1.6 but something newer instead. For reference, this is the approach that has been taken by buildroot: https://github.com/buildroot/buildroot/commit/b640c6d4cff5b6fd8480aeebc630443b6ca09b2a
- Leaving the Makefile as is, but instead backport the fix (and just that) into a patch file

Let me know what you think

BKPepe:
Hi,

I think backporting the fix into a patch file sounds good until `jq` will release a new version.

PR #16401, opened 2021-08-18
jq: Backport a bunch of commits in order to fix jq-1.6 terrible performance on startup

Fanfwe:
Maintainer: @ratkaj
Compile tested: MIPS(ATH79) on master
Run tested: MIPS(ATH79) on master. Test performed: echo '{}' | jq

Description:
As reported in #16353 there is a very big performance issue with jq-1.6 which was a regression compared to previous 1.5 version. jq simply takes a long time to start (over a second on an ATH79 based router) which makes scripts that rely on calling jq multiple times awfully slow.
upstream is aware of the issue and has in fact already fixed it in their master branch over two years ago, but there hasn't been any new official release since.
This PR hence cherry-picks the required patchset and puts it in the package so we can have a properly working jq until upstream makes a new release.


neheb:
Need to run make package/jq/{compile,refresh}

neheb:
PKG_RELEASE bump is needed.

PR #16403, opened 2021-08-19
Update info URL in Makefile for BCP38

tiagofreire-pt:
Maintainer: me / @tohojo 
Compile tested: N/A
Run tested: N/A

Description:

The url `https://tools.ietf.org/html/bcp38` redirects currently to the one proposed in this PR `https://www.rfc-editor.org/info/bcp38`.

Makes sense to point straight to the final one, where resides the useful information.


tohojo:
The change itself is fine, but please write a proper commit message, including a S-o-B line...

PR #16421, opened 2021-08-21
nfdump: fix compilation with PKG_ASLR

neheb:

Maintainer: @flyn-org 
Compile tested: ath79


flyn-org:
Looks fine. I will try to address this with upstream when I have time.

neheb:
The actual issue is that OpenWrt CFLAGS are not being passed because of this line. This is also a bugfix in disguise.

flyn-org:
Yes, let's merge.

flyn-org:
See https://github.com/phaag/nfdump/issues/304.

PR #16422, opened 2021-08-21
gerbera: update to 1.9.0

neheb:
Maintainer: 
Compile tested: ath79


neheb:
@micmac1 found the issue with GCC. Let's see how upstream fixes this.

neheb:
Added Os patch based on upstream feedback.

PR #16426, opened 2021-08-21
dnsdist: replace PKG_BUILD_DEPENDS with normal one

neheb:
It adds it there anyway.

Maintainer: @Habbie 
Compile tested: ath79

### DIFF
--- a/net/dnsdist/Makefile
+++ b/net/dnsdist/Makefile
@@ -17,7 +17,6 @@ PKG_INSTALL:=1
 PKG_BUILD_PARALLEL:=1
 
 PKG_ASLR_PIE:=0
-PKG_BUILD_DEPENDS:=boost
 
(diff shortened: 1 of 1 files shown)


Habbie:
> It adds it there anyway.

What adds what where anyway? I removed boost from the runtime dependencies to save (a few kilobytes) of disk space.

neheb:
AFAIK the boost package does not install anything.

All dependencies become build dependencies.

Habbie:
> AFAIK the boost package does not install anything.

Yes, I recall it is empty, but having it still costs a few kilobytes. I did this during a cleanup that yielded megabytes in other places. If you prefer a simpler Makefile that does not go as far as I did, this PR is fine with me :-)

neheb:
Eh it's fine.

PR #16436, opened 2021-08-22
[RFC/WIP/POC] bring up graphics stack

dangowrt:
Maintainer: me
Compile tested: mediatek/mt7623
Run tested: mediatek/mt7623, BananaPi R2

Description:
Bring up packages needed for graphics output, as some boards come with HDMI, LVDS or MIPI interfaces and may well serve a graphical user interface (such as LuCI inside a fullscreen webkit view). While this should all eventually end up in https://github.com/openwrt/video , however the problem is that meson is still only available in the packages feed until https://github.com/openwrt/openwrt/pull/4287 gets merged.

What works (on BPi-R2):
 * building everything needed for weston
 * running weston with `fbdev` backend: weston-terminal, weston-flower, ... work as expected (fonts and cursor icons are still missing though)
 * kmscube with vsync 60 FPS fullscreen on 1080p HDMI screen connected
 
What doesn't work yet:
 * running weston with `drm` freezes video output (but detection looks fine, see https://termbin.com/4zcq )
 * probably a lot more things...


aparcar:
🍿

neheb:
@dangowrt meson can be included just fine. The only issue with that PR is the pkgconf commit. For some reason, the meson build of pkgconf uses wrong paths.

PR #16446, opened 2021-08-23
mdio-tools: fix buildbot build

micmac1:
On buildbots the build fails because git isn't finding any git repo and
then AC_INIT refuses to run:

fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
configure.ac:5: error: AC_INIT should be called with package and version arguments

Address this by substituting the git command with $(PKG_VERSION).

Maintainer: @dmascord 
Compile tested: master
Run tested: N/A, build fix

Description:
Hi Damian,

A suggestion to get the build to complete on OpenWrt build bots.

Kind regards,
Seb

### DIFF
--- a/net/mdio-tools/Makefile
+++ b/net/mdio-tools/Makefile
@@ -35,4 +35,9 @@ define Package/mdio-tools/install
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/src/mdio/mdio $(1)/usr/bin/
 endef
 
+define Build/Prepare
+	$(call Build/Prepare/Default)
+	$(SED) 's/m4_esyscmd_s(.*)/$(PKG_VERSION)/' $(PKG_BUILD_DIR)/configure.ac
+endef
+
 $(eval $(call BuildPackage,mdio-tools))


dmascord:
Hi Sebastian,

Do you know why this only happens on the build bot and not on a local build ?  Is there some environmental sanity that should be checked when building locally so that it does build properly in the built bot ?

Cheers,

Damien

neheb:
Note that the buildbots use the SDK and not a normal cloned tree.

micmac1:
I think the real story is actually a bit funny. When "git describe --always --dirty" is running it checks for a git repo. In the directory "mdio-tools-2021-07-19-65f6898f" it can't find any, because the git info is stripped away from this directory. So it goes up in the hierarchy until it reaches "/". If it can find something it'll use that. If not, not. So in my git checkout it goes all the way up to the openwrt checkout and extracts the info from that. But then it ends up with the OpenWrt version, really, which has no relation to mdio-tools :)

micmac1:
So I guess you could say it never works as intended in OpenWrt. But the build may still succeed when there's some git repo on the way towards "/".

dmascord:
Interesting, can this be detected in the build system, so that it can fail out ?

micmac1:
Maybe check git docs to see if you can prevent going up the filesystem hierarchy and take it from there. But why waste more time on this?

dmascord:
Well, I don't mind just fixing this here and now, but I am sure this will be encountered for other packages in future, and solving it again and again with it not failing in the standard non-SDK build, or essentially failing silently.

micmac1:
I think "again and again" is a bit of an overstatement. I think the way mdio-tools chooses to handle the version is a bit special. Can't remember seeing anything similar in OpenWrt before.

dmascord (approved):
Based on our discussion, I am happy to Approve this

PR #16447, opened 2021-08-23
Htop sensors

graysky2:
Compile tested: x86_64
Run tested: bcm27xx/bcm2711/RPi4B
Description: Hitting F2 within htop should allow for an optional display of the CPU temp if compiled against lm-sensors but we currently do not build as such.  I tested the resulting package on my RPi4 but this hardware does not use kernel modules for temp so I did not see data displayed but I did see the missing prompt and "N/A" when enabled consistent with functional htop.

### DIFF
--- a/admin/htop/Makefile
+++ b/admin/htop/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=htop
 PKG_VERSION:=3.0.5
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/htop-dev/htop/tar.gz/$(PKG_VERSION)?
@@ -19,7 +19,7 @@ PKG_LICENSE:=GPL-2.0-or-later
 PKG_LICENSE_FILES:=COPYING
 PKG_CPE_ID:=cpe:/a:htop:htop
 
-
+PKG_BUILD_DEPENDS:=lm-sensors
 PKG_FIXUP:=autoreconf
 PKG_BUILD_PARALLEL:=1
 PKG_INSTALL:=1


graysky2:
Not sure why there are two commits in this PR.  It is clearly one on my side :confused:

hnyman:
At the first glance this looks suspicious to me.
I would expect it to depend on the libsensors library, also for runtime.

Based on your sentence about testing, it seems to me that this did not even work for you, so the quality of this change is untested... 

Only some processors have CPU temp monitoring, so the functionality would be for a narrow population.

I would like to keep htop as a lean processor monitoring tool, and not beef it up to an include-all system status behemoth. So I am negative toward this change (which would likely require some further polishing work).

graysky2:
I can't argue with the testing status.  If it is working, the change in size vs 3.0.5-1 is trivial:

3.0.5-1 is 154016
3.0.5-2 is 158112

So like 4K larger.

champtar:
@hnyman I looked at the http configure file and my understanding is that it's loading lm-sensors via dl-open if present.

jmarcet:
It is indeed dlopened. Building with lm-sensors installed, seems htop uses it, so the commit should be fine. Here on a i3770 based OpenWRT:

```
❯ ldd /usr/bin/htop
/lib/ld-musl-x86_64.so.1 (0x7f4bea8f4000)
libncursesw.so.6 => /usr/lib/libncursesw.so.6 (0x7f4bea869000)
libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x7f4bea855000)
libc.so => /lib/ld-musl-x86_64.so.1 (0x7f4bea8f4000)
```

image

champtar:
@graysky2 could you make it configurable and default enabled for x86 only ?

graysky2:
@champtar - I don't know the knowledge to do that.

champtar:
Ok, do you want me to send pointers so you can try, else I can do it ?

graysky2:
Feel free to take it on.  If you get a functional PR, we can close this one.

champtar:
See #16504

PR #16450, opened 2021-08-24
Compile with GCC-10 fails without -fcommon option

droncheg:
Without this it fails to build in latest trunk which uses gcc-10:
/openwrt/snapshot/staging_dir/toolchain-mips_24kc_gcc-10.3.0_musl/lib/gcc/mips-openwrt-linux-musl/10.3.0/../../../../mips-openwrt-linux-musl/bin/ld: ./.libs/libradiotap.a(radiotap.o):(.bss.__packed+0x0): multiple definition of `__packed'; lib/osdep/.libs/libaircrack_osdep_la-linux.o:(.bss.__packed+0x0): first defined here

Tested on ath79 platform

Description:


neheb:
https://github.com/openwrt/packages/pull/16405

PR #16453, opened 2021-08-24
grep: update to 3.7

paper42:
Maintainer: @Zokormazo
Compile tested: aarch64, Turris MOX, OpenWrt 19.07
Run tested: aarch64, Turris MOX, OpenWrt 19.07

Description: I couldn't test ppc, but it seems like it's broken, so I imported a patch from Void.


stintel:
@mangix this PR adds a patch, but the commit message does not mention what the patch does, nor why it was added, so this PR should have not been accepted in this state. Please do proper review before merging.

paper42:
> this PR adds a patch, but the commit message does not mention what the patch does, nor why it was added, so this PR should have not been accepted in this state. Please do proper review before merging.

The patch name should be self-explanatory - `ppc-musl.patch` and the patch contains a source, but I will make sure to mention patches in the commit message next time.

PR #16458, opened 2021-08-25
usbutils: bump to 014

kuanyili:
Remove dependency on `usbids` as `lsusb` binary does not use `usb.ids` file anymore, instead it uses udev hardware database.

Maintainer: @nbd168
Compile tested: aarch64, Raspberry Pi 4 Model B Rev 1.1, OpenWrt SNAPSHOT, r17399-6393ea1581
Run tested: as above

References:
https://github.com/gregkh/usbutils/commit/94986fc7abcc6bd932a3f45501e83f4f5a4126f2
https://github.com/gregkh/usbutils/commit/44c9cd24380531d3b6354499aee174ba23bb5e8c

### DIFF
--- a/utils/usbutils/Makefile
+++ b/utils/usbutils/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=usbutils
-PKG_VERSION:=013
-PKG_RELEASE:=2
+PKG_VERSION:=014
+PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@KERNEL/linux/utils/usb/usbutils
-PKG_HASH:=9e23494fcc78b7a80ee29a07dd179c95ae2f71509c35728dbbabc2d1cca41338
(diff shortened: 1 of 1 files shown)


neheb:
??? where is this database located?

kuanyili:
Umm.... This one is a bit tricky.

For distros like Debian, running regular systemd/udev, there is `/usr/lib/udev/hwdb.bin` compiled from `/lib/udev/hwdb.d`.

See
https://www.freedesktop.org/software/systemd/man/hwdb.html
https://packages.debian.org/stable/amd64/udev/filelist

But there is none in OpenWrt, because `libudev` in OpenWrt is actually [`libudev-zero`](https://github.com/openwrt/packages/blob/master/libs/libudev-zero/Makefile) from https://github.com/illiliti/libudev-zero. It simply _forwards_ the information from what's inside the descriptor of USB devices, if I understand correctly.

Despite the fact that `libudev-zero` doesn't come with a hardware database, I personally don't see a point to keep `usbids` as a dependency. (Neither `libudev-zero` nor binary version of `lsusb` make use of it.)

CanadianJeff:
this has got to be the worst idea ever usb.ids provides much more detail about a device than udev garbage....

see #18148

PR #16462, opened 2021-08-25
python3-package.mk: add check for dependencies in host-pip-requirements

commodo:
Maintainer: me, @jefferyto 
Compile tested: x86 https://github.com/openwrt/openwrt/commit/c0afe3a5d2239fb40aa566e4d81666b4461f07fb
Run tested: n/a

It often happens that we update a package to a new version (e.g. cffi) to a
newer version, but we forget to update the version for cffi in the
`lang/python/host-pip-requirements/cffi.txt` file.

This check adds a minimal check, so that when a build occurs for a Python
package, if there is a mention/listing of this package in
`lang/python/host-pip-requirements/` it will check that the versions match.

This way, when we update a package, we get a build failure and update the
host version as well.

This will omit packages (like Cython) that are not packaged for OpenWrt,
but are host-side dependencies only.
But until we find some mechanism to check for those, we will probably only
notice to update them when another build occurs (at the very least).


jefferyto:
Looks okay to me, though I think compiling the django package will cause it to error on host-pip-requirements/Django-1.11.txt?

It would be nice if we can choose to bypass this check and pin a dependency to a certain version, but in this case I think it would be fine to move Django-1.11.txt into the seahub folder (and use a relative path for `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS`, e.g. `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=./host-pip-requirements.txt`) since nothing else uses this dependency.

commodo:
> I'm a little hesitant to split pytz out of Django-1.11.txt (and to make it a general rule to have one PyPI package per requirements file), because then the "top-level" (OpenWrt) package (in this case seahub) would need to know/list its direct dependencies (django 1.11) and all of its indirect dependencies (pytz). (The same would apply for cffi.txt.)
> 
> I realize this may lead to duplicate packages/hashes in multiple requirements files, but for now we don't have any duplicates and I hope we can replace this system of separate requirements files (and reuse data from OpenWrt package makefiles) if/when there are more host-side pip dependencies.
> 
> What do you think?

So, I don't have any ideas (or opinions) yet (for this).
[ I am a little tired, currently running on autopilot and it's only Monday :p ]

I guess, we can go with moving Django-1.11. txt in the Seahub package folder for now.
And we can see what the future holds.

Currently, pytz isn't used anywhere else (host-side) other than Django-1.11
I guess when Seahub gets updated, Django-1.11 would also get updated.

commodo:
Changelog v1 -> v2:
* moved Django-1.11.txt in Seahub package folder

jefferyto (approved):
Thanks for making the Django change :pray:

commodo:
> Thanks for making the Django change pray

you're welcome :)

will merge this in a couple of hours then if no objections

Issue #16256, opened 2021-07-30
acme: Restart uhttpd after renewing certificate


K2IE:
I'm not certain if the issue is in acme proper or in luci-app-acme, but whenever my certificate is renewed I have an issue.  uhttpd is not restarted as part of the process and so at some point my certificate will expire, I must then ssh in and manually restart uhttpd.

This ought to be triggered at the end of the renewal process if "Use for uhttpd" is checked in luci-app-acme.

CodeFetch:
You could introduce a generic configuration parameter in uacme for triggering a shell-command after renewal. Hard-coding this to uhttpd is noch a good idea. Instead you could define a generic trigger script which like "uacme_restart_httpd" and then let uhttpd/nginx/apache whatever provide that file.

scde:
Hi,

I have the same issue and think `uhttpd` is not restarted but only reloaded. See https://github.com/openwrt/packages/blob/b03fe54e0e988f62c33503cb20cdb30b366870ad/net/acme/files/run.sh#L144
I guess for an easy fix you could just replace `reload` with `restart` in that line. Since `nginx` restarts every time as well you might want to do it like that.

But if you want to prevent unwanted `uhttpd` restarts you would need to look a bit closer into it and only trigger if there was a successful renewal. Maybe by passing a return value to the `post_checks()` function and using it to branch between reload and restart.

Best regards
scde

Neustradamus:
@scde, @CodeFetch, @scde: I think that this issue can be closed:
- https://github.com/openwrt/packages/pull/16482
- https://github.com/openwrt/packages/pull/16545
- https://github.com/openwrt/packages/pull/16546

scde:
Hi @Neustradamus,
yeah. From my point of view you can close it. I didn't observe the need for a manual uhttpd restart after a cert update since this change. The script seems to work correctly now.
Since I didn't open the issue I can't close it, sorry.
BR scde

Neustradamus:
@scde: Thanks for your reply!

Hope that an admin or @K2IE can close it...

PR #16482, opened 2021-08-28
acme: Fix uhttpd restart to load new certificates (issue #16256)

scde:
Maintainer: @tohojo
Compile tested: No
Run tested: No

Description:
You have to restart `uhttpd` after cert updates for it to load them.
Right now I have to manually restart `uhttpd` after a cert expires and a reload won't work (see #16256). I am guessing because the process still has the old cert files loaded (although they are already replaced on the filesystem). You can check and see that `/etc/init.d/uhttpd reload` does not start a new process (no new PID for `uhttpd`).

Sorry for not really testing this but I am guessing this change might be small enough for it to not break anything major. Although if there are concerns about the script restarting `uhttpd` too often you might want to make a few more changes (see [my comment in the issue](https://github.com/openwrt/packages/issues/16256#issuecomment-903186029))

Best regards
scde

### DIFF
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -145,7 +145,7 @@ post_checks() {
 			UHTTPD_LISTEN_HTTP=
 		fi
 		uci commit uhttpd
-		/etc/init.d/uhttpd reload
+		/etc/init.d/uhttpd restart
 	fi
 
 	if [ -e /etc/init.d/nginx ] && { [ "$NGINX_WEBSERVER" -eq 1 ] || [ "$UPDATE_NGINX" -eq 1 ]; }; then


stangri:
As per https://openwrt.org/submitting-patches#sign_your_work, please use your real name in the signed off by.

scde:
Hi @stangri,
sorry for not reading the contribution info correctly. It should be fixed now.
BR Dennis

tohojo (approved):

tohojo (changes requested):
Ah wait; please bump the package revision (in Makefile) along with the change...

neheb:
@scde @tohojo misspoke. Undo your change.

scde:
Hi,
thanks for merging.
But one more question. I am not familiar with the release process. Will this be backported to 19.07 and 21.02? Or more broadly when and where will this land?
BR Dennis

neheb:
Propose new PRs targeting whichever version. For 19.07, PKG_RELEASE will need to be bumped.

PR #16483, opened 2021-08-28
update to the  master

openips:

Description:


BKPepe:
- Removed PR template, which results in:
Missing run tested, compile tested and maintainer
- In your commit, there is a missing correct subject prefix.
It should start with the package name.
- In your commit, there is a missing `Signed-off-by`.

TL;DR: No indication of run testing, formal issues, etc.

In this case, I am closing this pull request until you read [submitting patches in OpenWrt documentation](https://openwrt.org/submitting-patches) and follow it.

PR #16484, opened 2021-08-28
Update Makefile

openips:

Description:


BKPepe:
Same as https://github.com/openwrt/packages/pull/16483#issuecomment-907592043

PR #16488, opened 2021-08-29
json-glib: bump to version 1.6.6

micmac1:

Maintainer: me
Compile tested: ath79 master & 21.02
Run tested: 21.02

Description: version bump, tested with rtpengine, kamailio on router and asterisk on laptop.

Thanks!


micmac1:
Looks like Github Actions now works fine without 01-remove-python-check.patch, so removing it.

PR #16503, opened 2021-08-31
acme: add haproxy support

lialosiu:
Maintainer: @tohojo
Compile tested: (x86-64, x86, 21.02.0-rc4)
Run tested: (x86-64, x86, 21.02.0-rc4, tests done)

Description:

acme: add haproxy support.

If haproxy was installed, deploy cert file to haproxy and restart haproxy.

### DIFF
--- a/net/acme/files/acme.config
+++ b/net/acme/files/acme.config
@@ -9,6 +9,7 @@ config cert 'example_wildcard'
 	option keylength 2048
 	option update_uhttpd 1
 	option update_nginx 1
+	option update_haproxy 0
 	list domains example.org
 	list domains sub.example.org
 	list domains *.sub.example.org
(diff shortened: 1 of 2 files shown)


tohojo:
Please fix the commit message: the title should be 'acme: add haproxy support', and you need to write a description as well. Also, your signed-off-by should use your real name.

You don't need to close & reopen the pull request to do this, you can just force-push to the source branch, and the PR will be updated...

lialosiu:
> Please fix the commit message: the title should be 'acme: add haproxy support', and you need to write a description as well. Also, your signed-off-by should use your real name.
> 
> You don't need to close & reopen the pull request to do this, you can just force-push to the source branch, and the PR will be updated...

Sorry, my English so poor🤣
I am trying to fix this.

tohojo:
Alright, that's better as far as the commit message is concerned! 

One more thing: could you please add the new variable to the example config as well (just setting it to 0 is fine)? That way it'll be easier to discover for people; and if you could open a separate pull request for luci-app-acme adding it there as well, that would be great! :)

lialosiu:
new variable added to example config file now :)
I am never try to coding for luci app, so I may take some time to learning it, I will try my best. 😁

tohojo (approved):

PR #16504, opened 2021-08-31
htop: Add HTOP_LMSENSORS config option

champtar:
Maintainer: me
Compile tested: ath79 / openwrt/openwrt@25cb37bc009341da8afff0335cd7f9590a6bb411
Run tested: no

Description:
Enabled by default for x86, this enables lm-sensors support in htop.
Also add --enable-linux-affinity to avoid autodetecting it

### DIFF
--- a/admin/htop/Makefile
+++ b/admin/htop/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=htop
 PKG_VERSION:=3.0.5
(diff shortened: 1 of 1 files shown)


champtar:
Unrelated failure: `perlbase-i18n has a NULL tmp_unpack_dir`

champtar:
Fix for the unrelated http://lists.openwrt.org/pipermail/openwrt-devel/2021-September/036286.html

PR #16516, opened 2021-09-01
treewide: Remove GO_PKG_LDFLAGS for stripping binaries

jefferyto:
Maintainer: @BKPepe (dnscrypt-proxy2), @1715173329 (dnslookup, dnsproxy, v2raya, xray-core, yq), @liudf0716 & @expiron (kcptun), @rs (nextdns), @gekmihesg (restic, restic-rest-server)
Compile tested: armvirt-32, 2021-08-29 snapshot sdk
Run tested: none

Description:
The `"-s -w"` flags in `GO_PKG_LDFLAGS` tells the Go compiler to strip the binaries it produces. Since the default Go package build process will strip binaries when `CONFIG_USE_STRIP` or `CONFIG_USE_SSTRIP` are selected, these flags are unnecessary.

When `CONFIG_NO_STRIP` is selected, these flags override the user's intention of building unstripped packages.

This removes these flags for all relevant packages.

### DIFF
--- a/net/dnscrypt-proxy2/Makefile
+++ b/net/dnscrypt-proxy2/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnscrypt-proxy2
 PKG_VERSION:=2.1.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=dnscrypt-proxy-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/DNSCrypt/dnscrypt-proxy/tar.gz/$(PKG_VERSION)?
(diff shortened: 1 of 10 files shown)


1715173329 (approved):

rs:
Just to make sure I understand, the default behavior is to strip or packages need to set `CONFIG_USE_STRIP` or `CONFIG_USE_SSTRIP` explicitly?

jefferyto:
@rs Those are config options for the build system and are not specific to any one package. The [default](https://github.com/openwrt/openwrt/blob/9e9057f4164a31bbe5a645878a9e0f083b533c0f/config/Config-build.in#L163-L165) is to strip binaries.

rs:
@jefferyto got it thanks. LGTM

jefferyto:
CI is failing because of the run-time test for dnslookup - I believe it will need to be upgraded to 1.4.9 to be compatible with Go 1.17.

1715173329:
> CI is failing because of the run-time test for dnslookup - I believe it will need to be upgraded to 1.4.9 to be compatible with Go 1.17.

Updated in #16519

PR #16520, opened 2021-09-02
boost: Bump to version 1.77.0

ClaymorePT:
Maintainer:  @ClaymorePT
Compile tested: BCM27xx on Master branch
Run tested: N/A

Description:
This commit updates boost to version 1.77.0

There are two new libraries in this version:
* [Describe](https://www.boost.org/libs/describe/): A C++14 reflection library, from Peter Dimov. Provides macros for describing enumerators and struct/class members, and primitives for querying this information.
* [Lambda2](https://www.boost.org/libs/lambda2/): A C++14, dependency-free, single header lambda library, from Peter Dimov. Allows simple function objects to be constructed via expressions such as: `_1 + 5, _1 % 2 == 0, _1 > _2, or _1 == ' ' || _1 == '\t'.`

More info about Boost 1.77.0 can be found at the [usual place](https://www.boost.org/users/history/version_1_77_0.html).


neheb:
2 failures that I see from this:

https://downloads.openwrt.org/snapshots/faillogs/mips64_octeonplus/packages/libtorrent-rasterbar/compile.txt
https://downloads.openwrt.org/snapshots/faillogs/mips64_octeonplus/packages/kea/compile.txt

ClaymorePT:
@neheb hey, sorry for the delay.
Both situations seem to be a fault of those packages sources. They need to be fixed on their side :/

neheb:
Yeah I got patches for them.

ClaymorePT:
Ah! I noticed that you have already created some PRs for those situations.

https://github.com/openwrt/packages/pull/16531
https://github.com/openwrt/packages/pull/16530

neheb:
Interesting failure: https://downloads.openwrt.org/snapshots/faillogs/arm_xscale/packages/boost/compile.txt

ClaymorePT:
@neheb There seems to be a fix for this upstream. Just not in 1.77. https://github.com/boostorg/math/pull/684

I'll add it ASAP.

ClaymorePT:
@neheb Can you provide me a target which uses arm_scale? I was unable to find it in the target menu using a fresh download of openwrt/master.

neheb:
Should be Marvell Kirkwood.

neheb:
One other thing I've noticed (problem has been in multiple versions) is:

```
ldd staging_dir/hostpkg/lib/libboost_filesystem.so.1.77.0
        linux-vdso.so.1 (0x00007ffdf25e5000)
        libboost_atomic.so.1.77.0 => not found
```

That's the only so with not found. Totally unused but still interesting.

ClaymorePT:
Hmm, I need to check that then. I know that I caught a similar issue during
the update and because of that, it was necessary to add another dependency
to one of the libs.

Thanks for the info. Updating this ASAP

A terça, 7/09/2021, 23:41, Rosen Penev ***@***.***> escreveu:

ClaymorePT:
@neheb 

> edit: oh libboost_regex.so links to the OS icu libraries as well...

Are you saying that it shouldn't ?

neheb:
AFAIK icu/host installs those libraries in hostpkg

ClaymorePT:
@neheb I think I understand what you meant. Regex links against libicu if it the boost build system detects icu from host instead of using the one for the target system. This may get tricky but I need to check this out.
In the mean time, I've already fixed the math issue (just a new patch).

In the boost update to 1.77, I've fixed the filesystem linkage issue with atomic. It seems that boost.Filesystem now depends on Boost.Atomic.

ClaymorePT:
@neheb  -> https://github.com/openwrt/packages/pull/16583

I've not included changes for the icu issue, but we can still discuss those and I can add them tomorrow.

PR #16524, opened 2021-09-02
[21.02] hplip: add a patch to fix PIE builds

paper42:
Maintainer: @luizluca

Description: PIE breaks the build:
```
ld: hp-hp.o: relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol `stderr' which may bind externally can not be used when making a shared object; recompile with -fPIC
ld: hp-hp.o(.text+0xb8): unresolvable R_AARCH64_ADR_PREL_PG_HI21 relocation against symbol `stderr'
ld: final link failed: bad value
```


champtar:
See my comment in the other PR

luizluca:
This should wait until the issue is fixed in master (#16523)

PR #16529, opened 2021-09-04
smartdns: update to version 1.2021.35

pymumu:
Compile tested: MIPS/ATH79
Run tested: master


pymumu:
```shell
*** No rule to make target '/home/build/openwrt/build_dir/target-x86_64_musl/linux-x86_64/cryptodev-linux-cryptodev-linux-1.11/ioctl.o', needed by '/home/build/openwrt/build_dir/target-x86_64_musl/linux-x86_64/cryptodev-linux-cryptodev-linux-1.11/cryptodev.o'.  Stop.
```
Compilation fails, the error message is as above, nothing to do with smartdns, it may be a problem with the compilation environment I think.

PR #16530, opened 2021-09-05
libtorrent-rasterbar: update to 2.0.4

neheb:
Fixes compilation with boost 1.77

Maintainer: @yangfl 
Compile tested: ath79


neheb:
Interesting that the patch is needed only on 64-bit architectures...

yangfl:
crazy, but... lgtm

PR #16533, opened 2021-09-06
libzip: fix compilation without SSL

neheb:
Linker error. No idea why.

Maintainer: @mhei 
Compile tested: ath79

In my local environment, something changed and now linking is more strict. I have no idea why.


mhei:
Guess due to newer cmake version...
Patch looks good to me, though. Do you care to open issue/PR in upstream?

neheb:
https://github.com/nih-at/libzip/pull/257

mhei:
Thanks!

PR #16538, opened 2021-09-06
unbound: backport fix for permission denied error

Ansuel:
Currently there is a problem with log spam when ipv6 network
is dropped. Fix this by backporting a patch to silence these errors
when verbose logging is not enabled.

Signed-off-by: Ansuel Smith <ansuelsmth@gmail.com>


Ansuel:
Any idea why the ci failed? Can't find the error... I local tested the patch and it does work correctly.

champtar:
The error is `Dirty patches detected, please refresh and review the diff`, likely line changed
`make package/unbound/refresh` should do the trick

Ansuel:
Aside from x86_64 complaining about cryptodev broken this is ready and tested

BKPepe:
cc: @EricLuehrsen as he is maintainer of the unbound

EricLuehrsen:
Looks good. Also I think backport to 21.02 branch.

champtar:
@Ansuel can you open a PR for 21.02 ?

Ansuel:
@champtar i notice that in 21.02 we have a old version and we are missing a commit. I backported all the 3 commits and open https://github.com/openwrt/packages/pull/16582

Issue #15096, opened 2021-03-11
podman: issues running applications in containers from non-root accounts


flyn-org:
Maintainer: @dangowrt
Environment: x86_64 master

Description:
The podman utility is meant to allow non-root users to run applications in containers. This bug is meant to track things preventing this from working on OpenWrt. (Note there is also ongoing work on uxc.)

1. The default OpenWrt kernel does not contain support for EXT4 security labels (EXT4_FS_SECURITY), and this seems to cause podman to fail under some circumstances. I reported this upstream, and upstream merged a fix. See issue #[9687](https://github.com/containers/podman/issues/9687) and pull request #[851](https://github.com/containers/storage/pull/851). For reference, here is the corresponding error:
```
Error: error creating container storage: error creating an ID-mapped copy of layer "[hash]": exit status 1: error during chown: storage-chown-by-maps: lgetxattr bin: operation not supported
```

2. /etc/containers/* is not readable by anyone but root. Fedora lets non-root users read files in this directory, so we could probably do the same on OpenWrt.

3. Non-root users cannot write to /sys/fs/cgroup/*. I am not sure how to safely handle this, and I have not yet figured out how other distributions do it. I am still trying to find information about how this is done.

4. Running `podman run ...` wants to mount /proc and so on in the container. This fails when run as non-root with:
```
mounting "/proc" to rootfs at "/proc" caused: operation not permitted
```
This might be related to the user namespaces, but I am not yet sure of this. I have installed shadow-newuidmap and -newgidmap, and I think I have set /etc/subid and /etc/subgid properly.

For reference, I found the following articles helpful for understanding how to build namespaces on Linux:

https://medium.com/@teddyking/linux-namespaces-850489d3ccf
https://medium.com/@teddyking/namespaces-in-go-basics-e3f0fc1ff69a
https://medium.com/@teddyking/namespaces-in-go-user-a54ef9476f2a
https://medium.com/@teddyking/namespaces-in-go-reexec-3d1295b91af8
https://medium.com/@teddyking/namespaces-in-go-mount-e4c04fe9fb29
https://medium.com/@teddyking/namespaces-in-go-network-fdcf63e76100
https://medium.com/@teddyking/namespaces-in-go-uts-d47aebcdf00e

dangowrt:
I'm working on updating to podman-3.0.1 here: https://github.com/openwrt/packages/pull/15264

Regarding the issues you discovered when using rootless containers:
[...]

PR #16547, opened 2021-09-07
podman: update to 3.3.1

flyn-org:
Earlier versions of podman did not make use of TMPDIR when running "podman run ...". Podman's default, /var/tmp, presents a problem to rootless use since OpenWrt's /var/tmp does not permit writes by non-root users. Podman 3.3.1 makes full use of TMPDIR.

This is part of an attempt to get rootless podman to work on OpenWrt. See https://github.com/openwrt/packages/issues/15096.

See also the upstream issue at https://github.com/containers/podman/issues/10698.

Maintainer:  @oskarirauta
Compile tested: x86_64
Run tested: x86_64

Description:


dangowrt (approved):

flyn-org:
@neheb, can you merge?

PR #16550, opened 2021-09-07
spidev-lib: add lib to access spidev devices

PolynomialDivision:
This package provides a C wrapper to linux user space spidev.

Maintainer: me
Compile tested: openwrt-trunk
Run tested: BPI-R64

### DIFF
--- a/libs/spidev-lib/Makefile
+++ b/libs/spidev-lib/Makefile
@@ -1,0 +1,39 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=spidev-lib
+PKG_SOURCE_DATE:=2014-10-31
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/milekium/spidev-lib.git
+PKG_SOURCE_VERSION:=10fc6f4f900104ddb5fbea1d442842f448c9fd75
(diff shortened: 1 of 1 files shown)


karlp:
Is there really nothing more up to date than a 2014 single shot package that builds a static library?  What's gonna use this?

karlp:
No, I actually think most people using spidev will either be using higher level languages or using it directly.  there's a pile of "example" spidev code around, without any "library" at all.  This just really feels like something that's adding some indirection, but as it's GPL and intended for static linking, people should just add these two files into their own projects themselves, it seems really irrelevant to be packaging it.

champtar:
@PolynomialDivision do you have any packages depending on that / related PR ?

PolynomialDivision:
Yeah. I write some library for some UWB chip. Initially it used `wiringpi` but we rewrite the driver to work with spidev. I thought it would be a good idea to support such a spidev wrapper officially. However, I'm also fine with just going for a private repositroy. I could also make the lib shared or fork it to go further with the lib. I'm not sure what is the best idea.

champtar:
Packaging a lib without a public package using it is a bit wasting buildbot ressources. And as @karlp as it's statically linked it's simpler for project to just 'vendor' the files. Once we get a user I'm happy merging a copy of this PR.

PolynomialDivision:
Okay. I will close. Thanks for feedback. :)

Issue #16548, opened 2021-09-07
grep: /usr/libexec/{egrep-gnu,fgrep-gnu} uses #!/bin/bash as interpreter


BlessDeix92:
Maintainer: Julen Landa Alustiza <julen@zokormazo.info>
Environment: package 3.6, OpenWRT 21.02.0

Description: /usr/libexec/{egrep-gnu,fgrep-gnu} improperly uses #!/bin/bash as interpreter. Should be using /bin/sh or make bash a dependency.

This problem was in RC1 too.

BlessDeix92:
Why is this still not actually fixed?

```
root@openwrt:~# ls /bin/*grep*
lrwxrwxrwx 1 root root 22 Mar 22 15:09 /bin/egrep -> /usr/libexec/egrep-gnu
lrwxrwxrwx 1 root root 22 Mar 22 15:09 /bin/fgrep -> /usr/libexec/fgrep-gnu
lrwxrwxrwx 1 root root 21 Mar 22 15:09 /bin/grep -> /usr/libexec/grep-gnu
root@openwrt:~# cat /bin/egrep 
#!/bin/bash
cmd=${0##*/}
echo "$cmd: warning: $cmd is obsolescent; using grep -E" >&2
exec grep -E "$@"
root@openwrt:~# 

```

PR #16551, opened 2021-09-07
grep: use posix shebang for egrep/fgrep

vgaetera:
Maintainer: @Zokormazo
Compile tested: -
Run tested: -
Related issue: https://github.com/openwrt/packages/issues/16548

Description:
Use POSIX shebang for egrep/fgrep instead of the variable/Bash.
The issue affects both master and openwrt-21.02.

Signed-off-by: Vladislav Grigoryev <vg.aetera@gmail.com>

### DIFF
--- a/utils/grep/Makefile
+++ b/utils/grep/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=grep
 PKG_VERSION:=3.7
-PKG_RELEASE:=1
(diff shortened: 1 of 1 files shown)


champtar:
you likely can pass the shell as configure option instead of patching the source

zhanhb:
It seems `MAKE_FLAGS += SHELL="/bin/sh"` should be declared after the inclusion of package.mk. For it will be overwritten otherwise.

PR #16452, opened 2021-08-24
python-simplejson: bump to version 3.17.5

commodo:
Maintainer: me
Compile tested: x86 https://github.com/openwrt/openwrt/commit/c0afe3a5d2239fb40aa566e4d81666b4461f07fb
Run tested: x86 https://github.com/openwrt/openwrt/commit/c0afe3a5d2239fb40aa566e4d81666b4461f07fb

Signed-off-by: Alexandru Ardelean <ardeleanalex@gmail.com>

PR #15098, opened 2021-03-11
fail2ban: initial package for fail2ban v0.11.2

erdoukki:
Maintainer: Kerma Gérald gandalf@gk2.net
Compile tested: (aarch64_cortex-a53, espressobin board, OpenWrt master)
Run tested: (aarch64_cortex-a53, espressobin board, OpenWrt master)

Description:
python3 package of pyinotify as requirement of fail2ban package
and
python3 fail2ban package

tested on snapshot : OK
```
root@RELAY:/# fail2ban-python --version                                                                                                                            
Python 3.9.2                                                                                                                                                       
root@RELAY:/# fail2ban-server --version                                                                                                                            
Fail2Ban v0.11.2                                                                                                                                                   
root@RELAY:/# uname -ar                                                                                                                                            
Linux RELAY 5.4.102 #0 SMP Sat Mar 13 13:41:04 2021 aarch64 GNU/Linux                                                                                              
root@RELAY:/# cat /etc/openwrt_release                                                                                                                             
DISTRIB_ID='OpenWrt'                                                                                                                                               
DISTRIB_RELEASE='SNAPSHOT'                                                                                                                                         
DISTRIB_REVISION='r16175-997ff740dc'                                                                                                                               
DISTRIB_TARGET='mvebu/cortexa53'                                                                                                                                   
DISTRIB_ARCH='aarch64_cortex-a53'                                                                                                                                  
DISTRIB_DESCRIPTION='OpenWrt SNAPSHOT r16175-997ff740dc'                                                                                                           
DISTRIB_TAINTS=''                                                                                                                                                  
```
Fail2ban scans log files (e.g. /var/log/apache/error_log) and bans IPs that show the malicious signs -- too many password failures, seeking for exploits, etc. Generally Fail2Ban is then used to update firewall rules to reject the IP addresses for a specified amount of time, although any arbitrary other action (e.g. sending an email) could also be configured. Out of the box Fail2Ban comes with filters for various services (apache, courier, ssh, etc). 
More @ http://www.fail2ban.org

Use case (POC):
Fail2Ban, with the help of centralized logs directly on a main router OpenWrt based, can ban unwanted access by their IP source.

PR #16552, opened 2021-09-07
python-awesomeversion: add it to the repository

BKPepe:
Maintainer: me
Compile tested: Turris Omnia, mvebu, cortexa9, OpenWrt 21.02
Run tested: Turris Omnia, mvebu, cortexa9, OpenWrt 21.02

Description:
- Add [python-awesomeversion](https://pypi.org/project/awesomeversion/)
It is one dependency for Home Assistant

### DIFF
--- a/lang/python/python-awesomeversion/Makefile
+++ b/lang/python/python-awesomeversion/Makefile
@@ -1,0 +1,42 @@
+#
+# Copyright (C) 2021 CZ.NIC, z. s. p. o. (https://www.nic.cz/)
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=python-awesomeversion
+PKG_VERSION:=21.8.1
+PKG_RELEASE:=$(AUTORELEASE)
+
+PYPI_NAME:=awesomeversion
+PKG_HASH:=a72f3fff12df257e30f5e25a8550fc4da62e7591d5fdb70714f5c96827c37645
+
+PKG_MAINTAINER:=Josef Schlehofer <josef.schlehofer@nic.cz>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENCE.md
+
+include ../pypi.mk
+include $(INCLUDE_DIR)/package.mk
+include ../python3-package.mk
+
+define Package/python3-awesomeversion
+  SECTION:=lang
+  CATEGORY:=Languages
+  SUBMENU:=Python
+  TITLE:=AwesomeVersion
+  URL:=https://github.com/ludeeus/awesomeversion
+  DEPENDS:= \
+      +python3-light \
+      +python3-logging
+endef
+
+define Package/python3-awesomeversion/description
+  Make anything a version object, and compare against a vast selection of other version formats.
+endef
+
+$(eval $(call Py3Package,python3-awesomeversion))
+$(eval $(call BuildPackage,python3-awesomeversion))
+$(eval $(call BuildPackage,python3-awesomeversion-src))


lang/python/python-awesomeversion/Makefile
```diff
@@ -0,0 +1,42 @@
...
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=python-awesomeversion
+PKG_VERSION:=21.8.1
+PKG_RELEASE:=1
```
commodo, line 12: should we use $(AUTORELEASE) here?
i'm not insisting on it, but i started switching some of my packages to it;
and i have no idea how good the mechanism is, but it's worth using it until it perfects itself;

lang/python/python-awesomeversion/Makefile
```diff
@@ -0,0 +1,42 @@
...
+      +python3-logging
+endef
+
+define Package/python3-awesomeversion/description
+  One version package to rule them all, One version package to find them, One version package to bring them all, and in the darkness bind them.
+endef
+
+$(eval $(call Py3Package,python3-awesomeversion))
+$(eval $(call BuildPackage,python3-awesomeversion))
+$(eval $(call BuildPackage,python3-awesomeversion))
```
commodo, line 42: duplicate `$(eval $(call BuildPackage,python3-awesomeversion))`
i think this was supposed to be `python3-awesomeversion-src`

commodo:
rest looks good to me;

lang/python/python-awesomeversion/Makefile
BKPepe: I wonder from which package I took it. :shrug: 
Thank you for noticing! Fixed.

lang/python/python-awesomeversion/Makefile
BKPepe: Yeah, sure! I will need to switch all my current maintained packages, which I maintain to it. Sometimes it is very handy.

commodo (approved):

lang/python/python-awesomeversion/Makefile
```diff
@@ -0,0 +1,42 @@
...
+  SUBMENU:=Python
+  TITLE:=AwesomeVersion
+  URL:=https://github.com/ludeeus/awesomeversion
+  DEPENDS:= \
+      +python3-light \
+      +python3-logging
+endef
+
+define Package/python3-awesomeversion/description
+  One version package to rule them all, One version package to find them, One version package to bring them all, and in the darkness bind them.
```
jefferyto, line 37:
```suggestion
  Make anything a version object, and compare against a vast selection of other version formats.
```

lang/python/python-awesomeversion/Makefile
```diff
@@ -0,0 +1,42 @@
...
+PKG_NAME:=python-awesomeversion
+PKG_VERSION:=21.8.1
+PKG_RELEASE:=$(AUTORELEASE)
+
+PYPI_NAME:=awesomeversion
+PKG_HASH:=a72f3fff12df257e30f5e25a8550fc4da62e7591d5fdb70714f5c96827c37645
+
+PKG_MAINTAINER:=Josef Schlehofer <josef.schlehofer@nic.cz>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE.md
```
jefferyto, line 19:
```suggestion
PKG_LICENSE_FILES:=LICENCE.md
```

commodo: wow, that's a good eye
any way to automate this check through CI?

or, is there a CI check for this?
pinging a few people on this question (for ideas)   @neheb @hnyman 

i feel like we may have missed a few along the way;

neheb: @aparcar is the CI person.

aparcar: Any ideas how to implement this? Wondering if this could become part of package/foobar/check? We'd need to unpack the downloaded file and see if the license exists in BUILD_DIR

commodo: > Any ideas how to implement this? Wondering if this could become part of package/foobar/check? We'd need to unpack the downloaded file and see if the license exists in BUILD_DIR

yes, i think that would be the approach to do it;
so, something like
```
foreach(license_files in $(PKG_LICENSE_FILES)) {
    if (not file.exists(PKG_BUILD_DIR +  license_file)
         fail_build
}
```

jefferyto: If you want to go crazy you can also use a tool like [identify](https://github.com/pre-commit/identify#identifying-license-files) or [lc](https://github.com/boyter/lc/) to verify the SPDX ids in `PKG_LICENSE` against the files in `PKG_LICENSE_FILES`.

aparcar: I'm currently stuck on how to figure out `PKG_BUILD_DIR` from the package name. So ideally something like `package/<foobar>/dumpinfo` would print such information. It's possible to use `make -C package/utils/<foobar>/ va.PKG_BUILD_DIR` however I don't have the full path, just the name...

commodo:
will merge this tomorrow if no objections;

jefferyto (approved):
Thanks!

aparcar:
I wonder why x86 fails...?

jefferyto:
@aparcar cryptodev-linux is failing to build in the sdk: https://github.com/openwrt/packages/pull/16452#issuecomment-904669067 https://github.com/openwrt/packages/pull/15098#issuecomment-909010215

aparcar:
> @aparcar cryptodev-linux is failing to build in the sdk: #16452 (comment) #15098 (comment)

I remember that being an issue in the past but it went silent for a while. Do you have any clues what could solve this?

commodo:
> > @aparcar cryptodev-linux is failing to build in the sdk: #16452 (comment) #15098 (comment)
> 
> I remember that being an issue in the past but it went silent for a while. Do you have any clues what could solve this?

not without investigating unfortunately;
it looks like it's in the base/core packages layer;

commodo:
> > > @aparcar cryptodev-linux is failing to build in the sdk: #16452 (comment) #15098 (comment)
> > 
> > 
> > I remember that being an issue in the past but it went silent for a while. Do you have any clues what could solve this?
> 
> not without investigating unfortunately;
> it looks like it's in the base/core packages layer;

i took some time to check this;
it still reproduces if i download the x86_64 SDK and do  `./scripts/feeds update -a ; ./scripts/feeds install -a`  && `make package/python-awesomeversion/compile V=s`

so, this is clearly an issue with the SDK;
i don't have a clear idea on the matter;
the `cryptodev-linux` package is an out-of-tree kmod;
looking into the `openwrt-sdk-x86-64_gcc-10.3.0_musl.Linux-x86_64/build_dir/target-x86_64_musl/linux-x86_64/linux-5.10.64 ` it looks like some kmods are shipped with the SDK;
in the SDK context, it seems that OpenSSL may select this cryptodev-linux package via `PACKAGE_libopenssl-devcrypto [=m] && !OPENSSL_ENGINE_BUILTIN [=n] && OPENSSL_ENGINE [=y] && PACKAGE_libopenssl [=m]`
doing the same in the OpenWrt directory works fine;

i wonder if the Linux directory (in the SDK) may be missing some Kbuild stuff for this kmod;
in the list of kmods in the Cryptographic API sub-menu, this seems to be the only kmod showing up;
or whether the SDK should ship this kmod?

i'd ping  @Ansuel to take over this issue, since he owns cryptodev-linux;

PR #16565, opened 2021-09-08
haproxy: Update HAProxy to v2.4.4

gladiac1337:
Maintainer: Thomas Heil <heil@terminal-consulting.de>, Christian Lachner <gladiac@gmail.com> (/me)
Run tested: ipq806x (r7800)

Description: Update HAProxy to v2.4.4
- **This update fixes CVE-2021-40346**; see: [JFrog: Critical Vulnerability in HAProxy (CVE-2021-40346): Integer Overflow Enables HTTP Smuggling](https://jfrog.com/blog/critical-vulnerability-in-haproxy-cve-2021-40346-integer-overflow-enables-http-smuggling/)
- Make use of $(AUTORELEASE) for PKG_RELEASE

### DIFF
--- a/net/haproxy/Makefile
+++ b/net/haproxy/Makefile
@@ -10,12 +10,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=haproxy
-PKG_VERSION:=2.4.3
-PKG_RELEASE:=1
+PKG_VERSION:=2.4.4
+PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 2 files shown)


paper42:
There is a linux-musl target since 2.2, it seems like it's the same as linux-glibc except it doesn't define USE_BACKTRACE. I think switching to that target would be nice for future and shouldn't break anything right now.

gladiac1337:
> There is a linux-musl target since 2.2, it seems like it's the same as linux-glibc except it doesn't define USE_BACKTRACE. I think switching to that target would be nice for future and shouldn't break anything right now.

Yes, they added that target a while ago. I will look into it with the next release. Thanks!

PR #16571, opened 2021-09-08
python3: bump version to 3.9.7

commodo:
Maintainer: me, @jefferyto 
Compile tested: x86 https://github.com/openwrt/openwrt/commit/de0c380a5f8289839ab970e794a45f0e04a466a3
Run tested: x86 https://github.com/openwrt/openwrt/commit/de0c380a5f8289839ab970e794a45f0e04a466a3

Refreshed patches.
And apply hack for line-endings in pep517 (from pip).

Hack comment:
```
  # FIXME: [1] get rid of this asap; 'patch' doesn't like Windows endings, and this file is full of them...
  #        I actually tried this in a number of ways and the only way to fix this is to implement
  #        a poor-man's dos2unix using sed.
  #        The issue is with the pip package; it seems that it throws in some Windows line-endings
  #        and 'patch' won't handle them. So, we do a "dos2unix" and then patch.
  #        We can get rid of this once this is solved upstream and in pip:
  #            https://github.com/pypa/pep517/pull/130
```


commodo:
and i thought i'd seen everything until this hack......

jefferyto (approved):
Thanks!

jefferyto:
If you remade the pep517 patch with Windows line endings, would `patch` accept it?

commodo:
> If you remade the pep517 patch with Windows line endings, would `patch` accept it?

i tried a number of combinations including the one you mentioned;
(which is why it took me some time to submit the PR)
for one thing,`git format-patch` did not seem to generate a patch with Windows line-endings;

for kicks i also tried a unix2dos conversion to the patch file and got this:
```
Applying ./patches-pip/001-pep517-pyc-fix.patch using plaintext: 
(Stripping trailing CRs from patch; use --binary to disable.)
patching file pip/_vendor/pep517/in_process/__init__.py
Hunk #1 FAILED at 10 (different line endings).
```
( `Hunk #1 FAILED at 10 (different line endings).`  is what i get for the patch without converting the file prior to patching )

when i googled this issue with `patch` and windows line-endings, it seems that i am not alone in this;
multiple people have faced issues with this;
and have solved it with dos2unix / unix2dos

what i am not clear btw is why these line-endings get changed;
they don't seem to appear in the git-repo of pip or pep517;
they show up in the released files;
i found reports from 3 years ago mentioning that there might be some voodoo magic with CRLF :   https://github.com/pypa/pipenv/issues/2061  https://github.com/pypa/pipenv/pull/2069  https://github.com/pypa/pip/issues/5383
maybe some of these are still valid today
or it may be some new bug?
well, best is to try to send them all patches we have so that we don't keep anything on our side and let them have the "fun" or CRLF patching

jefferyto:
I'm okay to merge this PR as-is, though I might try redoing the patch later if I have time (by redoing I mean editing the files again and using `diff` manually, not using tools like git or converting the patch file).

(I doubt upstream will merge the patch, not just because .py files are normally present in a Python installation, but because bytecode files are normally saved in a `__pycache__` directory and with extensions like `.cpython-39.pyc`. Our patch assumes the simpler `.pyc` bytecode naming/placement.)

commodo:
> I'm okay to merge this PR as-is, though I might try redoing the patch later if I have time (by redoing I mean editing the files again and using `diff` manually, not using tools like git or converting the patch file).

i think the more solid approach would be to make the `patch` command/syntax in the build to be a bit more robust;
maybe use the `--binary` option;
but, i don't want to do it just yet; it's a single case

> (I doubt upstream will merge the patch, not just because .py files are normally present in a Python installation, but because bytecode files are normally saved in a `__pycache__` directory and with extensions like `.cpython-39.pyc`. Our patch assumes the simpler `.pyc` bytecode naming/placement.)

whatever the case with upstream, hopefully we can resolve a discussion with them that fixes the issue for us as well;
it's how i got rid of a few patches in some of the packages in this feed;

PR #16584, opened 2021-09-08
vpn-policy-routing: update to 0.3.5-3

stangri:
Maintainer: me
Compile tested: x86_64, Sophos SG-105, 21.02.0

Description:
improvement: reintroduce vpnbypass conflict to Makefile
improvement: rename firewall hotplug script
bugfix: create iptables rules for tables when IPv6 is enabled
improvement: convert dnsmasqIpsetSupported from true/false to 1/0
bugfix: completely destroy old tables before creating new ones
improvement: better looks for default routing table entries
bugfix: use hash:ip ipsets instead of hash:net
bugfix: create interface ipsets/rules for user scritps when no policies require them
bugfix: force full restart when gateway information is missing
bugfix: insert kernel mods & calculate dnsmasq file hash only on full restart
bugfix: don't use hex padding for ifaceMark


net/vpn-policy-routing/Makefile
```diff
@@ -5,7 +5,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=vpn-policy-routing
 PKG_VERSION:=0.3.5
-PKG_RELEASE:=2
+PKG_RELEASE:=3
```
champtar, line 8: why not increase PKG_VERSION ?

PR #16610, opened 2021-09-10
prometheus-node-exporter-lua-nat_traffic: Improve node_nat_traffic prometheus metric.

cconstantine:
Maintainer: @mweinelt 
Compile tested: N/A, lua-only changes
Run tested: RaspberryPI CM4, OpenWrt 21.02-SNAPSHOT r15935-2db7fe792e

Description:
This change does 3 things:
 1) For each connection output 2 metrics; a `transmit` and a `receive` bytes counter.
 2) Update the node_nat_traffic metric type from `gauge` to `counter`.  The total number of bytes transferred is a natural counter.  See: https://prometheus.io/docs/concepts/metric_types/
 3) Adds the source and destination ports for each connection.  This ensures that when a src has multiple connections to a dest we don't get multiple lines with the same label values.


PolynomialDivision:
Please use only `prometheus-node-exporter-lua:` as commit prefix. :)

@champtar I think this is your plugin? Could you review?

champtar:
Cardinality would be too big IMO, it would be better to have an exporter for https://github.com/jow-/nlbwmon

PR #16628, opened 2021-09-13
[WIP] uradvd: add new package

mwarning:
Send IPv6 router advertisments.

Signed-off-by: Moritz Warning <moritzwarning@web.de>

Description:


net/uradvd/Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/uradvd
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Advertise an IPv6 prefix/route
+  DEPENDS:=+librt
```
champtar, line 17: I'm not sure you need explicit depency on librt (it added automatically)

mwarning:
@NeoRaider, @christf since I use this for other projects as well. It would make sense to send this upstream. Instead of having to copy it around.

I am not sure about the naming of options in /etc/config/uradvd though.

net/uradvd/Makefile
```diff
@@ -0,0 +1,42 @@
...
+endef
+
+define Package/uradvd/description
+	Advertise an IPv6 prefix from the node
+endef
+
+define Build/Prepare
+	mkdir -p $(PKG_BUILD_DIR)
+	$(CP) ./src/* $(PKG_BUILD_DIR)/
```
champtar, line 26: The source is pretty small and likely not to change too much, but is it available online ?
Maybe at least we should put a link to it.

net/uradvd/Makefile
NeoRaider: Yeah, the Makefile is ancient and this can be removed - it might have been necessary with uClibc in the past.

net/uradvd/Makefile
NeoRaider: At the moment the primary source is in Gluon's OpenWrt feed at https://github.com/freifunk-gluon/packages/tree/master/net/uradvd. I see two options here:
- We drop the package from our feed, making openwrt/packages the primary source
- We create a separate repository for the source in the freifunk-gluon Github org

The latter is probably preferable. I can take care of creating the repo if we want to go with that.

net/uradvd/files/etc/config/uradvd
```diff
@@ -0,0 +1,13 @@
...
+# simple IPv6 SLAAC advertisement
+config interface
+	# Interface to advertise on
+	option ifname 'br-lan'
+	# default lifetime in seconds (0: forever)
+	option lifetime 65000
+	# prefix on link
+	list prefix_p '300::/64'
+	# prefix not on link
+	#list prefix_a '200::/64'
```
NeoRaider, line 11: How about the following option names?
- `prefix_on_link`
- `prefix_off_link`
- `default_lifetime`
The "default" in that option name is very important - this is the lifetime of the default route!

net/uradvd/files/etc/config/uradvd
```diff
@@ -0,0 +1,13 @@
+
+# simple IPv6 SLAAC advertisement
+config interface
+	# Interface to advertise on
+	option ifname 'br-lan'
+	# default lifetime in seconds (0: forever)
```
NeoRaider, line 6: The comment is wrong. `0` means a lifetime of 0, meaning that no default route is announced.

net/uradvd/files/etc/config/uradvd
NeoRaider: For Gluon is would be nice to have something like `option enabled '0'` in the example config, so the initscript doesn't do anything by default.

net/uradvd/Makefile
mwarning: Sounds good to me. Please provide a link to get the source code from. :)

NeoRaider: I've pushed the code to https://github.com/freifunk-gluon/uradvd with minor Makefile improvements and a LICENSE file that can be referenced from the OpenWrt Makefile.

I don't think it makes sense to start versioning the source, so I propose we remove `PKG_VERSION` from the Makefile and use `PKG_SOURCE_DATE`-based package versions, as in https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob;f=package/network/config/netifd/Makefile;h=c33e0c2134d6f53bd8bbfff66d58c605763e5fc8;hb=HEAD

net/uradvd/files/etc/init.d/uradvd
```diff
@@ -0,0 +1,64 @@
...
+	config_get device $cfg 'device'
+	config_get interface $cfg 'interface'
+	config_get default_lifetime $cfg 'default_lifetime'
+
+	if [ "$enabled" != "1" ]; then
+		echo "Disabled in config."
+		exit 0
+	fi
+
+	if [ -n "$device" -a -n "$interface" ]; then
```
neheb, line 20: shellcheck will warn here. Replace with [] && []

net/uradvd/Makefile
mwarning: Please have a look. Do you want to be the maintainer? It would make sense. I do not care.

net/uradvd/Makefile
```diff
@@ -0,0 +1,48 @@
...
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL=https://github.com/freifunk-gluon/uradvd.git
+PKG_SOURCE_DATE:=2021-08-14
+PKG_SOURCE_VERSION:=9b0da60e27c67305d251b10163e388191d566d7a
+PKG_MIRROR_HASH:=70dac4c156cd98baea92c9f1fb14692737b6c55c2aa0442c272ec9abe8fe06e4
+
+PKG_MAINTAINER:=Moritz Warning <moritzwarning@web.de>
+PKG_LICENSE:=MIT
```
NeoRaider, line 13: The license is wrong, this should be `BSD-2-Clause`.

net/uradvd/Makefile
```diff
@@ -0,0 +1,48 @@
...
+include $(INCLUDE_DIR)/package.mk
+
+define Package/uradvd
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Advertise an IPv6 prefix/route
+endef
+
+define Package/uradvd/description
+	Advertise an IPv6 prefix from the node
```
NeoRaider, line 27: This description is pretty useless, and also too specific ("node" doesn't make any sense in a generic OpenWrt system).

net/uradvd/Makefile
```diff
@@ -0,0 +1,48 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=uradvd
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL=https://github.com/freifunk-gluon/uradvd.git
+PKG_SOURCE_DATE:=2021-08-14
```
NeoRaider, line 8:
```suggestion
PKG_SOURCE_DATE:=2021-09-14
```

net/uradvd/Makefile
```diff
@@ -0,0 +1,48 @@
...
+endef
+
+define Package/uradvd/description
+	Advertise an IPv6 prefix from the node
+endef
+
+define Build/Prepare
+	mkdir -p $(PKG_BUILD_DIR)
+	$(CP) ./src/* $(PKG_BUILD_DIR)/
+endef
```
NeoRaider, line 33: You can drop the whole Prepare step.

net/uradvd/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+define Build/Configure
+endef
+
+define Build/Compile
+	CFLAGS="$(TARGET_CFLAGS)" CPPFLAGS="$(TARGET_CPPFLAGS)" $(MAKE) -C $(PKG_BUILD_DIR) $(TARGET_CONFIGURE_OPTS)
+endef
+
+define Package/uradvd/install
+	$(CP) ./files/* $(1)/
```
NeoRaider, line 43: While we use it in Gluon a lot, I don't like `$(CP)` for the install step that much anymore. When installing only a few files, it is IMO advisable to do the `$(INSTALL_DIR)`, `$(INSTALL_BIN)` and `$(INSTALL_DATA)` manually, as this will ensure that no unexpected permission bits are set on files.

net/uradvd/Makefile
```diff
@@ -0,0 +1,48 @@
...
+	Advertise an IPv6 prefix from the node
+endef
+
+define Build/Prepare
+	mkdir -p $(PKG_BUILD_DIR)
+	$(CP) ./src/* $(PKG_BUILD_DIR)/
+endef
+
+define Build/Configure
+endef
```
NeoRaider, line 36: Can be dropped - the default Configure step already does nothing when no `configure` script exists.

net/uradvd/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+define Build/Compile
+	CFLAGS="$(TARGET_CFLAGS)" CPPFLAGS="$(TARGET_CPPFLAGS)" $(MAKE) -C $(PKG_BUILD_DIR) $(TARGET_CONFIGURE_OPTS)
+endef
```
NeoRaider, line 40: I *think* this Compile step can also be dropped completely - as far as I can tell from looking at the code, the default definition should be fine.

net/uradvd/src/Makefile
```diff
@@ -0,0 +1,4 @@
+all: uradvd
+
+uradvd: uradvd.c
+	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -Wall -o $@ $^ $(LDLIBS) -lrt
```
NeoRaider, line 4: You need to delete the `src` directory, as you are using a proper source repo now.

net/uradvd/files/etc/init.d/uradvd
```diff
@@ -0,0 +1,69 @@
+#!/bin/sh /etc/rc.common
+
+START=50
+
+USE_PROCD=1
+
+start_instance() {
+	local cfg="$1" enabled device ifname default_lifetime args
+
+	config_get enabled $cfg 'enabled'
```
NeoRaider, line 10:
```suggestion
	config_get_bool enabled $cfg 'enabled' 1
```
I think defaulting to 1 is fine as long as the example config explicitly sets enabled to 0.

net/uradvd/files/etc/init.d/uradvd
```diff
@@ -0,0 +1,69 @@
...
+	config_get device $cfg 'device'
+	config_get ifname $cfg 'ifname'
+	config_get default_lifetime $cfg 'default_lifetime'
+
+	if [ "$enabled" != "1" ]; then
+		echo "disabled in config"
```
NeoRaider, line 16: I'd avoid printing anything when an instance is disabled.

net/uradvd/files/etc/init.d/uradvd
```diff
@@ -0,0 +1,69 @@
...
+		exit 0
+	fi
+
+	if [ -n "$device" ] && [ -n "$ifname" ]; then
+		echo "either set device or ifname"
```
NeoRaider, line 21: - Log messages should contain the instance name
- Errors and warnings should go to stderr (`>&2`)

net/uradvd/files/etc/init.d/uradvd
```diff
@@ -0,0 +1,69 @@
...
+
+	if [ -n "$default_lifetime" ]; then
+		args="$args --default-lifetime $default_lifetime"
+	fi
+
+	args="$args -i $ifname"
+
+	append_prefix_off_link() {
+		args="$args -a $1"
+	}
```
NeoRaider, line 47: Do not declare this functions inside `start_instance`.

net/uradvd/Makefile
NeoRaider: With the removal of the source and most of the package Makefile, the only remaining part of the package is the initscript and the example config - both of which I don't care about much, so I'd prefer for you to keep the maintainership.

NeoRaider (changes requested):

net/uradvd/Makefile
mwarning: I actually like `$(CP)`, as it decouples the static files to be installed and the Makefile.
Nevertheless, I have chnaged it.

net/uradvd/files/etc/init.d/uradvd
mwarning: I agree.

mwarning: But the system log (`logread`) already contains the uradvd when it logs this output.

net/uradvd/files/etc/init.d/uradvd
mwarning: I kinda disagree here. When the uradvd is enabled via /etc/rc.d/, but doesn't start, then there should be a log message.
Anyway, I removed it, as I do not care that much about his detail.

net/uradvd/Makefile
mwarning: fine with me, it is low maintenance.

net/uradvd/Makefile
mwarning: Done. But the package does not compile right now:
```
make[4]: Entering directory '/home/mwarning/openwrt/build_dir/target-mipsel_24kc_musl/uradvd'
make[4]: *** No targets specified and no makefile found.  Stop.
make[4]: Leaving directory '/home/mwarning/openwrt/build_dir/target-mipsel_24kc_musl/uradvd'
make[3]: *** [Makefile:41: /home/mwarning/openwrt/build_dir/target-mipsel_24kc_musl/uradvd/.built] Error 2
make[3]: Leaving directory '/home/mwarning/openwrt/package/uradvd'
time: package/uradvd/compile#0.86#0.19#1.58
    ERROR: package/uradvd failed to build.
```

Any idea?

NeoRaider: Remove the `PKG_BUILD_DIR` line from the Makefile to fix this error.

net/uradvd/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+define Package/uradvd/description
+	Advertise an IPv6 prefix/route via SLAAC.
+endef
+
+define Package/uradvd/install
+	$(INSTALL_DIR) $(1)/etc/config/
+	$(INSTALL_DATA) ./files/etc/config/uradvd $(1)/etc/config/
+	$(INSTALL_DIR) $(1)/etc/init.d/
+	$(INSTALL_DATA) ./files/etc/init.d/uradvd $(1)/etc/init.d/
```
NeoRaider, line 33: The initscript needs to be installed using `INSTALL_BIN`, so it is made executable.

neheb:
this ready to go?

PR #16631, opened 2021-09-15
curl: update to 7.79.0

stangri:
Maintainer: me
Compile tested: x86_64, Sophos SG-105, OpenWrt 21.02.0
Run tested: x86_64, Sophos SG-105, OpenWrt 21.02.0, install curl & libcurl, use with https-dns-proxy, download via HTTPS.

Description:
* update to [7.79.0](https://curl.se/changes.html#7_79_0)

### DIFF
--- a/net/curl/Makefile
+++ b/net/curl/Makefile
@@ -8,14 +8,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=curl
-PKG_VERSION:=7.78.0
+PKG_VERSION:=7.79.0
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://dl.uxnr.de/mirror/curl/ \
 	https://curl.askapache.com/download/ \
 	https://curl.se/download/
-PKG_HASH:=be42766d5664a739c3974ee3dfbbcbe978a4ccb1fe628bb1d9b59ac79e445fb5
+PKG_HASH:=2a1420076f9ffc35c982c78e85b7a69e2ef5d532267895fdb2eac16ad9b680c9
 
 PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=COPYING


BKPepe (approved):
LGTM

champtar (approved):

stangri:
Should I cherry-pick for 21.02? curl releases updates every month, should I always cherry-pick ones with CVEs for last release branch?

paper42:
> Should I cherry-pick for 21.02? curl releases updates every month, should I always cherry-pick ones with CVEs for last release branch?

I was looking into this update earlier today. This release seems to fix 3 CVEs, but they only fix issues in IMAP, POP3, SMTP and FTP. Some of the CVEs are serious and each curl version fixes a few new CVEs, so I think it would be nice to have them fixed in stable releases, but we should probably properly test them. Maybe we could  test it on our routers and wait a few days to see if alpine (because they use musl) reports any problems? What do you think @stangri?

stangri:
> > Should I cherry-pick for 21.02? curl releases updates every month, should I always cherry-pick ones with CVEs for last release branch?
> 
> I was looking into this update earlier today. This release seems to fix 3 CVEs, but they only fix issues in IMAP, POP3, SMTP and FTP. Some of the CVEs are serious and each curl version fixes a few new CVEs, so I think it would be nice to have them fixed in stable releases, but we should probably properly test them. Maybe we could test it on our routers and wait a few days to see if alpine (because they use musl) reports any problems? What do you think @stangri?

Sounds like a great idea. I can only do limited real-world testing and my use case is very specific (https-dns-proxy on latest release). 

I would welcome feedback from testing on other platforms, maybe unless issues are reported, I'll update curl for master upon release (mid-month) and for stable 2 weeks later (early each month)?

@jow- what do you think?

acooler15:
Can UNIX socke support be enabled by default?

stangri:
> Can UNIX socke support be enabled by default?

I suggest you create an issue, better yet, send a PR to solicit additional feedback on this (and link this discussion). From my testing, support for unix sockets caused the libcurl to grow in size by 1Kb, which is not that much of a change given ~ 170Kb size already, however there may be other considerations wherever to enable it by default or not.

acooler15:
> > Can UNIX socke support be enabled by default?
> 
> I suggest you create an issue, better yet, send a PR to solicit additional feedback on this (and link this discussion). From my testing, support for unix sockets caused the libcurl to grow in size by 1Kb, which is not that much of a change given ~ 170Kb size already, however there may be other considerations wherever to enable it by default or not.

ok, thank you.

PR #16636, opened 2021-09-16
qemu: fix compilation on 64 bit platforms

neheb:
Last patch is for mips. Second is an upstream backport. First is needed
for at least musl 1.2.x

Maintainer: @yousong 
Compile tested: mips64el

### DIFF
--- a/utils/qemu/patches/0008-falloc.patch
+++ b/utils/qemu/patches/0008-falloc.patch
@@ -1,0 +1,12 @@
+--- a/block/export/fuse.c
++++ b/block/export/fuse.c
+@@ -31,6 +31,9 @@
+ #include <fuse.h>
+ #include <fuse_lowlevel.h>
+ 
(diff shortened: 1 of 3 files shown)


utils/qemu/patches/0010-mips-softfloat.patch
```diff
@@ -0,0 +1,33 @@
+This patch is needed due to our mips64 build uses softfloat. Qemu will not
+build without this patch.
+
+--- a/tests/fp/fp-bench.c
++++ b/tests/fp/fp-bench.c
```
yousong, line 5: Can we patch out `subdir('tests` in meson.build?

neheb: I guess so.

neheb: Done.

PR #16638, opened 2021-09-16
fail2ban: patch CVE-2021-32749

paper42:
Maintainer: @erdoukki

Description: The patch is imported from the project's master, so it will be fixed in the next release. This fix should also be backported to 21.02.


erdoukki (approved):
thanks

BKPepe:
Cherry-picked into OpenWrt 21.02: https://github.com/openwrt/packages/commit/e1c24c07f47ba577f163391173df028738370ebc

Thanks

PR #16011, opened 2021-07-02
miniupnpd: add force_forwarding option support

ptpt52:
This option replace the 'ext_ip_reserved_ignore', and replace the
patch with new one

as addr_is_reserved() behavior should not be changed, so drop the
ext_ip_reserved_ignore

use a force_forwarding option to make the port forwarding force to
work when the router is behind NAT

Maintainer: me / @\<github-user> (find it by checking history of the package Makefile)
Compile tested: (put here arch, model, OpenWrt version)
Run tested: (put here arch, model, OpenWrt version, tests done)

Description:

Issue #15258, opened 2021-03-25
miniupnpd: multiple unhelpful messages in the log


stangri:
Maintainer: @neheb 
Environment: x86-64, Sophos SG-105, 21.02-SNAPSHOT r15926

Description:

I keep seeing the messages like below in the log. Is it missing a `d` at the end (as in checking is ignored)? There's no indication on why is it being reported or what caused it. Is it possible to improve this message to make it more clear?

```
00:11:33 miniupnpd[5981]: private/reserved address checking is ignore
00:11:33 miniupnpd[5981]: private/reserved address checking is ignore
00:11:33 miniupnpd[5981]: private/reserved address checking is ignore
00:17:38 miniupnpd[5981]: private/reserved address checking is ignore
00:17:39 miniupnpd[5981]: private/reserved address checking is ignore
00:17:39 miniupnpd[5981]: private/reserved address checking is ignore
00:35:23 miniupnpd[5981]: private/reserved address checking is ignore
00:35:24 miniupnpd[5981]: private/reserved address checking is ignore
00:35:24 miniupnpd[5981]: private/reserved address checking is ignore
```

neheb:
I don't really maintain this. It looks like some upstream bug.

pali:
@neheb: this change is not part of upstream and it was accepted into openwrt by [you](https://github.com/openwrt/packages/pull/14031).

Moreover this patch is incorrect and and is breaking upnp protocol as desribed in [comments](https://github.com/miniupnp/miniupnp/pull/511).

So please do not instruct users to forward openwrt bugs and issues to upstream projects like it already [happened](https://github.com/miniupnp/miniupnp/issues/542) because upstream developers really do not have a time to investigate issues caused by downstream openwrt patches.

stangri:
@ptpt52 -- can you please advise me when have I started seeing these messages in the log? What do they mean?

ptpt52:
I am trying to drop this option, Please help test https://github.com/openwrt/packages/pull/16011

Neustradamus:
There is a new patch from @ptpt52:
- https://github.com/openwrt/packages/pull/16011#issuecomment-920538698

It replaces the old 301-ext_ip_reserved_ignore.patch.

Tested by @openips!

tiagogaspar8:
Hi @stangri, is this still vaild?

stangri:
I no longer use miniupnpd, not sure if this issue persists.

PR #16639, opened 2021-09-17
miniupnpd: add force_forwarding option support

ptpt52:
This option replace the 'ext_ip_reserved_ignore', and replace the
patch with new one

as addr_is_reserved() behavior should not be changed, so drop the
ext_ip_reserved_ignore

use a force_forwarding option to make the port forwarding force to
work when the router is behind NAT

by default force_forwarding is set to 0(disabled), where change
nothing with the original behavior, and users can turn it on
according to their wishes

related: https://github.com/openwrt/packages/pull/16011

### DIFF
--- a/net/miniupnpd/patches/301-options-force_forwarding-support.patch
+++ b/net/miniupnpd/patches/301-options-force_forwarding-support.patch
@@ -1,0 +1,209 @@
+From 09690d550a1ad3cc3a8cba79aa2e970c3b2b8fbe Mon Sep 17 00:00:00 2001
+From: Chen Minqiang <ptpt52@gmail.com>
+Date: Sun, 5 Jul 2020 10:42:52 +0800
+Subject: [PATCH] options: force_forwarding support
+
+This make the port forwarding force to work even
+when the router is behind NAT
+
+Signed-off-by: Chen Minqiang <ptpt52@gmail.com>
+---
+ miniupnpd.c      | 12 ++++++++----
+ miniupnpd.conf   |  2 ++
+ natpmp.c         |  2 +-
+ options.c        |  1 +
+ options.h        |  1 +
+ testgetifaddr.c  |  2 ++
+ testportinuse.c  |  2 ++
+ upnpdescgen.c    |  2 +-
+ upnpglobalvars.h |  2 ++
+ upnpredirect.c   |  2 +-
+ upnpsoap.c       |  6 +++++-
+ 11 files changed, 26 insertions(+), 8 deletions(-)
+
+diff --git a/miniupnpd.c b/miniupnpd.c
+index f13dd8d..21ce4d3 100644
+--- a/miniupnpd.c
++++ b/miniupnpd.c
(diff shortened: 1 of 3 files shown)


ptpt52:
for use case:
1. router get private IP but it is full cone nat environment, the force forwarding UPNP redirect works.

ptpt52:
Theoretically and technically, all IPs are equal. Don't make any difference because it is a private IP. Under the understanding of this concept, this option is useful

champtar:
You need to refresh the patches (make package/miniupnpd/refresh)

net/miniupnpd/patches/301-options-force_forwarding-support.patch
```diff
@@ -0,0 +1,209 @@
...
++++ b/upnpsoap.c
+@@ -348,7 +348,11 @@ GetExternalIPAddress(struct upnphttp * h, const char * action, const char * ns)
+ 			ext_ip_addr[0] = '\0';
+ 		} else if (addr_is_reserved(&addr)) {
+ 			syslog(LOG_NOTICE, "private/reserved address %s is not suitable for external IP", ext_ip_addr);
+-			ext_ip_addr[0] = '\0';
++			if (!GETFLAG(FORCEFORWARDINGMASK)) {
++				ext_ip_addr[0] = '\0';
++			} else {
++				syslog(LOG_NOTICE, "force_forwarding enable, private/reserved address %s used as external IP", ext_ip_addr);
```
stangri, line 202: If I understand the concept correctly, `force_forwarding enabled, private/reserved address %s used as external IP` is a correct way of saying this.

stangri (changes requested):

stangri:
> Theoretically and technically, all IPs are equal. Don't make any difference even if it is a private IP. Under this concept, this option is useful

Since this came up before, would it be possible to submit this upstream so that no modifications are needed for OpenWrt package?

Also, including @BKPepe as he has merged some PRs for this before.

BKPepe (changes requested):
In your latest changes, you tried to upstream something, which was refused to be merged by upstream as you were changing some logics, etc. It was clearly hacked, which was merged here, and your warnings, errors were cryptic. Nobody knows what they meant. (= https://github.com/openwrt/packages/issues/15258 )

My response from the previous PR with the same changes didn't change:
https://github.com/openwrt/packages/pull/16011#pullrequestreview-698143667

TL;DR: We can not carry patches that are refused to be part of upstream and are causing several (even security) issues.  (= https://github.com/miniupnp/miniupnp/issues/556#issuecomment-907191205 ). Cooperate with upstream to get it right.

In the meantime, NAK.

BKPepe:
This was not sent to upstream and because of that, I'm closing it.

rmilecki:
Our attempts to upstream this feature have failed, see:
* https://github.com/miniupnp/miniupnp/pull/511
* https://github.com/miniupnp/miniupnp/pull/741

It seems however a similar option (`ignore_private_ip`) was added in the upstream by https://github.com/miniupnp/miniupnp/pull/813 and is present in the 2.3.9 release.

PR #16647, opened 2021-09-18
dtc: switch to compiling with meson

neheb:
Faster and more parallel friendly.

Maintainer: @yousong 
Compile tested: ath79


yousong:
Thank you for the work.

PR #16651, opened 2021-09-19
ninja: remove

neheb:
There's no point to this package. Ninja files are typically generated by
meson or CMake, none of which are available as packages.

Maintainer: @dhewg 
Compile tested: ath79


dhewg:
Ack, but the commit message is confusing. We'll need ninja of course, it's just that this package has been superseeded by `tools/ninja` in the main repo

dhewg:
No wait, the host build was already removed because of that reason. This is about the target package. I didn't think of that as we removed the meson target package, we could've kept that I guess...

But since that's gone this can go too I guess, until someone actually requires it on the target.

So nevermind, commit msg is good, carry on :)

PR #16657, opened 2021-09-19
clamav: update to 0.104.0

lucize:

Maintainer: me 
Tested aarch64


lucize:
here again x86_64 fails without having anything to do with this PR

PR #16664, opened 2021-09-20
envsubst: initial package v1.2.0

erdoukki:
Environment variables substitution
Written in GO

Maintainer: Gérald Kerma @erdoukki Gandalf@Gk2.net

Compile tested: arm64 Marvell EspressoBin OpenWrt 21.02.x
Run tested: arm64 Marvell EspressoBin OpenWrt 21.02.x

Description:
Environment variables substitution

source: https://github.com/a8m/envsubst

@jefferyto 

### DIFF
--- a/utils/envsubst/Makefile
+++ b/utils/envsubst/Makefile
@@ -1,0 +1,76 @@
+#
+# Copyright (C) 2021 Gerald Kerma
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=envsubst
+PKG_VERSION:=1.2.0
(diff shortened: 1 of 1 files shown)


erdoukki:
@jefferyto
is it okay to use envsubst from this github ?
Can you look at this PR, please ?

jefferyto:
Can you explain why this package is necessary? Installing a several megabytes large package just to do some string replacements does not sound like a good idea.

erdoukki:
Okay, I agree...
I have not seen the size of the package.
It was proposed to help the Crowdsec security project packaging.
Their upstream tools use a script which use envsubst and other external tools, actually not present in OpenWrt.

I have actually done the replacement using sed on my own package.

The envsusbt was proposed only in case of the upstream wizard script use.

I will do sed replacements in future tools also.

PR #16670, opened 2021-09-20
prometheus: drop from packages.git

aparcar:
I'm no longer using Prometheus on a OpenWrt device and I don't find the
time to actively maintain it. If nobody else steps up I'd drop this
package.


ja-pa:
Hi,
I'm using Prometheus, I will take maintainership after you.

PR #16676, opened 2021-09-21
nut: Add the option to compile Nut NetXML drivers with the Nut package.

supersebbo:
Also modifies the nut-server init script to include options specific to the netxml driver.

Signed-off-by: Seb Belcher <sdjbelcher@gmail.com>

Maintainer: unknown
Compile tested: x86_64
Run tested: x86_64

Description:
Add the option to compile Nut NetXML drivers with the Nut package.

### DIFF
--- a/net/nut/Config.in
+++ b/net/nut/Config.in
@@ -30,3 +30,10 @@
 		help
 			If you have a UPS connected via serial cable, select this.
 		default y
+
+        config NUT_DRIVER_NEON
+                depends on PACKAGE_nut
+                bool "Build with suport for netxml drivers"
+                help
+                        If you have a UPS connected via netxml, select this.
+                default y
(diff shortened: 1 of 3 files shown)


neheb:
missing PKG_RELEASE bump.

supersebbo:
> 
> 
> missing PKG_RELEASE bump.

I deliberately didn't bump the release because this change does not alter the output binary for any of the other nut packages except for the netxml driver, which is a standalone binary that gets dropped into the driver folder.  It simply adds the option to compile said driver. 

If you still think it needs the PKG_RELEASE bump, I can amend.  

Thanks.

supersebbo:
To clarify, my thoughts are that there is no need for anyone with an exisitng nut installation to update to this version, as it does not change anything in their installation, but it will make the new driver avialable in the package repository and that can be installed seperately with a 'okpg install nut-driver-netxml-ups' if they need it.

PR #16679, opened 2021-09-21
mutt: update to 2.1.3

paper42:
Maintainer: @philenotfound

Description: just a bugfix release


paper42:
The CI failure should be a false positive.

PR #16680, opened 2021-09-21
[collectd]: stop warning non-existent directory

graysky2:
A fresh install of luci-app-statistics expects /etc/collectd/conf.d/ to be created on the file system but the package does not create this target.  You can see output in logread to this effect:
```
 configfile: stat (/etc/collectd/conf.d) failed: No such file or directory
```

Originally reported against luci-app-statistics here but this is probably the best place to fix it:   https://github.com/openwrt/luci/issues/5373

Maintainer: @hnyman

### DIFF
--- a/utils/collectd/Makefile
+++ b/utils/collectd/Makefile
@@ -352,6 +352,8 @@ define Package/collectd/conffiles
 endef
 
 define Package/collectd/install
+	$(INSTALL_DIR) $(1)/etc/collectd/conf.d
+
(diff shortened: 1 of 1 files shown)


hnyman:
Fixed in Luci with https://github.com/openwrt/luci/commit/2c27f5c9417ff5414a2730a9c310b68fdd5f585c

(Fix applies to a new default LuCI statistics config file. The line can be commented out / removed from an already existing config file.)

PR #16685, opened 2021-09-22
libuwsc: fix compilation with GCC11

neheb:
Bad if identation is now an error.

Maintainer: @zhaojh329 
Compile tested: ath79


neheb:
This is already upstream.

PR #16687, opened 2021-09-22
keepalived: fix build with IPVS disabled

stintel:
Maintainer: @feckert @scrpi 
Compile tested: master r17554-da5bb885e1 on octeon
Run tested: N/A

Description:
The genhash binary is only built when IPVS is enabled, so make its
installation depend on IPVS being enabled.

### DIFF
--- a/net/keepalived/Makefile
+++ b/net/keepalived/Makefile
@@ -208,9 +208,11 @@ define Package/keepalived/install
 	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/sbin/keepalived \
 		$(1)/usr/sbin/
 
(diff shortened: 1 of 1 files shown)


feckert:
Somehow the CI is not satisfied with the x86_64.
`ERROR: package/feeds/base/cryptodev-linux failed to build.`
But I don't think that's the reason. Therefore I will merge it -> thanks

stintel:
> Somehow the CI is not satisfied with the x86_64.
> `ERROR: package/feeds/base/cryptodev-linux failed to build.`
>  But I don't think that's the reason. Therefore I will merge it -> thanks

Yeah, this has been failing in CI on x86_64 for weeks. I don't know what is going on, and I cannot reproduce it.

Thanks for merging!

PR #16689, opened 2021-09-22
openvswitch: add option for failure mode and fix build with libunbound enabled

stintel:
Maintainer: @yousong 
Compile tested: master r17554-da5bb885e1 on x86/64
Run tested: master r17554-da5bb885e1 on x86/64


stintel:
Thanks @yousong!

PR #16690, opened 2021-09-22
nnn: update to 4.2

paper42:
Maintainer: @BKPepe

Description: They started using fts in some version, so it now needs musl-fts and a small patch. Alpine uses the same patch.


BKPepe:
Looks good, thanks. :+1:

BKPepe (approved):

pprindeville:
@BKPepe: What's odd about this PR is it doesn't include `-lmusl-fts` in the `TARGET_LDFLAGS`.

PR #16696, opened 2021-09-24
"wide links = yes" is VFS module now since verson 4.13 and above

scott0122:
https://www.samba.org/samba/docs/current/man-html/vfs_widelinks.8.html


GeorgeSapkin:
Looks like this change is already in the Makefile.

PR #16697, opened 2021-09-24
"wide links = yes" is VFS module now since verson 4.13 and above

scott0122:
https://www.samba.org/samba/docs/current/man-html/vfs_widelinks.8.html

### DIFF
--- a/net/samba4/Makefile
+++ b/net/samba4/Makefile
@@ -259,7 +259,7 @@ ifdef CONFIG_KERNEL_IO_URING
 	SAMBA4_VFS_MODULES_SHARED :=$(SAMBA4_VFS_MODULES_SHARED)vfs_io_uring,
 endif
 ifeq ($(CONFIG_SAMBA4_SERVER_VFS),y)
-	SAMBA4_VFS_MODULES_SHARED :=$(SAMBA4_VFS_MODULES_SHARED)vfs_fruit,vfs_shadow_copy2,vfs_recycle,vfs_fake_perms,vfs_readonly,vfs_cap,vfs_offline,vfs_crossrename,vfs_catia,vfs_streams_xattr,vfs_xattr_tdb,vfs_default_quota,
+	SAMBA4_VFS_MODULES_SHARED :=$(SAMBA4_VFS_MODULES_SHARED)vfs_fruit,vfs_shadow_copy2,vfs_recycle,vfs_fake_perms,vfs_readonly,vfs_cap,vfs_offline,vfs_crossrename,vfs_catia,vfs_streams_xattr,vfs_xattr_tdb,vfs_default_quota,vfs_widelinks,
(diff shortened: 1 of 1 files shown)


Andy2244:
So whats the use-case here for openwrt?
This was a obscure functionality before and after 4.13, that's why the docu states:

> Please note that "wide links = yes" functionality is a deliberately insecure option, and should never be used in Samba installations. On Linux, bind mounts can be used instead to implement anything "wide links = yes" can enable.

The new vfs module is just there for backward compatibility, so why would anyone ever use this?

gfrancesco:
> The new vfs module is just there for backward compatibility, so why would anyone ever use this?

Well, one would use it because it needs the feature, like with any other Samba option/module. I don't get the connection between something that is there to maintain _backward compatibility_ and being marked as useless or just removed? If that were to be the best solution they would have probably dropped it upstream.

Although the samba docs claim that you can achieve the same with bind mounts (if you are on linux), I found out that it's not exactly the same and there are drawbacks (see [here](https://unix.stackexchange.com/questions/49623/are-there-any-drawbacks-from-using-mount-bind-as-a-substitute-for-symbolic-lin)).

Anyway this broke my Samba conf after upgrading (here's the broken backward compatibility), I'll check what I can do with bind mounts but I'd have to abandon symlinks on the filesystem, so would be a workaround.

Andy2244:
If bind mounts as stated does not work in a similar way, sure i can add the vfs module. I assume its just a few kb, without any extra deps. Size is the main reason i try to avoid building the bigger, rare use case modules by default.

gfrancesco:
If you want I can let you know if I can fix it on my side, but of course feel free to add it regardless if you believe so since other people's use case might be different from mine. Thanks!

gfrancesco:
Sorry to bother again, I saw that this was merged in master but not in 21.02, should I switch to master to use it or there's a plan to port it to 21.02 at any time? Thanks.

PS: If of any interest I can confirm that bind mounts does not allow me to have the same outcome, I would need an fstab with hundreds of lines and even doing that I would loose the ability to symlink files or map to a different directory structure.

Andy2244:
try switch to master, not sure if i have time to backport (test) for stable releases.

gfrancesco:
Looks like I'm running into this: [FW compiled with older musl C library](https://forum.openwrt.org/t/openssl-util-1-1-1l-1-error-relocating-usr-lib-libssl-so-1-1-time64-symbol-not-found/108861/2).
Basically I get many errors similar to this one if I upgrade `samba4-server` and/or any of the deps:
`Error relocating /usr/sbin/smbd: __fstat_time64: symbol not found`

From what I understood, I'll need to wipe, flash a new snapshot fw and retry, does that sound like a plausible solution to you?
I'm currently on 21.02

PR #16703, opened 2021-09-24
rrdtool1: adjust for musl 64bit time_t

hnyman:
Musl 1.2.2 has always 64bit time_t which causes problems in 32bit systems, when %ld or %lu is used in printf formatting for time values (as %ld is only 32bit while time_t is now 64bit).

Fix the time value printf formatting to 64bit %lld or %llu.

I combed through the %ld and %lu usage in rrdtool1 sources, and fixed the necessary places.

Maintainer: @jow- 
Compile & run tested: ipq806x / R7800

cc @neheb 
Reference to 
* https://github.com/openwrt/openwrt/pull/3004
* https://musl.libc.org/time64.html (Application Compatibility section)

----

Before the patch I ran into segmentation faults:

```
root@router1:/etc/backup.stats# rrdtool dump /tmp/rrd/router1/load/load.rrd
<!-- Round Robin Database Dump -->
<rrd>
        <version> 0001 </version>
        <step> 30 </step> <!-- Seconds -->
Segmentation fault
```


hnyman:
Btw, I haven't tested this with glibc.

Any ideas about that compatiblity angle, @neheb  ?

neheb:
The proper solution is PRIu64 here. Otherwise you get warnings under 64-bit.

time_t == long long == 32-bit
time_t == long == 64-bit

Or you could just cast the time_t variable to long long. No actual difference. printf is a mess in C.

hnyman:
> The proper solution is PRIu64 here. Otherwise you get warnings under 64-bit.

Looks like you are right. Looking at the build log on aarch64_cortex-a53 (for mt7622) with the patch, there are warnings in the build log.  In style of:

```
rrd_dump.c:42:31: warning: format '%lld' expects argument of type 'long long int', but argument 2 has type 'time_t' {aka 'long int'} [-Wformat=]
   42 |     printf("\t<lastupdate> %lld </lastupdate> <!-- %s -->\n\n",
      |                            ~~~^
      |                               |
      |                               long long int
      |                            %ld
   43 |            rrd.live_head->last_up,somestring);
      |            ~~~~~~~~~~~~~~~~~~~~~~
      |                         |
      |                         time_t {aka long int}

```

That also prompted me look look more closely at the 32bit build logs, and I found a few more format strings to patch there also :-(

utils/rrdtool1/patches/100-fix-musl122-time_t-64bit.patch
```diff
@@ -0,0 +1,171 @@
...
+     }
+     
+     if (end_tmp < start_tmp) {
+-	rrd_set_error("start (%ld) should be less than end (%ld)", start_tmp, end_tmp);
++	rrd_set_error("start (%" PRId64 ") should be less than end (%" PRId64 ")", start_tmp, end_tmp);
+ 	return(-1);
+     }
+     
+@@ -265,7 +265,7 @@ rrd_fetch_fn(
+     start_offset = (long)(*start - rra_start_time) / (long)*step;
```
neheb, line 34: long casts for a time_t sounds wrong.

hnyman: Not my change.
I have so far only changed the printf stuff.

No compiler warnings from the current changes.

neheb: Sure. It's an issue for upstream. Although that seems to be dead...

hnyman: But you are right, those are time_t variables.
Sad, but possibly the rrd needs more work.

(we are using year 2003 version, so we are missing all the later modifications for 64bit.)

hnyman: > Although that seems to be dead...

????
current versions: https://github.com/oetiker/rrdtool-1.x

But we are using so ancient version from 2003 that they have not even brought that to the github repo.
Our version is really small and does not require pango and cairo libraries like the new versions do.

neheb: Ah OK.

jow-:
I suppose continue treating time_t as 32bit int within rrdtool 1.0.x is the way to go. We don't care as much about the absolute time stamp being correct but that the values relative to each other are okay. That leaves the problem of a potential wrap-around in 2038 but if the choice is between throwing away all databases now and redo them to store 64bit times internally or to defer the same task until ~2038 I'd personally opt for the latter.

So my approach would be  adding `#include <stdint.h>` (if not already there), then ensure that all `time_t`'s are casted to `(int32_t)` where needed.

hnyman:
> I suppose continue treating time_t as 32bit int within rrdtool 1.0.x is the way to go. 

How that "treat as 32bit in rrdtool" would work in practice?
Should we change all current time reads (from musl or glibc) to include casting to 32bit?  
And change those printf formatting items to be always 32bit? 

> between throwing away all databases now 

Well, I converted yesterday my RRD databases from 32 to 64 bit via XML export/import.
Apparently some further work is needed in rrdtool, as the human-readable dates are wrong in the dump output in 64bit. Curiously, the last item is correct. (Probably due to that kind of 32bit calculations that neheb noticed above.)
```
                        <!-- 2157-11-01 17:18:16 EET / 1632559800 --> <row><v> 2.2469683333e+03 </v></row>
                        <!-- 2157-11-01 17:28:16 EET / 1632560400 --> <row><v> 2.0824516667e+03 </v></row>
                        <!-- 2157-11-01 17:38:16 EET / 1632561000 --> <row><v> 2.0320500000e+03 </v></row>
                        <!-- 2021-09-25 12:20:00 EEST / 1632561600 --> <row><v> 2.0107300000e+03 </v></row>
```

hnyman:
Btw, I tested compiling the current rrdtool 1.7.2 with pango, cairo etc.:
That succeeds in the 32bit ipq806x without any patches to rrdtool (despite the 64bit time in musl), but it introduces several megabytes of libraries (e.g. font handling that is unncessary for our purposes). So, like discussed before, not a realistic option for flash-constrained devices. (I didn't try to optimize any library usage, but just tested for feasilibity)

With our small RRD item count, the graphs do not differ much from the old ones. But e.g. the background grid is more subtle and there are month end indicators in the grid, and similar small changes.

image

hnyman:
I created a 32bit time alternative as #16835

PR #16732, opened 2021-09-26
Openwrt 21.02

woshisbsbsbsb:
i want to use pillow with libwebp on, I don't knew how to do. Maybe it work

### DIFF
--- a/lang/python/pillow/Makefile
+++ b/lang/python/pillow/Makefile
@@ -28,7 +28,7 @@ define Package/python3-pillow
   CATEGORY:=Languages
   TITLE:=The friendly PIL fork
   URL:=https://python-pillow.org/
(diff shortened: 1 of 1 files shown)


jefferyto:
Some general notes:

* A change like this first needs to be proposed for the master branch. Once it has been merged, it can be considered for release branches like openwrt-21.02.
* Your commit message(s) need to have the [correct formatting](https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#commits-in-your-pull-requests-should). In this case, you will also need to squash your commits.
* You will need to update [`PKG_RELEASE`](https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#makefile-contents-should-contain).
* And you will want to ping the package maintainer, in this case @commodo

commodo:
will close this;
for 21.02 #16767
for master #16766

Issue #16544, opened 2021-09-06
python3-light: bluetooth disabled by default


divanikus:
Hi. 

We are working on `imx6` firmware and stumbled upon base `python3-light` isn't having bluetooth support enabled by default.  The platform, by itself, contains onboard bluetooth, and as far as it is more of IoT device, it would be really cool to have support of it through `python3`. 

What are proposed steps in that case? We don't really wish to fork `python3-light` with different name. Having a separate private feed for it is also doesn't seem to be good, because it might mess up things for users. So what can we do? Any suggestions are welcome.

jefferyto:
If you compile the Python package locally, you can enable Bluetooth support in `make menuconfig` (the python3-light package should have a "Configuration" option). I believe Bluetooth isn't enabled by default to save disk space.

@commodo Anything you would like to add?

divanikus:
@jefferyto We are aware of it, but how to distribute it? No one is able to decipher if that `python` is compiled with bluetooth support or not just by observing package name. I doubt how `opkg` would handle it if we add it into our custom feed - would it download our "proper" version or official one? Also lots of dependency hell lies here I guess.

According to our test difference between `python` with bluetooth enabled and disabled is something around mere 2-5 KB.

jefferyto:
>  I doubt how `opkg` would handle it if we add it into our custom feed - would it download our "proper" version or official one?

If your python3-light package has a "newer" version string, then opkg should favour it over the one provided in the packages feed.

> According to our test difference between `python` with bluetooth enabled and disabled is something around mere 2-5 KB.

Does this include the library dependency for Bluetooth support (bluez-libs)?

divanikus:
> If your python3-light package has a "newer" version string, then opkg should favour it over the one provided in the packages feed.

Well, I don't believe it is a right thing to do, because it is not a trully "newer" version, unless there is a specific semantic for that, maybe something like Debian's epochs.

> Does this include the library dependency for Bluetooth support (bluez-libs)?

Unfortunately not, I'm talking just about `python3-light` package itself.

jefferyto:
> Well, I don't believe it is a right thing to do, because it is not a trully "newer" version, unless there is a specific semantic for that, maybe something like Debian's epochs.

That's why I put "newer" in quotation marks. I believe opkg handles version numbers in a similar, if not the same, way as apt/dpkg does, including support for epochs and things like `+really`.

divanikus:
[...]

PR #16748, opened 2021-09-27
python3: enable bluetooth support by default

commodo:
Maintainer: me, @jefferyto 
Compile tested: x86 https://github.com/openwrt/openwrt/commit/fc050c7b53116b6e3f2c31797b27cfc76af60e74
Run tested: x86 https://github.com/openwrt/openwrt/commit/fc050c7b53116b6e3f2c31797b27cfc76af60e74

--------------------------------------------------------------------------------

Bluetooth support requires bluez-libs present, but they are only required
for the build, and don't seem to be needed to be present on the target.

There isn't any linking required to libbluetooth. It's only the bluetooth.h
header that is required for building BT support into Python.

For testing, this snippet was used from `Lib/test/test_socket.py` (inside
cpython):
```
def _have_socket_bluetooth():
    """Check whether AF_BLUETOOTH sockets are supported on this host."""
    try:
        # RFCOMM is supported by all platforms with bluetooth support. Windows
        # does not support omitting the protocol.
        s = socket.socket(socket.AF_BLUETOOTH, socket.SOCK_STREAM, socket.BTPROTO_RFCOMM)
    except (AttributeError, OSError):
        return False
    else:
        s.close()
    return True
```

Fixes: https://github.com/openwrt/packages/issues/16544

### DIFF
--- a/lang/python/python3/Config-python3-light.in
+++ b/lang/python/python3/Config-python3-light.in
@@ -1,9 +1,5 @@
 menu "Configuration"
 
-config PYTHON3_BLUETOOTH_SUPPORT
-	bool "Enable Bluetooth support"
-	default n
-
 config PYTHON3_HOST_PIP_CACHE_WORLD_READABLE
 	bool "Ensure host pip cache is world-readable"
 	default n
--- a/lang/python/python3/Makefile
+++ b/lang/python/python3/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 include ../python3-version.mk
 
 PKG_NAME:=python3
-PKG_RELEASE:=1
+PKG_RELEASE:=2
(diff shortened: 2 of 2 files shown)


jefferyto (approved):
Thanks!

jefferyto:
I wonder if we need to document somewhere that bluez-libs is required for Python to work with Bluetooth?

commodo:
> I wonder if we need to document somewhere that bluez-libs is required for Python to work with Bluetooth?

[i was a bit late with the reply here]
we could probably add a mention somewhere;

i think [not entirely sure] that Python doesn't care much about bluez{-libs} other than just needing a definition for AF_BLUETOOTH, so that it can open a socket to the kernel, to talk to a BT device; 
it's the same as opening a TCP IPv4 socket via AF_INET;

PR #16778, opened 2021-09-30
mutt: add smtp, sasl, gnutls, menuconfig options

clayface:
Maintainer: @philenotfound
Compile tested: bcm53xx, MX65, master
Run tested: same

Description:
Add support for SMTP, SASL, GnuTLS and POP, IMAP, OpenSSL in the
menuconfig. The SMTP/SASL configurable was required to successfully
send email on GMail. Tested separately using OpenSSL and GnuTLS.

### DIFF
--- a/mail/mutt/Makefile
+++ b/mail/mutt/Makefile
@@ -30,7 +30,7 @@ include $(INCLUDE_DIR)/package.mk
 define Package/mutt
   SECTION:=mail
   CATEGORY:=Mail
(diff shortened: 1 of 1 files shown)


neheb:
Isn't OpenSSL deprecated upstream?

clayface:
Can't see anything about that - unless I have missed something? Initially wanted to add SMTP and SASL support only for sending emails but figured maybe better to add the extra configurables for completeness.

neheb:
Ah nvm. That;s msmtp.

PR #16766, opened 2021-09-28
libwebp: bump to 1.2.1 and add myself as maintainer ; pillow: add webp support 

commodo:
Maintainer: me
Compile tested: x86 https://github.com/openwrt/openwrt/commit/fc050c7b53116b6e3f2c31797b27cfc76af60e74
Run tested: x86 https://github.com/openwrt/openwrt/commit/fc050c7b53116b6e3f2c31797b27cfc76af60e74

Requested via: #16732

Continuing from #16765  

Signed-off-by: Alexandru Ardelean <ardeleanalex@gmail.com>

PR #16784, opened 2021-10-01
libwebp: fix webpmux lib name for cmake linking

commodo:
Maintainer: me
Compile tested: x86 master
Run tested:  x86 master

-------------------------------------

Fixes: https://github.com/openwrt/packages/pull/16766#issuecomment-931761232

When using CMake the `<prefix>/usr/share/WebP/cmake/WebPConfig.cmake` gets
used to get the names of the libraries to link against.

Since version 1.2.1 of libwebp, libwebpmux is on by default.
This causes a linker error because the linker arg should be `-lwebpmux`
instead of `-llibwebpmux`.

This is easily fixable by correcting the generation of the WebPConfig.cmake
file to append `webpmux` when this lib is enabled.

### DIFF
--- a/libs/libwebp/patches/001-fix-cmake-webpmux-linking.patch
+++ b/libs/libwebp/patches/001-fix-cmake-webpmux-linking.patch
@@ -1,0 +1,55 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -471,19 +471,18 @@ endif()
+ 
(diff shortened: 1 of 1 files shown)


commodo:
Changelog v1 -> v2:
* multiple renames libwebpmux -> webpmux so tha the target installs

anomeome:
master - compile / test : mvebu / mamba

PR #16785, opened 2021-10-01
travelmate: support meta-refresh with single-quote

ayufan:
Currently `travelmate` only support `<meta` tag if it contains `"`.
This updates `travelmate.sh` to support`'` as well.

This works, because we want to emit `['"]` in a single arg:

- we close current string with `'`
- we open a new string with `"`, emit `'` (needed character), close a string `"`
- we open a new string `'` till the rest of string, and include `"` double quotes

```shell
root@slate:~# echo '['"'"'"]'
['"]
```

Maintainer: me / @dibdot
Compile tested: ath79/nand
Run tested: OpenWrt 21.02.0 r16279-5cc0535800, GL.iNet GL-AR750S (NOR/NAND)

Description:

### DIFF
--- a/net/travelmate/Makefile
+++ b/net/travelmate/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=travelmate
 PKG_VERSION:=2.0.7
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 PKG_LICENSE:=GPL-3.0-or-later
 PKG_MAINTAINER:=Dirk Brenken <dev@brenken.org>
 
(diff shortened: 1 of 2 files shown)


ayufan:
Fixed failing CI.

dibdot:
The sequence `'\'' `does the trick: it closes the single-quote literal, specifies the quote character (using an escape that is supported outside of single-quote literals) and then re-opens a new single-quote literal. So the new search term to handle both variants should look like this: `['\''"]` .... also please bump the PKG_RELEASE in the Makefile.

ayufan:
I thought I was testing that. Will check later. Thanks :)

ayufan:
@dibdot Updated

ayufan:
@dibdot Do you think could be picked into stable as a bug fix?

PR #16787, opened 2021-10-02
tailscale: update to version 1.16.0

kevin-david:
Maintainer: `@ja-pa`
Compile tested: mipsel_24kc / MT7621
Run tested: ~~mipsel_24kc / MT7621 on a RB750Gr3~~ TODO
Description: Updating to latest stable from https://pkgs.tailscale.com/stable/#static

### DIFF
--- a/net/tailscale/Makefile
+++ b/net/tailscale/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tailscale
-PKG_VERSION:=1.12.3
+PKG_VERSION:=1.16.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=tailscale-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/tailscale/tailscale/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=05e5b1d382cad7ac5737d87d0b61277791c75938468c2c662f21665998d431e9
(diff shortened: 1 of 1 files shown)


kevin-david:
@ja-pa Thanks for the guidance. 

>Please update patch (dirty patch detection is failing)
make package/tailscale/refresh V=s see https://openwrt.org/docs/guide-developer/build-system/use-patches-with-buildsystem

I think the required checks are passing now - do I still need to run that command? I think my build environment was a bit screwed up so I might have to repave it if that's still necessary.

>and move github action change to separate PR (you can ping @aparcar).

Done! #16849 which is now merged.

>Also from my point of view PKG_HASH comment is unnecessary because it is already described in https://openwrt.org/docs/guide-developer/packages#buildpackage_variables

I removed the comment - it was mostly a note for me since I didn't know how to find that! Thanks for the pointer.

### Now what?

I still can't test this on my install completely - it seems like the `tailscale` and `tailescaled` packages are getting even larger - now topping 7MB of `.ipk` files combined. When installing `tailscaled` I do get this error, but unclear if it's anything that matters:
```
Error relocating /usr/sbin/tailscaled: __nanosleep_time64: symbol not found
```

I've been trying to get extroot working on my RB750Gr3 but [haven't had much success yet](https://forum.openwrt.org/t/extroot-on-rb750gr3-loading-wrong-modules-at-boot/108725).

1715173329:
closed due to #17354

PR #16799, opened 2021-10-04
prometheus-node-exporter-lua: add dsl error vectors to ltq-dsl

dhewg:
Maintainer: @champtar 
    
These are useful stats to debug vector related line deteriorations, see [0].

NOTE: this depends on https://github.com/openwrt/openwrt/pull/4633

[0] https://forum.openwrt.org/t/vectoring-on-lantiq-vrx200-vr9-missing-callback-for-sending-error-samples/104046

### DIFF
--- a/utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/ltq-dsl.lua
+++ b/utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/ltq-dsl.lua
@@ -10,6 +10,7 @@ local function scrape()
   local dsl_max_datarate = metric("dsl_max_datarate", "gauge")
   local dsl_error_seconds_total = metric("dsl_error_seconds_total", "counter")
   local dsl_errors_total = metric("dsl_errors_total", "counter")
+  local dsl_erb_total = metric("dsl_erb_total", "counter")
 
   local u = ubus.connect()
   local m = u:call("dsl", "metrics", {})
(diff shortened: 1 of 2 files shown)


dhewg:
Anyone? Bueller?

utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/ltq-dsl.lua
```diff
@@ -21,13 +22,15 @@ local function scrape()
...
+    driver_version = m.driver_version,
   }, 1)
 
   -- dsl line settings information
   metric("dsl_line_info", "gauge", {
     annex = m.annex,
-    mode = m.mode,
+    standard = m.standard,
     profile = m.profile,
+    mode = m.mode,
```
champtar, line 33: There is no need to move `mode`

champtar (changes requested):
Sorry for the delay, can you not move `mode` and squash everything in 1 commit ?

dhewg:
np, and done! ;)

champtar:
If we move to autorelease commit should be `update to 2021.10.31`. Maybe we need an 'autoversion' for packages with source in tree. For now I think autorelease will produce weird results, so maybe drop it (or tell me what you get)

dhewg:
It's mostly to not care about the pkg version anymore, as that's increased with each commit for the package. I think it's useful to have and gets rid of the "you forgot to bump the version" churn on reviews.

But it's basically just for convenience, I removed it again as requested

champtar (approved):

dhewg:
Thanks!

champtar:
Thank you for the patch !

PR #16801, opened 2021-10-04
acme: update to 3.0.1

httpstorm:
Update acme client to the latest version 3.0.1

@tohojo @nbd168 @jow- @adschm


httpstorm:
The Test x86_64 which failed is unrelated to this commit. I just compiled acme for x64, WRT3200, TL-WR1043ND v4 and Orange PI. The firmware installed and everything seems fine.

Someone has to fix the failing library:
ERROR: package/feeds/base/cryptodev-linux failed to build

tohojo:
Hmm, v3.0 changes the default CA, seemingly for commercial reasons. That seems fishy; let's add a configuration option expose the 'server' directive for acme.sh, and default it to letsencrypt, before merging this.

httpstorm:
> Hmm, v3.0 changes the default CA, seemingly for commercial reasons. That seems fishy; let's add a configuration option expose the 'server' directive for acme.sh, and default it to letsencrypt, before merging this.

@tohojo
Hi Toke! Sure, just let me know when you update acme, and I will update the PR.
The reason I suggested to update acme is because a few days ago one root CA and one intermediary CA certificate from LetsEncrypt expired. As a result my web server httpstorm.com went offline for a few hours until I recompiled it to serve the full certificate chain. I hope the latest version of acme is adapted to these changes.

May I also suggest that you add `option days '60'` to the config, to make sure certificates are updated 30 days before they expire? This value seems reasonable and I saw other people suggesting it.

Oh, and thank you for creating this nice project! It's very comfortable that whenever I need new certificates, I can simply grab them from my router, and it takes care to keep them updated. 😎

tohojo:
Georgi Valkov ***@***.***> writes:

>> Hmm, v3.0 changes the default CA, seemingly for commercial reasons. That seems fishy; let's add a configuration option expose the 'server' directive for acme.sh, and default it to letsencrypt, before merging this.
>
> @tohojo
> Hi Toke! Sure, just let me know when you update acme, and I will
> update the PR.

Well, I was hoping I could rope you into doing it as part of this PR,
actually ;)

> Oh, and thank you for creating this nice project! It's very
> comfortable that whenever I need new certificates, I can simply grab
> them from my router, and it takes care to keep them updated. 😎

You're welcome! That's an interesting use case! I've only used it to get
a proper cert for Luci on my own routers...

httpstorm:
> Well, I was hoping I could rope you into doing it as part of this PR, actually ;)

I don't mind doing it, of course some instructions will be very helpful, since I don't have experience with the scripts and it would take me a long time to see how they work and what exactly you want me to do. You could also share the changed files on pastebin, and I will add them to the PR tomorrow.

> You're welcome! That's an interesting use case! I've only used it to get a proper cert for Luci on my own routers...

Especially with the DNS validation method, I can generate certificate without even accessing the actual servers.

@tohojo

tohojo:
Hmm, seems we actually already support specifying 'server' in the config, so it's just the default we need to add. Something like (completely untested):

```patch
diff --git a/net/acme/files/run.sh b/net/acme/files/run.sh
index 2401c95f146b..741e41971e47 100644
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -280,6 +280,9 @@ issue_cert() {
        if [ -n "$acme_server" ]; then
                log "Using custom ACME server URL"
                acme_args="$acme_args --server $acme_server"
+       else
+               # default to letsencrypt instead of upstreams' zerossl weirdness
+               acme_args="$acme_args --server letsencrypt"
        fi
 
        if [ -n "$days" ]; then
```

tohojo:
On 7 October 2021 19:54:26 CEST, Georgi Valkov ***@***.***> wrote:
>Yes, it would be safer to handle staging, since they might change it in the future.
>
>Does that look reasonable? Should I use `letsencrypt_test ` or `letsencrypttest`?
>```
># default to letsencrypt because the upstream default may change
>if [ "$use_staging" -eq "1" ]; then
>	acme_args="$acme_args --server letsencrypt_test"
>else
>	acme_args="$acme_args --server letsencrypt"
>fi
>```
>

Yup, exactly. And just remove the old check for staging entirely while you're at it. Don't think it matters which you use, but I find letsencrypt_test slightly more readable, so let's go with that :)

httpstorm:
@tohojo OK, changes applied. But I didn't get which old check for staging I should remove? It is always helpful when you provide a few patch lines and explanation. 😉 I also prefer the human readable form `letsencrypt_test `.
@theAeon Like Toke said, the intend of this commit is to update acme to 3.0.1. Adding features to the LuCI GUI sound like a good plan for a second PR. I have good experience in HTML, and no experience with LUA, though I might be able to implement this. Should I give it a try?

tohojo:
Alright, LGTM. Thanks! :)

tohojo:
Care to make another PR for the openwrt-21.02 branch backporting the change?

httpstorm:
@tohojo Thank you, Toke! 

> Care to make another PR for the openwrt-21.02 branch backporting the change?

So basically I checkout a tag, apply the same changes and make a PR, don't I?
That will be my first time, and new experience is welcome. 😀

@theAeon We made sure the defaults remain to use Let's Encrypt, because upstream acme changed its default.

httpstorm:
@tohojo 
https://github.com/openwrt/packages/pull/16855
Should we also update `luci-app-acme`?

httpstorm:
> worth pointing out that luci-app-acme has no conception of CAs outside of default, staging, and custom-so any changes here will need to be reflected over there.

@theAeon
On master and openwrt-21.02, LuCI does support setting a custom `ACME server URL`, which is saved in the `acme_server` config. No change needed.

PR #1653, opened 2015-08-12
Xl2tpd fix

yousong:
Hi,

This pull request depends on a OpenWrt netifd patch [1] just sent and should not be merged until that patch got acked.

I found that xl2tpd segfault fairly frequently after enabling coredump from the service script.  I think that should be the main reason why it fails to reconnect as reported in #1334.

This series tries to solve it by adding a new interface config option `checkup_interval` for netifd to check if the interface is really up after that many seconds since its last attempt to setup the interface so that even if xl2tpd dropped off during that period, netifd can still bring it up.

However, there was a time I noticed that xl2tpd was trapped in a dead loop without any syscall and cannot service control commands from xl2tpd-control and as such the interface cannot be brought up.  Cases like this have to be solved at the xl2tpd side.

[1] http://patchwork.ozlabs.org/patch/506592/

PR #16804, opened 2021-10-05
xl2tpd: add persist and maxfail options

stintel:
Maintainer: @yousong 
Compile tested: NA
Run tested: OpenWrt master r17520+18-5ef4608c02 on qoriq


stintel:
> The reconnect logic should be working out of the box: #1653 .

My L2TP connection dies and does not reconnect at all without these changes.

predators46:
Work for me `option checkup_interval '10'` reconnect

PR #16808, opened 2021-10-05
zabbix: switch dependencies to php8

mhei:
Maintainer: @champtar 
Compile tested: mxs
Run tested: **no** @champtar: please test (if possible)

Description:

This just switches package dependencies to php8 variants as preparation of getting rid of php7.
Since I'm not familiar with zabbix, I'd appreciate runtime testing by maintainer and/or others.


champtar:
I've never ran the GUI on OpenWrt, but never removed it in case there were some users for it :)
Let's wait to see if anyone wants to test else we will merge and wait for bug report

PR #16819, opened 2021-10-06
cgroupfs-mount: conflict with procd-ujail

stintel:
Maintainer: @G-M0N3Y-2503 / cc @dangowrt 
Compile tested: NA
Run tested: NA

Description:
The procd ujail feature expects the unified cgroup hierarchy to be
mounted on /sys/fs/cgroups. Normally, procd takes care of this, but
having cgroupfs-mount installed will cause the unified hierarchy to be
unmounted and mounts the legacy cgroup hierarchy instead. This
completely breaks procd-ujail, so add a conflict to avoid installing
this package when procd-ujail is enabled.

### DIFF
--- a/utils/cgroupfs-mount/Makefile
+++ b/utils/cgroupfs-mount/Makefile
@@ -25,6 +25,7 @@ define Package/cgroupfs-mount
   SECTION:=utils
   CATEGORY:=Utilities
   TITLE:=cgroup mount scripts
+  CONFLICTS:=procd-ujail
   DEPENDS:=+mount-utils
   MENU:=1
 endef


jmarcet:
There is one other thing, with the non legacy unified cgroup hierarchy, some metrics are missing.

Namely ctop does not report I/O stats anymore and cadvisor does not report docker containers' memory usage either.

stintel:
After discussing with @dangowrt my conclusion is this package should be dropped, as procd requires cgroup2 and this package simply breaks it. Instead, other packages should be fixed to use cgroup2.

PR #16835, opened 2021-10-07
rrdtool: Keep time as 'long' despite 64bit time_t in musl

hnyman:
Adjust to the 64 bit time_t in musl 1.2.2 by keeping time internally in RRD still as 'long', so that old databases
created during musl 1.1.x still work in 32 bit targets.

Alternative approach suggested by @jow-  in #16703

Time handling has been changed so that rrdtool internally keeps time as long, just like earlier. 
Some functions require intermediate variables as time_t, so that functions expecting a pointer to time_t work right (when time_t is 64bit).

----

@jow- 
Please check.

I haven't tested rrd_cgi, but otherwise the old database works, and graphs get updated and "rrdtool dump" works.

I haven't yet tested this with 64bit targets, but there 64bit long should match 64bit time_t in any case.

### DIFF
--- a/utils/rrdtool1/Makefile
+++ b/utils/rrdtool1/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=rrdtool1
 PKG_VERSION:=1.0.50
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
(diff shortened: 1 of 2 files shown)


hnyman:
Commit updated:
I haven't run-tested rrdcgi, but I think that I have now fixed also it sensibly and it compiles ok.

jow-:
That is exactly the kind of change I had in mind, thanks for working on it. From my pov it is good to go

PR #16836, opened 2021-10-07
atlas-probe: Fix compilation with gcc11

cotequeiroz:
Maintainer: @ja-pa 
Compile tested: ath79
Run tested: ath79/mips, openwrt master and a snapshot from 2021/08 with must 1.1.24.  

Description:

This uses some definitions from `<sys/cdefs.h>` in gcc 8.4.0, not present in musl or gcc11.

Also use `clock_gettime()` instead of `syscall(__NR_clock_gettime,...)`, which is not currently defined.

---

V1->V2: 
add run-test
rebased the actual patch sent upstream, plus a patch runit/runsv.c, not needed in master.

### DIFF
--- a/net/atlas-probe/Makefile
+++ b/net/atlas-probe/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=atlas-probe
 PKG_VERSION:=2.2.1
-PKG_RELEASE:=1
+PKG_RELEASE:=3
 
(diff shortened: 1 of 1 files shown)


cotequeiroz:
I've submitted the change upstream: https://github.com/RIPE-NCC/ripe-atlas-probe-busybox/pull/4 - after I had run some really basic tests, running the package's `date` with different versions of musl.

PR #16840, opened 2021-10-08
add multi-ip ddns support

zxrobin01:

Description:
If a domain is bind to more than 1 ip, the script will always shows update failed and tries to update every minute. I replace the checking part for the script to avoid that situation.


BKPepe:
Missing ``Signed-off-by``, etc. Closing.

Issue #16833, opened 2021-10-07
nginx: Dynamic modules and other changes to Makefile


alex9434:
Maintainer: @Ansuel @mhei 
Environment: all

Description: Suggested changes to NGINX Makefile for OpenWRT.

Dear all,

Can I suggest the following 3 changes to the NGINX makefile:

**1. Add MP4 pseudo-streaming:**
The makefile already contains the option `--with-http_flv_module` to add FLV pseudo streaming. NGINX has the comparable functionality to add MP4 pseudo streaming `--with-http_mp4_module`. The code for the module is included in the NGINX source code. Given the more widespread use of MP4 compared to FLV today that module might be very useful.

**2. Add Digest authentication**
Because some routers/boards do not have hardware encryption support it might be helpful to be able to stream media over HTTP instead of HTTPS. Digest Authentication offers better protection than Basic Authentication over HTTP because the password is not transmitted in cleartext.
https://www.nginx.com/resources/wiki/modules/auth_digest/

**3. Compile Dynamic Modules instead of Static Modules**
In NGINX 1.9.11 onwards selected modules can be loaded into NGINX at runtime based on configuration files. This can help to reduce memory requirements by only loading modules that are needed. Currently, all modules are statically compiled into NGINX. Can we change this to dynamic loading?
https://www.nginx.com/blog/compiling-dynamic-modules-nginx-plus/

**Next steps?**
I have already made the changes for 1 and 2 by myself to the Makefile and Config-ssl.in (they are actually trivial). If you agree with these suggestions, I am happy to work on a pull request (have never done this before). On point 3 it would be good to get your view and the suggested approach.

Thank you for your kind consideration!

Ansuel:
This night i will check the required changes for the makefile and post here a pof...

anyway in theory most of the standard module are in the nginx wiki and everyone is listed if can be loaded dynamically or not.

Ansuel:
@alex9434 i made a first version of this. Can you test it?

alex9434:
Regarding MP4 could you kindly add the following to the Makefile?
`
...
CONFIG_NGINX_MP4 \
...
  ifeq ($(CONFIG_NGINX_MP4),y)
    ADDITIONAL_MODULES += --with-http_mp4_module
  endif
...
 ADDITIONAL_MODULES += --with-ipv6 --with-http_stub_status_module --with-http_flv_module  --with-http_mp4_module
`
and to Config_ssl.in
`
config NGINX_MP4
	bool
[...]

PR #16842, opened 2021-10-08
nginx: introduce support for dynamic modules

Ansuel:
Start building sub package that provide dynamic modules.
Each module needs to be loaded using load_modules. Refer
to nginx documentation on how to use this.
This should result in lower memory usage as only used module
are loaded.

~~This is a first step to dynamic loading... Some testing is needed and i still need to add description to the new package.~~

Requested from: #16833

### DIFF
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -95,14 +95,14 @@ define Package/nginx-ssl
 	+NGINX_PCRE:nginx-ssl-util +!NGINX_PCRE:nginx-ssl-util-nopcre \
 	+NGINX_HTTP_GZIP:zlib +NGINX_DAV:libxml2
   EXTRA_DEPENDS:=nginx-ssl-util$(if $(CONFIG_NGINX_PCRE),,-nopcre) (>=1.5-1) (<2)
-  CONFLICTS:=nginx-all-module
+  CONFLICTS:=nginx-full
 endef
 
 Package/nginx-ssl/description = $(Package/nginx/description) \
   This variant is compiled with SSL support enabled. To enable additional module \
   select them in the nginx default configuration menu.
 
-define Package/nginx-all-module
+define Package/nginx-full
   $(Package/nginx/default)
   TITLE += with ALL module selected
   DEPENDS+=+libpcre +nginx-ssl-util +zlib +libxml2 \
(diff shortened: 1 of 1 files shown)


Ansuel:
@alex9434 Are you still interested? Sorry for the delay of 2 years :(

Ansuel:
@peter-stadler Are you ok with the change for the module? Or do you have suggestion on how to implement?

Ansuel:
@hgl If you want to check this...

hgl:
Awesome work! I'll try it out on my device soon.

A couple of questions from first glance:

1. Is using pcre instead of pcre2 a compromise for support of some dynamic modules?
2. Did you get any warnings from the lua module that the nginx it lives in is not from OpenResty?
3. Since new package names are introduced, do you think it would be a good opportunity to rename some packages as mentioned in my discussion with Peter? Specifically:
    1.  Rename nginx-all-module to nginx-full (a lot popular packages use the `-full` suffix, like dnsmasq-full, ip-full)
    2. Since both you and Peter agree that nginx should have a default config that supports HTTPS, then maybe nginx should be the main package and make nginx-ssl depend on it for compatibility?

Ansuel:
> 1. Is using pcre instead of pcre2 a compromise for support of some dynamic modules?

as said in the commit it's for lua module... since both are supported by nginx is it a problem to use pcre instead of pcre2?

> 2\. Did you get any warnings from the lua module that the nginx it lives in is not from OpenResty?

the lua module still needs some love as the openresty core and lrucache are still required (and also adding the correct path in the config) but if some user wants to use it we at least provide it. (manual modification are needed anyway since the module is loaded dynamically and that needs to be manually added to the nginx config)

> 3\. Since new package names are introduced, do you think it would be a good opportunity to rename nginx-all-module to nginx-full as mentioned in my discussion with Peter, since a lot of popular packages use the `-full` suffix, like dnsmasq-full, ip-full?

Yep sure! Also since we are doing this name change... wonder if we can just drop the package name... guess we had enough time to make every user transition to it?

Ansuel:
No idea why CI fails to build.

npenin:
Any chance to also have the njs module? And also a package for the njs standalone executable?

Ansuel:
@npenin for standalone no idea... but for the module... as you can see it's pretty easy to add additional external module now...

alex9434:
Great, thank you. Will try it out now!

PR #16846, opened 2021-10-09
libsodium: update maintainer email address

damianorenfer:
Maintainer: me
Compile tested: N/A
Run tested: N/A

Description:
Simply change maintainer email address


champtar:
Your old email will stay forever in git, but I don't see a problem using a relay email

PR #16853, opened 2021-10-10
transmission: update seccomp file

acooler15:
Maintainer:  @dangowrt  @neheb 
Compile tested: x86_64, openwrt master(e672d1b387)
Run tested: aarch64, raspberry pi 4B, OpenWrt SNAPSHOT(r17727-e672d1b387)

Description:
The transmision-daemon.json file is generated by using `/etc/init.d/transmission trace`, using the original seccomp failed to boot on my device.

### DIFF
--- a/net/transmission/Makefile
+++ b/net/transmission/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=transmission
 PKG_VERSION:=3.00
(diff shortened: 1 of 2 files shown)


acooler15:
In my tests, `procd_set_param env TRANSMISSION_WEB_HOME="$web_home"` was not passed into the transmission process,  but that doesn't seem to be a problem with the transmission program

PR #16856, opened 2021-10-10
python-pyopenssl: Update to 21.0.0

jefferyto:
Maintainer: me, @commodo 
Compile tested: armvirt-32, 2021-10-03 snapshot sdk
Run tested: none

Description:


jefferyto:
A problem with meson is causing all CI builds to fail (this is why I had to use an older sdk for local build).

jefferyto:
I agree that Rust will be necessary, whether it is a Rust package here or requiring Rust as a build requirement for the main repo is another question. It doesn't look like there are any security fixes in the new version of cryptography so I don't think there is any rush to update yet.

PR #16873, opened 2021-10-12
usteer: add package

blocktrron:
Maintainer: me / @blocktrron 
Compile tested: ath79/generic
Run tested: ath79/generic (UAP AC Lite)

Description:

This commits adds the new usteer package to the packages feed.

usteer is a daemon for steering wireless clients across frequency
bands as well as between multiple access points on a network.

### DIFF
--- a/net/usteer/Makefile
+++ b/net/usteer/Makefile
@@ -1,0 +1,42 @@
+# SPDX-License-Identifier: GPL-2.0-only
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=usteer
+PKG_SOURCE_DATE:=2021-09-21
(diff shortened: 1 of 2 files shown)


blocktrron:
@champtar usteer requires hostapd flavors with WNM enabled, which is not included in the default package set. Currently i do not see a benefit in having usteer in core, as it has to be manually installed anyhow.

PolynomialDivision:
Hmmm, still fails with glibc?

blocktrron:
@PolynomialDivision pushed shit yesterday, should work now.

Also added a SPDX header, although i don't really care for such a generic Makefile.

PR #16890, opened 2021-10-14
knot-resolver: update to 5.4.2

paper42:
Maintainer: @ja-pa
Run tested: aarch64, Turris MOX, OpenWrt 21.02

Description:
* refresh patches
* disabling kres_gen_test is not required anymore for cross compilation, it was fixed upstream with the 5.4.1 release


paper42:
The CI failure on powerpc_464fp is from LuaJIT, on x86_64 it is from cryptodev-linux, so they shouldn't be related to this PR.

PR #16906, opened 2021-10-17
[19.07] vpn-policy-routing: downgrade to 0.2.1-13

stangri:
I've received multiple reports that newer versions (0.3.x) are not working properly on _some_ 19.07 installs and the latest working-for-everyone version is 0.2.1-13. I'm downgrading the 19.07 version to last known good version as I don't want to spend time troubleshooting this package on an outdated release.

(cherry picked from commit 7bb2ccd4e173486d1c51a72374c55398c9c7e725)

### DIFF
--- a/net/vpn-policy-routing/Makefile
+++ b/net/vpn-policy-routing/Makefile
@@ -4,8 +4,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=vpn-policy-routing
-PKG_VERSION:=0.3.2
-PKG_RELEASE:=20
+PKG_VERSION:=0.2.1
+PKG_RELEASE:=13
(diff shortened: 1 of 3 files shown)


stangri:
@aparcar Paul, is it OK to use cherry-pick for the downgrade like this?

BKPepe:
I would not use cherry-pick for this downgrade as you did. Downgrade looks good OK. Even though, you should mention in the commit message, why you decided to downgrade it and reverted README to its previous state. 

> on an outdated release.

IMHO, which is not exactly true, though. It is still supported and should receive important bug fixes, security fixes, etc. But if you don't want to troubleshoot it and decided to downgrade it, then there could be some more issues in the future, and it is going to be a pity if you are not going to fix them. It is just a point of view.

PR #16910, opened 2021-10-17
ksmbd/ksmbd-tools: update to 3.4.2

socram8888:
Maintainer: @Andy2244
Compile tested: apm821xx, WD My Book Live single, https://github.com/openwrt/openwrt/commit/b118efa0d25f5b60226a9d316eb838dd6be22f78
Run tested: apm821xx, WD My Book Live single, https://github.com/openwrt/openwrt/commit/b118efa0d25f5b60226a9d316eb838dd6be22f78

Description:
Update Ksmbd module and tools to latest version.

The patch for the Ksmbd module has been deleted as it has been already integrated upstream, so it is no longer needed.

### DIFF
--- a/net/ksmbd-tools/Makefile
+++ b/net/ksmbd-tools/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ksmbd-tools
-PKG_VERSION:=3.4.1
+PKG_VERSION:=3.4.2
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


Andy2244:
> 
> 
> Could it be the failed compilation for x86-64 is an error on the pipeline? I've tried compiling it for "Generic x86/64" and sure enough, it built successfully.

yeah probably, looks fine otherwise.

neheb:
the problem is cryptodev does not compile with the x86 SDK. Nobody has bothered to fix it.

PR #16913, opened 2021-10-17
libreswan: compile without USE_DNSSEC

paulwouters:
The libunbound in openwrt is compiled without libevent support, and
libreswan needs that when compiled with USE_DNSSEC. This means
currently, 3.32 packages crash on boot with:

 ipsec pluto --stderrlog --nofork

[...]

Oct 17 18:31:33.970124: Failed to initialize unbound libevent ABI, please recompile libunbound with libevent support or recompile libreswan without USE_DNSSEC
Oct 17 18:31:33.970217: seccomp security for crypto helper not supported
Oct 17 18:31:33.970333: 2 crypto helpers shutdown
Oct 17 18:31:33.979417: ABORT: ASSERTION FAILED: event_initialized(&se->ev) (in free_signal_handlers() at server.c:635)
Aborted

Description:

### DIFF
--- a/net/libreswan/Makefile
+++ b/net/libreswan/Makefile
@@ -85,6 +85,7 @@ MAKE_FLAGS+= \
     FINALRUNDIR="/var/run/pluto" \
     FINALNSSDIR="/etc/ipsec.d" \
     MODPROBEARGS="-q" \
+    USE_DNSSEC=false \
     ARCH="$(LINUX_KARCH)" \
 
 define Build/Prepare


lucize:
yes, but if you disable the dnssec, then unbound is not needed, so maybe you need to correct the unbound package

paulwouters:
note that unbound is apparently fixed in 21.02.0, where libreswan works properly. So I guess this PR is wrong. I'm not sure how to file a PR that works only against 20.x versions ?

paulwouters:
thanks! Will close this one and open one on the branch. Thanks for the quick responses!

PR #16931, opened 2021-10-19
cloudreve: add new package

1715173329:
Maintainer: me
Compile tested: bcm27xx/bcm2711, rockchip/armv8, x86/64
Run tested: x86/64

Description:
cloudreve is a self-deployed file management and sharing system, supports multiple storage providers.
Documentation: https://docs.cloudreve.org

Compared to nextcloud, this one is much more thinner and easy-to-use. No third-party dependencies were needed.


1715173329:
> May want to use codeload for cloudreve but this is fine - thanks!

I tried but it seemed not to work, as this repo has a git submodule, which won't be packed into tarball.
Thanks for review!

PR #16935, opened 2021-10-19
miniflux: new package

paper42:
Maintainer: me

Description: miniflux is a feed reader using PostgreSQL, so it needs some setup, enabling the service is not enough. After starting the postgresql service, run
```
opkg install pgsql-cli-extra
createuser -U postgres -P miniflux
createdb -U postgres -O miniflux miniflux
psql -U postgres miniflux -c 'create extension hstore'
psql -U postgres -c 'ALTER USER miniflux WITH SUPERUSER;'
DATABASE_URL='user=postgres password=XXX dbname=miniflux sslmode=disable' miniflux.app -migrate
DATABASE_URL='user=postgres password=XXX dbname=miniflux sslmode=disable' miniflux.app -create-admin
DATABASE_URL='user=postgres password=XXX dbname=miniflux sslmode=disable' miniflux.app
```
then change the database password in /etc/config/miniflux and run the miniflux service.

I chose port 5433 as the default port to avoid conflicts, but maybe 8080 (or something similar) would make more sense (?)

### DIFF
--- a/utils/miniflux/Makefile
+++ b/utils/miniflux/Makefile
@@ -1,0 +1,59 @@
+#
+# Copyright (C) 2021 CZ.NIC, z. s. p. o. (https://www.nic.cz/)
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=miniflux
+PKG_VERSION:=2.0.33
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/miniflux/v2/tar.gz/${PKG_VERSION}?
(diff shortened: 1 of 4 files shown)


paper42:
> Would be nice if the database setup can be automated but I don’t know how much of it requires interactive input (if any).

"The database creation" process consists of 2 parts:

1. database setup (needs database password, but that could be randomly generated and automatically updated in uci). I don't think this should be automated for everyone, because a user might want to use a different device for the database.
```
createuser -U postgres -P miniflux
createdb -U postgres -O miniflux miniflux
psql -U postgres miniflux -c 'create extension hstore'
psql -U postgres -c 'ALTER USER miniflux WITH SUPERUSER;'
```

2. admin user creation - This one definitely needs user input, we probably don't want to choose the username and generate a password for the user.
```
DATABASE_URL='user=postgres password=XXX dbname=miniflux sslmode=disable' miniflux.app -create-admin
```

It would be possible to provide a script which the user would have to manually run that would "complete" the installation, but it should probably be in a subpackage and I don't know about any other packages which do this in OpenWrt, so it wouldn't be consistent. If there is interest, I will definitely try to create the script.

PR #16940, opened 2021-10-20
bossa: add bossa

PolynomialDivision:
Based on the work of Hirokazu MORIKAWA (nxhack):
https://github.com/nxhack/openwrt-arduino-packages/tree/master/bossa
    
BOSSA is a flash programming utility for Atmel's SAM family of flash-based ARM microcontrollers. The motivation behind BOSSA is to create a simple, easy-to-use, open source utility to replace Atmel's SAM-BA software.

Runtested with feather m0, flashing firmware to it on banana pi r64.
However, I needed to add a reset of serial connection with python:

```python
import serial

com = serial.Serial()
com.port = '/dev/ttyACM0'
com.baudrate = 1200
com.timeout = 1
com.stopbits = 1
com.parity = serial.PARITY_NONE
com.setDTR(False)
com.open()
com.close()
```

### DIFF
--- a/utils/bossa/Makefile
+++ b/utils/bossa/Makefile
@@ -1,0 +1,45 @@
+# SPDX-License-Identifier: GPL-2.0-only
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=bossa
+PKG_VERSION:=1.9.1
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/shumatech/BOSSA/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=ca650455dfa36cbd029010167347525bea424717a71a691381c0811591c93e72
+PKG_BUILD_DIR:=$(BUILD_DIR)/BOSSA-$(PKG_VERSION)
+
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
> I am not so sure about it. You can simply add an Arduino feed to your's `feeds.conf`. This is what we are doing with the node feed for Turris OS. @nxhack is doing great work. +1 It would be a pity if he does not have at least authorship here for it. Maybe with the video feed, we can have node feed (as well the arduino), but that's decision is up to @aparcar and other OpenWrt members.

Maybe we can put it under the hood of openwrt? With getting more devs board like banana pi r64 those tools are getting more and more interesting for openwrt.

PolynomialDivision:
@BKPepe As I understood @nxhack he is not longer intersted in maintaing bossa, see here:
https://github.com/nxhack/openwrt-arduino-packages/issues/16#issuecomment-948106919
But maybe I misunderstood it and he still wants to have it in his packages feed. But I think would be awesome to get the stuff to official OpenWrt commnity.

nxhack:
@aparcar 

I no longer do any maintenance on this package. I'm happy if @PolynomialDivision  will maintain it well, and I'm in favor of adding it to the community packages if it's needed by OpenWrt users.

aparcar:
@nxhack thanks for the prompt response.

@PolynomialDivision I'm fine adding this package if you want to become the maintainer.

PR #16947, opened 2021-10-21
apparmor: update to 3.0.3

neheb:
Fix compilation under Fedora. There's some weird systemd target. Just
remove it.

Maintainer: @oskarirauta 
Compile tested: mips64el


paper42:
What's the purpose of the 190 line patch called `060-openwrt-dnsmasq-profile.patch`? As far as I can tell, it just adds 2 lines and removes a few comments, otherwise it's just rearranging the lines.
```patch
+  /tmp/** r,
+  include <abstractions/user-tmp>
-  # Neighbor Discovery protocol (RFC 2461)
-  # NetworkManager integration
-  # Site-specific additions and overrides. See local/README for details.
-  # access to iface mtu needed for Router Advertisement messages in IPv6
-  # for the read-only TFTP server
-  # libvirt config and hosts file for dnsmasq
-  # libvirt lease helper
-  # libvirt pid files for dnsmasq
-  # lxc-net pid and lease files
-  # lxd-bridge pid and lease files
-    # libvirt lease and status files for dnsmasq
```

020-fix-ss-path.patch could be replaced by an upstream patch [use internal which implementation](https://gitlab.com/apparmor/apparmor/-/commit/c31d49234f0c7a81b69d525e249b154e11c486a9)

If we want this version of apparmor to support Python 3.10, we should import these two patches: https://gitlab.com/apparmor/apparmor/-/commit/06dc32a9f3ede351c46167afa5002a0fe062c0f7 https://gitlab.com/apparmor/apparmor/-/commit/6cfc6eeec22d3c9ffae413819654359b76680150.

PR #16950, opened 2021-10-21
vnstat2: fix monitored interfaces

janh:
Maintainer: me
Compile tested: x86-64, master
Run tested: x86-64, QEMU/KVM, master
(different configurations including update from legacy databases have been tested)

Description:

This fixes two issues where vnStat does not actually monitor the interfaces as configured:

- By default, vnstatd adds all available interfaces on startup when its database is empty. The --noadd option prevents this, but it breaks import of legacy databases, and causes vnstatd to exit immediately after startup, which breaks reloading.

  This changes the init script to add the --noadd option when no legacy databases need to be imported, and patches vnstatd to keep running even when no interfaces are configured.

- If an interface doesn't exist yet when vnStat is started, it won't be monitored, as only existing interfaces can be added to the database via the vnstat command.

  This adds a hotplug script which adds any configured interfaces to the vnStat database when it goes up.

### DIFF
--- a/net/vnstat2/Makefile
+++ b/net/vnstat2/Makefile
@@ -77,6 +77,8 @@ define Package/vnstat2/install
 	$(INSTALL_CONF) ./files/vnstat.config $(1)/etc/config/vnstat
 	$(INSTALL_DIR) $(1)/etc/init.d
 	$(INSTALL_BIN) ./files/vnstat.init $(1)/etc/init.d/vnstat
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/iface
+	$(INSTALL_DATA) ./files/vnstat.hotplug $(1)/etc/hotplug.d/iface/50-vnstat
 endef
 
 define Package/vnstati2/install
--- a/net/vnstat2/files/vnstat.hotplug
+++ b/net/vnstat2/files/vnstat.hotplug
@@ -1,0 +1,23 @@
+/etc/init.d/vnstat running || exit 0
+[ "$ACTION" = "ifup" ] || exit 0
+
+. /lib/functions.sh
+
+check_iface() {
+	local iface="$1"
+
+	[ "$iface" = "$DEVICE" ] || return
+
+	/usr/bin/vnstat --add -i "$iface" >/dev/null
+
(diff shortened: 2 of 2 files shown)


vergoh:
fyi @janh (and sorry that I hadn't seen this change earlier),

vnStat 2.12 (not released yet) will have something similar to `--noexit` you've implemented in your patch but named `--startempty` (https://github.com/vergoh/vnstat/commit/46c3b3f56d33ba382f61cd13dee09b461b506f9f). Without some additional changes, the daemon would end up polling the database every 5 seconds for interfaces as long as none were seen.

You may also find `--noremove` (https://github.com/vergoh/vnstat/commit/c9bfab279278eb2a293835c00719c4148d879994) useful for the second bullet you mentioned. If you know in advance the name of the interfaces that you plan on monitoring then you can `vnstat --add interface --force` those to the database regardless of those being present when being added and when `--noremove` is being used, the daemon isn't going to clean those out either.

As for other "probably worth mentioning" changes already included for version 2.12, there's finally 95th percentile output (which does however require `5MinuteHours` configuration to be at least 744). There's also `--invert-colors` option for image output which may be practical for dark mode UIs as it generates the images with a dark background when used without having to modify the color palette configuration.

janh:
@vergoh These new options work great, and [really simplify the OpenWrt package a lot](https://github.com/janh/packages/commit/dd95975c34b7bba2b24309033d83e1870ce15501). Thank you for adding them! I probably should have opened a feature request right away instead of just silently adding a downstream patch.

There is only one small imperfection remaining now, as the init script needs to check for unimported legacy database files and skip configuration of interfaces in that case (because `--initdb` does not do the import).

vergoh:
@janh Thanks for pointing out the issue with `--initdb` not trying to import legacy data. That's now fixed with https://github.com/vergoh/vnstat/commit/91a77f11284fa2f20dac73e11f5e9ec10e63f7c5. The old behaviour can still be enabled with `--initdb --noadd` if needed in some scenario.

janh:
@vergoh Thank you again, it works fine!

PR #16894, opened 2021-10-15
rclone: fix logic err in service file

1715173329:
Maintainer: @ElonH, me
Compile tested: -
Run tested: rk3328 nanopi-r2s

Description:
- Corrected `enable` detection.
- Granted config/log write permission for `rclone` user.

Fixes: d51823988847fc ("rclone: initial package")

PR #16968, opened 2021-10-24
[openwrt-21.02] rclone: bring up packages

1715173329:
Maintainer: @ElonH, me
Compile tested: bcm27xx/bcm2710
Run tested: RaspberryPi 3B

Description:
Rclone ("rsync for cloud storage") is a command line program to sync
files and directories to and from different cloud storage providers.

This is a backport for openwrt-21.02 based on PR #16871 #16894.

I got really positive feedback in https://forum.openwrt.org/t/getting-rclone-into-openwrt/108614,
however, I think it's worthy to have such a nice utility in stable release.

Ping @aparcar @BKPepe @jefferyto @neheb @PolynomialDivision

### DIFF
--- a/net/rclone-ng/Makefile
+++ b/net/rclone-ng/Makefile
@@ -1,0 +1,43 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Copyright (C) 2020 Elon Huang <elonhhuang@gmail.com>
+# Copyright (C) 2021 ImmortalWrt.org
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=rclone-ng
+PKG_VERSION:=0.5.0
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=RcloneNg-v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/ElonH/RcloneNg/releases/download/v$(PKG_VERSION)/
+PKG_HASH:=0b4916ddd0bacb5b358dc8d36b64c30f4182c0ace3eb06cda94b6578c419dcd9
+
+PKG_LICENSE:=MIT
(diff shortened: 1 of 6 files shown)


champtar:
Different packages have different release style, most of the time running latest upstream version is the right thing to do IMO. As long as there is no impact on other packages I don't see a reason not to backport new packages/libs either.

neheb:
> 
> 
> I am still unsure if that was ironic or not.

mostly.

hnyman:
> Overall I don't have an issue with adding packages to existing releases since those packages are not added by default and don't affect others. I remember however that people complained about it in the past, is there reasoning for that?

The main risk is that the new packages will be build by the buildbot, and thus the existing packages' configure scripts may found them, and possibly cause new dependencies in those other packages (unless the found functionality/capability of the new package is disabled in other packages). That is true especially for library-like packages.

The library dependency things should have naturally already materialised in master, and those issues should be fixed in the backport. The backport author should look into related commits in master.

Otherwise, backporting non-library packages that nothing may depend upon, is pretty harmless.

jefferyto:
While I’m not strictly against adding new packages to an existing release, for me a release implies a level of stability that master does not. My understand of why we want to merge packages into master first is so that they can be tested by users that can accept some level of instability. Adding new packages to an existing release may lead to a poor experience for users that expect stability.

(I’m not questioning the stability of this package in particular, just speaking about the practice of add new packages in general.)

This lack of a testing period in master may also lead to extra maintenance effort going to fixing issues in both the master and release versions of the package.

That said, this past release cycle (21.02) was not ideal, so perhaps some flexibility should be afforded for new packages.

I think I would like to see some sort of agreed-upon guidelines for merging new packages that everyone can abide by, rather than each committer basing their decisions on their own preferences. If everyone agrees that adding new packages to an existing release is fine, I’m okay to go along with that.

PR #17007, opened 2021-10-28
afuse: Add new package

miska:
Maintainer: me 
Compile tested: mvebu, Turris Omnia, Turris OS 5.3 = OpenWrt 19.07
Run tested: mvebu, Turris Omnia, Turris OS 5.3 = OpenWrt 19.07

Description:

This commit add new package afuse into packages feed.

afuse is an automounting file system implemented in user-space using
FUSE. It implements the most basic functionality that can be expected by
an automounter; that is it manages a directory of virtual directories.
If one of these virtual directories is accessed and is not already
automounted, afuse will attempt to mount a filesystem onto that
directory. If the mount succeeds the requested access proceeds as
normal, otherwise it will fail with an error.

### DIFF
--- a/utils/afuse/Makefile
+++ b/utils/afuse/Makefile
@@ -1,0 +1,53 @@
+#
+# Copyright (C) 2021 Michal Hrusecky <michal@hrusecky.net>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=afuse
+PKG_VERSION:=0.4.1
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=Michal Hrusecky <michal@hrusecky.net>
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=COPYING
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/pcarrier/$(PKG_NAME)/archive/refs/tags/v$(PKG_VERSION).tar.gz?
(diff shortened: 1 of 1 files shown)


utils/afuse/Makefile
```diff
@@ -0,0 +1,54 @@
...
+
+PKG_MAINTAINER:=Michal Hrusecky <michal@hrusecky.net>
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=COPYING
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/pcarrier/$(PKG_NAME)/archive/refs/tags/v$(PKG_VERSION).tar.gz?
+PKG_HASH:=8c7950cdeb24cdff828cdd3807537ee12cab87b18ecef4496b7bdedb2b8bd0e9
+
+
```
BKPepe, line 22:
```suggestion
```

Somehow, it happened that now there are two empty lines. Sorry for nitpicking.

BKPepe (approved):
Otherwise, looks good to me. :+1: 
I performed compile test, and it worked. It would be good to add `test.sh `in the next pull requests to know how run testing in CI works for this package.

BKPepe:
powerpc CI failure:
`WARNING: No feed for package 'afuse' found`

Interesting, it is the same as it happens with recent lighttpd update as well. :thinking:

miska:
CI would be challenging as it is a filesystem, so it would pose some requirements on the underlying system. Not sure whether it will work in CI. I can give it a try, but not sure whether it will work.

miska:
> Otherwise, looks good to me. +1 I performed compile test, and it worked. It would be good to add `test.sh `in the next pull requests to know how run testing in CI works for this package.

Seems like random error in CI, re-triggering fixed it.

PR #17021, opened 2021-10-29
openvswitch: bring up member ports

stintel:
Maintainer: @yousong 
Compile tested: NA
Run tested: OpenWrt 21.02-SNAPSHOT, r16273

Description:
Open vSwitch does not bring up ports automatically. This is not a
problem for wireless ports, or for ports configured in
/etc/config/network, but other ports will be down, and require manual
interaction to be brought up. Configuring them with proto none will
cause netifd to do some actions on them, which might cause undefined
results, and will also bloat the UCI config file.

The cleanest solution is to bring all member ports up as part of the
init script.

### DIFF
--- a/net/openvswitch/Makefile
+++ b/net/openvswitch/Makefile
@@ -17,7 +17,7 @@ include ./openvswitch.mk
 #
 PKG_NAME:=openvswitch
 PKG_VERSION:=$(ovs_version)
-PKG_RELEASE:=9
+PKG_RELEASE:=10
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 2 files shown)


dangowrt:
Is this an issue also when using `ovsd`?

stintel:
> Is this an issue also when using `ovsd`?

The uCentral implementation does not use ovsd, and I need this for a fix in uCentral.

dangowrt:
Sad, as ovsd would also solve dynamic reload without having to restart openvswitch. I'd love to see it being used for uCentral as well.

That being said, for static configurations your solution looks fine to me (and I don't see how it could hurt dynamic/ovsd based setups. @arkap what you think?)

yousong (approved):
LGTM

PR #17024, opened 2021-10-30
gnunet: fix invalid group spec for postgres

BKPepe:
Maintainer: @dangowrt
Compile tested: N/A
Run tested: N/A

Description:
Fixes warning in CI:
feeds/packages_ci/net/gnunet/Makefile: invalid group spec :postgres=5432
feeds/packages/net/gnunet/Makefile: invalid group spec :postgres=5432

### DIFF
--- a/net/gnunet/Makefile
+++ b/net/gnunet/Makefile
@@ -321,7 +321,7 @@ CONFLICTS_fs-mysql:=gnunet-fs-pgsql gnunet-fs-sqlite
 
 DEPENDS_pgsql:=+libpq +pgsql-server
 LIB_pgsql:=pq
-USERID_pgsql:=gnunet=958::postgres=5432
+USERID_pgsql:=gnunet=958:postgres=5432
 
 DEPENDS_dhtcache-pgsql:=+gnunet-pgsql
 PLUGIN_dhtcache-pgsql:=datacache_postgres


dangowrt:
NACK. This requires updated buildroot to define additional group membership. See also https://github.com/openwrt/openwrt/commit/b2aca61360c33ad3161f8ef35a1c28ebb2507600

BKPepe:
Hmm... interesting. Recent change. I would notice that if you were using GitHub pull requests to have it compile and run tested by CI or mention that in that commit message that it requires or depends on the commit in the main repository.

Hopefully, the Docker image will be updated soon so the warning gets rid of it ASAP.

Reference: https://github.com/openwrt/packages/commit/c3c6a2ff1d231e10e593956c004430af9f8687b5

dangowrt:
Yeah right, sorry for the confusion and for having wasted your time.

Unfortunately CI has never worked well for GNUnet, the dependency trail is too big, it takes hours and something always fails before it even gets to gnunet (last time I tried: cryptodev-linux failed mysteriously, some gstreamer stuff on other targets, ...) Hence I do only local tests (incl. run-time) for some days for gnunet...

BKPepe:
Yeah! Understood. :-( Everything has its issues. Nothing to worry about. I just noticed and wanted to try it. No problem at all, Daniel.

> cryptodev-linux failed mysteriously

Yes, this one fails for a month or maybe even longer for x86_64, I think somewhere was a fix for it, but I could not find it right now. Hopefully, the CI/CD will get better over time.

PR #17032, opened 2021-10-31
MariaDB cleanup

miska:
Maintainer: me
Compile tested: mvebu, Turris Omnia, Turris OS 5.3 = OpenWrt 19.07
Run tested: mvebu, Turris Omnia, Turris OS 5.3 = OpenWrt 19.07

Removed all my local configuration and local storage, started from scratch, installed packages, started database, it created new empty dtabase, daemon is running and I was able to connect to the database and was able to start using it.

Description:
Various changes as described in commits. Mainly pushing defaults from config file to the compile options and improving the init script and defaults to make it easier to get started and not to break that easily when somebody changes the defaults.


neheb:
I don't see a fix for liburing dependency

miska:
I don't see any dependency on liburing, are you sure this comment is related to this pull request?

neheb:
This is related to the latest version bump. Latest version introduces liburing support, which the buildbots pick up. I don’t see it being disabled here, which means it will still fail with the buildbots.

cotequeiroz:
#17064 takes care of liburing--not by disabling it, but accepting it as, supposedly, a better alternative to libaio.  @miska, feel free to cherry-pick the commits if you want to do it here instead.

PR #17034, opened 2021-10-31
telegraf: Update packate to version 1.20.3

JonnyTischbein:
Signed-off-by: Jonathan Pagel jonny_tischbein@systemli.org

Maintainer:
me

Compile tested:
on amd64 for aarch64 on https://git.openwrt.org/openwrt/openwrt.git commit 3bbc476e9306b1b0f51070d4dd24f90878458146

Run tested:
aarch64 (Banana PI R64), package installed, configured, starting via /etc/init.d/, configuring telegraf config and successfull working of input plugin prometheus and output plugin influxdb tested.

Description:
Update to version 1.20.3

Telegraf is a plugin-driven agent for collecting and sending metrics and events. It supports various inputs (including prometheus endpoints) and is able to send data into InfluxDB. https://www.influxdata.com/time-series-platform/telegraf/

### DIFF
--- a/utils/telegraf/Makefile
+++ b/utils/telegraf/Makefile
@@ -59,6 +59,12 @@ define Package/telegraf-full/description
 	(Full build. including all plugins)
 endef
 
+define Package/telegraf/conffiles
+/etc/telegraf.conf
+endef
+
+Package/telegraf-full/conffiles = $(Package/telegraf/conffiles)
+
 define Build/Prepare
 	$(call Build/Prepare/Default)
 ifeq ($(BUILD_VARIANT),small)
@@ -70,7 +76,7 @@ define Package/telegraf/install
 	$(call GoPackage/Package/Install/Bin,$(1))
 	$(INSTALL_DIR) $(1)/etc/init.d $(1)/etc/config
 	$(INSTALL_BIN) ./files/etc/init.d/telegraf $(1)/etc/init.d/telegraf
-	$(INSTALL_DATA) $(PKG_BUILD_DIR)/etc/telegraf.conf $(1)/etc/config/telegraf.conf
+	$(INSTALL_CONF) $(PKG_BUILD_DIR)/etc/telegraf.conf $(1)/etc/telegraf.conf
 endef
 
 define Package/telegraf-full/install
--- a/utils/telegraf/Makefile
+++ b/utils/telegraf/Makefile
@@ -3,12 +3,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=telegraf
-PKG_VERSION:=1.20.2
+PKG_VERSION:=1.20.3
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/influxdata/telegraf/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=3b0db9f48a76bb5076eded0ceab6ea568b295bbc525cdd8ad1ecf700ef317c18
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
Something is weird. ;) Your commit message cintains two times Signed off. Further please do two seperate commits or include everything in one commit. You do a rename of the config file. That will maybe cause some trouble for people, so please add this in a seperate commit message.

JonnyTischbein:
Thanks!
I fixed the commit message and seperate it into 2 commits.

Should / Can I announce somewhere the change of the config file location ?
My reason was that all other `/etc/config/` file do not have an ending `.conf` and since it is not neccessairy I removed it,

BKPepe (changes requested):
If I am not mistaken the config is not marked as configuration file. Thus it is going to be overwriten by each update. Is that on purpose? You can try to verify it by installing the package with force-reinstall.

Also, the path for the file is wrong. I take a look at upstream repository for `telegraf.conf` is not uci config.  If you take a look, all configs in `/etc/config` are in uci format. You should place it under `/etc`

JonnyTischbein:
No it is not on purpose.
How do I mark it as a config file ? Is it enough to change the Makefile to

```
define Package/telegraf/install
        .....
	$(INSTALL_CONF) $(PKG_BUILD_DIR)/etc/telegraf.conf $(1)/etc/telegraf.conf
```

Thanks for your hint with the uci config. I missed that!

BKPepe:
Yes, it should. 

______ 

You can mark it as configuration file by adding this to your Makefile:

```
define Package/telegraf/conffiles
/etc/telegraf.conf
endef

Package/telegraf-full/conffiles = $(Package/telegraf/conffiles)
```

But as always, don't blindly use copy&pasting, but perform run testing first.

BKPepe:
Yes, it should. 

______ 

You can mark it as configuration file by adding this to your Makefile:

```
define Package/telegraf/conffiles
/etc/telegraf.conf
endef

Package/telegraf-full/conffiles = $(Package/telegraf/conffiles)
```

But as always, don't blindly use copy&pasting, but perform run testing first.

JonnyTischbein:
I tested your example with my small and full version successfully.
With a modified config I get

```
root@Mesh1:~# opkg install --force-reinstall /tmp/telegraf-full_1.20.3-1_aarch64_cortex-a53.ipk 
Installing telegraf-full (1.20.3-1) to root...
Configuring telegraf-full.
Collected errors:
 * resolve_conffiles: Existing conffile /etc/telegraf.conf is different from the conffile in the new package. The new conffile will be placed at /etc/telegraf.conf-opkg.
```
which is the same behaviour I get when I change e.g. the `/etc/tor/torrc` and reinstall the tor package (as a comparison package).

utils/telegraf/Makefile
```diff
@@ -59,6 +59,13 @@ define Package/telegraf-full/description
 	(Full build. including all plugins)
 endef
 
+define Package/telegraf/conffiles
+/etc/telegraf.conf
+endef
+
+Package/telegraf-full/conffiles = $(Package/telegraf/conffiles)
+
+
```
BKPepe, line 68:
```suggestion
```

Just removing empty row.

BKPepe:
I would slightly change your commit subject as now it might belong.  E.g. "move config and mark it as conffile"

And in the commit message, you can have something like: "/etc/config is uci default config folder, so let's rather move it to /etc/ and while at it mark configuration file as conffile, which means that each update won't overwrite it"

But it's up to you. Otherwise LGTM!

JonnyTischbein:
Thank you, I added your suggestions!

BKPepe (approved):

PolynomialDivision:
Thanks! :)

PR #17048, opened 2021-11-02
libyang: et all to 2+

lucize:
Maintainer: @jsmolic 
Tested x86,kirkwood and ramips

update libyang and all the applications needed to 2.x
this is needed for frr 8+

### DIFF
--- a/libs/libyang/Makefile
+++ b/libs/libyang/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libyang
-PKG_VERSION:=1.0.225
+PKG_VERSION:=2.0.112
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/CESNET/libyang/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=1b736443d2c69b5d7a71ac412655e6edab0647b18f35f7bf504b0a24e06cb046
+PKG_HASH:=184dd67c66c1ad968a2ee4d0950fb6b103834917b04b17af9c7bca80967636ee
 
 PKG_MAINTAINER:=Jakov Smolic <jakov.smolic@sartura.hr>
 PKG_LICENSE:=BSD-3-Clause
@@ -30,15 +30,7 @@ define Package/libyang
   CATEGORY:=Libraries
   TITLE:=YANG data modeling language library
   URL:=https://github.com/CESNET/libyang
-  DEPENDS:=+libpcre +libpthread
-endef
-
-define Package/libyang-cpp
-  SECTION:=libs
-  CATEGORY:=Libraries
-  TITLE:=YANG data modeling C++ language library
-  URL:=https://github.com/CESNET/libyang
-  DEPENDS:=+libyang +libstdcpp
+  DEPENDS:=+libpcre2 +libpthread
 endef
 
 define Package/yanglint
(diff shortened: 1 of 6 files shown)


libs/libyang/Makefile
```diff
@@ -55,20 +55,17 @@ define Package/libyang/description
...
-	$(INSTALL_DIR) $(1)/usr/lib/libyang1
-	$(INSTALL_DIR) $(1)/usr/lib/libyang1/extensions
-	$(INSTALL_DIR) $(1)/usr/lib/libyang1/user_types
-	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/lib/libyang1/extensions/* $(1)/usr/lib/libyang1/extensions
-	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/lib/libyang1/user_types/* $(1)/usr/lib/libyang1/user_types
+#	$(INSTALL_DIR) $(1)/usr/lib/libyang1
+#	$(INSTALL_DIR) $(1)/usr/lib/libyang1/extensions
+#	$(INSTALL_DIR) $(1)/usr/lib/libyang1/user_types
+#	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/lib/libyang1/extensions/* $(1)/usr/lib/libyang1/extensions
+#	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/lib/libyang1/user_types/* $(1)/usr/lib/libyang1/user_types
```
jsmolic, line 68: can we remove these lines entinery instead of commenting them out?

lucize: yes, I'll do it in few minutes

lucize:
I'll verify again with a clean make to see why libyang compile is not starting

lucize:
@neheb I was unable to reproduce the build issue with a clean make on debian 10 for x86, kirkwood and ramips

libs/libyang/Makefile
```diff
@@ -55,20 +55,11 @@ define Package/libyang/description
 endef
 
 CMAKE_OPTIONS += \
-	-DENABLE_LYD_PRIV:BOOL=ON \
-	-DCMAKE_BUILD_TYPE:String="Release" \
-	-DGEN_LANGUAGE_BINDINGS=ON \
-	-DGEN_PYTHON_BINDINGS=OFF
+	-DCMAKE_BUILD_TYPE:String="Release"
```
neheb, line 58: This is default: https://github.com/openwrt/openwrt/blob/master/include/cmake.mk#L94

lucize:
no dice, should I try a separate PR ?

neheb:
Doubt it. I assume C++ bindings are compiled conditionally with some configuration option.

lucize:
I'll add one at time and see

lucize:
@neheb didn't help, I guess merge this and see what build bots are thinking ?

neheb:
No way. The buildbots would most likely fail in the same way. Note that every single target is failing. I'll look into this soon.

neheb:
I get the same failure locally. The issue is that there is no libyaml-cpp anymore with version 2. Which means everything referencing it needs to be removed. That CMAKE_OPTIONS like needs removal as well. It's already default.

neheb:
python3-sysrepo is probably gone too. Should probably remove as well.

lucize:
yes, it seems is a standalone package now

net/sysrepo/Makefile
```diff
@@ -8,24 +8,22 @@
...
 PKG_SOURCE_URL:=https://codeload.github.com/sysrepo/sysrepo/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=2cc7537a03f48dc3c955436e1e0ed077bc3b31a755d6979d24ca42e1187fce01
+PKG_HASH:=fe09da5f40fb53e3fb97268a134cc0ed3003f0018d0d117c73e81e1553a11f30
 
 PKG_MAINTAINER:=Jakov Smolic <jakov.smolic@sartura.hr>
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=LICENSE
 
 CMAKE_INSTALL:=1
 PKG_BUILD_DEPENDS:=swig/host
```
neheb, line 23: This was for python. Not needed now.

libs/libyang/Makefile
```diff
@@ -55,25 +47,11 @@ define Package/libyang/description
 endef
 
 CMAKE_OPTIONS += \
-	-DENABLE_LYD_PRIV:BOOL=ON \
-	-DCMAKE_BUILD_TYPE:String="Release" \
-	-DGEN_LANGUAGE_BINDINGS=ON \
-	-DGEN_PYTHON_BINDINGS=OFF
+	-DCMAKE_BUILD_TYPE=Release
```
neheb, line 50: Already default.

lucize:
@neheb, good review, as always !

PR #17056, opened 2021-11-03
boringssl: remove dependency on golang

martinschneider:
This commit removes the unnecessary runtime dependency on Go. Go is only
required during buildtime.

Maintainer: me
Compile tested: x86_64
Run tested: x86_64

### DIFF
--- a/libs/boringssl/Makefile
+++ b/libs/boringssl/Makefile
@@ -28,7 +28,7 @@ define Package/boringssl
 	CATEGORY:=Libraries
 	TITLE:=An implementation of the TLS protocol
 	URL:=https://boringssl.googlesource.com/boringssl/
-	DEPENDS:=+libstdcpp $(GO_ARCH_DEPENDS) @!(mips||mips64)
+	DEPENDS:=+libstdcpp @!(mips||mips64||powerpc||powerpc64||arc)
 endef
 
 define Package/boringssl/description


libs/boringssl/Makefile
```diff
@@ -28,7 +28,7 @@ define Package/boringssl
 	CATEGORY:=Libraries
 	TITLE:=An implementation of the TLS protocol
 	URL:=https://boringssl.googlesource.com/boringssl/
-	DEPENDS:=+libstdcpp $(GO_ARCH_DEPENDS) @!(mips||mips64)
```
1715173329, line 31: Well, `$(GO_ARCH_DEPENDS)` doesn't include any runtime dependencies actually, but just limit the package to build for architectures supported by golang only.

martinschneider:
Using the initial version `opkg` also tried to install Go (which isn't needed and also fails on most routers due to space limitations).

jefferyto:
If you download the [boringssl .ipk file](https://downloads.openwrt.org/snapshots/packages/x86_64/packages/boringssl_20210608-1_x86_64.ipk) and extract the `control` file, this is what you would see:

```
Package: boringssl
Version: 20210608-1
Depends: libc, libstdcpp6
Source: feeds/packages/libs/boringssl
SourceName: boringssl
License: OpenSSL ISC
LicenseFiles: LICENSE
Section: libs
SourceDateEpoch: 1635383364
Maintainer: Martin Schneider <martschneider@google.com>
Architecture: x86_64
Installed-Size: 881002
Description:  An implementation of the TLS protocol
```

Unless `libstdcpp6` has a hidden dependency on Go, installing this package will not install the Go compiler. I suggest you do an installation test for this package again.

martinschneider:
I stand corrected. Closing this PR.

PR #17057, opened 2021-11-03
python-yaml: add host build

rmelotte:
Maintainer: @BKPepe 
Compile tested: no
Run tested: no

Description:
Some packages require a host python-yaml to build.

Add a host build similar to how it's done for the python-six package,
to allow other packages to depend on it.

### DIFF
--- a/lang/python/python-yaml/Makefile
+++ b/lang/python/python-yaml/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-yaml
 PKG_VERSION:=5.3.1
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PYPI_NAME:=PyYAML
 PKG_HASH:=b8eac752c5e14d3eca0e6dd9199cd627518cb5ec06add0de9d32baeee6fe645d
@@ -19,6 +19,9 @@ PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=LICENSE
 PKG_CPE_ID:=cpe:/a:pyyaml_project:pyyaml
 
+HOST_BUILD_DEPENDS:=python3/host
+
+include $(INCLUDE_DIR)/host-build.mk
 include ../pypi.mk
 include $(INCLUDE_DIR)/package.mk
 include ../python3-package.mk
@@ -35,6 +38,14 @@ endef
 define Package/python3-yaml/description
   PyYAML is a YAML parser and emitter for the Python programming language.
 endef
+
+define Host/Compile
+	$(call HostPython3/ModSetup,,install --prefix="" --root="$(STAGING_DIR_HOSTPKG)")
+endef
+
+Host/Install:=
+
+$(eval $(call HostBuild))
 
 PYTHON3_PKG_SETUP_GLOBAL_ARGS:=--with-libyaml
 PYTHON3_PKG_SETUP_ARGS:=


jefferyto:
Which packages require host python-yaml?

Instead of adding a host build, would adding a host pip requirements file and setting `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=PyYAML` be sufficient?

rmelotte:
>Which packages require host python-yaml?

prplMesh (it's not yet part of OpenWrt): https://gitlab.com/prpl-foundation/prplmesh/prplMesh/

>Instead of adding a host build, would adding a host pip requirements file and setting HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=PyYAML be sufficient?

That's what I tried initially, but unfortunately with this approach I didn't get the `yaml` module built at all (only python3 without the yaml module).

What is the reason (except historical) for having between both  `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` and the ability to define a `package/host` package?
What's the difference between the two?
I tried to understand it from the Git history, but it's not yet clear to me.

jefferyto:
> prplMesh (it's not yet part of OpenWrt): https://gitlab.com/prpl-foundation/prplmesh/prplMesh/

Do you have an OpenWrt package makefile we can use to debug this? `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` should work but without something to compile it is hard to diagnose your issue.

> What is the reason (except historical) for having between both `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` and the ability to define a `package/host` package?
What's the difference between the two?
I tried to understand it from the Git history, but it's not yet clear to me.

Host build packages are built on the host, packages defined in `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` are installed on the host using pip.

I believe we transitioned to using `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` because it was easier to manage. @commodo may have more details.

I would prefer to remove the host build of python-six (and use `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS`  instead) but I am not familiar enough with the ovn/openvswitch packages to make this change.

commodo:
> prplMesh (it's not yet part of OpenWrt): https://gitlab.com/prpl-foundation/prplmesh/prplMesh/

I forgot that prpl still exists :p 
They were a bit active (for a while) around OpenWrt trying to get closer to the community.

> I believe we transitioned to using HOST_PYTHON3_PACKAGE_BUILD_DEPENDS because it was easier to manage. @commodo may have more details.

The `HOST_PYTHON[3]_PACKAGE_BUILD_DEPENDS` mechanism was very useful in the Python2 and Python3 era.
It  was created to fix the issue of Python2 and Python3 packages.
Host build VARIANTs are ok for target builds (in the OpenWrt build system context), but nearly impossible to maintain when you want to package (let's say python-cffi) for both Python2 & 3.
(cffi is one package that also requires a host-side build);
So, if you need python-cffi and python3-cffi, you need 2 more cffi builds in the `{build_dir,staging_dir}/hostpkg/python[3]-cffi` folders
In this case you end up with 2 solutions:
1. some complicated Makefile rule to support some sort of host-side build VARIANTs
2. keep 2 Makefiles for each Python2 & 3 package (which personally, I always wanted to avoid)

On point 2, I remember cringing when I wanted to maintain Python 2 & 3, then Python{2,3}-setuptools and Python{2,3}-pip.
So, whenever updating setuptools or pip [they were packaged separately from Python for a while], I would need to update 2 Makefiles all the time, which also required double-testing or double attention.
With one Makefile, even if it was 2 Python interpreters, I felt more comfortable testing just for Python2 or 3.

> I would prefer to remove the host build of python-six (and use HOST_PYTHON3_PACKAGE_BUILD_DEPENDS instead) but I am not familiar enough with the ovn/openvswitch packages to make this change.

[Funny story: I was the original maintainer of OVS in OpenWrt, before passing it to Yousong].
So, OVS could use `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS`.
But we need to try it again and tune it.
I really don't remember why I did not bother to convert it.
Maybe I decided to let it go before the `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` support and never bothered to spin-up a change for it.

I do remember it required python-six [host-side] during build-time to run some of the Python scripts, otherwise the build fails.
Maybe I'll try again.


Coming back to PyYAML, I would try to use `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` for it.
It may not be quite straightforward  as it sounds, if it needs libyaml.
Host-side builds can be tricky if dependencies are required.
But, it's always better to let `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` handle that build, so that you are sure that the correct Python2/3 interpreter handles the package.
Otherwise, you can risk that the host's Python interpreter is used; that used to be a common risk in the past.

rmelotte:
Thanks a lot for the comments and explanations, they have been very useful!

I realized I could use `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS` to install PyYAML, but since prplMesh is not a python package, I had to manually call `HostPy3PipInstall` (for 19.07) at configure time.

For anyone interested, the changes are at: https://gitlab.com/prpl-foundation/prplos/feed-prpl/-/merge_requests/93/diffs?commit_id=8c1dd813e5e3a26a2d5adae241e7f1a98d95504b

With this approach this PR is not needed anymore, so I'll close it.

PR #17058, opened 2021-11-03
gnutls: don't run aclocal --install

cotequeiroz:
Maintainer: @nmav 
Compile tested: mediatek, master
Run tested: N/A

Description:
Remove the `--install` parameter when running `aclocal`.  The `--install` argument is used to copy third-party files to the first `-I` directory.  
`gnutls` has `-I m4` first, which would copy files to its local build directory.  However, openwrt preppends the staging dir m4 directory, causing `aclocal --install` to copy an old definition of `ax_code_coverage.m4` into the staging dir.

If strace is built after gnutls, compilation will fail:

```
    Makefile:9303: *** missing separator.  Stop.

The version of ax_code_coverage.m4 that gets installed does not define
@CODE_COVERAGE_RULES@.
```
Removing the `--install` parameter in gnutls solves the issue.

@neheb 
I saw you got strace to not use `ax_code_coverage.m4`, which is fine on its own.  However, we may end up having trouble with other outdated files in the future if we don't change this behavior.

### DIFF
--- a/libs/gnutls/Makefile
+++ b/libs/gnutls/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gnutls
 PKG_VERSION:=3.7.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_USE_MIPS16:=0
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
--- a/libs/gnutls/patches/020-dont-install-m4-files.patch
+++ b/libs/gnutls/patches/020-dont-install-m4-files.patch
@@ -1,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Eneas U de Queiroz <cote2004-github@yahoo.com>
+Date: Mon, 25 Oct 2021 08:49:21 -0300
+Subject: Do not install m4 files
+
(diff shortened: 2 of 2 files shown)


neheb:
Interesting... I was wondering where this old copy came from.

cotequeiroz:
Not that I have seen.

cotequeiroz:
> 
> 
> Interesting... I was wondering where this old copy came from.

I thought that it was an outdated autoconf-archive, but then I saw you updated it on September.  I had this patch before that update, and haven't checked it since.  Is it still happening?  I saw your patch to strace and then remembered to push this update.

Nonetheless, we should not allow packages to add third-party m4 files to the system's archive, especially when it is not their intention!

neheb:
Target packages installing stuff to staging_dir/host is indeed a bug. I have not looked at how this could be solved.

nmav (approved):

cotequeiroz:
One way of solving this would be to just filter out `--install` when running `aclocal.real`:
``` diff
--- a/tools/automake/files/aclocal
+++ b/tools/automake/files/aclocal
@@ -1,2 +1,2 @@
-#!/usr/bin/env sh
-aclocal.real $ACLOCAL_INCLUDE $@
+#!/usr/bin/env bash
+aclocal.real $ACLOCAL_INCLUDE "${@#--install}"
```
I imagine the flag is mostly useful to create a release tarball, not to just build the package; moreover, changing the installation path defeats the intention of the author anyway.

PR #17063, opened 2021-11-04
conmon: update to 2.0.30

oskarirauta:
bug fixes:
 - Remove unreachable code path
 - exit: report if the exit command was killed
 - exit: fix race zombie reaper
 - conn_sock: allow watchdog messages through the notify socket proxy
 - seccomp: add support for seccomp notify

misc:
 - Add seccomp to build dependency

included patch removes unnecessary dependency of libdl and also allows a succesfull build

Run tested: x86_64, pc, quite recent snapshot

### DIFF
--- a/utils/conmon/Makefile
+++ b/utils/conmon/Makefile
@@ -1,18 +1,18 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=conmon
-PKG_VERSION:=2.0.29
+PKG_VERSION:=2.0.30
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://github.com/containers/$(PKG_NAME)/archive/v$(PKG_VERSION)
-PKG_HASH:=eb4d5e157671a61b88786e44094775194e30e1d0ad0b9d50035532ece78dbc28
+PKG_HASH:=4b0a98fbe8a63c42f60edac25c19aa6606caa7b1e4fe7846fc7f7de0b566ba25
 
 PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=LICENSE
 
-PKG_BUILD_DEPENDS:=golang/host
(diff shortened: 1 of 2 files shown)


utils/conmon/Makefile
```diff
@@ -1,18 +1,18 @@
...
 PKG_SOURCE_URL:=https://github.com/containers/$(PKG_NAME)/archive/v$(PKG_VERSION)
-PKG_HASH:=eb4d5e157671a61b88786e44094775194e30e1d0ad0b9d50035532ece78dbc28
+PKG_HASH:=4b0a98fbe8a63c42f60edac25c19aa6606caa7b1e4fe7846fc7f7de0b566ba25
 
 PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=LICENSE
 
-PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_DEPENDS:=golang/host libseccomp
```
PolynomialDivision, line 15: I'm not familiar with this package. But are you sure you just need that at build and not runtime? For example your package crun has that as dependency:
https://github.com/openwrt/packages/blob/master/utils/crun/Makefile#L29

oskarirauta: Yes. Build bot here failed because of missing libseccomp.h, so I had to add it as a build dependency. libseccomp is not available for arc, and that's why arc is disabled. Better would be to take parts of libseccomp that are required and integrate them to common, but that would possibly need building that system every time libseccomp updates- so better this way. libseccomp is not needed on runtime, just something from header file.

PolynomialDivision:
thanks. lgtm

jefferyto:
This is unrelated to this update, but I noticed that this package has a build dependency on `golang/host`. The CI passed even for platforms that Go doesn’t support (powerpc), so either it’s not compiling correctly for the target platform (instead compiling for the build platform), or no runtime component written in Go is being compiled (only components needed for build-time).

Looking at the upstream PR that added the Go dependency (https://github.com/containers/conmon/pull/41), it appears that the Go component(s) are only used when conman is a dependency of another Go program.

If this is true, you may want to patch out compiling the Go component(s) instead of including the build dependency on `golang/host`, that may speed up the compilation of this package.

PR #17064, opened 2021-11-04
mariadb, liburing: fix liburing dependency failure (use liburing if kernel supports it)

cotequeiroz:
Maintainer: @miska, @gladiac1337  
Compile tested: mediatek, Linksys EA8450
Run tested: needs run-testing

Description:
This fixes the problem of liburing dependency in mariadb.  If kernel io uring support is turned off, then mariadb will pick  libaio as a fallback.
I had to add the `KERNEL_IO_URING` dependency to liburing, otherwise it would get built for a system that will not support it.  Then mariadb will detect liburing, and we won't be able to build it with libaio without a patch.

Also included is a commit disabling columnstore engine.  At least here, it will pick-up the boost library, and will attempt to configure columnstore, failing.  I noticed that the bots are not trying to build it, even though they did find boost.  Here's the part where columnstore is involved:
```
== MariaDB-Columnstore 5.6.2
-- Could NOT find Boost: missing: thread date_time chrono (found /builder/shared-workdir/build/sdk/staging_dir/hostpkg/lib/cmake/Boost-1.77.0/BoostConfig.cmake (found suitable version "1.77.0", minimum required is "1.53.0"))
-- Required Boost libraries not found!
```

### DIFF
--- a/libs/liburing/Makefile
+++ b/libs/liburing/Makefile
@@ -22,6 +22,7 @@ define Package/liburing
   CATEGORY:=Libraries
   TITLE:=io_uring library
   URL:=https://git.kernel.dk/cgit/liburing
+  DEPENDS:=@KERNEL_IO_URING
 endef
 
 define Package/liburing/description
--- a/utils/mariadb/Makefile
+++ b/utils/mariadb/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mariadb
 PKG_VERSION:=10.6.4
-PKG_RELEASE:=1
+PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL := https://archive.mariadb.org/$(PKG_NAME)-$(PKG_VERSION)/source
@@ -36,6 +36,7 @@ MARIADB_SOCKET=/var/run/mysqld/mysqld.sock
 
 MARIADB_DISABLE_ENGINES := \
 	cassandra \
+	columnstore \
 	example \
 	mroonga \
 	oqgraph \
--- a/utils/mariadb/Makefile
+++ b/utils/mariadb/Makefile
@@ -82,6 +82,7 @@ MARIADB_SERVER_PLUGINS := \
 
 PKG_CONFIG_DEPENDS := \
 	$(patsubst %,CONFIG_PACKAGE_$(PKG_NAME)-server-plugin-%,$(subst _,-,$(MARIADB_SERVER_PLUGINS))) \
+	CONFIG_KERNEL_IO_URING \
 	CONFIG_PACKAGE_mariadb-server
 
 plugin-auth_ed25519             := PLUGIN_AUTH_ED25519
@@ -284,7 +285,8 @@ define Package/mariadb-server-base
   $(call Package/mariadb/Default)
   DEPENDS:=mariadb-common \
 	  $(MARIADB_COMMON_DEPENDS) \
-	  +libaio \
+	  +!KERNEL_IO_URING:libaio \
+	  +KERNEL_IO_URING:liburing \
 	  +liblzma \
 	  +libpcre2 \
 	  +resolveip


oskarirauta:
I had same issue on x86, but finally was not able to compile Mariadb-server package at all and gave up on it then..

cotequeiroz:
See if this fixes it for you, then run-test it if you're able, please.

miska:
I thought I already commented on this, looks great and I'm all for it, unfortunatelly I'm still running 19.07 and need to upgrade my router so can't run test at the moment :-(

cotequeiroz:
I got my HD back, so I may be able to test this.  I've installed it, but can't start it.  I've enabled the server in `/etc/config/mysqld`.  But then I can't start it, or install the system dbs.
```
# /etc/init.d/mysqld start
2021-11-19 15:55:21 0 [Warning] Could not open mysql.plugin table: "Table 'mysql.plugin' doesn't exist". Some options may be missing from the help text
2021-11-19 15:55:22 0 [Warning] Could not open mysql.plugin table: "Table 'mysql.plugin' doesn't exist". Some options may be missing from the help text
2021-11-19 15:55:22 0 [Warning] Could not open mysql.plugin table: "Table 'mysql.plugin' doesn't exist". Some options may be missing from the help text
2021-11-19 15:55:22 0 [Warning] Could not open mysql.plugin table: "Table 'mysql.plugin' doesn't exist". Some options may be missing from the help text
Cannot detect privileges table. You might need to run
'mysql_install_db "--force --basedir=/usr"'
to initialize the system tables.
Then hand it ower to MariaDB user
'chown -Rh "mariadb:mariadb" "/mnt/data/mysql/"'
# mysql_install_db "--force --basedir=/usr"

FATAL ERROR: Could not find mariadbd

The following directories were searched:

    /usr/bin
    /usr/libexec
    /usr/sbin
    /usr/bin

If you compiled from source, you need to either run 'make install' to
copy the software into the correct location ready for operation.
If you don't want to do a full install, you can use the --srcdir
option to only install the mysql database and privilege tables.

If you are using a binary release, you must either be at the top
level of the extracted archive, or pass the --basedir option
pointing to that location.

The latest information about mysql_install_db is available at
https://mariadb.com/kb/en/installing-system-tables-mysql_install_db
```
I've created a symlink, and tried it again, and then without any arguments:
```
# ln -s mysqld /usr/bin/mariadbd
# mysql_install_db "--force --basedir=/usr"
Could not find plugin directory.
Will continue with "/usr/lib/mariadb/plugin".

FATAL ERROR: Could not find /usr/share/mariadb/mysql_sys_schema.sql

If you compiled from source, you need to either run 'make install' to
copy the software into the correct location ready for operation.
If you don't want to do a full install, you can use the --srcdir
option to only install the mysql database and privilege tables.

If you are using a binary release, you must either be at the top
level of the extracted archive, or pass the --basedir option
pointing to that location.

The latest information about mysql_install_db is available at
https://mariadb.com/kb/en/installing-system-tables-mysql_install_db
# mysql_install_db
Could not find plugin directory.
Will continue with "/usr/lib/mariadb/plugin".

FATAL ERROR: Could not find /usr/share/mariadb/mysql_sys_schema.sql

If you compiled from source, you need to either run 'make install' to
copy the software into the correct location ready for operation.
If you don't want to do a full install, you can use the --srcdir
option to only install the mysql database and privilege tables.

If you are using a binary release, you must either be at the top
level of the extracted archive, or pass the --basedir option
pointing to that location.

The latest information about mysql_install_db is available at
https://mariadb.com/kb/en/installing-system-tables-mysql_install_db
```
Do you have any tips on how to set this up and do at least a basic test?

gladiac1337:
Hey there, sorry for the delay.
The liburing changes look good to me. We probably want kernel URING support if we intend to use this library. Thanks for that.
I cannot speak for the other changes though.

cotequeiroz:
Kernel URING support is already default for `!SMALL_FLASH`:

https://github.com/openwrt/openwrt/blob/f84053af5c0b0a159ea4d3e90b0c06574b4fde8d/config/Config-kernel.in#L429-L431

If this package is compiled without kernel `URING` support, it will fallback to libaio.

oskarirauta:
Works great on x86_64

oskarirauta (approved):
Works on x86_64

PR #8870, opened 2019-05-01
zstd: Fix compilation on uClibc-ng

neheb:
uClibc-ng defines __GLIBCC__ and does not compile in execinfo.h. Disable
it.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @ammubhave 
Compile tested: arc700

Issue #8858, opened 2019-04-30
zstd does not compile with uClibc-ng


neheb:
Maintainer: @ammubhave 

https://downloads.openwrt.org/snapshots/faillogs/arc_arc700/packages/zstd/compile.txt

The issue is that uClibc-ng in OpenWrt does not get compiled with backtrace support. Meaning execinfo.h does not get installed. A small look at the code indicates it's optional.

PR #8851, opened 2019-04-30
sane-backends: rebind printers to usblp after scan

luizluca:
While scanning, sane unbinds usb device from usblp module. This
patch creates a saned wrapper that keep track of all usb devices
binded to usblp before saned is launched and rebind them back
if not still bound after saned exits.

This workaround is only effective if saned is launched by a super
daemon like xinetd as it needs to exit after scan job finish. If
not, printing will not be available while saned service is running.

Write access to /sys/bus/usb/drivers/usblp/bind is required. So,
normally it would require root access.

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

Maintainer: me
Compile tested: ar71xx, arc_arc700, arm_xscale, x86, x86_64 (although there is no change in binaries)
Run tested: ar71xx

Description:

There is a conflict between user-mode USB drivers (SANE, CUPS) and usblp (used by p910nd). When saned runs while usblp is loaded, it unbinds the USB device (removing device /dev/usb/lp*) from the kernel driver. However, once saned exists, usblp does not reclaim it back. This way, p910nd cannot print after someone has scanned.

The previous workaround approach for this problem is to create a saned wrapper that unloads usblp module and reload it after saned exits. However, removing usblp will stop any printer from working during the scanning, even those not used by SANE. Also, some saned uses might not even touch the printer, like in listing devices. And besides, unloading/loading modules is a critical operation.

This patch also creates a wrapper but, instead of removing the module, it keeps track of every USB device attached to the driver before saned runs. If any of those are not bound to a driver after saned exists, it rebinds it to usblp.

It does need root access (or, at least write access to /sys/bus/usb/drivers/usblp/bind). If someone wants to run saned with a different user/group, it is possible that some hotplug magic would be enough to assign write access to saned user/group.

@cshoredaniel , would you please take a look if this alternative approach is enough for your needs. Feel free to suggests improvements, specially if it helps others run saned as a non-root user.

Another alternative approach would be to patch saned code. saned claims and release USB device, even detecting when kernel driver is using it:

```
Tue Apr 30 02:01:51 2019 daemon.info saned[18698]: io/hpmud/musb.c 427: Found interface conf=0, iface=1, altset=0, index=1
Tue Apr 30 02:01:51 2019 daemon.info saned[18698]: io/hpmud/musb.c 389: Active kernel driver on interface=1 ret=1
[...]

PR #8848, opened 2019-04-30
libre2: bump version to 2019-04-01

ammubhave:
bump version of libre2 to 2019-04-01 which is latest.

compile tested on nbg6817

Signed-off-by: Amol Bhave <ambhave@fb.com>

PR #17067, opened 2021-11-04
docker-compose: Update to version 2.1.0

jmarcet:
Maintainer: me
Compile tested: master x86_64
Run tested: master x86_64

Description:
Major upstream update, switching from Python to Golang.
I missed the previous one a few days ago.

Since I've had to redo the Makefile, I am not entirely sure I did it
in an optimal way. It is built locally, without resorting to docker
or git like the upstream Makefile tries to do.

It is now installed to `/usr/lib/docker/cli-plugins/docker-compose`
and symlinked to `/usr/bin/docker-compose`. That way it can now
be run as `docker-compose` or as `docker compose`.

Also, I wonder what to do with all the old python dependencies,
namely:

`python-cached-property, python-distro, python-docker,
python-dockerpty, python-docopt, python-dotenv,
python-jsonschema, python-paramiko, python-pynacl,
python-texttable & python-websocket-client`

I'm the current maintainer of all those packages, after this update
I'm dropping them from my regular builds since I only used them
with this package.

This update has recreated all my containers, and the only caveat
was I had to enclose in `""` a couple variables in an `.env` file.
Other than that, it has smoothly replaced the python version.

Docker Compose V2 is a major version bump release of Docker Compose.
It has been completely rewritten from scratch in Golang (V1 was in
Python).

Features:

 - Added support for running the plugin as a standalone program. Like
 docker-compose v1 would behave compute sha256 checksums while releasing

Bugfixes:

 - Allow combination of --status and --services
 - Fix build cache_from option
 - Fix compose up on README.md
 - Make --status a multi-flag
 - No longer fail when inferred .env is a directory

Misc

 - Stop ticker after use on ttyWriter
 - Use uname -m for cross platform suffixes
 - Add note about installing it system-wide
 - Bump containerd 1.5.5

What's Changed

 - Fix support for devices by @ndeloof in #8732
 - Make command descriptions consistent by @mat007 in #8739
 - Restore missing version commands by @Shikachuu in #8738
 - Add step in README to install on linux by @Yopadd in #8755
 - log --follow must stop when container get killed by @ndeloof in
 #8726
 - Fix index out of range on compose.buildContainerMountOptions by
 @ulyssessouza in #8750
 - Pass runtime option to containerCreate by @ndeloof in #8783
 - Fix compose down --timeout/-t flag by @debdutdeb in #8788
 - Fix network_mode "service:x" by @ulyssessouza in #8792
 - Make service>build>dockerfile a simple filename by @ulyssessouza
 in #8779
 - Compose exec cannot process more than 32KB of data by @resios in
 #8815
 - Actually fix Compose exec cannot process more than 32KB of data by
 @resios in #8816
 - Fix project settings' options order by @ulyssessouza in #8819
 - Update link to Docker Community Slack by @mat007 in #8824
 - Add support for DOCKER_DEFAULT_PLATFORM by @ndeloof in #8848
 - COMPOSE_COMPATIBILITY can be set by .env file by @ndeloof in #8847
 - Add support for EnableIPv6 by @ndeloof in #8851
 - Avoid test flakyness by ordering volumes before checking by
 @ulyssessouza in #8858
 - Update README.md: typographical edit of "About update..." by
 @youssefeldakar in #8838
 - Add support for classic builder by @ulyssessouza in #8818
 - Fix typo: netwok -> network by @Mygao in #8789
 - Bump compose-go to v1.0.5 by @ulyssessouza in #8870

### DIFF
--- a/utils/docker-compose/Makefile
+++ b/utils/docker-compose/Makefile
@@ -1,49 +1,47 @@
 include $(TOPDIR)/rules.mk
 
-PKG_NAME:=docker-compose
-PKG_VERSION:=1.29.2
+PKG_NAME:=compose
+PKG_VERSION:=2.1.0
 PKG_RELEASE:=$(AUTORELEASE)
-
-PYPI_NAME:=docker-compose
-PKG_HASH:=4c8cd9d21d237412793d18bd33110049ee9af8dab3fe2c213bbd0733959b09b7
-
-PKG_MAINTAINER:=Javier Marcet <javier@marcet.info>
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=LICENSE
 
-include ../../lang/python/pypi.mk
-include $(INCLUDE_DIR)/package.mk
-include ../../lang/python/python3-package.mk
+PKG_SOURCE:=v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/docker/compose/tar.gz/v${PKG_VERSION}?
+PKG_HASH:=48b20b1b75de0e517deb0e29122ecdb01982f136e68429bdf6a97121701d4274
 
-PYTHON3_PKG_SETUP_ARGS:=
+PKG_MAINTAINER:=Javier Marcet <javier@marcet.info>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/docker/compose/v2
+GO_PKG_BUILD_PKG:=github.com/docker/compose/v2/cmd
+GO_PKG_LDFLAGS_X:=github.com/docker/compose/v2/internal.Version=v$(PKG_VERSION)
+GO_PKG_TAGS:=e2e,kube
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
 
 define Package/docker-compose
   SECTION:=utils
   CATEGORY:=Utilities
   TITLE:=Docker Compose
   URL:=https://github.com/docker/compose
-  DEPENDS+=+docker \
-      +python3-light \
-      +python3-cached-property \
-      +python3-distro \
-      +python3-distutils \
-      +python3-docopt \
-      +python3-docker \
-      +python3-dockerpty \
-      +python3-dotenv \
-      +python3-jsonschema \
-      +python3-logging \
-      +python3-openssl \
-      +python3-pkg-resources \
-      +python3-requests \
-      +python3-texttable \
-      +python3-yaml
+  DEPENDS:=$(GO_ARCH_DEPENDS) +docker
 endef
 
 define Package/docker-compose/description
   Multi-container orchestration for Docker
 endef
 
-$(eval $(call Py3Package,docker-compose))
+define Package/docker-compose/install
+	$(INSTALL_DIR) $(1)/usr/lib/docker/cli-plugins/
+	$(INSTALL_BIN) $(GO_PKG_BUILD_BIN_DIR)/cmd $(1)/usr/lib/docker/cli-plugins/docker-compose
+	$(INSTALL_DIR) $(1)/usr/bin/
+	$(LN) ../lib/docker/cli-plugins/docker-compose $(1)/usr/bin/docker-compose
+endef
+
+$(eval $(call GoBinPackage,docker-compose))
 $(eval $(call BuildPackage,docker-compose))
-$(eval $(call BuildPackage,docker-compose-src))


jefferyto:
I suggest taking a look at [other packages that use golang-package.mk](https://github.com/openwrt/packages/search?q=golang-package) to see how they do things. (I don’t see a strong need to override `Build/Compile` or `Build/Install`.)

In particular, you can set `GO_PKG_LDFLAGS_X` for `github.com/docker/compose/v2/internal.Version` (you can set it to the git tag, e.g. `v2.1.0`), and set `GO_PKG_TAGS` for build tags. (There is no need for `-s -w` ldflags and PIE build mode is automatically handled.)

jmarcet:
> GO_PKG_TAGS

Much cleaner now, `GO_PKG_BUILD_PKG` was key, I hadn't seen it earlier.

jmarcet:
What am I missing? I can build it with no issues locally.

utils/docker-compose/Makefile
```diff
@@ -1,49 +1,47 @@
...
 endef
 
 define Package/docker-compose/description
   Multi-container orchestration for Docker
 endef
 
-$(eval $(call Py3Package,docker-compose))
+define Package/docker-compose/install
+	$(INSTALL_DIR) $(1)/usr/lib/docker/cli-plugins/
+	$(INSTALL_BIN) $(GO_PKG_BUILD_DIR)/bin/cmd $(1)/usr/lib/docker/cli-plugins/docker-compose
```
jefferyto, line 41:
```suggestion
	$(INSTALL_BIN) $(GO_PKG_BUILD_BIN_DIR)/cmd $(1)/usr/lib/docker/cli-plugins/docker-compose
```

or

```suggestion
	$(call GoPackage/Package/Install/Bin,$(PKG_INSTALL_DIR))
	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/cmd $(1)/usr/lib/docker/cli-plugins/docker-compose
```

paper42:
> Also, I wonder what to do with all the old python dependencies, namely:
> 
> `python-cached-property, python-distro, python-docker, python-dockerpty, python-docopt, python-dotenv, python-jsonschema, python-paramiko, python-pynacl, python-texttable & python-websocket-client`
> 
> I'm the current maintainer of all those packages, after this update I'm dropping them from my regular builds since I only used them with this package.

We are using python-distro and python-jsonschema in TurrisOS, so I will adopt these packages if you are not interested in them anymore.

utils/docker-compose/Makefile
```diff
@@ -1,49 +1,47 @@
...
 
-PYPI_NAME:=docker-compose
-PKG_HASH:=4c8cd9d21d237412793d18bd33110049ee9af8dab3fe2c213bbd0733959b09b7
+PKG_SOURCE:=v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/docker/compose/tar.gz/v${PKG_VERSION}?
+PKG_HASH:=48b20b1b75de0e517deb0e29122ecdb01982f136e68429bdf6a97121701d4274
 
 PKG_MAINTAINER:=Javier Marcet <javier@marcet.info>
-PKG_LICENSE:=Apache-2.0
-PKG_LICENSE_FILES:=LICENSE
```
jefferyto, line 12: Others may prefer `PKG_LICENSE`/`PKG_LICENSE_FILES` stay here, I don’t have a strong opinion on this.

jefferyto (approved):
LGTM - thanks!

jmarcet:
> > Also, I wonder what to do with all the old python dependencies, namely:
> > `python-cached-property, python-distro, python-docker, python-dockerpty, python-docopt, python-dotenv, python-jsonschema, python-paramiko, python-pynacl, python-texttable & python-websocket-client`
> > I'm the current maintainer of all those packages, after this update I'm dropping them from my regular builds since I only used them with this package.
> 
> We are using python-distro and python-jsonschema in TurrisOS, so I will adopt these packages if you are not interested in them anymore.

It's more about proper testing them when they're updated than upgrading the packages. I do not have any issue with the later, but I'm somewhat concerned about the former.

utils/docker-compose/Makefile
jmarcet: It's true I moved those two lines up while I did the move to a Go package, mimicking [runc](https://github.com/openwrt/packages/blob/master/utils/runc/Makefile).

While I do not see any reference to this matter in the [package-policies](https://openwrt.org/docs/guide-developer/package-policies), checking the [packages](https://github.com/openwrt/packages) tree reveals most of the packages have the three `PKG_` variables together.

I kind of liked them separate since it looks like the license is related somehow to the maintainer, while it is an upstream thing entirely, just like the version or the homepage.

So, for me I'd leave it as it is, if anyone else cares and really prefers to keep it like the majority of packages, please tell me and I'll change it.

jmarcet:
I saw #17049 and I liked how it used `$(LN) ..` instead of `ln -s /usr/` so I changed it.

PR #17068, opened 2021-11-04
crun: update to version 1.2

oskarirauta:
release notes:

0.20.1
 - container: ignore error when resetting the SELinux label for the keyring.

0.21
 - when compiled with krun, automatically use it if the current executable file is called "krun"
 - cgroup: lookup pids controller as well when the memory controller is not available
 - status: add fields for owner and created timestamp
 - honor memory swappiness set to 0

1.0
 - Fix symlink target mangling for tmpcopyup targets.
 - Makefile.am: fix link error when using directly libcrun.
 - cgroup: add support for setting memory.use_hierarchy on cgroup v1.
 - linux: treat pidfd_open failures EINVAL as ESRCH.
 - cgroup: chown the current container cgroup to root in the container.

1.1
 - utils: retry openat2 on EAGAIN. If the openat2 syscall is interrupted, try again.
 - criu: fix save of external descriptors. Now restored containers attach correctly their standard streams.
 - criu: Add support for external PID namespace.
 - container: Set primary process to 1 via LISTEN_PID by default if user configuration is missing.
 - exec: refuse to exec in a paused container/cgroup.
 - cgroup: use cgroup.kill when available. It is faster to kill a container through its cgroup as there is no need to recurse over the cgroup pids and terminate each one of them.

1.2
 - criu: add support for external ipc, uts and time namespaces.
 - exec: fix regression in 1.1 where containers are being wrongly reported as paused.

Run tested: x86_64, pc, recent snapshot


PolynomialDivision:
Do you maybe want to takeover maintainership? Just looked at commit history and you seem very active:
https://github.com/openwrt/packages/commits/master/utils/crun

PolynomialDivision:
lgtm. thanks

PR #17070, opened 2021-11-04
zerotier: update to 1.8.1

oskarirauta:
Release notes:

1.8.0
 - Upgrade json.hpp dependency to version 3.10.2
 - Check if DNS servers need to be applied on macOS
 - Set MAC address before bringing up Linux TAP link
 - Stop binding to temporary IPv6 addresses
 - Fix for mistakenly using v6 source addresses for v4 routes on some platforms
 - Fix for MacOS MTU capping issue on feth devices
 - Implement a workaround for one potential source of a "coma" bug, which can occur if buggy NATs/routers stop allowing the service to communicate on a given port. ZeroTier now reassigns a new secondary port if it's offline for a while unless a secondary port is manually specified in local.conf. Working around crummy buggy routers is an ongoing effort.
 - A completely rewritten desktop UI for Mac and Windows!

1.8.1
 - Fix an issue that could cause clobbering of MacOS IP route settings on restart.
 - Added additional hardening against address impersonation on networks (also in 1.6.6).
 - MacOS IPv6 no longer binds to temporary addresses as these can cause interruptions if they expire.
 - Remove support for REALLY ancient 1.1.6 or earlier network controllers.
 - Fix numerous UI issues from 1.8.0 (never fully released).

Changed GIT as source and added $(AUTORELEASE)

Maintainer: Moritz Warning / @mwarning
Run tested: x86_64, pc, recent snapshot


PolynomialDivision:
You further add autorelease and change to git as source. Maybe also say that in commit message. :)

mwarning:
thank you, looks good, LGTM

oskarirauta:
> You further add autorelease and change to git as source. Maybe also say that in commit message. :)

done

oskarirauta:
> thank you, looks good, LGTM

works great too. With previous versions, sometimes all nodes were not connectable between some others. This has not failed ever since after upgrading.

PolynomialDivision:
thanks. will merge after ci finishes.

PR #17075, opened 2021-11-05
v4l2tools: fix missing dependency

PolynomialDivision:
Should fix errors in the form of:

    Package v4l2tools is missing dependencies for the following libraries: libvpx.so.6

Maintainer: @mpromonet
Compile tested: tbd
Run tested: tbd

### DIFF
--- a/multimedia/v4l2tools/Makefile
+++ b/multimedia/v4l2tools/Makefile
@@ -26,7 +26,7 @@ define Package/v4l2tools
 	SECTION:=multimedia
 	CATEGORY:=Multimedia
 	TITLE:=v4l2tools
-	DEPENDS:=+libstdcpp +libjpeg
+	DEPENDS:=+libstdcpp +libjpeg +libvpx
 	URL:=https://github.com/mpromonet/v4l2tools
 endef
 


PolynomialDivision:
I did not runtest only compile test.

mpromonet:
Hi,

The project v4l2tools can compile with vpx or not (it is the same with x264/x265). The build problem seems CMake detect libvpx is available (using pkg-tools) but openwrt build not have in dependency. It may be nice to have a conditional dependency ?

Best Regards,
Michel.

PolynomialDivision:
Any suggestion how to do this?

PolynomialDivision:
@mpromonet Just to be sure, you mean that libvpx is available from the host?

BKPepe:
@PolynomialDivision Please, don't blindly add more dependencies to packages when they are autodetected to just try to fix their compilation. They can be disabled. If you add a dependency, then package sizes are going to be increased. It could somehow change the current behavior of the package, and users don't expect it. You need to verify if it works. 

TL;DR: Let's imagine that you added a new dependency that is autodetected. You need to consider ``pros`` and ``cons`` to add it or disable it. Sometimes it can be handy, sometimes not. If it is added, you need to verify that the package is working, but you need to check also if the new feature, which depends on the package you added, works. Sometimes, it does not worth the hassle, and it is good to disable it. Usually, you don't need all the features which the package offers.

PolynomialDivision:
Yeah sorry. We had this update of ipq40xx to 5.10er kernel. And I wanted to go through fail logs and look whats wrong and fails. Now I'm realizing I dig out already existing failures.

mpromonet (commented):
It may be better to use : 
`	DEPENDS:=+libstdcpp +libjpeg +PACKAGE_libvpx:libvpx`

stintel:
> It may be better to use :
> ` DEPENDS:=+libstdcpp +libjpeg +PACKAGE_libvpx:libvpx`

The best way is to add an option in the v4l2tools build to disable libvpx, and use that in the OpenWrt Makefile. Like this, packages build by the buildbots, and thus the packages in the repo, will still have the dependency.

PR #17076, opened 2021-11-05
node: fix compile error

PolynomialDivision:
Should fix errors in the form of

    fatal error: bits/c++config.h: No such file or directory

Maintainer: @nxhack
Compile tested: tbd
Run tested: tbd

### DIFF
--- a/lang/node/Makefile
+++ b/lang/node/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=node
 PKG_VERSION:=v14.18.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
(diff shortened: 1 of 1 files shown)


BKPepe:
Closing. This is the wrong approach. More details are in linked issues. You should have installed these two packages on your host system: ``gcc-multilib g++-multilib``, what are you trying to do is just trying to work around or solve it **how it should not be**. 

What is the best way is to check the Docker, which I did for luajit, if it compiles.
https://github.com/openwrt/docker/commit/713696a6a76dfa6a19e96aa827c956972c5c5e5b

If not, send the pull request to the Docker repository and then as well try to send the patch to [the buildbot repository](https://git.openwrt.org/?p=buildbot.git;a=summary) like it was done here https://git.openwrt.org/?p=buildbot.git;a=commit;h=56f05546ab38e0675586c853503a5013c39a0a32

PolynomialDivision:
Thanks. I will send patch to docker.

PR #16823, opened 2021-10-06
openblas: fix compilation with mips x4k series CPU

cotequeiroz:
Maintainer: @commodo 
Compile tested: ath79/TP-Link Archer C6, built with `CONFIG_CPU_TYPE="74kc"`, on master.
Run tested: same arch, using a small test program from https://www.hpc.ntnu.no/idun/software-on-idun/openblas/
``` C
/*  blas_test.c : Example using BLAS matrix multiply code sgemm */

#include <stdio.h>
#include <cblas.h>

int main(void)
{
  const int lda=3, ldb=3, ldc=3;
  int m, n, k;
  float alpha, beta;

  float a[] = { 0.11, 0.21, 0.12,
                0.15, 0.13, 0.17,
                0.22, 0.13, 0.23 };

  float b[] = { 1011, 1021, 1018,
                1031, 1012, 1021,
                1022, 1032, 1015 };

  float c[] = { 0.00, 0.00, 0,00,
                0.00, 0.00, 0,00,
                0.00, 0.00, 0,00 };

  m = 3; n = 3; k = 3;

  alpha = 1.0; beta = 0.0;

  cblas_sgemm (CblasColMajor, CblasNoTrans, CblasNoTrans,
                m, n, k, alpha, a, lda, b, ldb, beta, c, ldc);

  printf ("[ %g, %g, %g\n", c[0], c[1], c[2]);
  printf ("  %g, %g, %g\n", c[3], c[4], c[5]);
  printf ("  %g, %g, %g ]\n", c[6], c[7], c[8]);

  return 0;
}
```

Description:
Anything later than MIPS 24k can run MIPS 24k code.  Set OPENBLAS_TARGET
to MIPS24K in those cases.

Signed-off-by: Eneas U de Queiroz <cotequeiroz@gmail.com>

It should fix current build errors for mipsel_74kc, mipsel_24kc_24kf, and of course, if you have a custom build on mips_74kc, like I just did 😜

PR #17077, opened 2021-11-05
openblas: default to MIPS24K target for all mips32 targets

commodo:
Maintainer: me
Compile tested: master mips32 (Broadcom BMIPS)
Run tested: n/a

----------------------------------------------------

Reported via:
  https://github.com/openwrt/packages/pull/16823#issuecomment-945312460

This fails in mips_mip32 targets with the output listed below.

Using the MIPS24K target works fine.

```
mips-openwrt-linux-musl-gcc -c -Os -pipe -mno-branch-likely -mips32 -mtune=mips32 -fno-caller-saves -fno-plt -fhonour-copts -Wno-error=unused-but-set-variable -Wno-error=unused-result -msoft-float -ffile-prefix-map=/builder/shared-workdir/build/sdk/build_dir/target-mips_mips32_musl/OpenBLAS-0.3.18=OpenBLAS-0.3.18 -Wformat -Werror=format-security -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro  -I/builder/shared-workdir/build/sdk/staging_dir/toolchain-mips_mips32_gcc-11.2.0_musl/usr/include -I/builder/shared-workdir/build/sdk/staging_dir/toolchain-mips_mips32_gcc-11.2.0_musl/include/fortify -I/builder/shared-workdir/build/sdk/staging_dir/toolchain-mips_mips32_gcc-11.2.0_musl/include   -DMAX_STACK_ALLOC=2048 -Wall -mabi=32 -DF_INTERFACE_GFORT -fPIC -DNO_LAPACK -DNO_LAPACKE -DNO_AVX512 -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=2 -DMAX_PARALLEL_NUMBER=1 -DBUILD_SINGLE=1 -DBUILD_DOUBLE=1 -DBUILD_COMPLEX=1 -DBUILD_COMPLEX16=1 -DVERSION=\"0.3.18\" -UASMNAME -UASMFNAME -UNAME -UCNAME -UCHAR_NAME -UCHAR_CNAME -DASMNAME= -DASMFNAME=_ -DNAME=_ -DCNAME= -DCHAR_NAME=\"_\" -DCHAR_CNAME=\"\" -DNO_AFFINITY -I.  -DMAX_STACK_ALLOC=2048 -Wall -mabi=32 -DF_INTERFACE_GFORT -fPIC -DNO_LAPACK -DNO_LAPACKE -DNO_AVX512 -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=2 -DMAX_PARALLEL_NUMBER=1 -DBUILD_SINGLE=1 -DBUILD_DOUBLE=1 -DBUILD_COMPLEX=1 -DBUILD_COMPLEX16=1 -DVERSION=\"0.3.18\" -UASMNAME -UASMFNAME -UNAME -UCNAME -UCHAR_NAME -UCHAR_CNAME -DASMNAME=sgemm -DASMFNAME=sgemm_ -DNAME=sgemm_ -DCNAME=sgemm -DCHAR_NAME=\"sgemm_\" -DCHAR_CNAME=\"sgemm\" -DNO_AFFINITY -I.. -I. -UDOUBLE  -UCOMPLEX gemm.c -o sgemm.o
In file included from ../common.h:581,
                 from gemm.c:41:
gemm.c: In function 'sgemm_':
../param.h:3477:25: error: 'sgemm_p' undeclared (first use in this function); did you mean 'sgemm_'?
 3477 | #define SGEMM_DEFAULT_P sgemm_p
      |                         ^~~~~~~
../common_param.h:1334:25: note: in expansion of macro 'SGEMM_DEFAULT_P'
 1334 | #define SGEMM_P         SGEMM_DEFAULT_P
      |                         ^~~~~~~~~~~~~~~
../common_param.h:1482:33: note: in expansion of macro 'SGEMM_P'
 1482 | #define GEMM_P                  SGEMM_P
      |                                 ^~~~~~~
gemm.c:494:37: note: in expansion of macro 'GEMM_P'
  494 |   sb = (XFLOAT *)(((BLASLONG)sa + ((GEMM_P * GEMM_Q * COMPSIZE * SIZE + GEMM_ALIGN) & ~GEMM_ALIGN)) + GEMM_OFFSET_B);
      |                                     ^~~~~~
../param.h:3477:25: note: each undeclared identifier is reported only once for each function it appears in
 3477 | #define SGEMM_DEFAULT_P sgemm_p
      |                         ^~~~~~~
../common_param.h:1334:25: note: in expansion of macro 'SGEMM_DEFAULT_P'
 1334 | #define SGEMM_P         SGEMM_DEFAULT_P
      |                         ^~~~~~~~~~~~~~~
../common_param.h:1482:33: note: in expansion of macro 'SGEMM_P'
 1482 | #define GEMM_P                  SGEMM_P
      |                                 ^~~~~~~
gemm.c:494:37: note: in expansion of macro 'GEMM_P'
  494 |   sb = (XFLOAT *)(((BLASLONG)sa + ((GEMM_P * GEMM_Q * COMPSIZE * SIZE + GEMM_ALIGN) & ~GEMM_ALIGN)) + GEMM_OFFSET_B);
      |                                     ^~~~~~
make[5]: *** [Makefile:1295: sgemm.o] Error 1
make[5]: Leaving directory '/builder/shared-workdir/
```

### DIFF
--- a/libs/openblas/Makefile
+++ b/libs/openblas/Makefile
@@ -58,13 +58,9 @@ ifeq ($(ARCH),aarch64)
 else ifeq ($(ARCH),arm)
   OPENBLAS_TARGET:=ARMV5
 else ifeq ($(ARCH),mips)
-  ifneq ($(filter 24k% 74k%,$(CPU_TYPE)),)
-    OPENBLAS_TARGET:=MIPS24K
-  endif # CPU_TYPE == 24k* or 74k*
+  OPENBLAS_TARGET:=MIPS24K
 else ifeq ($(ARCH),mipsel)
-  ifneq ($(filter 24k% 74k%,$(CPU_TYPE)),)
-    OPENBLAS_TARGET:=MIPS24K
-  endif # CPU_TYPE == 24k* or 74k*
+  OPENBLAS_TARGET:=MIPS24K
 else ifeq ($(ARCH),powerpc)
   OPENBLAS_TARGET:=PPC440
 endif


commodo:
also signalling @cotequeiroz @neheb @brada4  here

brada4:
The output looks like empty KERNEL.$TARGET was used. Since you force existing target it should not fail at compiling. If you get sigill on a real CPU you need to patch assembly out from KERNEL.MIPS24K.

cotequeiroz:
I have no straight mips32 machines--or at least any that will not run mips24k code, so the following statements are all based on theorical observation.

I've just ran a build for mips 74kc. I just can't find where the msa sources are built.  From the gcc manual, in order to generate msa code, we would need to be compiling it with at least without `-msoft-float`

>
>GCC provides intrinsics to access the SIMD instructions provided by the MSA MIPS SIMD Architecture.  The interface is made available by including '<msa.h>' and using '-mmsa -mhard-float -mfp64 -mnan=2008'.

Here's my build log:
https://zerobin.net/?0a160776a9794bb8#BoeG5RpbbYqNblLDbMgyloM37B8A5DnUM4hvGnFuCo4=

Notice the `CFLAGS`  has `-msoft-float`, and none of the other flags mentioned in the gcc manual.  By looking at the code, the generation of MSA code is guarded by `HAVE_MSA`, which is not defined anywhere.

I also displayed the output of the msa test in `c_check`, and the check failure message confirms my suspicion:
```
cc1: error: '-mmsa' must be used with '-mfp64' and '-mhard-float'
```
That only confirms that no msa instructions will be generated from C code.  By looking at the sources, I can see that the code uses the GCC interface, not asm directly.  There are very few straight asm code, so I'm assuming it will respect the arch passed in `CFLAGS`.  I can't be 100% sure, but it appears to work with any MIPS32 CPU, as long as `-march` is set correctly `-mips3r2` and friends are a shortcut for `-march=`.  The `-mtune` override will not add foreign instruction, just CPU optimization.

My conclusion: @commodo's assessment appears to be valid.  Even though the openblas statement is that it shoul run only on 24kc and later CPUs, there does not appear to be any code that will not run on earlier mips CPUs.  Again, I don't have any mips machine that won't run mips24k code to verify my theory. 🤷‍♂️

cotequeiroz:
I forgot to post my `Makefile.conf`:
```
OSNAME=Linux
ARCH=mips
C_COMPILER=GCC
BINARY64=
BINARY32=1
CROSS_SUFFIX=mips-openwrt-linux-musl-
CROSS=1
CEXTRALIB=-L -L/usr/local/src/openwrt/staging_dir/toolchain-mips_74kc_gcc-11.2.0_musl/lib/gcc/mips-openwrt-linux-musl/11.2.0 -L/usr/local/src/openwrt/staging_dir/toolchain-mips_74kc_gcc-11.2.0_musl/lib/gcc/mips-openwrt-linux-musl/11.2.0/../../../../mips-openwrt-linux-musl/lib  -lssp_nonshared -lc
F_COMPILER=GFORTRAN
FC=gfortran
BU=_
NOFORTRAN=1
CORE=MIPS24K
LIBCORE=mips24K
NUM_CORES=12
MIPS24K=1
L1_DATA_SIZE=32768
L1_DATA_LINESIZE=32
L2_SIZE=32768
L2_LINESIZE=32
DTB_DEFAULT_ENTRIES=64
DTB_SIZE=4096
L2_ASSOCIATIVE=8
SBGEMM_UNROLL_M=8
SBGEMM_UNROLL_N=4
SGEMM_UNROLL_M=2
SGEMM_UNROLL_N=2
DGEMM_UNROLL_M=2
DGEMM_UNROLL_N=2
QGEMM_UNROLL_M=2
QGEMM_UNROLL_N=2
CGEMM_UNROLL_M=2
CGEMM_UNROLL_N=2
ZGEMM_UNROLL_M=2
ZGEMM_UNROLL_N=2
XGEMM_UNROLL_M=2
XGEMM_UNROLL_N=2
CGEMM3M_UNROLL_M=2
CGEMM3M_UNROLL_N=2
ZGEMM3M_UNROLL_M=2
ZGEMM3M_UNROLL_N=2
XGEMM3M_UNROLL_M=2
XGEMM3M_UNROLL_N=2
```

brada4:
There are some .S files. Unlike on x86, on mips GCC would also filter generated assembly for foreign instructions.

brada4:
Check that thread, it suggests starting off with P5600 instead.

martin-frbg:
@cotequeiroz do you get a /proc/cpuinfo on that system, and if so, does it show "msa" or similar anywhere ? Currently OpenBLAS treats the MSA capability as a property of the compiler rather than the actual hardware (for whatever "historical" reason) - see the c_check perl script that is run early in the build sequence.

cotequeiroz:
Here's a more detailed account of my findings:

None of the mips *4k series support MSA.  They all belong to the [MIPS Classic Processor Cores family](https://www.mips.com/products/classic/), which are all MIPS32R2.

`MSA` was implemented in mip32r5, while the mips classic processors implement the older DSP ASE (deprecated in MIPS32R5).  The "optional" "module/component/extensions" terms make it complicated for me to figure out what can be done in each architecture.  I can look at the code, though, and here are my findings:

None of the `.S` files are compiled for mips32.  They are `./cpuid.S`, which is x86-only, then many in `kernel/%CPU%`, where `CPU` can be `alpha,arm,arm64,loongarch64,mips64,power,sparc,x86,x86_64,zarch`.  There are no `.S` files under `kernel/mips`.  The SIMD code there is all implemented using gcc intrinsics:
https://gcc.gnu.org/onlinedocs/gcc/MIPS-SIMD-Architecture-_0028MSA_0029-Support.html
Those sources are not built though, because we are using soft-float, [see the `c_check` output above](https://github.com/openwrt/packages/pull/17077#issuecomment-962124176).

Here's my search of `__asm` in C code that's not specific for other platforms (unless there's code for one architecture in a file whose name contains another arch 🤷‍♂️):
```
src/OpenBLAS $ grep -lr __asm | grep -Ev 'mips64|power|sparc|arm|loongarch64|ia64|x86|zarch'
cmake/system_check.cmake
common_mips.h
c_check
driver/others/dynamic.c
driver/others/blas_server.c
driver/others/blas_server_win32.c
driver/others/blas_server_omp.c
common.h
kernel/mips/dgemm_kernel_8x4_msa.c
kernel/mips/sgemm_kernel_8x8_msa.c
kernel/mips/zgemm_kernel_4x4_msa.c
kernel/mips/cgemm_kernel_8x4_msa.c
kernel/mips/macros_msa.h
```
- `cmake/system_check.cmake` usage of asm is for x86-only.

- Files in `driver/others` are x86-only:  `dynamic.c` is buit only when `DYNAMIC_ARCH=1`, which is not compatible with mips; `blas_server*` asm code is guarded by `CONSISTENT_FPCSR`, which is x86-specific.

- Code  in `common_mips.h` is [not compiled for `MIPS24K`](https://github.com/xianyi/OpenBLAS/blob/2d7ca63e21e97dcd2db5c3aa7658f6cc2a504e55/common_mips.h#L46-L57).

- `c_check` is testing for `MSA` and `AVX512`.

-  Code in `common.h` is definitely not MSA 😉 
```
src/OpenBLAS $ grep __asm common.h
#define YIELDING        __asm__ __volatile__ ("nop;nop;nop;nop;nop;nop;nop;nop; \n");
#define YIELDING        __asm__ __volatile__ ("nop;nop;nop;nop;nop;nop;nop;nop;\n");
#define YIELDING        __asm__ __volatile__ ("nop;nop;nop;nop;nop;nop;nop;nop;\n");
#define YIELDING        __asm__ __volatile__ ("nop;nop;nop;nop;nop;nop;nop;nop;\n");
#define YIELDING        __asm__ __volatile__ ("nop;nop;nop;nop;nop;nop;nop;nop;\n");
```

- Files in `kernel/mips/*_msa.c` are not compiled unless `HAVE_MSA` is defined.  They are all guarded by `ifdef HAVE_MSA` in [`kernel/mips/KERNEL.P5600`](https://github.com/xianyi/OpenBLAS/blob/2d7ca63e21e97dcd2db5c3aa7658f6cc2a504e55/kernel/mips/KERNEL.P5600#L38-L261).  Finally, `HAVE_MSA` is [defined](https://github.com/xianyi/OpenBLAS/blob/2d7ca63e21e97dcd2db5c3aa7658f6cc2a504e55/c_check#L385) if the test in [c_check](https://github.com/xianyi/OpenBLAS/blob/2d7ca63e21e97dcd2db5c3aa7658f6cc2a504e55/c_check#L200-L222) passes, which it does not.

So that explains why it builds and runs correctly for any mips32 processor.

My conclusion is that MIPS24K is just a misnomer, and should really be renamed to just generic `MIPS32R2`, since this is what's being supported.  The detection of `MSA` by `c_check` can result in a false-positive, so I would add an option to set/unset `HAVE_MSA` manually, skipping the test, if possible.

My D-LINK DIR-615 (pure 24kc)  is offline.  It supports only the `mips16 ASE`. Here is a sample of my other MIPS boxes:

```
system type             : Qualcomm Atheros QCA956X ver 1 rev 0
machine                 : TP-Link Archer C6 v2 (US) / A6 v2 (US/TW)
processor               : 0
cpu model               : MIPS 74Kc V5.0
BogoMIPS                : 385.84
wait instruction        : yes
microsecond timers      : yes
tlb_entries             : 32
extra interrupt vector  : yes
hardware watchpoint     : yes, count: 4, address/irw mask: [0x0ffc, 0x0ffc, 0x0ffb, 0x0ffb]
isa                     : mips1 mips2 mips32r1 mips32r2
ASEs implemented        : mips16 dsp dsp2
Options implemented     : tlb 4kex 4k_cache prefetch mcheck ejtag llsc dc_aliases perf_cntr_intr_bit cdmm contextconfig perf
shadow register sets    : 1
kscratch registers      : 0
package                 : 0
core                    : 0
VCED exceptions         : not available
VCEI exceptions         : not available


system type             : MediaTek MT7628AN ver:1 eco:2
machine                 : TP-Link Archer C50 v3
processor               : 0
cpu model               : MIPS 24KEc V5.5
BogoMIPS                : 385.84
wait instruction        : yes
microsecond timers      : yes
tlb_entries             : 32
extra interrupt vector  : yes
hardware watchpoint     : yes, count: 4, address/irw mask: [0x0ffc, 0x0ffc, 0x0ffb, 0x0ffb]
isa                     : mips1 mips2 mips32r1 mips32r2
ASEs implemented        : mips16 dsp
Options implemented     : tlb 4kex 4k_cache prefetch mcheck ejtag llsc pindexed_dcache userlocal vint perf_cntr_intr_bit perf
shadow register sets    : 1
kscratch registers      : 0
package                 : 0
core                    : 0
VCED exceptions         : not available
VCEI exceptions         : not available


system type             : MediaTek MT7621 ver:1 eco:3
machine                 : TP-Link Archer C6 v3
processor               : 0
cpu model               : MIPS 1004Kc V2.15
BogoMIPS                : 586.13
wait instruction        : yes
microsecond timers      : yes
tlb_entries             : 32
extra interrupt vector  : yes
hardware watchpoint     : yes, count: 4, address/irw mask: [0x0ffc, 0x0ffc, 0x0ffb, 0x0ffb]
isa                     : mips1 mips2 mips32r1 mips32r2
ASEs implemented        : mips16 dsp mt
Options implemented     : tlb 4kex 4k_cache prefetch mcheck ejtag llsc pindexed_dcache userlocal vint perf_cntr_intr_bit cdmm perf
shadow register sets    : 1
kscratch registers      : 0
package                 : 0
core                    : 0
VPE                     : 0
VCED exceptions         : not available
VCEI exceptions         : not available
```
So none of them have `MSA` support.  I haven't tried to run openblas in all of them, only in the Archer C2, and it was just a basic test, really.

cotequeiroz:
BTW, the code in `c_check` can be compiled for mips32r2 and above, as long as you give the compiler the right flags.

``` C
#include <msa.h>
#include <stdio.h>


void main(void){
        __asm__ volatile("addvi.b $w0, $w1, 1");
        printf ("OK.\n");
}
```
```
mips-openwrt-linux-gcc -march=74kc msa_check.c -o msa_check -mmsa -mhard-float -mfp64
/home/openwrt/staging_dir/toolchain-mips_74kc_gcc-8.4.0_musl/bin/../lib/gcc/mips-openwrt-linux-musl/8.4.0/../../../../mips-openwrt-linux-musl/bin/ld: warning: msa_check uses -msoft-float (set by /home/openwrt/staging_dir/toolchain-mips_74kc_gcc-8.4.0_musl/bin/../lib/gcc/mips-openwrt-linux-musl/8.4.0/../../../../mips-openwrt-linux-musl/lib/crt1.o), /tmp/ccdXTv8y.o uses -mhard-float
```
It failed with a weird error, not the invalid instruction exception message I was expecting:
```
# /tmp/msa_check
-ash: /tmp/msa_check: not found
```

martin-frbg:
> My conclusion is that MIPS24K is just a misnomer, and should really be renamed to just generic MIPS32

Not quite true - it is just mainly a clone of the "P5600" target, which itself is fairly generic. The argument for a separate target type was just (a) automatic cpu model detection when building on that hardware (b) the ability to have the build system add suitable compiler flags and lastly (c) as a starting point for writing cpu-specific optimized BLAS kernels.

martin-frbg:
The "not found" error message could mean that the binary depends on a particular version of the loader (ld.so) that is not available, or simply that the loader entry is missing (if the compiler was totally confused by the conflicting options). Anyway it would certainly be better if the check could probe the cpu in addition to the compiler - if that is possible. Where /proc/cpuinfo is available, I guess it would already make sense to disable HAVE_MSA when the best MIPS ISA version given there  is too old to include msa.

commodo:
> The "not found" error message could mean that the binary depends on a particular version of the loader (ld.so) that is not available, or simply that the loader entry is missing (if the compiler was totally confused by the conflicting options). Anyway it would certainly be better if the check could probe the cpu in addition to the compiler - if that is possible. Where /proc/cpuinfo is available, I guess it would already make sense to disable HAVE_MSA when the best MIPS ISA version given there is too old to include msa.

it's not possible;
all packages on OpenWrt get built on x86 hosts for MIPS, PPC, ARM, x86, ARC (and all other flavors);
for now, this is fine as-is :)
if anyone wants more tweaking on a specific target, they can ask

Issue #17072, opened 2021-11-05
 v4l2tools: fails to build on arm_cortex-a7_neon-vfpv4


PolynomialDivision:
@mpromonet 

```
[0/1] Install the project...
-- Install configuration: "Release"
-- Installing: /builder/shared-workdir/build/sdk/build_dir/target-arm_cortex-a7+neon-vfpv4_musl_eabi/v4l2tools-0.1.5/ipkg-install/usr/bin/v4l2tools-0.1.5
Package v4l2tools is missing dependencies for the following libraries:
libvpx.so.6
make[3]: *** [Makefile:47: /builder/shared-workdir/build/sdk/bin/packages/arm_cortex-a7_neon-vfpv4/packages/v4l2tools_0.1.5-1_arm_cortex-a7_neon-vfpv4.ipk] Error 1
time: package/feeds/packages/v4l2tools/compile#8.64#1.44#32.62
```

https://downloads.openwrt.org/snapshots/faillogs/arm_cortex-a7_neon-vfpv4/packages/v4l2tools/compile.txt

PolynomialDivision:
https://github.com/openwrt/packages/pull/17075

neheb:
Should be fixed now.

PR #17082, opened 2021-11-05
v4l2tools: get rid of x264 and vpx

neheb:
Patched as it seems they can't be disabled.

Maintainer: @mpromonet 
Compile tested: ath79

### DIFF
--- a/multimedia/v4l2tools/patches/010-no-vpx-x264.patch
+++ b/multimedia/v4l2tools/patches/010-no-vpx-x264.patch
@@ -1,0 +1,18 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -39,7 +39,6 @@ target_link_libraries (${PROJECT_NAME} y
+ find_package(PkgConfig)
+ 
+ # x264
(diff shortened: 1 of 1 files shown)


PolynomialDivision:
Maybe some stupid question, but is this a general issue? I remember I had to fix gpsd to not use host library. Using pkg_check_modules should be fine?

(not against this PR just asking)

mpromonet:
Hi,

Maybe the openwrt build system doesnot use the right pkgconfig settings and use the host system. It should have a way settings some cmake variable to detect the target system ?

Best Regards,
Michel.

neheb:
It finds target packages. X264 should not be used.
.

mpromonet:
Hi,

Neither libx264 nor libx265 seems to be package of openwrt, but libvpx is, so it may be interesting to build v4l2tools with it.
I think https://github.com/openwrt/packages/pull/17084 is fixing the build problem describe in issue https://github.com/openwrt/packages/issues/17072. I am verifying this before submitting the PR.

Best Regards,
Michel.

PR #17084, opened 2021-11-06
v4l2tools: add vpx dependency if library installed

mpromonet:
Signed-off-by: Michel Promonet <michel.promonet@free.fr>

Maintainer: @mpromonet 
Compile tested: (arch:arm7, model:bcm2709, OpenWrt master)
Run tested: ( arch:arm7, model:bcm2709, OpenWrt master)
    * start v4l2tool to compress YUYV to VP80 on virtual v4l2loopback device
    * start v4l2rtspserver consuming virtual v4l2loopback device, get the stream remotely using ffplay

Description: 
Fix for https://github.com/openwrt/packages/issues/17072 adding vpx dependency when this library is selected

### DIFF
--- a/multimedia/v4l2tools/Makefile
+++ b/multimedia/v4l2tools/Makefile
@@ -7,13 +7,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=v4l2tools
-PKG_VERSION:=0.1.5
+PKG_VERSION:=0.1.7
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/mpromonet/v4l2tools.git
 PKG_SOURCE_VERSION:=v$(PKG_VERSION)
-PKG_MIRROR_HASH:=af63df7f88c1031aa5a438d175d77ca9c976e48eb55ff78a6ff6d2bd7272ac51
+PKG_MIRROR_HASH:=ded23eef11efbba8467ba58c5e69071ab4b34c0d29ad1a4fdc96683cfe2aff3c
 
 PKG_MAINTAINER:=Michel Promonet<michel.promonet@free.fr>
 PKG_LICENSE:=Unlicense
@@ -22,13 +22,21 @@ PKG_LICENSE_FILES:=LICENCE
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/cmake.mk
 
+PKG_CONFIG_DEPENDS := CONFIG_PACKAGE_libvpx
+
 define Package/v4l2tools
 	SECTION:=multimedia
 	CATEGORY:=Multimedia
 	TITLE:=v4l2tools
-	DEPENDS:=+libstdcpp +libjpeg
+	DEPENDS:=+libstdcpp +libjpeg +PACKAGE_libvpx:libvpx
 	URL:=https://github.com/mpromonet/v4l2tools
 endef
+
+CMAKE_OPTIONS += \
+       -DWITH_JPEG=ON \
+       -DWITH_VPX=$(if $(CONFIG_PACKAGE_libvpx),ON,OFF) \
+       -DWITH_X264=OFF \
+       -DWITH_X265=OFF 
 
 define Package/v4l2tools/install
 	$(INSTALL_DIR) $(1)/usr/bin


neheb:
This does not work as intended.

```
select v4l2tools
make package/libvpx/compile
make package/v4l2tools <-- should fail
```

mpromonet:
Hi,

What I tested is : 
- de-selecting libvpx and selecting v4l2tools => it builds v4l2tools without vpx support
- selecting libvpx and v4l2tools => it builds v4l2tools with vpx support
But I may misunderstand the DEPENDS feature.

Best Regards,
Michel.

stintel:
As I explained [here](https://github.com/openwrt/packages/pull/17075#issuecomment-965272858), this is not the right approach.

stintel:
With mpromonet/v4l2tools#42 we can do this:

```
diff --git a/multimedia/v4l2tools/Makefile b/multimedia/v4l2tools/Makefile
index 1c1f7c61d..78814bcad 100644
--- a/multimedia/v4l2tools/Makefile
+++ b/multimedia/v4l2tools/Makefile
@@ -30,6 +30,12 @@ define Package/v4l2tools
        URL:=https://github.com/mpromonet/v4l2tools
 endef

+CMAKE_OPTIONS += \
+       -DWITH_JPEG=OFF \
+       -DWITH_VPX=OFF \
+       -DWITH_X264=OFF \
+       -DWITH_X265=OFF \
+
 define Package/v4l2tools/install
        $(INSTALL_DIR) $(1)/usr/bin
        $(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/v4l2tools-$(PKG_VERSION) $(1)/usr/bin/v4l2tools
```

mpromonet:
Hi @stintel 

Thanks for the PR I merged it, I still not really understand the approach of openwrt package dependencies. How can we enable jpeg/vpx support for v4l2tools dynamically when libjpeg/libvpx are selected ?

Best Regards,
Michel.

neheb:
https://github.com/openwrt/packages/blob/master/libs/libextractor/Makefile#L25 and https://github.com/openwrt/packages/blob/master/libs/libextractor/Makefile#L61 are a hint.

mpromonet:
Hi,

I update v4l2tools Makefile to set CMAKE_OPTIONS & DEPENDS depending on CONFIG_PACKAGE_libvpx.
Does it match better the openwrt packages dependencies ?

Best Regards,
Michel.

multimedia/v4l2tools/Makefile
```diff
@@ -26,10 +26,16 @@ define Package/v4l2tools
 	SECTION:=multimedia
 	CATEGORY:=Multimedia
 	TITLE:=v4l2tools
-	DEPENDS:=+libstdcpp +libjpeg
+	DEPENDS:=+libstdcpp +libjpeg $(if $(CONFIG_PACKAGE_libvpx),+libvpx,)
```
neheb, line 29: This

neheb:
No. You need PKG_CONFIG_DEPENDS. The dependency should also be simplified as +PACKAGE_libvpx:libvpx

mpromonet:
Hi Rosen,

I did not understand PKG_CONFIG_DEPENDS was used by the package system, reading https://openwrt.org/docs/guide-developer/packages becomes clearer.
I update the pull request.

Best Regards,
Michel.

stintel:
It's still dynamic now, which still creates unpredictable builds. What you want to do is make those things an option. Check net/net-snmp for an example.

mpromonet:
Hi Stijn,

I am not sure to catch your idea. Why dynamic (depending on selected libraries) means unpredictable ?

Best Regards,
Michel.

stintel:
If I build today with `# CONFIG_PACKAGE_libvpx is not set`, v4l2tools will not have a dependency on libvpx. If tomorrow I enable `CONFIG_PACKAGE_libvpx=y`, v4l2tools will have a dependency on libvpx. For packages built by the buildbots, this will always result in the dependency being there, which we probably want to avoid, see https://github.com/openwrt/packages/pull/17075#issuecomment-961737893.

Unless a dependency is almost always needed for normal functioning of the tool, it is better to disable by default instead of making it dynamic, and if we do want to offer the feature to users, we should add config symbols for it, to allow people building their own images to enable it with e.g. CONFIG_V4L2TOOLS_WITH_VPX.

PR #17100, opened 2021-11-08
CI: add PKG-INFO metadata file

Habbie:
As briefly discussed in #17078 (cc @aparcar), this PR:

* adds PR numbers to the artifact file names
* shortens the git hashes in the file names (and prefixes them with `g`)
* adds a `PKG-INFO` file that is compatible with Python's packaging metadata, and therefore very easy to read with https://pypi.org/project/pkginfo/ (but also without, it's just a mail headers-like plain text file in the zip)

Comments welcome!

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -95,17 +95,57 @@ jobs:
       - name: Move created packages to project dir
         run: cp bin/packages/${{ matrix.arch }}/packages_ci/*.ipk . || true
 
+      - name: Collect metadata
+        run: |
+          MERGE_ID=$(git rev-parse --short HEAD)
+          echo "MERGE_ID=$MERGE_ID" >> $GITHUB_ENV
+          echo "BASE_ID=$(git rev-parse --short HEAD^1)" >> $GITHUB_ENV
+          echo "HEAD_ID=$(git rev-parse --short HEAD^2)" >> $GITHUB_ENV
+          PRNUMBER=${GITHUB_REF_NAME%/merge}
+          echo "PRNUMBER=$PRNUMBER" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


aparcar:
Please prefix this with `CI:` and not `GH CI artifacts`

aparcar:
Apart from that I think it's very cool, thank for working on that!

Habbie:
> Please remove the g, I think more people will think it's part of the hash and end up confused.

You're the second to say that! I'll do that :)

Habbie:
Removed (force push)

.github/workflows/multi-arch-test-build.yml
```diff
@@ -95,17 +95,60 @@ jobs:
       - name: Move created packages to project dir
         run: cp bin/packages/${{ matrix.arch }}/packages_ci/*.ipk . || true
 
+      - name: Collect metadata
+        run: |
+          VERSION=${BRANCH#openwrt-}
+          if [ $VERSION = master ]; then VERSION=99.99+master; fi
```
aparcar, line 101: The 99.99 is a bit confusing, shouldn't we just keep it as master/snapshot?

Habbie: I was looking for ways to make consumption of this data easy for users, and the python PKG-INFO format specification ( is strict about at least having a few numbers in there,  However, I did not check what the `pkginfo` tool requires, it might not be as strict. I'll check that soon, because I also prefer just sticking `master` or `snapshot` in there.

Habbie: Ok, it's not picky at all, so I'll remove all processing of $BRANCH and just stick it into Version directly.

aparcar:
Okay looks good! Thanks for you work. @neheb mind having a quick look?

neheb (approved):

PR #17101, opened 2021-11-08
wget: fix read hsts database (fix musl 64bit time_t)

zhanhb:
Maintainer: @neheb 
Compile tested: ath79/nand
Run tested: ath79/nand mips_24kc

Description:
- `time_t` on musl 1.2 is 64bit, while `long` is 32 bit. we will always get zero time when loading hsts database with the original source on 32bit big endian mips and must 1.2.
- The PR also upgrade wget to latest version.


net/wget/patches/100-fix-hsts-time.patch
```diff
@@ -0,0 +1,45 @@
+--- a/src/hsts.c
```
BKPepe, line 1: This patch is backported from upstream? If not, are you going to send it to them? It is not clear. There is missing commit header (author, commit messace, etc.)

BKPepe (changes requested):

net/wget/patches/100-fix-hsts-time.patch
zhanhb: https://savannah.gnu.org/bugs/index.php?61451

darnir:
Upstream is applying a much simpler patch instead:

```
diff --git i/src/hsts.c w/src/hsts.c
index ead5b728..9d03a9d0 100644
--- i/src/hsts.c
+++ w/src/hsts.c
@@ -293,12 +293,12 @@ hsts_read_database (hsts_store_t store, FILE *fp, bool merge_with_existing_entri
       if (*p == '#')
         continue;
 
-      items_read = sscanf (p, "%255s %d %d %lu %lu",
+      items_read = sscanf (p, "%255s %d %d %llu %llu",
                            host,
                            &port,
                            &include_subdomains,
-                           (unsigned long *) &created,
-                           (unsigned long *) &max_age);
+                           &created,
+                           &max_age);
 
       if (items_read == 5)
         func (store, host, port, created, max_age, !!include_subdomains);

```

This forces the read to be at a 64-bit value instead which is correct since the variables being read into are 64-bit and the portability layer bring used tries to ensure that `time_t` is 64-bit on nearly all systems that Wget runs on.

zhanhb:
@darnir Is `%llu` availiable on windows? I can find only three occurs of `%ll` in the C source now(including hsts.c).
https://github.com/mirror/wget/blob/f75fcf2985bf7ace36051f5d00a9f7c53e125a2b/src/metalink.c#L533-L535

darnir:
It generally isn't. But our gnulib portability layer provides support for the `ll` modifier on cygwin, mingw and msvc 14.

zhanhb:
> the portability layer bring used tries to ensure that `time_t` is 64-bit on nearly all systems that Wget runs on.

@darnir IMO, the portability layer won't change the size of `time_t`, at least on latest openwrt release the `time_t` is still 32 bit.

I compile wget with patch https://github.com/mirror/wget/commit/e1bacd2fa5f026f9493bd74e9179b859fa9ce8f1 applied using [openwrt-sdk-21.02.1-ath79/nand](https://downloads.openwrt.org/releases/21.02.1/targets/ath79/nand/openwrt-sdk-21.02.1-ath79-nand_gcc-8.4.0_musl.Linux-x86_64.tar.xz) and the feature HSTS broken.

darnir:
I thought you mentioned that `time_t` is 64-bit. It's just `long` that is 32-bit.
gnulib indeed does not fix the case where `time_t` is 32-bit only. But if that's the case, many more things will be broken. Not just HSTS.

I'll have to think about how to solve this issue in the correct way. I'll recommend you bring this issue up on the bug-wget mailing list so that we can resolve this upstream without you needing to carry patches. Simultaneously, I will look into what gnulib says about the y2038 problem in these cases.

zhanhb:
Size of `time_t` on latest openwrt release depends on the target machine. Openwrt only switched to musl1.2 (64-bit time_t) on the latest snapshot.

I'm not considering about 2038 problem, just on the device where I deployed the latest snapshot version of openwrt, the hsts file breaks like as following.
```
api.github.com	0	1	0	0
raw.githubusercontent.com	0	0	0	0
codeload.github.com	0	0	0	0
github-releases.githubusercontent.com	0	0	0	0
openwrt.org	0	0	0	0
archive.openwrt.org	0	0	0	0
github.com	0	1	1638520257	31536000
```

When `time_t` is 64 bit, it should be easy to handle even when `long` is only 32bit.
When only 32-bit `time_t` is available, the project gnulib should reimplement the 64-bit time system call like C lib and the type `time_t` should not be used directly in project `wget` to solve year 2038. It will be a huge refactor.

I've no sure what I can mail about 32-bit `time_t` and year 2038.

PR #17102, opened 2021-11-08
bossa: add patch to fix flash failures on serial

lynxis:
Maintainer: @PolynomialDivision 
Compile tested: master / omap
Run tested: 19.07 / omap

Description:
bossa only waited 1ms for the serial device instead calling
tcdrain().
Without this patch bossa is failing 8/10 times.
Patch has been submitted as PR#150 upstream but is still waiting
for the maintainers for feedback.


PolynomialDivision:
Please refresh the patch. ;)

lynxis:
Ok. The patch wasn't dirty just different :).

PR #17103, opened 2021-11-09
dnsproxy: Update to 0.39.10

1715173329:
Maintainer: me
Compile tested: ipq807x, rockchip/armv8
Run tested: rk3328 nanopi-r2s

Description:
Release note: https://github.com/AdguardTeam/dnsproxy/releases/tag/v0.39.10

PR #17105, opened 2021-11-09
CI: do not crash during PKG-INFO generation if there are no packages

Habbie:
Signed-off-by: Peter van Dijk <peter.van.dijk@powerdns.com>

reported in #17103, tested in https://github.com/PowerDNS/openwrt-packages/pull/1


Habbie:
@aparcar

PR #17112, opened 2021-11-12
cyrus-sasl: patch CVE-2019-19906

paper42:
Maintainer: @flyn-org

Description: https://nvd.nist.gov/vuln/detail/CVE-2019-19906


flyn-org:
Looks good. Recommend merge.

The fix is tracked upstream, but cyrus-sasl has been slow in the past to make new releases. I recommend moving forward with the local patch.

PR #17115, opened 2021-11-12
frr: update to 8.1.0

lucize:

Maintainer: me 
Tested x86,rampis,kirkwood


lucize:
I've opened an issue for the problematic targets

PR #17111, opened 2021-11-11
[21.02] php7: Update and clean up distributed php7.ini

thg2k:
This is a follow up of PR #17110 for branch 21.02, with the additional directives as of PHP 7.4.25 included in the change.

Please refer to that PR for explanation and discussion.

This same change should also be applied to master as the package is identical.

Maintainer: @mhei (Michael Heimpold)

PR #17119, opened 2021-11-13
php8: Clean up and update distributed php.ini for php 8.0.12

thg2k:
Follow up of PR #17111 for php8 package.

Notable changes from the php7 ini:
- Added `zend.exception_string_param_max_len` (new to PHP 8)
- Removed `highlight.bg` (that actually never existed, I missed it in the cleanup for php7, besides, we could drop the whole highlight block from the distributed file, who would ever want to change them?)

Maintainer: @mhei 

P.S.: This time I made the PR for master, but note that if you want to cherry pick it for 21 there is one patch behind, so the PKG_RELEASE bump will not work.

### DIFF
--- a/lang/php8/Makefile
+++ b/lang/php8/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=php
 PKG_VERSION:=8.0.12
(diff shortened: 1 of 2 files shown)


mhei:
Thanks!

Issue #996, opened 2015-03-07
Python: File not found (_bsddb.so), build failed


enigmagroup:
This happens since today, in all architectures tested so far (x86, x86_64, sunxi).

I also made ```make dirclean``` and restarted the build, same result.

```
rstrip.sh: /home/x/openwrt-32.mostrecent/build_dir/target-i386_i486_uClibc-0.9.33.2/Python-2.7.9/ipkg-x86/python-light/usr/lib/python2.7/lib-dynload/time.so: shared object
rstrip.sh: /home/x/openwrt-32.mostrecent/build_dir/target-i386_i486_uClibc-0.9.33.2/Python-2.7.9/ipkg-x86/python-light/usr/lib/python2.7/lib-dynload/operator.so: shared object
rstrip.sh: /home/x/openwrt-32.mostrecent/build_dir/target-i386_i486_uClibc-0.9.33.2/Python-2.7.9/ipkg-x86/python-light/usr/lib/python2.7/lib-dynload/bz2.so: shared object
rstrip.sh: /home/x/openwrt-32.mostrecent/build_dir/target-i386_i486_uClibc-0.9.33.2/Python-2.7.9/ipkg-x86/python-light/usr/lib/python2.7/lib-dynload/_io.so: shared object
rstrip.sh: /home/x/openwrt-32.mostrecent/build_dir/target-i386_i486_uClibc-0.9.33.2/Python-2.7.9/ipkg-x86/python-light/usr/lib/python2.7/lib-dynload/resource.so: shared object
(cd /home/x/openwrt-32.mostrecent/build_dir/target-i386_i486_uClibc-0.9.33.2/Python-2.7.9/ipkg-x86/python-light/CONTROL; ( echo "$CONTROL"; printf "Description: "; echo "$DESCRIPTION" | sed -e 's,^[[:space:]]*, ,g'; ) > control; chmod 644 control; ( echo "#!/bin/sh"; echo "[ \"\${IPKG_NO_SCRIPT}\" = \"1\" ] && exit 0"; echo ". \${IPKG_INSTROOT}/lib/functions.sh"; echo "default_postinst \$0 \$@"; ) > postinst; ( echo "#!/bin/sh"; echo ". \${IPKG_INSTROOT}/lib/functions.sh"; echo "default_prerm \$0 \$@"; ) > prerm; chmod 0755 prerm;  )
install -d -m0755 /home/x/openwrt-32.mostrecent/bin/x86/packages/packages
[...]

PR #17121, opened 2021-11-14
python3: bump to version 3.10.0

commodo:
Maintainer: me, @jefferyto 
Compile tested: x86 https://github.com/openwrt/commit/49f615022c921189ff4566b8b2bfdbf97a2a8787
Run tested: x86 https://github.com/openwrt/commit/49f615022c921189ff4566b8b2bfdbf97a2a8787

Manually re-applied:
  008-distutils-use-python-sysroot.patch
  016-adjust-config-paths.patch

Drop patch: 003-do-not-run-distutils-tests.patch
There is now a configure option '--disable-test-modules'
And seems we left the '_ctypes_test' around for quite some time.
Dropped now.
Refs:
  https://bugs.python.org/issue27640
  https://bugs.python.org/issue43282

Drop patch: 013-getbuildinfo-date-time-source-date-epoch.patch
Python build honors SOURCE_DATE_EPOCH pretty well now.

Drop setuptools patches. Setuptools should be reproducible with Python 3.6+
according to a mention here:
  https://github.com/pypa/setuptools/pull/1690#issuecomment-536517456
It's time to let upstream fix Setuptools reproduce-ability.

Drop patch: 010-do-not-add-rt-lib-dirs-when-cross-compiling.patch
I can't seem to fully remember why it's there.
And it seem to build fine without it.

Drop patch: 015-abort-on-failed-modules.patch
Python build supports a similar PYTHONSTRICTEXTENSIONBUILD=1 env-var
option.

Add patch: 026-openssl-feature-flags.patch
We need to keep this in our tree for a while.
See:
  https://bugs.python.org/issue45627

Backport patch: 027-bpo-43158-Use-configure-values-for-building-_uuid-ex.patch
Link: https://github.com/python/cpython/pull/29353
Fixes the build for uuid C module.

Add patch: 028-host-python-support-ssl-with-libressl.patch
We need the _ssl module working on the host-side with LibreSSL for pip to
work to download from https://pypi.org
Refs: https://github.com/openwrt/openwrt/pull/4749

Add patch: 029-disable-deprecation-warning.patch
Fixes apparmor build. The warning causes a configure error.

Refreshed the rest of patches.

Some old build-flags were removed. They don't seem to be necessary anymore.

Split python3-uuid from python3-light. To better manage the libuuid library
(if needed). Also, fixing the uuid C module build. Seems this was failing,
and was falling back to using hashlib.


jefferyto:
I won’t have time to review this until later this week, sorry 🙏

diizzyy:
This seems to break quite a bit of users in general (even at compilation),  is this tested with users in tree?

commodo:
> This seems to break quite a bit of users in general (even at compilation), is this tested with users in tree?

I tested with python3-cffi.
But since you raised this, I will do a test with all of them.
It's not often than a Python 3.X -> 3.(X + 1) version bump breaks things, but this one looks more special.

I will take a look.
Do you have any available logs?
If not, I will try to reproduce this locally.

diizzyy:
Unforunately not, I just noticed breakage in another repo and wanted to highlight it before getting committed.

commodo:
First error I got is in apparmor:
```
configure: error: cannot import Python module "distutils".
Please check your Python installation. The error was:
<string>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
make[3]: *** [Makefile:220: /home/aardelean/work/openwrt/openwrt/build_dir/target-i386_pentium4_musl/apparmor-3.0.3/.configured_a17fb5ef857664f03cd0ce37cc5ea591] Error 1
make[3]: Leaving directory '/home/aardelean/work/openwrt/openwrt/feeds/packages/utils/apparmor'
```

We can patch this out for now.
Let's see if this is the case for others.
If so, then ¯\_(ツ)_/¯  , packages will need to update after a while.

commodo:
Changelog v1 -> v2:
  * Add patch: 029-disable-deprecation-warning.patch
     Fixes apparmor build. The warning causes a configure error.

commodo:
I think I managed to find a few more.
Seems like some Python packages did not get any updates recently, and this newer Python version is bringing a few of them into the light.

Some PRs:  https://github.com/openwrt/packages/pull/17143  https://github.com/openwrt/packages/pull/17142 https://github.com/openwrt/packages/pull/17144

I think, we should be in a "good" condition to take this change and let the buildbots bring out more.
For PyYAML (python-yaml in PR https://github.com/openwrt/packages/pull/17142) I can't tell yet why it fails to build with the SDK.
My local-build works with the OpenWrt core repo.
Will try to find out.

paper42:
Python 3.10 removed some deprecated names from collections, so we should either update them or import upstream patches:

wscli
https://github.com/aws/aws-cli/commit/0c9c08856bff1e60cb14ea9a2fdfa640c9af9eeb
botocore
https://github.com/boto/botocore/commit/ebd12d8d539b2135db32c84b94f33319179e2eb3
dnspython
https://github.com/rthalley/dnspython/commit/62d3d41382d919af4662c12876f3de6ca854fd54
stem
https://github.com/torproject/stem/commit/36bcb170ba9097885902513640075eac2e6ce384

some packages don't have a patch upstream, but it should be trivial to fix:
micropython-lib
https://github.com/micropython/micropython-lib/blob/3a6ab0b46d6471bee00ae815444c624709dd4cdd/python-stdlib/http.client/http/client.py#L902
eventlet 0.30.2
https://github.com/eventlet/eventlet/blob/377b4fb39cc59273bd5ff461eb0388e3c3dffdb3/eventlet/green/http/client.py#L1051

commodo:
> Python 3.10 removed some deprecated names from collections, so we should either update them or import upstream patches:
> 
> wscli [aws/aws-cli@0c9c088](https://github.com/aws/aws-cli/commit/0c9c08856bff1e60cb14ea9a2fdfa640c9af9eeb)

the commit title on this one says `Fix collections.abc imports for Python 3.9 `
so, the breakage should have been noticed by now;

and yes, the packages should update or patch stuff;

paper42:
> the commit title on this one says `Fix collections.abc imports for Python 3.9 ` so, the breakage should have been noticed by now;

It was just a deprecation warning until 3.10.

commodo:
> > the commit title on this one says `Fix collections.abc imports for Python 3.9 ` so, the breakage should have been noticed by now;
> 
> It was just a deprecation warning until 3.10.
> 
> 3.9:
> 
> ```
> >>> from collections import MutableMapping
> <stdin>:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
> ```
> 
> 3.10:
> 
> ```
> >>> from collections import MutableMapping
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> ImportError: cannot import name 'MutableMapping' from 'collections' (/usr/lib/python3.10/collections/__init__.py)
> ```

ack;
makes sense;

anyway: if no objections, i would merge this update now and handle the breakage as it arrives;

this update has been in the works for 1+ month :)
and we're at a point where it's easier to merge [now] and fix fallout [later], rather than wait in fear of any breakage (that will happen anyway)

(i'm not frustrated or anything, i mostly want to forget about it and move forward; even if it means cleaning up breakage)

jefferyto:
I was working on a review for this PR but I see you have merged it already. My apologies for the delay but I have actual questions about this change. (I had questions, since this PR is in the past now.)

> we're at a point where it's easier to merge [now] and fix fallout [later]

What evidence is there that is easier to merge and fix later? I would argue it is easier to make things worse by creating bugs by rushing to merge, then rushing fixes and creating more bugs, repeat and repeat.

21.02 was just released, there won't be another release for at least 6 months. There is no urgent need to update master to the latest version.

> rather than wait in fear of any breakage (that will happen anyway)

Just because "breakage will happen anyway" doesn't mean we stop trying to fix problems before they happen.

Perhaps you don't agree, but I am of the belief that (to a reasonable extent) every commit in the repo history should be a working version of the codebase, i.e. if you checkout the code any at point in the history, it should still work. This is why we have development branches and pull requests and code reviews, to ensure that committed code is working code.

I appreciate your work on this @commodo but you wanting to "forget about it and move forward" is not a good enough reason to merge your own work without anyone else's review.

jefferyto:
This would have been my review:

> Drop setuptools patches. Setuptools should be reproducible with Python 3.6+
> according to a mention here:
> https://github.com/pypa/setuptools/pull/1690#issuecomment-536517456
> It's time to let upstream fix Setuptools reproduce-ability.

AFAICT that PR only addresses Provides-Extra entries (replacing 003-PKG-INFO-output-reproducible.patch), it does not cover the fixes in the other two patches.

Has upstream fixed those reproducibility issues? If not, is upstream aware of these issues (i.e. are there open issues in their tracker)? I copied those patches from Debian, I'm not sure if the Debian developers (or their bug reporters) actually reported the issues to upstream. If upstream is not aware of the issues then they cannot possibly fix them.

(There is also no need to keep the patches-setuptools directory if there are no more patches.)

> Drop patch: 010-do-not-add-rt-lib-dirs-when-cross-compiling.patch
> I can't seem to fully remember why it's there.
> And it seem to build fine without it.

The patch was added in c8c509bdadbdc5e6007fc5436f94485beb59de19 from #1006 to fix #996. (From #996 it appears you were not able to reproduce the issue locally, i.e. it built fine without the patch for you.) The commit message says there were build failures due to cross compilers not supporting the `-R` option. Do our cross compilers now support this option?

> Add patch: 026-openssl-feature-flags.patch
> We need to keep this in our tree for a while.
> See:
> https://bugs.python.org/issue45627

From https://bugs.python.org/msg405162:

> If you disable the builtin blake2 module, then certain features like MAC and tree hashing are not working.

I'm concerned that without the blake2 module, Python programs may start breaking with no obvious workaround. I think we should get BLAKE2 enabled by default in the main repo and have the Python package depend on `CONFIG_OPENSSL_WITH_BLAKE2`.

> Add patch: 028-host-python-support-ssl-with-libressl.patch
> We need the _ssl module working on the host-side with LibreSSL for pip to
> work to download from https://pypi.org
> Refs: openwrt/openwrt#4749

I'm concerned that this patch looks complex and difficult to maintain (though I am not a LibreSSL/OpenSSL developer, maybe it is simple for someone with that knowledge).

> Add patch: 029-disable-deprecation-warning.patch
> Fixes apparmor build. The warning causes a configure error.

I don't think we should disable this warning globally.

I'm curious, why does the warning cause the build failure? I would expect warnings to just warn (and let the build continue).

If we disable the warning here, then we miss the chance to see the warnings and report the deprecation to upstream packages. If we cannot stop the warning from triggering a build failure, then I would prefer to patch individual packages (if possible) to bypass the warning or not use distutils.

commodo:
firstly, thank you for the review :)

> This would have been my review:
> 
> > Drop setuptools patches. Setuptools should be reproducible with Python 3.6+
> > according to a mention here:
> > [pypa/setuptools#1690 (comment)](https://github.com/pypa/setuptools/pull/1690#issuecomment-536517456)
> > It's time to let upstream fix Setuptools reproduce-ability.
> 
> AFAICT that PR only addresses Provides-Extra entries (replacing 003-PKG-INFO-output-reproducible.patch), it does not cover the fixes in the other two patches.

It's an alternative to https://github.com/pypa/setuptools/pull/1305

> 
> Has upstream fixed those reproducibility issues? If not, is upstream aware of these issues (i.e. are there open issues in their tracker)? I copied those patches from Debian, I'm not sure if the Debian developers (or their bug reporters) actually reported the issues to upstream. If upstream is not aware of the issues then they cannot possibly fix them.

So, I'm not 100% sure whether they did.
I tested by twiddling the SOURCE_DATE_EPOCH value and got reproducible results.

On PR https://github.com/pypa/setuptools/pull/1305
This is a comment from a Debian core dev: https://github.com/pypa/setuptools/pull/1305#issuecomment-470736783
So, Debian devs should be aware of this. I did not dig too deep in their patches to see whether they got rid of them; they still apply.
Then the last comment on that PR https://github.com/pypa/setuptools/pull/1305#issuecomment-538810632 suggests that setuptools should be reproducible.

Again: not 100%, but it looks promising to take a risk.

> 
> (There is also no need to keep the patches-setuptools directory if there are no more patches.)

ok; I can do a PR to remove it;
mostly it's there to signal where to put other patches should we require them;

> 
> > Drop patch: 010-do-not-add-rt-lib-dirs-when-cross-compiling.patch
> > I can't seem to fully remember why it's there.
> > And it seem to build fine without it.
> 
> The patch was added in [c8c509b](https://github.com/openwrt/packages/commit/c8c509bdadbdc5e6007fc5436f94485beb59de19) from #1006 to fix #996. (From #996 it appears you were not able to reproduce the issue locally, i.e. it built fine without the patch for you.) The commit message says there were build failures due to cross compilers not supporting the `-R` option. Do our cross compilers now support this option?

Seems they do. Or maybe they have been supporting this for a while.
I think this was before we were failing the build with options like `PYTHONSTRICTEXTENSIONBUILD=1` (and what we had before this).
So, we were missing extension-build failures and getting reports later

> 
> > Add patch: 026-openssl-feature-flags.patch
> > We need to keep this in our tree for a while.
> > See:
> > https://bugs.python.org/issue45627
> 
> From https://bugs.python.org/msg405162:
> 
> > If you disable the builtin blake2 module, then certain features like MAC and tree hashing are not working.
> 
> I'm concerned that without the blake2 module, Python programs may start breaking with no obvious workaround. I think we should get BLAKE2 enabled by default in the main repo and have the Python package depend on `CONFIG_OPENSSL_WITH_BLAKE2`.

i thought about enabling Blake2 by default as well;
it was a discussion i avoided having with OpenSSL in OpenWrt core

> 
> > Add patch: 028-host-python-support-ssl-with-libressl.patch
> > We need the _ssl module working on the host-side with LibreSSL for pip to
> > work to download from https://pypi.org
> > Refs: [openwrt/openwrt#4749](https://github.com/openwrt/openwrt/pull/4749)
> 
> I'm concerned that this patch looks complex and difficult to maintain (though I am not a LibreSSL/OpenSSL developer, maybe it is simple for someone with that knowledge).

this will get simpler after LibreSSL 3.5.0
well, hopefully Python SSL won't require some even newer OpenSSL;
in reality, 4-5 functions that this patch adds are covered in LibreSSL as presented here: https://github.com/openwrt/openwrt/pull/4749#issuecomment-962898659
only `SSL_CTX_get_security_level()` needs patching, same as OpenBSD does  https://github.com/openbsd/ports/blob/master/lang/python/3.10/patches/patch-Modules__ssl_c

but since it's easier to patch in Python stuff for OpenWrt host build, I went that way

> 
> > Add patch: 029-disable-deprecation-warning.patch
> > Fixes apparmor build. The warning causes a configure error.
> 
> I don't think we should disable this warning globally.
> 
> I'm curious, why does the warning cause the build failure? I would expect warnings to just warn (and let the build continue).
> 
> If we disable the warning here, then we miss the chance to see the warnings and report the deprecation to upstream packages. If we cannot stop the warning from triggering a build failure, then I would prefer to patch individual packages (if possible) to bypass the warning or not use distutils.

this is one patch that needs to go away asap
interestingly enough, apparmor still fails:  https://downloads.openwrt.org/snapshots/faillogs/x86_64/packages/apparmor/compile.txt
but that's apparmor's job to fix compat with Python 3.10

commodo:
> I was working on a review for this PR but I see you have merged it already. My apologies for the delay but I have actual questions about this change. (I had questions, since this PR is in the past now.)

ok; apologies for the rush

> 
> > we're at a point where it's easier to merge [now] and fix fallout [later]
> 
> What evidence is there that is easier to merge and fix later? I would argue it is easier to make things worse by creating bugs by rushing to merge, then rushing fixes and creating more bugs, repeat and repeat.

it's a balance actually between rushing and letting PRs / issues open for long periods of time;
to give some examples:
the alternative is 110 PRs here https://github.com/openwrt/packages/pulls
and 240 PR here https://github.com/openwrt/openwrt/pulls
all sitting idle-y waiting for attention

FWIW, I also have a few open items (for long periods) on the packages feed, but I try to address things as I can
i'm not happy about them
they are festering on my brain
and i feel somewhat relieved when some things go away for a while

> 
> 21.02 was just released, there won't be another release for at least 6 months. There is no urgent need to update master to the latest version.
> 
> > rather than wait in fear of any breakage (that will happen anyway)
> 
> Just because "breakage will happen anyway" doesn't mean we stop trying to fix problems before they happen.

ack
it's still a balance;

> 
> Perhaps you don't agree, but I am of the belief that (to a reasonable extent) every commit in the repo history should be a working version of the codebase, i.e. if you checkout the code any at point in the history, it should still work. This is why we have development branches and pull requests and code reviews, to ensure that committed code is working code.
> 
> I appreciate your work on this @commodo but you wanting to "forget about it and move forward" is not a good enough reason to merge your own work without anyone else's review.

i understand that i upset you (again)
we seem to have these periodic arguments
i guess, we should come closer to a point where we shouldn't have them anymore :)

my sentiment is that i've overridden your control of things, and i understand that it can be upsetting

jefferyto:
> i understand that i upset you (again)
> we seem to have these periodic arguments
> i guess, we should come closer to a point where we shouldn't have them anymore :)

I don't mind having productive arguments. It is natural for different people to have different opinions; it is only through discussion, and sometimes debate, that we arrive at the best/most appropriate solution.

> my sentiment is that i've overridden your control of things, and i understand that it can be upsetting

Yes, because I was under the impression that we were working under a system of consensus, or at least acknowledged disagreement. In the past I would not have requested any change to be merged without asking for your opinion first; now with commit access I would not have merged anything without your input. I'm disappointed to learn that I wasn't afforded the same courtesy.

commodo:
> I'm disappointed to learn that I wasn't afforded the same courtesy.

ack ; i understand the disappointment;
i don't know how to fix disappointment [in this case], aside from stepping down and avoiding further situations;

i feel that this is not the first disappointment that I've caused you along the years;
whether they were my fault or not, i don't know/remember;
either way, if there are more than 2 cases, then [me] stepping down is not a bad idea;

jefferyto:
For future reference, regarding the other two setuptools patches removed in this PR:

* 001-reproducible.patch (concerning the content of [install_files.txt](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=848136)) - This does not appear to have been reported upstream, but since it appears this file is generated by `easy_install` and `easy_install` is now [officially deprecated](https://setuptools.pypa.io/en/latest/history.html#v58-3-0), I'm not sure if upstream would be interested in fixing this.
* 002-sorted-requires.patch (concerning the content of [requires.txt](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=804249)) - This was reported in https://github.com/pypa/setuptools/issues/458; upstream has fixed the issue by requiring ordered sequences (list or string) for `install_requires`.

PR #17124, opened 2021-11-14
sstp-client: fix register the sstp interface

pipihu:
Maintainer: @fededim 
Compile tested: (mt7621,x86_64)
Run tested: (mt7621,x86_64)

Description:
fix this [https://forum.archive.openwrt.org/viewtopic.php?id=58007](https://forum.archive.openwrt.org/viewtopic.php?id=58007)
at the same time, the client IP address can be displayed on the interface.
interface


neheb:
Get rid of that merge commit.

pipihu:
sorry, I'm not good at using git.

PR #17125, opened 2021-11-14
ddns-scripts: Fix typo in route53 update script

Atema:
Maintainer: @feckert 
Compile tested: N/A
Run tested: No

Description: Fix typo in the route53 ddns update script that caused it not to recognise a missing username key

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
-PKG_RELEASE:=15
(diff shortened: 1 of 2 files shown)


feckert:
Please bump PKG_RELEASE number
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L11

Do we need also have to backport this to openwrt-21.02 ?

Atema:
> Please bump PKG_RELEASE number https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L11

Done

> Do we need also have to backport this to openwrt-21.02 ?

I don't really think it's necessary: the script should work fine with a correct configuration. There may just be some weird errors now when username is missing from the config.

PR #17130, opened 2021-11-15
python-build: add support for pyproject.toml files

commodo:
Maintainer: me & @jefferyto
Compile tested: x86 https://github.com/openwrt/commit/49f615022c921189ff4566b8b2bfdbf97a2a8787
Run tested: x86 https://github.com/openwrt/commit/49f615022c921189ff4566b8b2bfdbf97a2a8787

    A new PEP 517 (https://www.python.org/dev/peps/pep-0517/) has defined that
    Python packages can be shipped without any `setup.py` file, and that a
    `pyproject.toml` file is sufficient.
    
    For these cases, we will add a support in the OpenWrt build-system to
    provide the default `setup.py` shim layer in case this file does not exist,
    but there is a `pyproject.toml` file.
    
    We also seem to need to tweak the shim layer with the PKG_VERSION,
    otherwise the detected version is 0.0.0.
    
### DIFF
--- a/lang/python/python3-host.mk
+++ b/lang/python/python3-host.mk
@@ -51,6 +51,12 @@ HOST_PYTHON3_PIP:=$(STAGING_DIR_HOSTPKG)/bin/pip$(PYTHON3_VERSION)
 
 HOST_PYTHON3_PIP_CACHE_DIR:=$(DL_DIR)/pip-cache
 
+define SetupPyShim
+	if [ -f $(1)/pyproject.toml ] && [ ! -f $(1)/setup.py ] ; then \
+		$(CP) $(python3_mk_path)setup.py.shim $(1)setup.py ; \
+	fi
+endef
+
 # Multiple concurrent pip processes can lead to errors or unexpected results: https://github.com/pypa/pip/issues/2361
 # $(1) => packages to install
 define HostPython3/PipInstall
@@ -75,8 +81,9 @@ endef
 # $(2) => additional arguments to setup.py
 # $(3) => additional variables
 define HostPython3/ModSetup
+	$(call SetupPyShim,$(HOST_BUILD_DIR)/$(strip $(1)))
 	$(call HostPython3/Run, \
 		$(HOST_BUILD_DIR)/$(strip $(1)), \
 		setup.py $(2), \
-		$(3))
+		$(3) PY_PKG_VERSION=$(PKG_VERSION))
 endef
--- a/lang/python/python3-package.mk
+++ b/lang/python/python3-package.mk
@@ -62,10 +62,11 @@ endef
 # $(3) => additional variables
 define Python3/ModSetup
 	$(INSTALL_DIR) $(PKG_INSTALL_DIR)/$(PYTHON3_PKG_DIR)
+	$(call SetupPyShim,$(PKG_BUILD_DIR)/$(strip $(1)))
 	$(call Python3/Run, \
 		$(PKG_BUILD_DIR)/$(strip $(1)), \
 		setup.py $(2), \
-		$(3))
+		$(3) PY_PKG_VERSION=$(PKG_VERSION))
 endef
 
 define Python3/FixShebang
--- a/lang/python/setup.py.shim
+++ b/lang/python/setup.py.shim
@@ -1,0 +1,6 @@
+import os
+import setuptools
+
+# FIXME: see about getting rid of PY_PKG_VERSION asap when setuptools handles this correctly
+if __name__ == "__main__":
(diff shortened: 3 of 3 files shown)


jefferyto:
I think the Python build change is important enough to have it's own PR, reviewing it as part of an update to jsonschema doesn't make sense to me.

Also, I'm not a fan of adding a setup.py shim as a solution to this problem; Python packages without setup.py is the future, we shouldn't be adding it back in just to make our lives easier. I'll try to find some time to look into this further.

commodo:
> I think the Python build change is important enough to have it's own PR, reviewing it as part of an update to jsonschema doesn't make sense to me.
> 

ack;
will change;
the idea was [as is done in the Linux kernel sometimes] that if you add a new subsystem/framework change, you also add a first user [which also serves as example]
otherwise, people add functions/changes that may never be used;
a first user is an indicator of how it will be used;
i'll admit that the case here may not be that similar;

> Also, I'm not a fan of adding a setup.py shim as a solution to this problem; Python packages without setup.py is the future, we shouldn't be adding it back in just to make our lives easier. I'll try to find some time to look into this further.

i should have probably referenced some other blogs;
looking around, i found that to use pyproject.toml, one needs to run setuptools.setup(), which essentially can end up looking like a minimal setup.py;
calling setuptools.setup() without any args, is buggy, because the version may be 0.0.0

commodo:
Changelog v1 -> v2:
* remove jsonschema patch

commodo:
> Also, I'm not a fan of adding a setup.py shim as a solution to this problem; Python packages without setup.py is the future, we shouldn't be adding it back in just to make our lives easier. I'll try to find some time to look into this further.

So, not to put pressure or anything, but even this solution is a little better (or a good middle-ground) than not doing anything.
While I agree that there are better solutions (than this), I will let this sit here for another ~5 days, then merge it as-is.
If something better comes along [after that] we can add it later.

Python's packaging mess (and we've come to know it for a while now) should not take too much of our time (if possible).

I agree that we want to do it perfect on the first iteration, but that seems to take longer than doing a middle-iteration (which this is).

As a bit of background: after struggling with the Python 3.10.0 (mostly the SSL) update, I was perfectly happy to do this simple shim layer :)

Again: I'm not frustrated or anything.
While it looks like I am half-assing stuff, the reality is that, I really am half-assing this stuff :)
It's the best I'm willing to do (right now).

jefferyto:
> Python's packaging mess (and we've come to know it for a while now) should not take too much of our time (if possible).

I disagree, if something is worth doing then it worth doing with as much time as it needs.

> I agree that we want to do it perfect on the first iteration, but that seems to take longer than doing a middle-iteration (which this is).

By merging a "middle-iteration", there is less incentive for the next person to work on the "perfect iteration" because "it already works". (Why fix something that already works?) Except there will be cases where it doesn't work (because the solution was not the correct one), but again there is less incentive to work on the correct solution because "it already mostly works".

By not working on correct solutions from the beginning, we introduce [technical debt](https://en.wikipedia.org/wiki/Technical_debt) that will need to be paid later (with interest).

> While it looks like I am half-assing stuff, the reality is that, I really am half-assing this stuff :)
> It's the best I'm willing to do (right now).

There is actually no urgent need to do anything right now. There are no other packages (other than pyyaml) that need this. 21.02 was just released. The only urgency is in your mind to get rid of this task, by half-assing a solution that someone else will need to fix later.

> Again: I'm not frustrated or anything.

I suggest taking a break and stepping away for a while.

commodo:
> By not working on correct solutions from the beginning, we introduce technical debt that will need to be paid later (with interest).

I understand technical debt :)
How's Seafile and dependencies doing these days?

>  There is actually no urgent need to do anything right now. 

well, urgent / not urgent
sometimes things pile up
and there's always a decision to address them (even half-way)
if people want to update packages in master and this is a change that will help things for now ¯\_(ツ)_/¯

even Python had it share of technical debt (via patches) that got a bit cleaned up
and I am optimistic that next updates will clean up more of those patches;

again: apologies for overriding your control; i understand it can be upsetting :)

commodo:
>  How's Seafile and dependencies doing these days?
> 

speaking of Seafile, my intention wasn't to do this:
https://downloads.openwrt.org/snapshots/faillogs/x86_64/packages/seafile-seahub/compile.txt

but  ¯_(ツ)_/¯

commodo:
I thought about doing this a few times, after our first 2 (or 3 ? ) arguments.

This has been brewing in my head for some time.
I guess now, I am bit more neutral than with the other arguments.

I also know 2 d*** moves don't make a not d*** move.
But https://github.com/openwrt/packages/pull/17184

Having said that, this PR can also be closed.

PR #17131, opened 2021-11-15
[21.02] python-build: add support for pyproject.toml files

commodo:
Maintainer: me, @jefferyto 
Compile tested: n/a
Run tested: n/a

Pulled from #17130 
In case we need to update a package to a version that does not have any `setup.py`(but has `pyproject.toml`) we should be ready.

A new PEP 517 (https://www.python.org/dev/peps/pep-0517/) has defined that
Python packages can be shipped without any `setup.py` file, and that a
`pyproject.toml` file is sufficient.

For these cases, we will add a support in the OpenWrt build-system to
provide the default `setup.py` shim layer in case this file does not exist,
but there is a `pyproject.toml` file.

We also seem to need to tweak the shim layer with the PKG_VERSION,
otherwise the detected version is 0.0.0.


commodo:
Closing this, as #17130 as closed.

PR #17136, opened 2021-11-15
ddns-scripts: Add script for TransIP.nl

Atema:
Maintainer: @chris5560 / @feckert
Compile tested: No
Run tested: Manual installation of the script on openwrt 21.02 with the current version of ddns-scripts, curl, and openssl-utils installed works as expected.

Description: Adds a custom script for updating DNS records at TransIP.nl (one of the largest registrars in The Netherlands) with ddns-scripts

### DIFF
--- a/net/ddns-scripts/files/usr/lib/ddns/update_transip_nl.sh
+++ b/net/ddns-scripts/files/usr/lib/ddns/update_transip_nl.sh
@@ -1,0 +1,134 @@
+#!/bin/sh
+#
+# 2021 Martijn Atema <martijn@atema.one>
+#
+# This script sends ddns updates using the TransIP API (see https://api.transip.nl/)
+# and is parsed by dynamic_dns_functions.sh inside send_update().
+#
+# The following options provided by ddns are used:
+# username  - Username of account used for logging in to TransIP
+# password  - Private key generated at https://www.transip.nl/cp/account/api/
+#            (make sure to accept non-whitelisted IP addresses)
+# domain    - Base domain name registered at TransIP
+#             ('domain.tld' when updating 'hostname.domain.tld')
+# param_enc - Name of DNS record to update
+#             ('hostname' when updating 'hostname.domain.tld')
+# param_opt - TTL of the DNS record to update (in seconds)
+#
+# Note: Make sure that there is exactly one record of type A (for IPv4) or
+#       AAAA (for IPv6) with the specified name and TTL. That record will be
+#       updated by this script.
+#
+# The script requires cURL with SSL and the openssl binary
+
+
+[ -z "${username}" ] && write_log 14 "Service config is missing 'username'"
+[ -z "${password}" ] && write_log 14 "Service config is missing 'password' (private key)"
+[ -z "${domain}" ] && write_log 14 "Service config is missing 'domain' (base domain name)"
+[ -z "${param_enc}" ] && write_log 14 "Service config is missing 'param_enc' (DNS record name)"
+[ -z "${param_opt}" ] && write_log 14 "Service config is missing 'param_opt' (DNS record TTL)"
+
+[ -z "${CURL_SSL}" ] && write_log 14 "TransIP update requires cURL with SSL"
+[ -z "$(openssl version)" ] && write_log 14 "TransIP update requires openssl binary"
+
+. /usr/share/libubox/jshn.sh
+
+
+# Re-format the private key and write to a temporary file
+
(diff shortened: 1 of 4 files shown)


net/ddns-scripts/files/usr/lib/ddns/update_transip_nl.sh
```diff
@@ -0,0 +1,138 @@
...
+
+[ -z "${CURL_SSL}" ] && write_log 14 "TransIP update requires cURL with SSL"
+[ -z "$(openssl version)" ] && write_log 14 "TransIP update requires openssl binary"
+
+. /usr/share/libubox/jshn.sh
+
+
+# Re-format the private key and write to a temporary file
+
+__tmp_keyfile="/tmp/ddns-transip-$(openssl rand -hex 4)"
```
feckert, line 39: Please use mktemp for this
https://github.com/openwrt/packages/search?q=mktemp

net/ddns-scripts/files/usr/lib/ddns/update_transip_nl.sh
```diff
@@ -0,0 +1,138 @@
...
+        -H "Content-Type: application/json" \
+        -H "Signature: ${__auth_signature}" \
+        -d "${__auth_body}" \
+        -w "%{http_code}\n" \
+        -o "${DATFILE}" 2>"${ERRFILE}")
+
+
+# Logging for error and debug
+
+if [ $? -ne 0 ]
```
feckert, line 81: Please use `if [ $? -ne 0 ]; then`

net/ddns-scripts/files/usr/lib/ddns/update_transip_nl.sh
```diff
@@ -0,0 +1,138 @@
...
+then
+        write_log 14 "Curl failed: $(cat "${ERRFILE}")"
+        return 1
+fi
+
+if [ -z ${__auth_status} ] || [ ${__auth_status} -ne 201 ]
```
feckert, line 87: Please use ` ]; then`

net/ddns-scripts/files/usr/lib/ddns/update_transip_nl.sh
```diff
@@ -0,0 +1,138 @@
...
+        -H "Content-Type: application/json" \
+        -H "Authorization: Bearer ${__auth_token}" \
+        -d "${__update_body}" \
+        -w "%{http_code}\n" \
+        -o "${DATFILE}" 2>"${ERRFILE}")
+
+
+# Logging for error and debug
+
+if [ $? -ne 0 ]
```
feckert, line 125: Please use ` ]; then`

net/ddns-scripts/files/usr/lib/ddns/update_transip_nl.sh
```diff
@@ -0,0 +1,138 @@
...
+then
+        write_log 14 "Curl failed: $(cat "${ERRFILE}")"
+        return 1
+fi
+
+if [ -z ${__update_status} ] || [ ${__update_status} -ne 204 ]
```
feckert, line 131: Please use ` ]; then`

feckert (changes requested):
Unfortunately, I can't test this because I don't have an account with this provider. Everything else looks good so far

Atema:
@feckert I've updated with the requested changes. Can I assume that `mktemp` is always available?

feckert:
In the openwrt repository the requested busybox app is set to `y`
https://github.com/openwrt/openwrt/blob/b6cbbbb6ef4deaba00f5c128ea66696fc64c5bf0/package/utils/busybox/Config-defaults.in#L703
I don't expect it to change, but to be on the safe side we could add this as a dependency to.
https://github.com/openwrt/packages/pull/17136/files#diff-c0e9e626f2ba612f4284522051479b373f8481892313fdc5c766dae5e70cbf1dR213

Have a look how this could be done in openwrt https://github.com/openwrt/openwrt/blob/master/package/network/utils/wireguard-tools/Makefile#L40

Atema:
@feckert I added a dependency for `coreutils-mktemp` only if the symbol `BUSYBOX_CONFIG_MKTEMP` is not set (using syntax from https://openwrt.org/docs/guide-developer/dependencies). That should make it work in any scenario.

feckert:
@Atema thanks for this :+1: . I have update the PKG_RELEASE number so I pushed it directly.
https://github.com/openwrt/packages/commit/aa820a2cae321311097e59de9b67dfa5c648cbeb

PR #17148, opened 2021-11-18
conntrack-tools: import patch to fix cache

PolynomialDivision:
As written in the commit message:

Depending on your conntrackd configuration, events might get lost,
leaving stuck entries in the cache forever. Skip checking the conntrack
ID to allow for lazy cleanup by when a new entry that is represented by
the same tuple is added.

Maintainer: @jow- 
Compile tested: Edge Router 4 (trunk)
Run tested: Edge Router 4

### DIFF
--- a/net/conntrack-tools/Makefile
+++ b/net/conntrack-tools/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=conntrack-tools
 PKG_VERSION:=1.4.6
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
Here, you can see the difference:
http://monitor.berlin.freifunk.net/detail.php?p=memory&t=memory&h=ohlauer-gw&s=86400

PolynomialDivision:
Okay, I saw that @neheb did the last merges. Could you merge this? This is a very important fix otherwise, the cache will grow without any bound and kill the synchronization. I run-tested that now in a asymmetric routing bgp structure for several days and it fixes everything.

PolynomialDivision:
Thanks! :)

PR #17158, opened 2021-11-19
[19.07] nextdns: Update to version 1.37.4

rs:
Update to version 1.37.4

### DIFF
--- a/net/nextdns/Makefile
+++ b/net/nextdns/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nextdns
-PKG_VERSION:=1.37.3
+PKG_VERSION:=1.37.4
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


1715173329:
From the error [1]:
```
build github.com/nextdns/nextdns: cannot load embed: malformed module path "embed": missing dot in first path element

Building targets
build .: cannot find module for path .
```

The `embed` package is part of go1.16+, but we only have go1.13 on 19.07 branch [2].

1. https://github.com/openwrt/packages/runs/4267308371?check_suite_focus=true#step:5:3409
2. https://github.com/openwrt/packages/blob/c559096e03e596077a60394cbe236332854aeb18/lang/golang/golang-version.mk

rs:
Is it stuck with this version? For what reason? Any plan tu upgrade the build system for this branch?

Most go packages will be harder and harder to keep up to date with this restriction.

neheb:
It's frozen, yes.

PR #17163, opened 2021-11-20
prometheus-node-exporter-lua: listen on ipv4 and ipv6

luizluca:
Permits the daemon to listen on both protocol family and
also on all interface addresses (using different instances)

* prometheus-node-exporter-lua script:
- Add new parameters (-4, -6) to select socket family in
prometheus-node-exporter-lua. By default it will use both
- Make bind optional. It will use :: or 0.0.0.0 as needed.

* prometheus-node-exporter-lua config:
- listen_family now defines the socket family (ipv4, ipv6 or any)
- listen_ipv6 is gone, although if present, it will be converted
  to listen_family, overhiding its settings

* prometheus-node-exporter-lua init script:
- when a listen_interface is defined (not '*'), it will launch an
  instance for each interface address, respecting listen_family

Maintainer: @champtar 
Compile tested: noarch
Run tested: x86_64

Description:

### DIFF
--- a/utils/prometheus-node-exporter-lua/Makefile
+++ b/utils/prometheus-node-exporter-lua/Makefile
@@ -4,7 +4,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=prometheus-node-exporter-lua
-PKG_VERSION:=2021.10.31
+PKG_VERSION:=2021.11.19
 PKG_RELEASE:=1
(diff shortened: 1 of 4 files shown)


champtar:
After a long delay ... in https://github.com/openwrt/packages/pull/18360 I switch to uhttpd, please have a look

PR #17171, opened 2021-11-21
geoipupdate: initial commit

clayface:
Maintainer: me
Compile tested: bcm53xx
Run tested: bcm53xx

Description:
Add MaxMind's geoipupdate utility. mmdb files are downloaded to /tmp/GeoIP
by default. The user should update /etc/GeoIP.conf with their API key and
DB choice, currently set to country only. So as not to exceed MaxMind's
download limitations, the user should manually run the utility or set up a
cron job.

### DIFF
--- a/net/geoipupdate/Makefile
+++ b/net/geoipupdate/Makefile
@@ -1,0 +1,67 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=geoipupdate
+PKG_VERSION:=4.8.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/maxmind/geoipupdate/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=ca718c3ffcc595ef441363699888d20150f1d3a6583ac2d60bcbd34f052db09f
+
+PKG_LICENSE:=Apache-2.0 MIT
+PKG_LICENSE_FILES:=LICENSE-APACHE LICENSE-MIT
+PKG_MAINTAINER:=Matthew Hagan <mnhagan88@gmail.com>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/maxmind/geoipupdate
+GO_PKG_LDFLAGS_X:= \
+	main.defaultConfigFile=/etc/GeoIP.conf \
+	main.defaultDatabaseDirectory=/var/GeoIP
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/geoipupdate
+  TITLE:=GeoIP Update
+  URL:=https://github.com/maxmind/geoipupdate
+  SECTION:=net
+  CATEGORY:=Network
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
+define Package/geoipupdate/description
+ The GeoIP Update program performs automatic updates of GeoIP2 and GeoIP
+ Legacy binary databases. CSV databases are not supported.
+
+ Users will need to update /etc/GeoIP.conf with their API key and DB
+ choice, currently set to country only. To update the DB, the user can
+ manually run the utility or set up a cron job.
+
+ This package contains the program and config file.
+endef
+
+define Package/geoipupdate/install
+	$(call GoPackage/Package/Install/Bin,$(PKG_INSTALL_DIR))
+	
+	$(INSTALL_DIR) \
+	$(1)/usr/sbin \
+	$(1)/etc/init.d
+
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/geoipupdate $(1)/usr/sbin/
+	$(INSTALL_CONF) $(PKG_BUILD_DIR)/conf/GeoIP.conf.default $(1)/etc/GeoIP.conf
+	$(INSTALL_BIN) ./files/geoipupdate.init $(1)/etc/init.d/geoipupdate
+
+	$(SED) "s/^EditionIDs GeoLite2-Country GeoLite2-City/EditionIDs GeoLite2-Country/" $(1)/etc/GeoIP.conf
+endef
+
+define Package/geoipupdate/conffiles
+/etc/GeoIP.conf
+endef
+
+$(eval $(call GoBinPackage,geoipupdate))
+$(eval $(call BuildPackage,geoipupdate))
--- a/net/geoipupdate/files/geoipupdate.init
+++ b/net/geoipupdate/files/geoipupdate.init
@@ -1,0 +1,9 @@
+#!/bin/sh /etc/rc.common
+
+START=99
+
+start() {
+	[ -d /var/GeoIP ] || {
+		mkdir -p /var/GeoIP
+	}
+}


net/geoipupdate/Makefile
```diff
@@ -0,0 +1,83 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=geoipupdate
+PKG_VERSION:=4.8.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=v$(PKG_VERSION).tar.gz
```
BKPepe, line 9: This is not good.

net/geoipupdate/Makefile
```diff
@@ -0,0 +1,83 @@
...
+PKG_SOURCE_URL:=https://github.com/maxmind/geoipupdate/archive/refs/tags/
+PKG_HASH:=ca718c3ffcc595ef441363699888d20150f1d3a6583ac2d60bcbd34f052db09f
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)
```
BKPepe, line 12: Already default.

```suggestion
```

net/geoipupdate/Makefile
```diff
@@ -0,0 +1,83 @@
...
+  TITLE:=GeoIP Update
+  URL:=https://github.com/maxmind/geoipupdate
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
+define Package/geoipupdate
+$(call Package/geoipupdate/Default)
+  SECTION:=net
+  CATEGORY:=Network
+endef
```
BKPepe, line 40: You can merge this, right? You don't need to use `Package/geoipupdate/Default`. It is ok to use just this: `define Package/geoipupdate`

net/geoipupdate/Makefile
```diff
@@ -0,0 +1,83 @@
...
+	
+	$(INSTALL_DIR) \
+	$(1)/usr/sbin \
+	$(1)/etc/init.d \
+	$(1)/lib/upgrade/keep.d
+
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/* $(1)/usr/sbin/
+	$(INSTALL_CONF) $(PKG_BUILD_DIR)/conf/GeoIP.conf.default $(1)/etc/GeoIP.conf
+	$(INSTALL_BIN) ./files/geoipupdate.init $(1)/etc/init.d/geoipupdate
+	$(INSTALL_DATA) ./files/geoipupdate.upgrade $(1)/lib/upgrade/keep.d/geoipupdate
```
BKPepe, line 64: Why this is required?

BKPepe (changes requested):

net/geoipupdate/Makefile
clayface: Thanks. I did likewise for the description

net/geoipupdate/Makefile
clayface: Primarily we need to maintain the user's API key across upgrades in `/etc/GeoIP.conf`. Is this what you are referring to?

BKPepe: Yes, but since the file is marked as conffile. This should not be necessary AFAIK.

clayface: That should be it now. Confirmed `/etc/GeoIP.conf` is retained across sysupgrades with `geoipupdate.upgrade` removed.

net/geoipupdate/Makefile
```diff
@@ -0,0 +1,72 @@
...
+PKG_NAME:=geoipupdate
+PKG_VERSION:=4.8.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/maxmind/geoipupdate/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=ca718c3ffcc595ef441363699888d20150f1d3a6583ac2d60bcbd34f052db09f
+
+PKG_LICENSE:=Apache-2.0, MIT
+PKG_LICENSE_FILES:=LICENSE-APACHE, LICENSE-MIT
```
jefferyto, line 14: `PKG_LICENSE` and `PKG_LICENSE_FILES` are space-separated (no commas).

net/geoipupdate/Makefile
```diff
@@ -0,0 +1,72 @@
...
+PKG_MAINTAINER:=Matthew Hagan <mnhagan88@gmail.com>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/maxmind/geoipupdate
+GO_PKG_LDFLAGS_X:= \
+	main.defaultConfigFile=/etc/GeoIP.conf \
+	main.defaultDatabaseDirectory=/tmp/GeoIP
```
jefferyto, line 24: Perhaps `/var/GeoIP` would be better? `/var` is usually a symlink to `/tmp` (unless the user wants to mount `/var` to permanent storage).

net/geoipupdate/Makefile
```diff
@@ -0,0 +1,72 @@
...
+endef
+
+define Package/geoipupdate/postinst
+#!/bin/sh
+
+[ -n "$$IPKG_INSTROOT" ] || {
+	/etc/init.d/geoipupdate enable
+	/etc/init.d/geoipupdate start
+}
+endef
```
jefferyto, line 69: This is already done by the [default postinst script](https://github.com/openwrt/openwrt/blob/25bc66eb40ea2c062940778fba601032b2579734/package/base-files/files/lib/functions.sh#L296-L305).

clayface:
@jefferyto thanks for the feedback. Those issues are hopefully now resolved.

net/geoipupdate/Makefile
```diff
@@ -0,0 +1,63 @@
...
+  CATEGORY:=Network
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
+define Package/geoipupdate/description
+ The GeoIP Update program performs automatic updates of GeoIP2 and GeoIP
+ Legacy binary databases. CSV databases are not supported.
+
+ This package contains the program and config file.
+endef
```
jefferyto, line 42: Just noticed the commit message - perhaps the details that the user needs to know:

* update /etc/GeoIP.conf with their API key and DB choice, currently set to country only
* manually run the utility or set up a cron job

can be included here in the package description?

jefferyto (commented):
Apologies for the delay

net/geoipupdate/Makefile
clayface: Makefile updated with description. Thanks

jefferyto (approved):
LGTM - thanks!

net/geoipupdate/Makefile
```diff
@@ -0,0 +1,67 @@
...
+	
+	$(INSTALL_DIR) \
+	$(1)/usr/sbin \
+	$(1)/etc/init.d
+
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/geoipupdate $(1)/usr/sbin/
+	$(INSTALL_CONF) $(PKG_BUILD_DIR)/conf/GeoIP.conf.default $(1)/etc/GeoIP.conf
+	$(INSTALL_BIN) ./files/geoipupdate.init $(1)/etc/init.d/geoipupdate
+
+	sed -i "s/^EditionIDs GeoLite2-Country GeoLite2-City/EditionIDs GeoLite2-Country/" $(1)/etc/GeoIP.conf
```
neheb, line 59: sed -i can be replaced with $(SED) I think.

clayface: @neheb done. Thanks

PR #17184, opened 2021-11-23
python-packages: remove myself as maintainer

commodo:
Maintainer: @jefferyto  /  was me
Compile tested: n/a
Run tested: n/a

There's been a bit of overlapping opinions on some of these packages.
The best thing to do here is to reduce ownership and relinquish my
control.

This patch does that.


jefferyto:
To be clear, this was not what I wanted when I suggested taking a break.

commodo:
> To be clear, this was not what I wanted when I suggested taking a break.

yeah, i know :)

PR #17191, opened 2021-11-24
openvswitch: document missing options

stintel:
Maintainer: @yousong 
Compile tested: NA
Run tested: NA

Description:
Commit 1038ac1 ("openvswitch: add support for definining bridge ports...")
added two new options:
- drop_unknown_ports
- ports
They are missing from the README, so add them.


yousong:
It seems that another uci option `ports` was also added in the same commit for `drop_unknown_ports`.  Can you please also add a line for that ?

PR #17204, opened 2021-11-25
conmon: remove unnecessary dependency

oskarirauta:
this patch removes unnecessary dependency to golang/host
as pointed out by @jefferyto earlier at #17063 

Compile tested: x86_64, git
Run tested: x86_64, git

### DIFF
--- a/utils/conmon/Makefile
+++ b/utils/conmon/Makefile
@@ -12,7 +12,7 @@ PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=LICENSE
 
-PKG_BUILD_DEPENDS:=golang/host libseccomp
+PKG_BUILD_DEPENDS:=libseccomp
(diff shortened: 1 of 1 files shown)


BKPepe:
* dependency

Otherwise, this comment: https://github.com/openwrt/packages/pull/17063#issuecomment-962192883 was more clear than your commit description, but from my side, it is nitpicking.

oskarirauta:
> * dependency
> 
> Otherwise, this comment: [#17063 (comment)](https://github.com/openwrt/packages/pull/17063#issuecomment-962192883) was more clear than your commit description, but from my side, it is nitpicking.

I agree.

oskarirauta:
@BKPepe how about approving it then?

jefferyto:
Apologies for not reviewing this PR in time.

@oskarirauta In the future, I suggest adding the actual comment link (`https://github.com/openwrt/packages/pull/17063#issuecomment-962192883`) in the commit message instead of the text `#17063 (comment)`, since that text does not link to the actual comment.

PR #17205, opened 2021-11-25
apparmor,domoticz: use $(PYTHON3_VERSION) instead of hardcoding it in the Makefile

cotequeiroz:
Maintainer: @oskarirauta, @dwmw2 
Compile tested: mediatek, aarch64
Run tested: none

Description:
The packages fail to build with python 3.10, because the python version was hardcoded to the packages' Makefile, and there was a recent python change from 3.9 to 3.10.

Instead of hardcoding the pyton version, use `PYTHON3_VERSION` from python3-version.mk (included by `python3-package.mk`).

I have not run-tested them, but I did check that the apparmor files where properly installed in `/usr/lib/python3.10/site-packages`.

### DIFF
--- a/utils/domoticz/Makefile
+++ b/utils/domoticz/Makefile
@@ -75,7 +75,6 @@ CMAKE_OPTIONS += \
 
 TARGET_CFLAGS += -flto
 TARGET_CXXFLAGS += -DWITH_GPIO -flto
-TARGET_LDFLAGS += -lpython3.9
 
 define Build/Prepare
 	$(call Build/Prepare/Default)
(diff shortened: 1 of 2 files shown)


oskarirauta (approved):
Looks good...

utils/domoticz/Makefile
```diff
@@ -75,7 +76,7 @@ CMAKE_OPTIONS += \
 
 TARGET_CFLAGS += -flto
 TARGET_CXXFLAGS += -DWITH_GPIO -flto
-TARGET_LDFLAGS += -lpython3.9
+TARGET_LDFLAGS += -lpython$(PYTHON3_VERSION)
```
neheb, line 79: Wrong. This just needs to be removed. This was a hack from before domoticz had proper python 3.9 support. It's not needed anymore.

neheb:
The egg-info thing is weird.

@jefferyto know anything about that? I don't see other packages installing that.

utils/domoticz/Makefile
cotequeiroz: OK

cotequeiroz:
Maybe @jefferyto knows their relevance, but the egg-info files or directories are quite common: 
```
src/openwrt/staging_dir/target-aarch64_cortex-a53_musl/root-mediatek $ find usr/lib/python3.9/site-packages/ -name '*egg-info' | wc -l
195
```
This was an from an older `CONFIG_ALL=y` staging dir that I had in hand, but it should not change that much with the python3.10 update.

BKPepe:
Let's merge this. We can take a look at ```egg-info``` files later.

jefferyto:
egg-info files are how on-device pip knows what Python packages are installed. They are usually packaged automatically by the Python package build process in python3-package.mk, so it's a little unusual to see them "manually" installed like here.

PR #17209, opened 2021-11-26
ddns-scripts: Fix wrong whitespace in preinst and postinst scripts

### LINKED ISSUE #16143: ddns-scripts: preinst exec format error
Maintainer: @feckert 
Environment: Turris Omnia, arch mvebu, Turris OS 5.2.3 based on 19.07

After updating to the newest version of the OS, I got this error:

```
ddns-scripts_no-ip_com/preinst: [31;1mDIE[0m:Failed to exec /usr/lib/opkg/info//ddns-scripts_no-ip_com.preinst: Exec format error

ddns-scripts/preinst: [31;1mDIE[0m:Failed to exec /usr/lib/opkg/info//ddns-scripts.preinst: Exec format error
```

peci1:
Maintainer: unkown?
Compile tested: no
Run tested: no

Description: Fixes #16143 by removing superfluous whitespace at the beginnings of preinst/postinst shell scripts. The scripts are apparently generated as is directly from the makefile, which means the scripts begin with `\t#!/bin/sh` and the interpreter is not happy with that.

Alternatively, it should be enough to unindent the shebang lines only. However, it would seem weird to me.

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -151,11 +151,12 @@ endef
 
 ###### *************************************************************************
 define Package/ddns-scripts/preinst
-	#!/bin/sh
-	# if NOT run buildroot then stop service
-	[ -z "$${IPKG_INSTROOT}" ] && /etc/init.d/ddns stop >/dev/null 2>&1
-	exit 0	# suppress errors
+#!/bin/sh
+# if NOT run buildroot then stop service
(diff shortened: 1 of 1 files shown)


feckert:
@BKPepe I don't have an openwrt-19.07 device anymore, so I can't do a run/build/install test. But from my point of view, the pullrequst looks good. All other preinst and postinst install scripts I know do not have a `tab` at the beginning. Therefore :+1:

feckert:
Let's give it a try.
Thanks for the fix.

PR #17214, opened 2021-11-26
backport artifact metadata generation to 21.02

Habbie:
backport of #17100 + #17105, with corrrected commit message prefixes (`CI:`)


Habbie:
friendly ping

BKPepe:
In my opinion, we should backport more changes from the ``master`` branch or even from the ``openwrt-22.03`` branch to keep it in sync as there are many other useful things, which could be useful for OpenWrt 21.02.

aparcar:
Thanks. Ideally we should create a GitHub action which is included, so we don't have to update things over multiple branches.

PR #17227, opened 2021-11-28
ksmbd-tools: Fix ksmbd service is semi-killed at system startup

httpstorm:
The configuration for the ksmbd service is auto-generated when the OpenWRT configuration changes, and also during startup, hence ksmbd.init has to reload the kernel module. It does that by calling kill_server, which does not perform cleanup. This results in ksmbd being killed but not restarted properly during boot.
This patch resolves the issue by using stop_service, which performs proper cleanup.

https://forum.openwrt.org/t/ksmbd-samba3-4-alternative-ex-cifsd-smbd-package-support-thread/51695/68

Maintainer: @Andy2244 
Compile tested: (x86_x64, OpenWrt version r18176)
Run tested: (x86_x64, OpenWrt version r18176, tests done: startup, change config)

@neheb

### DIFF
--- a/net/ksmbd-tools/Makefile
+++ b/net/ksmbd-tools/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ksmbd-tools
 PKG_VERSION:=3.4.2
-PKG_RELEASE:=1
+PKG_RELEASE:=$(AUTORELEASE)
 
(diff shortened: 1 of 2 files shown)


httpstorm:
It must be something wrong with the buildbot, because I compiled and deployed the latest master from an empty directory, and everything went smooth. Also this patch changes only the service script.

PR #17120, opened 2021-11-13
ddns-scripts: Rework DigitalOcean update script

giannoug:
Signed-off-by: George Giannou <giannoug@gmail.com>

Maintainer: @feckert
Compile tested: N/A
Run tested: ASUS RT-AC58U / ARMv7 Processor rev 5 (v7l) / OpenWrt 21.02.0-rc3 r16172-2aba3e9784 / LuCI openwrt-21.02 branch git-21.163.64918-ba57ec5. Tested that it updates hostnames after IP renewal

Description:
As described in #16162, the DigitalOcean update script needed some work. I changed the way it handles API updates, but I have only tested successfully renewing hostnames (which works).

I'd like to check other paths too, e.g. by removing parameters and using wrong API keys, but I can't find where `write_log` is shown 🤔

PR #17228, opened 2021-11-28
ddns-scripts: Fix DigitalOcean JSON payload

giannoug:
Signed-off-by: George Giannou <giannoug@gmail.com>

Maintainer: @feckert
Compile tested: N/A
Run tested: ASUS RT-AC58U / ARMv7 Processor rev 5 (v7l) / OpenWrt 21.02.0-rc3 r16172-2aba3e9784 / LuCI openwrt-21.02 branch git-21.163.64918-ba57ec5. Tested that it updates hostnames after IP renewal

Description:
This is a fix for a previous PR (#17120). It seems that the JSON object is created incorrectly and returned empty ({ }). Update returns 200 with an empty body, but the hostname doesn't update 🤕 Removing `json_add_object` and `json_close_object` fix the issue, the payload is correctly created and the IPs are updated with the new value.

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
-PKG_RELEASE:=16
+PKG_RELEASE:=17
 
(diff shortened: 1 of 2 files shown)


giannoug:
Bumped to 17

feckert:
Thanks

PR #17230, opened 2021-11-28
cni-route-override: new package

oskarirauta:
route-override IPAM works as meta CNI plugin to override IP route given by previous CNI plugins. It is useful in a case with network-attachment-definition.
Currently route-override verified its feature with podman and crio(with Kubernetes).


PolynomialDivision:
lgtm

PR #17239, opened 2021-11-30
[RFC] mwan3: convert src_ip and dest_ip hanling to uci list

feckert:
Maintainer: me / @aaronjg 
Compile tested: not needed script changes
Run tested: x86_64, APU3, OpenWrt latest master

Description:
The 'src_ip' and 'dest_ip' option on a rule cannot take only one value.
Until now it was possible to add several with comma separation. This
option is now converted into a list so that no extra handling has to be
applied during validation in the LuCI.

The following issue took me to turn into a list
https://github.com/openwrt/luci/issues/5333

Could you validate this in your setup?
@jamesmacwhite @ptpt52 

The LuCI change and an upgrade script must still be created TODO

### DIFF
--- a/net/mwan3/files/lib/mwan3/mwan3.sh
+++ b/net/mwan3/files/lib/mwan3/mwan3.sh
@@ -727,10 +727,8 @@ mwan3_set_user_iptables_rule()
 	config_get timeout "$1" timeout 600
 	config_get ipset "$1" ipset
 	config_get proto "$1" proto all
-	config_get src_ip "$1" src_ip
 	config_get src_iface "$1" src_iface
 	config_get src_port "$1" src_port
-	config_get dest_ip "$1" dest_ip
 	config_get dest_port "$1" dest_port
 	config_get use_policy "$1" use_policy
 	config_get family "$1" family any
@@ -742,13 +740,51 @@ mwan3_set_user_iptables_rule()
 	[ "$family" = "ipv4" ] && [ "$ipv" = "ipv6" ] && return
 	[ "$family" = "ipv6" ] && [ "$ipv" = "ipv4" ] && return
 
-	for ipaddr in "$src_ip" "$dest_ip"; do
-		if [ -n "$ipaddr" ] && { { [ "$ipv" = "ipv4" ] && echo "$ipaddr" | grep -qE "$IPv6_REGEX"; } ||
-						 { [ "$ipv" = "ipv6" ] && echo "$ipaddr" | grep -qE $IPv4_REGEX; } }; then
-			LOG warn "invalid $ipv address $ipaddr specified for rule $rule"
-			return
+	local ip_valid=1
+	mwan3_valid_rule_ip() {
+		local ipaddr="$1"
+		local ipv="$2"
+		local rule="$3"
+		local direction="$4"
+
+		local result=0
+
+		case "$ipv" in
+			"ipv6")
+				echo "$ipaddr" | grep -Eq "$IPv6_REGEX"
+				result="$?"
+				if [ "$result" = "1" ]; then
+					LOG warn "Invalid '$ipv' address '$ipaddr' specified for rule '$rule'"
+					export ip_valid=0
+				fi
+				;;
+			"ipv4")
+				echo "$ipaddr" | grep -Eq "$IPv4_REGEX"
+				result="$?"
+				if [ "$result" = "1" ]; then
+					LOG warn "Invalid '$ipv' address '$ipaddr' specified for rule '$rule'"
+					export ip_valid=0
+				fi
+				;;
+		esac
+
+		if [ "${direction}" = "scp_ip" ]; then
+			export scp_ip="${src_ip}${ipaddr},"
+		else
+			export dest_ip="${dest_ip}${ipaddr},"
 		fi
-	done
+	}
+
+	config_list_foreach "$1" src_ip mwan3_valid_rule_ip "$ipv" "$rule" "src_ip"
+	config_list_foreach "$1" dest_ip mwan3_valid_rule_ip "$ipv" "$rule" "dest_ip"
+
+	if [ "$ip_valid" -eq 0 ]; then
+		return
+	fi
+
+	# Strip last comma from string
+	dest_ip="${dest_ip%,}"
+	src_ip="${src_ip%,}"
 
 	if [ -n "$src_iface" ]; then
 		network_get_device src_dev "$src_iface"
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.10.12
+PKG_VERSION:=2.11.0
 PKG_RELEASE:=1
 PKG_MAINTAINER:=Florian Eckert <fe@dev.tdt.de>, \
 		Aaron Goodman <aaronjg@alumni.stanford.edu>


ptpt52:
@feckert  Could we have the src_ip/dest_ip as a textfield
e.g.
```
option src_ip "192.168.15.0/24,172.16.1.100-172.16.1.200,1.2.3.4"
```
and in mwan3.sh this option would be translate to:
```
src_ip="1.2.3.4/32,172.16.1.100/30,172.16.1.104/29,172.16.1.112/28,172.16.1.128/26,172.16.1.192/29,172.16.1.200/32,192.168.15.0/24"
```
than apply to iptables

ptpt52:
for validation in javascript:
```
function nets_validate(nets)
{
        var re_ip = /^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/
        var re_ip_cidr = /^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\/([0-9]|1[0-9]|2[0-9]|3[012])$/
        var re_ip_range = /^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)-(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/
        var net = nets.split(',');
        for (i = 0; i < net.length; i++) {
                if (net[i].match(re_ip)) continue;
                if (net[i].match(re_ip_cidr)) continue;
                if (net[i].match(re_ip_range) && ip_range_validate(net[i])) continue;
                return false;
        }
        return true
}
```

ptpt52:
uci list and list of text input in web ui is not comfortable
I like the textfield and we could type a lot of `ip/ip cidr/ip range` in to it.

aaronjg:
Related to this, we should not be creating rules for iptables with multiple ips, since this creates a bunch of iptable rules.  I had a branch that creates an ipset when multiple IPs are present in source and destination.:

https://github.com/aaronjg/openwrt-packages/tree/mwan3-ipset

Could we make this change in parallel to the UCI change?

ptpt52:
great!
Using ipset is an option, or hidden inside mwan3.sh, it depends on the actual situation whether it uses ipset or not
For the UI input interface, the user only needs to enter an ip, some ip, some ip range, or any combination
For mwan3.sh, it chooses to use an iptables rule or a rule with ipset according to the situation

feckert:
@ptpt52 I understand what you mean, but I'm in the process of cleaning up the mwan3 and I didn't like the handling of src_ip/dest_ip. I just want to get rid of all the special treatments.
The fact that the comma works the way it does is only because there was no `list` on writing this software and the previous maintainer, did not do a well validation I think.

The suggestion with the ipset is great. @aaronjg  can you rebase your change and make a pullrequest?

@ptpt52 If this works with the ipsets, then you can fill the ipset as you wish and do not have to use the changed list options src_ip/dest_ip via luci-app-mwan3. I have had a [WIP pullrequest ](https://github.com/openwrt/luci/pull/4833) for ipset handling open in LuCI for a while. But I haven't worked on it any more due to lack of time.

If this works for you I can then change this option to list, because then I can better validate the whole thing in the LuCI.

openips:
fllow this

jamesmacwhite:
Hello, I've been summoned!

I'm not sure I'll be able to test this currently, I'm currently remaining on 19.07 and the 2.8 mwan3 versions because of the wireless performance issues with mwlwifi (WRT3200ACM) on 21.02, although hopefully 21.02.2 might look promising to maybe take the jump with the regression hopefully fixed, but there is a fair bit of work with the DSA change too. I also can't run 2.10.x on 19.07 because of issues with the kernel and L2TP.

However, I think converting the comma syntax to uci list is the right move, I guess my only question, if we convert the sort of half unknown comma syntax to uci list, does this means these rules become ipset rules? Say if you have a just a few IPs with comma separated, as iptables rules now, does having an ipset created for a couple of IPs negate any performance or use more memory than the iptables example?

aaronjg:
I rebased my ipset change and pushed, though it still needs more testing.

sjtuross:
Hello, just my two cents. I think mwan3 should leverage firewall to create and manage ipset as documented here https://openwrt.org/docs/guide-user/firewall/fw3_configurations/fw3_config_ipset and simply support source ipset. On luci part, expose src ipset and dst ipset as columns in the rules tab. Keep src_ip and dest_ip as is just for simple scenarios.

GeorgeSapkin:
@aaronjg @feckert still relevant? If so it has merge conflicts.

feckert:
I have that on my patch set. I will update it when I get the chance.

BKPepe:
Kind reminder :)

net/mwan3/files/lib/mwan3/mwan3.sh
```diff
@@ -742,13 +740,51 @@ mwan3_set_user_iptables_rule()
...
+					export ip_valid=0
+				fi
+				;;
+		esac
+
+		if [ "${direction}" = "scp_ip" ]; then
+			export scp_ip="${src_ip}${ipaddr},"
+		else
+			export dest_ip="${dest_ip}${ipaddr},"
 		fi
```
openwrt-ai, line 775: Source addresses are stored in the wrong variable, so `src_ip` matching is silently dropped and the source list is merged into `dest_ip`.

`config_list_foreach` (line 778) passes `"src_ip"` as `direction`, but this branch tests for `"scp_ip"`, so it is never taken — every entry from *both* lists falls through to the `else` and is appended to `dest_ip`. Even if the test matched, the branch assigns to a stray `scp_ip` rather than `src_ip`. Downstream the rule is built with `${src_ip:+-s} $src_ip` / `${dest_ip:+-d} $dest_ip`, so `-s` is never emitted and `dest_ip` ends up containing the source addresses. Both the sentinel string and the target variable need to be `src_ip`:

```suggestion
		if [ "${direction}" = "src_ip" ]; then
			export src_ip="${src_ip}${ipaddr},"
		else
			export dest_ip="${dest_ip}${ipaddr},"
		fi
```

---

openwrt-ai (commented):
Reviewed 1 new commit.

## Commit checks
- Commit subject has a typo: "hanling" -> "handling". Otherwise the message accurately describes the diff (converting `src_ip`/`dest_ip` to UCI lists). Note this is marked `[RFC]` and the body lists the LuCI change and an upgrade/migration script as still-TODO, so the conversion is not yet complete on its own.

One inline issue flagged: the `direction`/`scp_ip` typos cause `src_ip` matching to be dropped and source addresses to be merged into `dest_ip`.

No CI checks are configured for this PR's head commit, so nothing to ground against there.

---

PR #17240, opened 2021-11-30
[19.07] nextdns: Update to version 1.37.5

rs:
Update to version 1.37.5


BKPepe:
The PR was created by mistake, right? While looking into this https://github.com/openwrt/packages/pull/17158 which didnt make it through.

rs:
It's created automatically by our CI. We might have to drop support for 19.07 if we can't find another solution. Feel free to close those two PRs for now.

PR #17243, opened 2021-11-30
python-rcssmin: bump to version 1.1.0

cotequeiroz:
Maintainer: me, @commodo 
Compile tested: mediatek, Linksys EA8450, master
Run tested: mediatek, Linksys EA8450, master, runn `python -mrcssmin </var/tmp/jsdoc-default.css`

Description:
Relevant changes:
 * quoted data urls which are not base64 encoded keep their spaces now
 * accept bytes and text as input. All other types now raise a TypeError
 * update python & gcc support
 * python version will only accept the C implementation if the versions match exactly. This should prevent using older installed C versions.

Along with the version bump:
 - update maintainer email address
 - use $(AUTORELEASE)

### DIFF
--- a/lang/python/python-rcssmin/Makefile
+++ b/lang/python/python-rcssmin/Makefile
@@ -6,15 +6,15 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-rcssmin
-PKG_VERSION:=1.0.6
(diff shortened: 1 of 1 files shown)


commodo:
thank you :)

this was on my to-do list, but i am on vacation this week

PR #17247, opened 2021-12-01
zerotier: add configuration reload trigger

jow-:
Maintainer: @mwarning 
Compile tested: -
Run tested: -

Description:

Ensure that the zerotier service is automatically restarted when the uci
configuration is modified.

Ref: https://forum.openwrt.org/t/reloading-a-service-when-config-changes/113488/3

### DIFF
--- a/net/zerotier/files/etc/init.d/zerotier
+++ b/net/zerotier/files/etc/init.d/zerotier
@@ -115,3 +115,12 @@ stop_service() {
 	config_foreach stop_instance 'zerotier'
 	rm -f ${CONFIG_PATH}
 }
+
+reload_service() {
+	stop
+	start
+}
+
+service_triggers() {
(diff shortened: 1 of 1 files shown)


mwarning:
How is this different from `procd_set_param respawn`?
Anyway, otherwise LGTM.

jow-:
> How is this different from procd_set_param respawn?

The respawn parameter controls whether if and how the process is supposed to get restarted by procd if it terminates. The reload trigger on the other hand instructs procd to re-invoke the init script reload action if /etc/config/zerotier changed.

mwarning:
understood, thank you.

mwarning:
please merge

BKPepe:
It should be backported to OpenWrt 21.02, right?

mwarning:
@BKPepe good idea

PR #17250, opened 2021-12-02
[19.07] libs/c-ares: fix domain hijacking CVE-2021-3672

ynezz:
Maintainer: @karlp 
Compile tested: arm_cortex-a9
Run tested: None

Description:

Missing input validation of host names returned by Domain Name Servers
in the c-ares library can lead to output of wrong hostnames (leading to
Domain Hijacking).

I've just taken [patch](https://github.com/c-ares/c-ares/compare/809d5e8..44c009b.patch) from the [advisory](https://c-ares.org/adv_20210810.html) and rebased it onto 1.15.0 version.

### DIFF
--- a/libs/c-ares/Makefile
+++ b/libs/c-ares/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=c-ares
 PKG_VERSION:=1.15.0
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 PKG_LICENSE:=MIT
 PKG_CPE_ID:=cpe:/a:c-ares_project:c-ares
 
--- a/libs/c-ares/patches/0001-ares_expand_name-should-escape-more-characters.patch
+++ b/libs/c-ares/patches/0001-ares_expand_name-should-escape-more-characters.patch
@@ -1,0 +1,88 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: bradh352 <brad@brad-house.com>
+Date: Fri, 11 Jun 2021 11:27:45 -0400
+Subject: [PATCH] ares_expand_name() should escape more characters
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+RFC1035 5.1 specifies some reserved characters and escaping sequences
+that are allowed to be specified.  Expand the list of reserved characters
+and also escape non-printable characters using the \DDD format as
+specified in the RFC.
+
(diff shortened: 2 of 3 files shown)


ynezz:
BTW I've looked at the security related fixes since 1.15.0 and I'm not sure if this CVE fix is going to cut it and make it less vulnerable:

1.16.1 fixes following security issues:
 * Prevent possible use-after-free and double-free in ares_getaddrinfo() if ares_destroy() is called prior to ares_getaddrinfo() completing. Reported by Jann Horn at Google Project Zero. 
 
1.17.0 fixes following security issues:
  * avoid read-heap-buffer-overflow in ares_parse_soa_reply found during fuzzing
  * Avoid theoretical buffer overflow in RC4 loop comparison
  * Empty hquery->name could lead to invalid memory access
  * ares_parse_{a,aaaa}_reply() could return a larger *naddrttls than was passed in 

1.17.2 fixes following security issues:
  * NodeJS passes NULL for addr and 0 for addrlen to ares_parse_ptr_reply() on systems where malloc(0) returns NULL. This would cause a crash.
  * When building c-ares with CMake, the RANDOM_FILE would not be set and therefore downgrade to the less secure random number generator
  * If ares_getaddrinfo() was terminated by an ares_destroy(), it would cause a crash
  * Crash in sortaddrinfo() if the list size equals 0 due to an unexpected DNS response
  * Expand number of escaped characters in DNS replies as per RFC1035 5.1 to prevent spoofing follow-up
  * Perform validation on hostnames to prevent possible XSS due to applications not performing valiation themselves

I'm wondering if it wouldn't be better to simply upgrade to 1.17.2, which is not vulnerable to this CVE and contains all the previous security fixes. According to the my manual check of the changes between 1.15.0->1.17.2 and [ABI Laboratory](https://abi-laboratory.pro/index.php?view=timeline&l=c-ares) it seems like they're doing good work regarding the API/ABI compatibility so it shouldn't probably cause much regressions, but we can't be sure until we test it.

The other option is to backport those security fixes, but this might introduce regressions as well and we would need to test it anyway.

karlp:
I'm still concerned about making the major version change this late in 19.07, there's only three months til EOL as it is.  see https://github.com/openwrt/packages/pull/16312#issuecomment-896997199 for comments on how much recent fixes was fixing breakages.  They might have kept ABI, but that's still two major version changes.

ynezz:
>  see #16312 (comment) for comments on how much recent fixes was fixing breakages

Opps, I've completely forget about this one, thanks for reminding me.

> I'm still concerned about making the major version change this late in 19.07, there's only three months til EOL as it is.

Fair enough, I understand.

BKPepe (approved):

karlp:
So you've both just agreed to take over this package then ok?

PR #17257, opened 2021-12-02
Add new lua module: ipops

ptpt52:
This is very usefull to handle ip calculation

for example convert an network strings to formated ipcidir string:
```
nets="1.2.3.4,192.168.1.0/24,192.168.100.100-192.168.200.222"
``` 
`lua ipops.lua netStrings2ipcidrStrings "1.2.3.4,192.168.1.0/24,192.168.100.100-192.168.200.222"`
OUTPUT:
```
1.2.3.4/32,192.168.1.0/24,192.168.100.100/30,192.168.100.104/29,192.168.100.112/28,192.168.100.128/25,192.168.101.0/24,192.168.102.0/23,192.168.104.0/21,192.168.112.0/20,192.168.128.0/18,192.168.192.0/21,192.168.200.0/25,192.168.200.128/26,192.168.200.192/28,192.168.200.208/29,192.168.200.216/30,192.168.200.220/31,192.168.200.222/32
```

also the ipops.lua could be imported to any lua program and use it as an module

### DIFF
--- a/lang/lua-ipops/Makefile
+++ b/lang/lua-ipops/Makefile
@@ -1,0 +1,40 @@
+#
+# Copyright (C) 2017-2019 Chen Minqiang <ptpt52@gmail.com>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lua-ipops
+PKG_VERSION:=0.0.1
+PKG_RELEASE:=1
+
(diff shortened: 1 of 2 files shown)


jow-:
Note that there already is https://openwrt.github.io/luci/api/modules/luci.ip.html and https://openwrt.github.io/luci/api/modules/luci.ip.cidr.html

ptpt52:
@jow- 
aha.
I do not notice that there is an API that can handle ip range string (`192.168.1.100-192.168.1.200`)
By the way, this ipops.lua is pure Lua

BKPepe:
So, this can be closed, right?

PR #17261, opened 2021-12-03
nfs-kernel-server: fix build on MacOs arm64

ptpt52:
|checking for prctl... no
  |configure: error: prctl syscall is not available

### DIFF
--- a/net/nfs-kernel-server/patches/host/10-macos-build-hack.patch
+++ b/net/nfs-kernel-server/patches/host/10-macos-build-hack.patch
@@ -1,0 +1,55 @@
+diff --git a/configure b/configure
+index 53b8cef..a30c839 100755
+--- a/configure
++++ b/configure
+@@ -5880,13 +5880,6 @@ fi
+ 
+ 
+ 
+-    ac_fn_c_check_func "$LINENO" "prctl" "ac_cv_func_prctl"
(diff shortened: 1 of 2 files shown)


ptpt52:
strang, my local build is good
but here buildbot failed
what I am trying to do is do patch for Host build only, how to achive it?

neheb:
Why not just make a patch for both builds?

As far as host only patches, I had a commit adding support for that but was rejected AFAIK.

neheb:
This one: https://github.com/openwrt/openwrt/pull/4030/commits/28b2786f2363cac382a259dc0445852fdbdc511b

svlobanov:
@ptpt52  @neheb I proposed another patch to solve this issue https://github.com/openwrt/packages/pull/17313

net/nfs-kernel-server/patches/host/10-macos-build-hack.patch
```diff
@@ -0,0 +1,55 @@
+diff --git a/configure b/configure
```
BKPepe, line 1: You should send this to upstream. You even mention that it is a hack in the file naming, but we should not have hacks forever, otherwise we will take care of it forever.

BKPepe (changes requested):

svlobanov:
due to #17313 has been merged let's close this PR

PR #17263, opened 2021-12-03
htop: update to 3.1.2-1

graysky2:
Build-tested: x86/64
Run-tested: bcm2711/RPi4B

Description: new upstream release


hnyman:
Any objections, @champtar  ?

(The x86 CI errors look like perl related, not due to this package)

champtar:
Sorry for the delay, can you just confirm during the configure step that nothing is autodetect ? We had some problem in the past with option added / renamed

graysky2:
@champtar - Following this [wiki](https://openwrt.org/docs/guide-developer/toolchain/single.package) page, my goal is to run `make -j1 V=s pacakge/htop/compile` and inspect the output.  But I am unable to get the package to compile.

I have my target architecture from `make menuconfig` selected.  Then, the following two commands complete successfully:
```
make tools/install
make toolchain/install
```

But I am unable to build either dependency listed by `opkg info htop` nor the suggested command: `make package/htop/compile` ... what am I doing wrong?

```
% make -j1 V=s pacakge/htop/compile
make[1]: Entering directory '/scratch/union'
make[1]: *** No rule to make target 'pacakge/htop/compile'.  Stop.
make[1]: Leaving directory '/scratch/union'
make: *** [/scratch/union/include/toplevel.mk:230: pacakge/htop/compile] Error 2
```

graysky2:
@champtar - [Here](https://gist.github.com/graysky2/6b25057ef73c9e5549a32f1b03ecfee9) is the build log from `make -j1 V=s package/htop/compile`.  Does it answer your question?

hnyman:
Looking at the htop configure.ac changelog back from 3.1.2,

https://github.com/htop-dev/htop/commits/d73783d6dbec4a8d9d20d744bf2b5e6a7073aa46/configure.ac

it seems that detection of libunwind has been added with commit

https://github.com/htop-dev/htop/commit/29e1fcfa0517a7c5770b2a95fe7fbc0bc197c360

Likely we should disable that detection by adding `--disable-unwind \` line to the CONFIGURE_ARGS in Makefile

That way we could avoid the libunwind dependency error in buildbot (that builds all libraries and so also libunwind will be present for detection, unlike in private build envs.).

graysky2:
@hnyman 
> Likely we should disable that detection by adding --disable-unwind \ line to the CONFIGURE_ARGS in Makefile

Force pushed with that modification, build was successful (`make -j1 V=s package/htop/compile` output is [here](https://gist.github.com/graysky2/334a61d3055bc74853a3fd333521ca31)).  Corresponding package installs and runs as expected.

champtar:
Thanks you both !

PR #17265, opened 2021-12-03
modemmanager: refactoring procd init script

feckert:
Maintainer: @aleksander0m @nickberry17 
Run tested: x86_64, APU3, latest OpenWrt master, started a modemmanger instance)

Description:
The way the init script is written now, we get a bad output when calling
the ubus service backend.

```
ubus call service list "{'verbose':true,'name':'modemmanager'}"
{
        "modemmanager": {
                "instances": {
                        "instance1": {
                                "running": true,
                                "pid": 20511,
                                "command": [
                                        "sh",
                                        "-c",
                                        ".
/usr/share/ModemManager/modemmanager.common;    \t
mkdir -m 0755 -p /var/run/modemmanager;          \t
mm_cleanup_interfaces;                            \t
( mm_report_events_from_cache ) >/dev/null 2>&1 & \t
/usr/sbin/ModemManager"
                                ],
                                "term_timeout": 5,
                                "respawn": {
                                        "threshold": 3600,
                                        "timeout": 5,
                                        "retry": 5
                                },
                                "pidfile":"/var/run/modemmanager/modemmanager.pid"
                        }
                }
        }
}
```

I also get the following output in the log that the PID file cannot be created:

`daemon.err procd: Failed to remove pidfile: :No such file or directory`

The changes in this commit fixes this issues

### DIFF
--- a/net/modemmanager/files/modemmanager.init
+++ b/net/modemmanager/files/modemmanager.init
@@ -22,14 +22,12 @@ start_service() {
 	# it starts.
 	#
 	# All these commands need to be executed on every MM start, even after
-	# procd-triggered respawns, which is why they're all included as instance command
+	# procd-triggered respawns, which is why this is wrapped in a startup
+	# wrapper script called '/usr/sbin/ModemManager-wrapper'.
 	#
+	. /usr/share/ModemManager/modemmanager.common
 	procd_open_instance
-	procd_set_param command sh -c ". /usr/share/ModemManager/modemmanager.common;    \
-	                               mkdir -m 0755 -p ${MODEMMANAGER_RUNDIR};          \
-	                               mm_cleanup_interfaces;                            \
-	                               ( mm_report_events_from_cache ) >/dev/null 2>&1 & \
-	                               /usr/sbin/ModemManager"
+	procd_set_param command /usr/sbin/ModemManager-wrapper
 	procd_set_param respawn "${respawn_threshold:-3600}" "${respawn_timeout:-5}" "${respawn_retry:-5}"
 	procd_set_param pidfile "${MODEMMANAGER_PID_FILE}"
 	procd_close_instance
--- a/net/modemmanager/files/modemmanager.init
+++ b/net/modemmanager/files/modemmanager.init
@@ -4,6 +4,8 @@
 USE_PROCD=1
 START=70
 
+LOG_LEVEL="INFO"
+
 stop_service() {
 	# Load common utils
 	. /usr/share/ModemManager/modemmanager.common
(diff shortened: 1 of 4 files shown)


net/modemmanager/files/modemmanager.init
```diff
@@ -5,32 +5,24 @@ USE_PROCD=1
...
-	#
-	# The cached events processing will wait for MM to be available in DBus
-	# and will make sure all ports are re-notified to ModemManager every time
-	# it starts.
-	#
-	# All these commands need to be executed on every MM start, even after
-	# procd-triggered respawns, which is why they're all included as instance command
-	#
+start_mm() {
+	. /usr/share/ModemManager/modemmanager.common
```
aleksander0m, line 13: This is more or less what we had before, with the daemon call as only command in the procd instance. But having it that way ended up failing because mm_report_events_from_cache was not being called if the service was restarted by procd itself e.g. after a daemon crash (as explained in the comment that was also removed: "All these commands need to be executed on every MM start, even after procd-triggered respawns, which is why they're all included as instance command").

How are these changes avoiding that? If these changes prevent that problem, I'm all for them, but I'm not really sure if they do. Isn't procd supposed to re-run the "command" configured only on daemon crashes?

aleksander0m (changes requested):

net/modemmanager/files/modemmanager.init
feckert: If that's the case then I'd rather put it all in a wrapper/startup script `mm-service`?
Then it looks better right away.
We also could fix the two issues as mentioned (`ubus service` and `pdifle`)

aleksander0m: Yes, that is probably the cleanest way forward

feckert: @aleksander0m I have now outsourced the startup sequence to a wrapper script.

feckert:
@aleksander0m Can you look at that again.
I have now run shellcheck and fixes 4 things.

1. Disable SC2064 warning https://github.com/openwrt/packages/pull/17265/files#diff-2a022c1a8090bf6224c600ad07b0756c4f811b1b2b15ab7612a7bca07f3ee346R6

2. Split this command to fix SC2174
https://github.com/openwrt/packages/pull/17265/files#diff-2a022c1a8090bf6224c600ad07b0756c4f811b1b2b15ab7612a7bca07f3ee346R21

3. Made CHILD variable global https://github.com/openwrt/packages/pull/17265/files#diff-2a022c1a8090bf6224c600ad07b0756c4f811b1b2b15ab7612a7bca07f3ee346R26

4. Removed unused variables

nickberry17:
ping @neheb LGTM

PR #17276, opened 2021-12-06
speedtestpp: add new package

oskarirauta:
SpeedTest++
Yet another unofficial speedtest.net client cli interface

For users who instead of python based speedtest client want
to use something that was written in c++...

Description:
great alternative for existing python3-speedtest-cli. Results as good as results with python based client and actually, really similar output of data also..

git version chosen because latest release(tag) is quite old and
there has been a rather many commits since it was released.

I made it to conflict with python3-speedtest-cli due to binary's filename being same with python3-speedtest-cli (/usr/bin/speedtest) - which could had been avoided by using default binary filename which was `SpeedTest` but for clarity reasons I made it to use all lowercased letters.. Choice to do this is open to converse here, should confliction be removed and original filename kept instead of doing this, since I only did this as it was my own opinion, but not absolute only available option.

### DIFF
--- a/net/speedtestpp/Makefile
+++ b/net/speedtestpp/Makefile
@@ -1,0 +1,42 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=speedtestpp
+PKG_VERSION:=1.14
+PKG_RELEASE:=$(AUTORELEASE)
+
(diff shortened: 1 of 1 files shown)


PolynomialDivision:
lgtm

PR #17282, opened 2021-12-07
openwisp-config: refactoring and update openwisp-config package

feckert:
Maintainer: @nemesisdesign 
Compile tested: x86_64, APU3, latest OpenWrt
Run tested: No (no source changes)

Description:
With the last upstream change, all build variants were removed so that we only have one package.
It is now called openwisp-config.

### DIFF
--- a/admin/openwisp-config/Makefile
+++ b/admin/openwisp-config/Makefile
@@ -5,48 +5,43 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openwisp-config
-PKG_SOURCE_VERSION:=0.5.0
 PKG_RELEASE:=1
 
-PKG_SOURCE_PROTO:=git
-PKG_SOURCE_URL:=https://github.com/openwisp/openwisp-config.git
-PKG_MIRROR_HASH:=337a3a9542a0898da9f951256b0d19b6bc87ced98f4ec6dc9646172b551880ef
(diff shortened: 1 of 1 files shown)


devkapilbansal:
Do we need to open PRs to include it in `OpenWRT 19.07` and `OpenWRT 21.02` ?

feckert:
@devkapilbansal We should not make such a big change in stable releases, because the stable branches are feature complete
We should only add bug fixes and security patches.

Not that we break something in a stable branch, that would be unpleasant.
In the master branch, people have time to test such big changes and give feedback.

I would suggest that we make this changes to the master branch first. And then we will see if we can roll this out to the stable branches as well.

admin/openwisp-config/Makefile
```diff
@@ -5,48 +5,43 @@
...
+  SECTION:=admin
+  SUBMENU:=openwisp
+  DEPENDS:=+curl \
+    +lua \
+    +libuci-lua \
+    +luafilesystem \
+    +luci-lib-nixio \
+    +ca-certificates
+  PKGARCH:=all
+  URL:=http://openwisp.org
```
BKPepe, line 34:
```suggestion
  URL:=https://openwisp.org/
```

nickberry17:
LGTM? @feckert @BKPepe

PR #17286, opened 2021-12-08
gg: add new package

1715173329:
Maintainer: me
Compile tested: ipq807x, rockchip/armv8, x86/64
Run tested: nanopi-r2s, x86_64 (via VMware)

Description:
gg is a command-line tool for one-click proxy in your research and development.
You can just add gg before another command to redirect its traffic to your proxy without installing v2ray or anything else. Usage example: gg python -m pip install torch.

Compared to proxychains or graftcp, it has the following advantages:
- Use it independently without any other proxy utils.
- UDP support.
- More friendly configuration.

For documentation please move to https://github.com/mzz2017/gg.

### DIFF
--- a/net/gg/Makefile
+++ b/net/gg/Makefile
@@ -1,0 +1,53 @@
+# SPDX-License-Identifier: GPL-3.0-only
+#
+# Copyright (C) 2021 ImmortalWrt.org
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gg
+PKG_VERSION:=0.2.1
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/mzz2017/gg/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=51d1ccfb10ba0d9e8fdd68f4a566b8ed772980be1cf1d03da1969313f88b8bb6
+
+PKG_MAINTAINER:=Tianling Shen <cnsztl@immortalwrt.org>
+PKG_LICENSE:=AGPL-3.0-only
+PKG_LICENSE_FILES:=LICENSE-AGPL
+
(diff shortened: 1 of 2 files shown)


jefferyto (approved):
The package looks fine to me, though their LICENSE file [says](https://github.com/mzz2017/gg/blob/main/LICENSE):

> The license for this project is based on AGPL as stated in another file in the repository, with an additional one:
> 
> The contributors allows the owner of the repository (mzz2017) to change the license of this repository.

IANAL so I don't know if this additional requirement:

* is actually allowable in a software license
* creates a new (but very similar to AGPL) license
* is compatible with AGPL

Not sure if this is a concern here.

1715173329:
Ping @mzz2017

mzz2017:
This is just an license for contributors, and the code completely follows AGPL license.

1715173329:
Changed `PKG_LICENSE_FILES`, now points to `LICENSE-AGPL`.

jefferyto:
> This is just an license for contributors, and the code completely follows AGPL license.

AFAIK you can’t have one license for contributors and another license for users, and you can’t arbitrarily modify GPL/AGPL and still call it GPL

mzz2017:
@jefferyto Thanks for your pointing. According to [Terms of Service of Github](https://docs.github.com/en/github/site-policy/github-terms-of-service#6-contributions-under-repository-license):
> Whenever you add Content to a repository containing notice of a license, you license that Content under the same terms, and you agree that you have the right to license that Content under those terms. If you have a separate agreement to license that Content under different terms, such as a contributor license agreement, that agreement will supersede.

`gg` just added a separate agreement to the license as the contributor license agreement.

But I agree to change the filename `LICENSE` to `contributor-agreement`, and `LICENSE-AGPL` to `LICENSE`.

jefferyto:
@mzz2017 

> `gg` just added a separate agreement to the license as the contributor license agreement.
> 
> But I agree to change the filename `LICENSE` to `contributor-agreement`, and `LICENSE-AGPL` to `LICENSE` to avoid ambiguity.

Thanks, that sounds fine to me.

PR #17295, opened 2021-12-09
nft-qos: fix include on build

wulfy23:
commit: f88485f572ec1ff9082106ccf0ccb20fc7af5801 which adds
IPKG_INSTROOT to source is ineffective for this package.

Source includes only when empty.

### DIFF
--- a/net/nft-qos/Makefile
+++ b/net/nft-qos/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nft-qos
 PKG_VERSION:=1.0.6
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 PKG_LICENSE:=GPL-2.0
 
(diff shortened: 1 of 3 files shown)


BKPepe:
Since you previously did such changes, the commit description should include:

`Fixes: commit hash ("commit title")` and also the commit description should be more clear.

Anyway, this package is not actively maintained, and it is a question if it should not be removed.

wulfy23:
Thankyou for the guidance

Added suggested  commit message improvements and repushed...

Valid point re: removing unmaintained package... considered this also...

Given OpenWrt is making movements towards nft... imho, its close but not
quite on it's last legs just yet...

There are a handful of users, and a handful of people capable of further adjustments
in the near future.

It's luci lua addon is perhaps more relavent in this regard

neheb:
PKG_RELEASE needs to be bumped.

PR #17296, opened 2021-12-09
php7/php7-pecl-*: drop PHP7 packages

mhei:
Maintainer: me + @flyn-org 
Compile tested: mxs, x64_64
Run tested: mxs

Description:

This PR migrates most of PHP 7 PECL packages to PHP 8 and finally removes PHP 7 support.
Since active support for PHP 7.4 branch ended some days before, and we have PHP 8 in place,
it seems reasonable to focus maintenance work on trunk on PHP 8.


flyn-org:
Sounds great! I recommend we merge. I will keep on eye on things after that is done.

Issue #17284, opened 2021-12-07
modemmanager: unable to start service on mips_24kc


feckert:
Maintainer: @aleksander0m  @nickberry17
Environment: lantiq_xrx200, latest OpenWrt master

Description:

I do not know exactly what the problem here is but I have two targets an lantiq_xrx200 device and an x86_64 (APU3) device.
I am using the same ModemManager version 1.18.2 but on my lantiq_xrx200 (this is mips_24kc target) the ModemManger could not start. I get the following messages in the logs

```
Tue Dec  7 17:02:19 2021 daemon.info [17068]: <info>  [1638892939.268189] ModemManager (version 1.18.2) starting in system bus...
Tue Dec  7 17:02:19 2021 daemon.debug [17068]: <debug> [1638892939.269515] [charsets] detecting platform iconv() support...
Tue Dec  7 17:02:19 2021 daemon.debug [17068]: <debug> [1638892939.270638] [charsets]   UTF-8: iconv conversion to/from charset is supported
Tue Dec  7 17:02:19 2021 daemon.debug [17068]: <debug> [1638892939.271902] [charsets]   UCS-2BE: iconv conversion to charset not supported
Tue Dec  7 17:02:19 2021 daemon.debug [17068]: <debug> [1638892939.273176] [charsets]   ASCII: iconv conversion to/from charset is supported
Tue Dec  7 17:02:19 2021 daemon.debug [17068]: <debug> [1638892939.274376] [charsets]   ISO8859-1: iconv conversion to charset not supported
Tue Dec  7 17:02:19 2021 daemon.debug [17068]: <debug> [1638892939.275581] [charsets]   CP437: iconv conversion to charset not supported
Tue Dec  7 17:02:19 2021 daemon.debug [17068]: <debug> [1638892939.276763] [charsets]   CP850: iconv conversion to charset not supported
Tue Dec  7 17:02:19 2021 daemon.debug [17068]: <debug> [1638892939.277928] [charsets]   UTF-16BE: iconv conversion to/from charset is supported
Tue Dec  7 17:02:19 2021 daemon.warn [17068]: <warn>  [1638892939.310430] could not acquire the 'org.freedesktop.ModemManager1' service name
Tue Dec  7 17:02:19 2021 daemon.info [17068]: <info>  [1638892939.311374] ModemManager is shut down
```
Can anyone confirm this? Could it be due to the different endianness of the processors?

- x86_64 little endieanness
- mips_24kc big endianness

[...]

PR #17300, opened 2021-12-10
dbus: add debug option and move service to procd and update/fix compile CMAKE_OPTIONS

feckert:
Maintainer: @neheb @diizzyy 
Compile tested: x86_64 and lantiq_xrx200, APU3, latest OpenWrt master
Run tested: lantiq_xrx200, latest OpenWrt master

Description:
I wanted to try the ModemManager which has the dbus as a dependency.
But the ModemManagter does have some problems with the DBus on my xrx200 target https://github.com/openwrt/packages/issues/17284
I noticed a few things during ModemManager debugging that I would like to fix or improve in the DBus packages by this pull request to simplify the debugging


utils/dbus/Makefile
```diff
@@ -74,7 +74,6 @@ endef
 CMAKE_OPTIONS += \
 	-DDBUS_SYSTEM_SOCKET=/var/run/dbus/system_bus_socket \
 	-DDBUS_SESSION_SOCKET_DIR=/tmp \
-	-DDBUS_SYSTEM_PID_FILE=/var/run/dbus.pid \
```
neheb, line 77: You sure about this?

feckert: The cmake overview gives me the following output when building.
```
        System bus address:       unix:path=/var/run/dbus/system_bus_socket
        System bus socket:        /var/run/dbus/system_bus_socket
        System bus PID file:      /var/run/dbus/pid
        System bus user:          root
```

This is also the default pid location.
The procd writes its own pid file that is located under /var/run/dbus.pid
if I specify the one from the dbus in the procd, then the dbus service cannot start!

```
root@VR2-106149 /tmp/run/dbus # ls -la
drwxr-xr-x    2 root     root            80 Dec 10 14:16 .
drwxr-xr-x   11 root     root           620 Dec 10 14:05 ..
-rw-r--r--    1 root     root             6 Dec 10 14:16 pid
srwxrwxrwx    1 root     root             0 Dec 10 14:16 system_bus_socket
```

neheb: hmm OK.

utils/dbus/files/dbus.init
```diff
@@ -3,15 +3,32 @@
...
-	service_start /usr/bin/dbus-daemon --system
+
+	procd_open_instance
+	procd_set_param command "${PROG}"
+	procd_append_param command --system
+	procd_append_param command --nofork
+	[ -n "$DEBUG" ] && procd_set_param env DBUS_VERBOSE=1
+	procd_set_param stdout 1
+	procd_set_param stderr 1
+	procd_set_param pidfile "$PIDFILE"
```
neheb, line 28: so my understanding of this is that it's for daemons that use a pidfile but don't create it themselves. Since dbus does this, this is unneeded.

feckert:
@neheb Then this should fit now.
I have now started and stopped the dbus again without the procd pid.
Works as expected.

```
root@VR2-106149 ~ # ps | grep dbus
10836 root      1416 S    grep dbus
root@VR2-106149 ~ # /etc/init.d/dbus start
root@VR2-106149 ~ # ubus call service list '{"name":"dbus","verbose":true}'
{
        "dbus": {
                "instances": {
                        "instance1": {
                                "running": true,
                                "pid": 10860,
                                "command": [
                                        "/usr/bin/dbus-daemon",
                                        "--system",
                                        "--nofork"
                                ],
                                "term_timeout": 5
                        }
                }
        }
}
root@VR2-106149 ~ # ps | grep dbus
10860 root      1696 S    /usr/bin/dbus-daemon --system --nofork
10899 root      1416 S    grep dbus
root@VR2-106149 ~ # /etc/init.d/dbus stop
root@VR2-106149 ~ # ubus call service list '{"name":"dbus","verbose":true}'
{

}
root@VR2-106149 ~ # ps | grep dbus
10934 root      1416 S    grep dbus
```

neheb:
btw are you seeing behavior like this? https://github.com/openwrt/packages/pull/15442#issuecomment-822160791

feckert:
I have now started and stopped the dbus service a few times and so far this has not happened.
From my point of view, the procd takes care of it and this does not happen.

neheb:
note that the comment is a result of the conversion to cmake. not really sure why.

If you want to backport the procd stuff to 21.02, feel free. I don't think this is an issue there though.

feckert:
@neheb Can you take another look at the last [commit](https://github.com/openwrt/packages/pull/17300/commits/4c1f20f4c999bc3505248de3c2e60c511b2b4e37) I added? From my point of view, it makes sense to remove the` dbus-update-activation-envirenment` tool, since we only use the system bus anyway, and add the `dbus-test-tool` to add debug possibility if dbus does not work as expected.

neheb:
Looks good. I don't use dbus anymore.

PR #17313, opened 2021-12-11
nfs-kernel-server: fix build on macos arm64

svlobanov:
Previously, @ptpt52 user propose https://github.com/openwrt/packages/pull/17261 but that PR is not compatible with non-macos building hosts.

This patch doesn't not break compilation on non-macos building hosts

1. prctl() check is not required for host-compile on macos so this
check is removed from configure script for darwin targets

2. __DARWIN_ONLY_64_BIT_INO_T is true on macos arm64 so struct stat64
and stat64() are not available. This patch defines stat64 as stat if
__DARWIN_ONLY_64_BIT_INO_T is true

Maintainer: Peter Wagner <tripolar@gmx.at>
Run tested: mipsel; tplink,tl-mr3020-v3; OpenWrt version 56b14fdeb24b63b26ae948aa2fa6b1ce4b9df1fd, tests done

### DIFF
--- a/net/nfs-kernel-server/patches/200-fix-macos-build.patch
+++ b/net/nfs-kernel-server/patches/200-fix-macos-build.patch
@@ -1,0 +1,17 @@
+fix stat64 issue for modern macos versions (including macos arm64)
+
+--- a/tools/rpcgen/rpc_main.c
++++ b/tools/rpcgen/rpc_main.c
+@@ -62,6 +62,12 @@
+ #define EXTEND	1		/* alias for TRUE */
+ #define DONT_EXTEND	0	/* alias for FALSE */
+ 
++#ifdef __APPLE__
++# if __DARWIN_ONLY_64_BIT_INO_T
++#  define stat64 stat
++# endif
++#endif
++
+ struct commandline
+   {
+     int cflag;			/* xdr C routines */
--- a/net/nfs-kernel-server/Makefile
+++ b/net/nfs-kernel-server/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nfs-kernel-server
 PKG_VERSION:=2.5.4
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 PKG_HASH:=546ce4b51eeebc66e354b6cc6ca0ce509437efbdef0caaf99389534eef0e598b
 
 PKG_SOURCE_URL:=@SF/nfs
@@ -150,6 +150,7 @@ HOST_CONFIGURE_VARS += \
 	ac_cv_header_nfsidmap_h=yes \
 	ac_cv_header_blkid_blkid_h=yes \
 	ac_cv_lib_resolv___res_querydomain=yes \
+	ac_cv_func_prctl=yes \
 	GSSGLUE_CFLAGS=" " \
 	GSSGLUE_LIBS=" " \
 	RPCSECGSS_CFLAGS=" " \


svlobanov:
@neheb please approve starting CI checks

neheb:
Requires patch to be refreshed and PKG_RELEASE bump.

svlobanov:
@neheb sorry, I forgot to bump. done. please approve again

svlobanov:
My patch is not dirty, I've checked it manually, it is applied normally without any warnings:

```
sergey@Sergeys-MacBook-Air nfs % rm -rf nfs-utils-2.5.4
sergey@Sergeys-MacBook-Air nfs % tar -xf nfs-utils-2.5.4.tar.xz 
sergey@Sergeys-MacBook-Air nfs % cd nfs-utils-2.5.4
sergey@Sergeys-MacBook-Air nfs % cp ../200-fix-macos-build.patch 200-fix-macos-build.patch 
sergey@Sergeys-MacBook-Air nfs-utils-2.5.4 % patch -p1 --dry-run < 200-fix-macos-build.patch 
patching file configure
patching file tools/rpcgen/rpc_main.c
sergey@Sergeys-MacBook-Air nfs-utils-2.5.4 % patch -p1  < 200-fix-macos-build.patch 
patching file configure
patching file tools/rpcgen/rpc_main.c
```

stangri:
> My patch is not dirty, I've checked it manually, it is applied normally without any warnings:

From my very limited experience with CI complaining about the patches it's when something (like an extra line) in the files that need to be patched changed, so locally you can still build the package, but the more strict CI checks complain about them.

Usually it's fixed by re-creating the patch-file from scratch (https://openwrt.org/docs/guide-developer/toolchain/use-patches-with-buildsystem).

svlobanov:
> > My patch is not dirty, I've checked it manually, it is applied normally without any warnings:
> 
> From my very limited experience with CI complaining about the patches it's when something (like an extra line) in the files that need to be patched changed, so locally you can still build the package, but the more strict CI checks complain about them.
> 
> Usually it's fixed by re-creating the patch-file from scratch (https://openwrt.org/docs/guide-developer/toolchain/use-patches-with-buildsystem).

thanks for the comment. the logs are here https://github.com/openwrt/packages/runs/4494891334?check_suite_focus=true , looks like patches were applied twice, maybe it happened because I forgot to bump PKG_VERSION at the first time

neheb:
run make/package/nfs-kernel-server/{clean,refresh}

ptpt52:
great!

svlobanov:
> run make/package/nfs-kernel-server/{clean,refresh}

thank you, refreshed. that was a 'diff --git ...' garbage in the patch file

net/nfs-kernel-server/patches/200-fix-macos-build.patch
```diff
@@ -0,0 +1,40 @@
+fixes two issue: prctl in configure and stat64 for modern macos 
+versions (including macos arm64)
```
BKPepe, line 2: Can you send this to upstream?

BKPepe (changes requested):

net/nfs-kernel-server/patches/200-fix-macos-build.patch
svlobanov: I have found really good solution to avoid ./cofigure patching. New patch doesn't have configure hack, it disables prctl() check via HOST_CONFIGURE_VARS

svlobanov: I will send stat64 fix to upstream

svlobanov: stat64 patch has been to upstream: https://bugzilla.linux-nfs.org/show_bug.cgi?id=372

Issue #17311, opened 2021-12-11
fail2ban: does not compile (_2to3) is invalid


BKPepe:
Maintainer: @erdoukki
Environment: aarch64, Turris MOX, OpenWrt 21.02.1

Description:

```
Applying ./patches/001-fail2ban-fix-python3-script.patch using plaintext: 
patching file config/filter.d/ignorecommands/apache-fakegooglebot

Applying ./patches/CVE-2021-32749.patch using plaintext: 
patching file config/action.d/complain.conf
patching file config/action.d/dshield.conf
patching file config/action.d/mail-buffered.conf
patching file config/action.d/mail-whois-lines.conf
patching file config/action.d/mail-whois.conf
patching file config/action.d/mail.conf
error in fail2ban setup command: use_2to3 is invalid.
make[2]: *** [Makefile:84: /home/beast/beast/workspace/turris-os-packages-dragons-mox/build/build_dir/target-aarch64_cortex-a53_musl/fail2ban-0.11.2/.built] Error 1
time: package/feeds/packages/fail2ban/compile#1.82#0.20#2.86
```

erdoukki:
Thanks for the report.

May be it is related to https://github.com/openwrt/packages/issues/15099 ?

 I will look at this issue...

erdoukki:
Fail2ban upstream look to have already resolved the issue : https://github.com/fail2ban/fail2ban/issues/3098

I will test and propose a new fail2ban package based on latest upstream release !

erdoukki:
> Fail2ban upstream look to have already resolved the issue : [fail2ban/fail2ban#3098](https://github.com/fail2ban/fail2ban/issues/3098)
> 
> I will test and propose a new fail2ban package based on latest upstream release !

I do not know if the OpenWrt packaging proposal must follow upstream development or upstream release.
So I just take care of the reported issue and look to fix it only !

BKPepe:
I think you should take time to analyze it, prefer how you want to fix it, and share the final proposal with us instead of three comments in a row because then you are making it very difficult to read and understandable. You should also consider using edit.

erdoukki:
Resolved-by: https://github.com/openwrt/packages/pull/17341

Issue #15099, opened 2021-03-11
fail2ban: new package, missing lib2to3 in checks of PR from buildroot


erdoukki:
https://github.com/openwrt/packages/pull/15098

Checks failed because of missing library lib2to3 from buildroot images.

Is there any possibilities to add a host python pypi library at compile to have checks validating this PR ?

aparcar:
Isn't the correct way to add it as a package to this repo and then depend on it?

erdoukki:
Sorry, I don’t understand what you mean.
I have made a PR for the package of fail2ban which is not ok on additional build tests.
It is okay on my host, and on CI test.
Lib2to3 is necessary at build time only for these buildbot tests.
On my ubuntu buildroot the python package is already here.
Can someone help me to fix this ?

erdoukki:
```
PKG_BUILD_DEPENDS:=python3/host
HOST_BUILD_DEPENDS:=python3-lib2to3/host
```
will this be correct and sufficient ?
What the difference between PKG_B and UILD_DEPENDS and HOST_BUILD_DEPENDS ?

erdoukki:
looks like to fix the problem !
PR is no check/building correctly.

erdoukki:
```
cd /home/build/openwrt/build_dir/target-arc_archs_glibc/fail2ban-0.11.2 && ./fail2ban-2to3
Traceback (most recent call last):
  File "/home/build/openwrt/staging_dir/hostpkg/bin/2to3", line 3, in <module>
    from lib2to3.main import main
ModuleNotFoundError: No module named 'lib2to3'
Fail!
```

erdoukki:
> looks like to fix the problem !
> PR is no check/building correctly.

Not fixed, wrking without understanding why, today KO again : 
```
Traceback (most recent call last):
  File "/home/build/openwrt/staging_dir/hostpkg/bin/2to3", line 3, in <module>
    from lib2to3.main import main
ModuleNotFoundError: No module named 'lib2to3'
Fail!
```

erdoukki:
https://github.com/openwrt/packages/pull/15098/files/ffa8cbf11d3899e69cd32968857aa3aad216a64e#r593729377
After removing python3-lib2to3 dependency, the checks build do not work no more and made error happen again

erdoukki:
still not resolved with adding : HOST_BUILD_DEPENDS:=python3-lib2to3 !
I cannot add to PKG_BUILD_DEPENDS because of an error if I try to : 
[...]

PR #17316, opened 2021-12-12
fail2ban: fix 2to3 error

erdoukki:
Maintainer: me / @erdoukki 
Run tested: (untested)
Compile tested: (arch=MVEBU EspressoBin, host=Linux Mint x64)

Description:
2to3 support has been removed in setuptools since version 58.0.0.

Upstream issue/fix:
[BR]: build fails against setuptools 58+ fail2ban/fail2ban#3098
fail2ban/fail2ban@5ac303d

Fix openwrt/packages#17311

To install/build for python3 from source, it is necessary to convert to py3 codebase before setup
(invoke 2to3 or ./fail2ban-2to3 firstly).
> ./fail2ban-2to3
> python3 setup.py build


BKPepe:
Hmm… It makes me wonder.

Why you didn't use [the upstream commit](https://github.com/fail2ban/fail2ban/commit/21d94ff178192116eb5288190d78dd4630534a7b.patch), it was not working for you?

erdoukki:
> Why you didn't use [the upstream commit](https://github.com/fail2ban/fail2ban/commit/5ac303df8a171f748330d4c645ccbf1c2c7f3497.patch), it was not working for you?

Because it simply do the opposite we need.
This patch removes the needed flag we need to use with the latest release and the latest python setuptools.
The upstream master hasnone year bug fixes and enhancment

erdoukki:
New error now:
```
***   File "/usr/lib/python3.10/site-packages/fail2ban/server/ipdns.py", line 312
    					mask = ~(0xFFFFFFFFL >> cidr)
    					                  ^
SyntaxError: invalid hexadecimal literal
```
Reported here: https://github.com/fail2ban/fail2ban/issues/2959

erdoukki:
We need to execute ./fail2ban-2to3 in the configure step, before the install...
Is there any Python3 expert I could ask help for ?

erdoukki:
Locally resolved like : 
https://github.com/fail2ban/fail2ban/issues/2959#issuecomment-791681678
`apt install 2to3 python3-lib2to3 python3-toolz`

But I cannot add it in my Makefile nor in the buildroot...

It was already the problem in the first PR https://github.com/openwrt/packages/issues/15099

erdoukki:
> (invoke 2to3 or ./fail2ban-2to3 firstly).

Th problem is the necessary call of the ./fail2ban-2to3 command before the python build

erdoukki:
@BKPepe 
It may be compile correctly now.

Can you give a try, please ?

BKPepe:
I already told you that it is better to think twice before writing multiple comments in a row as you make it long, leading to any misunderstanding. Take your time. There is no need to rush. There is an edit button if you prefer. 

Well, you made more changes than it is described in the commit subject and its description. 

1. Why you change python3-light dependency into python3?
2. For the file `lib2to3.txt`, which you created, there is a folder for these files called `host-pip-requirements`.
Feel free more to check this [README](https://github.com/openwrt/packages/blob/master/lang/python/README.md). I'd prefer if you stick to the standard instead of reinventing the wheel.
3. Patch(es) should be properly numbered and have a clear name.

TBH, I don't like the changes you made, but I think there should be a better way to handle this. For now, I don't have much time to look at the issues you are dealing with. Hopefully, at the end of the week, it will be better.

erdoukki:
> Why you change python3-light dependency into python3?

Reffer to README python3, must be python3-light only for few or less than 3 lib !


> For the file `lib2to3.txt`, which you created, there is a folder for these files called `host-pip-requirements`.
> Feel free more to check this [README](https://github.com/openwrt/packages/blob/master/lang/python/README.md). I'd prefer if you stick to the standard instead of reinventing the wheel.

The sample codes use the same way...
The README never talk of requirement !

> Patch(es) should be properly numbered and have a clear name.

In facts it's true, and this patch may be also useless...

> TBH, I don't like the changes you made, but I think there should be a better way to handle this. For now, I don't have much time to look at the issues you are dealing with. Hopefully, at the end of the week, it will be better.

For me I do not like the tone you always take in our exchanges !
What's the problem with you ?
Did the TURRIS become a MASTER on OpenWrt and make ITS own COMPANY RULES been required !!!
I made this fix for be pleasant, and also without any help, on a tricky tweak...

> I already told you that it is better to think twice before writing multiple comments in a row as you make it long, leading to any misunderstanding. Take your time. There is no need to rush. There is an edit button if you prefer.

Sorry again to be so dumb !

erdoukki:
It will be a very little effort to enhance the work...
It was a huge effort to find a way to fix...

Perfection is not of this world, I think !?

PR #17330, opened 2021-12-12
Update SElinux utilities

flyn-org:
python3-libselinux: update to 3.3
	semodule-utils: update to 3.3
	python3-networkx: update to 2.6.3

Maintainer: me and @tpetazzoni 
Compile tested: x86_64

Description:


neheb:
for the FTS stuff, I wonder if it should be made dependent on musl. glibc does not need a separate library.

BKPepe:
Seems ok, merging. Thanks.

PR #17333, opened 2021-12-13
ddns-scripts: add ns1.com provider

ysc3839:
Signed-off-by: Richard Yu <yurichard3839@gmail.com>

Maintainer: @feckert ?
Compile tested: N/A
Run tested: x86_64

Description:
[NS1.com](https://ns1.com/)'s record is based on the domain and reocrd type, all IPs are in record's answers array.
When updating, new "answers" will override the old ones. The only way to keep old IPs is to read first, modify and update.
Since DDNS users rarely need multiple IPs, I didn't implement such function. Other settings such as TTL will not change during update.

Another question: Is `PKG_VERSION` really useful for this package? Looks like it doesn't change for a year. Maybe a date-based version code is useful.

### DIFF
--- a/net/ddns-scripts/files/usr/lib/ddns/update_ns1_com.sh
+++ b/net/ddns-scripts/files/usr/lib/ddns/update_ns1_com.sh
@@ -1,0 +1,77 @@
+#!/bin/sh
+# Derived from update_gandi_net.sh
+
+. /usr/share/libubox/jshn.sh
+
+local __ENDPOINT="https://api.nsone.net/v1"
+local __TTL=600
+local __RRTYPE
+local __URL
+local __STATUS
+
+[ -z "$username" ] && write_log 14 "Service section not configured correctly! Missing zone as 'username'"
+[ -z "$password" ] && write_log 14 "Service section not configured correctly! Missing API Key as 'password'"
+
+[ $use_ipv6 -ne 0 ] && __RRTYPE="AAAA" || __RRTYPE="A"
+
+# Construct JSON payload
+json_init
+# {"answers":[{"answer":["1.1.1.1"]}]}
+json_add_array answers
+json_add_object
+json_add_array answer
+json_add_string "" "$__IP"
+json_close_array
+json_close_object
+json_close_array
+
+__URL="$__ENDPOINT/zones/$username/$domain/$__RRTYPE"
+
+__STATUS=$(curl -s -X POST "$__URL" \
+	-H "X-NSONE-Key: $password" \
+	-H "Content-Type: application/json" \
+	-d "$(json_dump)" \
+	-w "%{http_code}\n" -o $DATFILE 2>$ERRFILE)
+
+if [ $? -ne 0 ]; then
+	write_log 14 "Curl failed: $(cat $ERRFILE)"
+	return 1
+elif [ -z $__STATUS ] || [ $__STATUS != 200 ]; then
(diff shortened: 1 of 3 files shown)


ysc3839:
Set TTL to 600 when creating DNS record, because the default is 3600. Will not change existing record's TTL.

net/ddns-scripts/files/usr/lib/ddns/update_ns1_com.sh
```diff
@@ -0,0 +1,74 @@
...
+
+add_answer() { # {"answers":[{"answer":["1.1.1.1"]}]}
+	json_add_array answers
+		json_add_object
+			json_add_array answer
+				json_add_string "" "$1"
+			json_close_array
+		json_close_object
+	json_close_array
+}
```
feckert, line 14: Please do not move in here, it will only confuse and if we need to change the object record every think gets changed.

net/ddns-scripts/files/usr/lib/ddns/update_ns1_com.sh
```diff
@@ -0,0 +1,74 @@
...
+local __STATUS
+
+[ -z "$username" ] && write_log 14 "Service section not configured correctly! Missing zone as 'username'"
+[ -z "$password" ] && write_log 14 "Service section not configured correctly! Missing API Key as 'password'"
+
+[ $use_ipv6 -ne 0 ] && __RRTYPE="AAAA" || __RRTYPE="A"
+
+# Construct JSON payload
+json_init
+add_answer "$__IP"
```
feckert, line 29: Could you please remove this function.
I see no point in writing a sub-function here for two calls.

ysc3839:
@feckert Updated, please review.

net/ddns-scripts/files/usr/lib/ddns/update_ns1_com.sh
```diff
@@ -0,0 +1,77 @@
...
+
+[ -z "$username" ] && write_log 14 "Service section not configured correctly! Missing zone as 'username'"
+[ -z "$password" ] && write_log 14 "Service section not configured correctly! Missing API Key as 'password'"
+
+[ $use_ipv6 -ne 0 ] && __RRTYPE="AAAA" || __RRTYPE="A"
+
+# Construct JSON payload
+json_init
+# {"answers":[{"answer":["1.1.1.1"]}]}
+json_add_array answers
```
feckert, line 20: I still have one small change. Can you please leave out the indentation. It only irritates when reading the source code.

net/ddns-scripts/files/usr/lib/ddns/update_ns1_com.sh
```diff
@@ -0,0 +1,77 @@
...
+	write_log 4 "Request failed: $__STATUS, NS1 answered: $(cat $DATFILE)"
+	if [ $__STATUS = 404 ]; then
+		write_log 4 "Status is 404, trying to create a DNS record"
+
+		json_init
+		json_add_string "zone" "$username"
+		json_add_string "domain" "$domain"
+		json_add_string "type" "$__RRTYPE"
+		json_add_string "ttl" "$__TTL"
+		json_add_array answers
```
feckert, line 49: And here

ysc3839:
@feckert Updated, please review.

feckert:
thanks merged :+1:

PR #17335, opened 2021-12-13
bandwidthd: migrate dependencies to php8 (fixes #17296)

mhei:
Maintainer: @padre-lacroix 
Compile tested: mxs
Run tested: -

Description:

This package was forgotten when all other packages were upgraded to PHP8.
So let's upgrade the dependencies here as well.

Thanks to @feckert, for pointing this out.

### DIFF
--- a/utils/bandwidthd/Makefile
+++ b/utils/bandwidthd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bandwidthd
 PKG_VERSION:=2.0.1-35
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/NethServer/bandwidthd/tar.gz/$(PKG_VERSION)?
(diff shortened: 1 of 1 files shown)


feckert:
Is the source functional with php8? I have never used the package before.

padre-lacroix:
Bandwidthd by itself does not run in PHP.  In the bandwidthd-pgsql or bandwidthd-sqlite, there are PHP files that goes on directory on a computer (or router) running PHP and when querying, using a browser, the directory, it quewries the database (postgresql or sqlite) to present on the browser the bandwidth data.

The PHP code is old, and I do not know if it works fine on PHP8.

I have not touched bandwidthd in a while and I will have to check the impact on the changes on the makefile.

Last, it seems that one of thec hecks did not pass (x86_64).

mhei:
That the checks did not pass  for x86_64 is a common problem, unrelated to this PR here. I had a look at the PHP source files and my personal judgement is that it should run also with PHP 8. At least, I cannot see any obvious problems - but I must admit, that I did not run the code - which would be the best approach.

utils/bandwidthd/Makefile
```diff
@@ -48,14 +48,14 @@ define Package/bandwidthd-sqlite
...
 
 define Package/bandwidthd-php
 $(call Package/bandwidthd/Default)
   TITLE = PHP files to graph bandwidthd data from postgresql
   VARIANT:=php
-  DEPENDS:=+libpcre +libxml2 +php7 +php7-cgi +php7-mod-pgsql +php7-mod-gd
+  DEPENDS:=+libpcre +libxml2 +php8 +php8-cgi +php8-mod-pgsql +php8-mod-gd
 endef
 
 define Package/bandwidthd/description
```
ptpt52, line 61: it is better do not use `@ someone` in commit message

ptpt52 (changes requested):

utils/bandwidthd/Makefile
mhei: Why?

padre-lacroix:
I will be checking this weekend whether the PHP files in bandwidthd-php are working with PHP8.
I will come back later to tell whether or not it is working.

padre-lacroix:
OK: it works fine!
But, there is a couple of things to set for php:
- In the /etc/php.ini file, the date.timezone = should be edited to a valid value (ex:date.timezone = “America/Montreal”): there is none when php8 is installed on openwrt.
- If you are using uhttpd, the following line should be added to /etc/config/uhttpd: list interpreter “.php=/usr/bin/php-cgi”

Both of these were explained in the Wiki for Bandwidthd for Openwrt (https://openwrt.org/docs/guide-user/services/network_monitoring/bandwidthd), with the exception about the timezone that was explained for php5, but was not required in php7 as one was already there!
I will edit the wiki to reflect that!

**Therefore, I approve the commit!**

mhei:
Thanks for caring about the wiki, really appreciated!

PR #17341, opened 2021-12-14
fail2ban: fix 2to3 error

erdoukki:
Maintainer: me / @erdoukki 
Run tested: (test done on MVEBU arch53 EspressoBin & EspressoBin-ULTRA, OpenWrt 21.02.x)
Compile tested: (arch=cortexa53, host=amd64)

Issue: 2to3 support has been removed in setuptools since version 58.0.0.

Fix: openwrt#17311

Requirements: 2to3/host openwrt/packages#17429

To install/build for python3 from source, it is necessary to convert to py3
codebase before setup (invoke 2to3 or ./fail2ban-2to3 firstly).
> ./fail2ban-2to3
> python3 setup.py build

### DIFF
--- a/net/fail2ban/Makefile
+++ b/net/fail2ban/Makefile
@@ -1,6 +1,6 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
 #
-# This is free software, licensed under the GNU General Public License v2.
-# See /LICENSE for more information.
+# Copyright (C) 2021-2022 Gerald Kerma
 #
 
 include $(TOPDIR)/rules.mk
@@ -19,6 +19,8 @@ PKG_LICENSE_FILES:=COPYING
 
 include $(INCLUDE_DIR)/package.mk
 include ../../lang/python/python3-package.mk
+
+HOST_BUILD_DEPENDS:=2to3/host
 
 define Package/fail2ban
   SECTION:=net
@@ -73,6 +75,11 @@ define Py3Package/fail2ban/install
 
 	$(INSTALL_DIR) $(1)/etc/uci-defaults
 	$(INSTALL_BIN) ./files/fail2ban.defaults $(1)/etc/uci-defaults/99_fail2ban
+endef
+
+define Py3Build/Compile
+	cd $(PKG_BUILD_DIR) && ./fail2ban-2to3
+	$(call Py3Build/Compile/Default)
 endef
 
 $(eval $(call Py3Package,fail2ban))
--- a/net/fail2ban/patches/020-fix-2to3.patch
+++ b/net/fail2ban/patches/020-fix-2to3.patch
@@ -1,0 +1,62 @@
+From 196c55e93103d1e56d1336f27a5f2591b4e54fef Mon Sep 17 00:00:00 2001
+From: sebres <info@sebres.de>
+Date: Sun, 19 Sep 2021 18:49:18 +0200
+Subject: [PATCH] fix gh-3098: build fails with error in fail2ban setup
+ command: use_2to3 is invalid (setuptools 58+)
+
+---
+ setup.py | 16 +---------------
+ 1 file changed, 1 insertion(+), 15 deletions(-)
+
+--- a/setup.py
++++ b/setup.py
+@@ -56,7 +56,7 @@ import warnings
+ from glob import glob
+ 
+ from fail2ban.setup import updatePyExec
+-
++from fail2ban.version import version
+ 
+ source_dir = os.path.realpath(os.path.dirname(
+ 	# __file__ seems to be overwritten sometimes on some python versions (e.g. bug of 2.6 by running under cProfile, etc.):
+@@ -120,22 +120,12 @@ class install_scripts_f2b(install_script
+ # Wrapper to specify fail2ban own options:
+ class install_command_f2b(install):
+ 	user_options = install.user_options + [
+-		('disable-2to3', None, 'Specify to deactivate 2to3, e.g. if the install runs from fail2ban test-cases.'),
+ 		('without-tests', None, 'without tests files installation'),
+ 	]
+ 	def initialize_options(self):
+-		self.disable_2to3 = None
+ 		self.without_tests = not with_tests
+ 		install.initialize_options(self)
+ 	def finalize_options(self):
+-		global _2to3
+-		## in the test cases 2to3 should be already done (fail2ban-2to3):
+-		if self.disable_2to3:
+-			_2to3 = False
+-		if _2to3:
+-			cmdclass = self.distribution.cmdclass
+-			cmdclass['build_py'] = build_py_2to3
+-			cmdclass['build_scripts'] = build_scripts_2to3
+ 		if self.without_tests:
+ 			self.distribution.scripts.remove('bin/fail2ban-testcases')
+ 
+@@ -186,7 +176,6 @@ commands.'''
+ if setuptools:
+ 	setup_extra = {
+ 		'test_suite': "fail2ban.tests.utils.gatherTests",
+-		'use_2to3': True,
+ 	}
+ else:
+ 	setup_extra = {}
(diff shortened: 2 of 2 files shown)


erdoukki:
@BKPepe 
Can you check, please ?

lang/python/host-pip-requirements/lib2to3.txt
```diff
@@ -0,0 +1,2 @@
+2to3==1.0 --hash=sha256:958bc212c928bbdcbc778b72528e0a39ae4ee8040eda6af6c3b5dd640c98ce6d
```
jefferyto, line 1: I would prefer to do something similar to Debian/Ubuntu (their `2to3` comes from a simple [script](https://salsa.debian.org/cpython-team/python3-defaults/-/blob/e429bbce303776b0357ef2636fd4b43ee0945c57/debian/2to3.py) that they maintain) rather than install a random `2to3` from PyPI.

I'll try to find time in the coming week to add a 2to3 host package, then this package can do `HOST_BUILD_DEPENDS:=2to3/host`.

jefferyto: > …why do you say « random » ?

* Project doesn't have a readme or a public code repository, only has one maintainer and one release from 2018
* Maintainer only has 2 projects total, the other one is labelled as deprecated

erdoukki: Great... Thanks !

erdoukki: > If needed, I can also take a look, test, or make the proposed PR, but I don’t know when.

I try to make the package for 2to3/host...
Will propose a PR if I get success on it !

erdoukki: Package made and compiled on master...
Waiting for your advice and review.
;-)

erdoukki: https://github.com/openwrt/packages/pull/17429

erdoukki:
May not compile correctly without the required éto3/host package from PR https://github.com/openwrt/packages/pull/17429

net/fail2ban/Makefile
```diff
@@ -20,6 +20,8 @@ PKG_LICENSE_FILES:=COPYING
 include $(INCLUDE_DIR)/package.mk
 include ../../lang/python/python3-package.mk
 
+HOST_BUILD_DEPENDS:=python3/host 2to3/host
```
jefferyto, line 23:
```suggestion
HOST_BUILD_DEPENDS:=2to3/host
```

Since `2to3/host` already depends on `python3/host`.

net/fail2ban/Makefile
```diff
@@ -75,6 +77,11 @@ define Py3Package/fail2ban/install
 	$(INSTALL_BIN) ./files/fail2ban.defaults $(1)/etc/uci-defaults/99_fail2ban
 endef
 
+define Py3Build/Compile
+	$(SHELL) -c "cd $(PKG_BUILD_DIR)/$(PYTHON3_PKG_SETUP_DIR) && ./fail2ban-2to3";
```
jefferyto, line 81:
```suggestion
	cd $(PKG_BUILD_DIR) && ./fail2ban-2to3
```

`$(PYTHON3_PKG_SETUP_DIR)` is only needed if your package had set `PYTHON3_PKG_SETUP_DIR` earlier.

net/fail2ban/Makefile
erdoukki: okay, agreed, thanks, modified

erdoukki: But still non testable with actual master...
Will have to switch to my local 2to3 branch, because of still missing 2to3 host package.
https://github.com/openwrt/packages/pull/17429 still pending !

net/fail2ban/Makefile
erdoukki: Okay.
Thanks for the `PYTHON3_PKG_SETUP_DIR` coorection (and for the explanation also ;-) ).
I think I have already tested something like this, with direct cd command instead of $SHELL call.
`	cd $(PKG_BUILD_DIR) && ./fail2ban-2to3`
I will test with your modification, and if it do not work, I revert with your modification like:
` 	$(SHELL) -c "cd $(PKG_BUILD_DIR) && ./fail2ban-2to3";`

net/fail2ban/Makefile
erdoukki: It looks like to work, my test gets successful.
The package was built.
Working also on TestBuild...
But if the 2to3 package is missing, it might not work !
I do not understand why it is built without the 2to3 package...
I get some same issue/success with the initial package, without get the point to understand...
May be a CACHE problem or any compilation side effect !?
The issue https://github.com/openwrt/packages/issues/17311 came because of the missing 2to3 !!!
Why it might or may not work, is non-understandable for now.

net/fail2ban/Makefile
erdoukki: okay with your proposal, another issue I get before, looks to be fixed now, but I still do not understand why !

jefferyto:
Not sure why https://github.com/openwrt/packages/actions/runs/1662159088 passed even though the 2to3 package wasn't merged into master yet - ping @aparcar 

@erdoukki please rebase on current master to pick up 2to3

erdoukki:
It now works with the add of the 196c55e93103d1e56d1336f27a5f2591b4e54fef commit from fail2ban upstream

erdoukki:
TEST IN PROGRESS,
There is an issue on the OpenWrt target:
```
root@STARGATE:~# fail2ban-client --version
  File "/usr/bin/fail2ban-client", line 34, in <module>
    from fail2ban.client.fail2banclient import exec_command_line, sys
ModuleNotFoundError: No module named 'fail2ban'
```
I am looking at to solve it. Will update the PR with a fix.

jefferyto:
I suggest:

1. Clone the packages repo into a new directory and check out the `openwrt-21.02` (or `openwrt-19.07`) branch
2. Copy the new package and dependencies (i.e. 2to3) into the new packages directory
3. Set up the openwrt-21.02 sdk/buildroot to use the new directory as the packages feed

net/fail2ban/Makefile
```diff
@@ -14,12 +14,14 @@ PKG_SOURCE_URL:=https://codeload.github.com/fail2ban/fail2ban/tar.gz/$(PKG_VERSI
 PKG_HASH:=383108e5f8644cefb288537950923b7520f642e7e114efb843f6e7ea9268b1e0
 
 PKG_MAINTAINER:=Gerald Kerma <gandalf@gk2.net>
-PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE:=GPL-2.0-only
```
jefferyto, line 17: Where did you see it is GPL2 only? According to the [readme](https://github.com/fail2ban/fail2ban/blob/095aeda8407b433098df35424cde2764a09566a6/README.md?plain=1#L120-L121) it is GPL2 or later.

erdoukki: fixed... thanks

erdoukki:
BUILD and TEST compatibility issue fixed with the correct version of Python.

```
root@ULTRA-5G:~# opkg install /root/fail2ban_0.11.2-4_aarch64_cortex-a53.ipk --force-reinstall
Removing package fail2ban from root...
Installing fail2ban (0.11.2-4) to root...
Configuring fail2ban.
uci: Entry not found
```

```
root@ULTRA-5G:~# fail2ban-client  --version
Fail2Ban v0.11.2
```
Tests done.
Upgrade fine.
Looks to work fine.

erdoukki:
@jefferyto test successful on 21.02.
I will PR the cherry-picked of the commit to the 21.02.

Do we need tests and PR for 19.07 ?

erdoukki:
@jefferyto 
All tests works ok
Build : ok
Install : ok
Upgrade : ok

jefferyto (approved):
Thanks!

PR #17344, opened 2021-12-15
qemu: add libslirp dependency

1715173329:
Maintainer: @yousong 
Compile tested: x86_64
Run tested: N/A

Description:
It has been integrated in QEMU to provide user mode networking.

### DIFF
--- a/utils/qemu/Makefile
+++ b/utils/qemu/Makefile
@@ -197,7 +197,7 @@ define qemu-target
 	+QEMU_UI_VNC_SASL:libsasl2 \
 	+QEMU_UI_SPICE:libspice-server \
 	+QEMU_DEV_USB:libusb-1.0 \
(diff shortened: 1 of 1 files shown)


utils/qemu/Makefile
```diff
@@ -197,7 +197,7 @@ define qemu-target
 	+QEMU_UI_VNC_SASL:libsasl2 \
 	+QEMU_UI_SPICE:libspice-server \
 	+QEMU_DEV_USB:libusb-1.0 \
-	$(if $(filter %-softmmu,$(1)),+libncurses +libfdt +pixman +qemu-firmware-efi $(ICONV_DEPENDS))
+	$(if $(filter %-softmmu,$(1)),+libncurses +libfdt +libslirp +pixman +qemu-firmware-efi $(ICONV_DEPENDS))
```
yousong, line 200: Can you also add an explicit configure option `--enable-slirp=system`

PR #17346, opened 2021-12-15
openvswitch: Tidy dependency definitions, fix libunbound dependency check

clayface:
Maintainer: @yousong
Compile tested: ipq806x
Run tested: N/A
cc: @stintel 

Description:
Tidy the dependency lists by merging the appended libatomic dependencies into the definitions, and changing appends to define where no previous definition exists. Finally fix the libunbound dependency check to use the standard method defined in the OpenWrt developer guide.


stintel (approved):
LGTM

PR #17358, opened 2021-12-16
jose: remove libjose

neheb:
It can be built statically as nothing else uses it.

Maintainer: @Tiboris 
Compile tested: ath79


Tiboris:
+1

PR #17389, opened 2021-12-18
openldap: hide configurables when libopenldap is not enabled

clayface:
Maintainer: @flyn-org 
Compile tested: bcm53xx
Run tested: N/A

Description: When openldap is not enabled, the following configurables remain present
in the config:

CONFIG_OPENLDAP_DEBUG=y
\# CONFIG_OPENLDAP_CRYPT is not set
\# CONFIG_OPENLDAP_MONITOR is not set
\# CONFIG_OPENLDAP_DB47 is not set
\# CONFIG_OPENLDAP_ICU is not set

This patch adds a libopenldap dependency for each option, removing the
above from .config when libopenldap is not used.

### DIFF
--- a/libs/openldap/Config.in
+++ b/libs/openldap/Config.in
@@ -1,0 +1,44 @@
+if PACKAGE_libopenldap
+	config OPENLDAP_DEBUG
+		bool "Enable debugging information"
+		default y
+		help
+		  Enable debugging information. This option must be enabled
+		  for the loglevel directive to work.
+	config OPENLDAP_CRYPT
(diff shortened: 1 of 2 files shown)


feckert:
While we're at it, should we put the whole thing in a separate Config.in file?
Now we intent additional one tab!
```
define Package/libopenldap/config
  source "$(SOURCE)/Config.in"
endef
```

clayface:
@feckert done. Are those Config.in indents okay?

feckert:
Thanks merged!

PR #17390, opened 2021-12-18
[21.02] python3-speedtest-cli: update to 2.1.3

jamesmacwhite:
Description:

Cherry pick 2.1.3 package bump from master due to a breaking change with the API.

python3-speedtest-cli version < 2.1.3 will not work currently and instead throw a Python error.

https://forum.openwrt.org/t/speedtest-cli-doesnt-work/114940/2
https://unix.stackexchange.com/questions/644442/speedtest-cli-valueerror-invalid-literal-for-int-with-base-10

### DIFF
--- a/lang/python/python3-speedtest-cli/Makefile
+++ b/lang/python/python3-speedtest-cli/Makefile
@@ -1,11 +1,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python3-speedtest-cli
-PKG_VERSION:=2.1.2
-PKG_RELEASE:=2
+PKG_VERSION:=2.1.3
+PKG_RELEASE:=1
 
 PYPI_NAME:=speedtest-cli
-PKG_HASH:=cf1d386222f94c324e3125ba9a0d187e46d4a13dca08c023bdb9a23096be2e54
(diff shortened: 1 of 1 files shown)


BKPepe:
Was it run and compile tested for OpenWrt 21.02?

cc maintainer: @jempatel

jamesmacwhite:
@BKPepe Thanks for taking a look. I haven't got a 21.02 environment to test on, but in theory the 2.1.2 to 2.1.3 change is purely down to the Ookla speedtest API and hence shouldn't have any major breaking change, however this will need confirming.

BKPepe:
Thus this can not be accepted as it is. I am sorry! 

We can not push this change to the stable branch unless it is tested on the real device, and because of that, I'm closing this pull request.

jamesmacwhite:
@BKPepe I understand, however the package version currently in the stable branch (2.1.2) is broken and 2.1.3 has the fix with related to the API response. Granted, I can't verify that, but if the package maintainer can or someone else hopefully that will allow you to accept it.

jamesmacwhite:
@BKPepe Fair point, it did seem like a minor version bump, but obviously there are further issues. My only hardware available is a WRT3200ACM which absolutely isn't going to be running 21.02, because of WiFI performance and general stability issues.

Maybe the original maintainer can take a look at some point.

PR #17394, opened 2021-12-19
ookla-speedtest package

jamesmacwhite:
Ookla Speedtest has an official CLI client which is statically linked and can run on OpenWrt, however it is only supported for certain architectures and the source code for it is also not published due to Ookla potentially giving up "trade secrets" or certain inner workings of their product.

I used the following Makefile as a starting point, with the objective of being able to have a installable package in OpenWrt repositories.

https://raw.githubusercontent.com/pyther/openwrt-feed/master/net/ookla-speedtest/Makefile
Credit to Matthew Gyurgyik @pyther who wrote the original Makefile, however it is hard coded to the x86_64 binary.

It would be good to allow the Makefile to be able to pull the correct binary relative to the architecture, but I'm thinking it could get quite messy. Ookla's binary naming may not align to OpenWrt directly and the PKG_HASH will need to be changed for each different binary, along with restricting the package itself to only supported architectures. 

If OpenWrt devs or maintainers could offer any assistance with making this more multi arch aware so other targets can benefit, I'd be happy to look into it, but for now I've taken the original Makefile from @pyther and just updated it to the standards outlined here: https://openwrt.org/docs/guide-developer/packages

Any advice or guidance on achieving the above is welcome!

### DIFF
--- a/net/ookla-speedtest/Makefile
+++ b/net/ookla-speedtest/Makefile
@@ -1,0 +1,42 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=ookla-speedtest
+PKG_VERSION:=1.1.1
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION)-linux-x86_64.tgz
+PKG_SOURCE_URL:=https://install.speedtest.net/app/cli/
+PKG_HASH:=970477fdfee4d741e75f088faf648f9b51c7be04313fb88ffb5de07eb1a3040c
+
+PKG_MAINTAINER:=James White <james@jmwhite.co.uk>
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/ookla-speedtest
+    SECTION=:net
+    CATEGORY:=Network
+    DEPENDS:=@TARGET_x86_64
+    TITLE:=Ookla Speedtest CLI
+    URL:=https://www.speedtest.net/apps/cli
+endef
+
+define Package/speedtest-cli/description
+    Speedtest CLI brings the trusted technology and global server network behind
+    Speedtest to the command line. Built for software developers, system
+    administrators and computer enthusiasts alike, Speedtest CLI is the first
+    official Linux-native Speedtest application backed by Ookla.
+endef
+
+define Build/Prepare
+    $(Build/Prepare/Default)
+endef
+
+define Build/Compile
+endef
+
+define Package/ookla-speedtest/install
+    $(INSTALL_DIR) $(1)/usr/bin
+    $(INSTALL_BIN) $(BUILD_DIR)/speedtest $(1)/usr/bin/speedtest
+endef
+
+$(eval $(call BuildPackage,ookla-speedtest))


net/ookla-speedtest/Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+define Package/speedtest-cli/description
+Speedtest CLI brings the trusted technology and global server network behind
+Speedtest to the command line. Built for software developers, system
+administrators and computer enthusiasts alike, Speedtest CLI is the first
+official Linux-native Speedtest application backed by Ookla.
+endef
+
+define Build/Prepare
+	$(Build/Prepare/Default)
```
BKPepe, line 30: Why?

net/ookla-speedtest/Makefile
```diff
@@ -0,0 +1,41 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=ookla-speedtest
+PKG_VERSION:=1.1.1
+PKG_RELEASE:=1
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION)-linux-x86_64.tgz
+PKG_SOURCE_URL:=https://install.speedtest.net/app/cli/
+PKG_HASH:=970477fdfee4d741e75f088faf648f9b51c7be04313fb88ffb5de07eb1a3040c
+
+PKG_MAINTAINER:=
```
BKPepe, line 10: I am not sure if we should add new packages w/o maintainer.

net/ookla-speedtest/Makefile
```diff
@@ -0,0 +1,41 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=ookla-speedtest
+PKG_VERSION:=1.1.1
+PKG_RELEASE:=1
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION)-linux-x86_64.tgz
+PKG_SOURCE_URL:=https://install.speedtest.net/app/cli/
+PKG_HASH:=970477fdfee4d741e75f088faf648f9b51c7be04313fb88ffb5de07eb1a3040c
```
BKPepe, line 8:
```suggestion

PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION)-linux-x86_64.tgz
PKG_SOURCE_URL:=https://install.speedtest.net/app/cli/
PKG_HASH:=970477fdfee4d741e75f088faf648f9b51c7be04313fb88ffb5de07eb1a3040c
```

net/ookla-speedtest/Makefile
```diff
@@ -0,0 +1,41 @@
...
+PKG_MAINTAINER:=
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/ookla-speedtest
+	SECTION=:net
+ 	CATEGORY:=Network
+	DEPENDS:=@TARGET_x86_64
+	TITLE:=Ookla Speedtest CLI
+  	URL:=https://www.speedtest.net/apps/cli
```
BKPepe, line 19:
```suggestion
  SECTION=:net
  CATEGORY:=Network
  DEPENDS:=@TARGET_x86_64
  TITLE:=Ookla Speedtest CLI
  URL:=https://www.speedtest.net/apps/cli
```

net/ookla-speedtest/Makefile
```diff
@@ -0,0 +1,41 @@
...
+endef
+
+define Package/speedtest-cli/description
+Speedtest CLI brings the trusted technology and global server network behind
+Speedtest to the command line. Built for software developers, system
+administrators and computer enthusiasts alike, Speedtest CLI is the first
+official Linux-native Speedtest application backed by Ookla.
```
BKPepe, line 26:
```suggestion
  Speedtest CLI brings the trusted technology and global server network behind
  Speedtest to the command line. Built for software developers, system
  administrators and computer enthusiasts alike, Speedtest CLI is the first
  official Linux-native Speedtest application backed by Ookla.
```

BKPepe (changes requested):
Formal things:
You said in the pull request that you honored the original author for the Makefile, but it is nowhere in the commit. You could use cherry-pick and mention your changes below `Signed-off-by` in square brackets.  There is also another way by using `Co-authored-by`.

You removed the template, so it is hard to know if it was at least somehow compile and run tested before sending it here.

Now the Makefile itself:
- missing PKG_LICENSE, PKG_LICENSE_FILES
- no one is stated as maintainer (not good for newly added packages)
- only for x86_64 as said

If possible, you can add `test.sh` to have it run tested inside CI.

What bothers me is that there is no source code available.

jamesmacwhite:
@BKPepe Thanks for reviewing. 

- I've updated the commit to reference the original Makefile after the sign off.
- I've added myself as the maintainer
- Looking through the documentation, I cannot find any direct license for Ookla Speedtest CLI. I can only find licenses for related software used in the software itself.
- It technically isn't just x86_64 only, but in order to have the Makefile work for different targets, it's probably going to get really messy to achieve under the Makefile, unless there is a better way? Although, there is the consideration that using Ookla's speedtest CLI on less powerful hardware, you are likely to get invalid results. x86_64 at least makes this slightly less of an issue.
- I understand the concerns about the source code, there is an official comment on this from an Ookla employee: https://www.reddit.com/r/HomeNetworking/comments/dpalqu/speedtestnet_just_launched_an_official_c_cli/f5tmibx/
-

net/ookla-speedtest/Makefile
jamesmacwhite: I've added myself as the maintainer.

pyther:
I don't believe Ookla Speedtest should be packaged with OpenWrt.

- After the initial release (10/29/2019), no updates were released for two years
- The original release of version 1.0.0 linked against glibc. At some point in time, they released new binaries that statically linked against glibc, without updating the version number. This can be seen in the commit history for my speedtest docker container. https://github.com/pyther/speedtest/commits/master
- Source code is not available and binaries do not exist for all openwrt platforms

In addition, the license does not allow for redistribution based on my understanding. When running speedtest for the first time you must agree to the following.
```
[mg5@bug ~]$ ./speedtest
==============================================================================

You may only use this Speedtest software and information generated
from it for personal, non-commercial use, through a command line
interface on a personal computer. Your use of this software is subject
to the End User License Agreement, Terms of Use and Privacy Policy at
these URLs:

	https://www.speedtest.net/about/eula
	https://www.speedtest.net/about/terms
	https://www.speedtest.net/about/privacy

==============================================================================

Do you accept the license? [type YES to accept]:
``` 

On the EULA page in section 2: Restriction of use, it states
```
You shall not:
...
(e) rent, lease, lend, sell, sublicense, assign, distribute, publish, transfer, or otherwise make available the Software, or any features or functionality of the Software, to any third party for any reason, including by making the Software available on a network where it is capable of being accessed by more than one device at any time;
```

BKPepe:
So, that's it. Closing.

PR #17401, opened 2021-12-20
xray-core: Update to 1.5.1

1715173329:
Maintainer: me
Compile tested: ipq807x, rockchip
Run tested: rk3328 nanopi-r2s

Description:
Release note: https://github.com/XTLS/Xray-core/releases/tag/v1.5.1

This version contained some security fixes.
All patches were taken from upstream commits.


net/xray-core/Makefile
```diff
@@ -1,18 +1,17 @@
 include $(TOPDIR)/rules.mk
 
-PKG_NAME:=xray-core
-PKG_VERSION:=1.4.5
+PKG_NAME:=Xray-core
```
PolynomialDivision, line 3: why did you do a rename? Never saw that people use capital chars in the pkg-name.

PolynomialDivision:
My general opinion about this package would be to rename the pkg-name and directory to xray.

net/xray-core/Makefile
1715173329: It doesn't affect the real pkg-name:
- xray-core
- xray-example
- xray-geodata

In order to reuse default PKG_BUILD_DIR, I made this change.
[openblas](https://github.com/openwrt/packages/blob/master/libs/openblas/Makefile) also did the same thing.

BKPepe: Even you're right and you found the exception, the pkg_name should match the naming of the folder. So, I'd suggest to drop this change.

1715173329: Updated, thanks.

Issue #16698, opened 2021-09-24
adguardhome: How to install and run?


pkejval:
Installed package ` adguardhome (0.106.3-1)` from Luci and what next? No new option added in Luci. So I tried SSH and see what is going on:

`root@Router:~# /etc/init.d/adguardhome status` returns `running`
but I can't see any process running in top nor ps. There is no config file `/etc/adguardhome.yaml` nor in `/etc/config`. And there is no app listening at http://routerip:3000

Seems like package installation isn't working or am I doing something wrong?

My OpenWRT is running in KVM `OpenWrt 21.02.0 r16279-5cc0535800 / LuCI openwrt-21.02 branch git-21.231.26241-422c175`

pkejval:
I was playing around with it and I found that:

```
root@Router:~# /usr/bin/AdGuardHome
panic: qtls.CertificateRequestInfo doesn't match

goroutine 1 [running]:
github.com/marten-seemann/qtls-go1-16.init.0()
        github.com/marten-seemann/qtls-go1-16@v0.1.3/unsafe.go:17 +0x198
```

BUT then I downloaded binary from AdguardHome Github and replaced `/usr/bin/AdGuardHome` with downloaded binary version `v0.107.0-b.11`, created config `/etc/adguardhome.yaml` and now AdGuardHome is running!

Issues with package are:

- no config file is saved in `/etc/adguardhome.yaml` after installing package
- `/usr/bin/AdGuardHome` can't be run (error above) and `/etc/init.d/adguardhome status` shows running even service isn't running

alaviss:
From a cursory web search this seems to be the issue with a dependency of AdGuard Home having hard dependency on a given version of Go. See this similar issue: https://github.com/caddyserver/caddy/issues/4118

The release version of AdGuard Home is only compatible with Go 1.16 from the error message, and OpenWrt uses Go 1.17, hence why it doesn't work.

Issue #16594, opened 2021-09-09
adguardhome: panic: qtls.ClientHelloInfo doesn't match


foxcris:
Maintainer: @dobo90
Environment: 
Openwrt Version: OpenWrt 21.02.0, r16279-5cc0535800
Hardware: fritzbox 4040

Description:
adguardhome is listed as started but no adguardhome proess is running

Executing the AdGuardHome binary manually gives the following error
```
root@owrt1:/etc/openvpn# /usr/bin/AdGuardHome --help
panic: qtls.ClientHelloInfo doesn't match

goroutine 1 [running]:
github.com/marten-seemann/qtls-go1-15.init.0()
        github.com/marten-seemann/qtls-go1-15@v0.1.1/unsafe.go:20 +0x168
```

derterz:
Same here after upgrading from an old nightly 21.02 build to 21.02.0 r16279-5cc0535800 on my Linksys MR8300 (Dallas) [ARMv7 Processor rev 5 (v7l)].

Manually downloading and running the latest version works fine: https://github.com/AdguardTeam/AdGuardHome/releases/download/v0.106.3/AdGuardHome_linux_armv7.tar.gz

BKPepe:
I cherry-picked all the updates related to adguardhome from the ``master`` branch into ``openwrt-21.02``. Hopefully, it will solve your issues. If yes, then please close it. If not, feel free to discuss it with the maintainer.

predators46:
hmm. i tried on lede 17.01.x it runs fine

![Screenshot_2021-09-12_162401](https://user-images.githubusercontent.com/31931530/133251862-e547e1d9-e954-41c6-b604-935684cf8532.jpg)

BKPepe:
LEDE 17.01.x is not supported. Supported versions are OpenWrt 19.07 and OpenWrt 21.02.

predators46:
> LEDE 17.01.x tidak didukung. Versi yang didukung adalah OpenWrt 19.07 dan OpenWrt 21.02.



> LEDE 17.01.x is not supported. Supported versions are OpenWrt 19.07 and OpenWrt 21.02.

Yes, I know

HinataKato:
Same, I was able to use it at 21.02.0-rc4, after upgrading to 21.02.0 it doesn't work

foxcris:
@BKPepe how am i able to perform a test? Till now there is no updated package available via opkg. Do just have to wait a bit?

foxcris:
[...]

PR #17409, opened 2021-12-21
adguardhome: update to 0.107.0

alaviss:
Maintainer: @dobo90
Compile tested: arm64, Linksys E8450, OpenWrt SNAPSHOT r18371-5a4685cfa2
Run tested: arm64, Linksys E8450, OpenWrt SNAPSHOT r18371-5a4685cfa2, runs and functions normally.

Description:

Full changelog available at: https://github.com/AdguardTeam/AdGuardHome/releases/tag/v0.107.0

Also added the ability to configure working directory location.

This directory is used by AdGuard to store persistent data like query
logs, filter lists, etc.

Data stored in this directory can get really huge, as such allowing
this directory to be moved elsewhere (ie. an USB drive) is very
beneficial.

Fixes https://github.com/openwrt/packages/issues/16698
Fixes https://github.com/openwrt/packages/issues/16594

### DIFF
--- a/net/adguardhome/files/adguardhome.config
+++ b/net/adguardhome/files/adguardhome.config
@@ -1,0 +1,3 @@
+config adguardhome config
+	# Where to store persistent data by AdGuard Home
+	option workdir /var/adguardhome
--- a/net/adguardhome/Makefile
+++ b/net/adguardhome/Makefile
@@ -6,19 +6,19 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=adguardhome
-PKG_VERSION:=0.106.3
+PKG_VERSION:=0.107.0
 PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_VERSION:=v$(PKG_VERSION)
 PKG_SOURCE_URL:=https://github.com/AdguardTeam/AdGuardHome
-PKG_MIRROR_HASH:=bdf5ad833df90969cd82b9fb6a91846338c738ed01671fc768d867b592102f7c
+PKG_MIRROR_HASH:=7bb70a3cdbde0d4c451454369317e6265832bcf6743c44fae7df7e3f8a37a108
 
 PKG_LICENSE:=GPL-3.0-only
 PKG_LICENSE_FILES:=LICENSE.txt
 PKG_MAINTAINER:=Dobroslaw Kijowski <dobo90@gmail.com>
 
-PKG_BUILD_DEPENDS:=golang/host node/host node-yarn/host packr/host
+PKG_BUILD_DEPENDS:=golang/host node/host node-yarn/host
 PKG_BUILD_PARALLEL:=1
 PKG_USE_MIPS16:=0
 
@@ -46,6 +46,7 @@ endef
 
 define Package/adguardhome/conffiles
 /etc/adguardhome.yaml
+/etc/config/adguardhome
 endef
 
 define Package/adguardhome/description
@@ -56,7 +57,6 @@ define Build/Compile
 	( \
 		pushd $(PKG_BUILD_DIR) ; \
 		make js-deps js-build ; \
-		packr -z -v -i internal ; \
 		popd ; \
 		$(call GoPackage/Build/Compile) ; \
 	)
@@ -66,6 +66,9 @@ define Package/adguardhome/install
 	$(call GoPackage/Package/Install/Bin,$(1))
 	$(INSTALL_DIR) $(1)/etc/init.d
 	$(INSTALL_BIN) ./files/adguardhome.init $(1)/etc/init.d/adguardhome
+
+	$(INSTALL_DIR) $(1)/etc/config
+	$(INSTALL_DATA) ./files/adguardhome.config $(1)/etc/config/adguardhome
 endef
 
 $(eval $(call GoBinPackage,adguardhome))
--- a/net/adguardhome/files/adguardhome.init
+++ b/net/adguardhome/files/adguardhome.init
@@ -1,7 +1,6 @@
 #!/bin/sh /etc/rc.common
 
 PROG=/usr/bin/AdGuardHome
-WORK_DIR=/tmp/adguardhome
 
 USE_PROCD=1
 
@@ -11,6 +10,9 @@ START=21
 STOP=89
 
 start_service() {
+  config_load adguardhome
+  config_get WORK_DIR config workdir
+
   [ -d "$WORK_DIR" ] || mkdir -m 0755 -p "$WORK_DIR"
 
   procd_open_instance


dobo90 (changes requested):
Thanks for this pull request.

packr has been removed in
https://github.com/AdguardTeam/AdGuardHome/commit/c6888326b034daea9e69de5ee0a8f37aede45892#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L2

So please remove:
* "packr/host" from PKG_BUILD_DEPENDS
* "packr -z -v -i internal" from Build/Compile

I've compiled this with packr removed and it seems to be working on NanoPi R2S.

My diff:
```
commit c45eb29cf7433275a3b384c711289dc339718761
Author: Dobroslaw Kijowski <dobo90@gmail.com>
Date:   Tue Dec 21 19:38:32 2021 +0100

    adguardhome: drop packr

diff --git a/net/adguardhome/Makefile b/net/adguardhome/Makefile
index f58640818..b6bbf67d1 100644
--- a/net/adguardhome/Makefile
+++ b/net/adguardhome/Makefile
@@ -18,7 +18,7 @@ PKG_LICENSE:=GPL-3.0-only
 PKG_LICENSE_FILES:=LICENSE.txt
 PKG_MAINTAINER:=Dobroslaw Kijowski <dobo90@gmail.com>
 
-PKG_BUILD_DEPENDS:=golang/host node/host node-yarn/host packr/host
+PKG_BUILD_DEPENDS:=golang/host node/host node-yarn/host
 PKG_BUILD_PARALLEL:=1
 PKG_USE_MIPS16:=0
 
@@ -57,7 +57,6 @@ define Build/Compile
        ( \
                pushd $(PKG_BUILD_DIR) ; \
                make js-deps js-build ; \
-               packr -z -v -i internal ; \
                popd ; \
                $(call GoPackage/Build/Compile) ; \
        )
```

alaviss:
Confirmed working here as well @dobo90. I pushed a new commit with your patch.

BKPepe:
I have one nitpick. Can you add to the commit description that there was removed packr? Otherwise, this is very good contribution. Thank you for that. You even added `Co-authored-by`. 👍 More contributions like this.

net/adguardhome/files/adguardhome.config
```diff
@@ -0,0 +1,3 @@
+config adguardhome config
+	# Where to store persistent data by AdGuard Home
+	option workdir /tmp/adguardhome
```
jefferyto, line 3:
```suggestion
	option workdir /var/adguardhome
```

`/var` is usually symlinked to `/tmp`, and if the user decides to move `/var` to persistent storage then this directory would also become persistent.

mercyground:
The adguard team use /opt/AdGuardHome for their install folder. It would make things easier if you use same location so people using the AGH wiki get to use same locations? Also that also makes it easier to redirect that folder to a usb key or an exroot partition.

alaviss:
I don't think it is a good idea to keep the working directory in the flash by default. It experiences a lot of writes due to the aggressive default logging and filter updating policy and can easily fill up root and/or wear down the flash used by routers.

Knowledgeable users that setup their own extroot should be able to modify the UCI configuration to move the folder to their external devices.

Also, `/opt` is generally meant for software that cannot be fitted into the file system hierarchy, which is not the case here.

> It would make things easier if you use same location so people using the AGH wiki get to use same locations

For this I think we should just update the AGH wiki to explain about the package provided by OpenWRT, or we can provide a readme in this repo and link it there.

mercyground:
Agreed on the flash issues. 

> For this I think we should just update the AGH wiki to explain about the package provided by OpenWRT, or we can provide a readme in this repo and link it there.

Maybe just a wiki entry on OpenWrt's wiki explaining the differences?

Due to when i originally set out to use AGH, the opkg version was 104 and had serious issues with it which i detailed in this thread. https://forum.openwrt.org/t/howto-running-adguard-home-on-openwrt/51678/224?u=mercygroundabyss

More recently i wrote up how to install and use the 107 edge build as a clean install using AGH's script with tweaks done to swap Dnsmasq to just local ptr requests and use AGH for primary DNS. Mostly because i do not have much experience with doing PRs or fixing how packages are packaged for OpenWrt. That post is here. https://forum.openwrt.org/t/how-to-updated-2021-installing-adguardhome-on-openwrt/113904

Potentially some of the work done there could be used to improve the opkg version? I've detailed my choices and decisions and some of the background for those choices in that post.

mercyground:
A few notes from my explorations:

AGH still requires some settings to be manually done via their yaml file. There is requests to be able to configure those initial setup parameters on the setup page on port 3000 or be able to configure them better. At present you either have to select which interface to use singularly or use ALL interfaces. The ability to select and manage that would make configuring easier (by selecting ALL it also includes the WAN interface which is not ideal) I believe this is due in 108.

Another issue is AGH's webinterface clashing with Luci. In my install i just installed it on port 8080 as a secondary webinterface. This also has issues with SSL. You have to setup and run SSL AGH interface on a different port.

There did used to be a luci adguardhome plugin but this appears to be unmaintained and very out of date.

rebind_protection on dnsmasq needs disabling if you switch AGH to be primary DNS.

DHCP option 6: which DNS (Domain Name Server) to include in the IP configuration for name resolution. Used to redirect clients to use AGH DNS. (I set it up this way in case others wanted to run AGH separately on a different port)

What sort of default config do we offer to opkg installed AGH users? or do we leave it for them to configure from scratch?

BKPepe:
CI failure is related to cryptodev, which is not related to somehow, so merging.

alaviss:
@mercyground 

Your forum post is great, though there is only this tiny correction:

> The primary being that it installs to /tmp/AdGuardHome. This means every time you reboot your router you will have to re-setup AGH.

This is false, `/tmp/adguardhome` is used as the persistent data directory, but the configuration is stored in `/etc/adguardhome.yml`, making it persistent between reboots/upgrades. So you will only lose logs, stats and filter lists (which will be automatically re-downloaded) on reboot.

Your dnsmasq configuration script is a bit overkill, though. Turning off `peerdns` and turning on `localuse` should be enough to force DNS to be done solely on the router. I will comment on some potential improvements in the forum when I find some time.

The forum post is otherwise excellent, though.

> What sort of default config do we offer to opkg installed AGH users? or do we leave it for them to configure from scratch?

Changing users settings on installation is error prone and is a side-effect that most won't expect (especially since you can't revert them on uninstall). I think it might be better to either provide a wiki entry guiding them through the installation/usage like the ones for DoH and DoT: https://openwrt.org/docs/guide-user/services/dns/dot_dnsmasq_stubby. I think your forum post is an excellent starting point for that.

I can't commit on writing wiki entries for this though, due to a lack of time.

> Hope some of my explorations and details will make the opkg version easier to install for OpenWrt users instead of having to install via the install script and do manual changes.

For now we can only avoid the tools you need to run AGH installation script and the update script with the opkg package, unfortunately.

> Another issue is AGH's webinterface clashing with Luci. In my install i just installed it on port 8080 as a secondary webinterface. This also has issues with SSL. You have to setup and run SSL AGH interface on a different port.

FWIW, I solved this problem on my system by running LuCI via nginx and setup `/adguard` as a reverse-proxy to AGH, though I couldn't get DoH interface to work this way.

mercyground:
> This is false, `/tmp/adguardhome` is used as the persistent data directory, but the configuration is stored in `/etc/adguardhome.yml`, making it persistent between reboots/upgrades. So you will only lose logs, stats and filter lists (which will be automatically re-downloaded) on reboot.

This is referring to the /much/ older 104 build. I hadn't tried the opkg version since then. I should update and make that clear.

> Your dnsmasq configuration script is a bit overkill, though. Turning off `peerdns` and turning on `localuse` should be enough to force DNS to be done solely on the router. I will comment on some potential improvements in the forum when I find some time.

I deliberately set out to make the OpenWrt DNS more bullet proof. By repointing the router DNS to Cloudflare and ignoring ISP DNS then we avoid ISP dns failures and allow the router to do NTP updates and opkg updates while AGH loads. Clients are pointed to AGH DNS so filtering is effective. The router is excluded from that because we should not need to filter the router itself.

> The forum post is otherwise excellent, though.

Thanks. It was a combined effort however. I took the original 2yr long thread from https://forum.openwrt.org/t/howto-running-adguard-home-on-openwrt/51678 and summerised it and redid it with explainations to ensure that people could follow my directions and reasoning behind it. Blindly following instructions from randomers on the internet should be considered... unwise (rm / -r? ) :)

> Changing users settings on installation is error prone and is a side-effect that most won't expect (especially since you can't revert them on uninstall). I think it might be better to either provide a wiki entry guiding them through the installation/usage like the ones for DoH and DoT: https://openwrt.org/docs/guide-user/services/dns/dot_dnsmasq_stubby. I think your forum post is an excellent starting point for that.
> I can't commit on writing wiki entries for this though, due to a lack of time.

Agreed. I do have wiki edit so i could potentially re-write my instructions into a wiki post by using a backup router and trialing the opkg version.

> For now we can only avoid the tools you need to run AGH installation script and the update script with the opkg package, unfortunately.

Making things simpler for users should hopefully reduce questions.

> FWIW, I solved this problem on my system by running LuCI via nginx and setup `/adguard` as a reverse-proxy to AGH, though I couldn't get DoH interface to work this way.

maybe using Traefik instead? but i feel that's adding potentially more complex issues. I don't see a reason personally to do more than just use AGH on port 8080 to keep things simple.

alaviss:
> By repointing the router DNS to Cloudflare and ignoring ISP DNS then we avoid ISP dns failures and allow the router to do NTP updates and opkg updates while AGH loads.

I think the only part we need to be aware of is NTP updates since correct time is required for upstream DoT/DoH servers, of which you can `[/pool.ntp.org/]1.1.1.1` to bypass it by using insecure DNS. Granted, I haven't rebooted my router to test this theory :P

> The router is excluded from that because we should not need to filter the router itself.

Do note that LuCI uses reverse DNS to resolve IP into names in wireless overview, so not using router's DNS might break it.

I haven't fully verified this, only verified that it does use rDNS since my AGH reported an insane amount of rDNS requests coming from `127.0.0.1` whenever I keep LuCI open.

And you might want to use AGH for the router just so you benefit from the same DNS cache that is shared with your other clients.

> maybe using Traefik instead? but i feel that's adding potentially more complex issues. I don't see a reason personally to do more than just use AGH on port 8080 to keep things simple.

I looked into it but settled on nginx for two reasons:

- `luci-nginx-ssl` is officially supported and basically install-and-run.
- AGH wiki has [sample configuration](https://github.com/AdguardTeam/AdGuardHome/wiki/FAQ#nginx) for nginx reverse proxy.

mercyground:
> I think the only part we need to be aware of is NTP updates since correct time is required for upstream DoT/DoH servers, of which you can `[/pool.ntp.org/]1.1.1.1` to bypass it by using insecure DNS. Granted, I haven't rebooted my router to test this theory :P
> Do note that LuCI uses reverse DNS to resolve IP into names in wireless overview, so not using router's DNS might break it.
> I haven't fully verified this, only verified that it does use rDNS since my AGH reported an insane amount of rDNS requests coming from `127.0.0.1` whenever I keep LuCI open.
> And you might want to use AGH for the router just so you benefit from the same DNS cache that is shared with your other clients.

The way I approached it was to ensure OpenWrt's dnsmasq remains active but on port 5353 instead. I just point its upstream at Cloudflare. This means the routers DNS is insecure but avoids issues with ntp and luci.

To solve your issues with rdns lookups? you need to enable rDNS lookups on AGH and also a bit of interception (this is needed for multiple networks such as VLANS) This means AGH is primary DNS but dnsmasq is downstream for ptrs. Potentially in future AGH may just take over DHCP too but it is far inferior right now against OpenWrt's DHCP setup.

[https://forum.openwrt.org/uploads/default/original/3X/d/0/d0da99936ac7351015bef47ea7166be54d2009a3.png](url)

The yaml settings for this:

```
  resolve_clients: true
  use_private_ptr_resolvers: true
  local_ptr_upstreams:
  - 127.0.0.1:5353
```
Domain interception :

If you have the default OpenWrt DHCP then adding :

[/lan/]127.0.0.1:5353
[//]127.0.0.1:5353

to the upstream list that will intercept lan (internal domain) and domainless requests and pass those requests back to openwrt. "lan" is OpenWrt's default domain. (You may need to put your ip address instead of just local host.) This is necessary for multi ip networks. For simple IP networks rDNS should cover this.

> I looked into it but settled on nginx for two reasons:
> 
> * `luci-nginx-ssl` is officially supported and basically install-and-run.
> * AGH wiki has [sample configuration](https://github.com/AdguardTeam/AdGuardHome/wiki/FAQ#nginx) for nginx reverse proxy.

That is a simpler option and less work to document. Rubberstamp that idea!

mercyground:
https://forum.openwrt.org/t/request-to-create-page-for-adguard-home-under-dns-section/115582

standby for documentation /placeholder/ 👍

mercyground:
https://openwrt.org/docs/guide-user/services/dns/adguard-home

thanks to jamesmacwhite for his work with that.

PR #17415, opened 2021-12-21
Add exporter for dnsmasq dhcp leases

simonswine:
This adds a lua prometheus node exporter, which can expose the DHCP lease expiry time.

### DIFF
--- a/utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/dnsmasq_dhcp_leases.lua
+++ b/utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/dnsmasq_dhcp_leases.lua
@@ -1,0 +1,41 @@
+local ubus = require "ubus"
+local u = ubus.connect()
+
+local lease_file = "/tmp/dhcp.leases"
+
+-- get the configured lease file from uci
+local lease_file_cfg = u:call("uci", "get", {config="dhcp", section="@dnsmasq[0]", option="leasefile"})
+if lease_file_cfg ~= nil then
+    lease_file = lease_file_cfg.value
+end
+
+local function scrape()
+    local metric_dhcp = metric("dnsmasq_dhcp_lease_expiry_time_seconds", "gauge")
+    f = io.input(lease_file)
+    for line in f:lines() do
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
Pease preifx commit message with `prometheus-node-exporter-lua`, so `prometheus-node-exporter-lua: add exporter for dnsmasq dhcp leases`

utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/dnsmasq_dhcp_leases.lua
```diff
@@ -0,0 +1,30 @@
+local function scrape()
+    local metric_dhcp = metric("dnsmasq_dhcp_lease_expiry_time_seconds", "gauge")
+    f = io.input("/tmp/dhcp.leases")
```
PolynomialDivision, line 3: Maybe you should not assume that `option leasefile '/tmp/dhcp.leases'` is set in `/etc/config/dhcp`. Maybe just read the config option?

simonswine: Good point should be fixed now :+1:

utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/dnsmasq_dhcp_leases.lua
```diff
@@ -0,0 +1,41 @@
+local ubus = require "ubus"
+local u = ubus.connect()
+
+local lease_file = "/tmp/dhcp.leases"
+
+-- get the configured lease file from uci
+local lease_file_cfg = u:call("uci", "get", {config="dhcp", section="@dnsmasq[0]", option="leasefile"})
```
champtar, line 7: I think we can have multiple dnsmasq instances thus multiple lease files

champtar (changes requested):

feckert:
Nothing has happened here for a long time. Is it still relevant? If so, please rebase and open this PR again

PR #17416, opened 2021-12-21
Ensure mac address is lower cased

simonswine:
Most other exporter would use a lower-cased mac hex format for the `mac=` label. This changes allows to match them up using PromQL.

### DIFF
--- a/utils/prometheus-node-exporter-lua/Makefile
+++ b/utils/prometheus-node-exporter-lua/Makefile
@@ -4,7 +4,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=prometheus-node-exporter-lua
-PKG_VERSION:=2021.12.21
+PKG_VERSION:=2022.01.02
 PKG_RELEASE:=1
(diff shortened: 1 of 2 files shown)


BKPepe:
This requires to bump ``PKG_RELEASE`` in Makefile.

simonswine:
My feeling is that the test failure is unrelated:

> `ERROR: package/feeds/base/cryptodev-linux failed to build.`

champtar:
Looking at other exporter in this package, dawn exporter (@PolynomialDivision), uci_dhcp_host (@Tycale), hostapd_stations (@mweinelt / @blocktrron) might also have upper cases mac 
Are you all fine with forcing lower case mac addresses ?

mweinelt:
Making them the same case is important for match across metrics, so I'm all for that. I prefer lowercase just on the basis of being simpler to type.

feckert:
Nothing has happened here for a long time. Is it still relevant? If so, please rebase and open this PR again

PR #17431, opened 2021-12-25
adguardhome: add support for dropping privileges

alaviss:
Maintainer: @dobo90
Compile tested: arm64, Linksys E8450, OpenWrt SNAPSHOT r18371-5a4685cfa2
Run tested: arm64, Linksys E8450, OpenWrt SNAPSHOT r18371-5a4685cfa2, runs and functions normally.

Description:

AdGuard Home can be run as a normal user as long as it has the permission to bind to ports < 1024 and access rights to its files.

Configuration folder has to be moved from `/etc` to `/etc/adguardhome`, since AGH web UI creates temporary configs to atomically replace the main config via renames, which it could not do in `/etc` since only root can create files there.

This is draft for a couple reasons:
- Jail is very restrictive by default and might break configurations that use the HTTPS feature since the certificates are not mounted by default (and it is not feasible to do unless we parse AGH yaml, which I don't know how to). This could be remedied by doing a RO `/` bind mount.
- There are no UID allocated for the AGH user.
- The migration script is not very well tested yet.
- I don't know of any way to notify the user that they have to perform migrations if they want to use `ujail`.
- I'm not sure if the jail should be turned on by default.
- I have not tested the DHCP function.

While I have been running this for a week, I have not fully tested


net/adguardhome/files/adguardhome.init
```diff
@@ -9,15 +11,51 @@ START=21
...
+    chown adguardhome:adguardhome "$WORK_DIR"
+  fi
 
   procd_open_instance
-  procd_set_param command "$PROG" -c /etc/adguardhome.yaml -w "$WORK_DIR" --no-check-update
+  procd_set_param command "$PROG" -c "$CONFIG_FILE" -w "$WORK_DIR" --no-check-update
   procd_set_param stdout 1
   procd_set_param stderr 1
+
+  if [ "$jail" -ne 0 ]; then
```
wulfy23, line 44: this appears prone to the 'umdns ujail not installed bug'...

until a usable library exists, you are going to have to check for jail capability and consider what to do if it's not available...

wulfy23 (commented):
please test without ujail

jamesmacwhite:
@alaviss As this PR is looking at configuration paths and such, should the YAML file be named "AdGuardHome.yaml" not "adguardhome.yaml"?

Looking at the main AdGuard Home wiki it suggests, this is what the default is outside of the OpenWrt package.

https://github.com/AdguardTeam/AdGuardHome/wiki/Configuration#configuration-file

Should OpenWrt match upstream as close as possible?

alaviss:
> As this PR is looking at configuration paths and such, should the YAML file be named "AdGuardHome.yaml" not "adguardhome.yaml"?
> 
> Looking at the main AdGuard Home wiki it suggests, this is what the default is outside of the OpenWrt package.

Sure, renamed config to `AdGuardHome.yml`.

> * I don't know of any way to notify the user that they have to perform migrations if they want to use `ujail`.

I'm thinking about adding a `migrate` command to the init script to perform the one-time migration via `chown`, any opinion on this?

net/adguardhome/files/adguardhome.init
alaviss: Yea, fixed with a `-x /sbin/ujail` check.

GeorgeSapkin:
This PR is superseded by #26225 which has already been merged, so I think it can be closed.

PR #17459, opened 2021-12-30
transmission: add ca_bundle support with procd-ujail

clayface:
Maintainer: @dangowrt
Compile tested: bcm53xx
Run tested: bcm53xx

Description:

With procd-ujail enabled, it is not possible to use HTTPS URLs, for
example when either for downloading torrent files or blocklists. The
followig example occurs when downloading a URL from the "Upload Torrent
Files" dialogue box:

Error adding
"https://releases.ubuntu.com/21.10/ubuntu-21.10-desktop-amd64.iso.torrent":
gotMetadataFromURL: http error 0: No Response

syslog will also hint that no CA_BUNDLE is being used:

transmission-daemon[6683]: [2021-12-30 20:01:30.990] web will verify
tracker certs using envvar CURL_CA_BUNDLE: none (web.c:455)

This patch rectifies this issue by adding a ca_bundle configurable,
enabled by default. This explicitly fixes the ca_bundle file location
to /etc/ssl/certs/ca-certificates.crt and adds this file to the procd
jail. On subsequent testing, HTTPS URL download functionality is
restored.

### DIFF
--- a/net/transmission/Makefile
+++ b/net/transmission/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=transmission
 PKG_VERSION:=3.00
-PKG_RELEASE:=17
+PKG_RELEASE:=18
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@GITHUB/transmission/transmission-releases/master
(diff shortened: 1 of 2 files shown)


clayface:
Some additional notes on this: with procd-ujail disabled there is no need to specify `CURL_CA_BUNDLE`, but I wasn't able to replicate this adding /etc/ssl/cert.pem or /etc/ssl/certs/ca-certificates.crt to the procd jail. It would only work when CURL_CA_BUNDLE was explicitly declared.

Not sure whether we really need the ca_bundle configurable either. This could be omitted and we could do the rest unconditionally.

dangowrt (approved):

PR #17465, opened 2021-12-31
zerotier: add configuration reload trigger

kolbyjack:
Backporting #17247 to 21.02 branch as was suggested in the PR discussion

### DIFF
--- a/net/zerotier/Makefile
+++ b/net/zerotier/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=zerotier
 PKG_VERSION:=1.6.6
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


BKPepe:
You should use ``git cherry-pick -x``, so we know in the git history that it was cherry-picked. Otherwise LGTM

kolbyjack:
> You should use `git cherry-pick -x`, so we know in the git history that it was cherry-picked. Otherwise LGTM

I didn't know about -x, re-picked

BKPepe:
Now, there's is missing ``PKG_RELEASE`` bump, which was in previous commit.

BKPepe:
Superseded by this commit: https://github.com/openwrt/packages/commit/15b41a6751e9f7a81c0532765e73a3c507ef9a86 pushed directly.

PR #17467, opened 2021-12-31
netdata: Update init script to use -D rather than -nd

jamesmacwhite:

Maintainer: @BKPepe @diizzyy
Compile tested: N/A
Run tested: OpenWrt 19.07.8

Description:

When netdata is started via the current init script, the following deprecation notice is logged:

```
/usr/sbin/netdata: deprecated option -- -nd -- please use -D instead.
 2021-12-31 16:41:00: netdata INFO  : MAIN : SIGNAL: Not enabling reaper
```

This updates the init script to use `-D` rather than `-nd` as recommended by the log message.

### DIFF
--- a/admin/netdata/Makefile
+++ b/admin/netdata/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=netdata
 PKG_VERSION:=1.30.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


BKPepe:
There should be commit description.

jamesmacwhite:
@BKPepe Thanks. As it passed the DOCO requirements and the commit title was self explanatory I didn't think it needed anything else, but I have now added a commit description as requested.

BKPepe:
Looks good. :+1: I will run test it tomorrow and approve it.

jamesmacwhite:
@BKPepe Thank you for merging!

PR #17094, opened 2021-11-07
miniupnpd: bump version, drop igdv1 variant, add nftables support

stintel:
Maintainer: NA
Compile/run tested: OpenWrt master r18045-71af55ae2a on [qoriq](https://git.openwrt.org/?p=openwrt/staging/stintel.git;a=shortlog;h=refs/heads/qoriq)

Please do not merge this before someone tested the iptables variant. I have currently only tested the nftables variant.

PR #17469, opened 2022-01-01
miniupnp: update to version 2.2.3

shalzz:
With refreshed patches


shalzz:
This version fixes the MS IGDv1 issues (miniupnp/miniupnp#539) and so the IGDv1 variant should probably be removed.

net/miniupnpd/Makefile
```diff
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=miniupnpd
-PKG_VERSION:=2.2.2
+PKG_VERSION:=2.2.3
 PKG_RELEASE:=3
```
PolynomialDivision, line 12: reset PKG_RELEASE to 1 again.

shalzz:
Seems like just a patch refresh is not sufficient for building the upnpc package, unfortunately I'm not too familiar with the make/cmake build system to fix this build issue.

openips (approved):

PolynomialDivision:
Can you seperate that into two commits and prefix it with the correct package name, so `miniupnpc`. If actually the package miniupnp, we should maybe think about combining both packages.

shalzz:
> Can you seperate that into two commits and prefix it with the correct package name, so `miniupnpc`. If actually the package miniupnp, we should maybe think about combining both packages.

Done. Do you want me to create them as separate PRs?

PolynomialDivision:
> Done. Do you want me to create them as separate PRs?

No, everything fine. Even better since they originate from the same project. :) I will wait for CI to pass and then I will merge.

stintel:
Next time please search open PRs first: #17094

PolynomialDivision:
> Next time please search open PRs first: #17094

Sorry. Can you maybe add yourself as PKG_MAINTAINER?

stintel:
> Can you maybe add yourself as PKG_MAINTAINER?

Considering it!

stintel:
After looking at the existing issues for miniupnpd I think I'll pass.

PR #17473, opened 2022-01-02
jool: update to 4.1.6, add init files and default configuration

tiagogaspar8:
Compile tested: WRT3200ACM on the latest trunk
Run tested: WRT3200ACM on the latest trunk

Description:

This commit adds support for starting and running jool through init scripts, with default config files as examples.

This is also my first time with startup scripts, so all advices are welcome!

Signed-off-by: Tiago Gaspar <tiagogaspar8@gmail.com>

### DIFF
--- a/net/jool/files/jool.init
+++ b/net/jool/files/jool.init
@@ -1,0 +1,57 @@
+#!/bin/sh /etc/rc.common
+
+START=54
+STOP=54
+
+PROG_NAT64="/usr/bin/jool"
+PROG_SIIT="/usr/bin/jool_siit"
+
+CONFIGFILE_NAT64="/etc/jool/jool-nat64.conf.json"
+CONFIGFILE_SIIT="/etc/jool/jool-siit.conf.json"
+
+config_parser(){
+	enabled=0
+	enabled_nat64=0
+	enabled_siit=0
+
+	config_load "jool"
+
+	#verify if the services are enabled in the configuration and populate it's variables
+	config_get_bool enabled general enabled 0
+	config_get_bool enabled_nat64 nat64 enabled 0
+	config_get_bool enabled_siit siit enabled 0
+
+	#If the main service is not enabled exit
+	[ "$enabled" -eq 0 ] && return 1
+
+	#if nat64 is enabled continue
+	if [ "$enabled_nat64" -gt 0 ]; then
+		#check if the orer is to start or stop
+		if [ "$1" -gt 0 ]; then
+			#start jool
+			$PROG_NAT64 file handle $CONFIGFILE_NAT64
+		else
+			$PROG_NAT64 -f $CONFIGFILE_NAT64 instance remove
+		fi
+	fi
+
+	#if siit is enabled continue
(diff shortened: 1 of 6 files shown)


net/jool/files/jool.init
```diff
@@ -0,0 +1,51 @@
...
+	#verify if the services are enabled in the configuration and populate it's variables
+	config_get_bool enabled general enabled 0
+	config_get_bool enabled_nat64 nat64 enabled 0
+	config_get_bool enabled_siit siit enabled 0
+
+	#If the main service is not enabled exit
+	[ "$enabled" -eq 0 ] && return 1
+
+	#if nat64 is enabled continue
+	if [[ "$enabled_nat64" -gt 0 ]]; then
```
neheb, line 22: [[]] is a bashism. replace with [].

net/jool/files/jool.init
```diff
@@ -0,0 +1,51 @@
...
+	fi
+
+	#if siit is enabled continue
+	if [[ "$enabled_siit" -gt 0 ]]; then
+		#check if the orer is to start or stop
+		if [[ "$1" -gt 0 ]]; then
+			#start jool
+			jool_siit file handle /etc/jool/jool-siit.conf.json
+		else
+			/usr/bin/jool_siit -f /etc/jool/jool-siit.conf.json instance remove
```
neheb, line 39: interesting that sometimes it's /usr/bin and sometimes the name. Normally a PROG variable is created for this.

neheb:
I recommend shellcheck for the init script.

net/jool/files/jool.init
tiagogaspar8: I'll fix that.

net/jool/files/jool.init
tiagogaspar8: What do you mean by a PROG variable? I'm not used to creating scripts. 😄

tiagogaspar8:
`I recommend shellcheck for the init script.`
That was a very interesting tool, thanks!
I believe I have fixed the issues you pointed out, the ones pointed out by the tool were not applicable after that.

tiagogaspar8:
Can someone help me understand why this is failing on x86? It's very strange that it compiles in other archs, and I've seen this is not the only one that fails in the x86 target.

tiagogaspar8:
Well, it seems to be an issue with these test builds because on my vm with default config it builds fine

neheb:
x86 failure is a problem with the SDK. Unrelated to this PR..

net/jool/files/jool.init
neheb: In other unit scripts I've seen PROG= at the top and $PROG referenced later on.

tiagogaspar8:
Is there anything missing for this to get merged?

I'll take the time to add that #16818 has been noticed and was already mentioned to the package developer who said that this will be fixed this weekend maybe, and sometime later there will probably be a new release with this addressed.

tiagogaspar8:
Thank you very much for your help @neheb ! 😁

PR #17483, opened 2022-01-03
golang: simplify build on darwin/arm64 host

svlobanov:
Normal golang bootstrap is not supported on darwin/arm64. Currently
it is possible to use external bootstrap via CONFIG_ variable. This
patch adds an ability to use host golang without defining CONFIG_
variable to set external bootstrap

Maintainer: @jefferyto 

Description: see above

### DIFF
--- a/lang/golang/golang/Makefile
+++ b/lang/golang/golang/Makefile
@@ -12,7 +12,7 @@ GO_VERSION_PATCH:=5
 
 PKG_NAME:=golang
 PKG_VERSION:=$(GO_VERSION_MAJOR_MINOR)$(if $(GO_VERSION_PATCH),.$(GO_VERSION_PATCH))
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 GO_SOURCE_URLS:=https://dl.google.com/go/ \
                 https://mirrors.ustc.edu.cn/golang/ \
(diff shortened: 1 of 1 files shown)


jefferyto:
I'm not inclined to accept this because:

* I don't see why darwin/arm64 should get special treatment when there are other targets that bootstrap Go doesn't compile on, e.g. linux/arm64
* External bootstrap is meant as an advanced feature (i.e. the user trusts their system Go compiler) and having to fill in the config option is a way to ensure the user knows what they are doing

I am open to counterarguments, though.

svlobanov:
> I'm not inclined to accept this because:
> 
> * I don't see why darwin/arm64 should get special treatment when there are other targets that bootstrap Go doesn't compile on, e.g. linux/arm64
> * External bootstrap is meant as an advanced feature (i.e. the user trusts their system Go compiler) and having to fill in the config option is a way to ensure the user knows what they are doing
> 
> I am open to counterarguments, though.

I got your point, I don't have strong counterarguments. In this case I will change my patch just to show more informative message related to external bootstrap and add some description related to golang bootstrap issue to the wiki page https://openwrt.org/docs/guide-developer/toolchain/buildroot.exigence.macosx

svlobanov:
Added info to https://openwrt.org/docs/guide-developer/toolchain/buildroot.exigence.macosx how to use homebrew&bootstrap variable

I'm closing this issue

PR #17487, opened 2022-01-03
Update Makefile: add dependency of 'ip-full'

jfsgithub2020:
Maintainer: Florian Eckert
Compile tested: none, as just the dependencies are affected during opkg install
Run tested: 21.02 release with and without 'ip-full' install shows that 'ip-full' is required for mwan3 to work.

Description:
* a clean install of mwan3 in openwrt-21.02 does not work, as vital features are missing in the busybox implementation of 'ip' that is installed by default
* installation of 'ip-full' resolves this and mwan3 works as designed
* => 'ip-full' is mandatory and thus must be in the dependancies of 'mwan3' package.

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -27,7 +27,8 @@ define Package/mwan3
      +iptables \
      +iptables-mod-conntrack-extra \
      +iptables-mod-ipopt \
-     +jshn
+     +jshn \
+     +ip-full
    TITLE:=Multiwan hotplug script with connection tracking support
    MAINTAINER:=Florian Eckert <fe@dev.tdt.de>
    PKGARCH:=all


feckert:
Could you please change the header of the commit to
`mwan3: add missing dependency to ip-full`
I have already change this in your pullrequest header
And could you also bump this number by one
https://github.com/openwrt/packages/blob/master/net/mwan3/Makefile#L12

jow-:
Note that "ip-tiny" would likely work as well.

jfsgithub2020:
'ip' is already part of the dependencies and did not fullfill the job as the version in busybox is lacking some feature. You are probably right though, that libbpf is not needed and therefore ip-tiny is ok. But I have not tried this (running a live openwrt that with users needing it - no quick chance to check if it works as well...). If anyone knows how to add dependencies with alternatives, that would be cool. (sorry, I am a real newbie).

feckert:
> Note that "ip-tiny" would likely work as well. Also both `ip-full` and `ip-tiny` provide a virtual `ip` which allows users to freely chose. So if at all, the mwan3 package should depend on `ip`, not on a particular ip build variant.

I don't know exactly what the difference is between ip-full and ip-tiny.
I always use ip-full.
When I look at the [dependency](https://github.com/openwrt/openwrt/blob/master/package/network/utils/iproute2/Makefile#L50), it is only libbpf.
So I assume that the dependency setting to `ip` is ok as suggested by @jow-.
@jfsgithub2020 Could you please update the dependency from `ip-full` to `ip`?
So it doesn't matter whether you have installed ip-full or ip-tiny.

jfsgithub2020:
I did look at the Makefile again. The dependancy to 'ip' was always there (see the list of dependancies). It seems that the busybox also seems to provide this simple 'ip' implementation and that is definitely not sufficient. As you pointed out 'ip-tiny' does work, just as 'ip-full' does. If anyone can tell me how to enforce either 'ip-tiny' or 'ip-full' but not just the simple  'ip' from busybox, I will update it. This is the same question that I posted above 8 days ago, sorry.

feckert:
So I have looked at this again. Can we not change the dependency in mwan3 as follows?

https://github.com/openwrt/packages/blob/master/net/mwan3/Makefile#L25
DEPENDS:= @(PACKAGE_ip-full||PACKAGE_ip-tiny)
@jow- Does this works, so we exclude ip from busybox which does not full fill the needs?

jfsgithub2020:
I think this would solve the issue, as I pointed out earlier. About the missing DCO, I don't quite understand what to do. Can someone take care of this for me? Or explain better?

GeorgeSapkin:
21.02 is EOL. Closing as stale.

PR #17488, opened 2022-01-03
nut: fix typo in nut-monitor init script

coudu:
Maintainer: none
Compile tested: x86_64, OpenWrt master
Run tested: x86_64, OpenWrt master

Description:
There is a typo in the nut-monitor init script causing an uci option to be missed when building daemon config file.


BKPepe:
1. PKG_RELEASE in Makefile needs to be bumped
2. Signed-off-by requires your real name.

coudu:
Thanks,

thought it was using AUTORELEASE, I should have checked...
I've pushed 3 PRs about nut package, should I push only one PR with the 3 commits plus a PKG_RELEASE bump so that it is only bumped once?
Git config changed with real name.

PR #17493, opened 2022-01-04
stubby: explicitly disable static linking

rsalvaterra:
Although undocumented, there's a way to explicitly disable static linking in
Stubby, setting the CMake build option ENABLE_GETDNS_STATIC_LINK to OFF (ON by
default). Make it so.

@jonathanunderwood, @neheb, this will stop `stubby` from linking to anything but `getdns` and `libyaml`. I tried to reproduce @anomeome's build failure to no avail, but in any case this seems the right thing to do. Thoughts?

### DIFF
--- a/net/stubby/Makefile
+++ b/net/stubby/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=stubby
 PKG_VERSION:=0.4.0
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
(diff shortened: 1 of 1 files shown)


anomeome:
Resolves build issue.

dhewg:
Fixes the build for me too.

Without this I get:
```
Package stubby is missing dependencies for the following libraries:
libunbound.so.8
```

rsalvaterra:
Ping, @neheb, @jonathanunderwood. It would be nice to get this merged, since it's fixing a build failure.

PR #17497, opened 2022-01-04
dockerd: Add "log-driver" option

corna:
Signed-off-by: Nicola Corna <nicola@corna.info>

Maintainer: @G-M0N3Y-2503
Compile tested: **not built**, modified directly `/etc/init.d/dockerd` and `/etc/config/dockerd` on my router
Run tested: tested on x86-64 (PCEngines apu4d4); `/tmp/dockerd/daemon.json` and the logging driver reported by `docker info` change as expected

Description:
The [docker guide](https://docs.docker.com/config/containers/logging/configure/) suggests using the "local" log driver to avoid disk exhaustion, as the default one "json-file" does not perform log rotation. This commit adds the option "log_driver" to let the user choose the logging driver, but does not change default one.

### DIFF
--- a/utils/dockerd/files/dockerd.init
+++ b/utils/dockerd/files/dockerd.init
@@ -175,6 +175,7 @@ process_config() {
 
 	# Don't add these options by default
 	# omission == docker defaults
+	config_get log_driver globals log_driver ""
 	config_get bip globals bip ""
 	config_get registry_mirrors globals registry_mirrors ""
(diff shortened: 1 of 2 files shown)


G-M0N3Y-2503:
Looks good to me.
Thanks for the PR,
I Didn't know about this caveat.

G-M0N3Y-2503 (approved):

PR #17505, opened 2022-01-06
krb5: fix build on macos

svlobanov:
search_paths_first is not supported by GCC so this darwin-specific
option should not be used during build process

Maintainer: @flyn-org 

Description: see above

### DIFF
--- a/net/krb5/patches/002_remove_unnecessary_flag_in_macos_build.patch
+++ b/net/krb5/patches/002_remove_unnecessary_flag_in_macos_build.patch
@@ -1,0 +1,44 @@
+commit 3c66c1fec7ce4edeff284a5e4abe32d0b4398539
+Author: Nikhil Benesch <nikhil.benesch@gmail.com>
+Date:   Wed Dec 22 18:58:17 2021 -0500
+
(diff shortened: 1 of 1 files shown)


flyn-org:
Should this be removed entirely, upstream? It looks like the checks you deactivated are inside a check for GCC. Is there ever a case that they should be active?

svlobanov:
There is similar commit in upstream https://github.com/krb5/krb5/commit/3c66c1fec7ce4edeff284a5e4abe32d0b4398539 but in general all `uname -p` checks are invalid during cross compilation. I will try to find more upstreamable solution so I'm status of this PR to draft

svlobanov:
I've changed this PR. Now it is just a backport from upstream, I've checked, upstream fix is enough to build OpenWrt krb5 on MacOS. The remaining `uname -s` in configure checks do not affect building process

BKPepe (approved):

flyn-org:
Looks good, and corresponds with upstream commit 3c66c1fec7ce4edeff284a5e4abe32d0b4398539. Recommend merge.

PR #17513, opened 2022-01-07
Update aria2.init

qiuzi:
procd_add_jail_mount affects the content of the configuration file


BKPepe:
1. Why do you need this?
 - Which issue do you experience? Do you have any logs?
 Commit descripton: ``Comment out the configuration file is not sufficient``.
 
 2. There are missing formal things like missing `Signed-off-by`, commit subject should start with package package, that's why the DCO fails for you.
 
 Closing.

PR #17515, opened 2022-01-07
crowdsec: update to 1.2.2

erdoukki:
Maintainer: Kerma Gérald <gandalf@gk2.net> / @erdoukki
Run tested: (test done on MVEBU arch53 EspressoBin & EspressoBin-ULTRA, OpenWrt 19.07.x and 21.02.x)
Compile tested: (arch=cortexa53, host=amd64)

Description:

Update crowdsec to latest upstream release version 1.2.2

Makefile rework
- use tagged version for download

### DIFF
--- a/net/crowdsec/Makefile
+++ b/net/crowdsec/Makefile
@@ -1,19 +1,17 @@
 # SPDX-License-Identifier: MIT
 #
-# Copyright (C) 2021 Gerald Kerma
+# Copyright (C) 2021-2022 Gerald Kerma
 #
 
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=crowdsec
-PKG_VERSION:=1.2.1
+PKG_VERSION:=1.2.2
 PKG_RELEASE:=$(AUTORELEASE)
 
-PKG_SOURCE_PROTO:=git
-PKG_SOURCE_URL:=https://github.com/crowdsecurity/crowdsec
-PKG_SOURCE_VERSION:=v$(PKG_VERSION)
-PKG_SOURCE_DATE:=20211117
-PKG_MIRROR_HASH:=dfe50e5fb4d32fb6c21275b3f13a837ad64a9054e78076b44325376dc003fc64
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/crowdsecurity/crowdsec/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=b815ac4fafa25f0e657b56baaba9a093ef64fbe7336e4bc29208eb47ed1af5e6
 
 PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=LICENSE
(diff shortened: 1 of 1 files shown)


erdoukki:
@PolynomialDivision
Héllo, can you check please ?

net/crowdsec/files/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
BKPepe, line 1: Why it was added? No need. It takes LICENSE file from downloaded tarball, which is extracted.

erdoukki: Removed

PolynomialDivision:
@erdoukki Thanks. Happy New Year. :) I see @BKPepe already does the review. If you need help again, than just ping me.

net/crowdsec/files/LICENSE
BKPepe: I wondered why did you added in the first place and if you tested it.

BKPepe (changes requested):
1. As you touched the reordering in the Makefile, why it is not possible to use tagged releases?
2. Squash commits and use force-push.

net/crowdsec/files/LICENSE
erdoukki: A simple mistake !
Thinking it was forget at the first time, but have not think of it in the downloaded tar.
Just an error...

erdoukki:
> * As you touched the reordering in the Makefile, why it is not possible to use tagged releases?

Yes, modified.

> * Squash commits and use force-push.

Yes, planned to, thanks

> * Would be nice if you can run test it.

I will do some tests and report here, okay

BKPepe (approved):
It looks good. Thanks. 👍 

_I didn't do any compile or run test._

erdoukki:
Tests done.
Upgrade fine.
Looks to work fine.

erdoukki:
Thanks

net/crowdsec/Makefile
```diff
@@ -24,25 +22,24 @@ PKG_BUILD_PARALLEL:=1
...
 
 CWD_SYSTEM:=openwrt
-CWD_BUILD_VERSION?=$(PKG_SOURCE_VERSION)
+
+CWD_BUILD_VERSION?=v$(PKG_VERSION)
 CWD_BUILD_GOVERSION:=$(shell go version | cut -d " " -f3 | sed -E 's/[go]+//g')
 CWD_BUILD_CODENAME:=alphaga
-CWD_BUILD_TIMESTAMP:=$(SOURCE_DATE_EPOCH)
-CWD_BUILD_TAG:=openwrt
+CWD_BUILD_TIMESTAMP:=$(shell date +%F"_"%T)
```
jefferyto, line 29: You should use `$(SOURCE_DATE_EPOCH)` even when you want to use a different date format (i.e. pass it into `date`) for reproducibility (to have the exact same binary executable every time you compile the same package with the same buildroot/sdk).

erdoukki: @jefferyto 
Thanks, it makes sense !

I applied to https://github.com/openwrt/packages/pull/17989
on https://github.com/openwrt/packages/pull/17989/commits/3d451298593fe9270564e02ed74c46ede4b0f732

erdoukki: I applied the same to https://github.com/openwrt/packages/pull/18015 on https://github.com/openwrt/packages/pull/18015/commits/d54d2319a4602f6e34e3aaa0bb6d634b9c21f65a

PR #17521, opened 2022-01-07
luajit: fix build on macos (ldconfig issue)

svlobanov:
fix ldconfig build issue. This patch is a backport from upstream:
https://github.com/LuaJIT/LuaJIT/commit/18c9cf7d3788a8f7408df45df92fc4ae3bcc0d80

Maintainer: @milani 

Description: see above

### DIFF
--- a/lang/luajit/Makefile
+++ b/lang/luajit/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=luajit
 PKG_VERSION:=2.1.0-beta3
-PKG_RELEASE:=5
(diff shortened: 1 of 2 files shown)


milani:
there has been some work to support ARM64 too. I wish Mike Pall could do a new release. Then we wouldn't need this patch at all.

milani (approved):

svlobanov:
@milani yes, we will be able to just drop this patch in the next release because this patch is a backport from upstream

btw, this patch is not related to arm64 or macos arm64 support, it is related to non-linux build environments without ldconfig

PR #17525, opened 2022-01-08
aria2: download session can now survive service restart

AdorableJiang:
Signed-off-by: Shengjiang Quan <qsj287068067@126.com>

Maintainer: @kuoruan
Compile tested: MT7621,HC5962,R22.1.1
Run tested: MT7621, HC5962, R22.1.1, tests done

Description:
Aria2 session (downloading status) would loss every time aria2 service (re)starts, e.g. mannually restart aria2 service or after device reboot.
Related issue: coolsnowwolf/lede#5829

This problem is caused by unconditionally 
1. directory creation of `$config_dir` in function `aria2_start()`;
2. file creation of `$session_file` by `_create_file()` in function `aria2_start()`.

In this PR, the existence of directory `$config_dir` and `$session_file` is checked before creating new ones. Further, if `$session_file` exists, check read&write permission of it. In this way, the `$session_file` can now survive aria2 service restart and device reboot.

# Verification
Verification of this contribution can be found here coolsnowwolf/packages#350.

### DIFF
--- a/utils/bluez/patches/206-sync.patch
+++ b/utils/bluez/patches/206-sync.patch
@@ -1,0 +1,27 @@
+--- a/obexd/client/sync.c
++++ b/obexd/client/sync.c
+@@ -209,7 +209,7 @@ static void sync_remove(struct obc_sessi
+ 	g_dbus_unregister_interface(conn, path, SYNC_INTERFACE);
+ }
+ 
+-static struct obc_driver sync = {
++static struct obc_driver sync2 = {
+ 	.service = "SYNC",
+ 	.uuid = SYNC_UUID,
+ 	.target = OBEX_SYNC_UUID,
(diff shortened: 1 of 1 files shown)


BKPepe:
1. Commit message should start with prefix, which is the package name. That's why the CI fails for you.
2. ``Signed-off-by should contain a real name.
3. There should be a commit message, why you are doing such changes. You mentioned them in PR, so copy&paste should be sufficient. 
4. You try to merge 4 commits and two of them merge commits. You need to remove them by rebasing your branch.
5. You need to bump ``PKG_RELEASE`` in the Makefile.

___


Nitpick: Would be nice if those mentioned/linked PRs will be in English or at least what is in there (some summary would be ok) instead of using Google Translate.

AdorableJiang:
Thanks for your kind reply.

PR #17535, opened 2022-01-09
ddns-scripts-cloudflare: allow explicit zone_id

hgl:
Maintainer: @feckert

Description:

When zone id is explicitly provided, there is no need for the API token to have read permission. Inspired by acme.sh's cloudflare logic.


feckert:
Please bump PKG_RELEASE and also add your comment form the PR to the commit.
`When zone id is explicitly provided, there is no need for the API token to have read permission. Inspired by acme.sh's cloudflare logic.`

hgl:
Done.

feckert:
CI/CD is broken, but this does not have anything todo with this change.

feckert:
Thanks merged

PR #17556, opened 2022-01-09
mdio-tools: update to version 1.0.1

robimarko:
Maintainer: @dmascord 
Compile tested: mvebu, Armada 7040, RB5009, OpenWrt master
Run tested: mvebu, Armada 7040, RB5009, OpenWrt master, using read/write of the mdio tool

Description:

Version 1.0.1 brought the following changes:
[v1.0.1] - 2021-11-26

Primarily fixes a few issues in the kernel module that were found
during a quick review from Russell King:

https://lore.kernel.org/netdev/YYPThd7aX+TBWslz@shell.armlinux.org.uk/
https://lore.kernel.org/netdev/YYPU1gOvUPa00JWg@shell.armlinux.org.uk/

- mdio: The mvls subcommand now supports flushing the ATU

- mdio-netlink: Plug some glaring holes around integer overflows of
  the PC.
- mdio-netlink: Release reference to MDIO bus after a transaction
  completes.

So, update to the latest version and switch the kernel module back
to fetching tarballs like the userspace tool does.

Signed-off-by: Robert Marko <robimarko@gmail.com>


mhei:
I'm confused: when we have the very same source definition, i.e. same URL, same version etc. shouldn't be the mirror hash the same, too?

robimarko:
Trust me, it's confusing for me as well.
Because initially, I used the same hash and then the builder would complain, no idea how its possible as the tarball should be exactly the same.

mhei:
Yepp, I got it... actually, both packages are downloading the sources on their own, so we have two different tarballs with only almost the same content: the first-level directory differs: one time it is mdio-netlink-... and the second tarballs has mdio-tools-...

robimarko:
Ahh, yeah.
I forgot that its being packaged with a top level folder, that explains it-

mhei (approved):
LGTM

PR #17565, opened 2022-01-10
Fix to prevent ddns scripts to spoil logs

tdy91:
Maintainer: me / @\<github-user>
Run tested: Architecture ARMv8 Processor rev 3, Model Raspberry Pi 4 Model B Rev 1.1, OpenWrt SNAPSHOT r17084+13-8307da3dbd

Description:
Avoid error messages such as "cURL error '7'" and "Transfer failed - retry [count]/0 in 60 seconds" to fill logs.

This happens using an infinite loop as every 60s, to get the public address of a WAN2 interface which is voluntarily disabled because used as an mwan3 failover of a main working WAN1 interface.

This change makes it possible to record the first occurrence of error messages only, which is sufficient.


stangri:
@tdy91 Have you considered the scenario where log is filled by other messages (from other packages/events) so that the sole error message from ddns-scripts expires/disappears from the log? That would leave user without any error messages and non-working ddns-scripts.

Since there's no maintainer in the Makefile, I'm tagging @chris5560 @feckert @dibdot who contributed to ddns-scripts in the past.

tdy91:
@stangri To take into account your scenario, it might be possible to add a numeric parameter, for example 'retry_count_log' which would have the following operation:
-1 (default): systematic log, like the original operation
0: limit the log to the first error
N> 0: log modulo N
Thanks

GeorgeSapkin:
No updates in over three years, merge conflicts and not following the [submission guidelines](https://openwrt.org/submitting-patches#submission_guidelines). Closing as stale. Feel free to reopen if anything changes.

PR #17567, opened 2022-01-10
wg-installer: fix using symlinks for conf files

PolynomialDivision:
It is useful to symlink babeld and olsrd to /tmp/ if we frequently write to those config files.

### DIFF
--- a/net/wg-installer/Makefile
+++ b/net/wg-installer/Makefile
@@ -54,7 +54,7 @@ endef
 
 define Package/wg-installer-server-hotplug-babeld
 	$(call Package/wg-installer-server)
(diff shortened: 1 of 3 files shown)


PolynomialDivision:
Again this cryptdev error:
```
make[5]: *** No rule to make target '/home/build/openwrt/build_dir/target-x86_64_musl/linux-x86_64/cryptodev-linux-cryptodev-linux-1.12/ioctl.o', needed by '/home/build/openwrt/build_dir/target-x86_64_musl/linux-x86_64/cryptodev-linux-cryptodev-linux-1.12/cryptodev.o'.  Stop.
```

PR #17559, opened 2022-01-10
aria2: session file existence check

AdorableJiang:
Maintainer: @kuoruan
Compile tested: MT7621,HC5962,R22.1.1
Run tested: MT7621, HC5962, R22.1.1, tests done

Aria2 session (downloading status) would loss every time aria2 service (re)starts, e.g. mannually restart aria2 service or after device reboot.

This problem is caused by unconditionally
1. directory creation of `$config_dir` in function `aria2_start()`;
2. file creation of `$session_file` by `_create_file()` in function `aria2_start()`.

In this PR, the existence of directory `$config_dir` and `$session_file` is checked before creating new ones. Further, if `$session_file` exists, check read&write permission of it. In this way, the `$session_file` can now survive aria2 service restart and device reboot.

# Verification
Verification of this contribution can be found here coolsnowwolf/packages#350.

Signed-off-by: Shengjiang Quan <qsj287068067@126.com>

PR #17569, opened 2022-01-11
aria2: session file existence check

AdorableJiang:
Maintainer: @kuoruan
Compile tested: MT7621,HC5962,R22.1.1
Run tested: MT7621, HC5962, R22.1.1, tests done

Description:
Aria2 session (downloading status) would loss every time aria2 service (re)starts, e.g. mannually restart aria2 service or after device reboot.

This problem is caused by unconditionally
1. directory creation of `$config_dir` in function `aria2_start()`;
2. file creation of `$session_file` by `_create_file()` in function `aria2_start()`.

In this PR, the existence of directory `$config_dir` and `$session_file` is checked before creating new ones. Further, if `$session_file` exists, check read&write permission of it. In this way, the `$session_file` can now survive aria2 service restart and device reboot.

# Verification
Verification of this contribution can be found here coolsnowwolf/packages#350.

Signed-off-by: Shengjiang Quan <qsj287068067@126.com>

### DIFF
--- a/net/aria2/files/aria2.init
+++ b/net/aria2/files/aria2.init
@@ -168,15 +168,25 @@ aria2_start() {
 	config_file_tmp="$config_dir/$NAME.conf.tmp"
 	session_file="$config_dir/$NAME.session.$section"
 
-	_make_dir "$config_dir" || {
-		_err "Can't create config dir: $config_dir"
+	# check directory existence before creating it
+	if [ ! -e "$config_dir" ]; then
+		_make_dir "$config_dir" || {
+			_err "Can't create config dir: $config_dir"
+			return 1
+		}
+	fi
+
+	_create_file "$config_file" "$config_file_tmp" || {
+		_err "Can't create files: $config_file, $config_file_tmp"
 		return 1
 	}
 
-	_create_file "$session_file" "$config_file" "$config_file_tmp" || {
-		_err "Can't create files: $session_file, $config_file, $config_file_tmp"
-		return 1
-	}
+	# check session file existence before creating it
+	if [ ! -e "$session_file" ]; then
+	 	_create_file "$session_file"|| { _err "Can't create files: $session_file"; return 1; }
+	elif [ ! -r "$session_file" ] || [ ! -w "$session_file" ]; then
+		_change_file_mode 600 "$session_file"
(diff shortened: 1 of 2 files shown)


net/aria2/files/aria2.init
```diff
@@ -168,16 +168,26 @@ aria2_start() {
 	config_file_tmp="$config_dir/$NAME.conf.tmp"
 	session_file="$config_dir/$NAME.session.$section"
 
-	_make_dir "$config_dir" || {
-		_err "Can't create config dir: $config_dir"
-		return 1
-	}
+	# check directory existence before creating it
+	if [ ! -e "$config_dir" ]; then
```
kuoruan, line 172: use `-d` ?

kuoruan: I notice that you are checking `config_dir` and `session_file` with `-e`, do you mean that `config_dir` can be a link file?

AdorableJiang: My mistake. I will change them to `-d` for `config_dir` and `-f` for `session_file`.

kuoruan: Keep it to `2`

kuoruan (approved):

kuoruan (approved):

stangri:
Like I pointed out in the [PR you closed](https://github.com/openwrt/packages/pull/17559), the `_make_dir` and `_create_file` should work fine as is without your changes.

The `_make_dir` already checks if the directory exists and `_create_file` does not delete existing files.

Please address these and stop closing existing PRs (especially if they have non-resolved comments) and creating new ones.

AdorableJiang:
Thanks for your explanation for the principle of `_make_dir` and `_create_file`. I have reproduced your statement and you are right.
But the initial motivation of my commit is that not only I but also other users encountered this sesssion-lost problem many times, e.g. coolsnowwolf/lede#5829 (the original content is in Chinese). Refer to your statement, it seems that this problem is not caused by the behavior of `_make_dir` and `_create_file`.

Please close this PR, since my commit has nothing to do with this problem.

P.S. the former PRs are closed automatically when I tried to remove "merge" commits from my forked repo, of course, in a wrong way.

PR #17574, opened 2022-01-11
mblaze: fix build on macos

svlobanov:
redefine OS=Linux due to OpenWrt is always Linux

Run tested: not really tested, but build for Linux is not changed and binary checksums on Linux (Ubuntu 20.04) and MacOS are the same.

Description: see above

### DIFF
--- a/mail/mblaze/Makefile
+++ b/mail/mblaze/Makefile
@@ -29,6 +29,9 @@ define Package/mblaze/description
   and interacting with mail messages which are stored in maildir folders.
 endef
 
+MAKE_FLAGS += \
+	OS="Linux"
+
 define Package/mblaze/install
 	$(INSTALL_DIR) $(1)/usr
 	$(CP) $(PKG_INSTALL_DIR)/usr/local/bin $(1)/usr


mail/mblaze/patches/100_allow-to-override-Makefile-OS-variable.patch
```diff
@@ -0,0 +1,24 @@
...
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+--- a/GNUmakefile
++++ b/GNUmakefile
+@@ -2,7 +2,7 @@ CFLAGS?=-g -O2 -fstack-protector-strong
+ override CFLAGS:=-Wall -Wno-switch -Wextra $(CFLAGS)
+ LDLIBS=-lrt
+ 
+-OS := $(shell uname)
++OS ?= $(shell uname)
```
paper42, line 21: Is this really required? `make OS=Linux` will overwrite this even without this patch.

svlobanov: you are right. I have removed this patch from PR and closed PR to upstream as not required

PR #17583, opened 2022-01-11
python-curl: fix build on macos

svlobanov:
setup.py detects macos (darwin) and adds -flat_namespace flag. This
flag is not compatible with GCC that is used to compile target.
darwin detection disabled using sed in Makefile

Maintainer: @valdi74 
Run tested: (armvirt/64, OpenWrt trunk)

Description: see above


lang/python/python-curl/Makefile
```diff
@@ -32,6 +32,16 @@ define Package/python3-curl/description
 PycURL is a Python interface to libcurl, the multiprotocol file transfer library.
 endef
 
+define Build/Prepare
+	$(call Build/Prepare/Default)
+
+ifeq ($(CONFIG_HOST_OS_MACOS),y)
+	#Disable darwin detection. It is not required on cross-compile build with Linux target
+	sed -i 's/darwin/fakevalue/g' $(PKG_BUILD_DIR)/setup.py
```
jefferyto, line 40: This feels too general/uncertain to me. I would prefer a patch to remove the [two relevant lines](https://github.com/pycurl/pycurl/blob/REL_7_44_0/setup.py#L351-L352).

It's also not necessary to test for macOS, the change can be made for all platforms.

jefferyto (commented):
`PKG_RELEASE` would also need to be incremented.

lang/python/python-curl/Makefile
svlobanov: I will do this, also, I have started a discussion with upstream how to fix it in upstream. https://github.com/pycurl/pycurl/issues/727

svlobanov: done (PKG_VERSION, OpenWrt patch instead of using sed in Makefile)

PR #17590, opened 2022-01-12
serdisplib: fix build on macos

svlobanov:
./configure script detects if serdisplib is built on non-linux build
host and disables framebuffer driver. It blocks touchscreen_tool
compilation. This detection is not required on cross-compile build
so it is disabled via ac_cv_build=$(GNU_TARGET_NAME) in Makefile

Maintainer: @dangowrt 

Description: see above


dangowrt:
Thank you!

PR #17592, opened 2022-01-12
lcdproc: fix build on macos

svlobanov:
./configure script detects mtab file and fails build if mtab is not
found on build host. It is not required for OpenWrt build due to
mtab is always /etc/mtab on OpenWrt

MacOS doesn't have mtab file so disable it via ac_cv_mtab_file var

Maintainer: @haraldg @pprindeville 

Description: see above


haraldg:
Thanks for looking into this. I think this is a good change.

pprindeville (approved):
LGTM

PR #17594, opened 2022-01-12
samba4: macOS compatibility - incomplete

httpstorm:
On macOS, `/bin/sh -c echo -n` does not support parameters, which results in `-n` being sent to the output:
`Checking uname sysname type : -n Linux`
Resolved by using `printf`.

@Andy2244 @neheb @regae @svlobanov Can you please have a look?
We can safely accept this patch as it is, and resolve the remaining issues in another patch, or discuss and solve the remaining build errors on macOS:
* detection of poptGetContext
* detection of statfs
* possibly other errors
```
Checking for system popt                                                          : yes
Checking for header popt.h                                                        : yes
Checking for library popt                                                         : no
Checking for poptGetContext                                                       : not found
Checking for library popt                                                         : no
Checking for poptGetContext                                                       : not found
ERROR: System library popt of version 0.0.0 not found, and bundling disabled
```
Possible workaround: remove `checkfunctions='poptGetContext'` in `buildtools/wafsamba/samba_third_party.py:15`.

Maintainer: @Andy2244 
Compile tested: fails on host macOS 11.6.2 x64, target x64, OpenWRT master f2c3875dfcbf82d993ebe20f9563125c5fde2c60

https://github.com/openwrt/openwrt/pull/4638#issuecomment-990655125

### DIFF
--- a/net/samba4/patches/105-macos-compatibility.patch
+++ b/net/samba4/patches/105-macos-compatibility.patch
@@ -1,0 +1,11 @@
+--- a/buildtools/wafsamba/samba_cross.py
++++ b/buildtools/wafsamba/samba_cross.py
+@@ -134,7 +134,7 @@ class cross_Popen(Utils.subprocess.Popen
+                 cross_answers_incomplete = True
+                 add_answer(ca_file, msg, ans)
+             (retcode, retstring) = ans
+-            args = ['/bin/sh', '-c', "echo -n '%s'; exit %d" % (retstring, retcode)]
++            args = ['/bin/sh', '-c', "printf '%s'; exit %d" % (retstring, retcode)]
+         real_Popen.__init__(*(obj, args), **kw)
+ 
+ 


svlobanov:
Could you please send the same to upstream? 
Nobody wants to support patches for a long time...

When upstream accepts this, we will be able to remove your patch from openwrt.

Andy2244:
Thanks looks fine, i have no issue adding this patch.

httpstorm:
> Could you please send the same to upstream? Nobody wants to support patches for a long time...
> When upstream accepts this, we will be able to remove your patch from openwrt.

@svlobanov 
PR to upstream:
https://gitlab.com/samba-team/samba/-/merge_requests/2328

I also tried bumping `samba4` to a new version: 4.14.11 has the same issues, while 4.15.x require small changes to `002-dnsserver-4.7.0.patch` and `009-samba-4-11-fix-host-tools-checks.patch.patch`. Next in `Makefile`, `--without-ntvfs-fileserver` and `--without-dnsupdate` are not supported and should be removed. Followed to more build errors:
`staging_dir/host/bin/pkg-config` fails to find `pkg-config.real`, which is present in the same directory. So we should focus on fixing 4.14 first.

svlobanov:
I think we should bump samba to the latest release, after that fix the issues related related to macos build, otherwise it will be required to rebase patches and fixes for latest release and upstream trunk (we will need to send as much as possible to upstream "dev" branch)

If you already know how to bump to the latest release, please open a PR

httpstorm:
> I think we should bump samba to the latest release, after that fix the issues related related to macos build, otherwise it will be required to rebase patches and fixes for latest release and upstream trunk (we will need to send as much as possible to upstream "dev" branch)
> 
> If you already know how to bump to the latest release, please open a PR

@svlobanov Yes, but I would only bump to a version which builds on both Linux and macOS. I suspect 4.15.3 would require more work than 4.14.11, though we can try both. Once we have something that builds, it would be much easier to jump between commits and adapt it to the latest release.

regae:
build error fix caused by samba is not configured for linux target.
for me, i need to override sys.platform to linux in buildtools/bin/waf

svlobanov:
1. echo -n to printf fix: https://gitlab.com/samba-team/samba/-/merge_requests/2374 . Merged to upstream
2. sys.platform issue fix: https://gitlab.com/samba-team/samba/-/merge_requests/2380/diffs?commit_id=a844d6dd1b899374d26a93dcbd11d6d49bf843aa . Under review by upstream members

When second PR is merged, I will add two openwrt patches (backports) and change openwrt Makefile to add sys.platform redefining

svlobanov:
my variant to fix samba4 build on macos: https://github.com/openwrt/packages/pull/17874

neheb:
Closing because of the other PR.

PR #17599, opened 2022-01-12
perl-device-serialport: fix build on macos

svlobanov:
./configure script detects macos specific system headers
(IOKit/serial/ioss.h and sys/ttycom.h) that are not available
during compile time. There is no way to pass ac_cv_* vars to
./configure script due to perl wrappers

To fix this issue, fake(empty) headers provided during compile
time if build host is MacOS

Maintainer: Paul Oranje <por@xs4all.nl>

Description: see above

### DIFF
--- a/lang/perl-device-serialport/Makefile
+++ b/lang/perl-device-serialport/Makefile
@@ -41,7 +41,20 @@ define Build/Configure
 	$(call perlmod/Configure,,)
 endef
 
+MOD_CFLAGS_PERL += $(if $(CONFIG_HOST_OS_MACOS),-I$(PKG_BUILD_DIR)/macos_compat,)
+
 define Build/Compile
+
+ifeq ($(CONFIG_HOST_OS_MACOS),y)
+	#Zeroize macos specific system headers found by ./configure
+	mkdir -p $(PKG_BUILD_DIR)/macos_compat
+	mkdir -p $(PKG_BUILD_DIR)/macos_compat/sys
+	mkdir -p $(PKG_BUILD_DIR)/macos_compat/IOKit
+	mkdir -p $(PKG_BUILD_DIR)/macos_compat/IOKit/serial
+	echo '' > $(PKG_BUILD_DIR)/macos_compat/sys/ttycom.h
+	echo '' > $(PKG_BUILD_DIR)/macos_compat/IOKit/serial/ioss.h
+endif
+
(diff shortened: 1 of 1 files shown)


lang/perl-device-serialport/Makefile
```diff
@@ -41,7 +41,20 @@ define Build/Configure
...
+
 define Build/Compile
+
+ifeq ($(CONFIG_HOST_OS_MACOS),y)
+	#Zeroize macos specific system headers found by ./configure
+	mkdir -p $(PKG_BUILD_DIR)/macos_compat
+	mkdir -p $(PKG_BUILD_DIR)/macos_compat/sys
+	mkdir -p $(PKG_BUILD_DIR)/macos_compat/IOKit
+	mkdir -p $(PKG_BUILD_DIR)/macos_compat/IOKit/serial
+	echo '' > $(PKG_BUILD_DIR)/macos_compat/sys/ttycom.h
```
neheb, line 54: I'd replace with touch.

neheb:
ping @poranje

lang/perl-device-serialport/Makefile
svlobanov: I can replace, but touch will generate 'no endline' warnings

neheb: Hmm OK.

pprindeville:
I'd rather see an upstream fix for this.

svlobanov:
> I'd rather see an upstream fix for this.

I will try to concat the author, but the last change was in 2007... so I'm not sure if he replies

@pprindeville do you known how to communicate with perl cpan authors? Do they have mailing list or smth like this? I can't find it...

pprindeville:
> > I'd rather see an upstream fix for this.
> 
> I will try to concat the author, but the last change was in 2007... so I'm not sure if he replies
> 
> @pprindeville do you known how to communicate with perl cpan authors? Do they have mailing list or smth like this? I can't find it...

If you go to the `URL` value (`https://metacpan.org/dist/Device-SerialPort`) it shows you on the top right-hand side the author, `Kees Cook`.  You can click on that to get to his page, and on the left-side of that page will be his email address, `pause.cpan1@outflux.net`.

poranje:
As I do not have any better solution: go ahead.

BTW: According https://outflux.net/?see=outflux_net Kees Cook's email is homepage@outflux.net.

PR #17600, opened 2022-01-13
ddns-scripts: pass user agent string

### LINKED ISSUE #17507: ddns-scripts: user-agent for wget
Hello

when updating the current IP address via ip_url option it can happen that wget gets a 403. It would be nice to have the option to add a user-agent string for wget to prevent this.

Can I write a pull request for that? Or does it seem useless?

Best regards
Claudio

camarel:
This adds a user agent string to the wget request.

Fixes #17507

Signed-off-by: Claudio Marelli <camarelli@gmx.net>

### DIFF
--- a/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh
+++ b/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh
@@ -745,6 +745,13 @@ do_transfer() {
 		# disable proxy if no set (there might be .wgetrc or .curlrc or wrong environment set)
 		[ -z "$proxy" ] && __PROG="$__PROG --no-proxy"
 
+		# user agent string if provided
+		if [ -n "$user_agent" ]; then
+			# replace single and double quotes
+			user_agent=$(echo $user_agent | sed "s/'/ /g" | sed 's/"/ /g')
+			__PROG="$__PROG --user-agent='$user_agent'"
+		fi
+
 		__RUNPROG="$__PROG '$__URL'"	# build final command
 		__PROG="GNU Wget"		# reuse for error logging
 
--- a/net/ddns-scripts/samples/ddns.config_sample
+++ b/net/ddns-scripts/samples/ddns.config_sample
@@ -221,6 +221,11 @@ config service "myddns"
 #	option ip_source "script"
 #	option ip_script ""
 
+	# This option can be used in combination with ip_source "web" and ip_url.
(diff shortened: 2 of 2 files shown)


net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh
```diff
@@ -183,6 +183,7 @@ ERR_LAST=$?	# save return code - equal 0 if SECTION_ID found
 [ -z "$retry_count" ]	  && retry_count=0	# endless retry
 [ -z "$use_syslog" ]      && use_syslog=2	# syslog "Notice"
 [ -z "$use_https" ]       && use_https=0	# not use https
+[ -z "$user_agent" ]      && user_agent=''	# user agent for wget
```
feckert, line 186: The way I see it, we don't need a default here.
Because if this variable is not set, then no agent string is set for a wget and it uses the default value. Whatever that is!

net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh
```diff
@@ -745,6 +745,13 @@ do_transfer() {
 		# disable proxy if no set (there might be .wgetrc or .curlrc or wrong environment set)
 		[ -z "$proxy" ] && __PROG="$__PROG --no-proxy"
 
+		# user agent string if provided
+		if [ ! -z "$user_agent" ]; then
```
feckert, line 749: We could then use `-n` here then we wouldn't have to invert `-z`?
Can you verify, whether it works like that as I suggested?

feckert (changes requested):

camarel:
@feckert: Fixed your comments. You are right, it looks better this way.

feckert:
@camarel It's not because of your change. Something is wrong in the CI on x86_64

@aparcar I think you already know this, that the CI throws an error here with x86_64.
https://github.com/openwrt/packages/runs/4815064861?check_suite_focus=true#step:5:3714
I've noticed that several times now. Can I help you to fix this?

feckert:
@camarel please squash the two commits. This is only one change.
Thanks for your work

@aparcar thanks for the replay if I find some time I will also have a look on this ci issue.
I would like to take a closer look at the whole github ci anyway, so this is a good starting point to get to know it better.

camarel:
@feckert one commit now, hope its fine now

PR #17604, opened 2022-01-13
mc: fix mksh subshell

paper42:
Maintainer: -
Run tested: aarch64, Turris MOX, OpenWrt 21.02, run mksh, run mc, ctrl-o

Description: before this patch, mc failed to open a subshell with ctrl-o if the shell was mksh.


paper42:
> make package/mc/{clean,refresh}

Sorry for the delay, done

Issue #17602, opened 2022-01-13
mariadb - Uses sudo without having dependency to install it


najdanovicivan:
Maintainer: @miska 
Environment: x86_64 

Description:

Init script uses sudo when upgrading mysql.  Usage of sudo should be removed or sudo should be added as dependency

https://github.com/openwrt/packages/blob/dafb96c14837bf133e5215cf65fc7ed1439a375a/utils/mariadb/files/mysqld.init#L156

miska:
Fixed and merged

PR #17608, opened 2022-01-14
Mariadb owner fix & sudo dependency

miska:
Maintainer: me
Compile tested: Turris Omnia, mvebu/cortex-a9, OpenWrt 21.02
Run tested: Turris Omnia, mvebu/cortex-a9, OpenWrt 21.02, tested migration of datadir from different user

Description:
This makes sure that if in some previous version of MariaDB or MySQL datadir was owned by somebody else, it will have the owner corrected and the database will successfully start. Also fix dependency on sudo (#17602)

Tested-By: @BKPepe


BKPepe:
x86_64 failure is not related, merging.

PR #17612, opened 2022-01-15
ddns: fix $zone_id if statement error

wangkezun:

Description:
the if statement that check $zone_id is empty need to be reversed


feckert:
This is the wrong fix
It should be `if [ -n "$zone_id" ]; then`. There is missing a space before `]`

This was added by your change @hgl could you have a look

hgl:
#17617

Thanks for spotting it.

feckert:
Fixed with https://github.com/openwrt/packages/commit/aed4dbf34960543de5637ba97bb85a0d3b66874e 
Thanks @wangkezun for spotting this

PR #17619, opened 2022-01-15
[21.02] 2to3: add package host tool

erdoukki:
Maintainer: Kerma Gérald / @erdoukki
Compile tested: (LinuxMint amd64 host, OpenWrt master)
Run tested: (none)

Maybe python version will not change for 21.02.x release and this PR may be obsolete, but I prefer propose it, for review, in case it will be needed later or already needed.

Description:
2to3 is a Python program that reads Python 2.x source code and applies a
  series of fixers to transform it into valid Python 3.x code. The standard
  library contains a rich set of fixers that will handle almost all code. 2to3
  supporting library lib2to3 is, however, a flexible and generic library, so it
  is possible to write your own fixers for 2to3. lib2to3 could also be adapted
  to custom applications in which Python code needs to be edited automatically.

This tool is necessary for fail2ban package because of issue
  https://github.com/openwrt/packages/issues/17311
  https://github.com/openwrt/packages/pull/17341

Simple 2to3.py script from Debian, thanks to Matthias Klose <doko@ubuntu.com>
From: https://salsa.debian.org/cpython-team/python3-defaults

(cherry picked from commit 2f91e1c9c5876bcdac4630fdc0e417903d8f9eb9)
Co-authored-by: Jeffery To <jeffery.to@gmail.com>

### DIFF
--- a/lang/python/2to3/Makefile
+++ b/lang/python/2to3/Makefile
@@ -1,0 +1,51 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=2to3
+PKG_VERSION:=1.0
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=Gerald Kerma <gandalf@gk2.net>,Jeffery To <jeffery.to@gmail.com>
+PKG_LICENSE:=PSF-2.0
+PKG_LICENSE_FILES:=copyright
+
+PKG_HOST_ONLY:=1
+HOST_BUILD_DEPENDS:=python3/host
+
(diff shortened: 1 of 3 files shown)


erdoukki:
@jefferyto
What do you think of this PR ?

> Maybe python version will not change for 21.02.x release and this PR may be obsolete, but I prefer propose it, for review, in case it will be needed later or already needed.

jefferyto:
The issue isn’t the version of Python but the version of setuptools, and setuptools was updated to 58.1.0 with Python 3.9.9.

erdoukki:
Thanks, do you think this is also needed for 19.07.x ?

jefferyto:
No, 19.07 should receive security-related fixes only.

erdoukki:
Okay, thanks.
Good to know !

PR #17620, opened 2022-01-16
ddns-scripts: add dynv6.com prefix

Marionic0723:
Just update ipv6 prefix for dynv6 domains

Description:Just update ipv6 prefix for dynv6 domains


feckert:
That is the same as  https://github.com/openwrt/packages/blob/master/net/ddns-scripts/files/usr/share/ddns/default/dynv6.com.json except the ipv6. Shouldn't we then drop the IPv4?

Can't we change this provider name to ipv6-with-prefix.com? You also need to update the list file.

BKPepe:
Closing due to inacitivity.

PR #17630, opened 2022-01-17
ddns-scripts: Create update_one_com.sh

LugicoHDPlayer:
Signed-off-by: Lugico <main@lugico.de>

Maintainer: me (@LugicoHDPlayer)
Compile tested: -
Run tested: OpenWRT-21.02.1 x89-64 generic ext4 combined Virtual Machine (VirtualBox, Windows 10 Host)

Description:
ddns-script for one.com dns support as per https://github.com/LugicoHDPlayer/one.com-ddns-python-script/issues/2

### DIFF
--- a/net/ddns-scripts/files/usr/lib/ddns/update_one_com.sh
+++ b/net/ddns-scripts/files/usr/lib/ddns/update_one_com.sh
@@ -1,0 +1,112 @@
+#!/bin/sh
+
+# ONE.COM DDNS SCRIPT
+# REQUIRES CURL
+# $ opkg install curl
+
+# SCRIPT BY LUGICO
+# CONTACT: main@lugico.de
+
+[ -z "$CURL" ] && [ -z "$CURL_SSL" ] && write_log 14 "one.com communication require cURL with SSL support. Please install"
+[ -z "$domain" ]   && write_log 14 "Service section not configured correctly! Missing 'domain'"
+[ -z "$username" ] && write_log 14 "Service section not configured correctly! Missing 'username'"
+[ -z "$password" ] && write_log 14 "Service section not configured correctly! Missing 'password'"
+
+write_log 0 "one.com ddns script started"
+
+local __SUBDOMAIN __MAINDOMAIN __LOGINURL __RECORDID
+local __TTL=3600
+
+__SUBDOMAIN=$(echo $domain | sed -e 's/[^\.]*\.[^\.]*$//' -e 's/\.$//' )
+__MAINDOMAIN=$(echo $domain | sed -e "s/$__SUBDOMAIN\.//" )
+
+
+# LOGGING IN
+__LOGINURL=$( $CURL \
+  -RsSL \
+  --stderr $ERRFILE \
+  -c /tmp/one_com_cookiejar \
+  "https://www.one.com/admin/" \
+  | grep 'Login-form login autofill' \
+  | sed -e 's/.*action="//' -e 's/".*//' -e 's/\&amp;/\&/g' \
+)
+
+if ! [ "$( $CURL \
+  -RsSL \
+  --stderr $ERRFILE \
+  -c /tmp/one_com_cookiejar \
+  -b /tmp/one_com_cookiejar \
+  "$__LOGINURL" \
+  -H "Content-Type: application/x-www-form-urlencoded" \
+  -X POST \
+  -d "username=$username&password=$password&credentialId=" \
+  | grep "Invalid username or password.")" == "" ] ; then
+  write_log 14 "Invalid credentials"
+  return 1
+fi
+
+
+# SETTING DOMAIN
+$CURL -RsSL \
+  --stderr $ERRFILE \
+  -o $DATFILE \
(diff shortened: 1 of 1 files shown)


feckert:
Please update the package makefile with a new package and also add the missing json file.
For more detailed information, take a look at the commit history of the other scripts.

net/ddns-scripts/files/usr/lib/ddns/update_one_com.sh
```diff
@@ -0,0 +1,112 @@
...
+__LOGINURL=$( $CURL \
+  -RsSL \
+  --stderr $ERRFILE \
+  -c /tmp/one_com_cookiejar \
+  "https://www.one.com/admin/" \
+  | grep 'Login-form login autofill' \
+  | sed -e 's/.*action="//' -e 's/".*//' -e 's/\&amp;/\&/g' \
+)
+
+if ! [ "$( $CURL \
```
feckert, line 34: Couldn't we take the call from the if statement and then check the return value from the subshell to see if it is non-zero?

Example code:
```
$($CURL \
  -RsSL \
  --stderr $ERRFILE \
  -c /tmp/one_com_cookiejar \
  -b /tmp/one_com_cookiejar \
  "$__LOGINURL" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -X POST \
  -d "username=$username&password=$password&credentialId=" \
  | grep "Invalid username or password.")"
result="$?"

if [ "$result" != 0 ]; then
  write_log 14 "Invalid credentials"
  return 1
fi
```

net/ddns-scripts/files/usr/lib/ddns/update_one_com.sh
```diff
@@ -0,0 +1,112 @@
...
+  -RsSL \
+  --stderr $ERRFILE \
+  -c /tmp/one_com_cookiejar \
+  -b /tmp/one_com_cookiejar \
+  "$__LOGINURL" \
+  -H "Content-Type: application/x-www-form-urlencoded" \
+  -X POST \
+  -d "username=$username&password=$password&credentialId=" \
```
feckert, line 42: This line should be change to `-d "username=${username}&password=${password}&credentialId="`
so the variable is enclosed with breaks `{` `}`

net/ddns-scripts/files/usr/lib/ddns/update_one_com.sh
```diff
@@ -0,0 +1,112 @@
...
+fi
+
+
+# SETTING DOMAIN
+$CURL -RsSL \
+  --stderr $ERRFILE \
+  -o $DATFILE \
+  -c /tmp/one_com_cookiejar \
+  -b /tmp/one_com_cookiejar \
+  "https://www.one.com/admin/select-admin-domain.do?domain=$__MAINDOMAIN"
```
feckert, line 55: This should also go into an if else statement, what happens if this request is incorrect?

net/ddns-scripts/files/usr/lib/ddns/update_one_com.sh
```diff
@@ -0,0 +1,112 @@
...
+
+__SUBDOMAIN=$(echo $domain | sed -e 's/[^\.]*\.[^\.]*$//' -e 's/\.$//' )
+__MAINDOMAIN=$(echo $domain | sed -e "s/$__SUBDOMAIN\.//" )
+
+
+# LOGGING IN
+__LOGINURL=$( $CURL \
+  -RsSL \
+  --stderr $ERRFILE \
+  -c /tmp/one_com_cookiejar \
```
feckert, line 28: Can't we use a temporary file here throughout the script for `/tmp/one_com_cookiejar` ?See for example https://github.com/openwrt/packages/blob/38f01ad2c9738c6c5a5dbda85019bb9cf5bc10f8/net/speedtest-netperf/files/speedtest-netperf.sh#L268

feckert (changes requested):
Please squash the commits to one. So wie have one commit after the merge.
Also replace \<spaces\> with \<tabs\>. This saves characters

net/ddns-scripts/files/usr/lib/ddns/update_one_com.sh
LugicoHDPlayer: This request is used to select a domain in case the user owns more than one. If this is unsuccessful the script can still work so long as the user only owns one domain. Furthermore, if you try selecting a domain that isn't yours, it will automatically select your default domain instead and redirect to the control panel page. I could possibly implement a check if the correct domain was selected though.

LugicoHDPlayer:
I'm kind of new to git and accidentally merged with openwrt:master again so i can't squash the commits. Should i create a new pull request?

feckert:
I would recommend that you create a feature branch for such changes.
Then you can rebase and squash all you want.

Please read the workflow for git.

LugicoHDPlayer:
So that would still require a new pull request though, am i getting this right?

I've created a feature branch now and will make a new pull request from there unless there is a way to fix this pull request that i'm missing

feckert:
As I know you have to make a new pullrequest

PR #17634, opened 2022-01-18
lpc21isp: fix build on macos

svlobanov:
lpc21isp Makefile detects Darwin and defines __APPLE__ that is not
required for cross-compile build for OpenWrt

This patch sets OSTYPE="Linux" due to OpenWrt is always Linux

Maintainer: @Skeen 

Description: see above


Skeen:
LGTM

PR #17640, opened 2022-01-18
wg-installer: check if a key is already inserted

PolynomialDivision:
Check if a peer is already existing with a given public key. Introduce a response code for signaling why the server rejected the request.

### DIFF
--- a/net/wg-installer/wg-server/lib/wg_functions.sh
+++ b/net/wg-installer/wg-server/lib/wg_functions.sh
@@ -3,6 +3,12 @@
 . /usr/share/libubox/jshn.sh
 . /usr/share/wginstaller/wg.sh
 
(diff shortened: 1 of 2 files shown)


net/wg-installer/wg-server/lib/wg_functions.sh
```diff
@@ -40,6 +51,14 @@ wg_register () {
 	local mtu=$2
 	local public_key=$3
 
+	if [ "$(wg_key_exists $public_key)" == "1" ]; then
```
neheb, line 54: == is a bashism IIRC.

PolynomialDivision: Thanks. What do you think about the newest change?

PR #17644, opened 2022-01-19
ntpd: add option for enabling gpsd

PolynomialDivision:
Time synchronization can also be done via gpsd. Until now you have to patch the init.d-file. Add an option that allows using ntpd in combination with gpsd.

Replaces the tutorial:
https://openwrt.org/docs/guide-user/services/ntp/gps

Now it is enough to set:
    option enable_gpsd '1'

Maintainer: @tripolar 
Compile tested: tbd
Run tested: master


PolynomialDivision:
https://github.com/openwrt/openwrt/pull/4972

PR #17645, opened 2022-01-19
CI: disable x86/64 due to constant failures

aparcar:
Due to constant build failures of `cryptodev-linux` the SDK should no
longer be used for CI testing.

Disable testing until `cryptodev-linux` is fixed fox x86/64.

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -47,10 +47,6 @@ jobs:
             target: x86-geode
             runtime_test: true
 
-          - arch: x86_64
-            target: x86-64
-            runtime_test: true
-
     steps:
       - uses: actions/checkout@v2
         with:


champtar:
I tried to reproduced the failure yesterday and couldn't, most recent failure I saw was the ola PR, rerunning made it pass the x86-64, maybe a recent commit fixed it in master

champtar:
(on my phone, just saw today's failure)

jow-:
This happens when the SDK is built on a system without libelf support, in this case the kernel objtool won't get built and bundled. If the SDK is then later executed on a system where a host libelf is present, kernel stack validation will get enabled but building kmods will fail due to the missing objtool. This should affect all out of tree kmods.

> 21:07 < jow> in case the *SDK* got built by a host system which had no libelf support, then stack validation support must be forcibly disabled in  the SDK in order to prevent the SDK from picking up the host libelf from wahtever system it is executed on
> 21:08 < jow> since it will ship without a working kernel objtool
> 21:09 < jow> however I do not like the way it silently fails
> 21:09 < jow> if CONFIG_STACK_VALIDATION is enabled then kbuild should simply die if the prereqs are not satisfied
> 21:15 < jow> something like this (untested): https://pastebin.com/XktsEaxK

aparcar:
Sounds like an immediate step is to update all Buildbot runners? `libelf-dev` is installed on all workers.

> https://gitlab.com/openwrt/buildbot/-/blob/master/docker/buildworker/Dockerfile#L26

champtar:
I guess not all buildbot runners are running the same version, because the sdk images (so the sdk) was broken, then worked yesterday, and is now broken again. With `65ab664f4557` I always get a failure, with `980f31ecab6f` build is always succesfull

```
$ podman images
REPOSITORY                                        TAG            IMAGE ID      CREATED        SIZE
docker.io/openwrt/sdk                             broken         65ab664f4557  17 hours ago   1.81 GB
docker.io/openwrt/sdk                             x86_64-master  65ab664f4557  17 hours ago   1.81 GB
docker.io/openwrt/sdk                             working        980f31ecab6f  41 hours ago   1.82 GB
```
```
podman run --rm -ti docker.io/openwrt/sdk:broken bash

./scripts/feeds update -a
make defconfig
./scripts/feeds install cryptodev-linux
make package/cryptodev-linux/compile -j4
```

BKPepe:
We can close it, right? Based on recent discussion on #openwrt-adm/#openwrt-devel on OFTC, all buildbots were updated to the latest version, so this issue should no  longer appear.

champtar:
Yup, first 25 PR seems ok, let's call it fixed

PR #17646, opened 2022-01-19
ddns-scripts: add one.com provider

LugicoHDPlayer:
Signed-off-by: Lugico <main@lugico.de>

Maintainer: me (@LugicoHDPlayer)
Compile tested: N/A
Run tested: OpenWRT-21.02.1 x89-64 generic ext4 combined Virtual Machine (VirtualBox, Windows 10 Host)

Description:
ddns-script for one.com as per LugicoHDPlayer/one.com-ddns-python-script#2
(Since i messed up my first pull request when trying to merge from master i've cleaned things up a bit and merging from a feature branch instead now)


net/ddns-scripts/Makefile
```diff
@@ -237,6 +237,20 @@ define Package/ddns-scripts-transip/description
...
+	$(call Package/ddns-scripts/Default)
+	TITLE:=Extension for one.com Control Panel
+	DEPENDS:=ddns-scripts +curl
+endef
+
+define Package/ddns-scrtips-one/description
+	Dynamic DNS Client scripts extension for "one.com".
+	It requires:
+	"option username" to be a valid Email for one.com Control Panel
+  "option password" to be the matching one.com Control Panel password
```
feckert, line 250: Indentation

net/ddns-scripts/files/usr/lib/ddns/update_one_com.sh
```diff
@@ -0,0 +1,125 @@
...
+
+# GETTING RECORD ID
+__RECORDID=$( $CURL \
+	-RsSL \
+	--stderr $ERRFILE \
+	-c $COOKIEJAR \
+	-b $COOKIEJAR \
+	"https://www.one.com/admin/api/domains/${__MAINDOMAIN}/dns/custom_records" \
+	| sed 's/,/\n/g' \
+	| while read line ; do
```
feckert, line 77: I don't think that's very legible either! can't we rewrite that.

I understand your approach, that you process the data immediately without storing it in between.
How many lines are we talking about?
The maintenance of the code is not good. I was already willing to blame the last line with the `done \`.

feckert (changes requested):

net/ddns-scripts/files/usr/lib/ddns/update_one_com.sh
LugicoHDPlayer: The cURL request returns a single JSON line with all the custom DNS records the user has created. It can range anywhere from just a few to multiple thousands, though that's rather unlikely I think. It should be safe to store the data for a bit instead of immediately processing it. This was just my first solution.

I didn't notice the backslash in `done \` was unnecessary, I'll remove it.

feckert: If this is a json, then we can also process this with json_init?
Search for this string in the code and hopefully you will understand what I mean.

LugicoHDPlayer: Thanks for the idea. The jshn library is a lot more elegant, I wasn't aware that even existed. Let me know if I can improve anything on my implementation of it.

LugicoHDPlayer:
If everything is looking good so far, should I squash the commits for merge?

feckert:
Yes please.
So we only have one commit, because this is a new ddns provider

LugicoHDPlayer:
Commits have been squashed. Anything else I need to do for merge?

net/ddns-scripts/files/usr/lib/ddns/update_one_com.sh
```diff
@@ -0,0 +1,132 @@
...
+
+
+# SENDING PATCH
+$CURL \
+	-RsSL \
+	--stderr $ERRFILE \
+	-c $COOKIEJAR \
+	-b $COOKIEJAR \
+	-X PATCH \
+	-d "{\"type\":\"dns_service_records\",\"id\":\"${__RECORDID}\",\"attributes\":{\"type\":\"A\",\"prefix\":\"${__SUBDOMAIN}\",\"content\":\"${__IP}\",\"ttl\":${__TTL}}}" \
```
feckert, line 119: I have also noticed that we could use `json_*` there.
https://github.com/openwrt/openwrt/blob/ea49690ff47d14ee096d7c1368cff03f1cce1d76/package/network/config/netifd/files/lib/netifd/dhcp.script#L81

Everything else is ready to merge

LugicoHDPlayer: Done. Patch request is now being generated using jshn.

feckert:
CI/CD failed with the following issue. https://github.com/openwrt/packages/runs/4920796824?check_suite_focus=true
Please update your commit with a Signed-off-by line.
After that fix we could merge :+1:

LugicoHDPlayer:
Signed-off-by lines should be fixed now.

feckert:
Sorry to be so picky, but the commit message

```
ddns-scripts: add one.com provider

Squashed for merge

Signed-off-by: Luca Conte <main@lugico.de>
```
does not fit.

Either remove `Squashed for merge` or write what you did with the commit.

LugicoHDPlayer:
Removed the `Squashed for merge`  👍

feckert:
Thanks for your work :+1:

LugicoHDPlayer:
Had some trouble with previous commits being made using a different identity but I'm pretty sure things should be fine now.

feckert:
So now we ready for merge.
Thank you for your work.

PR #17650, opened 2022-01-19
postfix: fix build on macos

svlobanov:
macos build fails due to two reasons:
1. using build host ar and ranlib tools
2. using uname -r to get kernel version

First issue is fixed by specifying ar and ranlib from toolchain
Second issue is fixed by specifying kernel release major version=5
Using 'uname -r' from build host for cross-compiling is not a good
idea even for Linux build host

Maintainer: @Shulyaka 
Run tested: (armvirt/64, OpenWrt trunk)

Description: see above

### DIFF
--- a/mail/postfix/patches/500-crosscompile.patch
+++ b/mail/postfix/patches/500-crosscompile.patch
@@ -5,15 +5,17 @@
  case $# in
   # Officially supported usage.
 - 0) SYSTEM=`(uname -s) 2>/dev/null`
-+ 0) SYSTEM="OpenWRT"
++ 0) SYSTEM="OpenWrt"
      RELEASE=`(uname -r) 2>/dev/null`
      # No ${x%%y} support in Solaris 11 /bin/sh
      RELEASE_MAJOR=`expr "$RELEASE" : '\([0-9]*\)'` || exit 1
-@@ -242,6 +242,15 @@ case "$SYSTEM" in
+@@ -242,6 +242,17 @@ case "$SYSTEM" in
  esac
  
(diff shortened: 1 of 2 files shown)


Shulyaka (approved):

mail/postfix/patches/500-crosscompile.patch
```diff
@@ -9,11 +9,13 @@
      RELEASE=`(uname -r) 2>/dev/null`
      # No ${x%%y} support in Solaris 11 /bin/sh
      RELEASE_MAJOR=`expr "$RELEASE" : '\([0-9]*\)'` || exit 1
-@@ -242,6 +242,15 @@ case "$SYSTEM" in
+@@ -242,6 +242,17 @@ case "$SYSTEM" in
  esac
  
  case "$SYSTEM.$RELEASE" in
-+   OpenWRT*)    SYSTYPE=LINUX$RELEASE_MAJOR
++   OpenWRT*)    SYSTYPE=LINUX5
```
BKPepe, line 16:
```suggestion
+   OpenWrt*)    SYSTYPE=LINUX5
```

svlobanov: Will do.

TODO: also change OpenWRT->OpenWrt here https://github.com/openwrt/packages/blob/3900ec1f34e2a221c0531ae7ff595e83a2b5d683/mail/postfix/patches/500-crosscompile.patch#L8

svlobanov: done

PR #17670, opened 2022-01-21
jool: remove iptables support and update templates

tiagogaspar8:
Compile tested: WRT3200ACM on the latest master
Run tested: WRT3200ACM on the latest master

Description:

This pull request removes iptables backend support from jool and leaves only the netfilter backend support. This means that:
-  iptables and nftables firewall based systems (firewall3 and firewall 4) are supported trough the netfilter instance mode
-  the iptables/xtables mode support is disabled

For more information on the modes and how to use the new netfilter instance checkout https://www.jool.mx/en/intro-jool.html

This move is made out of the commit upstream that sets firewall4 as the default for new default builds and based on the conversation in #16818 and was decided that the netfilter interface is the priority since iptables support will be dropped in the foreseeable future.

While at it update the templates provided.

### DIFF
--- a/net/jool/Makefile
+++ b/net/jool/Makefile
@@ -8,20 +8,25 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=jool
-PKG_VERSION:=4.1.6
+PKG_VERSION:=2022-01-20
 PKG_RELEASE:=$(AUTORELEASE)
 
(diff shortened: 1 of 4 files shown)


tiagogaspar8:
I have no idea how this patch is dirty, I have done a `make package/jool/refresh` at least 15 times...

neheb:
make package/jool/{clean,refresh}

tiagogaspar8:
I've already done that yet this time it worked, thanks!

PR #17671, opened 2022-01-21
sqm-scripts: Change iptables dependency to iptables-nft

tohojo:
There's only one of the shaper scripts (simple.qos) that uses iptables, and
it only uses it for matching DSCP bits and setting the fwmark. So we should
be fine with just the basic iptables-nft dependency, and this will allow
sqm-scripts to be installed with the new default nft-based firewall.

Maintainer: me 

Looking for testers; I don't have anything that runs snapshots these days...

### DIFF
--- a/net/sqm-scripts/Makefile
+++ b/net/sqm-scripts/Makefile
@@ -24,8 +24,7 @@ include $(INCLUDE_DIR)/package.mk
 define Package/sqm-scripts
   SECTION:=net
   CATEGORY:=Base system
-  DEPENDS:=+tc +kmod-sched-cake +kmod-ifb +iptables \
-	+iptables-mod-ipopt +iptables-mod-conntrack-extra
+  DEPENDS:=+tc +kmod-sched-cake +kmod-ifb +iptables-nft +iptables-mod-ipopt
   TITLE:=SQM Scripts (QoS)
   PKGARCH:=all
 endef


hnyman:
Not sure if replacing the dependency of iptables with iptables-nft is enough, as the calling script might need to be changed also.

"iptables" command leads to the old iptables, while you need "iptables-nft" command to reach the nft enabled one.

```
 OpenWrt SNAPSHOT, r18614-46ce629fe0
 -----------------------------------------------------
root@router1:~# iptables -V
iptables v1.8.7 (legacy)
root@router1:~# iptables-nft -V
iptables v1.8.7 (nf_tables)
```

I think that the "ipt" function in the defaults.sh (at your own upstream repo) needs to be changed, too, so that calls iptables-nft.

tohojo:
Hannu Nyman ***@***.***> writes:

> Not sure if replacing the dependency of iptables with iptables-nft is
> enough, as the calling script might need to be changed also.
>
> "iptables" command leads to the old iptables, while you need
> "iptables-nft" command to reach the nft enabled one.

Oh, I see. I just assumed that the iptables-nft binary installed a
compatibility symlink to the iptables binary; otherwise, what's the
point of having a compatibility binary if applications have to select it
explicitly? Shouldn't this be fixed in the iptables-nft package instead?

hnyman:
> I just assumed that the iptables-nft binary installed a compatibility symlink to the iptables binary; otherwise, what's the point of having a compatibility binary if applications have to select it explicitly? Shouldn't this be fixed in the iptables-nft package instead?

Good question. 
So far there is not much info about the usage of iptables-nft in the OpenWrt context. 

@aparcar enabled some iptables option with https://github.com/openwrt/openwrt/commit/0637093e8cb0e4cfb94dfcfe7b57afc23bf88852 in order to provide some nftables compatibility, but I don't think that the option actually installs iptables-nft. I had to install it separately.  

And so far the "iptables" and "iptables-nft" lead to different binaries. I manually installed iptables-nft package, and after that it looks like this:
```
root@router1:/usr/sbin# ls -lt /usr/sbin/*tabl*
lrwxrwxrwx    1 root     root            17 Jan 21 17:48 /usr/sbin/iptables-nft -> xtables-nft-multi
lrwxrwxrwx    1 root     root            17 Jan 21 17:48 /usr/sbin/iptables-nft-restore -> xtables-nft-multi
lrwxrwxrwx    1 root     root            17 Jan 21 17:48 /usr/sbin/iptables-nft-save -> xtables-nft-multi
lrwxrwxrwx    1 root     root            17 Jan 21 17:48 /usr/sbin/iptables-restore-translate -> xtables-nft-multi
lrwxrwxrwx    1 root     root            17 Jan 21 17:48 /usr/sbin/iptables-translate -> xtables-nft-multi
lrwxrwxrwx    1 root     root            20 Jan 21 17:19 /usr/sbin/iptables -> xtables-legacy-multi
lrwxrwxrwx    1 root     root            20 Jan 21 17:19 /usr/sbin/iptables-restore -> xtables-legacy-multi
lrwxrwxrwx    1 root     root            20 Jan 21 17:19 /usr/sbin/iptables-save -> xtables-legacy-multi
-rwxr-xr-x    1 root     root         58631 Jan 21 17:19 /usr/sbin/xtables-legacy-multi
-rwxr-xr-x    1 root     root        149855 Jan 14 01:42 /usr/sbin/xtables-nft-multi
```

Would be great if somebody with actual knowledge about the nftables firewall infra would enlighten us about the transition tools. (And maybe that symlink of iptables --> iptables-nft should be added, like some desktop distros apparently do.

hnyman:
I think that it (currently) needs to also select the old iptables.
iptables-nft has for some strange reason the "at" dependency for iptables, instead of selecting it.

https://github.com/openwrt/openwrt/blob/0637093e8cb0e4cfb94dfcfe7b57afc23bf88852/package/network/utils/iptables/Makefile#L111

That causes a circular dependency with your patch.
```
Collecting package info: done
tmp/.config-package.in:38676:error: recursive dependency detected!
tmp/.config-package.in:38676:	symbol PACKAGE_libxtables is selected by PACKAGE_tc-mod-iptables
tmp/.config-package.in:108984:	symbol PACKAGE_tc-mod-iptables is selected by PACKAGE_tc-tiny
tmp/.config-package.in:108997:	symbol PACKAGE_tc-tiny is selected by PACKAGE_sqm-scripts
tmp/.config-package.in:1243:	symbol PACKAGE_sqm-scripts depends on PACKAGE_libxtables
```

Selecting also iptables in sqm fixes it.

```
--- a/net/sqm-scripts/Makefile
+++ b/net/sqm-scripts/Makefile
@@ -24,8 +24,7 @@ include $(INCLUDE_DIR)/package.mk
 define Package/sqm-scripts
   SECTION:=net
   CATEGORY:=Base system
-  DEPENDS:=+tc +kmod-sched-cake +kmod-ifb +iptables \
-       +iptables-mod-ipopt +iptables-mod-conntrack-extra
+  DEPENDS:=+tc +kmod-sched-cake +kmod-ifb +iptables +iptables-nft
   TITLE:=SQM Scripts (QoS)
   PKGARCH:=all
 endef
``` 

On the long the better solution would likely be to change dependency directly at the iptables-nft, but I am not even sure why there is that dependency, as AFAIK iptables-nft is more related to nftables than to iptables...

tohojo:
Looks like that circular dependency comes from 'tc' though? Guess that shouldn't depend on iptables either...

hnyman:
I  think that whole iptables/nftables/tc/whatever dependency universe is currently twisted, as the dependencies have been drafted from the "iptables as default" perspective.  The same goes also for the various kmods. (and ip6tables-nft and all that ipv6 side...)

Hopefully people with enough knowledge about the modules can straigthen the dependencies out.

tiagogaspar8:
The question is, even though iptables is switched to iptables-nft tc still selects iptables for some reason.

hnyman:
Looks like the automatic iptables-nft conversion does not quite work, at least yet. 
I have applied the PRs both on OpenWrt (netfilter and tc reorg, 5004 and 5007) and this SQM PR, so the "iptables" is actually "iptables-nft".
However,  the simple.qos does not set the requested dcsp marks.

Apparently the iptables --set-mark command does not (currently) translate well with iptables-nft:

The curious part is that based on googling network, it actually should work. Possibly we are missing some option toggle regarding iptables-nft extensions, or similar.

Here is SQM restart that looks successful, but in reality the dscp commands apparently fails. After the restart, I tried the key commands manually.

```
root@router1:~# /etc/init.d/sqm restart
SQM: Stopping SQM on eth0.2
SQM: Starting SQM script: simple.qos on eth0.2, in: 190000 Kbps, out: 57000 Kbps
SQM: simple.qos was started on eth0.2 successfully

root@router1:~# iptables -t mangle -A QOS_MARK_eth0.2 -m dscp --dscp-class CS1 -j MARK --set-mark 0x3
iptables v1.8.7 (nf_tables): Couldn't load match `dscp':No such file or directory

Try `iptables -h' or 'iptables --help' for more information.

root@router1:~# iptables-translate  -t mangle -A QOS_MARK_eth0.2 -m dscp --dscp-class CS1 -j MARK --set-mark 0x3
iptables-translate v1.8.7 (nf_tables): Couldn't load match `dscp':No such file or directory

Try `iptables-translate -h' or 'iptables-translate --help' for more information.

root@router1:~# iptables -t mangle -A QOS_MARK_eth0.2 -j MARK --set-mark 0x2
iptables v1.8.7 (nf_tables): unknown option "--set-mark"
Try `iptables -h' or 'iptables --help' for more information.

root@router1:~# iptables-nft -t mangle -A QOS_MARK_eth0.2 -j MARK --set-mark 0x2
iptables v1.8.7 (nf_tables): unknown option "--set-mark"
Try `iptables -h' or 'iptables --help' for more information.
```

tohojo:
Hannu Nyman ***@***.***> writes:

> Looks like the automatic iptables-nft conversion does not quite work, at least yet. 
> I have applied the PRs both on OpenWrt (netfilter and tc reorg, 5004 and 5007) and this SQM PR, so the "iptables" is actually "iptables-nft".
> However,  the simple.qos does not set the requested dcsp marks.
>
> Apparently the iptables --set-mark command does not (currently)
> translate well with iptables-nft:

The 'dscp' match should be usable through the iptables-translate tool:

https://wiki.nftables.org/wiki-nftables/index.php/List_of_available_translations_via_iptables-translate_tool

And it "just" requires the "nft_payload" module, which AFAICT is enabled
in the base openwrt nft install. Similar with the '-j MARK' which
requires the nft_meta module...

Just tried it on my Arch linux box with iptables-nft, and that seems to
work fine. Hmm...:

$ lsmod | grep nft
nft_counter            16384  0
nft_compat             20480  0
nft_ct                 20480  2
nf_conntrack          180224  1 nft_ct
nf_tables             266240  51 nft_ct,nft_compat,nft_counter
nfnetlink              20480  2 nft_compat,nf_tables
x_tables               57344  4 xt_dscp,nft_compat,ip_tables,xt_mark


$ cd build/linux/net/netfilter; grep nft_compat
Makefile:obj-$(CONFIG_NFT_COMPAT)       += nft_compat.o


So maybe we're missing CONFIG_NFT_COMPAT and the nft_compat module? I
was under the impression that iptables-nft did the full conversion in
userspace and wasn't dependent on anything in the kernel, but it appears
not? If I remove the nft_compat module it gets autoloaded by the next
iptables-nft invocation, so it's definitely needed...

hnyman:
> So maybe we're missing CONFIG_NFT_COMPAT and the nft_compat module? I

I think that we actually do have that, see below.

```
root@router1:~# lsmod | grep nf
libcrc32c              16384  1 nf_tables
nf_conntrack           86016  8 nft_redir,nft_nat,nft_masq,nft_flow_offload,nft_ct,nf_nat,nf_flow_table,nf_conntrack_netlink
nf_conntrack_netlink   36864  0
nf_defrag_ipv4         16384  1 nf_conntrack
nf_defrag_ipv6         20480  1 nf_conntrack
nf_flow_table          32768  4 nf_flow_table_ipv6,nf_flow_table_ipv4,nf_flow_table_inet,nft_flow_offload
nf_flow_table_inet     16384  0
nf_flow_table_ipv4     16384  0
nf_flow_table_ipv6     16384  0
nf_nat                 36864  4 nft_redir,nft_nat,nft_masq,nft_chain_nat
nf_reject_ipv4         16384  2 nft_reject_ipv4,nft_reject_inet
nf_reject_ipv6         16384  2 nft_reject_ipv6,nft_reject_inet
nf_tables             163840216 nft_fib_inet,nf_flow_table_ipv6,nf_flow_table_ipv4,nf_flow_table_inet,nft_reject_ipv6,nft_reject_ipv4,nft_reject_inet,nft_reject,nft_redir,nft_quota,nft_objref,nft_numgen,nft_nat,nft_masq,nft_log,nft_limit,nft_hash,nft_flow_offload,nft_fib_ipv6,nft_fib_ipv4,nft_fib,nft_ct,nft_counter,nft_compat,nft_chain_nat
nfnetlink              16384  4 nft_compat,nf_tables,nf_conntrack_netlink
nft_chain_nat          16384  2
nft_compat             20480  0
nft_counter            16384 18
nft_ct                 20480  3
nft_fib                16384  3 nft_fib_inet,nft_fib_ipv6,nft_fib_ipv4
nft_fib_inet           16384  0
nft_fib_ipv4           16384  1 nft_fib_inet
nft_fib_ipv6           16384  1 nft_fib_inet
nft_flow_offload       16384  0
nft_hash               16384  0
nft_limit              16384  5
nft_log                16384  0
nft_masq               16384  1
nft_nat                16384  0
nft_numgen             16384  0
nft_objref             16384  0
nft_quota              16384  0
nft_redir              16384  0
nft_reject             16384  3 nft_reject_ipv6,nft_reject_ipv4,nft_reject_inet
nft_reject_inet        16384  2
nft_reject_ipv4        16384  0
nft_reject_ipv6        16384  0
x_tables               24576  1 nft_compat
```

hnyman:
It might be that we do have the needed modules, but the iptables-nft translation tool fails to translate --set-mark.
Adding mark manually with nft seems to succeed...
(no idea about the proper command to mimic the simple.qos dscp conditional, but just testing the rudimentary)

```
root@router1:~# nft list ruleset | tail
table ip mangle {
        chain QOS_MARK_eth0.2 {
                meta mark set ct mark
                meta mark set 0x00000002
        }
}
table ip6 mangle {
        chain QOS_MARK_eth0.2 {
        }
}

root@router1:~# nft add rule mangle  QOS_MARK_eth0.2 meta mark set 3

root@router1:~# nft list ruleset | tail
        chain QOS_MARK_eth0.2 {
                meta mark set ct mark
                meta mark set 0x00000002
                meta mark set 0x00000003
        }
}
table ip6 mangle {
        chain QOS_MARK_eth0.2 {
        }
}
```

hnyman:
Maybe like this?

```
root@router1:~# nft add rule mangle  QOS_MARK_eth0.2 ip dscp cs1 meta mark set 3

root@router1:~# nft list ruleset | tail -n 10
}
table ip mangle {
        chain QOS_MARK_eth0.2 {
                ip dscp cs1 meta mark set 0x00000003
        }
}
table ip6 mangle {
        chain QOS_MARK_eth0.2 {
        }
}

```

champtar:
You actually need to keep the iptables-mod-* dependencies (and need my PR to have nft-compat)
https://manpages.debian.org/buster/iptables/iptables-nft.8.en.html
"These tools use the libxtables framework extensions and hook to the nf_tables kernel subsystem using the nft_compat module."

champtar:
And if you are looking how to modify current mark / do some bitwise operation, everything is in the quick ref
https://wiki.nftables.org/wiki-nftables/index.php/Quick_reference-nftables_in_10_minutes

champtar:
```
root@OpenWrt:/# iptables -t mangle -N QOS_MARK_eth0.2
root@OpenWrt:/# iptables -t mangle -A QOS_MARK_eth0.2 -m dscp --dscp-class CS1 -j MARK --set-mark 0x3

table ip mangle {
	chain QOS_MARK_eth0.2 {
		# xt_dscp counter packets 0 bytes 0 # xt_MARK
	}
}
```

hnyman:
@champtar 

Thanks for advice.

> You actually need to keep the iptables-mod-* dependencies

Sounds like we have a bit misleading iptables-mod-*  kmod names, probably due to historical reasons. Calling them as iptables-something sounds wrong in the nftables context. Should the common kmods be xtables-something?

Hopefully your PR 5004 gets polished enough that you could merge at least the first parts :-)
Maybe there could also be another PR to clarify the descriptions (or names?) of the kmods, so that also laymen (like me) understand more easily which kmods are still relevant in the nftables context.

champtar:
@hnyman for now I've reduced my PR to the strict minimum as I don't want to introduce regression without a full clean solution were we can really have everything splitted, but I'm hesitant to spend too much time on this, better work on moving to nft, that might be time better spent ;)

champtar:
BTW even linux has some legacy naming, for example ipt_REJECT.ko should be named xt_REJECT.ko

tohojo:
Alright, put the iptables-mod-ipopt dependency back (don't think we ever used the other one)...

hnyman:
The iptables fixes from champtar were merged today, so this should now be ok, and easier to test. (I will test later today)

hnyman:
I compiled with this now that the iptables fixes have been merged. Seems to work ok.
Merging.

tohojo:
@hnyman So someone on IRC pointed out that with this, it is no longer possible to select the old fw3 and sqm-scripts at the same time. Not sure what we can do about that - AFAIK we can't express "either or" dependencies, right? Any ideas?

hnyman:
> someone on IRC pointed out that with this, it is no longer possible to select the old fw3 and sqm-scripts at the same time

The old firewall3 is going to be abandoned. 
Note that buildbot includes the firewall4 by default into firmware images.

If somebody wants to use firewall3, he could also modify SQM Makefile sources

But, regarding 
> AFAIK we can't express "either or" dependencies, right? 

We can.

We could maybe say:
```
DEPENDS:=+tc +kmod-sched-cake +kmod-ifb +firewall4:iptables-nft +firewall3:iptables +iptables-mod-ipopt
```

Or, as iptables-mod-ipopt selects iptables  (since  https://github.com/openwrt/openwrt/commit/b0bd6599e840c443d8ccb8759315e4dd006fd6aa), also this should work:

```
DEPENDS:=+tc +kmod-sched-cake +kmod-ifb +firewall4:iptables-nft +iptables-mod-ipopt
```

hnyman:
@tohojo 
My error in the previous example.
The old firewall3 is still named just "firewall".
So, the correct one is: `DEPENDS:=+tc +kmod-sched-cake +kmod-ifb +firewall4:iptables-nft +firewall:iptables +iptables-mod-ipopt`

But the later, simpler proposal it still correct:
`DEPENDS:=+tc +kmod-sched-cake +kmod-ifb +firewall4:iptables-nft +iptables-mod-ipopt`

@jow- @aparcar 
Should the old iptables firewall package be named as "firewall3" just for clarity ?
It is a bit confusing that the abandoned firewall packages is called just as "firewall", while the current one is "firewall4".

aparcar:
> Should the old iptables firewall package be named as "firewall3" just for clarity ?

I'd prefer that but I can not estimate if this introduces issues down the line. @stintel thoughts?

champtar:
> > Should the old iptables firewall package be named as "firewall3" just for clarity ?
> 
> I'd prefer that but I can not estimate if this introduces issues down the line. @stintel thoughts?

Should be easy to fix breakage IMO
As firewall3 was installed by default likely nobody depends on it

pprindeville:
> > someone on IRC pointed out that with this, it is no longer possible to select the old fw3 and sqm-scripts at the same time
> 
> The old firewall3 is going to be abandoned. Note that buildbot includes the firewall4 by default into firmware images.
> 
> If somebody wants to use firewall3, he could also modify SQM Makefile sources

I think we should have used VARIANT to allow sqm scripts for both firewalls.

champtar:
Right now many packages have iptables dependencies, we might be able to switch to VARIANT once that's fixed

hnyman:
I think that the point is that there are people trying to stick to fw3 in master, by reverting the adoption of fw4. Introducing fw3 specific variants help them. It would be possible, but is somewhat extraneous in the long run.

E.g. there could be sqm-firewall3 package that would not depend on iptables-nft.

tohojo:
Hannu Nyman ***@***.***> writes:

> I think that the point is that there are people trying to stick to fw3 in master, by reverting the adoption of fw4. Introducing fw3 specific variants help them. It would be possible, but is somewhat extraneous in the long run.
>
> E.g. there could be sqm-firewall3 package that would not depend on
> iptables-nft.

But if we make the dependency on iptables-nft conditional on firewall4
as you suggested above, wouldn't it just work to install either version?

hnyman:
I think that the conditional only works at the build time, so it would help people self-compiling. But it would not work for people installing it later.

hnyman:
I could draft the variant (sqm-firewall3), but it is also a bit political question: how much we should spend effort in helping users trying to reverse/delay firewall4 adoption in the bleeding edge master.

tohojo:
No strong opinions either way; any idea what it is that make people
prefer to keep firewall3 around?

graysky2:
@hnyman
Is this the expected package set for adding `sqm scripts` to the [snapshot.config](https://downloads.openwrt.org/snapshots/targets/bcm27xx/bcm2711/config.buildinfo):
```
% diff snapshot.config new.config --color -u
--- snapshot.config	2022-02-08 06:54:47.490050115 -0500
+++ new.config	2022-02-08 06:54:50.790046544 -0500
@@ -17,11 +17,27 @@
 CONFIG_KERNEL_BUILD_DOMAIN="buildhost"
 CONFIG_KERNEL_BUILD_USER="builder"
 CONFIG_MAKE_TOOLCHAIN=y
+CONFIG_PACKAGE_iptables=y
+CONFIG_PACKAGE_iptables-mod-ipopt=y
+CONFIG_PACKAGE_iptables-nft=y
+CONFIG_PACKAGE_kmod-ifb=y
+CONFIG_PACKAGE_kmod-ipt-core=y
+CONFIG_PACKAGE_kmod-ipt-ipopt=y
+CONFIG_PACKAGE_kmod-nf-ipt=y
+CONFIG_PACKAGE_kmod-nft-compat=y
+CONFIG_PACKAGE_kmod-sched-cake=y
+CONFIG_PACKAGE_kmod-sched-core=y
 CONFIG_PACKAGE_libbpf=m
 CONFIG_PACKAGE_libelf=m
+CONFIG_PACKAGE_libip4tc=y
+CONFIG_PACKAGE_libip6tc=y
+CONFIG_PACKAGE_libxtables=y
+CONFIG_PACKAGE_libxtables-nft=y
 CONFIG_PACKAGE_qosify=m
+CONFIG_PACKAGE_sqm-scripts=y
 CONFIG_PACKAGE_tc-full=m
-CONFIG_PACKAGE_tc-mod-iptables=m
+CONFIG_PACKAGE_tc-mod-iptables=y
+CONFIG_PACKAGE_tc-tiny=y
 CONFIG_PACKAGE_zlib=m
 CONFIG_REPRODUCIBLE_DEBUG_INFO=y
 CONFIG_SDK=y
```

hnyman:
> Is this the expected package set

Why wouldn't it be?

There is iptables-nft plus its dependencies, cake, tc plus its dependencies, mod-ifb, iptables-mod-ipotp (including iptables) etc.

`DEPENDS:=+tc +kmod-sched-cake +kmod-ifb +iptables-nft +iptables-mod-ipopt`

You can traverse the dependency tree with menuconfig, and verify by yourself (what has selected what).

But it looks pretty normal to me.

graysky2:
@tohojo - I concerned that having two firewall packages could lead to problems is all.  If there is no danger in breakage, I am fine with things as they are.

hnyman:
> I was thinking that one wouldn't want to have the full iptables package co-existing with nftables. If this package set is expected ...

It is sadly expected, if any iptables-mod-xxx stuff is used.

Please read the last 20 comments or so from
https://github.com/openwrt/openwrt/pull/5004

PR #17672, opened 2022-01-21
gpsd: fix build on macos

svlobanov:
gpsd build fails on macos due to using sys.platform in build logic.

This patch allows to redefine target platform to fix the issue.

Maintainer: @psidhu 
Compile tested: (armivirt/64, OpenWrt trunk)

PR to upstream: https://gitlab.com/gpsd/gpsd/-/merge_requests/290

Description: see above


svlobanov:
After a discussion with upstream, now this fix in upstream (modified by gpsd author). Upstreamed version of the patch is backported and added to this PR

PR #17678, opened 2022-01-22
collectd: remove libip4tc dependency

tiagogaspar8:
Maintainer: Jo-Philipp Wich <jo@mein.io>, Hannu Nyman <hannu.nyman@iki.fi>
Compile tested: WRT3200ACM
Run tested: WRT3200ACM

Description:

Base collectd doesn't require libip4tc as stated in the collectd git:

* libiptc (optional)
For querying iptables counters
<http://netfilter.org/>

And libip4tc is already added as a dependency in the iptables plugin
because it needs it, so remove this dependency from the default collectd
to make it compatible with the new nftables firewall4 (by not selecting
any iptables components).

### DIFF
--- a/utils/collectd/Makefile
+++ b/utils/collectd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=collectd
 PKG_VERSION:=5.12.0
-PKG_RELEASE:=15
+PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=https://collectd.org/files/ \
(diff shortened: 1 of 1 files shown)


hnyman:
I tried to figure out why this has been added, but (due to SVN) there is no commit message from @nbd168  in the year 2012 commit:

https://git.openwrt.org/?p=openwrt/svn-archive/packages.git;a=commitdiff;h=ad8f143c93ab5af93e237aa3b433275f41bc15a8;hp=26a7a9b08cb5e4988d37e1339c72d37b19e62a92

But it looks like the libip4tc dependency has been added for the iptables plugin.

@jow- 
Any objections for this?  The CI builds seem run complete ok.

As a separate item, currently there is no official nftables plugin for collectd, but there is an abandoned upstream PR about it. I wonder if we should try to include it here.

hnyman:
@tiagogaspar8 
You need to also bump the PKG_RELEASE in Makefile.
Maybe we should now change it to `$(AUTORELEASE)` to get rid of the later bump needs.

tiagogaspar8:
>As a separate item, currently there is no official nftables plugin for collectd, but there is an abandoned upstream PR about it. I wonder if we should try to include it here.

I'd love to do that but that's beyond my abilities 😅

>Maybe we should now change it to $(AUTORELEASE) to get rid of the later bump needs.

Totally right, it's best way and it's done!

Issue #17641, opened 2022-01-18
ddns-scripts [nsupdate]: impossible to configure nsupdate with ipv6 without (a single) forced update


WildPenquin:
Maintainer:  feckert
Environment: 21.02.1

Description:

It is impossible to configure ddns-scripts-nsupdate for IPv6 protocol so that no forced updates happen *and* the ip address is till checked at intervals. This is because run-once is always enabled if and only if forced is disabled (set to 0). See: https://openwrt.org/docs/guide-user/services/ddns/client .

This is in conflict of nsupdate terms. They are very adamant on not updating the IP address if it has not been changed; from [nsupdate.info documentation](https://nsupdateinfo.readthedocs.io/en/latest/user.html):

> Note: please do not “update” your IP address if it did not change. Doing so is considered abusive use of the service. All sane dyndns2 clients only send an update if the IP address has changed.

Running the forced update when IP has not changed can cause an abuse flag to be enabled for the client.

Though, I'm not sure whether this will be an issue if the client sets a forced update, say, every 31th day. For cleanness sake, there should probably be a separate setting for _run-once_ and _no forced updates_.

WildPenquin:
A log snippet demonstrating a forced update when setting `forced_interval=0` (also containing my configuration):

```
$ dog mylogsnippet.log 
 205922       : ************ ************** ************** **************
 205922  note : PID '25878' started at 2022-01-18 20:59
 205922       : ddns version  : 2.8.2-12
 205922       : uci configuration:
ddns.nspudate.cacert='/etc/ssl/certs'
ddns.nspudate.check_interval='40'
ddns.nspudate.check_unit='minutes'
ddns.nspudate.domain='[MYHOSTNAME_REDACTED]'
ddns.nspudate.enabled='1'
ddns.nspudate.force_interval='0'
ddns.nspudate.force_unit='hours'
ddns.nspudate.interface='br-lan'
ddns.nspudate.ip_interface='br-lan'
ddns.nspudate.ip_source='interface'
ddns.nspudate.lookup_host='[MYHOSTNAME_REDACTED]'
ddns.nspudate.password='***PW***'
ddns.nspudate.retry_count='10'
ddns.nspudate.retry_interval='15'
ddns.nspudate.retry_unit='minutes'
ddns.nspudate.service_name='nsupdate.info'
ddns.nspudate.use_https='1'
ddns.nspudate.use_ipv6='1'
ddns.nspudate.use_syslog='2'
[...]

PR #17679, opened 2022-01-22
ddns-scripts: forced updates can now be properly disabled.

WildPenquin:
Maintainer: @feckert 
Compile tested: not applicable
Run tested: NONE

Description: As per bug #17641, this pull request actually enables disabling forced updates. The issue is not nsupdate specific but applies to the general scripts (however not all service providers might have gripes with updates when IP address has not been changed).

I haven't tested this yet, and I should before this gets merged. These are the test cases I can think of:

1. Forced updates on, checks done at intervals.
2. Forced updates on, single run only.
3. Forced updates off, checks done at intervals.
4. Forced updates off, single run only.

In the meanwhile I would like to hear if anyone has any objections to this kind of patch or the documentation changes. The code and comments should be self-explanatory.

I've made it so that there is a new run_once configuration, as this is the way I believe this should be fixed to retain all functionality. If it is not set (as it will not be in existing configurations) the old behavior is retained (i.e. always forced updates at intervals, but if forced_interval=0 -> run_once is internally set to 1 and forced_interval=1). This means that users who use their old configuration will not need to do anything.

### DIFF
--- a/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh
+++ b/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh
@@ -156,11 +156,11 @@ trap "trap_handler 15" 15	# SIGTERM	Termination
 #
 # force_interval	force to send an update to your service if no change was detected
 # force_unit		'days' 'hours' 'minutes' !!! force_interval="0" disables all forced updates, unless run_once=""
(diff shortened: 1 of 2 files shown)


feckert:
Please fixup your commits to one, because this is one change
Please add a proper header to this commit message `ddns-scripts: forced updates can now be properly disabled`

BKPepe:
Closing due to OP inactivity. If you are willing to continue on this, please address and reopen such changes. Thank you for any cooperation on this.

PR #17687, opened 2022-01-24
tcl: fix build on macos

svlobanov:
configure script detects Darwin and uses flags incompatible with
Linux target build.

This patch uses fakeuname tool if host OS is MacOS to avoid Darwin
detection on target build.

Maintainer: @mistachkin 
Compile tested: (armivirt/64, OpenWrt trunk)
Run tested: (armivirt/64, OpenWrt trunk)

Description: This PR should not be merged until ~~https://github.com/openwrt/openwrt/pull/4993~~ https://github.com/openwrt/packages/pull/17702 is merged


svlobanov:
fakeuname was merged to packages tree so this PR can be merged

PR #17688, opened 2022-01-24
stress-ng: fix build on macos

svlobanov:
This patch adds KERNEL=Linux to MAKE_FLAGS to avod Darwin detection.
If Makefile detects Darwin, it removes -lbsd from build flags, but
this flag is required due to target is always Linux, not bsd-like.

Maintainer: @commodo 

Description: see above


commodo (approved):

commodo:
Thank you :)

PR #17689, opened 2022-01-24
tvheadend: fix conffiles section

BKPepe:
Maintainer: @M95D
Compile and run tested: Turris 1.1, powerpc_8540, OpenWrt 21.02.1

Description:
The previous one was wrong, and it did not work. It could be checked
inside compiled package in control.tar.gz that there was missing
``conffiles`` file with content `/etc/config/tvheadend`

It is also possible to verify that the config is not overwritten on the router
by running ``opkg install tvheadend --force-reinstall``.

### DIFF
--- a/multimedia/tvheadend/Makefile
+++ b/multimedia/tvheadend/Makefile
@@ -230,7 +230,7 @@ define Build/Prepare
 		> $(PKG_BUILD_DIR)/debian/changelog
 endef
 
-define Package/conffiles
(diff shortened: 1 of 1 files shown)


M95D (approved):

M95D:
[Documentation](https://openwrt.org/docs/guide-developer/packages#packagedescription) should be improved.

BKPepe:
That's somehow true. It depends on the point of view.

PR #17705, opened 2022-01-26
reptyr: fix build on macos

svlobanov:
This patch sets UNAME_S=Linux due to OpenWrt is always Linux but
`uname -s` return Darwin on MacOS and fails target build.

Maintainer: @BKPepe 

Description: see above


BKPepe:
Sure. Why not.

PR #17709, opened 2022-01-27
Coova-chilli: improve init script

f00b4r0:
Maintainer: @neheb 

Description: This series of patches intends to improve the Coova-Chilli init script by:

- postponing service startup until wan becomes available
- executing firewall cleanup at service shutdown
- removing an hotplug trigger that interfered with e.g. disabling the service
- proposing an alternative to said hotplug trigger

These patches are fully independent and can thus be cherry-picked as desired.

HTH

### DIFF
--- a/net/coova-chilli/Makefile
+++ b/net/coova-chilli/Makefile
@@ -132,8 +132,6 @@ define Package/coova-chilli/install
 	$(INSTALL_CONF) $(PKG_INSTALL_DIR)/etc/chilli.conf $(1)/etc/
 	$(INSTALL_DIR) $(1)/etc/chilli
 	$(CP) $(PKG_INSTALL_DIR)/etc/chilli/* $(1)/etc/chilli/
-	$(INSTALL_DIR) $(1)/etc/hotplug.d/iface
-	$(INSTALL_DATA) ./files/chilli.hotplug $(1)/etc/hotplug.d/iface/30-chilli
 	$(INSTALL_DIR) $(1)/usr/sbin
 	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/sbin/chilli* $(1)/usr/sbin/
 	$(INSTALL_DIR) $(1)/usr/lib/
--- a/net/coova-chilli/files/chilli.hotplug
+++ b/net/coova-chilli/files/chilli.hotplug
@@ -1,7 +1,0 @@
-#!/bin/sh
-
-[ "$ACTION" == "ifup" ] || exit 0
-
-[ "$INTERFACE" = "wan" ] && {
-	/etc/init.d/chilli restart
-}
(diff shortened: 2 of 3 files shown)


neheb:
Ping @teslamint

teslamint:
The hotplug trigger is workaround of when wan interface is down and up for some reason and chilli does not recover from it, and that will make claims from real-world hotspot service users.

Since I no longer work with that service anymore, I wonder if it still happens and I think service trigger can replace the hotplug trigger.

f00b4r0:
> The hotplug trigger is workaround of when wan interface is down and up for some reason and chilli does not recover from it, and that will make claims from real-world hotspot service users.

That's how I understood it.

> Since I no longer work with that service anymore, I wonder if it still happens and I think service trigger can replace the hotplug trigger, but IIRC reload is not enough to recover chilli to back to service.

Ok. Then maybe replacing the current content of `reload_service()` with `stop; start` (i.e. force a restart on reload) and removing `reload_chilli` would be a more robust option?

FWIW the `reload` command of chilli_query is undocumented in the [manpage](https://coova.github.io/CoovaChilli/chilli_query(1).html), so maybe it's better not to use it at all?

f00b4r0:
I've just implemented the above proposal. HTH.

teslamint:
> Ok. Then maybe replacing the current content of `reload_service()` with `stop; start` (i.e. force a restart on reload) and removing `reload_chilli` would be a more robust option?
> 
> FWIW the `reload` command of chilli_query is undocumented in the [manpage](https://coova.github.io/CoovaChilli/chilli_query(1).html), so maybe it's better not to use it at all?

Restart chilli will remove in-memory user session data unless compile chilli with (binary) stat file compile option(or RADIUS server handles user session resumption), reload command may still needs for small configuration change while chilli still running.

f00b4r0:
> > Ok. Then maybe replacing the current content of `reload_service()` with `stop; start` (i.e. force a restart on reload) and removing `reload_chilli` would be a more robust option?
> > FWIW the `reload` command of chilli_query is undocumented in the [manpage](https://coova.github.io/CoovaChilli/chilli_query(1).html), so maybe it's better not to use it at all?
> 
> Restart chilli will remove in-memory user session data unless compile chilli with (binary) stat file compile option(or RADIUS server handles user session resumption), reload command may still needs for small configuration change while chilli still running.

Well, are "small configuration changes" a frequent enough scenario that trying to avoid a full restart is really needed? I'm also concerned that the lack of documentation of the "reload" command may suggest it may not be well supported.

But at the end this is your call. Still the hotplug script does not respect the disabled status of the service, and that is however a bug I think.

f00b4r0:
How about this?
The reload trigger is unchanged, and instead I added an explicit interface trigger that restarts the service.
HTH

teslamint (approved):

neheb:
this CI error has nothing to do with this package. Merging.

PR #17712, opened 2022-01-28
frp: improve frp init script

brvphoenix:
Maintainer: @ysc3839
Run tested: x86_64

Description:
1. Use anonymous sections and extra name option to define the proxy settings.  Some words such as `-` are not allowed in the uci section name.
2. Add `service_trigger` and define the `reload_service` explicitly without which the service will not reload the conf files if it is modified by the LuCI.

### DIFF
--- a/net/frp/files/frpc.init
+++ b/net/frp/files/frpc.init
@@ -37,6 +37,11 @@ config_cb() {
 	fi
 }
 
+service_triggers()
+{
+	procd_add_reload_trigger "$NAME"
+}
+
 start_service() {
 	local init_cfg=" "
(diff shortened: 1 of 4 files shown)


net/frp/files/frpc.init
```diff
@@ -35,6 +37,22 @@ config_cb() {
 	fi
 }
 
+upgrade() {
```
feckert, line 40: This should happen in a UCI default script. Otherwise it is always done at the startup of the service!

brvphoenix: Done

PR #17717, opened 2022-01-29
mandoc: fix build on macos

svlobanov:
./configure scripts compiles&runs test so it detects darwin
C-functions and defines that can not be used for Linux target.

This patches expicitly specifies Linux-specific parameters that
can not be detected correctly on Darwin.

This patch removes 010-mandoc-1.14.5-fix-gnu-make.patch. mandoc
provides its own way to redefine specific parameters (using
configure.local file). Defines from this patch are moved to
generated configure.local file.

CONFIGURE_VARS are removed by this patch due to ./configure script
does not use ENV vars specified by CONFIGURE_VARS

This commit does not change compilation on Linux hosts (the resulted
binary is exactly the same as before applying this commit). Binary
compiled on MacOS host is exactly the same as compiled on Linux.

Files generated by ./configure (config.h and Makefile.local) are
exactly the same on Linux and MacOS build hosts

Maintainer: @flyn-org 
Run tested: not really tested, but sha256sum is the same before&after this patch (mandoc binary produced on MacOS is exactly the same as produced on Linux)

Description: see above


flyn-org:
Looks good. Recommend merge.

PR #17723, opened 2022-01-29
tvheadend: fix build on macos

svlobanov:
tvheadend configure/make files detect Darwin build host and changes
build logic, but it fails compilation for OpenWrt target (Linux)

This patch explicitly specifies Linux as a target platfrom

Maintainer: @M95D 

Description: see above

### DIFF
--- a/multimedia/tvheadend/Makefile
+++ b/multimedia/tvheadend/Makefile
@@ -187,7 +187,17 @@ endif
 ##	CONFIGURE_ARGS += --disable-libopus
 ##endif
 
+#required to cross-compile hdhomerun on non-Linux build host
+MAKE_FLAGS += \
+	OS=Linux
+
+#required to always have "build.linux" dir, not "build.darwin" on macos
+CONFIGURE_VARS += \
+	PLATFORM=linux
+
+#--platfrom=linux is required to cross-compile tvheadend on non-Linux build host
 CONFIGURE_ARGS += \
+	--platform=linux \
 	--arch=$(ARCH) \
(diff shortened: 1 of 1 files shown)


neheb:
I think this PR somehow introduced a new error:

```
 Package tvheadend is missing dependencies for the following libraries:
libFLAC.so.8
libfdk-aac.so.2
libsane.so.1
```

can you verify this?

svlobanov:
> I think this PR somehow introduced a new error:
> 
> ```
>  Package tvheadend is missing dependencies for the following libraries:
> libFLAC.so.8
> libfdk-aac.so.2
> libsane.so.1
> ```
> 
> can you verify this?

I'm checking it. Will report here

svlobanov:
@neheb I double checked it and I can not reproduce the issue.

I rebuilt tvheadend on ubuntu20.04 for aarch64 target before my and after my commit and got likely the same binary (the only difference is version string and configure string (added --platform=linux))

Also, I don't see such errors in CI output. How did you get these errors? Can you rebuild with your .config before and after this commit? If this commit introduces this error, can you provide your .config?

neheb:
I got it from https://github.com/openwrt/packages/pull/17709

svlobanov:
@neheb 

> I got it from #17709

Ok, let's revert this change and restart CI for https://github.com/openwrt/packages/pull/17709

If it helps, then I will investigate the issue

neheb:
This PR may not even be the issue. It might be some configuration thing.

PR #17724, opened 2022-01-29
sane-backends: fix build on macos

svlobanov:
MacOS does not have /bin/true, but OpenWrt Makefile uses it so
build fails on macos

This patch removes absolute path (replaces /bin/true to true)

Maintainer: @luizluca 

Description: see above


luizluca:
LGTM. Thanks!

PR #17725, opened 2022-01-29
CircleCI: Add 21.02 GPG and usign public keys and update to Debian 10

BKPepe:
Maintainer: @champtar , @thess 

- Debian 9 (stretch) comes with Python 3.5 by default and End of Life is going to be
on 30th June 2022, let's update it to Debian 10 (buster).

Fixes issues in CircleCI:

1) OpenWrt master wants to use an equal or newer version than Python 3.6
```
Build dependency: Please install Python >= 3.6
Build dependency: Please install Python >= 3.6
Build dependency: Please install the Python3 distutils module

/home/build/build_dir/include/prereq.mk:9: recipe for target 'prereq' failed
```

That was fixed by updating it to Debian 10

2) It adds rsync as dependency to the Docker image required in the ``config.yml`` as there it is using: ``rsync -av "$SDK_HOST::downloads/$SDK_PATH/$SDK_FILE" .
sha256sum -c --ignore-missing sha256sums``

```
/bin/bash: line 23: rsync: command not found
```

It requires to tag a new release as Docker image on Docker hub by the following readme.

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:9
+FROM debian:10
 
 
 # Configuration version history
@@ -8,6 +8,7 @@ FROM debian:9
 # v1.0.3 - Verify usign signatures
 # v1.0.4 - Add support for Python3
 # v1.0.5 - Add 19.07 public keys, verify keys
+# v1.0.6 - Add 21.02 public keys, update Debian image to version 10, add rsync
 
 RUN apt update && apt install -y \
 build-essential \
@@ -23,6 +24,7 @@ python \
 python3 \
 signify-openbsd \
 subversion \
+rsync \
 time \
 unzip \
 wget \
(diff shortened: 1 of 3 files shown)


neheb:
Strange. Maybe newer python is available with debian-backports for 9.

Reminds me that I have a patch for MPD included (rejected by upstream) that fixes compilation under debian 9.

champtar:
Sorry for the delay, new container pushed
We sadly can't had more users to the free docker org, we have 5 and the limit is now 3
We could switch to anything else, maybe github registry so it's easier to give rights (never tried it)

BKPepe:
@neheb Yes, that's possible, but I don't want to fiddle with adding backports, doing almost insane things, when Debian 9 LTS support should end within ~4 months, so updating it to Debian 10 sounds like a good move to do it.

@champtar, Don't worry about it. Circle CI popped up to me once I submitted one PR until the Github Actions superseded somehow, and during that time I looked why the Circle CI failed.

BKPepe:
@champtar I just noticed that you pushed the tag ``latest`` instead of ``v1.0.6``. Shall I change this to use [the ``latest`` tag](https://hub.docker.com/r/openwrtorg/packages-cci/tags) in new PR?

champtar:
I followed the README without putting my brain on :)
Only ran `docker push docker.io/openwrtorg/packages-cci` and not `docker push docker.io/openwrtorg/packages-cci:1.0.6`
Now fixed

BKPepe:
Ah, that's ok. It is not even worth mentioning. I'm glad that I could help!

champtar:
I've used podman and the image shows a bit differently, maybe not multiarch, tell me if it causes any issue
image

BKPepe:
It seems to be alright [1], thanks! I need to figure out why it does not detect new/modified packages.

[1] https://app.circleci.com/pipelines/github/BKPepe/packages/463/workflows/975abab7-2e11-483a-8e0f-c7a840427e5c/jobs/245

PR #17729, opened 2022-01-29
libwebsockets: fix compilation with newer GCC

neheb:
Errors on uninitialized variable. No idea why only on powerpc64.

Maintainer: @karlp 
Compile tested: powerpc64

ping @stintel


karlp:
has this been filed upstream?  Otherwise looks fine, certainly harmless.  disabling werror is probably saner for downstream builds, but whatever

neheb:
Probably. I didn't look into disabling Werror as this is a really simple fix.

Still interesting that I only got this error with @stintel 's special powerpc64 config.

neheb:
Anyway, upstream PR: https://github.com/warmcat/libwebsockets/pull/2551

neheb:
Upstream fixed it differently. Backported.

PR #17736, opened 2022-01-30
nano: Add a plus variant with more features

hnyman:
Nano is by default built as "tiny" with most features disabled.

Add a new nano-plus variant that enables selected additional features in the build config: 
multiple files (multibuffer), justify, nanorc, help. 
Also some key bindings get enabled as the "tiny" configure option is removed.

Submitted as PR for CI testing...

Maintainer: me 
Run tested: ipq806x/R7800

### DIFF
--- a/utils/nano/Makefile
+++ b/utils/nano/Makefile
@@ -25,13 +25,25 @@ PKG_BUILD_PARALLEL:=1
 
 include $(INCLUDE_DIR)/package.mk
 
-define Package/nano
+define Package/nano/Default
   SUBMENU:=Editors
   SECTION:=utils
   CATEGORY:=Utilities
-  TITLE:=GNU nano - enhanced clone of the Pico text editor
   URL:=https://www.nano-editor.org/
   DEPENDS:=+libncurses
+endef
+
+define Package/nano
+  $(call Package/nano/Default)
+  TITLE:=GNU nano text editor (minimal features)
+  VARIANT:=tiny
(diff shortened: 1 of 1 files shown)


utils/nano/Makefile
```diff
@@ -22,18 +22,31 @@ PKG_CPE_ID:=cpe:/a:gnu:nano
 
 PKG_INSTALL:=1
 PKG_BUILD_PARALLEL:=1
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(BUILD_VARIANT)/$(PKG_NAME)-$(PKG_VERSION)
```
neheb, line 25: wasn't there an upstream commit that already does this?

hnyman: you are apparently right. Since 2019 there is a default line for that purpose. 
I will modify and test.

alexeys85: Is it possible to include unicode support in that plus variant?

alexeys85:
Is it possible to include unicode support in that plus variant?

hnyman:
It might be, but that would then increase the installed size quite a lot, I guess. (including the possible add-on packages)

hnyman:
I tested, and Unicode/UTF-8 support only adds 2 kB to the plus variant, as we already provide libncursesw and libncurses from the same file.

I added it to the plus variant.

PR #17738, opened 2022-01-30
knxd: fix build on macos

svlobanov:
knxd compilation fails on macos due to clang does not support
exit() builtin function that is used to detect build cc

This commit adds a patch to fix this issue (replaces `exit 0` by
`return 0` in conftest.c)

Maintainer: @tru7 

Description: see above

### DIFF
--- a/net/knxd/Makefile
+++ b/net/knxd/Makefile
@@ -12,7 +12,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=knxd
 PKG_VERSION:=0.14.53
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


neheb:
how strange...

does including stdlib,h fix it?

neheb:
oh this is an upstream GMP fix. never mind.

svlobanov:
> how strange...
> 
> does including stdlib,h fix it?

JFYI: exit(0) works without stdlib.h in GCC because it is a GCC built-in function (something like non-standard extension). Yes, I just picked upstream(i.e. GMP) variant to fix this issue

neheb:
sounds like bad behavior from GCC

tru7:
we can apply this patch while waiting for upstream to release a new version - the suggested patch was applied upstream today. I'll track upstream release.

PR #17739, opened 2022-01-31
libowfat: fix build on macos

svlobanov:
libowfat target-build fails on macos due to using system `ar` and
`ranlib` tools not compatible with the objects generated by
OpenWrt GCC toolchain.

This patch specifies CROSS= make flag that is used as a prefix for
`ar` and `ranlib` tools.

This patch also specifies CCC= make flag due to CCC= has invalid
value after specifying CROSS= make flag (CCC=$(CROSS)$(CC))

Maintainer: @dangowrt 

Description: see above

### DIFF
--- a/libs/libowfat/Makefile
+++ b/libs/libowfat/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libowfat
 PKG_VERSION:=0.32
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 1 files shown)


neheb:
interesting that ARC is not failing anymore. Currently it fails with https://downloads.openwrt.org/snapshots/faillogs/arc_archs/packages/libowfat/compile.txt

svlobanov:
@neheb do you think proposed change somehow related to the error? This PR is not applied yet... so I think the error and my PR are not related to each other

neheb:
Beats me. Might as well merge since it looks like it fixes things.

PR #17744, opened 2022-01-31
mktorrent: fix bad format for 64bit platforms

neheb:
The Makefile tries and fails to run a binary that sets it properly. Just
hardcode it.

Maintainer: @dangowrt 
Compile tested: ppc64


utils/mktorrent/patches/010-format.patch
```diff
@@ -0,0 +1,20 @@
+--- a/GNUmakefile
```
BKPepe, line 1: Shouldn't this be send to upstream?

PR #17748, opened 2022-01-31
xinetd: fix bad printf formats

neheb:
Easier to use the proper C macro.

Maintainer: @feckert 
Compile tested: ppc64

### DIFF
--- a/net/xinetd/Makefile
+++ b/net/xinetd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=xinetd
 PKG_VERSION:=2.3.15
-PKG_RELEASE:=12
+PKG_RELEASE:=13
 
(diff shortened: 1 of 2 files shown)


feckert:
LGTM
Should we send this to the upstream project?

neheb:
Isn't it dead upstream?

BKPepe:
Still it should be send to upstream.

neheb:
Upstream's last update was in 2016 with pending PRs. I don't think there's a point in trying to upstream.

BKPepe:
Well, atleast while sending to upstream, it is clear who authored it, why the change is required and so on. Then any GNU/Linux distribution can use it or maybe dealing with it, so it is always good idea to send it there even they are not active. Anyone can review it.

neheb:
meh alright: https://github.com/xinetd-org/xinetd/pull/35

PR #17749, opened 2022-01-31
mtdev: fix bad printf formats

neheb:
Easiest to just change the type.

Maintainer: @dangowrt 
Compile tested: ppc64


dangowrt:
Should we send this upstream as well?

neheb:
upstream seems dead.

PR #17758, opened 2022-01-31
python-psutil: update to version 5.9.0, fix build on macos

svlobanov:
1. updated to 5.9.0

2. psutil can not be built on macos due to build script detects Darwin
using sys.platform and changes build logic to build for Darwin, but
OpenWrt is Linux.
This commit add patch to allow redefining sys.platform and uses
env var TARGET_SYS_PLATFORM to specify linux as sys platfrom.

Maintainer: @ja-pa 
Compile tested: (armvirt/64, OpenWrt version r18681-5d110c0161)
Run tested: (armvirt/64, OpenWrt version r18681-5d110c0161, tests done)

```
root@OpenWrt:/# uname -a
Linux OpenWrt 5.10.92 #0 SMP Mon Jan 31 00:10:42 2022 aarch64 GNU/Linux
root@OpenWrt:/# cat /etc/openwrt_release 
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='SNAPSHOT'
DISTRIB_REVISION='r18681-5d110c0161'
DISTRIB_TARGET='armvirt/64'
DISTRIB_ARCH='aarch64_cortex-a53'
DISTRIB_DESCRIPTION='OpenWrt SNAPSHOT r18681-5d110c0161'
DISTRIB_TAINTS='no-all'
root@OpenWrt:/# python
Python 3.10.2 (main, Jan 31 2022, 00:10:42) [GCC 11.2.0] on linux
>>> import psutil
>>> print(psutil.__version__)
5.9.0
>>> psutil.cpu_times()
scputimes(user=0.93, nice=0.0, system=0.95, idle=501.24, iowait=0.0, irq=0.0, softirq=0.0, steal=0.0, guest=0.0, guest_nice=0.0)
>>> psutil.cpu_count()
2
>>> psutil.cpu_stats()
scpustats(ctx_switches=26232, interrupts=16379, soft_interrupts=19950, syscalls=0)
>>> 
```

PR to upstream: https://github.com/giampaolo/psutil/pull/2068/

### DIFF
--- a/lang/python/python-psutil/Makefile
+++ b/lang/python/python-psutil/Makefile
@@ -8,11 +8,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-psutil
-PKG_VERSION:=5.8.0
+PKG_VERSION:=5.9.0
 PKG_RELEASE:=1
 
 PYPI_NAME:=psutil
-PKG_HASH:=0c9ccb99ab76025f2f0bbecf341d4656e9c1351db8cc8a03ccd62e318ab4b5c6
+PKG_HASH:=869842dbd66bb80c3217158e629d6fceaecc3a3166d3d1faee515b05dd26ca25
 
 PKG_MAINTAINER:=Jan Pavlinec <jan.pavlinec1@gmail.com>
 PKG_LICENSE:=BSD 3-Clause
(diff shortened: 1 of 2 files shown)


jefferyto:
Apologies for the late comments, but I wanted to add:

* I think it would have been preferable to patch setup.py instead of psutil/_common.py, so that the changes affect only the build process and not the final package.
* Cross-compilation is not really supported by the Python interpreter/compiler and so I doubt many Python libraries would consider it a valid use case. While I think it's good to submit patches upstream in general, for Python libraries I don't hold much hope that they will be accepted.

PR #17763, opened 2022-01-31
chaosvpn: fix build on macos

svlobanov:
chaosvpn Makefile detects Darwin (macos) and changes compilation
flags for macos target, but OpenWrt is always Linux so build fails.
This patch redefines OS=Linux to use Linux compilation flags.

Maintainer: Norbert Summer <git@o-g.at> (I can't find github nickname)
Compile tested: (armvirt/64, OpenWrt version b21bc3479d46e6a4c3cc6bf7c245d4b0ddccb7db)

Description: see above


neheb:
maintainer is @TuxCoder

PR #6274, opened 2018-06-11
net/bonding: added protohandler package for netifd

he-ma:
Maintainer: me
Compile tested: only a shell script
Run tested: mips

Description:
This is a first version of a netifd proto handler which handles Link Aggregation (Channel Bonding)

PR #17768, opened 2022-02-01
ipmitool: add config options for ipmi kernel modules

he-ma:
To use the IPMI system interface some kernel modules are required.

This PR adds a configuration option to select the required kernel modules if the use of the IPMI system interface is desirable.

This package depends on PR in another repo ([openwrt/packages#6274](https://github.com/openwrt/openwrt/pull/5020))

Signed-off-by: Helge Mader <ma@dev.tdt.de>

### DIFF
--- a/admin/ipmitool/Config.in
+++ b/admin/ipmitool/Config.in
@@ -1,0 +1,15 @@
+if PACKAGE_ipmitool
+
+config IPMITOOL_KERNEL_MODULES
+	bool "Select system interface kernel modules for ipmitool"
+	default n
+	select PACKAGE_kmod-ipmi-handler
+	select PACKAGE_kmod-ipmi-si
+	select PACKAGE_kmod-ipmi-device-interface
+	help
+	  If not set, no kernel modules get installed on ipmitool package
+	  installation. The package ipmitool needs this kernel modules to
+	  work properly with the system interface.
+	  If in doubt, say "Y".
+
+endif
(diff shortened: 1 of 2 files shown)


he-ma:
@lynxis 
Would you kindly review this PR and merge it.

lynxis:
Hi @he-ma  

I'm sorry for the late reply.
The ipmitool is also quite usable without the kmod packages (e.g. small systems).
On debian those kernel modules would be an recommended dependency. Config options are usually meant for compile-time options and bitrot if not used regulary. The ipmitool binary isn't different as with your new Config option?
I would also try to avoid using Config option and prefer separate packages (e.g. wpad-basic).
Maybe you could create a meta package as ipmitool-kmods? I'm unsure what's the best solution for the dependencies in general and how we should fix those dependency. E.g. same for cryptsetup and all the different kmod-crypto-zoo modules.

Atm I would prefer not to merge your PR.

he-ma:
Hi @lynxis
Sorry for the late response but I have been quite busy over the last days.

Regarding your comment I absolutely agree with you.
The changes I wanted to make were only thought to "advise" an user (e.g. in the `make menuconfig`) in a simple way that it would be possible to add the kernel modules "in an easy way". Indeed there is absolutely no requirement to do any changes within your package.
Maybe the simplest solution is just to close this PR and have the user himself decide whether to build/include the kernel modules or not.

What do you think about it?

PR #17771, opened 2022-02-01
mezoura: Add new package

linosgian:
`mezoura` uses `tc` subsystem eBPF hook ([direct-action](https://man7.org/linux/man-pages/man8/tc-bpf.8.html)).
It's attached as a TC classifier for both ingress and egress.
The eBPF program is used to measure traffic (bandwidth and packets per
second) from/to the internet (non-private addresses).

It utilizes `uci` to parse static entries for DHCP (/etc/config/dhcp)
to map MAC addresses -> hostnames.

For each new IP/MAC address pair seen on the network, it attempts to resolve the
MAC address to a hostname. If there's no mapping, the entry is skipped.

It works for both v4 and v6 without having to maintain some weird
iptables rules to count packets/bytes as seen in other packages.

Lastly, it exposes a prometheus exporter for a widely used method to
consume the metrics held in the eBPF maps.

There's an option to disable exposing the embedded prometheus exporter.
The exporter component, flushes the metrics to `/tmp/prom_c`.
From that point, `prometheus-node-exporter-lua-textfile` can be used to
consume the metrics instead of exposing yet another service.

Maintainer: me
Run tested: `ramips_mt7621/xiaomi_r3g`


linosgian:
It seems like there's some issue with my package depending on `bpf-headers` from `base/kernel`.
Not sure whether or not it has to do with how the CI/Actions are configured or my package is misconfigured.

I've built this package locally, on openwrt's master branch, but it seems to fail on the CI tests here.

svlobanov:
I'm not sure, but maybe your issue somehow related to https://github.com/openwrt/openwrt/pull/5011 . CC: @aparcar

utils/mezoura/Makefile
```diff
@@ -0,0 +1,56 @@
...
+PKG_NAME:=mezoura
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_DATE:=2022-02-01
+PKG_SOURCE_URL:=https://github.com/linosgian/mezoura
+PKG_SOURCE_VERSION:=ae8969d207f9f68535cc9f5cabdf8345226352e2
+PKG_MIRROR_HASH:=3cd82bf5cf5754aaffbb7656fc3e0c525ac925f42a70eb23aa3004d1e2542096
+PKG_RELEASE:=1
+
+PKG_LICENSE:=GPL-2.0
+PKG_MAINTAINER:=Linos Giannopoulos <linosgian00@gmail.com>
```
BKPepe, line 20: This order is not good. Please take a look how it is done in other Makefiles.

1. Also, you can fix the deprecated SPDX License Identifier

utils/mezoura/Makefile
```diff
@@ -0,0 +1,56 @@
+#
+# Copyright (C) 2022 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
```
BKPepe, line 7: https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md

> Provide an up-to-date Copyright notice or none. Copyright should not be assigned to OpenWrt unless you are explicitly requested by or working under contract to OpenWrt. Assigning a Copyright to yourself or organization you represent is acceptable.

BKPepe (changes requested):
I wonder if we want to merge this to this repository since it seems that this is your project, there is no website/documentation, and there is only an initial commit done 15 hours ago. That leads to me question how many users will use this package. Because this package is going to be compiled for all targets in OpenWrt. It will take resources, storage, etc.

linosgian:
> I wonder if we want to merge this to this repository since it seems that this is your project, there is no website/documentation, and there is only an initial commit done 15 hours ago. That leads to me question how many users will use this package. Because this package is going to be compiled for all targets in OpenWrt. It will take resources, storage, etc.

I was thinking about that as well, but it seemed to me like a useful package for others that want to have bandwidth monitoring for their setups. I've been using this for a while, that's why I decided to open a PR to merge this. I've added a README to the repository and I'm planning on writing about this in the forum, once this is merged and people can use it (those that are on the master branch at least. Since 21.02 does not contain `bpf-headers`).

utils/mezoura/Makefile
linosgian: pushed a fixup for this. Let me know if I got it right

utils/mezoura/Makefile
linosgian: ditto

linosgian:
After @nbd168's latest commit https://github.com/openwrt/openwrt/commit/8c1d5129ee8464b3a240d73c48eb2d166537c1a8 on master, all (but one) builds aren't failing anymore. @BKPepe Is there anything else I should to get this merged?

The last build (arc_archs) is failing due to `stdarg.h` missing. I'm not sure how to debug this tbh

BKPepe:
With arc, I think @neheb might want to help you

linosgian:
> The last build (arc_archs) is failing due to `stdarg.h` missing. I'm not sure how to debug this tbh

Hey @neheb could you shed some light on this? I am at loss on how to debug this :/

neheb:
sounds like a problem of CFLAGS not being passed.

neheb:
Meh I'd ignore it. I doubt eBPF is supported under ARC anyway.

GeorgeSapkin:
@linosgian is this still relevant? There are no updates to either the PR or the underlying project in over 3 years.

GeorgeSapkin:
No updates in over three years to either the project or the PR. Project doesn't seem to have any prominence. Closing as stale. Feel free to reopen as necessary.

PR #17790, opened 2022-02-04
adguardhome: Adjust start/stop values of init script

jamesmacwhite:
Maintainer: @alaviss @dobo90
Compile tested: N/A
Run tested: Linksys WRT3200ACM armv7 19.07.8

Description:

The current startup value used by the AdGuard Home procd init script does not allow AdGuard Home to start on boot due to a race condition, possibly due to a certain service or dependency not being available at the time it tries to start. Running `service adguardhome start` when the router has booted and SSH is available works fine. I have personally experienced this but thought it might be specific to my setup, however the user report below suggests it can happen to others.

[A user reported this on the forum recently](https://forum.openwrt.org/t/adguardhome-ipk-autostart-not-working-after-reboot/119274/27) and changing the start value resolves the issue from a cold power on or reboot.

[Another forum member highlighted](https://forum.openwrt.org/t/adguardhome-ipk-autostart-not-working-after-reboot/119274/25) that the OpenWrt init script bundled by the AdGuard Home developers uses different values in the procd init script, therefore it makes sense that the opkg package matches this as well.

This should resolve AdGuard Home not starting despite being set to autostart, by tweaking the procd init start/stop values accordingly.


net/adguardhome/files/adguardhome.init
```diff
@@ -4,10 +4,9 @@ PROG=/usr/bin/AdGuardHome
 
 USE_PROCD=1
 
-# starts after network starts
-START=21
-# stops before networking stops
-STOP=89
+# Matches values used by the official AdGuardHome procd init script
+START=95
```
alaviss, line 8: 95 might be a bit late, since this provides the basic service of DNS.

From the forum post it seems that starting right after network doesn't work because network might still be busy bringing up interfaces so AGH couldn't attach to them (in case you use a specific subnet like `192.168.1.1` instead of `0.0.0.0`).

Maybe delaying until 50 is a bit better here to ensure that network fully started but not too late that other network-related services have trouble working (9x range seems to be filled with applications from my cursory look at this repo). This number is taken from [uhttpd](https://github.com/openwrt/openwrt/blob/8c1d5129ee8464b3a240d73c48eb2d166537c1a8/package/network/services/uhttpd/files/uhttpd.init#L4) of which I assume also have the requirement of waiting for network to fully activate.

net/adguardhome/files/adguardhome.init
```diff
@@ -4,10 +4,9 @@ PROG=/usr/bin/AdGuardHome
...
+STOP=1
```
alaviss, line 9: I think we should keep the old stop value. Stopping too early means that if any service send a message over the net on exit for example, there wouldn't be DNS for it to work.

alaviss (commented):
> [Another forum member highlighted](https://forum.openwrt.org/t/adguardhome-ipk-autostart-not-working-after-reboot/119274/25) that the OpenWrt init script bundled by the AdGuard Home developers uses different values in the procd init script, therefore it makes sense that the opkg package matches this as well.

I personally don't think that it's a good idea to match AGH's `START`/`STOP`, since these values are essentially hacks for the lack of dependencies in the init system and would have to be tailored to the conditions of other OpenWRT's init scripts.

I'm not an expert in init scripts though, so take these comments with a grain of salt.

net/adguardhome/files/adguardhome.init
jamesmacwhite: That seems reasonable. It seems to be that AdGuard Home can't start just after network, likely as you say due to interfaces not being available potentially.

jamesmacwhite:
@alaviss Updated the start/stop values per your suggestions.

Beginner-Go:
@jamesmacwhite Hi, Can `PKG_RELEASE` switch to `$(AUTORELESE)` ?

jamesmacwhite:
@Beginner-Go That might be best to be directed to the main maintainers of the package @dobo90 and @alaviss. I know the AdGuard Home team also recently implemented a change in [their build process with SOURCE_DATE_EPOCH](https://github.com/AdguardTeam/AdGuardHome/issues/4221)

jamesmacwhite:
@alaviss Do you think we can go with the new start value for the init.d script and merge this? Currently it's possible for adguardhome to fail on start up leaving a router with no DNS (depending on certain edge cases).

alaviss (approved):

jamesmacwhite:
@alaviss I've just recently moved to 21.02.2 and found that START=50 might still be too early, from testing as I've had to do several reboots while reconfiguring my network interfaces, we may have to bump it to 95, to be absolutely sure that all interfaces and network parameters are present for adguardhome to start successfully, even with setting it to 50, I've found adguardhome is not starting on boot.

alaviss:
I wonder if it is possible to wait for network to fully initialize then. I'm not really a fan of bumping magic numbers until things work.

jamesmacwhite:
@alaviss Yes, I've found what the issue is in my case, I'm using the ipset functionality and I found this on a startup:

> Sat Feb 19 15:13:04 2022 daemon.err AdGuardHome[5156]: 2022/02/19 15:13:04.096103 [fatal] dnsServer.Prepare: initializing ipset: ipset: getting ipsets from config line at idx 0: querying ipset "wanb": netfilter query: netlink receive: no such file or directory

That was with START=95 set, so it seems ipset isn't ready when AdGuardHome attempts to start. I am initialising ipset chains in my /etc/rc.local, but it would seem this gets run after AdGuardHome attempts to start for the first time.

I don't believe there is a way to make AdGuardHome initialise an ipset chain itself before hand. I could move my ipset commands to a custom initscript that's before AdGuardHome using start values potentially.

jamesmacwhite:
I assume this is the reason why S99 was used by the AdGuard Home developers in their init script, they potentially knew this may be a problem and moved it right to the end of the boot process to avoid it, but it is a bit of a hack and leaves the router without DNS for much longer than really needed. 

I'm wondering if by moving my ipset initialisation commands to a init script that's using something like S18, AdGuard Home might be able to boot under it's original init script value (21) without this change at all. I do wonder however if there is a possibility other network conditions may cause similar issues. e.g. if AdGuard Home has a bind to an IP address of which the network interface is not up yet.

So, summary, maybe this init start is actually needed for my issue, but I'm sure other race conditions may be possible.

WildByDesign:
One of the original issues prompting the AdGuard Dev team to go with 99 had something to do with sysntpd, if I recall correctly. It was causing problems with AGH encryption settings for OpenWrt users at the time. The certificates weren’t getting the correct date/time.

sysntpd is 98. We need a successful startup at boot for all AGH users on OpenWrt. I had always installed AGH as a service directly from their binary for a couple of years and always had 100% success at boot.

jamesmacwhite:
Ah yes, NTP will be another one. There is a potential workaround to that without going long on the service start value. something like this in your AdGuard Home configuration:

```
[/pool.ntp.org/]1.1.1.1
[/pool.ntp.org/]1.0.0.1
[/pool.ntp.org/]2606:4700:4700::1111
[/pool.ntp.org/]2606:4700:4700::1001
```

The only downside this bypasses any encrypted DNS, but having NTP requests being plain DNS, probably something you can live with.

In reality there's probably many edge cases, which is why it's S99 to avoid them all. The downside, it leaves your router without DNS resolution for longer than maybe you'd like, given dnsmasq starts way sooner.

zhanghua000:
Sorry to bother this closed pull request but I think maybe I find a better solution.
I have also experienced this problem while there is not any ipset chain, the log shows:
```
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.418900 [info] AdGuard Home, version v0.107.25
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.420762 [info] AdGuard Home updates are disabled
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.491593 [info] tls: using default ciphers
Tue Mar 21 21:43:32 2023 daemon.notice netifd: wan (3285): udhcpc: broadcasting select for 192.168.0.165, server 192.168.0.1
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.589270 [info] Initializing auth module: /tmp/adguardhome/data/sessions.db
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.591477 [info] auth: initialized.  users:1  sessions:0
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.592125 [info] web: initializing
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.607723 [info] dnsproxy: cache: enabled, size 4096 b
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.607990 [info] MaxGoroutines is set to 300
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.613732 [info] AdGuard Home is available at the following addresses:
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.614719 [info] go to http://192.168.8.1:3001
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.631717 [info] stopped
Tue Mar 21 21:43:32 2023 daemon.err AdGuardHome[2645]: 2023/03/21 13:43:32.633517 [fatal] listen tcp 192.168.8.1:3001: bind: cannot assign requested address
Tue Mar 21 21:43:32 2023 daemon.notice netifd: wan (3285): udhcpc: lease of 192.168.0.165 obtained from 192.168.0.1, lease time 86400
Tue Mar 21 21:43:32 2023 daemon.notice hostapd: Configuration file: /var/run/hostapd-phy0.conf (phy phy0-ap0) --> new PHY
Tue Mar 21 21:43:32 2023 kern.err kernel: [   31.641396] ksmbd: Failed to bind socket: -13
Tue Mar 21 21:43:32 2023 kern.err kernel: [   31.645940] ksmbd: Failed to shutdown socket: -134
Tue Mar 21 21:43:32 2023 kern.info kernel: [   31.657855] br-lan: port 4(phy0-ap0) entered blocking state
Tue Mar 21 21:43:32 2023 kern.info kernel: [   31.663547] br-lan: port 4(phy0-ap0) entered disabled state
Tue Mar 21 21:43:32 2023 kern.info kernel: [   31.670286] device phy0-ap0 entered promiscuous mode
Tue Mar 21 21:43:32 2023 kern.info kernel: [   31.675977] br-lan: port 4(phy0-ap0) entered blocking state
Tue Mar 21 21:43:32 2023 kern.info kernel: [   31.681680] br-lan: port 4(phy0-ap0) entered forwarding state
Tue Mar 21 21:43:32 2023 kern.info kernel: [   31.688123] IPv6: ADDRCONF(NETDEV_CHANGE): br-lan: link becomes ready
Tue Mar 21 21:43:32 2023 daemon.notice hostapd: phy0-ap0: interface state UNINITIALIZED->COUNTRY_UPDATE
Tue Mar 21 21:43:32 2023 daemon.notice netifd: bridge 'br-lan' link is up
Tue Mar 21 21:43:32 2023 daemon.notice netifd: Interface 'lan' has link connectivity
Tue Mar 21 21:43:32 2023 daemon.notice netifd: Interface 'lan' is setting up now
Tue Mar 21 21:43:32 2023 daemon.notice netifd: Interface 'lan' is now up
```
It seems that AdGuardHome starts before the network is ready on my Router. 
I was also agree with adjusting START/STOP value at first, however, I noticed that ksmbd has similar situation(failed to bind socket) while it works well at last. After checked its init script, I find that it added some triggers to restart itself when network changes. So I think maybe we can also modify AdGuardHome's init script to let procd restart it when network changes.

WildByDesign:
I have also been experiencing a bunch of these errors in the past few weeks where AdGuardHome failed to start. Same thing.

I ended up having to change the start back to 95 for now since it works every time.

But modifying init script to check for network sounds like a more ideal fix.

jamesmacwhite:
The issue with having AdGuardHome being set to 95 is that's it's very late in the boot process and DNS should ideally be available much earlier for other services. Comparing against dnsmasq its around S20, so it's quite a difference.

The issue is undoubtedly race conditions or specific scenarios which rely on network related aspects being available which aren't at that time. In the original case, it was because ipset chains were being referenced before they existed which is one example of a such a network condition. So the solution was a custom init script inserted before AdGuardHome that makes sure ipset chains required are always initialised in the sequence order.

Be careful when having AdGuardHome set to start at 95, because your router will not have DNS for most of the start up. It's less of an issue if AdGuardHome is not the primary DNS resolver, but in many cases it will be as it's intends to replace dnsmasq for full functionality, this has impact for other services relying on DNS as well.

For those using the opkg if you upgrade the adguardhome package, the changes to the init script will get overwritten when opkg overwrites files. For those experiencing race conditions a custom init script to ensure whatever requirement is present before AdGuardHome starts is likely best. You can insert any custom init script using: https://openwrt.org/docs/techref/initscripts

We could potentially bump the start value a bit, but I don't think the OpenWrt devs will accept setting the value to 95. AdGuardHome devs ship a different init script, but it's not necessarily the right thing to do either.

WildByDesign:
Thank you so much for your follow up and especially all of the great detail. I appreciate it.

I had done some testing over the past few weeks after noticing these errors recently. I tried S50 and also S55. Both worked the majority of the time, but did occasionally fail to start.

So the choice for me was S95 for 100% success at startup with a delayed DNS startup, or S55 with occasional startup failures and therefore potentially no Internet for my family at times when I may not be home to manually start the adguardhome service.

My setup has always been having dnsmasq disabled and having adguardhome claiming port 53 and therefore all DNS on my network. This has always worked great me for and is quite a simplistic setup.

I honestly am not familiar with Linux init scripts and such. It would be great if adguardhome could, for example, attempt to start again after say 20 seconds after service startup failure. Or wait for network to be ready entirely before startup.

Do you know if adguardhome `verbose` setting (debug verbose output) for log file would give more specific details on these network failures causing adguardhome service to fail to start?

jamesmacwhite:
Yes AdGuardHome does have a verbose flag, which can be set in the yaml config or when run from command line. For OpenWrt specifically, there is `logread -e AdGuardHome` but running something like this `AdGuardHome -v -c /etc/adguardhome.yaml -w /var/adguardhome --no-check-update`, will provide verbose logs directly to shell/console.

You can however set a verbose flag within the yaml so it always outputs verbose information, useful if you are debugging startup issues.

99.9% of the time it's likely some network related condition which fails and AdGuardHome doesn't start, then if you run `service adguardhome start` once SSH is available, it will then usually startup fine because whatever network condition wasn't present when it first tried, is now present.

To detect this, would need a bit of rework of the init script shipped in the opkg version, but not impossible, however to always ensure more bespoke or custom setups don't have race conditions with certain network related areas, custom init scripts are the way to go, inserted before the startup value in the shipped init script. That way you always ensure you have satisfied requirements. You can also add any custom init script to the backup file list to preserve across firmware upgrades/flashing and even with your own image build, automatically have it set to start.

init scripts aren't too complex, this is the basic outline you need

```
#!/bin/sh /etc/rc.common

START=18

start() {
  logger -s "Custom init.d script stuff"
  ...
}

Here you have a init script inserted at S18, just before AdGuardHome, this always ensures this is run first before AdGuardHome starts. You can add whatever you need here i.e. ipset initialisation.

Issue #17780, opened 2022-02-03
dbus: switch from cmake to autotools for packaging


arkanoid87:
Maintainer: @feckert 
Environment: mips_24kc, gl-x300b (all *big endian* devices affected), snapshot

Description:

while debugging a problem on ModemManager not booting on *big endian* platforms, it ended up as a build issue with `dbus` that does not define `WORDS_BIGENDIAN` when built with cmake instead of autotools

The current dbus build writes and reads malformed messages. This causes correctly formed messages (like the ones ModemManager writes using glib2) to not be accepted by dbus.

I've [opened an issue on dbus bug tracker,](https://gitlab.freedesktop.org/dbus/dbus/-/issues/375) and they suggested to use autotools instead of cmake, as AC_C_BIGENDIAN Autoconf macro handles this.

In the meanwhile just for a quick test I've been suggested tro try adding `TARGET_CFLAGS += $(if $(CONFIG_BIG_ENDIAN),-DWORDS_BIGENDIAN)` in current cmake Makefile, and that fixed the problem.

Please follow the whole story here:
https://github.com/openwrt/packages/issues/17284
https://gitlab.freedesktop.org/dbus/dbus/-/issues/375

aleksander0m:
> I've [opened an issue on dbus bug tracker,](https://gitlab.freedesktop.org/dbus/dbus/-/issues/375) and they suggested to use autotools instead of cmake, as AC_C_BIGENDIAN Autoconf macro handles this.
> 
> In the meanwhile just for a quick test I've been suggested tro try adding TARGET_CFLAGS += $(if $(CONFIG_BIG_ENDIAN),-DWORDS_BIGENDIAN) in current cmake Makefile, and that fixed the problem.

Worth noting that upstream stated that the cmake port was targeting Windows setups. I would definitely switch to autotools instead of adding the `TARGET_CFLAGS` workaround, which has all the proper bits and pieces to detect BE systems. Also, who knows what else is not covered by the cmake port :/

arkanoid87:
In the meanwhike, the ticket https://gitlab.freedesktop.org/dbus/dbus/-/issues/375
Changed title in "cmake build effectively assumes all platforms are little-endian"

feckert:
So that we can fix this as soon as possible.
Because this also effects the openwrt-21.02 stable branch.
I think we have to switch from cmake to autotools for the modemmanager packages.
How are we going to do that?
[...]

PR #17792, opened 2022-02-04
dbus: switching back to autotools to fix big endian issue

feckert:
Maintainer: @neheb 
Compile tested: x86_64(little endian) lantiq_xrx200(big endian), latest openwrt master
Run tested: x86_64(little endian) lantiq_xrx200(big endian), latest openwrt master

Description:
According to the upstream developers of dbus, cmake is only inteded for
windows. Unix like operating systems should continue to use the
autotools stack.

Since the changeover to cmake, the dbus was no longer usable on all big
endian systems because the name binding no longer worked.

In cmake, it is not evaluated whether it should run on a little endian
or big endian system.

This fixes issue #17780. The original problem occurred when using the
modem manager on the big endian system (MIPS).

Thanks to @arkanoid87 for for the accurate analysis

### DIFF
--- a/utils/dbus/Makefile
+++ b/utils/dbus/Makefile
@@ -19,11 +19,13 @@ PKG_MAINTAINER:=
 PKG_LICENSE:=AFL-2.1
 PKG_CPE_ID:=cpe:/a:freedesktop:dbus
 
+PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=1
+
 PKG_CONFIG_DEPENDS:= \
 	CONFIG_DBUS_VERBOSE
 
 include $(INCLUDE_DIR)/package.mk
-include $(INCLUDE_DIR)/cmake.mk
 
 define Package/dbus/Default
   SECTION:=utils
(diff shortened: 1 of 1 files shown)


neheb:
How unfortunate...

arkanoid87:
thank you!

utils/dbus/Makefile
```diff
@@ -78,27 +80,37 @@ define Package/dbus/config
...
+	--with-system-socket=/var/run/dbus/system_bus_socket \
+	--with-system-pid-file=/var/run/dbus.pid \
+	--with-dbus-user=root \
+	--without-x \
+	--enable-qt-help=no \
+	--disable-xml-docs
 
 ifeq ($(CONFIG_DBUS_VERBOSE),y)
-  CMAKE_OPTIONS += -DDBUS_ENABLE_VERBOSE_MODE=ON
+  CONFIGURE_ARGS += --disable-verbose-mode
```
smcv, line 113: This was probably meant to be `--enable-verbose-mode`?

smcv:
Thanks for swapping this back. I confirm that the upstream maintainers of dbus recommend building it with Autotools on all platforms except Windows.

We might switch from Autotools to Meson in a future release, but if we do, that would only happen in a new release series like 1.15.x; the intention is that 1.12.x will always be primarily Autotools and secondarily CMake.

utils/dbus/Makefile
feckert: I guess you're right. I'll have to look at that again. Thanks for the tip.

feckert: Fixed with commit https://github.com/openwrt/packages/commit/2cbe230b6a65fd00c04cd528f6e5d4651218e1f8 :+1:

PR #17801, opened 2022-02-04
bind: support compile-time exclusion of DNS-over-HTTPS support

nmeyerhans:
Maintainer: me
Compile tested: x86_64. Validated that the expected `configure` options were provided and that libnghttp2 was not linked when it was supposed to be disabled.
Run tested: x86_64. Basic named serving functionality inc. recursing resolution and authoritative service.

Description:

DoH is enabled by default, but disabling it removes the need to link against libnghttp2, which may be desirable more constrained environments. This change introduce a new Config option allowing support for DoH to be turned off if desired.

### DIFF
--- a/net/bind/Config.in
+++ b/net/bind/Config.in
@@ -20,3 +20,17 @@ config BIND_LIBXML2
 		format. Building with libjson support will require the
 		libxml2 package to be installed as well.
 endif
+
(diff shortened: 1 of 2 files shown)


pprindeville (approved):
LGTM

Issue #16684, opened 2021-09-22
lxc-create: requires getopts as a dep


graysky2:
Maintainer: @neheb 
It seems that `lxc-create` requires `getopt` as a dependency but it is not pulled down when the package is installed.

```
# lxc-create --name myLMS --template download
/usr/share/lxc/templates/lxc-download: line 241: getopt: not found
LXC container image downloader
...
lxc-create: myLMS: lxccontainer.c: create_run_template: 1616 Failed to create container from template
lxc-create: myLMS: tools/lxc_create.c: main: 319 Failed to create container myLMS
```

neheb:
I don't actually use or maintain LXC. Want to be a maintainer?

graysky2:
@neheb - My bad.  I saw you several times in the history of [Makefile](https://github.com/openwrt/packages/commits/master/utils/lxc/Makefile) plus your activity with it so assumed you were.  I see [Marko](https://github.com/openwrt/packages/blob/master/utils/lxc/Makefile#L19) hard-coded in the `Makefile` ... how can I find Marko's github nick?  Guess it is @robimarko

rmilecki:
I noticed this change while updating LXC in stable releases.

This change adds `getopt` as dependency of `lxc-create`. I see few problems here:
1. `lxc-create` requires more than just `getopt` to work
2. We already have `LXC_BUSYBOX_OPTIONS` which takes care of requirements (including `getopt`)
3. We ended up with
   * `lxc-create` selecting `getopt`
   * `LXC_BUSYBOX_OPTIONS` selecting `BUSYBOX_CONFIG_GETOPT`

PR #17817, opened 2022-02-06
lxc-create: update depends to include getopt

graysky2:
Fixes https://github.com/openwrt/packages/issues/16684
```
Build system: x86_64
Build-tested: bcm2711/RPi4B
Run-tested: bcm2711/RPi4B
```

Maintainer: @robimarko @BKPepe @cotequeiroz

### DIFF
--- a/utils/lxc/Makefile
+++ b/utils/lxc/Makefile
@@ -39,7 +39,7 @@ LXC_SCRIPTS += \
 
 DEPENDS_APPLETS = +libpthread +libcap +liblxc
 
-DEPENDS_create = +lxc-configs +lxc-hooks +lxc-templates +flock
+DEPENDS_create = +lxc-configs +lxc-hooks +lxc-templates +flock +getopt
 
 define Package/lxc/Default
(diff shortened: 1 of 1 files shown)


BKPepe:
- I see that luci-app-lxc depends on getopt. Reference: https://github.com/openwrt/luci/blob/4fc3cd8e42ff9245f028c521307fa45e6a0c76af/applications/luci-app-lxc/Makefile#L10

- Anyway, I removed getopt from my router, rebooted it and lxc-create still works for me. :thinking:  On the other hand, getopt is indeed used in LXC: https://github.com/lxc/lxc/blob/7e37cc96bb94175a8e351025d26cc35dc2d10543/templates/lxc-download.in#L150

So, in the end, approved. :+1:

BKPepe (approved):

graysky2:
Thanks, yes, some users do not have the luci-app installed (I do not) but as you referenced, the underlying code makes calls to it so it needs to be a depends.  Thanks.

PR #17794, opened 2022-02-04
ruby: update to 3.0.3

paper42:
Maintainer: @luizluca
Compile tested: aarch64, Turris MOX, OpenWrt 21.02
Run tested: aarch64, Turris MOX, OpenWrt 21.02

Description: fixes CVEs:
* [CVE-2021-41817: Regular Expression Denial of Service Vulnerability of Date Parsing Methods](https://www.ruby-lang.org/en/news/2021/11/15/date-parsing-method-regexp-dos-cve-2021-41817/)
* [CVE-2021-41816: Buffer Overrun in CGI.escape_html](https://www.ruby-lang.org/en/news/2021/11/24/buffer-overrun-in-cgi-escape_html-cve-2021-41816/)
* [CVE-2021-41819: Cookie Prefix Spoofing in CGI::Cookie.parse](https://www.ruby-lang.org/en/news/2021/11/24/cookie-prefix-spoofing-in-cgi-cookie-parse-cve-2021-41819/)

PR #17820, opened 2022-02-06
ruby: fix build on macos

svlobanov:
1. ruby/host build fails on macos due to Apple ld generates warning
if a folder from LDFLAGS is not exist. configure script catches this
warning and fails. This patch disables ld warnings for macos

2. ruby build fails on macos due /bin/true is not exist on macos.
This patch replaces /bin/true with true in OpenWrt Makefile

Maintainer: @luizluca 
Compile tested: (armvirt/64, OpenWrt 4d904524effc9eb0cc5094574c55d3a520803223)

Description: see above

### DIFF
--- a/lang/ruby/Makefile
+++ b/lang/ruby/Makefile
@@ -12,7 +12,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ruby
 PKG_VERSION:=3.0.3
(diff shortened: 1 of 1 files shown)


svlobanov:
CI issues are the same as for https://github.com/openwrt/packages/pull/17794#issuecomment-1030817073

PR #17841, opened 2022-02-09
libtalloc: fix build on macos

svlobanov:
This commit adds patch to fix the issue with 'echo -n' on macos

Maintainer: @thess 
Compile tested: (armvirt/64, OpenWrt 4d904524effc9eb0cc5094574c55d3a520803223)

Description: see above

### DIFF
--- a/libs/libtalloc/Makefile
+++ b/libs/libtalloc/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=talloc
 PKG_VERSION:=2.3.1
 MAJOR_VERSION:=2
-PKG_RELEASE:=3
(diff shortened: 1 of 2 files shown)


neheb:
interesting.

What's /bin/sh on macOS? Some POSIX shell?

svlobanov:
> interesting.
> 
> What's /bin/sh on macOS? Some POSIX shell?

Bash 3.2 (bigsur11.6)

neheb:
weird that it doesn't support echo -n. Even (d)ash supports that.

svlobanov:
> weird that it doesn't support echo -n. Even (d)ash supports that.

Actual bash versions support it(e.g. bash 5.1)

PR #17860, opened 2022-02-14
Update `coreutils-sort` location in adblock

jamesooo:
Maintainer: @dibdot 

Description:
after https://github.com/openwrt/packages/commit/df10e58a9c7c5ef88eae23c03e497d2f9ca9cedd the install location of `coreutils-*` plugins shifted to `/usr/libexec/` per https://github.com/openwrt/packages/pull/14257

This pull request updates the `adblock` package to use this new location.

### DIFF
--- a/net/adblock/files/adblock.sh
+++ b/net/adblock/files/adblock.sh
@@ -60,6 +60,7 @@ adb_action="${1:-"start"}"
 adb_packages=""
 adb_sources=""
 adb_cnt=""
+adb_sort="/usr/libexec/sort-coreutils"
 
 # load & check adblock environment
 #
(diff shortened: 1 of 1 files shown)


dibdot:
NAK from my side. Why should we replace the autodection via `command -v` with a hardcoded path?

jamesooo:
> NAK from my side. Why should we replace the autodection via `command -v` with a hardcoded path?

@dibdot The current result from the autodetect command is busybox ‘sort’ since in the pull request and commit I link to in the description coreutils has moved its installed bins to a new location (not on the default PATH and not named ‘sort’).

dibdot:
Then you've got a borked sort installation (most probably due to a busybox update via opkg). That's not a proper fix for that ...
Adblock has a dependency to coreutils-sort which creates a sort symlink in /usr/bin, e.g.:

```
root@2go-re650:~# command -v sort
/usr/bin/sort

root@2go-re650:~# ls -la /usr/bin/sort
lrwxrwxrwx    1 root     root            27 Feb 14 04:59 /usr/bin/sort -> /usr/libexec/sort-coreutils
``` 

Just fix your broken installation and re-install coreutils-sort.

PR #17861, opened 2022-02-14
opendoas: add package

paper42:
Maintainer: me
Compile tested: aarch64, Turris MOX, OpenWrt master
Run tested: aarch64, Turris MOX, OpenWrt master

Description: opendoas is a port of OpenBSD doas to Linux, it replaced sudo in the main alpine repository with 3.15.

if someone wants to test this, add this to `/etc/doas.conf`:
```
permit nopass <user> as root
```
and try running doas id -u as `<user>`

### DIFF
--- a/utils/opendoas/Makefile
+++ b/utils/opendoas/Makefile
@@ -1,0 +1,40 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=opendoas
+PKG_VERSION:=6.8.2
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=https://github.com/Duncaen/OpenDoas/releases/download/v$(PKG_VERSION)
+PKG_HASH:=4e98828056d6266bd8f2c93e6ecf12a63a71dbfd70a5ea99ccd4ab6d0745adf0
+
+PKG_MAINTAINER:=Michal Vasilek <michal.vasilek@nic.cz>
+PKG_LICENSE:=ISC, BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_INSTALL:=1
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/opendoas
+  SECTION:=utils
+  CATEGORY:=Utilities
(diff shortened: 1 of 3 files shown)


paper42:
Should this package go to utils/ or admin/?

neheb:
utils probably.

If this is a sudo replacement, wouldn't it make sense to add a symlink?

paper42:
> utils probably.

`sudo` is in admin which makes me a bit confused.

> If this is a sudo replacement, wouldn't it make sense to add a symlink?

It doesn't accept the same arguments, it's not just a sudo rewrite. Alpine has a wrapper script that parses sudo options and translates them to doas, but I don't think that should be added to this package. On OpenBSD and Void, many people don't use any wrappers and just write `doas <something>` or `alias sudo=doas`.

neheb:
wrapper script sounds interesting. I assume this can be handled with ALTERNATIVES.

I don't have a strong opinion on this though., I don't use non root users or sudo on OpenWrt.

paper42:
I added a package for the doas - sudo compatibility wrapper that alpine uses, but I am not sure how to use alternatives to avoid conflicts. Is there some documentation I missed?

I converted this PR to a draft for now because it's currently not in a mergeable state.

paper42:
I think I now know how alternatives work, but if we wanted to have doas-sudo-shim to replace sudo, /usr/bin/sudo would have to be a symlink to the real sudo. That sounds wrong and confusing to users to me, what do you think?
```
# opkg install ./doas-sudo-shim_0.1.1_arm_cortex-a9_vfpv3-d16.ipk 
Installing doas-sudo-shim (0.1.1) to root...
Configuring doas-sudo-shim.
Collected errors:
 * pkg_alternatives_update_path: /usr/bin/sudo exists but is not a symlink
```

paper42:
I dropped doas-sudo-shim from this PR

paper42:
I don't think the aarch64 CI failure is caused by this PR. I have been using this package on multiple routers since I created this PR and it seems to work fine.

neheb:
https://downloads.openwrt.org/snapshots/faillogs/x86_64/packages/opendoas/compile.txt

paper42:
> [downloads.openwrt.org/snapshots/faillogs/x86_64/packages/opendoas/compile.txt](https://downloads.openwrt.org/snapshots/faillogs/x86_64/packages/opendoas/compile.txt)

hopefully fixed in #19183

PR #17866, opened 2022-02-15
[21.02] ksmbd: update to 3.44

neheb:
Maintainer: @Andy2244 

ping @hauke


hauke:
@neheb  Did you test this? Are the dependency changes correct?

neheb:
Yes they are. Note that the dependencies are workarounds for inability to select various symbols. More info: https://github.com/openwrt/openwrt/pull/5060

Issue #17816, opened 2022-02-06
[21.02] tvheadend: does not execute first run


BKPepe:
Maintainer: @M95D
Environment: OpenWrt 21.02.1, [Seeedstudio mini router](https://www.seeedstudio.com/Dual-GbE-Carrier-Board-with-4GB-RAM-32GB-eMMC-RPi-CM4-Case-p-5029.html) (Raspberry Pi 4 CM, bcm27xx/aarch64_cortex-a7)

Description:

Once I started tvheadend for the first time, it was asking me for credentials, which I dont have.
```
root@OpenWrt:~# opkg install tvheadend
Installing tvheadend (4.2.8-5) to root...
Downloading https://downloads.openwrt.org/releases/21.02.1/packages/aarch64_cortex-a72/packages/tvheadend_4.2.8-5_aarch64_cortex-a72.ipk
Installing libffi (3.3-2) to root...
Downloading https://downloads.openwrt.org/releases/21.02.1/packages/aarch64_cortex-a72/packages/libffi_3.3-2_aarch64_cortex-a72.ipk
Configuring libffi.
Configuring tvheadend.
root@OpenWrt:~# /etc/init.d/tvheadend start
root@OpenWrt:~# ps w | grep tvheadend
11850 root     61036 S    /usr/bin/tvheadend -B -c /etc/tvheadend --ipv6
11952 root      1228 S    grep tvheadend
```

M95D:
Already fixed here:
https://github.com/openwrt/packages/pull/16475

BKPepe:
I am afraid that it is fixed only for OpenWrt master, which makes this issue still valid for OpenWrt 21.02.

M95D:
I don't think I can help with that.

BKPepe:
Hmm. So, the issue is in the stable branch. What can be done about it:

1. Someone can send a pull request against that branch
2. Should we, I mean me, cherry-pick it from ``master`` branch to ``openwrt-21.02`` branch?
I didn't look at it closely, but in the pull request, you added tvheadend user and use it. Is it possible that some permissions migration would be necessary or not?

walterav1984:
I can confirm this is currently an issue on Raspberry Pi 1 with 21.02.1 release, but I'm almost certain this worked at the end of august 2021 as in 4.0 (old/previous tvheadend 4.0.x on 19.07.x style/way meaning running as ROOT no hdhomerun nor satipclient) or maybe an early 4.2.x version without @M95D improvements?

[...]

PR #17872, opened 2022-02-15
tvheadend: fix issue #17816

M95D:
This is a temporary workaround until the tvheadend bug is fixed.

Maintainer: me
Compile tested: Not tested. I took it from master branch, removing the user and group options.
Run tested: r18404, I tested only the command, not whole package.

Description:
The command I replaced should create tvheadend initial admin account without password and save it to /etc/tvheadend. It does not. Tvheadend aborts execution before the configuration is saved to disk. It's a tvheadend bug.
This was fixed in master branch https://github.com/openwrt/packages/pull/16475, but that pull request makes lots of other changes and it was not included in this branch.

### DIFF
--- a/multimedia/tvheadend/files/tvheadend.init
+++ b/multimedia/tvheadend/files/tvheadend.init
@@ -10,7 +10,12 @@ TEMP_CONFIG=/tmp/tvheadend
 PERSISTENT_CONFIG=/etc/tvheadend
 
 execute_first_run() {
-	"$PROG" -c "$1" -B -C -A >/dev/null 2>&1
+	# This should create a new configuration including an admin account with no name and no password,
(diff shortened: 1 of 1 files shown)


BKPepe:
- Commit subject is wrong. It should start with the prefix package name. It means: tvheadend
and it should be more clear than Fix issue "link". 
I can imagine something like: tvheadend: fix first run

- There is a missing commit description.
You can here mention the issue, etc.

- Formal things: missing ``Signed-off-by`` in the commit description.

You are still referring to that this is a tvheadend issue. Did you report it to upstream? I am afraid they will not fix it in the stable release, but they can in the development version (4.3). If the upstream does not know about it, then they can not fix it.

M95D:
Sorry. I did it from the github file editor.

PR #4119, opened 2017-03-07
Cleanup: avoid reuse of $(PKG_NAME)

stintel:
For consistency, use full name instead of $(PKG_NAME) in define and eval lines for all packages.
I've seen reviews that asked to do this before, and I am asking the same during reviews now. To avoid this in the future, fix this treewide so when people use existing packages as example, we will not have to request this change anymore.

This makes all packages consistent with both LEDE and OpenWrt base repositories (check with ``git grep PKG_NAME| egrep -i 'define|call|eval'``).

There is one exception for PHP pecl packages, where ``PKG_NAME:=php7-pecl-$(PECL_NAME)``. Do we want to change these, too?

PR #17876, opened 2022-02-16
bandwhich: terminal bandwidth monitoring utility

Grommish:
Initial Commit - version 0.20.0

bandwhich is a terminal bandwidth monitoring utility

This package requires rust-lang (https://github.com/openwrt/packages/pull/13916)

Description:

### DIFF
--- a/net/bandwhich/Makefile
+++ b/net/bandwhich/Makefile
@@ -1,0 +1,40 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=bandwhich
+PKG_VERSION:=0.20.0
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_DATE:=2020-10-15
+PKG_SOURCE_VERSION:=a703513412d4da71085e55fd9d62a080527b2800
+PKG_SOURCE_URL:=https://github.com/imsnif/bandwhich.git
+PKG_MIRROR_HASH:=647e1412a84b224a2490c316fc536ee838dab1957d4c5bfb1dbdfe49b870d8d9
+
+PKG_BUILD_DEPENDS:=rust/host
+
+include $(INCLUDE_DIR)/package.mk
+include $(TOPDIR)/package/feeds/packages/rust/rustc_environment.mk
+
+define Build/Compile
+	$(call RustPackage/Cargo/Update)
+	$(call RustPackage/Cargo/Compile)
+endef
+
+define Package/bandwhich
+    SECTION:=net
+    CATEGORY:=Network
+    DEPENDS:=@!SMALL_FLASH @!LOW_MEMORY_FOOTPRINT +zlib +libncurses +libncurses-dev
+    TITLE:=Terminal bandwidth utilization tool
+    URL:=https://github.com/imsnif/bandwhich
+endef
+
+define Package/bandwhich/description
+ Rust-lang based Terminal bandwidth utilization tool
+endef
+
+define Package/bandwhich/install
+	$(INSTALL_DIR) $(1)/bin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/target/$(RUSTC_TARGET_ARCH)/release/bandwhich $(1)/bin/bandwhich
+endef
+
+$(eval $(call BuildPackage,bandwhich))


BKPepe:
Does it make sense to open so many PRs if they depend on Rust? Rust is not getting to be merged anytime soon. You could add those packages into the existing PR.

net/bandwhich/Makefile
```diff
@@ -0,0 +1,91 @@
...
+# 3) Save the Makefile, and run the following to update the feeds index:
+#    ./scripts/feeds update -i
+#    ./scripts/feeds install <pkg_name>
+# 4) Call 'make package/feeds/packages/to/app/{download,check} FIXUP=1'
+# 5) cargo packages typically do not have a ./configure step, so you
+#    can leave that entire section blank.
+# 6) Build/Compile is used to call cargo and/or rustc - See the below
+#    example on how to use the variables and format
+# 7) Be sure to run `make menuconfig` PRIOR to building, so that it updates the
+#    dependencies.
```
BKPepe, line 21: This does not belong here.

net/bandwhich/Makefile
```diff
@@ -0,0 +1,91 @@
...
+PKG_VERSION := 0.20.0
+PKG_RELEASE := 1
+
+PKG_SOURCE_PROTO := git
+PKG_SOURCE_DATE := 2020-10-15
+PKG_SOURCE_VERSION := a703513412d4da71085e55fd9d62a080527b2800
+PKG_SOURCE_URL := https://github.com/imsnif/bandwhich.git
+PKG_MIRROR_HASH := 647e1412a84b224a2490c316fc536ee838dab1957d4c5bfb1dbdfe49b870d8d9
+
+PKG_BUILD_DEPENDS := rust/host
```
BKPepe, line 43: You can remove the whitespace between ``:=``

net/bandwhich/Makefile
```diff
@@ -0,0 +1,91 @@
...
+CONFIGURE_ARGS += \
+  	--host=$(REAL_GNU_TARGET_NAME)
+
+define Build/Compile
+        cd $(PKG_BUILD_DIR) && $(TARGET_CONFIGURE_OPTS) $(CONFIGURE_VARS) cargo update && \
+	  $(TARGET_CONFIGURE_OPTS) $(CONFIGURE_VARS) cargo build -v --release \
+	  --target $(REAL_GNU_TARGET_NAME)
+endef
+
+define Package/$(PKG_NAME)
```
BKPepe, line 74: Please avoid to re-use ``$(PKG_NAME)``

https://github.com/openwrt/packages/pull/4119

BKPepe (changes requested):

net/bandwhich/Makefile
Grommish: Is there a reason for this other than looks?  `nano` syntax highlights for GNU Makefile, but only when it's properly spaced.

BKPepe: Well, you are doing something here special. You might want to take a look at it. ;-) All the Makefiles here does not have it.

Grommish: I've made the changes, but I did it that way because of the Reference https://www.gnu.org/software/make/manual/make.html#Reference that outlines the Simple Expanded Variables.  I'll keep it in mind for the future though.  I never bothered to check the spacing in previous PR 👍 

This isn't going anywhere until I can get traction on testing rust across the arches, but I'm hoping to generate at least some interest if it's more than a "hypothetical"

Grommish:
@BKPepe Will it help to tag these with `Blocked`?  Will that keep the CI from running?  I know I can do it with `@BROKEN` in package, but then I have to use a non-standard config (turning on broken packages) to test it.

BKPepe:
@Grommish The CI works for you. The label Blocked was added because of:
> This package requires rust-lang (https://github.com/openwrt/packages/pull/13916)

Still, it does not make sense to me that you are opening multiple PRs which depends on Rust, which is not getting here any time soon.

Grommish:
> Still, it does not make sense to me that you are opening multiple PRs which depends on Rust, which is not getting here any time soon.

Because it works and shouldn't be limited to just me.  If it never gets accepted into the mainline, so be it. However, with the inclusion of rust into the kernel and other people approaching the project in interest, I'm hoping it won't be overly long.  Is there testing to do? Yes, however, I have no qualms about limiting the arches to only tested ones over on the rust package itself.

Grommish:
Update `Makefile` to use the `cargo` handling calls and environmental file.

Grommish:
This update corrects for how the `rustc`/`cargo` calls are being used by the build system.

net/bandwhich/Makefile
Grommish: Removed extraneous whitespace

Grommish:
image

gyrusdentatus:
> Does it make sense to open so many PRs if they depend on Rust? Rust is not getting to be merged anytime soon. You could add those packages into the existing PR.

Why isn't Rust getting merged anytime soon? Any serious reason for that? Rust is inevitable! Either add Rust or your  thing will become obsolete in the next 2-5 years. There are many people that actually need Rust and this amazing person @Grommish has done a wonderful job and is doing his best. I don't understand why you're being mean.

Grommish:
Mean?  No, no one is being mean 🤣  

This package won't work without the rust package, so it's in limbo until/if rust is at the point it can be included.  I'm happy to draft this to get it off the screen and still make it findable by folks who are testing rust.

In the future, you might even be able to have rustup in the build system or on the host, at which point anyone with rustup would be able to compile this.  The tuples are still in the process of being upstreamed into the rust master (`mips64-openwrt-linux-musl` is already available), but I'd guess if you built for a mips64 target, you could just use rustup on your build machine since it knows what to do with it.  Generating the interest in people to test is the hard part.  "Run this random ipk on your hardware" does generally get many volunteers, so these packages are the carrot to bring source-building testers in because it gives them something to actually use.

I might just move the apps off openwrt's tree and let people find them via a link in the rust pr.

gyrusdentatus:
> Mean? No, no one is being mean rofl
> 
> This package won't work without the rust package, so it's in limbo until/if rust is at the point it can be included. I'm happy to draft this to get it off the screen and still make it findable by folks who are testing rust.
> 
> In the future, you might even be able to have rustup in the build system or on the host, at which point anyone with rustup would be able to compile this. The tuples are still in the process of being upstreamed into the rust master (`mips64-openwrt-linux-musl` is already available), but I'd guess if you built for a mips64 target, you could just use rustup on your build machine since it knows what to do with it. Generating the interest in people to test is the hard part. "Run this random ipk on your hardware" does generally get many volunteers, so these packages are the carrot to bring source-building testers in because it gives them something to actually use.
> 
> I might just move the apps off openwrt's tree and let people find them via a link in the rust pr.

Can we please get in touch? Please check your mailbox in ~ 5 min. I'm kinda desperate here, like for real ! Please ! :)

BKPepe:
Closing, OP is inactive.

PR #17884, opened 2022-02-17
uacme: proper ualpn integration

lucize:
ualpn works only with openssl and gnutls, so default build will not have this included

Maintainer: me / @
Tested mips64

So this is a hackish way to include the package based on libcurl ssl library, if someone has a better idea, please do share!


lucize:
I'll add mbedtls 2.28 and if accepted in base, the hack install can go away

ndilieto:
@lucize Now that the update to mbedtls 2.28 has been merged, you may want to remove the hack...

PR #17887, opened 2022-02-17
libdaq3: Enable NFQ DAQ

miska:
Maintainer: @flyn-org 
Compile tested: mvebu, Turris Omnia, Turris SO 6.0 ~ OpenWrt 21.02
Run tested: mvebu, Turris Omnia, Turris SO 6.0 ~ OpenWrt 21.02, snort seems to able to use, started and processed some packets

Description:
Enable support for DAQ module that uses NFQUEUE.


flyn-org:
Looks good. Recommend merge.

Eventually we might want to conditionally compile some of this, but I think most routers that aim to run Snort will be pretty well-resources.

Issue #16500, opened 2021-08-30
tvheadend: satip-client discovery not working(workaround)


walterav1984:
Maintainer: @M95D
Environments tested: 
bcm2708, Raspberry Pi 1, (OpenWrt version snapshot r17454-f08b76b32f,  old-stable 19.07.8, stable 21.02.2,
OpenWrt SNAPSHOT r18801-56256259a1, OpenWrt SNAPSHOT r18807-7fae1e5677)
lantiq/xrx200, ZyXEL P-2812HNU-F1, (OpenWrt 21.02.2 r16495-bf0c965af0)

Devices routers/tuners and configs:
-RPi1 running OpenWrt ext4 image, only disabled DHCP server on LAN for this device:
*passive (static ip no DHCP server no default route using the LAN interface)
*active (LAN interface gets ip by DHCP-client not the WAN, and default route/dns from main router)

-ZyXEL P-2812HNU-F1(for testing besides the Raspberry Pi1, because it has more interfaces):
*default OpenWrt Lan/Wan Router config only changed WAN "dsl0" to "eth0.2" 

-FritzBox 6490 Sat-IP tuner disabled DHCP server on LAN for this device: 
*passive (static ip matching subnet of main router no default route)
*active(LAN interface gets DHCP from openwrt router or main router depending or setup)

-HDHomeRun:
*active(LAN gets DHCP ip-address from main router or OpenWrt Router Rpi1/Zyxel)

Issue Description:
Discovery of SAT>IP server based network tuners is not working out of the box with Tvheadend web-interface on OpenWrt.

You can let the satip-client function scan for satip-server based tuners by pressing the following button in the Tvheadend webinterface "Configuration>General>SATIP-Server>Discover SATIP-Servers". This would normally list (multiple) network satip and HDhomerun tuners in the "Configuration>DVB Inputs>TV Adapters" section.

Assuming that all devices are connected with each other via ethernet and configured for the same subnet as the OpenWrt "Lan" interface is using. Only the Raspberry or the Zyxel were tested in a setup as being the OpenWrt device not both together!
Both test cases of having a default OpenWrt router WAN/LAN config with a 192.168.1.1 subnet on its LAN with a DHCP server or a OpenWrt config not running as a router but as a passive/active LAN client were tested a long the way. All configuraties have this issue of not detecting SATIP tuners by default, except the config were the RPI1 has the one and only interface have its default route 0.0.0.0 and gateway set on its own LAN interface. 

[...]

PR #17896, opened 2022-02-18
tvheadend: bind to LAN IP by default

M95D:
Maintainer: me
Compile tested: not tested, compilation is not affected
Run tested: r18404

Description:
If config parameter is not set, tvheadend will bind to LAN IP address by default. If IP not found, the switch is not added and tvheadend will bind to 0.0.0.0, as it is now.
Fixes issue #16500 without requiring user to edit the config file. It's also more secure if firewall becomes disabled. There are no drawbacks that I can think of.

### DIFF
--- a/multimedia/tvheadend/files/tvheadend.config
+++ b/multimedia/tvheadend/files/tvheadend.config
@@ -5,6 +5,8 @@
 
 config tvheadend server
 	option ipv6 '1'
+	# If bindaddr is not set, it will default to LAN IP address (uci get network.lan.ipaddr).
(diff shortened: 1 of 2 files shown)


walterav1984:
Tested successful both on RaspberryPi1(LAN-ip client active/passive) and Zyxel P-2812HNU-F1(normal router WAN/LAN config) running OpenWrt 21.02.2 are able to instantly detect SATIP-tuner (fritzbox 6490 v07.29, with ip-address matching the LAN interface subnet of the OpenWrt devices) with this patch.

PR #17908, opened 2022-02-20
coreutils: update to 9.0

graysky2:
Removed unneed patch:
```
  002-ls-restore-831-behavior-on-removed-directories.patch
```
After updating ran `make package/coreutils/refresh` from `$BUILDROOT` to clean dirty patches.
```
Build system: x86_64
Build-tested: bcm2711/RPi4B
Run-tested: bcm2711/RPi4B
```

Maintainer: @neheb @jow- @champtar


graysky2:
I am a little confused since there is coreutils here and also under the openwrt repo.  Do they need to be bumped simultaneously?

neheb:
I don't think so. tools/coreutils is equivalent to a coreutils/host package.

PR #17924, opened 2022-02-22
Stop taxing the resources of the FreeBSD project with htpdate requests

allanjude:
Remove `www.freebsd.org` from the list of hosts that receive a constant deluge of HTTP requests from clients looking for the current date

FreeBSD has taken to removing the date header from its HTTP response due to this abuse.


jow-:
@allanjude @marcin1j - I took the liberty to fixup and push this change directly, hope this is okay.

allanjude:
Thank you.

PR #17929, opened 2022-02-23
cni-plugins-nft: change naming to support fw4

oskarirauta:
added patch with following features:

 - forces table family always to inet instead of ip or ip6.
 - force fw4 table using
 - change default chain names for fw4


oskarirauta:
Anyone to review? This patch actually makes this useful in Openwrt.

oskarirauta:
@neheb

BKPepe:
Both patches should have some description, what they are doing. Did you reach upstream if there is something what they can do? Right now, it seems like we need to carry such patches forever and maintain them, of course. 

Prefix, which you used such as ``010-``, ``020-`` means that they are upstream backports. Is that true?

PR #17931, opened 2022-02-23
[21.02] expat: import patches for CVEs

paper42:
Maintainer: @thess
Compile tested: powerpc 8540, Turris 1.X, OpenWrt 19.07
Run tested: powerpc 8540, Turris 1.X, OpenWrt 19.07 only a bit, I would appreciate help with testing this.

Description: It's a bit unfortunate that we have an unsupported expat version in 21.02, but I don't know if it would be idea to switch to 2.4 in a stable branch. I got the patches from [alpine 3.13](https://gitlab.alpinelinux.org/alpine/aports/-/tree/3.13-stable/main/expat) and refreshed them. This commit should be backported to 19.07, but there we should also update expat to .10.


neheb:
What's expat used for?

paper42:
> What's expat used for?

In 14 packages if I am counting correctly, most notably avahi, libarchive and unbound, but I am not sure under which circumstances it gets used there.

neheb:
xml parser. great...

PR #17934, opened 2022-02-24
tvheadend: fix group /dev/dvb permissions, add /etc/tvheadend to conffiles

clayface:
Maintainer: @M95D 
Compile tested: mvebu
Run tested: mvebu/rb5009

Description: Fix group permissions so that tvheadend user can access adapters. Also add the /etc/tvheadend dir to conffiles to preserve it on sysupgrades.

### DIFF
--- a/multimedia/tvheadend/Makefile
+++ b/multimedia/tvheadend/Makefile
@@ -242,6 +242,7 @@ endef
 
 define Package/tvheadend/conffiles
 /etc/config/tvheadend
+/etc/tvheadend
 endef
 
 define Package/tvheadend/install
--- a/multimedia/tvheadend/files/dvb.hotplug
+++ b/multimedia/tvheadend/files/dvb.hotplug
@@ -1,7 +1,7 @@
 #!/bin/sh
 case "$ACTION" in
   add)
-    chown -R root:dvb /dev/dvb/*
-    chmod -R 660 /dev/dvb/*
+    chgrp -R dvb /dev/dvb/*
+    chmod -R g+rwX /dev/dvb/*
     ;;
 esac
--- a/multimedia/tvheadend/files/tvheadend.init
+++ b/multimedia/tvheadend/files/tvheadend.init
@@ -79,7 +79,8 @@ start_service() {
 	ensure_config_exists
 	procd_open_instance
 	procd_set_param file /etc/config/tvheadend
-	chown -R root:$TVH_GROUP /dev/dvb/*
+	chgrp -R $TVH_GROUP /dev/dvb/*
+	chmod -R g+rwX /dev/dvb/*
 	procd_set_param command "$PROG" -B -u $TVH_USER -g $TVH_GROUP
 	load_uci_config
 	procd_close_instance


clayface:
By the way, is it necessary to have `chown -R root:$TVH_GROUP /dev/dvb/*` in tvheadend.init if it's already in the hotplug script? Do I need to put the chmod command here too?

M95D:
Yes, I made some mistakes when I wrote those permissions. The dvb adapter directory (/dev/dvb/adapter0) has x permission by default. With my **chmod -R 660**, x will be removed.
But your command, **chmod -R g+rwx** will set x on device files too. I think it should be "X" instead:
chmod -R g+rwX /dev/dvb/*

About tvheadend.init:
I don't know if hotplug script executes after reboot if the dvb stick is already plugged-in. Probably both chown and chmod should be put in init script.

clayface:
> I think it should be "X" instead

fixed

> I don't know if hotplug script executes after reboot if the dvb stick is already plugged-in

In all my tests the hotplug script was used on reboot as well. I think we don't need to do it twice unless you want to be very safe.

walterav1984:
Maybe slightly offtopic but @clayface and @M95D which dvb adapters and interfaces (USB/PCI/PCI-E) did you test and especially which kernel modules for modulators etc. Do you have any resources to those patches enabling these devices dvb?

clayface:
I tested using a Hauppauge WinTV-DualHD USB-2.0 adapter.

As for enabling these in OpenWrt it's possible and something I would like as well - this is a useful feature that others may use if it wasn't difficult to do. However testing will be difficult without possessing devices from multiple vendors. When I have time I will look into it in more detail.

walterav1984:
You are right about diffculty of enabling that hardware kernel module support. Did a custom hack a while ago for 18.06.x with usb Mircrosoft Xbox tuner / Astrometa dvb-t2 (both easy available) using the dvb-usb,dvb_usb_dib0700,mn88472/3 using a customized "openwrt-dvb" github repository (sorry cannot find that one anymore) it worked fine with tvheadend. But I rather see all those modules come available by default in the OpenWrt kmod opkg repository, since doing it for a newer release without the upstream openwrt-dvb was very hard.

For Info on USB/PCI/PCI-E tuners I have access too for testing see the [hardware table](https://ubuntuforums.org/showthread.php?t=2336566) in this link:

M95D:
I use Astrometa USB tuner (Sony demod). I build my own openwrt firmware from source with all needed kernel modules built-in. They are all included in openwrt kernel sources. No external tree needed. But there's a catch! The kernel menuconfig options only become visible after i2c bus multiplexing (I2C_MUX) option is selected.

@clayface
I want to be extra-safe and have them in both places. For example, the user may plug-in the tuner before tvheadend (and dvb hotplug script) is installed.

clayface:
> I want to be extra-safe and have them in both places. For example, the user may plug-in the tuner before tvheadend (and dvb hotplug script) is installed.

Done.

M95D (approved):

clayface:
@M95D one more thing if you don't mind. Should `chown` be changed to `chgrp` since we aren't touching the user?

M95D:
Yes, why not. Go ahead.

clayface:
@M95D done.

@neheb believe this is ready. Thanks.

walterav1984:
@M95D
> They are all included in openwrt kernel sources. No external tree needed. But there's a catch! The kernel menuconfig options only become visible after i2c bus multiplexing (I2C_MUX) option is selected.
> Kernel config options are: DVB_CXD2841ER, DVB_RTL2832, MEDIA_TUNER_R820T.

When building the linuxtv mediatree modules/drivers this way, will they all be available as seperate 'kmod-xxxxx' files or be included by default in the kernel for your image? 

Since this PR was merged it seems that OpenWrt with Tvheadend or even Minisatip might only miss a [dvb-firmware](https://github.com/tmn505/openwrt-dvb/blob/ca8b5c7bca4c5eae5454127af438f52e3f8cc563/dvb-firmware/Makefile) and include libdvben50221 for Commen Interface based hardware. I'm aware that dependencies of modules are not automaticly solved but atleast they will be available to install by the end user themself, comparable with intel wifi which does not auto install firmware requires 2 packages.

Would it be intime to ask a pull request on [github.com/openwrt/openwrt 22.03 branch](https://github.com/openwrt/openwrt/issues/9248) consider enabling the default build of "i2c bus multiplexing (I2C_MUX) option" and "all other related media-tree tuner modules" for the upcoming release? @M95D @neheb

M95D:
I build my own image with everything included in kernel or as kernel modules preinstalled in squashfs. I don't use openwrt packages for drivers or firmware.

You may want to install firmwares manually. There are many tuners and package mentainance may be too much. There are also copyright issues with many of them.

walterav1984:
> You may want to install firmwares manually. There are many tuners and package mentainance may be too much. There are also copyright issues with many of them.

Does it mean as in although the driver code may already be in the current openwrt linux source, compiling and shipping the modules might be an issue?

> Dvb module drivers would make sense only for targets with platform dvb (are there any of these even supported?) or devices with usb.

By platform do you mean architectures like x86/64, ath79 and ramips that does have these module sources by default, or that some interfaces like pci or usb are lacking on those platforms and block building at all?

> I don't think there in much interest for dvb on openwrt...

It will be niche offcourse, people might try a raspberry pi or even a pc but the hardware could be perfectly usable as in minisatip server or simple tvheadend setup.

M95D:
- Not the drivers. Those are free. The firmwares. They are usually extracted from the windows binary driver.
- I was referring to embedded devices specialised as TV tuners. See the kernel menuconfig. There is a submenu with various TV tuners that are not PCI and not USB - "platform" DVB. Devices with PCI(e) usually support booting a full Linux distro, and then why use OpenWrt on them?
- I don't know how many users need DVB drivers. I don't even know how popular tvheadend is! (Hey! Package download statistics from OpenWrt servers would be nice!!!) Here on github there's only 3-4 participants. In the forum - even less. I updated the tvheadend package for myself. I pushed the updates to OpenWrt git just because it was already here. I didn't expect it to be actually used by anyone. Adding DVB drivers and firmwares is a lot of work for very few users.

PR #17937, opened 2022-02-25
shadowsocks-libev: convert to using nft

yousong:
Maintainer: me
Compile tested: x86/64, reboot-18956-g86b5e651d6 
Run tested: x86/64, reboot-18956-g86b5e651d6, tcp/udp redir, local redir, nft set match

Description:

See #16818


yousong:
v3 <- v2: verbose commit messages.

jeffpeng78:
ss-rules is not able to be disabled on OpenWRT 22.03
After further investigation, found the file "90-ss-rules.nft" is left remain in directory /etc/nftables.d/ when the ss-rule is disabled.  So the fw4 still forwards the traffic.
Please remove the "90-ss-rules.nft" when disable the ss-rule.

jeffpeng78 (commented):
ss-rules is not able to be disabled on OpenWRT 22.03
Found the file "90-ss-rules.nft" is left remain in directory /etc/nftables.d/ when the ss-rule is disabled.  So the fw4 still forwards the traffic.
Please remove the "90-ss-rules.nft" when disable the ss-rule.

yousong:
> ss-rules is not able to be disabled on OpenWRT 22.03 Found the file "90-ss-rules.nft" is left remain in directory /etc/nftables.d/ when the ss-rule is disabled. So the fw4 still forwards the traffic. Please remove the "90-ss-rules.nft" when disable the ss-rule.

I opened a pull request for this.  Please give it a try and see if it can fix your case.  For further updates, let's move on there.
https://github.com/openwrt/packages/pull/19132

I am going to lock the session here.

PR #17940, opened 2022-02-25
mwan3: update to version 2.11.0

feckert:
Maintainer: me @aaronjg 
Run tested: x86_64, APU3, OpenWrt latest

Changes:
* Add ip6tables dependency to fix issue that was added by the iptables package renaming.
* Add debugging dump into `/var/run/mwan3/iptables_log` to see all iptables/ipset command that mwan3 do for debugging
* Move the iptables and ipset command definition to `/lib/mwan3/common.sh`.
* Since nftables does not support ipsets in ipsets (list:set), there is now a separate ipset and iptables rule for each ipset in mwan3. This is not mandatory, but it helps when porting from mwan3 to native nftables syntax without iptables-nft backend usage, to verify the rules.

@champtar @stangri @jamesmacwhite @brianjmurrell @ptpt52 
Could someone please verify my change with nftables before I merge it?

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -25,6 +25,7 @@ define Package/mwan3
      +ip \
      +ipset \
      +iptables \
+     +IPV6:ip6tables \
      +iptables-mod-conntrack-extra \
      +iptables-mod-ipopt \
      +jshn
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.10.13
+PKG_VERSION:=2.11.0
 PKG_RELEASE:=1
 PKG_MAINTAINER:=Florian Eckert <fe@dev.tdt.de>, \
 		Aaron Goodman <aaronjg@alumni.stanford.edu>
--- a/net/mwan3/files/etc/hotplug.d/iface/15-mwan3
+++ b/net/mwan3/files/etc/hotplug.d/iface/15-mwan3
@@ -3,7 +3,6 @@
 . /lib/functions.sh
 . /lib/functions/network.sh
 . /lib/mwan3/mwan3.sh
-. /lib/mwan3/common.sh
 
 initscript=/etc/init.d/mwan3
 . /lib/functions/procd.sh
--- a/net/mwan3/files/etc/init.d/mwan3
+++ b/net/mwan3/files/etc/init.d/mwan3
@@ -1,6 +1,5 @@
 #!/bin/sh /etc/rc.common
 
-. "${IPKG_INSTROOT}/lib/mwan3/common.sh"
 . "${IPKG_INSTROOT}/lib/functions/network.sh"
 . "${IPKG_INSTROOT}/lib/mwan3/mwan3.sh"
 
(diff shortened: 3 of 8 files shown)


champtar:
Hi @feckert 
Why do you need to force iptables-legacy ?

feckert:
@champtar This change is a preparation to use iptables-nft later. I actually want to do two variations one with iptables and the dependencies and one with nftables and the dependencies. The problem I see are the IPsets I want to translate them with the new tool `ipset-translate restore` and then use them. For this to work I have staged a change for ipset version update to v7.15 https://patchwork.ozlabs.org/project/openwrt/list/?series=287940
But I am not that far yet.

I would like to see that the mwan3 works again after the dependency change with iptables.

champtar:
ipsets should be usable with iptables-nft.
The idea is that if you don't need to insert you rules in the middle of the OpenWrt ones (maybe mwan3 does), you should be able to switch between iptables-legacy + fw3 and iptables-nft + fw4 more or less transparently.
`iptables` is now provided by both `iptables-legacy` and `iptables-nft` so either leave the code referencing the generic command, or change the dependencies.
If you have examples of broken iptables invocations I can try to have a look

feckert:
@champtar you are welcome to look at it. Unfortunately, I'm not really familiar in nftables yet.
But in order to be able to support nftables and iptables at once, I have to clean up the mwan3 first with this pullrequest.
Translating the iptables rules with iptables-nft seems to work. But my problem is the ipsets conected, dynamic, custom and sticky.

These cannot simply be translated. The problem here is, that there is no type in the nftables universe that creates ipsets in ipsets. So for each ipset I use, I have to swap it out into a separate ipset to use this set in an iptables rule with this commit
https://github.com/openwrt/packages/pull/17940/commits/bf7e276d0b4494684db375e395253ba4007b397a.

If this works, I think I can then translate these sets into nftables-set if mwan3 uses nft and use them with iptables-nft translation.
But I'm not that far yet

stangri:
> ipsets should be usable with iptables-nft. The idea is that if you don't need to insert you rules in the middle of the OpenWrt ones (maybe mwan3 does), you should be able to switch between iptables-legacy + fw3 and iptables-nft + fw4 more or less transparently. `iptables` is now provided by both `iptables-legacy` and `iptables-nft` so either leave the code referencing the generic command, or change the dependencies. If you have examples of broken iptables invocations I can try to have a look

Is the goal to also provide two flavors of dnsmasq-full, one with the ipset support and another with nftset support?

ptpt52:
are you sure tested ok? make sure all ipset destroyed ok after mwan3 stop?

ptpt52:
```
ipset create listset list:set
ipset create ipset1 iphash
ipset add listset ipset1
ipset add listset ipset1
ipset add listset ipset1
ipset add listset ipset1
ipset list listset
ipset destroy listset
ipset destroy ipset1

ipset v7.6: Set cannot be destroyed: it is in use by a kernel component
```

feckert:
@ptpt52 Thank you for your test
 But I don't think that applies here anymore.
I no longer have an ipset in ipset (list:set). Each ipset now has its own iptables rule.

ptpt52:
> @ptpt52 Thank you for your test But I don't think that applies here anymore. I no longer have an ipset in ipset (list:set). Each ipset now has its own iptables rule.

ok, I notice that it is bug on my custom kernel. so it is not bug here.

feckert:
No objection, then let's give it a try

jamesmacwhite:
One regression that's been spotted after merge is related to sticky chain names and their length https://github.com/openwrt/packages/issues/18155

For the mwan3_sticky chains, the change to use `ipv4` and `ipv6` in the chain name has added two extra characters which exceeds the 31 limit when a rule is more than 13 characters (max being 15).

PR #17952, opened 2022-02-28
mariadb: fix compilation with ppc64

neheb:
MariaDB's PPC64 CRC32 support requires AltiVec, which QoriQ does not
support. Disable it.

Maintainer: @miska 
Compile tested: ppc64


BKPepe:
Would be nice if you can send this to upstream repository, so we dont need to carry patches forever.

neheb:
where's upstream in this case?

BKPepe:
Talking about this: https://github.com/MariaDB/server/pulls

miska:
Seems fine to me for us, but would make sense to send to upstream with better condition that would rule out the unsupported CPUs better. So go ahead from me.

PR #17956, opened 2022-02-28
perl: add powerpc64 support

neheb:
Needed by the QoriQ target.

Maintainer: @pprindeville 
Compile tested: ppc64

### DIFF
--- a/lang/perl/Makefile
+++ b/lang/perl/Makefile
@@ -11,7 +11,7 @@ include perlver.mk
 
 PKG_NAME:=perl
 PKG_VERSION:=$(PERL_VERSION)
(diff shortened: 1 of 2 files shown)


pprindeville (approved):
LGTM

pprindeville:
What's required to get CI/CD coverage of this target?  The `powerpc_464fp` and `powerpc_8540` are both 32-bit, aren't they?

neheb:
Not possible right now. QoriQ is source only so no SDK available. Maybe @stintel can also test.

stintel:
I cannot test that as the current version doesn't build at all.

PR #17961, opened 2022-02-28
mksh: resign as maintainer

alive4ever:
Maintainer: @mirabilos 
Compile tested: none
Run tested: none

Description:
Remove myself from maintainers.

### DIFF
--- a/utils/mksh/Makefile
+++ b/utils/mksh/Makefile
@@ -10,10 +10,9 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mksh
 PKG_VERSION:=59c
-PKG_RELEASE:=1
+PKG_RELEASE:=$(AUTORELEASE)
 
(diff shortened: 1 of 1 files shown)


mirabilos:
Sad to see you go, thanks for the help so far.

I asked in IRC if an active OpenWrt user can pick this up; otherwise I’ll occasionally test in qemu as before.

alive4ever:
Personally, I love `mksh` as login shell, with good balance between size and feature. In my opinion, `mksh` is more than `busybox sh` and less than `bash`.

In fact, `mksh` as default shell on Android is what drove me to adopt the package.

Unfortunately, I haven't been doing OpenWrt related stuffs recently, so the best step forward is resigning from maintaining such an essential package.

alive4ever:
While you are here, would you look at [mksh-makefile](https://github.com/alive4ever/mksh-makefile), which I started a few years ago.

PR #17963, opened 2022-03-01
sqm-scripts: switch back to iptables

champtar:
Maintainer: @tohojo 
Compile tested: no
Run tested: no

Description:
Following recent dependency rework, we can switch
between iptables-legacy and iptables-nft, and they both
PROVIDES iptables. Make it easier for user that want/need to
stick to firewall3/iptables-legacy to do so.


tohojo (approved):

champtar:
@tohojo can I go ahead and merge ?

tohojo:
Sure, go ahead!

PR #17965, opened 2022-03-01
cloudflared: add new package

1715173329:
Maintainer: me
Compile tested: ramips/mt7621, rockchip/armv8, x86/64
Run tested: nanopi-r2s, x86_64 (via VMware)

Description:
Contains the command-line client for Cloudflare Tunnel, a tunneling
daemon that proxies traffic from the Cloudflare network to your origins.

This daemon sits between Cloudflare network and your origin (e.g. a
webserver). Cloudflare attracts client requests and sends them to you
via this daemon, without requiring you to poke holes on your firewall
--- your origin can remain as closed as possible.

Docs: https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/install-and-setup/tunnel-guide

GitHub: https://github.com/cloudflare/cloudflared

### DIFF
--- a/net/cloudflared/files/cloudflared.config
+++ b/net/cloudflared/files/cloudflared.config
@@ -1,0 +1,9 @@
+
+config cloudflared 'config'
+	option enabled '0'
+	option config '/etc/cloudflared/config.yml'
+	option origincert '/etc/cloudflared/cert.pem'
+	option region ''
+	option loglevel 'info'
+	option logfile '/var/log/cloudflared.log'
+
--- a/net/cloudflared/files/sample_config.yml
+++ b/net/cloudflared/files/sample_config.yml
@@ -1,0 +1,3 @@
+url: http://localhost:8000
+tunnel: <Tunnel-UUID>
+credentials-file: /etc/cloudflared/<Tunnel-UUID>.json
--- a/net/cloudflared/Makefile
+++ b/net/cloudflared/Makefile
@@ -1,0 +1,65 @@
+# SPDX-License-Identifier: GPL-3.0-only
+#
+# Copyright (C) 2021 ImmortalWrt.org
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=cloudflared
+PKG_VERSION:=2022.3.1
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/cloudflare/cloudflared/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=e3459bc83d9749d89d93e6ab2dcbc51872568d1fb144d7ab7d3656b29785e364
+
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE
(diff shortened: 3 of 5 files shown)


BKPepe:
cc: @jefferyto

1715173329:
Updated to latest tag 2022.3.0.

jefferyto (approved):
LGTM - thanks!

1715173329:
Updated to 2022.3.1, added conffiles.

jefferyto:
> > Any way to keep the config files in /etc?
> 
> It would be possible with patching the source code directly: https://github.com/cloudflare/cloudflared/blob/5352b3cf0489e6568e148f40cd6e89660a7233cc/config/configuration.go#L35

It may be easier to use the [`config`](https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/configuration/arguments/#config) command line option.

> 
> > Also, does this need an init script?
> 
> I'll take a look on it.

Thanks :pray:

1715173329:
> > > Any way to keep the config files in /etc?
> > 
> > 
> > It would be possible with patching the source code directly: https://github.com/cloudflare/cloudflared/blob/5352b3cf0489e6568e148f40cd6e89660a7233cc/config/configuration.go#L35
> 
> It may be easier to use the [`config`](https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/configuration/arguments/#config) command line option.

Yes, `config` option is available. However it requires login at first which will generate `<uuid>.json` and `ca.pem` in default folder.

jefferyto:
> Yes, `config` option is available. However it requires login at first which will generate `<uuid>.json` and `ca.pem` in default folder.

Actually, looking at https://github.com/cloudflare/cloudflared/blob/5352b3cf0489e6568e148f40cd6e89660a7233cc/config/configuration.go there seems to be some support for loading config files from /etc/cloudflared? Will let you figure it out :joy:

1715173329:
> Actually, looking at https://github.com/cloudflare/cloudflared/blob/5352b3cf0489e6568e148f40cd6e89660a7233cc/config/configuration.go there seems to be some support for loading config files from /etc/cloudflared? Will let you figure it out 😂

maybe we can put a sample config in `/etc/cloudflared` so cloudflared will use this folder by default.🌚

1715173329:
Added init script and sample config.

net/cloudflared/files/cloudflared.config
```diff
@@ -0,0 +1,9 @@
+
+config cloudflared 'config'
+	option enabled '0'
+	option config '/etc/cloudflared/config.yml'
+	option origincert '/etc/cloudflared/cert.pem'
+	option region ''
+	option loglevel 'info'
+	option logfile '/tmp/cloudflared.log'
```
jefferyto, line 8:
```suggestion
	option logfile '/var/log/cloudflared.log'
```

net/cloudflared/files/sample_config.yml
```diff
@@ -0,0 +1,3 @@
+url: http://localhost:80
```
jefferyto, line 1:
```suggestion
url: http://localhost:8000
```

To match Cloudflare's documentation, and to not accidentally tunnel luci.

1715173329: Thank you for your suggestion!

jefferyto (approved):
Thanks again!

PR #17967, opened 2022-03-01
strongswan: add missing dependency

stintel:
Compile tested: master r19033-495c4f4e19 on x86/64
Run tested: NA

Description:

Fix the following build failure by adding the missing dependency:

Package strongswan-mod-connmark is missing dependencies for the following libraries:
libip4tc.so.2


pprindeville (commented):
LGTM

PR #17968, opened 2022-03-01
ffmpeg: update to 5.0

graysky2:
Bump to latest upstream version.

Manually rebased: `030-h264-mips.patch`
Backported: FFmpeg/FFmpeg/commit/2f0a214a6202516b4dda2bb22b6b3ac20e465d6d

Tested using:
```
ffmpeg -f lavfi -i testsrc=duration=10:size=1280x720:rate=30 testsrc.mpg
```
Resulting mpg was good.

Build system: x86_64
Build-tested: bcm2711/RPi4B
Run-tested: bcm2711/RPi4B

Maintainer: @BKPepe @bkuhls @lipnitsk @antonlacon @neheb

### DIFF
--- a/multimedia/ffmpeg/Makefile
+++ b/multimedia/ffmpeg/Makefile
@@ -9,12 +9,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ffmpeg
-PKG_VERSION:=4.3.3
+PKG_VERSION:=5.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://ffmpeg.org/releases/
-PKG_HASH:=9f0a68fbd74feb4e50dc220bddd59d84626774a53687fb737806ae00e5c6e9e6
+PKG_HASH:=51e919f7d205062c0fd4fae6243a84850391115104ccf1efc451733bc0ac7298
 PKG_MAINTAINER:=Ted Hess <thess@kitschensync.net>, \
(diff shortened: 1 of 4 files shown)


graysky2:
We can close this given the number of non-x86_64 and ARM failures.  I do not know what to do to fix the build on the others.  Open to suggestions.

neheb:
Sounds like missing -latomic

graysky2:
@neheb - Thoughts?

https://github.com/openwrt/packages/runs/5383554839?check_suite_focus=true#step:5:5862

neheb:
Yeah. libatomic needs to be added to dependencies.

diizzyy:
So now we have a lot of broken packages?

graysky2:
> So now we have a lot of broken packages?

Can you elaborate?

> As for migration we have two options, ether make sure all users / packages in tree works with ffmpeg 5.0 and/or deprecate (remove) incompatible ones or move current version of ffmpeg in tree to ffmpeg4 (along with users) and slowly migrate to "main port".

Are you suggesting that there are packages that depend on ffmpeg and need to be build against it?

neheb:
https://downloads.openwrt.org/snapshots/faillogs/powerpc_464fp/packages/gst1-libav/ is the only one I see.

diizzyy:
@graysky2 
5.0 is not a drop-in replacement to 4.X series
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=261302#c4 (and the bugs report itself) etc

neheb:
@diizzyy it's handled with ABI_VERSION, which is unfortunately not implemented here. In any case, the buildbots rebuild all packages daily, which prevents issues like this. opkg update is another matter.

PR #17970, opened 2022-03-02
shadowsocks-libev: add tcp_extra/udp_extra options

azuwis:
Maintainer: @yousong
Compile tested: (ramips/mt7621, r19033-495c4f4e19)
Run tested: (ramips/mt7621, r19033-495c4f4e19, tests done)

Description: Add extra statement to tcp/udp forward rule, example:

```
config ss_rules 'ss_rules'
    ...
    option tcp_extra 'tcp dport { 80, 443 }' # tcp only forward connections with dport 80 or 443
    option udp_extra 'udp dport { 53 }' # udp only forward connections with dport 53
```

This somewhat restores the old ipt_args functionality.

### DIFF
--- a/net/shadowsocks-libev/README.md
+++ b/net/shadowsocks-libev/README.md
@@ -71,6 +71,11 @@ We can have multiple instances of component and `server` sections.  The relation
 
 	local_default		[bypass], forward, checkdst
 
+	--- extra nftables statement added to the final tcp/udp forward rule
+
+	nft_tcp_extra		e.g. use 'tcp dport { 80, 443 }' to only forward connections with dport 80 or 443
+	nft_udp_extra		e.g. use 'udp dport { 53 }' to only forward connections with dport 53
+
 ss-rules now uses nft set for storing addresses/networks.  Those set names are also part of the API and can be populated by other programs, e.g. dnsmasq with builtin nft set support
 
 Note also that `src_ips_xx` and `dst_ips_xx` actually also accepts cidr network representation.  Option names are retained in its current form for backward compatibility coniderations
@@ -154,6 +159,8 @@ As for other options, change them only when you know the effect.
 		option src_default 'checkdst'
 		option dst_default 'forward'
 		option local_default 'forward'
+		# option nft_tcp_extra 'tcp dport { 80, 443 }' # tcp only forward connections with dport 80 or 443
+		# option nft_udp_extra 'udp dport { 53 }' # udp only forward connections with dport 53
 
 Restart shadowsocks-libev components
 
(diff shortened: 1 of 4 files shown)


net/shadowsocks-libev/README.md
```diff
@@ -154,6 +154,8 @@ As for other options, change them only when you know the effect.
 		option src_default 'checkdst'
 		option dst_default 'forward'
 		option local_default 'forward'
+		# option tcp_extra 'tcp dport { 80, 443 }' # tcp only forward connections with dport 80 or 443
+		# option udp_extra 'udp dport { 53 }' # udp only forward connections with dport 53
```
yousong, line 158: Maybe name it with `nft_` prefix , e.g. `nft_tcp_extra` to indicate that they are nft-specific.

Please also update README.md mentioning the additions like we used to have for `ipt_args`.

net/shadowsocks-libev/files/ss-rules/chain.uc
```diff
@@ -97,7 +97,7 @@ chain ss_rules_dst_{{ proto }} {
 
 {%   if (proto == "tcp"): %}
 chain ss_rules_forward_{{ proto }} {
-	meta l4proto tcp redirect to :{{ redir_port }};
+	meta l4proto tcp {{ o_tcp_extra }} redirect to :{{ redir_port }};
```
yousong, line 100: chain `ss_rules_local_out` also needs to be filtered.

net/shadowsocks-libev/README.md
azuwis: > Maybe name it with `nft_` prefix , e.g. `nft_tcp_extra` to indicate that they are nft-specific.

Done.
 
> Please also update README.md mentioning the additions like we used to have for `ipt_args`.

I don't quite understand, searched the history of `README.md` with `git log -p -- README.md`, and did not find the document of `ipt_args`.

I think the added notes about `nft_tcp_extra/nft_udp_extra` should explain the usage well.

net/shadowsocks-libev/files/ss-rules/chain.uc
azuwis: I've checked, it seems to work fine for local out without changing `ss_rules_local_out`.

net/shadowsocks-libev/README.md
yousong: My bad, checked again and found that I only added description in `incompatible changes` section for `ipt_args` after the [transition](https://github.com/openwrt/packages/commit/fdaf2de2aee717dc6ae9a39ef9ba6f0388993e5d)...

Anyway, please add proper description for the two options in README.md.  It's supposed to help users understand how components work together and how changing uci options will actually determine behaviors

azuwis: > Anyway, please add proper description for the two options in README.md. It's supposed to help users understand how components work together and how changing uci options will actually determine behaviors

Done, please check the new changes.

yousong:
I edited the README.md a bit and applied with https://github.com/openwrt/packages/commit/2de9035bda5464a3990b15ef582f6df243ea4f9c .   Thank you for the contribution!

PR #17982, opened 2022-03-03
openssh: update to 8.9p1

paper42:
Maintainer: @tripolar
Compile tested: aarch64, Turris MOX, OpenWrt master
Run tested: aarch64, Turris MOX, OpenWrt master

Description: According to reports from other distributions, using openssh server 8.9p1 on 32bit architectures without 32bit-seccomp.patch (which is already upstream in the 8.9 maintenance branch) does not accept connections. I didn't test this if OpenWrt is affected by this bug yet.

### DIFF
--- a/net/openssh/Makefile
+++ b/net/openssh/Makefile
@@ -8,18 +8,19 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openssh
-PKG_VERSION:=8.8p1
+PKG_VERSION:=8.9p1
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/ \
 		https://ftp.spline.de/pub/OpenBSD/OpenSSH/portable/
-PKG_HASH:=4590890ea9bb9ace4f71ae331785a3a5823232435161960ed5fc86588f331fe9
+PKG_HASH:=fd497654b7ab1686dac672fb83dfb4ba4096e8b5ffcdaccd262380ae58bec5e7
 
 PKG_LICENSE:=BSD ISC
 PKG_LICENSE_FILES:=LICENCE
 PKG_CPE_ID:=cpe:/a:openssh:openssh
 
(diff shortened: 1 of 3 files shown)


neheb:
Sounds like Turris Omnia would be impacted.

paper42:
I was not able to reproduce the bug on my Omnia and I am not sure why. The code with the bug is compiled, but maybe I need to somehow enable the sandbox and it's enabled by default on other distributions?

hardfalcon:
FYI: At least with GCC 11, OpenSSH 8.9 fails to build on non-x86 platforms because of the following issue:
https://www.mail-archive.com/gcc-patches@gcc.gnu.org/msg281299.html

My quick and dirty hack to get things working was to add the following as `./feeds/packages/net/openssh/patches/000-openssh-8.9p1-non-x86.patch`:
```
--- a/configure	2022-02-23 12:33:36.000000000 +0100
+++ b/configure	2022-02-23 12:33:36.000000000 +0100
@@ -6832,9 +6832,9 @@ rm -f core conftest.err conftest.$ac_obj
 	{ $as_echo "$as_me:${as_lineno-$LINENO}: checking if $CC supports compile flag -fzero-call-used-regs=all" >&5
 $as_echo_n "checking if $CC supports compile flag -fzero-call-used-regs=all... " >&6; }
 	saved_CFLAGS="$CFLAGS"
-	CFLAGS="$CFLAGS $WERROR -fzero-call-used-regs=all"
+	CFLAGS="$CFLAGS $WERROR -fzero-call-used-regs=skip"
 	_define_flag=""
-	test "x$_define_flag" = "x" && _define_flag="-fzero-call-used-regs=all"
+	test "x$_define_flag" = "x" && _define_flag="-fzero-call-used-regs=skip"
 	cat confdefs.h - <<_ACEOF >conftest.$ac_ext
 /* end confdefs.h.  */
```

However, before being added to the repo, this should probably get some more polishing to ensure `-fzero-call-used-regs=all` is not disabled on x86 platforms (it's a desirable security feature). Sadly, configure scripts are not my expertise at all, so somebody more knowledegable would have do to that polishing. The details of what this flag does and what values it can be set to can be found [here](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html) (search for `zero_call_used_regs`).

PR #17983, opened 2022-03-03
Update ModemManager and dependencies

kuanyili:
Maintainer: @nickberry17
Compile tested: aarch64, Raspberry Pi 4 Model B Rev 1.1, OpenWrt SNAPSHOT, r19039-8235723c78
Run tested: as above. checked basic functionalities.

Description:

Update mobile broadband connectivity packages including

- `libqrtr-glib` (1.0.0 -> 1.2.2)
- `libqmi` (1.30.2 -> 1.30.4)
- `libmbim` (1.26.0 -> 1.26.2)
- `modemmanager` (1.18.4 -> 1.18.6)

```
Overview of changes in libqrtr 1.2.2
----------------------------------------

 * libqrtr-glib:
   ** Fixed 'exported_packages' in GIR setup

 * doc:
   ** Added missing references.


Overview of changes in libqrtr 1.2
----------------------------------------

 * Build fully switched to meson.
   ** There are no longer source tarballs released, builds should be done
      exclusively from git repository checkouts, e.g. based on release tags.
   ** The GNU autotools based build is no longer supported.
   ** See https://modemmanager.org/docs/libqrtr-glib/building/building-meson/

 * Build now requires:
   ** glib2 >= 2.56.

 * libqrtr-glib:
   ** Added new qrtr-types.h header including all types in the library, which
      avoids cross references between the different objects.
```

```
Overview of changes in libqmi 1.30.4
----------------------------------------

 * Build:
   ** meson: switch to use the new python module in meson.
   ** meson: added a new boolean 'man' option in the meson setup to explicitly
      enable or disable building the man pages.
   ** meson: removed the option to detect if rmnet is supported.
   ** meson: multiple updates to use newer meson features like install_dir(),
      install_mode() or summary().
   ** meson: options 'mbim_qmux' and 'qrtr' are enabled by default and must be
      explicitly disabled if they're not needed, there is no attempt to
      autodetect whether they can be enabled or not.

 * qmi-proxy:
   ** Remove assert when attempting to close ghost device.

 * qmi-firmware-update:
   ** Use defaults if FLASH variables not reported, enabling support to flash
      the new Sierra Wireless EM9190 and EM9191 modules.

 * Several other minor improvements and fixes.
```

```
Overview of changes in libmbim 1.26.2
----------------------------------------

 * The meson build setup was updated as follows:
   ** Added new 'man' boolean option to allow disabling the generation of man
      pages explicitly; 'true' by default.
   ** Updated the 'introspection' option to be a boolean; 'true' by default.
      This effectively removes the automatic enabling of the introspection
      support based on whether the dependencies are found or not.
   ** Fixed static build failure with toolchains that don't support dynamic
      library.
   ** Add version script linker flag only when available.
   ** Fixed double quotes in udev rules file generated.

 * Added support for a new Quectel service, including the following operations:
   ** MBIM_CID_QUECTEL_RADIO_STATE

 * mbim-codegen:
   ** Explicitly require python3.

 * mbimcli:
   ** New '--quectel-query-radio-state' and '--quectel-set-radio-state'
      operations, which provide a way to FCC unlock EM120 modules in new Lenovo
      laptops.

 * Several other minor improvements and fixes.
```

```
ModemManager 1.18.6

 * Build:
   ** meson: fix gdbus code generation with meson 0.61.
   ** meson: fix i18n.merge_file() call with meson 0.60.

 * Core:
   ** plugin-manager: longer extra probing time when not using udev.
   ** filter: added 'option' driver name to the TTY driver filter list.
   ** udev: avoid wrong [0-9]* as shell globbing patterns.

 * QMI:
   ** Added new ID_MM_QMI_FACILITY_LOCK_QUERY_REDUCED udev flag to reduce the
      facility lock querying operations on devices that don't support it
      properly.
   ** Fixed capabilities and modes loading in multimode devices.
   ** Improved capabilities and modes handling for devices with 5G capabilities.

 * MBIM:
   ** Fixed logic reporting the number of remaining unlock attempts.

 * 3GPP interface:
   ** Fix memleak in network scan result building.

 * libmm-glib:
   ** Fix memleak when processing network scan results.

 * mmcli:
   ** Fixed operator mcc/mnc field names in human output of location info.

 * Plugins:
   ** telit: added port type hints for LE910C1-EUX.
   ** fibocom: added port type hints for FM101.
   ** foxconn: firmware info loading only of MBIM-QMUX is enabled in libqmi.
   ** quectel: ignore carrier revision.
   ** quectel: use +QGMR to set version.
   ** quectel: add new 'NAME' tag in device ids, so that firmware updates can be
      performed differently depending on the device name.

 * Several other improvements and fixes.
```


nickberry17:
Hey, thanks for sending this through!  

Normally we prefer the libraries committed first so that the build bots do not run into any issues, because they may build one commit at a time, _i.e._ if it tries to build ModemManager but a lib (_e.g._ libqmi) is too old, the buildbot may fail.  Granted, I do not see this being a problem in this particular instance, but just for consistency could you please rearrange the commits so that the libs are updated per the order of their dependency graph?

It should be:
libqrtr > libqmi > libmbim > modemmanager

Also, to add support for MM to actually use QRTR, `WITH_QRTR` needs to be configurable in the build options for MM.  Can you have a try at adding that?
https://github.com/openwrt/packages/blob/f2b03a0a296ad480959e2ec31b911d87b7a4cd79/net/modemmanager/Makefile#L44

See:
https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/blob/main/meson.build#L258

kuanyili:
@nickberry17 I checked again, and the commit is already in the order you suggested, could you have another look?

Currently, packages other than `libqrtr-glib` are kept using GNU autotools generated tarballs, not meson. So the default `modemmanager` build behavior is to use QRTR automatically since the library is available.

nickberry17:
Ah, yes the commits are in the correct order, sorry I forgot that Github reverses them when looking at a PR.  

I don't think  libqrtr-1.2.2 is able to be built without meson though.  I think the makefile needs to look something like this:

```
#
# Copyright (C) 2016 Velocloud Inc.
# Copyright (C) 2016 Aleksander Morgado <aleksander@aleksander.es>
#
#

include $(TOPDIR)/rules.mk

PKG_NAME:=libqrtr-glib
PKG_VERSION:=1.2.2
PKG_RELEASE:=$(AUTORELEASE)

PKG_SOURCE_PROTO:=git
PKG_SOURCE_URL:=https://gitlab.freedesktop.org/mobile-broadband/libqrtr-glib.git
PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
PKG_SOURCE_VERSION:=$(PKG_VERSION)
PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION)-$(PKG_SOURCE_VERSION).tar.gz

PKG_BUILD_DEPENDS:=meson/host

PKG_INSTALL:=1
PKG_BUILD_PARALLEL:=1

include $(INCLUDE_DIR)/nls.mk

define Package/libqrtr-glib
  SECTION:=libs
  CATEGORY:=Libraries
  DEPENDS:=+glib2
  TITLE:=Helper library to talk to QRTR enabled modems
  URL:=https://gitlab.freedesktop.org/mobile-broadband/libqrtr-glib
  LICENSE:=LGPL-2.0-or-later
  LICENSE_FILES:=COPYING.LIB
endef

define Package/libqrtr-glib/description
  Helper library talk to QRTR enabled modems.
endef

MESON_ARGS += \
	--prefix=/usr                       \
	--libdir=/usr/lib/x86_64-linux-gnu  \
	--buildtype=release                 \
	-Dgtk_doc=true                      \
	-Dintrospection=false

define Build/InstallDev
	$(INSTALL_DIR) $(1)/usr/include
	$(CP) \
		$(PKG_INSTALL_DIR)/usr/include/libqrtr-glib \
		$(1)/usr/include/

	$(INSTALL_DIR) $(1)/usr/lib
	$(CP) \
		$(PKG_INSTALL_DIR)/usr/lib/libqrtr-glib*.so* \
		$(1)/usr/lib/

	$(CP) \
		$(PKG_INSTALL_DIR)/usr/lib/pkgconfig/qrtr-glib.pc \
		$(1)/usr/lib/pkgconfig
endef

define Package/libqrtr-glib/install
	$(INSTALL_DIR) $(1)/usr/lib
	$(CP) \
		$(PKG_INSTALL_DIR)/usr/lib/libqrtr-glib*.so.* \
		$(1)/usr/lib/
endef

$(eval $(call BuildPackage,libqrtr-glib))

```

nickberry17 (approved):

nickberry17:
Ignore previous comment about meson, everything looks good!

Issue #17681, opened 2022-01-23
crowdsec/geoipupdate: enhancement request for offline database download


erdoukki:
Maintainer: Gérald Kerma / @erdoukki
Compile tested: (aarch64_cortex-a53, espressoBin, OpenWrt master and 21.02)
Run tested: (aarch64_cortex-a53, espressoBin, OpenWrt 21.02.x, tests done)

Description:

As refereed in [Idea/GeoIP: offline/update/share database](https://github.com/crowdsecurity/crowdsec/issues/1196)

> CrowdSec OpenWrt package is downloading the GeoIP databases at postinstall, this may be a huge download on some users ISP, but it also may not work at all in case users want to integrate the crowdsec package in the OpenWrt firmware image.

Thanks to @clayface the package geoipupdate is available since PR: https://github.com/openwrt/packages/pull/17171

The question here is like the same from https://github.com/crowdsecurity/crowdsec/issues/1196

Note: Sorry for crosslink between OpenWrt/CrowdSec but the both communities may be interested

erdoukki:
Build: ok
Install: ok

```
gke@PRECISION:~/DEVEL/OWRT/packages$ git push upstream
Énumération des objets: 16, fait.
Décompte des objets: 100% (16/16), fait.
Compression par delta en utilisant jusqu'à 8 fils d'exécution
Compression des objets: 100% (14/14), fait.
Écriture des objets: 100% (14/14), 37.45 Mio | 109.00 Kio/s, fait.
Total 14 (delta 2), réutilisés 0 (delta 0)
remote: Resolving deltas: 100% (2/2), completed with 2 local objects.
remote: warning: See http://git.io/iEPt8g for more information.
remote: warning: File net/crowdsec-geoipdatabase/files/GeoLite2-City_20220118/GeoLite2-City.mmdb is 69.78 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
remote: warning: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.
remote: 
remote: Create a pull request for 'geoipupdate' on GitHub by visiting:
remote:      https://github.com/erdoukki/packages/pull/new/geoipupdate
remote: 
To github.com:erdoukki/packages.git
 * [new branch]          geoipupdate -> geoipupdate
```

Is there any recommendation about maximum file size ?

Another potential issue is the license about redistribution of the DB, can anyone help to understand if there is any issue and if the package may be valid to be proposed upstream ?

[...]

Issue #17406, opened 2021-12-21
crowdsec-firewall-bouncer: remove crowdsec package dependency


erdoukki:
Maintainer: @erdoukki 
Environment: (OpenWrt 21.02)

Description:
https://forum.openwrt.org/t/crowdsec-initial-packages-v1-2-0-for-openwrt/102648/46

> using crowdsec-firewall-bouncer on many OpenWRT devices connected with my domain LAPI server (which collect many crowdsec machines, mostly nginx), it works great. Actually, crowdsec package is not mandatory for that usage, it would be great if it was not a dependency.

erdoukki:
Resolved by https://github.com/openwrt/packages/pull/17407

iamnague:
Hi @erdoukki, not tested yet,  don't you think you should also change /etc/firewall.cs file as it enables and restarts crowdsec service ?

erdoukki:
sure...
good advice, thanks.
I modify the script and send an update to this PR.

ne20002:
Do we have a chance to get this into 21.02 as well? The size of the GeoIp database downloaded with the main program is too much for my WRT3200ACM and I just need the baouncer anyway.

erdoukki:
> Do we have a chance to get this into 21.02 as well?
Sure, thanks @ne20002 to ask, it was just missed for 21.02.

> The size of the GeoIp database downloaded with the main program is too much for my WRT3200ACM and I just need the baouncer anyway.

I also plan to propose today, actually testing the fix
see https://github.com/openwrt/packages/issues/17681#issuecomment-1058958900

erdoukki:
> > Do we have a chance to get this into 21.02 as well?
> 
> Sure, thanks @ne20002 to ask, it was just missed for 21.02.

Done: https://github.com/openwrt/packages/pull/17986

ne20002:
Hi
Dependency is gone and space seems to be no problem now.

The installation ends with an error message.

![grafik](https://user-images.githubusercontent.com/39007279/156923655-aeb15bda-80d5-4ecc-8ca3-93dc5218eb10.png)

erdoukki:
two points here.
first:
crowdsec, because of cscli tool *is* required !
the crowdsec service may be disabled in your case.
I will revert the dependency remove !
Packaging only the cscli tool is not a good alternative but I will look also at it.

second:
your problem with ipset is an openWrt issue, maybe...
 not specific to CrowdSec.

ne20002:
[...]

PR #17989, opened 2022-03-05
crowdsec: new upstream, add plugins, add cli variant, and install fixes

erdoukki:
Maintainer: Gérald Kerma @erdoukki [gandalf@gk2.net](mailto:gandalf@gk2.net)
Compile tested: (mvebu, EspressoBin & EspressoBin-Ultra, OpenWrt Master & 21.02.x)
Run tested: (mvebu, EspressoBin & EspressoBin-Ultra, OpenWrt Master & 21.02.x, tests done)

Description:
 -  crowdsec: fix post install for offline mode and small devices
 -  crowdsec: new upstream and install fixes
 -  crowdsec: add notifications
 -  crowdsec: add cscli command line package only
 -  crowdsec: clean Makefile
 -  crowdsec: crowdsec-cli optimize package size
 -  crowdsec: define variants of crowdsec packages
 -  crowdsec: fix crowdsec plugins lib path in crowdsec default config
 - crowdsec: keep configs and data from the generic defaults
 - crowdsec: extend configs for user friendly generic usage
 - crowdsec: add & use external lib functions script
 - crowdsec: fix some copyright strings
 - crowdsec: fix BUILD_TIMESTAMP

Fixes:
 - [crowdsec-firewall-bouncer: remove crowdsec package dependency #17406](https://github.com/openwrt/packages/issues/17406)
 - [crowdsec/geoipupdate: enhancement request for offline database download #17681](https://github.com/openwrt/packages/issues/17681)
 - https://github.com/openwrt/packages/pull/17515#discussion_r827791506


erdoukki:
@jefferyto @PolynomialDivision 
Héllo,
can you look at this PR please ?
Thanks in advance

jefferyto:
Some general comments/questions:

* There are too many large changes here to review at once. With no high-level summary of what is being changed or for what purpose, it is hard to understand what is happening.
* Why are you switching back to using the upstream Makefile instead of using golang-package.mk?
* crowdsec vs crowdsec-cli: Why can't crowdsec depend on crowdsec-cli? Usually variants are used when each variant uses different compilation options (e.g. built with different SSL libraries), but in this case cscli is the same and the same files are duplicated in both packages.
* Why is the same code being run in both the uci-defaults script and the init.d script? If something needs to be done once, it should be in the uci-defaults script; if something needs to be done every time the service is started, it should be in the init.d script.
* Also on the config script(s):
  * There is no need to resave the default values back into the uci config, this just adds unnecessary writes to flash memory.
  * Modifying the config files in /etc/crowdsec also adds writes to flash memory, this is why packages generate config files in /var (/tmp).

I think some of the commits should be reorganized (dependent/atomic changes should happen in the same commit), but I think the above points should be addressed first.

erdoukki:
> Some general comments/questions:

Thanks, I will try to answer my best.

>     * There are too many large changes here to review at once. With no high-level summary of what is being changed or for what purpose, it is hard to understand what is happening.

Sorry, about this, I will make another reorganization of this PR.
- I mainly aggregate some bugfixes and users feedbacks from the Forum.
- Upstream version synchronize
- Rework the Makefile to add new plugins notifications
- some cleaning
- try a variant with the cscli only (which may be removed and pretty "useless")
- propose some OpenWrt specifics features and fixes for parameters and package install, upgrade

>     * Why are you switching back to using the upstream Makefile instead of using golang-package.mk?

External plugins, for notifications features are not built whith the default golang-package.mk, so I switch back to use the upstream Makefile.

>     * crowdsec vs crowdsec-cli: Why can't crowdsec depend on crowdsec-cli? Usually variants are used when each variant uses different compilation options (e.g. built with different SSL libraries), but in this case cscli is the same and the same files are duplicated in both packages.

Yes, cscli is always the same. It was proposed to get a smallest package for cli usage.
I may remove the crowdsec-cli variant which complexify too much for a very little use case.

>     * Why is the same code being run in both the uci-defaults script and the init.d script? If something needs to be done once, it should be in the uci-defaults script; if something needs to be done every time the service is started, it should be in the init.d script.

The package need some pre-config and network need also to be up.
I cannot use only uci-defaults because of internet access and network needs.
I can use init.d script only and have think also of an external libs for reusable functions, but may be it is a little too much.

>     * Also on the config script(s):
>       
>       * There is no need to resave the default values back into the uci config, this just adds unnecessary writes to flash memory.

I porpose this to be able to modify the defaults.
Do you mean to remove the defaults and only take care of defined values ?

>       * Modifying the config files in /etc/crowdsec also adds writes to flash memory, this is why packages generate config files in /var (/tmp).

I will revert this part and rethink how to minimize writes to /etc/ and flash memory and move all to /var/.
It may be need a patch on the upstream sources to be more efficient.

> I think some of the commits should be reorganized (dependent/atomic changes should happen in the same commit), but I think the above points should be addressed first.

I will redo another new rework on this PR.
And do the same with the CrowdSec-Firewall-Bouncer, also.

erdoukki:
.

BKPepe:
What are you talking about?

GeorgeSapkin:
No updates in over three years, merge conflicts and still marked as draft. Closing as stale. Feel free to reopen as necessary.

erdoukki:
Hello,


I have no time for now, but still planned to upstream for soon ( it is 
still in my TODO)...


I will reopen when I get some free/available time.


Thanks you

Gerald

Le 10/12/2025 à 17:03, George Sapkin a écrit :

PR #18001, opened 2022-03-06
chrony: add conflict between without NTS and NTS variant

BKPepe:
Maintainer: @mlichvar
Compile and run tested: N/A

Both packages provide the same files:
/usr/bin/chronyc
/usr/sbin/chronyd
/etc/chrony/chrony.conf
/etc/hotplug.d/iface/20-chrony
/etc/init.d/chronyd

They should not be installed side by side.


mlichvar:
LGTM. Thanks.

PR #18002, opened 2022-03-06
haproxy: Makefile polishing and add conflict to SSL/non-SSL variant

BKPepe:
Maintainer: @gladiac
Compile tested: Turris 1.1, powerpc8540, OpenWrt 21.02.2
Run tested: N/A

Description:

- There are two commits:
    - One commit tries to have Makefile in some better shape, so it can be understood faster. :-)
    - The second commit adds conflict between SSL and non-SSL variant


gladiac:
LGTM, thanks!

PR #18007, opened 2022-03-06
python-jsonschema: Update to 4.4.0

jefferyto:
Maintainer: @jmarcet 
Compile tested: armvirt-32, 2022-03-06 snapshot sdk
Run tested: armvirt-32 (qemu), 2022-03-06 snapshot

Description:


jefferyto:
This is basically #16863 (including the same setup.py patch) but updating to the latest version of jsonschema.

While I'm still not a fan of adding setup.py back in, I haven't had time to work on the Python package build system yet. I think it's better to use this workaround for now instead of blocking updates for this package.

PR #18008, opened 2022-03-06
seafile: Remove packages

jefferyto:
Maintainer: me
Compile tested: N/A
Run tested: N/A

Description:
I believe these packages should be removed (and [imported into the abandoned packages repo][1]) as Seafile Server and Seahub have been marked as broken for some time, and I do not believe I will have time to fix or update these packages in the foreseeable future.

[1]: https://github.com/openwrt/packages-abandoned/pull/22


jefferyto:
This only removes the core Seafile packages - I will follow up with a PR for other packages that only exist to support Seafile (libevhtp, Django packages) after this has been merged.

PR #18009, opened 2022-03-07
catatonit: add new package

oskarirauta:
podman uses catatonit now for container init, so new package for
catatonit is required.

### DIFF
--- a/utils/catatonit/Makefile
+++ b/utils/catatonit/Makefile
@@ -1,0 +1,43 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=catatonit
+PKG_VERSION:=0.1.7
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/openSUSE/catatonit.git
+PKG_SOURCE_DATE:=2022-03-07
+PKG_SOURCE_VERSION:=d8d72fea155c144ed3bf298a35a1aba5625a5656
+PKG_MIRROR_HASH:=5dfec105de2b1e674db55e12007aa66cf67769d38e3f294fbca54fc3e9b78674
+
+PKG_BUILD_PARALLEL:=1
+PKG_FIXUP:=autoreconf
+PKG_INSTALL:=1
+
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+PKG_LICENSE:=GPL-3.0-or-later
+PKG_LICENSE_FILES:=COPYING
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/catatonit
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=A container init that is so simple it's effectively brain-dead.
+  URL:=https://github.com/openSUSE/catatonit
+  DEPENDS:=
+endef
+
+define Package/catatonit/description
+  This is a rewrite of initrs in C, because we found that it is not
+  possible to statically compile Rust binaries without using musl.
+  That was, in turn, a reimplementation of other container inits like
+  tini and dumb-init.
+endef
+
+define Package/catatonit/install
+	$(INSTALL_DIR) $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/catatonit $(1)/usr/bin/
+endef
+
+$(eval $(call BuildPackage,catatonit))
--- a/utils/catatonit/test.sh
+++ b/utils/catatonit/test.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+case "$1" in
+	"catatonit")
+		test $(/usr/bin/catatonit -V | grep '^tini version' | wc -l) -gt 0
+		;;
+esac


oskarirauta:
I will push a new PR for podman soon after this is accepted with decency for catatonit.

utils/catatonit/Makefile
```diff
@@ -0,0 +1,43 @@
...
+
+PKG_NAME:=catatonit
+PKG_VERSION:=0.1.7
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/openSUSE/catatonit.git
+PKG_SOURCE_DATE:=2022-03-07
+PKG_SOURCE_VERSION:=d8d72fea155c144ed3bf298a35a1aba5625a5656
+PKG_MIRROR_HASH:=skip
```
BKPepe, line 11: This does not look good.

utils/catatonit/Makefile
```diff
@@ -0,0 +1,43 @@
...
+PKG_LICENSE_FILES:=COPYING
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/catatonit
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=A container init that is so simple it's effectively brain-dead.
+  URL:=https://github.com/openSUSE/catatonit
+  DEPENDS:=
```
BKPepe, line 28: It's weird, but there are no dependencies for catatonit?

utils/catatonit/Makefile
BKPepe, line 11: Why don't use tagged release?

utils/catatonit/Makefile
```diff
@@ -0,0 +1,43 @@
...
+PKG_SOURCE_DATE:=2022-03-07
+PKG_SOURCE_VERSION:=d8d72fea155c144ed3bf298a35a1aba5625a5656
+PKG_MIRROR_HASH:=skip
+
+PKG_BUILD_PARALLEL:=1
+PKG_FIXUP:=autoreconf
+PKG_INSTALL:=1
+
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+PKG_LICENSE:=GPL-2.0-or-later
```
BKPepe, line 18: It's actually GPL-3.0-or-later.
See: https://github.com/openSUSE/catatonit/blob/d8d72fea155c144ed3bf298a35a1aba5625a5656/COPYING

BKPepe (changes requested):

utils/catatonit/Makefile
oskarirauta: Sorry, I always compile it with tests here to get hash, because hash that I compile locally usually does not match with hash here. Just had to do some work between, so I did not have time to fix this until now.. Anyway, test build has finished and was cancelled due to this and reported correct hash and it's now in it's place.

utils/catatonit/Makefile
oskarirauta: AFAIK, no.

utils/catatonit/Makefile
oskarirauta: Hey, that's right. Sorry, my bad. Probably was a typo in the morning (had 5 minutes time before had to start work so it was a bit sloppy work) - fixing it now..

utils/catatonit/Makefile
oskarirauta: because in podman instructions (build script for catatonit in unpacked source), git tree is fetched instead of release; can't wonder if there is a reason they did it like that. Although, I didn't bother to check how close of latest commit is that release, nor how far their chosen commit is from current. Anyway, never touch a working system...

utils/catatonit/Makefile
oskarirauta: done..

utils/catatonit/Makefile
oskarirauta: It seems that there are not too many changes between 0.1.7 release and latest commit, but all those seem to be related to issues with autoconf files, so I suggest we go with them. Latest commit is dated in 2021, so most likely commit used by build script included with podman sources, is using the same version (without actually checking this..)

oskarirauta:
@BKPepe all checks have passed.

utils/catatonit/Makefile
oskarirauta: No, there are no dependencies.

oskarirauta:
What changes are still being requested?

utils/catatonit/Makefile
BKPepe: From my point of view, using tagged releases is preferred and this commit https://github.com/openSUSE/catatonit/commit/99bb9048f532257f3a2c3856cfa19fe957ab6cec could be backported manually, but this is fine as it is.

BKPepe:
Maybe I have one small thing. Would it be possible to add file ``test.sh`` like it is done for some packages here to check the version during runtime testing in Docker?

Otherwise, this looks good.

oskarirauta:
> Maybe I have one small thing. Would it be possible to add file `test.sh` like it is done for some packages here to check the version during runtime testing in Docker?
> 
> Otherwise, this looks good.

done..

BKPepe:
It seems that something is wrong with your's ``test.sh``, because CI fails with this output:
```
Testing package catatonit in version 0.1.7 from catatonit
Installing catatonit (0.1.7-1) to root...
Configuring catatonit.
Use package specific test.sh
Test failed
```

oskarirauta:
What is the purpose of test.sh - what should it do? (Meaning how can I test it..)

oskarirauta:
@BKPepe ping

I made changes to test.sh --

PR #18013, opened 2022-03-07
coova-chilli: remove kmod dep on binary package

f00b4r0:
Maintainer: @neheb @teslamint 

Description: Remove kmod dep on binary package

There is no reason for the kmod to depend on the binary package
itself, neither for building nor for installing.

That dependency prevents phase1 from building the kmod even though
support is enabled in the binary.


teslamint (approved):

f00b4r0:
If merged, it would be great if this could be backported to 21.02.

PR #18016, opened 2022-03-07
python-packages: Remove unused library packages

jefferyto:
Maintainer: me, @commodo, @cotequeiroz
Compile tested: N/A
Run tested: N/A

Description:
With the removal of Seafile, these library packages no longer have any in-repo users. They will be imported into the [abandoned packages repo][1].

[1]: https://github.com/openwrt/packages-abandoned/pull/24

* * *

So Seafile was removed in #18008. (Apologies @cotequeiroz I know you were interested in maintaining it, but I felt it was better to remove it rather than keep broken packages in the repo.)

These packages appear to be in the feed to support Seafile only, so with Seafile gone I am proposing to remove them:

Maintained by @commodo:
* django-appconf
* django-compressor
* django-jsonfield
* django-picklefield
* django-postoffice
* django-restframework39
* django-statici18n
* django1
* gunicorn
* ~openpyxl~ (not removed)
* ~pillow~ (not removed)
* ~pymysql~ (not removed)
* ~python-lxml~ (not removed)

Maintained by @cotequeiroz:
* django-formtools
* django-ranged-response
* django-simple-captcha
* django-webpack-loader
* pyjwt
* ~python-defusedxml~ (not removed)
* python-oauthlib
* python-qrcode
* python-requests-oauthlib

Maintained by @commodo and @cotequeiroz:
* ~python-et_xmlfile~ (not removed)
* ~python-jdcal~ (not removed)
* python-rcssmin

Maintained by me:
* django-jsonfield2

@commodo @cotequeiroz Please let me know if any should be kept in the repo (or if they can all be removed).

@BKPepe Are there any packages you would like to keep?

(Anyone else I should ask about keeping these packages?)


BKPepe:
Hi,

I'd say that package pillow is required for Home Assistant, which was not merged here, yet. I focused on higher-priority things.

While looking at the list and comparing it with Turris OS packages, I don't see any dependencies, so these packages could be removed.

jefferyto:
Thanks @BKPepe - I have removed pillow from the removal.

commodo:
from my libraries: i'd probably keep:

* `python-lxml` is an XML parser/lib that is mostly written in C/C++, and has regular updates; if anyone wants XML with OpenWrt + Python, it is one of the best; and it is pretty well maintained;
* `pymysql` - this one seems interesting to keep around, but since it's pure-python, it can be installed via pip (or manually) on system; not sure about whether to keep it in the feeds or not;  a bit neutral on this; it also doesn't get many updates;
* `openpyxl` is also interesting to keep, as it allows for Excel file manipulation; not sure how many use it though

the rest can be removed from my side;

commodo:
> Maintained by @commodo and @cotequeiroz:
> 
>     * python-et_xmlfile
> 
>     * python-jdcal
> 
>     * python-rcssmin

regarding these 3; if seafile goes away, these can definitely go away (from my side);

they are not well maintained, and exist only for seafile;

jefferyto:
Thanks @commodo - I have removed python-lxml, pymysql, openpyxl and its dependencies (python-defusedxml, python-et_xmlfile, python-jdcal) from the removal.

cotequeiroz:
I am OK with the removal of all seafile remnants I maintain.  Thanks again for taking care of it all.

jefferyto:
Thanks @cotequeiroz for your confirmation.

I will mark this PR as ready for review.

PR #18017, opened 2022-03-07
frr: update to 8.2

lucize:

Maintainer: me
Tested x86_64, ramips


lucize:
seems that archs doesn't work from a while

PR #18032, opened 2022-03-09
libtirpc: don't build host shared libraries

neheb:
Maintainer: @Andy2244

### DIFF
--- a/libs/libtirpc/Makefile
+++ b/libs/libtirpc/Makefile
@@ -28,7 +28,7 @@ define Package/libtirpc
 endef
 
 CONFIGURE_ARGS += --disable-gssapi
-HOST_CONFIGURE_ARGS += --disable-gssapi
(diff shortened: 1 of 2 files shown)


neheb:
```
configure: error: Missing library functions needed for IPv6.
```

what?

neheb:
testing on Fedora 35 (has host libtirpc) and Alpine Linux (no host libtirpc) and ldd shows this is working as intended. Wonder why it's failing here.

net/samba4/Makefile
```diff
@@ -5,6 +5,7 @@ PKG_NAME:=samba
 PKG_VERSION:=4.14.12
 PKG_RELEASE:=$(AUTORELEASE)
 
+
```
feckert, line 8: Is that intentional?

neheb: Yeah it's just a test commit to get it to build. samba4 relies on libtirpc/host I believe.

PR #18037, opened 2022-03-09
tvheadend: bind to LAN IP by default

M95D:
If config parameter is not set, tvheadend will bind to LAN IP address by default.
Fixes issue #16500 without requiring user modifications to config file. It's also more secure if firewall becomes disabled.

See https://github.com/openwrt/packages/pull/17896


BKPepe:
The commit should contain in the commit message that it was cherry-picked. That can be achieved by writing it manually, which you don't want or rather use ``git cherry-pick -x hash``

M95D:
Is it ok now?

PR #18038, opened 2022-03-09
podman: updates podman package

oskarirauta:
Description:

- add missing runtime depency catatonit ( #18009 )
 - removed SELinux variant, add config option for it instead, I do not believe we need variant of EVERY SELinux support capable software
 - add config option for iptables firewall setup in default cni network config, otherwise skip iptables part (part of nftables transition, use cni-plugins-nft for nftables fw support or better; forget about cni fw completely, instead use openwrt's own as that way your rules for containers do not disappear on firewall restart), adjusts depends accordingly

### DIFF
--- a/utils/podman/Makefile
+++ b/utils/podman/Makefile
@@ -36,36 +36,30 @@ define Download/default-policy
   HASH:=cddfaa8e6a7e5497b67cc0dd8e8517058d0c97de91bf46fff867528415f2d946
 endef
 
-define Package/podman/Default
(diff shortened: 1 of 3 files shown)


oskarirauta:
Pending for PR #18009 which contains required catatonit package.

oskarirauta:
@ping neheb

Discuss with me about these changes - or review, please...

oskarirauta:
@neheb ping

PR #18041, opened 2022-03-10
syslog-ng: update to version 3.36.1 and add test.sh

BKPepe:
Maintainer: me
Compile and run tested: Turris 1.1, powerpc_8540, OpenWrt 21.02.2

Release notes:
- https://github.com/syslog-ng/syslog-ng/releases/tag/syslog-ng-3.36.1

### DIFF
--- a/admin/syslog-ng/Makefile
+++ b/admin/syslog-ng/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=syslog-ng
-PKG_VERSION:=3.35.1
+PKG_VERSION:=3.36.1
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_MAINTAINER:=Josef Schlehofer <josef.schlehofer@nic.cz>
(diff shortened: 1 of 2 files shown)


admin/syslog-ng/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+syslog-ng --version | head -n 1 | grep "$2"
```
pprindeville, line 3: I would use:

```
syslog-ng --version | awk 'NR == 1 { print $2; }'
```

instead.

pprindeville (changes requested):
Minor change

admin/syslog-ng/test.sh
BKPepe: I run tested it on router, but it shows only number 3.

```
root@turris:~# syslog-ng --version | awk 'NR == 1 { print $2; }'
3
```

But, thanks for suggestion anyway. I think I will drop to add test.sh for now and push the update. I will look at this next week.

BKPepe:
Dropped the ``test.sh`` and merging.

admin/syslog-ng/test.sh
pprindeville: Sorry, wrote that too fast:

```
syslog-ng --version | awk "NR == 1 && \$0 =~ \"$2\" { print $0; }"
```

PR #18047, opened 2022-03-12
mtr: Enable JSON output

foxtrot:
Prior to commit e9a695b3 JSON output was present, but since disabling
jansson, the `--json` help text is present, but functionality is
missing.

Instead, split the MTR package into two new packages: `mtr-json` and `mtr-nojson`.

This commit also bumps the version of MTR to 0.95.

Signed-off-by: Marc Egerton <marc@malloc.me>

Maintainer: @jmccrohan 
Compile tested: ramips, 21.02.2
Run tested: ramips, 21.02.2

### DIFF
--- a/net/mtr/Makefile
+++ b/net/mtr/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mtr
-PKG_VERSION:=0.94
-PKG_RELEASE:=1
+PKG_VERSION:=0.95
+PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://www.bitwizard.nl/mtr/files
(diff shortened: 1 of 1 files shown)


foxtrot:
Looks like the tests failed due to some LUCI feed issue? I'm not sure what the cause of that is.

BKPepe:
You need to bump ``PKG_RELEASE`` in the Makefile.

foxtrot:
Oops, thanks. Should be fixed now :)

BKPepe (approved):

foxtrot:
> ... If that JSON output is urgently needed then it should be probably added to a new "mtr-full" variant.

I don't think this is a bad suggestion. I am happy to rework my patch to offer a "mtr-full" option instead, if that is more desirable.

jow-:
> I don't think this is a bad suggestion. I am happy to rework my patch to offer a "mtr-full" option instead, if that is more desirable.

I would much appreciate that :) Maybe call it `mtr-json` if the only difference to the non-full variant is the JSON support. That would be analogous to `nftables` vs. `nftables-json`.

foxtrot:
I've also taken the liberty to update the package to 0.95. Please let me know if there are further desired changes or anything I missed :)

jow-:
LGTM!

PR #18049, opened 2022-03-12
nano: add one even more feature rich variant

oskarirauta:
nano-plus was a very welcome addition, something similar to what I had used for years already; since I do some development natively, I really enjoy c/c++ syntax colouring. And why not some other added features as well, not maybe suitable for devices with minimal flash, but for those cases where _size does not matter_ ... So I integrated my personal version of package to nano package.

So here's nano-full with nearly all the bells and whistles, but still - biggest and most visible change to nano-plus, is syntax colouring.

Some features for quite obvious reasons (from my point of view) were left out, like mouse support. This PR is not critically important; not so sure if it's that useful either, could be just me who wants more of his favourite editor, well anyway, here it is, incase that there is some interest..

Maintainer:  Hannu Nyman / @hnyman 
Run tested: x86_64, master

### DIFF
--- a/utils/nano/Makefile
+++ b/utils/nano/Makefile
@@ -46,6 +46,12 @@ define Package/nano-plus
   VARIANT:=plus
 endef
 
+define Package/nano-full
+  $(call Package/nano/Default)
+  TITLE:=GNU nano text editor (most features, Unicode)
+  VARIANT:=full
+endef
+
 define Package/nano/description
   Nano is a small and simple text editor for use on the terminal.
 
@@ -63,7 +69,34 @@ define Package/nano-plus/description
   $(call Package/nano/description)
 endef
 
-ifeq ($(BUILD_VARIANT),plus)
+define Package/nano-full/description
+  nano-full - most features enabled, including syntax highlighting,
+  multibuffer, Unicode/UTF-8, help, justify, nanorc, some key bindings)
+
+  $(call Package/nano/description)
+endef
+
+ifeq ($(BUILD_VARIANT),full)
+# full variant with most features included
+  CONFIGURE_ARGS += \
+	--enable-help \
+	--enable-justify \
+	--enable-linenumbers \
+	--enable-multibuffer \
+	--enable-nanorc \
+	--enable-utf8 \
+	--enable-browser \
+	--enable-color \
+	--enable-comment \
+	--disable-extra \
+	--disable-histories \
+	--disable-libmagic \
+	--disable-mouse \
+	--disable-operatingdir \
+	--disable-speller \
+	--disable-tabcomp \
+	--disable-wordcomp
+else ifeq ($(BUILD_VARIANT),plus)
 # plus variant with more features included
   CONFIGURE_ARGS += \
 	--enable-help \
@@ -104,6 +137,13 @@ define Package/nano-plus/install
 	$(call Package/nano/install,$1)
 endef
 
+define Package/nano-full/install
+	$(call Package/nano/install,$1)
+	$(INSTALL_DIR) $(1)/etc $(1)/usr/share/nano
+	$(INSTALL_CONF) ./files/nanorc $(1)/etc/nanorc
+	$(CP) $(PKG_INSTALL_DIR)/usr/share/nano/*.nanorc $(1)/usr/share/nano/
+endef
+
 $(eval $(call BuildPackage,nano))
 $(eval $(call BuildPackage,nano-plus))
-
+$(eval $(call BuildPackage,nano-full))
--- a/utils/nano/files/nanorc
+++ b/utils/nano/files/nanorc
@@ -1,0 +1,2 @@
+## === Syntax coloring ===
+include "/usr/share/nano/*.nanorc"


utils/nano/Makefile
```diff
@@ -63,7 +69,36 @@ define Package/nano-plus/description
...
+	--enable-linenumbers \
+	--enable-multibuffer \
+	--enable-nanorc \
+	--enable-utf8 \
+	--enable-browser \
+	--enable-color \
+	--enable-comment \
+	--disable-extra \
+	--disable-histories \
+	--enable-justify \
```
hnyman, line 93: At least justify and linenumbers are defined twice...

oskarirauta: Oops..

oskarirauta: duplicate configure arguments removed.

oskarirauta: @hnyman ping

okibcn:
I did a test using this patch modifying the nano-plus profile:
```
diff --git a/utils/nano/Makefile b/utils/nano/Makefile
index 156af90da..f74d20fa1 100644
--- a/utils/nano/Makefile
+++ b/utils/nano/Makefile
@@ -44,6 +44,7 @@ define Package/nano-plus
   $(call Package/nano/Default)
   TITLE:=GNU nano text editor (more features, Unicode)
   VARIANT:=plus
+  DEPENDS:=+libncurses +zlib
 endef
 
 define Package/nano/description
@@ -72,17 +73,7 @@ ifeq ($(BUILD_VARIANT),plus)
   --enable-multibuffer \
   --enable-nanorc \
   --enable-utf8 \
-  --disable-browser \
-  --disable-color \
-  --disable-comment \
-  --disable-extra \
-  --disable-histories \
-  --disable-libmagic \
-  --disable-mouse \
-  --disable-operatingdir \
-  --disable-speller \
-  --disable-tabcomp \
-  --disable-wordcomp
+  --enable-color
 else
 # default tiny variant
   CONFIGURE_ARGS += \
@@ -96,8 +87,11 @@ CONFIGURE_VARS += \
   ac_cv_header_regex_h=no \
 
 define Package/nano/install
-  $(INSTALL_DIR) $(1)/usr/bin
-  $(CP) $(PKG_INSTALL_DIR)/usr/bin/$(PKG_NAME) $(1)/usr/bin/
+    [ $(BUILD_VARIANT) = plus ] && \
+    $(INSTALL_DIR) $(1)/usr/share/$(PKG_NAME) && \
+    $(CP) $(PKG_INSTALL_DIR)/usr/share/$(PKG_NAME)/* $(1)/usr/share/$(PKG_NAME); \
+    $(INSTALL_DIR) $(1)/usr/bin; \
+    $(CP) $(PKG_INSTALL_DIR)/usr/bin/$(PKG_NAME) $(1)/usr/bin
 endef
 
 define Package/nano-plus/install
```
It provides all the features mentioned here and it is only 49 KB more than the nano-plus. Maybe a nano-plusplus variant?
Untitled
Screenshot 2022-03-15 203421

okibcn:
By the way, shouldn't this nano-full be called nano-color? There are other interesting features disabled such as undo history, etc that are quite handy if you want to do some things. Try the set of features I did in the patch and see if that is better for a nano-full release.

okibcn:
I have compared the IPK size of the four variants (nano, nano-plus, nano-color, nano-full) coming from this patch:
```
diff --git a/utils/nano/Makefile b/utils/nano/Makefile
index 156af90da..97d626932 100644
--- a/utils/nano/Makefile
+++ b/utils/nano/Makefile
@@ -46,6 +46,19 @@ define Package/nano-plus
   VARIANT:=plus
 endef
 
+define Package/nano-color
+  $(call Package/nano/Default)
+  TITLE:=GNU nano text editor (more features, color, Unicode)
+  VARIANT:=color
+endef
+
+define Package/nano-full
+  $(call Package/nano/Default)
+  TITLE:=GNU nano text editor (all features, Unicode)
+  VARIANT:=full
+  DEPENDS:=+libncurses +zlib
+endef
+
 define Package/nano/description
   Nano is a small and simple text editor for use on the terminal.
 
@@ -58,12 +71,51 @@ endef
 
 define Package/nano-plus/description
   nano-plus - Additional features enabled, larger size than default nano.
-  (multibuffer, Unicode/UTF-8, help, justify, nanorc, some key bindings)
+  (multibuffer, Unicode/UTF-8, help, justify, nanorc)
+
+  $(call Package/nano/description)
+endef
+
+define Package/nano-color/description
+  nano-full - most features enabled, including syntax highlighting,
+  multibuffer, Unicode/UTF-8, help, justify, nanorc, comments)
+
+  $(call Package/nano/description)
+endef
+
+define Package/nano-full/description
+  nano-full - all features enabled, including syntax highlighting,
+  multibuffer, Unicode/UTF-8, help, justify, nanorc, some key bindings)
 
   $(call Package/nano/description)
 endef
 
-ifeq ($(BUILD_VARIANT),plus)
+ifeq ($(BUILD_VARIANT),full)
+# full variant with all features included
+  CONFIGURE_ARGS += \
+	--enable-all \
+	--enable-utf8 
+else ifeq ($(BUILD_VARIANT),color)
+# full variant with most features included
+  CONFIGURE_ARGS += \
+	--enable-help \
+	--enable-justify \
+	--enable-linenumbers \
+	--enable-multibuffer \
+	--enable-nanorc \
+	--enable-utf8 \
+	--enable-browser \
+	--enable-color \
+	--enable-comment \
+	--disable-extra \
+	--disable-histories \
+	--disable-libmagic \
+	--disable-mouse \
+	--disable-operatingdir \
+	--disable-speller \
+	--disable-tabcomp \
+	--disable-wordcomp
+else ifeq ($(BUILD_VARIANT),plus)
 # plus variant with more features included
   CONFIGURE_ARGS += \
 	--enable-help \
@@ -104,6 +156,19 @@ define Package/nano-plus/install
 	$(call Package/nano/install,$1)
 endef
 
+define Package/nano-full/install
+	$(call Package/nano/install,$1)
+	$(INSTALL_DIR) $(1)/etc $(1)/usr/share/nano
+	$(INSTALL_CONF) ./files/nanorc $(1)/etc/nanorc
+	$(CP) $(PKG_INSTALL_DIR)/usr/share/nano/* $(1)/usr/share/nano
+endef
+
+define Package/nano-color/install
+	$(call Package/nano-full/install,$1)
+endef
+
 $(eval $(call BuildPackage,nano))
 $(eval $(call BuildPackage,nano-plus))
+$(eval $(call BuildPackage,nano-color))
+$(eval $(call BuildPackage,nano-full))
``` 
The variant proposed by @oskarirauta is named `nano-color` and the variant `nano-full` has all the features plus utf8. I have compiled all four variants for the Lynksys E8450 (aarch64) and I got this:
```
164 -rw-r--r-- 1 oki oki 165493 Mar 16 13:54 nano-color_6.2-1_aarch64_cortex-a53.ipk
172 -rw-r--r-- 1 oki oki 175346 Mar 16 13:54 nano-full_6.2-1_aarch64_cortex-a53.ipk
120 -rw-r--r-- 1 oki oki 120529 Mar 16 13:53 nano-plus_6.2-1_aarch64_cortex-a53.ipk
 76 -rw-r--r-- 1 oki oki  73915 Mar 16 13:53 nano_6.2-1_aarch64_cortex-a53.ipk
```
Given the size of the packages, I believe a full featured nano is possible, adding interesting things such as syntax checker, infinite undo history, etc.

oskarirauta:
I'm perfectly ok with version by @okibcn  - syntax support is why I made my version of nano-full (or nano-color), so it's not all about binary size; - if @okibcn's version is considered as better proposal, or name nano-full doesn't describe it enough, I still think we should name it something else than nano-color - somehow it doesn't sound right - just in my opinion though, but wanted to bring it up without any better ideas on my behalf..

okibcn:
I agree with naming it `nano-full` if it is actually the full version with all the features. I just named your set of features `nano-color` just to have an idea of the size difference. 

So, if we agree on the full-featured nano, the final patch for nano should be something like this:
```makefile
#
# Copyright (C) 2007-2016 OpenWrt.org
#
# See /LICENSE for more information.
#

include $(TOPDIR)/rules.mk

PKG_NAME:=nano
PKG_VERSION:=6.2
PKG_RELEASE:=$(AUTORELEASE)

PKG_SOURCE_URL:=@GNU/nano
PKG_HASH:=2bca1804bead6aaf4ad791f756e4749bb55ed860eec105a97fba864bc6a77cb3

PKG_LICENSE:=GPL-3.0-or-later
PKG_LICENSE_FILES:=COPYING
PKG_MAINTAINER:=Hannu Nyman <hannu.nyman@iki.fi>
PKG_CPE_ID:=cpe:/a:gnu:nano

PKG_INSTALL:=1
PKG_BUILD_PARALLEL:=1

define Package/nano/Default
  SUBMENU:=Editors
  SECTION:=utils
  CATEGORY:=Utilities
  URL:=https://www.nano-editor.org/
  DEPENDS:=+libncurses
endef

define Package/nano
  $(call Package/nano/Default)
  TITLE:=GNU nano text editor (minimal features)
  VARIANT:=tiny
  DEFAULT_VARIANT:=1
endef

define Package/nano-plus
  $(call Package/nano/Default)
  TITLE:=GNU nano text editor (more features, Unicode)
  VARIANT:=plus
endef

define Package/nano-full
  $(call Package/nano/Default)
  TITLE:=GNU nano text editor (all features, Unicode)
  VARIANT:=full
  DEPENDS:=+libncurses +zlib
endef

define Package/nano/description
  Nano is a small and simple text editor for use on the terminal.

  Nano started as an enhanced clone of the Pico text editor.
  Nowadays Nano wants to be a generally useful editor with sensible
  defaults (linewise scrolling, no automatic line breaking).

  Nano is an official GNU package.
endef

define Package/nano-plus/description
  nano-plus - Additional features enabled, larger size than default nano.
  (multibuffer, Unicode/UTF-8, help, justify, nanorc)

  $(call Package/nano/description)
endef

define Package/nano-full/description
  nano-full - all features enabled, including syntax highlighting,
  multibuffer, Unicode/UTF-8, help, justify, nanorc, some key bindings)

  $(call Package/nano/description)
endef

ifeq ($(BUILD_VARIANT),full)
# full variant with all features included
  CONFIGURE_ARGS += \
	--enable-all \
	--enable-utf8
else ifeq ($(BUILD_VARIANT),plus)
# plus variant with more features included
  CONFIGURE_ARGS += \
	--enable-help \
	--enable-justify \
	--enable-linenumbers \
	--enable-multibuffer \
	--enable-nanorc \
	--enable-utf8 \
	--disable-browser \
	--disable-color \
	--disable-comment \
	--disable-extra \
	--disable-histories \
	--disable-libmagic \
	--disable-mouse \
	--disable-operatingdir \
	--disable-speller \
	--disable-tabcomp \
	--disable-wordcomp
else
# default tiny variant
  CONFIGURE_ARGS += \
	--enable-tiny \
	--enable-linenumbers \
	--disable-color \
	--disable-utf8
endif

CONFIGURE_VARS += \
	ac_cv_header_regex_h=no \

define Package/nano/install
	$(INSTALL_DIR) $(1)/usr/bin
	$(CP) $(PKG_INSTALL_DIR)/usr/bin/$(PKG_NAME) $(1)/usr/bin/
endef

define Package/nano-plus/install
	$(call Package/nano/install,$1)
endef

define Package/nano-full/install
	$(call Package/nano/install,$1)
	$(INSTALL_DIR) $(1)/etc $(1)/usr/share/nano
	$(INSTALL_CONF) ./files/nanorc $(1)/etc/nanorc
	$(CP) $(PKG_INSTALL_DIR)/usr/share/nano/* $(1)/usr/share/nano
endef

$(eval $(call BuildPackage,nano))
$(eval $(call BuildPackage,nano-plus))
$(eval $(call BuildPackage,nano-full))
```

And using the nanorc from above.

okibcn:
@hnyman, I have modified the Makefile above to remove the need for the zlib dependency. It was a bug in nano's build system. It was requesting zlib even when it is only needed for libmagic, and libmagic is desabled in all the variants (soon to be deprecated). In order to avoid this annoying bug, just adding --disable-libmagic to the configure args is enough to remove that dependency while keeping all the features.

I have also modified the app colors of nano in the nanorc above to match the OpenWrt blue/white colors.

okibcn:
The problem can be fixed with this patch. I have confirmed it works in three platforms: mips, arm, aarch64.

```diff
diff --git a/package/network/utils/iproute2/Makefile b/package/network/utils/iproute2/Makefile
index 55c00a0d6d..465f107575 100644
--- a/package/network/utils/iproute2/Makefile
+++ b/package/network/utils/iproute2/Makefile
@@ -75,7 +75,7 @@ $(call Package/iproute2/Default)
   VARIANT:=tcfull
   PROVIDES:=tc
   ALTERNATIVES:=400:/sbin/tc:/usr/libexec/tc-full
-  DEPENDS:=+kmod-sched-core +(PACKAGE_devlink||PACKAGE_rdma):libmnl +libbpf +libxtables +tc-mod-iptables
+  DEPENDS:=+kmod-sched-core +libbpf +libxtables +tc-mod-iptables +(PACKAGE_devlink||PACKAGE_rdma):libmnl
 endef
 
 define Package/tc-mod-iptables
```

hnyman:
I think that I will implement nano-full along the lines of the model by @okibcn 

But I will not include that large nanorc by default
The values in the possibly included nanorc should match the nano default, unless there is a clear reason.  E.g. tabsize is 8 by default, so let's not define it to 4.  
Similarly, the keymappings should not deviate from the nano defaults. 
Most people also use nano from a traditional SSH console, not from a non-standard console window in browser, so the surrounding colors are likelt balck and white ;-)
Likely I will just include a small/minimal nanorc.

Its location will also need to be investigated, preferably into /etc, and it should also be declared as conffiles, so that it (possible personal modifications) will be preserved in an OpenWrt sysupgrade.

okibcn:
@hnyman 

The problem you mention with coloring the environment doesn't exist. You can test it yourself doing this:
```bash
export TERM=vt100
ssh root@192.168.1.1   # IP of your device
```
In such a case (basic VT100 and the basic ssh protocol), nano uses black&while. The color adapts to the terminal as it is handled by ncurses. you can see the same when using midnight commander, htop, etc... so, that impact of coloring the app is not factor, and taking advantage of 256 color terminals is a plus.

Regarding keymap, it is ok to maintain the default keymap, however there should be a nanorc under /etc/ with the instructions to do that, making the life easier for the en user. It is typical to see that predefined commented out config files, such as vnstat.conf.

In that sense, I have prepared this [/etc/nanorc](https://github.com/openwrt/packages/files/8317960/nanorc.txt). just comment out those options you think should be disabled/enabled.

I have also prepared an additional [config.nanorc](https://github.com/openwrt/packages/files/8317983/config.nanorc.txt) syntax file adapted to the folders and type of config files OpenWrt has in /etc and several of its subfolders. Just place it in /user/share/nano along with the rest of the syntax files. You will see the difference when editing any config file, something quite common when dealing with OpenWrt.

hnyman:
@okibcn 
See #18126 of my proposed changes.

I did not use the "enable-all" config option you tries to use, as that has been removed in 2014.

I provided only a minimal nanorc, and also created a color scheme for the OpenWrt uci config files in /etc/config

oskarirauta:
@hnyman 

#18126 looks good to me..

hnyman:
Superseded by #18126

PR #18056, opened 2022-03-14
isc-dhcp: Update to 4.4.3

pprindeville:
Maintainer: me
Compile tested: x86_64, generic, HEAD (236c3ea730)
Run tested: same, installed and running on production router

Description:

### DIFF
--- a/net/isc-dhcp/Makefile
+++ b/net/isc-dhcp/Makefile
@@ -9,8 +9,8 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=isc-dhcp
 UPSTREAM_NAME:=dhcp
-PKG_VERSION:=4.4.1
-PKG_RELEASE:=20
+PKG_VERSION:=4.4.3
+PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_LICENSE:=BSD-3-Clause
 PKG_LICENSE_FILES:=LICENSE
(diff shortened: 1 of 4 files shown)


neheb:
Conflicts.

pprindeville:
Conflicts are resolved, but I need to find a way to pass `--disable-backtrace` to `bind/configure` when running on 32-bit ARM hardware and using MUSL.

neheb:
Strange that it's only 32-bit ARM. libunwind supports more platforms.

pprindeville:
> Strange that it's only 32-bit ARM. libunwind supports more platforms.

Maybe the problem isn't with isc-dhcp but with libunwind then?

Seems like the problem is seen elsewhere:

https://lore.kernel.org/all/20210103205534.GQ2997@scaer/T/
https://www.mail-archive.com/bind-users@lists.isc.org/msg30748.html

neheb:
Might as well disable backtrace then. Isn't it just a debugging feature?

pprindeville:
> Might as well disable backtrace then. Isn't it just a debugging feature?
> 
> edit: maybe this helps: https://github.com/openwrt/packages/blob/master/net/openvswitch/patches/0007-build-only-link-libopenvswitch-with-libunwind-libunb.patch

What's odd is that Bind builds without disabling backtracing.

PR #18061, opened 2022-03-15
isc-dhcp: detect whether ddns transfers are IPv4 only

pprindeville:
Maintainer: me
Compile tested: x86_64, generic, HEAD (236c3ea730)

Description:

If named is configured to not listen on any IPv6 interfaces,
then we should run 'nsupdate' with the '-4' argument.

Also:

* cleanup RFC-1918 address detection;
* don't generate PTR records for domain entries that aren't
  RFC-1918 addresses or these will generate NOTAUTH failures

We're assuming that we're doing DNS split-horizon and that
internal addresses aren't routable.


pprindeville:
@dwfreed Ended up using your `sed` suggestion.

PR #18063, opened 2022-03-15
netdata: disable LZ4 compression support

BKPepe:
Maintainer: me and @diizzyy
Compile tested: Turris 1.1, powerpc_8540, OpenWrt master
Run tested: N/A

Description:

To avoid unnecessary dependency, let's disable it for now.
LZ4 can be also used for DB engine and HTTPS

Fixes:
Package netdata is missing dependencies for the following libraries:
liblz4.so.1

Reported by @neheb in https://github.com/openwrt/packages/pull/16339#issuecomment-1067518013

### DIFF
--- a/admin/netdata/Makefile
+++ b/admin/netdata/Makefile
@@ -57,6 +57,7 @@ CONFIGURE_ARGS += \
 	--without-libcap \
 	--disable-https \
 	--disable-dbengine \
+	--disable-compression \
 	--disable-plugin-nfacct \
 	--disable-plugin-freeipmi \
 	--disable-plugin-cups \


neheb:
Does this get rid of the zlib dependency too?

BKPepe:
By looking at Netdata's [configure.ac](https://github.com/netdata/netdata/blob/908d369195205d25b4599be4a7ef89b804b172fc/configure.ac) zlib should be included and no, it does not get rid of the zlib dependency. That could be disabled by ``--without-zlib``.

> You are building without zlib. zlib allows netdata to transfer a lot less data with web clients. It should be enabled.

diizzyy:
Thanks :-)

PR #18065, opened 2022-03-15
python-asn1crypto: drop package

commodo:
Maintainer: me
Compile tested: n/a
Run tested: n/a

---------------------------

    I can't seem to see any package that needs it.
    This was added for cryptography, since it was needed up to version 2.7
    asn1-crypto doesn't have a user since commit 9d892e3cf88bd7c5f2c61117df2f3c2fd6c0e960
    So, remove it.

Abandoned packaged PR: https://github.com/openwrt/packages-abandoned/pull/23


commodo:
@jefferyto    fyi

commodo:
@BKPepe    i forgot to ask: any users in Turris for this package?

if no objections, i will merge this in 2 days

BKPepe:
@commodo Thanks for asking. We don't include and use this package in Turris OS packages feed nor depend on it in Turris OS, so I would say that it is enough to remove it. It might be possible that a few users have this installed, but who knows if they are using this, and just for a few people, I don't think that it is a good idea to keep this package. They can compile it locally and on the other hand, we will get faster builds. :-)

BKPepe (approved):

commodo:
i think we got faster builds after Seafile went away :P 
but yeah, i'm trying to remove some of my packages that have bit-rotted around; 

i may try to remove others in the meantime;
there was a time when i took some packages under my wing (because they were abandoned by others), but after a while it looks to have more sense to start removing them (as the natural things in life do);

PR #18073, opened 2022-03-15
vim: remove vim/host

neheb:
The only use of it is for xxd. tools/xxd was added in base so this can
go.

Maintainer: @ratkaj


BKPepe:
Nitpick: Just for sake of completness, would it be possible to include the reference in your commit description to the commit in the base?

neheb:
Added,

PR #18076, opened 2022-03-16
chrony: enable LTO

stintel:
Maintainer: @mlichvar
Compile tested: OpenWrt master r19124-832b90216f on qoriq
Run tested: OpenWrt master r19124-832b90216f on qoriq

Description:
Build chrony with LTO


mlichvar:
Makes sense to me. On mips it seems to decrease the binary size by about 6%.

PR #18089, opened 2022-03-18
openvswitch/ovn/spice-protocol/xl2tpd: bump versions

yousong:
Maintainer: me
Compile tested: x86/64 on QEMU
Run tested: x86/64 on QEMU

Description:

PR #18096, opened 2022-03-18
Openvswitch: bump to 2.15.3

stintel:
Maintainer: @yousong
Compile tested: OpenWrt master r19136-83ca16fc43 on x86/64
Run tested: OpenWrt master r19136-83ca16fc43 on x86/64

Description:
Bump openvswitch to 2.15.3. While there are newer major versions available, I've heard reports of problems with 2.16 and as I have no time to properly debug those for now, let's start with just bumping to the latest patch version of the current minor release.


yousong:
Hi, I was about to bump openvswitch version to 2.17 (https://github.com/openwrt/packages/pull/18089).  Is the issue still there for that version.  And is there a link to the issue?

yousong:
> Unfortunately I never received any useful information. I'd prefer to have this PR merged first instead of bumping to 2.17 directly.

Indeed, I wonder how we will move on if the issue is an unknown one.  But I am fine with 2.15.3.  Can you also backport the bump to openwrt-22.03 branch here?

yousong:
The other thing is, do you mind taking ownership of openvswitch and ovn packages here?  I do not work on them for a while now and my dev machine for the work is not regularly available.

stintel:
> Indeed, I wonder how we will move on if the issue is an unknown one. But I am fine with 2.15.3. Can you also backport the bump to openwrt-22.03 branch here?

We can do 2.17 right after 2.15.3, but in case something pops up it can simply be reverted.


> The other thing is, do you mind taking ownership of openvswitch and ovn packages here? I do not work on them for a while now and my dev machine for the work is not regularly available.

Unfortunately my focus has shifted.

PR #18126, opened 2022-03-21
nano: provide nano-full with most nano features enabled (also uci syntax coloring)

hnyman:
Provide a new variant, nano-full, that enables almost all functionality of nano. Only libmagic file type detection
has been left out.

Ship with a minimal /etc/nanorc that the user can modify.

Provide color highlighting for the uci config files. I drafted a minimal color scheme for them.

Maintainer: me

### DIFF
--- a/utils/nano/Makefile
+++ b/utils/nano/Makefile
@@ -46,6 +46,12 @@ define Package/nano-plus
   VARIANT:=plus
 endef
 
+define Package/nano-full
+  $(call Package/nano/Default)
+  TITLE:=GNU nano text editor (all features, Unicode)
+  VARIANT:=full
+endef
+
 define Package/nano/description
   Nano is a small and simple text editor for use on the terminal.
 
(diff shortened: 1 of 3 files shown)


okibcn:
@hnyman 

Having a full featured is great, and you are right, --enable-all was deprecated since that it is now the default nano feature set. 

I would place a more user friendly nanorc file, and a better config file that could take into consideration .conf files and other config variants inside a typical etc folder in OpenWrt. 

If you want to optimize the end size, I would sacrifice some syntax files that will be unlikely used in a router (haskell, ada, c, povray,...) leaving space for useful files such as the proposed [/etc/nanorc](https://github.com/openwrt/packages/files/8317960/nanorc.txt) (uncommenting only the features you want to keep) and syntax config file such as [config.nanorc](https://github.com/openwrt/packages/files/8317983/config.nanorc.txt) that could handle more types of config files used in OpenWrt or by the packages used in OpenWrt. The end size would be slightly smaller, while boosting the ease of customization and config files coloring.

PR #18108, opened 2022-03-19
poemgr: add package

blocktrron:
Maintainer: me
Compile tested: ramips-mt7621
Run tested: ramips-mt7621 / USW Flex

Description:

```
Add the poemgr package. This package is required to control the PSE
controller for the PoE outputs on the Ubiquiti UniFi Switch Flex.

Signed-off-by: David Bauer <mail@david-bauer.net>
```

PR #18131, opened 2022-03-22
poemgr: add package

blocktrron:
Maintainer: me
Compile tested: ramips-mt7621
Run tested: ramips-mt7621 / USW Flex

Description:

```
Add the poemgr package. This package is required to control the PSE
controller for the PoE outputs on the Ubiquiti UniFi Switch Flex.

```

### DIFF
--- a/utils/poemgr/Makefile
+++ b/utils/poemgr/Makefile
@@ -1,0 +1,40 @@
+# SPDX-License-Identifier: GPL-2.0-only
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=poemgr
+PKG_SOURCE_DATE:=2022-03-19
+PKG_SOURCE_VERSION:=77395242f9b3cb6ef003c5ed74f9b5be78404450
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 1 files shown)


utils/poemgr/Makefile
```diff
@@ -0,0 +1,40 @@
...
+
+define Package/poemgr
+  SECTION:=utils
+  CATEGORY:=Utilities
+  DEPENDS:=+libuci +libjson-c
+  TITLE:=Utility to control PoE ports on the UniFi Flex switch
+endef
+
+define Package/poemgr/conffiles
+/etc/config/usteer
```
galeksandrp, line 27: Nope.

BKPepe: This package is already in OpenWrt master branch. See https://github.com/openwrt/packages/pull/18108 So, it needs to be fixed there and here as well.

PolynomialDivision: It is already fixed upstream.

BKPepe: Yes, but the commit is not backported here, so this makes this PR as stalled.

blocktrron: fixed

PR #18132, opened 2022-03-22
afalg_engine,gost_engine: adapt to new openssl UCI config

cotequeiroz:
Maintainer: me & @6ec123321 (Artur Petrov, turned into @ghost by github)
Compile tested: mediatek/mt7622/Linksys E8450, mvebu/cortexa9/Linksys WRT3200ACM, master
Run tested: same as above

Description:
Since https://github.com/openwrt/openwrt/commit/0134f845dab95e0e7f885f4e227d9a41aedf909d (openssl: configure engines with uci), OpenSSL uses UCI to enable and disable engines.  Each engine now uses its own config file.  This adapts the engine packages.

This depends on openwrt/openwrt#9535.

### DIFF
--- a/libs/gost_engine/Makefile
+++ b/libs/gost_engine/Makefile
@@ -13,29 +13,27 @@ PKG_LICENSE:=OpenSSL
 
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/cmake.mk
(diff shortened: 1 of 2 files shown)


cotequeiroz:
This breaks builds with SDK.  I'll need to move `engine.mk` to `$(INCLUDE_DIR)` before I can make this work.

PR #18133, opened 2022-03-22
miniupnpd: Bump to latest git source

msylgj:
Maintainer: 
Compile tested: OpenWrt master SNAPSHOT, build on x86/64
Run tested: OpenWrt master SNAPSHOT, test on armv8(rockchip R4S)

Description:
Bump to latest git source, follow upstream change to modify the init script: use scripts to create nft tables for firewall rules. The ruleset look like this:
```
table inet filter {
	chain forward {
		type filter hook forward priority -25; policy accept;
		jump miniupnpd
	}

	chain miniupnpd {
		iif "pppoe-wan" th dport 6881 @nh,128,32 0x<ipv4 public ip> @nh,72,8 0x6 accept
		iif "pppoe-wan" th dport 6881 @nh,192,128 0x<ipv6 public ip> @nh,48,8 0x6 accept
		iif "pppoe-wan" th dport 6881 @nh,128,32 0x<ipv4 public ip> @nh,72,8 0x11 accept
		iif "pppoe-wan" th dport 6881 @nh,192,128 0x<ipv6 public ip> @nh,48,8 0x11 accept
		iif "pppoe-wan" th dport 8083 @nh,128,32 0x<ipv4 public ip> @nh,72,8 0x6 accept
		iif "pppoe-wan" th dport 8083 @nh,192,128 0x<ipv6 public ip> @nh,48,8 0x6 accept
	}

	chain prerouting {
		type nat hook prerouting priority dstnat; policy accept;
		jump prerouting_miniupnpd
	}

	chain postrouting {
		type nat hook postrouting priority srcnat; policy accept;
		jump postrouting_miniupnpd
	}

	chain prerouting_miniupnpd {
		iif "pppoe-wan" @nh,72,8 0x6 th dport 6881 dnat ip to 192.168.2.3:6881
		iif "pppoe-wan" @nh,72,8 0x11 th dport 6881 dnat ip to 192.168.2.3:6881
		iif "pppoe-wan" @nh,72,8 0x6 th dport 8083 dnat ip to 192.168.2.3:8083
	}

	chain postrouting_miniupnpd {
	}
}
```
Remaining problem: The 'inet' family of nftables con't set nat rule both ipv4 and ipv6. But I con't use ip&ip6 table for rule sets.(miniupnpd doesn't support, throw errors: send_batch failed -4) So we can see only ipv4 redirect rules in the nat table


stintel:
* I'm not the maintainer
* please don't bump versions and add unrelated changes in a single commit
* I'm pretty sure these scripts are not needed, and you're not explaining anything about why they might be in the commit message either
So based on that: NACK

msylgj:
- I'm not the maintainer
 ---Sorry about that.There have not maintainer info in the Makfile, and I saw your commits.
- please don't bump versions and add unrelated changes in a single commit
- I'm pretty sure these scripts are not needed, and you're not explaining anything about why they might be in the commit message either
 ---Because of this commit(https://github.com/miniupnp/miniupnp/commit/3129683cb335fa8559928d4b8557248f3bc1bc51), we must use scripts to create rule tables start from v2.3.0. If I want to bump to new version, I must add scripts file, modify the init script, that's why they are in a single commit. 
But I think we could use upstream repo's scripts directly but not custom one. I will remove them and include scripts from source code in Makefile, maybe waiting for answer about what I called 'remaining problem'. So what should I do now will be better? Close this pr temporary, or can you add a 'work needed' label plz?

stintel:
Thanks for clarifying. I'll try to have a closer look soonish but can't promise anything. To be honest, I'm somewhat tempted to stop using miniupnpd altogether if scripts are going to be a hard requirement.

msylgj:
> Thanks for clarifying. I'll try to have a closer look soonish but can't promise anything. To be honest, I'm somewhat tempted to stop using miniupnpd altogether if scripts are going to be a hard requirement.

My pleasure. Perhaps I have a mistake about this, hope we can use it simply

stintel:
@neheb can you please stop merging things that clearly have changes that require explanation in commit messages and not mention any of it in there?

BKPepe:
Reverted - https://github.com/openwrt/packages/commit/0bc25077b37d74065fd107dc9f85fbb11ccba56b.

net/miniupnpd/patches/201-change-default-chain-rule-to-accept.patch
```diff
@@ -0,0 +1,13 @@
+--- a/netfilter_nft/scripts/nft_init.sh
```
BKPepe, line 1: 1. Who is the author of this change?
2. Why is this change required?
- There is a missing commit description!
3. Was this discussed with upstream?
- Is there any pending pull request? If not, submit it there first.

msylgj: 1. I do this change(Author: ZiMing Mo <msylgj@immortalwrt.org>)
2. We can find the reason in https://github.com/miniupnp/miniupnp/blob/master/miniupnpd/netfilter_nft/README.md
```
Custom Chains
NFTables is very flexible but it comes with some restrictions because of that. If there is a second filter chain than all packets that were passed before with the miniupnpd chain will be reevaluated. This also means that if the chain is a drop chain you loose the packets. In that case you really want to use a custom chain and jump to it in your filter chain. miniupnpd should save all accept rules in that custom chain. For NAT it is the same, a second chain will also evaluate the packets again and therefore it is possible that a second SNAT or DNAT is performed.

The following is used in miniupnpd for a table setup but it can be customized
```
There have a filter chain in table inet fw4(created by firewall4), it's a drop chain. if the filter chain of miniupnpd is also a drop chain, all packets will be loose(In my testing, it happend). So i do the change for miniupnpd work on firewall4.
3. Based on the description above, this change doesn't need an upstream PR or discuss, in my mind.

BKPepe: All of this, should be included in the patch itself. Then it is a question, why it should not be a default for miniupnpd and we do need to modify this in OpenWrt.

tiagogaspar8:
Hi @msylgj, what issues are still present for this to merge?
For the PCP issue, I'm already figuring it out with upstream [here](https://github.com/miniupnp/miniupnp/issues/611)
Besides that is there anything left to do? 😄

msylgj:
> Hi @msylgj, what issues are still present for this to merge? For the PCP issue, I'm already figuring it out with upstream [here](https://github.com/miniupnp/miniupnp/issues/611) Besides that is there anything left to do? 😄

We have some different views of the patch about 'nft_init' script. (You can see it in the review conversation above.)

tiagogaspar8:
Yeah, I'm waiting for upstream to fix master and then I'll make the pull request.

tiagogaspar8:
Hi,
So, it seems upstream is taking its sweet time to merge the fix to the master branch. Until they have the time we have 3 options:
1. Wait it out
2. use the fix branch as the source for miniupnpd
3. use the latest commit in the master branch and backport ourselves the patch, later, when merged in upstream master this can be dropped when upgraded

What are your thoughts @stintel @BKPepe?

PR #18142, opened 2022-03-24
poemgr: fix conffiles path

stintel:
Maintainer: @blocktrron 
Compile tested: NA
Run tested: NA


stintel:
Strange breakage ... I don't see how this can be caused by my changes.

BKPepe (approved):

PR #18149, opened 2022-03-25
uacme: always include ualpn

lucize:
with mbetls update we can have ualpn on any variant

Maintainer: me
Tested: ramips


lucize:
guess something is wrong with the CI ?

neheb:
```
Package uacme-ualpn is missing dependencies for the following libraries:
libev.so.4
```

PR #18152, opened 2022-03-26
libmbim: switch to meson build tools

Linaro1985:

Maintainer: Nicholas Smith <nicholas@nbembedded.com>

Compile tested:
on amd64 for mipsel_24kc on https://git.openwrt.org/openwrt/openwrt.git commit c7bcbcd49280a79b287cc072cd0ca7de777a7ac4

Run tested:
Zbtlink ZBT-WG3526

Description:
Using https://gitlab.freedesktop.org/mobile-broadband/libmbim.git to download the source code.
Enabled lto and additional gcc flags for perfomance and less size
Modified to use meson as upstream has abandoned autotools.
Removed BUILD_PARALLEL options. These are default with ninja/meson.

### DIFF
--- a/libs/libmbim/Makefile
+++ b/libs/libmbim/Makefile
@@ -8,27 +8,29 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libmbim
-PKG_VERSION:=1.26.2
+PKG_SOURCE_VERSION:=1.26.2
 PKG_RELEASE:=$(AUTORELEASE)
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_SOURCE_URL:=https://www.freedesktop.org/software/libmbim
-PKG_HASH:=10c77bf5b5eb8c92ba80e9b519923ad9b898362bc8e1928e2bc9a17eeba649af
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://gitlab.freedesktop.org/mobile-broadband/libmbim.git
+PKG_MIRROR_HASH:=47eb6511818271be66336fb67ec97de7925119b09ddbd0ffd06b3f76893a461e
 
 PKG_MAINTAINER:=Nicholas Smith <nicholas@nbembedded.com>
 
 PKG_INSTALL:=1
-PKG_BUILD_PARALLEL:=1
 
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/nls.mk
+include $(INCLUDE_DIR)/meson.mk
 
-CONFIGURE_ARGS += \
-	--disable-static \
-	--disable-gtk-doc \
-	--disable-gtk-doc-html \
-	--disable-gtk-doc-pdf \
-	--disable-silent-rules
+TARGET_CFLAGS += -ffunction-sections -fdata-sections -fno-merge-all-constants -fmerge-constants
+TARGET_LDFLAGS += -Wl,--gc-sections
+
+MESON_ARGS += \
+	-Dintrospection=false \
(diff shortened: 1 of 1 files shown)


Linaro1985:
Ping @nickberry17

aleksander0m:
Same comment as in https://github.com/openwrt/packages/pull/18153#issuecomment-1079961909. When using meson, I would switch to using the git protocol to download a specific tag, instead of relying on the gitlab-generated tarball.

Linaro1985:
@aleksander0m thanks! switched to git.

aleksander0m:
LGTM

libs/libmbim/Makefile
```diff
@@ -11,24 +11,28 @@ PKG_NAME:=libmbim
...
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://gitlab.freedesktop.org/mobile-broadband/libmbim.git
+PKG_SOURCE_VERSION:=$(PKG_VERSION)
+PKG_MIRROR_HASH:=47eb6511818271be66336fb67ec97de7925119b09ddbd0ffd06b3f76893a461e
 
 PKG_MAINTAINER:=Nicholas Smith <nicholas@nbembedded.com>
 
 PKG_INSTALL:=1
-PKG_BUILD_PARALLEL:=1
+PKG_BUILD_DEPENDS:=python3/host
```
neheb, line 22: Why?

libs/libmbim/Makefile
```diff
@@ -11,24 +11,28 @@ PKG_NAME:=libmbim
 PKG_VERSION:=1.26.2
 PKG_RELEASE:=$(AUTORELEASE)
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_SOURCE_URL:=https://www.freedesktop.org/software/libmbim
-PKG_HASH:=10c77bf5b5eb8c92ba80e9b519923ad9b898362bc8e1928e2bc9a17eeba649af
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://gitlab.freedesktop.org/mobile-broadband/libmbim.git
+PKG_SOURCE_VERSION:=$(PKG_VERSION)
```
neheb, line 16: The interesting part here is that you can set this to 1.26.2 and remove PKG_VERSION as the latter is set from the former.

libs/libmbim/Makefile
Linaro1985: Meson build requires python3 https://gitlab.freedesktop.org/mobile-broadband/libmbim/-/blob/main/meson.build#L48
Is this dependency not needed in this case?

libs/libmbim/Makefile
Linaro1985: Thanks. Should I replace PKG_VERSION with PKG_SOURCE_VERSION and remove line 16? I relied on the Makefile of other packages (for example libqrtr-glib)

libs/libmbim/Makefile
neheb: I don't think it's needed here. Try removing and see if it builds.

libs/libmbim/Makefile
neheb: I'd remove PKG_VERSION and put it in PKG_SOURCE_VERSION. Keep in mind the latter is needed.

Linaro1985: Ok, done.

libs/libmbim/Makefile
aleksander0m: @Linaro1985 could you open a new PR to remove that dep?

aleksander0m: Ah! great, thanks

PR #18153, opened 2022-03-26
libqrtr-glib: Makefile polishing

Linaro1985:
Signed-off-by: Maxim Anisimov [maxim.anisimov.ua@gmail.com](mailto:maxim.anisimov.ua@gmail.com)

Maintainer: Nicholas Smith <nicholas@nbembedded.com>

Compile tested:
on amd64 for mipsel_24kc on https://git.openwrt.org/openwrt/openwrt.git commit c7bcbcd49280a79b287cc072cd0ca7de777a7ac4

Run tested:
Zbtlink ZBT-WG3526

Description:
Enabled lto and additional gcc flags for perfomance and less size
Removed BUILD_PARALLEL options. These are default with ninja/meson

PR #18154, opened 2022-03-26
libqmi: switch to meson build tools

Linaro1985:

Maintainer: Nicholas Smith <nicholas@nbembedded.com>

Compile tested:
on amd64 for mipsel_24kc on https://git.openwrt.org/openwrt/openwrt.git commit c7bcbcd49280a79b287cc072cd0ca7de777a7ac4

Run tested:
Zbtlink ZBT-WG3526

Description:
Using https://gitlab.freedesktop.org/mobile-broadband/libqmi.git to download the source code.
Enabled lto and additional gcc flags for perfomance and less size.
Added support for selecting qmi message collection.
Selected basic qmi message collection by default (modemmanager requires it).
Modified to use meson as upstream has abandoned autotools.
Removed BUILD_PARALLEL options. These are default with ninja/meson.

### DIFF
--- a/libs/libqmi/Makefile
+++ b/libs/libqmi/Makefile
@@ -11,17 +11,22 @@ PKG_NAME:=libqmi
 PKG_VERSION:=1.30.4
 PKG_RELEASE:=$(AUTORELEASE)
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_SOURCE_URL:=https://www.freedesktop.org/software/libqmi
-PKG_HASH:=00d7da30a4f8d1185f37cba289cfaf1dfcd04a58f2f76d6acfdf5b85312d6ed6
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://gitlab.freedesktop.org/mobile-broadband/libqmi.git
+PKG_SOURCE_VERSION:=$(PKG_VERSION)
+PKG_MIRROR_HASH:=537eae29c36aba9757afd86e48b91c37c3fe3232037ad11fdd426297f6040a6b
 
 PKG_MAINTAINER:=Nicholas Smith <nicholas@nbembedded.com>
 
 PKG_INSTALL:=1
-PKG_BUILD_PARALLEL:=1
+PKG_BUILD_DEPENDS:=python3/host
 
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/nls.mk
+include $(INCLUDE_DIR)/meson.mk
+
+TARGET_CFLAGS += -ffunction-sections -fdata-sections -fno-merge-all-constants -fmerge-constants
+TARGET_LDFLAGS += -Wl,--gc-sections
 
 define Package/libqmi/config
   source "$(SOURCE)/Config.in"
@@ -59,26 +64,31 @@ define Package/libqmi-utils/description
   Utils to talk to QMI enabled modems
 endef
 
-CONFIGURE_ARGS += \
-	--disable-static \
-	--disable-gtk-doc \
-	--disable-gtk-doc-html \
-	--disable-gtk-doc-pdf \
-	--disable-silent-rules \
-	--enable-firmware-update \
-	--without-udev \
-	--without-udev-base-dir
+MESON_ARGS += \
+	-Dudev=false \
+	-Dintrospection=false \
+	-Dman=false \
+	-Dbash_completion=false \
+	-Db_lto=true
 
 ifeq ($(CONFIG_LIBQMI_WITH_MBIM_QMUX),y)
-	CONFIGURE_ARGS += --enable-mbim-qmux
+	MESON_ARGS += -Dmbim_qmux=true
 else
-	CONFIGURE_ARGS += --disable-mbim-qmux
+	MESON_ARGS += -Dmbim_qmux=false
 endif
 
 ifeq ($(CONFIG_LIBQMI_WITH_QRTR_GLIB),y)
-	CONFIGURE_ARGS += --enable-qrtr
+	MESON_ARGS += -Dqrtr=true
 else
-	CONFIGURE_ARGS += --disable-qrtr
+	MESON_ARGS += -Dqrtr=false
+endif
+
+ifeq ($(CONFIG_LIBQMI_COLLECTION_MINIMAL),y)
+	MESON_ARGS += -Dcollection=minimal
+else ifeq ($(CONFIG_LIBQMI_COLLECTION_BASIC),y)
+	MESON_ARGS += -Dcollection=basic
+else
+	MESON_ARGS += -Dcollection=full
 endif
 
 define Build/InstallDev
@@ -99,12 +109,15 @@ define Build/InstallDev
 endef
 
 define Package/libqmi/install
-	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DIR) \
+		$(1)/usr/lib \
+		$(1)/usr/libexec
+
 	$(CP) \
 		$(PKG_INSTALL_DIR)/usr/lib/libqmi*.so.* \
 		$(1)/usr/lib/
 
-	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/lib/qmi-proxy $(1)/usr/lib/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/libexec/qmi-proxy $(1)/usr/libexec/
 endef
 
 define Package/qmi-utils/install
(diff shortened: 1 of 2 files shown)


Linaro1985:
Ping @nickberry17

aleksander0m:
Same comment as in https://github.com/openwrt/packages/pull/18153#issuecomment-1079961909. When using meson, I would switch to using the git protocol to download a specific tag, instead of relying on the gitlab-generated tarball.

Linaro1985:
@aleksander0m thanks! switched to git.

aleksander0m:
> Selected basic qmi message collection by default (modemmanager requires it).

No, default should be FULL, or otherwise we'll get tons of complaints.

aleksander0m:
>  Enabled lto and additional gcc flags for perfomance and less size.

Open question about the need for this in here: https://github.com/openwrt/packages/pull/18167#issuecomment-1080448972

libs/libqmi/Makefile
```diff
@@ -59,26 +64,31 @@ define Package/libqmi-utils/description
...
 endif
 
 ifeq ($(CONFIG_LIBQMI_WITH_QRTR_GLIB),y)
-	CONFIGURE_ARGS += --enable-qrtr
+	MESON_ARGS += -Dqrtr=true
+else
+	MESON_ARGS += -Dqrtr=false
+endif
+
+ifeq ($(CONFIG_LIBQMI_COLLECTION_MINIMAL),y)
```
aleksander0m, line 86: I'm not sure about having separate config fields for these, as they're mutually exclusive.
Can we not have a single `CONFIG_LIBQMI_COLLECTION` field and then the user can select between "minimal", "basic" and "full"?

Linaro1985: > No, default should be FULL, or otherwise we'll get tons of complaints.

Do you mean the reduced functionality of the library? But as I know basic is enougth for modemmanager. There is no any packages which also uses libqmi. OpenWrt works on different devices. Most of them have a limited amount of resources. This makes things a little better.

aleksander0m: > But as I know basic is enough for modemmanager.

Yes, "basic" is what MM expects. But there may be others integrating other custom applications out there that rely on qmicli commands not included in the basic collection. I'm not fully convinced that we should use that as default. @nickberry17 what's your take on this?

aleksander0m: > I didn't find better solutions. It would be great if someone could point me a better way.

Switching to a single "string" type config option could be a way to do that, with "full" as default. And then explain that libqmi provides 3 different collections by default, "minimal", "basic" and "full".

The name of the collection is not limited to those 3 in libqmi, because users could also use a custom different collection e.g. named "something-else", as long as `data/qmi-collection-something-else.json` file exists in the tree.

libs/libqmi/Makefile
```diff
@@ -11,17 +11,22 @@ PKG_NAME:=libqmi
...
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://gitlab.freedesktop.org/mobile-broadband/libqmi.git
+PKG_SOURCE_VERSION:=$(PKG_VERSION)
+PKG_MIRROR_HASH:=537eae29c36aba9757afd86e48b91c37c3fe3232037ad11fdd426297f6040a6b
 
 PKG_MAINTAINER:=Nicholas Smith <nicholas@nbembedded.com>
 
 PKG_INSTALL:=1
-PKG_BUILD_PARALLEL:=1
+PKG_BUILD_DEPENDS:=python3/host
```
aleksander0m, line 22: Why is this a required change now?

aleksander0m: I'm asking because I tried a build with these patches and it broke because of some failure building `python3/host`

Linaro1985: Meson build requires python3 https://gitlab.freedesktop.org/mobile-broadband/libqmi/-/blob/main/meson.build#L48
It worked for me. I will check it again.

Linaro1985: > I'm asking because I tried a build with these patches and it broke because of some failure building `python3/host`

All build checks have just passed successfully. No problem.

libs/libqmi/Makefile
Linaro1985: > Switching to a single "string" type config option could be a way to do that

But there is no choice in this case. The user will select a collection by typing its name manually. This may result in an typo error. There are three collections. It is quite logical to have a choice between them. It's not that hard to add a 4th choice, where the user himself can specify the collection manually. I'd also like to hear @nickberry17 opinion.

neheb: This can be simplified to a single line.

libs/libqmi/Makefile
aleksander0m: glib2 requires meson and it doesn't depend on `python3/host`. @neheb is there any rationale here to require this?

aleksander0m:
@neheb sorry, I'm not very convinced about this merge you did here, there are still open points here I think, and @nickberry17 didn't even talk yet :/

libs/libqmi/Makefile
aleksander0m: So now we have 3 different boolean options that are mutually exclusive. Is that really a thing? Shouldn't we at least try to e.g. error out somewhere if the user sets 2 booleans to TRUE or something?

aleksander0m:
> > Selected basic qmi message collection by default (modemmanager requires it).
> 
> No, default should be FULL, or otherwise we'll get tons of complaints.

I'm quite against this change. Using basic is reducing functionality of the library and the qmicli a lot. ModemManager is not the only user of libqmi.

libs/libqmi/Makefile
aleksander0m: @Linaro1985 could you post a new PR removing the explicit `python3/host` dep?

Linaro1985:
> ModemManager is not the only user of libqmi.

At the moment, these "users" are not in the source code of the OpenWrt repositories.

libs/libqmi/Makefile
aleksander0m: "minimal" is contained in "basic". "basic" is contained in "full".

Selecting this config wouldn't make any sense, and it would completely ignore the FULL request and only enable the MINIMAL collection:
```
CONFIG_LIBQMI_COLLECTION_MINIMAL=y
CONFIG_LIBQMI_COLLECTION_BASIC=n
CONFIG_LIBQMI_COLLECTION_FULL=y
```
There has to be a better way to handle this... or maybe there isn't?

nickberry17:
@aleksander0m @neheb @Linaro1985 
Hey!  I won't be able to test anything for the rest of the week because I am currently on an island with a pretty slow internet connection (but an internet connection nonetheless, thanks to ModemManager and friends!)  I was looking into migrating these packages to Meson a few weeks ago but at the time I did not see a way to support all of the build time options.  Has that changed in the latest release? (sorry, haven't had a chance to look at the release notes yet).  I'll comment on what I can for now, but don't let me hold up an obviously good merge

aleksander0m:
> @aleksander0m @neheb @Linaro1985 Hey! I won't be able to test anything for the rest of the week because I am currently on an island with a pretty slow internet connection (but an internet connection nonetheless, thanks to ModemManager and friends!) I was looking into migrating these packages to Meson a few weeks ago but at the time I did not see a way to support all of the build time options. Has that changed in the latest release? (sorry, haven't had a chance to look at the release notes yet). I'll comment on what I can for now, but don't let me hold up an obviously good merge

:+1:

libs/libqmi/Makefile
nickberry17: @aleksander0m is there any harm in passing all of these at build time?  Because if that is okay, then one thing we could do is just make each selection progressively select all the other ones that are a subset of it using `select`.  I haven't tested this but it looks like it may work.

For example:
https://github.com/openwrt/packages/blob/38f01ad2c9738c6c5a5dbda85019bb9cf5bc10f8/utils/lcd4linux/Config.in#L220

```
config LCD4LINUX_CUSTOM_DRIVER_picoLCDGraphic
	bool
	prompt "picoLCDGraphic"
	select LCD4LINUX_CUSTOM_NEEDS_libgd
	select LCD4LINUX_CUSTOM_NEEDS_libusb
```

@neheb  what do you think?

aleksander0m: auto-selecting CONFIG_LIBQMI_COLLECTION_MINIMAL if CONFIG_LIBQMI_COLLECTION_BASIC is selected, and auto-selecting CONFIG_LIBQMI_COLLECTION_BASIC if CONFIG_LIBQMI_COLLECTION_FULL is selected, would make a lot of sense config-wise.

Then, the Makefile could look for those options in reverse order, first check for FULL, then check for BASIC, and then check for MINIMAL. Right now it's looking at those options in the reverse order.

aleksander0m: @Linaro1985 what do you think of that suggestion?

PR #18158, opened 2022-03-26
[openwrt-22.03] xray-core: Update to 1.5.4

1715173329:
Maintainer: me
Compile tested: -
Run tested: redmi ax6s

Description:
Backported from:
- #18113


1715173329:
> https://github.com/openwrt/packages/runs/5711331203?check_suite_focus=true#step:4:26
> Building gnunet xray-core horst

Looks like it needs a rebase.

Issue #18155, opened 2022-03-26
mwan3: setname 'mwan3_sticky_ipv4_default_rule_v4' is longer than 31 characters


RasimKalimullin:
Maintainer: @feckert
Environment: mediatek, Belkin rt3200, snapshot)

Description:
Fw4, empty mwan3_rules chain
```
Chain mwan3_rules (1 references)
 pkts bytes target     prot opt in     out     source               destination         
```
because
```
Sat Mar 26 11:35:04 2022 user.err mwan3-init[15552]: set_sticky_ipset (default_rule_v4): ipset v7.15: Error in line 2: Syntax error: setname 'mwan3_sticky_ipv4_default_rule_v4' is longer than 31 characters
Sat Mar 26 11:35:04 2022 user.err mwan3-init[15552]: set_user_rules (ipv4): iptables-restore v1.8.7 (nf_tables): setname `mwan3_sticky_ipv4_default_rule_v4' too long, max 31 characters. Error occurred at line: 36 Try `iptables-restore -h' or 'iptables-restore --help' for more information.
```
mwan3 version: 2.11.0

jamesmacwhite:
By default, the default_rule_v4 doesn't have sticky enabled, so I assume this had been enabled in your own configuration?

You are correct though the chain names do exceed 31 characters. I believe this is due to the different chain names created in 2.11, before they were prefixed as `mwan3_sticky_v4` `mwand_sticky_v6` but now it's `mwan3_sticky_ipv4` and `mwan3_sticky_ipv6`.

Combined with the default rule, `mwan3_sticky_ipv4_` (18 characters) and the rule name `default_rule_v4` (15), exceeds the limit by 2 characters.

I believe rule names remain fixed at 15 characters, so it means the mwan3 rule chain prefix can not exceed 16 characters to stay within the limit, relative to a rule name.

feckert:
@jamesmacwhite Thanks for this spotting and sorry, I didn't think of that at all and it slipped through my fingers!

There are 2 characters too many now because of **ip** :-(
I would suggest that we have to check if this restriction still exist in iptables?
That would be the simpleset change to change the max char from 31 to 33 

I don't want to revert the change, because the source code it is now generic.
Everywhere ipv4 or ipv6 is used now.

jamesmacwhite:
[...]

PR #18172, opened 2022-03-28
mwan3: fix sticky chain length

feckert:
Maintainer: me
Run tested: x86_64, APU3, OpenWrt master and Openwrt 22.03

Description:
This should fixes issue #18155 
@RasimKalimullin please test this change and give me feedback if this fixes your issue
@jamesmacwhite I have decided to change `sticky` with `rule`.
From my point of view, this has less of an impact. If I had changed `mwan3` to `mw3` as you suggested then I would have to do that everywhere. That would be a pretty big change in the code.

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.11.0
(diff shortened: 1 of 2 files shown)


jamesmacwhite:
@feckert That's fine, the main thing is reducing any prefix by two characters so it never exceeds 16 characters, so your suggestions works fine and doesn't require a code wide change.

PR #18173, opened 2022-03-28
vala: update to version 0.56.0 (LTS version)

BKPepe:
Maintainer: N/A
Compile tested on Turris Omnia, OpenWrt master, mvebu/cortex-a9

According, to the project website [1], we were not using long-term
version, but stable one. Let's use LTS version.

Changelog can be found on their GitLab [2].

[1] https://wiki.gnome.org/Projects/Vala
[2] https://gitlab.gnome.org/GNOME/vala/raw/0.56/NEWS


neheb:
Last I remember updating major vala versions I broke some packages depending on it. Would be good to test compilation of all vala/host dependant packages.

neheb:
I tested all except lcdringer. Compiles.

PR #18187, opened 2022-03-30
acme: switch from iptables to nft

tohojo:
Use nft instead of iptables to open port 80 in the firewall when getting a
cert. Since nft doesn't allow deleting a rule by its contents, capture and
save the handle when creating the rule, and use that to delete.

-------

Maintainer: me
Compile tested: N/A
Run tested: nope

### DIFF
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -21,6 +21,7 @@ NGINX_WEBSERVER=0
 UPDATE_NGINX=0
 UPDATE_UHTTPD=0
 UPDATE_HAPROXY=0
+NFT_HANDLE=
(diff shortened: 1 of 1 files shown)


champtar:
Maybe we can use procd firewall objects (exemples https://github.com/search?p=2&q=org%3Aopenwrt+json_add_array+firewall&type=Code) ?
Never used it myself but that would be the cleanest

tohojo:
Ah, oops, it's missing a `$` before the opening parenthesis :)

neheb:
Any update?

tohojo:
Heh, seems I forgot to actually update the PR, sorry about that...

PR #18195, opened 2022-03-31
kea: bump to 2.0.2

stintel:
Maintainer: @rosysong
Compile tested: OpenWrt master on mvebu/cortexa53
Run tested: OpenWrt master on mvebu/cortexa53


neheb:
```
2022-03-31T03:59:29.6309996Z checking OpenSSL version... LibreSSL 3.4.2
2022-03-31T03:59:29.8268333Z configure: error: missing EVP entry for SHA-2
```

stintel:
I really don't get how this doesn't build ... in my log I see this:

```
checking OpenSSL version... LibreSSL 3.4.2
checking support of SHA-2... yes
```

As I do not have LibreSSL installed on my host, it must be using the version from OpenWrt tools, so why does this build fine for me but not in CI?

neheb:
You have OpenSSL installed?

stintel:
> You have OpenSSL installed?

I do, but the configure clearly shows it's detecting LibreSSL 3.4.2. OpenSSL is 1.1.1n...

stintel:
@tiagogaspar8 any reason why you enabled `--enable-generate-messages` in 03cf09333254fa9f58de1d35556a3286849137ff?

neheb:
I think the reason it was failing is because —with-openssl was specified with a wrong path.

neheb:
Cool.

PR #18215, opened 2022-03-31
domoticz: update to 2022.1

lucize:
change to codeload
backport python 3.10 plugin loading

Maintainer: @dwmw2 
Tested x86_64

### DIFF
--- a/utils/domoticz/Makefile
+++ b/utils/domoticz/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=domoticz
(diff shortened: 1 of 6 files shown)


jefferyto:
There has been a number of [bug fixes](https://github.com/domoticz/domoticz/search?o=desc&q=python&s=committer-date&type=commits) since the stable ABI patch, you may want to backport some of them and/or continue to follow their progress (since it has only been a few weeks since the change).

lucize:
yes, I'll see about them
the compile crash is related to too many cpu and not enough memory

PR #18228, opened 2022-04-03
[master] adguardhome: Patch quic-go to fix compilation on Go 1.18

jamesmacwhite:

Maintainer: @dobo90 @alaviss 
Compile tested: mvebu and bcm53xx OpenWrt master
Run tested: SNAPSHOT r19355-80f79beb95 bcm53xx/generic

Description:

With the help of @alaviss regarding a suitable patch, we propose this patch to fix the current build failure of the adguardhome package occurring on master.

Since the Go 1.18 update, the adguardhome package has been broken due to a compilation failure. This is due to the quic-go component present in 0.107.5 using an older version that does not have Go 1.18 support. Support for Go 1.18 was added in quic-go later. This patch bumps the component to a version that does have support.

~~This is requirement for master and openwrt-22.03. openwrt-21.02 remains unaffected as the Go version remains at 1.17.~~

I have also compiled tested 0.108.0b5 which builds OK with Go 1.18 but as it's currently BETA, this patch allows the stable release to still be available. Although it's a good sign that the 0.108.x can build fine without any patches, but in the meantime this patch is needed to fix the build error.


jamesmacwhite:
Looks like it's just master that needs this patch, the 22.03 branch is still using Go 1.17.8 so the patch isn't needed.

jamesmacwhite:
Run tested the package on an R7000 running SNAPSHOT r19355-80f79beb95, working fine:

```
root@OpenWrt:~# opkg install /tmp/adguardhome_0.107.5-2_arm_cortex-a9.ipk
Installing adguardhome (0.107.5-2) to root...
Configuring adguardhome.
root@OpenWrt:~# AdGuardHome --version
AdGuard Home, version v0.107.5
```

SiNONiMiTY:
If this is good, please merge :)

alaviss (approved):
LGTM, not tested on my side (yet), but I think it's a fairly safe change.

PR #18230, opened 2022-04-03
ubnt-manager/prometheus-node-exporter-lua: add airos monitoring

PolynomialDivision:
Add ubnt-manager for managing airos devices. Based on that package ubnt-manager-collector collects statistics.

### DIFF
--- a/utils/prometheus-node-exporter-lua/Makefile
+++ b/utils/prometheus-node-exporter-lua/Makefile
@@ -168,6 +168,17 @@ define Package/prometheus-node-exporter-lua-textfile/install
 	$(INSTALL_BIN) ./files/usr/lib/lua/prometheus-collectors/textfile.lua $(1)/usr/lib/lua/prometheus-collectors/
 endef
 
+define Package/prometheus-node-exporter-lua-ubnt-manager
+  $(call Package/prometheus-node-exporter-lua/Default)
+  TITLE+= (ubnt-manager collector)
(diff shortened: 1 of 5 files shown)


PolynomialDivision:
@champtar

champtar:
In the Prometheus makefile you are missing `call BuildPackage` and you need to bump the version

PolynomialDivision:
Thanks. Fixed it.

aparcar:
Looks fine to me, overall I recommend rewriting the whole Lua code with https://github.com/jow-/ucode

PolynomialDivision:
Thanks for your reviews. I would like to merge this first and then go on rewriting again.

It was suggested that I use:
- snmp instead of ssh (I am unsure if I would like to do this)
- rewrite with ucode

Ubnt-manager should not only be used for "monitoring". We focus on automation of tasks and that's why I focus in creating some app, that configures ubiquiti devices automatically.

aparcar:
If this is useful please merge, I'd recommend a combination of SNMP and Ansible. That approach is less invasive (compared to frequent SSH logins) and doesn't store all credential on an Internet facing device.

PolynomialDivision:
For me it is useful, so I will merge. Let's see how it will go on. :) Thanks for your feedback.

PolynomialDivision:
I am very sorry. I was very sure I fixed it. o.O
https://github.com/openwrt/packages/pull/18234

PR #18231, opened 2022-04-04
node: Major update from v14 to v16

nxhack:
Maintainer: me @ianchi 

Description:
Major update from v14 to v16

This update also changes npm from v6 to v8.
This change also requires node module packages to be modified.
Each package will be updated later.

### DIFF
--- a/lang/node/Makefile
+++ b/lang/node/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=node
-PKG_VERSION:=v14.18.3
+PKG_VERSION:=v16.14.2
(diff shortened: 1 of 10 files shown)


nxhack:
@openwrtfirmwarecostum 
For 21.02, the following patch is required.
https://github.com/openwrt/packages/blob/openwrt-21.02/lang/node/patches/999-fix_building_with_system_c-ares_on_Linux.patch

However, I intend to keep v14 for 21.02.

For 22.03 I will go with v16.

openwrtfirmwarecostum:
thanks for the patches. but for openwrt 22.03 it looks like it will take a long time to release there is a lot of work on 22.03

PR #18253, opened 2022-04-06
node-javascript-obfuscator: Support for npm@8

nxhack:
Maintainer: @zokl

Description:

With the upgrade of node.js to version 16, the npm version will also change to version 8.
This fix is to support npm@8. npm@6 can also build without problems.

The modification method is different from other node modules.
The reason is due to the npm@8 issue.
https://github.com/npm/cli/issues/4027


zokl:
Thank you @nxhack for the modification. 👍👍👍

PR #18254, opened 2022-04-06
node-serialport-bindings: Support for npm@8

nxhack:
Maintainer: me 

Description:
With the upgrade of node.js to version 16, the npm version will also change to version 8.
This fix is to support npm@8. npm@6 can also build without problems.


nxhack:
It builds well on my side, but it is failing in CI...

nxhack:
https://github.com/openwrt/packages/runs/5843706905?check_suite_focus=true#step:5:53
```
USER root
```
Isn't this an incorrect setting?
npm refuses to run as root, so this seems to be an error.

neheb:
I believe it runs as root because it's a docker container.

PR #18262, opened 2022-04-06
strongswan: do not force to use iptable-legacy

feckert:
Compile tested: x86_64, APU3, latest OpenWrt master
Run tested: no

Description:
The default firewall is the fw4, which uses nft. In order to not
install the legacy implementation when installing strongswan, the build
system should decide which firewall backend to use.

While we are at it, I have also added the dependency packages for IPV6.

https://github.com/openwrt/openwrt/pull/9648


pprindeville (approved):
Looks good to me.  Has it been tested?

feckert:
@pprindeville I found an [article](https://thermalcircle.de/doku.php?id=blog:linux:nftables_demystifying_ipsec_expressions) on this topic that is worth reading. 
maybe we should extend the script to use nft right away without *tables-nft

pprindeville:
> @pprindeville I found an [article](https://thermalcircle.de/doku.php?id=blog:linux:nftables_demystifying_ipsec_expressions) on this topic that is worth reading.
> maybe we should extend the script to use nft right away without *tables-nft

I'd rather confirm this first, merge it, and then have it as a checkpoint for rolling back if the above is a-bridge-too-far...

Thermi:
I am not aware of any development for a replacement or addition that uses nft instead.

feckert:
@Thermi I started a discussion about this topic in the strongswan [Discussion-Page](https://github.com/strongswan/strongswan/discussions/1019) on github.

Issue #16963, opened 2021-10-23
aria2: unable to start up


gw826943555:
Maintainer: @Imre Kaloz <kaloz@openwrt.org>, @Hsing-Wang Liao <kuoruan@gmail.com>
Environment: latest OpenWrt snapshot 

Description:

aria2 is unable to start up. 
![image](https://user-images.githubusercontent.com/22762473/138545116-6fdf0f13-4dc3-4cd1-ba8a-42cb76f0892a.png)
The aria2.conf.main is empty.
![image](https://user-images.githubusercontent.com/22762473/138545077-ee0425aa-dedd-4787-8083-d65c30cb9f94.png)
I tried factory reset and failed.

Fil010:
Hello greetings

doing what I mentioned, let it run but on this side I have had problems to download 
```
`2021-12-09` 20:42:40.394553 [INFO] [Context.cc:179] <<--- --- --- ---
2021-12-09 20:42:40.394589 [INFO] [Context.cc:180]   --- --- --- ---
2021-12-09 20:42:40.394605 [INFO] [Context.cc:181]   --- --- --- --->>
2021-12-09 20:42:40.394620 [INFO] [Context.cc:182] aria2 1.36.0
2021-12-09 20:42:40.394642 [INFO] [Context.cc:183] gcc 11.2.0
  built by  x86_64-pc-linux-gnu
  targeting x86_64-openwrt-linux-gnu
  on        Dec  5 2021 17:49:14
2021-12-09 20:42:40.394662 [INFO] [Context.cc:184] Linux 5.10.84 #0 SMP Sun Dec 5 17:49:14 2021 x86_64
2021-12-09 20:42:40.394680 [INFO] [Context.cc:185] zlib/1.2.11 OpenSSL/1.1.1l
2021-12-09 20:42:40.394701 [INFO] [Context.cc:186] Logging started.
2021-12-09 20:42:40.394717 [DEBUG] [Context.cc:215] Not setting rlimit NO_FILE: 1024 >= 1024
2021-12-09 20:42:40.394730 [INFO] [SocketCore.cc:1659] Checking configured addresses
2021-12-09 20:42:40.395257 [INFO] [SocketCore.cc:1710] Not considered: 127.0.0.1
2021-12-09 20:42:40.395281 [INFO] [SocketCore.cc:1707] Found configured address: 172.20.0.43
[...]

PR #18270, opened 2022-04-08
aria2: fix aria2.init re-mount issue

Narakuku:
The original procd_add_jail_mount "$config_file" will cause the init script unable to start the aria2 process.

Maintainer: me
Compile tested: (Linux MIPS, OpenWrt 21.02.1)
Run tested: (Linux MIPS, OpenWrt 21.02.1)

Description:
Re-mount `$config_file` inside `$config_dir` in jail will cause all files unreadable, thus unable to start the service, besides the second `procd_add_jail_mount_rw "$dir" "$config_dir"` already includes the `$config_file`.

Also, it is necessary to read-only all certificate files, so update `procd_add_jail_mount "$config_file"` to `procd_add_jail_mount "$ca_certificate" "$certificate" "$rpc_certificate" "$rpc_private_key"`.

### DIFF
--- a/net/aria2/Makefile
+++ b/net/aria2/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=aria2
 PKG_VERSION:=1.36.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://github.com/aria2/aria2/releases/download/release-$(PKG_VERSION)/
--- a/net/aria2/files/aria2.init
+++ b/net/aria2/files/aria2.init
@@ -341,7 +341,7 @@ aria2_start() {
 		procd_set_param user "$user"
 
 	procd_add_jail "$NAME.$section" log
-	procd_add_jail_mount "$config_file"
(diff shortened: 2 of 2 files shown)


leonghui:
Fixed #16963

Run tested: mvebu/Linksys WRT1900AC v2 OpenWrt SNAPSHOT r19417

Narakuku:
@leonghui Thank you! Exactly what is the issue and how I solved it!

But I wonder how come the package that I installed from latest OpenWrt 21.02.1 still in the wrong format. 

Nobody ever committed the fix?

leonghui:
This PR was created against the `master` branch (which is used for OpenWrt SNAPSHOT and future branches).

If the issue is also affecting 21.02, the same change will need to be backported to the `openwrt-21.02` branch (usually in a separate PR). Once the PR has been accepted and merged, the change will be picked up by the buildbots before the package gets rebuilt and updated.

I noticed that `openwrt-22.03` has branched, so the fix will need to be ported there eventually.

The review process will take time (testers and reviewers volunteer their time), so please be patient and continue enjoying OpenWrt!

Narakuku:
@leonghui 
Alright, we sit tight and assess! Got it!

BKPepe (changes requested):
``PKG_RELEASE`` needs to be bumped in the Makefile.

Acris:
Hi @Narakuku , could you please update `PKG_RELEASE` in the Makefile to merge this pull request?

Narakuku:
@Acris @BKPepe Am I doing this correctly?

PR #18272, opened 2022-04-08
uacme: add libev dependency to uacme-ualpn

cotequeiroz:
Maintainer: @lucize 
Compile tested: mediatek/mt7622, aarch64-cortexa53, master
Run tested: *none*

Description:
The dependency has a PACKAGE_uacme-ualpn condition so that libev won't
be unnecessarily built if uacme-ualpn is not selected.


lucize:
from what I recall if we add this one and don't use the one from the package itself, we can remove `PKG_USE_MIPS16:=0`, this needs to be tested on mips/ramips (just compile would give an error, no need to run test)

lucize:
LGTM

cotequeiroz:
I've successfully built this for ramips and ath79, covering mips_24kc and mipsel_74kc.

PR #18281, opened 2022-04-10
libobsd: add new package

guijan:

Maintainer: me
Compile tested: sdk-22.03-SNAPSHOT-x86-64 (built April 7th)
Run tested: openwrt-22.03-snapshot-r19235-d0965dc174-x86-64 (built April 7th)

Description:
```
This adds the libobsd package. Libobsd is a transparent OpenBSD
compatibility library; the BSDs provide a similar set of libc functions,
so it can also serve as an alternative to the similar libbsd if programs
that make use of it are compiled against libobsd.
```

I actually did run test this package, I wrote a 2nd package for a program I wrote that uses it. Check out my [add-dictpw](https://github.com/guijan/packages/tree/add-dictpw/utils/dictpw) branch.

Here's a screenshot of the package in use:
tmux_1649572387-N

I'm pinging @neheb as he showed interest in the package.

### DIFF
--- a/libs/libobsd/Makefile
+++ b/libs/libobsd/Makefile
@@ -1,0 +1,53 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2022 Guilherme Janczak <guilherme.janczak@yandex.com>
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=libobsd
+PKG_VERSION:=1.1.1
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=https://github.com/guijan/libobsd/releases/download/v$(PKG_VERSION)/
+PKG_HASH:=332b72ba5f9a76c40f8a526771b78dae3ac3388f689f818c0ac4ab77ef52809d
+
+PKG_MAINTAINER:=Guilherme Janczak <guilherme.janczak@yandex.com>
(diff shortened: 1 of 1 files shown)


libs/libobsd/Makefile
```diff
@@ -0,0 +1,55 @@
+#
+# Copyright (C) 2022 Guilherme Janczak <guilherme.janczak@yandex.com>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
```
PolynomialDivision, line 5: please use spdx

neheb:
This should be submitted to https://github.com/openwrt/openwrt and it should replace libbsd. Probably should mention that libobsd is smaller and has fewer dependencies.

Would also help to mention that all packages have been tested as compiling.

neheb:
I don't see any real issue with this PR. Impressive.

libs/libobsd/Makefile
guijan: done

guijan:
> Would also help to mention that all packages have been tested as compiling.

Did you change the Makefile to provide the `libbsd` dependency and tell the build system to generate a _libbsd-overlay.pc_ when testing that? They might have picked up the base libbsd if not. 

> This should be submitted to https://github.com/openwrt/openwrt and it should replace libbsd.

Would the providing a virtual dependency (like [net/libreswan](https://github.com/openwrt/packages/blob/master/net/libreswan/Makefile#L39-L52=) does) be the proper way to do this?

neheb:
PROVIDES is a solution yes. This is done for libudev-zero for example.

guijan:
Closing this to move the conversation to
https://github.com/openwrt/openwrt/pull/9714

PR #18293, opened 2022-04-13
privoxy: prevent killing the init script itself when stopping service

hexian000:
Maintainer: @ja-pa  (find it by checking history of the package Makefile)
armv7, Linksys wrt32x, 21.02.2

Description:
This command is not working:
```sh
/etc/init.d/privoxy restart
```
because the init script kills itself when trying to stop the service


hexian000:
Could anyone please review this? It's only a one line change. Test failures seems not related to this change. Without this fix, once the service stopped in case of ifup/ifdown, it will not be restarted automatically. @PolynomialDivision @ja-pa

BKPepe (changes requested):
1.  CI fails and until then nobody will review it.
 - Missing **Signed-off-by**
 2. ``PKG_RELEASE`` needs to be bumped.
 
 @ja-pa is not maintainer for this package.

hexian000:
> 

Thank you for your patient explanation. Those 2 problems has been fixed.

1. I don't know why CI is failing, I believe it's not introduced by my patch. I need some help with this problem.
2. How can I find the current maintainer of this package. I just followed the guide and find ja-pa in the Makefile's commit log.

hexian000:
> > 
> 
> Thank you for your patient explanation. Those 2 problems has been fixed.
> 
>     1. I don't know why CI is failing, I believe it's not introduced by my patch. I need some help with this problem.
> 
>     2. How can I find the current maintainer of this package. I just followed the guide and find ja-pa in the Makefile's commit log.



> 1. CI fails and until then nobody will review it.
> 
> 
>     * Missing **Signed-off-by**
> 
> 
>     2. `PKG_RELEASE` needs to be bumped.
> 
> 
> @ja-pa is not maintainer for this package.

Thank you for your patient explanation. Those 2 problems has been fixed.

1. I don't know why CI is failing, I believe it's not introduced by my patch. I need some help with this problem.
2. How can I find the current maintainer of this package. I just followed the guide and find ja-pa in the Makefile's commit log.

hexian000:
> 1. CI fails and until then nobody will review it.
> 
> 
>     * Missing **Signed-off-by**
> 
> 
>     2. `PKG_RELEASE` needs to be bumped.
> 
> 
> @ja-pa is not maintainer for this package.

Thank you for your patient explanation. Those 2 problems has been fixed.

1. I don't know why CI is failing, I believe it's not introduced by my patch. I need some help with this problem.
2. How can I find the current maintainer of this package. I just followed the guide and find ja-pa in the Makefile's commit log.

aparcar:
@jefferyto both Buildbot SDKs and CI SDKs use the user [`buildbot`](https://gitlab.com/openwrt/buildbot/-/blob/master/docker/buildworker/Dockerfile#L56) or [`build`](https://gitlab.com/openwrt/docker/-/blob/master/Dockerfile.base#L38), not `root` if I see this correctly.

I'm not aware of other upstream package install scripts running `chown` or something similar on their own, that seems to be a task for the package maintainers of the distribution. [Here](https://github.com/openwrt/openwrt/blob/master/package/utils/busybox/Makefile#L29) is an example on how to downstream *own* a package by a specific user.

hexian000:
Sorry for the mistake, I fixed formality problems this time.
Because this is a script-only change, I only tested this with imagebuilder (by overwritting the init script from FILES).
I'll try to build it later.

hexian000:
> Can you squash the commits? (`PKG_RELEASE` should be incremented in the same commit as changes to the package.)

Sure, btw I found there is a newer bugfix release of privoxy. I'd like to include it in my changes, too. The package has been built from source and tested on my device as well.

jefferyto:
Umm sure... but now you should mention you are also upgrading the package in the commit message (preferably in the commit title), and `PKG_RELEASE` should not be incremented.

Better still, can you set `PKG_RELEASE` to `$(AUTORELEASE)`? That would make on-going maintenance easier.

jefferyto (approved):
Thanks for sticking with this :pray:

jefferyto:
Would you be open to cherry picking this change (with `git cherry-pick -x`) for the openwrt-22.03 and openwrt-21.02 branches?

hexian000:
> Would you be open to cherry picking this change (with `git cherry-pick -x`) for the openwrt-22.03 and openwrt-21.02 branches?

Got it, I will open pull requests soon.

PR #18304, opened 2022-04-15
Fix for DNS API mode

perceival:
Maintainer: me / @tohojo 
Compile tested: NA
Run tested: ipq806x, Netgear Nighthawk X4S R7800, 22.03-SNAPSHOT

Description:
Without quoting run-acme throws the following error:
run-acme /usr/lib/acme/run-acme: export: line 257: /api key/: bad variable name daemon err err 1b

### DIFF
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -257,7 +257,6 @@ issue_cert()
 
 	handle_credentials() {
 		local credential="$1"
-		eval export `$credential`
 	}
 	config_list_foreach "$section" credentials handle_credentials
 
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -257,7 +257,7 @@ issue_cert()
 
 	handle_credentials() {
 		local credential="$1"
-		eval export $credential
+		eval export `$credential`
 	}
 	config_list_foreach "$section" credentials handle_credentials
 


tohojo:
Erm, backticks? That doesn't seem right. What does your config look like, exactly? (please use placeholders instead of any actual credentials when posting!)

perceival:
/etc/config/acme:
```
config acme
	option state_dir '/etc/acme'
	option account_email 'email'
	option debug '0'

config cert 'moodsy'
	option use_staging '0'
	option keylength '2048'
	option enabled '1'
	option update_uhttpd '1'
	option update_nginx '1'
	list domains 'myhost.duckdns.org'
	option validation_method 'dns'
	option dns 'dns_duckdns'
	list credentials 'token'
	option dalias 'myhost.duckdns.org'
```
/etc/acme/myhost.duckdns.org/myhost.duckdns.org.conf
```
Le_Domain='myhost.duckdns.org'
Le_Alt='no'
Le_Webroot='no'
Le_PreHook=''
Le_PostHook=''
Le_RenewHook=''
Le_Keylength='2048'
Le_OrderFinalize='URL'
Le_LinkOrder='password'
Le_LinkCert='URL'
Le_CertCreateTime='1649997958'
Le_CertCreateTimeStr='Fri Apr 15 04:45:58 UTC 2022'
Le_Listen_V6='1'
Le_NextRenewTimeStr=7654321558'
Le_API='URL'

```

perceival:
I agree it should work without them. BTW better way would be probably to use brackets:
```
	handle_credentials() {
		local credential="$1"
		eval export $(credential)
	}
	config_list_foreach "$section" credentials handle_credentials
```

tohojo:
Ah, I see. Well, you're supposed to add the variable name to the credentials field. Each DNS provider script shipped with acme.sh expects different credentials, so there's no way we can handle this from run.sh. You'll need to know which DNS provider script you're using and set the credentials appropriately.

In your case, instead of 
```
	list credentials 'token'
```

you'd need to set:

```
	list credentials 'DuckDNS_Token=token'
```

Closing the PR, as changing the code is not needed :)

PR #18316, opened 2022-04-17
telldus-core: use proper cmake argp

neheb:
Avoids linking to argp-standalone with glibc.

Some other minor fixes.

Maintainer: @PeterFromSweden 
Compile tested: mips64

### DIFF
--- a/utils/telldus-core/Makefile
+++ b/utils/telldus-core/Makefile
@@ -23,6 +23,18 @@ PKG_BUILD_DEPENDS:=argp-standalone
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/cmake.mk
 
+define Download/argp-cmake
+	URL:=@GITHUB/alehaa/CMake-argp/e7e77e68d062708edf055f944d3094b0ce0b11b8/cmake
+	FILE:=Findargp.cmake
+	HASH:=b74e961260d17bc9af868be59053e50739dc94e06cbe73e3fced414a070a29fd
+endef
+
+define Build/Prepare
+	$(eval $(call Download,argp-cmake))
+	$(Build/Prepare/Default)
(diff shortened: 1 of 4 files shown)


utils/telldus-core/Makefile
```diff
@@ -23,6 +23,18 @@ PKG_BUILD_DEPENDS:=argp-standalone
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/cmake.mk
 
+define Download/argp-cmake
```
neheb, line 26: This downloads Findargp.cmake into dl/ . Not a big deal but maybe would be better to download into the download directory only. Not sure how to do that.

utils/telldus-core/patches/980-auto-ptr.patch
```diff
@@ -0,0 +1,20 @@
+--- a/service/DeviceManager.cpp
```
BKPepe, line 1: Why this patch is necessary? It is not mentioned in the commit message.

neheb: minor fix for a compiler warning.

BKPepe: Was this submitted to upstream? Why there is missing commit description in the patch itself? Why it is not mentioned in the commit description?

neheb: Upstream is dead. Minor fixes are mentioned.

BKPepe: In that case, we might want to remove this package. 

"Minor fixes" are too vague.

BKPepe: But still it should be submitted to upstream first. Maybe it will be revived. Who knows. I don't like to carry patches forever.

PR #18147, opened 2022-03-25
golang: Update to 1.18, update patch

jefferyto:
Maintainer: me
Compile tested: armvirt-32 / x86-64, 2022-03-17 snapshot sdk
Run tested: armvirt-32 (qemu), 2022-03-17 snapshot

Description:
Signed-off-by: Jeffery To <jeffery.to@gmail.com>

PR #18319, opened 2022-04-17
passh: an sshpass alternative

clarkwang:
Sshpass does not work well in a few scenarios. Passh is an sshpass
alternative which tries to fix the issues.

Signed-off-by: Clark Wang <dearvoid@gmail.com>

Maintainer: Clark Wang <dearvoid@gmail.com>
Compile tested: 21.02.0-ramips-mt7621
Run tested: 18.06.4-ramips-mt7621

Description:

### DIFF
--- a/utils/passh/Makefile
+++ b/utils/passh/Makefile
@@ -1,0 +1,39 @@
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME    := passh
+PKG_VERSION := 1.0.1
+PKG_RELEASE := $(AUTORELEASE)
+
+PKG_SOURCE     := $(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL := https://codeload.github.com/clarkwang/passh/tar.gz/v$(PKG_VERSION)?
+PKG_HASH       := f6efc7127515b3716108fa28439caca561d16923404bb1a8f734b41340b3f14e
+
+PKG_MAINTAINER    := Clark Wang <dearvoid@gmail.com>
+PKG_LICENSE       := GPL-3.0
+PKG_LICENSE_FILES := LICENSE
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/passh
+  SECTION  := utils
+  CATEGORY := Utilities
+  TITLE    := an sshpass alternative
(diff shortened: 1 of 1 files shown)


PolynomialDivision:
lgtm

jefferyto:
> Run tested: 18.06.4-ramips-mt7621

This doesn't inspire confidence.

PolynomialDivision:
> > Run tested: 18.06.4-ramips-mt7621
> 
> This doesn't inspire confidence.

You are right. Workflows looked good.

clarkwang:
> > Run tested: 18.06.4-ramips-mt7621
> 
> This doesn't inspire confidence.

@jefferyto I was wondering how I could test it on newer openwrt releases since my router is still on 18.06. Last night I tried to upgrade to 21.02 but the upgrade path is not supported. I'll find some time to reinstall it with latest release.

Anyway, is there a way I can test on newer releases without actually installing on my router? Like a simulator?

jefferyto:
@PolynomialDivision The CI will do a run-test if the package has a `test.sh` file (and only on platforms with QEMU support), otherwise it is a compile-test only. Given [recent](https://github.com/openwrt/gh-action-sdk/pull/7) [issues](https://github.com/openwrt/gh-action-sdk/pull/4#issuecomment-1092023756) (the former issue led to both false negatives (build failures) and [false positives](https://github.com/openwrt/packages/pull/18147#issuecomment-1084958154) (build successes)), I would take the current CI with a grain of salt.

@clarkwang You can use QEMU to run virtual instances on your computer. There is more detail on the [wiki](https://openwrt.org/docs/guide-user/virtualization/qemu) (though I don't know how accurate/up-to-date that information is), and the main repo has a [script](https://github.com/openwrt/openwrt/blob/master/scripts/qemustart) you can use/reference.

clarkwang:
> @clarkwang You can use QEMU to run virtual instances on your computer. There is more detail on the [wiki](https://openwrt.org/docs/guide-user/virtualization/qemu) (though I don't know how accurate/up-to-date that information is), and the main repo has a [script](https://github.com/openwrt/openwrt/blob/master/scripts/qemustart) you can use/reference.

Thanks for the info.

And today I upgraded my Netgear R6220 to the latest 21.02.3 and verified passh works fine.

PR #18323, opened 2022-04-18
prometheus-node-exporter-lua: improve ubnt-manager

PolynomialDivision:
It is costly in transmissions to add all information to each metric.
Instead, only use the "device" as a label and add all other important
labels to the "uptime" metric.

@champtar

### DIFF
--- a/utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/ubnt-manager.lua
+++ b/utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/ubnt-manager.lua
@@ -8,9 +8,9 @@ local function get_devices()
     return devices
 end
 
-local function get_metric_airos6(device_data, label)
+local function get_metric_airos6(device_data, label, label_full)
     -- host
-    metric("ubnt_uptime", "counter", label, device_data['host']['uptime'])
+    metric("ubnt_uptime", "counter", label_full, device_data['host']['uptime'])
     metric("ubnt_totalram", "gauge", label, device_data['host']['totalram'])
     metric("ubnt_freeram", "gauge", label, device_data['host']['freeram'])
     metric("ubnt_cpuload", "gauge", label, device_data['host']['cpuload'])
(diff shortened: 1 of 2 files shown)


champtar:
Best fix would be to be behind a real http server with gzip no ?

PolynomialDivision:
What do you mean? Can you give me more information?

champtar:
Right now we are just opening a TCP socket and implementing a poor man http server. switching to a proper http server (uhttpd ?) so we can have IPv4+IPv6, https and gzip might be a better move than limiting the data available.

PolynomialDivision:
> Right now we are just opening a TCP socket and implementing a poor man http server. switching to a proper http server (uhttpd ?) so we can have IPv4+IPv6, https and gzip might be a better move than limiting the data available.

However, the commit is still useful, since repeating the same labels is just unnecessary. I think you are right but for now it does its job. :)

utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/ubnt-manager.lua
```diff
@@ -118,19 +118,23 @@ local function get_metric(device)
...
-        essid = essid
+    local label_short = {
+       device = device
     }
 
+    local label_full = {
+       device = device,
+       hostname = hostname,
+       devmodel = devmodel,
+       fwversion = fwversion,
```
champtar, line 129: Should devmodel and fwversion be metrics instead of labels ?

champtar:
> However, the commit is still useful, since repeating the same labels is just unnecessary. I think you are right but for now it does its job. :)

I don't want to block this PR, we can merge it as is, just wanted to be sure you had considered other options ;)

utils/prometheus-node-exporter-lua/files/usr/lib/lua/prometheus-collectors/ubnt-manager.lua
PolynomialDivision: Hmmmm, they are strings. But could be interesting. How would you do that?

PolynomialDivision:
> > However, the commit is still useful, since repeating the same labels is just unnecessary. I think you are right but for now it does its job. :)
> 
> I don't want to block this PR, we can merge it as is, just wanted to be sure you had considered other options ;)

I will merge. However, lets focus on optimizing that further with gzip. :)

PR #18332, opened 2022-04-19
[22.03] sexpect: Expect for Shells

clarkwang:
Sexpect is another implementation of Expect which is specifically
designed for shells. It's lightweight and has no dependency on other
packages.

Signed-off-by: Clark Wang <dearvoid@gmail.com>

Maintainer: Clark Wang <dearvoid@gmail.com>
Compile tested: 21.02.3-ramips-mt7621
Run tested: 21.02.3-ramips-mt7621 (Netgear R6220)

Description:

### DIFF
--- a/utils/sexpect/Makefile
+++ b/utils/sexpect/Makefile
@@ -1,25 +1,25 @@
 
 include $(TOPDIR)/rules.mk
 
-PKG_NAME    := sexpect
-PKG_VERSION := 2.3.8
-PKG_RELEASE := $(AUTORELEASE)
+PKG_NAME:=sexpect
+PKG_VERSION:=2.3.8
+PKG_RELEASE:=$(AUTORELEASE)
 
-PKG_SOURCE     := $(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL := https://codeload.github.com/clarkwang/sexpect/tar.gz/v$(PKG_VERSION)?
-PKG_HASH       := a586283210a76f03b9cce9f09aac28977d6fc3e314355e22c30d6f42524d9a42
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/clarkwang/sexpect/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=a586283210a76f03b9cce9f09aac28977d6fc3e314355e22c30d6f42524d9a42
 
-PKG_MAINTAINER    := Clark Wang <dearvoid@gmail.com>
-PKG_LICENSE       := GPL-3.0
-PKG_LICENSE_FILES := LICENSE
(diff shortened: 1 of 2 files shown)


utils/sexpect/Makefile
```diff
@@ -0,0 +1,40 @@
...
+PKG_NAME    := sexpect
+PKG_VERSION := 2.3.8
+PKG_RELEASE := $(AUTORELEASE)
+
+PKG_SOURCE     := $(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL := https://codeload.github.com/clarkwang/sexpect/tar.gz/v$(PKG_VERSION)?
+PKG_HASH       := a586283210a76f03b9cce9f09aac28977d6fc3e314355e22c30d6f42524d9a42
+
+PKG_MAINTAINER    := Clark Wang <dearvoid@gmail.com>
+PKG_LICENSE       := GPL-3.0
```
BKPepe, line 13: This is outdated SPDX LICENSE Identifier.

BKPepe (changes requested):
I know that this was merged to master, but I am not sure who reviewed this.

1. Would you please create file ``test.sh`` to be run-tested in GitHub Actions?
2. I know that you are using a different editor than most of us, but can you please remove additional spaces/tabs between variable and ``:=``? This is a minor thing and mostly a cosmetic one, but we need to stick to some format, which we have here. It might be wrong or not, but this is what we have here. 

3. I am not sure why this package is necessary or what the usage is.

clarkwang:
Hi BKPepe,

I've updated the license to GPL-3.0-only.

> I know that this was merged to master, but I am not sure who reviewed this.
> 
> 1. Would you please create file `test.sh` to be run-tested in GitHub Actions?

Done.

> 2. I know that you are using a different editor than most of us, but can you please remove additional spaces/tabs between variable and `:=`? This is a minor thing and mostly a cosmetic one, but we need to stick to some format, which we have here. It might be wrong or not, but this is what we have here.

Done.

> 3. I am not sure why this package is necessary or what the usage is.

It's like Expect (for Tcl) or pexpect (for Python)but does not depend on any other packages so you don't need to install Tcl or Python. It's used to automate interactive commands.

A few discussions:

* https://forum.archive.openwrt.org/viewtopic.php?id=57624
* https://forum.archive.openwrt.org/viewtopic.php?id=57254

jefferyto:
I think it would be better to make changes to master first, then cherry pick the commits for the openwrt-22.03 branch.

clarkwang:
Hi @BKPepe ,

I've made the change on master branch. Not sure if I need to combine 2 commits into 1 commit before merging to 22.03 branch. Please advise.

BKPepe:
Hi,

They should be cherry-picked from master. Right now, it is not clear from the commit description, that they were. You might want to use ``git cherry-pick -x hash`` or you can add it there manually.

PR #18335, opened 2022-04-20
keepalived: add uci support

fjamarquez:
Maintainer: @scrpi, @feckert
Compile tested: ath79
Run tested:  check that several options are parsed correctly to keepalived.conf  and that the keepalived daemon is started 
properly.

Description:

These two commits fix the uci support for the package keepalived, allowing its configuration through uci system.

**fix uci support commit:**

> This commit fix three issues in keepalived.init script:
>     
>     1: Modify the position of config_foreach_wrapper() function because
>        the previous functions are called inside it
>     2: Add config_foreach to read the alt_config_file value correctly
>     3: Rename globals section to global_defs

**add uci support for track_bfd configuration commit:**

> Allow the configuration of track_bfd for vrrp instance throught uci. 
> Use the function bfd_instance() to parse the bfd_instance section to keepalived.conf file

### DIFF
--- a/net/keepalived/files/keepalived.init
+++ b/net/keepalived/files/keepalived.init
@@ -218,6 +218,22 @@ print_track_elem_indent() {
 	printf '\n' >> "$KEEPALIVED_CONF"
 }
 
+print_track_bfd_indent() {
+	local section="$1"
+	local curr_track_elem="$2"
+	local indent="$3"
+	local name
+
+	config_get name "$section" name
+	[ "$name" != "$curr_track_elem" ] && return 0
+
+	config_get weight "$section" weight
+
+	printf '%b%s' "$indent" "$name" >> "$KEEPALIVED_CONF"
+	[ -n "$weight" ] && printf ' weight %s' "$weight" >> "$KEEPALIVED_CONF"
+	printf '\n' >> "$KEEPALIVED_CONF"
+}
+
 static_routes() {
 	local route
 	config_get route "$1" route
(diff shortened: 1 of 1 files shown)


feckert:
Could you please first fix indention we use `tabs`? It looks good at first glance

feckert:
> ok, I think that tabs indention is correct now

Yes its ok now

**To your first commit:**
As I understand it, there is a named `globals` section and there I can specify an alternative configuration in keepalived syntax without using the openwrt uci.
If we do this now as described with your chagne, then I now have several anonymous `globla_defs` section. But I can only use one alternative native keepalived config.

Can you please explain to me what the point is behind this?

fjamarquez:
The `alt_config_file`  (defined in `global_defs` section) parameter defines a configuration file in keepalived format to be used as configuration file instead of the uci configuration (ex. /etc/keepalived/keepalived.conf). With the current init file, that parameter is not used and leaves the final file empty. With the modification, if that parameter is defined, the configuration of the indicated file is applied. 

As I understand it, either the uci configuration or the alternative configuration of the file defined in the `alt_config_file` variable should be applied, but not both at the same time.

feckert:
> As I understand it, either the uci configuration or the alternative configuration of the file defined in the alt_config_file variable should be applied, but not both at the same time.

Yes exactly

But this should work with the current code as I can see.
If the named section `globals` is found in the keepalived uci config, then the `alt_config_file` option is read and a link is set to this file and the function returns and the daemon starts.
```
config globals 'globals'
    option alt_config_file '/etc/keepalived.conf'
```

https://github.com/openwrt/packages/blob/master/net/keepalived/files/keepalived.init#L501

fjamarquez:
That's right, I'm very sorry. I have tested everything again with a full firmware with the latest changes and it worked correctly. I guess in the previous test I missed some library and it didn't parse the configuration properly. I delete the first commit and leave only the second one that includes the tracking configuration with BFD in the uci.

feckert:
Thanks merged!!

PR #18337, opened 2022-04-20
Revert "netatalk: update to version 3.1.13"

BKPepe:
Maintainer: @dangowrt
Compile tested: N/A
Run tested: N/A

Description:

**It seems that only possible solution for users is to downgrade netatalk, which was working even though it is vulnerable to multiple CVEs issues (see reverted commit)**

We received a report from Turris user on Turris support department that
netatalk version 3.1.13 does not work properly.

Process afpd says: INTERNAL ERROR Signal 11
because of that Apple Time Machine does not work as it should

This was already reported to netatalk by different people on various
GNU/Linux distributions like CentOS, AlmaLinux [1] [2]

netatalk developer states [3]:
```
Generally, at this point I can only advice to stop using Netatalk. There
are more pending CVEs that I currently don't have the bandwidth to work on.
```

And he also adds that the fix is working on.

[1] https://sourceforge.net/p/netatalk/bugs/669/
[2] https://sourceforge.net/p/netatalk/bugs/670/
[3] https://sourceforge.net/p/netatalk/mailman/message/37638871/


dangowrt:
Wouldn't it be more appropriate to really just drop netatalk completely, ie. cherry pick commit 51e67786c8dd7eb1463f1b32b0067fd49e225de4 instead of reverting to a known-to-be-vulnerable version? I'm aware that netatalk is hardly ever used on the open Internet, but given that ransomware is *the* thing of the 21st century, even exploits carried out using a local infected machine as a proxy doesn't seem to be too much of an absurd scenario...

BKPepe:
Since we are talking about OpenWrt 19.07, I am not so sure that we want to introduce more drastic changes like dropping this package. I think it is not a wrong solution to revert it, to have it working as it should, and once the fix is available update it one more time. Yes, it is going to be vulnerable. While checking it in Debian, I see that Stretch, Buster, Bullseye has this vulnerable version, too.

Users do not care who is responsible for it or even why it does not work, when it was working, though. I don't think I want to say, user, I am so sorry, but the developer said that you should not use it anymore since it is vulnerable, I know it was working, but there isn't anything I can do.

Well, IMHO, the better approach is indeed to fix it properly, but I don't have time to look at it, and I don't use this. I don't have any Apple devices. On the other hand, this was quick change to get it working and have satisfied users and customers.

BKPepe:
By looking at https://github.com/Netatalk/Netatalk/pull/174 , which should address reported issue (https://github.com/Netatalk/Netatalk/issues/175) and mention above in OP. I think for a few days, we can go back to previous version until the new version appears. :-)

jefferyto:
Users who have already installed 3.1.13 will not see this "update" - what you want is a package version like 3.1.13-2really3.1.12.

diizzyy:
@BKPepe 
https://cgit.freebsd.org/ports/commit/?id=50daf3d28888226dc32e493b53ffc588a65cca5e ?

BKPepe:
@jefferyto You are right. That's somehow different than what it is in Turris OS. Thanks for spotting this! I will fix this.
@diizzyy Well, it could help, but I think there are more issues than just this (ref my previous comment). E.g. OpenBSD decided to downgrade it to 3.1.12 as well until fixes will appear. Hopefully soon.

autobakterie:
There is also a fix in progress in a Pull Request to Netatalk GitHub, seems to be nearing finalization, but not yet optimal:
https://github.com/Netatalk/Netatalk/pull/174

autobakterie:
The main question how I see it is - is it more broken when it allows root access to the router from any device in local network or when it crashes sometimes?

And possibly a second one: Do we find it beneficial to actively re-include a vulnerability in the repo?

BKPepe:
It crashes all the time once you try to authorize yourself. Netatalk should be used only in a trusted local network as I going to quote https://github.com/openbsd/ports/commit/68fb848d49768e4e9e56defbd34b1131c3a1b6d2. The question is do we want to have working software for users or not? If not, then they will throw it to trash bin. :) That's common response from users on our support department IMHO.

autobakterie:
By "crashes sometimes" I primarily referred to the two mentioned not much tested 3.1.13 fixes which can be applied temporarily just now instead of straight revert.

autobakterie:
I don't think "doing something useful while endangering the device I run on unless the device get's compromised" == "working".

Trusted network AFAIK doesn't usually mean everything inside has root access to the router. The mostly accepted security in depth approach is based on layered protection, which in this case, I think, usually means being inside local network should't be equal to having root on the router  - that should be the next protection layer.

Neustradamus:
19.07 / Re-introduce Netatalk 3.1.13 with:
- https://github.com/openwrt/packages/pull/18696

Linked to:
- https://github.com/openwrt/packages/pull/18687

PR #18361, opened 2022-04-24
https-dns-proxy: 2021-11-22-3: add support for Canary Domains

stangri:
Maintainer: me
Compile tested: x86_64, Sophos SG-135, OpenWrt 21.02.2
Run tested: x86_64, Sophos SG-135, OpenWrt 21.02.2, test the Canary Domain

Description:
Fixes https://github.com/stangri/source.openwrt.melmac.net/issues/155.
Background: https://support.mozilla.org/en-US/kb/canary-domain-use-application-dnsnet

### DIFF
--- a/net/https-dns-proxy/Makefile
+++ b/net/https-dns-proxy/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=https-dns-proxy
 PKG_VERSION:=2021-11-22
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/aarond10/https_dns_proxy/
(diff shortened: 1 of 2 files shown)


BKPepe:
I'd like to rather see commit subject as ``https-dns-proxy: add support for Canary Domains``, because right now it seems like versioning in the commit subject could be misleading.  Just my opinion on this thing.

stangri:
> I'd like to rather see commit subject as `https-dns-proxy: add support for Canary Domains`, because right now it seems like versioning in the commit subject could be misleading. Just my opinion on this thing.

Thanks for review this!

BKPepe:
> I've added "add", could you please elaborate why the package version in the commit subject is bad?

From my point of view, based on other commits here, it could mean that you try to update it to a newer version and, at the same time, add support for Canary Domains. You are not doing anything with the versioning except the bump, which is necessary.

stangri:
> > I've added "add", could you please elaborate why the package version in the commit subject is bad?
> 
> From my point of view, based on other commits here, it could mean that you try to update it to a newer version and, at the same time, add support for Canary Domains. You are not doing anything with the versioning except the bump, which is necessary.

I see. My idea was to "tag" the commit/PR with the corresponding version number (so it's easier to find), that's why I put it in the subject. Is there another way of doing that?

PR #18372, opened 2022-04-25
[21.02] prometheus-node-exporter-lua: improve ubnt-manager

PolynomialDivision:
It is costly in transmissions to add all information to each metric.
Instead, only use the "device" as a label and add all other important
labels to the "uptime" metric.

(cherry picked from commit 9f3064a11ccb4c04f610faa56b4695adbbab224d)


champtar:
I just noticed this weekend that 21.02 is missing a lots of commits from master.
Having the same version is confusing, I think I'm going to cherry-pick everything that is missing with the uhttpd change.
I believe one of your cherry-pick you had to fix a conflict around the version, we should have discussed that then.

PR #18390, opened 2022-04-26
strongswan: support child rekey by bytes and packets

lowjoel:
Compile tested: N/A
Run tested: aarch64-cortexa53 OpenWrt 22.03-rc1 (Linksys E8400)

Description:

This adds support for the child SA to be rekeyed through the byte/packet
threshold. The default is blank (which disables the byte/packet thresholds).

Not sure how to update the main OpenWrt documentation to include these new fields. Guidance welcome.

Test output
```
# swanctl --list-conns
<conn>: IKEv2, no reauthentication, rekeying every 14400s, dpd delay 30s
  local:  %any
  remote: vpn1.example.com
  local pre-shared key authentication:
    id: vpn2.example.com
  remote pre-shared key authentication:
    id: vpn1.example.com
  <child>: TUNNEL, rekeying every 3600s or 4294967296 bytes, dpd action is restart
    local:  0.0.0.0/0
    remote: 0.0.0.0/0
```

### DIFF
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -233,6 +233,10 @@ config_child() {
 	local startaction
 	local if_id
 	local rekeytime
+	local rekeybytes
+	local lifebytes
+	local rekeypackets
+	local lifepackets
 
 	config_get startaction "$1" startaction "route"
 	config_get local_nat "$1" local_nat ""
@@ -247,6 +251,10 @@ config_child() {
 	config_get interface "$1" interface ""
 	config_get hw_offload "$1" hw_offload ""
 	config_get priority "$1" priority ""
+	config_get rekeybytes "$1" rekeybytes ""
+	config_get lifebytes "$1" lifebytes ""
+	config_get rekeypackets "$1" rekeypackets ""
+	config_get lifepackets "$1" lifepackets ""
 
 	config_list_foreach "$1" local_subnet append_var local_subnet ","
 	config_list_foreach "$1" remote_subnet append_var remote_subnet ","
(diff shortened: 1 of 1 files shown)


lowjoel:
Hey @pprindeville, gentle bump. Hope you can have this reviewed for inclusion soon.

net/strongswan/files/swanctl.init
```diff
@@ -227,26 +227,34 @@ config_child() {
...
+	local life_bytes
+	local rekey_packets
+	local life_packets
 
 	config_get startaction "$1" startaction "route"
 	config_get local_nat "$1" local_nat ""
 	config_get updown "$1" updown ""
 	config_get firewall "$1" firewall ""
-	config_get lifetime "$1" lifetime ""
+	config_get life_time "$1" lifetime ""
```
pprindeville, line 245: Changing the names of the settings breaks backwards compatibility.  Please don't do this.

pprindeville (changes requested):
Don't break compatibility with existing configs.

net/strongswan/files/swanctl.init
lowjoel: Hey @pprindeville - I have tested this with my existing configuration using `lifetime`, and AFAICS the first argument to `config_get` is the name of the bash local, not the name of the variable in the config. Therefore this doesn't break backwards compatibility.

The naming is rather inconsistent in this function, and so I can't figure out if just `concatenatingallthewords` is preferred, or if using the swanctl.conf `snake_case` keyword is preferred. The `/etc/config/ipsec` keys are all still `singleword`.

net/strongswan/files/swanctl.init
```diff
@@ -227,26 +227,34 @@ config_child() {
 	local updown
 	local firewall
 	local remote_subnet
-	local lifetime
+	local life_time
```
lowjoel, line 230: Likewise, if it's preferred to combine all the local down below with the additions, let me know.

net/strongswan/files/swanctl.init
pprindeville: Using single words is preferred, unless you're willing to write a filter that handles the up-conversion...

lowjoel: @pprindeville fixed and rebased against master. Please have a look.

pprindeville (approved):
LGTM

PR #18398, opened 2022-04-27
antfs: require kernel version 5.10 to build

john-tho:
ANTFS does not compile with OpenWrt's current testing kernel 5.15,
as it needs to be modified for the Linux 5.12 idmapped mounts changes
2f221d6f7b88 ("attr: handle idmapped mounts")
https://lore.kernel.org/all/20210121131959.646623-1-christian.brauner@ubuntu.com/

Signed-off-by: John Thomson <git@johnthomson.fastmail.com.au>

---

Maintainer: @dengqf6
Compile tested: ath79, master, testing kernel (5.15): no longer stop ALL_KMODS=y build
Run tested: N/A

---

excluding kernel 5.15 may be preferable to requiring 5.10?
`@!LINUX_5.15`

### DIFF
--- a/kernel/antfs/Makefile
+++ b/kernel/antfs/Makefile
@@ -20,7 +20,7 @@ define KernelPackage/fs-antfs
 	TITLE:=AVM NTFS Read/Write Driver
 	FILES:=$(PKG_BUILD_DIR)/antfs.ko
 	AUTOLOAD:=$(call AutoLoad,30,antfs,1)
-	DEPENDS:=+kmod-nls-base
+	DEPENDS:=+kmod-nls-base @LINUX_5_10
 endef
 
 define KernelPackage/fs-antfs/description


dengqf6:
As 5.15 has in-tree ntfs3, I think this should be excluded from 5.10.

neheb:
So you want to backport that or?

john-tho:
@dengqf6 Could you please clarify for us:
> As 5.15 has in-tree ntfs3, I think this should be excluded from 5.10.

I am reading this as exclude this ANTFS package after kernel 5.10?
In that case, you support the change made by PR?

dengqf6:
> I am reading this as exclude this ANTFS package after kernel 5.10?

Yes. It was a typo.

neheb:
Will anyone be adding kmod-fs-ntfs3 support to base?

john-tho:
Thanks @neheb

> Will anyone be adding kmod-fs-ntfs3 support to base?

This is done:
https://github.com/openwrt/openwrt/blob/ab4eafbed974c6ead13b11a50b3032670c26540a/package/kernel/linux/modules/fs.mk#L646

Some are not so happy with the apparent lack of maintenance of ntfs3 though: https://lore.kernel.org/all/da20d32b-5185-f40b-48b8-2986922d8b25@stargateuniverse.net/

PR #18405, opened 2022-04-28
dockerd: fix dependencies missing

ggg17226:
Maintainer:  @G-M0N3Y-2503 

Compile tested:
tested on build Openwrt master x86_64 with musl on ubuntu 20.04 (amd64)

Run tested:
tested on amd64 qemu kvm vm 

Description:
fix dockerd dependencies missing, dockerd runtime require system has legacy iptables command and firewall nat support


ggg17226:
by the way, branch openwrt-22.03 also has this problem. should i make another pr?

G-M0N3Y-2503:
> by the way, branch openwrt-22.03 also has this problem. should i make another pr?

I'm working to get docker working with `iptables-nft`, just having a little trouble.
So this work will be redundant once that is working, but if it's helpful for you in the meantime this looks ok to go into master and openwrt-22.03

PR #18408, opened 2022-04-28
usbip: remove unused libbsd and pthreads deps

guijan:
Maintainer: @nunojpg 
Compile tested: x64 glibc toolchain built from commit f757a8a09885e3c8bb76371e037b8c0731111980
Run tested: None

This package doesn't actually depend on libbsd and pthreads on glibc. The Makefile was adding them to the build, I don't undersand why, but I don't see any mention of either in usbip's source.

Found while working on https://github.com/openwrt/openwrt/pull/9714.

### DIFF
--- a/net/usbip/Makefile
+++ b/net/usbip/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=usbip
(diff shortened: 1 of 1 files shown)


cotequeiroz:
I *think* it may have been because the old `libudev-fbsd` required `-lbsd -lpthread` when linking to glibc.  It's gone now, so this can go as well.

PR #18414, opened 2022-04-29
hfsprogs: update to 540.1.linux3

neheb:
Updates hfsprogs from 2011 to 2013 version.

Maintainer:
Compile tested: mips64


BKPepe:
So many patches were removed or some of them were added/refreshed, but it is not mentioned in the commit message. Why?

neheb:
I think what happened was the old patches were taken from Debian and when Debian updated to this version, they threw them out and replaced them with half of these. This PR has them and a few extra.

BKPepe:
I don't have anything against that, but it should be documented and that's why you should add it the commit description or do it in additional commit.

PR #18427, opened 2022-04-30
docker: fix compilation with glibc

ggg17226:
Maintainer: @G-M0N3Y-2503 

Compile tested:
tested on build Openwrt master x86_64 with glibc on ubuntu 20.04

Run tested:
tested on amd64 qemu kvm vm

Description:
fix docker build error when using glibc on tartget


ggg17226:
This change can also apply to branch 22.03

PR #18428, opened 2022-04-30
dockerd: fix compilation with glibc

ggg17226:
Maintainer: @G-M0N3Y-2503 

Compile tested:
tested on build Openwrt master x86_64 with glibc on ubuntu 20.04

Run tested:
tested on amd64 qemu kvm vm

Description:
fix dockerd build error when using glibc on tartget.
This change can also apply to branch 22.03

### DIFF
--- a/utils/dockerd/Makefile
+++ b/utils/dockerd/Makefile
@@ -51,6 +51,7 @@ endef
 GO_PKG_BUILD_VARS += GO111MODULE=auto
 TAR_OPTIONS:=--strip-components 1 $(TAR_OPTIONS)
 TAR_CMD=$(HOST_TAR) -C $(1) $(TAR_OPTIONS)
+TARGET_LDFLAGS += $(if $(CONFIG_USE_GLIBC),-lc -lgcc_eh)
 
 # $(1) = path to dependent package 'Makefile'
 # $(2) = relevant dependency '.installer' file


ggg17226:
@neheb 
it seems like auto ci has some problem

castillofrancodamian:
Still not compiled.

ggg17226:
> Still not compiled.

what version of gcc/binutils you are using?
you the best to create an new issue with error log and config file on it.

G-M0N3Y-2503:
> Still not compiled.

If your issue is the `runc` version mismatch compile issue should be addressed by https://github.com/openwrt/packages/pull/18441.

Entropy512:
I'm kind of scratching my head as to how the runc mismatch didn't break my builds until this landed.  I am POSITIVE that I have done a dirclean in the two months since the runc bump was committed, but my build didn't break until this touched the Makefile and forced a rebuild.

castillofrancodamian:
> > Still not compiled.
> 
> what version of gcc/binutils you are using? you the best to create an new issue with error log and config file on it.

How do I provide you with the necessary information?

castillofrancodamian:
> > Still not compiled.
> 
> If your issue is the `runc` version mismatch compile issue should be addressed by #18441.

Then we will have to wait for it to merge.

BKPepe:
We should really stop merging stuff to our repository without a commit description because the commit itself just adds something, no one knows why, and when I was digging deeper, I got to this pull request and am not sure, if it still applies - how I can reproduce it, why other Go packages do not require, etc.

If I would be applied to reproduce it - what is the exact error? Because if you know, you can meet something else. You did not know if it is related or not.

ggg17226:
because when build openwrt with glibc, the build config will filter out all gcc* implicit link libraries, so these packages must add this link config.

BKPepe:
And only 4 packages have exactly the same link config. Interesting.

OpenWrt with Glibc is not enabled and supported by default. Is it?

Dne čt 5. 10. 2023 11:46 uživatel aghost ***@***.***> napsal:

ggg17226:
yeah, build these package without this patch on openwrt with glibc will get link error.
with musl, docker can be run, but could not use general docker image.

BKPepe:
Which error exactly to be able to track it down? What needs to be done? So
many questions, which should be answered in the beginning.

That's what commit description is for.

Dne čt 5. 10. 2023 12:08 uživatel aghost ***@***.***> napsal:

Entropy512:
> with musl, docker can be run, but could not use general docker image.

The whole point of Docker is to isolate the host library environment from that of any containers.

I've been running Docker compiled against musl in OpenWRT for two years at this point, and I have no problem running containers that use libc.

I'm trying to figure out why this has suddenly come up again - was there a regression that broke Docker compilation against libc again?

It seems like this commit was done to fix a very rare use case, since as @BKPepe points out, glibc is not a normal configuration for OpenWRT.  So other packages might be broken when trying to build against libc but no one has noticed.  @BKPepe what problem are you trying to track down that caused you to wind up here?  Another package failing to compile against glibc?

In my case, looking at history, the only reason I commented on this is because the change to the Makefile forced a rebuild, which caused another issue (runc mismatch) to become apparent.  My PR to fix that is linked from a comment above.

BKPepe:
> I'm trying to figure out why this has suddenly come up again - was there a regression that broke Docker compilation against libc again?

People are copy&pasting things from different Makefiles and they don't have any clues what they are doing. So, yes. I am trying to find: **Why it was added and also if it can be removed (preferred).**

PR #18436, opened 2022-05-01
mtools: avoid picking libbsd dependency

cotequeiroz:
Maintainer: @oskarirauta 
Compile tested: ramips/rt3883, mipsel_74kc, openwrt master--also host-compiled under gentoo to ensure glibc does not need it.
Run tested: ramips/rt3883, mipsel_74kc, openwrt master

Description:

Even though commit 96ec989ed ("mtools: update to 4.0.39") correctly dropped libbsd dependency, the package's configure script will still link to it if the library is found.

Tell configure not to look for it by setting `ac_cv_lib_bsd_main=no`.

---

CC: @guijan

### DIFF
--- a/utils/mtools/Makefile
+++ b/utils/mtools/Makefile
@@ -29,7 +29,9 @@ define Package/mtools/description
   Mtools is a collection of utilities to access MS-DOS disks from GNU and Unix without mounting them.
 endef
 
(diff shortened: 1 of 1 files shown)


guijan:
Mtools upstream should be told about this. Seems they forgot to remove libbsd from configure.in.

PR #18452, opened 2022-05-05
syncthing: update to 1.20.1

paper42:
Maintainer: @aparcar

Description:
I am getting this warning, but I don't think this can be solved in OpenWrt because the value would have to be device specific:
```
2022/05/05 14:11:27 failed to sufficiently increase receive buffer size (was: 176 kiB, wanted: 2048 kiB, got: 352 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
```


aparcar:
Thanks!

PR #18454, opened 2022-05-05
audit: Fix compilation with kernel 5.15

cotequeiroz:
Maintainer: @tpetazzoni 
Compile tested:  mediatek/mt7622 (linux 5.15), aarch64, master
Run tested: none, but it is a trivial patch

Description:
Linux 5.15 does not have the linux/ipx.h header.

The patch is a partial cherry-pick (skipped ChangeLog) of upstream commit https://github.com/linux-audit/audit-userspace/commit/6b09724c6 ("Make IPX packet interpretation dependent on the ipx header file existing").

----

I've stumbled upon this while doing a `CONFIG_ALL` build to spot kernel-5.15 build problems.

For background, IPX support code was removed from linux in 2018;  the ipx.h header was left there until August 2021, when it was finally removed by https://github.com/torvalds/linux/commit/6c9b40844751ea30c72f7a2f92f4d704bc6b2927.

### DIFF
--- a/utils/audit/Makefile
+++ b/utils/audit/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=audit
 PKG_VERSION:=2.8.5
-PKG_RELEASE:=2
+PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=http://people.redhat.com/sgrubb/audit
--- a/utils/audit/Makefile
+++ b/utils/audit/Makefile
@@ -24,7 +24,6 @@ PKG_BUILD_DIR=$(BUILD_DIR)/$(PKG_NAME)-packages/$(PKG_NAME)-$(PKG_VERSION)
 PKG_USE_MIPS16:=0
 
 include $(INCLUDE_DIR)/package.mk
-include $(INCLUDE_DIR)/host-build.mk
 
 define Package/audit/Default
   TITLE:=Audit Daemon
@@ -139,7 +138,6 @@ define Package/audit/install
 	$(CP) $(PKG_INSTALL_DIR)/usr/sbin/auditd $(1)/usr/sbin/
 endef
 
-$(eval $(call HostBuild))
 $(eval $(call BuildPackage,libauparse))
 $(eval $(call BuildPackage,audit-utils))
 $(eval $(call BuildPackage,audit))
--- a/utils/audit/Makefile
+++ b/utils/audit/Makefile
@@ -19,6 +19,7 @@ PKG_LICENSE_FILES:=COPYING
 PKG_CPE_ID:=cpe:/a:linux_audit_project:linux_audit
 
 PKG_FIXUP:=autoreconf
+PKG_BUILD_DIR=$(BUILD_DIR)/$(PKG_NAME)-packages/$(PKG_NAME)-$(PKG_VERSION)
 
 PKG_USE_MIPS16:=0
 
--- a/utils/audit/patches/0003-Make-IPX-packet-interpretation-dependent-on-th.patch
+++ b/utils/audit/patches/0003-Make-IPX-packet-interpretation-dependent-on-th.patch
@@ -1,0 +1,52 @@
+From 6b09724c69d91668418ddb3af00da6db6755208c Mon Sep 17 00:00:00 2001
+From: Steve Grubb <sgrubb@redhat.com>
+Date: Thu, 2 Sep 2021 15:01:12 -0400
+Subject: [PATCH] Make IPX packet interpretation dependent on the ipx header
+ file existing
+
+--- a/auparse/interpret.c
++++ b/auparse/interpret.c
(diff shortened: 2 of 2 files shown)


jefferyto:
CI has a weird error:

```
make  install-data-hook
make[8]: Entering directory '/home/build/openwrt/build_dir/target-mipsel_24kc_musl/audit-2.8.5/audisp/plugins/builtins'
mkdir -p -m 0750 /home/build/openwrt/build_dir/target-mipsel_24kc_musl/audit-2.8.5/ipkg-install/etc/audisp/plugins.d
mkdir: cannot create directory '/home/build/openwrt/build_dir/target-mipsel_24kc_musl/audit-2.8.5/ipkg-install/etc/audisp': Not a directory
make[8]: Leaving directory '/home/build/openwrt/build_dir/target-mipsel_24kc_musl/audit-2.8.5/audisp/plugins/builtins'
make[8]: *** [Makefile:503: install-data-hook] Error 1
make[7]: *** [Makefile:438: install-data-am] Error 2
```

cotequeiroz:
I saw that.  The problem happens [a bit earlier](https://github.com/openwrt/packages/runs/6394458892?check_suite_focus=true#step:5:5887):

```
/home/build/openwrt/staging_dir/host/bin/install -c -m 644 -D -m 640 ./audispd.conf /home/build/openwrt/build_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/audit-2.8.5/ipkg-install/etc/audisp
```
It appears to expect that the `ipkg-install/etc/audisp` directory had already been created.  Instead, it is installing a regular file named `ipkg-install/etc/audisp`.  I could swear that my patch would not touch that stuff.

Is this a candidate for `PKG_BUILD_PARALLEL:=0`?  I'll give it a try in a bit...

cotequeiroz:
Nevermind.  I found more intriguing information about the problem.  Could it be caused because of a different version of `make`.  Here, at least, `make` runs `install` in `plugins/builtins` before it does `init.d`.  We can create the directory before we call `make install`.  Let me try that next.

cotequeiroz:
The "upstream" `audit` project is built by two openwrt packages:  `libaudit` in the base feed, then `audit` here.  The problematic file is created when `libaudit` is built.  Then, when the `audit` is built, it picks up the install tree from the previous run.  

The best way to avoid such interference is to have them built in separate directories.

cotequeiroz:
That made the Ci happy!

cotequeiroz:
I'll propose an equivalent change in base, to avoid (cause?🤔) further confusion.

jefferyto:
I tried building the existing audit package locally and it succeeded. Having separate build directories for different packages is a good thing, but I'm still puzzled by why it's a problem for the CI but not for local builds.

jefferyto:
Since `CONFIG_AUTOREMOVE` defaults to `y`, the ipkg-install directory created by libaudit should have been removed when the build for audit started :thinking:

cotequeiroz:
Before calling `make package/audit/compile`, the CI runs `make package/audit/refresh`, which activates `QUILT` and prevents the cleanup of the build dir.  Then, I imagine, `libaudit/compile` will not clean the shared build dir, and `audit/compile` will use it again.  I've just run it here to give it a try, and it failed exactly like the CI did.

cotequeiroz:
@aparcar 
If you run a `clean` job in between `refresh` and `compile`, it will compile the usual way, and problems with `QUILT` being active will be hidden.  This can be a good or a bad thing.  What do you think?
The down side of this, aside from not picking up `QUILT` build problems--I consider a shared build directory a problem--is that it will take longer to run.

jefferyto:
`clean` should definitely be run after a `refresh`.

neheb:
I think audit should be moved to base to have just one package.

feckert:
@neheb I have also played around with audit to solve another problem. I felled over the same problem and didn't understand why we maintain audit in two packages to build different packages.
Also it is not helpful if we do this in different feeds!
If we update one we have to update the other too. Otherwise libaudit will no longer fit together.
I in favour of unifying this and building the openwrt packages from one source package, which would improve everything for audit

 Here are all my pullrequests, which I closed because nothing went further.
https://github.com/openwrt/packages/pull/16342
https://github.com/openwrt/openwrt/pull/4447

jefferyto:
I read through the various PRs and it seemed the reason to split audit into two packages is to [avoid moving libev into the core](https://github.com/openwrt/packages/pull/10664#issuecomment-684016402) - is there a particular reason to avoid this, @dangowrt?

I also think it would be best to have only one package (in the main repo), but as an intermediate fix we can change the build directory (i.e. this PR). I think we can also add a check in this package to ensure the version matches the version of libaudit in core.

BTW https://github.com/openwrt/openwrt/pull/4447 limits this package to certain architectures - does that still apply?

feckert:
> BTW [openwrt/openwrt#4447](https://github.com/openwrt/openwrt/pull/4447) limits this package to certain architectures - does that still apply?

Audit does not exist for all architectures. It builds, but makes no sense from my point of view, as there are no syscall tabs for the non-supported architectures.
The following archs are supported.
https://github.com/linux-audit/audit-userspace/tree/master/lib

neheb:
Let's merge for now.

PR #18464, opened 2022-05-07
add nut driver config option "cable"

robje:
Maintainer: RobJE @robje
Compile tested: none
Run tested: 19.07.x and 21.02.3 (both hot-patched)

Description:
at least driver apcsmart-old (maybe more) allow for specifying the
type of cable used. My old UPS does will not function when cable type
is not specified.

This will add support for configuration option 'cable'


jefferyto:
You need to increment `PKG_RELEASE` in the package Makefile.

robje:
> You need to increment `PKG_RELEASE` in the package Makefile.

Thanks for the hint. TIL

jefferyto:
Please squash the commits (the `PKG_RELEASE` change should be in the same commit as other package changes).

PR #18467, opened 2022-05-07
ddns-scripts: mydns.ip to mydns.jp

AdorableJiang:
The provider is `mydns.jp` not `mydns.ip` which is wrong and misleading.


jefferyto:
I think you should increment `PKG_RELEASE` instead of `PKG_VERSION`.

AdorableJiang:
> I think you should increment `PKG_RELEASE` instead of `PKG_VERSION`.

Fixed.

jefferyto:
Would it be possible to squash the commits?

AdorableJiang:
I believe we are all set now.

jefferyto (approved):
Thanks!

PR #18472, opened 2022-05-08
kcptun: add tcp option

sotux:
Add a tcp option in the init.d script and config file.
This option is supported by the current version of kcptun.
I hope it can be cherry-picked to the branch openwrt-21.02 & openwrt-22.03.


jefferyto:
Ping package maintainers @liudf0716 @expiron

expiron:
LGTM

PR #18485, opened 2022-05-09
crun: update to version 1.4.5

oskarirauta:
- CRIU: add support for different manage cgroups modes.
 - the hook processes inherit the crun process environment if there is no environment block specified in the OCI configuration.
 - exec: fix double free when using --apparmor and --process-label.

Compile tested: x86_64
Run tested: x86_64


jefferyto:
I see from this PR and from the package history, that it is always not the commit tagged by upstream that is packaged but another commit (e.g. upstream tagged https://github.com/containers/crun/commit/c381048530aa750495cf502ddb7181f2ded5b400 as 1.4.5 but this PR has `PKG_SOURCE_VERSION:=85649a8c561f7fbf018e5c5640a7e9370c80ce73`). Is there a reason for this?

oskarirauta:
@jefferyto 

Only reason is that I am not always updating package _on-time_ - so I review changes on master 'till current moment when I update the package, if nothing experimental seems to be there- I use latest commit (which still runs with same version) to _compensate late updating_

This time it favours us, since openwrt doesn't have journald and some commits take crun further away from using journald- although alternative solution isn't a native solution for openwrt either... But nevertheless, there's some good stuff after initial 1.4.5 like improved group handling, newer support libraries being used and more, still nothing that would make it incompatible with version from exact 1.4.5 commit.

So, no valid reason, like broken version 1.4.5 which is fixed in a later commit, just a attempt to add some of that _good stuff_..

jefferyto:
I don't know about this project in particular, but I think in general projects tag a release when they feel the code at that moment is ready for general use (the code is stable). Upstream probably expects most users to be using stable releases, and most users probably want/expect to be using stable releases.

I don't want to discount your judgement, you already identified a change after 1.4.5 that makes the software usable on OpenWrt. But I think the less we deviate from stable releases, the more predictable and unsurprising it will be for everyone (upstream, end users, us).

In the case of fixes released after a stable release, I would suggest backporting the changes (adding the commits as patches). That way we don't include any other changes that may introduce new issues.

I don't want to block this update because of this though, just hope to give you some ideas to consider :smile:

jefferyto:
Apologies if I sound too preachy, I know you don't do this for most of your other packages.

Issue #15267, opened 2021-03-26
workflows: seeing multi-arch-test-build.yml compute wrong changed packages


pprindeville:
Maintainer: @aparcar 
Environment: Test Builds in GitHub

Description:

On PR #14711 two files are changed from `master`:

- `net/strongswan/Makefile`
- `net/strongswan/files/ipsec.init`

Yet looking at [this test build](https://pipelines.actions.githubusercontent.com/5yKFRfuu84LeYF9JWXeroTFD7H8ygjTZsUcmjLvv85kJt4ah0c/_apis/pipelines/1/runs/4397/signedlogcontent/11?urlExpires=2021-03-25T06%3A09%3A01.4076076Z&urlSigningMethod=HMACV1&urlSignature=cPtMpRoQUKoDyR2XAD%2B%2BAq84DKz4as%2F%2BYtDPSv%2FXAec%3D) we see the following:

```
...
2021-03-25T01:10:09.5633256Z ##[group]Run # only detect packages with changes
2021-03-25T01:10:09.5634146Z # only detect packages with changes
2021-03-25T01:10:09.5634988Z PKG_ROOTS=$(find . -name Makefile | grep -v ".*/src/Makefile" | sed -e 's@./\(.*\)/Makefile@\1/@')
2021-03-25T01:10:09.5635886Z CHANGES=$(git diff --diff-filter=d --name-only origin/master)
2021-03-25T01:10:09.5636699Z 
2021-03-25T01:10:09.5637130Z for ROOT in $PKG_ROOTS; do
2021-03-25T01:10:09.5637714Z   for CHANGE in $CHANGES; do
2021-03-25T01:10:09.5638227Z     if [[ "$CHANGE" == "$ROOT"* ]]; then
2021-03-25T01:10:09.5638864Z       PACKAGES+=$(echo "$ROOT" | sed -e 's@.*/\(.*\)/@\1 @')
2021-03-25T01:10:09.5639394Z       break
2021-03-25T01:10:09.5639781Z     fi
2021-03-25T01:10:09.5640223Z   done
2021-03-25T01:10:09.5640657Z done
2021-03-25T01:10:09.5641409Z 
2021-03-25T01:10:09.5642154Z # fallback to test packages if nothing explicitly changes this is
2021-03-25T01:10:09.5642983Z # should run if other mechanics in packages.git changed
2021-03-25T01:10:09.5643763Z PACKAGES="${PACKAGES:-vim tmux bmon}"
[...]

PR #18512, opened 2022-05-11
ci: Look for changed packages in the PR branch only

jefferyto:
Maintainer: @aparcar (ping @neheb, @BKPepe)
Compile tested: N/A
Run tested: N/A

Description:
Fixes https://github.com/openwrt/packages/issues/15267.


jefferyto:
The [nested loop](https://github.com/openwrt/packages/blob/7de587566e285322a76d997ab99d01e4d001c272/.github/workflows/multi-arch-test-build.yml#L73-L80) directly below is also the kind of O(n^2) code they tell you to avoid in programming school, but it doesn't annoy me enough for me to fix right now.

neheb (approved):

PR #18516, opened 2022-05-11
nut: add cable type nut driver config

robje:
retake of MR #18464

Maintainer: RobJE @robje
Compile tested: none
Run tested: 19.07.x and 21.02.3 (both hot-patched)

Description:
at least driver apcsmart-old (maybe more) allow for specifying the
type of cable used. My old UPS does will not function when cable type
is not specified.

This will add support for configuration option 'cable'

### DIFF
--- a/net/nut/Makefile
+++ b/net/nut/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nut
 PKG_VERSION:=2.7.4
-PKG_RELEASE:=27
+PKG_RELEASE:=28
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=http://www.networkupstools.org/source/2.7/
--- a/net/nut/files/nut-server.init
+++ b/net/nut/files/nut-server.init
@@ -147,6 +147,7 @@ build_driver_config() {
 	echo "[$cfg]" >>"$UPS_C"
 
 	get_write_driver_config "$cfg" bus
+	get_write_driver_config "$cfg" cable
 	get_write_driver_config "$cfg" community
 	get_write_driver_config "$cfg" desc
 	get_write_driver_config "$cfg" driver "usbhid-ups"


jefferyto:
Apologies for not noticing earlier: please add the package name to the front of the commit title, i.e. "nut: add cable type nut driver config"

robje:
> I mean the title in the commit message, not the title of the PR.

getting this simple 1 line change merged feels is becoming quite the challenge. I am pretty darn close to just giving up :disappointed: 

I'm also not sure how jumping through all these hoops for a small change would help attract people "passing by" to help.

As requested I have rewritten history. The commit now has the package name in it.

jefferyto:
> As requested I have rewritten history. The commit now has the package name in it.

Thanks.

> I'm also not sure how jumping through all these hoops for a small change would help attract people "passing by" to help.

Incrementing `PKG_RELEASE` updates the package version string so that this appears as an upgrade for existing users. Having it in the same commit makes the changes atomic (there is no state where the functional change exists but `PKG_RELEASE` is unchanged). Including the package name as a prefix for the commit title makes it easier for people looking at the git log to understand what package(s) are affected by each commit.

There are reasons, if you care to understand them. They improve the quality of the project. They are also included in the [contribution guidelines](https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md).

> getting this simple 1 line change merged feels is becoming quite the challenge. I am pretty darn close to just giving up :disappointed:

Then you're not going to like what I have to say next. Your commit is missing the `Signed-off-by` tag. Git can add it for you if you do `git commit --amend -s`.

Apologies for not doing a thorough-enough review to catch all of these changes in one go :pray:

robje:
First of, thanks for the detailed explanation.

> > I'm also not sure how jumping through all these hoops for a small change would help attract people "passing by" to help.
> 
> Incrementing `PKG_RELEASE` updates the package version string so that this appears as an upgrade for existing users. Having it in the same commit makes the changes atomic (there is no state where the functional change exists but `PKG_RELEASE` is unchanged). Including the package name as a prefix for the commit title makes it easier for people looking at the git log to understand what package(s) are affected by each commit.
> 
> There are reasons, if you care to understand them. They improve the quality of the project. They are also included in the [contribution guidelines](https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md).
>
> > getting this simple 1 line change merged feels is becoming quite the challenge. I am pretty darn close to just giving up disappointed
> 
> Then you're not going to like what I have to say next. Your commit is missing the `Signed-off-by` tag. Git can add it for you if you do `git commit --amend -s`.

Aren't there procedural ways, like hooks and/or checks, that can help?

For example a warning when the commit message doesn't start with `<word>:` and a check if `Signed-off-by` is in the commit message?

> Apologies for not doing a thorough-enough review to catch all of these changes in one go

Accepted. Lets make things better :smile:

jefferyto:
> Aren't there procedural ways, like hooks and/or checks, that can help?

Actually, there is :sweat_smile: I should have triggered the CI run right at the start, it would have flagged these corrections.

jefferyto (approved):
Thanks for sticking with this!

PR #18536, opened 2022-05-14
emailrelay: upgrade to v2.6 and add more options

stokito:
Maintainer: @fededim
Compile tested: x86_64

Description: fix of a bug when server_tls option break daemon start. Add more options.
Upgrade to v2.6. Support building with MbedTLS


stokito:
Also I think we should support --interface which is list. It may have a prefix so maybe it should be a separate section with advanced configuration.

Another big problem is that users still must edit emailrelay.auth and pop.auth manually.
This makes the UCI conf useless because ideally everything must be configurable from LUCI without a terminal.

There is a lot of rules
http://emailrelay.sourceforge.net/index.html#reference_md_Authentication

Looks complicated for me.
Maybe we can put something like PAM auth there by default.

stokito:
Also there is a new version v2.3. Could you please update it?

rodajc:
All options seem to work but one.

E-MailRelay V2.2 does not accept --server-tls-key option (tested on OpenWrt v22.03.2):
`emailrelay: error: invalid option: "--server-tls-key"`

nor second --server-tls-certificate:
`emailrelay: error: duplicate use of "--server-tls-certificate"`

So only way to pass both fullchain.pem and privkey.pem is to cat them into one file and pass that to E-MailRelay.

stokito:
yes, the --server-tls-key was added in v2.3
But the PR can be merged separately: just don't use the unsupported (yet) option

rodajc:
Ok. Well, looking forward to E-MailRelay V2.3 for OpenWrt then.

stokito:
There is a newer version 2.4 with a cool feature of direct delivery to SMTP by MX lookup.
The maintainer of the package doesn't response so maybe we should created a PR with a newer version ourselves.

rodajc:
@[stokito](https://github.com/stokito) I don't know how to do that. Do you?

And do you know where I can get the new sources to compile them myself meanwhile?

stokito:
@rodajc if you have a time you can try to make the PR yourself. I never did it but steps should be:
1. git clone https://github.com/openwrt/packages.git
2. Update version in the openwrt/packages/blob/master/mail/emailrelay/Makefile
3. PKG_VERSION: 2.4.1   PKG_HASH=35447b3783f07ace62ac4f8afd43d8ebe1112edd
4. Remove mail/emailrelay/patches/010-gcc11.patch
5. Build and test it on a router
6. Send the PR

rodajc:
I do have time but I don't have a clue.
After step one in packages/mail/emailrelay I have:
```
./Makefile
./files/emailrelay.config
./files/emailrelay.init
./files/emailrelay.auth
./patches/010-gcc11.patch

```

I can modify Makefile and remove the patch, but I don't see the new source to compile. I can't find the source either in the openwrt directory that I used to compile the firmware with emailrelay 2.2 that I am using currently, so you can safely say that I don't know what I'm doing...

stokito:
@rodajc I created the PR with version update.

You don't see sources of the EmailRelay because they are downloaded by OpenWrt's makefile during a build. This is a complicated procedure. You may find on wiki "OpenWrt Create package" or something like that. I'm not experienced here too so I don't want to give a bad instruction.

stokito:
If you just want to use tls you can do that even with the old version. You just need to merge privkey and fullchain and specify tls option with ext_command e.g. pass them manualy

rodajc:
I know. The new version is more compatible with Let's Encrypt exactly because you don't need to merge keys anymore.
One more thing, emailrelay does not appear in the available updates on my OpenWrt router. Do you know why not?

stokito:
Maybe there is no a package for your version of the OpenWrt. If you have a last version... my PR wasn't merged yet so you wont see it soon :(

rodajc:
I'll wait for the next OpenWrt stable release then.

rodajc:
So I waited for the new OpenWrt stable release which is OpenWrt 22.03.3.
I checked it out and compiled it, I flashed the result to the router and then there was a deep disappointment. E-MailRelay is still on V2.2!
What now? Will it ever merge into OpenWrt at all? If so, when? Isn't there any workaround to compile it already?

rodajc:
I finally managed to compile E-MailRelay V2.4.1.
So now I could test the config and the init script.
There seems to be an error in the init script for E-MailRelay:

`[ -n "$server_tls_key" ] && procd_append_param command --server-tls-certificate "$server_tls_certificate"`
should be:
`[ -n "$server_tls_key" ] && procd_append_param command --server-tls-certificate "$server_tls_key"`

The same may apply for $client_tls_key, but I don't use that, so I can't confirm.

stokito:
you right, will change that latter. Meanwhile author of the ER preparing v2.5 with many other changes that I waiting https://sourceforge.net/p/emailrelay/support-requests/75/
Going to test it latter.

stokito:
@rodajc I added a fix for the missing var. Please test. I can't compile for my router now

rodajc:
> @rodajc I added a fix for the missing var. Please test. I can't compile for my router now

Sorry for the delay. I will check the modification. Is that still with:
`PKG_VERSION:=2.4.1`
in Makefile?

stokito:
yes, the change is in the openwrt scripts but not in the ER itself.
Currently the next version of ER v2.5 is prepared and tested. Soon it should be released then I'll send a PR with the version upgrade

rodajc:
With:
`PKG_VERSION:=2.4.1`
in Makefile exactly the same package is compiled. Containing the same init script. Without the modification.

I assume this is as expected. The modification will be in a new version, right?

stokito:
This is a pull (merge) request (PR) with my patch.
OpenWrt devs didn't reviewed it yet and not merged into sources.

BKPepe (approved):

PR #18543, opened 2022-05-15
[openwrt-22.03] privoxy: update to 3.0.33 and fix the init script

hexian000:
Maintainer: @jefferyto 

Description:
Backport bug fixes from master.


jefferyto:
Thanks for your help on this!

PR #18559, opened 2022-05-17
strongswan: use own updown script

feckert:
Maintainer: @pprindeville, @Thermi
Compile tested: x86_64, APU3, OpenWrt 21.02
Run tested: x86_64, APU3, OpenWrt 21.02, start a connection

Description:

Stongswan does provide an own updown script in its source package.
This should serve as a basis and can be extended by the user. With
openwrt it is usual that a new updown script is installed with a
new version.

To fit better into the hotplug script handling of openwrt and because of
using iptables commands directly this commits moves the current source
updown script its own ipsec hotplug script.

Additionally a nftables updown package is added to fix strongswan with
fw4 (nftables).
From my point of view we should also update the ipsec init script to use
the new `_updown-openwrt` as default and not the `_updown` script from the
strongswan source

The following subtasks must be merged before hand:

- [x] https://github.com/openwrt/openwrt/pull/9909 merged https://github.com/openwrt/openwrt/commit/9379bc2fcf905568ef329a121c8c8a11fc98b02c
- [x] [fw4 include feature](https://patchwork.ozlabs.org/project/openwrt/patch/20220627105920.2317450-1-fe@dev.tdt.de/)

### DIFF
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -418,6 +418,22 @@ $(call Package/strongswan/description/Default)
  eap-peap, tnc-tnccs
 endef
 
+define Package/strongswan-updown-iptables
+  $(call Package/strongswan/Default)
+  TITLE+= updown iptables
+  DEPENDS:=strongswan \
+    +strongswan-mod-updown \
+    +iptables \
+    +IPV6:ip6tables \
+    +iptables-mod-ipsec \
+    +kmod-ipt-ipsec
+endef
+
+define Package/strongswan-updown-iptables/description
+  $(call Package/strongswan/description/Default)
+  This package contains the hotplug script for updown to add iptables rules.
+endef
+
 define BuildPlugin
   define Package/strongswan-mod-$(1)
     $$(call Package/strongswan/Default)
@@ -583,6 +599,12 @@ define Package/strongswan-libtls/install
 	$(CP) $(PKG_INSTALL_DIR)/usr/lib/ipsec/libtls.so.* $(1)/usr/lib/ipsec/
 endef
 
+define Package/strongswan-updown-iptables/install
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/ipsec
+	$(CP) ./files/etc/hotplug.d/ipsec/02-iptables \
+		$(1)/etc/hotplug.d/ipsec/02-iptables
+endef
+
 define Plugin/duplicheck/install
 	$(INSTALL_DIR) $(1)/usr/lib/ipsec/plugins
 	$(CP) $(PKG_INSTALL_DIR)/usr/lib/ipsec/duplicheck $(1)/usr/lib/ipsec/
@@ -649,6 +671,7 @@ $(eval $(call BuildPackage,strongswan-pki))
 $(eval $(call BuildPackage,strongswan-swanctl))
 $(eval $(call BuildPackage,strongswan-gencerts))
 $(eval $(call BuildPackage,strongswan-libtls))
+$(eval $(call BuildPackage,strongswan-updown-iptables))
 $(eval $(call BuildPlugin,addrblock,RFC 3779 address block constraint support,))
 $(eval $(call BuildPlugin,aes,AES crypto,))
 $(eval $(call BuildPlugin,agent,SSH agent signing,))
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -434,6 +434,20 @@ define Package/strongswan-updown-iptables/description
   This package contains the hotplug script for updown to add iptables rules.
 endef
 
+define Package/strongswan-updown-nftables
+  $(call Package/strongswan/Default)
+  TITLE+= updown nftables
+  DEPENDS:=strongswan \
+    +strongswan-mod-updown \
+    +nftables \
+    +kmod-nft-xfrm
+endef
+
+define Package/strongswan-updown-nftables/description
+  $(call Package/strongswan/description/Default)
+  This package contains the hotplug script for updown to add nftables rules.
+endef
+
 define BuildPlugin
   define Package/strongswan-mod-$(1)
     $$(call Package/strongswan/Default)
@@ -605,6 +619,24 @@ define Package/strongswan-updown-iptables/install
 		$(1)/etc/hotplug.d/ipsec/02-iptables
 endef
 
+define Package/strongswan-updown-nftables/install
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/ipsec
+	$(CP) ./files/etc/hotplug.d/ipsec/02-nftables \
+		$(1)/etc/hotplug.d/ipsec/02-nftables
+
+	$(INSTALL_DIR) $(1)/usr/share/nftables.d/chain-pre/input
+	$(LN) /tmp/strongswan/nftables.d/pre-input.nft \
+		$(1)/usr/share/nftables.d/chain-pre/input/10_strongswan.nft
+
+	$(INSTALL_DIR) $(1)/usr/share/nftables.d/chain-pre/output
+	$(LN) /tmp/strongswan/nftables.d/pre-output.nft \
+		$(1)/usr/share/nftables.d/chain-pre/output/10_strongswan.nft
+
+	$(INSTALL_DIR) $(1)/usr/share/nftables.d/chain-pre/forward
+	$(LN) /tmp/strongswan/nftables.d/pre-forward.nft \
+		$(1)/usr/share/nftables.d/chain-pre/forward/10_strongswan.nft
+endef
+
 define Plugin/duplicheck/install
 	$(INSTALL_DIR) $(1)/usr/lib/ipsec/plugins
 	$(CP) $(PKG_INSTALL_DIR)/usr/lib/ipsec/duplicheck $(1)/usr/lib/ipsec/
@@ -672,6 +704,7 @@ $(eval $(call BuildPackage,strongswan-swanctl))
 $(eval $(call BuildPackage,strongswan-gencerts))
 $(eval $(call BuildPackage,strongswan-libtls))
 $(eval $(call BuildPackage,strongswan-updown-iptables))
+$(eval $(call BuildPackage,strongswan-updown-nftables))
 $(eval $(call BuildPlugin,addrblock,RFC 3779 address block constraint support,))
 $(eval $(call BuildPlugin,aes,AES crypto,))
 $(eval $(call BuildPlugin,agent,SSH agent signing,))
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -610,8 +610,11 @@ endef
 
 define Plugin/updown/install
 	$(INSTALL_DIR) $(1)/usr/lib/ipsec/plugins
-	$(CP) $(PKG_INSTALL_DIR)/usr/lib/ipsec/_updown $(1)/usr/lib/ipsec/
 	$(CP) $(PKG_INSTALL_DIR)/usr/lib/ipsec/plugins/libstrongswan-updown.so $(1)/usr/lib/ipsec/plugins/
+
+	# Install OpenWrt updown script to call all scripts in '/etc/hotplug.d/ipsec'
+	$(CP) ./files/usr/lib/ipsec/_updown-owrt $(1)/usr/lib/ipsec
+
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/ipsec
 	$(CP) ./files/etc/hotplug.d/ipsec/01-user \
 		$(1)/etc/hotplug.d/ipsec/01-user
@@ -717,7 +720,7 @@ $(eval $(call BuildPlugin,sshkey,SSH key decoding,))
 $(eval $(call BuildPlugin,stroke,Stroke,+strongswan-charon +strongswan-ipsec))
 $(eval $(call BuildPlugin,test-vectors,crypto test vectors,))
 $(eval $(call BuildPlugin,unity,Cisco Unity extension,))
-$(eval $(call BuildPlugin,updown,updown firewall,+iptables +IPV6:ip6tables +iptables-mod-ipsec +kmod-ipt-ipsec))
+$(eval $(call BuildPlugin,updown,script execution,))
 $(eval $(call BuildPlugin,vici,Versatile IKE Configuration Interface,))
 $(eval $(call BuildPlugin,whitelist,peer identity whitelisting,))
 $(eval $(call BuildPlugin,wolfssl,WolfSSL crypto,+PACKAGE_strongswan-mod-wolfssl:libwolfssl))
--- a/net/strongswan/files/etc/hotplug.d/ipsec/02-nftables
+++ b/net/strongswan/files/etc/hotplug.d/ipsec/02-nftables
@@ -1,0 +1,488 @@
+#!/bin/sh
+# default updown script
+#
+# Copyright (C) 2003-2004 Nigel Meteringham
+# Copyright (C) 2003-2004 Tuomo Soini
+# Copyright (C) 2002-2004 Michael Richardson
+# Copyright (C) 2005-2007 Andreas Steffen <andreas.steffen@strongswan.org>
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 2 of the License, or (at your
+# option) any later version.  See <http://www.fsf.org/copyleft/gpl.txt>.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# for more details.
+
+# CAUTION:  Installing a new version of strongSwan will install a new
+# copy of this script, wiping out any custom changes you make.  If
+# you need changes, make a copy of this under another name, and customize
+# that, and use the (left/right)updown parameters in ipsec.conf to make
+# strongSwan use yours instead of this default one.
+#      PLUTO_VERSION
+#              indicates  what  version of this interface is being
+#              used.  This document describes version  1.1.   This
+#              is upwardly compatible with version 1.0.
+#
+#       PLUTO_VERB
+#              specifies the name of the operation to be performed
+#              (prepare-host, prepare-client, up-host, up-client,
+#              down-host, or down-client).  If the address family
+#              for security gateway to security gateway communica-
+#              tions is IPv6, then a suffix of -v6 is added to the
+#              verb.
+#
+#       PLUTO_CONNECTION
+#              is the name of the  connection  for  which  we  are
+#              routing.
+#
+#       PLUTO_INTERFACE
+#              is the name of the ipsec interface to be used.
+#
+#       PLUTO_REQID
+#              is the reqid of the AH|ESP policy
+#
+#       PLUTO_PROTO
+#              is the negotiated IPsec protocol, ah|esp
+#
+#       PLUTO_IPCOMP
+#              is not empty if IPComp was negotiated
+#
+#       PLUTO_UNIQUEID
+#              is the unique identifier of the associated IKE_SA
+#
+#       PLUTO_ME
+#              is the IP address of our host.
+#
+#       PLUTO_MY_ID
+#              is the ID of our host.
+#
+#       PLUTO_MY_CLIENT
+#              is the IP address / count of our client subnet.  If
+#              the  client  is  just  the  host,  this will be the
+#              host's own IP address / max (where max  is  32  for
+#              IPv4 and 128 for IPv6).
+#
+#       PLUTO_MY_SOURCEIP
+#       PLUTO_MY_SOURCEIP4_$i
+#       PLUTO_MY_SOURCEIP6_$i
+#              contains IPv4/IPv6 virtual IP received from a responder,
+#              $i enumerates from 1 to the number of IP per address family.
+#              PLUTO_MY_SOURCEIP is a legacy variable and equal to the first
+#              virtual IP, IPv4 or IPv6.
+#
+#       PLUTO_MY_PROTOCOL
+#              is the IP protocol that will be transported.
+#
+#       PLUTO_MY_PORT
+#              is  the  UDP/TCP  port  to  which  the IPsec SA  is
+#              restricted on our side.  For ICMP/ICMPv6 this contains the
+#              message type, and PLUTO_PEER_PORT the message code.
+#
+#       PLUTO_PEER
+#              is the IP address of our peer.
+#
+#       PLUTO_PEER_ID
+#              is the ID of our peer.
+#
+#       PLUTO_PEER_CLIENT
+#              is the IP address / count of the peer's client sub-
+#              net.   If the client is just the peer, this will be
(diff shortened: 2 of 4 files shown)


feckert:
@pprindeville I have just set up an environment with which I am testing the strongswan with the new fw4 (nft) of openwrt.

Currently it is not possible in my setup to ping the remote site from the router with the current *tables-nft updown script! 
Something is wrong here! If I stop the firewall with /etc/init.d/firewall then the tunnel works out of the box.
Unfortunately, I am still at the beginning of my investigation.
But the first thing from my point of view is to clean up the current `_updown` script and move it to the ipsec hotplug handling of openwrt.
That way, we can more easily replace the iptables script with an nftables script later.

net/strongswan/files/usr/lib/ipsec/_updown
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
pprindeville, line 1: What's wrong with the `./src/_updown/_updown` script in `$(PKG_INSTALL_DIR)`?

pprindeville (changes requested):
Don't add files that are already in the source-tree.

net/strongswan/files/usr/lib/ipsec/_updown
feckert: As explained above, the script only works with iptables so far. But in the openwrt-master and openwrt-22.03 branch we have **nft** in use as the default firewall backend with `fw4`. The `_updown` script is a sample implementation from strongswan.
I extended it years ago in openwrt with this [patch](https://github.com/openwrt/packages/commit/6cd8fcabe6d1727192bf447c7adc8e1eb42ab8f7) so that we have a hotplug mechanism and don't always have to change the `_updown` script on every system upgrade. This is because it is overwritten with every system upgrade.

With this pull request, I want to change the `_updown` script so that we trigger a hotplug event in this script and then all scripts under `/etc/hotplug.d/ipsec` are processed. To ensure that the rules for the old iptables are also executed, I have moved the `iptables` stuff of the upstream `_updown` script for `iptables` to the `02-iptables`.
With this change, the `_updown` handling is modular and I can now work on the nft integration.
The nft commands will then be included in the 02-nft script, which I am writing just now.
It works as expected. I will update this pullrequest when I am ready with my change

jow-:
@feckert - support for fw4 auto-includes has been merged into master and openwrt-22.03 now. The final path is `/usr/share/nftables.d/` - if you update it accordingly in your PR it should be ready to merge.

feckert:
I am on vacation till monday. I do not have the possibility to update this
PR. Could to this on thusday.

Best regards

Flo

Jo-Philipp Wich ***@***.***> schrieb am Do., 25. Aug. 2022,
10:24:

feckert:
@pprindeville All dependencies have now been merge.
If there are no objections from your side, then I would like to merge this and also backport it to owrt-22.03.

In my opinion, otherwise the strongswan for generated traffic from the router is broken, since the required firewall rules are not added for fw4 (nftables). This pullrequest fixes this and leaves the user the option of using nftables via fw4 or iptables for fw3.

stintel:
Honestly I don't see why this is needed. Use xfrm interfaces and add them in a firewall zone. No custom firewall needed at all.

feckert:
> Honestly I don't see why this is needed. Use xfrm interfaces and add them in a firewall zone. No custom firewall rules and updown scripts needed at all.

@stintel You may be right to use in the feature xfrm interfaces, but I currently I have the following setup in the field that previously worked with fw3 (iptables) with `strongswan-mod-updown`. And now it no longer works with the fw4 (nftables) setup.

I also added this two new packages to install either `strongswan-updown-iptables` or `strongswan-updown-nftables` independently of the `strongswan-mod-updown`.

Because if I change in the feature to xfrm interfaces I only need `strongswan-mod-updown` for hotplgug events `/etc/hotplug.d/strongswan` handling to trigger some actions on client up/down events (Enable LEDs on client connection for example)
But Currently the package `strongswan-mod-updown` is also installing on this event the the iptabels rules which I do not need if I use xfrm interfaces.

So I think we anyway have to move the related firewall stuffs for `strongswan-mod-updwon` for iptables into its own package `strongswan-updown-iptables` to fix the explained issues. Therefore, we can also fix the handling for the fw4 (nftables) and therefore outsource the setting of nftables rules into a separate package `strongswan-updown-nftables`.

stintel:
I won't be using it, so I don't care to review it. My suggestion would be to contribute this upstream, and then backport to OpenWrt.

feckert:
Anyone else from the package maintainers @pprindeville @Thermi using this?

> I won't be using it, so I don't care to review it.

Too bad but I can understand if you don't use it and are not interested in a review.

> My suggestion would be to contribute this upstream, and then backport to OpenWrt.

I can try, but it takes a long time to get it into the openwrt. I would suggest to merge this and if it is not working we could revert this.

The fact is currently it doesn't work without xfrm interfaces because the new fw4 (nftables) is not supported by `strongswan-mod-updown`

Thermi:
The updown script isn't getting any love upstream because it's considered deprecated because it's only badly suited for handling of any configuration because it breaks on the first reauthentication or rekeying of CHILD_SAs. I am working on an updownv2 plugin that enables proper handling of such scenarios.

pprindeville:
> The updown script isn't getting any love upstream because it's considered deprecated because it's only badly suited for handling of any configuration because it breaks on the first reauthentication or rekeying of CHILD_SAs. I am working on an updownv2 plugin that enables proper handling of such scenarios.

I'm inclined to wait until something happens upstream also...

feckert:
@Thermi @pprindeville Ok never the less we should at least change the dependency of strongswan-mod-updown.
Otherwise we will always install the iptables dependencies when installing strongswan-mod-updown. Even though I don't need it. 

If strongswan does not support nft from upstream or openwrt downstream then we should at least install only strongswan-mod-updown without any dependencies. If some on needs the iptables stuff then the new packeges strongswan-updown-iptables must also gets installed (See my pullrequest). Because I am using updown for other things not only for the firewall handling. The updown script for the firewall is only called if the option left|right firewall is set in the strongswan configuration.

pprindeville:
> Anyone else from the package maintainers @pprindeville @Thermi using this?

I'm using `xfrm` interfaces because they integrate well into the firewall per-interface rules...

net/strongswan/Makefile
```diff
@@ -441,6 +441,36 @@ $(call Package/strongswan/description/Default)
...
+endef
+
+define Package/strongswan-updown-iptables/description
+$(call Package/strongswan/description/Default)
+ This package contains the hotplug script for updown to add iptables rules.
+endef
+
+define Package/strongswan-updown-nftables
+$(call Package/strongswan/Default)
+  TITLE+= updown iptables
```
openwrt-ai: nit: copy/paste leftover — the nftables package title reads `updown iptables`.

```suggestion
  TITLE+= updown nftables
```

---

net/strongswan/files/etc/hotplug.d/ipsec/02-nftables
```diff
@@ -0,0 +1,378 @@
...
+				"${PLUTO_REQID}" \
+				"${PLUTO_PEER}" \
+				"${PLUTO_ME}" >> "${PRE_INPUT_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
+			fi
+			log_connection "up-host-v6" "128" ""
+			;;
+		down-host-v6:iptables)
+			rm -rf "${PRE_INPUT_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
+			rm -rf "${PRE_OUTPUT_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
+			log_connection "down-host" "32" ""
```
openwrt-ai: The `up-host-v6` case writes its rules to `ipv6-${PLUTO_REQID}.nft`, but this teardown removes the `ipv4-` files. The IPv6 host rules are therefore never cleaned up on disconnect (and the matching IPv4 host rule would be removed instead). The log label/mask are also the v4 ones.

```suggestion
			rm -rf "${PRE_INPUT_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
			rm -rf "${PRE_OUTPUT_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
			log_connection "down-host-v6" "128" ""
```

---

net/strongswan/files/etc/hotplug.d/ipsec/02-nftables
```diff
@@ -0,0 +1,378 @@
...
+				rm -rf "${PRE_FORWARD_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
+			fi
+			if [ -n "$PLUTO_MY_SOURCEIP" -o -n "$PLUTO_HOST_ACCESS" ]; then
+				rm -rf "${PRE_INPUT_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
+				rm -rf "${PRE_OUTPUT_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
+			fi
+			if [ -n "$PLUTO_IPCOMP" ]; then
+				rm -rf "${PRE_INPUT_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
+			fi
+			log_connection "down-client-v6" "128" "$PLUTO_MY_CLIENT"
```
openwrt-ai: Same v4/v6 mix-up as `down-host-v6`: `up-client-v6` created the rules under `ipv6-${PLUTO_REQID}.nft`, but this teardown removes the `ipv4-` files, so the IPv6 forward/input/output rules are never removed on disconnect. The `$PLUTO_MY_SOURCEIP/32` guard should also be `/128` for IPv6.

```suggestion
			if [ "$PLUTO_PEER_CLIENT" != "$PLUTO_MY_SOURCEIP/128" ]; then
				rm -rf "${PRE_FORWARD_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
			fi
			if [ -n "$PLUTO_MY_SOURCEIP" -o -n "$PLUTO_HOST_ACCESS" ]; then
				rm -rf "${PRE_INPUT_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
				rm -rf "${PRE_OUTPUT_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
			fi
			if [ -n "$PLUTO_IPCOMP" ]; then
				rm -rf "${PRE_INPUT_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
			fi
			log_connection "down-client-v6" "128" "$PLUTO_MY_CLIENT"
```

---

net/strongswan/files/etc/hotplug.d/ipsec/02-nftables
```diff
@@ -0,0 +1,378 @@
...
+			fi
+			# allow IPIP traffic because of the implicit SA created by the kernel if
+			# IPComp is used (for small inbound packets that are not compressed).
+			if [ -n "$PLUTO_IPCOMP" ]; then
+				printf "protocol %s meta iifname %s meta ipsec exists ipsec in reqid %s ip6 saddr %s ip6 daddr %s\n" \
+					"41" \
+					"${PLUTO_INTERFACE}" \
+					"${PLUTO_REQID}" \
+					"${PLUTO_PEER}" \
+					"${PLUTO_ME}" >> "${PRE_INPUT_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
```
openwrt-ai: Two issues in this `up-client-v6` IPComp rule: it is written to the `ipv4-` file (the rest of `up-client-v6` uses `ipv6-`), and the rule has no verdict — every other rule ends in `accept`, so as written this line is a no-op and inbound IPComp/IPIP6 packets won't be accepted. (The IPv4 `up-client` IPComp rule is missing `accept` too.) Also note the guard on line 296 uses `/32` rather than `/128`.

```suggestion
				printf "protocol %s meta iifname %s meta ipsec exists ipsec in reqid %s ip6 saddr %s ip6 daddr %s accept\n" \
					"41" \
					"${PLUTO_INTERFACE}" \
					"${PLUTO_REQID}" \
					"${PLUTO_PEER}" \
					"${PLUTO_ME}" >> "${PRE_INPUT_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
```

---

net/strongswan/files/etc/hotplug.d/ipsec/02-nftables
```diff
@@ -0,0 +1,378 @@
...
+					"${PLUTO_PEER_CLIENT}" \
+					"${PLUTO_MY_CLIENT}" > "${PRE_INPUT_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
+				printf "meta oifname %s ipsec out reqid %s ip saddr %s ip daddr %s accept\n" \
+					"${PLUTO_INTERFACE}" \
+					"${PLUTO_REQID}" \
+					"${PLUTO_MY_CLIENT}" \
+					"${PLUTO_PEER_CLIENT}" > "${PRE_OUTPUT_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
+			fi
+			# allow IPIP traffic because of the implicit SA created by the kernel if
+			# IPComp is used (for small inbound packets that are not compressed).
```
openwrt-ai: This IPComp rule has no verdict, unlike the `up-host` equivalent which ends in `accept`. Without a terminal statement the rule matches but does nothing, so inbound IPIP packets for compressed traffic won't be accepted.

```suggestion
				printf "protocol %s meta iifname %s meta ipsec exists ipsec in reqid %s ip saddr %s ip daddr %s accept\n" \
```

---

openwrt-ai (commented):
Reviewed 3 new commits. The refactor into separate `strongswan-updown-iptables`/`strongswan-updown-nftables` packages looks reasonable; comments below focus on the new `02-nftables` hotplug script, where the IPv6 code paths appear to have been copied from the IPv4 ones without fully adjusting the `ipv4-`/`ipv6-` state filenames and `/32` vs `/128` masks.

## Commit checks
- `strongswan: add new nftables updown content to own package` — the commit body reads "Move iptables related _updown stuff to its own package. This package should be enabled and used, if fw4 is the default firewall." This looks copied from the iptables commit; it should refer to nftables.

---

net/strongswan/files/etc/hotplug.d/ipsec/02-nftables
```diff
@@ -0,0 +1,488 @@
...
+				busybox sed -e "$line" -i "${PRE_OUTPUT_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
+			fi
+			if [ -n "$PLUTO_IPCOMP" ]; then
+				line="$(printf "/protocol %s meta iifname %s meta ipsec exists ipsec in reqid %s ip6 saddr %s ip6 daddr %s/d" \
+					"41" \
+					"${PLUTO_INTERFACE}" \
+					"${PLUTO_REQID}" \
+					"${pluto_peer}" \
+					"${pluto_me}")"
+				busybox sed -e "$line" -i "${PRE_INPUT_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
```
openwrt-ai: This IPComp teardown in the `down-client-v6` case deletes from the `ipv4-` state file, but every other rule in this case targets `ipv6-` (forward/input/output just above). The paired insertion in `up-client-v6` also writes this rule to the `ipv4-` file ([02-nftables#L406](https://github.com/openwrt/packages/blob/c9d4ad59441e551380577567d6969dbc93ed9200/net/strongswan/files/etc/hotplug.d/ipsec/02-nftables#L406)), so the two currently cancel out — but they diverge from the host path, where the v6 IPComp rule correctly lives in `ipv6-` ([up-host-v6 #L169](https://github.com/openwrt/packages/blob/c9d4ad59441e551380577567d6969dbc93ed9200/net/strongswan/files/etc/hotplug.d/ipsec/02-nftables#L169) / [down-host-v6 #L193](https://github.com/openwrt/packages/blob/c9d4ad59441e551380577567d6969dbc93ed9200/net/strongswan/files/etc/hotplug.d/ipsec/02-nftables#L193)). Net effect: the v6 client IPComp rule ends up in a spuriously-named `ipv4-<reqid>.nft` file. No suggestion block, because the fix must change both this line and the insertion at line 406 to `ipv6-` together — changing only one would leave a stale rule on teardown.

---

openwrt-ai (commented):
Reviewed the updated `02-nftables`: the teardown logic was reworked from `rm -rf`-ing the whole per-reqid state file to deleting individual rule lines with `busybox sed`, matching the switch to appending rules with `>>`. That direction is sound. One inline comment below on an `ipv4`/`ipv6` state-file mismatch in the v6 IPComp path.

---

net/strongswan/files/etc/hotplug.d/ipsec/02-nftables
```diff
@@ -0,0 +1,488 @@
...
+				"${PLUTO_INTERFACE}" \
+				"${PLUTO_REQID}" \
+				"${pluto_peer}" \
+				"${pluto_me}")"
+			busybox sed -e "$line" -i "${PRE_INPUT_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
+		fi
+		log_connection "down-host" "32" ""
+		;;
+	up-client-v6:iptables)
+		if [ "$PLUTO_PEER_CLIENT" != "$PLUTO_MY_SOURCEIP/32" ]; then
```
openwrt-ai: In the IPv6 client paths the host-to-host guard compares against `$PLUTO_MY_SOURCEIP/32`, but for IPv6 `PLUTO_PEER_CLIENT` is a `/128`, so this test can never be true and the forward rules are always added even for a pure host connection. The sibling iptables script in this same PR uses `/128` here ([`02-iptables` L467](https://github.com/openwrt/packages/blob/60375e5c458bbf135e0e877fea5783d317c2f9c6/net/strongswan/files/etc/hotplug.d/ipsec/02-iptables#L467) and [L519](https://github.com/openwrt/packages/blob/60375e5c458bbf135e0e877fea5783d317c2f9c6/net/strongswan/files/etc/hotplug.d/ipsec/02-iptables#L519)).

```suggestion
		if [ "$PLUTO_PEER_CLIENT" != "$PLUTO_MY_SOURCEIP/128" ]; then
```

The same `/32` → `/128` change is needed in the `down-client-v6` guard at line 411 so the up/down rules stay symmetric.

---

net/strongswan/files/etc/hotplug.d/ipsec/02-nftables
```diff
@@ -0,0 +1,488 @@
...
+			printf "meta oifname %s ipsec out reqid %s ip saddr %s ip daddr %s accept\n" \
+				"${PLUTO_INTERFACE}" \
+				"${PLUTO_REQID}" \
+				"${PLUTO_MY_CLIENT}" \
+				"${PLUTO_PEER_CLIENT}" >>"${PRE_OUTPUT_RULES_DIR}/ipv4-${PLUTO_REQID}.nft"
+		fi
+		# allow IPIP traffic because of the implicit SA created by the kernel if
+		# IPComp is used (for small inbound packets that are not compressed).
+		if [ -n "$PLUTO_IPCOMP" ]; then
+			printf "protocol %s meta iifname %s meta ipsec exists ipsec in reqid %s ip saddr %s ip daddr %s\n" \
```
openwrt-ai: This IPComp/IPIP exception rule for `up-client` is emitted without a verdict — the format string ends at `ip daddr %s\n` with no trailing `accept`. Every other rule in this file (e.g. the `up-host` IPComp rule at line 209) ends with `accept`, and the iptables sibling appends `-j ACCEPT` for the same case ([`02-iptables` L338-L345](https://github.com/openwrt/packages/blob/60375e5c458bbf135e0e877fea5783d317c2f9c6/net/strongswan/files/etc/hotplug.d/ipsec/02-iptables#L338-L345)). As written the packet just falls through instead of being accepted, so the "allow IPIP traffic" intent is not achieved.

The same omission is present in the `up-client-v6` rule at line 401. Note that if you add `accept`, the matching teardown `sed` patterns (`down-client` at line 311 and `down-client-v6` at line 441) must also gain a trailing `accept` so the delete pattern still matches the emitted rule.

---

net/strongswan/files/etc/hotplug.d/ipsec/02-nftables
```diff
@@ -0,0 +1,488 @@
...
+		if [ -n "$PLUTO_IPCOMP" ]; then
+			line="$(printf "/protocol %s meta iifname %s meta ipsec exists ipsec in reqid %s ip6 saddr %s ip6 daddr %s accept/d" \
+				"41" \
+				"${PLUTO_INTERFACE}" \
+				"${PLUTO_REQID}" \
+				"${pluto_peer}" \
+				"${pluto_me}")"
+			busybox sed -e "$line" -i "${PRE_INPUT_RULES_DIR}/ipv6-${PLUTO_REQID}.nft"
+		fi
+		log_connection "down-host" "32" ""
```
openwrt-ai: nit: this is the `down-host-v6` case, but it logs the action as `down-host` with mask `32`. The matching `up-host-v6` case logs `128` (line 345), and the iptables sibling logs `down-host-v6`/`128` ([`02-iptables` L461](https://github.com/openwrt/packages/blob/60375e5c458bbf135e0e877fea5783d317c2f9c6/net/strongswan/files/etc/hotplug.d/ipsec/02-iptables#L461)).

```suggestion
		log_connection "down-host-v6" "128" ""
```

---

openwrt-ai (commented):
The branch was force-pushed since my last review; the previously reviewed `02-nftables` (with the `>>`/`sed`-per-line teardown) has been replaced by an earlier variant that rebuilds each per-reqid file and regenerates the include files. Re-reviewed the current head against the sibling `02-iptables` in this PR. The IPv4/IPv6 state-file mismatch I flagged earlier is no longer present, but a few IPv6/IPComp adaptation issues remain — see inline comments.

## Commit checks
- `strongswan: add new nftables updown content to own package` (60375e5c) — the body's first sentence still reads "Move **iptables** related _updown stuff to its own package", which is copied from the iptables commit; this commit adds the nftables script, so it should say nftables (the "if fw4 is the default firewall" part is correct).

---

openwrt-ai (commented):
Reviewed 3 new commits; no new issues found.

The force-push resolves the items from my previous reviews: the client-side IPComp rules now carry `accept` (with matching `sed` teardown), the v6 client source-IP checks use `/128`, `down-host-v6` logs the correct verb/mask, and the nftables commit body now refers to nftables/fw4.

---

feckert:
@BKPepe So, I've now fixed all the issues the AI flagged. Just to be clear: Should I mark this as "resolved" like I've done before? Or what's the procedure here now?

BKPepe:
Hi Florian,

I think I know where the misunderstanding is. We currently have two bots in this repository (sooner in this org). One is openwrt [Bot], which checks for formalities and other basic things, and the second is @openwrt-ai. The latter is actually set up as a regular account, but it's basically powered by Claude Code right now and is meant for code reviews. Ultimately, it doesn't really matter. A bot is a bot, and lately, it's been absolutely killing it here. 🥳 🎉 

The openwrt [bot] cleans up after itself, meaning that if you fix everything it pointed out, its comment will automatically delete itself, so that's pretty much all good. What you're running into here is actually @openwrt-ai, which is capable of leaving a ton of comments in a pull request. Personally, I don't click "resolved" on them. It feels unnecessary, but if it's your workflow to do so and you want to mark them as resolved in your pull requests or wherever you come across them, go right ahead. I'm not aware of anyone else using a different or specific workflow. You could always bring it up on the mailing list, but it's up to you.

PR #18573, opened 2022-05-19
wifi-presence: Add config for process user/group

awilliams:
On systems using seccomp, the hostapd socket files will be owned by the
'network' user/group ([source][0]). In this case, if wifi-presence is
run as root/root, then it does not have permissions to open the
hostapd socket files. 

This was discussed in awilliams/wifi-presence#3.

This change allows the process user/group to be specified in
`/etc/config/wifi-presence`. If no explicit user/group is set, then the
init script will use the owner of the socket files in `/var/run/hostapd/`
to determine the appropriate process user/group.

[0]: https://github.com/openwrt/openwrt/blob/ec6293febc244d187e71a6e54f44920be679cde4/package/network/services/hostapd/files/wpad.init#L35-L36

### DIFF
--- a/net/wifi-presence/Makefile
+++ b/net/wifi-presence/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=wifi-presence
 PKG_VERSION:=0.1.2
-PKG_RELEASE:=1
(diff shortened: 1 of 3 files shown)


awilliams:
cc/ @jmccrohan

PR #18576, opened 2022-05-19
curl: bump to 7.83.1

stangri:
Maintainer: me
Compile tested: x86_64, Sophos SG-135, OpenWrt 21.02.3

Description:
* https://curl.se/changes.html#7_83_1


stangri:
I don't understand why `wolfssl` is failing to build and would appreciate some guidance.

neheb:
it is curl that's failing: https://github.com/openwrt/packages/blob/master/net/curl/patches/0001-wolfssl-fix-compiler-error-without-IPv6.patch

Beginner-Go:
Just refresh the [patch](https://github.com/Beginner-Go/packages/commit/ef064b47dc3e61483f10ffe360f1614ec83b2e57)

Beginner-Go:
Just drop the [patch](https://github.com/openwrt/packages/blob/master/net/curl/patches/0001-wolfssl-fix-compiler-error-without-IPv6.patch)

wecand0:
Stangri, did you manage to use the latest curl and wolfssl, I failed to do it, I do always right, my code works perfectly with OpenSSL, but I do not know what is hell when I try use wolfssl.

stangri:
@wecand0 yes, forgot to delete an extra patch file as @neheb has pointed out.

PR #1538, opened 2015-07-04
zmq: Make 2 variants, refresh patches, fix C++ build issues

thess:
* Correct package md5sum
* Create 2 variants of libzmq (libzmq-nc without CurveZMQ/libsodium; libzmq-curve with CurveZMQ)
* Re-number/refresh patches
* Add patch to correct C++ include file order errors

Signed-off-by: Ted Hess <thess@kitschensync.net>

Issue #12793, opened 2020-07-12
golang: Makefile:380: *** go-bootstrap cannot be installed on linux/arm64.  Stop.


alex9434:
Maintainer: @jefferyto @kuoruan 
Environment: aarch64, Raspberry Pi 4B, Snapshot
Build system: Linux debian-buster-64 4.19.118-v8+ #1311 SMP PREEMPT Mon Apr 27 14:32:38 BST 2020 aarch64 GNU/Linux

Description:
Package docker-ce does not build on Debian 10 Aarch64. error message:
```
make[3]: Leaving directory '/home/pi/Downloads/openwrt/feeds/packages/utils/cgroupfs-mount'
time: package/feeds/packages/cgroupfs-mount/compile#0.50#0.54#1.03
make[3]: Entering directory '/home/pi/Downloads/openwrt/feeds/packages/lang/golang/golang'
Makefile:380: *** go-bootstrap cannot be installed on linux/arm64.  Stop.
make[3]: Leaving directory '/home/pi/Downloads/openwrt/feeds/packages/lang/golang/golang'
time: package/feeds/packages/golang/host-compile#0.19#0.06#0.28
make[2]: *** [package/Makefile:111: package/feeds/packages/golang/host/compile] Error 2
make[2]: Leaving directory '/home/pi/Downloads/openwrt'
make[1]: *** [package/Makefile:107: /home/pi/Downloads/openwrt/staging_dir/target-aarch64_cortex-a72_musl/stamp/.package_compile] Error 2
make[1]: Leaving directory '/home/pi/Downloads/openwrt'
make: *** [/home/pi/Downloads/openwrt/include/toplevel.mk:235: world] Error 2

```

jefferyto:
1. Install Go on your build system, either using your distro’s packages or from golang.org.
2. In `make menuconfig`, find the “ External bootstrap Go root directory” option for the golang package and enter the path for the working Go tree installed in the previous step.
3. Build as normal.

jefferyto:
If you installed Go using the Debian package, then the working Go tree path would be `/usr/lib/go-1.11` (or `/usr/lib/go-1.14` if you are using buster-backports).

The instructions I provided should have resolved your issue. Please comment if you are able to build Go/Docker successfully or if you have encountered additional issues (or if the instructions were unclear).

[...]

PR #18584, opened 2022-05-20
golang: some makefile fixes

mpratt14:
Maintainer: @jefferyto 
Compile tested: ramips master

Host:
Linux 5.15.32-v8+ #1538 SMP PREEMPT Thu Mar 31 19:40:39 BST 2022 aarch64 GNU/Linux
Raspbian GNU/Linux 11 (bullseye)

my system is a Raspberry Pi 4 and Go was failing to build in 64-bit since the Raspbian version is 32-bit

also some other small fixes


lang/golang/golang/Makefile
```diff
@@ -14,7 +14,8 @@ PKG_NAME:=golang
 PKG_VERSION:=$(GO_VERSION_MAJOR_MINOR)$(if $(GO_VERSION_PATCH),.$(GO_VERSION_PATCH))
 PKG_RELEASE:=1
 
-GO_SOURCE_URLS:=https://dl.google.com/go/ \
+GO_SOURCE_URLS:=https://go.dev/dl/ \
```
jefferyto, line 17: `https://go.dev/dl/go1.18.2.src.tar.gz` is a 302 redirect to `https://dl.google.com/go/go1.18.2.src.tar.gz` so this is unnecessary.

lang/golang/golang/Makefile
```diff
@@ -202,17 +213,21 @@ endif
 $(eval $(call GoCompiler/AddProfile,Host,$(HOST_BUILD_DIR),$(HOST_GO_PREFIX),$(HOST_GO_VERSION_ID),$(GO_HOST_OS_ARCH),$(HOST_GO_INSTALL_SUFFIX)))
 
 HOST_GO_VARS= \
+	GOHOSTOS="$(GO_HOST_OS)" \
+	GOHOSTARCH="$(GO_HOST_ARCH)" \
```
jefferyto, line 217: Normally `GOHOSTARCH` (and `GOHOSTOS`) are auto-detected by go itself, so there should be no need to set it. What error messages are you getting?

lang/golang/golang/Makefile
```diff
@@ -202,17 +213,21 @@ endif
...
 	CC="$(HOSTCC_NOCACHE)" \
 	CXX="$(HOSTCXX_NOCACHE)"
 
-define Host/Compile
+define Host/Configure
 	$(call GoCompiler/Bootstrap/CheckHost,$(BOOTSTRAP_GO_VALID_OS_ARCH))
 	$(call GoCompiler/Host/CheckHost,$(HOST_GO_VALID_OS_ARCH))
 
 	mkdir -p "$(GO_BUILD_CACHE_DIR)"
+endef
```
jefferyto, line 228: Is there any advantage in having these checks (and creating the cache directory) in the configure step instead of compile?

lang/golang/golang/Makefile
mpratt14: ok I dropped it

lang/golang/golang/Makefile
mpratt14: it makes the makefile and output easier to understand
basically just to be more organized

lang/golang/golang/Makefile
mpratt14: it is autodetected as "arm64" but I am using external bootstrap from a Go install that is "arm" and build fails

I'll post the error soon

mpratt14: in the meantime I can try dropping GOHOSTOS= and see if it still works

mpratt14:
@jefferyto this is the original error during the last part of host compile,
(fails before it can get to go-cross)
it doesn't really give much useful information anyway...

I have also seen it happen during the earlier part
`Building Go toolchain3 using go_bootstrap and Go toolchain2.`

my first build attempt failed at bootstrap compile just like in #12793
which is how I found your directions there
but it still didn't work after that because the local Go at `/usr/lib/go/bin/go` is "arm" not "arm64"
but it autodetects "arm64" which is from the kernel being "aarch64"

```
Building packages and commands for linux/arm64.
container/list
container/ring
...
...
...
cmd/vendor/golang.org/x/tools/go/analysis/unitchecker
cmd/vet
go tool dist: FAILED: /media/SDStorage/builds/github-openwrt/openwrt/build_dir/target-mipsel_24kc_musl/go-1.18.2/pkg/tool/linux_arm64/go_bootstrap install -gcflags=all= -ldflags=all= -v std cmd: exit status 2
make[2]: *** [Makefile:391: /media/SDStorage/builds/github-openwrt/openwrt/build_dir/target-mipsel_24kc_musl/go-1.18.2/.built] Error 2
make[2]: Leaving directory '/media/SDStorage/builds/github-openwrt/openwrt/feeds/packages/lang/golang/golang'
time: package/feeds/packages/golang/compile#1128.52#149.96#594.71
    ERROR: package/feeds/packages/golang failed to build.
make[1]: *** [package/Makefile:116: package/feeds/packages/golang/compile] Error 1
make[1]: Leaving directory '/media/SDStorage/builds/github-openwrt/openwrt'
make: *** [/media/SDStorage/builds/github-openwrt/openwrt/include/toplevel.mk:233: package/feeds/packages/golang/compile] Error 2
```


by the way I should mention I also see warnings on successful builds

```
loadinternal: cannot find runtime/cgo
```

to build I run `nohup make package/feeds/packages/golang/{clean,compile} V=s &`
and for some reason it goes to toolchain libs, is it a package dependency?
```
make[2]: Entering directory '/media/SDStorage/builds/github-openwrt/openwrt/package/libs/toolchain'
Makefile:733: WARNING: skipping libgomp -- package has no install section
```

no idea if this is a problem or not
```
readelf: Error: Unable to find program interpreter name
readelf: Error: no .dynamic section in the dynamic segment
readelf: Error: Unable to find program interpreter name
readelf: Error: no .dynamic section in the dynamic segment
readelf: Warning: Unrecognized form: 0x23
readelf: Warning: Unrecognized form: 0x22
readelf: Warning: Unrecognized form: 0x22
readelf: Warning: Bogus end-of-siblings marker detected at offset 83 in .debug_info section
readelf: Warning: Bogus end-of-siblings marker detected at offset 84 in .debug_info section
readelf: Warning: Bogus end-of-siblings marker detected at offset 97 in .debug_info section
readelf: Warning: Further warnings about bogus end-of-sibling markers suppressed
```

mpratt14:
I dropped `GOHOSTOS` and it still works 👍🏼 

does it look ok now?

jefferyto:
I think this is an issue with Go and should be fixed upstream instead of just worked around here - I'm looking into it now.

> and for some reason it goes to toolchain libs, is it a package dependency?

Every package build involves checking/building the toolchain even when it isn't used.

The other error messages: I need more context (i.e. where they appear) to know what they mean. Would be good if you can post a complete build log.

mpratt14:
> I think this is an issue with Go

I agree



> should be fixed upstream

not sure...
I think this workaround is one of the possible appropriate fixes as it seems that it's not possible to update the bootstrap

from their docs
https://go.dev/doc/install/source

```
Bootstrap toolchain from C source code
To build a bootstrap toolchain from C source code, use either the git branch release-branch.go1.4 or
go1.4-bootstrap-20171003.tar.gz which contains the Go 1.4 source code plus accumulated fixes to keep the tools
running on newer operating systems. (Go 1.4 was the last distribution in which the toolchain was written in C.)
After unpacking the Go 1.4 source, cd to the src subdirectory, set CGO_ENABLED=0 in the environment,
and run make.bash (or, on Windows, make.bat).

Once the Go 1.4 source has been unpacked into your GOROOT_BOOTSTRAP directory, you must keep this
git clone instance checked out to branch release-branch.go1.4. Specifically, do not attempt to reuse this git clone
in the later step named "Fetch the repository." The go1.4 bootstrap toolchain must be able to properly traverse the
go1.4 sources that it assumes are present under this repository root.

Note that Go 1.4 does not run on all systems that later versions of Go do. In particular,
Go 1.4 does not support current versions of macOS.
On such systems, the bootstrap toolchain must be obtained using one of the other methods.
```

note the combination of these two issues:

 - **"Go 1.4 was the last distribution in which the toolchain was written in C"**
 - **"Go 1.4 does not run on all systems that later versions of Go do"**


the "arm" bootstrap cannot compile for "arm64", and I think that just makes sense because its not a cross-compiler...


however, they have a different script for cross-compile of the bootstrap

```
To cross-compile a bootstrap toolchain from source, which is necessary on systems Go 1.4 did not target
(for example, linux/ppc64le), install Go on a different system and run bootstrap.bash
```

basically the problem is, Go 1.4 does not support arm64, and thats all we have for bootstrap source
if you like the cross-compile idea better I can try that

there would be a new list of valid OS_ARCH for the bootstrap
and if it doesnt match, cross-compile it with the script, requiring that an external Go tree is specified
(which on my system is at /usr/lib/go/src/bootstrap.bash)
and otherwise the make job would error on purpose to warn the user the current config is not possible

mpratt14:
and to make things simple for the user, maybe auto detection of the local Go tree...

mpratt14:
I see that we have `BOOTSTRAP_GO_VALID_OS_ARCH`

we just need to use it better....

mpratt14:
after some testing with the package installation of Go, it seems like it has to be a git tree for some reason

```
/usr/share/go/src $ sudo GOOS=linux GOARCH=arm64 ./bootstrap.bash

#### Copying to ../../go-linux-arm64-bootstrap

#### Cleaning ../../go-linux-arm64-bootstrap

#### Building ../../go-linux-arm64-bootstrap

Building Go cmd/dist using /usr/lib/go-1.15. (go1.15.15 linux/arm)
go tool dist: FAILED: not a Git repo; must put a VERSION file in $GOROOT
```

mpratt14:
I discovered that cross-compiling had to be done simply in another copy of source, not that it had to be a live git tree

also I had to specify both `GOHOSTARCH` and `GOARCH`

I successfully got a cross-compiled bootstrap with the copy of source in openwrt

going to work on a patch to implement this with conditions....

```
openwrt/build_dir/hostpkg/go-1.18.2/src $ GOOS=linux GOARCH=arm64 GOHOSTARCH=arm ./bootstrap.bash
#### Copying to ../../go-linux-arm64-bootstrap

#### Cleaning ../../go-linux-arm64-bootstrap

#### Building ../../go-linux-arm64-bootstrap

Building Go cmd/dist using /usr/local/go. (go1.15.7 linux/arm)
Building Go toolchain1 using /usr/local/go.
Building Go bootstrap cmd/go (go_bootstrap) using Go toolchain1.
Building Go toolchain2 using go_bootstrap and Go toolchain1.
Building Go toolchain3 using go_bootstrap and Go toolchain2.
Building packages and commands for host, linux/arm.
Building packages and commands for target, linux/arm64.
----
Bootstrap toolchain for linux/arm64 installed in /home/pi/openwrt/openwrt/build_dir/hostpkg/go-linux-arm64-bootstrap.
Building tbz.
-rw-r--r-- 1 pi pi 135752384 May 24 02:14 /home/pi/openwrt/openwrt/build_dir/hostpkg/go-linux-arm64-bootstrap.tbz
```

jefferyto:
> if you like the cross-compile idea better I can try that

No... I mean the issue is Go detecting `GOHOSTARCH=arm64` when it should use `GOHOSTARCH=arm`. The bootstrap Go 1.4 compiler is not the issue.

> and to make things simple for the user, maybe auto detection of the local Go tree...

Not going to be done in this PR.

> I successfully got a cross-compiled bootstrap with the copy of source in openwrt
> 
> going to work on a patch to implement this with conditions....

Please don't - if you have Go to compile the bootstrap, you already have Go to compile Go. There is no need to compile the bootstrap separately.

I'm working on instructions on what I would like done. I need some time to compile this list.

mpratt14:
> if you have Go to compile the bootstrap, you already have Go to compile Go

my understanding from reading Go documentation was that this is only true if the bootstrap and target are the same arch...but I could be wrong

I thought I had already tried to set `GOHOSTARCH=arm GOARCH=arm64` on a normal build with the external Go but it didnt work, but I dont' remember, so I'll try it

mpratt14:
yes, you are right, the host compile was successful with a manual hack to both `GOARCH` and `GOHOSTARCH`

```
diff --git a/lang/golang/golang-compiler.mk b/lang/golang/golang-compiler.mk
index c6c836017..148136205 100644
--- a/lang/golang/golang-compiler.mk
+++ b/lang/golang/golang-compiler.mk
@@ -67,10 +67,10 @@ define GoCompiler/Default/Install/Bin

        $(INSTALL_DIR) "$(2)/lib/go-$(3)/bin"

-  ifeq ($(4),$(GO_HOST_OS_ARCH))
+  ifeq ($(4),linux_arm)
        $(INSTALL_BIN) -p "$(1)/bin"/* "$(2)/lib/go-$(3)/bin/"
   else
-       $(INSTALL_BIN) -p "$(1)/bin/$(4)"/* "$(2)/lib/go-$(3)/bin/"
+       $(INSTALL_BIN) -p "$(1)/bin/linux_arm64"/* "$(2)/lib/go-$(3)/bin/"
   endif

        $(INSTALL_DIR) "$(2)/lib/go-$(3)/pkg"
diff --git a/lang/golang/golang/Makefile b/lang/golang/golang/Makefile
index 30457910a..3f0efbdcd 100644
--- a/lang/golang/golang/Makefile
+++ b/lang/golang/golang/Makefile
@@ -221,6 +221,8 @@ define Host/Compile
                GOROOT_BOOTSTRAP="$(BOOTSTRAP_ROOT_DIR)" \
                $(if $(HOST_GO_ENABLE_PIE),GO_LDFLAGS="-buildmode pie") \
                $(HOST_GO_VARS) \
+               GOHOSTARCH="arm" \
+               GOARCH="arm64" \
        )
 endef
```

however now I have a different issue.... my system has a 32-bit shell and is refusing to execute the go binary built for arm64

```
Building target Go first stage
( cd "/media/SDStorage/builds/github-openwrt/openwrt/build_dir/target-mipsel_24kc_musl/go-1.18.2/src" ; GOROOT_FINAL="/usr/lib/go-1.18"  GOROOT_BOOTSTRAP="/media/SDStorage/builds/github-openwrt/openwrt/staging_dir/hostpkg/lib/go-cross" GO_GCC_HELPER_CC="gcc" GO_GCC_HELPER_CXX="g++" GOCACHE="/media/SDStorage/builds/github-openwrt/openwrt/tmp/go-build" GOENV=off GO_GCC_HELPER_PATH="$PATH" CC=gcc CXX=g++ PKG_CONFIG=pkg-config PATH="/media/SDStorage/builds/github-openwrt/openwrt/staging_dir/hostpkg/lib/go-cross/openwrt:$PATH"  bash make.bash --no-banner ; )
make.bash: line 187: /media/SDStorage/builds/github-openwrt/openwrt/staging_dir/hostpkg/lib/go-cross/bin/go: No such file or directory
Building Go cmd/dist using /media/SDStorage/builds/github-openwrt/openwrt/staging_dir/hostpkg/lib/go-cross. ()
make.bash: line 198: /media/SDStorage/builds/github-openwrt/openwrt/staging_dir/hostpkg/lib/go-cross/bin/go: No such file or directory
make[2]: *** [Makefile:374: /media/SDStorage/builds/github-openwrt/openwrt/build_dir/target-mipsel_24kc_musl/go-1.18.2/.built] Error 127
make[2]: Leaving directory '/media/SDStorage/builds/github-openwrt/openwrt/feeds/packages/lang/golang/golang'
time: package/feeds/packages/golang/compile#2.94#5.12#25.40
    ERROR: package/feeds/packages/golang failed to build.
make[1]: *** [package/Makefile:116: package/feeds/packages/golang/compile] Error 1
make[1]: Leaving directory '/media/SDStorage/builds/github-openwrt/openwrt'
make: *** [/media/SDStorage/builds/github-openwrt/openwrt/include/toplevel.mk:245: package/feeds/packages/golang/compile] Error 2
```

```
~/openwrt/openwrt $ stat /media/SDStorage/builds/github-openwrt/openwrt/staging_dir/hostpkg/lib/go-cross/bin/go
  File: /media/SDStorage/builds/github-openwrt/openwrt/staging_dir/hostpkg/lib/go-cross/bin/go
  Size: 14958648        Blocks: 29224      IO Block: 131072 regular file
Device: 27h/39d Inode: 130027026   Links: 1
Access: (0755/-rwxr-xr-x)  Uid: ( 1000/      pi)   Gid: ( 1000/      pi)
Access: 2022-05-24 04:43:09.784126179 -0400
Modify: 2022-05-24 04:43:15.274251260 -0400
Change: 2022-05-24 05:01:35.038975694 -0400
 Birth: -
~/openwrt/openwrt $ /media/SDStorage/builds/github-openwrt/openwrt/staging_dir/hostpkg/lib/go-cross/bin/go version
-bash: /media/SDStorage/builds/github-openwrt/openwrt/staging_dir/hostpkg/lib/go-cross/bin/go: No such file or directory
~/openwrt/openwrt $
```

so for my system, unless I change back to 32-bit kernel, I am actually forced to make Go build for 32-bit manually, since it uses the kernel to automatically detect `GOHOSTARCH`

is the original idea here ok? or would you prefer an option to just manually specify the host arch in `.config`?

jefferyto:
> so for my system, unless I change back to 32-bit kernel, I am actually forced to make Go build for 32-bit manually, since it uses the kernel to automatically detect `GOHOSTARCH`

In your case you want both `GOHOSTARCH` and `GOARCH` to be `arm`. You should only need to set `GOHOSTARCH=arm`, `GOARCH` should default to `GOHOSTARCH`.

> is the original idea here ok? or would you prefer an option to just manually specify the host arch in .config?

I would prefer to manually specify the host arch as a config option, as I believe it also applies when building/using the 1.4 bootstrap compiler (e.g. x86_64 kernel with i386 userspace).

I'm also wondering, since `GO_HOST_ARCH` is [derived from `HOST_ARCH`](https://github.com/openwrt/packages/blob/35961ec90d66f0615f541d5dcd45696efb9751ca/lang/golang/golang-values.mk#L135-L138) and is in turn [derived from `uname -m`](https://github.com/openwrt/openwrt/blob/90e4c8c6e6fe060d849a5b96bc7595345ce3d6ea/include/toplevel.mk#L25), perhaps `HOST_ARCH` should be fixed as well. I'm not sure if `HOST_ARCH` is suppose to mean "kernel arch only" or "kernel and userspace arch".

But to be honest, I'm not sure how supported your use case is. You may want to open an issue at https://github.com/openwrt/openwrt/issues. (Actually I might open an issue.) Do you have problems building any other packages or full images?

mpratt14:
@jefferyto after thinking about what you said and testing, I realized that I have the same problem with ruby as well (and ruby uses a `configure` script with GNU tools)

I also see now that `HOST_ARCH` and the other host arch variable for GNU build tools `GNU_HOST_NAME` are defined together in the main OpenWrt tree, so it is best to handle a custom case there

so the solution here would simply be to force the Go scripts to use the make variable instead of relying on its own host detection function

does the new patch look good?

jefferyto (approved):
LGTM thanks!

PR #18604, opened 2022-05-23
stubby: restart on trigger interfaces

notgood:
Force restart stubby if any of the trigger interfaces goes up or down.
Avoids DoT DNS lookup timeouts when default route changes, in case of multiple
upstream interfaces.

Run tested: latest 22.03, two upstream wan interfaces in stubby "trigger" config in /etc/config/stubby

### DIFF
--- a/net/stubby/files/stubby.init
+++ b/net/stubby/files/stubby.init
@@ -270,7 +270,7 @@ service_triggers()
 
     for trigger_item in $trigger
     do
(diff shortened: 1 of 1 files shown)


neheb:
ping @flyn-org

flyn-org:
@neheb, I don't recall having any experience with stubby.

neheb:
I confused with snort. whoops.

PR #18605, opened 2022-05-24
openconnect: three misc fixes

rsalvaterra:
Details in each commit message. I don't believe any of them is controversial. :)

Maintainer: @nmav

Run tested: mediatek/mt7622 (Belkin RT3200)

### DIFF
--- a/net/openconnect/Config.in
+++ b/net/openconnect/Config.in
@@ -12,6 +12,7 @@ config OPENCONNECT_GNUTLS
 
 config OPENCONNECT_OPENSSL
 	bool "OpenSSL"
+	select OPENSSL_WITH_DTLS
 
 endchoice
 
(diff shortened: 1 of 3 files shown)


rsalvaterra:
> Why? Is nslookup guarranteed to be present in every installation?

Yes. Both [`awk`](https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob;f=package/utils/busybox/Config-defaults.in;h=abe6d5431a70d4a61d03edcd30f9201be746b870;hb=HEAD#l1019) and [`nslookup`](https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob;f=package/utils/busybox/Config-defaults.in;h=abe6d5431a70d4a61d03edcd30f9201be746b870;hb=HEAD#l2499) are part of BusyBox and always enabled.

nmav (approved):

nmav:
Thank you!

PR #18620, opened 2022-05-27
udpspeeder: add key option

xfan1024:
udpspeeder support a simple xor encryption, add the option to its init script

```
common options, must be same on both sides:
    -k,--key              <string>        key for simple xor encryption. if not set, xor is disabled
```

### DIFF
--- a/net/udpspeeder/files/udpspeeder-init
+++ b/net/udpspeeder/files/udpspeeder-init
@@ -17,6 +17,7 @@ validate_udpspeeder_section() {
 		'timeout:integer:8'		\
 		'local:string'			\
 		'remote:string'			\
+		'key:string'			\
 		'report:integer:10'		\
 		'disable_obscure:bool:0'	\
 		'interval:integer:0'		\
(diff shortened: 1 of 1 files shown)


net/udpspeeder/files/udpspeeder-init
```diff
@@ -79,6 +80,11 @@ start_instance() {
 		procd_append_param command --disable-obscure
 	fi
 
+	if [ ! -z "${key}" ];
```
neheb, line 83: You can simplify with

```
[ -z "${key}"  || procd_append_param command --key "${key}"
```

xfan1024: Done. It is looks inconsistent with other condition branch in this file.

neheb:
```
: === Checking commit '12f93cb41c172698b54e7f3518726f4d669af8d2'
: Author name (xiaofan) need to be your real name 'firstname lastname'
: Commit subject line seems ok (udpspeeder: add key option)
: Signed-off-by match author
```

xfan1024:
Fixed

xfan1024:
> ```
> : === Checking commit '12f93cb41c172698b54e7f3518726f4d669af8d2'
> : Author name (xiaofan) need to be your real name 'firstname lastname'
> : Commit subject line seems ok (udpspeeder: add key option)
> : Signed-off-by match author
> ```

Fixed

neheb:
https://patch-diff.githubusercontent.com/raw/openwrt/packages/pull/18620.patch is a hint.

It's the committer name in git, changeable with git config user.name . Afterwards, git commit --reset-author --ammend

PR #18621, opened 2022-05-27
haproxy: Update haproxy.cfg mode health deprecated

howl:
mode health is deprecated and removed in actual haproxy versions.
Replace it with mode http and return status 200
Haven't investigate how to replicate returning "OK" in body, this works as
uncommenting option httpchk in the previous versions.
Without this change haproxy won't start because errors in config.

Signed-off-by: David Santamaría Rogado <howl.nsp@gmail.com>


GeorgeSapkin:
Looks like this is already the case in the current config.

PR #18624, opened 2022-05-28
i2pd: Update to 2.42.1

r4sas:
Maintainer: @yangfl
Compile tested: mipsel_24k, MT7621, 21.02.1
Run tested: mipsel_24k, MT7621, 21.02.1

Description:
* Update to 2.42.1
* Replace spaces with tabulation in init file


neheb:
@r4sas for the next bump, would be great to switch to cmake. currently I see occasional build failures when parallel building i2pd.

r4sas:
Hm, never seen problems when built locally in docker, but I'll try to check if it possible...

PR #18657, opened 2022-06-01
xtables-addons: update

PolynomialDivision:
Update and fix Makefile.

Maintainer: @jow-


BKPepe:
Same as pointed out in OpenWrt repository. "Fix Makefile" should be rather "Makefile polishing" or rather "Add PKG_LICENSE_FILES and improve SPDX"

PolynomialDivision:
@BKPepe Updatet.

PolynomialDivision:
updatet

PR #18659, opened 2022-06-01
open-iscsi: update to 2.1.7

lucize:

Maintainer: me 
Tested x86

Makefile would need a better approach


neheb:
```
 /bin/sh: 1: /usr/bin/sed: not found
```

that's uhhh weird.

lucize:
I know about it,they changed the options for path, but didn't had the time to finish, it's hard coded in the files, renaming to simply sed should work, but was lazy and left one file out

neheb:
> I know about it,they changed the options for path, but didn't had the time to finish, it's hard coded in the files, renaming to simply sed should work, but was lazy and left one file out

should be overriden to

```
$(STAGING_DIR_HOST)/bin/sed
```

PR #18662, opened 2022-06-01
pciutils: update to 3.8.0

lucize:

Maintainer: me
Tested x86


neheb:
Given that a separate pciids list is used, does it make sense to get rid of update-pciids?

BKPepe:
I wonder why do we need so many patches for pciutils and nobody bothered to take a look at them, upstream them and we are still rebasing them. I would just throw them away!

BKPepe:
ping @lucize

lucize:
hopefully this weekend

lucize:
removed pci ids related things

PR #18680, opened 2022-06-04
htop: update to 3.2.1

graysky2:
Build-tested: x86/64
Run-tested: bcm2711/RPi4B


jefferyto:
CI is failing for i386/x86_64 because:

1. htop selects lm-sensors by default for x86
2. lm-sensors depends on perl
3. CI tries to test all ipk files it finds
4. CI gets an error when removing perlbase-i18n during test

Unsure what exactly is causing the error; my guess would be how the CI test code assumes all ipks files are independent, ignoring the fact that one package can create multiple, dependent ipks.

@aparcar

PR #18709, opened 2022-06-07
tailscale: fix --version

paper42:
Maintainer: @ja-pa
Run tested: armv7l, Turris Omnia, OpenWrt 21.02

Description: --version reporting doesn't work well if it's not specified in go ldflags. Is it possible to test two binaries from two packages like this?

### DIFF
--- a/net/tailscale/test.sh
+++ b/net/tailscale/test.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+if command -v tailscale; then
+    tailscale version | grep "$2" || exit 1
+fi
+
+if command -v tailscaled; then
+    tailscaled -version | grep "$2"
+fi
--- a/net/tailscale/Makefile
+++ b/net/tailscale/Makefile
@@ -27,6 +27,8 @@ PKG_USE_MIPS16:=0
 GO_PKG:=\
 	tailscale.com/cmd/tailscale \
 	tailscale.com/cmd/tailscaled
+GO_PKG_LDFLAGS:=-X 'tailscale.com/version.Long=$(PKG_VERSION)-$(PKG_RELEASE) (OpenWrt)'
+GO_PKG_LDFLAGS_X:=tailscale.com/version.Short=$(PKG_VERSION)
 
 include $(INCLUDE_DIR)/package.mk
 include ../../lang/golang/golang-package.mk


paper42:
tests in CI fail with this error, but I don't think that's caused by this PR:
```
Installing tailscale (1.24.2-2) to root...
Collected errors:
 * pkg_hash_check_unresolved: cannot find dependency kernel (= 5.10.120-1-2e0e3355e21afcace96a06a6ad73321c) for kmod-tun
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for tailscale:
 * 	kernel (= 5.10.120-1-2e0e3355e21afcace96a06a6ad73321c)
 * opkg_install_cmd: Cannot install package tailscale.
Use package specific test.sh
/ci/net/tailscale/test.sh: line 2: tailscale: not found
/ci/net/tailscale/test.sh: line 3: tailscaled: not found
Test failed
```

net/tailscale/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+tailscale -version | grep "$2"
+tailscaled -version | grep "$2"
```
jefferyto, line 3: I suggest testing if `tailscale` / `tailscaled` exists before running it, since the CI will install each ipk separately. (Can also check if at least one is installed, if you want to get fancy.)

jefferyto:
> tests in CI fail with this error, but I don't think that's caused by this PR:

That error should not be present the next time the CI is run, the openwrt/rootfs Docker images were updated after the CI run.

net/tailscale/test.sh
jefferyto: Although, since tailscale depends on tailscaled, when the tailscale ipk is installed by opkg, tailscaled (the buildbot-built package, not the ipk built during the CI run) will also be installed but it will not have the correct version information, so the test will still fail (for this PR, should pass in the future).

jefferyto: The current tailscaled in the feed returned `1.24.2-dev` in the [test](https://github.com/openwrt/packages/runs/6881214373?check_suite_focus=true#step:14:41):

```
Testing package tailscale in version 1.24.2 from tailscale
Installing tailscale (1.24.2-2) to root...
2022/06/14 13:37:16 logtail started
2022/06/14 13:37:16 Program starting: v1.24.2-dev-t, Go 1.18.3: []string{"/usr/sbin/tailscaled", "--cleanup"}
2022/06/14 13:37:16 LogID: f17db0ab8cc78d82059348224cf738fedea0137f6e14d646bbc0e2418c509f20
2022/06/14 13:37:16 logpolicy: using UserCacheDir, "/root/.cache/Tailscale"
logpolicy.Read /root/.cache/Tailscale/tailscaled.log.conf: open /root/.cache/Tailscale/tailscaled.log.conf: no such file or directory
2022/06/14 13:37:16 dns: resolvedIsActuallyResolver error: resolv.conf doesn't point to systemd-resolved; points to [168.63.129.16]
2022/06/14 13:37:16 dns: [rc=resolved resolved=not-in-use ret=direct]
2022/06/14 13:37:16 dns: using *dns.directManager
2022/06/14 13:37:16 flushing log.
2022/06/14 13:37:16 logger closing down
Failed to connect to ubus
Installing kmod-tun (5.10.120-1) to root...
Downloading https://downloads.openwrt.org/snapshots/targets/x86/64/kmods/5.10.120-1-596cd7216a51a1d449b810129b5cce13/kmod-tun_5.10.120-1_x86_64.ipk
Installing tailscaled (1.24.2-1) to root...
Downloading https://downloads.openwrt.org/snapshots/packages/x86_64/packages/tailscaled_1.24.2-1_x86_64.ipk
Configuring kmod-tun.
Configuring tailscaled.
Configuring tailscale.
Use package specific test.sh
/usr/sbin/tailscale
1.24.2
/usr/sbin/tailscaled
1.24.2-dev
Test successful
Removing package tailscale from root...
```

net/tailscale/test.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+if command -v tailscale; then
+    tailscale version | grep "$2"
```
jefferyto, line 3: If this fails and the tailscaled test below succeeds, then the test script will exit with a success exit value. I suggest either exiting immediately if this test fails, or adding a line at the end of the script that evaluates the exit value of both tests.

paper42: fixed

paper42:
In the last push, I added "(OpenWrt)" to the end of the long version similar to how many other distributions do it. I had to use `GO_PKG_LDFLAGS` (not `GO_PKG_LDFLAGS_X`) because _X replaces every space with " -X ".

jefferyto:
> In the last push, I added "(OpenWrt)" to the end of the long version similar to how many other distributions do it. I had to use `GO_PKG_LDFLAGS` (not `GO_PKG_LDFLAGS_X`) because _X replaces every space with " -X ".

I think this is fine, though alternatively you could set the long version to `$(PKG_VERSION)-$(PKG_RELEASE)-openwrt`.

paper42:
> I think this is fine, though alternatively you could set the long version to `$(PKG_VERSION)-$(PKG_RELEASE)-openwrt`.

Some distributions use ` (distribution)` and some `-distribution`. In my opinion, ` (OpenWrt)` looks prettier in the web UI, but if you or the maintainer would prefer `-distribution`, I can change that.

paper42:
I split tests to a second commit, is this better?

BKPepe:
Yes, but it is still not good in my eyes. It would help to have in the git history, what whas wrong and why are you fixing it like this. Mostly in the first commit.

paper42:
> Yes, but it is still not good in my eyes. It would help to have in the git history, what whas wrong and why are you fixing it like this. Mostly in the first commit.

fixed

BKPepe (approved):

PR #18712, opened 2022-06-07
glib2: replace intl patch with upstream solution

neheb:

Maintainer: @tripolar 
Compile tested: mips24

@micmac1 does this work fine?


neheb:
@micmac1 so I just took a look at json-glib. I see that it's using intl APIs: bindtextdomain, bind_textdomain_codeset, and textdomain. Sounds like the meson.build file should be patched to add an intl dependency.

micmac1:
> @micmac1 does this work fine?

It'd break json-glib right now for BUILD_NLS=y. Weren't there other packages with the same issue? They'd break as well I reckon.

micmac1:
> @micmac1 so I just took a look at json-glib. I see that it's using intl APIs: bindtextdomain, bind_textdomain_codeset, and textdomain. Sounds like the meson.build file should be patched to add an intl dependency.

Maybe. I could raise an issue upstream. But I think that'd only make sense if the patch you want to include here gets applied upstream. I didn't see it anywhere.

micmac1:
Found my old comment, it's [here](https://github.com/openwrt/packages/pull/13038#issuecomment-671085706). I tried to build libsoup with BUILD_NLS=y and your new patch. It fails the same as in 2020.

`soup-init.c:(.text.startup+0x1a): undefined reference to `libintl_bindtextdomain'`

I'm guessing the other packages might also fail again, but I'm not gonna try, don't have the time.

It looks to me like the packages depending on glib2 expect glib2 to setup libintl/libiconv. I don't see any checks in the meson build file from json-glib nor libsoup, and they're both GNOME projects. And honestly why would they all duplicate these checks if glib2 provided a proper pkg-config file with the proper linker flag?

neheb:
@micmac1 I noticed json-glib actually uses bindtext through. Wouldn't it make more sense to patch it?

neheb:
@micmac1 turns out the issue was the TARGET_LDFLAG adding -lintl and messing everything up. In the pkgconfig file I now see:

```
Libs: -L${libdir} -lglib-2.0 -lintl
```

micmac1:
> @micmac1 I noticed json-glib actually uses bindtext through. Wouldn't it make more sense to patch it?

Hi Rosen,

I don't know. Not for me to say. That's for upstream to decide. I left a comment on the Gnome Gitlab [here](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2734#note_1474341).

Kind regards,
Seb

neheb:
Another nice thing about this patch is that it removes the dependency on libiconv-stub. I have it removed locally and it compiles fine whereas before it would error.

micmac1:
Glad you got it sorted. I hope I'll never have to thing about this again. :)

neheb:
:). The nice thing is that since this will go upstream soon, we can stop carrying this with version 2.74 or something.

eli-schwartz:
> When using autotools glib2 relies on glib-gettext.m4 to add the proper flags. The meson build obviously doesn't. In the glib2 source there's actually a comment regarding these differences:
> 
> _"FIXME: glib-gettext.m4 has much more checks to detect broken/uncompatible
> implementations. This could be extended if issues are found in some platforms." (meson.build)_
> 
> So in my opinion glib2 should have CONFIG_BUILD_NLS in PKG_CONFIG_DEPENDS _and_ add "-lintl" to it's pkg-config file when needed. This way one doesn't need to add workarounds to packages that depend on glib2.

Ultimately, Meson is learning more about how to find gettext in ways that handle what glib-gettext.m4 used to do, and thus tempts projects to stop rolling their own gettext checks (as is the story with what is happening for glib).

I too hope that people never have to think about intl themselves again, but if you ever find yourself needing to, it's probably a Meson bug in the intl detection so please do report it!

Other than extending the compiler search path to add the INTL_FULL prefix, this should never have to be handled by individual builders if Meson can handle it for you. :)

PR #18713, opened 2022-06-08
lighttpd: update to lighttpd 1.4.65 release hash

gstrauss:
Maintainer: @flyn-org

Description:
* update to lighttpd 1.4.65 release hash
* specify lua version w/ -Dlua_version=lua to avoid patching meson.build
* include brotli and zstd with mod_deflate

Signed-off-by: Glenn Strauss [gstrauss@gluelogic.com](mailto:gstrauss@gluelogic.com)


net/lighttpd/files/lighttpd.conf
```diff
@@ -24,9 +24,6 @@ server.feature-flags       += ("server.graceful-shutdown-timeout" => 5)
 #server.errorlog-use-syslog  = "enable"
 #server.network-backend      = "writev"
 
-### Use IPv6 if available
```
BKPepe, line 27: Btw, this was removed and I could not see why, which results that you are making more changes than saying in the commit itself.

flyn-org:
I do agree with @BKPepe, so I would prefer the change to "... Use IPv6 ..." be omitted, unless the new lighttpd version made the (commented, I think) block obsolete. This is just to keep things tight. Other than that, this looks good to me.

HSXX:
good

PR #18714, opened 2022-06-08
rsyslog: add uci support

gsanchietti:
Maintainer: none
Compile tested: x86, 22.03.0-rc3
Run tested: x86, 22.03.0-rc3, replaced logd with rsyslog

Description:

Support dynamic configuration with UCI.
Default rsyslog configuration is preserved and moved inside `/etc/config/rsyslog`.

Enable TCP receiver:
```
uci set rsyslog.syslog.tcp_input='1'
uci commit
/etc/init.d/rsyslog restart
```

Setup forwarding to another rsyslog:
```
uci set rsyslog.remote=forwarder
uci set rsyslog.remote.source=*.*
uci set rsyslog.remote.protocol=udp
uci set rsyslog.remote.port=10514
uci set rsyslog.remote.rfc=3164
uci set rsyslog.remote.target=192.168.122.2
uci commit
/etc/init.d/rsyslog restart
```

If `rfc` is set to `5424`, the template will be accordingly and TCP framing will be set to `octed-counted`.
Setup forwarding to promtail:
```
uci set rsyslog.promtail=forwarder
uci set rsyslog.promtail.source=*.*
uci set rsyslog.promtail.protocol=tcp
uci set rsyslog.promtail.port=10514
uci set rsyslog.promtail.rfc=5424
uci set rsyslog.promtail.target=192.168.122.2
uci commit
/etc/init.d/rsyslog restart
```


jefferyto:
> Maintainer: gsanchietti

If you want to become the maintainer, please set `PKG_MAINTAINER` in the makefile with your name and email. If not, you can just write "Maintainer: none".

jefferyto (commented):
Instead of keeping /etc/rsyslog.conf and overwriting it on every start (and wearing out flash storage), perhaps it would be better to write the generated config file to /var/etc (in RAM) and tell rsyslogd to use the generated config file.

gsanchietti:
> > Maintainer: gsanchietti
> 
> If you want to become the maintainer, please set `PKG_MAINTAINER` in the makefile with your name and email. If not, you can just write "Maintainer: none".

Reading the doc, this wasn't clear to me. Thanks for the suggestion. I've changed Maintainer to `none`

gsanchietti:
> Instead of keeping /etc/rsyslog.conf and overwriting it on every start (and wearing out flash storage), perhaps it would be better to write the generated config file to /var/etc (in RAM) and tell rsyslogd to use the generated config file.

This is indeed a good idea. I'm new to OpenWrt so I'd need a couple of suggestion.

1. If I move the config file inside `/var/etc`, should I leave also `/etc/rsyslog.conf`? What the original should contain? Just a placeholder with some comments?
2. How to handle upgrades for existing installations? Maybe I can assume that if `/etc/rsyslog.conf` is not a placeholder, just use it: otherwise use the config from UCI

What do you think?

jefferyto:
Thinking about this a little more, I think you can combine the uci config and /etc/rsyslog.conf by adding `$IncludeConfig /etc/rsyslog.conf` in the generated config file. This would allow "simple" configuration to be done with uci and "advanced" configuration to be added in /etc/rsyslog.conf, as well as allow upgrades of existing installations to continue working.

Installing a placeholder with comments (perhaps with commented-out example config) to /etc/rsyslog.conf would be good.

To make upgrading existing installations as seamless as possible, I suggest instead of installing a default uci config file, add a uci-defaults script that will add the default uci config if there is no uci config already and if /etc/rsyslog.conf does not contain any valid config (does not have anything other than comments or blank lines).

admin/rsyslog/files/rsyslog.conf
```diff
@@ -1,18 +1,18 @@
...
-
-mail.*                                    /var/log/maillog
-
-cron.*                                    /var/log/cron
-
-local7.*                                  /var/log/boot.log
+#
+# The preferred way to configure rsyslogd is now UCI.
+#
+# Thi file can be still used and will be included
```
jefferyto, line 4:
```suggestion
# This file can be still used and will be included
```

admin/rsyslog/files/rsyslog.init
```diff
@@ -5,8 +5,102 @@ START=20
...
+        modules="${modules} imudp"
+        config_get udp_port syslog udp_input_port
+        input_u="input(type=\"imudp\" port=\"${udp_port}\")"
+
+    fi
+    config_get template syslog default_template
+    config_foreach handle_selector selector
+    config_foreach handle_forwarder forwarder
+
+    > ${CONFIG_FILE}
```
jefferyto, line 78: Suggest to add `mkdir -p /var/etc` before creating the config file.

jefferyto (commented):
Suggest to also check the init script with shellcheck (with `--shell=sh` or `--shell=dash`) and run-test (if you haven't already).

admin/rsyslog/files/rsyslog.conf
gsanchietti: Fixed, thank you for pointing it out!

admin/rsyslog/files/20_rsyslog
```diff
@@ -0,0 +1,33 @@
+[ "$(sed -e '/^#/d' -e '/^\s*$/d' /etc/rsyslog.conf)" != "" ] && exit 0
```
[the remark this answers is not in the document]
gsanchietti: I will adjust the regexp to match also lines starting with whitespace.
Still, I'd prefer sed because IMO it's more readable.

admin/rsyslog/files/rsyslog.init
gsanchietti: Done.

gsanchietti:
> Suggest to also check the init script with shellcheck (with `--shell=sh` or `--shell=dash`) 

I've never found shellcheck really helpfull, still I followed your suggestion.
This is the commit: https://github.com/openwrt/packages/pull/18714/commits/68f780adeb5e42c9b70230330f029deeddd1f6ec
I still think, it doesn't worth the effort.

> and run-test (if you haven't already).

I would like to do it. Do you have any pointer on how to run the tests?

admin/rsyslog/files/20_rsyslog
```diff
@@ -0,0 +1,33 @@
+grep -qv -e '^\s*#' -e '^\s*$' /etc/rsyslog.conf 2>/dev/null && exit 0
```
feckert, line 1: Do we not needed a shebang here?
`#!/bin/sh`

feckert (commented):
Here are a few more comments from my side

admin/rsyslog/files/rsyslog.init
```diff
@@ -1,12 +1,115 @@
...
 
+UCI_CONF="rsyslog"
+CONFIG_FILE="/var/etc/rsyslog.conf"
+BASE_CONFIG_FILE="/etc/rsyslog.conf"
+
+modules=""
+selectors=""
+forwarders=""
+
+handle_selector() {
```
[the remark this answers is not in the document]
gsanchietti: You just broke my heart :D

admin/rsyslog/files/rsyslog.init
```diff
@@ -5,8 +5,100 @@ START=20
...
+    fi
+    config_get template syslog default_template
+    config_foreach handle_selector selector
+    config_foreach handle_forwarder forwarder
+
+    > ${CONFIG_FILE}
+    for m in ${modules}; do
+        echo "module(load=\"${m}\")" >> ${CONFIG_FILE}
+    done
+    echo ${input_t} >> ${CONFIG_FILE}
```
neheb: could be simplified with

```
{
    echo ${input_t}
    echo ${input_u}
    echo "\$ActionFileDefaultTemplate ${template}"
    echo -e ${selectors}
    echo -e ${forwarders}
} >> ${CONFIG_FILE}
```

admin/rsyslog/files/rsyslog.init
```diff
@@ -5,8 +5,100 @@ START=20
...
+    local rfc
+    local opts
+
+    config_get src ${config} source
+    config_get target ${config} target
+    config_get protocol ${config} protocol "udp"
+    config_get port ${config} port "514"
+    config_get rfc ${config} rfc "3164"
+
+    if [ "$rfc" == "5424" ]; then
```
neheb: == is a bashism. Please run through shellcheck.

admin/rsyslog/files/rsyslog.init
gsanchietti: I know but I'd prefer current implementation because IMO it's more readable.
Is this change really required to get the PR accepted?

admin/rsyslog/files/rsyslog.init
gsanchietti: I've already done that, but disabled the check because it works with `sh`.
Still, I've added the change as request: https://github.com/openwrt/packages/pull/18714/commits/3db02aa66210cf623bff61f845774594ab8b7a78

Also fixed the commit formalities: https://github.com/openwrt/packages/runs/7208594812?check_suite_focus=true#step:4:53

gsanchietti:
Thank you for accepting it, I'm very pleased to contribute to the project!

PR #18720, opened 2022-06-09
cshark: fix compilation with GCC12

neheb:

Maintainer: @lperkov 
Compile tested: mips24

### DIFF
--- a/net/cshark/patches/010-gcc12.patch
+++ b/net/cshark/patches/010-gcc12.patch
@@ -1,0 +1,11 @@
+--- a/src/uclient.c
++++ b/src/uclient.c
+@@ -150,7 +150,7 @@ static void cshark_ustream_ssl_init(void
+ 
+ 	ssl_ctx = ssl_ops->context_new(false);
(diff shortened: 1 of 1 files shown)


BKPepe:
My usual questions related to the patch, which you want to add.

Patch:
- Who is the author
- Where is the commit subject and commit description in the patch
- Was the patch upstreamed? If yes, it is unclear from the PR or the commit itself.

neheb:
I wrote it. It's a fairly simple patch. Upstream seems dead.

BKPepe:
:-( Sending it upstream and then backporting the patch has many advantages, like a commit description in the patch, etc. Even though it is dead, it forces you to do such things. Otherwise, we will need to carry this patch for ages, and we do not know what it is doing, who the author is, etc.

neheb:
Honestly this package probably doesn't even work. I agree generally speaking and I do upstream most of my patches. For some stuff I don't even want to touch it (*cough* CMake Iconv *cough*)

PR #18722, opened 2022-06-09
openvpn: Enable managment server support by default

MaxSKash:
It makes no sense to disable support for this parameter.
Compiled files are the same length.
In luci-app-openvpn, it is in the settings.

Signed-off-by: Max S Kash <asukms@ya.ru>


MaxSKash:
hm, openvpn-wolfssl does not support management?...

PR #18727, opened 2022-06-09
libsoup3: add

neheb:

Maintainer: none
Compile tested: mips24

@flyn-org

### DIFF
--- a/libs/libsoup3/Makefile
+++ b/libs/libsoup3/Makefile
@@ -1,0 +1,76 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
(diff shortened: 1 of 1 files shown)


flyn-org:
@neheb, please add me as the maintainer. The danger is that one library (libx) will pick up libsoup3, while another library (liby) uses libsoup. I think this will brek an application that uses both libx and liby.

I have been watching for libsoup3 support in the applications and libraries OpenWrt ships. I think grilo/grilo-plugins and libdmapsharing might be only to require libsoup 2. I have to port libdmapsharing, and I think the newest Grilo supports libsoup3.

neheb:
Added maintainer.

As far as I can tell, nothing uses libsoup3 yet, even with the buildbots.

PR #18730, opened 2022-06-09
clamav: use standard CMake find_package for iconv

neheb:
Fixes compilation without libiconv-stub.

Maintainer: @lucize 
Compile tested: mips24


neheb:
Shocking that this fails...

PR #18731, opened 2022-06-09
ksmbd-tools: only add -liconv with BUILD_NLS

neheb:
Fixes compilation without libiconv-stub.

Maintainer: @Andy2244 
Compile tested: mips24


neheb:
Added meson build instead. Should work better.

neheb:
this was merged upstream and can go away next version. Merging.

PR #18732, opened 2022-06-09
elektra: remove custom iconv handling

neheb:
CMake's find_package works fine. No need to override.

Maintainer: @haraldg 
Compile tested: mips24


neheb:
Actually fixed by porting hidapi's iconv patch.

PR #18733, opened 2022-06-09
irqbalance: fix iconv handling

neheb:
Fixes compilation when libiconv is missing.

Also remove pthread from glibc. No longer needed.

Maintainer: @hnyman 
Compile tested: mips24


hnyman:
Fails in CI ?

neheb:
Relies on https://github.com/openwrt/openwrt/pull/10024 getting merged.

neheb:
Decided to just write a meson build. Pretty simple.

hnyman:
Adding meson stuff seems strange to me.  

I have no experience with meson, and adding meson to a package that does not natively use it, sounds like a complication to me.

neheb:
I'm working on upstreaming it.

neheb:
this was upstreamed.

hnyman:
>> adding meson to a package that does not natively use it, sounds like a complication to me.

> this was upstreamed.

Upstreamed? 
To me it looks like upstream accepted your file into irqbalance's contrib/ with hesitation after a long discussion, and didn't find obvious benefits for irqbalance itself.  Meson is not yet used by upstream for anything by default. And your PR does not use that "upstream meson", instead it still creates a local file.

The PR commit description only talks about adding Meson, but does not mention iconv, that was the original aim of your PR.

Too bad that upstream is opposed to having "partially statically linked libraries", and would like to have either all libs static or all libs dynamically linked.   (I have orignally complained about the large glib2 in 2017 in https://github.com/Irqbalance/irqbalance/issues/40  )

Might be that meson is easier for maintaining the "OpenWrt hack" regarding the glib2 static linking, but I am still not quite comfortable with adding meson. I fear that meson adds additional workload in future if upstream changes their configure.ac and we should convert those to meson.

neheb:
That's true. However upstream's build is tiny and does not have many moving parts.

hnyman:
@neheb 
Is this now ready for merge?

neheb:
Yes

PR #18738, opened 2022-06-10
hidapi: update to 0.12.0

neheb:
Switch to cmake.

Maintainer: @paulfertser 
Compile tested: mips24

### DIFF
--- a/libs/hidapi/Makefile
+++ b/libs/hidapi/Makefile
@@ -8,29 +8,26 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=hidapi
-PKG_VERSION:=0.10.1
+PKG_VERSION:=0.12.0
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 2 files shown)


neheb:
and find_package(Iconv) is broken again. What a surprise...

neheb:
Added a basic meson build. Took care to match the cmake build.

neheb:
@dhewg does PKG_INSTALL do anything with meson packages?

neheb:
@paulfertser updated based on upstream feedback.

dhewg:
> @dhewg does PKG_INSTALL do anything with meson packages?

I don't think so. Looking at it again:
if PKG_INSTALL is set, Build/Install will default to Build/Install/Default.
cmake provides Build/Install/Default and sets PKG_INSTALL so that it gets used.
meson provides Build/Install/Meson and sets Build/Install to default to it, so there's no need for PKG_INSTALL.

It's inconsistent, may be that could be cleaned up. But I don't recall if there was a specific reason to not go the cmake route.

neheb:
Yeah I removed it: https://github.com/openwrt/packages/commit/df5d7245b9f888b21e2c9c02fd002c606e9c7c6e

PR #18746, opened 2022-06-12
rrdbackup: back up luci-statistics RRD data to flash during shutdown 

jtkohl:
Using @sqrwf 's repository as the source, but maybe he would be happy to put the rrdbackup init script directly into a package file and skip the git repo downloads in this Makefile?
(can't quite figure out how to tag him)

Definitely this is a draft form, glad to take comments/suggestions

I find it useful to have this as a full package, so that I can install it into my image-builder images and manage upgrade easier.
Still have to do this, I think, since it seems sysupgrade doesn't do an orderly reboot/shutdown:
```
/etc/init.d/luci_statistics stop
/etc/init.d/rrdbackup stop
```
<then run sysupgrade>

### DIFF
--- a/utils/rrdbackup/Makefile
+++ b/utils/rrdbackup/Makefile
@@ -1,0 +1,44 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=rrdbackup
+PKG_VERSION:=1
+PKG_RELEASE:=1
+
+PKG_LICENSE:=Unlicense
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/rrdbackup
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Enable the backup and restore of the RRD database for collectd during orderly reboots
+  URL:=https://github.com/sqrwf/openwrt-rrdbackup
+  PKGARCH:=all
+  MAINTAINER:=John Kohl <jtk.git@bostonpog.org>
+  EXTRA_DEPENDS:=luci-app-statistics
+endef
+
+define Package/rrdbackup/description
+ rrdbackup provides a startup/shutdown script that backs up the RRD collectd
+ database to flash and restores it to volatile storage during orderly reboots.
+endef
+
+define Package/rrdbackup/conffiles
+/etc/collectd/rrdbackup.tgz
+endef
+
+define Build/Compile
+endef
+
+define Build/Configure
+endef
+
+define Build/Install
+endef
+
+define Package/rrdbackup/install
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_BIN) files/rrdbackup $(1)/etc/init.d/rrdbackup
+endef
+
+$(eval $(call BuildPackage,rrdbackup))
--- a/utils/rrdbackup/LICENSE.md
+++ b/utils/rrdbackup/LICENSE.md
@@ -1,0 +1,24 @@
+This is free and unencumbered software released into the public domain.
+
+Anyone is free to copy, modify, publish, use, compile, sell, or
+distribute this software, either in source code form or as a compiled
+binary, for any purpose, commercial or non-commercial, and by any
+means.
+
+In jurisdictions that recognize copyright laws, the author or authors
+of this software dedicate any and all copyright interest in the
+software to the public domain. We make this dedication for the benefit
+of the public at large and to the detriment of our heirs and
+successors. We intend this dedication to be an overt act of
+relinquishment in perpetuity of all present and future rights to this
+software under copyright law.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+OTHER DEALINGS IN THE SOFTWARE.
+
+For more information, please refer to <https://unlicense.org>
--- a/utils/rrdbackup/files/rrdbackup
+++ b/utils/rrdbackup/files/rrdbackup
@@ -1,0 +1,46 @@
+#!/bin/sh /etc/rc.common
+
+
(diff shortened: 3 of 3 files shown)


sqrwf:
I feel flattered that you consider my rather trivial shell script worthy of making it into a package. However, and please take this in the least confrontional way possible, I don't appreciate naming me as the package maintainer without my consent. Feel free to fork or copy my code and run with it, but don't pin a role on me I didn't ask for.

jtkohl:
> I feel flattered that you consider my rather trivial shell script worthy of making it into a package. However, and please take this in the least confrontional way possible, I don't appreciate naming me as the package maintainer without my consent. Feel free to fork or copy my code and run with it, but please don't pin a role on me I didn't ask for.

No offense taken, I apologize for the goof.  (I'm still learning the practices and norms of the OpenWrt community.)

Mind if I ask for your guidance on package maintainers?  Who's eligible to be named--must it be someone "officially" part of the OpenWrt project, or can it be anybody with an email address?

And on this particular item, it would be simpler to just copy the one init script into the OpenWrt package, rather than clone the (tiny) git repository with the init script.  Since it's your script, may I ask you to put an appropriate license statement in the script or the repo, and then I could rework the packaging to include the script directly?

sqrwf:
> may I ask you to put an appropriate license statement in the script or the repo

Not a problem at all, it is now "licensed" as public domain. Have at it.

sqrwf:
FYI, I cleaned up the script (proper variables quoting, comments, no functional changes)

jtkohl:
> FYI, I cleaned up the script (proper variables quoting, comments, no functional changes), although I see you already cleaned up your copy for the most part

Thanks, I pulled in your update.

utils/rrdbackup/Makefile
```diff
@@ -0,0 +1,44 @@
...
+include $(INCLUDE_DIR)/package.mk
+
+define Package/rrdbackup
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Enable the backup and restore of the RRD database for collectd during orderly reboots
+  URL:=https://github.com/sqrwf/openwrt-rrdbackup
+  PKGARCH:=all
+  MAINTAINER:=John Kohl <jtk.git@bostonpog.org>
+  EXTRA_DEPENDS:=collectd-mod-rrdtool luci-app-statistics
```
sqrwf, line 18: The script depends on neither (those two packages are for visual output in LuCI), it actually doesn't depend on anything really. You can of course make the case that it doesn't make much sense without `collectd` itself.

jtkohl: so what are the "right" packages to depend upon?  I'd want it to depend on the packages that generate files in /tmp/rrd.  That's not strictly collectd itself, from what I have examined.

The writing to /tmp/rrd comes from the rrdtool plugin into collectd, and the configuration file generated to use that plugin.
I see that /etc/init.d/luci_statistics generates /var/etc/collectd.conf which refers to the rrdtool plugin and configures it to store files in /tmp/rrd, from where rrdbackup saves/restores the files.

So then really I guess the dependency is best expressed on /etc/init.d/luci_statistics (which generates the configuration for collectd to use /tmp/rrd), so I'd say it should depend on luci-app-statistics.
And then what controls whether luci_statistics generates an output to /tmp/rrd ?  I'd say that is when it's configured to use the rrd output plugin, which is I think collectd-mod-rrdtool...but  the .js file with the configuration panel for rrdtool is in luci-app-statistics, which itself depends on collectd-mod-rrdtool, so again this points to luci-app-statistics.

And, this script only works correctly if /tmp/rrd is configured as the pathname in /etc/config/luci_statistics...so maybe there's a change needed to get that name through something like `$(uci get luci_statistics.collectd_rrdtool.DataDir)`
?

and the depends can be trimmed to just luci-app-statistics

feckert:
Sorry if I'm interfere in your discussion.
But making a package and a service for this is not necessary from my point of view.

We can simply extend the collected so that when it is stopped cleanly it creates a backup and when it finds a backup it unpacks it again. See https://github.com/openwrt/packages/blob/master/utils/collectd/files/collectd.init

@hnyman any objection?

sqrwf:
> But making a package and a service for this is not necessary

I would tend to agree. It's a singular script after all. But it's not my place to object if somebody wants to make it a package.

> We can simply extend the collected

Sounds good to me, but then I would suggest an UCI option to turn it off/on (default tbd) and a sensibly chosen default location for the backup tgz (and a UCI option for that, too).

hnyman:
If collectd is extended, the feature should be optional and non-default.
Likely most users do not care about it.
Also the backup location might be configurable ?

Just for reference, personally I am using a manual backup/restore to `/etc/backup.stats/` defined as conffile location.
(And I also have a cron job to do the backup once per day to minimize data loss at a possible router crash.)

`/etc/storeStats.sh`
```
#!/bin/sh
/etc/init.d/collectd stop
logger -t "LuCI statistics" Create backup archive
mkdir -p /etc/backup.stats
cd /tmp/rrd/$(uname -n)
tar c -zvf /etc/backup.stats/stats.tar.gz *
cp /etc/backup.stats/stats.tar.gz /etc/backup.stats/stats-$(date +%Y%m%dT%H%M).tar.gz
sync
/etc/init.d/collectd start
```

`/etc/restoreStats.sh`
```
#!/bin/sh

/etc/init.d/collectd stop
logger -t "LuCI statistics" collectd stopped, stats being restored
mkdir -p /tmp/rrd/$(uname -n)
cd /tmp/rrd/$(uname -n)
tar x -zvf /etc/backup.stats/stats.tar.gz
sync
/etc/init.d/collectd start
```

jtkohl:
> Sorry if I'm interfere in your discussion. But making a package and a service for this is not necessary from my point of view.
> 
> We can simply extend the collected so that when it is stopped cleanly it creates a backup and when it finds a backup it unpacks it again. See https://github.com/openwrt/packages/blob/master/utils/collectd/files/collectd.init


I'm fine with doing this inside the collectd init file instead of as a new package.  My personal purpose here is to have something that makes upgrading easier and preserves the rrd history.  I use imagebuilder to create my images, so a package is nice so that during restart after upgrade, the history is restored before collectd/luci_statistics starts up.

jtkohl:
Hmm, so if we do it in `collectd.init` then to be self-contained it would need to parse /etc/config/collectd, and/or the alt_config_file that is configured in there, plus any Include directives, to find the rrdplugin configuration.
Or "cheat" and use uci to find the configured pathname from luci_statistics rrd output plugin config?

hnyman:
> Or "cheat" and use uci to find the configured pathname from luci_statistics rrd output plugin config?

LuCI statistics uses collectd, but not vice versa. Collectd package itself is not aware about LuCI statistics and can also be used independently.

If the backup is mainly targeting LuCI stats, it might be better to have the functionality there in the LuCI statistics app, which defines the config items like directories for LuCI stats (and should also define the items preserved in sysupgrade).

jtkohl:
Discussing this on the forums, I'll rework this into luci_statistics directly.
https://forum.openwrt.org/t/saving-dynamically-generated-files-during-sysupgrade/175227/14

PR #18774, opened 2022-06-18
ncdu: update to 1.17 and add blue option

graysky2:
Upstream bump and add a patch to use a blue highlight color for the dark color scheme (`--color dark`) as some users find the default bright green is too intense.  Note that invoking ncdu without the `--color` switch at use uses the upstream default bright green, so users will need to call ncdu with `--color dark` in order to get the blue color.

@neheb @clehner


graysky2:
Not sure why the build checks are failing.  It completed without errors on my system.

PR #18778, opened 2022-06-19
collectd: Fix various issues with lantiq dsl metrics.

bluebrother:
- Handle bool values correctly. We get the values as bool, but collectd
  requires a bool value to be numeric in the range [0, 1].
- Remove profile and mode values. Those are strings, which cannot be
  represented by collectd.
- Update collectd type for some values.
- Fix latn value being present twice.

Signed-off-by: Dominik Riebeling <bluebrother@posteo.de>

### DIFF
--- a/utils/collectd/files/lua-scripts/dsl.lua
+++ b/utils/collectd/files/lua-scripts/dsl.lua
@@ -44,7 +44,7 @@ local line_vars = {
 		type = "bitrate"
 	},
 	{
-		name = "latn",
+		name = "interleave_delay",
 		type = "latency"
 	}
 }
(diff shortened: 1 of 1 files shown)


utils/collectd/files/lua-scripts/dsl.lua
```diff
@@ -157,7 +156,7 @@ local function read()
 			get_values(hostname, errors, far_errors, "far")
 			get_values(hostname, line_vars, down_line, "down")
 			get_values(hostname, line_vars, up_line, "up")
-			get_values(hostname, erb_vars, erb, "")
+			-- get_values(hostname, erb_vars, erb, "")
```
feckert, line 159: Why did you comment that out? It's not in your description of the change?

feckert: This was added in [master](commit https://github.com/openwrt/openwrt/commit/70729d345414cdf1463dec042811813ff9a94e7a ) and [openwrt-22.03](https://github.com/openwrt/openwrt/commit/70729d345414cdf1463dec042811813ff9a94e7a).
The branch openwrt-21.02 does not have this commit https://github.com/openwrt/openwrt/commits/openwrt-21.02/package/network/config/ltq-vdsl-app/src/src/dsl_cpe_ubus.c
So we should not comment this out because you made a pullrequest against the master branch.

bluebrother: Thanks for the explanation. Updated the PR accordingly.

feckert:
LGTM thanks

PR #18787, opened 2022-06-21
dnsdist: switch from liblua to luajit

Habbie:
Maintainer: me
Compile tested: openwrt master, built on Debian 11; openwrt 19.07, built on Debian 11
Run tested: openwrt master on edgerouter lite; openwrt 19.07 on tp-link archer c7 v5 (after locally backporting this change to 19.07)

Description:
luajit provides higher performance for requests handled in Lua hooks.
It also enables access to dnsdist functionality only exposed via FFI,
and allows configurations/hooks to call functions in any C library
without providing separate bindings.

### DIFF
--- a/net/dnsdist/Makefile
+++ b/net/dnsdist/Makefile
@@ -115,7 +115,7 @@ define Package/dnsdist
 	  +libedit \
 	  +libstdcpp \
 	  +lmdb \
-	  +liblua \
+	  +luajit \
 	  +tinycdb
   URL:=https://dnsdist.org/
 endef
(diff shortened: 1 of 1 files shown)


Habbie:
The `powerpc_464fp` build failure is because luajit does not build for it. It's been broken since at least 22.03-rc4, probably longer.

neheb:
Yeah it's a GCC problem. It built with a previous version. I have no idea how to fix it.

neheb:
You can rebase to fix it.

Habbie:
> I fixed it here

Thank you!



> You can rebase to fix it.

That worked locally, and I force pushed this PR. Thanks again!

PR #18792, opened 2022-06-21
gcc: update 11 minor version 

Ansuel:
Maintainer: @noblepepper

Description:
- Update 11 minor version to match the upstream
- Rename patches dir to reflect new gcc 11 minor version
- Drop upstream patch


Ansuel:
@neheb Here

neheb:
@flyn-org fails under 22.04 with that bool error: https://downloads.openwrt.org/releases/faillogs-22.03/x86_64/packages/gcc/compile.txt

Should all of this work be backported?

flyn-org:
I think a backport would be the way ahead that would cost the least amount of work. It looks like the master and openwrt-22.03 openwrt/openwrt GCC definitions are similar, so I think the openwrt/package definitions could be made the same.

neheb:
Done.

PR #18793, opened 2022-06-21
luajit: backport softfloat ppc support

neheb:

Maintainer: @milani

### DIFF
--- a/lang/luajit/Makefile
+++ b/lang/luajit/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=luajit
 PKG_VERSION:=2.1.0-beta3
-PKG_RELEASE:=6
+PKG_RELEASE:=7
 
 PKG_SOURCE:=LuaJIT-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://luajit.org/download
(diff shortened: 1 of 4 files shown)


neheb:
@BKPepe merge or remove powerpc support? I know you guys use this on a powerpc platform.

lang/luajit/Makefile
```diff
@@ -24,7 +24,7 @@ define Package/luajit
  CATEGORY:=Languages
  TITLE:=LuaJIT
  URL:=https://www.luajit.org
- DEPENDS:=@(i386||x86_64||arm||armeb||aarch64||(powerpc&&HAS_FPU)||mips||mipsel||mips64)
+ DEPENDS:=@(i386||x86_64||arm||armeb||aarch64||powerpc||mips||mipsel||mips64)
```
neheb, line 27: given that this is basically everything except ARC, probably should just change this to @!arc

BKPepe:
We would like to use LuaJIT on PowerPC, so we can use Knot Resolver by default instead of unbound, but there were still issues that prevent us from using it how we want even though we tried as many forks as there were in the past (https://github.com/moonjit/moonjit/issues/101,  https://github.com/openresty/luajit2/issues/118) 

Sadly, this support will be removed and unsupported soon, but I don't think there is anything we can do about it. :cry: 

I can give it a last try to see what happens.

neheb:
I ran into the same compilation issues in that issue and fixed them here. I have not done any runtime testing on powerpc.

neheb:
guess I should also ping @Cynerd

BKPepe:
> I have not done any runtime testing on powerpc.

I did it right now. It seems like the Knot Resolver works with that on Turris 1.1, p8540 (Freescale P2020), OpenWrt 21.02.3. @paper42 Can you check it on your end as well? I still find it somehow weird.

BKPepe (approved):

BKPepe:
That was fast. You could wait a little bit until it is tested by my colleague.

PR #18805, opened 2022-06-23
cyrus-sasl: install pkgconfig file

paper42:
Maintainer: @flyn-org
Compile tested: aarch64, Turris MOX, OpenWrt 19.07
Run tested: -

Description: I needed the pkgconfig file for an unrelated build that won't go to openwrt, but I think it makes sense to upstream this change.


flyn-org:
Would you mind doing the same to the master branch? With that, this looks fine to me. Thank you!

flyn-org:
Great! Looks good. Recommend merge.

PR #18821, opened 2022-06-26
wifitoggle: set button to wlan, disable auto-disable timer by default

msdos03:
Maintainer: not quite sure
Run tested: OpenWRT snapshot

Description:
I think the default config of wifitoggle is unreasonable, so I made some changes to it. 
Wifitoggle turned wifi off several minutes after turning it on by default. It is confusing for users.

I have added signoff to my commit message.

Sorry, I don't know I have to use real name to sign it off, fixed.

### DIFF
--- a/utils/wifitoggle/files/wifitoggle.config
+++ b/utils/wifitoggle/files/wifitoggle.config
@@ -1,8 +1,8 @@
 config wifitoggle
-	option button	'wps'
+	option button	'wlan'
 
 	option persistent	'0' 
-	option timer		'600'
+	option timer		'0'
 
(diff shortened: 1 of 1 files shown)


LoakAudio:
Hi all, 3 years later but I agree with @msdos03
Putting the timer to 0 would be less confusing for users as, without a LuCi configuration page, users can miss this parameter and might struggle to find the reason for the recurring shutdown of their WiFi after triggering a WiFi OFF/WiFi ON just one time.

msdos03:
@LoakAudio Actually now I think this entire package should be removed. We have rfkill to do this job.

PR #18823, opened 2022-06-26
php8: update to 8.1.7

mhei:
Maintainer: me
Compile tested: mxs
Run tested: mxs

Description:

This fixes:
    - CVE-2022-31625
    - CVE-2022-31626

### DIFF
--- a/lang/php8/Makefile
+++ b/lang/php8/Makefile
@@ -6,7 +6,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=php
-PKG_VERSION:=8.1.6
+PKG_VERSION:=8.1.7
 PKG_RELEASE:=1
 
 PKG_MAINTAINER:=Michael Heimpold <mhei@heimpold.de>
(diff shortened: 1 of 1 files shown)


neheb:
Can you take a look at php8 in the context of https://github.com/openwrt/openwrt/pull/10024 ?

php8 seems to really want an external libiconv library.

mhei:
Yes and no. It really wants a library which provides the libiconv **API**. While the libiconv-stub used to provide that without actually doing the full magic, after applying your draft PR to openwrt php8's build fails as expected. The simplest solution would of course be to switch in php's makefile to libiconv-full. Is there any discussion ongoing about https://github.com/openwrt/openwrt/pull/10024 beside the Github PR? I'm just asking because I'm curious about removing this "feature". Well, I'm aware that newer systems provide more and more disk space and the need to use a faking libiconv-stub might be gone.

neheb:
musl supports iconv just fine. More limited charset but supports it.

libiconv-stub is a holdover from uClibc-ng days. There’s no need for it with musl. Unfortunately from what I’ve seen, php8 seems to want an external libiconv.

mhei:
Yes, you're right - PHP could be tricked with the stubs, but it fails with musl's implementation and thus needs libiconv-full. I'll prepare a PR for php8. But I would mark at as draft as long as https://github.com/openwrt/openwrt/pull/10024 is not yet merged.

neheb:
Right. There’s probably a way to edit the configure scripts and whatnot to work with libc iconv but I don’t quite know how to.

PR #18829, opened 2022-06-28
[19.07] openwisp-config: update to 1.0.1

nemesisdesign:
Maintainer: @nemesisdesign
Compile tested: ath79, ramips
Run tested: ath79, ramips

### DIFF
--- a/admin/openwisp-monitoring/Makefile
+++ b/admin/openwisp-monitoring/Makefile
@@ -1,0 +1,116 @@
+# openwisp.org
+#
+# This is free software, licensed under the GNU General Public License v3.
+# See /LICENSE for more information.
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=openwisp-monitoring
+PKG_RELEASE:=1
+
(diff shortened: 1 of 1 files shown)


feckert:
Could we please use a cherry-pick commit from master
So we have a reference in this commit where this is comming from?

The same should be done for openwrt-21.02 and openwrt-22.03

feckert:
@nemesisdesign This would be the correct command with an `-x`.
This adds a rev to your commit message. So every on know the parrent commit.
This is import if you work with different branches. And you want to know if a bug is fixed or a security issue is fixed in other branches too.
`git cherry-pick -x <commit-hash-from-master>`
see `man git-cherry-pick`

nemesisdesign:
> @nemesisdesign This would be the correct command with an `-x`. This adds a rev to your commit message. So every on know the parrent commit. This is import if you work with different branches. And you want to know if a bug is fixed or a security issue is fixed in other branches too. `git cherry-pick -x <commit-hash-from-master>` see `man git-cherry-pick`

Thanks @feckert! I updated the PR. If you can confirm that's good I will do the same on the other PRs for the other branches.

PR #18831, opened 2022-06-28
[22.03] acme: switch from iptables to nft (backport)

ngehrsitz:
Use nft instead of iptables to open port 80 in the firewall when getting a
cert. Since nft doesn't allow deleting a rule by its contents, capture and
save the handle when creating the rule, and use that to delete.

Signed-off-by: Toke Høiland-Jørgensen <toke@toke.dk>
(cherry picked from commit d926da5182a4547e5248dad3e3e59abcd41d126f)

Maintainer: @tohojo 
Compile tested: N/A
Run tested: Fritzbox 3370, 22.03.0-rc1

Backported to 22.03 since nft is the default there


tohojo:
Thanks!

PR #18834, opened 2022-06-29
prometheus-node-exporter-lua: Added selinux.lua.

jpds:
Maintainer: @champtar


champtar:
Hi @jpds,
Thanks for the contribution. As selinux is not compiled by default, can you move this plugin to a separate sub package ?
Also can you make sure you return the same values as the official prometheus node exporter ?

PR #18835, opened 2022-06-29
libv4l: Change dependency from libudev-zero to libudev

Vavooon:
Maintainer: Ted Hess <thess@kitschensync.net>
Compile tested: (mipsbe, 8dev Rambutan, OpenWrt v22.03.0-rc4)
Run tested: (mipsbe, 8dev Rambutan, OpenWrt v22.03.0-rc4)

Description:
`libudev-zero` dependency causes issues when other `libudev` provider is present, e.g. `eudev`. Especially, it was trying to select both `libudev-zero` and `libudev` what caused the conflict in my case.
Instead, when `libudev` is used, it just looks for any available package which provides the library.

### DIFF
--- a/libs/libv4l/Makefile
+++ b/libs/libv4l/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=v4l-utils
 PKG_VERSION:=1.20.0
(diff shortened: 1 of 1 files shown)


neheb:
I think PKG_RELEASE should be incremented.

Issue #18850, opened 2022-07-02
shadowsocks-libev: udp tproxy can't work


time-river:
## Description:

The udp relay can't work,  **`ss-redir` can't receive any udp packets.** But the tproxy rule works.

### Hardware:
NanoPi R5S

### Software:
openwrt 22.03 rc3
kernel - 5.10.116-1-46ad65b4062101b6af7b599cdea01149
overture 1.7-1
shadowsocks-libev 3.3.5-5

### Configuration
shadowsocks config:
```
config ss_local 'socks_tcp'
        option local_port '1080'
        option timeout '30'
        option fast_open '1'
        option no_delay '1'
        option reuse_port '1'
        option local_address '192.168.86.1'
        option mode 'tcp_only'
        option server 'tcp'

config ss_local 'socks_udp'
        option local_port '1080'
        option timeout '30'
        option fast_open '1'
        option no_delay '1'
        option reuse_port '1'
        option local_address '192.168.86.1'
        option mode 'udp_only'
        option server 'udp'
        option verbose '1'

config ss_redir 'redir_tcp'
        option local_port '1100'
        option timeout '60'
        option fast_open '1'
        option reuse_port '1'
        option no_delay '1'
        option local_address '::'
        option mode 'tcp_only'
        option server 'tcp'
        option disabled '0'

config ss_redir 'redir_udp_v6'
        option local_port '1100'
        option timeout '60'
        option fast_open '1'
        option reuse_port '1'
        option no_delay '1'
        option local_address '::'
        option mode 'udp_only'
        option verbose '1'
        option server 'udp'
        option disabled '0'

config ss_redir 'redir_udp_v4'
        option local_port '1100'
        option timeout '60'
        option fast_open '1'
        option reuse_port '1'
        option no_delay '1'
        option local_address '0.0.0.0'
        option mode 'udp_only'
        option verbose '1'
        option server 'udp'
```

Both ipv4 and ipv6 rules are configured, nftables shows:

```
table inet fw4 {
        chain ss_rules_pre_udp {
                type filter hook prerouting priority mangle; policy accept;
                meta l4proto udp goto ss_rules_pre_src_udp
        }

        chain ss_rules_pre_src_udp {
                ip daddr @ss_rules_dst_bypass_ accept
                ip6 daddr @ss_rules6_dst_bypass_ accept
                goto ss_rules_src_udp
        }

        chain ss_rules_src_udp {
                ip saddr @ss_rules_src_bypass accept
[...]

PR #18852, opened 2022-07-02
shadowsocks-libev: support ipv6 udp relay

time-river:
`ip` is equal to `ip -4`, which cause only v4 rules can be set.
Here using `ip -4` to replace `ip` apparently, using `ip -6` to
set v6 rules.

Background: #18850 

@yousong 

Signed-off-by: fu.lin <river@vvl.me>


yousong:
Thank you.  The patch looks fine with me.  Please amend the commit message suggested by ci workflow "Test Formalities"

yousong:
@time-river friendly ping ;)

yousong:
Replaced by https://github.com/openwrt/packages/pull/19142 .   Thank you for the heads-up

Issue #9682, opened 2019-08-09
[Meta] Prefered look of Makefiles


ja-pa:
@BKPepe @val-kulkov @neheb 
Hi,
I would like to start a discussion about the preferred look of Makefiles. 

This is an extension of discussion from PR https://github.com/openwrt/packages/pull/9667#issuecomment-518674778

My idea is, if we could define some guidelines how should Makefile look (variable orders, tabs usage etc.) in a more explicit way. I believe that it can have advantages of in case of change review and possibly in future, this could be used as a specification for a tool which could check that automatically.

ja-pa:
I personally prefer template described here by @BKPepe  https://github.com/openwrt/packages/pull/9399#issuecomment-508727872

BKPepe:
I agree we should have a template to which all of we will stick, otherwise, somebody will do it how he/she likes it and we can start with that one, which was mention in that PR. Thanks to @neheb in most Makefiles, there is order in them.

However, it's not updated in OpenWrt documentation either in [mwarning's openwrt-examples](https://github.com/mwarning/openwrt-examples). We should update both of them.

Another point of this discussion should be about PKG_SOURCE_URL. Do we prefer there .git or not? Talking about this https://github.com/openwrt/packages/pull/9626#discussion_r310020579 . Maybe this point should be discussed with OpenWrt main repository. From my point of view, there should be .git.

I think we are catching mixed indentation and would be good if somewhere can be mention this quote

> Jul 28 21:57:57 <KanjiMonster> ynezz: https://www.gnu.org/software/make/manual/html_node/Recipe-Syntax.html - line starts with tab => shell command; spaces => make syntax

On the other hand, we can do it in a community maintained repository how we want to do it, but it should be discussed and updated in OpenWrt main repository as well.

neheb:
The lack of .git is a GitHub feature. Other git repositories require it.

Maybe download.mk can be modified to pattern match on the URL. If there’s.git at the end, use git by default. Otherwise use curl/wget

karlp:
I feel this is just an invitation to keep churning things that aren't really relevant 99% of the time.

ja-pa:
I think it would be beneficial even just as a recommendation. Right now these pieces of information are written in different issues and MR (like https://github.com/openwrt/packages/pull/8738#issuecomment-485930530  ...) and some maintainers insist on them.

val-kulkov:
[...]

PR #18853, opened 2022-07-03
openwisp: cleanup Makefiles

PolynomialDivision:
Not sure if you want gpl-3.0-only or gpl-3.0-or-later for the makefile?
Maintainer: @nemesisdesign

### DIFF
--- a/admin/openwisp-config/Makefile
+++ b/admin/openwisp-config/Makefile
@@ -1,14 +1,14 @@
-# openwisp.org
+# SPDX-Identifier-License: GPL-3.0-only
 #
-# This is free software, licensed under the GNU General Public License v3.
-# See /LICENSE for more information.
+# openwisp.org
+
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openwisp-config
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_MAINTAINER:=Federico Capoano <f.capoano@openwisp.io>
-PKG_LICENSE:=GPL3.0-or-later
+PKG_LICENSE:=GPL-3.0-or-later
 
 PKG_SOURCE_URL:=https://github.com/openwisp/openwisp-config.git
 PKG_MIRROR_HASH:=7daa10a9d170e665f33a5555a246b4da2223c2d8d0e8a047edb01701c8886986
--- a/admin/openwisp-monitoring/Makefile
+++ b/admin/openwisp-monitoring/Makefile
@@ -1,14 +1,14 @@
-# openwisp.org
+# SPDX-Identifier-License: GPL-3.0-only
 #
-# This is free software, licensed under the GNU General Public License v3.
-# See /LICENSE for more information.
+# openwisp.org
+
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openwisp-monitoring
-PKG_RELEASE:=1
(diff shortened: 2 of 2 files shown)


admin/openwisp-config/Makefile
```diff
@@ -1,14 +1,14 @@
-# openwisp.org
+# SPDX-Identifier-License: GPL-3.0-only
 #
-# This is free software, licensed under the GNU General Public License v3.
-# See /LICENSE for more information.
+# openwisp.org
```
BKPepe, line 3: IMHO this should be done treewide not only for these two packages. Both approaches are OK for me.

admin/openwisp-config/Makefile
```diff
@@ -1,14 +1,14 @@
...
+
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openwisp-config
-PKG_RELEASE:=1
+PKG_RELEASE:=2
```
BKPepe, line 8: No need to bump this. You just touch SPDX License Identifier and for that you don't want to really wear the flash memory on the routers.

admin/openwisp-config/Makefile
```diff
@@ -1,14 +1,14 @@
...
 
 PKG_MAINTAINER:=Federico Capoano <f.capoano@openwisp.io>
-PKG_LICENSE:=GPL3.0-or-later
```
BKPepe, line 11: While at it, add license file.

admin/openwisp-config/Makefile
```diff
@@ -1,14 +1,14 @@
...
+PKG_LICENSE:=GPL-3.0-or-later
 
 PKG_SOURCE_URL:=https://github.com/openwisp/openwisp-config.git
```
BKPepe, line 13: If you want to clean it up properly, please use tagged releases.

admin/openwisp-monitoring/Makefile
```diff
@@ -1,14 +1,14 @@
-# openwisp.org
+# SPDX-Identifier-License: GPL-3.0-only
```
BKPepe, line 1: Same as above.

BKPepe (changes requested):
This is a just a partial change to clean it up. We should clean it once for all.

admin/openwisp-config/Makefile
PolynomialDivision: Can you be a bit more specific? Not sure about which change you are talking.

BKPepe:
Nope. I'd say that the @PolynomialDivision should do such changes treewide or not do it one by one, because there exists such issue how it should be preferable to look like https://github.com/openwrt/packages/issues/9682 , but most Makefiles in this repository were polished and somehow standardized by @neheb.

admin/openwisp-config/Makefile
BKPepe: Your's change related to SPDX License Identifier. Now, you are making it more confusing, at least from my side. Because some packages have variant A, some of them have your variant. Thus it should be done for all packages at the same time.

nemesisdesign (commented):
Anything I can do let me know. The changes look good to me.
Thanks @PolynomialDivision @BKPepe!

admin/openwisp-config/Makefile
PolynomialDivision: Yeah.

PR #18866, opened 2022-07-04
[22.03] collectd: Fix various issues with lantiq dsl metrics.

bluebrother:
- Handle bool values correctly. We get the values as bool, but collectd
  requires a bool value to be numeric in the range [0, 1].
- Remove profile and mode values. Those are strings, which cannot be
  represented by collectd.
- Update collectd type for some values.
- Fix latn value being present twice.

Cherry-picks eb632da5bbd732bbffc2761fbb191267ef2375d0

Signed-off-by: Dominik Riebeling <bluebrother@posteo.de>

### DIFF
--- a/utils/collectd/files/lua-scripts/dsl.lua
+++ b/utils/collectd/files/lua-scripts/dsl.lua
@@ -44,7 +44,7 @@ local line_vars = {
 		type = "bitrate"
 	},
 	{
-		name = "latn",
+		name = "interleave_delay",
 		type = "latency"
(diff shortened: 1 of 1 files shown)


feckert:
In the commit message, the reference to the commit in the master branch is missing.
Can you redo this with the command `git cherry-pick -x <hash>`?
So we know that this commit is also applied to the openwrt-22.03 branch

feckert:
Thanks merged!

PR #18868, opened 2022-07-05
  gnutls: adjust to new configure.ac syntax

PolynomialDivision:
Configure.ac syntax changed to:
  Old: --without-libbrotli --without-libzstd (also --with-*)
  New: --without-brotli --without-zstd (also --with-*)
  https://github.com/gnutls/gnutls/commit/6b794e49d1a14e43f9e08023f958364712c3c89a

Fixes: 6385813ddfb4 ("gnutls: update to 3.7.5")

Maintainer: @nmav

### DIFF
--- a/libs/gnutls/Makefile
+++ b/libs/gnutls/Makefile
@@ -135,8 +135,8 @@ CONFIGURE_ARGS+= \
 	--with-librt-prefix="$(LIBRT_ROOT_DIR)/" \
 	--with-pic \
 	--with-system-priority-file="" \
-	--without-libzstd \
-	--without-zlib
+	--without-zlib \
+	--without-zstd
 
 ifneq ($(CONFIG_GNUTLS_EXT_LIBTASN1),y)
 CONFIGURE_ARGS += --with-included-libtasn1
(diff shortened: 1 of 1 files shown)


nmav:
@PolynomialDivision shouldn't we add `--without-brotli` as well?

PolynomialDivision:
> @PolynomialDivision shouldn't we add `--without-brotli` as well?

It was not used before, so I did not add it. But maybe it will prevent further errors.

BKPepe:
Would you please change the commit subject to be more clear instead of cryptic "fix compiling"?

PolynomialDivision:
> Would you please change the commit subject to be more clear instead of cryptic "fix compiling"?

Done

BKPepe:
To be honest, I am not sure if I like it this way, because still the subject requires you to check commit description even though subject should be good on its first sight.

It's better than it was, but you are fixing primary this error, which prevents full builds:
```
Package libgnutls is missing dependencies for the following libraries:
libzstd.so.1
``` 

You should say something, for example, fix disable zstd. And in another commit, you can disable other  options, but it is just my opinion.

nmav (approved):

PR #18875, opened 2022-07-06
treewide: replace GPL-2.0 boilertext with SPDX

PolynomialDivision:
Replaced with:
---
grep -rnw $1 -e 'General Public License v2.' | awk -F: {' print $1 '}

for file in $gplv2; do
	sed -i '1s/^/# SPDX-Identifier-License: GPL-2.0-only\n/' $file
	sed -i '/GNU General Public License v2/d' $file
	sed -i '/LICENSE for more information/d' $file
	echo $file
done
---

@neheb @BKPepe


neheb:
Fine by me.

BKPepe:
Before merging, I have a few questions. 

1. This does not go through in the master branch: https://github.com/openwrt/openwrt/pull/4661 and there is no summary of your discussion with @aparcar on this matter - how it went, what is necessary to do.

2. From the pull commit and the commit description, it is unclear why we need to have such a change. 

Please don't take me wrong, but even the OpenWrt master repository is not united on this matter. Even formal things should comply with the [contributing policy](https://openwrt.org/submitting-patches#submission_guidelines).

PolynomialDivision:
@BKPepe Updated. Thanks for feedback.

BKPepe:
The commit description and the pull request description also seem better than they were. Thank you! Now, what remains to be solved is the pull request for the main repository.

jefferyto:
I'm not against this change but just want to note that the packages feed is userland and not the kernel; I think we're allowed to diverge from kernel style when appropriate.

BKPepe:
One year has passed, so is there anything new regarding this? What is the progress in the main repo? There was this PR https://github.com/openwrt/openwrt/pull/4661, but nothing new.

GeorgeSapkin:
No updates in over three year. Multiple merge conflicts. Closing as stale. Feel free to reopen if anything changes.

PR #18880, opened 2022-07-07
netsniff-ng: Update to version 0.6.8

xdarklight:
Maintainer: @utoni 
Compile tested: lantiq xRX200
Run tested: n/a

Description:
The update fixes GCC-10 (or newer) builds. Remove 010-uclibc.patch as it has been applied upstream in this new version.

note: for me the package doesn't build with this (and it didn't before) because `libcli` is missing so mausezahn (which is what I am actually interested in) won't build.
I can create a separate PR for adding a `libcli` package to fix the mausezahn build (which is part of netsniff-ng).


xdarklight:
That `libcli` issue is probably a TODO that I need to solve: it's coming up because `netsniff-ng` detects that I have `libcli` availabe -> then the build fails.
If I don't provide `libcli` the build works fine.
So please ignore that comment for now.

PR #18912, opened 2022-07-11
zerotier: implement netifd protocol support

Vavooon:
Signed-off-by: Volodymyr Nazarchuk <vavooon@gmail.com>

Maintainer: @mwarning 
Compile tested: (mipsbe, 8dev Rambutan, OpenWrt v22.03.0-rc4)
Run tested: (mipsbe, 8dev Rambutan, OpenWrt v22.03.0-rc4)

Description: Currently Zerotier package uses own config file located under `/etc/config/zerotier`. Instead, this PR adds native `zerotier` protocol support which is based on existing service script, so it can be properly handled by `netifd`. 

Few new features compared to `zerotier` service are:
- `zerotier-one` daemon is executed for each network separately with random port (`-p0` argument)
- It supports consistent network interface naming, so you can have interface names like `zt0`

I am not sure whether it is a good idea to remove existing `zerotier` service so I'm keeping it for now for backward compatibility, but let me know if it's better to remove it.

Here's example interface definition inside `/etc/config/network`:
```
config interface 'zt0'
        option disabled '0'
        option proto 'zerotier'
        option network_id '**************'
        option config_path '/root/.zt-**************'
        option copy_config_path '1
```

### DIFF
--- a/net/zerotier/files/lib/netifd/proto/zerotier.sh
+++ b/net/zerotier/files/lib/netifd/proto/zerotier.sh
@@ -1,0 +1,210 @@
+#!/bin/sh
+
+DAEMON=/usr/bin/zerotier-one
+CONFIG_PATH=/var/lib/zerotier-one
+
+[ -n "$INCLUDE_ONLY" ] || {
+	. /lib/functions.sh
+	. ../netifd-proto.sh
+	init_proto "$@"
+}
+
+proto_zerotier_init_config() {
+	available=1
+	no_device=1
+	proto_config_add_string "network_id"
+	proto_config_add_string "secret"
+	proto_config_add_string "port"
+	proto_config_add_string "config_path"
+	proto_config_add_boolean "copy_config_path"
+}
+
+proto_zerotier_setup() {
+	local interface="$1"
+
+	local network_id secret port config_path copy_config_path
+	local args=""
+
+	config_load network
+	config_get network_id "${interface}" "network_id"
+	config_get secret "${interface}" "secret"
+	config_get port "${interface}" "port"
+	config_get config_path "${interface}" "config_path"
+	config_get_bool copy_config_path "${interface}" "copy_config_path" 0
+
+	path=${CONFIG_PATH}_$interface
+
+	# Remove existing link or folder
+	rm -rf "${path}"
+
+	# Create link or copy files from CONFIG_PATH to config_path
+	if [ -n "$config_path" -a "$config_path" != "$path" ]; then
+		if [ ! -d "$config_path" ]; then
+			echo "ZeroTier config_path does not exist: $config_path" 1>&2
+			return
+		fi
+
+		# ensure that the target exists
+		mkdir -p $(dirname "${path}")
+
+		if [ "$copy_config_path" = "1" ]; then
+			cp -r "$config_path" "${path}"
+		else
+			ln -s "$config_path" "${path}"
+		fi
+	fi
+
+	mkdir -p "${path}"/networks.d
+	touch "${path}"/networks.d/"${network_id}".conf
+
+	# enforce service to use given interface name
+	touch "${path}"/devicemap
+	echo "${network_id}=${interface}" >> "${path}"/devicemap
+
+	if [ -n "$port" ]; then
+		args="$args -p${port}"
+	else
+		args="$args -p0"
+	fi
+
+	local secret
+	secret="$(uci get network."${interface}".secret)"
+
+	if [ -z "$secret" ]; then
+		echo "Generate secret - please wait..."
+		local sf="/tmp/zt.${interface}.secret"
+
+		zerotier-idtool generate "$sf" > /dev/null
+		[ $? -ne 0 ] && return 1
+
+		secret="$(cat $sf)"
+		rm "$sf"
+
+		uci set network."${interface}".secret="$secret"
+		uci commit network
+	fi
+
+	if [ -n "$secret" ]; then
+		echo "$secret" > "${path}"/identity.secret
+		# make sure there is not previous identity.public
+		rm -f "${path}"/identity.public
+	fi
+
+	proto_run_command "${interface}" "${DAEMON}" -p0 "${path}"
+
+	local status
+	while true; do
+		status="$(zerotier-cli -D${path} get ${network_id} status 2>/dev/null)"
+		[ "$status" != "REQUESTING_CONFIGURATION" -a -n "$status" ] && break
+		sleep 1
+	done
+
+	if [ "$status" != "OK" ]; then
+		echo "Unable to connect"
+		proto_zerotier_teardown
+		proto_notify_error "$interface" "$status"
+		return 1
+	fi
+
+
+	IP4ADDRS=
+	IP6ADDRS=
+
+	local addresses="$(ip -json address list dev "$interface")"
+	json_init
+	json_load "{\"addresses\":${addresses}}"
+
+	if json_is_a addresses array; then
+		json_select addresses
+		json_select 1
+
+		if json_is_a addr_info array; then
+			json_select addr_info
+
+			local i=1
+			while json_is_a ${i} object; do
+				json_select ${i}
+				json_get_vars scope family local prefixlen broadcast
+
+				if [ "${scope}" == "global" ]; then
+					case "${family}" in
+						inet)
+							append IP4ADDRS "$local/$prefixlen/$broadcast/"
+							;;
+
+						inet6)
+							append IP6ADDRS "$local/$prefixlen/$broadcast///"
+							;;
+					esac
+				fi
+
+				json_select ..
+				i=$(( i + 1 ))
+			done
+		fi
+	fi
+
+
+	IP4ROUTES=
+	IP6ROUTES=
+
+	local routes="$(ip -json route list dev "$interface")"
+	json_init
+	json_load "{\"routes\":${routes}}"
+
+	if json_is_a routes array;then
+		json_select routes
+
+		local i=1
+		while json_is_a ${i} object; do
+			json_select ${i}
+			json_get_vars dst gateway metric prefsrc
+
+			case "${dst}" in
+				*:*/*)
+					append IP6ROUTES "$dst/$gateway/$metric///$prefsrc"
+					;;
+				*.*/*)
+					append IP4ROUTES "$dst/$gateway/$metric///$prefsrc"
+					;;
+				*:*)
+					append IP6ROUTES "$dst/128/$gateway/$metric///$prefsrc"
+					;;
+				*.*)
+					append IP4ROUTES "$dst/32/$gateway/$metric///$prefsrc"
+					;;
+			esac
+
+			json_select ..
+			i=$(( i + 1 ))
+		done
+	fi
+
+	proto_init_update "${interface}" 1
+
+	PROTO_IPADDR="${IP4ADDRS}"
+	PROTO_IP6ADDR="${IP6ADDRS}"
+
+	PROTO_ROUTE="${IP4ROUTES}"
+	PROTO_ROUTE6="${IP6ROUTES}"
+
+	echo "Zerotier is up"
+
+	proto_send_update "${interface}"
+}
+
+proto_zerotier_teardown() {
+	local interface="$1"
+
+	proto_kill_command "${interface}"
+
+	# Wait for daemon to stop before removing config path otherwise there might be 'file not found' errors
+	sleep 2
+
+	# Remove existing link or folder
+	rm -rf ${CONFIG_PATH}_"${interface}"
+}
+
+[ -n "$INCLUDE_ONLY" ] || {
+	add_protocol zerotier
+}
--- a/net/zerotier/Makefile
+++ b/net/zerotier/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=zerotier
 PKG_VERSION:=1.16.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/zerotier/ZeroTierOne/tar.gz/$(PKG_VERSION)?


mwarning:
hi @Vavooon , thank you for your contribution! I will take a look at your code. Since I am rather new to netifd, maybe someone else can take a look as well.

mwarning:
@Vavooon hi, can you tell what advantages netifd support has for the zerotier package? And how does it work along the current init script?

Vavooon:
@mwarning 
The main point for me is unification, so Zerotier interfaces will be handled by `netifd` in the same way as all other interface types (e.g. `openvpn`, `wireguard` and other VPNs have their own `netifd` protocol extensions). It should improve netifd awareness about Zerotier interfaces. There should probably be other hidden advantages, e.g. it should be easier to implement Zerotier support in LUCI in the future.

Regarding existing Zerotier service, it is still present in the package and it wouldn't interfere with new logic since it's allowed to run multiple Zerotier instances at once on different ports. 
Although I think it's better to remove the service and only use new logic, we might keep it in order to provide some backward compatibility. On the other hand, users can be confused by having 2 different ways to add a network, so the decision is up to you.

stangri:
@Vavooon this is amazing, I've struggled with creating a proper protocol for nebula package, do you mind if I use yours as a template?

@mwarning one of the biggest reasons I wanted to implement nebula as a protocol, it's because you can then use the appropriate interface name in the `/etc/config/dropbear` to let dropbear listen for incoming SSH connections on that interface. This requires an interface with netifd support. As far as I understand how Zerotier works, this might also be a very desired feature, on top of the reasons listed by @Vavooon.

Vavooon:
@stangri NP, feel free to use it.

oskarirauta:
@Vavooon 

[luci-proto-zerotier](https://github.com/oskarirauta/packages2/tree/main/luci/luci-proto-zerotier)

Vavooon:
@oskarirauta Thanks a lot!
Have you tested it in pair with my PR? I don't have a change to try it by myself

oskarirauta:
@Vavoon

Yes, I have. Using it as we speak. It's not very useful though, as it's unaware of zerotier's connection state, but atleast protocol is known by LuCi.

### Some comments about protocol

Ok, script needs some fine tuning- for example, this patch doesn't make it executable, which helps at start..
Second, config path is removed _too soon_, zerotier has _not yet exited_ and this floods errors to syslog..
So, add sleep 1 before removing files..

before issuing proto_run_command when initializing connection you also should add sleep 1, because previous ifdown might still be processing, so those files may not yet have been deleted-

### Further development

Also, it's a shame, that protocol cannot fill in details about connection for netifd, I've tried it all, hotplug doesn't trigger with zerotier and there's no good way to do this, so I proposed author of zerotierone, to provide support for such feature, AND made my own fork that has such support initially- unfortunatelly so far, I haven't had a chance to test it in live environment, but if interested, it's here: https://github.com/oskarirauta/ZeroTierOne/tree/notifier

I'll test it soon..

This fork adds a new parameter, -n which is used like this -n/root/zt-notifier.sh - and it is supposed to execute script/executable with either up or down argument when state of connection changes.

ofcourse /root/zt-notifier.sh is just for testing, but with changes, my zerotier.sh protocol file has executable flag and following parts have changes:
```
        fi

        sleep 1
        proto_run_command "${interface}" "${DAEMON}" -p0 -n/root/zt-notifier.sh "${path}"

        proto_send_update "${interface}"
}

proto_zerotier_teardown() {
        local interface="$1"

        proto_kill_command "${interface}"

        sleep 1
        # Remove existing link or folder
        rm -rf ${CONFIG_PATH}_"${interface}"
}

```

actual code for notification support doesn't grow heap a lot, I kept it simple, but since I wrote it in a bit of haste- I can't say it's perfect, it's more like a concept and I am not even sure- if it really works **when** it should post a notification through the script. So I am hoping that it gives a nice idea for feature and author of zerotier would write something similar and merge to mainstream. Atleast they were interested in the idea at the first that much that they asked for details, not sure though if they feel idea boring enough, to be not worth notice. That will remain to be seen...

oskarirauta:
output of ifstatus zt0 after using script:

```
root@gateway:/lib/netifd/proto# ifstatus zt0
{
	"up": true,
	"pending": false,
	"available": true,
	"autostart": true,
	"dynamic": false,
	"uptime": 1964,
	"l3_device": "zt0",
	"proto": "zerotier",
	"updated": [
		"addresses",
		"data"
	],
	"metric": 0,
	"dns_metric": 0,
	"delegation": true,
	"ipv4-address": [
		{
			"address": "172.28.28.1",
			"mask": 32
		}
	],
	"ipv6-address": [
		{
			"address": "fe80::XXXX:XXXX:fe4c:XXXX",
			"mask": 64
		}
	],
	"ipv6-prefix": [
		
	],
	"ipv6-prefix-assignment": [
		
	],
	"route": [
		
	],
	"dns-server": [
		
	],
	"dns-search": [
		
	],
	"neighbors": [
		
	],
	"inactive": {
		"ipv4-address": [
			
		],
		"ipv6-address": [
			
		],
		"route": [
			
		],
		"dns-server": [
			
		],
		"dns-search": [
			
		],
		"neighbors": [
			
		]
	},
	"data": {
		"code": 200,
		"hwaddr": "36:36:XX:XX:XX:XX",
		"id": "XXXXXXXXXXXXXXXX",
		"mtu": 2800,
		"name": "XXX_NETWORK"
	}
}

```

some personal data is redacted from output.

oskarirauta:
Latest commit on my fork fixes the issue...

oskarirauta:
One more thing is missing from my notifier fork- it should also give ifd name as arg... I know how to retrieve it- no problem, but if someone uses multiple networks, then we have a issue- it needs some further studying, though, I have a idea- but.. maybe later..

oskarirauta:
@Vavoon check out how it plays out at https://github.com/zerotier/ZeroTierOne/issues/1858

Vavooon:
@oskarirauta Thanks a lot for the contribution.
I've checked it and while I like idea of getting connection state updates, I'm not really sure that ZT devs are willing to merge them. So we should probably merge our current progress and then try to improve everything further.

For example, this protocol handler just waits 10 seconds before sending an update: https://github.com/openwrt/packages/blob/38f01ad2c9738c6c5a5dbda85019bb9cf5bc10f8/net/sstp-client/files/lib/netifd/proto/sstp.sh#L102

I think it should be sufficient to have a simple loop which would run until interface is not running and then we can call `proto_send_update`.

May I also see your `zt-notifier.sh`?

oskarirauta:
@Vavoon

I am not sure if they are interested in it either. There was a previous version of this in the past, but idea was forgotten/became unnecessary. He shared his PR to me and I worked through it and I've got it all working now _perfectly_..

I had issues that zt interfaces did not come up, so I changed your protocol- now it depends on wan and works just fine.
You can also have multiple connections simultaneously through network config and they all works fine.

I have created a script for network event changes on zerotier that handles it all. I have branched packages but won't be doing a PR, because this uses my _patched_ version of ZeroTier instead of mainstream.. 
https://github.com/oskarirauta/packages/tree/zerotier-event-scripting
(pushing in few minutes..)

Here's though my work on ZeroTier-one
da0ed006e871c4c188839cfbdb7c51b89df3f0e5 - add network event script support for zerotier

Usage is on release description:
https://github.com/oskarirauta/ZeroTierOne/releases/tag/1.10.2

And then a nice illustration picture as a proof of concept :)
zt-interfaces

oskarirauta:
> @oskarirauta Thanks a lot for the contribution. I've checked it and while I like idea of getting connection state updates, I'm not really sure that ZT devs are willing to merge them. So we should probably merge our current progress and then try to improve everything further.

What do you mean by merging? To packages? That's a no go- openwrt cannot go with my patched zt, or could- but in that case I - or someone - or a team, would need to maintain that patch- keep it up to date when ever new version of zt comes. That's not difficult. Unless it changes fundamentally, which is unlikely- but where do we get the resources for that? I got to be honest, I am between jobs and have some time now... Later on... Propably not that much and I'm already on a ton of projects and shuffle between them from time to time. Or did you mean merge progress between us 2? That suites me fine, you can find everything from the repo links that I've posted.

Right about that, they gently declined. Notification script seems to be huge security vulnurability. I wonder what then hotplug.d is - a huge backdoor..? Well enough about that, but I am pretty sure that anything like this can ever be merged there. But I really can't see another way except that zt would notify in some way- if not in a script like this, then the other- they kept rambling about API and polling it periodically which in my opinion... Well sounded really stupid- another software, or a complex cronjob for the task that runs alongside with zt and cannot be aware if it's out for a reason :O I mean, if I call ifdown zt0 - for a reason, I then should propably tell my _watchdog_ about my intentions aswell.. Okay, I could argue about it eternally but I'll stop now, it's pointless.

Let's try to be more productive- what we now do have, is something that works flawlessly and yet, keeps possibility of improvement- so let's try to find some light from the darkness. Current progress works as a proof of concept, that this is do-able. It just isn't do-able **with** zerotier's maintainers. So we need to think outside of the box. Maybe some ideas come by, or if not, then this must be abandoned, but let's try..

Base: we **only** need interface and action of event. Same details that would be available for hotplug.d - as with that information, all can be sorted out. Best would be that update information (network_update) since that was the phase where connection was established and ip was assigned. On up stage, only ifd was up without any crucial information. Extra information in data section was nice add-on, but when taking in consideration of netifd - it's just trivial useless extra information. Won't even show anywhere, without running ifstatus zt0.

 - So, their offer is to poll API constantly, which is heavy and pointless task, as in the end - even though my devices are packed with power- most routers are not. Lighter we go, more steady system- right? In my opinion there are numerous ways to do this- one being a cron job, which I don't like at all. It would mean a **huge** script and temporary file(s) containing hashes of previous json's, used to compare if there has been differences, and then a function that determines if connection state should be updated. And yes, this is a big decision, as everytime- it triggers firewall restart. Meaning that your connections behind upnpd etc - die, in cases where port is even momentarily closed. This would definetly be a huge and complex script, and I don't like them- it's difficult to edit and they often are more or less hacks when they become bloated. But that's what they offered.
 - Second solution is same as abowe, except without script- actual program instead that does it and even skips whole zerotier-cli as it can do it on it's own... Once again, bad idea. If they change their API, and there's no update done to the software......... Not a long term solution, atleast I don't have all the time in the world and unfortunately I tend to move from idea to another, meaning that I won't be updating it full time. Though, with good programming, dividing it to enough many files, comments, Source code propably could look something that other than author can work on as well- and finally, it would be possible to run it with pretty small footprint.

In my opinion, both of these are bad ideas. Still the approach I took before, was the best- this isn't even realtime. When netifd reaches information that we have a connection, it is possible that it is already gone as we cannot use all the cpu time for monitoring zt.

### But...

Here I throw the ball to you.. I've been burning the midnight oil here to do the progress to this stage and as it took a lot of time and effort to get through it, and without forggeting a huge disappointment that the whole idea was killed (I never said that they should merge my code, it was supposed to be a fully working example) by zt - I might have a small gridlock, maybe you have some way around this in your mind. But I remind you, idea must... be different. Zt was so much against the whole idea, that I am quite sure, that nothing even similar from far, will receive interest. I think that door was pretty much closed for good.

I got one idea, but that's pretty dead too- a _mini_ version of the patch... Nope... Too much changes still as the information in zt code so scattered, that few lines to correct place just doesn't do the job, because where the script is executed, it does not have enough information to provide. Second one, maybe we could link it to netifd shared library (if one exists..) and communicate directly to netifd.... No, won't work that way as I have no idea where to get interface name...

Your turn... Throw a idea.

> For example, this protocol handler just waits 10 seconds before sending an update:

I added delay of 3 seconds. 2 was sufficient, some times. But 3 was everytime. I think there could be a setting called delay, and if not used at all, it would default to 3- but at the moment this is our smallest problem.

> I think it should be sufficient to have a simple loop which would run until interface is not running and then we can call `proto_send_update`.

You are looking at wrong place. It's in the teardown. Zt doesn't _die_ that fast, for API it still tries to write some details to files that no longer existed.

> May I also see your `zt-notifier.sh`?

Sure thing. It's in the [branch](https://github.com/oskarirauta/packages/blob/zerotier-event-scripting/net/zerotier/files/lib/netifd/zerotier.script) - it's all out.

oskarirauta:
@mwarning join the conversation... I am out of ideas, but would love to hear if others aren't...

Vavooon:
@oskarirauta Alright, thanks for all the explanations.
I'll setup a test router next week in order to try your current progress and maybe to find some workarounds for protocol status update.
Just to be sure, is the main issue you're trying to solve the requirement to run `proto_send_update` only after the network/interface is fully up and has IP address(es) assigned?

oskarirauta:
@Vavooon 

> Just to be sure, is the main issue you're trying to solve the requirement to run `proto_send_update` only after the network/interface is fully up and has IP address(es) assigned?

Yes. What happens when you call ifup zt0 on the system:
 - first nothing (zerotier client works on it's magic for a moment)
 - interface comes up
 - initial details about connection become available through cli
 - zerotier (patched) posts network_up - current script test if json contains attribute status with value OK, if not - exit
 - Connection is established
 - cli data is updated
 - zerotier (patched) posts network_update - current script tests again if json contains attribute status with value OK, if not - exit
....
 - update posts keep coming - if connection state is still up and status says OK - script ignores cli data
 - script keeps firing every time notification is posted, as ifd is up and status is OK, it keeps ignoring cli data

This is when openwrt stops monitoring/working with API- I called ifdown wan and zerotier did not fire event "down" or "update" - one minute later, I called up again, and zerotier still was working (even though internet was offline for a minute), and no notification - so it's not important to follow up on notifications. That's why others are empty.

oskarirauta:
I might be on to something.

If you do...
`
echo change >/sys/devices/virtual/net/zt0/uevent
`

hotplug get's triggered.
I made 2 small hotplug scripts for iface and net classes that catch all events and output env to a temp file along with output of ifconfig zt0.
I also changed my protocol to not use script I created. When I bring zt0 up... hotplug is triggered when interface is created, but unfortunately - connection is not yet ready in that phase - so we don't have information about connection or it's state.
But when I wrote change to uevent in later phase when connection was up, I got promising results...

```
USER=root
ACTION=change
SHLVL=1
HOME=/
SEQNUM=2142
boot_variant=sysroot
IFINDEX=22
HOTPLUG_TYPE=net
DEVPATH=/devices/virtual/net/zt0
LOGNAME=root
DEVICENAME=zt0
TERM=linux
SUBSYSTEM=net
PATH=/usr/sbin:/usr/bin:/sbin:/bin
INTERFACE=zt0
PWD=/
SYNTH_UUID=0
zt0       Link encap:Ethernet  HWaddr 36:36:E8:4C:10:2D  
          inet addr:172.28.28.1  Bcast:172.28.28.255  Mask:255.255.255.0
          inet6 addr: fe80::3436:e8ff:fe4c:102d/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:2800  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:7 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:0 (0.0 B)  TX bytes:746 (746.0 B)
```

What here is important, is that we have INTERFACE mentioned in env, and also change as action. If we would get this to run, meaning that change event to uevent in sysfs - we could work something out.

Now, check [ManagedRoute.cpp](https://github.com/oskarirauta/ZeroTierOne/blob/notifier3/osdep/ManagedRoute.cpp) file in ZeroTierOne/osdep at lines 529 and 584. On around these lines are functions that seem to setup zerotier routing.
In both cases, _device is mentioned and this should be the interface name, as you know when you set up routes, you may used param dev <ifname> - so by making a much SMALLER patch that simply would, around this lines, write change to uevent file, we could trigger a hotplug event.

So, how about we write a small file that is compiled alongside with zerotier, that has a single purpose: trigger the event for selected ifd. And then patch this files with only 3 lines, that would trigger that function in the smaller file?
That shouldn't be very difficult to keep up to date, as it's simply just 3 lines-
 - #include for hotplug triggerer
 - trigger on route add
 - trigger on route removal

I am quite confident that it's not even worth the effort at all to make a PR to ZeroTierOne.git - but if change is this small...
This propably _could_ be approved as a patch with openwrt packages, as change is so minor and system would greatly benefit on it.

Some reading for you: [sysfs-uevent](https://github.com/torvalds/linux/blob/master/Documentation/ABI/testing/sysfs-uevent)

Now, we could just write to file, but... I am pretty sure there are other methods to send this change through it. We should figure out the best way to write to that file, most simple and reliable way. But in the end, we just could write to that file..

Unless- you come up with better solution... I'm planning to do some tests on this a bit later, it's possible that this isn't usable idea at all, but I have some hope left for it.

EDIT:
Another good place for patch would be at line 273. Propably even best place for it.
op is just char* with either add or del in it and localInterface contains if device. And we can forget about extra file, if we just follow their lead and use execl and echo to trigger the event.

It would be a very small patch then. I'll check later. I'm feeling a bit tired and think that before beginning, a small nap could make everything better ;) - afterall, I got this idea last night while sleeping and woke up to it :D

oskarirauta:
That was so good idea. Unfortunately it never executes. Define LINUX is not set. Let's see if routing is somewhere else too.. Maybe LinuxNetLink.cpp..

oskarirauta:
Ok. Redirecting with > is impossible with ::execl- ::system works find.
Line 821 on LinuxNetLink.cpp:
`

        close(fd);
`

If following is added after that line..
`

        ::system(std::string("echo change >/sys/devices/virtual/net/" +
                std::string(ifaceName) + "/uevent &").c_str());
`

hotplug.d is registered and output test output follows:

```
USER=root
ACTION=change
SHLVL=1
HOME=/
SEQNUM=2210
boot_variant=sysroot
IFINDEX=29
HOTPLUG_TYPE=net
DEVPATH=/devices/virtual/net/zt0
LOGNAME=root
DEVICENAME=zt0
TERM=linux
SUBSYSTEM=net
PATH=/usr/sbin:/usr/bin:/sbin:/bin
INTERFACE=zt0
PWD=/
SYNTH_UUID=0
zt0       Link encap:Ethernet  HWaddr 36:36:E8:4C:10:2D  
             inet addr:172.28.28.1  Bcast:172.28.28.255  Mask:255.255.255.0
             inet6 addr: fe80::3436:e8ff:fe4c:102d/64 Scope:Link
             UP BROADCAST RUNNING MULTICAST  MTU:2800  Metric:1
             RX packets:0 errors:0 dropped:0 overruns:0 frame:0
             TX packets:7 errors:0 dropped:0 overruns:0 carrier:0
             collisions:0 txqueuelen:1000 
             RX bytes:0 (0.0 B)  TX bytes:746 (746.0 B)
```

So with this single line patched to code, we can receive a hotplug when route was added and can work with it. I would still though change it a bit, to check that sysfs path exists. 

This though will update very frequently...
```
Sun Jan 29 19:31:47 2023 daemon.notice netifd: Interface 'zt0' is setting up now
Sun Jan 29 19:31:49 2023 user.notice root: hotplug-net activated with add
Sun Jan 29 19:31:54 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:31:54 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:32:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:32:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:32:49 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:32:49 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:32:59 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:32:59 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:33:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:33:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:33:49 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:33:49 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:34:04 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:34:04 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:34:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:34:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:34:49 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:34:49 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:35:09 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:35:09 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:35:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:35:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:35:49 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:35:49 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:36:14 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:36:14 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:36:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:36:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:36:49 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:36:49 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:37:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:37:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:37:19 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:37:19 2023 user.notice root: hotplug-net activated with change
```

I wonder, if this would be better if that line would be added to line 1055 where ip address is set. I'll try that..

oskarirauta:
Ok. I moved it to line 1055 and got promising results, exactly the same results. I added test for routes since I was worried that routing information isn't updated fast enough- though it is not even necessary for this, and propably that could be fixed with sleep 1 in the script, as it propably will be added very soon after ip address is set. But atleast while testing, this didn't seem to be necessary, route was already updated.

Here's output, with irrelevant information redacted:

```
root@gateway:~# cat /tmp/net-zt-hotplug 
ACTION=change
...
IFINDEX=31
HOTPLUG_TYPE=net
DEVPATH=/devices/virtual/net/zt0
DEVICENAME=zt0
SUBSYSTEM=net
INTERFACE=zt0
SYNTH_UUID=0
ifconfig:
zt0       Link encap:Ethernet  HWaddr 36:36:E8:4C:10:2D  
             inet addr:172.28.28.1  Bcast:172.28.28.255  Mask:255.255.255.0
             inet6 addr: fe80::3436:e8ff:fe4c:102d/64 Scope:Link
...
             RX bytes:0 (0.0 B)  TX bytes:746 (746.0 B)

route:
...
172.28.28.0/24 dev zt0 proto kernel scope link src 172.28.28.1
root@gateway:~#
```

And here's some syslog output:
```
Sun Jan 29 19:46:11 2023 daemon.notice netifd: Interface 'zt0' is now down
Sun Jan 29 19:46:38 2023 user.notice root: CUT
Sun Jan 29 19:46:52 2023 daemon.notice netifd: Interface 'zt0' is setting up now
Sun Jan 29 19:46:54 2023 user.notice root: hotplug-net activated with add
Sun Jan 29 19:47:06 2023 user.notice root: hotplug-net activated with change
Sun Jan 29 19:53:35 2023 user.notice root: CUT
Sun Jan 29 19:54:13 2023 daemon.notice ttyd[4406]: [2023/01/29 19:54:13:4104] N: rops_handle_POLLIN_netlink: DELADDR
Sun Jan 29 19:54:13 2023 daemon.notice ttyd[4406]: [2023/01/29 19:54:13:4126] N: rops_handle_POLLIN_netlink: DELADDR
Sun Jan 29 19:54:13 2023 user.notice root: hotplug-net activated with remove
Sun Jan 29 19:54:14 2023 daemon.notice netifd: Interface 'zt0' is now down
```
zt0 was down on 19:46, and then I brought it back up at 19:46:52. 2 seconds later, interface was "up", and 12 seconds later from that at 19:47:06 interface received ip address, and script had fired.
I then left it at running, and no messages were received. On 19:54 I asked it to go down.
Looks pretty promising?

EDIT:
not putting more logs here, but next time connection establishment took only 2 seconds, so it varies, though- if we try with this- it doesn't matter how long it takes, whether it was fast or slow, necessary details are available and whole thing can be done by hotplug.d script, though I would just add a small script to there which compares action to change,
and then compares uci get network.${INTERFACE}.proto == "zerotier" and in this case, executes the update script.

Got other ideas?

oskarirauta:
@Vavooon 

Patched zerotier is at branch notifier3 on this [repository](https://github.com/oskarirauta/ZeroTierOne/tree/notifier3)
and commit details are here [2ee2bc9](https://github.com/oskarirauta/ZeroTierOne/commit/2ee2bc949e7ead3ca0b8831f55292725d91542a0)

It first checks that path to uevent in sysfs exists and is writable, then it checks if running user is capable of running commands (has shell), and then executes triggering and backgrounds that process (& at the end).

oskarirauta:
@Vavooon 

I have updated scripts accordingly, and I have made a new branch for it. It contains local patch for uevent support and has removed traditional initscript and /etc/config/zerotier config file.
In the commit message is full usage and description on what it does.
It is complete out-of-the-box solution.

branch is here: [zerotier-proto](https://github.com/oskarirauta/packages/tree/zerotier-proto)
commit: 6024dc5629e9d56b87c9610ede3597683a8aca91

Happy building. If you use x86-64 arch, I have a pre-compiled package.

Vavooon:
@oskarirauta Your work looks really awesome!

Few notes:

You likely don't need to call `echo` just for writing to a file, instead you could write to it directly:

```

ofstream ueventInterfaceFile ("/sys/devices/virtual/net/" + std::string(iface) + "/uevent");
if (ueventInterfaceFile.is_open())
{
  ueventInterfaceFile << "change";
  ueventInterfaceFile.close();
}
```

oskarirauta:
I know, but I tried to keep it as short as possible in lines of code- and ::system("echo...") felt like shortest way to that..

oskarirauta:
@Vavoon - whether I change it from echo to that, do you think there's any chance to push it to here? Because I cannot push it to zerotier mainstream, and even if I could- I highly doubt they would be interested.

So- what do you think- would it be better if I just write to file? End result and benefits are pretty much the same, though calling exec (or system in this case) is ofcourse always something that is looked like a bit crooked..

oskarirauta:
Line count though is identical- 5..

oskarirauta:
@Vavooon 

Is this the direction how we should approach this, or are any other ideas floating around? This definetly works though.
Did you have chance to try it?

Vavooon:
@oskarirauta I haven't had a chance, was going to check it this evening though.
I'll probably use a `mipsbe` board for my tests.

oskarirauta:
@Vavooon that works just as well, but unfortunately I cannot provide your pre-compiled build, because I don't have suitable toolchain installed.

Vavooon:
@oskarirauta 

NP, I can compile it by myself, just need to set up everything.

I'm wondering whether triggering `change` event isn't too dirty hack, and I'm either wondering what other protocol handlers do in order to collect proper info about connection state. I've checked it on my PC and there are no any `change` events when I unplug/plug my LAN cable and IP address gets removed and assigned to the interface. So I'm not sure we should really do it for Zerotier.

From very beginning, could you explain why a simple `while-sleep` inside the protocol handler won't work in our case?
E.g. 
```
while [[ "$(zerotier-cli get "$network_id" status)" -ne "OK" ]]; do
  sleep 1;
done
// Read all the protocol information here
```
I know, this will read the information just once but that should be enough in most cases, unless there's a chance that your IP will get changed during some reconnect and you really need to see this IP change in your UI.

oskarirauta:
Look at qmi protocol. While initiating connection, all the data is gathered during the process. But it's full of check ups for errors - in every turn, it is ready to exit if somethings not looking good enough for go. I counted 21 cases when script decides to fail connection attempt.

To me, it feels like waster of resources, especially since it blocks zerotier- remember how I added & to the system call? Without it, it caused a full stop in zt and a failure to get that data- I suspect polling does the same thing- though sleep 1 - wouldn't block very often and for very long.

It's just that, how we then handle failures? Do we want the netifd system to stuck on a single protocol waiting eternally?
Or do we add a counter of sort, for example that loop will automaticly exit after 10 times = ~10 seconds? What if wan connection is real slow?

And then again, whilst you are on a mips- you know there are routers not very powerful? I still have at the attic (not running though) few original Linksys WRT54G's- can't remember specs and not looking up for them, what you offer, is basicly the same they offered at zerotier, constantly polling until results necessary are ready.

Also, status = OK is not sufficient. It just means we have authenticated, not that we have ip address, or route.
Though.... We could have extra checking for sufficient details, like this..

```
zt_ok=0
zt_ip=""
zt_cnt=0
while [[ "$zt_ok" -eq 0 -a "$zt_qnt" -lt 10 ]]; do
  sleep 1
  zt_status=$(zerotier-cli -D${CONFIG_PATH}_${cfg} get ${network_id} status)
  [ "$zt_status" -eq "OK" ] && {
      # parse ip to zt_ip, then if ok, parse other details as well, if not ok, clear zt_ip
  }
  [ -z "$zt_ip" ] && zt_ok=0
  zt_cnt=$((zt_cnt+1))
done

[ "$zt_ok" -eq 0 ] && {
   echo "failed to setup zerotier"
   return 1
}

```

That's just a mockup of script, propably with full of errors, but gives some idea on how to do some extra checkup which is necessary to succeed.
Personally, I don't like long scripts, when they get long enough, they are really difficult to maintain, especially, by 3rd party, but if enough time has passed, even by original author. Though I also hate perl and python and can't figure out why anyone would do things in something else than c/c++ :D
Also if you look at development of openwrt, complex scripts are reduced year by year, by providing custom software to avoid need to rely completely on scripts, such as netifd, procd, uqmi... Even though they all still use scripts, and it's good that they do- included support scripts are much easier to read and use, for example, /lib/netifd/netifd-proto.sh - it's fitted in 478 lines and it's not tightened, instead it has been kept very readable.

Check for example banip script (broken, due to lack of nft support, but yet works as example, if I remember scripts complexivity and length correctly) - or some protocols for some other vpn's, hundreds of rows of very complex things and real pain in the *ss to adapt to work when situation deviates.

This approach by far is most likely code/scripting wisely the best solution- **if** it would come from mainstream of zt.
Unfortunately answer is no.

Most common binary based protocol handlers, such as uqmi: qmi proto - build their software around in a symbiosis of script, supporting netifd. In this case it's no wonder, as both pieces, netifd and uqmi are designed for Openwrt.

Then there are those script hells, like for few other vpns...

And then there is some others, using such scripts as you proposed, but in this case- if zt gets blocked with zerotier-cli requests- connection originally failed, no ip - ever. I see possibility of this to happen on lesser routers, especially when connection is slow- that blocking with cli, would cause a failure to connect in once in a while due to incomplete/timed out handshake or what ever zt protocol has in it.. And I think it's heavier, especially when you have multiple networks, even though most users need just one.

I really am not the right person to say which approach would be best, because of my favouring of code instead of scripting and thinking that it would be more problem-free if notification would come from zt itself- 

I also like the idea that it uses less resources- current version with path, initiates zerotier connection and leaves it in a pending state, and does other things- until zerotier informs that it's ready. I run 2 servers, both have web servers, and proxies, php, etc-- used to have exim also for mail, but now days apple handles it for me and I am pretty sure they have better high availability rate that I could ever provide ;)
Point to run openwrt is very simple- it's the most light weightest solution (and also very simple), even when those vps/kvm servers have multiple cores and 2 or 4 gb of ram available, I do not understand who wants a server with full blown desktop support- or even why the system would need to hold os that requires gigabytes, when my setup needs only 106mb packed with stuff, such as podman and vpn connectivity along with numerous tools- after all, server is supposed to serve, in this case, provide content in form of web pages- that content should be the heaviest disk using part, not the system, why so many depencies when they clearly are unnecessary in so many occasions, isn't it so, that every piece of software exposes it's threat in it's bugs- and we know... that there is very little of bug free software, especially if they are bigger applications. So, why sasl, openldap, gpg, dbus, systemd, journald, X11/X-Free86, gnome and kde, when you need in the end need nginx for http, and caddy to serve it as https?

Yes, propably well explained point- which doesn't make patching zerotier's code any more approvable. Patch itself isn't that dirty in my opinion, it's very simple and clean- and would harm no one, even if they used different technique. But I cannot vouch for it's approval. I think judgement should come from zerotier maintainer mwarning here at openwrt.

oskarirauta:
@mwarning
Could you please be kind and checkout commit [f99ad3e](https://github.com/oskarirauta/packages/commit/f99ad3e369238f2cd1bb638b132600a53b471467) and possibly even test it?

Description and usage are on previous [commit](https://github.com/oskarirauta/packages/commit/fbeb858ef6ddf53e18808362c3c592b586771da3)

Our concern is this [patch](https://github.com/oskarirauta/packages/blob/zerotier-proto/net/zerotier/patches/0003-add-uevent-trigger.patch) - do you consider it as un-acceptable modification?

How about rest of the protocol support/idea? Rubbish, or on-to-something?

Vavooon:
@oskarirauta You're right, it will still fail in some cases, for example when you don't make the client authorized for long time at Zerotier web dashboard.
On the other hand, neither the uevent change hook handles all the cases e.g. disconnects etc., so if we really want to cover all the cases, we need to implement a C/C++ daemon to watch all the connection changes.

While I also like such maximalist approach, I'd rather go with 80/20 rule and handling only bare minimum here.

oskarirauta:
my routine also executes just once, adding it to route routine, it executes... often... Details should remain pretty much the same, unless you change ipaddress of client through dashboard, but that does get registered by uevent trigger.

If you change configuration there, for example, set a bridging mode - you will have to restart zerotier client anyway to get it working. If connection is disconnected, zerotier will automaticly resolve the issue- or die, and ip address remains the same, unless changes made in dashboard.

Unfortunately zerotier seems to lack some sort of keepalive & monitor method, or does not reply on those cases, as when I disconnected wan- it seemed to be completely unaware that network connection is lost, even when I used network by pinging other devices on the network, which ofcourse failed since wan was not available- when I brought wan up, zerotier was up as well. Nevertheless, if we'd want to cover all the cases, we would need to restart zerotier every time, as it seems to handle connection restore on it's own...

maximalist approach? I considered this as minimal approach o_O - just my opinion but I considered this as minimal add to processing + minimal failure rate..

oskarirauta:
Could we get a script as comparison then? Hopefully from you as scripting is my weakness... Tens of commits until scripts work when uevent triggering in c++ succeeded after few tries and that just because I added the routine to incorrect place.. One reason for why I am bias to judge which would be better approach... Scripts just aren't my thing and I try almost always to avoid them at all cost..

Vavooon:
@oskarirauta Alright, will try to write one, hopefully in this case I don't need to compile ZeroTier from sources so should be much faster.

oskarirauta:
No you don't- idea behind the script is that makes the patch obsolete.

mwarning:
@oskarirauta I will test it. :-)

oskarirauta:
@mwarning enjoy :)

Somewhere on this thread is also luci protocol if you are into pretty things...

oskarirauta:
@Vavooon you should add atleast requirement for wan connection, as you have a eternal loop there- your router might in worst case become in-accessible if you rely on connection through network, when other network methods won't start because it stucks to the loop. That's what happened to me when zerotier-cli was stuck, that's why I added that &.......

oskarirauta:
Suggestion:

Put your loop to a script that eventually exits if no success, to another file and execute it with & so it backgrounds.
When initiating connection or tearing down, kill that script if it is running....

Vavooon:
@oskarirauta Thanks, will improve loop handling.

Could you explain what `( proto_add_host_dependency "$cfg" '' "wan" )` does?
I couldn't find it in any examples.

Vavooon:
@oskarirauta I've added an error check, it throws an error on status other than `OK`
https://github.com/zerotier/ZeroTierOne/blob/ab503902c87b262f4a78d0a386b7529be12171f5/attic/macui/ZeroTier%20One/Network.h#L21

oskarirauta:
Adds dependancy on ifup wan…Lähetetty iPhonestaVavooon ***@***.***> kirjoitti 1.2.2023 kello 14.58:﻿
@oskarirauta Thanks, will improve loop handling.
Could you explain what ( proto_add_host_dependency "$cfg" '' "wan" ) does?
I couldn't find it in any examples.

oskarirauta:
That won’t work. Connection establishment delay varies a lot. On LTE modem, average delays are between 2s-20s on 150mbps (max, never get that much though) connection. Status will always be !ok and you need to take care of situation where it does not answer at all (hangs) - that is why that separated script running in background would be a savior..Lähetetty iPhonestaVavooon ***@***.***> kirjoitti 1.2.2023 kello 16.07:﻿
@oskarirauta I've added an error check, it throws an error on status other than OK
https://github.com/zerotier/ZeroTierOne/blob/ab503902c87b262f4a78d0a386b7529be12171f5/attic/macui/ZeroTier%20One/Network.h#L21

Vavooon:
@oskarirauta 
re: proto_add_host_dependency "$cfg" '' "wan"
Thanks, could you please show me any examples where it's being used? I haven't seen any other vpn protocol handler does that.

re: That won’t work. Connection establishment delay varies a lot. 
I have tested it locally and it was waiting about 15 seconds for `status` to become `OK` without any problems.

Vavooon:
@oskarirauta Also, could you please create a PR with your `luci-proto-zerotier` package? Looks good enough to me.

oskarirauta:
@ Vavoon

> I have tested it locally and it was waiting about 15 seconds for `status` to become `OK` without any problems.
Looks actually pretty good solution; but add also that it won't exit when status is "REQUESTING_CONFIGURATION"... Otherwise it exits literally every time... Though I wonder.. Why it did works without support for that status... It should had exited, as that should had been status during setup.. It might be that it is because of "not backgrounding" and process has a problem and actually it took 15 seconds to get first answer which happened to be OK..

But you should also support situation where cli didn't answer at all (something is wrong) - or connection establishment takes more than minute: trigger respawn..

> Thanks, could you please show me any examples where it's being used? I haven't seen any other vpn protocol handler does that.

I Updated my previous posts, because I was away and answered with email which made answers a bit messy, there's examples and explanations on what is what.

> Also, could you please create a PR with your luci-proto-zerotier package? Looks good enough to me.

You cannot attach that this, it must be separate, as it must be to made as a PR against [luci](https://github.com/openwrt/luci) repository.

oskarirauta:
@Vavooon 

explanation for [proto_add_host_dependency](https://openwrt.org/docs/guide-developer/network-scripting)

oskarirauta:
@Vavooon 

I just finished creating tailscale protocol. A little different approach with that, but- in the script I made it timeout like this:

```
local status

while true; do
	status=$($TAILSCALE status)
	[ "$status" = "Tailscale is stopped." ] || break
	sleep 1
	counter=$((counter+1))
	[ "$counter" -ge 60 ] && {
		export "$2=1"
		return
	}
done
```

As an example where in a minute it decides to fail the connection.

Vavooon:
@oskarirauta

> But you should also support situation where cli didn't answer at all (something is wrong) - or connection establishment takes more than minute: trigger respawn..

Could you show me an example when cli doesn't answer? When it returns an error (e.g. it can't connect to the daemon), there's a fallback to empty string which executes the loop again.

> explanation for [proto_add_host_dependency](https://openwrt.org/docs/guide-developer/network-scripting)

Thanks, although I haven't seen any specific usages with `wan` as an argument. There's only such usages in some protocols:
https://github.com/openwrt/routing/blob/dc0adf7dced5f6de8bcf096e3bb6b2ec67f91666/batman-adv/files/lib/netifd/proto/batadv_hardif.sh#L30

Vavooon:
@oskarirauta 

> You cannot attach that this, it must be separate, as it must be to made as a PR against [luci](https://github.com/openwrt/luci) repository.

Exactly, it has to be a separate PR, so since it's entirely your code, it would be better if you create this PR.

oskarirauta:
@Vavooon - sure.. But it depends on this, so let's hold it for a moment..

Vavooon:
@oskarirauta 

> Example of cli stalling? Well, it just doesn't do anything, cli connects to socket, but since socket doesn't respond, it waits that response and seems to have pretty long timeout val..

I'm afraid there's nothing we can do here since protocol script has to wait for `zerotier-cli` command to exit and we can't kill it during execution.

Re: the timeout you've suggested
I've checked your references to QMI proto and they have timeout for SIM card/device initialization, but Zerotier does not rely on any hardware.
On the other hand, I'm not sure that it requires any timeouts to restart since it might be ok for the service to wait for long time e.g. when wan connection is not yet available, even if it does require some care, we can't do it effectively because we need to have a background process to watch it.

oskarirauta:
@Vavooon 
> Re: the timeout you've suggested I've checked your references to QMI proto and they have timeout for SIM card/device initialization, but Zerotier does not rely on any hardware. On the other hand, I'm not sure that it requires any timeouts to restart since it might be ok for the service to wait for long time e.g. when wan connection is not yet available, even if it does require some care, we can't do it effectively because we need to have a background process to watch it.

I'm working on something, I'll get back later if it produces anything good- but this has some pointers from your script, and is supposed to run single zerotier instance even if multiple networks would be in use.. as this **does not** use my patch or later version of it, there's some issues that I am trying to sort out... and a huge cleanup process..

Vavooon:
@oskarirauta Thanks for your finding.
I was only ensuring that it works by checking it's IP address.
Alright, I'll probably implement the same hack I for routes as I did for addresses.

oskarirauta:
It needs some other minor changes as well...
Here..

```
	local addresses="$(ip a list dev "$cfg" scope global 2>/dev/null | grep inet | awk '{print $2}')"

	[ -z "$addresses" ] && {
		echo "zerotier failed to receive ip address from ${network_id}"
		proto_notify_error "$cfg" $"NO_IPADDRESS"
                return 1
	}

	for address in ${addresses}; do
		case "${address}" in
			*:*/*)
				proto_add_ipv6_address "${address%%/*}" "${address##*/}"
				;;
			*.*/*)
				proto_add_ipv4_address "${address%%/*}" "${address##*/}"
				;;
			*:*)
				proto_add_ipv6_address "${address%%/*}" "128"
				;;
			*.*)
				proto_add_ipv4_address "${address%%/*}" "32"
				;;
		esac
	done

	local routes="$(ip r list dev $cfg | awk '{ print $1 }')"
	local routevia routesrc routemetric

	for route in ${routes}; do
		case "${route}" in
			# ipv6 routes not supported at this time
			#*:*/*)
			#	proto_add_ipv6_route "${route%%/*}" "${route##*/}" "$routevia" "" "" "$routesrc"
			#	;;
			*.*/*)
				routevia="$(ip r show $route dev $cfg | grep via | awk '{ print $3 }')"
				routesrc="$(ip r show $route dev $cfg | grep src | awk '{ print $7 }')"
				routemetric="$(ip r show $route dev $cfg | grep metric | awk '{ print $9 }')"
				proto_add_ipv4_route "${route%%/*}" "${route##*/}" "$routevia" "$routesrc"
				;;
		esac
	done
```

Vavooon:
@oskarirauta 
I managed to improve IP addresses parsing.


In fact, `netifd` does not require routes for work (at least for me) and it wasn't working rather because wrong `/32` prefix was set for IP addresses. 

Now it sets masks properly and I like it more because it uses JSON instead of raw strings.

net/zerotier/files/lib/netifd/proto/zerotier.sh
```diff
@@ -0,0 +1,173 @@
...
+		rm -f "${path}"/identity.public
+	fi
+
+	proto_run_command "${interface}" "${DAEMON}" -p0 "${path}"
+
+	sleep 1
+
+	local status
+	while true; do
+		status="$(zerotier-cli -D${path} get ${network_id} status)" || ""
```
mwarning, line 99: Not sure of `|| ""` is needed. But maybe a redirect of stderr to /dev/null might be a good idea.

net/zerotier/files/lib/netifd/proto/zerotier.sh
```diff
@@ -0,0 +1,173 @@
...
+		[ "$status" != "REQUESTING_CONFIGURATION" -a "$status" != "" ] && break
```
mwarning, line 100: Proposal: `-a "$status" != ""` => `-a -n "$status"`

oskarirauta:
> @oskarirauta I managed to improve IP addresses parsing.
> 
> In fact, `netifd` does not require routes for work (at least for me) and it wasn't working rather because wrong `/32` prefix was set for IP addresses.

Maybe netifd doesn't but linux seem to does, I wasn't able to connect to any other host in my zerotier network. Not directly, and not by managed routes either.
 
> Now it sets masks properly and I like it more because it uses JSON instead of raw strings.
Great.

I also got my script set finished, it's at [here](https://github.com/oskarirauta/packages/commit/ef483b323a63ff7b7d85ff27d84be953e09aea70) if interested, it's rather complex, but I tried my best to also write it to be readable as well. It has different approach as it runs only one instance of server, so it cannot be used with multiple secret identities, though I don't see why anyone would use multiple identities but who am I to judge if someone misses such feature.

I came to conclusion that timeout support is necessary, it seemed to be my problem with init script as well why my connections weren't succesful. I made it to retry connection if not succeeded in one minute- in that time, connection surely has stalled. I have a 150mb mobile broadband connection, but reception is poor and network is very busy, so I never get more than 10mb, and on busy hours, I might not get even 1mb and there are about 45 second gaps sometimes on connection and these are times when some connections fail, zerotier being one. This is 3rd edition of this script now, and I built those 3 editions for about a week, so to test it, I restarted it numerous times and I got ~10 times when connection failed and never finished during that time. After implementing a retry, I only get retry notifications to logs and eventually connections are up.

net/zerotier/files/lib/netifd/proto/zerotier.sh
oskarirauta: I noticed same. I used some pointers from your script and also had changed it to form that mwarning proposed.

Vavooon:
@oskarirauta 

> Maybe netifd doesn't but linux seem to does, I wasn't able to connect to any other host in my zerotier network. Not directly, and not by managed routes either.
> 
> EDIT: nope - verified; it does NOT work without routes and netifd erased them again. Please add routes. Connection to 10.100.0.0/24 or 10.95.0.0/24 does not work without routes.

That's really weird because when I run `ip route list dev zt0` once it is connected, I can see all the routes there.
I can also ping other hosts in my ZT network.

Vavooon:
@oskarirauta 

> I tried with your script. No go. I added routes manually with ip command, and all worked out. I used ifdown to put it down, and then ifup and routes appear, but before protocol is finished, they vanish - and no routes again. I didn't add that routing support for nothing- but it won't hurt to put it there to make it work in all environments, maybe it's something else that has that side-effect, but adding routes fixes it. If I manually when no routes are there, renew connection with zerotier-cli leave/join routine without interfering with netifd, I get routes.

Alright, if it still doesn't work for you, I'll try to implement routes parsing and adding in the same way I have done with addresses. I already have started with routes, but decided not to proceed since it was working to me.

> Yes, it stucks at requesting_configuration. Good luck with them, I'm banned and they don't seem to be interested in that much improving it. But it's pretty easy to time it out, and restart connection attempt, like I show'd you in mine.
> It's pretty much just few lines. Geez, I'm getting 18kb/s... This must be a record of poor connection..
> It's just that one thing I cannot understand- why make it limited to optimal environment when it's not far from working, please make it goooood :)

Might be some small watchdog script running in the background and only responsible for killing the daemon when it stucks in this state for too long. Or a patch for ZT to kill itself using the same logic.
After all, it can stuck not only at the startup but also some time later, e.g. I've seen some bug reports complaining that connection can't be re-established when switching from one WiFi network to another.

oskarirauta:
@Vavoon - thanks. They are available in the same json. No, do not kill the daemon, it's unnecessary - just re-join with client instead. Daemon ends connection process when asked to leave and re-join.

How about this? Add a counter to loop, which doesn't exit - just use client to re-join and then resets value of counter, so it tries it eternally...? I would ask you to just pull your ethernet cable in the middle.. but I guess you would never be fast enough...

oskarirauta:
Now I have good connection (for awhile) - I triggered both my networks up and down constantly, and it worked out every time (with my own script, but it doesn't matter as end-result is the same)- so it's definetly related to network speed. I tried to overload the network but I just exceeded packet count of connection on other uses of networking.

Screenshot 2023-02-13 at 13 52 49

Now I'll go watch some netflix as it appears that I might be able to receive some fine image quality instead of pixel art ;)

Vavooon:
@mwarning Thanks for your hints, I've fixed it as you asked.

@oskarirauta I have added routes re-applying, please let me know whether it works for you.

oskarirauta:
@Vavooon
it does not work. You need to set it to route managed routes through local zt address, now it routes through zt dev, but via is missing.

if managed route to is 10.100.0.0/24, through zt0 interface that has ip address 172.28.28.1 route would be
10.100.0.0/24 via 172.28.28.1 dev zt0, your script skips via. It needs some trickery, as it's not exposed through json from cli and this is only for managed networks. Check from my script, it's that "gw" I used.....

I made a discovery- if netifd "removes" routes upon connection, zerotier eventually re-creates routes, apparently when this uevent _change_ event would occur, if it would be supported. Problem is though that this happens after unknown time, I noticed that when I used your script, and routes failed; I started to work on my cni protocol, I noticed that about 20 minutes later routes appeared. So eventually zerotier _renews_ - but for example, in my case- this sucks as if I reboot my remote server, I more or less prefer it available after reboot, not 30 minutes later- and is this certainly even happening in specific time or is it related to some other thing happening on zt network, for example that server rebooting/re-newing it's zt connection...

oskarirauta:
@Vavooon 

this is how routing table (wwan0 and default redacted) looks with my script:
```
root@gateway:/etc/config# ip r show
10.0.0.0/16 dev br-lan proto kernel scope link src 10.0.0.1 
10.95.0.0/24 via 172.28.28.95 dev zt0 proto static metric 5000 
10.100.0.0/24 via 172.28.28.100 dev zt0 proto static metric 5000 
172.28.28.0/24 dev zt0 proto static scope link src 172.28.28.1 
172.28.28.0/24 dev zt0 proto static scope link metric 5000 
192.168.191.0/24 dev zt1 proto static scope link src 192.168.191.1  
```

Identical to what zerotier does...
What if you would change it a bit?
Get routes from `ip -json r show dev $interface` - would that be useful?

oskarirauta:
I have ip-full - does ip-tiny support -json flag? If not ip-full would need to be added as depend instead of ip. I unfortunately have no host available with nothing that wouldn't be using "maximal" versions of everything (except that I do stick with musl instead of glibc)..

oskarirauta:
This might have something to do with that when network speed is very slow...

https://github.com/zerotier/ZeroTierOne/commit/3c9ea2b6674819612cd88a2e141e0f8f633b06cd

Vavooon:
@oskarirauta I just checked how it works on my desktop PC and I do not see any `via` routes:
```
10.244.0.0/16 dev zt7nnmxtuc proto kernel scope link src 10.244.166.134 
172.16.42.0/24 dev ztkse2sbso proto kernel scope link src 172.16.42.128 
172.30.0.0/16 dev ztnfahr3y4 proto kernel scope link src 172.30.166.134
```

I can confirm that I can establish a SSH connection with another peer in my network right after Zerotier start either without any `via` entries.

Do you have any exotic network configuration?

oskarirauta:
@Vavooon if advanced is exotic.. Meaning that on my.zerotier.com, it's under advanced title.

There is section "managed routes" - which... Well, I'll show example as I think it describes better what it does..

**router1**: provides lan 192.168.1.0/24 with it's own ip 192.168.1.1, is connected to zerotier network with ip address 172.28.28.1
**router2**: provides lan 192.168.8.1/24 with it's own ip 192.168.8.1, zerotier ip: 172.28.28.8

Now with managed routes, I can se 192.168.1.0/24 accessible **as** 192.168.1.0/24. Meaning that my router2, and all machines in it's subnet, can accss 192.168.1.0/24 as is.

Machine at 192.168.8.50 can ping/ssh/whatever, machine at 192.168.1.5, and if I set router2 as managed as well, vice versa. That's when via's are used.

I might have explained it in confusing, let me know, if I did and I'll try again..

Vavooon:
@oskarirauta Alright, should be working now, I added either `gateway` and `metric` to routes.

oskarirauta:
@Vavooon 

Sorry, I've been busy doing other stuff. But I just tested this. It works almost perfectly now, it just didn't work unless I manually restarted it **after** my wan connection is ready. If zerotier is initiated before wan is ready, log contains this:
```
root@gateway:~# logread | grep zt0
Tue Mar 21 17:20:51 2023 daemon.notice netifd: Interface 'zt0' is setting up now
Tue Mar 21 17:20:51 2023 daemon.notice netifd: zt0 (3936): sendto: Network unreachable
```
And it won't start unless I do
```
root@gateway:~# ifdown zt0
root@gateway:~# ifup zt0
```

oskarirauta:
I gave some recommendations. Seems that setting delay to even 1 is enough. As my network is from time to time (read almost all the time) so slow; I added delay 2 to be sure.

Still, it would be best if there would be a single instance of service daemon, so you could run just one instance in a dedicated port (for example, recommended 9993) for all your network; but I think that usually people only have 1 network so it's not maybe then a real world problem. Or do they connect to multiple networks? I don't know and I can only represent myself.

Vavooon:
@oskarirauta I'm trying to figure out the correct way to implement WAN availability check.

Looks like `proto_add_host_dependency` is a good option, although I'm not sure how to use it:
If I got this right, this command makes protocol handler to wait until specific host becomes available.

In your example you just use `proto_add_host_dependency "$interface" '' "wan"` but I haven't seen any examples of such usage, especially I'm not sure whether it's good to use `wan` as interface, since not every setup has interface with such name.

I'm not sure what is the purpose of `delay` since `proto_add_host_dependency` should do all the job.
I expect to try such usage
https://github.com/openwrt/chaos_calmer/blob/22375a2fb87271dadcdb30f5e825c168ba73d502/package/network/ipv6/6to4/files/6to4.sh#L40

oskarirauta:
@Vavooon

Yes, it is based on documentation and other samples that do **not** rely on wan, instead they rely on different tunlinks, I didn't find any depending on wan- but it wouldn't mean that it wouldn't work- it does.

Unfortunately, not the way I thought it would. It would seem that it is not that good option at all; it triggers when interface is ready. Not when it's online. Meaning in my case, when interface is **present**.

That's why in my case I needed that delay, cdc-wdm0 is used to initiate connection, qmi brings wan up but handshake has not yet finished and connection details are still on their way, so zerotier fails because there's no connection yet available.

> I'm trying to figure out the correct way to implement WAN availability check.

So actually, another solution would be far better. Also- if you find another version, one that really works- you don't need that _delay_ option anymore..

One could be a test for wan interface's ip address, that it is present and is not a private address; though, some connections do not provide public ip address, so that's not then maybe a good idea after all- well, if ip address then is anything but 0.0.0.0/0.... But I am confident on your imagination, so far your scripting talent has been all good- so you probably will figure out a best solution for this as well.

oskarirauta:
Usually proto_add_host_dependency **seems** to be a sort of "delay" automation, it brings up its own interface/tunnel and with that, it waits until it's complete. Meaning like: bring zt0 up with external method, and proto_add_host_dependency wait before it is up...

But you might need to consult on someone else about that. It was only partially documented.
Document available [https://openwrt.org/docs/guide-developer/network-scripting](here)

oskarirauta:
But that does place delay + connection more precisely on occasion where connection is initiated.

oskarirauta:
@Vavooon 

Please check changes I proposed. Those make this perfect. I've been using this for ages (with those changes though).

@mwarning 
This should be merged. It's superb. I also have a luci protocol support for it.

mwarning:
@oskarirauta then let's merge it after the merge conflicts have been resolved. But I do not have merge permissions.

oskarirauta:
@mwarning neither do I. Let's invite someone. It works without proposed changes as well, but those optional delay's there might help people with low bandwidth speeds, so I recommend them. But yes, for moderate and better speeds, works as is perfectly.

@1715173329
Could you join the conversation? It is about a netifd protocol, with this, zerotier(vpn) configuration is moved to /etc/config/network file. My opinion is that it clarifies setup quite a lot. I have a luci protocol for this as well. Very simple and eases up a lot to manage firewalling with zerotier interface included.

1715173329:
This intention looks good to me, but I'm not familiar with the netifd proto. Cc feckert for review.

oskarirauta:
@feckert ping

Neustradamus:
Any progress on this PR?

Vavooon:
I'd love to get it merged eventually. I can fix the PKG_RELEASE number in order to fix conflict if you're happy with anything else.

mwarning:
@Vavooon the configuration structure has changed. If you can adapt your MR, then I can put it in the 1.14.2 update.

nemesifier:
@mwarning are these changes going to be backward compatible with the old format? To allow systems like OpenWISP a bit of time before migrating to the new format?

Vavooon:
@mwarning Could you please point me out to what exactly has been changed?

mwarning:
@Vavooon /etc/config/zerotier changed from

```

config zerotier sample_config
	option enabled 0
	option config_path '/etc/zerotier'
	option copy_config_path '1'
	option port '9993'
	option local_conf '/etc/zerotier.conf'
	option secret ''
	list join '8056c2e21c000001'
	list join '1234567890123456'
```
to
```
config zerotier 'global'
	option enabled 0
	option port '9993'
	option secret ''
	option local_conf_path '/etc/zerotier.conf'
	option config_path '/etc/zerotier'
	option copy_config_path '1'

config network 'earth'
	option id '8056c2e21c000001'
	option allow_managed '1'
	option allow_global '0'
	option allow_default '0'
	option allow_dns '0'

config network
	option id '1234567890123456'
	option allow_managed '1'
	option allow_global '0'
	option allow_default '0'
	option allow_dns '0'
```
So the `join` setting was replaces by discrete setions for each network to allow setting specific  settings per network.

mwarning:
@nemesifier they are not backwards compatible. Which might be a error on my part. I will see if we can make them backwards compatible again.

Vavooon:
As I remember, there's existing Zerotier init.d service bundled with `zerotier` package which handles existing `/etc/config/zerotier` settings.
It should be able to work simultaneously with my patch.

sppmasterspp:
Can this patch be applied to zt 1.14.1 on openwrt main.

nemesifier (commented):
So if one only has the old configuration format and would upgrade to this version, would zerotier continue to work or not?

Vavooon:
> Can this patch be applied to zt 1.14.1 on openwrt main.

Yes, just updated it.



> So if one only has the old configuration format and would upgrade to this version, would zerotier continue to work or not?

Yes, both configuration formats will work simultaneously.

miyatsu:
What is the current status of this PR? Are we close to get this merged yet? Will `[luci-proto-zerotier](https://github.com/oskarirauta/packages2/tree/main/luci/luci-proto-zerotier)` be merged as well?

Neustradamus:
Have you progressed on it?

Vavooon:
I'm not quite sure I understand what is stopping from merging it. It would be nice if someone could point me to the right direction. 
The patch works quite fine for me many years.

Neustradamus:
@Vavooon: Maybe good to contact OpenWrt team members for merging...

GeorgeSapkin:
I rebased, fixed merge conflicts, and added a commit message.

PR #18931, opened 2022-07-14
libxml2: add libiconv path to pkg-config file

alinnastac:
Needed for applications that use libxml-2.0.pc Libs.private params
to link with -liconv library.

Signed-off-by: Alin Nastac <alin.nastac@gmail.com>

### DIFF
--- a/libs/libxml2/Makefile
+++ b/libs/libxml2/Makefile
@@ -100,6 +100,7 @@ CONFIGURE_ARGS += \
 	--with-xpath \
 	--with-xptr \
 	--with-zlib=$(STAGING_DIR)/usr \
+	--with-iconv$(if $(ICONV_PREFIX),="$(ICONV_PREFIX)") \
 	--without-lzma
 
 HOST_CONFIGURE_ARGS += \


libs/libxml2/Makefile
```diff
@@ -100,6 +100,7 @@ CONFIGURE_ARGS += \
 	--with-xpath \
 	--with-xptr \
 	--with-zlib=$(STAGING_DIR)/usr \
+	--with-iconv="$(ICONV_PREFIX)" \
```
mhei, line 103: In case `CONFIG_BUILD_NLS` is n (unset, or in other words "Compile with full language support" is not selected), then `$ICONV_PREFIX` is empty string and thus this line results in wrong syntax.
I think we must use:
`--with-iconv$(if $(ICONV_PREFIX),="$(ICONV_PREFIX)")`

mhei (changes requested):
Will have a deeper look later.

libs/libxml2/Makefile
alinnastac: From what I see in current include/nls.mk, $(ICONV_PREFIX) is never empty:

ifeq ($(CONFIG_BUILD_NLS),y)
	ICONV_PREFIX:=$(STAGING_DIR)/usr/lib/libiconv-full
	ICONV_FULL:=1

	INTL_PREFIX:=$(STAGING_DIR)/usr/lib/libintl-full
	INTL_FULL:=1

# iconv stub
else
	ICONV_PREFIX:=$(STAGING_DIR)/usr/lib/libiconv-stub
	ICONV_FULL:=

	INTL_PREFIX:=$(STAGING_DIR)/usr/lib/libintl-stub
	INTL_FULL:=
endif

neheb: will not be true soon.

alinnastac: Would $(STAGING_DIR)/usr/lib/libiconv.so be available when $(ICONV_PREFIX) is empty?

neheb: No.

mhei:
Now you disable iconv support completely, when `$(ICONV_PREFIX)` is empty? musl has iconv support, so why not use it? The prefix is empty in this case since its part of libc itself.

alinnastac:
> Now you disable iconv support completely, when `$(ICONV_PREFIX)` is empty? musl has iconv support, so why not use it? The prefix is empty in this case since its part of libc itself.

Technically speaking there will be no need to pass **-with-iconv** argument to configure _IF_ configure script would be fixed to append -L argument when needed (i.e. when libiconv.so is found somewhere else than `$(STAGING_DIR)/usr/lib/` or `$(STAGING_DIR)/lib/`), but that would require a little more than one additional configure parameter.

This PR tries to fix the inclusion of -liconv in libxml-2.0.pc Libs.private params without adding the path to with with `-L$(ICONV_PREFIX)`. I asked above if  `$(STAGING_DIR)/usr/lib/libiconv.so` would be available when `$(ICONV_PREFIX)` is empty and the answer was **no**. If I would do what you asked there will be 2 issues:

1. The same issue will pop up again if libiconv.so path is not found in one of the directories mentioned above
2. When libiconv.so is not available (e.g. musl not be installed), configure will likely fail with "iconv not found" (or some similar error)

mhei:
To summarize my understanding - not fully tested:

| Used libc | `CONFIG_BUILD_NLS`   | `$(ICONV_PREFIX)` | `$(STAGING_DIR)/usr/lib/libiconv.so`available | Current behavior of libxml2 | If your PR was merged |
| --------- | -------------------- | ----------------  | --------------------------------------------- | ---------------- | -------------------- |
| musl      | `CONFIG_BUILD_NLS=n` | empty             | no  | iconv support enabled :heavy_check_mark:, Libs.private is empty :heavy_check_mark: (not needed, musl's internal is used) | iconv support **disabled** :x: |
| musl      | `CONFIG_BUILD_NLS=y` | is set            | yes | iconv support enabled :heavy_check_mark:, Libs.private is **empty** :x: (but would be required since libiconv is used as you suggests) | iconv support enabled :heavy_check_mark:, Libs.private is **non-empty**, pointing to libiconv-full :heavy_check_mark: |
| glibc     | `CONFIG_BUILD_NLS=n` | empty             | no  | iconv support enabled :heavy_check_mark:, Libs.private is **empty** :question: (I don't know whether it is needed here, I guess not) | iconv support **disabled** :x: |
| glibc     | `CONFIG_BUILD_NLS=y` | is set            | yes | iconv support enabled :heavy_check_mark:, Libs.private is **empty** :x: (but would be required since libiconv is used as you suggests) | iconv support enabled :heavy_check_mark:, Libs.private is **non-empty**, pointing to libiconv-full :heavy_check_mark: |

My point is just, that there is a behavior change with this PR as it disables iconv support for the `$(ICONV_PREFIX)` empty cases, right?
We do not have a reason to do so.

> ... no need to pass -with-iconv argument to configure IF configure script would be fixed
> ... but that would require a little more than one additional configure parameter.

Yes, I share your opinion, but it seems that this hard way is required.
Don't get me wrong, but in my opinion this is an upstream issue, not an OpenWrt one.
Do you involved upstream already? We really should bring this to upstreams attention.

I don't see where my proposal fails, but maybe I missed something?

alinnastac:
Hold on, are you saying that all the other libraries that install libiconv.so put it in `$(STAGING_DIR)/usr/lib`? If that's true then your variant will fix everything, as `-L$(STAGING_DIR)/usr/lib` is not needed (toolchain will scan this directory by default).

I asked
> Would `$(STAGING_DIR)/usr/lib/libiconv.so `be available when `$(ICONV_PREFIX)` is empty?

to which Rosen replied **no**. If libiconv.so could only be found in this directory when $(ICONV_PREFIX) is empty, I have no problems adopting your variant.


> Do you involved upstream already? We really should bring this to upstreams attention.

Sorry, I'm not involved with upstream.

mhei:
musl has *included* iconv support, it is included in `/lib/libc.so`. It does not install a dedicated `libiconv.so` file in any directory. Since your applications/libs (must) already link to this libc, you get iconv "for free" - without any external library. Thus there should no specific search path required.
glibc is similar, here the file is `/lib/libc.so.6`.
Only if you really want to link against the full libiconv external library, then this library brings dedicated `libiconv.so` file which is installed in `/usr/lib/libiconv.so` on the target and in staging dir under `/usr/lib/libiconv-full/lib` - which is `$(ICONV_PREFIX)`.

mhei (approved):

PR #18939, opened 2022-07-14
sudo: don't build with MIPS16

neheb:
GCC12 doesn't implement some security flags used by sudo.

Maintainer: @commodo 
Compile tested: mipsel


commodo:
thanks :)

PR #18945, opened 2022-07-15
libuwsc: fix compiltation with wolfSSL.

cotequeiroz:
Maintainer: @zhaojh329
Compile tested: mediatek/mt7622, aarch64_cortex-a53, master
Run tested: none

Description:
wolfssl/options.h needs to be included before the other wolfssl headers
to enable OpenSSL API required to build the package.

---

Upstream has already handled this a while ago, but there's not been a new release since then.
Makefile has `PKG_RELEASE:=$(AUTORELEASE)`.

### DIFF
--- a/libs/libuwsc/patches/100-cmake-fix-wolfssl-detection.patch
+++ b/libs/libuwsc/patches/100-cmake-fix-wolfssl-detection.patch
@@ -1,0 +1,10 @@
+--- a/src/ssl.c
++++ b/src/ssl.c
+@@ -54,6 +54,7 @@ struct uwsc_ssl_ctx {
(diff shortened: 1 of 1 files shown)


neheb:
Backported to 21.02.

neheb:
options.h has been a thorn in wolfssl packages it seems. I wonder if transmission even works anymore...

cotequeiroz:
Curl now fails to detect wolfSSL because of it.  Wifidog is also failing.  I wonder if there is a reason, or use case for upstream not including `<options.h>` in `ssl.h`, for example, or any of the API headers.

PR #18948, opened 2022-07-16
slang2: try to fix compilation without BUILD_NLS

neheb:
Using --with-iconv like this causes -liconv to get added, which is not
desirable.

Maintainer: @jefferyto

### DIFF
--- a/libs/slang2/Makefile
+++ b/libs/slang2/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=slang
 PKG_VERSION:=2.3.2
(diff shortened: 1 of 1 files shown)


Ansuel:
@neheb was thinking... why keep incrementing instead of using autorelease macro? or doing that would mess up the numbering?
(sorry for OT)

neheb:
No real reason. I only switch to AUTORELEASE on version bump.

jefferyto (approved):
Thanks!

jefferyto:
Should this be backported for release branches?

neheb:
No. This PR is related to the recent libiconv-stub removal.

PR #18953, opened 2022-07-16
mpd-full: remove leftover guard on ffmpeg

pacien:
mpd-full: remove partial BUILD_PATENTED on ffmpeg

    This removes the guard "CONFIG_BUILD_PATENTED" for the ffmpeg build
    feature for mpd-full. ffmpeg was already unconditionally registered as a
    dependency (at line 51).

    The ffmpeg package already handles whether or not to include patented
    codecs internally using the same config flag, so guarding its use as a
    dependency here does not appear necessary anymore.

    (For reference: guard "BUILD_PATENTED" was introduced in 44eb12f for
    ffmpeg, then partially removed by 326c973).

---

CC: @neheb 

This patch should probably also be back-ported to openwrt-22.03 as well.


neheb:
Yeah...no.

It's fine as is. You're comparing old commits to modern day. Keep in mind PKG_BUILD_DEPENDS still has CONFIG_BUILD_PATENTED.

pacien:
The version currently on master (63ba2cc) has at line 51 and 183:

https://github.com/openwrt/packages/blob/63ba2ccbdaa27e128cfcbcc5d85af547328ad92d/sound/mpd/Makefile#L47-L54

and

https://github.com/openwrt/packages/blob/63ba2ccbdaa27e128cfcbcc5d85af547328ad92d/sound/mpd/Makefile#L177-L184

It looks to me that `ffmpeg` is always registered as a dependency while the build flag is conditional.

I don't see `PKG_BUILD_DEPENDS` either. Did I miss something?

neheb:
https://github.com/openwrt/packages/blob/master/sound/mpd/Makefile#L21

dhewg:
To me the patch looks good. For mpd-full the inconsistency is:
`DEPENDS+=+libffmpeg` vs `-Dffmpeg=$(if $(CONFIG_BUILD_PATENTED),en,dis)abled`
It either should be
`DEPENDS+=+BUILD_PATENTED:libffmpeg` or `-Dffmpeg=on`
I'd say the latter since the ffmpeg package itself handles CONFIG_BUILD_PATENTED to toggle various codecs?

Either way, PKG_CONFIG_DEPENDS is missing CONFIG_AUDIO_SUPPORT too ;)

neheb:
The whole point of adding BUILD_PATENTED was that ffmpeg contains patented functionality.

BUT, fedora seems to have ffmpeg included now with all patented codecs disabled. One thing I've been meaning to do is go through https://src.fedoraproject.org/rpms/ffmpeg/blob/rawhide/f/enable_decoders and https://src.fedoraproject.org/rpms/ffmpeg/blob/rawhide/f/enable_encoders to see where the openwrt package deviates and fix it. I haven't gotten around to doing so. They also seem to have some useful patches that could probably be ported here.

dhewg:
And that's all fine, but this patch is about mpd-full and if that should use ffmpeg.
Right now it's not clear, because it always depends on +libffmpeg no matter what, but only builds mpd with ffmpeg support it if CONFIG_BUILD_PATENTED is selected?

pacien:
The ffmpeg package already seems to contain lists of patented codecs which are excluded when building without `CONFIG_BUILD_PATENTED`, so I think it's reasonable not to have it guarded here again anymore.

I updated the commit message, which should now be clearer.
I also added the missing `CONFIG_AUDIO_SUPPORT`.

Issue #18681, opened 2022-06-04
fail2ban: 22.03 - ERROR cannot import name 'Mapping' from 'collections'


diepeterpan:
Maintainer: @erdoukki   ? ? ?
Environment: (mediatek, mt7622, OpenWrt 22.03.0-rc3 r19378-9f415792e1)

Description:

fail2ban-client start 
2022-06-04 15:58:42,964 fail2ban                [8967]: ERROR   cannot import name 'Mapping' from 'collections' (/usr/lib/python3.10/collections/__init__.pyc)

PR #18781, opened 2022-06-20
fail2ban: fix patch for python 3.10

paper42:
Maintainer: @erdoukki
Compile tested: x86_64, OpenWrt master
Compile tested: x86_64, docker, OpenWrt master

Description: 89d5d2e091c8da10a1c963158a6de92b05079116 only patched importing MutableMapping from collections, but importing Mapping has to be patched too

Closes: #18681

This should also be backported to 22.03

cc @ryzhovau

PR #18964, opened 2022-07-18
fail2ban: bpo for python3.10

erdoukki:
Maintainer: @erdoukki
Compile untested

May replace https://github.com/openwrt/packages/pull/18781

May fix: https://github.com/openwrt/packages/issues/18681

cc @ryzhovau
cc @paper42

### DIFF
--- a/net/fail2ban/patches/011-python3.10_compat_amend.patch
+++ b/net/fail2ban/patches/011-python3.10_compat_amend.patch
@@ -1,0 +1,23 @@
+From 42dee38ad2ac5c3f23bdf297d824022923270dd9 Mon Sep 17 00:00:00 2001
+From: "Sergey G. Brester" <serg.brester@sebres.de>
+Date: Mon, 8 Feb 2021 17:25:45 +0100
+Subject: [PATCH] amend for `Mapping`
+
+---
+ fail2ban/server/actions.py | 5 ++++-
+ 1 file changed, 4 insertions(+), 1 deletion(-)
+
+--- a/fail2ban/server/actions.py
++++ b/fail2ban/server/actions.py
+@@ -28,7 +28,10 @@ import logging
+ import os
+ import sys
+ import time
+-from collections import Mapping
++try:
++	from collections.abc import Mapping
++except ImportError:
++	from collections import Mapping
+ try:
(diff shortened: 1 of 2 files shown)


BKPepe:
Why are there now two PRs for the same thing? Both look identical and in my point of view, I prefer one patch instead of three patches, though. Also, your PR is totally untested instead of https://github.com/openwrt/packages/pull/18781

- In your commit message, you don't follow what is here:
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

and by each force push you are spamming this: https://github.com/openwrt/packages/issues/18681

- Both commit messages (yours and in the referenced PR), there is missing proper ``Fixes`` as this was wrong in the previous PR, though, but at least from @paper42 there is reference for the commit, which was only partial.

erdoukki:
it was a last try to help.
you said differently in the past for patches !
In french we say "allez vous faire foutre !"

I give you 2 hours of my precious time.
I will never more and already regret !

paper42:
I don't care much which approach will be merged, both should work fine and both do the same changes, so it functionally shouldn't matter. I slightly prefer one patch for fixing one issue even if it contains multiple commits because I personally find it cleaner, but as I said, both are ok from my side.

It would be nice if you told me to update my PR instead of opening your own though, this way we are just spamming everyone who follows this repository with extra notifications for no reason.

BKPepe:
@erdoukki I don't want anything new from you. You don't need to reinvent the wheel again. Please check this out:
- [Submitting patches (OpenWrt Wiki)](https://openwrt.org/submitting-patches) - there are all formal things explained.
- [CONTRIBUTING.md of this repository](https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md) says:
> Have been tested to compile with the correct includes and dependencies. Please also test with "Compile with full language support" found under "General Build Settings" set if language support is relevant to your package.

Yes, usually, patches should be done in separate files, but these patches are related to each other, thus it should be done in patch series in one file or all changes in one files as it was done by @paper42 . Seeing this in big picture, I see that this small package has already 4 patches and you want to introduce another 2. Also, your commit message is somehow cryptic, tbh.

PR #18977, opened 2022-07-18
wifidog: fix compilation with more recent wolfssl

neheb:
Needs an extra header. pthread.h is also no longer implicitly included.

Maintainer: @mhaas 
Compile tested: mipsel


neheb:
ping @cotequeiroz

PR #18981, opened 2022-07-19
python3: update patches for libressl to fix hashlib build error

dhewg:
Based upon patches from OpenBSD[0] and Gentoo[1].

Switched to the LibreSSL provided LIBRESSL_VERSION_NUMBER macro, which
makes OPENWRT_HOST_BUILD redundant.

[0] https://github.com/openbsd/ports/tree/master/lang/python/3.10/patches
[1] https://github.com/gentoo/libressl/tree/master/dev-lang/python/files

Signed-off-by: Andre Heider <a.heider@gmail.com>

Maintainer: @jefferyto 
Compile tested: ipq40xx/master
Run tested: no

This fixes the build error reported in https://github.com/openwrt/openwrt/pull/10299#issuecomment-1188755924
Just compile time tested.

### DIFF
--- a/lang/python/python3/Makefile
+++ b/lang/python/python3/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 include ../python3-version.mk
 
 PKG_NAME:=python3
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_VERSION:=$(PYTHON3_VERSION).$(PYTHON3_VERSION_MICRO)
 
 PKG_SOURCE:=Python-$(PKG_VERSION).tar.xz
@@ -326,10 +326,6 @@ HOST_CONFIGURE_VARS += \
 HOST_MAKE_VARS += \
 	USE_PYTHON_CONFIG_PY=1
 endif
-
-# FIXME: remove when removing patch '028-host-python-support-ssl-with-libressl.patch'
-HOST_CFLAGS += \
-	-DOPENWRT_HOST_BUILD
 
 HOST_CONFIGURE_ARGS+= \
 	--enable-optimizations \
--- a/lang/python/python3/patches/026-openssl-feature-flags.patch
+++ b/lang/python/python3/patches/026-openssl-feature-flags.patch
@@ -1,12 +1,47 @@
 --- a/Modules/_hashopenssl.c
 +++ b/Modules/_hashopenssl.c
-@@ -48,7 +48,9 @@
+@@ -45,10 +45,12 @@
+ 
+ #define MUNCH_SIZE INT_MAX
+ 
++#if !defined(LIBRESSL_VERSION_NUMBER)
  #define PY_OPENSSL_HAS_SCRYPT 1
  #define PY_OPENSSL_HAS_SHA3 1
  #define PY_OPENSSL_HAS_SHAKE 1
-+#ifndef OPENSSL_NO_BLAKE2
  #define PY_OPENSSL_HAS_BLAKE2 1
 +#endif
  
  #if OPENSSL_VERSION_NUMBER >= 0x30000000L
  #define PY_EVP_MD EVP_MD
+@@ -119,6 +121,7 @@ static const py_hashentry_t py_hashes[]
+     PY_HASH_ENTRY(Py_hash_sha256, "SHA256", SN_sha256, NID_sha256),
+     PY_HASH_ENTRY(Py_hash_sha384, "SHA384", SN_sha384, NID_sha384),
+     PY_HASH_ENTRY(Py_hash_sha512, "SHA512", SN_sha512, NID_sha512),
++#if !defined(LIBRESSL_VERSION_NUMBER)
+     /* truncated sha2 */
+     PY_HASH_ENTRY(Py_hash_sha512_224, "SHA512_224", SN_sha512_224, NID_sha512_224),
+     PY_HASH_ENTRY(Py_hash_sha512_256, "SHA512_256", SN_sha512_256, NID_sha512_256),
+@@ -133,6 +136,7 @@ static const py_hashentry_t py_hashes[]
+     /* blake2 digest */
+     PY_HASH_ENTRY(Py_hash_blake2s, "blake2s256", SN_blake2s256, NID_blake2s256),
+     PY_HASH_ENTRY(Py_hash_blake2b, "blake2b512", SN_blake2b512, NID_blake2b512),
++#endif
+     PY_HASH_ENTRY(NULL, NULL, NULL, 0),
+ };
+ 
+@@ -873,11 +877,15 @@ py_evp_fromname(PyObject *module, const
+         goto exit;
+     }
+ 
++#if defined(LIBRESSL_VERSION_NUMBER)
++    type = get_hashlib_state(module)->EVPtype;
++#else
+     if ((EVP_MD_flags(digest) & EVP_MD_FLAG_XOF) == EVP_MD_FLAG_XOF) {
+         type = get_hashlib_state(module)->EVPXOFtype;
+     } else {
+         type = get_hashlib_state(module)->EVPtype;
+     }
++#endif
+ 
+     self = newEVPobject(type);
+     if (self == NULL) {
--- a/lang/python/python3/patches/028-host-python-support-ssl-with-libressl.patch
+++ b/lang/python/python3/patches/028-host-python-support-ssl-with-libressl.patch
@@ -1,117 +1,27 @@
 --- a/Modules/_ssl.c
 +++ b/Modules/_ssl.c
-@@ -67,7 +67,18 @@
+@@ -67,6 +67,12 @@
  #  error "OPENSSL_THREADS is not defined, Python requires thread-safe OpenSSL"
  #endif
  
-+#ifdef OPENWRT_HOST_BUILD
-+#undef TLS1_3_VERSION
-+static STACK_OF(X509) *SSL_get0_verified_chain(const SSL *s)
-+{
-+    return NULL;
-+}
- 
++#ifdef LIBRESSL_VERSION_NUMBER
 +static int SSL_CTX_get_security_level(const SSL_CTX *ctx)
 +{
-+    return 0;
++    return 1;
 +}
 +#endif
  
+ 
  struct py_ssl_error_code {
-     const char *mnemonic;
-@@ -291,8 +302,10 @@ typedef struct {
-     int post_handshake_auth;
- #endif
-     PyObject *msg_cb;
-+#ifndef OPENWRT_HOST_BUILD
-     PyObject *keylog_filename;
-     BIO *keylog_bio;
-+#endif
-     /* Cached module state, also used in SSLSocket and SSLSession code. */
-     _sslmodulestate *state;
- } PySSLContext;
-@@ -2358,8 +2371,14 @@ _ssl__SSLSocket_write_impl(PySSLSocket *
- 
-     do {
-         PySSL_BEGIN_ALLOW_THREADS
-+#ifdef OPENWRT_HOST_BUILD
-+        retval = SSL_write(self->ssl, b->buf, (int)b->len);
-+        if (retval > 0) count = retval;
-+        err = _PySSL_errno(retval <= 0, self->ssl, retval);
+@@ -169,7 +175,11 @@ extern const SSL_METHOD *TLSv1_2_method(
+  * Based on Hynek's excellent blog post (update 2021-02-11)
+  * https://hynek.me/articles/hardening-your-web-servers-ssl-ciphers/
+  */
++#ifndef LIBRESSL_VERSION_NUMBER
+   #define PY_SSL_DEFAULT_CIPHER_STRING "@SECLEVEL=2:ECDH+AESGCM:ECDH+CHACHA20:ECDH+AES:DHE+AES:!aNULL:!eNULL:!aDSS:!SHA1:!AESCCM"
 +#else
-         retval = SSL_write_ex(self->ssl, b->buf, (size_t)b->len, &count);
-         err = _PySSL_errno(retval == 0, self->ssl, retval);
++  #define PY_SSL_DEFAULT_CIPHER_STRING "ECDH+AESGCM:ECDH+CHACHA20:ECDH+AES:DHE+AES:!aNULL:!eNULL:!aDSS:!SHA1:!AESCCM"
 +#endif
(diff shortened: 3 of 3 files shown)


dhewg:
The patches for libressl unfortunately grew.

With https://github.com/openwrt/openwrt/pull/10305 they'll shrink though: [0001-python3-update-patches-for-libressl-v3.5.patch.txt](https://github.com/openwrt/packages/files/9141414/0001-python3-update-patches-for-libressl-v3.5.patch.txt).

Looking at the current OpenBSD patches, `028-host-python-support-ssl-with-libressl.patch` will be redundant with a future LibreSSL release.

jefferyto:
This looks okay to me though I am not familiar with the existing LibreSSL patches. @commodo not sure if you want to take a look.

mpratt14:
should mention your host machine in commit message

I tested on armv7l (kernel is 64-bit, userspace is 32-bit)
Linux 5.15.32-v8+ SMP PREEMPT Thu Mar 31 19:40:39 BST 2022 aarch64 GNU/Linux

Might need to look into whether it is correctly compiling for 32-bit, but I think it is...

it does result in improvements: with these changes to the patches, both modules now compile, but fail to import

```
*** WARNING: renaming "_ssl" since importing it failed: /media/SDStorage/builds/github-openwrt/openwrt/build_dir/hostpkg/Python-3.10.4/build/lib.linux-aarch64-3.10/_ssl.cpython-310.so: unexpected reloc type 0x03
*** WARNING: renaming "_hashlib" since importing it failed: /media/SDStorage/builds/github-openwrt/openwrt/build_dir/hostpkg/Python-3.10.4/build/lib.linux-aarch64-3.10/_hashlib.cpython-310.so: unexpected reloc type 0x03

Python build finished successfully!
The necessary bits to build these optional modules were not found:
_dbm                 _gdbm                _tkinter
To find the necessary bits, look in setup.py in detect_modules() for the module's name.


The following modules found by detect_modules() in setup.py, have been
built by the Makefile instead, as configured by the Setup files:
_abc                 pwd                  time
zlib


Following modules built successfully but were removed because they could not be imported:
_hashlib             _ssl


Could not build the ssl module!
Python requires a OpenSSL 1.1.1 or newer
Custom linker flags may require --with-openssl-rpath=auto

```

mpratt14:
@jefferyto is it possible that python requires that the library be a shared object instead of static?

mpratt14:
I should also mention I am testing on a branch from master where libressl is still version 3.4.3

mpratt14:
libressl at version 3.5.3 makes no difference

libressl with shared libraries enabled causes more problems

```
/media/SDStorage/builds/github-openwrt/openwrt/build_dir/hostpkg/Python-3.10.4/Modules/_ssl.c:72:24: error: static declaration of 'SSL_get0_verified_chain' follows non-static declaration
   72 | static STACK_OF(X509) *SSL_get0_verified_chain(const SSL *s)
      |                        ^~~~~~~~~~~~~~~~~~~~~~~
In file included from /media/SDStorage/builds/github-openwrt/openwrt/build_dir/hostpkg/Python-3.10.4/Modules/_ssl.c:60:
/media/SDStorage/builds/github-openwrt/openwrt/staging_dir/host/include/openssl/ssl.h:515:17: note: previous declaration of 'SSL_get0_verified_chain' was here
  515 | STACK_OF(X509) *SSL_get0_verified_chain(const SSL *s);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~

```

```
/media/SDStorage/builds/github-openwrt/openwrt/build_dir/hostpkg/Python-3.10.4/Modules/_hashopenssl.c:55:1: error: static declaration of 'FIPS_mode' follows non-static declaration
   55 | FIPS_mode(void)
      | ^~~~~~~~~
In file included from /media/SDStorage/builds/github-openwrt/openwrt/staging_dir/host/include/openssl/bio.h:69,
                 from /media/SDStorage/builds/github-openwrt/openwrt/staging_dir/host/include/openssl/evp.h:67,
                 from /media/SDStorage/builds/github-openwrt/openwrt/build_dir/hostpkg/Python-3.10.4/Modules/_hashopenssl.c:33:
/media/SDStorage/builds/github-openwrt/openwrt/staging_dir/host/include/openssl/crypto.h:499:5: note: previous declaration of 'FIPS_mode' was here
  499 | int FIPS_mode(void);
      |     ^~~~~~~~~

```

```
Python build finished successfully!
The necessary bits to build these optional modules were not found:
_dbm                 _gdbm                _tkinter
To find the necessary bits, look in setup.py in detect_modules() for the module's name.


The following modules found by detect_modules() in setup.py, have been
built by the Makefile instead, as configured by the Setup files:
_abc                 pwd                  time
zlib


Failed to build these modules:
_hashlib             _ssl


Could not build the ssl module!
Python requires a OpenSSL 1.1.1 or newer
Custom linker flags may require --with-openssl-rpath=auto

```

dhewg:
Thanks for the test! For v3.5 you need the additional patch from comment 1.
I'm not sure why that reloc error occurs, sounds like some CFLAGS/LDFLAGS mismatch for your arm64 host?
And that doesn't occur when you use your host openssl?

dhewg:
From a quick search it sounds like libressl needs to be build with PIC because it's only build static.
But its Makefile already contains `HOST_CFLAGS += $(HOST_FPIC)`.
Can you verify if it's actually compiling with those flags?

dhewg:
Untested random idea:
```
diff --git a/tools/libressl/Makefile b/tools/libressl/Makefile
index 0679961d17..e3878ebcd1 100644
--- a/tools/libressl/Makefile
+++ b/tools/libressl/Makefile
@@ -23,9 +23,8 @@ HOST_BUILD_PARALLEL:=1
 
-HOSTCC := $(HOSTCC_NOCACHE)
+HOSTCC := $(HOSTCC_NOCACHE) $(HOST_FPIC)
 HOST_CONFIGURE_ARGS += --enable-static --disable-shared --disable-tests
-HOST_CFLAGS += $(HOST_FPIC)
 
 ifeq ($(GNU_HOST_NAME),x86_64-linux-gnux32)
 HOST_CONFIGURE_ARGS += --disable-asm
```

commodo:
> This looks okay to me though I am not familiar with the existing LibreSSL patches. @commodo not sure if you want to take a look.

The only note is: Host Python3 Pip should be able to install packages.

To properly test this, I would also build any package that has a `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS`.
Here's a quick list:

```
lang/python/bcrypt/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=cffi  # cffi>=1.1
lang/python/numpy/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=Cython
lang/python/python-apipkg/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-augeas/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=cffi  # cffi>=1.0.0
lang/python/python-bidict/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-cryptography/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=cffi  # cffi>=1.12
lang/python/python-dateutil/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-execnet/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-iniconfig/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-jsonschema/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-pluggy/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-psutil/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-py/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-pycparser/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=ply  # ply==3.10
lang/python/python-pynacl/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=cffi  # cffi>=1.4.1
lang/python/python-pytest-forked/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-pytest-xdist/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-pytest/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm
lang/python/python-yaml/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=Cython
lang/python/python-zipp/Makefile:HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=setuptools-scm toml  # setuptools_scm[toml] >= 3.4.1
```

If any of these packages build fine, then the SSL patches are fine.

I don't know about needing to support `_hashlib` on Host-Python3.
IIRC, _hashlib was never needed on the host-python3 side.

commodo:
> From a quick search it sounds like libressl needs to be build with PIC because it's only build static.

Yes. LibreSSL (well, most libraries in `tools/`) get built as static libraries, so as to avoid some ABI issues that show up with shared libraries. This is more evident in the OpenWrt SDK, when you try to use the SDK on different machines and get random error reports.
It is intentional that all libs in `tools/` have the shared-libs disabled (and only static enabled).

Python3 can use (and uses) the static libressl.

commodo:
General question: is `_hashlib` needed for Host-Python3?

I've never seen it used/needed for OpenWrt's host build. It's only the _ssl module that's needed. 
We can fix `_hashlib` as well, but I wouldn't make it a priority ; I never made it a priority anyway.
There's a few host-Python3 modules that randomly fail/don-t-fail to build depending on what's installed on your host (like ncurses)

The host-Python3 is mostly needed to build target-Python3 + python packages on the target, so whatever minimal form we can get to build these, is fine (well, that's my thought anyway).

commodo:
> @jefferyto is it possible that python requires that the library be a shared object instead of static?

it can build with both; but on OpenWrt, we should build with the static library; this is a requirement for building a solid OpenWrt SDK; also refers to my comment earlier   https://github.com/openwrt/packages/pull/18981#issuecomment-1189908917

commodo:
i'll also refer to an old PR I did a few months ago:  https://github.com/openwrt/openwrt/pull/4749

as you can probably guess, i am amused now;  because, now i look (or feel) like a prophet 

well, anyway, my current (Github) avatar is ` ¯\_(ツ)_/¯ `

dhewg:
> General question: is `_hashlib` needed for Host-Python3?

Maybe not ;) But:
We assumed that the build error we're fixing here fixes @mpratt14 's issue to import the host ssl module. Turns out it doesn't. Whatever that issue is about in the end, there's the fact that the python/libressl patches requires maintenance. This gets that patch to almost match the corresponding ones in OpenBSD/Gentoo, so it's more like a shared burden then.
I.e. it's worth merging this, independent of if hashlib is used in the end or not.

mpratt14:
> doesn't occur when you use your host openssl?

correct, as far as I can tell



> its Makefile already contains `HOST_CFLAGS += $(HOST_FPIC)`.
> Can you verify if it's actually compiling with those flags?

also yes

> I don't know about needing to support `_hashlib`

not needed but...it would be nice if everything "just worked" right? 😄 

both modules do depend on one of the libraries, which would be the sign of some issue with them, whether minor or not

dhewg:
The libressl v3.5 bump has been merged, so I squashed the patch from comment 1 in here.

dhewg:
> > doesn't occur when you use your host openssl?
> 
> correct, as far as I can tell

Okay, and does it work if you build that static only? Usually via `--enable-static --disable-shared`?

Another idea, look at the libressl Makefile:
```
ifeq ($(GNU_HOST_NAME),x86_64-linux-gnux32)
HOST_CONFIGURE_ARGS += --disable-asm
endif
```

Sound like you have the arm equivalent of that, does that switch fix it?

In any case, please attach out.log of `make V=cs tools/libressl/{clean,compile} 2>&1|tee out.log`.

dhewg:
The CI action logs at https://github.com/openwrt/packages/actions/runs/2697959763 for aarch64_cortex-a53, arm_cortex-a15_neon-vfpv4 and arm_cortex-a9_vfpv3-d16 all don't show that error.
The modules have been built and there isn't any `Following modules built successfully but were removed because they could not be imported` in `compile.log` either.
So I guess it indeed is related to your rather uncommon 64bit kernel/32bit userland setup.

dhewg:
> > I.e. it's worth merging this, independent of if hashlib is used in the end or not.
> 
> if you can validate you're not breaking any python package builds, i agree; see comment [#18981 (comment)](https://github.com/openwrt/packages/pull/18981#issuecomment-1189906239)

Thanks for that list, random choice looks good on my amd64 host: [python-iniconfig.log](https://github.com/openwrt/packages/files/9152140/python-iniconfig.log)
Is that sufficient as validation?

mpratt14:
> Sound like you have the arm equivalent of that, does that switch fix it?

--disable-asm did the trick, it works

I don't think any of my full logs are useful, unless we want to show libressl maintainers (they probably know about this but don't document it well)

going to make a PR for that, but not getting rid of the other one, I believe openssl should be optional anyway

mpratt14:
> `¯\_(ツ)_/¯`

I agree this mindset is sometimes the big brain move 🤣

mpratt14:
> The CI action logs at https://github.com/openwrt/packages/actions/runs/2697959763 for aarch64_cortex-a53, arm_cortex-a15_neon-vfpv4 and arm_cortex-a9_vfpv3-d16 all don't show that error.

keep in mind we are talking about ARM host, I think those builds are regarding ARM targets, but not 100% certain

mpratt14:
a small project of interest, although I don't know how much we are interested in having patches of assembly code...

https://github.com/q66/libressl-portable-asm

dhewg:
> > Is that sufficient as validation?
> 
> yes (from my side)

Nice, and thanks for all your feedback!

> > Sound like you have the arm equivalent of that, does that switch fix it?
> 
> --disable-asm did the trick, it works

Nice, again!

Issue #18980, opened 2022-07-19
`mutt`: Fails to build with missing dependencies for `libcrypto.so.1.1`, `iblmdb.so`, `libsasl2.so.3` and `libssl.so.1.1` in 22.03.


dreirund:
### Maintainer:

@philenotfound

### Environment:

* Building for Xunlog OrangePi Plus (sun8i):  
  `grep ^CONFIG_TARGET .config`:
  ```
  CONFIG_TARGET_sunxi=y
  CONFIG_TARGET_sunxi_cortexa7=y
  CONFIG_TARGET_sunxi_cortexa7_DEVICE_xunlong_orangepi-plus=y
  CONFIG_TARGET_BOARD="sunxi"
  CONFIG_TARGET_SUBTARGET="cortexa7"
  CONFIG_TARGET_PROFILE="DEVICE_xunlong_orangepi-plus"
  CONFIG_TARGET_ARCH_PACKAGES="arm_cortex-a7_neon-vfpv4"
  [...]
  ```
* OpenWrt version:
  - Git branch `openwrt-22.03` from `git://git.openwrt.org/openwrt/openwrt.git`,
  - initially checked out on 2022-07-05,
  - latest `git pull` and `./scripts/feeds update -a && ./scripts/feeds update -a` on 2022-07-14, ca. 12:00 UTC  
    *(I give up on posting latest commit date and hash, since `git log` gives me some old dates, and I don't know how I can make `git log` printing out the last operation. I don't know much about `git`.)*
* Building on GNU/Linux on x86_64-architecture (Distribution: [Artix Linux](http://artixlinux.org)).
* OpenWrt configuration: [&rarr;&nbsp;`config.txt`](https://github.com/openwrt/packages/files/9140489/config.mutt-builderror.txt).

* `grep -i mutt .config`:  
  ```
  CONFIG_PACKAGE_mutt=m
  CONFIG_MUTT_POP=y
  CONFIG_MUTT_IMAP=y
  CONFIG_MUTT_SMTP=y
  CONFIG_MUTT_HCACHE=y
  CONFIG_MUTT_SASL=y
  # CONFIG_MUTT_GNUTLS is not set
  CONFIG_MUTT_OPENSSL=y
  ```
* `grep -i lmdb .config`:  
  ```
  CONFIG_PACKAGE_lmdb=m
  # CONFIG_PACKAGE_lmdb-test is not set
  # CONFIG_PACKAGE_lmdb-utils is not set
  ```

### Description:

[...]

PR #18986, opened 2022-07-20
mutt: add PKG_CONFIG_DEPENDS

neheb:
Should fix missing dependencies when manupulating options.

Maintainer: @philenotfound 

Fixes: https://github.com/openwrt/packages/issues/18980


felixbecker2:
I can confirm that this changes solve issue #18980 for me.

dreirund:
I have to add that there still is one problem:

When `CONFIG_MUTT_HCACHE=y`, then build still fails for me with

```
Package mutt is missing dependencies for the following libraries:
liblmdb.so
```

When `# CONFIG_MUTT_HCACHE is not set` build succeeds.

PR #18987, opened 2022-07-20
[RFC] mwan3: add online_metric for ipsec failover

feckert:
Maintainer: me / @aaronjg 

Description:

The mwan3  with IPsec does not work in a failover scenario as expected.
Because the default table is used for an lookup wich traffic should be encrypt.
The traffic for encryption is sent out on the interface with the lowest metric.
No matter which interface is currently seen as connected by the mwan3.

In order for this that mwan3 could work with IPsec, an additional metric is set that indicates which interface is currently connected. And the interface with the lowest online_metric is used as the IPsec interface.
If the interface is not considered connected by the mwan3, then the `online_metric` for this interface is not set.

### DIFF
--- a/net/mwan3/files/etc/hotplug.d/iface/13-mwan3
+++ b/net/mwan3/files/etc/hotplug.d/iface/13-mwan3
@@ -1,0 +1,116 @@
+#!/bin/sh
+
+. /lib/functions.sh
+. /lib/functions/network.sh
+. /lib/mwan3/mwan3.sh
+
+initscript=/etc/init.d/mwan3
+. /lib/functions/procd.sh
+
+logging() {
+	local level="$1"
+	local message="$2"
+
+	logger -t "mwan3-hotplug[$$]" -p "daemon.${level}" "${message}"
+}
+
+[ "$ACTION" = "connected" ] || [ "$ACTION" = "disconnected" ] || exit 1
+[ -n "$INTERFACE" ] || exit 2
+
+if [ "$ACTION" = "connected" ]; then
+	[ -n "$DEVICE" ] || exit 3
+fi
+
+procd_lock
+
+config_load mwan3
+config_get enabled $INTERFACE enabled 0
+[ "$enabled" == "1" ] || {
+	exit 0
+}
+
+config_get online_metric $INTERFACE online_metric 0
+if [ "$online_metric" = 0 ]; then
+	logging notice "No online metric for interface "$INTERFACE" found"
+	exit 0
+fi
+
+mwan3_add_failover_metric() {
+	local iface="$1"
+	local device="$2"
+	local metric="$3"
+
+	local route_args ret
+
+	config_get family $iface family ipv4
+
+	if [ "$family" == "ipv4" ]; then
+		if ubus call network.interface.${iface}_4 status 1>/dev/null 2>&1; then
+			network_get_gateway route_args ${iface}_4
+		else
+			network_get_gateway route_args $iface
+		fi
+
+		if [ -n "$route_args" -a "$route_args" != "0.0.0.0" ]; then
+			route_args="via $route_args"
+		else
+			route_args=""
+		fi
+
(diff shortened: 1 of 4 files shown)


net/mwan3/files/etc/hotplug.d/iface/13-mwan3
```diff
@@ -0,0 +1,116 @@
...
+mwan3_add_failover_metric() {
+	local iface="$1"
+	local device="$2"
+	local metric="$3"
+
+	local route_args ret
+
+	config_get family $iface family ipv4
+
+	if [ "$family" == "ipv4" ]; then
```
patrakov, line 47: In POSIX sh, "==" is undefined. Please use "=".

net/mwan3/files/etc/hotplug.d/iface/13-mwan3
```diff
@@ -0,0 +1,116 @@
...
+		$IP4 route add default $route_args dev $device proto static metric $metric 1>/dev/null 2>&1
+		ret="$?"
+		if [ "$ret" -ne 0 ]; then
+			logging warn "Unable to set online metric for interface '$iface' (IPv4)"
+		else
+			logging info "Online metric for interface '$iface' set to '$metric' (IPv4)"
+		fi
+	fi
+
+	if [ "$family" == "ipv6" ]; then
```
patrakov, line 69: Same here.

net/mwan3/files/etc/hotplug.d/iface/13-mwan3
```diff
@@ -0,0 +1,116 @@
...
+}
+
+mwan3_del_failover_metric() {
+	local iface="$1"
+	local device="$2"
+	local metric="$3"
+
+	config_get family $iface family ipv4
+
+	if [ "$family" == "ipv4" ]; then
```
patrakov, line 98: And here.

net/mwan3/files/etc/hotplug.d/iface/13-mwan3
```diff
@@ -0,0 +1,116 @@
...
+		$IP4 route del default dev $device proto static metric $metric 1>/dev/null 2>&1
+	fi
+
+	if [ "$family" == "ipv6" ]; then
```
patrakov, line 102: And here, too.

patrakov (changes requested):

BKPepe:
Kind reminder.

net/mwan3/files/etc/hotplug.d/iface/13-mwan3
```diff
@@ -0,0 +1,116 @@
...
+		fi
+
+		if [ -n "$route_args" -a "$route_args" != "::" ]; then
+			route_args="via $route_args"
+		else
+			route_args=""
+		fi
+
+		$IP6 route add default $route_args dev $device proto static metric $metric 1>/dev/null 2>&1
+		if [ "$ret" -ne 0 ]; then
```
openwrt-ai, line 83: The IPv6 branch checks `$ret` but never assigns it. Unlike the IPv4 branch above (which does `ret="$?"` after `$IP4 route add`), there is no `ret="$?"` after `$IP6 route add`. For an `ipv6`-family interface the IPv4 branch never runs, so `ret` stays empty and `[ "$ret" -ne 0 ]` evaluates `[ "" -ne 0 ]`, which errors out and is treated as false — meaning failures are silently logged as success (and no warning is ever emitted for IPv6).

```suggestion
		$IP6 route add default $route_args dev $device proto static metric $metric 1>/dev/null 2>&1
		ret="$?"
		if [ "$ret" -ne 0 ]; then
```

---

openwrt-ai (commented):
Reviewed 1 new commit; found one inline issue in the new `13-mwan3` hotplug script (missing `ret="$?"` in the IPv6 branch). Note that the existing unresolved review comments about POSIX `==` vs `=` still apply to this revision.

---

Issue #18989, opened 2022-07-20
lxc: out of date and new version requires meson/ninja


graysky2:
Maintainer: @ratkaj @paper42 @robimarko @cotequeirozz @BKPepe @aparcar 
Environment: all

Description:  Latest upstream release requires meson/ninja to build.  Modifying our Makefile to do that is beyond my abilities.  Do any of you guys have some experience to do this?

graysky2:
I tried to update the Makefile, see https://github.com/graysky2/packages/commits/lxc

I am getting an error when building ([full build log](https://gist.github.com/graysky2/40e4a2ee5174c24175ae784e9925d0ad)):
```
...
[1/1527] Compiling C object src/lxc/liblxc_static.a.p/cgroups_cgfsng.c.o
FAILED: src/lxc/liblxc_static.a.p/cgroups_cgfsng.c.o 
[...]

PR #18990, opened 2022-07-20
lxc: update to 5.0.1

graysky2:
Bump to latest and update Makefile to use meson which is upstream's standard.
Deleted unneeded 010-Remove-distro-check.patch (reference to configure).

Build system: x86_64
Build-tested: bcm2711/RPi4B
Run-tested: bcm2711/RPi4B

Maintainers: @ratkaj


graysky2:
This closes issue #18989

graysky2:
My test build did not include all of the lxc tools.  I am not sure why lxc-monitord is failing.  Any suggestions are appreciated.

graysky2:
Is this something we can fix in our `Makefile` or is this something upstream?

I see the expected file, `lxc-monitord` is getting put in the the libexe path:
```
Installing src/lxc/cmd/lxc-monitord to /scratch/union/build_dir/target-aarch64_cortex-a72_musl/lxc-5.0.0/ipkg-install/usr/libexec/lxc
```

But later on, `install` is directed to another path:
```
install -m0755 /scratch/union/build_dir/target-aarch64_cortex-a72_musl/lxc-5.0.0/ipkg-install"/usr/lib/lxc"/lxc-monitord /scratch/union/build_dir/target-aarch64_cortex-a72_musl/lxc-5.0.0/ipkg-aarch64_cortex-a72/lxc-monitord"/usr/lib/lxc"/
install: cannot stat '/scratch/union/build_dir/target-aarch64_cortex-a72_musl/lxc-5.0.0/ipkg-install/usr/lib/lxc/lxc-monitord': No such file or directory
```

utils/lxc/Makefile
```diff
@@ -25,6 +25,19 @@ PKG_BUILD_PARALLEL:=1
 PKG_USE_MIPS16:=0
 
 include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/meson.mk
+
+TARGET_CFLAGS += -fPIC
+TARGET_LDFLAGS += -fPIC
```
neheb, line 31: meson has -Db_pic=true

neheb: And actually you probably mean to use -Db_pie.

utils/lxc/Makefile
```diff
@@ -25,6 +25,19 @@ PKG_BUILD_PARALLEL:=1
...
+
+MESON_ARGS += \
+	-Dwerror=false \
+	-Dinit-script=sysvinit \
+	-Dman=false \
+	-Dapparmor=false \
+	-Dselinux=false \
+	-Dseccomp=$(if $(CONFIG_LXC_SECCOMP),true,false) \
```
neheb, line 39: this should probably be added to PKG_CONFIG_DEPENDS.

neheb:
```
PKG_FIXUP:=autoreconf
PKG_INSTALL:=1
PKG_BUILD_PARALLEL:=1
```

should all be removed. the last two are default and the first is useless.

graysky2:
@ratkaj - do you approve of this?

utils/lxc/Makefile
BKPepe: I see this discussion is marked as resolved, but I dont see any change related to it?

graysky2:
Is the maintainer of this package active?

neheb:
No. Anyway, no idea why

```
TARGET_CFLAGS += -fPIC
TARGET_LDFLAGS += -fPIC
```

is insisted on. b_pic is default on.

graysky2:
@neheb - I removed the two lines:
```
TARGET_CFLAGS += -fPIC
TARGET_LDFLAGS += -fPIC
```

and added `-Db_pie=true` to `MESON_ARGS` and it test built fine for bcm27xx.  Let's see if the build tests for others pass.

graysky2:
@neheb -

https://github.com/openwrt/packages/runs/8246554550?check_suite_focus=true#step:5:4137

Seems to want to have `-fPIC` ... do you have any ideas?  Bug in upstream code?  Revert to defining `-fPIC` for now?

AreYouLoco:
Is there anything I could do to help out testing lxc 5.0? I really need support for vlans added in that version

graysky2:
@AreYouLoco - just pull this PR and build.  I have been using it since I opened the PR.

AreYouLoco:
I'd rather wait until its upstreamed. I am totally not familiar with openwrt build system

neheb:
I'll look into this failure. Sounds like an upstream bug.

eli-schwartz:
Why does an ***executable*** want -fPIC? Meson's `b_pic` option is for libraries.

neheb:
I think I found the issue. LTO bug.

neheb:
Add this patch:

```
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,6 @@ project(
     version: '5.0.1',
     license: 'LGPLv2+',
     default_options: [
-        'b_lto=true',
         'b_lto_mode=thin',
         'b_colorout=always',
         'b_asneeded=true',
```

and

```
--- a/utils/lxc/Makefile
+++ b/utils/lxc/Makefile
@@ -25,6 +25,7 @@ include $(INCLUDE_DIR)/package.mk

 MESON_ARGS += \
+       -Db_lto=$(if $(findstring mips,$(CONFIG_ARCH)),false,true) \
        -Dwerror=false \
        -Dinit-script=sysvinit \
        -Dman=false \
```

eli-schwartz:
If you're overriding the setting anyway, you don't care what its project default_options is.

neheb:
oh ok. Edited.

graysky2:
I applied @neheb's patch and also added back `-Db_pie` which I removed as a test.  Let's see how the checks go.

utils/lxc/Makefile
```diff
@@ -9,22 +9,31 @@
...
+MESON_ARGS += \
+	-Db_lto=$(if $(findstring mips,$(CONFIG_ARCH)),false,true) \
+	-Dwerror=false \
+	-Dinit-script=sysvinit \
+	-Dman=false \
+	-Dapparmor=false \
+	-Dselinux=false \
+	-Dseccomp=$(if $(CONFIG_LXC_SECCOMP),true,false) \
+	-Dexamples=false \
+	-Db_pie
```
neheb, line 36: =true. I think this is already default though.

graysky2:
That worked... trying now with the `-Db_pie=true` removed since it could be defaulted and therefore superfluous.

AreYouLoco:
So this can be merged now?

graysky2:
Pull it: `git pull origin pull/18990/head --no-edit` or put the word `.patch` in your browser's url to get a patch

AreYouLoco:
Ok thanks @graysky2 I got the patch.

AreYouLoco:
Could it be somehow backported to 21.02?

neheb:
21.02 needs meson/host as tools/meson was not around.

AreYouLoco:
@neheb could you point me into direction on how to build it for 21.02?

graysky2:
Do you have a good reason to stay with the LTS release?  Building from master has been rock solid for me for a long time now.

AreYouLoco:
Actually I am running OpenWRT fork that is based on 21.02 and I could move to developer branches based on master but I am a bit worried about stability

neheb:
@AreYouLoco 

```
make package/meson/host/compile
make package/lxc/compile
```

should work.

AreYouLoco:
I started with clean sources. Added a patch from here to update lxc to 5.0.1 and the first command finished but the second complained about "no rule to make target [...]/meson.mk" Is this the right place to ask for backport?

neheb:
Oh right. The meson.mk line needs to be changed. See https://github.com/openwrt/packages/blob/openwrt-21.02/sound/mpd/Makefile#L29

AreYouLoco:
```
ERROR: Meson version is 0.56.0 but project requires >= 0.61
```

So I guess I need to port meson somehow

graysky2:
Not sure how involved that will be (does meson have deps) but do that under your openwrt build dir not under `feeds/packages`

neheb:
meson in 21.02 lives in devel/meson so update that.

AreYouLoco:
I tried but my changes were overwritten and I got the same error. How do I do that?

PR #18867, opened 2022-07-04
emailrelay: Support compiling with MbedTLS and update to v2.4.1

stokito:
Maintainer: @fededim
Compile tested: not yet :(

The EmailRelay has MbedTLS support. Ideally it would be great to have a separate package emailrelay-mdtls but at least for now we can make a first step.
The OpenWrt will get back from WolfSSL to MbedTLS by default.
Then installing of the EmailRelay bey be so cheap so that can be installed to OpenWrt by default.

Also please check my #18536 and update EmailRealy version. In a day or two it should be released v2.3.1

PR #19008, opened 2022-07-22
curl: switch default SSL to mbedTLS

neheb:
The decision to switch the default to wolfSSL was taken because of
hostapd back from when curl was in base. Unfortunately, not only is
wolfSSL bigger but it has also been causing issues recently. There's
also no relation between hostapd and curl.

Maintainer: @stangri 

https://github.com/openwrt/openwrt/issues/10302 as an example.

### DIFF
--- a/net/curl/Config.in
+++ b/net/curl/Config.in
@@ -4,7 +4,7 @@ comment "SSL support"
 
 choice
 	prompt "Selected SSL library"
-	default LIBCURL_WOLFSSL
+	default LIBCURL_MBEDTLS
 
 	config LIBCURL_MBEDTLS
 		bool "mbed TLS"


stangri:
@neheb Rosen, I've enquired about possibility of switching to a different SSL library in the past (I'll try to find and link that discussion when I have a moment) and I believe the common thinking was that the benefits of staying with WolfSSL outweighed the disadvantage of depending on a different library in curl/libcurl.

I can honestly say that I don't know enough about OpenWrt to have a say in this, I'd very much appreciate if other core developers comment on this.

neheb:
Note that the previous used library was mbedTLS with curl. The only reason it was switched to wolfSSL was because of hostapd. Speaking of which there's work to transition that to use mbedTLS as well: https://github.com/openwrt/openwrt/issues/10303

There are also some packages (mostly because of me) that use libcurl's SSL selection to select their own libraries. For example: https://github.com/openwrt/packages/blob/master/net/uacme/Makefile#L92

That supports openssl, gnutls, mbedtls, but not wolfssl.

https://github.com/openwrt/packages/blob/master/net/transmission/Makefile#L104

That supports, openssl, mbedtls, wolfssl, but not gnutls.

Given that openssl is a big library, the only other common denominator is mbedtls.

neheb:
I just realized uacme does not have a CONFIG_LIBCURL_NOSSL case...

neheb:
Sure I'll wait a day or two. I have no desire to backport this change (although things with wolfssl are currently broken on stable branches...)

TDL3:
I have no idea what you guy are talking about, but you guys are awesome, you made a great project!

stokito:
Sorry, I'm incompetent here but please clarify.
Is it possible in opkg to specify alternatives in dependencies i.e. make the libcurl to suggest/recommend the wolfssl but to be able to use the mbedtls?
In Debian the libcurl4 just doesn't depend on any library directly. Looks like it lookups for a backend dynamically [curl-library: Enabled multiple SSL backends](https://curl.se/mail/lib-2017-08/0118.html).
So maybe it's possible to do the same. How big the libcurl becomes if compiled with openssl, wolfssl and mbedtls support? If it's too big then yes, it may be better to compile with one library only. Or to create a different packages like libcurl4-mbedtls. Still the libcurl is not part of a base packages so it's fine if it will be slightly bloated.

The idea to use the CONFIG_LIBCURL_MBEDTLS, CONFIG_LIBCURL_GNUTLS, CONFIG_LIBCURL_OPENSSL, CONFIG_LIBCURL_WOLFSSL as a reference to compile the other packages looks clever but not solid. Packages that doesn't have a direct dependency to libcurl may use other library.
Maybe instead it is possible to make a separate global config like `CONFIG_DEFAULT_TLSLIB` and use it inside of the libcurl.

Also on package level it must be overridable e.g. the lighttpd had some problems with wolfssl so users may want to change this.

Some packages like `emailrelay` may already have support of other libraries but currently it's not possible to use them. I created a PR https://github.com/openwrt/packages/pull/18867 but wondering how many others may still be changed similarly. This is an ideal place to use such `CONFIG_DEFAULT_TLSLIB` variable.

Also sometimes you just want to have a separate packages compiled with a specific library. So it would be great to have a unified Makefile style for all packages.

The change that you proposed seems dangerous because it affects not only libcurl but other packages. Also since the WolfSSL didn't removed from a base then people who installs libcurl now will have two libraries.
As far I understood the libcurl works fine with the current wolfssl so there's no particular reason to change the tls backend.

BTW the mbedtls doesn't support TLS1.3 so it may broke something potentially.

stangri:
> As far I understood the libcurl works fine with the current wolfssl so there's no particular reason to change the tls backend.

As far as I understand, curl/libcurl and dependent packages (like https-dns-proxy) are currently broken in snapshots and 22.03, hence this PR.

neheb:
@stokito as implemented, the libcurl package only supports one SSL library. Implementing multiple backends as is done with libustream-ssl is beyond the scope of this PR.

mbedTLS not supporting TLS 1.3 is a worthwhile tradeoff. There's still plenty of TLS 1.2 out there.

openips:
the main bug is wolfssl 5.4 . only change default SSL to mbedTLS in curl can not fixed the wolfssl 5.4 bug

neheb:
Right. It avoids the problem with curl. OTOH I would argue that this change is still corrent given transmission and uacme.

openips:
openssl maybe better

neheb:
not by default. size is a concern on many routers.

Neustradamus:
For 22.03: https://github.com/openwrt/packages/pull/19058.

ynezz:
> mbedTLS not supporting TLS 1.3 is a worthwhile tradeoff.

FYI there are some industrial environments, where it's no longer possible to use anything lower than TLS v1.3.

> There's still plenty of TLS 1.2 out there

There are also environments where you're not able to start connections using TLS v1.2.

ynezz:
> openssl maybe better

IMO this would be the most sensible thing to do. If you've resources for curl, then you should've few more for OpenSSL.

gstrauss:
> > mbedTLS not supporting TLS 1.3 is a worthwhile tradeoff.
> 
> FYI there are some industrial environments, where it's no longer possible to use anything lower than TLS v1.3.
> 
> > There's still plenty of TLS 1.2 out there
> 
> There are also environments where you're not able to start connections using TLS v1.2.

@ynezz  I do not doubt that such environments exist.

> > openssl maybe better
> 
> IMO this would be the most sensible thing to do. If you've resources for curl, then you should've few more for OpenSSL.

That is a flippant statement.  Just because you do not care does not mean resource use does not matter to anyone else.

While openwrt may be used on systems which do not have memory constraints, openwrt is also used on smaller, embedded systems which do have memory constraints.  I hope that openwrt developers keep this in mind and would prefer a default TLS library aimed at smaller systems, e.g. mbedTLS or wolfSSL.

Similar to hostapd which has multiple variants for use with different TLS libraries, maybe there should be multiple variants for curl offering dependencies on different TLS libraries?  If not, maybe the curl menuconfig could offer the option to choose a TLS library against which to build, similar to what I have added to the lighttpd menuconfig for the base lighttpd crypto library?

champtar:
@gstrauss you can already select TLS library using menuconfig https://github.com/openwrt/packages/blob/6164a8bab74617aab75de3984719809b99eb83c1/net/curl/Config.in#L6

neheb:
@ynezz if you feel the need to have TLS 1.3, a better option would be to update mbedtls. It needs to be done eventually anyway.

champtar:
> @champtar IIUC then the suggestion is to have {lib}curl packages available in all crypto variants, so the users could decide whether they want to install `curl-mbedtls`, `curl-openssl` or `curl-wolfssl`. I don't think, that such changes would be acceptable for backport into 22.03, although it would offer some kind of a fix/workarounds.

Must all software that depend on libcurl then have variants, or the different libcurl variants have the same ABI and we can use PROVIDES:=libcurl ?

PR #19025, opened 2022-07-22
ruby: fix build with LibreSSL over v3.5

msylgj:
Maintainer: @luizluca 
Compile tested: armv8(rockchip r4s) OpenWrt master SNAPSHOT
Run tested: armv8(rockchip r4s) OpenWrt master SNAPSHOT

Description:
LibreSSL 3.5 and later (https://github.com/openwrt/openwrt/commit/5451b03b7ceb2315445c683fe174e28bbdd49c2f) provide and need to use
PEM_write_bio_PrivateKey_traditional()

upstream commit:
https://github.com/ruby/openssl/commit/e25fb0d0d86da5a9398ebdc9216b2ea89f80fa3d
https://github.com/ruby/openssl/commit/b02815271fcc295cb8b07ef740684b88a10f2760

### DIFF
--- a/lang/ruby/patches/001-fix-build-with-libressl-3.5.patch
+++ b/lang/ruby/patches/001-fix-build-with-libressl-3.5.patch
@@ -1,0 +1,20 @@
+From e25fb0d0d86da5a9398ebdc9216b2ea89f80fa3d Mon Sep 17 00:00:00 2001
+From: Jeremy Evans <code@jeremyevans.net>
+Date: Fri, 25 Mar 2022 13:11:31 -0700
+Subject: [PATCH] Fix build with LibreSSL 3.5
+
+---
+ ext/openssl/ossl_pkey.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+--- a/ext/openssl/ossl_pkey.c
++++ b/ext/openssl/ossl_pkey.c
+@@ -670,7 +670,7 @@ ossl_pkey_export_traditional(int argc, V
+ 	}
+     }
+     else {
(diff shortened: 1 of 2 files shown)


lang/ruby/patches/001-fix-build-with-LibreSSL-over-v3.5.patch
```diff
@@ -0,0 +1,11 @@
+--- a/ext/openssl/ossl_pkey.c
```
BKPepe, line 1: This patch comes from upstream? 
- If not, are you going to upstream this? 
- If yes, who is the author? Where is the commit message?

luizluca: I believe the upstream fix is this. And it is a little bit different. https://github.com/ruby/openssl/commit/e25fb0d0d86da5a9398ebdc9216b2ea89f80fa3d

msylgj: I just found the patch from openbsd source. Gladly to see that upstream was fixed it. May i close this pr and wait for sync from upsteam? @luizluca

luizluca: If you are facing the issue, you can just update this PR with the upstream patch. It might take some time until a new ruby is released with the fix incorporated and a little bit more to have the ruby package updated. I don't see any problem backporting fixes until then. The problem was only the use of a patch without any reference, comment, source, author.

msylgj: Okay, Thanks. I will refresh the patch with upstream commit.
Done it.

PR #19030, opened 2022-07-23
ddns-scripts: Add Simply.com provider

tomsommer:
Maintainer: @tomsommer

Description:
Add Simply.com ddns provider. Large danish/scandinavian DNS provider


feckert:
Please merge your commit to one commit. And also add a valid prefix to your commit.
For example:
`ddns-scripts: add simply.com provider`

PR #19033, opened 2022-07-23
acme.sh Support for --reloadcmd

stokito:
Maintainer: @tohojo

Description: I need to restart lighttpd and emailrelay but this is not supported out of the box. Not a big deal because we can use the `--reloadcmd` option with `/etc/init.d/lighttpd restart`. But the reloadcmd is not supported yet.
The PR adds it's support.

But it doesn't work if the command contains a space like in the `/etc/init.d/lighttpd restart`:
```
acme: Running acme.sh as '/usr/lib/acme/acme.sh --home /etc/acme --renew -d example.com --debug --reloadcmd "/etc/init.d/lighttpd restart" --ecc'
Unknown parameter : restart"
```

This is some problem here:
```sh
run_acme() {
	debug "Running acme.sh as '$ACME $*'"
	$ACME "$@"
}
```

I have no idea how to fix that.

### DIFF
--- a/net/acme/files/acme.config
+++ b/net/acme/files/acme.config
@@ -28,6 +28,7 @@ config cert 'example'
 	option update_uhttpd 1
 	option update_nginx 1
 	option update_haproxy 0
+	# option reloadcmd ""
 	list domains example.org
 	list domains sub.example.org
 	option webroot ""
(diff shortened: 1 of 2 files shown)


feckert:
I think this could be closed.
Because this already pullrequest does the same.
https://github.com/openwrt/packages/commit/e84f6514538f0638507ce8066f801374db6f4150

stokito:
Details about the hotplug change https://github.com/openwrt/packages/pull/17721

PR #19045, opened 2022-07-24
nextdns: fix init.d creation when the package is built-in

drizzt:
Signed-off-by: Timothy Redaelli <tredaelli@redhat.com>

Maintainer: @rs 
Compile tested: compiled openwrt last-master for linksys_e8450 (mt7622) with nextdns selected as built-in (=y)
Run tested:
tested on openwrt last-master on linksys_e8450 (mt7622) by just build an image with nextdns selected as built-in (=y) and checking that nextdns is running and it's working correctly

Description:
postinst, in this case, it's not executed then the package is installed as built-in.
uci-defaults it's executed on the first boot after a firmware update, so just use it

### DIFF
--- a/net/nextdns/Makefile
+++ b/net/nextdns/Makefile
@@ -50,13 +50,14 @@ define Package/nextdns/install
 
 	$(INSTALL_DIR) $(1)/etc/config
 	$(INSTALL_CONF) ./files/nextdns.config $(1)/etc/config/nextdns
+
+	$(INSTALL_DIR) $(1)/etc/uci-defaults
+	$(INSTALL_BIN) ./files/nextdns.defaults $(1)/etc/uci-defaults/nextdns
 endef
 
 define Package/nextdns/postinst
 #!/bin/sh
-if [ -z "$${IPKG_INSTROOT}" ]; then
-  nextdns install
-fi
+[ -n "$${IPKG_INSTROOT}" ] || (. /etc/uci-defaults/nextdns) && rm -f /etc/uci-defaults/nextdns
 endef
 
 define Package/nextdns/prerm
--- a/net/nextdns/files/nextdns.defaults
+++ b/net/nextdns/files/nextdns.defaults
@@ -1,0 +1,2 @@
+nextdns install
+service nextdns enable


feckert:
I don't know if that's such a good thing. I don't know the service exactly, but shouldn't it be configured first?

drizzt:
I only copied the same behavior the package have when you install it using a package instead of installing built in. If you objection is valid, it should be valid also when you install it as package. Isn't it?

fantom-x:
I use this service and I remember a discussion here a while back when the authors did not want to maintain the config in each repo/product they support: it was too much effort. They added all that logic into their binary via `nextdns install`, which generates an init.d script as well as a new (it is being overwritten) /etc/config/nextdns: this config file which is in this repo is stale and does not get updated. The service is activate after this step and mostly configured.

Once those files are installed, the service needs to be enabled (/etc/init.d/nextdns enable), which also does not happen if an image builder is used. So the installation of the package is broken if (1) it is built-in and (2) when the image builder is used.

@drizzt , this fix will be really appreciated by users. The uci-defaults file might need `#!/bin/sh` at the beginning just like other scripts, though.

fantom-x:
Found it https://github.com/openwrt/packages/pull/12360

fantom-x:
@drizzt , if I already have nextdns installed/configured and pull an updated version (with your change), I will get a new file under `/etc/uci-defaults` that will get executed and would overwrite my config?

drizzt:
@fantom-x no, the config file is not overwritten... I only configured it once when I installed the package the first time... then I built openwrt by myself and found that the init file was gone, so I applied this patch and I only have nextdns enabled and working at boot, but the config file is not overwritten.
This only launch nextdns install, that is exactly the same command than the package launch when you install or upgrade it + service nextdns enable since usually when you install a package on openwrt the service is enabled automatically

fantom-x:
Below are two use cases that I am thinking about. I could be completely wrong here, though.

First one:

1. opkg install nextdns
2. It now works
3. opkg update nextdns `<<<< with you changes`
4. Now I have a new file: /etc/uci-defaults/nextdns
5. Reboot
6. /etc/uci-defaults/nextdns gets executed and overwrites my config

Second one

1. opkg install nextdns
2. It now works
3. I do not like nextdns and disable it
4. opkg update nextdns `<<<< with you changes`
5. Now I have a new file: /etc/uci-defaults/nextdns
6. Reboot
7. /etc/uci-defaults/nextdns gets executed and overwrites my config
8. It also enabled nextdns service that was disabled in step 3

There is a default nextdns.config in this package https://github.com/openwrt/packages/tree/master/net/nextdns/files/nextdns.config that gets installed by opkg and to my knowledge it always gets overwriten by `nextdns install`.

Also, `service nextdns enable` will always enable the service even if it was disabled before.

I think you need to increment `PKG_RELEASE` for `opkg update` to work...

drizzt:
@fantom-x in my previous comment I told you, by pointing the source code, that nextdns install doesn't change the configuration so first point is not a problem and like any files in /etc/config, opkg doesn't overwrite /etc/config/nextdns if it's modified, but it can install the "plain"/"package" version as /etc/config/nextdns-opkg.
You are right in the second point, but since you have enabled config option and uci-defaults it's only executed the first time you reboot after an update I guess it's not a big deal

fantom-x:
I just tried and the file does get overwritten by "nextdns install", but in a way that seems to be preserving my config: I missed this part the first time. You can reorder config items in /etc/config/nextdns, run `nextdns install` and compare the files before and after. Also look at the timestamp. `nextdns install` also removes all comments I might have added to `/etc/config/nextdns`.
"nextdns install" actually starts nextdns (`NextDNS installed and started using procd init`), which is not desirable if it was disabled before.
As far as I am concerned, the second point is a big deal, because name resolution may stop working if there are conflicts, but I am not a maintainer to decide. Good luck!

```
~# ls -la /etc/config/nextdns 
-rw-r--r--    1 root     root           498 Jul 28 09:21 /etc/config/nextdns
~# nextdns install
NextDNS installed and started using procd init
~# ls -la /etc/config/nextdns 
-rw-r--r--    1 root     root           496 Jul 28 09:28 /etc/config/nextdns
```

drizzt:
nextdns install calls uci that rewrites the file (you have the same behavior if you change something via luci or by using uci set and uci commit, so configuration is not lost

fantom-x:
Right, but in this case I did not want to run uci and would not expect anything to change at all: I was not making any changes, yet the file now looks different even though the config is the same. I still think this is a problem (a stealth file change), but I am not a maintainer. May maintainers go easy on you :)

net/nextdns/Makefile
```diff
@@ -50,13 +50,14 @@ define Package/nextdns/install
...
+	$(INSTALL_DIR) $(1)/etc/uci-defaults
+	$(INSTALL_BIN) ./files/nextdns.defaults $(1)/etc/uci-defaults/nextdns
 endef
 
 define Package/nextdns/postinst
 #!/bin/sh
-if [ -z "$${IPKG_INSTROOT}" ]; then
-  nextdns install
-fi
+[ -n "$${IPKG_INSTROOT}" ] || (. /etc/uci-defaults/nextdns) && rm -f /etc/uci-defaults/nextdns
```
jow-, line 60: That logic is already covered by the [default postinstall procedure](https://github.com/openwrt/openwrt/blob/master/package/base-files/files/lib/functions.sh#L284), so you can remove the `postinst` section entirely from the Makefile.

fantom-x:
I just tried this approach with image builder and `service nextdns enable` failed with `service: unknown ACTION 'nextdns'` during the boot. This command was in a custom script under `/etc/uci-defaults` and the script ran last.
The command ran fine after the router completely booted, though. I changed it to `/etc/init.d/nextdns enable` and it worked from `/etc/uci-defaults`.

hnyman:
superseded by #19315

PR #19048, opened 2022-07-25
stress-ng: update to 0.14.03

guijan:

Maintainer: @commodo
Compile tested: x64, latest snapshot SDK as of today

### DIFF
--- a/utils/stress-ng/patches/010-dont_define_fortify.patch
+++ b/utils/stress-ng/patches/010-dont_define_fortify.patch
@@ -1,0 +1,13 @@
+--- a/stress-ng.h
++++ b/stress-ng.h
+@@ -34,10 +34,6 @@
+ #ifndef _LARGEFILE64_SOURCE
+ #define _LARGEFILE64_SOURCE
+ #endif
+-#if !defined(__PCC__) && 	\
+-    !defined(__TINYC__)
+-#define _FORTIFY_SOURCE 2
+-#endif
+ 
+ #if !defined(_FILE_OFFSET_BITS)
+ #define _FILE_OFFSET_BITS 	(64)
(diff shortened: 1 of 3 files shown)


commodo:
is this build error new?  

https://github.com/openwrt/packages/runs/7495316112?check_suite_focus=true

i am not sure if this was present;
tbh: the ARC toolchains have always been problematic

i'll need to see if we can somehow fix it (if it wasn't broken before);

commodo:
other than ARC, rest is LGTM

guijan:
It's new, the old version succeeded: https://github.com/openwrt/packages/runs/4019861942?check_suite_focus=true

Odd that this fails on ARC, the relevant Linux headers seem to be arch-independent...

neheb:
Probably a result of glibc update.

guijan:
Sorry, I didn't test adding the patch, I'll do it properly.

utils/stress-ng/patches/010-dont_define_fortify.patch
```diff
@@ -0,0 +1,13 @@
+--- a/stress-ng.h
++++ b/stress-ng.h
+@@ -34,10 +34,6 @@
+ #ifndef _LARGEFILE64_SOURCE
+ #define _LARGEFILE64_SOURCE
+ #endif
+-#if !defined(__PCC__) && 	\
+-    !defined(__TINYC__)
+-#define _FORTIFY_SOURCE 2
+-#endif
```
commodo, line 10: this patch is fine, but i don't see it fixing errors;
it's fixing warnings;

if possible i would send it upstream to have it discussed with them;
they are reasonably quick;

i would not carry patches in our tree if they don't fix anything more grave;

PR #19053, opened 2022-07-26
zmq: remove libbsd dependency

neheb:
Not really needed for anything. Patch taken from upstream.

Maintainer: @srdgame 
Compile tested: mipsel

ping @guijan


guijan:
Thanks for sending that patch to OpenWRT, it'll make things easier for me.

PR #19054, opened 2022-07-26
tar: add PKG_CONFIG_DEPENDS

neheb:
fixes compilation when deselecting options.

Maintainer: @Noltari 

note that compiling without CONFIG_PACKAGE_TAR_GZIP will result in an unupgradable system. See: https://github.com/openwrt/luci/issues/5893


hnyman:
The change itself looks good to me, but would it make sense to add at the same time a note about "gzip support is needed for OpenWrt sysupgrade and backup functionality" to that config item's explanation text.

neheb:
Done.

PR #19058, opened 2022-07-26
[22.03] curl: switch default SSL to mbedTLS

stangri:
Maintainer: me

Description:
The decision to switch the default to wolfSSL was taken because of
hostapd back from when curl was in base. Unfortunately, not only is
wolfSSL bigger but it has also been causing issues recently. There's
also no relation between hostapd and curl.

(cherry picked from commit 0a65e4c6fa84e13a1732c26260615c1ad5efea5b)

### DIFF
--- a/net/curl/Config.in
+++ b/net/curl/Config.in
@@ -4,7 +4,7 @@ comment "SSL support"
 
 choice
 	prompt "Selected SSL library"
-	default LIBCURL_WOLFSSL
+	default LIBCURL_MBEDTLS
 
 	config LIBCURL_MBEDTLS
 		bool "mbed TLS"


neheb (approved):

neheb:
22.03 is unreleased so I feel comfortable with this.

jow-:
There is a pending fix for the wolfSSL issue. Changing the crpyot provider of curl in the middle of the relase between two RCs seems quite suboptimal.

stangri:
@jow- curl was broken in 22.03 and people complained. Is the plan to implement the fix for wolfssl in snapshots as well? Can it then be cherry-picked for 22.03?

PR #19061, opened 2022-07-27
HTOP: add temperature sensor for Marvell Armada XP

M95D:
Maintainer: none, last updated by @graysky2 
Compile tested: Gentoo Linux on x86-64, OpenWrt r19971
Run tested: Linksys WRT-1900AC v1, OpenWrt r19971, htop works and shows the same CPU temperature as luci-statistics.

Description:
Add Marvell Armada XP temperature sensor to htop list of sensors. This will enable htop to display CPU temperature on these routers in the same way it does on x86 platforms.

To show temperature on Armada XP routers these options must be enabled:
- kernel CONFIG_ARMADA_THERMAL (already selected),
- kernel CONFIG_THERMAL_HWMON (already selected),
- openwrt CONFIG_HTOP_LMSENSORS and dependencies (not selected by default afaik).

### DIFF
--- a/admin/htop/patches/010-add-temp-sensor-mvebu-armada-xp.patch
+++ b/admin/htop/patches/010-add-temp-sensor-mvebu-armada-xp.patch
@@ -1,0 +1,10 @@
+--- a/linux/LibSensors.c
++++ b/linux/LibSensors.c
+@@ -130,6 +130,7 @@ static int tempDriverPriority(const sens
+       { "cpu_thermal", 0 },
+       { "k10temp",     0 },
+       { "zenpower",    0 },
++      { "f10182b0.thermal",0},	// Marvell Armada XP
+       /* Low priority drivers */
+       { "acpitz",      1 },
+    };


hnyman:
> Maintainer: none

???

`  MAINTAINER:=Etienne CHAMPETIER <champetier.etienne@gmail.com>`

cc maintainer @champtar

M95D:
> ` MAINTAINER:=Etienne CHAMPETIER <champetier.etienne@gmail.com>`
Sorry. I missed that.

> @M95D can you send that upstream first ? Distro patch should only fix build in general
I don't think they would be interested in routers. How many of us (openwrt users) build htop with lmsensors, anyway?

champtar:
@M95D Just to start the discussion / so maybe they implement something generic that display all *.thermal

neheb:
```
thermal: thermal@182b0
```

That's the last part. f0 is from...something. Reading kernel code, I see

```
clk_register_clkdev(clk, NULL, "e07008c4.thermal");
```

So it seems to be a normal format. Probably would be nice to find out how to rename though...

Issue #19012, opened 2022-07-22
transmission-daemon: Couldn't save temporary file ".../transmission/settings.json.tmp.XXXXXX": Permission denied (variant.c:1235)


siwind:
Maintainer: @\<github-user> (find it by checking history of the package Makefile)
Environment: arm_cortex-a7_neon-vfpv4,  Openwrt master

Description:

After update to latest master version,(kernel version: 5.10.131) 
my transmission failed to add torrent file:
(the storage disk is my USB disks)

```
daemon.err transmission-daemon[7929]: Couldn't save temporary file "/mnt/udisk/transmission/settings.json.tmp.XXXXXX": Permission denied (variant.c:1235)
daemon.err transmission-daemon[7929]: Unable to create session lock file (2): No such file or directory (session-id.c:96)
```

I use another user/group to run transmission:
and directories have correct permissions:

```
# ps | grep tran
 8164 root      2408 SN   {transmission} /sbin/ujail -t 5 -n transmission -S /etc/seccomp/transmission-daemon.json -U wangwang -G wangwang -l -e CURL_CA_BUNDLE -r /etc
 8166 wangwang  6156 SN   /usr/bin/transmission-daemon -f -g /mnt/udisk/transmission
 8214 root      1144 S    grep tran

#
root@OpenWrt:/mnt/udisk/transmission# ll
drw-rw-rw-    5 wangwang wangwang      4096 Feb 21 19:43 ./
drwxr-xr-x   17 wangwang wangwang      4096 Mar 30  2021 ../
drwxr-xr-x    2 wangwang wangwang      4096 Dec 26  2020 blocklists/
-rw-------    1 wangwang wangwang      6828 Feb 21 19:43 dht.dat
drwxr-xr-x    2 wangwang wangwang      4096 Jul 22 18:47 resume/
-rw-rw-rw-    1 wangwang wangwang      1506 Jul 22 18:46 settings.json
-rw-------    1 wangwang wangwang       163 Feb 21 19:43 stats.json
drwxr-xr-x    2 wangwang wangwang      4096 Jul 22 18:47 torrents/

```

siwind:
Transmission version:

```
root@OpenWrt:/mnt/udisk/transmission# opkg list-installed | grep transmission
luci-app-transmission - git-22.029.54222-e3fbfe9
luci-i18n-transmission-zh-cn - git-22.181.29827-f2e6f28
transmission-daemon - 3.00-19
transmission-web - 3.00-19
```

[...]

PR #19066, opened 2022-07-28
transmission-daemon: change judge logic to 'true/false' value in transmission.init file

siwind:
Maintainer:  @\<github-user> (find it by checking history of the package Makefile)
Compile tested:  Arch Linux x86-64, OpenWrt master
Run tested: (arm_cortex-a7_neon-vfpv4, OpenWrt master, tests done)

Description:

The configuation file of tranmsission-daemon is settings.json,   and its enabled value in this file is 'true/false',  but not '0/1'.
so the judgement logic in  transmission  init file need  change from '0/1'  to 'true/false'.

See this issue:
https://github.com/openwrt/packages/issues/19012


neheb:
ping @dangowrt

dangowrt:
In a way it would be more coherent with OpenWrt to change the values in `/etc/config/transmission` to be 0/1 instead of "true"/"false" and then generate settings.json based on that. But I can see that your suggestion will fix the problem is the user spells "false" exactly like that (and not "FALSE" or "False" or anything like that...)

salim-b:
This PR doesn't make much sense. See https://github.com/openwrt/packages/pull/20484 for a proper solution.

salim-b:
I'd strongly suggest to close this now that https://github.com/openwrt/packages/pull/20484 is merged.

dangowrt:
Closed in favor of #20484

PR #19068, opened 2022-07-28
hfsprogs: remove unused libbsd

guijan:
Wikipedia says "[Starting with macOS 10.15, HFS disks can no longer be read](https://en.wikipedia.org/wiki/MacOS_10.15)". hfsprogs also deals with the related HFS+ filesystem, Wikipedia says HFS+ was "[replaced with the Apple File System (APFS), released with macOS High Sierra](https://en.wikipedia.org/wiki/HFS_Plus)". This means support for HFS was removed from macOS in 2019, and HFS+ was deprecated in 2017.
Maybe we could drop this package instead? Tell me if so.

Maintainer: nobody
Compile tested: x64 glibc master

Description:
```
hfsprogs uses libbsd for strlcpy(), but the strlcpy() calls are inside
`#ifdef` and don't show up on the OpenWRT package.
```

### DIFF
--- a/utils/hfsprogs/Makefile
+++ b/utils/hfsprogs/Makefile
@@ -28,7 +28,7 @@ define Package/hfsprogs/Default
   SECTION:=utils
   CATEGORY:=Utilities
   SUBMENU:=Filesystem
-  DEPENDS:=+libopenssl +USE_GLIBC:libbsd
+  DEPENDS:=+libopenssl
 endef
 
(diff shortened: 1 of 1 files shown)


neheb:
works for me.

guijan:
Does "works for me" mean "drop the package" or that you tested it?

neheb:
Neither. I was referring more to the libbsd removal.

PR #19075, opened 2022-07-29
auc: use recommended User-Agent header and sent client

aparcar:
Maintainer: @dangowrt 
Compile tested: CI
Run tested: CI

Description:

The usual format for the User-Agent header is `name/version`, replace the
current format which is `name (version)`.

Additionally send the value within the image request JSON data as `client`. For
web clients it is not possible to modify the User-Agent header, it's based on
the browser. To avoid a special case for `auc`, add the `client` value.


dangowrt:
Merged User-Agent string formatting part via update to 0.3.0

PR #19077, opened 2022-07-31
tree: update upstream source

graysky2:
Update Makefile to use new source URL and corresponding variables due to old source url stating, "notice this site is likely going to be shutdown after over 28 years. I will likely be moving all my code-bases to https://gitlab.com/OldManProgrammer"[1]

1. http://mama.indstate.edu/users/ice/tree

Maintainer: @neheb @hbl0307106015

### DIFF
--- a/utils/tree/Makefile
+++ b/utils/tree/Makefile
@@ -8,12 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tree
-PKG_RELEASE:=1
 PKG_VERSION:=2.0.2
-PKG_SOURCE_URL:=http://mama.indstate.edu/users/ice/tree/src
-PKG_HASH:=7d693a1d88d3c4e70a73e03b8dbbdc12c2945d482647494f2f5bd83a479eeeaf
+PKG_RELEASE:=$(AUTORELEASE)
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tgz
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/Old-Man-Programmer/$(PKG_NAME)/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=753af251ba6446f4d40cc00064e73c55e910d95245a23d63a31ae1f36d444c1f
+
 PKG_MAINTAINER:=Banglang Huang <banglang.huang@foxmail.com>
 
 include $(INCLUDE_DIR)/package.mk


neheb:
that would be @hbl0307106015

graysky2:
Thanks @neheb ... I tried an advanced github search by name (from the Makefile).  How did you find the github username?

neheb:
History of Makefile > Original PR.

graysky2:
> History of Makefile > Original PR.

I did this, but I only see the name that is not a hyperlink whereas other commits are the github name/hyperlinked: 
https://github.com/openwrt/packages/commits/master/utils/tree

utils/tree/Makefile
```diff
@@ -8,12 +8,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tree
-PKG_RELEASE:=1
 PKG_VERSION:=2.0.2
-PKG_SOURCE_URL:=http://mama.indstate.edu/users/ice/tree/src
-PKG_HASH:=7d693a1d88d3c4e70a73e03b8dbbdc12c2945d482647494f2f5bd83a479eeeaf
+PKG_RELEASE:=2
```
BKPepe, line 12: Would it be possible to use autorelease? :)

utils/tree/Makefile
```diff
@@ -8,12 +8,14 @@
...
+
+PKG_SOURCE:=unix-$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://gitlab.com/OldManProgrammer/unix-$(PKG_NAME)/-/archive/$(PKG_VERSION)
+PKG_HASH:=fda48f7443c6883c599cd9e87e93fd843c1cd162e68794d23e4dc6d10efc6250
+PKG_BUILD_DIR:=$(BUILD_DIR)/unix-$(PKG_NAME)-$(PKG_VERSION)
```
BKPepe, line 17: I am thinking if you don't want to download this from [GitHub](https://github.com/Old-Man-Programmer/tree), which is a backup (as said in project's homepage: http://mama.indstate.edu/users/ice/tree/) and the repository (with the tarball) is named as tree.

utils/tree/Makefile
graysky2: I initially converted this to autorelease but found that it did not increment the PKG_RELEASE number.  I will switch it over and force push.

utils/tree/Makefile
graysky2: It's actually gitlab.  I am not sure how else to setup the download.  I modeled this after [apparmor](https://github.com/openwrt/packages/blob/master/utils/apparmor/Makefile).  Please suggest the edit to achieve what you're thinking.

BKPepe: I think you did not understand me. From the project homepage, it is clear that he is moving code to GitLab, which he will be using as a primary repository, but there is also a backup on GitHub, and with using the GitHub, you can avoid doing this craziness. :-)

graysky2: Ah!  I understand now.  Done.

utils/tree/Makefile
```diff
@@ -8,12 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tree
-PKG_RELEASE:=1
 PKG_VERSION:=2.0.2
-PKG_SOURCE_URL:=http://mama.indstate.edu/users/ice/tree/src
-PKG_HASH:=7d693a1d88d3c4e70a73e03b8dbbdc12c2945d482647494f2f5bd83a479eeeaf
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_SOURCE:=$(PKG_VERSION).tar.gz
```
neheb, line 14: Absolutely not.

graysky2: The tag is simply the version.

graysky2: I reverted back to the original layout to allow the source filename to confirm to the standard.

utils/tree/Makefile
graysky2: See the comment below from @neheb regarding the filename.  I switched back to this method.

graysky2: @BKPepe - Finally got the syntax right, see latest commit :+1:

graysky2:
Bah, powerpc_464fp threw an unrelated error.  I cannot trigger it to re-run.

utils/tree/Makefile
BKPepe: Yeah, it seems alright. :)

utils/tree/Makefile
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2017 OpenWrt.org
```
BKPepe, line 2: Have you affiliated anyhow with OpenWrt? It would be best if you kept it as it was.

graysky2: Done.

PR #19083, opened 2022-08-01
lcdproc: don't enable LPT support on non-x86 architectures

pprindeville:
Maintainer: me, @haraldg 
Compile tested: x86_64, generic, HEAD (a7be143646)
Run tested: no install

Description:

Only support parallel-port drivers on x86/x86_64 hardware.

### DIFF
--- a/utils/lcdproc/Makefile
+++ b/utils/lcdproc/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lcdproc
 PKG_VERSION:=0.5.9
-PKG_RELEASE:=7
+PKG_RELEASE:=8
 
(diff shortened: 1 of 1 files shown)


diizzyy:
I think that's a much better approach

pprindeville:
> Yeah, that's much better. However I believe this will run into the same build error, you tried to fix initially. You probably have to make sure sdeclcd driver is only included on x86 too. Maybe by adding a x86 check to this code:
> 
> ```
> ifeq ($(CONFIG_PACKAGE_kmod-lp),y)
> LCDPROC_OTHER_DRIVERS_TEXT+=sdeclcd sed1330 sed1520 stv5730 t6963
> endif
> ```

Done in latest push.

haraldg:
Looks good to me know. Thanks!

pprindeville:
Merged

PR #19087, opened 2022-08-02
rsync: fix CVE-2022-29154

graysky2:
The rsync package is vulnerable to CVE-2022-29154[1], which is not yet in a non-preview release. This commit applies the upstream commit to fix it and several subsequent commits needed to fix bugs the initial fix introduced[2].

1. https://rsync.samba.org/ftp/rsync/NEWS#SECURITY_FIXES-3.2.5
2. https://bugs.archlinux.org/task/75558

Maintainer: @mstorchak

### DIFF
--- a/net/rsync/patches/010-fix-CVE-2022-29154.patch
+++ b/net/rsync/patches/010-fix-CVE-2022-29154.patch
@@ -1,0 +1,385 @@
+From b7231c7d02cfb65d291af74ff66e7d8c507ee871 Mon Sep 17 00:00:00 2001
+From: Wayne Davison <wayne@opencoder.net>
+Date: Sun, 31 Jul 2022 16:55:34 -0700
+Subject: [PATCH] Some extra file-list safety checks.
+
+---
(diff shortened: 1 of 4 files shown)


paper42:
Could you also backport this to 22.03, 21.02 and 19.07? This seems important.

graysky2:
I do not understand why there are dirty patches detected.

https://github.com/openwrt/packages/runs/7745277468?check_suite_focus=true#step:5:3448

graysky2:
@mstorchak - thoughts on suitability for merging?

BKPepe:
I'm taking the initiative to merge this and include it to OpenWrt 22.03 before there is tagged the final release. Would you mind to create pull requests for older releases? Thanks.

graysky2:
@BKPepe - Since the fix itself introduced a few bugs, I think it's better to use an official upstream tagged version particularly since there is a merge window for inclusion in a OW release.  For example, 3.2.5pre2 was tagged a few days ago which includes the fix.  I will send a PR after I do some testing on my RPi4.

Issue #19086, opened 2022-08-02
tailscale: improve packaging


CAFxX:
Maintainer: @BKPepe (?)
Environment: OpenWrt 21.02.3, r16554-1d4dea6d4f

Description: 

The tailscale and tailscaled binaries are quite large, even when squashfs'd:

![image](https://user-images.githubusercontent.com/408018/182301167-0c3ca83d-9857-4aea-b8cf-e4fb741590f1.png)

The tailscale [documentation](https://tailscale.com/kb/1207/small-tailscale/) mentions that, exactly to address this, tailscaled can be optionally built including the cli functionality, so that it can be symlinked busybox-style to support both the tailscaled and tailscale cli roles:

```
go build -o tailscale.combined -tags ts_include_cli ./cmd/tailscaled
ln -s tailscale.combined tailscale
ln -s tailscale.combined tailscaled
```

As currently tailscale depends on tailscaled, we could maybe move from the current two packages

- tailscaled 
- tailscale (depends: tailscaled)

to two packages and a virtual package

- tailscaled
- tailscale-combined (provides: tailscale, tailscaled)

where tailscale-combined is the tailscaled command built using the ts_include_ci tag.

Quickly building the packages locally and compressing with upx (as a stand-in for squashfs) seems to indicate that this approach would save ~1.3MB in a firmware image that includes tailscale (e.g. such as those built via attended-sysupgrades).

```
        File size         Ratio      Format      Name
   --------------------   ------   -----------   -----------
   9961472 ->   2308728   23.18%   linux/mips    tailscale
  26083328 ->   4790500   18.37%   linux/mips    tailscale.combined
  17235968 ->   3765896   21.85%   linux/mips    tailscaled
```

BKPepe:
> As currently tailscale depends on tailscaled, we could maybe move from the current two packages

Feel free to create a pull request to address this.

> Quickly building the packages locally and compressing with upx (as a stand-in for squashfs) seems to indicate that this approach would save ~1.3MB in a firmware image that includes tailscale (e.g. such as those built via attended-sysupgrades).

Please see: https://github.com/openwrt/packages/issues/17984#issuecomment-1059674191

CAFxX:
> Please see: https://github.com/openwrt/packages/issues/17984#issuecomment-1059674191

[...]

PR #19090, opened 2022-08-02
[WIP] tailscale: add tailscale-combined

CAFxX:
https://github.com/openwrt/packages/issues/19086

### DIFF
--- a/net/tailscale/Makefile
+++ b/net/tailscale/Makefile
@@ -52,12 +52,28 @@ define Package/tailscale
   DEPENDS+= +tailscaled
 endef
 
+define Package/tailscale-combined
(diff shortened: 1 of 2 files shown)


BKPepe:
PR is in draft for more than one year, closing.

wrenix:
that is really sad - it will helps alot

CAFxX:
Agreed, unfortunately I was unable to find a way to express in the Makefiles the relationship between the packages that I described above, and I was also lacking a way to quickly test this. If the documentation and testing infrastructure improves I can give it another go, otherwise I just hope someone else will be able to step in and make this happen... As you point out, it would be very useful.

PR #19097, opened 2022-08-03
zabbix: add fping dependency

feckert:
Maintainer: @champtar

Description:
If fping is not installed on the system the following message is show if
hosts should be monitored via icmp.

Log-Message:
"At least one of '/usr/sbin/fping', '/usr/sbin/fping6' must exist. Both
are missing in the system."

To fix this also, add a dependency to 'fping' for 'zabbix-server' and
'zabbix-proxy' for installation to allow icmp hosts monitoring.

### DIFF
--- a/admin/zabbix/Makefile
+++ b/admin/zabbix/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=zabbix
 PKG_VERSION:=5.0.18
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://cdn.zabbix.com/zabbix/sources/stable/5.0/
@@ -113,7 +113,10 @@ endef
 define Package/zabbix-server
   $(call Package/zabbix/Default)
   TITLE+= server
-  DEPENDS += +ZABBIX_POSTGRESQL:libpq +ZABBIX_MYSQL:libmariadbclient +libevent2
+  DEPENDS += +ZABBIX_POSTGRESQL:libpq \
+    +ZABBIX_MYSQL:libmariadbclient \
+    +libevent2 \
+    +fping
 endef
 
 define Package/zabbix-server-frontend
@@ -127,7 +130,10 @@ endef
 define Package/zabbix-proxy
   $(call Package/zabbix/Default)
   TITLE+= proxy
-  DEPENDS += +ZABBIX_POSTGRESQL:libpq +ZABBIX_MYSQL:libmariadbclient +libevent2
+  DEPENDS += +ZABBIX_POSTGRESQL:libpq \
+    +ZABBIX_MYSQL:libmariadbclient \
+    +libevent2 \
+    +fping
 endef
 
 define Package/zabbix-extra-mac80211/description


borkra:
Since only 2 not default packages are using fping, fping dependency needs to be conditional on package selection

feckert:
@borkra Can you please describe this in detail? I don't understand your answer, sorry.

> Since only 2 not default packages are using fping, 

Which default packages.

borkra:
@feckert Usually when package recipe, there is one default/main package which is always must be built and sometimes optional .sub-packages. The dependencies that are needed only for these optional sub-packages needs to be conditional on the sub-package selection, to avoid always building them. I admit the terminology I used is home grown, but hope the idea got across.

feckert:
I understand your words, but fping is only built as a dependency and needed at runtime when I install zabbix-proxy or zabbix-server. If the router is only used as a client then no fping is installed and is also not needed.
The problem is that if I use the `zabbix-proxy` or the `zabbix-server` and want to monitor a client with icmp, then this is done with fping in zabbix. If I do not have installed this on my openwrt system where `zabbix-server` or `zabbix-proxy` is running, I get the following error message that fping is missing on the system. 
I think we should also install this tool and improve the user experience.

borkra:
You added fping to server subpackage specific dpends, not top level, unfortunately, using subpacjage specific depends by itself does not make dependency subpackage specific, one need explicitly add the logic to the dependency itself, otherwise openwrt adds this just to the top level dependency list.

feckert:
So from my point of view, it works as intended.
If I do not select the packages `zabbix-proxy` and `zabbix-server` in `make menuconfig`, then `fping` is also not selected and so not built.
All other enable zabbix-* packages are build.
I do not understand your objection!

borkra:
Unfortunately, that is not how it works, if zabbix-proxy and zabbix-server is not selected fping will be fetched and built but not used.

feckert:
This is my openwrt .config with my patch applied. Before and after selecting `zabbix-server` and `zabbix-proxy`.
As you can see fping is not selected.
So I assume that it will not be built.
I have now started a new build.
I'm curious whether it will be built as you claim. If that is the case, then I find that strange.
```
--- a/configs/x86_64.config
+++ b/configs/x86_64.config
@@ -1704,15 +1704,15 @@ CONFIG_PACKAGE_zabbix-extra-mac80211=m
 CONFIG_PACKAGE_zabbix-extra-network=m
 CONFIG_PACKAGE_zabbix-extra-wifi=m
 CONFIG_PACKAGE_zabbix-get=m
-CONFIG_PACKAGE_zabbix-proxy=m
+# CONFIG_PACKAGE_zabbix-proxy is not set
 CONFIG_PACKAGE_zabbix-sender=m
-CONFIG_PACKAGE_zabbix-server=m
+# CONFIG_PACKAGE_zabbix-server is not set

 #
 # Database Software
 #
 # CONFIG_ZABBIX_MYSQL is not set
-# CONFIG_ZABBIX_POSTGRESQL is not set
+CONFIG_ZABBIX_POSTGRESQL=y
 # CONFIG_PACKAGE_zabbix-server-frontend is not set
 # end of Zabbix

@@ -4368,7 +4368,7 @@ CONFIG_PACKAGE_libelf=m
 # CONFIG_PACKAGE_libestr is not set
 CONFIG_PACKAGE_libev=m
 CONFIG_PACKAGE_libevdev=y
-CONFIG_PACKAGE_libevent2=m
+# CONFIG_PACKAGE_libevent2 is not set
 CONFIG_PACKAGE_libevent2-core=m
 # CONFIG_PACKAGE_libevent2-extra is not set
 # CONFIG_PACKAGE_libevent2-openssl is not set
@@ -7079,7 +7079,7 @@ CONFIG_PACKAGE_ethtool=m
 # CONFIG_PACKAGE_fakepop is not set
 # CONFIG_PACKAGE_family-dns is not set
 # CONFIG_PACKAGE_foolsm is not set
-CONFIG_PACKAGE_fping=m
+# CONFIG_PACKAGE_fping is not set
 # CONFIG_PACKAGE_generate-ipv6-address is not set
 # CONFIG_PACKAGE_gensio-bin is not set
 # CONFIG_PACKAGE_geoipupdate is not set
```

feckert:
@borkra I have made a new build.
And there is no fping being built in my new build.
I have selected all packages from `zabbix-` except `zabbix-server` and `zabbix-proxy` which have a `fping` dependency.
So from my point of view, the buildsystem of openwrt does exactly what it should do.
It only enables fping to get build if `zabbix-server` or `zabbix-proxy` is enable for building!

borkra:
@feckert I merged this PR into my local branch, and enabled through menuconfig zabbix-agentd and only zabbix-agentd

The following extra packages were built as a result (output from make command provided): 

```
 make[3] -C feeds/packages/net/fping compile
 make[3] -C feeds/packages/libs/pcre2 compile
 make[3] -C feeds/packages/libs/expat compile
 make[3] -C feeds/packages/lang/php8 compile
 make[3] -C package/libs/libevent2 compile
 make[3] -C feeds/packages/libs/postgresql host-compile
 make[3] -C package/libs/readline compile
 make[3] -C feeds/packages/libs/postgresql compile
 make[3] -C feeds/packages/admin/zabbix compile
```

As far as I can tell most of them are not needed for zabbix-agentd and should not be built.

champtar:
@borkra this seems like a bug in OpenWrt build system, was there an attempt to fix it there ? (not a fan of manually sprinkling conditional everywhere)

champtar:
@feckert you are actively using zabbix-server on OpenWrt or know people who are ? I've only ever used the agent so would love testers I can ping (or someone to take maintainership of zabbix ;) )

borkra:
@champtar I do not know if it is a bug with openwrt build system, I see that every package who want this to work correctly specifies explicit precondition on each dependency and that includes all core openwrt packages. This issue was discussed lately during wolfssl compilation failure and was fixed just like that, by adding explicit precondition on every dependency.

feckert:
@borkra I repeat my statement. If I do not select `zabbix-server` or `zabbix-proxy` in my own build, then fping will not get built. There is no fping directory under `build_dir/target-<arch>/`.
I cannot say why fping compile is displayed. But the build dir of fping is not there.
Could you verified this?

> @feckert you are actively using zabbix-server on OpenWrt or know people who are ? I've only ever used the agent so would love testers I can ping (or someone to take maintainership of zabbix ;) )

@champtar I am currently verifying whether zabbix is better than checkmk for our purposes.
Unfortunately, we are not yet ready to use zabbix conclusively.
So maintainership maybe in the future but you are welcome to ping me for testing when I have time.

borkra:
@feckert fping exist in dl/ and buid_dir/target-<arch>. Had to clean it after too.  I already verified similar issue multiple times on multiple projects and fixed it on few.

PR #19104, opened 2022-08-03
python-paho-mqtt: add missing dependency

janh:
The uuid module has been split out into a separate package with the update to Python 3.10, so it needs to be added as a dependency here.

This also applies to the 22.03 branch.

Maintainers: @BKPepe, @commodo

### DIFF
--- a/lang/python/python-paho-mqtt/Makefile
+++ b/lang/python/python-paho-mqtt/Makefile
@@ -25,7 +25,7 @@ define Package/python3-paho-mqtt
   SUBMENU:=Python
   TITLE:=MQTT version 5.0/3.1.1 client class
   URL:=http://eclipse.org/paho
-  DEPENDS:=+python3-light
+  DEPENDS:=+python3-light +python3-uuid
 endef
 
 define Package/python3-paho-mqtt/description


BKPepe (approved):
Oh, in my case, I have installed python3 (not the light version) and that package depends on python3-uuid, that's why I was not unable to reproduce it.

Tested:
```
root@turris:~# python3
Python 3.10.5 (main, Aug  3 2022, 03:51:33) [GCC 8.4.0] on linux
>>> import paho.mqtt.client as mqtt
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/site-packages/paho/mqtt/client.py", line 24, in <module>
ModuleNotFoundError: No module named 'uuid'
```

Thanks!

commodo:
Thanks from me as well :)

janh:
Can you cherry-pick this change into 22.03, or should I open a separate pull request for that?

BKPepe:
Cherry-picked as requested - https://github.com/openwrt/packages/commit/0fd2eb51af1da1c80d40e065ab921b03c24d7ab4

PR #19110, opened 2022-08-04
python3: backport and fix target musl libc detection

autobakterie:
Patch 030:
Backported from Python main branch[^1] for Python to distinguish between glibc and musl libc SOABI.

Patch 131:
Changes PLATFORM_TRIPLET -gnu/-musl suffix detection (performed by the backported patch)
to be based on the target OS instead of the building OS.

See included patches for more detailed descriptions.

Specifically this fixes cross-compilation for mpc8548 CPUs with SPE instructions[^2] enabled.

[^1]: merged to python:main as https://github.com/python/cpython/pull/24502 'bpo-43112: detect musl as a separate SOABI'
[^2]: https://www.nxp.com/docs/en/reference-manual/SPEPEM.pdf

Co-authored-by: Pali Rohár <pali@kernel.org>
Signed-off-by: Šimon Bořek <simon.borek@nic.cz>

Maintainer: @jefferyto
Compile and run tested: mpc85xx/p2020, Turris 1.x, OpenWrt 21.02 and master

### DIFF
--- a/lang/python/python3/patches/030-bpo-43112-detect-musl-as-a-separate-SOABI-GH-24502.patch
+++ b/lang/python/python3/patches/030-bpo-43112-detect-musl-as-a-separate-SOABI-GH-24502.patch
@@ -1,0 +1,75 @@
+From 3f79de7b8411c76a1fcd1ca850ea62500be7a881 Mon Sep 17 00:00:00 2001
+From: Natanael Copa <ncopa@alpinelinux.org>
+Date: Sat, 29 Jan 2022 00:02:54 +0100
+Subject: [PATCH 1/2] bpo-43112: detect musl as a separate SOABI (GH-24502)
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+musl libc and gnu libc are not ABI compatible so we need set different
+SOABI for musl and not simply assume that all linux is linux-gnu.
+
+Replace linux-gnu with the detected os for the build from config.guess
+for linux-musl*.
+
+(cherry picked from commit 1f036ede59e2c4befc07714cf76603c591d5c972)
+Signed-off-by: Šimon Bořek <simon.borek@nic.cz>
+---
+ Lib/test/test_sysconfig.py                                | 8 ++++----
+ .../next/Build/2021-02-10-17-54-04.bpo-43112.H5Lat6.rst   | 1 +
+ configure                                                 | 5 +++++
+ configure.ac                                              | 5 +++++
+ 4 files changed, 15 insertions(+), 4 deletions(-)
+ create mode 100644 Misc/NEWS.d/next/Build/2021-02-10-17-54-04.bpo-43112.H5Lat6.rst
+
+--- a/Lib/test/test_sysconfig.py
++++ b/Lib/test/test_sysconfig.py
+@@ -425,11 +425,11 @@ class TestSysConfig(unittest.TestCase):
+             self.assertTrue('linux' in suffix, suffix)
+         if re.match('(i[3-6]86|x86_64)$', machine):
+             if ctypes.sizeof(ctypes.c_char_p()) == 4:
+-                self.assertTrue(suffix.endswith('i386-linux-gnu.so') or
+-                                suffix.endswith('x86_64-linux-gnux32.so'),
+-                                suffix)
++                expected_suffixes = 'i386-linux-gnu.so', 'x86_64-linux-gnux32.so', 'i386-linux-musl.so'
+             else: # 8 byte pointer size
+-                self.assertTrue(suffix.endswith('x86_64-linux-gnu.so'), suffix)
++                expected_suffixes = 'x86_64-linux-gnu.so', 'x86_64-linux-musl.so'
++            self.assertTrue(suffix.endswith(expected_suffixes),
++                            f'unexpected suffix {suffix!r}')
+ 
+     @unittest.skipUnless(sys.platform == 'darwin', 'OS X-specific test')
+     def test_osx_ext_suffix(self):
+--- /dev/null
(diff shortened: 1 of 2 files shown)


autobakterie:
As mentioned in https://github.com/python/cpython/pull/24502#issuecomment-1021290819 incorporating this might break loading modules unless they are rebuilt along with the update, therefore this change won't be backported to older versions of upstream cpython. I'm not sure if this is a relevant issue in the context of OpenWrt package.

Could somebody more knowledgeable clarify this?

lang/python/python3/patches/030-bpo-43112-detect-musl-as-a-separate-SOABI-GH-24502.patch
```diff
@@ -0,0 +1,75 @@
...
+             if ctypes.sizeof(ctypes.c_char_p()) == 4:
+-                self.assertTrue(suffix.endswith('i386-linux-gnu.so') or
+-                                suffix.endswith('x86_64-linux-gnux32.so'),
+-                                suffix)
++                expected_suffixes = 'i386-linux-gnu.so', 'x86_64-linux-gnux32.so', 'i386-linux-musl.so'
+             else: # 8 byte pointer size
+-                self.assertTrue(suffix.endswith('x86_64-linux-gnu.so'), suffix)
++                expected_suffixes = 'x86_64-linux-gnu.so', 'x86_64-linux-musl.so'
++        self.assertTrue(suffix.endswith(expected_suffixes),
++                        f'unexpected suffix {suffix!r}')
```
merwok, line 39: You need to backport the text fix too 🙂 

```suggestion
+            self.assertTrue(suffix.endswith(expected_suffixes),
+                            f'unexpected suffix {suffix!r}')
```

autobakterie: Thanks for the reminder. Should be corrected by the fixup.

BKPepe:
cc: @commodo as well

commodo:
this probably means that we should re-visit patch `lang/python/python3/patches/014-remove-platform-so-suffix.patch` ?

commodo:
> this probably means that we should re-visit patch `lang/python/python3/patches/014-remove-platform-so-suffix.patch` ?

(we may have to also re-visit `lang/python/python3/patches/016-adjust-config-paths.patch` )

so, to detail my side on this (from memories):
* i've been meaning to remove patches 014 & 016 for a while, but it requires re-building the entire Python packages (which grew over time)
* as-far-as-i-remember, in OpenWrt, we just disconsider `PLATFORM_TRIPLET` for the target build; and we don't really need it, because when we build for a target, the files get put in the target build/staging folder of that target;
* that being said, it's also a good idea to not keep any OpenWrt specifics, if CPython could handle these elegantly; but it does require some time to re-build everything a couple of times until it successfully compiles (everything with all Python packages)
* AFAICT: this patch could be a no-op, because patches 014 & 016 will just remove the PLATFORM_TRIPLET anyway? I could be wrong, but I would first start by removing those 2 and then trying to see if these new patches fix anything

autobakterie:
> this probably means that we should re-visit patch lang/python/python3/patches/014-remove-platform-so-suffix.patch ?
> (we may have to also re-visit lang/python/python3/patches/016-adjust-config-paths.patch )

I do not think that is necessary, as the change in `PLATFORM_TRIPLET` construction (swapping '-gnu' with '-musl' in relevant cases) does not affect the operation of patches making Python ignore `PLATFORM_TRIPLET` when naming files/directories (if I understand them correctly) - it will still be ignored in this context. (although even with patches 014 and 016 applied, the backport still fixes the problem described below)


> i've been meaning to remove patches 014 & 016 for a while, but it requires re-building the entire Python packages (which grew over time)
> as-far-as-i-remember, in OpenWrt, we just disconsider PLATFORM_TRIPLET for the target build; and we don't really need it, because when we build for a target, the files get put in the target build/staging folder of that target;
> that being said, it's also a good idea to not keep any OpenWrt specifics, if CPython could handle these elegantly; but it does require some time to re-build everything a couple of times until it successfully compiles (everything with all Python packages)

It looks like a good idea to stick to upstream Python build process behaviour where possible, although I don't think I have deep enough insight into OpenWrt's Python packages build process to assess the effects removal of patches 14 and 16 will have.

> AFAICT: this patch could be a no-op, because patches 014 & 016 will just remove the PLATFORM_TRIPLET anyway? I could be wrong, but I would first start by removing those 2 and then trying to see if these new patches fix anything

I'm proposing this as a working fix for an existing issue (tested on Turris OS branches based on OpenWrt master and 21.02, Turris 1.x device, as stated in the description). I'm currently in the process of submitting patches allowing OpenWrt to be compiled for mpc85xx targets with Power ISA SPE extension enabled (enables CPU native floating point arithmetic among other things; work already tested and merged in development branches of Turris OS[^1][^2]). Without the backport (plus its fix[^3]), compilation of Python for mpc8548 musl-spe fails.

[^1]: https://gitlab.nic.cz/turris/os/build/-/merge_requests/542
[^2]: https://gitlab.nic.cz/turris/os/build/-/merge_requests/544
[^3]: already suggested upstream https://github.com/python/cpython/issues/95855

commodo:
@autobakterie 

thanks for the info :)

so, from my side these patches are fine if they fix mpc85xx and don't break others (which seems to be the case)

i'll let @jefferyto have the last word on this;

autobakterie:
It's been almost 2 weeks since the last update in this thread. Are further adjustments needed? @jefferyto, do you think this can be merged?

neheb:
Formalities need to be fixed.

jefferyto:
Apologies for being absent on this - I would like ask regarding 131-configure_ac-switch-PLATFORM_TRIPLET-suffix-to-musl.patch:
* Is this an original patch for OpenWrt or was it originally submitted somewhere else?
* Is this suitable to be upstreamed? If yes, has it been submitted upstream?

PR #19112, opened 2022-08-04
gerbera: remove

neheb:
This was meant to be a less buggy replacement to minidlna.
Unfortunately, the opposite seems to have happened over time. In
addition, it currently doesn't build.

I don't see many people actually using this. Probably better to just
remove.

Maintainer: nobody

I'll merge in two weeks if no one complains.


BKPepe:
Oh, that's too bad. If I am not mistaken some people from Turris community were using it. :-(

neheb:
There was one guy that was using it. He switched to running it from a container.

PR #19118, opened 2022-08-05
luajit: patch: PPC/e500 SPE: use soft float instead of failing

autobakterie:
makes LuaJit builds for mpc85xx targets with SPE ISA extension
enabled possible

Quoting inner commit message:

This allows building LuaJit for systems with Power ISA SPE
extension[^1] support by using soft float on LuaJit side.

While e500 CPU cores support SPE instruction set extension
allowing them to perform floating point arithmetic natively,
this isn't required. They can function with software floating
point to integer arithmetic translation as well,
just like FPU-less PowerPC CPUs without SPE support.

Therefore I see no need to prevent them from running LuaJit
explicitly.

[^1]: https://www.nxp.com/docs/en/reference-manual/SPEPEM.pdf

Signed-off-by: Pali Rohár <pali@kernel.org>
Signed-off-by: Šimon Bořek <simon.borek@nic.cz>

Maintainer: @milani
Compile and run tested: mpc85xx/p2020, Turris 1.x, OpenWrt 21.02 and master

### DIFF
--- a/lang/luajit/patches/300-PPC-e500-with-SPE-enabled-use-soft-float.patch
+++ b/lang/luajit/patches/300-PPC-e500-with-SPE-enabled-use-soft-float.patch
@@ -1,0 +1,58 @@
+From 2c8b9d5807a515a0ff859899065f512beabcbaed Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?=C5=A0imon=20Bo=C5=99ek?= <simon.borek@nic.cz>
+Date: Sat, 16 Jul 2022 18:39:55 +0200
+Subject: [PATCH] PPC/e500 with SPE enabled: use soft float instead of failing
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+This allows building LuaJit for systems with Power ISA SPE
+extension[^1] support by using soft float on LuaJit side.
+
+While e500 CPU cores support SPE instruction set extension
+allowing them to perform floating point arithmetic natively,
+this isn't required. They can function with software floating
+point to integer arithmetic translation as well,
+just like FPU-less PowerPC CPUs without SPE support.
+
+Therefore I see no need to prevent them from running LuaJit
+explicitly.
+
+[^1]: https://www.nxp.com/docs/en/reference-manual/SPEPEM.pdf
+
+Signed-off-by: Pali Rohár <pali@kernel.org>
+Signed-off-by: Šimon Bořek <simon.borek@nic.cz>
+---
+ src/lj_arch.h | 7 ++-----
+ 1 file changed, 2 insertions(+), 5 deletions(-)
+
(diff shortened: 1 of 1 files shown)


autobakterie:
Corresponding PR to Openresty's LuaJit 2 here: https://github.com/openresty/luajit2/pull/180

neheb:
I'm confused here. https://downloads.openwrt.org/snapshots/faillogs/powerpc_8540/packages/ does not list luajit as a failed package.

The previous commit that added softfloat PPC luajit support was https://github.com/openwrt/packages/commit/24c0007ea2561611776e50c8876a7b040ffd6fdc

What am I missing?

autobakterie:
The thing is that mpc85xx target devices have CPUs with e500 cores that support Power ISA SPE extension as described in the description, more specifically - all supported subtargets - p1010, p1020, p2020 - use mpc8548 CPUs with e500v2 cores.

Regarding those e500v2 cores:

They don't have an FPU with dedicated floating point registers. Instead, they support the already mentioned SPE instruction set extension with instructions allowing floating point arithmetic on floats stored in GPRs[^1]. The extension instructions also allow utilizing the whole size of the GPRs, which are, even though the CPUs are generally considered to be 32-bit, of 64 bits size (only halves are accessible using the base instruction set) enabling direct storage and operations on not only single, but double precision floats and vectors as well.

Regardless, the current configuration of OpenWrt treats mpc8548 CPUs as mpc8540 and incapable of floating point arithmetic. The system and packages are utilizing translation of floating point operations to integer ones reducing performance unnecessarily on native floating point operations capable CPU cores.

If SPE ISA extension is enabled - which is a change we are just incorporating into Turris OS for Turris 1.x devices[^2] - this leads to LuaJit build process having `__NO_FPRS__` defined (there are no floating point registers) and `_SOFT_FLOAT` undefined (although there are no floating point registers, the CPU is still capable of floating point operations) in the same time which in turn currently leads to compile time error "No support for PPC/e500 anymore (use LuaJIT 2.0)" and the build exiting (that doesn't happen in current mpc85xx OpenWrt, because it always uses soft float, even though it's not needed, leading to `_SOFT_FLOAT` being defined).

While it would be a lot of work to restore SPE extension support in LuaJit, a small change - as suggested in this PR - will make it possible for OpenWrt and packages to utilize SPE (=> better performance) while LuaJit stays with soft float translation (the performance should remain the same, better than not-JIT compiled Lua).

I am just in the process of creating pull requests to OpenWrt based on the mentioned Turris OS patches[^2] with more rigorous descriptions of and reasonings for particular changes regarding the mpc85xx target optimizations.

This PR by itself just stops LuaJit from unnecessarily aborting compilation if the target mpc8548 CPU is properly recognized as floating point operations capable while not altering the current OpenWrt for mpc85xx build process.

[^1]: General Purpose Registers
[^2]: https://gitlab.nic.cz/turris/os/build/-/merge_requests/542/diffs

Issue #1383, opened 2015-06-16
picocom fails to compile with musl


joaochainho:
Hi,
picocom fails to compile with musl as C library.
Arch: ar71xx
Rev: trunk r45995

```
make[3]: Entering directory `/home/openwrt/trunk/build_dir/target-mips_24kc_musl-1.1.9/picocom-1.7'
mips-openwrt-linux-musl-gcc -Os -pipe -march=24kc -mtune=24kc -mno-branch-likely -fno-caller-saves -fhonour-copts -Wno-error=unused-but-set-variable -msoft-float -mips16 -minterlink-mips16 -DVERSION_STR=\"1.7\" -DUUCP_LOCK_DIR=\"/var/lock\" -DHIGH_BAUD  -c -o term.o term.c
term.c:37:20: fatal error: termio.h: No such file or directory
 #include <termio.h>
                    ^
compilation terminated.
make[3]: *** [term.o] Error 1
```

joaochainho:
Thanks for quickly solving this issue. It's running fine on the real device.

karlp:
fwiw, filed upstream as https://code.google.com/p/picocom/issues/detail?id=28

Issue #1323, opened 2015-05-29
vips again not build (error)


build000:
(...)
[...]

PR #19126, opened 2022-08-07
gnutls: update to 3.7.7

PolynomialDivision:
- libgnutls: Fixed double free during verification of pkcs7 signatures.
  Reported by Jaak Ristioja (#1383). [GNUTLS-SA-2022-07-07, CVSS: medium]
  [CVE-2022-2509]

- libgnutls: gnutls_hkdf_expand now only accepts LENGTH argument less than or
  equal to 255 times hash digest size, to comply with RFC 5869 2.3.

- libgnutls: Length limit for TLS PSK usernames has been increased
  from 128 to 65535 characters (#1323).

- libgnutls: AES-GCM encryption function now limits plaintext
  length to 2^39-256 bits, according to SP800-38D 5.2.1.1.

- libgnutls: New block cipher functions have been added to transparently
  handle padding.  gnutls_cipher_encrypt3 and gnutls_cipher_decrypt3 can be
  used in combination of GNUTLS_CIPHER_PADDING_PKCS7 flag to automatically
  add/remove padding if the length of the original plaintext is not a multiple
  of the block size.

- libgnutls: New function for manual FIPS self-testing.

API and ABI modifications:
- gnutls_fips140_run_self_tests: New function
- gnutls_cipher_encrypt3: New function
- gnutls_cipher_decrypt3: New function
- gnutls_cipher_padding_flags_t: New enum

Maintainer: @nmav

### DIFF
--- a/libs/gnutls/Makefile
+++ b/libs/gnutls/Makefile
@@ -6,13 +6,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gnutls
-PKG_VERSION:=3.7.6
+PKG_VERSION:=3.7.7
 PKG_RELEASE:=$(AUTORELEASE)
 PKG_USE_MIPS16:=0
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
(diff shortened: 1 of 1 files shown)


nmav (approved):

nmav:
LGTM

PR #19128, opened 2022-08-07
strongswan: Update to 5.9.7

pprindeville:
Maintainer: me, @Thermi 
Compile tested: x86_64, generic, HEAD (a7be143646)

Description:


pprindeville:
Seeing the following:

```
root@OpenWrt2:~# swanctl -l
plugin 'mgf1': failed to load - mgf1_plugin_create not found and no plugin file available
plugin 'kdf': failed to load - kdf_plugin_create not found and no plugin file available
plugin 'drbg': failed to load - drbg_plugin_create not found and no plugin file available
root@OpenWrt2:~#
```

Thermi:
[0001-gmpdh-Plugin-that-implements-gmp-DH-functions-in-an.txt](https://github.com/openwrt/packages/files/9277911/0001-gmpdh-Plugin-that-implements-gmp-DH-functions-in-an.txt)

Try to use this patch (adjust numbering etc so it's inline with the numbering in the patch set already in use) instead of the 904 one. It should work.

pprindeville:
Need a tester for `gmpdh` ...

PR #19132, opened 2022-08-08
shadowsocks-libev: ss-rules: nft rule cleanup on reload

yousong:
Maintainer: me
Compile tested: n/a
Run tested: x86/64 qemu

Description:

Remove nft rules file generated by ss-rules if ss-rules was or should be
turned off for by configuration.  Use "fw4 restart" instead of "fw4
reload" to force the runtime rule reloading

Ref: https://github.com/openwrt/packages/pull/17937#issuecomment-1207357037
Signed-off-by: Yousong Zhou <yszhou4tech@gmail.com>


jeffpeng78:
@yousong It works excellent now! Thanks a million!

PR #18716, opened 2022-06-09
libowfat: try to fix compilation under ARC

neheb:
CFLAGS are not being passed.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @dangowrt 
Compile tested: mips24

PR #19136, opened 2022-08-08
gatling: add package gatling

Akira25:
Gatling is a high-performance webserver from fefe. It gives a
fairly decent feature-set at really small size. And its fast.

Signed-off-by: Martin Hübner <martin.hubner@web.de>

Maintainer: me
Compile tested: OpenWrt 21.02.3, ath79/generic, TP-Link Archer C5 v1 and ramips/mt76x8, GL-MT300N v2
Run tested: OpenWrt 21.02.3, ath79/generic, TP-Link Archer C5 v1

### DIFF
--- a/net/gatling/Makefile
+++ b/net/gatling/Makefile
@@ -1,0 +1,40 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gatling
+PKG_VERSION:=0.16
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=https://www.fefe.de/gatling/
+PKG_HASH:=5f96438ee201d7f1f6c2e0849ff273b196bdc7493f29a719ce8ed08c8be6365b
+
+PKG_BUILD_DEPENDS:=px5g-mbedtls libowfat
+
+PKG_MAINTAINER:=Martin Hübner <martin.hubner@web.de>
+PKG_LICENSE:=GPL-2.0-only
+PKG_LICENSE_FILES:=LICENSE
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/gatling
+  SECTION:=net
+  CATEGORY:=Network
+  SUBMENU:=Web Servers/Proxies
+  TITLE:=gatling - a high performance web server
+  URL:=https://www.fefe.de/gatling/
+  DEPENDS:=+zlib +libcap +libopenssl
+endef
+
+define Package/gatling/description
+  gatling - a high performance web server
+  with a a rather small binary size but a decent set of extra features
+  like ftp, basic htaccess and basic DoS-Protection.
+  Gatling is particularly good in situations with very high load.
+endef
+
+define Package/gatling/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/gatling $(1)/usr/bin/
+endef
+
+$(eval $(call BuildPackage,gatling))
--- a/net/gatling/patches/010-decrease_optimisation.patch
+++ b/net/gatling/patches/010-decrease_optimisation.patch
@@ -1,0 +1,25 @@
+gatling: decrease optimisation
+
+When compiling on MIPS-architectures, in -O2 there happen some
+errors on interlinking. Linker will will fail with:
+
+```
+Unsupported jump between ISA modes; consider recompiling with interlinking enabled
+```
+
+Reducing optimisation works around these errors and looks like
+a fairly common solution to this problem.
+https://forum.openwrt.org/t/unsupported-jump-between-isa-modes-consider-recompiling-with-interlinking-enabled/55786
+
+Signed-off-by: Martin Hübner <martin.hubner@web.de>
+--- a/GNUmakefile
++++ b/GNUmakefile
+@@ -47,7 +47,7 @@ ifneq ($(DEBUG),)
+ CFLAGS+=-g -Og
+ LDFLAGS+=-g
+ else
+-CFLAGS+=-O2 -fomit-frame-pointer -I/usr/local/include
++CFLAGS+=-O1 -fomit-frame-pointer -I/usr/local/include
+ LDFLAGS+=-s
+ ifneq ($(DIET),)
+ DIET+=-Os
--- a/net/gatling/patches/020-configure_features.patch
+++ b/net/gatling/patches/020-configure_features.patch
@@ -1,0 +1,30 @@
+gatling: configure features
+
+Disables SMB-support, as only deprecated SMBv1 is supported,
+which most users wouldn't use anyway. Avoids potential
+vulnerabilities.
+
+Activates threaded opening: Thus gatling will open files in
+threads to open kernel I/O scheduling opportunities.
+
+Signed-off-by: Martin Hübner <martin.hubner@web.de>
+--- a/gatling_features.h
++++ b/gatling_features.h
+@@ -7,7 +7,7 @@
+ 
+ #define SUPPORT_SERVERSTATUS
+ // #define SUPPORT_DAV
+-#define SUPPORT_SMB
++// #define SUPPORT_SMB
+ #define SUPPORT_FTP
+ #define SUPPORT_PROXY
+ /* #define DEBUG to enable more verbose debug messages for tracking fd
+@@ -36,7 +36,7 @@
+ #define SUPPORT_FALLBACK_REDIR
+ 
+ /* open files in threads to open kernel I/O scheduling opportunities */
+-#undef SUPPORT_THREADED_OPEN
++#define SUPPORT_THREADED_OPEN
+ 
+ /* try to divine MIME type by looking at content */
+ #define SUPPORT_MIMEMAGIC
--- a/net/gatling/test.sh
+++ b/net/gatling/test.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+gatling -h


net/gatling/Makefile
```diff
@@ -0,0 +1,45 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gatling
+PKG_VERSION:=0.16
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/gatling-$(PKG_VERSION)
```
BKPepe, line 7:
```suggestion
```

It is already default.

net/gatling/Makefile
```diff
@@ -0,0 +1,45 @@
...
+PKG_SOURCE:=gatling-$(PKG_VERSION).tar.xz
```
BKPepe, line 8:
```suggestion
PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
```

net/gatling/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/gatling
+	SECTION:=net
+	CATEGORY:=Network
+	SUBMENU:=Web Servers/Proxies
+	TITLE:=gatling - a high performance web server
+	URL:=https://www.fefe.de/gatling/
+	PKG_BUILD_DEPENDS:=openssl px5g-mbedtls libowfat
```
BKPepe, line 24: This seems wrong to me.

net/gatling/Makefile
```diff
@@ -0,0 +1,45 @@
...
+define Package/gatling/description
+	gatling - a high performance web server
+	with a a rather small binary size but a decent set of extra features
+	like ftp, basic htaccess and basic DoS-Protection.
+	Gatling is particularly good in situations with very high load.
+endef
+
+define Build/Configure
+	$(call Build/Configure/Default)
+endef
```
BKPepe, line 38: Why?

net/gatling/patches/series
```diff
@@ -0,0 +1,2 @@
+010-decrease_optimisation.patch
+020-configure_features.patch
```
BKPepe, line 2: What's for?

net/gatling/patches/010-decrease_optimisation.patch
```diff
@@ -0,0 +1,13 @@
+Index: gatling-0.16/GNUmakefile
+===================================================================
+--- gatling-0.16.orig/GNUmakefile
++++ gatling-0.16/GNUmakefile
+@@ -47,7 +47,7 @@ ifneq ($(DEBUG),)
+ CFLAGS+=-g -Og
+ LDFLAGS+=-g
+ else
+-CFLAGS+=-O2 -fomit-frame-pointer -I/usr/local/include
++CFLAGS+=-O1 -fomit-frame-pointer -I/usr/local/include
```
BKPepe, line 10: There is missing commit description, I wonder why you decreased optimisation.

net/gatling/patches/020-configure_features.patch
```diff
@@ -0,0 +1,22 @@
+Index: gatling-0.16/gatling_features.h
+===================================================================
+--- gatling-0.16.orig/gatling_features.h
++++ gatling-0.16/gatling_features.h
```
BKPepe, line 4: Git header and commit description, please.

net/gatling/Makefile
```diff
@@ -0,0 +1,45 @@
...
+	DEPENDS:=+zlib +libcap +libopenssl
+#	DEPENDS:=+zlib +mbedtls +libcap +libopenssl
+endef
+
+define Package/gatling/description
+	gatling - a high performance web server
+	with a a rather small binary size but a decent set of extra features
+	like ftp, basic htaccess and basic DoS-Protection.
+	Gatling is particularly good in situations with very high load.
+endef
```
BKPepe, line 34: Two spaces here.

net/gatling/Makefile
```diff
@@ -0,0 +1,45 @@
...
+PKG_SOURCE_URL:=https://www.fefe.de/gatling/gatling-0.16.tar.xz
+PKG_HASH:=5f96438ee201d7f1f6c2e0849ff273b196bdc7493f29a719ce8ed08c8be6365b
+
+PKG_MAINTAINER:=Martin Hübner <martin.hubner@web.de>
+PKG_LICENSE:=GPL-2.0-only
+PKG_LICENSE_FILES:=LICENSE
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/gatling
```
BKPepe, line 18: Two spaces here as well.

BKPepe (changes requested):

net/gatling/Makefile
Akira25: It was in the example makefile in the Wiki. Is that unnecessary too?

net/gatling/patches/series
Akira25: Thats the file produced by quilt. As far as I got it from the wiki, one should use quilt for having patches in OpenWrt.

net/gatling/patches/010-decrease_optimisation.patch
Akira25: If compiling with `-O2`, it will brake on mips-architectures, as the linker can't link the different modules together. It asks for enabling interlinking, which afaict enabled on the target but default. But it wasn't working for me. So decreasing the optimization was a quite easy workaround for me.

net/gatling/patches/020-configure_features.patch
Akira25: The pathing was done with quilt. Should it be done with git instead?

net/gatling/Makefile
Akira25: should be fine now. Thank you for spotting.

net/gatling/Makefile
Akira25: applied.

net/gatling/Makefile
Akira25: It looks a bit strange in the beginning, but if one omit the px5g-mbedtls, compilation will fail with this error:
```
pssl.c:9:10: fatal error: mbedtls/config.h: No such file or directory
    9 | #include <mbedtls/config.h>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [GNUmakefile:139: pssl.o] Error 1
```
So I assume it was intended by the author, to have openssl and mbedtls mixed somehow.

net/gatling/Makefile
feckert: This is not necessary and is done Implicity

net/gatling/patches/020-configure_features.patch
feckert: Yes please. For other people on rebasing on a new package version it is not bad if there is a clean commit description. What this patch is for.

net/gatling/Makefile
BKPepe: Interesting, but PKG_BUILD_DEPENDS should be elsewhere not in this section.

net/gatling/patches/series
BKPepe: I understand, but no other packages have it, so feel free to remove it.

feckert: This file is not necessary.
If you build the package with `make ./packages/feeds/gatling/{clean,compile} V=s QUILT=1` and then create a new patch with qullt in the build directory and then update the patches into the openwrt packages feed with `make ./packages/feeds/gatling/update`, then this file should not be there.

net/gatling/patches/010-decrease_optimisation.patch
BKPepe: This seems weird to me. In any case, this needs to be said in the commit description, because no else would know why it was decreased.

net/gatling/Makefile
Akira25: Fixed in recent version. Hopefully I got it rigth.

net/gatling/Makefile
Akira25: I removed it now.

net/gatling/patches/series
Akira25: Okay, I deleted it from the PR.

net/gatling/patches/010-decrease_optimisation.patch
Akira25: I added a description in the patch-header now.

net/gatling/patches/020-configure_features.patch
Akira25: Okay. I added one in the recent version of this PR. Thank you for the advice.

Akira25:
I plan on squashing these commits, once I got everything properly with the CI.

Is there maybe a chance, to get this package in the coming 22.03-release still?

net/gatling/Makefile
```diff
@@ -0,0 +1,40 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gatling
+PKG_VERSION:=0.16
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=https://www.fefe.de/gatling/gatling-0.16.tar.xz
```
BKPepe, line 8:
```suggestion
PKG_SOURCE_URL:=https://www.fefe.de/gatling/
```

Shouldn't this work too?

net/gatling/Makefile
```diff
@@ -0,0 +1,40 @@
...
+PKG_HASH:=5f96438ee201d7f1f6c2e0849ff273b196bdc7493f29a719ce8ed08c8be6365b
+
+PKG_BUILD_DEPENDS:=openssl px5g-mbedtls libowfat
```
BKPepe, line 11: openssl is already defined in DEPENDS, thus it can be removed here

net/gatling/Makefile
Akira25: Compile-tested it. Works fine.

net/gatling/Makefile
Akira25: adjusted in recent version.

Akira25:
Adjusted dirty-patch-failure by refreshing and recomitting the patches. Should hopefully work now.

Akira25:
@feckert @BKPepe It looks like the Testbuild for `arc_archs` (never heard of this architecture) is failing for some error in `libowfat`-package, which is not part of this PR. As it builds in all other architectures perfectly, I'd really be in favour of merging this. Would that be okay for you to?

For the error on libowfat, I'd hand over to the package maintainer @dangowrt or maybe @neheb, who fixed that error in #18716

neheb:
libiwfat issue is CFLAGS not getting passed properly. I don't know how to fix.

Akira25:
@BKPepe Could you maybe mark the change-requests as resolved? Or do you want me to make further adjustments?

net/gatling/patches/020-configure_features.patch
```diff
@@ -0,0 +1,30 @@
+gatling: configure features
+
+Disables SMB-support, as only deprecated SMBv1 is supported,
+which most users wouldn't use anyway. Avoids potential
+vulnerabilities.
+
+Activates threaded opening: Thus gatling will open files in
+threads to open kernel I/O scheduling opportunities.
```
BKPepe, line 8: Is this proposed to upstream?

net/gatling/patches/010-decrease_optimisation.patch
```diff
@@ -0,0 +1,25 @@
+gatling: decrease optimisation
+
+When compiling on MIPS-architectures, in -O2 there happen some
```
BKPepe, line 3: This can be disabled only for MIPS, right? But I still dont think that this is a good solution at all as you referenced two years old issue and it seems only one person has this. Maybe let's cc: @neheb on this one.

BKPepe (commented):
Makefile looks good! Thanks.

- What I am not sure are those patches, which are there. Does not look good to me.

Anyway, maybe there could be added file ``test.sh`` to be run tested inside CI for version. There are also some test.sh files in this repository, which you can use as reference.

BKPepe:
Also as it does not compile for arc, you can put it in depends to avoid building it there until it is solved. Not blocker from my side.

net/gatling/patches/020-configure_features.patch
Akira25: No, it isn't. The patch only tweaks Pre-processor directives in `gatling_features.h`. As far as I can see, this file is intended, to give the compiling user the ability to include or exclude functionality in the webserver. The corresponding webserver code is surrounded by `ifdefs`, so it can be activated/deactivated easily.

I consider that one mere configuration options for the user. I don't plan to bring them upstream, as poeple on x86-machines might have other demands on their special setups.

net/gatling/patches/010-decrease_optimisation.patch
Akira25: As far as I saw it on my compile test, MIPS would be the only target that needs this patch. I don't know how to apply a patch for specific target only. Would be great, if you could provide me with some instructions on that.

For the one issue 2 years ago: I could have linked some 3 website more. If the users did solve the issue at all, they mostly decreased optimization level. This site served as one example. You can find all of these by googling the linkers error.
https://www.google.com/search?q=gcc+Unsupported+jump+between+ISA+modes%3B+consider+recompiling+with+interlinking+enabled

If it helps, there's an overview on compiler-flags, that get activated by `-O2`:
https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

Akira25:
> Also as it does not compile for arc, you can put it in depends to avoid building it there until it is solved. Not blocker from my side.

What do you mean by that? How to do that?

Akira25:
@BKPepe I added a `test.sh` file. As there is no possibility to print the version number only, I used `-h` for printing the help-message instead.

Is there anything urgent left? As @neheb didn't respond for 20 days, there must be either no better approach to this or he has not the time for providing one.

This code is running on my OpenWrt-Machine since over 2 months without any issues. I would really like to get it merged soon.

neheb:
There's the ARC failure, which is a big hint that host headers are leaking in.

neheb:
Please rebase against current master.

Akira25:
> Please rebase against current master.

rebased against current master.

Akira25:
Thank you very much for your help and feedback. :)

BKPepe:
You are welcome. I tried to do my best. No worries. :)

Dne pá 23. 9. 2022 23:13 uživatel Martin Hübner ***@***.***>
napsal:

PR #18532, opened 2022-05-14
gcc: update to allow compiling different versions, based on toolchain gcc

flyn-org:
Signed-off-by: W. Michael Petullo <mike@flyn.org>

Maintainer: @noblepepper
Compile tested: x86_64

Description:
This is a proof-of-concept update that fixes issue #18529. I simply increased the version number to 8.4.0 and pulled the patches from the host GCC package (toolchain/gcc). GCC compiles again with this applied.

PR #19140, opened 2022-08-09
gcc: update to allow compiling different versions

BKPepe:
Maintainer: @flyn-org 

This is backport of https://github.com/openwrt/packages/pull/18532 for Openwrt 21.02:

```
This is based on the toolchain GCC, and aims to share as much of its
Makefile and patches with that definition. The package requires one additional patch:

(1) 003-dont-choke-when-building-32bit-on-64bit.patch, which fixes the
`error: size of array 'test_real_width' is negative` error that occurs
when building a 32-bit GCC on a 64-bit host. (Search the Internet for
examples of this error appearing.)
```

With slight modifications:
- added gcc version 9.3.0
- downgraded gcc to 10.2.0
- removed gcc version 11.3.0 as it is now in sync with OpenWrt 21.02

Compile tested only gcc 8.4.0: All Turris routers
Run tested only gcc 8.4.0: mpc85xx, Turris 1.1, powerpc8540


flyn-org:
Are the failed tests a regression when compared to the current package? If you are unconcerned with these tests, then I recommend we merge.

BKPepe:
That's a good question. GitHub actions run testing fails with this: 
``Collected errors:
 * pkg_run_script: package "gcc" postrm script returned status 1.`` and I will need to check, if this happens with the current package before merging.

BKPepe:
Confirmed that I did not break it, merging.

PR #19147, opened 2022-08-11
cgi-io: allow underscore character in filename/mimetype

ptpt52:
underscore character in filename/mimetype should be allowed

### DIFF
--- a/net/cgi-io/patches/10-allow-underscore-character-in-filename.patch
+++ b/net/cgi-io/patches/10-allow-underscore-character-in-filename.patch
@@ -1,0 +1,30 @@
+--- a/main.c
++++ b/main.c
+@@ -547,11 +547,11 @@ main_download(int argc, char **argv)
+ 		return failure(403, 0, "Requested path is not a regular file or block device");
+ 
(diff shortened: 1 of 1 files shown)


jow-:
I can't see a single media type with underscore here: http://www.iana.org/assignments/media-types/media-types.xhtml

Whats the use-case for this? Allowing underscores in filenames is fine.

ptpt52:
@jow- 

I'm just looking at theoretically to see if it's allowed, if the minitype allows underscore, it should.

I'm going to invent a mimetype with an underscore, right?

ynezz:
FYI this is not a proper place submitting changes into `cgi-io` package, the upstream is at https://git.openwrt.org/project/cgi-io.git, so [Sending patches by git send-email](https://openwrt.org/docs/guide-developer/working-with-git-email) applies here.

ptpt52:
> FYI this is not a proper place for submitting changes into `cgi-io` package, the upstream is at https://git.openwrt.org/project/cgi-io.git, so [Sending patches by git send-email](https://openwrt.org/docs/guide-developer/working-with-git-email) applies here.

done

PR #19154, opened 2022-08-12
miniupnpd: update and fix nftables variant

tiagogaspar8:
Description:

Update the package to a commit that fixes an issue with removing PCP
mappings from nftables.
This also allows us to fix the nftables miniupnpd implementation on
openwrt.
In this new implementation, a table is created at the start of miniupnpd
and it is dedicated to miniupnpd with a priority above the firewall4
table. This allows miniupnpd to go ahead of the drop rules of firewall4
and forward traffic as needed. There was the possibility of adding a
chain inside the firewall4 table, but this would raise an issue where
if firewall4 was reloaded the port forwardings would be lost and
miniupnpd could be out of sync. When miniupnpd is stopped the table is
deleted, taking the port forwardings with it.

Some of this commit is based of msylgj's work, mainly the logic of the
init/hotplug scripts and the makefile build parameters.

@stintel @BKPepe and @msylgj I'd love some feedback on this 😄 
And testing also, might work for me but not others.

Signed-off-by: Tiago Gaspar <tiagogaspar8@gmail.com>

Maintainer: N/A
Compile tested: WRT3200ACM
Run tested: WRT3200ACM


stintel:
Ping @jow-

net/miniupnpd/Makefile
```diff
@@ -8,12 +8,15 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=miniupnpd
-PKG_VERSION:=2.2.3
+PKG_VERSION:=3.0.1
```
stintel, line 11: Where did you get this version from?

tiagogaspar8: That was what I used for testing, forgot to change this.
May I propose 2.3.0-1? Since 2.3.0 is the latest tag.

stintel: We usually don't use PKG_VERSION when we build packages from a git hash. I'd just drop it.

tiagogaspar8: Done!

net/miniupnpd/files/miniupnpd.init
```diff
@@ -171,12 +172,29 @@ upnpd() {
 
 		config_foreach conf_rule_add perm_rule
 
+		if [ "Z$FW" = "Zfw4" ]; then
+			#When using nftables configure miniupnpd to use it's own table and chains
```
stangri, line 176: "its own".

tiagogaspar8: Fixed!

PR #19158, opened 2022-08-13
[21.02] gcc: bump PKG_RELEASE because of upcoming update

BKPepe:
Compile tested: it compiles just fine for OpenWrt 21.02.
Run tested: N/A

Description:
- This pull request is for testing purposes, if there is already broken postrm script or not.  (Related to https://github.com/openwrt/packages/pull/19140 )


BKPepe:
Ok, [aarch64_cortex-a53](https://github.com/openwrt/packages/runs/7817791266?check_suite_focus=true) fails with the same error:
```
No test.sh script available
Removing package gcc from root...
Collected errors:
 * pkg_run_script: package "gcc" postrm script returned status 1.
No packages removed.
Error: Process completed with exit code 255.
```

Unrelated to my update, thus merging the pull request with gcc update.

PR #19159, opened 2022-08-13
rsync: bump to 3.2.5

graysky2:
With the new OW release approaching, it might be better to get an officially tagged upstream release in as PR#19087 just contained a fix for CVE-2022-29154 which itself introduced a few bugs.

Maintainer: @mstorchak @BKPepe


graysky2:
@BKPepe - if you agree with my thinking I can make PRs for 19.07 and 21.03

net/rsync/Makefile
```diff
@@ -8,12 +8,12 @@
...
 
 PKG_NAME:=rsync
-PKG_VERSION:=3.2.4
+PKG_VERSION:=3.2.5pre2
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://download.samba.org/pub/rsync/src
-PKG_HASH:=6f761838d08052b0b6579cf7f6737d93e47f01f4da04c5d24d3447b7f2a5fad1
+PKG_SOURCE_URL:=https://codeload.github.com/WayneD/$(PKG_NAME)/tar.gz/v$(PKG_VERSION)?
```
BKPepe, line 15: I don't like this change, why we can not use https://download.samba.org/pub/rsync/src-previews/rsync-3.2.5pre2.tar.gz ?

graysky2: Initially, I did not see the pre-release on samba.org but I modified per your suggestion.

graysky2:
@BKPepe - Version 3.2.5 has been released.  Updated PR.

BKPepe:
Please avoid committing subjects as "to the latest version", because it is not good for history. Just use the exact version, and I will merge it.

graysky2:
Fixed.  Are you doing the backports to 20.02 and to 19.07 or would you like me to do them?

Neustradamus:
@graysky2, @BKPepe: rsync 3.2.5 has been released:
- https://twitter.com/neustradamus/status/1559124379241512962

Neustradamus:
Oups, you have already updated this PR...
I had not refreshed the page!

BKPepe:
> Fixed. Are you doing the backports to 20.02 and to 19.07 or would you like me to do them?

I backported it to OpenWrt 22.03 https://github.com/openwrt/packages/commit/90f35f2625f00fced493f96ff4350558fe19425f , what remains is OpenWrt 21.02 and perhaps 19.07 as well, but there are some conflicts during the merging and there is older version of rsync, so I would say that it is a good to have rather run tested this version there before pushing it.

PR #19178, opened 2022-08-15
bluld: Do not include kernel headers into user space application

hauke:
Maintainer: @ktgeek 
Compile tested: aarch64 / glibc
Run tested: no

Description:

Remove the extra include for kernel headers from this user space
application. These extra includes are causing compile errors when
compiling with glibc. User space applications should not need such
headers.

Signed-off-by: Hauke Mehrtens <hauke@hauke-m.de>

### DIFF
--- a/utils/bluld/Makefile
+++ b/utils/bluld/Makefile
@@ -16,9 +16,6 @@ include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/nls.mk
 include $(INCLUDE_DIR)/kernel.mk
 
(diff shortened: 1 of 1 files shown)


ktgeek:
I will be surprised if this compiles without the headers as this does depend on a linux kernel header.  It's been over a year since I've touched this, but maybe I used the wrong solutoin here.

The daemon uses the uleds_user_dev structure from the kernel's public header linux/uleds.h to register a user space led.  Is there a better way to get to those public interface headers?

hauke:
The kernel uapi headers are still available to the application. This removes access to in kernel headers, they should only be used by kernel modules.

PR #19181, opened 2022-08-16
usteer: bump to git HEAD

stintel:
Maintainer: @blocktrron 
Compile tested: ramips/mt7621, TP-Link EAP615-Wall, master r20335-f37a7fa4e8
Run tested: ramips/mt7621, TP-Link EAP615-Wall, master r20335-f37a7fa4e8


stintel:
Hmmm looks like my git-src symlink is causing trouble.

PR #19183, opened 2022-08-16
opendoas: fix build with PAM

paper42:
Maintainer: me
Compile tested: armv7l, Turris Omnia, OpenWrt master
Run tested: no

Description: When PAM is available, the build system will detect and use it, but the
package dependency was missing.

cc @neheb


neheb:
This did not fix it: https://downloads.openwrt.org/snapshots/faillogs/mipsel_mips32/packages/opendoas/compile.txt

PR #19192, opened 2022-08-17
host-pip-requirements: update Cython to version 0.29.28

BKPepe:
Maintainer: @commodo 
Compile tested: N/A
Run tested: N/A

Description:

I saw you are updating Cython with numpy package, which I did not. Please check this out and if it works as it should. In Turris OS packages feed, we do have [cython package](https://gitlab.nic.cz/turris/os/packages/-/blob/develop/lang/cython/Makefile) which currently fails to build with OpenWrt 22.03 as it says:

```
Please update version of Cython to 0.29.28 in 'host-pip-requirements'/
```


commodo:
Thanks.

I did not update cython host, because I have an issue with numpy on x86{_64}. It has some AVX512 assembly that it (unconditionally) tries to enable. And I haven't made time to understand the numpy build to fix it.

PR #19193, opened 2022-08-18
boost: Updates package to version 1.80.0

ClaymorePT:
Maintainer: @ClaymorePT 
Compile tested: bcm27xx bcm2711 gcc 11.3.0 musl 
Run tested: N/A

Description:
This commit updates boost to version 1.80.0

There are no new libraries in this version

The following patches were removed due to upstream update:
* 010-mips1.patch [[1]]
* 910-json-array-erase-relocate.patch [[2]]

More info about Boost 1.80.0 can be found at the usual place [[3]].

[1]: https://github.com/boostorg/config/pull/390
[2]: https://github.com/boostorg/json/pull/695
[3]: https://www.boost.org/users/history/version_1_80_0.html


ClaymorePT:
@brvphoenix okay so, regarding the PKG_SOURCE_URL, I'm following the SF convention which was set some years ago. Was this changed in the meantime? @neheb Do you have any info on this?

Regarding the commit message, I will update it. I didn't notice about the requirement.

neheb:
https://downloads.openwrt.org/snapshots/faillogs/mips64_octeonplus/packages/pdns-recursor/compile.txt fails as a result.

ClaymorePT:
Yeahh, that's unfortunate. I'll need to take a look at that. I was hoping that the mips update present in upstream would be enough but it seems something is still incomplete...

ClaymorePT:
@neheb restored mips patch in https://github.com/openwrt/packages/pull/19221

PR #19205, opened 2022-08-18
qemu: enable usb passthrough by default

corna:
The total uncompressed size increase (qemu-system-x86_64 and the new libusb dependency) is less than 1%.

Maintainer: @yousong
Compile tested: built with the current SDK docker image on x86_64 for x86_64
Run tested: `qemu-system-x86_64 -device help` now returns `usb_host` and I can correctly pass a physical device to the VM.

### DIFF
--- a/utils/qemu/Makefile
+++ b/utils/qemu/Makefile
@@ -261,6 +261,7 @@ config QEMU_UI_SPICE
 
 config QEMU_DEV_USB
 	bool "QEMU USB passthrough support"
+	default y
 
 config QEMU_SECCOMP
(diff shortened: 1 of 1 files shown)


corna:
@yousong Any update on this? I've been using this for a while now and it works flawlessly

supersebbo:
I could have sworn I submitted a PR for a same a long time ago.  I agree this needs to be default, the package size impact is minimal and anyone running virtualization on an OpenWRT host is highly unlikely to be concerned about disk space.   Please merge.

PR #19221, opened 2022-08-22
boost: restores patch for libboost_context for mips64

ClaymorePT:
Maintainer: @ClaymorePT 
Compile tested: octeonplus mips64 using musl
Run tested: N/A

Description:

This commit restores the patch for libboost_context for mips64 given
that the upstream update added in 1.80.0 does not fully fixes the issue.

### DIFF
--- a/libs/boost/patches/010-mips1.patch
+++ b/libs/boost/patches/010-mips1.patch
@@ -1,0 +1,38 @@
+--- a/boostcpp.jam
++++ b/boostcpp.jam
+@@ -634,7 +634,7 @@ rule address-model ( )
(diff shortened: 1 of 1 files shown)


neheb:
ping @FlyGoat know what's going on here?

FlyGoat:
Cc @wzssyqa as well.
So the problem is boost::context didn't compiled into booster package?

neheb:
Specifically for mips64.

wzssyqa:
The reason is due to that https://github.com/boostorg/boost/pull/540 was forgot to merge before the release.
So a release of boost is broken.

PR #19238, opened 2022-08-23
python-pycares: add new package

dynasticorpheus:
Maintainer: @dynasticorpheus 

Compile / Run tested: Linksys_wrt32x, mvebu (cortex-a9), OpenWrt 22.03.0-rc6

Description: Add new package [pycares](https://pypi.org/project/pycares/)

### DIFF
--- a/lang/python/python-pycares/Makefile
+++ b/lang/python/python-pycares/Makefile
@@ -1,0 +1,38 @@
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=python-pycares
+PKG_VERSION:=4.2.2
+PKG_RELEASE:=$(AUTORELEASE)
+
+PYPI_NAME:=pycares
+PKG_HASH:=e1f57a8004370080694bd6fb969a1ffc9171a59c6824d54f791c1b2e4d298385
+
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=Fabian Lipken <dynasticorpheus@gmail.com>
+
+PKG_BUILD_DEPENDS:=libffi/host
+
+HOST_PYTHON3_PACKAGE_BUILD_DEPENDS:=cffi  # cffi>=1.5.0
+
+include ../pypi.mk
+include $(INCLUDE_DIR)/package.mk
+include ../python3-package.mk
+
+define Package/python3-pycares
+  SECTION:=lang
+  CATEGORY:=Languages
+  SUBMENU:=Python
+  TITLE:=Python interface for c-ares
+  URL:=https://github.com/saghul/pycares
+  DEPENDS:=+python3-light +python3-cffi
+endef
+
+define Package/python3-pycares/description
+Python interface for c-ares
+endef
+
+$(eval $(call Py3Package,python3-pycares))
+$(eval $(call BuildPackage,python3-pycares))
+$(eval $(call BuildPackage,python3-pycares-src))


dynasticorpheus:
cc python maintainers: @commodo , @jefferyto

lang/python/python-pycares/Makefile
```diff
@@ -0,0 +1,48 @@
...
+define Package/python3-pycares
+  SECTION:=lang
+  CATEGORY:=Languages
+  SUBMENU:=Python
+  TITLE:=Python interface for c-ares
+  URL:=https://github.com/saghul/pycares
+  DEPENDS:=+python3-light
+endef
+
+define Package/python3-pycares-src
```
commodo, line 32: the `define Package/python3-pycares-src` definition is redundant;

you get it for free when you define `Package/python3-pycares`

you can remove it;

lang/python/python-pycares/Makefile
```diff
@@ -0,0 +1,48 @@
...
+endef
+
+define Package/python3-pycares/description
+Python interface for c-ares
+endef
+
+$(eval $(call Py3Package,python3-pycares))
+$(eval $(call BuildPackage,python3-pycares))
+$(eval $(call BuildPackage,python3-pycares-src))
+
```
commodo, line 48: since i'm complaining anyway: you can remove this empty last-line; but it's more of a nitpick (or personal feeling)

feel free to disregard it;

lang/python/python-pycares/Makefile
```diff
@@ -0,0 +1,48 @@
...
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=python-pycares
+PKG_VERSION:=4.2.2
+PKG_RELEASE:=1
```
commodo, line 6: depends on who you ask, you can also just do `PKG_RELEASE:=$(AUTORELEASE)`

i'm not sure how robust the mechanism is, but i use it;
it makes my life simple when updating a package, as there is one less line to edit; 

otherwise, the rule is: when updating to a new version, PKG_RELEASE gets reset to 1; otherwise, if something gets changed/updated, you need to bump PKG_RELEASE, which (as it sounds) is a pretty manual and (from experience) a bit of an error-prone process;

commodo:
rest looks good to me;

i'm assuming you tested with just `python3-light`

lang/python/python-pycares/Makefile
dynasticorpheus: Thanks, I'm happy to follow your lead here so updated accordingly next to the other suggestions

commodo (approved):

commodo:
will merge this in 1-2 days if no objection

dynasticorpheus:
> will merge this in 1-2 days if no objection

Sure no problem, did add python3-cffi dependency just now based on additional bare install testing. Apologies for the force-push mess.

commodo:
Force-pushing is fine.

Since you're doing it, might as well just squash commits into a single one :)

dynasticorpheus:
> Force-pushing is fine.
> 
> Since you're doing it, might as well just squash commits into a single one :)

I learned something today! :)

commodo:
> > Force-pushing is fine.
> > Since you're doing it, might as well just squash commits into a single one :)
> 
> I learned something today! :)

well, (often in various projects) the idea is that fix-commits are not needed until they are merged;
once something is in master (or merged), only then fix-commits are more acceptable;
up until then, it's WIP

this is because (over time), nobody will care (when looking into the git history) about stuff that was fixed before it was in master;
also, i only search for fix-commits (in git history), if i find something broken (that has appeared in master or a release)

but, each repo/project has different rules

PR #19252, opened 2022-08-25
vim: update to 9.0

PolynomialDivision:
Release Notes:
https://www.vim.org/vim90.php

Maintainer: @ratkaj
Runtested: Linksys E8450


neheb:
Personally I'd rather see neovim here but 🤷

PolynomialDivision:
> Personally I'd rather see neovim here but shrug

Why don't you add it? Just add a PROVIDES:=vim to both packages?

jmarcet:
I've had to add two deps to build vim-fuller:

```
diff --git a/utils/vim/Makefile b/utils/vim/Makefile
index a88759bd2..0c91775f8 100644
--- a/utils/vim/Makefile
+++ b/utils/vim/Makefile
@@ -48,6 +48,7 @@ endef
 
 define Package/vim-fuller
   $(call Package/vim/Default)
+  DEPENDS:=+libncursesw +libsodium
   TITLE+= (Big)
   PROVIDES:=vim vim-full
   CONFLICTS:=vim vim-full
```

PolynomialDivision:
@jmarcet Can you try again? I also tried to add `--without-tlib=ncursesw` but this did not work. Do you have some special config that uses ncursesw?

https://www.linuxfromscratch.org/blfs/view/svn/postlfs/vim.html

PR #19254, opened 2022-08-25
openvpn-2.5.7: explicitly disable engine parameter for openssl variant

AuthorReflex:
Maintainer: me / @mkrkn 
Compile tested: ath79/generic, ramips/mt7620, ramips/mt7621
Run tested: ath79/Archer C7v2, mt7620/Xiaomi Mi-Mini, mt7621/Xiaomi Mi Router 3 Pro

Engine support is deprecated in OpenSSL 3.0 and for OpenSSL 3.0 the default
is to disable engine support as engine support is deprecated. For ath79 architecture
build with autodetection engine support fails, so explicitly set off for now.

### DIFF
--- a/net/openvpn/Makefile
+++ b/net/openvpn/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=openvpn
 
 PKG_VERSION:=2.5.7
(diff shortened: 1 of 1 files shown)


mkrkn:
Good idea, ack from me.

PR #19258, opened 2022-08-26
restic: update to 0.14.0

Torxgewinde:
Maintainer: Tom Stöveken <tom@naaa.de>
Compile tested: SDK for OpenWrt 21.02.3
Run tested: xRX200 rev 1.2, AVM FRITZ!Box 7360 V2, OpenWrt 21.02.3

Description:
Updated to version 0.14.0

Signed-off-by: Tom Stöveken <tom@naaa.de>


ahuse:
Thanks for maintaining this package. Is there a reason why the pull requests only for the master-branch and not openwrt-22.03? Would be nice to have the latest restic version in the latest stable openwrt release as well and not only in snapshots.

Torxgewinde:
There is no other reason than timing. 22.03 was branched before I updated the package. To get restic 0.14.0 on a 22.03 device there are two options:

- download the *.ipk files from snapshots and install with opkg. As example binaries for x86 are here: https://downloads.openwrt.org/snapshots/packages/x86_64/packages/restic_0.14.0-1_x86_64.ipk Go binaries are quite easy to exchange, because they link a lot statically into the binary.
- AFAIK we could also cherry pick restic from main to 22.03 branch. But this requires some of my time and learning if I do it since I have not done it before. Perhaps someone experienced can do that before I get to it.

ahuse:
Sorry for the late reply. For cherry-picking (best option) the version jump is maybe too high according to the [contributing guidelines](https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md).

For now, i am just using `restic self-update` to get the latest version.

Issue #15060, opened 2021-03-08
podman: should require runc package


flyn-org:
Maintainer: @dangowrt
Environment: x86_64 master

Description:
I installed the podman package, but I found that running `podman pull alpine` produced the error "Error: no valid executable found for OCI runtime /usr/sbin/runc: invalid argument." I installed the runc package, and then the `pull` succeeded. Should the podman package depend on the runc package?

dangowrt:
Actually that still doesn't make it a strict dependency, as one can use other runtime environments as well. I'm working on `uxc` to be suitable for that (it's not yet).

dangowrt:
Now we got both `runc` and `crun` in available with https://github.com/openwrt/packages/commit/01d1a4969e258f7b08a1284e30bfa9bceba62b9c , so users can choose.

paper42:
@dangowrt @oskarirauta Do you think it would be possible to make podman depend on crun (because it's 18x smaller than runc) and optionally on runc? Right now users have to see the error and manually install either runc or crun which isn't ideal.

```
size    package
 169442 crun_0.20-1_arm_cortex-a9_vfpv3-d16.ipk
3176827 runc_1.0.1-1_arm_cortex-a9_vfpv3-d16.ipk
```

PR #19266, opened 2022-08-27
podman: add crun dependency

marcov:
Signed-off-by: Marco Vedovati <mv@sba.lat>

Maintainer: @oskarirauta 

Compile tested: ---
Run tested: ---

Description: podman needs crun to spawn containers, so explicitly add it as package dependency

### DIFF
--- a/utils/podman/Makefile
+++ b/utils/podman/Makefile
@@ -41,7 +41,7 @@ define Package/podman
   CATEGORY:=Utilities
   TITLE:=Podman
   URL:=https://podman.io
-  DEPENDS:=$(GO_ARCH_DEPENDS) +conmon +cni +cni-plugins +btrfs-progs +glib2 +gnupg2 +uci-firewall +libgpg-error +libseccomp +libgpgme +nsenter +zoneinfo-simple +kmod-veth +catatonit +PODMAN_SELINUX_SUPPORT:libselinux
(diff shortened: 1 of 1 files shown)


paper42:
relevant discussion: #15060

oskarirauta:
It does not explicitly need crun, you can also use runs, and others.

marcov:
> It does not explicitly need crun, you can also use runs, and others.

The default config file installed with podman requires `crun`:
```console
# podman info
Error: default OCI runtime "crun" not found: invalid argument
```

oskarirauta:
True. But I left it out as requirement, because other options are available and it would be stupid to require crun, just because default options offer it. Maybe I should instead change the configuration file to not define crun as default? It is more or less a example configuration file, surely you’ve made your own changes to your system too..

This was discussed ages ago and requiring crun was decided to be unnecessary due to that openwrt is a minimal distribution and requiring package that is not a real requirement or possibly would not be used was the reason to exclude it from dependencies.

marcov:
I understand you want to keep the footprint to a minimum, but podman already requires 10+ packages installed, so in this case it sounds like a weak argument. Especially since you need a runtime installed to use podman, and `crun` is the smallest runtime around.

BTW, to be coherent with the minimal footprint philosophy, `catatonit` should also be ditched: podman runs just fine without it.

marcov:
> Only if you have alternative to catatonic

Nope, you can run containers just fine without any init. 

Thanks for taking the time to answering :-)

PR #19269, opened 2022-08-27
schroot: update to version 1.6.13, with new maintainer

jmarcet:
Maintainer: me
Compile tested: master x86_64
Run tested: master x86_64

Description:
New maintainer, new homepage and cmake based instead of autotols.

The forked project has changed its name to `reschroot`, I've kept the package name in the tree since the executable is still called `schroot`.

Also I doubted the init script, since we now can have a persistent `/var`, I was not sure whether to add a check to decide whether to install the init script or install the contents directly.


neheb:
Need to run make package/schroot/{clean,refresh} looks like

BKPepe:
Also, maintainer could be somehow confusing with OpenWrt maintainer. You should use something like switch to fork/update to fork, etc.

jmarcet:
@neheb @BKPepe Better now? I used the description from the new web.

stangri:
@jmarcet Is there's ever a plan to backport this to a stable release (where people may have the old `schroot` package already installed)? If so, would a `CONFLICTS` statement in a package definition make sense?

jmarcet:
@stangri The external package name (within OpenWRT) is the same, as is the main executable, so externally it is just an update. The same goes for the version number. Hence I don't think a `CONFLICTS` statement makes sense in this case.

admin/schroot/Makefile
```diff
@@ -1,23 +1,25 @@
...
-PKG_RELEASE:=3
+PKG_NAME:=reschroot
+PKG_VERSION:=1.6.13
+PKG_RELEASE:=1
 
-PKG_SOURCE:=$(PKG_NAME)_$(PKG_VERSION).orig.tar.xz
-PKG_SOURCE_URL:=http://deb.debian.org/debian/pool/main/s/schroot
-PKG_HASH:=3ce8dfd9cb97b099e4b6d4ccec421d6cc8c9ef84574681e928a12badb5643d0b
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeberg.org/shelter/reschroot/archive/release
```
BKPepe, line 8: Uhh, at first glance, it looks like this repository and the other one on Debian has the same recent commits of the same author, but when looking here https://salsa.debian.org/debian/schroot/-/commit/37f5af211fedc939593f6dbb2992f9d8bbaa866b he is referring to codeberg, thus I it is alright.

admin/schroot/Makefile
```diff
@@ -1,23 +1,25 @@
...
 PKG_LICENSE:=GPL-3.0-only
 PKG_LICENSE_FILES:=COPYING
 
-PKG_INSTALL:=1
-PKG_BUILD_PARALLEL:=1
-PKG_FIXUP:=autoreconf
+CMAKE_INSTALL:=1
+
+CMAKE_BINARY_SUBDIR:=build
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
```
BKPepe, line 18: @neheb this is ok for you? If yes, then it can be merged.

neheb: It is not.

neheb: Maybe

```
PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)/$(PKG_NAME)
```

PR #19271, opened 2022-08-28
xz: update to 5.2.6

PolynomialDivision:
Release Notes:
https://git.tukaani.org/?p=xz.git;a=blob;f=NEWS;hb=HEAD

Maintainer: noone?


BKPepe:
Is this run tested or compile tested as it was removed from the template?

PR #19275, opened 2022-08-28
gnupg: update to 2.2.39

PolynomialDivision:
Update to latest LTS version.

Maintainer: @dangowrt 
Compile tested: tbd
Run tested: tbd


dangowrt:
Please squash the *update to ...* commits into a single one, imho in this case there is no useful extra information added in the intermediate version steps.

PolynomialDivision:
Done

neheb:
I see two commits.

PolynomialDivision:
The second adds a pkg-cpe-id.

PR #19280, opened 2022-08-29
python-pycares: add new package

dynasticorpheus:
Signed-off-by: Fabian Lipken <dynasticorpheus@gmail.com>

Description:


dynasticorpheus:
@commodo  Hope this is correct way to cherry-pick into openwrt-22.03 branch

commodo:
so, i don't know the policy about adding new packages into releases (at this point in time)

that's for someone else to answer; usually in releases (after they get created) no new major stuff gets in;

from my side it looks ok;
though, some people would prefer to see a `git cherry-pick -x` so that the original commit shows up in the commit;

BKPepe:
> though, some people would prefer to see a git cherry-pick -x so that the original commit shows up in the commit;

I am one of them! :laughing: 
Added.

> so, i don't know the policy about adding new packages into releases (at this point in time)

It is not a big deal right now, because final version was not tagged yet and even if we would not make it, then it can be added later and then it could be part of fixup release even though it is not recommended.

dynasticorpheus:
@commodo @BKPepe Thanks for guidance on this. I have force-pushed the "git cherry-pick -x" variant

PR #19293, opened 2022-09-02
treewide: compatibility fixes for new GNU Autotools

mpratt14:
following up on https://github.com/openwrt/openwrt/pull/10394

these packages require specific changes in order to build with new versions of Autoconf and Automake

all the new patches introduced are backports


net/nfdump/Makefile
```diff
@@ -40,7 +40,9 @@ define Package/nfdump/conffiles
 endef
 
 TARGET_CFLAGS += -flto
+
 TARGET_LDFLAGS += -Wl,--as-needed $(if $(CONFIG_USE_MUSL),-lfts)
+TARGET_LDFLAGS += -L$(STAGING_DIR)/usr/lib
```
neheb, line 45: Isn't this added by default?

mpratt14: as far as I can tell, no. I only see the toolchain library directory. It would not build on my system until I did this.

mpratt14: I'm going to take a deeper look really quick because I'm not 100% certain what makes this necessary now...

mpratt14: nfdump uses macro

`AC_PROG_LEX(yywrap)`

and autoconf has these changes (second one partially reverts the first)

https://git.savannah.nongnu.org/cgit/autoconf.git/commit/?id=29ede6b96feee29c0c477d1659081bbdb82cd8b3
https://git.savannah.nongnu.org/cgit/autoconf.git/commit/?id=46f384f850b455cdb366bafbcf4e992606b3f2c0

mpratt14: uci has the same thing

```
$ git grep TARGET_LDFLAGS | grep usr
package/libs/openssl/Makefile:  [ -n "$(TARGET_LDFLAGS)" ] && $(SED) 's#$(TARGET_LDFLAGS)##g' $(1)/usr/lib/pkgconfig/{openssl,libcrypto,libssl}.pc || true
package/system/uci/Makefile:TARGET_LDFLAGS += -L$(STAGING_DIR)/usr/lib
rules.mk:    TARGET_LDFLAGS+= -L$(TOOLCHAIN_DIR)/usr/lib -L$(TOOLCHAIN_DIR)/lib
```

neheb: that...sounds like a bug.

mpratt14:
updated commit messages

net/nfdump/Makefile
mpratt14: do you think `-L$(STAGING_DIR)/usr/lib` should be in rules.mk? I kind of had that thought but didnt want to bother with it

neheb: I thought it already was. Weird.

neheb: Gonna ask @jow- about this.

jow-:
We patch the gcc specs to make `${getenv("STAGING_DIR")}/use/lib` a default search path, see `scripts/patch-specs.sh`, invoked by toolchain/gcc/final/Makefile

mpratt14:
> We patch the gcc specs to make `${getenv("STAGING_DIR")}/usr/lib` a default search path, see `scripts/patch-specs.sh`, invoked by toolchain/gcc/final/Makefile

thanks I had no idea, I tried removing and it still works

I didn't realize it at first, but I did make an assumption. Before I figured out how to add `InstallDev` to flex I tried using `$(STAGING_DIR_ROOT)` and I had to add that to TARGET_LDFLAGS for it to build, so I thought I would equally need `$(STAGING_DIR)` in TARGET_LDFLAGS after flex `InstallDev`, and I didn't see it in TARGET_LDFLAGS already, or in the build logs, and I saw uci with that line as well...

mpratt14:
if #19308 gets merged first, then the patch for boinc here is not needed

mpratt14:
also

conserver can be bumped instead of patched

minidlna *literally just* had a release right after I prepared that commit, so it can also be updated instead

is there any preference one way or the other?

neheb:
Update

PR #19302, opened 2022-09-03
gnupg: add PKG_CPE_ID

PolynomialDivision:
Add PKG_CPE_ID:=cpe:/a:gnupg:gnupg to allow scanning for CVEs.


neheb:
why not fold in the other gnupg PR?

PolynomialDivision:
> why not fold in the other gnupg PR?

That other was a version bump and that is security related. So I thought I would do two seperate realted PRs. Should I change it?

neheb:
Might as well. I don't think such a change deserves its own PR.

PolynomialDivision:
Okay, I added it now to the other PR.

PR #19308, opened 2022-09-04
boinc: Update to 7.20.2

AenBleidd:
Signed-off-by: Vitalii Koshura <lestat.de.lionkur@gmail.com>

### DIFF
--- a/net/boinc/Makefile
+++ b/net/boinc/Makefile
@@ -6,14 +6,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=boinc
-PKG_VERSION:=7.16.16
+PKG_VERSION:=7.20.2
 PKG_VERSION_SHORT:=$(shell echo $(PKG_VERSION)| cut -f1,2 -d.)
 PKG_RELEASE:=1
 
-PKG_SOURCE_DATE:=2020-02-25
+PKG_SOURCE_DATE:=2022-09-04
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/BOINC/boinc/tar.gz/client_release/$(PKG_VERSION_SHORT)/$(PKG_VERSION)?
-PKG_HASH:=0d5656a9f8ed1048936a5764270848b892d63f27bdb863d0ace447f1eaae6002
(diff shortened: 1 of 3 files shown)


mpratt14:
adding yourself as a maintainer should be done in a separate commit and should be approved by the other maintainers

for just bumping the version, editing copyright years is not necessary

mpratt14:
ping @N30dG 

updating boinc would make the patch in my PR for autotools not necessary

AenBleidd:
@mpratt14, any way to get why patch was applied not clean enough? I can't reproduce this locally on regular linux

mpratt14:
it's hard to catch with so much output from the CI, but it's this

`Refreshed patch 002-hosttypeRespected`

keyword is "refreshed" meaning some lines are offset and it needs to be refreshed

there is a build-target for automatically refreshing patches

`make feeds/packages/net/boinc/refresh`

but I prefer to do it manually

`make feeds/packages/net/boinc/prepare QUILT=1 V=s`

cd into the build_dir of boinc and

`quilt push -f`
`quilt refresh`

until all patches processed then copy that patches directory back to openwrt source patches directory
(but without the "series" file)

AenBleidd:
This seems to be more complicated than I thought before. I'm closing this for now and will open a new PR once I have all this done on my fork. @mpratt14, thanks for the hints.

mpratt14:
@AenBleidd please dont open a new PR, the members dont like that

if you need help with Git or quilt just ask 👍🏼 

basically all you need to do is refresh the patch with `quilt`

test build of boinc is simple:

`make feeds/packages/net/boinc/clean feeds/packages/net/boinc/compile`

all you need to do is run the prepare stage alone with quilt

`make feeds/packages/net/boinc/clean feeds/packages/net/boinc/prepare QUILT=1`

then cd to the build directory

```
cd build_dir
cd target*
cd boinc*
```

and run quilt

```
quilt push -f
quilt refresh
quilt push -f
quilt refresh
....
...
```
then copy the patches in `patches` directory back to `net/boinc/patches` in openwrt/packages

mpratt14:
I don't remember which way works but, you may have to use the symlink as the make target instead of direct path

`make package/feeds/packages/boinc/clean package/feeds/packages/boinc/prepare QUILT=1`

PR #19310, opened 2022-09-04
hping3: add new package

patrakov:
The new package would help measuring one-way delays using ICMP type 13
packets. This is important for various scripts that automatically adjust
CAKE shaper bandwidth based on the observed bufferbloat. They need to
understand whether the delay is on the way up or on the way down, so
that they can adjust the bandwidth of the proper part of the shaper.

https://forum.openwrt.org/t/cake-w-adaptive-bandwidth-historic/108848
https://forum.openwrt.org/t/cake-w-adaptive-bandwidth/135379

Signed-off-by: Alexander E. Patrakov <patrakov@gmail.com>

Description:


patrakov:
I have refreshed the patches, please rerun the CI.

Issue #19287, opened 2022-08-31
nextdns: nextdns service lost after firmware update using image builder


MatMaul:
Maintainer: @rs 

Environment: ipq8068, Netgear R7800, 22.03-rc6

Description:

After updating with a new firmware with nextdns pre-installed (built with the image builder), nextdns service is not available anymore, and I need to run `nextdns install` manually.
I think this is because nextdns init file is installed by `nextdns install` and doesn't live in the openwrt package itself, and `/etc/init.d` is not part of the the path that are restored after a firmware update.

PR #19315, opened 2022-09-05
nextdns: initialize nextdns from /etc/uci-defaults

fantom-x:
Maintainer: @rs 

When nextdns is prepackaged into firmware, the post install section does not run. This PR moves that logic into /etc/uci-defaults/nextdns to cover both cases.

The same as https://github.com/openwrt/packages/pull/19045 , but addresses the comments in the original MR.

Signed-off-by: Marc Benoit <marcb62185@gmail.com>

### DIFF
--- a/net/nextdns/Makefile
+++ b/net/nextdns/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nextdns
 PKG_VERSION:=1.37.11
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=nextdns-$(PKG_VERSION).tar.gz
 PKG_SOURCE_VERSION:=v$(PKG_VERSION)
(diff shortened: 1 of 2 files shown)


hnyman:
The uci-defaults approach differs from the normal practice, so I wonder about that a bit.

In any case, the uci-defaults script would need to have a proper "exit 0" statement, so that it surely does not run again at the next boot.

fantom-x:
Added `exit 0`. Yeah, the package does not follow the normal practices: it does NOT provide any configuration or init scripts via the repo. It is all generated by the binary on `nextdns install` and the init script has to be enabled explicitly.

rs:
LGTM

MatMaul:
Amazing thanks, closing #19287.

MatMaul:
is it possible to backport that to 22.03 ?

fantom-x:
> is it possible to backport that to 22.03 ?

That is the plan, but I am gonna let it settle in `master` for a few weeks first.

Issue #19292, opened 2022-09-02
lxc-auto: errors during an Image Builder build


fantom-x:
Maintainer: @ratkaj
Environment: 22.03-SNAPSHOT x86_64

Getting these errors when using Image Builder.

```
Finalizing root filesystem...
/build/openwrt-imagebuilder/build_dir/target-x86_64_musl/root-x86/etc/init.d/lxc-auto: line 3: /lib/functions.sh: No such file or directory
Enabling boot
Enabling cgroupfs-mount
Enabling collectd
Enabling cron
Enabling ddns
Enabling dnsmasq
Enabling done
Enabling dropbear
Enabling firewall
Enabling fstab
Enabling gpio_switch
Enabling led
Enabling lm-sensors
Enabling log
Enabling luci_statistics
./etc/init.d/lxc-auto: line 3: /lib/functions.sh: No such file or directory
Enabling lxc-auto
Enabling network
Enabling nlbwmon
Enabling odhcpd
Enabling rpcd
Enabling sqm
Enabling sysctl
Enabling sysfixtime
Enabling sysfsutils
Enabling sysntpd
Enabling system
Enabling ucitrack
Enabling uhttpd
Enabling umount
Enabling urandom_seed
Enabling urngd
```

BKPepe:
Which exact commit hash are you using for packages repository? This was already fixed here, so I am thinking that you are just using older HEAD.

fantom-x:
I do not know the commit for the package repo, but `OPENWRT_RELEASE="OpenWrt 22.03-SNAPSHOT r19677-ae1786e543"`

Looking at the latest source code in both master & 22.03 and they both have the same issue:

https://github.com/openwrt/packages/blob/master/utils/lxc/files/lxc-auto.init
https://github.com/openwrt/packages/blob/openwrt-22.03/utils/lxc/files/lxc-auto.init

PR #19316, opened 2022-09-05
lxc-auto: made init script compatible with image builder

fantom-x:
Maintainer: @ratkaj @BKPepe 

Fixes this message from image builder:
```
Finalizing root filesystem...
/build/openwrt-imagebuilder/build_dir/target-x86_64_musl/root-x86/etc/init.d/lxc-auto: line 3: /lib/functions.sh: No such file or directory
Enabling boot
Enabling cgroupfs-mount
Enabling collectd
```
Closes https://github.com/openwrt/packages/issues/19292

Signed-off-by: Marc Benoit <marcb62185@gmail.com>

### DIFF
--- a/utils/lxc/files/lxc-auto.init
+++ b/utils/lxc/files/lxc-auto.init
@@ -1,6 +1,6 @@
 #!/bin/sh /etc/rc.common
 
-. /lib/functions.sh
+. "${IPKG_INSTROOT}"/lib/functions.sh
 
 START=99
 STOP=00


utils/lxc/files/lxc-auto.init
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh /etc/rc.common
 
-. /lib/functions.sh
+. "${IPKG_INSTROOT}"/lib/functions.sh
```
BKPepe, line 3: Are you sure that this https://github.com/openwrt/packages/blob/master/utils/lxc/Makefile#L65 does not help for you?  That's why I asked for your commit hash of the packages feed, which you did not provide me neither the LXC Makefile.

Because it was fixed by https://github.com/openwrt/packages/commit/039912dec5d3ba2b0f6f53ab8330ab9fea2f7adf

BKPepe (changes requested):

utils/lxc/files/lxc-auto.init
fantom-x: That is a very different failure: postinstall does NOT run by image builder.

I did provide a hash of OpenWrt to show that it is the latest with the latest of packages. Either way:
```
OPENWRT_RELEASE="OpenWrt 22.03-SNAPSHOT r19691-39c115028a"
packages: commit 85e6f9348a3df7ec7d8136476b688b56fea7ad0a (HEAD -> openwrt-22.03, origin/openwrt-22.03)
```
I just ran a image builded build again and got this:

```
Finalizing root filesystem...
/build/openwrt-imagebuilder/build_dir/target-x86_64_musl/root-x86/etc/init.d/lxc-auto: line 3: /lib/functions.sh: No such file or directory
Enabling boot
Enabling cgroupfs-mount
Enabling collectd
Enabling cron
Enabling ddns
Enabling dnsmasq
Enabling dockerd
Enabling done
Enabling dropbear
Enabling firewall
Enabling fstab
Enabling gpio_switch
Enabling led
Enabling lm-sensors
Enabling log
Enabling luci_statistics
./etc/init.d/lxc-auto: line 3: /lib/functions.sh: No such file or directory
Enabling lxc-auto
```

fantom-x: BTW, the comment in the commit you referenced explicitly states that the image builder build is not fixed.

```
Sourcing /lib/functions.sh is not needed, as /etc/rc.common does so
already. Unfortunately removing that line from the init script is not
enough to fix the problem. The postinst script should also check
IPKG_INSTROOT. As these two changes are unrelated, they should go in
separate commits, and the solution to the image build problem is to
revert the commit that introduced the breakage.```

Comment (fantom-x) [reply utils/lxc/files/lxc-auto.init:3]: Try this:

```
wget https://downloads.openwrt.org/releases/22.03-SNAPSHOT/targets/x86/64/openwrt-imagebuilder-22.03-SNAPSHOT-x86-64.Linux-x86_64.tar.xz
tar zxf openwrt-imagebuilder-22.03-SNAPSHOT-x86-64.Linux-x86_64.tar.xz
cd openwrt-imagebuilder-22.03-SNAPSHOT-x86-64.Linux-x86_64
make image PROFILE=generic BIN_DIR=./firmware FILES=./files PACKAGES="lxc-auto"
```

fantom-x: BTW, the same issue will be happening if you just build a firmware with `lxc-auto` built in.

fantom-x: Here: https://openwrt.org/docs/guide-developer/procd-init-scripts#init_scripts_during_compilation

fantom-x: @BKPepe, are there still open questions?

fantom-x: @BKPepe , do you still have concerns about this change?

PR #19320, opened 2022-09-06
miniupnpd: rework firewall4 integration

jow-:
Maintainer: currently unassigned
Compile tested: r20484+1-6790e1a892 x86/64 QEMU, r20484+1-6790e1a892 mt7621 er-x
Run tested: mt7621, er-x

Description from the commit:

 - Bump to the latest Git version in order to increase the package version
   for simpler opkg upgrade of the broken version

 - (Re-)Introduce PKG_RELEASE into the package, omitting it may lead to
   opkg segmentation faults under certain circumstances

 - Utilize automatic include hooks to drop the isolated miniupnpd table
   in favor to chains within the main inet fw4 table, otherwise PCP is
   unreliable as the upnp table might accept traffic which is later
   rejected by fw4

 - Install a fw4 script hook to restart miniupnpd on fw4 restarts and
   reloads in order to repopulate the upnp chains with forward rules

 - Register the used miniupnpd configuration file and the firewall uci
   configuration as change sources, otherwise `/etc/init.d/miniupnpd reload`
   has no effect if the firewall or upnpd config was changed

Discussion:

While using a separate nft table is cleaner in theory, it has a number of practical issues, mainly the fact that `ACCEPT` verdicts for flows are not carried over from one table to another. This means that when the upnp table accepts certain traffic, e.g. due to
PCP staged rules or uPNP port forwards, the traffic might still end up getting dropped by generic fw4 policies. One solution proposal for this was making fw4 unconditionally accept flows whose related conntrack is in DNAT state, but that would only solve uPNP port forward cases, not PCP forward allow rules that do not involve any DNAT or SNAT.

Another consideration is that `fw4 restart` might change its behavior to flush the *entire* ruleset and not just the fw4 table in the future, in which case we would need to re-introduce rule-restoration logic anyway, regardless of a potentially separate upnp table.

Due to this, I decided to rework the firewall integration approach to mostly mirror what we did with iptables in the past, but using
modern include facilities provided by fw4 - means the chain declarations and jump rules are static nft snippets which we include into appropriate places of the ruleset. What remains in the script include is a simple restart of the miniupnpd service to ensure
that firewall reloads or restarts lead to a reinitialization of the miniupnpd service and the restoration of nft rules from the leasefile.

### DIFF
--- a/net/miniupnpd/Makefile
+++ b/net/miniupnpd/Makefile
@@ -11,10 +11,12 @@ PKG_NAME:=miniupnpd
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/miniupnp/miniupnp.git
-PKG_SOURCE_DATE:=2022-08-06
-PKG_SOURCE_VERSION:=fa42d8f9316bf9c1ca14317e5a6e0d4a21365629
-PKG_MIRROR_HASH:=06662c7cf8f553f625cd968d12ea732db4193706510ed0db6e8bdd1c6b935c50
+PKG_SOURCE_DATE:=2022-08-31
+PKG_SOURCE_VERSION:=68c8ec508a421f4f4af67a63e3eb6f497d2531e1
+PKG_MIRROR_HASH:=68a3170ec73149c4cf4855b1ce6e031557cc12bff85a58421bb94785daaf225d
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(BUILD_VARIANT)/$(PKG_NAME)-$(PKG_VERSION)/miniupnpd
+
+PKG_RELEASE:=1
 
 PKG_MAINTAINER:=
 PKG_LICENSE:=BSD-3-Clause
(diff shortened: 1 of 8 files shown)


jow-:
@tiagogaspar8 - pinging you since you did the rework towards a separate table. I don't want to erase your efforts without prior discussion

jow-:
Since there's been no negative feedback I decided to merge it for now so that we can pick it into 22.03 to have working version there again. If we find a cleaner approach we can still partially revert this and/or rework it again

tiagogaspar8:
Hi @jow-! sorry for the time I took to see your mention.
I can see your point and had no idea of the issues you mentioned, I thought that nftables would follow the chain of rules until it found one that matched and apply it.
Either way, this seems fine to me, but I have some remarks:

1. I see that you added an include in the firewall config, but can't this be achieved some other way? Cause for example if one builds an image with miniupnpd, and then another without it, it will leave residues behind in the firewall configuration file.
2. When you say it restarts with fw4 and restores with the forward rules, are you including PCP? because those aren't stored in a lease file and are instead written directly to the chains, with no tracking whatsoever

I'm yet to test this, but have you tested this with IPv6?
I'm building a fw right now and soon I'll test this 😄

tiagogaspar8:
Hi @jow- , the restoration system works only for the IPv4 port forwards and not PCP port mappings.
My proposal, is that at miniupnpd stop the rules are dumped into a file in /tmp and when started, if there is a file present it will load them, otherwise boot as normally.
What do you think?

Dopam-IT:
hi everybody i try to make work the upnp on belkin rt3200 with the last git 31 08 2022 and last luci-app-upnp but doesn't work , i didn't understand ? i will do a special manip ?

firewire10000:
I'm just leaving this comment here to say that I have successfully compiled v22.03.0 with these file changes and miniupnpd and luci-app-upnp are both working successfully on IPv4.

PR #19321, opened 2022-09-06
php8: update to 8.1.10

mhei:
Maintainer: me
Compile tested: mxs
Run tested: mxs

Description:


mhei:
@neheb : Any idea why CI downloads and installs the 8.1.9 package for some architectures when the correct version should actually be available through this PR?

neheb:
I'd ignore those failures

PR #19329, opened 2022-09-08
keepalived: add status rpc and service improvement

jempatel:
keepalived: add status rpc and service improvement
    
- enable json by default to generate json stats
- add rpc to generate json status
- add kmod-nf-ipvs dependencies for virtual servers
- set default vip labels on virtual interfaces
- set process name for keepalived child processes

Signed-off-by: Jaymin Patel <jem.patel@gmail.com>

Maintainer: @feckert 
Run tested: x86-64, aarch64, virtual box, master

Description:
- add RPC to generate JSON status and fix dependencies for ipvs

**We need this for https://github.com/openwrt/luci/pull/5965**

### DIFF
--- a/net/keepalived/Makefile
+++ b/net/keepalived/Makefile
@@ -33,7 +33,6 @@ PKG_CONFIG_DEPENDS:= \
 	CONFIG_KEEPALIVED_SNMP_RFC3 \
 	CONFIG_KEEPALIVED_SNMP_REPLY_V3_FOR_V2 \
 	CONFIG_KEEPALIVED_DBUS \
-	CONFIG_KEEPALIVED_JSON \
 	CONFIG_KEEPALIVED_ROUTES \
 	CONFIG_IPV6
 
@@ -58,6 +57,8 @@ define Package/keepalived
     +libnl-genl \
     +libmagic \
     +libkmod \
+    +kmod-nf-ipvs \
+    +libjson-c \
     +KEEPALIVED_NFTABLES:libnftnl \
     +KEEPALIVED_VRRP:kmod-macvlan \
     +KEEPALIVED_VRRP:libnl-route \
@@ -68,7 +69,6 @@ define Package/keepalived
     +KEEPALIVED_IPTABLES:libxtables \
     +KEEPALIVED_IPTABLES:libipset \
     +(KEEPALIVED_SNMP_VRRP||KEEPALIVED_SNMP_CHECKER||KEEPALIVED_SNMP_RFC2||KEEPALIVED_SNMP_RFC3):libnetsnmp \
-    +KEEPALIVED_JSON:libjson-c \
     +KEEPALIVED_DBUS:glib2
 endef
 
@@ -83,6 +83,7 @@ define Package/keepalived/conffiles
 endef
 
 CONFIGURE_ARGS+= \
+	--enable-json \
 	--with-init=SYSV \
 	--disable-track-process \
 	--runstatedir="/var/run"
@@ -146,11 +147,6 @@ CONFIGURE_ARGS += \
 	--disable-snmp-reply-v3-for-v2
 endif
 endif # CONFIG_KEEPALIVED_SNMP_RFC3
-
-ifeq ($(CONFIG_KEEPALIVED_JSON),y)
-CONFIGURE_ARGS += \
-	--enable-json
-endif
 
 ifeq ($(CONFIG_KEEPALIVED_VRRP_AUTH),)
 CONFIGURE_ARGS += \
@@ -240,6 +236,10 @@ endif
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/keepalived
 	$(INSTALL_DATA) ./files/hotplug-user \
 		$(1)/etc/hotplug.d/keepalived/01-user
+
+	$(INSTALL_DIR) $(1)/usr/libexec/rpcd
+	$(INSTALL_BIN) ./files/usr/libexec/rpcd/keepalived \
+		$(1)/usr/libexec/rpcd/keepalived
 
 ifneq ($(CONFIG_KEEPALIVED_SNMP_VRRP)$(CONFIG_KEEPALIVED_SNMP_CHECKER)$(CONFIG_KEEPALIVED_SNMP_RFC2)$(CONFIG_KEEPALIVED_SNMP_RFC3),)
 	$(INSTALL_DIR) $(1)/usr/share/snmp/mibs
--- a/net/keepalived/files/usr/libexec/rpcd/keepalived
+++ b/net/keepalived/files/usr/libexec/rpcd/keepalived
@@ -1,0 +1,96 @@
+#!/bin/sh
+
+. /lib/functions.sh
+. /usr/share/libubox/jshn.sh
+
+RPC_SCRIPTS=/usr/libexec/keepalived/rpc
+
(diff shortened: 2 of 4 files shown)


net/keepalived/Makefile
```diff
@@ -241,6 +242,10 @@ endif
 	$(INSTALL_DATA) ./files/hotplug-user \
 		$(1)/etc/hotplug.d/keepalived/01-user
 
+	$(INSTALL_DIR) $(1)/usr/libexec/rpcd
```
feckert, line 245: Since the output depends on the compile option. So we should not always install the rpcd?
But I am generally in favour of always switching it on and removing the compile option.
Especially since we need the ubus output in the LuCI, right?

net/keepalived/files/usr/libexec/rpcd/keepalived
```diff
@@ -0,0 +1,96 @@
+#!/bin/sh
+
+. /lib/functions.sh
+. /usr/share/libubox/jshn.sh
+
+RPC_SCRIPTS=/usr/libexec/keepalived/rpc
```
feckert, line 6: What do we need this for?
If I read this correctly, is it possible to load additional ubus plugins for the keepalived?

net/keepalived/Makefile
```diff
@@ -58,6 +58,7 @@ define Package/keepalived
     +libnl-genl \
     +libmagic \
     +libkmod \
+    +kmod-nf-ipvs \
```
feckert, line 61: For what does we need this dependency?

feckert (changes requested):
Can you please split this into different commits?
The current commit message with the title `keepalived: fix dependencies` is not correct.

net/keepalived/Makefile
jempatel: keepalived uses Linux kernel's ipvs module for the virtual server load balancing, and modules are already being loaded from the script, due to missing dependency, keepalive was crashing when virtual servers are configured.

https://github.com/openwrt/packages/blob/1aa0d0878039295ed5d65066b74de3d06d15115c/net/keepalived/files/keepalived.init#L509

net/keepalived/files/usr/libexec/rpcd/keepalived
jempatel: Yes, shell script-based RPC must be installed in /usr/libexec/keepalived/rpc directory. A similar approach has already been implemented in apinger package.

https://github.com/openwrt/packages/blob/1aa0d0878039295ed5d65066b74de3d06d15115c/net/apinger/Makefile#L86

jempatel:
> Can you please split this into different commits? The current commit message with the title `keepalived: fix dependencies` is not correct.

I can update the commit message, It just happened in my testing, so I fixed it.
https://github.com/openwrt/packages/pull/19329#discussion_r965962603

net/keepalived/Makefile
jempatel: Yes, Enabled JSON compilation in the default builds and removed the option.

jempatel:
> Can you please split this into different commits? The current commit message with the title `keepalived: fix dependencies` is not correct.

updated commit message

feckert:
I have now had a look at it. Everything fits so far.
But I'm not sure if the `status` of the ubus is correct.
I think it would be better to use `dump` here.
Because not only the stats are output, but also the keepalived configuration.

jempatel:
> I have now had a look at it. Everything fits so far. But I'm not sure if the `status` of the ubus is correct. I think it would be better to use `dump` here. Because not only the stats are output, but also the keepalived configuration.

Actually, it dumps the run time status of keepalived and not reading anything from UCI or keepalived config file.

jempatel:
> I have now had a look at it. Everything fits so far. But I'm not sure if the `status` of the ubus is correct. I think it would be better to use `dump` here. Because not only the stats are output, but also the keepalived configuration.

yes, that makes sense. change `status` to `dump`

net/keepalived/files/usr/libexec/rpcd/keepalived
```diff
@@ -0,0 +1,96 @@
...
+
+[ -d $RPC_SCRIPTS ] && include $RPC_SCRIPTS
```
feckert, line 8: I am currently trying to write a plugin for the ubus keepalived.
To verify your pull request.
Can you give me an example?
I'm not sure if the shell (openwrt uses ash) supports this `include` buildin word.

jempatel: Its default busysbox shell ash only.

`include` comes from `/lib/functions.sh`

Here are two examples of the keepalived rpcd plugin.

1. Valid RPC with a success response.
```sh
cat /usr/libexec/keepalived/rpc/example.sh 
#!/bin/sh

example() {
	json_init
	json_add_string key value
	json_dump
}

example_help() {
	json_add_object example
	json_close_object
}
```

2. Invalid RPC, partial RPC.
`touch /usr/libexec/keepalived/rpc/example2.sh`

3. Restart rpcd and list keepalived RPCs.
```sh
/etc/init.d/rpcd restart
ubus -v list keepalived
'keepalived' @19c942e2
	"dump":{}
	"example":{}
	"example2":{}
```

4. RPC call response
```sh
ubus call keepalived example
{
	"key": "value"
}
ubus call keepalived example2
{
	"error": "invalid call example2"
}
```

net/keepalived/files/usr/libexec/rpcd/keepalived
```diff
@@ -0,0 +1,96 @@
...
+
+__function__() {
+    type "$1" > /dev/null 2>&1
```
feckert, line 11: Wrong indentation please use tabs

net/keepalived/files/usr/libexec/rpcd/keepalived
```diff
@@ -0,0 +1,96 @@
...
+
+	json_init
+
+	json_add_object dump
+	json_close_object
+
+	foreach_extra list_extra ${1}
+
+	json_dump
+}
```
feckert, line 83: Space after the bracket

net/keepalived/files/usr/libexec/rpcd/keepalived
```diff
@@ -0,0 +1,96 @@
...
+
+__function__() {
+    type "$1" > /dev/null 2>&1
+}
+
+foreach_extra() {
+	local file obj
+
+	[ ! -d $RPC_SCRIPTS ] && return
+
```
feckert, line 18: Please remove this tab

feckert (changes requested):
Looks good. Can you please add my style note. When that is done. Then we are ready to merge.

net/keepalived/files/usr/libexec/rpcd/keepalived
jempatel: Updated

net/keepalived/files/usr/libexec/rpcd/keepalived
jempatel: Updated

net/keepalived/files/usr/libexec/rpcd/keepalived
jempatel: Updated

jempatel:
> Looks good. Can you please add my style note. When that is done. Then we are ready to merge.

comments resolved.

feckert:
Thanks merged :+1:

PR #19337, opened 2022-09-10
liboil: small cleanups

neheb:
Fixes a compile error when glib2 is missing.

get rid of nls.mk. iconv/gettext is not used. This was previously here because of glib2.

Maintainer: @flyn-org 
Compile tested: mips64 WSL


neheb:
wait a minute...

this is supposedly used by gst1-plugins-base. But I can't find a reference to it in the code.

PR #19339, opened 2022-09-10
grep: update to 3.8

PolynomialDivision:
Manually refresh:
- ppc-musl.patch

Release Notes:
https://lists.gnu.org/archive/html/info-gnu/2022-09/msg00001.html

Maintainer: @Zokormazo
Compile tested: ipq40xx
Run tested: ipq40xx avm fritz!box 4040


hnyman:
@PolynomialDivision  @neheb 
Seems to cause dependency errors like these:

https://downloads.openwrt.org/snapshots/faillogs/arm_cortex-a7_neon-vfpv4/packages/grep/compile.txt
https://downloads.openwrt.org/snapshots/faillogs/mipsel_74kc/packages/grep/compile.txt

```
checking for PCRE... yes
checking for pcre2_compile... yes
...
Package grep is missing dependencies for the following libraries:
libpcre2-8.so.0
```

You probably need to disable libpcre detection in the configure script / configure.ac

PolynomialDivision:
Lets just switch to libprce2 https://github.com/openwrt/packages/pull/19394

PR #19342, opened 2022-09-11
python3-iperf3: add iperf3 python wrapper

PolynomialDivision:
This wrapper gives us an easy to use api for using iperf3 in python.

Needs: https://github.com/openwrt/packages/pull/19344


PolynomialDivision:
Runtested Belkin rt3200.

PolynomialDivision:
Cherry-picked into https://github.com/openwrt/packages/pull/19344

PR #19355, opened 2022-09-13
[WIP] flent: add flent package

PolynomialDivision:
As described on GitHub page [0]:
  Flent is a Python wrapper to run multiple simultaneous
  netperf/iperf/ping instances and aggregate the results.

[0] - https://github.com/tohojo/flent

Maintainer: me (maybe @tohojo )
Compile tested: linksys e8450
Run tested: linksys e8450

Something is still wrong, because command always fails with
```
flent rrul -p all_scaled -l 60 -H netperf-west.bufferbloat.net -t text-to-be-included-in-plot -o filename.png
Starting Flent 2.0.1 using Python 3.10.5.
ERROR: Found no hostnames on lookup of netperf-west.bufferbloat.net
```


PolynomialDivision:
ping @neheb ? ;)

lang/python/python3-flent/Makefile
```diff
@@ -0,0 +1,44 @@
...
+  URL:=https://github.com/tohojo/flent
+  DEPENDS:= \
+    +python3-light \
+    +python3-email \
+    +python3-urllib \
+    +python3-uuid \
+    +python3-logging \
+    +python3-xml \
+    +python3-decimal \
+    +python3-distutils \
```
tohojo, line 32: The list of dependencies seems a bit long, are they all really needed?

lang/python/python3-flent/Makefile
```diff
@@ -0,0 +1,44 @@
...
+endef
+
+define Package/python3-flent/description
+  Flent is a Python wrapper to run multiple simultaneous
+  netperf/iperf/ping instances and aggregate the results.
+endef
+
+$(eval $(call Py3Package,python3-flent))
+$(eval $(call BuildPackage,python3-flent))
+$(eval $(call BuildPackage,python3-flent-src))
```
tohojo, line 44: What's in the -src package?

lang/python/python3-flent/Makefile
```diff
@@ -0,0 +1,44 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=python3-flent
```
tohojo, line 3: Not sure what the convention is on OpenWrt for Python packages, but on other distros the 'python-' prefix is for libraries, which Flent is not. So I think it would be better to just name the package 'flent'

lang/python/python3-flent/Makefile
```diff
@@ -0,0 +1,44 @@
...
+PKG_VERSION:=2.0.1
```
tohojo, line 4: I should really tag a new version, there are a couple of bugs in this one...

lang/python/python3-flent/Makefile
```diff
@@ -0,0 +1,44 @@
...
+    +python3-light \
+    +python3-email \
+    +python3-urllib \
+    +python3-uuid \
+    +python3-logging \
+    +python3-xml \
+    +python3-decimal \
+    +python3-distutils \
+    +netperf \
+    +fping
```
tohojo, line 34: There's an existing flent-tools package with the C helper utils. Maybe consolidating that into this package would make sense (and adding a dependency as well)?

tohojo (commented):
Thanks for doing this! A couple of comments...

tohojo:
Okay, not waiting for comments, I see :/

Please address them in a follow-up, then...

lang/python/python3-flent/Makefile
PolynomialDivision: I guess it's the python files. Normally, you only have those `.pyc` on OpenWrt. However, I am not sure. Just copy + paste some other tool and used it as template.

lang/python/python3-flent/Makefile
PolynomialDivision: Yes. At least I needed them. Maybe fping can be replaced. Or maybe python-light.

lang/python/python3-flent/Makefile
PolynomialDivision: That would be awesome! :)

lang/python/python3-flent/Makefile
PolynomialDivision: I am also not sure. Maybe I looked at the wrong packages, for example python3-speedtest-cli is also using the naming I used: https://github.com/openwrt/packages/blob/master/lang/python/python3-speedtest-cli/Makefile

Maybe @jefferyto can say something to that.

> So I think it would be better to just name the package 'flent'

Of course I can do that.

lang/python/python3-flent/Makefile
PolynomialDivision: I would not add it as dependency directly, but maybe as new variant? `flent-with-addons` or something like this?

PR #19357, opened 2022-09-13
treewide: change argp-standalone to a direct dependency

langerth:
As preparation to build argp-standalone as dynamic library, change all packets with a build dependency for it to a direct dependency. Without this the build will link against the dynamic library and the buildsystem will complain about the missing dependency.

Signed-off-by: Thomas Langer <tlanger@maxlinear.com>

### DIFF
--- a/utils/readsb/Makefile
+++ b/utils/readsb/Makefile
@@ -18,7 +18,6 @@ PKG_LICENSE_FILES:=COPYING LICENSE
 
 PKG_MAINTAINER:=Stijn Tintel <stijn@linux-ipv6.be>
 
-PKG_BUILD_DEPENDS:=!USE_GLIBC:argp-standalone
 PKG_BUILD_PARALLEL:=1
 
 include $(INCLUDE_DIR)/package.mk
@@ -26,9 +25,9 @@ include $(INCLUDE_DIR)/package.mk
 define Package/readsb/default
   SECTION:=utils
   CATEGORY:=Utilities
-  DEPENDS:=+libncurses
+  DEPENDS:=+libncurses +!USE_GLIBC:argp-standalone
   TITLE:=Mode-S/ADSB/TIS decoder for various devices
-  URL:=https://github.com/Mictronic/readsb
+  URL:=https://github.com/Mictronics/readsb
 endef
 
 define Package/readsb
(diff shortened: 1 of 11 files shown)


utils/readsb/Makefile
```diff
@@ -18,17 +18,16 @@ PKG_LICENSE_FILES:=COPYING LICENSE
...
 include $(INCLUDE_DIR)/package.mk
 
 define Package/readsb/default
   SECTION:=utils
   CATEGORY:=Utilities
-  DEPENDS:=+libncurses
+  DEPENDS:=+libncurses +!USE_GLIBC:argp-standalone
   TITLE:=Mode-S/ADSB/TIS decoder for various devices
-  URL:=https://github.com/Mictronic/readsb
+  URL:=https://github.com/Mictronics/readsb
```
langerth, line 30: This change is not strictly related and I can drop it if necessary

ynezz: Indeed, you either need to drop it or preferably include the fix as separate commit.

ynezz:
You forget to include more context here and in the commit message, so I assume, that this change itself doesn't make much sense to other reviewers.

It would probably be fair to warn upfront about [external dependency](https://patchwork.ozlabs.org/project/openwrt/patch/mailman.16158.1663084212.4154159.openwrt-devel@lists.openwrt.org/) and that it's probably not going to work until that it's merged.

BTW is there any reason, why you didn't followed the [PR template](https://github.com/openwrt/packages/blob/master/.github/pull_request_template)? I would recommend to read the [fine manual](https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md).

> As preparation to build argp-standalone as dynamic library, change all packets with a build dependency for it to a direct dependency. Without this the build will link against the dynamic library and the buildsystem will complain about the missing dependency.

Quoting from my [mail review comment](https://lists.infradead.org/pipermail/openwrt-devel/2022-September/039478.html):

> Can't this be solved in some less intrusive way? I can imagine, that this is
not going to be bisect friendly and could provide other woes in the future.
>
> What about introducing shared library in conjunction to the static one, thus
not breaking anything? Then if it makes sense, we could convert existing
static users to shared version in a next step.

diizzyy:
We should most likely make sure it's switched to https://github.com/argp-standalone/argp-standalone in openwrt's repo before proceeding

langerth:
As we found a different way for supporting this as part of our platform integration to prplos, I decline this PR now.

PR #19358, opened 2022-09-13
boinc-wrapper: add new package

AenBleidd:
Signed-off-by: Vitalii Koshura <lestat.de.lionkur@gmail.com>

### DIFF
--- a/net/boinc-wrapper/Makefile
+++ b/net/boinc-wrapper/Makefile
@@ -1,0 +1,70 @@
+# SPDX-Identifier-License: GPL-3.0-or-later
+#
+# Copyright (C) 2023 by Vitalii Koshura <lestat.de.lionkur@gmail.com>
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=boinc-wrapper
+PKG_VERSION:=26018
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/BOINC/boinc/tar.gz/refs/tags/wrapper/$(PKG_VERSION)?
+PKG_HASH:=a93ae0a9e640a893e78f523c6d93f31b1d5812092f85af4e9ce964846373f55d
+
+PKG_MAINTAINER:=Vitalii Koshura <lestat.de.lionkur@gmail.com>
+PKG_LICENSE:=GPL-3.0-or-later
+PKG_LICENSE_FILES:=COPYING
+PKG_CPE_ID:=cpe:/a:boinc-wrapper:boinc-wrapper
+
+PKG_INSTALL:=1
+PKG_BUILD_PARALLEL:=0
+PKG_FIXUP:=autoreconf
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/target.mk
+
+define Package/boinc-wrapper
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=BOINC wrapper
+  DEPENDS:=+libstdcpp
+  URL:=https://github.com/BOINC/boinc/
+endef
+
+define Package/boinc-wrapper/description
+  The Berkeley Open Infrastructure for Network Computing (BOINC) is a
+  software platform for distributed computing: several initiatives of
+  various scientific disciplines all compete for the idle time of
+  desktop computers. The developers' web site at the University of
+  Berkeley serves as a common portal to the otherwise independently run
+  projects.
+
+  This package provides the BOINC wrapper that runs the Project applications as
+  subprocesses, and handles all communication with the BOINC client
+  (e.g., to report CPU time and fraction done).
+endef
+
+CONFIGURE_ARGS += \
+	--disable-server --disable-manager --disable-client --enable-libraries \
+	--enable-boinczip \
+	--with-boinc-platform=$(REAL_GNU_TARGET_NAME) \
+	--with-boinc-alt-platform=$(ARCH)-$(BOARD)-$(DEVICE_TYPE)-openwrt-$(TARGET_SUFFIX)
+
+TARGET_CFLAGS += -Wno-format -Wno-format-security
+TARGET_CPPFLAGS += -Wno-format -Wno-format-security
+
+define Build/Compile
+	$(call Build/Compile/Default)
+	$(MAKE_VARS) $(MAKE) \
+		-C $(PKG_BUILD_DIR)/samples/wrapper \
+		$(MAKE_FLAGS)
+endef
+
+define Package/boinc-wrapper/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/samples/wrapper/wrapper $(1)/usr/bin/boinc-wrapper
+endef
+
+$(eval $(call BuildPackage,boinc-wrapper))
--- a/net/boinc-wrapper/test.sh
+++ b/net/boinc-wrapper/test.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+boinc-wrapper
+# Non-zero exit status is expected so always return 0 to avoid false failure.
+exit 0


fantom-x:
Is this different from the existing package? 

```
# opkg info boinc
Package: boinc
Version: 7.16.16-1
Depends: libc, curl, bzip2, libstdcpp6, libopenssl1.1, zlib
Status: unknown ok not-installed
Section: net
Architecture: x86_64
Size: 792911
Filename: boinc_7.16.16-1_x86_64.ipk
Description: The Berkeley Open Infrastructure for Network Computing (BOINC) is a
 software platform for distributed computing: several initiatives of
 various scientific disciplines all compete for the idle time of
 desktop computers. The developers' web site at the University of
 Berkeley serves as a common portal to the otherwise independently run
 projects.
 
 This package provides the BOINC core client program that is
 required to participate in any project that uses BOINC to control what
 projects to join and to determine constraints for the computation
 like the percentage of CPU time. OpenWrt does not
 provide the graphical BOINC Manager, but you can connect to this
 machine from the BOINC Manager of your desktop computer.
```

hnyman:
How does this related to the already existing boinc package?
(you have already copied the description from that, and even forgotten to adapt package name in the section definition)

Should this be combined into the same Makefile?

cc @smoe @N30dG

AenBleidd:
Nice catch! Thank you, I'll fix the description.
BOINC Wrapper is the application that could run with BOINC client (another OpenWRT package that is called here just BOINC).

smoe:
The wrapper is expected to be provided by the project maintainers, while the boinc-client is the rool that auto-fetches that wrapper (and other executables) from the respective project web page.

The wrapper is important for OpenWRT, and even more important are clear build instructions for the wrapper, so the respective project maintainers can get their patches in. I had failed to build it which would have widened the support for OpenWRT. The particular challenge is to have a static build of the wrapper (and all the other executables) because of glibc/musl differences in default installations. Projects are unlikely to provide multiple ARM32 binaries - it should be one for all OSs.


Knowing @AenBleidd very close to BOINC upstream, it would indeed be good if the regular BOINC client and the BOINC wrapper could be generated from the same source tree. I would appreciate that and certainly also would @N30dG .

A difficulty could be that the wrapper may be tied closer with the BOINC server release, while the boinc client is tagged separately. But again, with @AenBleidd 's active participation that should be a non-issue.

What would I need to do to see the two efforts merged?

AenBleidd:
@hnyman, description fixed. Please review

net/boinc-wrapper/Makefile
```diff
@@ -0,0 +1,78 @@
...
+
+PKG_NAME:=boinc-wrapper
+PKG_VERSION:=26018
+PKG_RELEASE:=1
+
+PKG_SOURCE_DATE:=2022-09-07
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/BOINC/boinc/tar.gz/refs/tags/wrapper/$(PKG_VERSION)?
+PKG_HASH:=a93ae0a9e640a893e78f523c6d93f31b1d5812092f85af4e9ce964846373f55d
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)
```
BKPepe, line 16: This is already default, so no need to explicitly state it again, right?

net/boinc-wrapper/Makefile
```diff
@@ -0,0 +1,78 @@
...
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=boinc-wrapper
+PKG_VERSION:=26018
+PKG_RELEASE:=1
+
+PKG_SOURCE_DATE:=2022-09-07
```
BKPepe, line 12: Isn't this used mostly when you are cloning Git repositories? This could be removed. 

Reference: https://openwrt.org/docs/guide-developer/package-policies#versioning

net/boinc-wrapper/Makefile
```diff
@@ -0,0 +1,78 @@
...
+PKG_HASH:=a93ae0a9e640a893e78f523c6d93f31b1d5812092f85af4e9ce964846373f55d
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)
+
+PKG_MAINTAINER:=Vitalii Koshura <lestat.de.lionkur@gmail.com>
+PKG_LICENSE:=GPL-3.0-or-later
+PKG_LICENSE_FILES:=COPYING
+PKG_CPE_ID:=cpe:/a:boinc-wrapper:boinc-wrapper
+
+PKG_INSTALL:=1
+PKG_BUILD_PARALLEL:=0
```
BKPepe, line 24: Was it causing any issues while building it parallelly?

net/boinc-wrapper/Makefile
```diff
@@ -0,0 +1,78 @@
...
+	--with-boinc-platform=$(REAL_GNU_TARGET_NAME) \
+	--with-boinc-alt-platform=$(ARCH)-$(BOARD)-$(DEVICE_TYPE)-openwrt-$(TARGET_SUFFIX)
+
+define Build/Compile
+	$(MAKE) -C $(PKG_BUILD_DIR) \
+		CC="$(TARGET_CC)" \
+		CXX="$(TARGET_CXX)" \
+		CFLAGS="$(TARGET_CFLAGS) $(TARGET_CPPFLAGS) -Wno-format -Wno-format-security " \
+		CPPFLAGS="$(TARGET_CFLAGS) $(TARGET_CPPFLAGS) -Wno-format -Wno-format-security " \
+		LDFLAGS="$(TARGET_LDFLAGS) "
```
BKPepe, line 63: For this one, isn't it possible to call [defaults](https://github.com/openwrt/openwrt/blob/master/include/package-defaults.mk#L142) and then build wrapper?

net/boinc-wrapper/Makefile
```diff
@@ -0,0 +1,78 @@
...
+  This package provides the BOINC wrapper that runs the Project applications as
+  subprocesses, and handles all communication with the BOINC client
+  (e.g., to report CPU time and fraction done).
+endef
+
+CONFIGURE_ARGS += \
+	--disable-server --disable-manager --disable-client --enable-libraries \
+	--enable-boinczip \
+	--with-boinc-platform=$(REAL_GNU_TARGET_NAME) \
+	--with-boinc-alt-platform=$(ARCH)-$(BOARD)-$(DEVICE_TYPE)-openwrt-$(TARGET_SUFFIX)
```
BKPepe, line 55: Nitpicking:
I'm thinking. Correct me if I am wrong. If this is not case sensitive, then you can use the proper spelling of ``OpenWrt``.

BKPepe (changes requested):
If possible, consider to add ``test.sh`` as it is done for other packages in this repository. It ensures that it is run tested inside GitHub Actions.

net/boinc-wrapper/Makefile
AenBleidd: we have all configurations in lowercase, so just for consistency with other platforms it would be nice to leave it as is.

BKPepe: Ok, ok. Thanks for explanation!

net/boinc-wrapper/Makefile
AenBleidd: I will check this, probably that would be an easier solution

net/boinc-wrapper/Makefile
AenBleidd: Yes, it does. It's not always reproducible, so for the stability we're now building it in one thread.
We have an open ticket to fix parallel build but it's not done yet, so once it will be fixed on our side, I'll update this Makefile

net/boinc-wrapper/Makefile
AenBleidd: True, will remove that

net/boinc-wrapper/Makefile
AenBleidd: I'll check this and remove if it's not needed, thank you for the hint

AenBleidd:
> If possible, consider to add `test.sh` as it is done for other packages in this repository. It ensures that it is run tested inside GitHub Actions.

Yes, technically this could be done, I'll add this script

net/boinc-wrapper/Makefile
AenBleidd: Done

net/boinc-wrapper/Makefile
AenBleidd: Done

net/boinc-wrapper/Makefile
AenBleidd: Done

AenBleidd:
@BKPepe, please review

BKPepe:
Approved CI, let's see how it goes. I could not review it during last week as I was at Battlemesh conference, sorry for that! Hopefully, I will get into that in a few days.

AenBleidd:
@BKPepe, friendly reminder ;)

AenBleidd:
@BKPepe, any chance to have this reviewed?

BKPepe:
I am not sure, if this is expected:
```
Use package specific test.sh
12:53:51 (433): wrapper (7.21.26018): starting
12:53:51 (433): can't open job file job.xml
12:53:51 (433): can't parse job file: -108
12:53:51 (433): called boinc_finish(-108)
Test successful
Removing package boinc-wrapper from root...
```

net/boinc-wrapper/Makefile
```diff
@@ -0,0 +1,74 @@
+# Copyright 2022 by Vitalii Koshura
```
1715173329, line 1: Please use SPDX.

net/boinc-wrapper/Makefile
```diff
@@ -0,0 +1,74 @@
...
+		CXX="$(TARGET_CXX)" \
+		CFLAGS="$(TARGET_CFLAGS)" \
+		CPPFLAGS="$(TARGET_CPPFLAGS)" \
+		LDFLAGS="$(TARGET_LDFLAGS) " \
+		all
+endef
+
+define Package/boinc-wrapper/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/samples/wrapper/wrapper $(1)/usr/bin/
```
1715173329, line 71: Emmm I would prefer using `bonic-wrapper`. `wrapper` is too common.

1715173329 (changes requested):

1715173329:
Otherwise LGTM, thanks for your work!

net/boinc-wrapper/Makefile
AenBleidd: What is SPDX?

1715173329: https://spdx.org/licenses/
e.g.

https://github.com/openwrt/packages/blob/1acf423ef5891eb88c8e2bbdf8dad7896b8db398/devel/flex/Makefile#L1-L3

AenBleidd: Fixed

net/boinc-wrapper/Makefile
AenBleidd: Fixed

AenBleidd:
Hello, @1715173329, sorry for the delay.
Requested changes were done. Please review again.
Thank you in advance

net/boinc-wrapper/Makefile
```diff
@@ -0,0 +1,75 @@
...
+		CFLAGS="$(TARGET_CFLAGS)" \
+		CPPFLAGS="$(TARGET_CPPFLAGS)" \
+		LDFLAGS="$(TARGET_LDFLAGS) " \
+		all
+endef
+
+define Package/boinc-wrapper/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(CP) $(PKG_BUILD_DIR)/samples/wrapper/wrapper $(PKG_BUILD_DIR)/samples/wrapper/boinc-wrapper
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/samples/wrapper/boinc-wrapper $(1)/usr/bin/
```
1715173329, line 72:
```suggestion
	$(INSTALL_BIN) $(PKG_BUILD_DIR)/samples/wrapper/wrapper $(1)/usr/bin/boinc-wrapper
```

net/boinc-wrapper/Makefile
```diff
@@ -0,0 +1,75 @@
...
+
+define Build/Compile
+	$(call Build/Compile/Default)
+	$(MAKE) -C $(PKG_BUILD_DIR)/samples/wrapper \
+		CC="$(TARGET_CC)" \
+		CXX="$(TARGET_CXX)" \
+		CFLAGS="$(TARGET_CFLAGS)" \
+		CPPFLAGS="$(TARGET_CPPFLAGS)" \
+		LDFLAGS="$(TARGET_LDFLAGS) " \
+		all
```
1715173329, line 66:
```suggestion
	+$(MAKE_VARS) $(MAKE) \
		-C $(PKG_BUILD_DIR)/samples/wrapper \
		$(MAKE_FLAGS)
```

this should be enough.

net/boinc-wrapper/Makefile
```diff
@@ -0,0 +1,70 @@
...
+	--enable-boinczip \
+	--with-boinc-platform=$(REAL_GNU_TARGET_NAME) \
+	--with-boinc-alt-platform=$(ARCH)-$(BOARD)-$(DEVICE_TYPE)-openwrt-$(TARGET_SUFFIX)
+
+TARGET_CFLAGS += -Wno-format -Wno-format-security
+TARGET_CPPFLAGS += -Wno-format -Wno-format-security
+
+define Build/Compile
+	$(call Build/Compile/Default)
+	$(MAKE_VARS) $(MAKE)
```
1715173329, line 60:
```suggestion
	$(MAKE_VARS) $(MAKE) \
```

small typo

1715173329 (approved):
lgtm - thanks!

net/boinc-wrapper/Makefile
AenBleidd: Done

net/boinc-wrapper/Makefile
AenBleidd: Done

net/boinc-wrapper/Makefile
AenBleidd: Fixed

1715173329:
Merged, thank you!

PR #19362, opened 2022-09-14
prometheus/collectd: adapt dependencies for the dsl collector 

dhewg:
This depends on https://github.com/openwrt/openwrt/pull/10586

It simplifies and cleans up the dependencies of consumers of the dsl ubus stats.

Maintainers: @champtar @jow- @hnyman 
I tested prometheus, but not collectd, so cc @jekkos @feckert

### DIFF
--- a/utils/collectd/Makefile
+++ b/utils/collectd/Makefile
@@ -528,4 +528,4 @@ $(eval $(call BuildPlugin,write-graphite,Carbon/Graphite output,write_graphite,+
 $(eval $(call BuildPlugin,write-http,HTTP POST output,write_http,+PACKAGE_collectd-mod-write-http:libcurl))
 
 $(eval $(call BuildScriptPlugin,sqm,SQM/qdisc collection,sqm_collectd,+PACKAGE_collectd-mod-sqm:collectd-mod-exec))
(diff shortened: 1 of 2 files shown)


feckert:
I have not tested it now. But I see nothing that would speak against it. I assume the ubus json format has not changed between vr9 and vr11

jekkos:
I'm using the collectd already with vr11, so I already did this change locally otherwise I was not able to select the packages. It works so good to go.

jekkos:
Actually I replaced vrx200 with this `@(TARGET_lantiq_xrx200||TARGET_ipq40xx)`

feckert:
> Actually I replaced `@TARGET_lantiq_xrx200` with `@(TARGET_lantiq_xrx200||TARGET_ipq40xx)`
I like that better. Now you can see right away in the source for which targets this applies.

dhewg:
Yeah, I had that too at first, but there're way more ipq40xx boards than those with a dsl chip.
With the patches here, the package is only selectable when the dsl daemon is available, which makes more sense to me.

feckert:
@dhewg @jekkos Well then let's do it like this

dhewg:
The PR this depends on has been merged, so this can go in too now. Thanks!

PR #19368, opened 2022-09-14
iperf3: fix install section of Makefile and dependencies

PolynomialDivision:
In the Makefile the library installation was accidentally called "Package/iperf3/install" and not "Package/libiperf3/install". Fix this typo. Thanks to Hartmut spotting this.

Also the iperf3-ssl does not need to depend on libiperf3.

Fixes ae48be8e2157 ("iperf3: add shared libiperf library and link iperf3 dynamically")

Maintainer: @nbd168 
Compile tested: Belkin
Run tested: Belkin

### DIFF
--- a/net/iperf3/Makefile
+++ b/net/iperf3/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=iperf
 PKG_VERSION:=3.11
-PKG_RELEASE:=1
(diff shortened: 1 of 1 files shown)


neheb:
this was posted on the openwrt-devel mailing list btw.

PolynomialDivision:
> this was posted on the openwrt-devel mailing list btw.

Updated. There was a second mistake. Also fixed that.

PR #19374, opened 2022-09-15
keepalived: high-availability files and data sync

jempatel:
keepalived: high-availability files and data sync

add new package keepalived-sync to synchronize files and data between master and backup node. The master node uses SSH over rsync to send and the backup node will use inotifywatch to watch received files.

The master node can track rsync.sh script to send configuration file on a backup node based on the vrrp_script configuration of the same script.

The backup node will have a keepalived-inotify service, which would watch for newly received files and it would call hotplug event. Each service can keep its respective script under the keepalived hotplug directory and executes commands to stop, start service or update any config in real-time.

Whenever a switchover will happen, the backup node would have the latest config and data files from the master node.

Hotplug events can be used to apply config when files are received.

Signed-off-by: Jaymin Patel <jem.patel@gmail.com>

Maintainer: @feckert
Run tested: x86-64, virtual box, master

Description:
OpenWrt HA synchronization support

**We need this for https://github.com/openwrt/luci/pull/5965**

### DIFF
--- a/net/keepalived/Makefile
+++ b/net/keepalived/Makefile
@@ -274,4 +274,103 @@ endif
 
 endef
 
+define Package/keepalived-sync
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Keepalived Master and Backup Synchronization
+  DEPENDS:= +keepalived +rsync +inotifywait +sudo +@BUSYBOX_CUSTOM +@BUSYBOX_CONFIG_TIMEOUT
+endef
+
+define Package/keepalived-sync/description
+ Keepalived HA with Master to Backup files and data Synchronization
+endef
+
+define Package/keepalived-sync/conffiles
+/etc/keepalived/scripts
+/etc/keepalived/keys
+endef
+
+define Package/keepalived-sync/install
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_BIN) ./files/etc/init.d/keepalived-inotify \
+		$(1)/etc/init.d/keepalived-inotify
+
+	$(INSTALL_DIR) $(1)/usr/share/keepalived/scripts
+	$(INSTALL_BIN) ./files/usr/share/keepalived/scripts/rsync.sh \
+		$(1)/usr/share/keepalived/scripts/rsync.sh
+
+	$(INSTALL_DIR) $(1)/etc/keepalived/scripts
+	$(LN) /usr/share/keepalived/scripts/rsync.sh \
+		$(1)/etc/keepalived/scripts/rsync.sh
+
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) ./files/usr/bin/keepalived-rsync-inotify \
+		$(1)/usr/bin/keepalived-rsync-inotify
+
+	$(INSTALL_DIR) $(1)/lib/functions/keepalived
+	$(INSTALL_DATA) ./files/lib/functions/keepalived/hotplug.sh \
+		$(1)/lib/functions/keepalived/hotplug.sh
+	$(INSTALL_DATA) ./files/lib/functions/keepalived/common.sh \
+		$(1)/lib/functions/keepalived/common.sh
+
+	$(INSTALL_DIR) $(1)/usr/libexec/keepalived/rpc
+	$(INSTALL_DATA) ./files/usr/libexec/keepalived/rpc/sync.sh \
+		$(1)/usr/libexec/keepalived/rpc/sync.sh
+
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/keepalived
+	$(CP) ./files/etc/hotplug.d/keepalived/* \
+		$(1)/etc/hotplug.d/keepalived
+endef
+
+USER=keepalived
+USER_ID=60001
+USER_HOME=/usr/share/keepalived/rsync
+SUDO_DIR=/etc/sudoers.d
+SUDO_FILE=$(SUDO_DIR)/$(USER)
+KEYS_DIR=/etc/keepalived/keys
+
+define Package/keepalived-sync/postinst
+	#!/bin/sh
+
+	mkdir -p "$${IPKG_INSTROOT}/etc/uci-defaults"
+	DEFAULT_SCRIPT="$${IPKG_INSTROOT}/etc/uci-defaults/99-keepalived-sync"
+
+	cat << EOF > $${DEFAULT_SCRIPT}
+#!/bin/sh
+
+. /lib/functions.sh
+
+mkdir -p $(KEYS_DIR)
+
+group_add "$(USER)" "$(USER_ID)"
+user_add "$(USER)" "$(USER_ID)" "$(USER_ID)" "$(USER)" "$(USER_HOME)" "/bin/ash"
+
+mkdir -m 700 -p "$(USER_HOME)"
+mkdir -m 700 -p "$(USER_HOME)/.ssh"
+chown "$(USER)":"$(USER)" "$(USER_HOME)" -R
+
+[ ! -d "$(SUDO_DIR)" ] && mkdir "$(SUDO_DIR)"
+echo "$(USER) ALL= NOPASSWD:/usr/bin/rsync" > "$(SUDO_FILE)"
+EOF
+
+	[ -z "$${IPKG_INSTROOT}" ] && [ -f "$${DEFAULT_SCRIPT}" ] && sh "$${DEFAULT_SCRIPT}"
+
+	exit 0
+endef
+
+define Package/keepalived-sync/postrm
+	#!/bin/sh
+
+	[ -n "$${IPKG_INSTROOT}" ] && exit 0
+
+	[ -d "$(KEYS_DIR)" ] && rm -rf "$(KEYS_DIR)"
+	[ -d "$(USER_HOME)" ] && rm -rf "$(USER_HOME)"
+	[ -f "$(SUDO_FILE)" ] && rm -f "$(SUDO_FILE)"
+
+	sed -i -e "/^$(USER):/d" /etc/passwd /etc/shadow /etc/group
+
+	exit 0
+endef
+
 $(eval $(call BuildPackage,keepalived))
+$(eval $(call BuildPackage,keepalived-sync))
--- a/net/keepalived/files/etc/init.d/keepalived-inotify
+++ b/net/keepalived/files/etc/init.d/keepalived-inotify
@@ -1,0 +1,65 @@
+#!/bin/sh /etc/rc.common
+
+START=99
+USE_PROCD=1
+PROG="/usr/bin/keepalived-rsync-inotify"
+
+KEEPALIVED_USER=keepalived
+KEEPALIVED_HOME=$(awk -F: "/^$KEEPALIVED_USER/{print \$6}" /etc/passwd)
+
+start_instance() {
+	local cfg=$1
+	local vrrp_instance=$2
+	local peer=$3
+
+	config_get name $cfg name
+	[ -z "$name" ] && return
+
+	[ "$name" != "$peer" ] && return
+
+	config_get sync $cfg sync 0
+	[ "$sync" = "0" ] && return
+
+	config_get sync_mode $cfg sync_mode
+	[ "$sync_mode" != "receive" ] && return
+
+	config_get sync_dir $cfg sync_dir $KEEPALIVED_HOME
+	[ -z "$sync_dir" ] && return
+
+	[ ! -d "$sync_dir" ] && mkdir -m 755 -p "$sync_dir"
+
+	procd_open_instance "$name"
+	procd_set_param command /bin/sh "$PROG" "$vrrp_instance" "$name" "$sync_dir"
+	procd_set_param pidfile /var/run/keepalived-inotify-$name.pid
+	procd_close_instance
+}
+
+process_unicast_peer() {
+	local peer=$1
+	local vrrp_instance=$2
+
+	config_foreach start_instance peer "$vrrp_instance" "$peer"
+}
+
+process_vrrp_instance() {
+	local cfg=$1
+	local peer_instance=$2
+	local name unicast_peer
+
+	config_get name $cfg name
+	config_get unicast_peer $cfg unicast_peer
+
+	if [ -n "$peer_instance" ]; then
+		list_contains unicast_peer "$peer_instance" || return
+		process_unicast_peer "$peer_instance" "$name"
+	else
+		config_list_foreach $cfg unicast_peer process_unicast_peer "$name"
+	fi
+}
+
+start_service() {
+	local peer_instance=$1
+
+	config_load keepalived
+	config_foreach process_vrrp_instance vrrp_instance "$peer_instance"
+}
--- a/net/keepalived/files/usr/bin/keepalived-rsync-inotify
+++ b/net/keepalived/files/usr/bin/keepalived-rsync-inotify
@@ -1,0 +1,54 @@
+#!/bin/sh
+
+# shellcheck shell=ash
+
+# shellcheck source=/dev/null
+. /lib/functions/keepalived/common.sh
+
+if [ $# -lt 3 ]; then
+	echo "$0 <vrrp_instance> <peer> <rsync_dir>"
+	exit 1
+fi
+
+VRRP_INSTANCE=$1
+PEER=$2
+RSYNC_DIR=$3
+
+INOTIFY_ACTIONS="create,delete,modify,move,moved_to,moved_from"
+INOTIFY_PID=""
+TMP_DIR=/tmp/keepalived
+FIFO_FILE="$TMP_DIR"/inotifywait-$PEER.fifo
+
+daemonize_inotifywait() {
+	/usr/bin/inotifywait -q -r --exclude '/\..+' -o "$FIFO_FILE" -m "$RSYNC_DIR" -e ${INOTIFY_ACTIONS} 2> /dev/null &
+	INOTIFY_PID="$!"
+}
+
+main() {
+	local inotify_action inotify_dir inotify_file
+	local source_file target_file
+
+	[ ! -d "$TMP_DIR" ] && mkdir "$TMP_DIR"
+	mkfifo "${FIFO_FILE}" || exit 1
+
+	daemonize_inotifywait
+
+	while read -r inotify_dir inotify_action inotify_file; do
+		source_file="${inotify_dir}${inotify_file}"
+		target_file=$(echo "${inotify_dir}" | sed -e "s:${RSYNC_DIR}::g")"${inotify_file}"
+
+		log_debug "received $target_file ($inotify_action) in $source_file"
+
+		ACTION=NOTIFY_SYNC TYPE=peer NAME=$PEER INSTANCE=$VRRP_INSTANCE \
+			RSYNC_SOURCE="${source_file}" RSYNC_TARGET="${target_file}" \
+			/sbin/hotplug-call keepalived
+	done < "$FIFO_FILE"
+}
+
+TRAP() {
+	[ -n "$INOTIFY_PID" ] && kill "$INOTIFY_PID"
+	[ -e "$FIFO_FILE" ] && rm -f "$FIFO_FILE"
+}
+
+trap TRAP TERM INT
+main "$@"
--- a/net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
+++ b/net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
@@ -1,0 +1,162 @@
+#!/bin/sh
+
+# shellcheck disable=SC2039
+
+# shellcheck source=/dev/null
+. /lib/functions.sh
+# shellcheck source=/dev/null
+. /lib/functions/keepalived/common.sh
+
+RSYNC_USER=$(get_rsync_user)
+RSYNC_HOME=$(get_rsync_user_home)
+
+utc_timestamp() {
+	date -u +%s
+}
+
+update_last_sync_time() {
+	uci_revert_state keepalived "$1" last_sync_time
+	uci_set_state keepalived "$1" last_sync_time "$(utc_timestamp)"
+}
+
+update_last_sync_status() {
+	local cfg="$1"
+	shift
+	local status="$*"
+
+	uci_revert_state keepalived "$cfg" last_sync_status
+	uci_set_state keepalived "$cfg" last_sync_status "$status"
+}
+
+ha_sync_send() {
+	local cfg=$1
+	local address ssh_key ssh_port sync_list sync_dir sync_file count
+	local ssh_options ssh_remote dirs_list files_list
+	local changelog="/tmp/changelog"
+
+	config_get address "$cfg" address
+	[ -z "$address" ] && return 0
+
+	config_get ssh_port "$cfg" ssh_port 22
+	config_get sync_dir "$cfg" sync_dir "$RSYNC_HOME"
+	[ -z "$sync_dir" ] && return 0
+	config_get ssh_key "$cfg" ssh_key "$sync_dir"/.ssh/id_rsa
+	config_get sync_list "$cfg" sync_list
+
+	for sync_file in $sync_list $(sysupgrade -l); do
+		[ -f "$sync_file" ] && {
+			dir="${sync_file%/*}"
+			list_contains files_list "${sync_file}" || append files_list "${sync_file}"
+		}
+		[ -d "$sync_file" ] && dir="${sync_file}"
+		list_contains dirs_list "${sync_dir}${dir}" || append dirs_list "${sync_dir}${dir}"
+	done
+
+	ssh_options="-y -y -i $ssh_key -p $ssh_port"
+	ssh_remote="$RSYNC_USER@$address"
+
+	# shellcheck disable=SC2086
+	timeout 10 ssh $ssh_options $ssh_remote mkdir -m 755 -p "$dirs_list /tmp" || {
+		log_err "can not connect to $address. check key or connection"
+		update_last_sync_time "$cfg"
+		update_last_sync_status "$cfg" "SSH Connection Failed"
+		return 0
+	}
+
+	# shellcheck disable=SC2086
+	if rsync --out-format='%n' --dry-run -a --relative ${files_list} -e "ssh $ssh_options" --rsync-path="sudo rsync" "$ssh_remote":"$sync_dir" > "$changelog"; then
+		count=$(wc -l "$changelog")
+		if [ "${count%% *}" = "0" ]; then
+			log_debug "all files are up to date"
+			update_last_sync_time "$cfg"
+			update_last_sync_status "$cfg" "Up to Date"
+			return 0
+		fi
+	else
+		log_err "rsync dry run failed for $address"
+		update_last_sync_time "$cfg"
+		update_last_sync_status "$cfg" "Rsync Detection Failed"
+		return 0
+	fi
+
+	# shellcheck disable=SC2086
+	rsync -a --relative ${files_list} ${changelog} -e "ssh $ssh_options" --rsync-path="sudo rsync" "$ssh_remote":"$sync_dir" || {
+		log_err "rsync transfer failed for $address"
+		update_last_sync_time "$cfg"
+		update_last_sync_status "$cfg" "Rsync Transfer Failed"
+	}
+
+	log_info "keepalived sync is compeleted for $address"
+	update_last_sync_time "$cfg"
+	update_last_sync_status "$cfg" "Successful"
+}
+
+ha_sync_receive() {
+	local cfg=$1
+	local ssh_pubkey
+	local name auth_file home_dir
+
+	config_get name "$cfg" name
+	config_get sync_dir "$cfg" sync_dir "$RSYNC_HOME"
+	[ -z "$sync_dir" ] && return 0
+	config_get ssh_pubkey "$cfg" ssh_pubkey
+	[ -z "$ssh_pubkey" ] && return 0
+
+	home_dir=$sync_dir
+	auth_file="$home_dir/.ssh/authorized_keys"
+
+	if ! grep -q "^$ssh_pubkey$" "$auth_file" 2> /dev/null; then
+		log_notice "public key not found. Updating"
+		echo "$ssh_pubkey" > "$auth_file"
+		chown "$RSYNC_USER":"$RSYNC_USER" "$auth_file"
+	fi
+
+	/etc/init.d/keepalived-inotify enabled || /etc/init.d/keepalived-inotify enable
+	/etc/init.d/keepalived-inotify running "$name" || /etc/init.d/keepalived-inotify start "$name"
+}
+
+ha_sync_each_peer() {
+	local cfg="$1"
+	local c_name="$2"
+	local name sync sync_mode
+
+	config_get name "$cfg" name
+	[ "$name" != "$c_name" ] && return 0
+
+	config_get sync "$cfg" sync 0
+	[ "$sync" = "0" ] && return 0
+
+	config_get sync_mode "$cfg" sync_mode
+	[ -z "$sync_mode" ] && return 0
+
+	case "$sync_mode" in
+		send) ha_sync_send "$cfg" ;;
+		receive) ha_sync_receive "$cfg" ;;
+	esac
+}
+
+ha_sync_peers() {
+	config_foreach ha_sync_each_peer peer "$1"
+}
+
+ha_sync() {
+	config_list_foreach "$1" unicast_peer ha_sync_peers
+}
+
+main() {
+	local lockfile="/var/lock/keepalived-rsync.lock"
+
+	if ! lock -n "$lockfile" > /dev/null 2>&1; then
+		log_info "another process is already running"
+		return 1
+	fi
+
+	config_load keepalived
+	config_foreach ha_sync vrrp_instance
+
+	lock -u "$lockfile"
+
+	return 0
+}
+
+main "$@"
--- a/net/keepalived/files/etc/hotplug.d/keepalived/501-rpcd
+++ b/net/keepalived/files/etc/hotplug.d/keepalived/501-rpcd
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+# shellcheck source=/dev/null
+. /lib/functions/keepalived/hotplug.sh
+
+set_service_name rpcd
+
+set_reload_if_sync
+
+add_sync_file /etc/config/rpcd
+
+keepalived_hotplug
--- a/net/keepalived/files/etc/hotplug.d/keepalived/505-system
+++ b/net/keepalived/files/etc/hotplug.d/keepalived/505-system
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+# shellcheck source=/dev/null
+. /lib/functions/keepalived/hotplug.sh
+
+set_service_name system
+
+set_reload_if_sync
+
+add_sync_file /etc/config/system
+
+keepalived_hotplug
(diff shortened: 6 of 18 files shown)


net/keepalived/Makefile
```diff
@@ -204,38 +204,34 @@ TARGET_CFLAGS += -I$(LINUX_DIR)
 
 define Package/keepalived/install
 	$(INSTALL_DIR) $(1)/usr/sbin
-	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/sbin/keepalived \
-		$(1)/usr/sbin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/sbin/keepalived $(1)/usr/sbin/
```
feckert, line 207: Nothing has changed there?

net/keepalived/Makefile
```diff
@@ -204,38 +204,34 @@ TARGET_CFLAGS += -I$(LINUX_DIR)
...
 
 ifeq ($(CONFIG_KEEPALIVED_LVS),y)
 	$(INSTALL_DIR) $(1)/usr/bin
-	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/genhash \
-		$(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/genhash $(1)/usr/bin/
```
feckert, line 211: Nothing has changed there?

net/keepalived/Makefile
```diff
@@ -204,38 +204,34 @@ TARGET_CFLAGS += -I$(LINUX_DIR)
...
 endif
 
 	$(INSTALL_DIR) $(1)/etc/keepalived
 	$(INSTALL_CONF) $(PKG_INSTALL_DIR)/etc/keepalived/keepalived.conf.sample \
 		$(1)/etc/keepalived/keepalived.conf
 
 	$(INSTALL_DIR) $(1)/etc/uci-defaults
-	$(INSTALL_BIN) ./files/etc/uci-defaults/keepalived \
-		$(1)/etc/uci-defaults/keepalived
+	$(INSTALL_BIN) ./files/etc/uci-defaults/keepalived $(1)/etc/uci-defaults/keepalived
```
feckert, line 219: Nothing has changed there?

net/keepalived/Makefile
```diff
@@ -204,38 +204,34 @@ TARGET_CFLAGS += -I$(LINUX_DIR)
...
 
 	$(INSTALL_DIR) $(1)/etc/init.d
-	$(INSTALL_BIN) ./files/keepalived.init \
-		$(1)/etc/init.d/keepalived
+	$(INSTALL_BIN) ./files/etc/init.d/keepalived $(1)/etc/init.d/keepalived
```
feckert, line 222: Nothing has changed there?

net/keepalived/Makefile
```diff
@@ -204,38 +204,34 @@ TARGET_CFLAGS += -I$(LINUX_DIR)
...
 
 	$(INSTALL_DIR) $(1)/etc/config
-	$(INSTALL_CONF) ./files/keepalived.config \
-		$(1)/etc/config/keepalived
+	$(INSTALL_CONF) ./files/etc/config/keepalived $(1)/etc/config/keepalived
```
feckert, line 225: Nothing has changed there?

net/keepalived/Makefile
```diff
@@ -204,38 +204,34 @@ TARGET_CFLAGS += -I$(LINUX_DIR)
...
 
 	$(INSTALL_DIR) $(1)/etc
-	$(INSTALL_CONF) ./files/keepalived.user \
-		$(1)/etc/keepalived.user
+	$(INSTALL_CONF) ./files/etc/keepalived.user $(1)/etc/keepalived.user
```
feckert, line 228: Nothing has changed there?

net/keepalived/Makefile
```diff
@@ -204,38 +204,34 @@ TARGET_CFLAGS += -I$(LINUX_DIR)
...
 
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/keepalived
-	$(INSTALL_DATA) ./files/hotplug-user \
-		$(1)/etc/hotplug.d/keepalived/01-user
+	$(INSTALL_DATA) ./files/etc/hotplug.d/keepalived/01-user $(1)/etc/hotplug.d/keepalived/01-user
+
+	$(INSTALL_DIR) $(1)/usr/libexec/rpcd
+	$(INSTALL_BIN) ./files/usr/libexec/rpcd/keepalived $(1)/usr/libexec/rpcd/keepalived
```
feckert, line 234: Nothing has changed there?

net/keepalived/Makefile
```diff
@@ -204,38 +204,34 @@ TARGET_CFLAGS += -I$(LINUX_DIR)
...
 
 	$(INSTALL_DIR) $(1)/etc
-	$(INSTALL_CONF) ./files/keepalived.user \
-		$(1)/etc/keepalived.user
+	$(INSTALL_CONF) ./files/etc/keepalived.user $(1)/etc/keepalived.user
 
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/keepalived
-	$(INSTALL_DATA) ./files/hotplug-user \
-		$(1)/etc/hotplug.d/keepalived/01-user
+	$(INSTALL_DATA) ./files/etc/hotplug.d/keepalived/01-user $(1)/etc/hotplug.d/keepalived/01-user
```
feckert, line 231: Nothing has changed there?

net/keepalived/Makefile
```diff
@@ -274,4 +270,27 @@ endif
...
+
+define Package/keepalived-sync/install
+	$(INSTALL_DIR) $(1)/etc/init.d \
+		$(1)/etc/keepalived/scripts \
+		$(1)/etc/keepalived/keys \
+		$(1)/usr/bin
+
+	$(INSTALL_BIN) ./files/etc/init.d/keepalived-inotify $(1)/etc/init.d/keepalived-inotify
+	$(INSTALL_BIN) ./files/etc/keepalived/scripts/rsync.sh $(1)/etc/keepalived/scripts/rsync.sh
+	$(INSTALL_BIN) ./files/usr/bin/keepalived-rsync-inotify.sh $(1)/usr/bin/keepalived-rsync-inotify.sh
```
feckert, line 292: Can you please wrap the block

net/keepalived/files/etc/init.d/keepalived-inotify
```diff
@@ -0,0 +1,42 @@
...
+	procd_open_instance "$name"
+	procd_set_param command /bin/sh "$PROG" "$name" "$sync_dir"
+	procd_set_param pidfile /var/run/keepalived-inotify-$name.pid
+	procd_close_instance
+}
+
+stop_keepalived_inotify_instance() {
+	local cfg=$1
+
+	killall inotifywait
```
feckert, line 31: I don't think that's so good idea.
You'll kill all the other scripts which use inotifywait!
Do you even need it? Doesn't the procd do that for you?

net/keepalived/files/etc/keepalived/scripts/rsync.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/sh
```
feckert, line 1: I would suggest to move this scripts to` /use/share/keepalived` or an other place and not to `/etc`

net/keepalived/files/usr/bin/keepalived-rsync-inotify.sh
```diff
@@ -0,0 +1,14 @@
+#/bin/sh
```
feckert, line 1: Please remove `.sh` from the file name `keepalived-rsync-inotify` that is not common in  `bin` directory

net/keepalived/files/usr/bin/keepalived-rsync-inotify.sh
```diff
@@ -0,0 +1,14 @@
...
+
+peer=$1
+rsync_dir=$2
+INOTIFY_ACTIONS="create delete modify move moved_to moved_from"
+
+ARGS="${INOTIFY_ACTIONS:+ -e ${INOTIFY_ACTIONS// / -e }}"
+
+/usr/bin/inotifywait -q -r --exclude '/\..+' -m $rsync_dir $ARGS 2>/dev/null | while read -r dir action file; do
+	RSYNC_TARGET_DIR=$(echo "${dir}" | sed -e "s:${rsync_dir}::g")
+	ACTION=SYNC RSYNC_SOURCE="${dir}${file}" RSYNC_TARGET="${RSYNC_TARGET_DIR}${file}" /sbin/hotplug-call keepalived inotify
```
feckert, line 13: Don't we have to check whether a sync has already been started?

net/keepalived/files/etc/init.d/keepalived
```diff
@@ -256,6 +256,21 @@ print_track_bfd_indent() {
 	printf '\n' >> "$KEEPALIVED_CONF"
 }
 
+print_unicast_peer_indent() {
```
feckert, line 259: Can you please move this change into a separate commit?
I assume that does have nothing to do with sync

feckert (changes requested):

net/keepalived/Makefile
jempatel: Updated

net/keepalived/Makefile
jempatel: Updated

net/keepalived/Makefile
jempatel: Updated

net/keepalived/Makefile
jempatel: Updated

net/keepalived/Makefile
jempatel: Updated

net/keepalived/Makefile
jempatel: Updated

net/keepalived/Makefile
jempatel: Updated

net/keepalived/Makefile
jempatel: Updated

net/keepalived/Makefile
jempatel: done

net/keepalived/files/etc/init.d/keepalived-inotify
jempatel: > I don't think that's so good idea.
> You'll kill all the other scripts which use inotifywait!

handled with the signal handler and fifo.

>  Do you even need it? Doesn't the procd do that for you?

I am not sure if it's possible to trigger hotplug event with it or not. 

This service keeps watch on the rsync directory for added/deleted/updated files and then it triggers hotplug event.

In the keepalived hotplug event it passes an extra argument `inotify` and variables
```sh
ACTION=SYNC
RSYNC_SOURCE=<source_file> # location of the file on the local host received from the master node
RSYNC_TARGET=<actual_file> # actual location of the file when this file should be replaced
```

If `/etc/config/dhcp` is changed on the master node, rsync.sh script will rsync that file to a backup node on configured interval.

On backup node, keepalived-rsync-inotify service, will get a modify event and it would trigger
```sh
ACTION=SYNC RSYNC_SOURCE="/usr/share/keepalived/rsync/etc/config/dhcp" \
 RSYNC_TARGET="/etc/config/dhcp" \
 /sbin/hotplug-call keepalived inotify
```

Whenever the master/backup switchover happens, to avoid delay, this hotplug event can be used to copy files to the target location and execute any action like start/stop service or execute any other command.

net/keepalived/files/usr/bin/keepalived-rsync-inotify.sh
jempatel: removed.

net/keepalived/files/etc/init.d/keepalived
jempatel: This one is actually needed.

Whenever keepalived peers are configured, to sync files, we also configure peers sync configuration. Instead of IP only peers can be saved with the name `master` or `backup` or 'other name`. which would be used as a reference in the vrrp_instance configuration same as track_interface, track_scripts, etc. And the same peer config is used in rsync.sh and inotify also.

In the `sender` (master) node, the peer saves SSH configuration, and additional files to sync.
In the `receiver` (backup) node, the peer saves SSH public key configuration.

net/keepalived/files/etc/keepalived/scripts/rsync.sh
jempatel: so in that case wouldn't It needs to be added to conf files so that it can be backed up?

net/keepalived/files/usr/bin/keepalived-rsync-inotify.sh
jempatel: rsync is triggered from the peer(which is master), on the backup node, It only keeps watch on `sync_dir`.

net/keepalived/files/etc/keepalived/scripts/rsync.sh
feckert: The question is rather do you want the file to be taken along with a system upgrade, or should it be installed by the package. As I see it, the script belongs to the `keepalived-sync` package and not to the runtime config.

Where is the rsync.sh used at all?

net/keepalived/Makefile
```diff
@@ -274,4 +274,33 @@ endif
...
+endef
+
+define Package/keepalived-sync/description
+ Keepalived Master Backup sync files
+endef
+
+define Package/keepalived-sync/install
+	$(INSTALL_DIR) $(1)/etc/init.d \
+		$(1)/etc/keepalived/scripts \
+		$(1)/etc/keepalived/keys \
```
feckert, line 291: Why do you create this directory. We are not installing anything there.

feckert (changes requested):

net/keepalived/Makefile
jempatel: We created these directories to keep all scripts and keys in this location. From the Luci application, we are allowing the upload and selection of scripts/keys from this directory only in vrrp_script and peer configuration respectively, also as /etc/keepalived/ is made part of conf files, It would also be backed up with config backup.

rsync.sh is installed in /etc/keepalived/scripts.

jempatel: >> Where is the rsync.sh used at all?

rsync needs UCI configuration of vrrp_script and track_script.

First Its configured as vrrp_script and then track_script
Now when we configure vrrp_instance, we add rsync script in a list of track_script configuration

We did not add any cron job/service because it only has to be executed when keepalived is running and as keepalived already provides a function to periodically execute the script, we leveraged that feature

Here is an example working UCI config from the backup node.

```sh
config globals 'globals'

config ipaddress
	option name 'eth0_ha'
	option address '192.168.111.101'
	option device 'eth0'
	option label_suffix 'ha'

config ipaddress
	option name 'eth3_ha'
	option address '192.168.5.1'
	option device 'eth3'
	option label_suffix 'ha'

config vrrp_script
	option name 'sync'
	option script '/etc/keepalived/scripts/rsync.sh'
	option interval '60'
	option weight '100'

config static_ipaddress
	list address 'eth0_ha'
	list address 'eth3_ha'

config track_script
	option name 'ha_sync'
	option value 'sync'
	option weight '100'

config track_interface
	option name 'eth0_ha'
	option value 'eth0'
	option weight '100'

config track_interface
	option name 'eth3_ha'
	option value 'eth3'
	option weight '100'

config peer
	option name 'master'
	option address '100.100.100.1'
	option sync '1'
	option sync_mode 'receive'
	option sync_dir '/usr/share/keepalived/rsync'
	option ssh_pubkey 'authorized_key_string'

config vrrp_instance
	option name 'backup'
	option state 'BACKUP'
	option interface 'eth1'
	option virtual_router_id '100'
	option priority '50'
	option advert_int '1'
	option nopreempt '0'
	list virtual_ipaddress 'eth0_ha'
	list virtual_ipaddress 'eth3_ha'
	option unicast_src_ip '100.100.100.2'
	list unicast_peer 'master'
	option auth_type 'PASS'
	option auth_pass 'admin'
	list track_interface 'eth0_ha'
	list track_interface 'eth3_ha'
	list track_script 'ha_sync'
```
rsync works differently based on if the node is master/backup. If the node is master, it sends configuration files to the backup node. If the node is back up it checks and updates authorized_keys.

net/keepalived/files/etc/keepalived/scripts/rsync.sh
jempatel: https://github.com/openwrt/packages/pull/19374#discussion_r974193740

net/keepalived/files/etc/keepalived/scripts/rsync.sh
```diff
@@ -0,0 +1,89 @@
...
+		list_contains DIRS ${sync_dir}${dir} || append DIRS ${sync_dir}${dir}
+	done
+
+	changed_files=$(rsync --out-format='%n' --dry-run -a --relative $FILES -e "ssh -y -i $ssh_key -p $ssh_port" $ssh_user@$address:$sync_dir | wc -l)
+	[ $changed_files -le 0 ] && return 0
+
+	ssh -y -i $ssh_key -p $ssh_port $ssh_user@$address mkdir -p $DIRS
+	rsync -a --relative $FILES -e "ssh -y -i $ssh_key -p $ssh_port" $ssh_user@$address:$sync_dir
+
+	unset DIRS FILES
```
feckert, line 33: Is that necessary? can't we make it locale variables?

net/keepalived/files/usr/bin/keepalived-rsync-inotify
```diff
@@ -0,0 +1,27 @@
...
+[ $# -lt 3 ] && exit 1
+
+vrrp_instance=$1
+peer=$2
+rsync_dir=$3
+
+INOTIFY_ACTIONS="create delete modify move moved_to moved_from"
+ARGS=${INOTIFY_ACTIONS:+ -e ${INOTIFY_ACTIONS// / -e }}
+
+fifo=/tmp/inotifywait.$$
```
feckert, line 12: Can we change the name. When I read this in the system, I don't know what it's for.
Maybe move this into a subdirectory /tmp/keepalived/ ?

feckert (changes requested):

net/keepalived/files/etc/keepalived/scripts/rsync.sh
jempatel: Got you, moved `/etc/keepalived/scripts/rsync.sh` to `/usr/share/keepalived/scripts/rsync.sh` and created soft link to `/etc/keepalived/scripts/rsync.sh` from the Makefile

net/keepalived/files/etc/init.d/keepalived-inotify
```diff
@@ -0,0 +1,50 @@
...
+
+	config_foreach start_instance peer "$vrrp_instance" "$peer"
+}
+
+process_vrrp_instance() {
+	local cfg=$1
+	local name
+
+	config_get name $cfg name
+	config_list_foreach $1 unicast_peer process_unicast_peer $name
```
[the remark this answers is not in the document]
jempatel: updated

net/keepalived/files/etc/keepalived/scripts/rsync.sh
```diff
@@ -0,0 +1,89 @@
...
+	local cfg=$1
+	local ssh_pubkey
+	local auth_file=/etc/dropbear/authorized_keys
+
+	config_get ssh_pubkey $cfg ssh_pubkey
+	[ -z "$ssh_pubkey" ] && return 0
+
+	if ! grep -q "$ssh_pubkey" "$auth_file"; then
+		echo "$ssh_pubkey" >> "$auth_file"
+		sed -i '$!N; /^\(.*\)\n\1$/!P; D' "$auth_file"
```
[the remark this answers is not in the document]
jempatel: Yes got your point,
Created regular user, used sudo rsync to take this out from global authorized_keys, and added logic to update it for the respective user's authorized_keys file only.

net/keepalived/files/etc/keepalived/scripts/rsync.sh
jempatel: updated

net/keepalived/files/usr/bin/keepalived-rsync-inotify
jempatel: updated

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
```diff
@@ -0,0 +1,153 @@
...
+	rsync -a --relative $FILES -e "ssh -y -i $ssh_key -p $ssh_port" --rsync-path="sudo rsync" $ssh_user@$address:$sync_dir || {
+		update_last_sync_time "$cfg"
+		update_last_sync_status "$cfg" "Rsync Transfer Failed"
+	}
+
+	update_last_sync_time "$cfg"
+	update_last_sync_status "$cfg" "Successful"
+}
+
+ha_sync_receive() {
```
feckert, line 75: I like this solution :+1: 

I would define a [USERID](https://github.com/openwrt/packages/blob/6d49ad9e5c5a93014cc99f1c956eaa51b36ee95f/net/gnunet/Makefile#L64) in the Makefile to add a userid and groupid for keepalived. Make a suggestion of the numbers. I do not know what is already in use in the repostitories. I would prefer to use the group `keepalived` and user `keepalived` instead of `keepalived-sync`

Instead of generation the files in `/etc/sudoers.d/*` and `<user>/.ssh` we could use a [postinst](
https://github.com/openwrt/packages/search?q=%2Fpostinst&type=) script in the Makefile. We could also remove this files by usind [postrm](https://github.com/openwrt/packages/search?q=%2Fpostrm&type=)

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
```diff
@@ -0,0 +1,153 @@
...
+ha_sync_peers() {
+	config_foreach ha_sync_each_peer peer "$1"
+}
+
+ha_sync() {
+	config_list_foreach "$1" unicast_peer ha_sync_peers
+}
+
+
+LOCK=/var/run/keepalived_rsync.lock
```
feckert, line 146: If all this is called keepalived-sync then we should also use this as a look file.

net/keepalived/files/usr/libexec/keepalived/hotplug.sh
```diff
@@ -0,0 +1,216 @@
...
+	value=${value:-0}
+	[ "$value" = "0" ] && return 1
+
+	return 0
+}
+
+_service() {
+	local rc="/etc/init.d/$SERVICE_NAME"
+
+	[ -d $rc ] && return
```
feckert, line 28: Must this not be `[ -f $rc ] || return`

net/keepalived/files/usr/libexec/keepalived/hotplug.sh
```diff
@@ -0,0 +1,216 @@
+#!/bin/sh
```
feckert, line 1: This is not an executable its a lib so it should go to for example to `/lib/functions` or to `/usr/share/keepalived/`

feckert (changes requested):
Apart from the last review changes, I would say that this now ready for testing :+1:

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
jempatel: > I would define a [USERID](https://github.com/openwrt/packages/blob/6d49ad9e5c5a93014cc99f1c956eaa51b36ee95f/net/gnunet/Makefile#L64) in the Makefile to add a userid and groupid for keepalived. Make a suggestion of the numbers.

I don't think this is helping much because it creates a user with /bin/false shell and a home directory in /var/run and there is no option to pass custom values for home and shell to internal function `add_group_and_user` to handle it, 

So even if we set it in Makefile, we still need home directory creation, permission, and ownership to be set from the same code because even if we use this, It still needs to update in `/etc/passwd` for home and shell

net/keepalived/files/usr/libexec/keepalived/hotplug.sh
jempatel: Moved to `/lib/functions/keepalived/hotplug.sh`

net/keepalived/files/usr/libexec/keepalived/hotplug.sh
jempatel: Updated.

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
jempatel: I did not get this point clearly, can you pls descripe bit more?

feckert: The lock belongs to a script for keepalived-inotify and everywhere we use a `-` .
I would like to say that we should somehow make it clearer which script owns the lock, and we should work with `-` as a separation and not with `_`.

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
feckert: If that is the case with `USERID` then I would suggest that we move the creation/deletion of a user and the other stuff to the postinst/postrm scripts.
We don't have to do that every time on script calling.
I would only add/change the SSH key in the script.

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
jempatel: changed lock filename

jempatel:
> Apart from the last review changes, I would say that this now ready for testing +1

@feckert Can you review it again? All pending comments are addressed now.

feckert:
@jempatel Sorry I'm busy at the moment, but I haven't forgotten the pullrequest.
I hope I will be able to do it next week.
I first have to set up two system with keepalived and your changes!

feckert (commented):
I have now set up and tested it.
And it works.
But I noticed that it might make sense to make the sync adjustable.
I noticed that when I switched to backup, the hostname also changed.
This is because it is in the uci system file, which is synced.
I would not have expected that.
I got totally confused. I thought I was on the wrong machine. 

The question is, does it make sense to make this configurable via the uci?
Or do you want a clone that only differs by IP (uci network config)?

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
```diff
@@ -0,0 +1,136 @@
...
+	home_dir=$sync_dir
+	auth_file="$home_dir/.ssh/authorized_keys"
+
+	if ! grep -q "^$ssh_pubkey$" "$auth_file" 2>/dev/null; then
+		echo "$ssh_pubkey" > "$auth_file"
+		chown $KEEPALIVED_USER:$KEEPALIVED_USER "$auth_file"
+	fi
+
+	/etc/init.d/keepalived-inotify enabled || /etc/init.d/keepalived-inotify enable
+	/etc/init.d/keepalived-inotify running || /etc/init.d/keepalived-inotify start
```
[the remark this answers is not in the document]
jempatel: Yes, updated the check against the instance, to start a specific instance.

jempatel:
> I have now set up and tested it. And it works. But I noticed that it might make sense to make the sync adjustable.

Yes, sync is configurable, in my config rsync script gets triggered every 60 seconds, interval can be used to adjust how frequent rsync should be triggered. To manually trigger rsync, this script can be executed and ha_sync track script can be removed from vrrp_instance configuration.
```
config vrrp_script
	option name 'ha_sync'
	option script '/etc/keepalived/scripts/rsync.sh'
	option interval '60'
	option weight '100'
```

> I noticed that when I switched to backup, the hostname also changed. This is because it is in the uci system file, which is synced. I would not have expected that. I got totally confused.I thought I was on the wrong machine.

I understand that, for that we already have taken care of it in its luci app, where we display configured mode and active mode in keepalived overview page, if luci-app-keepalived is not installed then interfaces config can be is used to identify if its active or backup.

> 
> The question is, does it make sense to make this configurable via the uci? Or do you want a clone that only differs by IP (uci network config)?

Talking about what files to sync, I actually dont want to add any service/package specific files in keepalived to sync but instead, Its better to maintain those files at package level, i.e `package-keepalived` but as this is first changeset, I have added it in this package it self, so it can be referenced and moved after feedback. Its still open for discussion, and we can change based on whatever is preferred.

For granular UCI sync configuration, I am still figuring out a better approach for this, because in `/etc/config/network`, it would still have ip rules, static routes and tunneling interfaces, which might be needed in cases of HA but it also needs consideration to not overwrite existing interfaces/device config. So I have excluded `/etc/config/network`. Other than that I am not sure what other files can be synced, but Its preffered that package it self maintains its hotplug with sync files in its own package.

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
```diff
@@ -0,0 +1,137 @@
+#!/bin/sh
```
feckert, line 1: Can you please run the script through the shellcheck.
I got a few warnings.
We should start right away and not fix it afterwards.

net/keepalived/files/usr/bin/keepalived-rsync-inotify
```diff
@@ -0,0 +1,28 @@
...
+
+cleanup() {
+	[ -n "$pid" ] && kill $pid
+	rm -f $fifo
+}
+trap cleanup TERM INT
+
+while read -r dir action file; do
+	RSYNC_TARGET_DIR=$(echo "${dir}" | sed -e "s:${rsync_dir}::g")
+	ACTION=NOTIFY_SYNC TYPE=peer INSTANCE=$vrrp_instance NAME=$peer RSYNC_SOURCE="${dir}${file}" RSYNC_TARGET="${RSYNC_TARGET_DIR}${file}" /sbin/hotplug-call keepalived
```
feckert, line 27: Can you break the line so that it is not so long?

I also think it wouldn't be bad if we could add a few more log messages / debug messages here.
I did that afterwards, to understand the code better, when what happens in the code.

I also think it would be good if we outsourced everything to a main function.
Then it looks the same as a C program?

You could also run shellcheck against this.

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
```diff
@@ -0,0 +1,137 @@
...
+}
+
+ha_sync() {
+	config_list_foreach "$1" unicast_peer ha_sync_peers
+}
+
+
+LOCK=/var/lock/keepalived-rsync.lock
+
+lock -n $LOCK > /dev/null 2>&1 || exit 0
```
feckert, line 132: I have now noticed a few times during testing that the lock has not been released.
Can we not use the locking facility from [procd](https://github.com/openwrt/openwrt/blob/master/package/system/procd/files/procd.sh#L48)? Like I do with the [mwan3](https://github.com/openwrt/packages/blob/master/net/mwan3/files/etc/hotplug.d/iface/15-mwan3#L7)?

While we're at it, you can outsource a main. Then it looks like a C program

feckert (changes requested):
Can you please look at my comments again? Everything else looks good so far. I'd say it's ready to merge then.

feckert:
> > The question is, does it make sense to make this configurable via the uci? Or do you want a clone that only differs by IP (uci network config)?
> 
> Talking about what files to sync, I actually dont want to add any service/package specific files in keepalived to sync but instead, Its better to maintain those files at package level, i.e `package-keepalived` but as this is first changeset, I have added it in this package it self, so it can be referenced and moved after feedback. Its still open for discussion, and we can change based on whatever is preferred.

I agree, we can do that later if necessary.

> For granular UCI sync configuration, I am still figuring out a better approach for this, because in `/etc/config/network`, it would still have ip rules, static routes and tunneling interfaces, which might be needed in cases of HA but it also needs consideration to not overwrite existing interfaces/device config. So I have excluded `/etc/config/network`. Other than that I am not sure what other files can be synced, but Its preffered that package it self maintains its hotplug with sync files in its own package.

We could consider to use uci section to synchronize and define which to keep
However, this would not be done via a file base but more with uci commands.
I think we will have to discuss this at a later date.
Thanks again for your create work :+1:

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
jempatel: > I have now noticed a few times during testing that the lock has not been released.
This might happen in case the process is already running or is stuck in any comment, possibly wrong ssh key might cause this I can look into

> Can we not use the locking facility from [procd](https://github.com/openwrt/openwrt/blob/master/package/system/procd/files/procd.sh#L48)? Like I do with the [mwan3](https://github.com/openwrt/packages/blob/master/net/mwan3/files/etc/hotplug.d/iface/15-mwan3#L7)?
rync is an independent script, and that was the reason I have not used it with procd also it does not collide with the init script, So can you elaborate on why we need that?

> While we're at it, you can outsource a main. Then it looks like a C program
I am still not clear on this, do you mean to add the main function in the same file or in and lib file (in the case of lib, its an independent script)

feckert: > > I have now noticed a few times during testing that the lock has not been released.
> 
> This might happen in case the process is already running or is stuck in any comment, possibly wrong ssh key might cause this I can look into
> 

We can also fix that later. For now, we need a better test depth.

> > Can we not use the locking facility from [procd](https://github.com/openwrt/openwrt/blob/master/package/system/procd/files/procd.sh#L48)? Like I do with the [mwan3](https://github.com/openwrt/packages/blob/master/net/mwan3/files/etc/hotplug.d/iface/15-mwan3#L7)?
> 
> rync is an independent script, and that was the reason I have not used it with procd also it does not collide with the init script, So can you elaborate on why we need that?
> 

As mentioned above, I had a problem with the lock, so I would have suggested using the API from procd.That was just a suggestion. If it comes to problems then we can still change it. So leave it as it is now. 


> > While we're at it, you can outsource a main. Then it looks like a C program
> 
> I am still not clear on this, do you mean to add the main function in the same file or in and lib file (in the case of lib, its an independent script)

I mean in the same file.

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
jempatel: Updated

net/keepalived/files/usr/share/keepalived/scripts/rsync.sh
jempatel: Updated

feckert (approved):

net/keepalived/files/usr/bin/keepalived-rsync-inotify
jempatel: Added log functions in `/lib/functions/keepalived/common.sh`. Default logging for info/err is added and a few more with debug-level logging.

By default log_debug would not log anything, if the `KEEPALIVED_DEBUG=1` variable is set in the script, debug logs would be logged in the logger.

feckert (approved):
Thanks for your work :+1: run another test over the weekend. If everything looks good so far, then we can merge on Monday.

feckert:
Then let's try it. Thank you for your work! :+1:

net/keepalived/Makefile
```diff
@@ -274,4 +274,103 @@ endif
 
 endef
 
+define Package/keepalived-sync
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Keepalived Master and Backup Synchronization
+  DEPENDS:= +keepalived +rsync +inotifywait +sudo +@BUSYBOX_CUSTOM +@BUSYBOX_CONFIG_TIMEOUT
```
feckert, line 281: @jempatel I just noticed that you are activating `@BUSYBOX_CUSTOM` here to use `@BUSYBOX_CONFIG_TIMEOUT`. Can't we achieve this in another way?

feckert: Can we not use the timeout of the [coreutils](https://github.com/openwrt/packages/blob/master/utils/coreutils/Makefile#L36) (coreutils-timeout).
I don't want to change the default settings of the busybox.
I will prepare a pullrequest for this.

jempatel: Yes, we can. I did not use it to avoid adding another additional package.

PR #19378, opened 2022-09-15
Revert "curl: update to 7.85.0"

stangri:
Reverts openwrt/packages#19377

* Accidentally changed the default SSL lib

### DIFF
--- a/net/curl/Config.in
+++ b/net/curl/Config.in
@@ -4,7 +4,7 @@ comment "SSL support"
 
 choice
 	prompt "Selected SSL library"
-	default LIBCURL_WOLFSSL
+	default LIBCURL_MBEDTLS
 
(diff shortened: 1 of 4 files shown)


BKPepe:
Missing Signed-off-by.

stangri:
> Missing Signed-off-by.

I've created this revert from WebUI which I now regret. I'm struggling to push an update containing SOB. Should I have created the revert in my own repo and sent a PR for that? Can I just create a PR with the proper SSL lib dependency instead of reverting this and then sending another PR?

Thanks!

BKPepe:
Yes, force-push is not allowed inside OpenWrt repositories, thus you can push revert with SoB directly or create pull request, that's up to you. Firstly, do the revert and then you can do whatever you want.. :)

PR #19385, opened 2022-09-16
collectd: disable nut plugin (incompatibility with nut 2.8.0)

hnyman:
Temporarily disable nut plugin by marking it BROKEN. Apparently the upstream collectd plugin is not yet compatible with nut 2.8.0.

I tried the upstream PR 4043, but it does not help, as it apparently requires at least one specific UPS type to be defined/compiled.

        nut . . . . . . . . . no (required data types for NUT API were not detected) (dependency error)

If there is a requirement to specify some UPS data type when compiling nut, it might be problematic for the generic OpenWrt buildbot context.

Maintainer: me


cotequeiroz:
@hnyman I think #19373 fixes this. The data type detection fails because `time_t` is used without `time.h` in `upsclient.h`, failing to compile no regardless of the actual data type being checked.

PR #19387, opened 2022-09-16
bcp38: Migration to nftables

ne20002:
Maintainer: Toke Høiland-Jørgensen [toke@toke.dk](mailto:toke@toke.dk)
Compile tested: no compile, script
Run tested: mvebu, WRT3200ACM, Openwrt 22.03.0

Description: this is a migration of bcp38 package to nftables

This implementation creates a new ip table with priority 2 (can be set in options) to filter spoofed packages.


ne20002:
This is the updated pull request for master

cc maintainer @tohojo

champtar:
For the IPv4 part, can't this be implemented with reverse path filtering + blockhole routes ?

net/bcp38/files/run.sh
```diff
@@ -22,83 +26,91 @@ add_bcp38_rule()
...
-    	local enabled
+	local section="$1"
+	local enabled
 	local interface
+	local priority
 	local detect_upstream
 	config_get_bool enabled "$section" enabled 0
 	config_get interface "$section" interface
 	config_get detect_upstream "$section" detect_upstream
+	config_get priority "$section" priority "2"
```
tohojo, line 55: Does this really need to be configurable? Presumably we know the priority of the default openwrt firewall chains, and we always want to hook in before this? So I think just setting the priority as a fixed value at the top of the file (with the chain name) is fine...

tohojo (changes requested):
Thanks for doing the conversion! This looks mostly reasonable, with just one small nit...

Oh, and also please fix the commit author and signoff so the bot is happy...

tohojo:
> For the IPv4 part, can't this be implemented with reverse path filtering + blockhole routes ?

This filtering is interface-specific. You don't generally want to blackhole all of 10.0.0.0/8 :)

But yeah, for IPv6, source-specific routing means this is not needed at all...

net/bcp38/files/run.sh
ne20002: Actually the priority 2 adds the bcp38 chains **after** the standard fw4 chains. 
I believe, this is a thing that can't be decided as it depends on the users situation. With nftables we place the hook either in front of or after the fw4 chains.
I the old fw3 implementation we needed to use the user_rules which have been placed before the rest of the fw3 rules but have been part of the fw4 rules.
With nftables chains are processed one after each other and only if a packet is accepted it is passed to the next chain. If it passes all chains accepted, it is accepted. Any drop or reject skips processing of further chains.
So the question is: do we need/want to filter this before or after the fw4 firewall? Placing it before means, each packet (that is not offloaded) is passing through the bcp filter before it is processed by fw4 firewall. Placing it after means, only packets that have not been already dropped in fw4 will be checked. In either way the filtering is applied correctly. 
So it depends on how many of those packets come in and I believe that fw4 is filtering more packets than bcp38, so I prefer to place this filter after fw4. But with the configuration option, the user may choose another solution that fits better for him. 
 
Having dedicated chains means that the filter does nor interfere with fw4 in any way as fw4 firewall only acts on fw4 table.

tohojo: Right, if the semantics is such that a packet needs an accept verdict by all chains, it's fine to execute the BCP filter after the regular firewall. I still don't think there's any real need to make it configurable, though; we can always add a knob if someone turns out to have a real use case for that...

ne20002:
I'll do another set of changes. Having now looked a bit deeper into nftables I believe the rules can be optimized.

Issue #19352, opened 2022-09-13
Bandwidthd  version 2.0.1-35-5 sending SIGSEGV to bandwidthd for invalid read access from 00000018


frankgrant:
Maintainer: @\<github-user> (find it by checking history of the package Makefile)
Environment: 
arch MediaTek MT7621 ver:1 eco:3
model Ubiquiti EdgeRouter X
OpenWrt version  OpenWrt 22.03.0 r19685-512e76967f / LuCI openwrt-22.03 branch git-22.245.77528-487e58a

Description:
Getting kernel error messages:
do_page_fault(): sending SIGSEGV to bandwidthd for invalid read access from 00000018
epc = 77e76534 in libc.so[77dea000+a9000]
ra  = 77e76410 in libc.so[77dea000+a9000]

No graphs are being produced.
It worked fine with OpenWrt 22.02

padre-lacroix:
OK: I have to check on a 32 bit virtual OpenWRT router (or if I can get my jand on a Ubiquiti Edge router) that the fix #19392 fixes that problem.  Unless somehow has confirmed that it works properly.

frankgrant:
Any outlook when this fix will be added to the openwrt packages?

padre-lacroix:
Sorry,  I have been pretty sick int he last 10 days.  The opened pull request (19392) is for the Master branch.  I will approve the pull request and it will be available in the nightly build.

Fort the stable version (openwrt-22.03) another pull request will have to be made in that branch to be available in the packages in the stable version of openwrt.

frankgrant:
Updated to the latest  version 2.0.1-35-6 and still get the error
do_page_fault(): sending SIGSEGV to bandwidthd for invalid read access from 00000018
epc = 77e9e534 in libc.so[77e12000+a9000]
ra  = 77e9e410 in libc.so[77e12000+a9000]

padre-lacroix:
Sorry that there is still a bug!
I will have to investigate this a little bit further (obviously).
I will try to replicate the bug in a router (I have a couple siting on a shelf) and then I will try to fix it when I will be able to replicate it.

As I mentionned previously, I have been pretty sick lately, and I am still a little bit sick.  So bear with me if it takes some time.

padre-lacroix:
OK, I replicated the bug in a Archer C7 v2.
Now to fix it!  I will start investigating the problem this weekend.

PR #19392, opened 2022-09-18
bandwidthd: fix format warnings

neheb:
Should fix crashing errors under musl 1.2

Maintainer: @padre-lacroix 

Fixes: https://github.com/openwrt/packages/issues/19352

I think.

### DIFF
--- a/utils/bandwidthd/Makefile
+++ b/utils/bandwidthd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bandwidthd
 PKG_VERSION:=2.0.1-35
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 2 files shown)


neheb:
Oh cool there's an up to date fork: https://codeberg.org/post-factum/bandwidthd

padre-lacroix:
OK, I will look at it.  Thismake takes a few days has I have to create a new virtual machine for building OpenWRT (the current one that I have is old).

For the up to date fork, I will look at it for the solution delpoyed.  But, it seems to be the same problem (but I am not an expert).

@neheb : did you try to execute the patch that you made?  This is why I have to compile it myself to see if the executable works on a 32 bit virtual machine (that is the problem isn't it: on a 32 bit router, this would not work properly).

neheb:
It executes fine under malta.

padre-lacroix:
OK: I approve the pull request.

Sorry that it took me so long: I have been pretty sick in the last 10 days.

frankgrant:
Updated to the latest version 2.0.1-35-6 and still get the error
do_page_fault(): sending SIGSEGV to bandwidthd for invalid read access from 00000018
epc = 77e9e534 in libc.so[77e12000+a9000]
ra = 77e9e410 in libc.so[77e12000+a9000]

PR #19395, opened 2022-09-18
realtek-poe: Update package to v1.0

mrnuke:
v1.0 fixes a bug with MCU communication, and allows shutting down individual ports over ubus. A summarry of commands can be shown with:

	ubus -v list poe

CC: @hurricos @bmork  @aparcar @walterav1984 @svanheule


walterav1984:
Is testing this auto build package "mips_4kec-packages " of [CI58](https://github.com/Hurricos/realtek-poe/actions/runs/2941724495) good enough compared to building this snapshot openwrt package?

walterav1984:
Just tested this CI58 package quickly on 22.03.0 GS1900-hp24v1 and I can confirm `ubus -v list poe` shows a extra `manage` option. 

However it can still show [status "unknown"](https://github.com/Hurricos/realtek-poe/issues/10) for configured ports in config after cold/re-boot.

Hurricos:
I just tested [CI58](https://github.com/Hurricos/realtek-poe/actions/runs/2941724495) on a GS1900-24HPv1 and can't reproduce after two cold-reboots.

ACK this PR.

walterav1984:
Sometimes I needed more like 5 reboots/cold boots before it manifests. It looks like the final 22.03.0 is booting faster than RC (just running default config) I think it has something todo with timing starting the POE to early?

Or maybe its my own specific switch since I share the same model like @Hurricos.

PR #19411, opened 2022-09-19
lua-eco: adapt SSL choice to libwolfsslcpu-crypto

cotequeiroz:
Maintainer: @zhaojh329
Compile tested: mediatek/mt7622 aarch64_cortex-a53, master
Run tested: tested with `make menuconfig`--does not change the installed binaries.

Description:
libwolfsslcpu-crypto has to be taken into consideration when selecting the default SSL backend.

### DIFF
--- a/lang/lua-eco/Makefile
+++ b/lang/lua-eco/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lua-eco
 PKG_VERSION:=1.0.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


lang/lua-eco/Makefile
```diff
@@ -82,7 +84,7 @@ define Package/lua-eco-ssl/config
 		depends on PACKAGE_libopenssl
 	config LUA_ECO_WOLFSSL
 		bool "wolfSSL"
-		depends on PACKAGE_libwolfssl
+		depends on PACKAGE_libwolfssl || PACKAGE_libwolfssl-cpu-crypto
```
McGiverGim, line 87: It seems the package has changed to:

```suggestion
		depends on PACKAGE_libwolfssl || PACKAGE_libwolfsslcpu-crypto
```
This is the commit
https://github.com/openwrt/openwrt/commit/d08c9da43cf364712e947d5faa3ab84d995dd0ec

cotequeiroz: I was forgetting about this.  Thanks.

PR #19420, opened 2022-09-21
bind: bump to 9.18.7

nmeyerhans:
Fixes multiple security issues:

CVE-2022-38178 - Fix memory leak in EdDSA verify processing

CVE-2022-3080 - Fix serve-stale crash that could happen when
			stale-answer-client-timeout was set to 0 and there was
			a stale CNAME in the cache for an incoming query

CVE-2022-2906 - Fix memory leaks in the DH code when using OpenSSL 3.0.0
			and later versions. The openssldh_compare(),
			openssldh_paramcompare(), and openssldh_todns()
			functions were affected

CVE-2022-2881 - When an HTTP connection was reused to get
			statistics from the stats channel, and zlib
			compression was in use, each successive
			response sent larger and larger blocks of memory,
			potentially reading past the end of the allocated
			buffer

CVE-2022-2795 - Prevent excessive resource use while processing large
			delegations

Signed-off-by: Noah Meyerhans <frodo@morgul.net>

Maintainer: me 
Compile tested: x86_64
Run tested: x86_64

Description:

### DIFF
--- a/net/bind/Makefile
+++ b/net/bind/Makefile
@@ -9,7 +9,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bind
-PKG_VERSION:=9.18.4
+PKG_VERSION:=9.18.7
 PKG_RELEASE:=$(AUTORELEASE)
 USERID:=bind=57:bind=57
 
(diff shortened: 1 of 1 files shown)


BKPepe (approved):
LGTM

PR #19436, opened 2022-09-24
snowflake: add package

dangowrt:
Package [Tor's Snowflake](https://snowflake.torproject.org) system components so users can offer e.g. a standalone Snowflake proxy on their routers or other devices.

Compile tested: arm_cortex-a53
Run tested: Belkin RT3200, snapshot

More users should try `snowflake-proxy`, esp. also on MIPS systems.

### DIFF
--- a/net/snowflake/Makefile
+++ b/net/snowflake/Makefile
@@ -1,0 +1,144 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=snowflake
+PKG_VERSION:=2.3.0
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL=https://git.torproject.org/pluggable-transports/snowflake.git
+PKG_SOURCE_VERSION:=v$(PKG_VERSION)
+PKG_MIRROR_HASH:=e323598c959ea336927bf8dc4a96b02e88de73cb8cf165221465ca2b76cd2c48
+
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=Daniel Golle <daniel@makrotopia.org>, Nick Hainke <vincent@systemli.org>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=git.torproject.org/pluggable-transports/snowflake.git/v2
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/snowflake/Default
+  TITLE:=Snowflake
+  URL:=https://snowflake.torproject.org/
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+  SECTION:=net
+  CATEGORY:=Network
+  USERID:=snowflake:snowflake
+endef
+
+define Package/snowflake-broker
+$(call Package/snowflake/Default)
+  TITLE+= Broker
+endef
+
+define Package/snowflake-client
+$(call Package/snowflake/Default)
+  TITLE+= Client
+endef
+
+define Package/snowflake-distinctcounter
+$(call Package/snowflake/Default)
+  TITLE+= Distinct Counter
+endef
+
+define Package/snowflake-probetest
+$(call Package/snowflake/Default)
+  TITLE+= Probe test
+endef
+
+define Package/snowflake-proxy
+$(call Package/snowflake/Default)
+  TITLE+= Proxy
+endef
+
+define Package/snowflake-server
+$(call Package/snowflake/Default)
+  TITLE+= Server
+endef
+
+define Package/snowflake/description/Default
+Snowflake is a system that allows people from all over the world to
+access censored websites and applications. Similar to how VPNs assist
+users in getting around Internet censorship, Snowflake helps you avoid
+being noticed by Internet censors by making your Internet activity appear
+as though you're using the Internet for a regular video or voice call.
+endef
+
+define Package/snowflake-broker/description
+$(call Package/snowflake/description/Default)
+
+This package provides the Snowflake broker service.
+endef
+
+define Package/snowflake-client/description
+$(call Package/snowflake/description/Default)
+
+This package contains the Snowflake client which provides the bridge to TOR.
+endef
+
+define Package/snowflake-distinctcounter/description
+$(call Package/snowflake/description/Default)
+
+This package provides the Snowflake distinct counter service.
+endef
+
+define Package/snowflake-probetest/description
+$(call Package/snowflake/description/Default)
+
+This package provides the Snowflake probe test.
+endef
+
+define Package/snowflake-proxy/description
+$(call Package/snowflake/description/Default)
+
+This package provides the standalone Snowflake proxy.
+endef
+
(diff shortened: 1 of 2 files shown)


net/snowflake/files/snowflake-proxy.init
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh /etc/rc.common
+
+START=99
+
+USE_PROCD=1
+NAME=snowflake-proxy
+PROG=/usr/bin/$NAME
```
PolynomialDivision, line 7: `$Name`?

net/snowflake/Makefile
```diff
@@ -0,0 +1,144 @@
...
+	$(INSTALL_BIN) $(GO_PKG_BUILD_BIN_DIR)/distinctcounter $(1)/usr/bin/snowflake-distinctcounter
+endef
+
+define Package/snowflake-probetest/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(GO_PKG_BUILD_BIN_DIR)/probetest $(1)/usr/bin/snowflake-probetest
+endef
+
+define Package/snowflake-proxy/install
+	$(INSTALL_DIR) $(1)/usr/bin $(1)/etc/init.d
```
PolynomialDivision, line 129: I don't know why, but all(?) other go packages do something like this:
`$(call GoPackage/Package/Install/Bin,$(PKG_INSTALL_DIR))`
in front of installing the binaries.

net/snowflake/Makefile
```diff
@@ -0,0 +1,144 @@
...
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL=https://git.torproject.org/pluggable-transports/snowflake.git
+PKG_SOURCE_VERSION:=v$(PKG_VERSION)
+PKG_MIRROR_HASH:=e323598c959ea336927bf8dc4a96b02e88de73cb8cf165221465ca2b76cd2c48
+
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=Daniel Golle <daniel@makrotopia.org>
```
PolynomialDivision, line 14: You can also add me as maintainer so I can bump versions too, if you want. ;)

net/snowflake/Makefile
dangowrt: Because most Go package builds only result in a single package containing everything. In order to split up the components into several packages I choose to not use `GoPackage/Package/Install/Bin`.

PolynomialDivision: Than maybe some other packages are wrong? Look at the [Makefile from nebula](https://github.com/openwrt/packages/blob/master/net/nebula/Makefile). They have two modules, nebula and nebula-cert. They copy the nebula binary from [here](https://github.com/openwrt/packages/blob/master/net/nebula/Makefile#L63) and for cert [here](https://github.com/openwrt/packages/blob/master/net/nebula/Makefile#L71). However, they use `$(PKG_INSTALL_DIR)` instead of `$(GO_PKG_BUILD_BIN_DIR)`.

I don't know why they do it. I think your solution is better since it seems unnecessary to install and than copy again the same binary. Or maybe I misinterpret the Makefile code of nebula.

dangowrt:
Also posted to anti-censorship-team mailing list:
https://lists.torproject.org/pipermail/anti-censorship-team/2022-September/000252.html

PolynomialDivision:
I would also go for updating the tor package. Or are you also planing to do this? ;)

net/snowflake/Makefile
dangowrt: Yeah, I've seen that and then looked at golang-package.mk and the scripts in lang/golang/ and concluded that `$(call GoPackage/Package/Install/Bin,$(1))` is a useful thing to do if really everything should end up in one package. But in case things should be split up there is no need for all the moving/copying of files as accessing them directly is just as easy and convenient.

dangowrt:
> I would also go for updating the tor package. Or are you also planing to do this? ;)

Oh yes, please go ahead.

net/snowflake/Makefile
PolynomialDivision: Thanks. Resolved.

PolynomialDivision:
Okay. However, I'm on the flixtrain. Internet is very slow right now, but I will do my best. :)

I runtested the proxy stuff already the whole day. I would also runtest the PR again, however that will take some time due to internet access. So ACK from my side if the CI also is fine with it. :)

PolynomialDivision (approved):

PolynomialDivision:
WTF, Flixbtrain blocks the torproject.org website???
```
Cloning into 'snowflake-2.3.0'...
fatal: unable to access 'https://git.torproject.org/pluggable-transports/snowflake.git/': Failed to connect to git.torproject.org port 443 after 5 ms: Connection refused
make[2]: *** [Makefile:140: /home/nick/openwrt/dl/snowflake-2.3.0.tar.xz] Error 128
make[2]: Leaving directory '/home/nick/packages/net/snowflake'
time: package/bla/net/snowflake/compile#0.27#0.12#1.96
    ERROR: package/bla/net/snowflake failed to build.
make[1]: *** [package/Makefile:116: package/bla/net/snowflake/compile] Error 1
```

dangowrt:
> WTF, Flixtrain blocks the torproject.org website???

Hm, maybe use `torsocks` then to access it ;) And if they really block TOR itself as well, use a bridge...

PolynomialDivision:
Somehow they answer the dns queries with: `torproject.org.		900	IN	A	0.0.0.0`.

```
dig torproject.org

; <<>> DiG 9.18.7 <<>> torproject.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 59568
;; flags: qr rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 51921273ea4a2f0b (echoed)
;; QUESTION SECTION:
;torproject.org.			IN	A

;; ANSWER SECTION:
torproject.org.		900	IN	A	0.0.0.0

;; Query time: 76 msec
;; SERVER: 1.1.1.1#53(1.1.1.1) (UDP)
;; WHEN: Sat Sep 24 18:11:35 CEST 2022
;; MSG SIZE  rcvd: 71

dig @8.8.8.8 torproject.org         

; <<>> DiG 9.18.7 <<>> @8.8.8.8 torproject.org
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 4158
;; flags: qr rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: c7a3b434ce14c612 (echoed)
;; QUESTION SECTION:
;torproject.org.			IN	A

;; ANSWER SECTION:
torproject.org.		900	IN	A	0.0.0.0

;; Query time: 26 msec
;; SERVER: 8.8.8.8#53(8.8.8.8) (UDP)
;; WHEN: Sat Sep 24 18:14:39 CEST 2022
;; MSG SIZE  rcvd: 71


```

PR #19445, opened 2022-09-24
[21.02] snowflake: add package

PolynomialDivision:
Package Tor's Snowflake system components so users can offer e.g. a standalone Snowflake proxy on their routers or other devices.

Signed-off-by: Daniel Golle [daniel@makrotopia.org](mailto:daniel@makrotopia.org)
(cherry picked from commit https://github.com/openwrt/packages/commit/cf120a7effd5d13a7f705b5eb9d22410b73d71f3)

Maintainer: @dangowrt, me
Compile tested: tbd
Run tested: tbd

### DIFF
--- a/net/snowflake/Makefile
+++ b/net/snowflake/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=snowflake
 PKG_VERSION:=2.3.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=https://git.torproject.org/pluggable-transports/snowflake.git
--- a/net/snowflake/Makefile
+++ b/net/snowflake/Makefile
@@ -1,0 +1,144 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=snowflake
+PKG_VERSION:=2.3.0
+PKG_RELEASE:=1
+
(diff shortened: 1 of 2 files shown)


BKPepe:
Is there a reason, why you want to add this package to version, which is a feature complete or we can say also old stable?

PolynomialDivision:
21.02 is still not dead. Even some devices like wdr4900v1 are not supported anymore in 22.03. So I thought we keep it in sync with 22.03 also. If a user wants to add his router to "snowflake-network", we should make that as easy as possible.

BKPepe:
Compile and run tested on 21.02?

Akira25:
> Compile and run tested on 21.02?

Compiling on 21.02 was done by the CI. As all the runs are green, I assume it was successful. I run-tested the artifacts for mips-24kc on my `tplink,archer-c5-v1`. I didn't see any issues with this, so we should be safe on merging this.

BKPepe:
It is the stable branch, you needs to be tested on the real device, which OP does not do, that's why I was asking that. Thanks @Akira25 for testing it! Quality over quantity, guys. :)

What should be done is to add ``test.sh`` file to be even run tested in CI, when we have it. On the other hand, you should not count on CI, too much. It just compiles a few packages, it does not do full builds. It can reveal some missing dependencies, etc.

1715173329:
> It just compiles a few packages, it does not do full builds. It can reveal some missing dependencies, etc.

Anyway, this is a standalone and dependency-free golang package ;)

PR #19458, opened 2022-09-26
gnunet: add without-libjose to fix buildbot compilation

PolynomialDivision:
The build process picks up the libjose library although it is not selected as dependency by gnunet.

Fixes errors in the form of:
  Package gnunet-rest is missing dependencies for the following libraries:
  libjose.so.0

Maintainer: @dangowrt 
Compile tested: banana pi r64
Run tested: no

### DIFF
--- a/net/gnunet/Makefile
+++ b/net/gnunet/Makefile
@@ -44,7 +44,8 @@ CONFIGURE_ARGS+= \
 	--with-png=$(STAGING_DIR)/usr \
 	--with-pulse=$(STAGING_DIR)/usr \
 	--with-libunistring-prefix=$(STAGING_DIR)/usr \
-	--with-microhttpd=$(STAGING_DIR)/usr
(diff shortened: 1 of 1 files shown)


cotequeiroz:
Careful with this.  According to [configure.ac:683](https://git.gnunet.org/gnunet.git/tree/configure.ac?h=v0.17.5#n683), the dependency is legit.  Without it, `libgnunetrest_openid_connect.so`, will not be built.  We are _**trying** to install it_ in https://github.com/openwrt/packages/blob/671594bec2f108588d365e99d0f3fca6363d4162/net/gnunet/Makefile#L298

I highlighted **trying** above because the package does not fail if any files are missing except the last one.  They are copied inside a for loop that does not check exit status.  I have a patch series fixing all of this that I'll post shortly.

PolynomialDivision:
Thanks. :)

PR #19459, opened 2022-09-26
libmariadb: update to 3.1.18

PolynomialDivision:
The official download mirror does not host 3.1.12 anymore. Update to newest version to solve the issue.

Maintainer: noone
Compile tested: banana pi r64
Run tested: no


BKPepe:
I don't think it is a good idea to update libmaria without mariadb, thus it needs to be run tested.

neheb:
ping @miska

miska:
It should be fine. It is a minor bump and given the nature, the connector and the library are meant to work and tested across various versions.

PR #19465, opened 2022-09-26
zabbix: update to 6.2.3

ttocsr:
Updated to 6.2.3 version.
Refreshed patches.

Compile tested: aarch64 cortex-a53
Run tested: aarch64 cortex-a53

Signed-off-by: Scott Roberts <ttocsr@gmail.com>

### DIFF
--- a/admin/zabbix/Makefile
+++ b/admin/zabbix/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=zabbix
-PKG_VERSION:=5.0.18
+PKG_VERSION:=6.2.3
 PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://cdn.zabbix.com/zabbix/sources/stable/5.0/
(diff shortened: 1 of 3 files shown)


champtar:
Thanks! Can you tell us what was tested?

ttocsr:
Mostly tested agentd with TLS against a server running zabbix-server 6.2.3.
What are you looking for and I can run specific tests.

champtar:
That's fine, just want to know if someone comes and say zabbix-server is now broken if it was tested or not.

ttocsr:
@champtar what do you think about changing the default from ZABBIX_NOSSL to OpenSSL or GnuTLS?
Right now if you have a proper implementation of zabbix (one with ssl), then you can't opkg upgrade zabbix-agentd.  If you do, it breaks your implementation.  The only way for someone to get agentd with ssl, is to compile their own.

Or should we make two(or three?) versions of agentd available on opkg? 
zabbix-agentd-openssl, zabbix-agentd-gnutls, zabbix-agentd

champtar:
Someone started the multiple variant approach but never finished: https://github.com/openwrt/packages/pull/17846
I'm happy to review a new PR

ttocsr:
Thanks @champtar I put in a new PR #19486
I think it corrects everything  for multiple variants.

PR #19470, opened 2022-09-27
ddns-scripts: clean up code a bit to make it easier to understand

rmilecki:
```
ddns-scripts: add explicit "-d" switch for Dry Run

It was a bit confusing to use *verbosity* level for Dry Run mode. Add
explicity switch for it and designed DRY_RUN variable to make code
easier to understand.
```
```
ddns-scripts: rename variable: s//ERR_UPDATE/RETRY_COUNT/

Rename variable to make code easier to understand. This variable
specifies how many times in row ddns script tried to update IP without a
success.

Previous name ("ERR_UPDATE") didn't suggest it was for counting
anything. It also didn't specify was error was it related to.
```
```
ddns-scripts: replace IP type (name) "local" with "current"

Local suggests something related to the local network or available
locally only. All that code related to the "local" IP was actually
dealing with *current* device external IP address. Using name "current"
should make code a bit easier to understand.
```
```
ddns-scripts: rename variable: s/retry_count/retry_max_count/

Rename variable to make code easier to understand. This variable
specifies how many times ddns script should try to send a request.

Previous name ("retry_count") suggested it was for *counting* attempts.
```

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
-PKG_RELEASE:=26
+PKG_RELEASE:=27
 
 PKG_LICENSE:=GPL-2.0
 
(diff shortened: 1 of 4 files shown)


Ansuel (approved):

feckert:
@rmilecki 

LGTM from my side. But we should not forget to increment the [PKG_RELASE](
https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L11).

rmilecki:
@feckert thanks! Should that be one increment per Pull Request or per commit?

feckert:
Its ok to increment by one

PR #19473, opened 2022-09-27
openvpn: Applied Tunnleblik XOR patch

hamid-nazari:
Maintainer: @hamid-nazari
Compile tested: (mips_24kc, TP-Link Archer C5 v1 (ath79-generic-c5-v1), OpenWrt v22.03)
Run tested: None

Description:
I've applied tunnelblick XOR patches from https://github.com/lawtancool/openvpn-windows-xor/tree/master/tunnelblick-patches


feckert:
Can you please describe this in more in detail for what it is needed and also add this in the commit message for this repository?
Also, the patches must/should have a clean header and a valid `Signed-off-by` with a proper description so that others know what the patches are for.
The best way to do this is checkout the correct openvpn version found in the openwrt Makefile and apply the changes on top.  Then generate patches with `git format patch HEAD~5`. And add this changes to this pullrequest.

You also have to increment the PKG_RELEASE in the openwrt Makefile.

The correct maintainer could be found in the commit history so please ping them. To review your changes.
https://github.com/openwrt/packages/commits/master/net/openvpn

AuthorReflex:
Those patches is not approved by OpenVPN developers, so in my opinion is no good idea include this patches by default in OpenWRT

PR #19475, opened 2022-09-28
ddns-scripts: store device's current IP in <section>.curr_ip file

rmilecki:
```
This should be useful for user interfaces (e.g. LuCI) to display DDNS
status. If current IP differs from registered IP it may suggest user
that there is a problem. It could happen if DNS change hasn't propagated
yet or DDNS provider refused change for some reason.
```

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
-PKG_RELEASE:=27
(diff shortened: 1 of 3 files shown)


Ansuel:
@rmilecki instead of saving the current ip, can't we just use ddns functions to get the current one at runtime?

I agree that this would be a handy feature to also understand what ddns is detecting with the current configuration as the ip to comunicate with the ddns provider.

BKPepe:
I see that this needs to be rebased.

GeorgeSapkin:
No updates in over three years and merge conflicts. Closing as stale. Feel free to reopen as necessary.

PR #19476, opened 2022-09-28
lcd4linux: minor init improvments

feckert:
Maintainer: me 
Run tested: x86_64, APU3 OpenWrt master, tests done

Description:

* Until now, the configuration must be stored under '/etc/lcd4linux.conf'. So that the configuration can also be changed dynamically, it makes sense to store this under /tmp and load them from this directory. The init script first checks whether there is a configuration under '/etc/lcd4linux.conf' and only then does it try to find it under '/tmp/lcd4linux.conf'. If there is no configuration, an error message is shown.

* Setting the DEBUG variable in the init script to '1' enables the
lcd4linux verbose mode, by setting the arg '-vv'. The option also
redirects the error and stdout to the syslog.

### DIFF
--- a/utils/lcd4linux/Makefile
+++ b/utils/lcd4linux/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lcd4linux
-PKG_RELEASE:=7
+PKG_RELEASE:=8
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/feckert/lcd4linux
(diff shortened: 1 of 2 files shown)


feckert:
@aparcar @ynezz I don't think this is a problem with my pull request. 
Since you are the CI experts can you take a look at this?
So on my build there is a link to ./compile

```
ls -la /home/feckert/workspace/openwrt/build_dir/target-x86_64_musl/lcd4linux-full/lcd4linux-2019-07-19-0b83bb3a | grep compile
lrwxrwxrwx  1 feckert feckert    130 Oct  4 12:17 compile -> /home/feckert/workspace/openwrt/build/openwrt/staging_dir/target-x86_64_musl/../host/share/automake-1.15/compile
```

PR #19480, opened 2022-09-29
modemmanager: Adding support for 'rat' option to set the Radio access technology for the device.

fjamarquez:
Maintainer: @mips171  <nicholas@nbembedded.com>, @aleksander0m 
Compile tested: ath79
Run tested: Tested with RM500Q-AE modem

Description:

With 'rat' option you can select the Radio access technology for the device. The RAT options supported are 3G, 4G and 5G.

Signed-off-by: Francisco Jose Alvarez <francisco.alvarez@galgus.net>

### DIFF
--- a/net/modemmanager/README.md
+++ b/net/modemmanager/README.md
@@ -24,6 +24,7 @@ Once installed, you can configure the 2G/3G/4G modem connections directly in
         option iptype      'ipv4'
         option lowpower    '1'
         option signalrate  '30'
+        option rat  	   '4G'
 
 Only 'device' and 'proto' are mandatory options, the remaining ones are all
 optional.
@@ -38,3 +39,6 @@ It will default to 'ipv4' if not given.
 
 The 'signalrate' option set's the signal refresh rate (in seconds) for the device.
 You can call signal info with command: mmcli -m 0 --signal-get
+
+The 'rat' option set's the Radio access technology for the device.
+You can get the RAT info with command: mmcli -m 0 or with AT+QENG="servingcell AT command.
--- a/net/modemmanager/files/modemmanager.proto
+++ b/net/modemmanager/files/modemmanager.proto
@@ -343,6 +343,7 @@ proto_modemmanager_init_config() {
 	proto_config_add_string	 password
 	proto_config_add_string	 pincode
 	proto_config_add_string	 iptype
+	proto_config_add_string	 rat
 	proto_config_add_int	 signalrate
 	proto_config_add_boolean lowpower
 	proto_config_add_defaults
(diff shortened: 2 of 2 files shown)


net/modemmanager/files/modemmanager.proto
```diff
@@ -404,6 +406,33 @@ proto_modemmanager_setup() {
 		return 1
 	}
 
+	# set the RAT
+	if [ -n "${rat}" ]; then
+		if [ "${rat}" == "automatic" ]; then
+			echo "setting modem RAT to automatic mode"
+			if [ "$(mmcli --modem="${device}" | grep supported | grep 5gnr)" ]; then
```
aleksander0m, line 413: This is not completely right I'm afraid. The "5gnr", "lte" and "gsm-umts" strings being grepped here are reporting "capabilities", not "modes", it is not exactly the same in the context of ModemManager.

For "automatic", the setting to use is simple: `--set-allowed-modes=any --set-preferred-mode=none`. This setting is built-in ModemManager and will try to automatically map to a setting that means "auto", regardless of the proto in use. There is no need to guess anything here.

Then, for the "5G" rat selection, the modem is assuming that 4G is also supported, while that may not be true. A 5GNR-only device doesn't have to support 4G.

And then, for the "4G" and "3G" rat selections, the user is selecting **only** those rats. E.g. a 4G/3G/2G modem selecting 4G would be selecting 4G-only, which may not be completely desired.

And 2G is completely ignored in the patch, while it shouldn't. There are a lot of 4G+2G modems (without 3G) as well.

My suggestion here would be:
 * First, let the proto handler receive exactly which allowed modes and preferred modes the user wants. So not a "rat" setting, but explicitly two settings saying "allowed-modes" and "preferred-modes", and pass those settings to ModemManager with the `--set-preferred-mode` and `--set-allowed-modes` commands without doing any explicit validation. The validation of what is supported or not is done by the modem itself anyway, so if a user is manually configuring those in the config file, the user should be aware of which settings are supported.
 * Second, in the luci-proto-modemmanager package, we can expose a list of which are the allowed/preferred combinations the user can choose, as reported in the `mmcli -m $device` output. Any of those combinations listed will be supported by the modem (and any combination NOT in the list will NOT be supported).

I'm afraid that this patch has made a lot of assumptions about a very specific modem model being used, I think it can be improved to make it more generic.

net/modemmanager/README.md
```diff
@@ -38,3 +39,6 @@ It will default to 'ipv4' if not given.
 
 The 'signalrate' option set's the signal refresh rate (in seconds) for the device.
 You can call signal info with command: mmcli -m 0 --signal-get
+
+The 'rat' option set's the Radio access technology for the device.
+You can get the RAT info with command: mmcli -m 0 or with AT+QENG="servingcell AT command.
```
aleksander0m, line 44: AT+QENG is Quectel specific, I would document here that `mmcli -m 0` reports the list of allowed/preferred modes the specific modem supports, and so the given settings should be one of the combinations given in the list.

fjamarquez:
I think you are right, this patch may cause conflicts with other modem models. I will try to make the new modification when I have some time. 
Thank you for your attention.

feckert:
Lets close them and waiting for another implementation try from @fjamarquez

PR #19486, opened 2022-09-29
zabbix: add variants for SSL support

ttocsr:
opkg does not offer ssl varients of zabbix-agentd. resolve this by adding ssl varients.

Compile tested: aarch64 cortex-a53
Run tested: aarch64 cortex-a53

Signed-off-by: Scott Roberts <ttocsr@gmail.com>

Supersedes: https://github.com/openwrt/packages/pull/17846

### DIFF
--- a/admin/zabbix/Makefile
+++ b/admin/zabbix/Makefile
@@ -24,32 +24,11 @@ PKG_INSTALL:=1
 PKG_BUILD_PARALLEL:=1
 
 PKG_CONFIG_DEPENDS:= \
-  CONFIG_ZABBIX_GNUTLS \
-  CONFIG_ZABBIX_OPENSSL \
   CONFIG_ZABBIX_MYSQL \
   CONFIG_ZABBIX_POSTGRESQL
 
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/nls.mk
-
-define Package/zabbix-agentd/config
-comment "SSL support"
-
-choice
-        prompt "Selected SSL library"
-        default ZABBIX_NOSSL
-
-        config ZABBIX_OPENSSL
-                bool "OpenSSL"
-
-        config ZABBIX_GNUTLS
-                bool "GnuTLS"
-
-        config ZABBIX_NOSSL
-                bool "No SSL support"
-
-endchoice
-endef
 
 define Package/zabbix-server/config
 comment "Database Software"
@@ -74,12 +53,31 @@ define Package/zabbix/Default
   TITLE:=Zabbix
   URL:=https://www.zabbix.com/
   USERID:=zabbix=53:zabbix=53
-  DEPENDS += $(ICONV_DEPENDS) +libpcre +zlib +ZABBIX_GNUTLS:libgnutls +ZABBIX_OPENSSL:libopenssl
+  DEPENDS+=$(ICONV_DEPENDS) +libpcre +zlib
 endef
 
 define Package/zabbix-agentd
   $(call Package/zabbix/Default)
   TITLE+= agentd
+  PROVIDES:=zabbix-agentd
+  VARIANT:=nossl
+  DEFAULT_VARIANT:=1
+endef
+
+define Package/zabbix-agentd-openssl
+  $(call Package/zabbix/Default)
+  TITLE+= agentd (with OpenSSL)
+  DEPENDS+= +libopenssl
+  PROVIDES:=zabbix-agentd
+  VARIANT:=openssl
+endef
+
+define Package/zabbix-agentd-gnutls
+  $(call Package/zabbix/Default)
+  TITLE+= agentd (with GnuTLS)
+  DEPENDS+= +libgnutls
+  PROVIDES:=zabbix-agentd
+  VARIANT:=gnutls
 endef
 
 define Package/zabbix-extra-mac80211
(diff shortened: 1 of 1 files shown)


feckert:
I have tested your PR in my own build.
I get now the following error message:
```
Package zabbix-sender is missing dependencies for the following libraries:
libcrypto.so.1.1
libssl.so.1.1
```

So I think we also need to add a build variant for zabbix-sender.
```
diff --git a/admin/zabbix/Makefile b/admin/zabbix/Makefile
index 19c2a898e23317b2ee8bdb6631e88202879c91d1..7a0b615eaa21cca012b8b850574517aaf152f5a2 100644
--- a/admin/zabbix/Makefile
+++ b/admin/zabbix/Makefile
@@ -101,6 +101,25 @@ endef
 define Package/zabbix-sender
   $(call Package/zabbix/Default)
   TITLE+= sender
+  PROVIDES:=zabbix-sender
+  VARIANT:=nossl
+  DEFAULT_VARIANT:=1
+endef
+
+define Package/zabbix-sender-openssl
+  $(call Package/zabbix/Default)
+  TITLE+= sender (with OpenSSL)
+  DEPENDS+= +libopenssl
+  PROVIDES:=zabbix-sender
+  VARIANT:=openssl
+endef
+
+define Package/zabbix-sender-gnutls
+  $(call Package/zabbix/Default)
+  TITLE+= sender (with GnuTLS)
+  DEPENDS+= +libgnutls
+  PROVIDES:=zabbix-sender
+  VARIANT:=gnutls
 endef

 define Package/zabbix-get
@@ -300,6 +319,8 @@ endef
 define Package/zabbix-sender/install
        $(call Package/zabbix/install/bin,$(1),sender)
 endef
+Package/zabbix-sender-openssl/install = $(Package/zabbix-sender/install)
+Package/zabbix-sender-gnutls/install = $(Package/zabbix-sender/install)

 define Package/zabbix-get/install
        $(call Package/zabbix/install/bin,$(1),get)
@@ -327,6 +348,8 @@ $(eval $(call BuildPackage,zabbix-extra-mac80211))
 $(eval $(call BuildPackage,zabbix-extra-network))
 $(eval $(call BuildPackage,zabbix-extra-wifi))
 $(eval $(call BuildPackage,zabbix-sender))
+$(eval $(call BuildPackage,zabbix-sender-openssl))
+$(eval $(call BuildPackage,zabbix-sender-gnutls))
 $(eval $(call BuildPackage,zabbix-server))
 $(eval $(call BuildPackage,zabbix-server-frontend))
 $(eval $(call BuildPackage,zabbix-proxy))
--
```

ttocsr:
zabbix-sender has never had ssl support in the make file.
I'm not sure why you're getting that error that ssl is a dependency. 
Are you just looking to also add ssl support for zabbix-sender like we do for zabbix-agentd?

feckert:
You're right about that.
I had another commit in my patch set.
From my point of view, I would say that this works and could be merged.
If there are no objections from maintainer of this package @champtar , I would merge this in the next few days.

We can then continue with the following variants:
* Add ssl variant also for zabbix-sender, zabbix-proxy and zabbix-server
* Add variants for zabbix-server and zabbix-proxy with the SQL backen (postgres and mysql) are also interesting
* Add a variant for zabbix-proxy with sqlite3 backend as I know this does not work for zabbix-server

The SQL variants could be problematic since there are already the individual ssl variants!

ttocsr:
Does it make sense for me to just add all the ssl variants here in this PR?
Then you could add the PR for SQL work?

champtar:
@feckert ack to merge this when you feel it's ready

ttocsr:
@feckert let me know if I missed something.

admin/zabbix/Makefile
```diff
@@ -74,68 +53,194 @@ define Package/zabbix/Default
...
+  DEPENDS+= +libgnutls
+  PROVIDES:=zabbix-agentd
+  VARIANT:=gnutls
 endef
 
 define Package/zabbix-extra-mac80211
   $(call Package/zabbix/Default)
   TITLE+= discovery/userparameters for mac80211
-  DEPENDS = +zabbix-agentd @PACKAGE_MAC80211_DEBUGFS @KERNEL_DEBUG_FS
+  DEPENDS= +zabbix-agentd @PACKAGE_MAC80211_DEBUGFS @KERNEL_DEBUG_FS
```
feckert, line 86: That's a small thing and a nitpick from my side but, this is not part of the commit and is not necessary, so either don't do it or outsource it to a separate commit.

admin/zabbix/Makefile
```diff
@@ -74,68 +53,194 @@ define Package/zabbix/Default
...
 endef
 
 define Package/zabbix-extra-network
   $(call Package/zabbix/Default)
   TITLE+= discovery/userparameters for network
-  DEPENDS = +zabbix-agentd +libubus-lua +lua
+  DEPENDS= +zabbix-agentd +libubus-lua +lua
```
feckert, line 92: See my comment above

admin/zabbix/Makefile
```diff
@@ -74,68 +53,194 @@ define Package/zabbix/Default
...
 endef
 
 define Package/zabbix-extra-wifi
   $(call Package/zabbix/Default)
   TITLE+= discovery/userparameters for wifi
-  DEPENDS = +zabbix-agentd +libiwinfo-lua +libubus-lua +lua
+  DEPENDS= +zabbix-agentd +libiwinfo-lua +libubus-lua +lua
```
feckert, line 98: See my comment above

admin/zabbix/Makefile
```diff
@@ -74,68 +53,194 @@ define Package/zabbix/Default
...
+  PROVIDES:=zabbix-get
+  VARIANT:=gnutls
+endef
+
+define Package/zabbix-server/Default
   $(call Package/zabbix/Default)
   TITLE+= server
-  DEPENDS += +ZABBIX_POSTGRESQL:libpq \
+  DEPENDS+= +ZABBIX_POSTGRESQL:libpq \
     +ZABBIX_MYSQL:libmariadbclient \
```
feckert, line 153: See my comment above

admin/zabbix/Makefile
```diff
@@ -74,68 +53,194 @@ define Package/zabbix/Default
...
+  PROVIDES:=zabbix-server
+  DEPENDS+= +libgnutls
+  VARIANT:=gnutls
+endef
+
+define Package/zabbix-server-frontend/Default
   $(call Package/zabbix/Default)
   TITLE+= server-frontend
-  DEPENDS += +php8 +php8-cgi +ZABBIX_POSTGRESQL:php8-mod-pgsql +ZABBIX_MYSQL:php8-mod-mysqli \
+  DEPENDS+= +php8 +php8-cgi +ZABBIX_POSTGRESQL:php8-mod-pgsql +ZABBIX_MYSQL:php8-mod-mysqli \
```
feckert, line 184: See my comment above

admin/zabbix/Makefile
```diff
@@ -74,68 +53,194 @@ define Package/zabbix/Default
...
+  PROVIDES:=zabbix-server-frontend
+  DEPENDS+= +libgnutls
+  VARIANT:=gnutls
+endef
+
+define Package/zabbix-proxy/Default
   $(call Package/zabbix/Default)
   TITLE+= proxy
-  DEPENDS += +ZABBIX_POSTGRESQL:libpq \
+  DEPENDS+= +ZABBIX_POSTGRESQL:libpq \
```
feckert, line 215: Move this is a separate commit. See my comment above.

admin/zabbix/Makefile
```diff
@@ -74,68 +53,194 @@ define Package/zabbix/Default
...
 
-define Package/zabbix-proxy
+define Package/zabbix-server-frontend
+  $(call Package/zabbix-server-frontend/Default)
+  PROVIDES:=zabbix-server-frontend
+  VARIANT:=nossl
+  DEFAULT_VARIANT:=1
+endef
+
+define Package/zabbix-server-frontend-openssl
```
feckert, line 196: The server frontend is a PHP application. And as I understand it, there is no difference between openssl and gnutls. If there is a variant needed, then the only difference is whether we use MySQL or PostgreSQL (but this is the next setup on update zabbix)

feckert (changes requested):
Looks pretty good so far. Please have a look at my comments

feckert (approved):
Very nice. Works as expected :+1:

PR #19494, opened 2022-09-30
An explanation for the force_dns option

birdie-github:
Signed-off-by: Artem S. Tashkinov <aros@gmx.com>

Maintainer: @stangri
Compile tested: not needed, it's a comment
Run tested: not needed, it's a comment

Description: adding an explanation for the force_dns option

### DIFF
--- a/net/https-dns-proxy/files/https-dns-proxy.config
+++ b/net/https-dns-proxy/files/https-dns-proxy.config
@@ -1,6 +1,6 @@
 config main 'config'
 	option update_dnsmasq_config '*'
-	option force_dns '1'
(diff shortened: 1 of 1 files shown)


stangri:
The description is:

* inaccurate: it implies that ports 53 and 853 are always blocked which is not the case.
* unnecessary: the [README](https://docs.openwrt.melmac.net/https-dns-proxy/#force_dns) already covers the `force_dns` option in a great detail; on top of that the ports affected by that setting are listed right there in the config file.

I'd be inclined to accept the PR if the comment contains a link to the README listed above instead of the suggested description.

birdie-github:
Suit yourself.

PR #19496, opened 2022-09-30
jq: fix hash mismatch problem

bradfordzhang:
Maintainer: @ratkaj 
Compile tested: x86 openwrt 22.03.0
Run tested: x86 openwrt 22.03.0 test done

Description:
change jq source from github to openwrt source, because openwrt hacked/changed jq

[#lede10195](https://github.com/coolsnowwolf/lede/issues/10195) mentioned this problem.

### DIFF
--- a/utils/jq/Makefile
+++ b/utils/jq/Makefile
@@ -12,11 +12,14 @@ PKG_VERSION:=1.6
 PKG_RELEASE:=2
 PKG_LICENSE:=BSD
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://github.com/stedolan/jq/releases/download/jq-$(PKG_VERSION)/
-PKG_HASH:=9625784cf2e4fd9842f1d407681ce4878b5b0dcddbcd31c6135114a30c71e6a8
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/stedolan/jq.git
+PKG_SOURCE_DATE:=2018-11-02
+PKG_SOURCE_VERSION:=2e01ff1fb69609540b2bdc4e62a60499f2b2fb8e
+PKG_MIRROR_HASH:=ee52d2dce67dfbea633020d11a908d5818145a2bf87e5a0a17c0ff6885a428a7
 PKG_MAINTAINER:=Marko Ratkaj <markoratkaj@gmail.com>
 
+PKG_FIXUP:=autoreconf
 PKG_INSTALL:=1
 
 include $(INCLUDE_DIR)/package.mk


Ansuel:
@bradfordzhang this happen when someone force push stuff to the repo and we have same release but different tar... Something strange is happening there...

bradfordzhang:
jq-1.6.tar(1).gz from https://github.com/stedolan/jq/releases/download/jq-1.6/jq-1.6.tar.gz
jq-1.6.tar(2).gz from https://sources.cdn.openwrt.org/jq-1.6.tar.gz
jq-1.6.tar(3).gz from https://sources.openwrt.org/jq-1.6.tar.gz

bradfordzhang:
> @bradfordzhang this happen when someone force push stuff to the repo and we have same release but different tar... Something strange is happening there... 

so weird😂

Ansuel:
@bradfordzhang very suspicious not wired

bradfordzhang:
> @bradfordzhang very suspicious not wired 

my bad english😂

Ansuel:
> 
> can we set two hash so that any one of the two hash match can pass the test.

This is a problematic situation honestly... and actually a security risk of some sort... 
If you want the direct answer NO. 

We can't have 2 hash for the same package... This is because one release have an unique hash and if that changes there could be only 2 reason:
- the download is corrupted
- someone did something strange and the package changed silently

Now we don't know what has changed and sometimes this is really bad... (push a clean zip... then modify it and insert virus in it and propose the same release like nothing happen)

In short the package maintainer should handle this and decide what should be done... force push should never be used and in any case a release should not be force pushed to prevent these kind of security concern.

bradfordzhang:
> > 
> > can we set two hash so that any one of the two hash match can pass the test.
> 
> This is a problematic situation honestly... and actually a security risk of some sort... 
> If you want the direct answer NO. 
> 
> We can't have 2 hash for the same package... This is because one release have an unique hash and if that changes there could be only 2 reason:
> - the download is corrupted
> - someone did something strange and the package changed silently
> 
> Now we don't know what has changed and sometimes this is really bad... (push a clean zip... then modify it and insert virus in it and propose the same release like nothing happen)
> 
> In short the package maintainer should handle this and decide what should be done... force push should never be used and in any case a release should not be force pushed to prevent these kind of security concern.

you're right. but jq has force pushed tag. that causes problem😂

bradfordzhang:
@Ansuel hi! should aria2c continue download from next mirror when hash mismatch?

bradfordzhang:
@Ansuel can we update cdn jq package to github version?

neheb:
> @bradfordzhang this happen when someone force push stuff to the repo and we have same release but different tar... Something strange is happening there...

this tarball is not a github generated one. Still weird.

1715173329:
image
By checking the date, you can see the author re-uploaded the archives 18 days after the release, but 0099bad30b22fe895223393bcc60f44b30fa86f9 was commited just 4 days after.

So this outdated hash in Makefile has been kept for near 4 years!
And even nobody noticed it as download.pl would automatically fallback to OpenWrt's mirror host which stored the old archive until recently aria2c support broke this feature.

Compared to old archive, the new one added missing git submodule which won't be packed by GitHub's tarball API:
```
$ git diff --stat --no-index jq-1.6-old jq-1.6
/dev/null => jq-1.6/modules/oniguruma/.git         |     1 +
/dev/null => jq-1.6/modules/oniguruma/.gitignore   |    27 +
/dev/null => jq-1.6/modules/oniguruma/AUTHORS      |     1 +
.../modules/oniguruma/CMakeLists.txt               |    86 +
/dev/null => jq-1.6/modules/oniguruma/COPYING      |    28 +
/dev/null => jq-1.6/modules/oniguruma/ChangeLog    |     0
/dev/null => jq-1.6/modules/oniguruma/HISTORY      |  2138 ++
/dev/null => jq-1.6/modules/oniguruma/INSTALL      |   370 +
/dev/null => jq-1.6/modules/oniguruma/Makefile.am  |    33 +
/dev/null => jq-1.6/modules/oniguruma/NEWS         |     0
/dev/null => jq-1.6/modules/oniguruma/README       |   189 +
/dev/null => jq-1.6/modules/oniguruma/README.ja    |   195 +
/dev/null => jq-1.6/modules/oniguruma/README.md    |   203 +
.../modules/oniguruma/cmake/dist.cmake             |   321 +
--- cut here ---
```
... but this took no effect on OpenWrt as oniguruma was [explicitly disabled](https://github.com/openwrt/packages/commit/0099bad30b22fe895223393bcc60f44b30fa86f9#diff-771fbc71be743bb9e37a80b67fe4e8ad67b3fd4f55098c0a164fbf066c8c92c7R32).

I would suggest updating the PKG_HASH in OpenWrt Makefile & removing the old archive from mirror host to keep it constant with upstream.

neheb:
Given that upstream has done this, I think it makes more sense to switch to PKG_SOURCE_PROTO:=git

1715173329:
> Given that upstream has done this, I think it makes more sense to switch to PKG_SOURCE_PROTO:=git

+1

utils/jq/Makefile
```diff
@@ -12,9 +12,10 @@ PKG_VERSION:=1.6
 PKG_RELEASE:=2
 PKG_LICENSE:=BSD
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://github.com/stedolan/jq/releases/download/jq-$(PKG_VERSION)/
-PKG_HASH:=9625784cf2e4fd9842f1d407681ce4878b5b0dcddbcd31c6135114a30c71e6a8
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/stedolan/jq.git
+PKG_SOURCE_DATE:=2018-11-02
+PKG_SOURCE_VERSION:=2e01ff1fb69609540b2bdc4e62a60499f2b2fb8e
```
1715173329, line 18: missing PKG_MIRROR_HASH

1715173329:
hummmm... need autoreconf?

bradfordzhang:
I can't.... I'm so noob

utils/jq/Makefile
```diff
@@ -12,9 +12,11 @@ PKG_VERSION:=1.6
...
+PKG_MIRROR_HASH:=ee52d2dce67dfbea633020d11a908d5818145a2bf87e5a0a17c0ff6885a428a7
 PKG_MAINTAINER:=Marko Ratkaj <markoratkaj@gmail.com>
 
 PKG_INSTALL:=1
```
1715173329, line 22:
```suggestion
PKG_FIXUP:=autoreconf
PKG_INSTALL:=1
```

Issue #18997, opened 2022-07-21
`php8:` sodium extension is missing


Livy-Inverse:
Maintainer: @mhei @flyn-org 
Environment: arm_cortex-a7_neon-vfpv4, MobiPromo CM520-79F, 22.03.0-rc5

Description:
My PHP installation requires Composer package manager. While trying to install Composer locally using the command:
```
php-cli composer.phar install --no-dev --prefer-source;
```
It complains about a lot of missing extensions. Most of them can be downloaded from separate packages, except for `sodium`. [According to PHP homepage](https://www.php.net/manual/en/sodium.installation.php), that extension is no longer available since PHP 7.2.0, and should be bundled with PHP by default. I am using 8.1.6-1 and it is not available. Please add support for that extension.

flyn-org:
Some time ago I produced a php8-pecl-sodium package for master. The application that I use that optionally relies on sodium is NextCloud, and I find that that application for some reason does not detect the sodium package despite it being installed. So it might be possible to backport php8-pecl-sodium, but there also might be something wrong in that package.

mhei:
The php8-pecl-sodium package is also available in 22.03 branch. @Livy-Inverse do you have a chance to test it? To be honest, I don't checked yet in detail, whether there is a difference between the pecl package and the extension shipped with PHP itself. However, I guess that the latter one might be more maintained than the package (last update 2020-12-06) - though since its only a wrapper, more updates might not be necessary...
@flyn-org : What do you think: would it make sense to drop the pecl package in favour of adding a package which bundles the extension?

flyn-org:
I have been poking around, and it is not clear to me which module is preferred, the PECL version or the version in PHP proper. To be honest, I have not yet found a clear description of the relationship between the two. The last update to the PECL version seems to have occurred in December of 2020. There are much newer mentions in the PHP proper Git log, so my instinct is to move away from the PECL module.

I am happy to drop the php8-pecl-sodium if the base PHP build provides a sodium module.

mhei:
Since we have `php8-pecl-sodium` in `22.03` available, I don't see what here is strictly required to do other than testing by @Livy-Inverse. In master we meanwhile moved away from the pecl package and replaced it with the vanilla packaged extension. I'm not yet sure, whether we should cherry-pick this to the `22.03` branch. Opinions?

esaaprillia:
@mhei 
sorry I commented in this thread 
are you helping me?
i want to make php7 in master but i have problem when i build it
[...]

PR #19497, opened 2022-09-30
php8: php8-pecl-sodium with php8-mod-sodium (refs #18997)

mhei:
Maintainer: me / @flyn-org 
Compile tested: mxs
Run tested: mxs

Description:

Since PHP8 meanwhile ships with a sodium extension, we can drop the PECL package in favour of it.


flyn-org:
The CI checks are still running, but I like this idea.

PR #19513, opened 2022-10-03
treewide: fix security issues by bumping all packages using libwolfssl

ynezz:
Maintainer:  @zhaojh329, @teslamint, @mhaas
Compile tested: n/a
Run tested: n/a
Description:

  
  As wolfSSL is having hard time maintaining ABI compatibility between
  releases, we need to manually force rebuild of packages depending on
  libwolfssl and thus force their upgrade. Otherwise due to the ABI
  handling we would endup with possibly two libwolfssl libraries in the
  system, including the patched libwolfssl-5.5.1, but still have
  vulnerable services running using the vulnerable libwolfssl-5.4.0.
  
  So in order to propagate update of libwolfssl to latest stable release
  done in commit [ec8fb542ec3e4](https://github.com/openwrt/openwrt/commit/ec8fb542ec3e4f584444a97de5ac05dbc2a9cde5) ("wolfssl: fix TLSv1.3 RCE in uhttpd by
  using 5.5.1-stable (CVE-2022-39173)") which fixes several remotely
  exploitable vulnerabilities, we need to bump PKG_RELEASE of all packages
  using wolfSSL library.
  
  Same bump has been done in buildroot in commit [f1b7e1434f66](https://github.com/openwrt/openwrt/commit/f1b7e1434f66a3cb09cb9e70b40add354a22e458) ("treewide:
  fix security issues by bumping all packages using libwolfssl").

### DIFF
--- a/net/coova-chilli/Makefile
+++ b/net/coova-chilli/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=coova-chilli
 PKG_VERSION:=1.6
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_RELEASE:=$(AUTORELEASE).1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/coova/coova-chilli/tar.gz/$(PKG_VERSION)?
--- a/lang/lua-eco/Makefile
+++ b/lang/lua-eco/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lua-eco
 PKG_VERSION:=1.0.0
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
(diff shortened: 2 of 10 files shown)


net/coova-chilli/Makefile
```diff
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=coova-chilli
 PKG_VERSION:=1.6
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_RELEASE:=$(AUTORELEASE).1
```
neheb, line 12: uhhh

ynezz: Is there any better solution how to bump package version without introducing dummy commit?

neheb: No. I wonder if AUTORELEASE should get treewide functionality.

ynezz: IMO this is not relevant in context of this PR and can't be backported to 22.03 and 21.02 either.

teslamint (approved):

PR #19498, opened 2022-10-01
lighttpd: remove deprecated modules

gstrauss:
Maintainer: @flyn-org
Compile tested: arm_cortex-a9 OpenWrt master

Description:
lighttpd: remove deprecated modules

All of 2022, each lighttpd release has repeated announcements deprecating some minor modules for which full replacements written in lua are available and can be run using [lighttpd mod_magnet](https://wiki.lighttpd.net/Docs_ModMagnet).  The latest release announcement can be found at: https://www.lighttpd.net/2022/9/17/1.4.67/

From the **Future Scheduled Behavior Changes** section of the annoucement:

*    Continue gradual deprecation of “mini-application” lighttpd modules
    for which mod_magnet lua implementations are better and more flexible.
    Please post on lighttpd forums to share feedback if you use these modules.
    Forums: https://redmine.lighttpd.net/projects/lighttpd/boards

*    Deprecated: mod_evasive will be removed.
    mod_evasive can be replaced by mod_magnet and a few lines of lua:
    Replacement: https://wiki.lighttpd.net/ModMagnetExamples#lua-mod_evasive
    https://wiki.lighttpd.net/AbsoLUAtion#Fight-DDoS
    https://wiki.lighttpd.net/AbsoLUAtion#Mod_Security

*    Deprecated: mod_secdownload will be removed.
    mod_secdownload can be replaced by mod_magnet and a few lines of lua:
    Replacement: https://wiki.lighttpd.net/ModMagnetExamples#lua-mod_secdownload
    mod_secdownload historically uses insecure MD5 though SHA1, SHA256 available

*    Deprecated: mod_uploadprogress will be removed.
    mod_uploadprogress can be replaced by mod_magnet and a few lines of lua:
    Replacement: https://wiki.lighttpd.net/ModMagnetExamples#lua-mod_uploadprogress

*   Deprecated: mod_usertrack will be removed.
    mod_usertrack can be replaced by mod_magnet and a few lines of lua:
    Replacement: https://wiki.lighttpd.net/ModMagnetExamples#lua-mod_usertrack
    mod_usertrack historically uses insecure MD5.

Issue #18851, opened 2022-07-02
lighttpd authentication modules depends on libnettle


stokito:
Maintainer: @flyn-org
Environment: OpenWrt 21.02.1   lighttpd/1.4.64 (ssl)

Description:

OpenWrt now uses wolfssl by default. But when I installed the lighttpd with mod-authn_file I found that now I have [libnettle8](https://openwrt.org/packages/pkgdata/libnettle8) installed:

```
# opkg whatdepends libnettle8
Root set:
  libnettle8
What depends on root set
        lighttpd-mod-authn_file 1.4.64-1        depends on libnettle8
        lighttpd 1.4.64-1       depends on libnettle8
        lighttpd-mod-webdav 1.4.64-1    depends on lighttpd
        lighttpd-mod-auth 1.4.64-1      depends on lighttpd
        lighttpd-mod-rewrite 1.4.64-1   depends on lighttpd
```

As far I understood the libnettle8 is just another crypto library.
From the [mod_authn_file source](https://github.com/lighttpd/lighttpd1.4/blob/master/src/mod_authn_file.c) it looks like it just includes [sys-crypto.h](https://github.com/lighttpd/lighttpd1.4/blob/master/src/sys-crypto.h) which may be configured any crypto lib. I.e. it can use the already installed wolfssl.

In the [lighttpd makefile](https://github.com/openwrt/packages/blob/openwrt-21.02/net/lighttpd/Makefile#L176) I see that the libnettle8 is the only needed dependency.

Is it possible to avoid installing the libnettle8?

I also have Turris Omnia which is a flavor of OpenWrt and there they are compiled the Lighttpd only with OpenSSL but the lighttpd-mod-authn_file doesn't have a dependency even to it:
```
# opkg depends lighttpd-mod-authn_file
lighttpd-mod-authn_file depends on:
        libc
        lighttpd
```

neheb:
ping @gstrauss

gstrauss:
tl;dr: The OpenWRT lighttpd base package **could be modified** to depend on WolfSSL instead of on Nettle, but that would not be my preference or recommendation for OpenWRT.

lighttpd is modular and provides multiple options for crypto and TLS.
However, there is no one-size-fits all to packaging.  More details:

[...]

PR #19537, opened 2022-10-06
lighttpd: document crypto lib options in Makefile

gstrauss:
Maintainer: @flyn-org

Description:
lighttpd: document crypto lib options in Makefile

Follow-up from https://github.com/openwrt/packages/issues/18851

I suspect there will be some discussion about how to do this, and perhaps to create an option in the config, so I will mark this as a draft.

---

Please note that due to a merge (https://github.com/openwrt/packages/commit/845d81ca0976c82829addc23e9e8b95885c910ee)  which occurred after https://github.com/openwrt/packages/pull/19498 was submitted but before the PR was merged, PKG_RELEASE was not bumped twice, as it should have been.  I do not think it a big deal since the end result is that the package was rebuilt.  However, the change made in the PR might not get to people until the next package release if they had downloaded PKG_RELEASE=2 before the PR was merged.

### DIFF
--- a/net/lighttpd/Makefile
+++ b/net/lighttpd/Makefile
@@ -9,9 +9,9 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lighttpd
 PKG_VERSION:=1.4.67
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 # release candidate ~rcX testing; remove for release
-#PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-1.4.67
+#PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://download.lighttpd.net/lighttpd/releases-1.4.x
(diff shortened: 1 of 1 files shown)


gstrauss:
@flyn-org would you please review and provide feedback?  Thanks.

flyn-org:
I think this is fine, but I do also think we should have a "make menuconfig" option (with libnettle as default). That said, I suppose we could merge this as is an add the config option later. From what I can tell, this will not change anything unless someone edits the lighttpd Makefile to redefine cryptolib.

gstrauss:
Thanks.  I'll have to refresh myself on the Makefile syntax to generate menu options.  I am open to suggestions if someone else gets to it before I do.

I am going to push a small change to allow for the option of no cryptolib, in which case lighttpd includes its own implementation of MD5 and SHA1, but SHA256 is not available in mod_auth in that case.

flyn-org:
You might want to look at net/wget and its use of variants. The utils/nano package uses the DEFAULT_VARIANT flag, which I suspect lighttpd also ought to use.

gstrauss:
I am not convinced there should be multiple variants of the lighttpd package just for the crypto library to use for MD5, SHA1, and SHA256

I pushed a change that uses 'choice' in the menu to select a crypto lib.  It would have saved me lots of time if https://openwrt.org/docs/guide-developer/packages documented 'choice' somewhere, or linked to another doc for 'choice'.  Searching the site (using the search-box on the site) for 'endchoice' turns up zero results!  'endchoice' should be an exact hit on documentation for the Makefile options.

gstrauss:
Any idea why this did not work and I needed to add back an explicit `+libpcre2` to `DEPENDS` for the build to succeed?
```
  DEPENDS:=+libpthread +LIGHTTPD_LOGROTATE:logrotate \
           $(if $(CONFIG_LIGHTTPD_PCRE2),+libpcre2,) \
           $(if $(cryptolib),+$(cryptolib),)
```

flyn-org:
https://openwrt.org/docs/guide-developer/packages says this:

PKG_BUILD_DEPENDS does not use + or @, but otherwise uses the same syntax. You can write FOO:bar to build bar if CONFIG_FOO is defined (if you want to build bar if package FOO is selected, use PACKAGE_FOO:bar). You may use !, || and && as explained above. PKG_BUILD_DEPENDS uses the name from the PKG_NAME, not the individual packages. For example, if you want to have openssl to be a build-dependency, you would write PKG_BUILD_DEPENDS:=openssl, whereas if your package depends and selects the openssl library, you'd have DEPENDS:=+libopenssl. Notice that there's no installable package named openssl: the library is libopenssl, the utility is openssl-util, but their PKG_NAME is openssl. If you need a host-built package, append /host to the PKG_NAME, e.g. PKG_BUILD_DEPENDS:=openssl/host. A package listed in PKG_BUILD_DEPENDS will be built even if it is not selected in make menuconfig.

EXTRA_DEPENDS does not accept the conditional or select dependency syntax. However, unlike DEPENDS and PKG_BUILD_DEPENDS, it is generated at package-build-time, so you may use Makefile functions to add them conditionally. For example, to get the equivalent of DEPENDS:=@SOMESYMBOL:foo +PACKAGE_somepkg:bar, you can write EXTRA_DEPENDS:=$(if $(CONFIG_SOMESYMBOL),foo) $(if $(CONFIG_PACKAGE_somepkg),bar). :!: Note that neither foo or bar will be selected in menuconfig, or guaranteed to be built before your package, even if selected somewhere else. For this reason, it is seldom used in packages.

EXTRA_DEPENDS is more often used to depend on specific versions, by adding the desired version specification in parentheses, using >=,>,<,<=,=. Make sure you add the PKG_RELEASE number if you're using '=', such as EXTRA_DEPENDS:=foo (=2.0.0-1): where foo's PKG_VERSION is 2.0.0, and PKG_RELEASE is 1.

flyn-org:
Looks good to me. @neheb, does the integration with the menu system look reasonable?

net/lighttpd/Makefile
```diff
@@ -22,13 +22,58 @@ PKG_LICENSE:=BSD-3-Clause
...
+  LIGHTTPD_CRYPTOLIB_NETTLE:nettle \
+  LIGHTTPD_CRYPTOLIB_MBEDTLS:mbedtls \
+  LIGHTTPD_CRYPTOLIB_WOLFSSL:wolfssl
 
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/meson.mk
 
+# choose crypto lib for lighttpd to use for crypto algorithms
+# (separate from lighttpd TLS modules, which are each standalone)
+ifdef CONFIG_LIGHTTPD_CRYPTOLIB_NONE)
```
rmilecki, line 65: That extra `')` seems to be a typo to me.

rmilecki:
@gstrauss, @neheb: those new dependencies don't seem to work at all. I've first hit this in January, today again.

Issue:
```
Package lighttpd is missing dependencies for the following libraries:
libnettle.so.8
```
For some reason all those `ifdef CONFIG_...` rules never evaluate to true. I have no idea why as code looks correct to me.

---

In my case it's:
```
> grep CONFIG_LIGHTTPD_CRYPTOLIB_NETTLE .config
CONFIG_LIGHTTPD_CRYPTOLIB_NETTLE=y
> grep -m 1 "^Depends:" tmp/info/.packageinfo-feeds_packages_lighttpd
Depends: +libc +libpthread +LIGHTTPD_LOGROTATE:logrotate
```
above is incorrect as there should be `+libnettle` in the `Depends:` line.

---

For a quick test I've added
```
else
  cryptolib=just-a-test
```
and that seems to work:
```
> grep -m 1 "^Depends:" tmp/info/.packageinfo-feeds_packages_lighttpd
Depends: +libc +libpthread +LIGHTTPD_LOGROTATE:logrotate +just-a-test
```
so it seems `cryptolib` variable works as expected but `ifdef`s don't.

PR #19538, opened 2022-10-06
libzip: update to 1.9.2

mhei:
Maintainer: me
Compile tested: mxs
Run tested: mxs

Description:

Also drop already mainlined patch.


mhei:
@neheb: I did some runtime testing using the tools, maybe you can test using idevicerestore?

Issue #19547, opened 2022-10-07
libwolfssl/libcurl: CVE-Fix (PR#19513) will crash https-dns-proxy


realizelol:
Maintainer: @tapper82 , @stangri , @ynezz , @neheb
Environment: OpenWrt 21.02.3 r16554-1d4dea6d4f - x86_64 - Generic v0 (ImageBuilder)

Description:
------------

I think since PR https://github.com/openwrt/packages/pull/19513 the actual packages @ repo [downloads.openwrt.org](https://downloads.openwrt.org/releases/21.02.3/packages/x86_64/packages/) will stop SSL support for curl.

Maybe in case of the missing commits in branch `openwrt-21.02` ?:
https://github.com/openwrt/packages/commits/master/net/curl/Makefile
-> https://github.com/openwrt/packages/commit/312ab68be498f0d895d8d4dc5b2f0fc857e45eb9
-> https://github.com/openwrt/packages/commit/b814eec3b59beec0e5610c3f0eda4429ba23132c
-> https://github.com/openwrt/packages/commit/03a32717bc15d8dd0c99e200dd64ae0bbd558c35
-> https://github.com/openwrt/packages/commit/c812153f8d4f73b3f82cb19e3b98c84ca680eecb
https://github.com/openwrt/packages/commits/openwrt-21.02/net/curl/Makefile

<br/>

<br/>

```
curl -v https://openwrt.org
* Protocol "https" not supported or disabled in libcurl
curl: (1) Protocol "https" not supported or disabled in libcurl
```

#### Syslog:
```
root@OpenWrt:~# grep -i 'https-dns-proxy' /var/log/syslog
Fri Oct  7 12:37:22 2022 user.notice https-dns-proxy: Starting service ✓
Fri Oct  7 12:37:22 2022 daemon.info https-dns-proxy[4494]: [W] 1665139042.858021 main.c:236 HTTP/3 is not supported by current libcurl
Fri Oct  7 12:37:23 2022 daemon.info https-dns-proxy[4494]: [F] 1665139043.083623 https-dns-proxy-2021-11-22/src/https_client.c:317 08C8: CURLOPT_SSLVERSION error 4: Error
Fri Oct  7 12:37:28 2022 daemon.info https-dns-proxy[6972]: [W] 1665139048.086884 main.c:236 HTTP/3 is not supported by current libcurl
[...]

PR #19548, opened 2022-10-07
[21.02] fix broken libcurl by backporting curl: fix compilation with wolfSSL

ynezz:
Maintainer: @stangri 
Compile tested: x86_64
Run tested: N/A

Description:

Fixes: #19547


stangri:
@ynezz I've removed the wolfssl patch from master and 22.03, is this the 21.02-specific or should the patch be re-added there as well?

ynezz:
> Do you think the patch should be re-added to master and 22.03?
Its [fixed upstream](https://github.com/curl/curl/commit/6657602f504e49d27b5aa667a5ed04076ac2c4f6) since version 7.85.0 so it's not needed.

PR #19560, opened 2022-10-08
new package: cpusage

thuehn:
This PR adds the new package "cpusage" to OpenWrt packages. Cpusage is a small utility that prints cpu usage per second as output.

Signed-off-by: Thomas Huehn <thomas.huehn@hs-nordhausen.de>

Maintainer: Thomas Huehn <thomas.huehn@hs-nordhausen.de>
Compile tested: lastest trunk OpenWrt apu2, r20885 , APU & BananaPi R64
Run tested: latest trunk  OpenWrt apu2, r20885

Description:
Cpusage is a small utility that prints cpu usage per second as output.
It makes it handy to collect cpu usage while performing router/wifi experiments.


PolynomialDivision:
Please change the commit title to `cpusage: add new package`. You also commited some .svn, please remove that.

PR #19574, opened 2022-10-10
arp-scan: update to latest 1.9.8 

Gingernut1978:
Maintainer: Sergey Urushkin <urusha.v1.0@gmail.com>

Compile tested: mediatek:mt7622
Run tested: mediatek:mt7622

Signed-off-by: Andrew Sim <andrewsimz@gmail.com>

### DIFF
--- a/net/arp-scan/Makefile
+++ b/net/arp-scan/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=arp-scan
-PKG_VERSION:=1.9.7
+PKG_VERSION:=1.9.8
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/royhills/arp-scan/tar.gz/$(PKG_VERSION)?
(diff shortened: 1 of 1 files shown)


feckert:
Why do we not use a stable version? The latest is https://github.com/royhills/arp-scan/tree/1.9.8

Gingernut1978:
No objection

feckert:
Thank you for the update.
But if we take the stable version, then we should also leave the URL and not use git.

Gingernut1978:
> Thank you for the update. But if we take the stable version, then we should also leave the URL and not use git.

But I couldn't find a recent stable release, stable 1.9.8 release is from 2013. There is also a 1.9.9 stable but it's also from 2013. 

The link you posted was a tag afaik. 

I'm quite new to this.

feckert:
If you enter this command `wget https://codeload.github.com/royhills/arp-scan/tar.gz/1.9.8` then you are downloading the version to the directory with the file name `1.9.8`. And with command `sha256sum 1.9.8 ` you are calculating the hash of the downloaded tar.gz.

Now you only have to update the `PKG_VERSION:=1.9.7` to` PKG_VERSION:=1.9.8` and the `PKG_HASH:=` value from the sha256sum command.

Changing this values is enough to update arp-scan to an new version

Gingernut1978:
Understood, thanks.

Gingernut1978:
Done

PR #19578, opened 2022-10-10
python3: Upgrade setuptools and pip

QuintinHill:
Maintainer: @jefferyto 
Compile tested: not yet
Run tested: not yet (but upgrading setuptools and pip using pip worked fine)

Description:  dbus-fast requires the latest version of setuptools currently only a rather old version of setuptools is available in the opkg repository so bring this up to date.  Also bump pip version (on next line) to latest while changing this file.


BKPepe:
This is not how it works. setuptools and pip comes from Python tarball, which is downloaded. You can not upgrade it like it. Closing.

Issue #19005, opened 2022-07-21
curl 7.83.1 can not work with https


openips:
i build x64 with last master tree ,and curl 7.83.1 with commit b0f213a726a019fa864420fc9d50523ab8ec8901
then i find  curl 7.83.1 cannot work with https
log below 
```
curl 7.83.1 (x86_64-openwrt-linux-gnu) libcurl/7.83.1
Release-Date: 2022-05-11
Protocols: file ftp http mqtt
Features: alt-svc IPv6 Largefile
root@OpenBSD:~# curl https://123.down/123
curl: (1) Protocol "https" not supported or disabled in libcurl
```
check and fix  thks

openips:
@stangri

Livy-Inverse:
I am using snapshot r20146-bc43ad88ed (2022-07-19). This bug is present on my Newifi D2 (mipsel_24kc), and make my `git clone` command over https not work.

But on my MobiPromo CM520-79F (arm_cortex-a7_neon-vfpv4) using the same snapshot version, everything works well.

ksverdlov:
I have the same issue with xiaomi ax6s (mediatek/mt7622) and snapshot "openwrt-37e86780cc34-mediatek-mt7622-xiaomi_redmi-router-ax6s-squashfs-sysupgrade".

With OpenWrt 22.03.0-rc5 r20151-a0b7fef0ff I can not install curl:

```
root@OpenWrt:~# opkg install curl
Installing curl (7.83.1-2) to root...
Downloading https://downloads.openwrt.org/releases/22.03.0-rc5/packages/aarch64_cortex-a53/packages/curl_7.83.1-2_aarch64_cortex-a53.ipk
Collected errors:
 * pkg_hash_check_unresolved: cannot find dependency libwolfssl5.4.0.ee39414e for libcurl4
 * pkg_hash_fetch_best_installation_candidate: Packages for libcurl4 found, but incompatible with the architectures configured
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for curl:
 * 	libwolfssl5.4.0.ee39414e
 * opkg_install_cmd: Cannot install package curl.
```

openips:
use openssl lib can fix this issues.

paulhybryant:
can you elaborate a bit more on how to fix it (what you mean by "use openssl lib")?

awsong:
[...]

PR #19579, opened 2022-10-10
curl: error out if wolfSSL is not usable

ynezz:
Maintainer: @stangri 
Compile tested: x86/64 21.02
Run tested: None

Description:

When we explicitly declare, that we would like to have curl built with wolfSSL support using `--with-wolfssl` configure option, then we should make sure, that we either endup with curl having that support, or it shouldn't be available at all, otherwise we risk, that we end up with regressions like following:

```
  configure:25299: checking for wolfSSL_Init in -lwolfssl
  configure:25321: x86_64-openwrt-linux-musl-gcc -o conftest [snip]
  In file included from target-x86_64_musl/usr/include/wolfssl/wolfcrypt/dsa.h:33,
                   from target-x86_64_musl/usr/include/wolfssl/wolfcrypt/asn_public.h:35,
                  from target-x86_64_musl/usr/include/wolfssl/ssl.h:35,
                   from conftest.c:47:
  target-x86_64_musl/usr/include/wolfssl/wolfcrypt/integer.h:37:14: fatal error: wolfssl/wolfcrypt/sp_int.h: No such file or directory
       #include <wolfssl/wolfcrypt/sp_int.h>
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
  compilation terminated.
```

and in the end thus produce curl without https support:

```
 curl: (1) Protocol "https" not supported or disabled in libcurl
```

So fix it, by making the working wolfSSL mandatory and error out in configure step when that's not the case:

```
 checking for wolfSSL_Init in -lwolfssl... no
 configure: error: --with-wolfssl but wolfSSL was not found or doesn't work
```
References: #19005, #19547
Upstream-Status: Accepted [curl/curl/pull/9682]

### DIFF
--- a/net/curl/patches/300-curl-wolfssl.m4-error-out-if-wolfSSL-is-not-usable.patch
+++ b/net/curl/patches/300-curl-wolfssl.m4-error-out-if-wolfSSL-is-not-usable.patch
@@ -1,0 +1,57 @@
+From 5c91ff8853cbe08fb30462f99d3261383de9b98e Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Petr=20=C5=A0tetiar?= <ynezz@true.cz>
+Date: Mon, 10 Oct 2022 07:36:56 +0200
+Subject: [PATCH] curl-wolfssl.m4: error out if wolfSSL is not usable
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+When I explicitly declare, that I would like to have curl built with
+wolfSSL support using `--with-wolfssl` configure option, then I would
+expect, that either I endup with curl having that support, for example
+in form of https support or it wouldn't be available at all.
+
+Downstream projects like for example OpenWrt build curl wolfSSL variant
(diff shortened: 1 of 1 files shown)


BKPepe (approved):

stangri:
@ynezz thank you! Do I understand this correctly that since this was accepted upstream, this patch is only needed for the current release, so it can be removed for the next upstream curl release which will have this already?

ynezz:
> Do I understand this correctly that since this was accepted upstream, this patch is only needed for the current release, so it can be removed for the next upstream curl release which will have this already?

Yep, but we should backport this down to 21.02 in order to prevent future regression like #19005 or #19547

stangri:
> > Do I understand this correctly that since this was accepted upstream, this patch is only needed for the current release, so it can be removed for the next upstream curl release which will have this already?
> 
> Yep, but we should backport this down to 21.02 release in order to prevent future regression like #19005 or #19547

Thank you for your prompt reply. When you say down to 21.02 do you also mean backporting to 22.03 as well?

ynezz:
> When you say down to 21.02 do you also mean backporting to 22.03 as well?

Indeed, down it this context usually means master -> 22.03 -> 21.02.

PR #19586, opened 2022-10-11
nebula: implement netifd support

stangri:
Maintainer: me
Compile tested: x86_64, Sophos SG-135, OpenWrt 22.03.1
Run tested: x86_64, Sophos SG-135, OpenWrt 22.03.1, bring interface up, test connectivity between nodes

Description:
This commit contains the following:
* Update binary to version 1.6.1
* Update README URLs in the Makefile to link OpenWrt-specific info
* Separate the binary, the init script and netifd script into 3 packages: nebula, nebula-service and nebula-proto accordingly
* implement yml parser for init script to fetch variables from it
* add the netifd script for nebula protocol
* update test file to address all built packages

### DIFF
--- a/net/nebula/Makefile
+++ b/net/nebula/Makefile
@@ -4,12 +4,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nebula
-PKG_VERSION:=1.6.0
+PKG_VERSION:=1.6.1
 PKG_RELEASE:=1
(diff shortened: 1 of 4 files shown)


stangri:
If anyone can help me understand why the service test is failing, I'd be very grateful. I'm using the same method in other packages/init scripts and it's working there.

PR #19588, opened 2022-10-11
smartdns: update to 1.2022.38.1

pymumu:
Maintainer: me
Compile tested: TL-WR703N/22.03-SNAPSHOT r19160+542-7ea2f3d6e2


net/smartdns/Makefile
```diff
@@ -7,12 +7,12 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=smartdns
 PKG_VERSION:=1.2022.38
```
feckert, line 9: It is, after all, a source c code change?
Shouldn't we then update the PKG_VERSION instead?
PKG_RELEASE should be update if changes where made in the openwrt Makefile or in files by openwrt provided.

pymumu: change to PKG_VERSION: 1.2022.38.1, PKG_RELEASE:=1

feckert: Please update the PKG_MIRROR_HASH https://github.com/openwrt/packages/actions/runs/3234170838/jobs/5296948001#step:5:3433

pymumu: done.

PR #19589, opened 2022-10-11
Update xfrpc.init

vpei:
Both the ip and the domain names are supported.

Signed-off-by: web <42698055+vpei@users.noreply.github.com>

Description:
Both the ip and the domain names are supported.

### DIFF
--- a/net/xfrpc/files/xfrpc.init
+++ b/net/xfrpc/files/xfrpc.init
@@ -41,7 +41,7 @@ handle_xfrpc() {
 
 	if [ "$name" = "common" ]; then
 		uci_validate_section xfrpc xfrp "$name" \
-				'server_addr:ipaddr' \
+				'server_addr:string' \
 				'server_port:uinteger' \
 				'token:string:'	
 


feckert:
Please update also the [PKG_RELASE](https://github.com/openwrt/packages/blob/master/net/xfrpc/Makefile#L12)
Please also change the [commit](https://github.com/openwrt/packages/pull/19589/commits/a8bf18128443640039408fe5310290c9d8817e96) message to `xfrpc: change uci  server_addr validation`

feckert:
You don't have to close the pullrequest right away.
Just change the pullrequest locally and then do a git push --force to github to your branch.
The pullrquest gets updated right away.

net/xfrpc/files/xfrpc.init
```diff
@@ -41,7 +41,7 @@ handle_xfrpc() {
 
 	if [ "$name" = "common" ]; then
 		uci_validate_section xfrpc xfrp "$name" \
-				'server_addr:ipaddr' \
+				'server_addr:string' \
```
1715173329, line 44: > Both the ip and the domain names are supported.

Using `host` here might be better then.

PR #19597, opened 2022-10-12
ddns-scripts: add PROVIDES for old package names

paper42:
Maintainer: I can't find anyone in the Makefile and looking at history doesn't help
Run tested: -

Description:
* ddns-scripts-services: provide ddns-scripts_service
* ddns-scripts-cloudflare: provide ddns-scripts_digitalocean.com-v2
* ddns-scripts-freedns: provide ddns-scripts_freedns_42_pl
* ddns-scripts-godaddy: provide ddns-scripts_godaddy.com-v1
* ddns-scripts-noip: provide ddns-scripts_no-ip_com
* ddns-scripts-nsupdate: provide ddns-scripts_nsupdate
* ddns-scripts-route53: provide ddns-scripts_route53-v1
* ddns-scripts-cnkuai: provide ddns-scripts_cnkuai_cn

https://github.com/openwrt/packages/pull/13509 by @feckert renamed many ddns-scripts packages, but didn't include a PROVIDES for the old package names to make updates work well.

This is should be applied to all stable branches because the rename happened with 21.02.

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
-PKG_RELEASE:=27
+PKG_RELEASE:=28
 
 PKG_LICENSE:=GPL-2.0
 
@@ -50,6 +50,7 @@ endef
 define Package/ddns-scripts-services
   $(call Package/ddns-scripts/Default)
   TITLE:=Common ddns providers
+  PROVIDES:=ddns-scripts_service
 endef
 
 define Package/ddns-scripts-services/description
@@ -61,6 +62,7 @@ define Package/ddns-scripts-cloudflare
   $(call Package/ddns-scripts/Default)
   TITLE:=Extension for cloudflare.com API v4
   DEPENDS:=ddns-scripts +curl
+  PROVIDES:=ddns-scripts_digitalocean.com-v2
 endef
 
 define Package/ddns-scripts-cloudflare/description
@@ -72,6 +74,7 @@ define Package/ddns-scripts-freedns
   $(call Package/ddns-scripts/Default)
   TITLE:=Extension for freedns.42.pl
   DEPENDS:=ddns-scripts +curl
+  PROVIDES:=ddns-scripts_freedns_42_pl
 endef
 
 define Package/ddns-scripts-freedns/description
@@ -83,6 +86,7 @@ define Package/ddns-scripts-godaddy
   $(call Package/ddns-scripts/Default)
   TITLE:=Extension for godaddy.com API v1
   DEPENDS:=ddns-scripts +curl
+  PROVIDES:=ddns-scripts_godaddy.com-v1
 endef
 
 define Package/ddns-scripts-godaddy/description
@@ -123,6 +127,7 @@ define Package/ddns-scripts-noip
   $(call Package/ddns-scripts/Default)
   TITLE:=Extension for no-ip.com
   DEPENDS:=ddns-scripts
+  PROVIDES:=ddns-scripts_no-ip_com
 endef
 
 define Package/ddns-scripts-noip/description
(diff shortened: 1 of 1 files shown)


feckert:
I am not sure how the workflow should be.
Does it really make sense to always support all old package names?

paper42:
> I am not sure how the workflow should be. Does it really make sense to always support all old package names?

I agree that it might be a bit painful to keep providing old package names on every rename, but users won't be surprised when a package is seemingly randomly removed with a major update or an update doesn't run because the package no longer exists. This way we only add one line mentioning the old package name and users will just get the new package, nothing will break for them and they won't even notice the rename.

BKPepe:
Maybe @jow- can share his opinion on this one?

LorbusChris:
This definitely surprised me, thanks for filing the PR @paper42!

Is keeping the `PROVIDES` around really a maintenance burden?

FWIW in Fedora/RHEL it's standard practice to add those when an RPM package is renamed as to not unnecessarily break consumers of the package. That said I'm not familiar with how `opkg` resolves those names.

feckert:
@LorbusChris Sounds good then let's do it like this
@paper42 thanks merged!

BKPepe:
We should backport this to OpenWrt 22.03 and OpenWrt 21.02 releases as well.

net/ddns-scripts/Makefile
```diff
@@ -61,6 +62,7 @@ define Package/ddns-scripts-cloudflare
   $(call Package/ddns-scripts/Default)
   TITLE:=Extension for cloudflare.com API v4
   DEPENDS:=ddns-scripts +curl
+  PROVIDES:=ddns-scripts_digitalocean.com-v2
```
BKPepe, line 65: This should be: ``ddns-scripts_cloudflare.com-v4``

net/ddns-scripts/Makefile
```diff
@@ -50,6 +50,7 @@ endef
 define Package/ddns-scripts-services
   $(call Package/ddns-scripts/Default)
   TITLE:=Common ddns providers
+  PROVIDES:=ddns-scripts_service
```
BKPepe, line 53: No need. I could not find any package in OpenWrt 19.07 with this naming. The same applies for OpenWrt 21.02.

net/ddns-scripts/Makefile
```diff
@@ -180,6 +187,7 @@ define Package/ddns-scripts-cnkuai
   $(call Package/ddns-scripts/Default)
   TITLE:=CnKuai API
   DEPENDS:=ddns-scripts +curl +giflib-utils
+  PROVIDES:=ddns-scripts_cnkuai_cn
```
BKPepe, line 190: - This is not necessary. I could not find this package in OpenWrt 19.07. 
- Also... this provides should be for package, which is below, right?

net/ddns-scripts/Makefile
paper42: It seems like the package was added in 42696aa16234b466de54cde77fbef0220e5c3fff and renamed in a05030b9ca76c5c6bf73a7b8c720cb531fcf8853 which I guess happened only in master and never got to a release. I went through the renames in that PR and didn't check if it got to a release or not, I will check next time if you prefer to only handle renames in releases.

net/ddns-scripts/Makefile
paper42: similar to https://github.com/openwrt/packages/pull/19597#discussion_r998089071

feckert:
> We should backport this to OpenWrt 22.03 and OpenWrt 21.02 releases as well.

@BKPepe Sorry did not see your review.
I have already backported this to [openwrt-22.03](https://github.com/openwrt/packages/commit/b0e52085578de4b62e3060efbce2683819f86ccb) and [openwrt-21.02](https://github.com/openwrt/packages/commit/687e1373358c9fa8371bb3a43dfa6e3cab252daa).

net/ddns-scripts/Makefile
paper42: oops, I am not sure how that happened, fixed in #19628

PR #19600, opened 2022-10-13
btop: add new package

1715173329:
Maintainer: me
Compile tested: ramips/mt7621, rockchip/armv8, x86/64
Run tested: NanoPi R2S

Description:
Resource monitor that shows usage and stats for processor, memory, disks, network and processes.
https://github.com/aristocratos/btop

### DIFF
--- a/admin/btop/Makefile
+++ b/admin/btop/Makefile
@@ -1,0 +1,47 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=btop
+PKG_VERSION:=1.2.12
+PKG_RELEASE:=$(AUTORELEASE)
+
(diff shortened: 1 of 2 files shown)


admin/btop/Makefile
```diff
@@ -0,0 +1,51 @@
+# SPDX-License-Identifier: GPL-3.0-only
```
PolynomialDivision, line 1: Why GPL-3.0? It's your choice but whole openwrt is gpl-2.0-only? ^^

1715173329: It's a long story... I will remove this header later today ;)

PolynomialDivision: You can leave it GPL-3.0. I don't care. ;)
I was just interested.

PR #19603, opened 2022-10-14
net-snmp: init script improvments

feckert:
Maintainer: @stintel 
Compile tested: no only script changes
Run tested: x86_64, APU3, OpenWrt master, done

- net-snmp: use -c snmpd option for config read
- net-snmp: explicitly restart the service when reloading
- net-snmp: add pid file
- net-snmp: load pass from json

### DIFF
--- a/net/net-snmp/files/snmpd.init
+++ b/net/net-snmp/files/snmpd.init
@@ -360,6 +360,10 @@ service_stopped() {
 	procd_set_config_changed firewall
 }
 
+reload_service() {
+	restart
+}
+
 service_triggers(){
 	local script=$(readlink "$initscript")
 	local name=$(basename ${script:-$initscript})
--- a/net/net-snmp/files/snmpd.init
+++ b/net/net-snmp/files/snmpd.init
@@ -175,6 +175,20 @@ snmpd_pass_add() {
 	echo "$pass $priority $miboid $prog" >> $CONFIGFILE
 }
 
+snmpd_json_pass_add() {
+	. /usr/share/libubox/jshn.sh
+	[ -d /usr/share/snmp/pass/ ] || return
+	for file in /usr/share/snmp/pass/*.json; do
+		local pass='pass'
+		local miboid prog persist priority
+		json_load_file "${file}"
+		json_get_vars miboid prog persist priority
+		[ -z "$persist" ] || pass='pass_persist'
+		priority=${priority:+-p $priority}
+		echo "$pass $priority $miboid $prog" >> $CONFIGFILE
+	done
+}
+
 snmpd_exec_add() {
 	local cfg="$1"
 
@@ -305,6 +319,7 @@ start_service() {
 	config_foreach snmpd_access_HostName_add access_HostName
 	config_foreach snmpd_access_HostIP_add access_HostIP
 	config_foreach snmpd_pass_add pass
+	snmpd_json_pass_add
 	config_foreach snmpd_exec_add exec
 	config_foreach snmpd_extend_add extend
 	config_foreach snmpd_disk_add disk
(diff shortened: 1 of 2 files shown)


net/net-snmp/files/snmpd.init
```diff
@@ -336,7 +336,7 @@ start_service() {
 	append_parm trapsess trapsess trapsess
 
 	procd_set_param command $PROG -Lf /dev/null -f -r -p "$pid_file"
-	procd_set_param file $CONFIGFILE
```
stintel, line 321: Why is this removed? This changes reload behaviour, has nothing to do with the commit subject nor is it explained in the commit message.

feckert: I just saw that I also have to remove this [line](https://github.com/openwrt/packages/blob/master/net/net-snmp/Makefile#L270) from the Makefile install target, when I add the procd line `procd_append_param command -C -c $CONFIGFILE` to load the generated config file.

feckert:
@stintel I have now stumbled across this pullrequest of mine again. How about it, can we at least merge a few of my commits? Then my patchset would become smaller.

 Or please tell me what I should change. I have answered your question

net/net-snmp/files/snmpd.init
```diff
@@ -175,6 +176,20 @@ snmpd_pass_add() {
...
+	. /usr/share/libubox/jshn.sh
+	[ -d /usr/share/snmp/pass/ ] || return
+	for file in /usr/share/snmp/pass/*.json; do
+		local pass='pass'
+		local miboid prog persist priority
+		json_load_file "${file}"
+		json_get_vars miboid prog persist priority
+		[ -z "$persist" ] || pass='pass_persist'
+		priority=${priority:+-p $priority}
+		echo "$pass $priority $miboid $prog" >> $CONFIGFILE
```
openwrt-ai, line 189: Two issues in this loop compared with the UCI `snmpd_pass_add` above:

1. **`persist` false is treated as true.** With a JSON boolean (`"persist": true|false` per the commit message), `json_get_vars` yields `1`/`0`, so `[ -z "$persist" ] || pass='pass_persist'` sets `pass_persist` even when `persist` is `false` (`0` is non-empty). `snmpd_pass_add` uses `config_get_bool persist ...; [ $persist -ne 0 ]`, and e.g. `ppp.sh` similarly branches on the numeric value. Compare the value, not just presence.

2. **No `miboid`/`prog` guards.** `snmpd_pass_add` does `[ -n "$miboid" ] || return 0` / `[ -n "$prog" ] || return 0`. Here, an incomplete JSON file, or the unmatched-glob case (dir exists but no `*.json`, where busybox ash leaves `$file` as the literal pattern and `json_load_file` fails), writes a malformed `pass   ` line to `$CONFIGFILE`.

```suggestion
	for file in /usr/share/snmp/pass/*.json; do
		[ -e "$file" ] || continue
		local pass='pass'
		local miboid prog persist priority
		json_load_file "${file}"
		json_get_vars miboid prog persist priority
		[ -n "$miboid" ] || continue
		[ -n "$prog" ] || continue
		[ "${persist:-0}" = "1" ] && pass='pass_persist'
		priority=${priority:+-p $priority}
		echo "$pass $priority $miboid $prog" >> $CONFIGFILE
```

---

openwrt-ai (commented):
Reviewed 4 new commits; one inline comment on `snmpd_json_pass_add` (`persist` handling and missing `miboid`/`prog` guards). The `-C -c $CONFIGFILE` change, pid-file, and symlink removal look consistent with their commit messages.

---

feckert:
I'll close this, rebase everything, and then open a new PR.

PR #19605, opened 2022-10-14
zabbix: add sqlite3 support

feckert:
Maintainer: @champtar 
Compile tested: x86_64, APU3, OpenWrt master

**Description:**
This change makes it possible to build zabbix with sqlite3 support.
**By selcting "SQLite" as "Database Software" you are not able to build the zabbix-server and zabbix-server-frontend package anymore. This database is not supported by this services.**

This replaces the pullrequest #17878

@ttocsr If you want, you can take a look at it and see if it fits.

**Tests:**
```
root@st-dev-07 ~ # zabbix_proxy --version
zabbix_proxy (Zabbix) 6.2.3
Revision 98ee88fc19d 21 September 2022

Copyright (C) 2022 Zabbix SIA
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it according to
the license. There is NO WARRANTY, to the extent permitted by law.
```
```
root@st-dev-07 ~ # ldd /usr/sbin/zabbix_proxy
        /lib/ld-musl-x86_64.so.1 (0x7f91533da000)
        libsqlite3.so.0 => /usr/lib/libsqlite3.so.0 (0x7f915307c000)
        libz.so.1 => /usr/lib/libz.so.1 (0x7f9153066000)
        libevent-2.1.so.7 => /usr/lib/libevent-2.1.so.7 (0x7f9153031000)
        libpcre.so.1 => /usr/lib/libpcre.so.1 (0x7f9152fd5000)
        libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x7f9152fc1000)
        libc.so => /lib/ld-musl-x86_64.so.1 (0x7f91533da000)
```

If I select sqlite for zabbix_proxy, then I can no longer build zabbix-server and zabbix-server-frotend.


ttocsr:
@feckert do you plan on adding another variant PR to handle the DB choices?

feckert:
That would be the next step.
The only problem is that I then have a lot of variations.
In total, there are 9.

- nossl-sqlite (without zabbix-server)
- openssl-sqlite (without zabbix-server)
- gnussl-sqlite (without zabbix-server)
- nossl-postgres
- openssl-postrgres
- gnussl-postgres
- nossl-mysql
- openssl-mysql
- gnussl-mysql

I'm not quite sure yet how I can best depict this. Do you @champtar have any suggestions?

champtar:
Keep in mind not to overuse the buildbot for packages that are used by ~10 users
we could drop gnussl from the default built variant (if buildroot make it possible)
Also maybe mysql/postgresql already pull openssl as a dependency and we can drop nossl variant

That would give us:
- nossl-sqlite (agent/proxy)
- openssl-sqlite (agent/proxy)
- openssl-postgres (proxy/server)
- openssl-mysql (proxy/server)

To help decide build time and size would be good

feckert:
@champter Next I will to  the build variant.
I will drop the gnussl version as you suggested

@ttocsr it would be nice if you could test my next pullrequest

PR #19606, opened 2022-10-14
samba4: change perl dependency to fix menuconfig

Gingernut1978:
Adding perlbase-json-pp to samba4-libs dependencies was the wrong approach and caused samba packages not to be offered by menuconfig. AFAIK perlbase-json-pp is a perl helper to building samba4 and seems to be already included in perl/host so use that instead to fix the menuconfig issues.

Signed-off-by: Andrew Sim <andrewsimz@gmail.com>


feckert:
Thanks

PR #19617, opened 2022-10-17
treewide: fix procd service inactive

1715173329:
Maintainer: me
Compile tested: N/A
Run tested: NanoPi R2S

Description:
Exit directly will result procd service inactive and uci configuration changes are no longer monitored.

### DIFF
--- a/net/cloudflared/files/cloudflared.init
+++ b/net/cloudflared/files/cloudflared.init
@@ -18,7 +18,7 @@ start_service() {
 
 	local enabled
 	config_get_bool enabled "config" "enabled"
-	[ "$enabled" -eq "1" ] || exit 1
+	[ "$enabled" -eq "1" ] || return 1
 
 	procd_open_instance "$CONF"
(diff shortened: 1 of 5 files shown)


neheb:
That is…interesting. Given that AUTORELEASE doesn’t really work with treewide, I think it makes more sense to make multiple commits with each package.

wonder how many other packages are affected…

1715173329:
> That is…interesting. Given that AUTORELEASE doesn’t really work with treewide

https://github.com/openwrt/packages/actions/runs/3265617495/jobs/5368113799#step:8:54
emmm.. seems okay?

> I think it makes more sense to make multiple commits with each package.

Yes, that's what I actually did before.
But to me it's just too redundant to have multiple commits with the same description...

> wonder how many other packages are affected…

adguardhome and hs20 at least.

stangri:
> wonder how many other packages are affected…

Thanks to @1715173329 I've just fixed simple-adblock. PR pending.

PR #19642, opened 2022-10-20
frr: update to latest 8.3.1 branch

lucize:

Maintainer: me
Tested on x86-64


lucize:
Arc is failing from some time and upstream doesn't do testing on so many targets

PR #19647, opened 2022-10-20
uacme: update to 1.7.3

lucize:

Maintainer: me 
Tested x86-64


hgl:
@lucize would you consider making uacme adopt the newly acme architecture that is developed in acme-common and acme-acmesh, so acme.sh and uacme can reuse lots of common code? I designed the architecture with @tohojo. Would be happy to answer any questions you might have.

PR #19648, opened 2022-10-20
[21.02] Backport fixes for modem-manager and its dependencies

nemesisdesign:
Maintainer: @mips171 
Compile tested: ramips MediaTek MT7620A ver:2 eco:6, Zbtlink ZBT-WE1026-5G (16M), OpenWrt 21.02
Run tested: same as above

Description:

Modem-manager is unusable on OpenWrt 21.02 due to frequent disconnections which leave the device disconnected from the internet, hence I am backporting the fixes to 21.02 which will still be preferred by some people until all the packages impacted by the nftables change are sorted out.


nemesisdesign:
I am not sure what is wrong with the `mips_24kc` build?

nemesisdesign:
Ping @aleksander0m @mips171 @feckert @neheb

mips171:
LGTM

nemesisdesign:
@hauke WDYT?

feckert:
I can't decide that on my own.
I'm not so excited when we make so many changes in a stable release branch.
This usually creates regressions in down stream projects.
Since they work with a different code base. If so, only bug fixes should be included.

nemesisdesign:
@feckert for changes you mean bumps of minor versions or what exactly? I backported only fixes and version bumps which are required by the more up to date version of ModemManager. In my experience the ModemManager on OpenWrt 21.02 release is pretty unusable, it disconnects and never reconnects again on its own without a manual restart or a watchdog.

nemesisdesign:
@Linaro1985 all the commits are cherry-picked with `git cherry-pick -x` except for two commits to bump the version of modem manager and libqmi which I couldn't cherry-pick and I had to do manually due to the fact that master has diverged from the code on the 21.02 branch and hence it was unpractical to cherry-pick and fix a lot of conflicts, these two being [056ad9f](https://github.com/openwrt/packages/pull/19648/commits/056ad9f504262ae5c3443cbd25a23ea86d76c7ad) and [0cccbcb](https://github.com/openwrt/packages/pull/19648/commits/0cccbcb0c76f2d2fe74fc0acae02c69aee0d8c52).
What other commits do you see as being completely different?

Linaro1985:
@nemesisdesign for example, meson build support which I added sometime ago to master branch. For openwrt-21.02 meson.mk is [here](https://github.com/openwrt/packages/blob/openwrt-21.02/devel/meson/meson.mk).

nemesisdesign:
@Linaro1985 is the meson build support being backported to all the OpenWrt 21.02 packages, if yes, I can look into.
Before spending time in making that change though, I would like to know if there are other opinions against merging these fixes, because if so I wouldn't want to waste my time on it.

BKPepe:
Yep, meson works with OpenWrt 21.02, but you need to be careful about the path for meson.mk, but let's pull the trigger and let's see if it will be broken. This has been sitting here for some time.

nemesisdesign:
Thanks @BKPepe, how can I see if it break anything?

feckert:
@BKPepe I am not entirely happy with the way the situation went here. I have some problems with my builds now. Because the checksums of the package have changed. For whatever reason!
After this [fix
](https://github.com/openwrt/packages/commit/2e35f82f117d89fd3925e0a26aa4e47515b942e7) I now have also problems with the libqmi.

```
SHELL= flock /home/feckert/workspace/openwrt/LDM-owrt-21.02/build/openwrt/tmp/.libqmi-1.30.8.tar.xz.flock -c '          /home/feckert/workspace/openwrt/LDM-owrt-21.02/build/openwrt/scripts/download.pl "/home/feckert/dl" "libqmi-1.30.8.tar.xz" "862482ce9e3ad0bd65d264334ee311cdb94b9df2863b5b7136309b41b8ac1990" "" "https://www.freedesktop.org/software/libqmi"    '
Hash of the local file libqmi-1.30.8.tar.xz does not match (file: a0fa33a89011bdb593f66fd0b674f2a7c31f87e43ffd7f3e9a515b00864c4a91, requested: 862482ce9e3ad0bd65d264334ee311cdb94b9df2863b5b7136309b41b8ac1990) - deleting download.
```

I think this is because of switching to meson.
Additional  the situation is that the owrt-22.03 is now behind the owrt-21.02 of the package versions!

BKPepe:
Yes, I do have the same issue with libqmi, though.  :-(

> Because I am using a download directory for all of my package of all build branches

Yep... We are doing the same for Turris OS.

BKPepe:
> Additional the situation is that the owrt-22.03 is now behind the owrt-21.02 of the package versions!

I thought that someone would look into it. 🤦 
It seems that one user of Turris router (https://gitlab.nic.cz/turris/os/build/-/issues/439), noticed that as well.
Fixed by https://github.com/openwrt/packages/commit/3c8a12e897d09e551ae87c9509f5e63ba7743780. Sorry!

PR #19657, opened 2022-10-21
net-snmp: change persistent directory and add support v3 in compile

liaoqingxin:

Description:
net/net-snmp/Makefile: change persistent directory and add v3 compile 
net/net-snmp/files/snmpd.conf: add v3 user config
net/net-snmp/files/snmpd.init: add v3 user init

to fix bug:  
    1. Using net-snmp v3, the EngineBoots no increase auto in persistent file when restart snmpd server
    2. After changed V3 user config, it no changed in persistent file when restart snmpd server

### DIFF
--- a/net/net-snmp/Makefile
+++ b/net/net-snmp/Makefile
@@ -41,7 +41,7 @@ define Package/libnetsnmp
 $(call Package/net-snmp/Default)
   SECTION:=libs
   CATEGORY:=Libraries
-  DEPENDS:=+libnl-tiny +libpci +libpcre
+  DEPENDS:=+libnl-tiny +libpci +libpcre +libopenssl
   TITLE:=Open source SNMP implementation (libraries)
 endef
 
@@ -203,7 +203,7 @@ CONFIGURE_ARGS += \
 	--with-mib-modules="$(SNMP_MIB_MODULES_INCLUDED)" \
 	--with-out-transports="$(SNMP_TRANSPORTS_EXCLUDED)" \
 	--with-transports="$(SNMP_TRANSPORTS_INCLUDED)" \
-	--without-openssl \
+	--with-openssl=yes \
 	--without-libwrap \
 	--without-mysql \
 	--without-rpm \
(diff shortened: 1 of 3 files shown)


karlp:
How does changing the directory fix this?

karlp:
while snmpv3 is nice, pulling openssl into the default build is probably not going to work for people...

liaoqingxin:
> Maintainer: me / @<github-user> (find it by checking history of the package Makefile) Compile tested: (put here arch, model, OpenWrt version) Run tested: (put here arch, model, OpenWrt version, tests done)
> 
> Description: net/net-snmp/Makefile: change persistent directory and add v3 compile net/net-snmp/files/snmpd.conf: add v3 user config net/net-snmp/files/snmpd.init: add v3 user init
> 
> to fix bug: 1. Using net-snmp v3, the EngineBoots no increase auto in persistent file when restart snmpd server 2. After changed V3 user config, it no changed in persistent file when restart snmpd server



> How does changing the directory fix this?

When "NETSNMP_PERSISTENT_DIRECTORY = /usr/lib/snmp/", it's the same as SNMPLIBPATH, then "/usr/lib/snmp/snmpd.conf" will be reload twice when restart snmpd server. And then, every persistent parameter will called twice, "oldEngineID" is the same. In net-snmp source code, when oldengineID_conf function called twice, oldEngineIDLength will always be 0, because read_config_read_octet_string_const function has a hidden bug. In function read_config_read_octet_string_const , it alloc iLen+1 byte memory but only return iLen when first call. When a parameter call it next time, it always be failed becuase iLen >= len.
Function  init_snmpv3_post_config will reset engineBoots to 1 if oldEngineID is NULL.

liaoqingxin:
> while snmpv3 is nice, pulling openssl into the default build is probably not going to work for people...

OK, V3 probably shouldn't be added to the build by default

GeorgeSapkin:
No updates in over three years and PR has merge conflicts. Closing as stale. Feel free to reopen if anything changes.

Issue #19641, opened 2022-10-20
MariaDB: Using mysql.sock causes permission denied error (with a fix)


brackenhill-mob:
Maintainer: @miska 
Environment: x86 running on Intel i3 with OpenWrt v22.03.01

Description:

Using any PHP connect call (ie WordPress or a simple test script), connections to the database fail with mysqli or pdo with the message "Permission denied"

Solution is to chmod 0777 ```/var/run/mysql``` (ie the folder holding the socket file)

I'm using nginx with uci disabled (I can't think this would make a difference but...). I've had this issue under both the release and nightly versions of OpenWrt and with different x86 hardware. My "fix" solves it every time.

Also just noticed that /init.d/mysqld sets ```rundir=/var/run/mysqld``` but the actual socket ends up in ```/var/run/mysql``` making ```/var/run/mysqld``` redundant.

miska:
Other possible solution is to make a user you are using to access the MariaDB database part of `mariadb` group. Thank you for your report, will look into that and the directories.

miska:
Hmmm, seems like run dir is never used and comment suggests that is was about socketdir, so yes, you are right, not 777, but 755, but still. Will fix that.

brackenhill-mob:
Changed my system to 755 and it works. Thanks for the quick fix in #19666

PR #19666, opened 2022-10-21
mariadb: Drop unused rundir fix socket dir perms

miska:
Maintainer: me
Compile tested: mvebu, Turris Omnia, OpenWrt 21.02
Run tested: mvebu, Turris Omnia, OpenWrt 21.02, runs, client is able to connect

Description:
We had been creating "rundir" but it was never used, probably leftover from some removed function. At the same time, we were setting quite strict rights to the socket directory (while comments sugested otherwise).

Fix for #19641

### DIFF
--- a/utils/mariadb/files/mysqld.init
+++ b/utils/mariadb/files/mysqld.init
@@ -67,7 +67,6 @@ mysqld_status() {
 start_service() {
 	conf=/etc/mysql/my.cnf
 	logdir=/var/log/mysql
-	rundir=/var/run/mysqld
 	version="$(mysqld --version | sed -n 's|.*Ver[[:blank:]]*\([0-9.]*\)-.*|\1|p')"
 
 	# Few basic checks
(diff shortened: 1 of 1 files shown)


utils/mariadb/files/mysqld.init
```diff
@@ -134,13 +133,13 @@ start_service() {
...
 	# Make sure all required directories exists and have correct rights
-	for i in "$logdir" "$rundir" "$sockdir"; do
+	for i in "$logdir" "$sockdir"; do
 		opts="-m 0750"
 		if ! [ -e "$i" ]; then
 			# $rundir needs to be accessible for
 			# clients
-			if [ "$i" = "$rundir" ]; then
-				opts=
+			if [ "$i" = "$sockdir" ]; then
```
neheb, line 141: maybe it'd be better to make it a one liner"

```
[ "$i" = "$sockdir" ] && opts="-m 0755"
```

neheb: The comment probably needs adjusting too.

miska: Done

PR #19669, opened 2022-10-22
acme: refactor

hgl:
- Move notifying logic out of acme hooks
- Prevent concurrent running of `acme get`
- Use procd instead of hotplug to reload services (leaving hotplug for user scripts only)
- Fix a bug where switching from staging to production didn't trigger the renew event

Maintainer: @tohojo

Description:

Using hotplug to reload a service can inadvertently restarting it if it was stopped, as pointed out by @ptpt52.

This PR make services use procd instead, which is how it should be properly done IMO. It also moves the actual notifying logic out of hooks, since they really shouldn't be concerned about how to actually do that.

@tohojo, if you think the design is good I will change other related services too.

### DIFF
--- a/net/acme-common/files/acme.sh
+++ b/net/acme-common/files/acme.sh
@@ -131,7 +131,6 @@ get)
 	config_load acme
 	config_foreach load_globals acme
 
-	mkdir -p /etc/ssl/acme
 	trap cleanup EXIT
 	config_foreach get_cert cert
 	;;
--- a/net/acme-common/files/acme.sh
+++ b/net/acme-common/files/acme.sh
@@ -59,6 +59,8 @@ load_options() {
 	export days
 	config_get standalone "$section" standalone 0
 	export standalone
+	config_get dns_wait "$section" dns_wait
+	export dns_wait
 
 	config_get webroot "$section" webroot
 	export webroot
--- a/net/acme-common/files/acme.sh
+++ b/net/acme-common/files/acme.sh
@@ -36,7 +36,7 @@ load_options() {
 	section=$1
 
 	# compatibility for old option name
-	config_get_bool use_staging "$section" staging
+	config_get_bool staging "$section" use_staging
 	if [ -z "$staging" ]; then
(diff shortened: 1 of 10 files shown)


net/acme-common/files/acme.sh
```diff
@@ -128,11 +134,16 @@ fi
 
 case $1 in
 get)
+	if [ -e "$run_dir"/lock ]; then
+		log err "Another ACME instance is already running."
+		exit 1
+	fi
+	touch "$run_dir/lock"
+	trap cleanup EXIT
```
tohojo, line 142: This is racy; two processes could run the check at the same time and both end up continuing. You'll have to use an atomic operation to actually create the lock; for shell scripts that's usually mkdir. See this example from sqm-scripts for how to do it right:

https://github.com/tohojo/sqm-scripts/blob/master/src/run-openwrt.sh#L80-L103

tohojo (changes requested):
General refactoring is OK by me; however, please split each change out into a separate commit, with a commit message explaining the problem it's fixing, and how it's being fixed.

Also, the locking is racy, see below...

net/acme-common/files/acme.sh
hgl: Fixed. I decided to use flock. Seems it's designed specifically for this use case.

hgl:
Updated as requested.

net/acme-common/files/acme.sh
tohojo: I'm pretty sure this doesn't work the way you think it does; how did you test it?

hgl: Sorry, I tested it in a small test script, but when tried to expand it in the real script, I got it wrong. It should be fixed.

I also fixed one more bug during testing, and took the liberty to add support for a new option. Let me know what you think.

tohojo: Yeah, that's better!

tohojo (approved):

Issue #19655, opened 2022-10-21
softflowd: do_page_fault(): sending SIGSEGV to softflowd for invalid read access from .....


HolgerHees:
Maintainer: @jesferman , @rvandegrift
Environment: OpenWrt 22.03.0, TP-Link Archer C7 v2, Qualcomm Atheros QCA9558 ver 1 rev 0

@rvandegrift this is not related to my ticket from yesterday. It is a different device running at a friend's house.

Description:

I'm running softflowd on a router where I get no flows when I call "softflowctl dump-flows"

When I call "softflowctl statistics" I see

```
root@OpenWrt:~# softflowctl statistics
softflowd[6688]: Accumulated statistics since 2022-10-21T07:12:06 UTC:
Number of active flows: 112
Packets processed: 20685
Fragments: 42
Ignored packets: 2 (2 non-IP, 0 too short)
Flows expired: 46 (0 forced)
Flows exported: 46 (79 records) in 4 packets (0 failures)
Packets received by libpcap: 21226
Packets dropped by libpcap: 0
Packets dropped by interface: 0

Expired flow statistics:  minimum       average       maximum
  Flow bytes:                  60        120147       1738147
  Flow packets:                 1           344          6350
  Duration:                  0.00s        47.31s        63.06s

Expired flow reasons:
       tcp =        14   tcp.rst =         1   tcp.fin =         0
       udp =         6      icmp =         0   general =         0
   maxlife =        25
over 2 GiB =         0
  maxflows =         0
   flushed =         0

Per-protocol statistics:     Octets      Packets   Avg Life    Max Life
           icmp (1):          10920          130      60.01s      60.02s
            tcp (6):         715552         2075      43.78s      63.06s
           udp (17):        4800272        13600      43.77s      63.01s
```

but when I call "softflowctl dump-flows" I get

```
root@OpenWrt:~# softflowctl dump-flows
softflowd[6688]: Dumping flow data:
```

an a kernel log message like 

```
Fri Oct 21 07:13:47 2022 kern.info kernel: [745743.417947] do_page_fault(): sending SIGSEGV to softflowd for invalid read access from 00000326
Fri Oct 21 07:13:47 2022 kern.info kernel: [745743.427175] epc = 77de0dc0 in libc.so[77d62000+a9000]
Fri Oct 21 07:13:47 2022 kern.info kernel: [745743.432682] ra  = 77de2b04 in libc.so[77d62000+a9000]
[...]

PR #19674, opened 2022-10-22
[22.03] softflowd: use %lld for time_t always

micmac1:
Add upstreamed patch to fix segmentation fault reported in [1].

[1] https://github.com/openwrt/packages/issues/19655

(cherry picked from commit 7d019635b1490248e0efc487b3723a4cd07b39d7)

Maintainer: @rvandegrift
Compile tested: 22.03 ath79 SDK
Run tested: ath79 22.03, also tested by @HolgerHees (see issue #19655)

Description:
Fix time64 issue resulting in segfault. Patch applied upstream.

### DIFF
--- a/net/softflowd/Makefile
+++ b/net/softflowd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=softflowd
 PKG_VERSION:=1.0.0
-PKG_RELEASE:=2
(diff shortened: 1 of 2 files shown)


rvandegrift:
This looks good - thanks for preparing the fix!

micmac1:
Updated "cherry picked from" ID to the one that is in master.

Issue #19665, opened 2022-10-21
privoxy: incorrectly formatted entry in /etc/services


powersj:
Maintainer: none listed :(
Environment: master

Description:

In the [privoxy makefile](https://github.com/openwrt/packages/blob/f4826a0f7461b26bf6ca090dad3d7a78af6b163b/net/privoxy/Makefile#L161), it will write an incorrectly formatted entry into `/etc/services`: `privoxy    8188`. This omits the protocol. The `/etc/services` manpage does not show the protocol as optional, like aliases are. If both TCP and UDP are supported, then there should be two lines one for each protocol.

This was discovered as a part of #19663 and https://github.com/influxdata/telegraf/issues/12084. I have filed an [upstream issue](https://gitlab.com/cznic/libc/-/issues/25) with the Go library that had a panic due to this. Looking at the [git blame](https://github.com/openwrt/packages/blame/f4826a0f7461b26bf6ca090dad3d7a78af6b163b/net/privoxy/Makefile#L161), it has been like this for 7 years, but I believe the right thing to do is fix this here as well.

PR #19694, opened 2022-10-24
privoxy: correctly format /etc/services additions

powersj:
Maintainer:

Description:

This adds the missing protocol (e.g. /tcp and /udp) to the entry in /etc/services. If the entry already exists, it will add the /tcp to it. Otherwise, it will look and add the tcp and udp entries if either is missing.

fixes: #19665

Signed-off-by: Josh Powers <powersj@fastmail.com>


neheb:
that's a useless cat call.

PR #19702, opened 2022-10-25
lighttpd: backport patch to fix dummy Sec-WebSocket-Key

BKPepe:
Maintainer: @gstrauss 
Compile tested: all Turris routers
Run tested: N/A

Description:
It was requested in https://gitlab.nic.cz/turris/os/packages/-/issues/873
  - Backports: https://github.com/lighttpd/lighttpd1.4/commit/cda9b71653bb6a633957f653fa08e819b32e601e


gstrauss:
LGTM.

Sorry about the bug.  The bug was introduced in lighttpd 1.4.65 (released Jun 2022) but only recently did someone report the bug due to a strict backend framework rejecting the invalid Sec-WebSocket-Key.

gstrauss:
(FYI: I missed this yesterday: you probably should have bumped the package version number from -3 to -4)

BKPepe:
Oh! I should. I am used to that many packages are using variable ``$(AUTORELEASE)``, because these things were happening so often and this variable takes care of it.

gstrauss:
I was not aware of `$(AUTORELEASE)` and will look to use it in the future.

PR #19704, opened 2022-10-25
atlas-probe: fix SIGSEGV error on 32bit system

Ansuel:
Changes to time_t cause SIGSEGV error on 32bit system and cause ripe atlas malfunction. (registration successful but no traffic) Also introduce minor patch to fix some compilation warning.

While at it move PKG_RELEASE to AUTORELEASE macro.

---

These patch are pushed upstream and under review (https://github.com/RIPE-NCC/ripe-atlas-software-probe/issues/74 and .https://github.com/RIPE-NCC/ripe-atlas-probe-measurements/pull/11) These were tested on a mt7621 WAX202

### DIFF
--- a/net/atlas-probe/Makefile
+++ b/net/atlas-probe/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=atlas-probe
 PKG_VERSION:=2.4.1
-PKG_RELEASE:=1
+PKG_RELEASE:=$(AUTORELEASE)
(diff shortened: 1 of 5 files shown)


neheb:
*WAX202.

I tried using -Werror=format=2 to find these bad format strings, but it adds more warnings than I'd like.

Ansuel:
@BKPepe there was another fix to add for this that i forgot but i will make another pr. sorry my fault for following too much project at once.

PR #19714, opened 2022-10-26
libwebsockets: fix recursive dependency

BKPepe:
Maintainer: @karlp 
Compile tested: no, only verified by ``make menuconfig``
Run tested: no, only verified by ``make menuconfig``

While running `make menuconfig`, it was discovered then there is a recursive dependency like this:
```
tmp/.config-package.in:59138:error: recursive dependency detected! tmp/.config-package.in:59138:	symbol PACKAGE_libwebsockets-openssl is selected by PACKAGE_libwebsockets-mbedtls tmp/.config-package.in:59122:	symbol PACKAGE_libwebsockets-mbedtls depends on PACKAGE_libwebsockets-openssl
```

It is not possible with the recently added conflicts that two packages (OpenSSL and full variant, which uses OpenSSL as well), which are almost the same provides the same named package libwebsockets as their conflict - Mbed TLS.

Fixes: 676c5c72b5eeb583da2603e399fac085fa442c59 ("libwebsockets: OpenSSL and mbedTLS variants should conflict")

Reported by: @hnyman and @zxlhhyccc 
Thanks!

___

It was almost the last minute change, mea culpa on this one guys!

### DIFF
--- a/libs/libwebsockets/Makefile
+++ b/libs/libwebsockets/Makefile
@@ -48,7 +48,6 @@ define Package/libwebsockets/Default
 	DEPENDS:=+zlib +libcap
 	URL:=https://libwebsockets.org
 	MAINTAINER:=Karl Palsson <karlp@etactica.com>
-	PROVIDES:= libwebsockets
 endef
 
 define Package/libwebsockets-openssl
@@ -64,6 +63,7 @@ define Package/libwebsockets-mbedtls
 	TITLE += (mbedTLS)
 	DEPENDS += +libmbedtls
 	VARIANT:=mbedtls
+	PROVIDES:=libwebsockets
 	CONFLICTS:=libwebsockets-openssl
 endef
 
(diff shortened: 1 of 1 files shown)


karlp:
You're plan is then to revert the original changes in 1806 and 1907 as they're EOL then, right? and apply this just to the supported 21.02, 22.03 and master that are broken?

hnyman:
This seems to fix the recursive error in master, so ACK for that ;-)

BKPepe:
Well, it needs to be said that it is not exactly broken. Packages can be compiled. They can be installed, but in my eyes, it just throws a harmless warning while using ``make menuconfig``.  Thus why I don't have any plans to revert it in the EoL branches because some users are using it or some devices are only supported there. It is quite nice to have it fixed, even for old versions. 

@hnyman Thanks for confirming and once again, sorry for causing it!

karlp:
Is there something you're waiting on here? this is still your mistake to clean up.  You were in a super hurry to commit the last changes, why has this fix been dangling now?

BKPepe:
Yep, I was waiting for your ACK/approval. Thus merging, though.

Issue #19510, opened 2022-10-02
bandwidthd: parially broken in 22.03 and master


padre-lacroix:
Maintainer: @padre-lacroix 
Environment: 
arch MediaTek MT7621 ver:1 eco:3
model Ubiquiti EdgeRouter X
OpenWrt version OpenWrt 22.03.0 r19685-512e76967f / LuCI openwrt-22.03 branch git-22.245.77528-487e58a

Also on an Archer C7 v2, same OpenWrt version (OpenWrt 22.03.0 r19685-512e76967f / LuCI openwrt-22.03 branch git-22.245.77528-487e58a.

Description:
from issue #19352 (opened by https://github.com/frankgrant) that is now closed:
Getting kernel error messages:
do_page_fault(): sending SIGSEGV to bandwidthd for invalid read access from 00000018
epc = 77e76534 in libc.so[77dea000+a9000]
ra = 77e76410 in libc.so[77dea000+a9000]

No graphs are being produced.
It worked fine with OpenWrt 22.02

I (@padre-lacroix ) have the same error on a Archer C7 v2.

A pull request has been made by @neheb (https://github.com/openwrt/packages/pull/19392) but did not fix the issue!

https://github.com/frankgrant reported on 2022-09-29 that the problem still exists and this was confirmed by @padre-lacroix on the Archer C7 v2.

I looked that the changes between 21.02 and 22.03, and maybe musl 1.2x with the time_t type from 32 to 64 bit is in cause, but I am really not sure: I need help here!  The error happens whether bandwidthd or bandwidthd-sqlite is used. bandwidthd-pgsql has not been tested, but it should behave the same way as bandwidthd-sqlite.

It happens when trying to make the graph at 192.168.1.1/bandwidthd: every time you ask the web page to make the graph the error occurs.  Maybe related to graph.c (I have not thoroughly checked).

Investigate if using bandwidthd-sqlite would make graph from the data in sqlite and PHP and it works fine.  But, installing bandwidthd-sqlite was a little bit problematic has a conflict between libgd and libgd-ful prevented its installation.  Has to install libgd-full first and then was able to install bandwidthd-sqlite and thsi is working properly (with the exception of when asking 192.168.1./bandwidthd to make the graph: the error occurs.  But the PHP graph at 192.168.1.1/phphtdocs are working properly.

[...]

PR #19721, opened 2022-10-26
[22.03] bandwidthd: fix time_t problem

padre-lacroix:
Maintainer: me
Compile tested: mips_24kc on 22.03
Run tested: Qualcomm Atheros QCA9558 ver 1 rev 0, TP-Link Archer C7 v2, OpenWRT 22.03.0 r19685-512e76967f

Description:
Fixes issue #19510 with the change in time_t with musl 1.2

Signed-off-by: Jean-Michel Lacroix <lacroix@lepine-lacroix.info>
(cherry picked from commit 61d1a86)

### DIFF
--- a/utils/bandwidthd/Makefile
+++ b/utils/bandwidthd/Makefile
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2006-2017 OpenWrt.org
+# Copyright (C) 2006-2020 OpenWrt.org
 #
 # This is free software, licensed under the GNU General Public License v2.
 # See /LICENSE for more information.
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bandwidthd
 PKG_VERSION:=2.0.1-35
-PKG_RELEASE:=6
+PKG_RELEASE:=7
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/NethServer/bandwidthd/tar.gz/$(PKG_VERSION)?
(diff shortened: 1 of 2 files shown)


BKPepe:
Was this submitted to upstream?

padre-lacroix:
I tries to follow the process that was given to me last week when I made a pull request directly in 22.03 (this was put on pause and I started again).
This was merged in Master by @neheb and I cherry-picked it in 22.03.

neheb:
Upstream is basically dead.

BKPepe:
Still it should be sent to upstream, we can not carry patches forever. Someone can use those changes, when forking repositories, etc. The patch itself is missing any description, git header.

padre-lacroix:
Ok,
I finally understand by sending it upstream.  Sorry it took me so long!
The source code of bandwidthd is at https://github.com/NethServer/bandwidthd.  This repository has been archived by the owner and is read-only.  I cannot therefore send anything upstream.
I understand that this change should be sent upstream to NethServer for them to update their code, and this makes perfect sense!
I am not involved in the development of bandwidthd: I make sure that it is working in openWRT by patching the source code when required.  Should a new version is released upstream, I update also in openWRT and patch it if required.
Bandwidthd is an old software and the original one in source forge (https://bandwidthd.sourceforge.net/) has not been updated since January 2005.  In 2016, I found a new source that was also working with mysql and I updated bandwidthd in openWRT to use that one instead, as this was maintained code.  But, this code is no longer maintained!
I may have to either find a new source that is maintained for bandwidthd (there are a few) or take what is in NethServer and make a copy in my own repository and maintain it myself.  If I would do the later, I would need other people to participate as it could be useful for bandwidthd to:
- work with IpV6 (there are version that do that may be useful)
- word also with MarianaDB in addition to postgresql

PR #19723, opened 2022-10-28
attendedsysupgrade-common: update ASU CA pubkey

aparcar:
The old key was overwritten during a machine migration and the current worker key is no longer valid. To create new valid worker certificates this commit adds a new CA pubkey.

To update running OpenWrt devices via a client one needs to force install it or upgrade to the latest version of
`attendedsysupgrade-common` - I'm sorry for the inconvenience.


dangowrt:
Maybe mention that this is about expired **ucert** and has nothing to do with TLS what-so-ever.

PR #19725, opened 2022-10-28
oui: add new packages

zhaojh329:
Signed-off-by: Jianhui Zhao <zhaojh329@gmail.com>

Maintainer: me
Compile tested: (x86, x86, master)
Run tested: (x86, x86, master, tests done)

Description:
This is a `framework` used to develop Web interface for OpenWrt. Use Nginx + Vue3 + Lua.
https://github.com/zhaojh329/oui
https://zhaojh329.github.io/oui/


BKPepe:
I don't think that we should add ~150 files to this repository to develop a new web interface for OpenWrt. Overall, it will be confusing for other packages, which we have here. I think this should stay in [your own repository](https://github.com/zhaojh329/oui), which you already have. I don't see reasons why we should have it here.

@diizzyy Your opinion on this one?

zhaojh329:
@BKPepe 
May I add it to feeds.conf.default?

BKPepe:
I doubt that it is going to work.

diizzyy:
There are very few occations where we include complete source code for externally projects in the tree and this doesn't fall under that category. I share BKPepe's opinion that this should reside in its own repo and unless it's close to LuCIs feature set and people want to hack on it I doubt you'll get it approved as a separate feed. Please don't take this is as discouragement, it would certainly help if you could document more why there's a need for oui, what it fixes and that further goals are.

PR #19729, opened 2022-10-29
btop: use utf-force by default

msylgj:

Maintainer: Tianling Shen <cnsztl@immortalwrt.org> @1715173329 
Compile tested: rockchip r4s/r2s OpenWrt master SNAPSHOT
Run tested: rockchip r4s/r2s OpenWrt master SNAPSHOT

Description:
Because of OpenWrt haven't LANG environment variable, use `btop` will return `No UTF-8 locale detected!`. so I add a patch that set `utf_force` to true by default.

### DIFF
--- a/admin/btop/patches/001-use-utf-force-default.patch
+++ b/admin/btop/patches/001-use-utf-force-default.patch
@@ -1,0 +1,25 @@
+From be911ed522726d54fb7bdcfc76d9473ccc09dee8 Mon Sep 17 00:00:00 2001
+From: ZiMing Mo <msylgj@immortalwrt.org>
+Date: Sat, 29 Oct 2022 18:55:14 +0800
+Subject: [PATCH] btop: use utf-force by default
+
+Because of OpenWrt haven't LANG environment variable, use btop will
+return No UTF-8 locale detected!. This patch set utf_force to true by
+default.
+
+Signed-off-by: ZiMing Mo <msylgj@immortalwrt.org>
+---
+ src/btop.cpp | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+--- a/src/btop.cpp
++++ b/src/btop.cpp
+@@ -93,7 +93,7 @@ namespace Global {
+ 
+     bool debuginit{};   // defaults to false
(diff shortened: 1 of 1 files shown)


1715173329 (approved):
LGTM, thanks!

admin/btop/patches/001-use-utf-force-default.patch
```diff
@@ -0,0 +1,11 @@
+--- a/src/btop.cpp
++++ b/src/btop.cpp
```
BKPepe, line 2: This patch should be upstreamed, or at least there should be a pull request to upstream, where they reject it. I don't prefer to have patches in this repository, which we are going to carry forever. Also, when the patch is sent upstream, it will at least have the basic formal things like who is the author of the change, the commit subject, the commit description, etc.

BKPepe (changes requested):

admin/btop/patches/001-use-utf-force-default.patch
1715173329: > This patch should be upstreamed, or at least there should be a pull request to upstream, where they reject it. I don't prefer to have patches in this repository, which we are going to carry forever.

This is a openwrt related issue, there are also lots of openwrt-specific hacks in tree, so do you mean we should send all of them to upstream and wait for the rejection or complain?

> Also, when the patch is sent upstream, it will at least have the basic formal things like who is the author of the change, the commit subject, the commit description, etc.

`git format-patch` is enough then.

BKPepe: There are many hacks in this repository, that's true, but why should we have more? This package does not have any patches. Why should we change it?

1715173329: So you guess why we added this patch

1715173329:
Another idea, we can create an alias in profile.d, so this however avoids patches.
```
$ /etc/profile.d/btop.sh
alias btop="btop --force-utf8"
```

@BKPepe

1715173329:
I will replace this hack with the profile alias recently. Thank you again.

PR #19749, opened 2022-10-31
apcupsd: remove header file patch

neheb:
Seems to be from a time when pkgconfig returned the wrong path.

Maintainer: @tru7 
Compile tested: mt7621


tru7:
Fine with me. Thank you

PR #19756, opened 2022-10-31
openvpn: add possibility to set param "compress" without algorithm

sch-m:
Maintainer: @AuthorReflex  / @mkrkn
Compile tested: lantiq/xrx200, x86/64
Run tested: lantiq/xrx200, x86/64

Description:
In some situations you need to set the compress param without an algorithm. Compression will be turned off, but the packet framing for compression will still be enabled, allowing a different setting to be pushed later.

As it is not possible to have options with optional values at the moment, I've introduced a pseudo value "frames_only" which will be removed in the init script.

I'm re-sending this PR after openvpn moved from openwrt core to packages feed: https://github.com/openwrt/openwrt/pull/2694

After this got merged, I'll create a PR for luci as well.

### DIFF
--- a/net/openvpn/Makefile
+++ b/net/openvpn/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=openvpn
 
 PKG_VERSION:=2.5.7
-PKG_RELEASE:=3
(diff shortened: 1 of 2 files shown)


mkrkn:
Ack from me.

PR #19761, opened 2022-10-31
libunistring: Update to 1.1

ejurgensen:
Signed-off-by: Espen Jürgensen <espenjurgensen+openwrt@gmail.com>

Maintainer: me
Compile tested: x86_64, r18609
Run tested: no

Description:
Version update


ejurgensen:
I'm not going to run-test this. The last 8 years, I have been maintainer of 4 OpenWrt packages and haven't seen this requirement before?

If it is now a hard requirement then I think you are doing yourself a disservice. It adds quite a bit work to me when upgrading packages, and I would question how much assurance it actually gives you that I write "yes" to run testing. I also think the likelyhood that a version update suddenly causes it not to run is very slim. So in the end, it's more likely to mean that I and other maintainers don't bother making PRs on upstream updates, and then OpenWrt doesn't stay up to date. Probably a much greater danger than non-running packages.

It's not that I'm against run testing as such, but you need a framework for it. You have excellent Github actions build checks -  perhaps you could come up with a concept where I can add a run test script to a Makefile, and then the Github actions can execute it? That would also give you transparency into what the test was, evidence that it was actually executed and means that a new maintainer doesn't need to reinvent a run test. Maybe you already have this, and I have missed the info, then please educate me.

PR #19763, opened 2022-10-31
pbr: initial commit

stangri:
Maintainer: me
Run tested: x86_64, Sophos SG-135, OpenWrt 22.03.2, test basic functionality with dnsmasq.nftset and simple policies

Description
* The makefile produces the nft and iptables capable `pbr` package and the `pbr-iptables` package for legacy setups
* This replaces `vpnbypass` and `vpn-policy-routing` packages
* I'm soliciting feedback on this package and my intention is to update the version to 1.0.0 before this is merged, but I need the feedback on this and luci-app-pbr before then.


net/pbr/Makefile
```diff
@@ -0,0 +1,198 @@
+# Copyright 2017-2022 Stan Grishin (stangri@melmac.ca)
+# This is free software, licensed under the GNU General Public License v3.
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=pbr
+PKG_VERSION:=0.9.9
+PKG_RELEASE:=24
```
BKPepe, line 8:
```suggestion
PKG_RELEASE:=1
```

net/pbr/Makefile
```diff
@@ -0,0 +1,198 @@
...
+
+define Package/pbr/default
+	SECTION:=net
+	CATEGORY:=Network
+	SUBMENU:=VPN
+	PROVIDES:=pbr
+	TITLE:=Policy Based Routing Service
+	URL:=https://docs.openwrt.melmac.net/pbr/
+	DEPENDS:=+ip-full +jshn +jsonfilter +resolveip
+	CONFLICTS:=vpnbypass vpn-policy-routing
```
BKPepe, line 22: Shouldn't pbr provide these packages?

net/pbr/Makefile
```diff
@@ -0,0 +1,198 @@
...
+	PKGARCH:=all
+endef
+
+define Package/pbr
+$(call Package/pbr/default)
```
BKPepe, line 27: Why this package provides itself based from ``define Package/pbr/default``?

net/pbr/Makefile
```diff
@@ -0,0 +1,198 @@
...
+
+PKG_NAME:=pbr
+PKG_VERSION:=0.9.9
+PKG_RELEASE:=24
+PKG_LICENSE:=GPL-3.0-or-later
+PKG_MAINTAINER:=Stan Grishin <stangri@melmac.ca>
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/pbr/default
```
BKPepe, line 14: This section should has two spaces as indentation.

net/pbr/Makefile
```diff
@@ -0,0 +1,198 @@
...
+This version supports OpenWrt with fw3/ipset/iptables.
+endef
+
+define Package/pbr-netifd/description
+This service enables policy-based routing for WAN interfaces and various VPN tunnels.
+This version supports OpenWrt with both fw3/ipset/iptables and fw4/nft.
+This version uses OpenWrt native netifd/tables to set up interfaces. This is WIP.
+endef
+
+define Package/pbr/default/conffiles
```
BKPepe, line 59:
```suggestion
define Package/pbr/conffiles
```

and then you need to modify the conffiles below.

net/pbr/Makefile
```diff
@@ -0,0 +1,198 @@
...
+Package/pbr-netifd/conffiles = $(Package/pbr/default/conffiles)
+
+define Build/Configure
+endef
+
+define Build/Compile
+endef
+
+define Package/pbr/default/install
+	$(INSTALL_DIR) $(1)/etc/config
```
BKPepe, line 74: What rather to create directory first and then what files should be in that directory and etc? Right now, it is not very clear unless going through into it deeply what goes where.

net/pbr/files/README.md
```diff
@@ -0,0 +1,3 @@
+# README
```
BKPepe, line 1: Hmm, readme should be here or it should be removed.

BKPepe (changes requested):

net/pbr/Makefile
stangri: > * I'm soliciting feedback on this package and my intention is to update the version to 1.0.0 before this is merged, but I need the feedback on this and luci-app-pbr before then.

stangri:
@BKPepe thank you for your review, I'll address your suggestions within a few days!

net/pbr/Makefile
stangri: Will be fixed in the next push.

net/pbr/Makefile
stangri: What's a better way to organize two packages: `pbr` and `pbr-iptables` from the same Makefile?

net/pbr/Makefile
stangri: Thanks, will be fixed in the next push. It's just the package definition section which needs two spaces vs tabs for indentation, correct?

net/pbr/Makefile
stangri: Sorry, I'm not sure I'm getting your point, could you rephrase it?

net/pbr/files/README.md
stangri: I strongly prefer to host the README and the screenshots it includes together, which is a no-go for this repo. Currently the README (and the screenshots) are hosted in another github repo and published thru github pages. 

Some people may discover/look for the README within the package directory that's why it's here.

net/pbr/Makefile
stangri: If I understood you correctly, this will be fixed in the next push.

net/pbr/Makefile
oldium: I think the idea was to group `INSTALL_DIR` with respective `INSTALL_DATA`/`BIN` so that it is clear which files go to which directories. Meaning to have something like this (like it is in `package/libs/libselinux/Makefile` for example):

```bash
$(INSTALL_BIN) ./files/etc/init.d/pbr.init $(1)/etc/init.d/pbr
$(SED) "s|^\(readonly PKG_VERSION\).*|\1='$(PKG_VERSION)-$(PKG_RELEASE)'|" $(1)/etc/init.d/pbr
$(INSTALL_DIR) $(1)/etc/hotplug.d/iface
$(INSTALL_DATA) ./files/etc/hotplug.d/iface/70-pbr $(1)/etc/hotplug.d/iface/70-pbr
…
```

This way it will be visible which directories are created empty for example.

BKPepe: Hey @oldium, thanks! This is exactly what I meant.

stangri: Thanks, I've reorganized that in 0.9.9-37 and will include this in the next push.

PR #19776, opened 2022-11-02
tailscale: preserve tailscaled state file

### LINKED ISSUE #19774: tailscale: /etc/tailscale/tailscaled.state not preserved
Maintainer: @ja-pa 
Environment: OpenWrt 21.02.3 r16554-1d4dea6d4f

Description:

The [tailscaled state file](https://github.com/openwrt/packages/blob/db71747697934f5da65c458926fa5def60e19411/net/tailscale/files/tailscale.conf#L5) is not included in configuration backups and is not preserved during sysupgrades.

CAFxX:
Signed-off-by: Carlo Alberto Ferraris <cafxx@strayorange.com>

Maintainer: @ja-pa 
Compile tested: N/A
Run tested: N/A (but I manually added the same line in the backup/restore configuration, and tailscale continued working correctly after a sysupgrade)

Description: Fixes #19774

### DIFF
--- a/net/tailscale/Makefile
+++ b/net/tailscale/Makefile
@@ -77,6 +77,7 @@ endef
 
 define Package/tailscaled/conffiles
 /etc/config/tailscale
+/etc/tailscale/tailscaled.state
 endef
 
 $(eval $(call BuildPackage,tailscale))


net/tailscale/Makefile
```diff
@@ -77,6 +77,7 @@ endef
 
 define Package/tailscaled/conffiles
 /etc/config/tailscale
+/etc/tailscale
```
BKPepe, line 80: You want to preserve state file, which is /etc/tailscale/tailscaled.state based on https://github.com/openwrt/packages/blob/db71747697934f5da65c458926fa5def60e19411/net/tailscale/files/tailscale.conf#L5 right? Thus I am not sure, why you want to preserve only /etc/tailscale

CAFxX:
```suggestion
/etc/tailscale/tailscaled.state
```

1715173329:
is this ready to go?

BKPepe:
Needs to be tested on the router itself.

1715173329:
This is a trivial fix. There's no syntax error, so it will just.work.

1715173329:
after having this package installed you can find `/lib/upgrade/keep.d/tailscaled`, and that's all.
prebuilt ipks are available at https://github.com/openwrt/packages/actions/runs/3394994736

PR #19733, opened 2022-10-29
Improvements for Pulseaudio package (conflicts, autorelease, simplifying)

BKPepe:
Maintainer: @tripolar
Compile tested: Raspberry PI 4 CM, aarch64_cortex-a72_musl, OpenWrt snapshot
Run tested: not yet

Description:
- Many changes, see the commit descriptions

PR #19783, opened 2022-11-03
pulseadio: fix conffiles for pulseadio-daemon-avahi package

BKPepe:
Maintainer: @tripolar
Compile tested: verified that conffiles in the packages are correct now
Run tested: N/A

Description:

- Fixes: https://github.com/openwrt/packages/pull/19733#issuecomment-1298842720
Thanks @hnyman for reporting!


hnyman (approved):
looks ok to me.

(didn't test the simple change to the bug I had noticed earlier)

PR #19803, opened 2022-11-04
remove various gerbera packages

neheb:
Maintainer:


BKPepe:
I think we can merge this, right?

neheb:
I'll probably merge later today. I need to add these to packages-abandoned.

PR #19810, opened 2022-11-04
Revert "fcgi: remove"

ynezz:
Compile tested: OpenWrt 22.03 x86/64, ipq40xx, mvebu/cortexa9
Run tested: OpenWrt 22.03 x86/64, ipq40xx, mvebu/cortexa9

This reverts commit 65ef6c646af7906f066badde93521b25c8445cdc where @neheb removed fcgi library for "Nothing uses this." reason.

I've noticed this missing library while rebasing downstream project from OpenWrt 19.07 release base to OpenWrt 22.03, as there is [amx-fcgi](https://gitlab.com/prpl-foundation/components/ambiorix/applications/amx-fcgi) system component which depends on this library.

So I would like to fix that regression by adding back the removed library package.

### DIFF
--- a/libs/fcgi/Makefile
+++ b/libs/fcgi/Makefile
@@ -1,0 +1,77 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=fcgi
+PKG_VERSION:=2.4.2
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)2-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/FastCGI-Archives/fcgi2/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=1fe83501edfc3a7ec96bb1e69db3fd5ea1730135bd73ab152186fd0b437013bc
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)2-$(PKG_VERSION)
+
+PKG_MAINTAINER:=Jacob Siverskog <jacob@teenageengineering.com>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE.TERMS
+
+PKG_FIXUP:=autoreconf
+PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=1
+
+include $(INCLUDE_DIR)/uclibc++.mk
+include $(INCLUDE_DIR)/package.mk
+
+define Package/fcgi/Default
+  SECTION:=libs
+  CATEGORY:=Libraries
+  URL:=https://fastcgi-archives.github.io/
+endef
+
+define Package/fcgi
+  $(call Package/fcgi/Default)
+  MENU:=1
+  DEPENDS:=+libpthread
+  TITLE:=Shared library of FastCGI
+endef
+
+define Package/fcgixx
+  $(call Package/fcgi/Default)
+  DEPENDS:=fcgi $(CXX_DEPENDS)
+  TITLE:=Shared library of FastCGI++
+endef
+
+define Package/fcgi/description
+ FastCGI is a language independent, scalable, open extension to
+ CGI that provides high performance without the limitations of
+ server specific APIs.
+endef
+
+TARGET_CXXFLAGS += -fno-rtti -flto
+TARGET_LDFLAGS += -Wl,--gc-sections,--as-needed
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/include
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/fastcgi.h $(1)/usr/include/
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/fcgi{app,_config,misc,o,os,_stdio}.h $(1)/usr/include/
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libfcgi{,++}.{a,so*} $(1)/usr/lib/
+endef
+
+define Package/fcgi/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/cgi-fcgi $(1)/usr/bin/
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libfcgi.so.* $(1)/usr/lib/
+endef
+
+define Package/fcgixx/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libfcgi++.so.* $(1)/usr/lib/
+endef
+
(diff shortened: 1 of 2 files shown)


neheb:
I probably also removed it as it was failing in some other way (maybe nls or autoconf issues).

Where is this amx-fcgi used?

ynezz:
> I probably also removed it as it was failing in some other way (maybe nls or autoconf issues).

I can't find anything regarding this in the issue tracker, it compiles and works fine, CI here doesn't complain either.

> Where is this amx-fcgi used?

[tr181-rest-api](https://gitlab.com/prpl-foundation/prplos/feed-prpl/-/blob/prplos/network/utilities/tr181-rest-api/Makefile#L13) for example.

ynezz:
Here is the regression exposed by the latest OpenWrt 22.03.2 x86/64 SDK: 

```cc
x86_64-openwrt-linux-musl-gcc -Os -pipe -fno-caller-saves -fno-plt -fhonour-copts -Wno-error=unused-but-set-variable -Wno-error=unused-result -ffile-prefix-map=/home/build/openwrt/build_dir/target-x86_64_musl/amx-fcgi-v0.2.3=amx-fcgi-v0.2.3 -Wformat -Werror=format-security -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro  -I/home/build/openwrt/staging_dir/toolchain-x86_64_gcc-11.2.0_musl/usr/include -I/home/build/openwrt/staging_dir/toolchain-x86_64_gcc-11.2.0_musl/include/fortify -I/home/build/openwrt/staging_dir/toolchain-x86_64_gcc-11.2.0_musl/include  -Werror -Wall -Wextra -Wformat=2 -Wshadow -Wwrite-strings -Wredundant-decls -Wmissing-declarations -Wno-attributes -Wno-format-nonliteral -fPIC -g3 -I ../include_priv -I /home/build/openwrt/staging_dir/target-x86_64_musl/include -I /home/build/openwrt/staging_dir/target-x86_64_musl/usr/include -Wstrict-prototypes -Wold-style-definition -Wnested-externs -std=gnu11 -c -o ../output/x86_64-openwrt-linux-musl/object/amx_fcgi.o amx_fcgi.c
In file included from ../include_priv/amx_fcgi.h:69,
                 from amx_fcgi.c:64:
../include_priv/amx_fcgi_types.h:69:10: fatal error: fcgiapp.h: No such file or directory
   69 | #include <fcgiapp.h>
      |          ^~~~~~~~~~~
compilation terminated.
```

It is working fine with OpenWrt 19.07 SDK.

champtar:
@ynezz could you maybe add a comment in the makefile pointing to the external users of the lib ? It's a bit weird to keep libraries with no users.
Also should you set yourself as maintainer ?

ynezz:
>  It's a bit weird to keep libraries with no users.

Fair enough, I wasn't aware about this rule.

champtar:
IMO It's fine to keep it in as there is at least one known user, but it's better to make it explicit so we don't delete it again in the future.

neheb:
I think if introduced it should also be removed from packages-abandoned.

ynezz:
> IMO It's fine to keep it in as there is at least one known user, but it's better to make it explicit so we don't delete it again in the future.

Do we really want  to start adding this kind of build system unrelated "hey, I'm using it, don't delete it" noise into the Makefiles? How often you would need to update it, so you know, that the users are still active? Once a year? IMO it's going to bit rot, folks can simply miss the information and delete the package anyway. It should probably be documented as well.

What about simply assuming, that there always might be some users out there, so don't delete the package in the first place? If you really need to remove it for some reason, then simply make that reason clear in the commit message and allow adding the package back.

ynezz:
> Also should you set yourself as maintainer ?

Done.

ynezz:
> I think if introduced it should also be removed from packages-abandoned.

[Prepared](https://github.com/openwrt/packages-abandoned/pull/28).

dhewg:
Not directly related, but:
I've tried to get some prpl hosted firmware files released under a compatible license for OpenWrt. I've used their issue tracker for that some month ago. Now they've shut down access to it.
Apparently they don't care about open source communities, so why should OpenWrt care about them and maintain this without in-tree users?

ynezz:
> I've tried to get some prpl hosted firmware files released under a compatible license for OpenWrt. I've used their issue tracker for that some month ago. 

As you've probably already noticed, those files reside under `Intel` vendor/member subgroup and are `Copyright (c) 2016 Lantiq Beteiligungs GmbH & Co KG.`. Basically prpl itself can't do much about it since unfortunately none of those files are being used/distributed in any form by prpl. Otherwise it would've need to follow [IP Policy](https://prplfoundation.org/about/ip-policy/) and thus have FOSS friendly license already.

What prpl can do in this case is to politely ask MaxLinear representatives to consider re-licensing those files, which was done and AFAIK it's currently waiting for final approval from their legal department, but you're hopefully already aware.

> Now they've shut down access to it.

I believe, that its not intentional, IMO its just being caused by the ongoing Jira system migration from on-premise to the new cloud hosted instance. Thank you for letting me know, I'll try to ping responsible folks and ask them for a fix.

> so why should OpenWrt care about them and maintain this without in-tree users?

IMO OpenWrt is a FOSS project with a clear contribution rules and guidelines, which anyone could reference in exactly such ambiguities.

IIUC then in FOSS projects "Upstream First" philosophy means, that whenever you solve a problem in your copy of the upstream code which others could benefit from, you contribute these changes back upstream, for example you send a patch or open a PR in the upstream project.

IMO thats exactly what I'm doing, I'm trying to upstream [a fix](https://gitlab.com/prpl-foundation/prplos/feed-prpl/-/commit/67a54ad6218e869f20a7acb7fb4513bf81313d11) for a downstream project breakage, so anyone else using `fcgi` library in OpenWrt 19.07 based project and migrating to any later versions could get the fix.

dhewg:
> What prpl can do in this case is to politely ask MaxLinear representatives to consider re-licensing those files, which was done and AFAIK it's currently waiting for final approval from their legal department, but you're hopefully already aware.

I was aware of an initial ping from a few months ago. _Final approval_ sounds better than that though.

> I believe, that its not intentional

I checked their website first, but couldn't find any info on any migration, so I assumed it is intentional. 

Thanks, let's hope that gets resolved.

ynezz:
> Thanks, let's hope that gets resolved.

https://prplfoundationcloud.atlassian.net/browse/PCI-16

dhewg:
Nice, thanks!
If that got resolved that quick due to you poking around, I urge you to poke some more to get that the desired response from MaxLinear.

libs/fcgi/Makefile
```diff
@@ -0,0 +1,77 @@
...
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=fcgi
+PKG_VERSION:=2.4.2
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)2-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/FastCGI-Archives/fcgi2/tar.gz/$(PKG_VERSION)?
```
Ansuel, line 13: Considering some recent problem with github codeload I think it's better to use git clone directly. Can you change that?

ynezz: Done.

GeorgeSapkin:
@ynezz still relevant?

PR #19837, opened 2022-11-06
crun: Update to version 1.6

hauke:
Maintainer: @oskarirauta  
Compile tested: mips_24kc_glibc
Run tested: none

This fixes compilation problems with glibc 2.36.

Full changelog:
* crun-1.6

- runc compatibility: -v now prints the version string.
- build: fix build with glibc 2.36.
- container: drop intermediate userns custom feature.
- cgroup: change the delegate cgroup semantic so that the cgroup is created in the container payload after the cgroup namespace is created.
- seccomp: use helper process to send file descriptor to the listener socket.  It enables to be notified on every syscall without hanging the main process.
- linux: add a fallback to using kill(2) if pidfd_send_signal(2) fails with ENOSYS.
- krun: add support for krun-sev.
- wasmtime: always grant file system capability for workdir inside the container.
- wasmtime: inherit arguments list from the handler instead of the current process.
- wasmedge: use released wasmedge library instead of libwasmedge_c.so.

* crun-1.5

- add mono based native .NET handler
- new Wasmtime backend for running WebAssembly
- add support for wasmedge 0.10 and dropping support for wasmedge 0.9.x
- dropping support for experimental `WasmEdgeProcess` from wasmedge handler
- honor process user's uid when setting the HOME environment variable
- create the current working directory if it is missing in the container
- fallback to using a tmpfs mount if umount of /sys and /proc fails
- fallback to netlink to setup lo device
- fix creating devices in the rootfs
- fallback to using io.weight if io.bfq.weight doesn't exist
- remove tun/tap from the default allow list
- linux: devices mounts have noexec and nosuid
- fix copyup of files from the container to the tmpfs
- honor $PATH for newgidmap and newguidmap
- krun: limit the number of vCPUs to 8
- cgroup: add support for cpu.idle

Signed-off-by: Hauke Mehrtens <hauke@hauke-m.de>


oskarirauta:
Thanks for doing the work for me, I've been super busy last six months doing other things. I am okay for this PR. Please proceed.

PR #19849, opened 2022-11-06
expat: update to 2.5.0

neheb:

Maintainer: @thess

### DIFF
--- a/libs/expat/Makefile
+++ b/libs/expat/Makefile
@@ -6,12 +6,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=expat
-PKG_VERSION:=2.4.9
+PKG_VERSION:=2.5.0
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_SOURCE_URL:=@SF/expat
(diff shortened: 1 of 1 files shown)


diizzyy:
Release tarball is also offered on GitHub
https://github.com/libexpat/libexpat/releases/tag/R_2_5_0

neheb:
Nasty part here is the URL is https://github.com/libexpat/libexpat/releases/download/R_2_5_0/expat-2.5.0.tar.xz

need to find some way to underscorify the version number.

1715173329:
`$(subst .,_,$(PKG_VERSION))` should be okay.

neheb:
The standard is to use tarballs when possible. git tarballs whether GitHub or OpenWrt ones have historically had issues with changing hashes.

1715173329:
> The standard is to use tarballs when possible.

Got it.

> git tarballs whether GitHub or OpenWrt ones have historically had issues with changing hashes.

Hummm... But this package uses sourceforge releases?

neheb:
Right. The tarball @diizzyy mentioned is the same one as the sourceforge one. The goal is to have mirrors in case sourceforge is not available.

1715173329:
Okay I understood, thanks!
`subst` should be fine in this case.

BKPepe:
In your commit description, you should mention that you added a mirror otherwise, you are doing more changes than necessary. Also, it would help if you mentioned [CVE fix](https://www.xml.com/news/2022-10-expat-250/) in the commit description.

PR #19850, opened 2022-11-06
taglib: update to 1.13

neheb:

Maintainer: nobody


diizzyy:
There's a release tarball available for taglib
https://github.com/taglib/taglib/releases/tag/v1.13

neheb:
So that was reverted in https://github.com/openwrt/packages/commit/b0dc454c0ec20d3e2a4994d30657de2f4e21fe55

I think it makes sense to keep codeload to avoid having to constantly update the URL.

PR #19866, opened 2022-11-09
isc-dhcp: Fix logic to detect if the config file is unchanged

pprindeville:
Maintainer: me
Compile tested: x86_64, generic, HEAD (c3b9f00aaa)

Description:

Compare config files after regenerating the new version.


pprindeville:
Cc @jempatel

Issue #19875, opened 2022-11-10
Nginx: support dynamic wan IP


hgl:
Maintainer: @Ansuel @@peter-stadler

Description:

Since Nginx [doesn't have the ability to listen on interfaces](https://stackoverflow.com/questions/39536714/nginx-bind-to-a-specific-network-interface-regardless-of-ip-address), and WAN having dynamic IPs is fairly common in residential networks, it'd be great if the Nginx could listen on WAN and still work when its IPs change.

I think one solution could be that this package natively supports template config, where WAN IPs could be replaced whenever they change. The implementation involves using iface hotplug to gather changed WAN IPs, writing them to a standard location, and generating a procd event, and then in Nginx's init.d file, the templates are rendered and imported whenever the procd event is triggered.

Templates could just be simple shell scripts that read the written WAN IPs, and write Nginx conf to stdout.

Do you think this feature is a good idea? If so I'd be happy to implement it and send a PR.

Let me know if the description is clear.

hgl:
Of course, templates are totally an addon. Only when users put scripts in certain location, say, `/etc/nginx/dynamic.conf.d/site.sh`, do they take effect.

Ansuel:
@hgl honestly i have a similar problem and currently my solution is to disable and use a custom nginx.conf... I like the idea of adding an hotplug script and generate the site config. Ideally this should be configurable to listen for a specific interface

hgl:
Cool, sounds like the feature is warranted.

There is one thing I'd like to add though. 

As I read the source code of the nginx package I realized it depends on nginx-util, which I believe handles UCI and also provides SSL cert and config auto management. I have no doubt it's a useful utility to have, and am grateful to @peter-stadler for the efforts. But may I propose dropping this dependency, and instead make it an optional companion package for nginx, for the following reasons:

1. Making nginx UCI compatible isn't that useful

    I think the major benefit of UCI is automatic LuCI app support, but given the complexity of nginx's config system, a satisfying luci-app-nginx is unlikely to happen. Having to have a mental translation between the UCI and native syntax also makes it harder to config nginx with UCI.

2. It makes it harder to contribute to this package

    In my case, in order to contribute to this package in a deeper way, I need to understand how nginx-util works, which isn't very well documented and isn't written in a language that this package is primarily written with (shell script). This can put quite a burden on future contributors to this package, since you need to understand cpp in order to understand how nginx conf is generated. No other major packages like dnsmasq and dropbear work like this.

[...]

PR #19878, opened 2022-11-11
nginx: drop dependency on nginx-util

hgl:
Maintainer: @Ansuel @peter-stadler

Description:

The main feature of nginx-util is to make Nginx config UCI compatible. However, for the nginx package, the benefit of being UCI compatible is debatable.

There is currently no luci-app-nginx. Given the complexity of Nginx config system, one that can thoroughly support all Nginx config features is unlikely to happen. If it were to be developed, and only serves to provide basic web server features, it really should be luci-app-uhttpd instead. So, without the prospect of a corresponding LuCI app, being UCI compatible adds complexity unnecessarily.

This change simplifies this package by providing a basic native Nginx config that optionally supports serving LuCI. Users are expect to directly edit the nginx config.

A few design choices/changes:

- Since there is no reliable method to read Nginx SSL options and determine if a certificate should be generated automatically, no default https server is provided.

- Not specifying config files via "procd_set_param file" in /etc/init.d/nginx, since they modifications shouldn't cause Nginx to restart.

I'd like to make this PR serve as a base for #19875. Let me know if it's a terrible idea.


hgl:
@Ansuel I'd like to implement #19875, would appreciate you feedback on this PR. If you accept it, that's great and I can continue implementing. If not, let's work something out?

hgl:
Continued on #19884

PR #19881, opened 2022-11-11
nginx: update to 1.22.1

hgl:
Maintainer: @Ansuel

Description:

### DIFF
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nginx
-PKG_VERSION:=1.21.3
-PKG_RELEASE:=3
+PKG_VERSION:=1.22.1
+PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=nginx-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://nginx.org/download/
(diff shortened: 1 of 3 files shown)


hgl:
@Ansuel Could you review this PR? It's unrelated to the other PR I sent. One test failed because the target is unsupported, it's already excluded in the Makefile.

hgl:
Forgot to cc @peter-stadler

hgl:
Forgot mention, many modules have to be updated also, since the old versions don't support the new nginx any more.

BKPepe:
Where is the PR template? You removed any indication of compile and run testing on real device.

hgl:
@BKPepe, thanks for the reminding. I was busy getting it to compile and forgot to test it on a real device.

@peter-stadler @Ansuel, it seems [lua-nginx-module](https://github.com/openresty/lua-nginx-module) is now very picky about the nginx environment it's run in. For starters, it generates a warning if luajit isn't OpenResty's fork, and after the warning, nginx [fails with this error](https://github.com/openresty/lua-nginx-module/issues/1533), without any Lua related directives in the nginx conf.

I'm not an expert on lua-nginx-module, but from my research, openresty really wants people to just use their version of nginx with bundled Lua, and provides little instructions on to compile with official nginx.

I wonder if it's the reason this package hasn't been updated for a long time?

If Lua must be supported, I'd appreciate some help.

If Lua support could be dropped, I'd be happy to remove it and give it another try.

PR #19893, opened 2022-11-14
libassuan:fix gpg_error unkown

doorxp:

Description:


BKPepe:
All 11 checks are failing for you. Would you mind to fix that?

PR #19907, opened 2022-11-16
ooniprobe: Update to 3.16.5

1715173329:
Maintainer: @ja-pa 
Compile tested: CI (immortalwrt/packages#527)
Run tested: CI (immortalwrt/packages#527)

Description:
Fix build with Go 1.19. #19652

### DIFF
--- a/net/ooniprobe/Makefile
+++ b/net/ooniprobe/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ooniprobe
-PKG_VERSION:=3.14.2
+PKG_VERSION:=3.16.5
 PKG_RELEASE:=1
 
 PKG_SOURCE:=probe-cli-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


BKPepe:
From my point of view, having something tested inside CI is just additional or, let's say, something nice to have, but that still can not replace the indication of run testing this on real device. The ``test.sh`` just ensures you that your application does not segfault immediately once issuing ``--help/--version``, but that does not mean that it works with the old setup.

1715173329:
I don't use this package, so can't do a real test and also don't know how to do it. This PR just makes it compile.

Ping @ValdikSS would you mind testing it?

ValdikSS:
> Ping @ValdikSS would you mind testing it?

Sorry, I don't have access to my laptop in this and next week. Try to run it. If it runs, then it is probably fine.

1715173329:
> Sorry, I don't have access to my laptop in this and next week. Try to run it. If it runs, then it is probably fine.

No problem, thank you!

1715173329:
emmmm... it does run... but looks like in this package there're lots of unrelated stuffs, even includes sample binaries... are they really necessary?
https://github.com/ooni/probe-cli/tree/master/internal/tutorial/measurex

PR #19912, opened 2022-11-16
luarocks: update to version 3.9.2

TjeuKayim:
Updating luarocks. They changed the build system upstream, so the old patches no longer worked.

Patches:

- Fixing the shebang just like the previous patch did

- Adding '--with-lua-bin=/usr/bin --with-lua=/usr/bin' to HOST_CONFIGURE_ARGS did not work, so patching the GNUmakefile as workaround.

- The OpenWRT Lua package has no header files, so patching the lua_h_exists() function to continue instead of error. Older versions of luarocks were also able to run without a C compiler and include files.

Tested this merge request with the SDK targetting ramips-mt7621 and OpenWRT version 22.03.2, it can run `luarocks install luassert` successfuly.

@amrhassan, could you please have a look, as you are the maintainer of this package.

Unresolved questions:

- [ ] It does not work with https://openwrt.org/packages/pkgdata/luaposix. If that package is not installed, luarocks falls back to another implementation that does work. If the package is installed, `fs_use_modules=false` in the config file is a workaround. Should this package patch the default for that config option?


TjeuKayim:
As 4 months passed, I rebased on top of master. Does someone want to give a code review?
Maybe @neheb

GeorgeSapkin:
Looks like luarocks is still outdated in the main branch, so still relevant. @BKPepe maybe?

Related:
- #27884

PR #19919, opened 2022-11-16
CI: deprecate $(AUTORELEASE) via comments

aparcar:
Autorelease causes some issues like heavy bandwidth usage as well as non-deterministic package releases whenever someone doesn't use the full git log.

With this comment all modified packages are checked and if they use the autorelease feature, kindly comment to the user to change that.

Signed-off-by: Paul Spooren <paul.spooren@rhebo.com>

### DIFF
--- a/.github/workflows/check-autorelease-deprecation.yml
+++ b/.github/workflows/check-autorelease-deprecation.yml
@@ -1,0 +1,86 @@
+name: Check autorelease deprecation
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, converted_to_draft, ready_for_review, edited]
+
+jobs:
+  build:
+    name: Check autorelease deprecation
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+
+    permissions:
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: Determine branch name
+        run: |
+          BRANCH="${GITHUB_BASE_REF#refs/heads/}"
+          echo "Building for $BRANCH"
+          echo "BRANCH=$BRANCH" >> $GITHUB_ENV
+
+      - name: Determine changed packages
+        run: |
+          # only detect packages with changes
+          PKG_ROOTS=$(find . -name Makefile | \
+            grep -v ".*/src/Makefile" | \
+            sed -e 's@./\(.*\)/Makefile@\1/@')
+          CHANGES=$(git diff --diff-filter=d --name-only origin/$BRANCH...)
+
+          for ROOT in $PKG_ROOTS; do
+            for CHANGE in $CHANGES; do
+              if [[ "$CHANGE" == "$ROOT"* ]]; then
+                if grep -q '$(AUTORELEASE)' "$ROOT/Makefile"; then
+                  CONTAINS_AUTORELEASE+="$ROOT"
+                fi
+                break
+              fi
+            done
+          done
+
+          if [ -n "$CONTAINS_AUTORELEASE" ]; then
+            cat > "$GITHUB_WORKSPACE/pr_comment.md" << EOF
+          Please do no longer set *PKG_RELEASE* to *AUTORELEASE* as the
+          feature is deprecated. Please use an integer instead. Below is a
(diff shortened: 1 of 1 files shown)


Ansuel:
@aparcar what is the idea to deprecate it? Keep it till we update them and hardcode it to a dummy high number? Parse the current autorelease value for the specific package and hardcode to the new one?

aparcar:
Ideally the bot comments people to drop AUTORELEASE by suggesting the correct PKG_RELEASE, it would look similar to this:
image

I'd prefer that to a massive PR changing everything. In another iteration this PR can be changed to check that people bump the PKG_RELEASE the right way, as it's done in the Alpine Linux CI.

Ansuel:
Ok so AUTORELEASE will still be there and used, just flag it as deprecated... IMHO a good idea would be add a time where the feature will be dropped and after that replace everything. What do you think? (just to promote the change even more for active packages)

Ansuel:
Also create an issue and pin it ?

1715173329:
No idea why CI is failing.

Ansuel:
require write permission

aparcar:
I'll fix that tomorrow

Ansuel:
https://github.com/nf-core/tools/issues/765#issuecomment-723339546 ok now it makes sense....

Ansuel:
@aparcar ok looks to work https://github.com/Ansuel/packages/pull/2
(tested by pushing the ci to master and opening a pr with the single testing commit)

We need to use the different event as pull_request doesn't permit write access to pull request when run from another fork (due to security reason)
So we need to use a different target but that event to work needs to be present on master or the event won't be triggered.

So this is ready... tell me how we want to proceed with this.

aparcar:
Looks good to me. Let's remove the Test commit and merge it?

Ansuel:
@aparcar ok for me. I can drop the commit if you want 
I would test this in this repo and then also include it in the openwrt repo? (or for that we should directly change every entry to a release?)

aparcar:
@Ansuel please proceed and yea I'd like to test it here and then move it to all other repos

BKPepe:
What about ``COMMITCOUNT``? :thinking:

hnyman:
@Ansuel @aparcar 
Suggestion regarding the "positive" text.
The automatic message is pretty confusing in a package that is not using AUTORELEASE, as there is suddenly a notice that "no usage of AUTORELEASE found". 
If you are not aware of this discussion, the message is unclear.
Is that meant to trigger the author to start using AUTORELEASE ????

image

I suggest that the "not found" text is changed to clearly :

`echo "No usage of *AUTORELEASE* found in changes" `

-->

`echo "No usage of *AUTORELEASE* found in changes. Good, as *AUTORELEASE* is being deprecated." `
or
`echo "No usage of deprecated *AUTORELEASE* found in changes. Good." `

Or something like that, so that the message is clear.

Ansuel:
@hnyman yes good idea...
I would use something less friendly like
`*AUTORELEASE* is now deprecated and should be dropped. These changes DOESN'T USE *AUTORELEASE* and DOESN'T REQUIRE additional deprecation fixup.`

What do you think?

hnyman:
Please say somehow positively that "this PR is ok".  
There is no reason to be "less friendly" in the target PR, (but we want to promote the idea of changing other packages).

e.g.
echo "No usage of *AUTORELEASE* found in changes. Good, as *AUTORELEASE* is being deprecated." 
echo "*AUTORELEASE* is being deprecated. Great that packages in this PR do not use it." 
echo "Notice: *AUTORELEASE* is being deprecated and should be dropped.\nGreat that packages in this PR do not use it." 

something like that

Ansuel:
OK could be that i'm too cold today ahhaha

Anyway I would chose the last one , will prepare a pr (anyway my concern was that an user by reading "good" would indirectly assume that the changes are entirely good)

BKPepe:
Guys, I find this weird. Why are we deprecating autorelease only in the packages repository (= this repository)? It does not make sense because there are several packages in the main repository, which are using autorelease. Shouldn't the main repository have some standards or policies that other repositories follow afterward?

Ansuel:
@BKPepe the idea was to test this on package repo and then port it to other repo...

My idea was that since we have total control on the core packages we could just directly drop AUTORELEASE use in core packages while the feature is dropped slowly here while packages gets updated/fixed.

PR #19921, opened 2022-11-16
Update uhubctl to 2.5.0

pesa1234:
update to new version 2.5.0

Signed-off-by: Andrea Pesaresi<andreapesaresi82@gmail.com>

Maintainer: Sorry for that I don't know, should be Steven Honson (on makefile right?)
Compile tested: (arm_cortex-a9_vfpv3-d16, WRT1900ACS v2, OpenWrt master)
Run tested: (arm_cortex-a9_vfpv3-d16, WRT1900ACS v2, OpenWrt master)

Description:

- Added support for Linux sysfs based power switching provided in Linux kernel 6.0+ - it allows to solve reliability issues when turning power off on Linux (https://github.com/mvp/uhubctl/pull/450).

-     Added option --nodesc to skip querying device string descriptors (necessary for some buggy devices which otherwise would completely freeze).

-     New simpler way to configure udev rules on Linux (one rule works for any USB hub).

-     Even more supported devices.


pesa1234:
Sorry, I close an I try again...

PR #19927, opened 2022-11-17
acme: make certificates use .crt as the extension

hgl:

Maintainer: @tohojo

Description:
This was an omission from my part in the initial PR.

Since uhttpd uses `.crt` as the certification extension name, I think the standard name in acme package should follow that convention, and we should make the change when there aren't many acme packages that have adopted the new architecture.

This change should be backward-compatible, since it only affects new certificates.

### DIFF
--- a/net/acme-acmesh/files/hook.sh
+++ b/net/acme-acmesh/files/hook.sh
@@ -46,7 +46,8 @@ get)
 			case $status in
 			0)
 				$NOTIFY renewed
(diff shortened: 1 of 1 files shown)


tohojo:
Sure, can do, but please look into the failing tests...

hgl:
My bad, I thought I should update the package version, and updated acme.sh's instead.

PR #19351, opened 2022-09-13
telegraf: Update to version 1.24.0

JonnyTischbein:
Signed-off-by: Jonathan Pagel [jonny_tischbein@systemli.org](mailto:jonny_tischbein@systemli.org)

Maintainer:
me

Compile tested:
on amd64 for aarch64_cortex-a53

Run tested:
only compile test this time

Description:
Add package with version 1.24.0

Telegraf is a plugin-driven agent for collecting and sending metrics and events. It supports various inputs (including prometheus endpoints) and is able to send data into InfluxDB. https://www.influxdata.com/time-series-platform/telegraf/

PR #19931, opened 2022-11-17
telegraf: Update to version 1.28.1

theatischbein:
Signed-off-by: Jonathan Pagel [jonny_tischbein@systemli.org](mailto:jonny_tischbein@systemli.org)

Maintainer:
me

Compile tested:
on amd64 for aarch64_cortex-a53

Run tested:
only compile test this time

Description:
Add package with version 1.28.1

Telegraf is a plugin-driven agent for collecting and sending metrics and events. It supports various inputs (including prometheus endpoints) and is able to send data into InfluxDB. https://www.influxdata.com/time-series-platform/telegraf/

### DIFF
--- a/utils/telegraf/Makefile
+++ b/utils/telegraf/Makefile
@@ -3,12 +3,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=telegraf
-PKG_VERSION:=1.23.4
+PKG_VERSION:=1.28.1
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/influxdata/telegraf/tar.gz/v$(PKG_VERSION)?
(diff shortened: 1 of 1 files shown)


JonnyTischbein:
I tried to add an commit with config similar to `PKG_USE_MIPS16 = 0` which excludes theses targets, since in since version and the previous one (https://github.com/openwrt/packages/pull/19351) only arm cortex a9 is failing while linking.
https://github.com/openwrt/openwrt/blob/25b65f548dfd93cae87781276bfff9a27cd3ebd4/include/package.mk#L27-L36

Do you have any recommendations what I did wrong ?

utils/telegraf/Makefile
```diff
@@ -18,6 +18,10 @@ PKG_BUILD_DEPENDS:=golang/host
 PKG_BUILD_PARALLEL:=1
 PKG_USE_MIPS16:=0
 
+#Filter out cortex-a9 due to build errors
```
BKPepe, line 21: IMHO This is wrong.

PolynomialDivision:
I would suggest sending the linking error to telegraf maintainers https://github.com/influxdata/telegraf/issues, but first try bumping to `1.24.4`.

Issue #19880, opened 2022-11-11
libassuan/libksba: build error against libgpg-error


feckert:
Maintainer: @neheb
Environment: x86_64, APU3, latest openwrt build

Description:
Since the following [commit](https://github.com/openwrt/packages/commit/b83c56576a79a8be01d5080f72cf540e17feb6a2) I have build error for the libs [libksba](https://github.com/openwrt/packages/blob/master/libs/libksba/Makefile) and [libassuan](https://github.com/openwrt/packages/blob/master/libs/libassuan/Makefile). I have remove the your commit I get an build error for [libgcrypt](https://github.com/openwrt/packages/blob/master/libs/libgcrypt/Makefile).

libksba:
```
...
checking for cc for build... gcc
checking if gcc supports -Wpointer-arith... yes
checking if gcc supports -Wno-pointer-sign... yes
checking whether the visibility attribute is supported... yes
checking for broken visibility attribute... no
checking for broken alias attribute... no
checking if gcc supports -fvisibility=hidden... yes
checking for gpg-error-config... /home/feckert/workspace/openwrt/LDM-master-x86_64/build/openwrt/staging_dir/target-x86_64_musl/host/bin/gpg-error-config
checking for gpgrt-config... /home/feckert/workspace/openwrt/LDM-master-x86_64/build/openwrt/staging_dir/target-x86_64_musl/host/bin/gpgrt-config
./configure: line 14388: test: : integer expression expected
./configure: line 14391: test: : integer expression expected
checking for GPG Error - version >= 1.8... no
configure: error: libgpg-error is needed.
                See ftp://ftp.gnupg.org/gcrypt/libgpg-error/ .
make[3]: *** [Makefile:75: /home/feckert/workspace/openwrt/LDM-master-x86_64/build/openwrt/build_dir/target-x86_64_musl/libksba-1.6.2/.configured_68b329da9893e34099c7d8ad5cb9c940] Error 1
```
libassuan:
```
...
checking for an ANSI C-conforming const... yes
checking for inline... inline
checking for size_t... yes
checking for sys/socket.h... yes
checking for socklen_t... yes
checking for struct cmsghdr.cmsg_len... yes
checking for gpg-error-config... /home/feckert/workspace/openwrt/LDM-master-x86_64/build/openwrt/staging_dir/target-x86_64_musl/host/bin/gpg-error-config
[...]

PR #19932, opened 2022-11-18
libgpg-error: copy gpg-error.pc to staging directory

flyn-org:
Packages like libassuan and libksba would not compile after updating libgpg-error to 1.46. This was because gpg-error.m4 (and thus the affected packages' configure scripts) relies on gpgrt-config, which in turn needs access to gpg-error.pc. This modifies the libgpg-error build process to copy gpg-error.pc to OpenWrt's staging directory, so that it is available for subsequent dependent package builds.

Fixes: https://github.com/openwrt/packages/issues/19880

Compile tested: x86_64 master

Description:

### DIFF
--- a/libs/libgpg-error/Makefile
+++ b/libs/libgpg-error/Makefile
@@ -91,6 +91,11 @@ define Build/InstallDev
 	$(INSTALL_DATA) \
 		$(PKG_INSTALL_DIR)/usr/share/aclocal/gpgrt.m4 \
 		$(1)/usr/share/aclocal/
+
+	$(INSTALL_DIR) $(1)/usr/lib/pkgconfig
+	$(CP) \
+		$(PKG_INSTALL_DIR)/usr/lib/pkgconfig/gpg-error.pc \
+		$(1)/usr/lib/pkgconfig
 endef
 
 define Package/libgpg-error/install


flyn-org:
@feckert, can you confirm this fixes https://github.com/openwrt/packages/issues/19880? I was able to build libassuan and libksba with this change applied.

libs/libgpg-error/Makefile
```diff
@@ -91,6 +91,11 @@ define Build/InstallDev
 	$(INSTALL_DATA) \
 		$(PKG_INSTALL_DIR)/usr/share/aclocal/gpgrt.m4 \
 		$(1)/usr/share/aclocal/
+
+	$(INSTALL_DIR) $(STAGING_DIR)/usr/lib/pkgconfig
```
feckert, line 95: Could you please replace `$(STAGING_DIR) `with `$(1)`. This will add the file to the package list and not copy it directly to the staging directory. The advantage is that during development, when a package clean-up call is made, all installed files are properly cleaned up from the staging directory for this packages
Take a look at the openwrt buildsystem https://github.com/openwrt/openwrt/blob/master/include/package.mk#L231

libs/libgpg-error/Makefile
```diff
@@ -91,6 +91,11 @@ define Build/InstallDev
...
+	$(CP) \
+		$(PKG_INSTALL_DIR)/usr/lib/pkgconfig/gpg-error.pc \
+		$(STAGING_DIR)/usr/lib/pkgconfig
```
feckert, line 98: See my comment above.

feckert (changes requested):
Your fix has repaired the build. If you can implement my comment, then we can merge it. :+1:

libs/libgpg-error/Makefile
flyn-org: I pushed this change.

1715173329: > I pushed this change.

Really? Looks like nothing happended. Forgot '--force'?

flyn-org: I did indeed forget `--force`. I just pushed again.

egorenar:
Hmm, i tested this change and it didn't help me, still getting the following error:
```
checking for gpg-error-config... /home/egorenar/Repositories/openwrt-rel/staging_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/host/bin/gpg-error-config
checking for gpgrt-config... /usr/bin/gpgrt-config
configure: Use gpgrt-config with /usr/lib as gpg-error-config
checking for GPG Error - version >= 1.8... yes (1.45)
configure: WARNING:
***
*** The config script "/usr/bin/gpgrt-config --libdir=/usr/lib gpg-error" was
*** built for x86_64-unknown-linux-gnu and thus may not match the
*** used host arm-openwrt-linux-gnu.
*** You may want to use the configure option --with-libgpg-error-prefix
*** to specify a matching config script or use $SYSROOT.
***
```

egorenar:
Adding this to libksba Makefile helped me:
```
CONFIGURE_ARGS += \
	ac_cv_path_GPGRT_CONFIG="$(STAGING_DIR)/host/bin/gpgrt-config"
```

But still getting this WARNING:
```
checking for gpg-error-config... /home/egorenar/Repositories/openwrt-rel/staging_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/host/bin/gpg-error-config
checking for gpgrt-config... (cached) /home/egorenar/Repositories/openwrt-rel/staging_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/host/bin/gpgrt-config
configure: Use gpgrt-config with /home/egorenar/Repositories/openwrt-rel/staging_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/host/lib as gpg-error-config
checking for GPG Error - version >= 1.8... yes (1.46-unknown)
configure: WARNING:
***
*** The config script "/home/egorenar/Repositories/openwrt-rel/staging_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/host/bin/gpgrt-config --libdir=/home/egorenar/Repositories/openwrt-rel/staging_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/host/lib gpg-error" was
*** built for arm-unknown-linux-gnueabi and thus may not match the
*** used host arm-openwrt-linux-gnu.
*** You may want to use the configure option --with-libgpg-error-prefix
*** to specify a matching config script or use $SYSROOT.
***
```

egorenar:
The same issue with gpgme package.

hnyman:
Let's merge this, as it seems to fix things (at least partially).
Hopefully buildbot will then show the remaining problems.

PR #19933, opened 2022-11-18
vallumd: updates

stintel:
Maintainer: me
Download from Codeberg, bump to latest version.

### DIFF
--- a/net/vallumd/Makefile
+++ b/net/vallumd/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=vallumd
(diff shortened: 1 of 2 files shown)


hnyman:
@stintel 
Vallumd currently breaks in buildbot.
Too old cmake declared as minimum...  Cmake 4.x doesn't accept compatibility claim with below 3.5 in CMakelists.txt upstream. 
Change to at least 3.10 (like being done to core OpenWrt)

Ps. And the same naturally goes for all other cmake packages that you are involved with.

stintel:
@hnyman thanks for the heads-up but please don't necro old threads, a new issue would have been more appropriate.

PR #19935, opened 2022-11-19
net/mwan3: add support for src ipsets in mwan3 rules

kaifabian:
Maintainer: @feckert
Compile tested: - (script change only, doesn't affect compilation)
Run tested: OpenWRT 21.02.3 (r16554) on x86_64 (QEMU)

Description:
This change allows users to specify an ipset matching on the source address (rather than the existing implementation for destination addresses) in a rule definition.

This change adds another uci config option to `rule`s, called `ipset_src`, which is handled similarly to the existing `ipset` option, but rather than generating a `-m set --match-set <x> dst`, it generates a `-m set --match-set <x> src` iptables expression.


hnyman:
You have apparently tested this in 21.02, but are trying to merge this into master.
You should first test this in the current master.

Note that master and 22.03 have been dropped iptables as the backbone of the firewall, and now use nftables.

kaifabian:
Given that the whole mwan3 package still relies on `iptables` even though 22.03 has dropped iptables as its default firewall backend, I'd like to ask the current maintainer, @feckert, what his opinion regarding the future of mwan3 is. I can possibly also backport this change to 21.02 and 22.03, but IMHO it should go into master first (if accepted at all).

feckert:
As already mentioned, the mwan3 does run with the fw4 that nft uses, but this is not optimal but it works. This works because it only marks the packets. If the system firewall fw4 uses nft, then all other packages should also use nft. I'm trying to implement a mwan4 for nft anyway, but it may take a while. Therefore, from my point of view, there is nothing to be said against integrating it into the master and then backporting the whole thing. Until the next mwan3 release, I would like to replace mwan3(iptables) with mwan4(nft).

Translated with www.DeepL.com/Translator (free version)

openwrtdiy:
@kaifabian @feckert 

I have added mwan3 rules in packages and luci to support src ipsets
Now I don't know how to use ipset src

kaifabian:
@openwrtdiy this PR only implements the mwan3 part (scripts configuring the firewall). The mwan3 ui is implemented by package `luci-app-mwan3`. I will look into implementing the necessary change(s) to that particular package once this PR is out of draft state.

As a temporary workaround, you could set the necessary field via [UCI CLI](https://openwrt.org/docs/guide-user/base-system/uci):
`uci set mwan3.172_31_26_1.ipset_src=<your_ipset>`, followed by a `uci commit`, or by manually editing `/etc/config/mwan3`.

Ovear:
Hi,

Please revisit this PR that has been open for some time now. Implementing this feature would be highly beneficial, particularly as source-based-policy routes play a crucial role in MWAN3, and the integration of ipset simplifies the process significantly.

Given that this is a straightforward PR and is unlikely to introduce any issues, could we consider merging it? As of OpenWrt 23.05, MWAN3 still relies on iptables, and it may still need take more time to migrate to NFT.

Considering this situation, it seems like a good time to merge this 2-year-old PR to mainstream which I believe would benefit all MWAN3 users.

Thanks

GeorgeSapkin:
No updates in over three years and still marked as draft. Closing as stale. Feel free to reopen as necessary.

PR #19936, opened 2022-11-19
haproxy: use .crt as extension for certificates

hgl:

Maintainer: @tohojo 

Description:


hgl:
This will break renewed combined certificates. Do you think we should detect `.cer` certificates and rename them for renewed certificate in the acme package? 

Given that the official openwrt hasn't used the new acme package, and only haproxy combined certificates might break, the number of affected certificates might not be very large.

tohojo:
Hmm, how about detecting if there's a .cer, and if there is, replace it with a symlink to the .crt?

hgl:
This PR should also be good to merge.

tohojo:
Needs a package revision bump, probably?

hgl:
Done.

tohojo (approved):

PR #19938, opened 2022-11-19
acme: ensure symlinks for renewed certificates

hgl:

Maintainer: @tohojo

Description:
This should make #19936 backward-compatible.

### DIFF
--- a/net/acme-acmesh/files/hook.sh
+++ b/net/acme-acmesh/files/hook.sh
@@ -45,6 +45,20 @@ get)
 
 			case $status in
 			0)
+				mkdir -p /etc/ssl/acme
+				if [ ! -e "/etc/ssl/acme/$main_domain.crt" ]; then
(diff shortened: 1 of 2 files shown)


net/acme-acmesh/files/hook.sh
```diff
@@ -45,6 +45,20 @@ get)
 
 			case $status in
 			0)
+				mkdir -p /etc/ssl/acme
+				if [ ! -e "/etc/ssl/acme/$main_domain.crt" ]; then
+					ln -s "$domain_dir/$main_domain.cer" "/etc/ssl/acme/$main_domain.crt"
+				fi
+				if [ ! -e "/etc/ssl/acme/$main_domain.key" ]; then
+					ln -s "$domain_dir/$main_domain.key" /etc/ssl/acme
```
tohojo, line 53: Let's add the filename to the target path here as well, so it's consistent with the other `ln` invocations.

hgl: Done.

tohojo (approved):

PR #19960, opened 2022-11-22
[RFC] librespeed-go: add new package

1715173329:
Maintainer: me
Run tested: nanopi-r2s

Description:
Go backend for LibreSpeed.
https://github.com/librespeed/speedtest-go

### DIFF
--- a/net/librespeed-go/Makefile
+++ b/net/librespeed-go/Makefile
@@ -1,0 +1,78 @@
+# SPDX-License-Identifier: GPL-3.0-only
+#
+# Copyright (C) ImmortalWrt.org
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=librespeed-go
+PKG_VERSION:=1.1.5
+PKG_RELEASE:=1
+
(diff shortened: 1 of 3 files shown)


net/librespeed-go/Makefile
```diff
@@ -0,0 +1,78 @@
...
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/librespeed/speedtest
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+TAR_OPTIONS+= --strip-components 1
+TAR_CMD=$(HOST_TAR) -C $(1) $(TAR_OPTIONS)
```
BKPepe, line 29: Why this is required? It shouldn't be, right?

1715173329: The repo name is `speedtest-go`, so does the folder. I don't wanna override `PKG_BUILD_DIR` as [there's a project](https://github.com/showwin/speedtest-go) uses the same name and it can cause confusing and conflict.

So add a tar option to have it in the "correct" bulid dir.

PR #19962, opened 2022-11-23
unbound: fix local_subnet for IPv6 addresses that contain a port

markmentovai:
Maintainer: @EricLuehrsen
Compile tested: x86-64/apu2, OpenWrt https://github.com/openwrt/openwrt/commit/eaba63cddee44e0761cee1b0f82e3dd359e2dac1, feeds/packages 971d8ac3e565
Run tested: x86-64/apu2, OpenWrt https://github.com/openwrt/openwrt/commit/eaba63cddee44e0761cee1b0f82e3dd359e2dac1, feeds/packages 971d8ac3e565

Description:

This prevents a forwarding server named like `::1@5453` from being added to unbound.conf as a `forward-host` instead of the correct `forward-addr`. `forward-host` requires the name to be resolved, which is impossible in the absence of another nameserver. Thus, forwarding-only configurations referencing only the IPv6 loopback address with a port number were broken.

Signed-off-by: Mark Mentovai \<mark@mentovai.com>

Additional information:

For a forwarding configuration in `/etc/config/unbound` like this:

```
config zone 'fwd'
        option enabled '1'
        option fallback '0'
        option zone_type 'forward_zone'
        option dns_assist 'unprotected-loop'
        list server '127.0.0.1@5453'
        list server '::1@5453'
        list zone_name '.'
```

The forwarding servers listed in the generated `/var/lib/unbound/unbound.conf` wound up like this:

```
forward-zone:
  name: .
  forward-host: ::1@5453
  forward-addr: 127.0.0.1@5453
  forward-first: no
  forward-tls-upstream: no
```

instead of the correct:

```
forward-zone:
  name: .
  forward-addr: 127.0.0.1@5453
  forward-addr: ::1@5453
  forward-first: no
  forward-tls-upstream: no
```


1715173329:
Ping @EricLuehrsen

EricLuehrsen (approved):
sorry been busy with other things. good to merge.

PR #19966, opened 2022-11-24
samba4: support both user map and valid users on same time

jjm2473:
Compile tested: arm64, Easepi ARS2, openwrt 21.02.3
Run tested: arm64, Easepi ARS2, openwrt 21.02.3

Description:
support both user map and valid users on same time.
`valid users` is for validate logining user, `user map` is map user to root after login, there is not conflicts.


1715173329:
Please add your description into commit message too.

1715173329:
ping @jjm2473

jjm2473:
@1715173329 It's a simple bug fix, just go ahead

PR #19974, opened 2022-11-27
add an iptables extension module to obfuscate WireGuard

infinet:
Maintainer: @infinet
Compile tested: (ramips-mt7620 19.07.9 and 22.03.2, x86_64 19.07.9)
Run tested: (x86_64 19.07.9)

Port from https://github.com/infinet/xt_wgobfs, this extension obfuscates WireGuard by:

- The first 16 bytes of WG message is obfuscated.

- The mac2 field is also obfuscated, if it is all zeros.

- Padding WG message with random long random bytes.

- Drop keepalive message with 80% probability.

- Change the Diffserv field to zero.

See https://github.com/infinet/xt_wgobfs/blob/main/README.md for usage.

Example as client:

Add two iptables rules to firewall.user:

iptables -t mangle -I INPUT -p udp -m udp --sport 6789 -j WGOBFS \
    --key mysecretkey --unobfs

iptables -t mangle -I OUTPUT -p udp -m udp --dport 6789 -j WGOBFS \
    --key mysecretkey --obfs

Here assuming remote server is listening on port 6789.

Signed-off-by: Wei Chen <weichen302@gmail.com>


infinet:
I changed the tile. Could you please re-run the Formalities test? Thanks!

Ansuel:
You need to change commit title also

jefferyto:
Not sure why but only 1 of the usual 10+ checks ran, I suggest rebasing on current master and doing a force push.

net/xtables-wgobfs/Makefile
```diff
@@ -0,0 +1,89 @@
...
+PKG_NAME:=xtables-wgobfs
+PKG_VERSION:=0.4.2
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/xt_wgobfs-$(PKG_VERSION)
+PKG_SOURCE:=xt_wgobfs-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:= https://github.com/infinet/xt_wgobfs/releases/download/v$(PKG_VERSION)/
+PKG_HASH:=09fa493d8305e1fa3224a940cab607b1860a9b5d9d395615105c7009e2bec767
+
+PKG_LICENSE:=GPL-2.0
```
jefferyto, line 17: Do you intend your software to be GPL v2 or later (`GPL-2.0-or-later`) or v2 only (`GPL-2.0-only`)?

net/xtables-wgobfs/Makefile
```diff
@@ -0,0 +1,89 @@
...
+include $(INCLUDE_DIR)/kernel.mk
+include $(INCLUDE_DIR)/package.mk
+
+define Package/xtables-wgobfs
+	SECTION:=net
+	CATEGORY:=Network
+	SUBMENU:=Firewall
+	TITLE:=iptables WireGuard obfuscation extension
+	URL:=https://github.com/infinet/xt_wgobfs
+	MAINTAINER:=Wei Chen <weichen302@gmail.com>
```
jefferyto, line 34: No need for `MAINTAINER` here, `PKG_MAINTAINER` above is enough.

net/xtables-wgobfs/Makefile
```diff
@@ -0,0 +1,89 @@
...
+	An iptables extension for WireGuard obfuscation
+endef
+
+# uses GNU configure
+CONFIGURE_ARGS+= \
+	--with-kbuild="$(LINUX_DIR)" \
+	--with-xtlibdir="/usr/lib/iptables" \
+
+define Build/Compile
+	+$(MAKE) $(PKG_JOBS) -C $(PKG_BUILD_DIR) \
```
jefferyto, line 48: I think you can add the values you need to `MAKE_FLAGS`, then you wouldn't need to override the default `Build/Compile` and `Build/Install`. (You can see the default values in [package-defaults.mk](https://github.com/openwrt/openwrt/blob/c35ff41ccc31f2ecaea7406830de04bd77940d48/include/package-defaults.mk).)

net/xtables-wgobfs/Makefile
```diff
@@ -0,0 +1,89 @@
...
+	$(MAKE) -C $(PKG_BUILD_DIR) \
+		ARCH="$(LINUX_KARCH)" \
+		CROSS_COMPILE="$(TARGET_CROSS)" \
+		DESTDIR="$(PKG_INSTALL_DIR)" \
+		DEPMOD="/bin/true" \
+		install
+endef
+
+define Package/iptables-mod-wgobfs
+	$(call Package/xtables-wgobfs)
```
jefferyto, line 67: Since `Package/xtables-wgobfs` is only used once (here), you can set the values here directly.

net/xtables-wgobfs/Makefile
infinet: I am not sure the differences. However, since Jason A. Donenfeld recently confirmed wireguard license is GPL-2.0, it is seems reasonable to use the same license for a kernel module works with WG.

net/xtables-wgobfs/Makefile
infinet: Thank you for the review! I have made some changes accordingly.

net/xtables-wgobfs/Makefile
jefferyto: > I am not sure the differences.

GPL-2.0-or-later software allows users to apply the terms of GPL 2.0 or 3.0 or any later version to how they use the software. GPL-2.0-only software allows users to apply the terms of GPL 2.0 only. The Linux kernel is GPL-2.0-only.

> However, since Jason A. Donenfeld recently confirmed wireguard license is GPL-2.0, it is seems reasonable to use the same license for a kernel module works with WG.

I'm not sure what the license for WireGuard is exactly, but `GPL-2.0` is the [deprecated](https://spdx.org/licenses/GPL-2.0.html) identifier for `GPL-2.0-only`.

infinet: Thank you for the explanation! I have update it to `GPL-2.0-only`.

valga:
@infinet It seems that your ChaCha8 hash function implementation has some crypto related flaws compared to reference ChaCha20 one:

1. Although your macro called `EIGHT_ROUNDS` it runs only 4 rounds, so your hash is actually ChaCha4 - https://github.com/infinet/xt_wgobfs/blob/440a2ff505fec95a393c4a316ae2811c60ae0948/src/chacha8.c#L76-L79 - you have to call `DOUBLE_ROUND` 4 times to get 8 rounds, see reference https://git.zx2c4.com/wireguard-linux-compat/tree/src/crypto/zinc/chacha20/chacha20.c#n71
2. You use ChaCha state part related to key in return value, whereas ChaCha20 skips it to nonce one - https://git.zx2c4.com/wireguard-linux-compat/tree/src/crypto/zinc/chacha20/chacha20.c#n150
3. ChaCha state after rounds gets ORed with the initial state - https://github.com/infinet/xt_wgobfs/blob/440a2ff505fec95a393c4a316ae2811c60ae0948/src/chacha8.c#L101 - considering previous point it gets ORed with 4 well known constants and first half of the key. ChaCha20 hash doesn`t do OR before return at all.

infinet:
> 

Thank you for the insight! It will take some time to review and correct that part of code. In general, this module is trying to remove wireguard packets signature with minimal overhead, while wireguard is taking care of the communication security.

infinet:
I've taken another look at chacha8 implementation.

1. You're correct; the chacha8 indeed only does 4 rounds. Need to test the performance of 8 rounds. If the overhead is similar to current 4 rounds, then revise the code to 8 rounds.

2. Wireguard uses XChaCha variant. It uses 128 bits nonce and original 256 bits key to generate a derived key, which is used by chacha20 with 96 bits nonce to encrypt message. The derived key generation takes first 32 bits and last 32 bits of multiple quarter rounds output, as you mentioned in remark 2.

3. The keystream used by wireguard to encrypt message is generated by ChaCha20 block function. Its output is different from the key derive function. It adds the original state to the quarter rounds output. This module use the same approach.
https://git.zx2c4.com/wireguard-linux-compat/tree/src/crypto/zinc/chacha20/chacha20.c#n84

valga:
@infinet I also reviewed everything again and found this useful piece of information - https://crypto.stackexchange.com/a/11215

In short, original state addition (feed-forward) is required when you take parts of result that is being added with key (64 bits in the middle). In other words, you can omit feed-forward if you take first and last 32 bits, because they are being added with known input, so it will improve performance a bit.

So the only concern is number of rounds.

infinet:
> @infinet I also reviewed everything again and found this useful piece of information - https://crypto.stackexchange.com/a/11215
> 
> So the only concern is number of rounds.

I have made few modifications.
    
  - The hash function is now chacha6 instead of chacha4.
  - Input to chacha is 16 bytes instead of 8 bytes.
  - All inputs for chacha are from the 16th to 31st bytes of WG message. First byte is used as counter and incremented each time to generate a new PRN.

The performance is almost the same.

infinet:
I have rebased it on the latest master branch.

GeorgeSapkin:
@infinet still relevant? If so commits need to be formatted according to the [submission guidelines](https://openwrt.org/submitting-patches#submission_guidelines). And since this is adding a new package, this can be just one commit.

infinet:
Yes, it’s still relevant. I have updated the PR and squashed them into a single commit. Thanks!

GeorgeSapkin:
I think the commit subject should be something like:

```
xtables-wgobfs: add new package
```

And the rest go into the commit message.

GeorgeSapkin:
You need to address the formality check failures.

net/xtables-wgobfs/Makefile
```diff
@@ -0,0 +1,74 @@
...
+PKG_VERSION:=0.6.2
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/xt_wgobfs-$(PKG_VERSION)
+PKG_SOURCE:=xt_wgobfs-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:= https://github.com/infinet/xt_wgobfs/releases/download/v$(PKG_VERSION)/$(PKG_SOURCE)
+PKG_HASH:=ba4c410c9dc304360d944249d5314ef4987515de381b1274873f8597928cb67f
+
+PKG_LICENSE:=GPL-2.0-only
+PKG_LICENSE_FILES:=COPYING
```
GeorgeSapkin: Doesn't look there's COPYING in that repo. Did you mean LICENSE?

net/xtables-wgobfs/Makefile
```diff
@@ -0,0 +1,74 @@
+#
+# Copyright (C) 2022-2025 Wei Chen <weichen302@gmail.com>
+#
```
GeorgeSapkin, line 3: Consider adding a SPDX license identifier so it's clear what the file's license is.

infinet:
Thanks! I have made two changes. Do they look reasonable?

net/xtables-wgobfs/Makefile
```diff
@@ -0,0 +1,76 @@
...
+# This is free software, licensed under the GNU General Public License v2.
```
GeorgeSapkin: This is not a proper SPDX license identifier. A proper one based on your license would be:

```suggestion
# SPDX-License-Identifier: GPL-2.0-only
```

infinet:
Hopefully I have fixed all the formalities error in CI checks.

PR #19979, opened 2022-11-27
tvheadend: test "Optimize for speed"

M95D:
DO NOT COMMIT!

There is a bug report that "Optimize for speed" doesn't work on some platforms. This is a test to see which platforms fail.

Signed-off-by: Marius Dinu <m95d+git@psihoexpert.ro>


Ansuel:
(still this couldn't be tested on your local repo? instead of creating a pr? but whatever)

M95D:
I don't know how to automate it locally. I don't even know where to look for a list of all platforms.
I asked this question in the [forum](https://forum.openwrt.org/t/build-one-package-for-all-targets/112732), but nobody answered.

PR #19982, opened 2022-11-28
ddns-scripts: update_gandi_net: improve logging & add timeout

golddranks:
Maintainer: @feckert
Compile tested: (dynamic script, so no compilation)
Run tested:
 - arch: ath79/generic
 - model: TP-Link Archer C7 v2
 - version: OpenWrt 22.03.2 r19803-9a599fee93 / LuCI openwrt-22.03 branch git-22.288.45147-96ec0cd
 - tests done: Replaced the script file, `rm /var/run/ddns/* /var/log/ddns/* && service ddns restart`, checked that the logs look like expected. Tried triggering an update when rate-limited to test the added timeout.

Description:
- Some gandi.net updates seemed flaky, but it was hard to know from the logs what went wrong, so I added to the logs
  - the executed curl command to be able to test it manually
  - the curl exit status
- Turns out sometimes the request would just hang, and eventually curl would return with exit status `35`. I think that it's better to clearly be able to distinguish flaky connections, so I added 30 second timeout, which seems reasonable to me.

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
-PKG_RELEASE:=29
+PKG_RELEASE:=30
 
 PKG_LICENSE:=GPL-2.0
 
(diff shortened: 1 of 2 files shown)


golddranks:
@feckert Thanks for the pointer! Added!

1715173329:
could you please shorten your commit title and move the remaing part into commit message?

golddranks:
@1715173329 Done :)

PR #19989, opened 2022-11-29
snort3: unified configs: local.lua and homenet.lua

graysky2:
This commit adds `/etc/snort/local.lua` and `/etc/snort/homenet.lua` for user defined config options which is more simplistic than modifying upstream files directly. That can be tedious and decisive to maintain in sync with upstream changes.  The init script has been adjusted accordingly.

Acknowledgment to amish who maintains the Arch Linux [snort-nfqueue package](https://aur.archlinux.org/packages/snort-nfqueue) for these ideas and initial code.

Another modification is dropping the following args in the call to `/usr/bin/snort` by the init system as these options are provided in `/etc/snort/local.lua`:

 * --daq-dir /usr/lib/daq/
 * -A "$alert_module"

Instructions to configure snort3:
```
1. Edit /etc/snort/homenet.lua and redefine HOME_NET and EXTERNAL_NET, for example:    
       HOME_NET = [[ 10.9.8.0/24 192.168.1.0/24 ]]
       EXTERNAL_NET = "!$HOME_NET"
    
2. Edit /etc/snort/local.lua to setup options unique to your use case of snort.
   The default ones I included should be sane for the role of IDS (alert only), 
   but users may easily uncomment some options therein to use IPS (drop) mode.
    
3. Install or symlink rules to /etc/snort/rules/snort.rules edit optionally edit
   /etc/snort/local.lua to define extra rules files if not using unified 'snort.rules'
```

### DIFF
--- a/net/snort3/Makefile
+++ b/net/snort3/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=snort3
 PKG_VERSION:=3.1.48.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://github.com/snort3/snort3/archive/refs/tags/
@@ -90,7 +90,8 @@ define Package/snort3/install
 		$(PKG_INSTALL_DIR)/usr/include/snort/lua/snort_plugin.lua \
 		$(1)/usr/share/lua/
 
-	$(INSTALL_DIR) $(1)/etc/snort
+	$(INSTALL_DIR) $(1)/etc/snort/{rules,lists,builtin_rules,so_rules}
+
 	$(INSTALL_CONF) \
 		$(PKG_INSTALL_DIR)/usr/etc/snort/*.lua \
 		$(1)/etc/snort
@@ -107,6 +108,25 @@ define Package/snort3/install
 	$(INSTALL_CONF) \
 		./files/snort.config \
 		$(1)/etc/config/snort
+	$(INSTALL_CONF) \
+		./files/local.lua \
+		$(1)/etc/snort
+	$(INSTALL_CONF) \
+		./files/homenet.lua \
+		$(1)/etc/snort
+	sed \
+		-i -e "/^EXTERNAL_NET\\s\\+=/ a include 'homenet.lua'" \
+		-e "/^HOME_NET\\s\\+=/ i -- we set HOME_NET and EXTERNAL_NET here or via an included file" \
+		-e 's/^\(HOME_NET\s\+=\)/--\1/g' \
+		-e 's/^\(EXTERNAL_NET\s\+=\)/--\1/g' \
+		$(1)/etc/snort/snort.lua
+	sed \
+		-i -e "s/^\\(RULE_PATH\\s\\+=\\).*/\\1 'rules'/g" \
+		-e "s/^\\(BUILTIN_RULE_PATH\\s\\+=\\).*/\\1 'builtin_rules'/g" \
+		-e "s/^\\(PLUGIN_RULE_PATH\\s\\+=\\).*/\\1 'so_rules'/g" \
+		-e "s/^\\(WHITE_LIST_PATH\\s\\+=\\).*/\\1 'lists'/g" \
+		-e "s/^\\(BLACK_LIST_PATH\\s\\+=\\).*/\\1 'lists'/g" \
+		$(1)/etc/snort/snort_defaults.lua
 endef
 
 $(eval $(call BuildPackage,snort3))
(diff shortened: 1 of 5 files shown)


flyn-org:
Looks fine: the use of % in Bourne shell variable substitution will remove a trailing '/'. Recommend merge.

1715173329:
not sure if somebody likes using `/` as the confdir... (xD
afaict `/etc/snort//snort.lua` equals to `/etc/snort/snort.lua` in linux?

graysky2:
That is true.... about a user using `/` is the confdir.  I do not know about the equivalence of `//` to `/` but it does seem to work without my PR so you might be right.

1715173329:
> I do not know about the equivalence of `//` to `/`

ref https://unix.stackexchange.com/questions/1910/how-does-linux-handle-multiple-consecutive-path-separators-home-username

flyn-org:
Yes, any number of adjacent slashes within a path behave the same as one slash. @graysky, your use of "cosmetic" in the pull-request title made me think you were merely fixing appearances, which you are.

graysky2:
Yes, if you guys think this PR will do more harm than good, please close it.

1715173329:
> not sure if somebody likes using `/` as the confdir... (xD

I do apologize that this doesn't really matter...
As `"${config_dir%/}/snort.lua"` has a slash already, it would be `/snort.lua`, which is the expected behavior.

I'm okay with this, or it can be included in the next version bumping / actual fixes.

graysky2:
@flyn-org @1715173329 - I modified this PR including the syntax correction but expanding it to make configuration more simplistic.  What do you think?

graysky2:
Let's hold off on merging this until I can get the need to modify `/etc/config/snort_defaults.lua` ... that is messy and I am not sure that `HOME_NET` and `EXTERNAL_NET` are getting picked-up in `local.lua` ... I am seeing tons of lines in the log file which I do not believe is normal.

graysky2:
OK, I think this is good/instructions in commit message.

1715173329:
Makefile needs a rebase.

graysky2:
> Makefile needs a rebase.

Done!

1715173329:
cc @flyn-org

flyn-org:
I like the concepts behind the latest push. I saw the following error when I tried to run a fresh install of snort3:
```
# snort -q -i eth0 -c /etc/snort/snort.lua --tweaks local
ERROR: local.lua:1 can't init /etc/snort/snort.lua: /etc/snort/local.lua:54: '}' expected (to close '{' at line 46) near '<eof>'

ERROR: local.lua:1 Could not find requested DAQ module: pcap

FATAL: see prior 2 errors (0 warnings)
Fatal Error, Quitting..
```
I fixed this by adding a closing brace to the end of local.lua. The indentation needs fixing too.

That left me with this:
```
# snort -q -i eth0 -c /etc/snort/snort.lua --tweaks local
ERROR: /etc/snort/snort.lua: invalid daq.modules.mode = 'tap'
FATAL: see prior 1 errors (0 warnings)
Fatal Error, Quitting..
```
I am presently fiddling around with what I need to change in local.lua.

graysky2:
Odd, thought I confirmed it was right.  Let me see.

graysky2:
> This is looking pretty good. I have one more question: if we use the syslog logging facility by default (I think we should), then does that mean we rely on the system log daemon to determine where the logs go? If so, then do we need this:
```output.logdir = '/var/log/snort'```

Yes, that was present in the last commit: 
https://github.com/graysky2/packages/commit/eada025d60101015970c73382c68170c96157bd8#diff-27f6d353ad520c988bb2ee5c2106f2fcd34ddcb64f220a479d480bb161a459f6R2

>and this:
```$(INSTALL_DIR) $(1)/var/log/snort```

That was also there: 
https://github.com/graysky2/packages/commit/eada025d60101015970c73382c68170c96157bd8#diff-9dc9397d7dfb45bf1e3feba330ab49533d25e7c8ca9b94c6c8e4805cb86f81eeR93

Users can either edit `local.lua` to redefine the log location if it is on external media or they can create a symlink within the live filesystem.

**Comment by flyn-org** [comment]
I think you took my comment the wrong way.

Your default `local.lua` includes this:
```
alert_syslog = {
  level = 'info',
}
```
According to the documentation, this "sends alerts to the syslog facility". From what I understand, this means that `/var/log/snort` is irrelevant in your default case. I agree that the use of syslog should be the default.

Please consider removing `$(INSTALL_DIR) $(1)/var/log/snort` and commenting out `output.logdir = '/var/log/snort'` with commentary that explains its use requires: (1) removing the comment characters from the line, (2) removing the comment characters from the `alert_fast` statement, and (3) creating the `/var/log/snort` directory.

graysky2:
@flyn-org - I misunderstood, sorry.  Please see if the latest push is to your liking.

flyn-org:
Looks good. @neheb, are you willing to merge for us?

PR #19991, opened 2022-11-29
uacme: use UCI instead of iptables for firewall configuration

pesintta:
Maintainer: @lucize 

Description: Here is an attempt to make uacme compatible with nft firewall implementation.
This has been missing as identified in https://github.com/openwrt/packages/issues/16818

### DIFF
--- a/net/uacme/files/run.sh
+++ b/net/uacme/files/run.sh
@@ -37,6 +37,7 @@ NGINX_WEBSERVER=0
 UPDATE_NGINX=0
 UPDATE_UHTTPD=0
 UPDATE_HAPROXY=0
+FW_RULE=
 USER_CLEANUP=
 
 . /lib/functions.sh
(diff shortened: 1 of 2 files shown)


lucize:
don't recall iptables being removed from master, so I think this should be a detect at run time, as in my case I still run iptables on some master devices (not necessary with uacme)

pesintta:
> don't recall iptables being removed from master, so I think this should be a detect at run time, as in my case I still run iptables on some master devices (not necessary with uacme)

Latest stable release does not ship with iptables command by default which makes uacme effectively broken on it.

You can see iptables utility removal being mentioned in [22.03.0 release notes](https://openwrt.org/releases/22.03/notes-22.03.0#firewall4_based_on_nftables).

One can of course install iptables via opkg but even then the commands run by uacme are broken as there no longer exists a chain called input_rule.

I figured it would be better to adopt to using netfilter natively as the nft command is built-in.

lucize:
still think that would be best to include both usages and use whatever utility is installed, some use case is when you self update the master from when iptables was the only option and because of legacy you still use that. So from mpov until there is iptables available in master then we still need to keep that option

pesintta:
I've updated the PR with another approach where we utilize UCI to add a temporary rule to OpenWrt's firewall. This should be compatible with both iptables (fw3) in older versions and netfilter (fw4) in later versions.

@lucize what do you think?

lucize:
I very like it now!
LGTM

lucize:
But first, please see about AUTORELEASE, it's deprecated

pesintta:
> But first, please see about AUTORELEASE, it's deprecated

I added a commit removing it to solve the deprecation warning.

net/uacme/files/run.sh
```diff
@@ -135,24 +136,30 @@ pre_checks()
...
-    fi
+    FW_RULE=$(uci add firewall rule) || return 1
+    uci set firewall."$FW_RULE".name='uacme: temporarily allow incoming http'
+    uci set firewall."$FW_RULE".enabled='1'
+    uci set firewall."$FW_RULE".target='ACCEPT'
+    uci set firewall."$FW_RULE".src='wan'
+    uci set firewall."$FW_RULE".proto='tcp'
+    uci set firewall."$FW_RULE".dest_port='80'
+    uci commit firewall
+    /etc/init.d/firewall reload
```
champtar, line 147: You can open firewall ports when you start a procd service, here an example: https://github.com/openwrt/packages/blob/96b2cd70e108d3ca4b9327881b00f1300c7f8cd0/net/nebula/files/nebula.init#L79-L91

PR #19997, opened 2022-11-30
libqmi: fix build regressions

feckert:
Maintainer: @BKPepe @nemesisdesign @aleksander0m 
Compile tested: x86_64, latest owrt-21.02
Run tested: no

Description:
Fix compilation for libqmi. Since merging of https://github.com/openwrt/packages/pull/19648 the libqmi does not build anymore. I got the following error.
`
SHELL= flock /home/feckert/workspace/openwrt/LDM-owrt-21.02/build/openwrt/tmp/.libqmi-1.30.8.tar.xz.flock -c '          /home/feckert/workspace/openwrt/LDM-owrt-21.02/build/openwrt/scripts/download.pl "/home/feckert/dl" "libqmi-1.30.8.tar.xz" "862482ce9e3ad0bd65d264334ee311cdb94b9df2863b5b7136309b41b8ac1990" "" "https://www.freedesktop.org/software/libqmi"
`
I already updated the HASH but then I got another error that `No rule to make target` !
Only the backport of the latest master changes fixes my issues.
If I am backporting the latest master changes for owrt-21.02 my build works again.

### DIFF
--- a/libs/libqmi/Config.in
+++ b/libs/libqmi/Config.in
@@ -13,4 +13,19 @@ config LIBQMI_WITH_QRTR_GLIB
 	help
 	  Compile libqmi with QRTR support
 
+choice
+	prompt "Select QMI message collection to build"
+	default LIBQMI_COLLECTION_BASIC
+
+	config LIBQMI_COLLECTION_MINIMAL
+		depends on !MODEMMANAGER_WITH_QMI
+		bool "minimal"
+
+	config LIBQMI_COLLECTION_BASIC
+		bool "basic (default)"
+
+	config LIBQMI_COLLECTION_FULL
+		bool "full"
+endchoice
+
 endmenu
--- a/libs/libqmi/Makefile
+++ b/libs/libqmi/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libqmi
-PKG_SOURCE_VERSION:=1.30.6
+PKG_SOURCE_VERSION:=1.30.8
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://gitlab.freedesktop.org/mobile-broadband/libqmi.git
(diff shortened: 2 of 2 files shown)


BKPepe:
I don't think that we need to have the first commit, which is revert, right? We can leave it as it is and then have only two changes, which are required, and that's switch to meson and polish the Makefile. Also, when doing a revert, please describe the reason, why you are reverting it.

1715173329:
The reason is already inside commit message, isn't it ;)
tbh this is the right way to backport stuffs: cherry-pick upstream commits instead of creating your own one.

BKPepe:
Nope. It isnt. The revert commit does  not say anything, why it is being
reverted.

Dne čt 1. 12. 2022 20:44 uživatel Tianling Shen ***@***.***>
napsal:

feckert:
From my point of view, I would say that the way I did it was right.

BKPepe:
> Revert the last libqmi bump to backport the latest master changes.

Sorry, but what's the meaning of that? If I am going to ask you one month later or you will check the Git history in the future, it will not say anything. 
In one PR, you are reverting the version, which should be there, and at the same time, you are bringing it back and the only reason what you said is: backport the latest master changes. :shrug: 

The correct reasoning for the revert commit is somehow described in OP, actually, what should be there is that while cherry-picking there was missed one commit and because of that, we are going to introduce 5 commits instead of 2? Guys, c'mon. :) We can do it better. 

There is nothing wrong with cherry-picking the switch to meson, solve the merge conflict, add it to the commit description in square brackets and that's it. 

So, the reason why you are doing is that you were not able to solve the merge conflict and that the version 1.30.8, which is in the stable branch requires meson. That's why I was asking it again...

feckert:
@BKPepe fair enough
I agree with you about the history! I've had that happen to me too. Good advice.
That's why I have now adjusted the message and made a push force.
I hope it fits now, if not then let me know.

BKPepe:
It's better, so let's merge it. I would prefer to have only 2 commits for this issue, but this one works, too.

PR #19998, opened 2022-11-30
natmap: add new package

ysc3839:
Maintainer: me
Compile tested: ipq4019 22.03.2
Run tested: ipq4019 22.03.2

Description:
NATMap is a program for opening port behind full cone NAT (NAT-1), without the need for using UPnP or another port forward settings.
More details can be found at original repo: https://github.com/heiher/natmap

~This PR is working in progress.~ I have some questions:
1. Can I use procd `capabilities` without the whole ujail? Just like systemd's `AmbientCapabilities=CAP_NET_BIND_SERVICE`. To allow non-root user bind port less than 1024. It looks like ujail have to manually setup file mounts. https://github.com/openwrt/openwrt/blob/366bcffa0e2101362a1856591dad2e71ebb0f3a7/package/utils/busybox/files/sysntpd#L73-L84
2. Is there any place to store runtime data which can be read by luci? I'm making a luci app shows current IP and port.


ysc3839:
@heiher natmap build failed on latest OpenWrt mips, do you know what's wrong with it?
https://github.com/openwrt/packages/actions/runs/3587027495/jobs/6036905309#step:5:3677

It works well using OpenWrt 22.03.2 mt7621 sdk.

heiher:
> @heiher natmap build failed on latest OpenWrt mips, do you know what's wrong with it? https://github.com/openwrt/packages/actions/runs/3587027495/jobs/6036905309#step:5:3677
> 
> It works well using OpenWrt 22.03.2 mt7621 sdk.

Looks like it's not compatible with mips16.

```diff
diff --git a/net/natmap/Makefile b/net/natmap/Makefile
index b6068e108..1f3b66be0 100644
--- a/net/natmap/Makefile
+++ b/net/natmap/Makefile
@@ -14,6 +14,8 @@ PKG_MAINTAINER:=Richard Yu <yurichard3839@gmail.com>
 PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=License
 
+PKG_USE_MIPS16:=0
+
 include $(INCLUDE_DIR)/package.mk
 
 define Package/$(PKG_NAME)
```

ysc3839:
@heiher Issue solved, thanks!

heiher:
Hmmm. I will add support for powerpc and arc architectures.

net/natmap/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+PKG_NAME:=natmap
+PKG_VERSION:=20221111
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/heiher/natmap.git
+PKG_SOURCE_DATE:=2022-11-11
+PKG_SOURCE_VERSION:=$(PKG_VERSION)
+PKG_MIRROR_HASH:=4359bdddd16d7f97af4ee03f814ea4e649554c30eeed93e5f4fc586151d2476e
```
1715173329, line 11: is it possible to download source code via git codeload?

net/natmap/Makefile
```diff
@@ -0,0 +1,38 @@
...
+PKG_USE_MIPS16:=0
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/$(PKG_NAME)
+	SECTION:=net
+	CATEGORY:=Network
+	TITLE:=TCP/UDP port mapping for full cone NAT, any host can access internal services of private address broadband
+	MAINTAINER:=heiher
+	URL:=https://github.com/heiher/natmap
```
1715173329, line 26:
```suggestion
define Package/natmap
  SECTION:=net
  CATEGORY:=Network
  TITLE:=TCP/UDP port mapping for full cone NAT, any host can access internal services of private address broadband
  URL:=https://github.com/heiher/natmap
```

- Do not use $(PKG_NAME), as it reduces readability.
- Use two `<space>` instead of `<tab>`.
- `MAINTAINER` can be removed, this is actually meant `PKG_MAINTAINER`.

net/natmap/Makefile
```diff
@@ -0,0 +1,38 @@
...
+endef
+
+define Package/$(PKG_NAME)/install
```
1715173329, line 29:
```suggestion
define Package/natmap/install
```

same above

net/natmap/Makefile
```diff
@@ -0,0 +1,38 @@
...
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/bin/natmap $(1)/usr/bin/
+	$(INSTALL_DIR) $(1)/etc/config/
+	$(INSTALL_CONF) ./files/natmap.config $(1)/etc/config/natmap
+	$(INSTALL_DIR) $(1)/etc/init.d/
+	$(INSTALL_BIN) ./files/natmap.init $(1)/etc/init.d/natmap
+endef
+
+$(eval $(call BuildPackage,$(PKG_NAME)))
```
1715173329, line 38:
```suggestion
$(eval $(call BuildPackage,natmap))
```

same above

net/natmap/Makefile
```diff
@@ -0,0 +1,38 @@
...
+PKG_SOURCE_URL:=https://github.com/heiher/natmap.git
+PKG_SOURCE_DATE:=2022-11-11
+PKG_SOURCE_VERSION:=$(PKG_VERSION)
+PKG_MIRROR_HASH:=4359bdddd16d7f97af4ee03f814ea4e649554c30eeed93e5f4fc586151d2476e
+
+PKG_MAINTAINER:=Richard Yu <yurichard3839@gmail.com>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=License
+
+PKG_USE_MIPS16:=0
```
1715173329, line 17: if this program can be compiled parallelly, please consider adding `PKG_BUILD_PARALLEL:=1`

https://github.com/heiher/natmap/blob/0ff5125dd04386d0abed2d8e0d2fa6875e21fbd4/build.mk#L10
given that `.git` will not be packed into tarball anyway, you may wanna manually pass `COMMIT_ID` cflag.

1715173329 (changes requested):

net/natmap/Makefile
1715173329: oh i see that, thanks for clarification.

ysc3839:
@1715173329 Fixed, please review.

net/natmap/Makefile
```diff
@@ -0,0 +1,40 @@
...
+include $(INCLUDE_DIR)/package.mk
+
+define Package/natmap
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=TCP/UDP port mapping tool for full cone NAT
+  URL:=https://github.com/heiher/natmap
+endef
+
+TARGET_CFLAGS+= -DCOMMIT_ID=\"$(PKG_VERSION)\"
```
1715173329, line 29:
```suggestion
TARGET_CFLAGS+= -DCOMMIT_ID="$(PKG_VERSION)"
```

hummm

ysc3839: Here should be `-DCOMMIT_ID=\\\"$(PKG_VERSION)\\\"`

heiher:
> Hmmm. I will add support for powerpc and arc architectures.

done: https://github.com/heiher/natmap/commit/dc9560df0d53c1333a987e7a26311e2e447478e8

ysc3839:
@1715173329 Thanks!
The "any place" I said is somewhere in memory. I guess it's able to save data in procd service instance, and can be read using ubus. After digging luci code I found there's no such place.

1715173329:
> I guess it's able to save data in procd service instance, and can be read using ubus.

Sadly procd does not support it ;(

ysc3839:
I think this PR is ready.

1715173329:
It would be nice if @heiher can create a new tag for arc/ppc fixes, or consider disabling build for them currently.

ysc3839:
Updated.

1715173329 (approved):

net/natmap/Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+PKG_NAME:=natmap
+PKG_VERSION:=20221202
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/heiher/natmap.git
+PKG_SOURCE_DATE:=2022-11-11
+PKG_SOURCE_VERSION:=$(PKG_VERSION)
+PKG_MIRROR_HASH:=c41ee4a4525c59103e66ee96afac4a636728d714c3f6af4a25f38a71eff71bd9
```
BKPepe, line 11: I wonder, why you are not download tarballs from tagged releases or using codeload at all?

BKPepe (changes requested):
If possible would be nice to add ``test.sh`` and copy&paste your PR description to the commit description, so it is clear why this package is being added.

net/natmap/Makefile
1715173329: It would be nice if you can take a look at my previous review so you can get the answer.

BKPepe: Hmm, I see, submodule, thanks for pointing it out! However, I think we can drop PKG_VERSION and rely on PKG_SOURCE_VERSION. Also, no need to have PKG_SOURCE_DATE, I will do suggestion soon.

1715173329: Good idea.

net/natmap/patches/100-remove-commit-id.patch
```diff
@@ -0,0 +1,15 @@
+--- a/build.mk
```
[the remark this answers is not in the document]
ysc3839: But there's a lot of patches in this repo without descriptions. I think it's obvious what this patch do.

net/natmap/Makefile
ysc3839: What's the suggestion?

net/natmap/patches/100-remove-commit-id.patch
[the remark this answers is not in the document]
ysc3839: What do you mean about `test.sh`?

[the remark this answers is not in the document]
BKPepe: I know it might be obvious for you, and for me, but not for others. The patch should have a description. It is pity that other patches do not have it, but they should have it. It's wrong, but we should not continue with the bad approach.

About test.sh, you might want to take a look at this repository, how others are doing that. Basically, it ensures you that your package is run-tested inside the GitHub Actions. The ``test.sh`` prevents that it does not segfault while checking for help, version, etc.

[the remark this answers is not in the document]
ysc3839: I renamed the patch file to `100-dont-get-commit-id-from-git.patch`. What do you think?

net/natmap/Makefile
BKPepe: Yup, that's a good idea - proposed the same https://github.com/heiher/natmap/issues/6

net/natmap/patches/100-remove-commit-id.patch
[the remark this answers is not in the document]
ysc3839: Added `test.sh`.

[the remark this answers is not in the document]
BKPepe: Naming is good, but still it is missing the formal stuff - author, subject, description, etc.

net/natmap/Makefile
1715173329: imo this is the standard format and good enough, or you can drop `PKG_VERSION` and `PKG_SOURCE_DATE` so it saves 2 lines, several bytes hhh

net/natmap/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+"$1" | grep "$2"
```
1715173329, line 3: `"$1" 2>&1 | grep "$2"`

net/natmap/Makefile
BKPepe, line 11:
```suggestion
PKG_NAME:=natmap
PKG_RELEASE:=1

PKG_SOURCE_PROTO:=git
PKG_SOURCE_URL:=https://github.com/heiher/natmap
PKG_SOURCE_VERSION:=20221202
PKG_MIRROR_HASH:=c41ee4a4525c59103e66ee96afac4a636728d714c3f6af4a25f38a71eff71bd9
```

Also, you need to fix line 29.

BKPepe: https://github.com/openwrt/packages/pull/19998#pullrequestreview-1202914938

net/natmap/patches/100-remove-commit-id.patch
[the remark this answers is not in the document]
ysc3839: OK, updated.

net/natmap/test.sh
ysc3839: FIxed.

net/natmap/patches/100-remove-commit-id.patch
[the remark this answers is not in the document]
1715173329: This is a trivial diff, having a proper name is enough. From my side it's not worthy to waste time / bytes for that "normal stuff".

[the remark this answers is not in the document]
BKPepe: Well, the description should contain that it fixes this:

```
make[3]: *** [Makefile:74: build/hev-stun.o] Error 1
<command-line>: error: "COMMIT_ID" redefined [-Werror]
<command-line>: note: this is the location of the previous definition
cc1: all warnings being treated as errors
make[3]: *** [Makefile:74: build/hev-sock.o] Error 1
```

Otherwise, package is not compiled.

net/natmap/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+"$1" 2>&1 >/dev/null | grep "$2"
```
1715173329, line 3: `>/dev/null` should be dropped, or stdout will get nothing

net/natmap/patches/100-don-t-get-commit-id-from-git.patch
```diff
@@ -0,0 +1,30 @@
+From ac0492d4d480948a3cd2790de3f9eeaccb0527ff Mon Sep 17 00:00:00 2001
```
1715173329, line 1: run `make package/natmap/refresh` to refresh your patch.

net/natmap/patches/100-remove-commit-id.patch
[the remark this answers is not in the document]
1715173329: It's redefined because of line 29.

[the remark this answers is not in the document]
BKPepe: Yes, it is, but that's not IMHO correct. Well, I would say, that this is not good at all. I'm sure that there is a better solution than providing patch, which is ugly hack and because of that, we have errors... :shrug:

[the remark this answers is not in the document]
1715173329: There's no error. Simple things are getting complex here.

net/natmap/test.sh
ysc3839: FIxed.

net/natmap/patches/100-don-t-get-commit-id-from-git.patch
ysc3839: Thanks.

net/natmap/Makefile
ysc3839: Fixed.

net/natmap/patches/100-don-t-get-commit-id-from-git.patch
```diff
@@ -0,0 +1,25 @@
...
+ 1 file changed, 9 deletions(-)
+
+--- a/build.mk
++++ b/build.mk
+@@ -6,12 +6,3 @@ rwildcard=$(foreach d,$(wildcard $1*), \
+ 
+ SRCFILES=$(call rwildcard,$(SRCDIR)/,*.c *.S)
+ SRCFILES=$(call rwildcard,$(SRCDIR)/,*.c *.S)
+-
+-TAG_ID=$(shell git -C $(SRCDIR) tag --points-at HEAD)
```
1715173329, line 18: Try to use `TAG_ID?=`, then it can be passed via make flags. This change should be acceptable on upstream, also makes BKPepe happy.

ysc3839: Updated.

1715173329:
Can't believe I created a PR for a PR.

heiher:
Is ready to create a tag? It missing something or not?

1715173329:
Looks everything is ready to go.

heiher:
> Looks everything is ready to go.

Done: https://github.com/heiher/natmap/releases/download/20221203/natmap-20221203.tar.gz

ysc3839:
There's another issue that runtime data in `/var/run/natmap` would not be removed.
I found a package using `instance.start` procd trigger: https://github.com/openwrt/packages/blob/091681e1523d2943e665c392b09f79d6959fe632/net/mstpd/files/etc/init.d/mstpd.init#L93
But according to my test, there's no `instance.start` and `instance.stop` triggers, only a `instance.update` trigger is present. And `instance.update` is only triggered when manually start/stop the service.
Is there a procd trigger for process stop event? So I can delete `/var/run/natmap/${PID}.json`.

1715173329:
fyi https://openwrt.org/docs/guide-developer/procd-init-scripts#stopping_services

ysc3839:
@1715173329 But that's also only triggered by manually start/stop, it can't be triggered by process exit (for example when process crashed).

1715173329:
> @1715173329 But that's also only triggered by manually start/stop, it can't be triggered by process exit (for example when process crashed).

Iirc there's no such logic to process this. It would be better to use a fixed file name.

1715173329:
Cc @jow-

ysc3839:
Updated.

ysc3839:
@1715173329 Can I backport this package to 22.03?

1715173329:
No problem ;)

BKPepe:
As requested, I backported the package to OpenWrt 22.03.
Commit: https://github.com/openwrt/packages/commit/ec2fbd57eeccc47ef33eba17a500a0f269454db3

PR #20002, opened 2022-12-01
ci: only comment AUTORELEASE deprecation if exists

aparcar:
If it doesn't exists, don't confuse the contributors.


aparcar:
Jokes on me, by using `pull_request_target` as target the CI uses the *upstream* YAML and not the one I modified here.

PR #20009, opened 2022-12-02
docker-compose: Update to version 2.14.0

jmarcet:
Maintainer: me
Compile tested: master x86_64
Run tested: master x86_64

Description:
New upstream release.

Since version 2.13.0, the cpu limit option in compose files version 2, does not work anymore.

I get [this error](https://github.com/SUSE/docker/blob/3e0025e2fc137f51e95758618e1b42627fe6ea1c/daemon/daemon_unix.go#L539):

"does not support CPU percent. Percent discarded."

I verified it happens just by changing compose, so it is not the kernel or the system.

Other than that it is working just fine, but I'd like to solve the issue.

### DIFF
--- a/utils/docker-compose/Makefile
+++ b/utils/docker-compose/Makefile
@@ -1,14 +1,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=compose
-PKG_VERSION:=2.12.2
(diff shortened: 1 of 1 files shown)


1715173329:
Please set `PKG_RELEASE` to `1`

PR #20014, opened 2022-12-03
frp：add enable options

zxlhhyccc:
Because luci-app-frpc and luci-app-frps do not have the enable option, you cannot control whether to enable or disable the plug-in.

### DIFF
--- a/net/frp/Makefile
+++ b/net/frp/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=frp
 PKG_VERSION:=0.45.0
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/fatedier/frp/tar.gz/v${PKG_VERSION}?


net/frp/files/frpc.init
```diff
@@ -49,6 +49,10 @@ start_service() {
 	> "$conf_file"
 	config_load "$NAME"
 
+	local enabled
+		config_get_bool enabled "common" 'enabled' '0'
+		[ "$enabled" = "1" ] || return 1
```
feckert, line 54: can you please change the indentation. This is wrong here!

net/frp/files/frps.init
```diff
@@ -47,6 +47,10 @@ start_service() {
 	> "$conf_file"
 	config_load "$NAME"
 
+	local enabled
+		config_get_bool enabled "common" 'enabled' '0'
+		[ "$enabled" = "1" ] || return 1
+
```
feckert, line 53: can you please change the indentation. This is wrong here!

ysc3839:
What do you want to achieve? You put the enable check in `start_service()`, which disables the whole service, instead of a proxy. If you want to disable the whole service, you can just `/etc/init.d/frpc disable` (or use System Service in luci).

This init script simply coverts uci config options to ini config options, since frp's config is complicated.
If you want to disable a single proxy but keep it in config file, currently frp doesn't provide a `enable` option in config file.
A solution is changing config type `conf` to other, because the init script ignores types other than `conf` and `init`: https://github.com/openwrt/packages/blob/23e6200e4d0a435915ab4ef9700a7297e89b68b3/net/frp/files/frpc.init#L14-L38 This solution theoretically possible in luci.

BKPepe:
Closing. Review was given, it was not addressed.

PR #20016, opened 2022-12-03
stress-ng: bump to version 0.15.00

commodo:
Maintainer: me
Compile tested: x86 https://github.com/openwrt/openwrt/commit/cf361b85097216538dfac5ad7b22050390b0bc67
Run tested: x86 https://github.com/openwrt/openwrt/commit/cf361b85097216538dfac5ad7b22050390b0bc67


commodo:
@hnyman @BKPepe

hnyman:
Is that xxhash really needed?
I thought that you might patch the detection away, but instead you added the explicit dependency.

Did you look into the role of the lib in stresss-ng?
To me the upstream changes look like xxhash is one new optional hash algorithm than can be used, but not mandatory.
https://github.com/ColinIanKing/stress-ng/commit/9b9d837887ce1183303e7d936d0ef1e42851f2a8

In general, we should avoid mandatory library dependency bloat, so I might prefer disabling the optional feature.

commodo:
> Is that xxhash really needed? I thought that you might patch the detection away, but instead you added the explicit dependency.
>

i'll remove it then; no worries from my side;
 
> Did you look into the role of the lib in stresss-ng? To me the upstream changes look like xxhash is one new optional hash algorithm than can be used, but not mandatory. [ColinIanKing/stress-ng@9b9d837](https://github.com/ColinIanKing/stress-ng/commit/9b9d837887ce1183303e7d936d0ef1e42851f2a8)
> 

yeah, stress-ng has a few optional features/libs that pull in various libs; xxhash is another one of them;

> In general, we should avoid mandatory library dependency bloat, so I might prefer disabling the optional feature.

well, in the packages feed, we are a bit passed the point of bloat-ware; 
there are a few packages that pull in a lot of deps, and require too many deps;
i think you are among the few that care about minimalism (and i am fine with it)  :)

commodo:
> Fanciest would a be build config option defaulting to off, so that the user could select at compile-time to include libxxhash feature if he wants it, but the feature (and depedency lib) would not be there by default.

yeah, i know; i'm a bit lazy about doing it onw;

> But that might be an overkill for this kind of package.

it may make sense at some point;

> I have started with OpenWrt in 2010 with 8 or 16 MB of flash in WNDR3700v1/3700v2/3800, so my perspective is still on sensible flash space usage ;-) The other angle is that having additional packages installed in the router always adds complexity, increases crash probability and makes exotic config problems more possible.

i started with Openwrt in 2014, while working for Ocedo (which was later acquired by Riverbed)
all that's left from that are the Ocedo Koala/Raccoon/Ursus in OpenWrt (which were donated to the Fritz Project sometime after the acquisition);
they had/have 16 MB of flash, but it was split into 2 to be able to do A/B partition-scheme updates;
so, we cared about size too because of those devices;
but we also had devices with 128 MB of space;

when openwrt dropped the 4/32 device support, i was a bit upset, but i understand;
but, it also made me think that it starts to make room for a pico-wrt sub-project/derivative;
maybe take another kernel (zephyr/rtos/etc) and a smaller userland    ¯\_(ツ)_/¯

but recently, a lot of "embedded" devices have sd-card storage, which allows for yocto and other things that will kill your PC if you want to build with it

hnyman:
@commodo 
CI fails it for dirty patch...
You need to run refresh once more.

commodo:
> @commodo CI fails it for dirty patch... You need to run refresh once more.

yep; apologies
i was multi-tasking stuff last night and did not finish this

commodo:
so, AFAICT: the failing builds are some other issue;

hnyman:
Looks like that. 
Likely a kernel 5.15 bump affecting SDK versioning/dependencies.

commodo:
so, i will try to remember to check the buildbots;

but if i forget, or you see an error sooner than me, ping me

i feel i'm just barely getting my head out of all work i need to finish

PR #20020, opened 2022-12-04
irqbalance: Managing the package status only on services

tiagofreire-pt:
Maintainer: @hnyman 
Compile tested: IPQ807x, Xiaomi AX9000, r21394-1d82c6b4e3
Run tested: IPQ807x, Xiaomi AX9000, r21394-1d82c6b4e3
Signed-off-by: Tiago Freire <redalert2004@gmail.com>

Description:
The ```as is``` package configuration leads to misleading managing, after its installation. The ```service``` command won't run unless you dig on the documentation to find this flag to be enabled priory the ```service```, under the configuration file.

The proposal is to manage the package status only under the ```services``` method, making the enable flag always ```true```.

### DIFF
--- a/utils/irqbalance/Makefile
+++ b/utils/irqbalance/Makefile
@@ -5,7 +5,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=irqbalance
-PKG_VERSION:=1.9.3
+PKG_VERSION:=1.9.2
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE_PROTO:=git
--- a/utils/irqbalance/Makefile
+++ b/utils/irqbalance/Makefile
@@ -5,7 +5,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=irqbalance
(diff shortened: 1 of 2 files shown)


hnyman:
NAK.

First, there is no 1.9.3 version in upstream, so it makes no sense to update to a non-existing version...

Having a disabled/enabled flag in the uci config file is quite normal. Irqbalance does not help in all cases, so offering an easy way to manage its settings is essential. uci config file is the default config method in OpenWrt.

Default being 0 enables a safe addition of the package to a firmware build, and the user then needs to explicitly enable it in the uci config. 

Note that the "service method" is unreliable, as all found services are enabled after flashing a new firmware (as they are enabled in the compilation process of a new firmware.

tiagofreire-pt:
The version bump was reverted.

tiagofreire-pt:
The irqbalance not always help is due to its default configuration being agnostic to the hardware itself, putting on the same level and cpu core the interrupts for WAN and LAN interfaces, e.g. A properly configured package is virtually always helpful, as long you use it on a modern and powerful enough SoC.

PR #20022, opened 2022-12-04
prometheus-node-exporter-lua: Add "node_textfile_mtime_seconds" metric

hoelzro:
…for textfile collector, to make it more consistent with the upstream Prometheus node-exporter

Signed-off-by: Rob Hoelz <rob@hoelz.ro>

Maintainer: @champtar
Compile tested: Archer C7 and Netgear R7800, OpenWrt 22.03.0
Run tested: Archer C7 and Netgear R7800, OpenWrt 22.03.0 - dropped in my change and verified that the existing collector still worked and the new metric was present

Description:  This adds a new metric to the textfile collector which exposes the mtime of each scraped text file - the official node exporter has this metric too, and it's handy in alerting on things like textfile-updating cronjobs that fail to run

### DIFF
--- a/net/haproxy/Makefile
+++ b/net/haproxy/Makefile
@@ -10,12 +10,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=haproxy
-PKG_VERSION:=3.0.0
+PKG_VERSION:=3.0.1
 PKG_RELEASE:=1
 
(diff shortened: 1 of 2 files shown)


champtar:
It allowed me to force push but closed the PR ...

champtar:
Pushed to master as 4edae8449946f46bbb3943eb0af066048ef37c54

PR #20032, opened 2022-12-06
django: bump version 4.1.3

peter-stadler:
Maintainer: @commodo and me
Compile tested: x86_64, qemu, master snapshot
Run tested: x86_64, qemu, master snapshot, serve etebase

Description: fix CVE-2022-41323

### DIFF
--- a/lang/python/django/Makefile
+++ b/lang/python/django/Makefile
@@ -8,11 +8,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=django
-PKG_VERSION:=4.1.1
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_VERSION:=4.1.3
+PKG_RELEASE:=1
 
 PYPI_NAME:=Django
-PKG_HASH:=a153ffd5143bf26a877bfae2f4ec736ebd8924a46600ca089ad96b54a1d4e28e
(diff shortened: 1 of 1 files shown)


1715173329:
> 
> * lang/python/django/Makefile: PKG_RELEASE:=3

Correct value should be `1`. @aparcar would you mind taking a look?

lang/python/django/Makefile
```diff
@@ -8,11 +8,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=django
-PKG_VERSION:=4.1.1
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_VERSION:=4.1.3
+PKG_RELEASE:=3
```
commodo, line 12: PKG_RELEASE:=1

it's a version bump

commodo:
> > Please do no longer set _PKG_RELEASE_ to _AUTORELEASE_ as the feature is deprecated. Please use an integer instead. Below is a list of affected packages including correct _PKG_RELEASE_:
> > 
> > * lang/python/django/Makefile: PKG_RELEASE:=3
> 
> Correct value should be `1`. @aparcar would you mind taking a look?

i did not miss these review comments :P

lang/python/django/Makefile
peter-stadler: done.

PR #20033, opened 2022-12-06
django: bump version 4.0.8

peter-stadler:
Maintainer: @commodo and me
Compile tested: none
Run tested: none

Description: fix CVE-2022-41323


lang/python/django/Makefile
```diff
@@ -8,11 +8,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=django
-PKG_VERSION:=4.0.7
+PKG_VERSION:=4.0.8
 PKG_RELEASE:=$(AUTORELEASE)
```
commodo, line 12: we'd need to change this to `PKG_RELEASE:=1`

now :)

commodo: as a general topic, there is a discussion to drop this AUTORELEASE feature
so, we will slowly drop it i guess

peter-stadler: done.

PR #20041, opened 2022-12-08
ruby: update to 3.1.3

bradfordzhang:
Add a comment about confusing code
---
 configure.ac | 1 +
 1 file changed, 1 insertion(+)

Signed-off-by: Bradford Zhang <zyc@zyc.name>

### DIFF
--- a/lang/ruby/Makefile
+++ b/lang/ruby/Makefile
@@ -11,7 +11,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ruby
-PKG_VERSION:=3.1.2
+PKG_VERSION:=3.1.3
 PKG_RELEASE:=$(AUTORELEASE)
 
 # First two numbes
@@ -19,7 +19,7 @@ PKG_ABI_VERSION:=$(subst $(space),.,$(wordlist 1, 2, $(subst .,$(space),$(PKG_VE
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://cache.ruby-lang.org/pub/ruby/$(PKG_ABI_VERSION)/
-PKG_HASH:=ca10d017f8a1b6d247556622c841fc56b90c03b1803f87198da1e4fd3ec3bf2a
+PKG_HASH:=4ee161939826bcdfdafa757cf8e293a7f14e357f62be7144f040335cc8c7371a
 PKG_MAINTAINER:=Luiz Angelo Daros de Luca <luizluca@gmail.com>
 PKG_LICENSE:=BSD-2-Clause
 PKG_LICENSE_FILES:=COPYING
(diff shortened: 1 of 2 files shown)


lang/ruby/patches/001-fix-build-with-libressl-3.5.patch
```diff
@@ -1,6 +1,6 @@
-From e25fb0d0d86da5a9398ebdc9216b2ea89f80fa3d Mon Sep 17 00:00:00 2001
-From: Jeremy Evans <code@jeremyevans.net>
-Date: Fri, 25 Mar 2022 13:11:31 -0700
+From b1641baee099805a07c87a57b6d86f43002d8139 Mon Sep 17 00:00:00 2001
+From: Bradford Zhang <zyc@zyc.name>
```
1715173329, line 2: what? this is not good. please keep original authorship.

lang/ruby/patches/001-fix-build-with-libressl-3.5.patch
```diff
@@ -9,7 +9,7 @@ Subject: [PATCH] Fix build with LibreSSL 3.5
 
 --- a/ext/openssl/ossl_pkey.c
 +++ b/ext/openssl/ossl_pkey.c
-@@ -670,7 +670,7 @@ ossl_pkey_export_traditional(int argc, V
+@@ -710,7 +710,7 @@ ossl_pkey_export_traditional(int argc, VALUE *argv, VALUE self, int to_der)
```
1715173329, line 12: run `make package/ruby/refresh V=s` to refresh patches.

1715173329 (changes requested):

lang/ruby/Makefile
```diff
@@ -11,15 +11,15 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ruby
-PKG_VERSION:=3.1.2
+PKG_VERSION:=3.1.3
 PKG_RELEASE:=$(AUTORELEASE)
```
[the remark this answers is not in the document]
bradfordzhang: Thanks but I give up

luizluca:
Thanks @bradfordzhang . I'll take it from here (#20064). But, please, try it again next time. Minor version bumps are normally straightforward.  

The $(AUTORELEASE) change hit you mid air. It is not directly related to ruby (or any other package) but it was recently deprecated by openwrt team. Any changes that touches a package that uses it should replaced that with a fixed number,  manually incremented on each package change.  As we are changing the version number as well, the release number rewinds back to 1.

And the `make package/ruby/refresh V=s` would fix the patches automatically for you. It looks like you did it manually because the line was not truncated. And we don't change author when a patch is refreshed. If you needed to modify it, you can explicitly add your info after the original commit message (before the ---).

Your commit message had a strange content: "Add a comment about confusing code". Version bumps do not need to be very verbose but it is nice to add CVE numbers when they are fixed.

bradfordzhang:
> Thanks @bradfordzhang . I'll take it from here (#20064). But, please, try it again next time. Minor version bumps are normally straightforward.
> 
> The $(AUTORELEASE) change hit you mid air. It is not directly related to ruby (or any other package) but it was recently deprecated by openwrt team. Any changes that touches a package that uses it should replaced that with a fixed number, manually incremented on each package change. As we are changing the version number as well, the release number rewinds back to 1.
> 
> And the `make package/ruby/refresh V=s` would fix the patches automatically for you. It looks like you did it manually because the line was not truncated. And we don't change author when a patch is refreshed. If you needed to modify it, you can explicitly add your info after the original commit message (before the ---).
> 
> Your commit message had a strange content: "Add a comment about confusing code". Version bumps do not need to be very verbose but it is nice to add CVE numbers when they are fixed.

Thanks a lot. I'll do it better next time. :-)

PR #20056, opened 2022-12-10
bind: update bind.keys and db.root to the latest version

rucketeg:
New root trust anchor file downloaded from:
https://downloads.isc.org/isc/bind9/keys/9.11/bind.keys.v9_11

This fixes an issue where DNSSEC would fail after showing following message in the logs:
managed-keys-zone: DNSKEY set for zone '.' could not be verified with current keys

Also, update root name servers hints file, while we're at it.

Maintainer: @nmeyerhans 
Compile tested: bcm27xx, Rpi4
Run tested: bind works correctly with this file

### DIFF
--- a/net/bind/Makefile
+++ b/net/bind/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bind
 PKG_VERSION:=9.18.7
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_RELEASE:=3
 USERID:=bind=57:bind=57
 
 PKG_MAINTAINER:=Noah Meyerhans <frodo@morgul.net>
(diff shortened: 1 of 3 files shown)


rucketeg:
> 
>     * net/bind/Makefile: PKG_RELEASE:=3

This is fixed in 7899dc25051c0970742cf8ead382e76e328c7760.

nmeyerhans:
The keys in the bind.keys change aren't actually changed. The ones we're currently shipping are still current and haven't changed since 2017.  I don't have any trouble validating DNSSEC with the key configuration we ship, and I don't think key issues are the root cause of your problem.

The root NS updates are worth having, though.

rucketeg:
I just spotted a39a8372b37e695d20f287d26c3cd8179100a97b where the `trust-anchors` stance was explicitly enabled, I should have checked the history before making this PR. I will look further for the source of my DNSSEC problems.

pprindeville:
Why all the trailing spaces?

PR #20057, opened 2022-12-11
prometheus-node-exporter-lua: add thermal zone exporter

daya0576:
Maintainer: @champtar 
Compile tested: Ubuntu 22.04 x64
Run tested: nanopi-r2s(OpenWrt 21.02.3)

Description: 
Adding thermal zone exporter to monitor temperature of device:
```shell
# TYPE node_thermal_zone_temp gauge
node_thermal_zone_temp{name="0",type="soc-thermal"} 49.09
node_scrape_collector_duration_seconds{collector="thermal_zone"} 0.00050687789916992
node_scrape_collector_success{collector="thermal_zone"} 1
```
image

---

Official node exporter implementation: 
- https://github.com/prometheus/node_exporter/blob/master/collector/thermal_zone_linux.go
- https://github.com/prometheus/procfs/blob/master/sysfs/class_thermal.go#L42

### DIFF
--- a/utils/prometheus-node-exporter-lua/Makefile
+++ b/utils/prometheus-node-exporter-lua/Makefile
@@ -4,7 +4,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=prometheus-node-exporter-lua
-PKG_VERSION:=2022.08.08
+PKG_VERSION:=2022.12.11
 PKG_RELEASE:=1
 
 PKG_MAINTAINER:=Etienne CHAMPETIER <champetier.etienne@gmail.com>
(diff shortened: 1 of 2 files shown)


mnlipp:
Not all devices use thermal zone for reporting. Notably the Fritz!Box 4040 has the information only in hwmon. I provide it like this (compliant to general linux node exporter):

```lua
#!/usr/bin/lua

local fs = require "nixio.fs"

local function scrape()
  for input in fs.glob("/sys/class/hwmon/hwmon*/temp1_input") do
    local name = input:gsub("/sys/class/hwmon/",""):gsub("/temp1_input","")
    local temp = get_contents(input)
    labels={}
    labels["sensor"]=name
    metric("node_hwmon_temp_celsius", "gauge", labels, tonumber(temp) / 1000.0)
  end
end

return { scrape = scrape }
```

It would make sense to merge this and provide all thermal information with one exporter plugin. Note that neither information is redundant. Although the [kernel documentation](https://www.kernel.org/doc/Documentation/thermal/sysfs-api.txt) suggests that hwmon is redundant if thermal zone is available, I have a router where the values reported by hwmon have no corresponding readings in the thermal zones.

champtar:
I ended up merging #21005 a while ago, so closing this

PR #20058, opened 2022-12-11
Update tailscale to 1.34

jwischka:
Signed-off-by: jwischka <jeff@wisch.org>

Maintainer: Jan Pavlinec <jan.pavlinec1@gmail.com> (find it by checking history of the package Makefile)
Compile tested: 22.03, mvebu, ipq40xx, mt762x/1
Run tested: Multiple devices, tested builds and functionality.

Description:

### DIFF
--- a/net/tailscale/Makefile
+++ b/net/tailscale/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tailscale
-PKG_VERSION:=1.32.3
+PKG_VERSION:=1.34.0
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


jwischka:
I'm new here, so I'm not exactly sure what the process is for getting this run again. The build errors are quite odd, especially since I tested builds both on aarch64 and cortex-a15. The errors seem to indicate that the kernel can't be found for those builds which... seems wrong. 

Any help would be appreciated.

BKPepe:
Let's solve the formalities check first, shall we?

```
: Author name (jwischka) need to be your real name 'firstname lastname'
: Commit subject line MUST start with '<package name>: ' (Update tailscale to 1.34)
```

PR #20059, opened 2022-12-11
acme: deprecate state_dir

hgl:
Maintainer: @tohojo 

Description:

state_dir is actually a hardcoded value in conffiles. Allowing users to customize it could result in losing certificates after upgrading if they don't also specify the dir as being preserved. We shouldn't default to this dangerous behavior.

With the new ACME package, certificates live in the standard location /etc/ssl/acme, users who need to do certificate customizations should look for them in that dir instead.


net/acme-common/files/acme.sh
```diff
@@ -8,10 +8,7 @@
 #
 # Authors: Toke Høiland-Jørgensen <toke@toke.dk>
 
-export state_dir=/etc/acme
-export account_email=
-export debug=0
-export run_dir=/var/run/acme
+run_dir=/var/run/acme
```
tohojo, line 11: Why stop exporting the run_dir? No reason the hook scripts can't use that...

tohojo (changes requested):
A few nits; also, you need to bump the package version...

net/acme-common/files/acme.sh
hgl: IMHO, the contract that acme-common has with acme-clients is `/var/run/acme`, not the variable `run_dir`, they shouldn't depend on a implementation detail.

net/acme-common/files/acme.sh
```diff
@@ -107,11 +104,19 @@ load_globals() {
 		log err "account_email option is required"
 		exit 1
 	fi
+	export account_email
+
+	config_get state_dir "$section" state_dir
+	if [ "$state_dir" ]; then
```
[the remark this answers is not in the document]
hgl: I thought about this one, and found warning on non-empty value to be better, since that means the user has the option set in the config, and could give an impression that it could be changed to other values.

net/acme-common/files/acme.sh
hgl: Actually, I'm not sure about this. It's mostly used to expose `/var/run/acme/challenge`. What do you think, should we expose `/var/run/acme/challenge` or `$run_dir/challenge`?

PR #20062, opened 2022-12-12
pbr: detect missing iptables

stangri:
Maintainer: me
Run tested: x86_64, Sophos SG-135, OpenWrt 22.03.2, install/start


stangri:
Given that 5 out of 9 platforms built and that the release CI test only failed on two platforms, I'd still like to merge it before the end of the day Monday. Please voice your concerns if I shouldn't.

PR #20083, opened 2022-12-13
miniflux: update to 2.0.41

paper42:
Maintainer: me


paper42:
>     * utils/miniflux/Makefile: PKG_RELEASE:=2

I am not sure why the bot suggests 2, this is an update, so there shouldn't  be a PKG_RELEASE:=1 with this version anywhere, right?

1715173329:
Yup.

PR #20088, opened 2022-12-14
acme: Export the canonical paths for certificates and challenges

tohojo:
This adds an export of the canonical paths for certificates and challenges so
the contract for consumers becomes explicit. Also contains a couple of cleanups
to make use of those exports in the hook scripts.

Cc @hgl


hgl:
LGTM

PR #20103, opened 2022-12-18
libdaq3: update to 3.0.10

graysky2:
Upstream bump

Signed-off-by: John Audia <therealgraysky@proton.me>

Maintainer: @flyn-org

PR #20104, opened 2022-12-18
snort3: update to 3.1.50.0

graysky2:
Upstream bump
```
    Build system: x86_64
    Build-tested: bcm2711/RPi4B
    Run-tested: bcm2711/RPi4B
```

Maintainer: @flyn-org

### DIFF
--- a/net/snort3/files/local.lua
+++ b/net/snort3/files/local.lua
@@ -2,7 +2,7 @@
 
 -- switch tap to inline in ips and uncomment the below to run snort in inline mode
 --snort = {}
---snort["-Q"] = ''
+--snort["-Q"] = true
 
 ips = {
   mode = tap,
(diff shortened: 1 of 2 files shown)


net/snort3/files/local.lua
```diff
@@ -50,3 +50,10 @@ file_policy = {
     }
   }
 }
+
```
BKPepe, line 53: I would suggest to split these changes to separate commit.

graysky2: done!

graysky2:
As I mentioned in https://github.com/openwrt/packages/pull/20103, libdaq3 needs to be updated before this snort3 update will successfully build so the failure in the check is expected.

flyn-org:
Please drop the "-1" from "snort3: update to 3.1.49.0-1". Other than this, the request looks good. Recommend merge.

flyn-org:
It is frustrating to have a merge request change out from under me after I recommend merging it. Please revert this to the commit I reviewed. You can then submit another merge request to take the package to 3.1.50.0. I am grateful for your help, @graysky2, but I prefer we go about this in a more deliberate way.

graysky2:
Ah, my bad.  I thought it made no sense to merge in the bump to 3.1.49.0 since it was superseded by 3.1.50.0 prior to getting merged.

PR #20105, opened 2022-12-18
dos2unix: Deactivate NLS support

hauke:
Maintainer: @1715173329 
Compile tested: malta/be
Run tested: none

Description:

Without this patch the build uses the msgfmt application which is
provided by the host tool gettext in OpenWrt. Instead of adding the
dependency to gettext remove NLS support.

This fixes the following build error:
msgfmt -c po/da.po -o po/da.mo
make[4]: msgfmt: No such file or directory
make[4]: *** [Makefile:472: po/da.mo] Error 127

### DIFF
--- a/utils/dos2unix/Makefile
+++ b/utils/dos2unix/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dos2unix
 PKG_VERSION:=7.4.3
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://waterlan.home.xs4all.nl/dos2unix/ \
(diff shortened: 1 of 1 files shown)


utils/dos2unix/patches/010-deactivate-nls.patch
```diff
@@ -0,0 +1,10 @@
+--- a/Makefile
++++ b/Makefile
+@@ -50,7 +50,6 @@ MAC2UNIX        = mac2unix
+ UNIX2MAC        = unix2mac
+ 
+ # Native Language Support (NLS)
+-ENABLE_NLS      = 1
```
1715173329, line 7: Could this be replaced by `MAKE_FLAGS+= ENABLE_NLS=`? So we won't need to keep a patch here ;)
Thank you!

hauke: Thanks for the hint, yes that works too.  I updated the pull request.

1715173329 (approved):
LGTM - Thanks!

PR #20111, opened 2022-12-19
Fix motion gettext

nrobert13:
Maintainer:  @mips171 (original Roger D <rogerdammit [at] gmail.com>)
Compile tested: (ARM, Linksys, OpenWrt version 22.03.2)

Description:
Package compile fails without.


multimedia/motion/Makefile
```diff
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=motion
-PKG_VERSION:=4.3.2
+PKG_VERSION:=4.5.0
```
1715173329, line 11: Please split this change to separate commit.

nrobert13: thanks for the hint. updated PR.

PR #20149, opened 2022-12-21
squeezelite: bump to 1.9.9-1415

robho:
Signed-off-by: Robert Högberg <robert.hogberg@gmail.com>

Maintainer: @thess 
Compile tested: ramips, Asus RT-AC51U, OpenWrt 22.03
Run tested: ramips, Asus RT-AC51U, OpenWrt 22.03


neheb:
out of curiosity, what's the difference between mpg123 and libmad here?

robho:
> out of curiosity, what's the difference between mpg123 and libmad here?

(I guess you noticed that I haven't changed any functionality. libmad was disabled by an openwrt patch before this change while now libmad is disabled with a compile time option.)

I don't know why mpg123 was chosen over libmad. It's been like this since #4766.

I found some benchmarks on the net indicating that mpg123 was faster and some people had noticed audio quality problems with libmad. But this could have changed over the years and I don't know if it even was the reason for choosing mpg123.

The openwrt libmad package is 20 kb smaller than the libmpg123 package.

---

Since squeezelite supports dynamic loading of libraries I had an idea to create a package (modify the squeezelite-mini package, I think) where squeezelite tries to load both libmad and libmpg123 and use whatever it finds. Then the user can install the decoder library it wants to use. For squeezelite-full we could depend on both decoders, but it might be a waste of space to install two separate mp3 decoders when installing squeezelite-full.

neheb:
libmad is smaller and faster on mips, slower elsewhere.

Just wondering since they both decode mp3.

PR #20152, opened 2022-12-21
[22.03 backport] telnet-bsd: add package

f00b4r0:
The telnet-bsd package currently only exist in master: this PR aims to bring it into current stable release.

It is a leaf package, sole provider and not affecting existing things, so hopefully backporting should be acceptable.


aparcar:
Fine with me

PR #20157, opened 2022-12-22
libarchive: update to 3.6.3

paper42:
Maintainer: @morgenroth
Compile tested: armv7, Turris Omnia, OpenWrt master
Run tested: no, unfortunately I don't currently have a router running OpenWrt master

Description:
* fixes CVE-2022-36227
* drop upstreamed patch
* removed deprecated AUTORELEASE


paper42:
The CI failure seems to be an issue in CI itself, it looks like it tries to install all packages and if they conflict, it fails.

neheb:
It's 3.6.2, not .3.

PR #20159, opened 2022-12-22
treewide: refresh GCC options in Makefiles

Spudz76:
> several Makefiles still used `CONFIG_GCC_USE_VERSION_*` to select various compilation options, for GCC versions that are antiquated
> 
> convert to parsing the major from the `CONFIG_GCC_VERSION` which will always exist and can also be used with range logic
> 
> some intent seemed to be "10 or newer" and has been adjusted as such. GCC 11 or 12 would likely revert to the default (6 through 9) options with those, because 10 was the newest at the time

Requesting comments that cross-check the logic transformations -- generally "<=9" or ">=10", but some were using inverted logic so I may have double-inverted something in conversion to ranges using filter-out, but the more I stare at it the less my eyes can be trusted.

The methods for parsing version-major were borrowed from `toolchain/gcc/Makefile` in the main repo.

### DIFF
--- a/utils/gnupg2/Makefile
+++ b/utils/gnupg2/Makefile
@@ -104,7 +104,7 @@ CONFIGURE_ARGS += \
 	--disable-tests \
 	--disable-zip
 
-TARGET_CFLAGS += $(if $(CONFIG_GCC_USE_VERSION_10),-DEXTERN_UNLESS_MAIN_MODULE=static)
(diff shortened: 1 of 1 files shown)


Spudz76:
The four failures don't appear to be related at all to these changes, and unclear why they are fiddling with `bluez` components anyway.

Spudz76:
I split into a separate commit per package assuming that is more palatable than a treewide single-hit with improved descriptives

neheb:
ping @ClaymorePT

ClaymorePT:
Boost changes look good to me. Thank you for the contribution!

PR #20165, opened 2022-12-23
libgpiod: add missing kernel build options via KCONFIG

feckert:
Maintainer: @mhei or @neheb 
Compile tested: x86_64, APU3, master
Run tested: x86_64, APU3, OpenWrt master,tests done)

Description:

This library needs the kernel config option 'CONFIG_GPIO_CDEV_V1=y' to be set. If this is not set, the tool 'gpioinfo' produces the error message 'error creating line iterator'. Add the missing kernel config option to build CDEV with API Version 1 fixes this. Other distributions does have the same issue: https://bugs.gentoo.org/807334

```
gpiochip3 - 7 lines:
        line   0: "front-led1" "apu:green:1" output active-low [used]
        line   1: "front-led2" "apu:green:2" output active-low [used]
        line   2: "front-led3" "apu:green:3" output active-low [used]
        line   3: "front-button" "front button" input active-low [used]
        line   4:    "simswap"       unused  output  active-high
        line   5: "mpcie2_reset" unused output active-high
        line   6: "mpcie3_reset" unused output active-high
```

### DIFF
--- a/libs/libgpiod/Makefile
+++ b/libs/libgpiod/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libgpiod
 PKG_VERSION:=1.6.3
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@KERNEL/software/libs/libgpiod/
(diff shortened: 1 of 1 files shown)


mhei:
Good catch, LGTM, thanks!

karlp:
This is certainly better than nothing, and probably suits us for now, but I'd been proposing that we add this to the actual default configs, as the kernel actually defaults to =y, but our current config stubs turn it off.

feckert:
@karlp We could also enable this in the kernel config for all targets, as you suggest. But I think it will not harm to add the missing Kconfig to this package, because this package explicit needs this config to work!

PR #20171, opened 2022-12-25
curl: update to 7.87.0

stangri:
Maintainer: me
Run tested: x86_64, Sophos SG-135, OpenWrt 22.03.2, test with https-dns-proxy

Description:
* Changelog https://curl.se/changes.html#7_87_0

### DIFF
--- a/net/curl/Makefile
+++ b/net/curl/Makefile
@@ -9,15 +9,15 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/nls.mk
 
 PKG_NAME:=curl
-PKG_VERSION:=7.86.0
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_VERSION:=7.87.0
+PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://github.com/curl/curl/releases/download/curl-$(subst .,_,$(PKG_VERSION))/ \
 	https://dl.uxnr.de/mirror/curl/ \
 	https://curl.askapache.com/download/ \
 	https://curl.se/download/
-PKG_HASH:=2d61116e5f485581f6d59865377df4463f2e788677ac43222b496d4e49fb627b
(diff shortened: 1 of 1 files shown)


trippleflux:
With update perl-www-curl failed to compile

stangri:
> With update perl-www-curl failed to compile

I only compile curl and dependent packages when testing. Do you care to share details about the perl-www-curl compilation failure?

Tagging @Naoir @ryzhovau @ecsv @champtar in case they can shed any light.

trippleflux:
Full log attached [output.log](https://github.com/openwrt/packages/files/10302780/output.log), it's not clear or clarified in the curl 7.87 changelogs.

stangri:
@trippleflux having looked at patches in the `perl-www-curl` directory, I think a patch may be needed to support this version. Hopefully by the early-mid January one of the people I've tagged could look into creating a patch for this version and this can be merged then.

trippleflux:
@stangri 
Yes, I have tried on my own to fixes it also but it seems need more indepth knowledges on curl and perl, the most important line is : 
```
Curl.xs:688:25: error: void value not ignored as it ought to be
  688 |                         curl_easy_setopt(clone->curl, 10000 + i, clone->strings[i]);
```

I don't know about curl API.

diizzyy:
I think replacing it with perl-Net-Curl is probably a good idea however it too needs a patch as of right now

stangri:
@diizzyy so what should be the strategy for this (and cherrypicked release) PR then?

paper42:
Other distributions are importing this commit: https://github.com/curl/curl/commit/728400f875e845f72ee5602edb905f6301ade3e7, could openwrt do that as well? I didn't try to reproduce these issues, but according to other reports, it should  fix openvpn and https://github.com/curl/curl/commit/e2aed004302e51cfa5b6ce8c8ab65ef92aa83196 should fix zabbix and probably also perl-www-curl.

stangri:
If there are already reported issues with 7.87.0 I suggest we skip this release and wait for a next point release.

diizzyy:
Agreed, if you haven't upgraded to 7.87.0 just wait on next release

PR #20174, opened 2022-12-26
natmap: fix some issues and update to 20221225

ysc3839:
Maintainer: me
Compile tested: x86-64, 22.03.2
Run tested: x86-64, 22.03.2

Description:
add Package/conffiles to prevent config file from override on update.
fix service not stopped when disabled in luci.
update to 20221225.


1715173329:
Also cherry-picked to the 22.03 branch ;)

PR #20175, opened 2022-12-26
sshuttle: Add new meta package for opkg install

tcuje:
Signed-off-by: Thomas Cujé <dev.cuje@freenet.de>

Maintainer: me / @tcuje
Compile tested: ARMv8 Processor rev 4, FriendlyElec NanoPi R2S, OpenWrt 22.03.2 r19803-9a599fee93
Run tested: ARMv8 Processor rev 4, FriendlyElec NanoPi R2S, OpenWrt 22.03.2 r19803-9a599fee93, tests: install/uninstall/opkg info, installed service start/stop/restart/reload, config correctly forwarded as args

Description:
This is a new meta package to conveniently install sshuttle (https://github.com/sshuttle/sshuttle) as a PROCD service in OpenWRT routers. Details like version, title and description kept as seen in sshuttle. All mandatory dependencies are referenced (see https://sshuttle.readthedocs.io/en/stable/openwrt.html) and available in OpenWRT. Example config with explanation is provided and installed in an inactive state.

Most arguments of sshuttle supported (compare https://sshuttle.readthedocs.io/en/stable/manpage.html#synopsis). Argument `--daemon` not being supported, since PROCD requires services to run in the foreground (see https://openwrt.org/docs/guide-developer/procd-init-scripts#procd_init_script_parameters). Argument `--pidfile` not supported, to have pid-filenames in a well known format, which can be controlled for all processes from the service. Arguments `--firewall` and `--hostwatch` not supported, for being sshuttle internal use only. Arguments ` --sudoers-no-modify` and ` --sudoers-user` not supported, because they don't apply for most OpenWRT systems.

I'll be actively using this package and sshuttle for my own purposes in the future. I'll be happy to maintain this package, update it with new sshuttle releases or take care of maintenance work with the package itself.

First time contributing here; Hi all! Any feedback welcome

### DIFF
--- a/net/sshuttle/Makefile
+++ b/net/sshuttle/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=sshuttle
 PKG_VERSION:=1.1.1
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_RELEASE:=1
 
 PYPI_NAME:=$(PKG_NAME)
 PKG_HASH:=f5a3ed1e5ab1213c7a6df860af41f1a903ab2cafbfef71f371acdcff21e69ee6
--- a/net/sshuttle/Makefile
+++ b/net/sshuttle/Makefile
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2022 Thomas Cujé
+# Copyright (C) 2022-2023 Thomas Cujé
 #
 # This is free software, licensed under the GNU General Public License v2.
 # See /LICENSE for more information.
@@ -23,20 +23,20 @@ include $(INCLUDE_DIR)/package.mk
 include ../../lang/python/python3-package.mk
 
 define Package/sshuttle
-  SUBMENU:=SSH
-  SECTION:=net
-  CATEGORY:=Network
-  TITLE:=Full-featured VPN over an SSH tunnel
-  URL:=https://github.com/sshuttle/sshuttle
-  EXTRA_DEPENDS:= \
-  	python3-light \
-  	python3-ctypes \
-  	python3-urllib
-  PKGARCH:=all
+	SUBMENU:=SSH
+	SECTION:=net
+	CATEGORY:=Network
+	TITLE:=Full-featured VPN over an SSH tunnel
+	URL:=https://github.com/sshuttle/sshuttle
+	EXTRA_DEPENDS:= \
+		python3-light \
+		python3-ctypes \
+		python3-urllib
+	PKGARCH:=all
 endef
 
 define Package/sshuttle/description
-  Transparent proxy server that works as a poor man's VPN. Forwards over ssh. Doesn't require admin. Works with Linux and MacOS. Supports DNS tunneling.
+	Transparent proxy server that works as a poor man's VPN. Forwards over ssh. Doesn't require admin. Works with Linux and MacOS. Supports DNS tunneling.
 endef
 
 define Package/sshuttle/conffiles
--- a/net/sshuttle/Makefile
+++ b/net/sshuttle/Makefile
@@ -32,6 +32,7 @@ define Package/sshuttle
   	+python3-light \
   	+python3-ctypes \
   	+python3-urllib
+  PKGARCH:=all
 endef
 
 define Package/sshuttle/description
--- a/net/sshuttle/Makefile
+++ b/net/sshuttle/Makefile
@@ -37,18 +37,22 @@ define Package/sshuttle
 endef
 
 define Package/sshuttle/description
-  Transparent proxy server that works as a poor man's VPN. Forwards over ssh. Doesn't require admin. Works with Linux and MacOS. Supports DNS tunneling.
+Transparent proxy server that works as a poor man's VPN. Forwards over ssh. Doesn't require admin. Works with Linux and MacOS. Supports DNS tunneling.
 endef
 
 define Package/sshuttle/conffiles
 /etc/config/sshuttle
 endef
 
-define Package/sshuttle/install
-	$(INSTALL_DIR) $(1)/etc/init.d
-	$(INSTALL_BIN) ./files/sshuttle.init $(1)/etc/init.d/sshuttle
+define Py3Package/sshuttle/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/sshuttle $(1)/usr/bin/
+
 	$(INSTALL_DIR) $(1)/etc/config
 	$(INSTALL_CONF) ./files/sshuttle.config $(1)/etc/config/sshuttle
+
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_BIN) ./files/sshuttle.init $(1)/etc/init.d/sshuttle
 endef
 
 $(eval $(call Py3Package,sshuttle))
--- a/net/sshuttle/Makefile
+++ b/net/sshuttle/Makefile
@@ -16,6 +16,8 @@ PKG_HASH:=f5a3ed1e5ab1213c7a6df860af41f1a903ab2cafbfef71f371acdcff21e69ee6
 
 PKG_MAINTAINER:=Thomas Cujé <dev.cuje@freenet.de>
 PKG_LICENSE:=LGPL-2.1-only
+PKG_LICENSE_FILES:=LICENSE
+
 
 include ../../lang/python/pypi.mk
 include $(INCLUDE_DIR)/package.mk
--- a/net/sshuttle/Makefile
+++ b/net/sshuttle/Makefile
@@ -28,10 +28,10 @@ define Package/sshuttle
   CATEGORY:=Network
   TITLE:=sshuttle: where transparent proxy meets VPN meets ssh
   URL:=https://github.com/sshuttle/sshuttle
-  DEPENDS:= \
-  	+python3-light \
-  	+python3-ctypes \
-  	+python3-urllib
+  EXTRA_DEPENDS:= \
+  	python3-light \
+  	python3-ctypes \
+  	python3-urllib
   PKGARCH:=all
 endef
 
--- a/net/sshuttle/Makefile
+++ b/net/sshuttle/Makefile
@@ -11,22 +11,29 @@ PKG_NAME:=sshuttle
 PKG_VERSION:=1.1.1
 PKG_RELEASE:=$(AUTORELEASE)
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://github.com/sshuttle/sshuttle/archive/refs/tags/v$(PKG_VERSION).tar.gz
-PKG_HASH:=87b6343da4b7ea164cc1638a3ff25c41cf9dbc70bbc84bc4df8c41f963a4b0b9
+PYPI_NAME:=$(PKG_NAME)
+PKG_HASH:=f5a3ed1e5ab1213c7a6df860af41f1a903ab2cafbfef71f371acdcff21e69ee6
 
 PKG_MAINTAINER:=Thomas Cujé <dev.cuje@freenet.de>
 PKG_LICENSE:=LGPL-2.1-only
 
+include ../../lang/python/pypi.mk
 include $(INCLUDE_DIR)/package.mk
+include ../../lang/python/python3-package.mk
 
 define Package/sshuttle
   SUBMENU:=SSH
   SECTION:=net
   CATEGORY:=Network
   TITLE:=sshuttle: where transparent proxy meets VPN meets ssh
-  DEPENDS:=+python3 +python3-pip +iptables-mod-extra +iptables-mod-nat-extra +iptables-mod-ipopt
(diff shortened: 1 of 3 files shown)


net/sshuttle/Makefile
```diff
@@ -0,0 +1,55 @@
...
+define Package/sshuttle/install
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_BIN) ./files/sshuttle.init $(1)/etc/init.d/sshuttle
+	$(INSTALL_DIR) $(1)/etc/config
+	$(INSTALL_CONF) ./files/sshuttle.config $(1)/etc/config/sshuttle
+endef
+
+define Package/sshuttle/postinst
+#!/bin/sh
+/usr/bin/pip3 install sshuttle==$(PKG_VERSION)
```
BKPepe, line 52: I wonder why did you choose this approach. Wouldn't it be better to have sshuttle as Python package? Check https://github.com/openwrt/packages/blob/master/lang/python/README.md and then you can add to the install section init script and config.

BKPepe (changes requested):

net/sshuttle/Makefile
tcuje: From the docs at https://openwrt.org/docs/guide-developer/packages I got the conclusion, that python packages are only for python library packages and not for python based applications. Thanks for the link to the README, I'll transform this into a python package and mark this as draft in the meantime.

tcuje: Ok, done. I kept the package in the `net` group, since it's not closely related to python. Just an application that happens to use python as runtime environment.

net/sshuttle/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+PYPI_NAME:=$(PKG_NAME)
+PKG_HASH:=f5a3ed1e5ab1213c7a6df860af41f1a903ab2cafbfef71f371acdcff21e69ee6
+
+PKG_MAINTAINER:=Thomas Cujé <dev.cuje@freenet.de>
+PKG_LICENSE:=LGPL-2.1-only
+
+include ../../lang/python/pypi.mk
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/python/python3-package.mk
```
BKPepe, line 22: You should put your package to python folder.

net/sshuttle/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+define Package/sshuttle
+  SUBMENU:=SSH
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=sshuttle: where transparent proxy meets VPN meets ssh
```
BKPepe, line 28: Isn't this too long for ``make menuconfig``?

net/sshuttle/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+define Package/sshuttle/description
+  Transparent proxy server that works as a poor man's VPN. Forwards over ssh. Doesn't require admin. Works with Linux and MacOS. Supports DNS tunneling.
+endef
+
+define Package/sshuttle/conffiles
+/etc/config/sshuttle
+endef
+
+define Package/sshuttle/install
```
BKPepe, line 47: This is still wrong, because in the package is currently only init script and config.

BKPepe (changes requested):

net/sshuttle/Makefile
tcuje: I think it's ok
grafik

net/sshuttle/Makefile
tcuje: I think the application is only losely related to python. Would you place all Java applications in `lang/java/`, too?  `fail2ban` is also using python and not in the python folder. On the functional level sshuttle perfectly relates to network, ssh and I would not look for an application based on the implementation language. Whats the structure behind here?

net/sshuttle/Makefile
tcuje: Sorry? I think the `fail2ban` package has the exact same instructions. What am I missing here?

1715173329: should use `Py3Package/sshuttle/install`, and please take care of executable files located under `$(PKG_INSTALL_DIR)/usr/bin`

tcuje: Ok, I think it's done with my last commit, please have a look.

net/sshuttle/files/sshuttle.init
```diff
@@ -0,0 +1,108 @@
...
+}
+
+start_service() {
+  _log "Starting sshuttle instances"
+  config_load "sshuttle"
+  config_foreach validate_gateway_section "gateway" load_gateway
+}
+
+stop_service() {
+  _log "Stopping sshuttle instances"
```
1715173329, line 99: I'm not sure if it is required. procd should handle this already, and `stop_service` is called even before procd kills the processes[^1].

[^1]: https://openwrt.org/docs/guide-developer/procd-init-scripts#stopping_services

net/sshuttle/Makefile
1715173329: lgtm

net/sshuttle/Makefile
```diff
@@ -0,0 +1,60 @@
...
+
+PKG_NAME:=sshuttle
+PKG_VERSION:=1.1.1
+PKG_RELEASE:=1
+
+PYPI_NAME:=$(PKG_NAME)
+PKG_HASH:=f5a3ed1e5ab1213c7a6df860af41f1a903ab2cafbfef71f371acdcff21e69ee6
+
+PKG_MAINTAINER:=Thomas Cujé <dev.cuje@freenet.de>
+PKG_LICENSE:=LGPL-2.1-only
```
BKPepe, line 18:
```suggestion
PKG_MAINTAINER:=Thomas Cujé <dev.cuje@freenet.de>
PKG_LICENSE:=LGPL-2.1-only
PKG_LICENSE_FILES:=LICENSE

```

net/sshuttle/Makefile
```diff
@@ -0,0 +1,60 @@
...
+  CATEGORY:=Network
+  TITLE:=sshuttle: where transparent proxy meets VPN meets ssh
+  URL:=https://github.com/sshuttle/sshuttle
+  DEPENDS:= \
+  +python3-light \
+  +python3-ctypes \
+  +python3-urllib \
+  +iptables-mod-extra \
+  +iptables-mod-nat-extra \
+  +iptables-mod-ipopt
```
BKPepe, line 36: Since OpenWrt 22.03 moved from iptables to nftables, thus could be somehow confusing. I think there are two options:
1. Make and do nftables variant only
2. Do it as e.g. miniupnpnd and do two variants one for iptables and second one for nftables

net/sshuttle/Makefile
tcuje: Thanks! I added the line for the license file

net/sshuttle/files/sshuttle.init
tcuje: Thanks for the hint, I will look into this.

I'm not sure, how procd handles the (possibly) multiple processes for each connetion and terminates all of them correctly.

net/sshuttle/Makefile
tcuje: Good point! I'll check this.

- [ ] is sshuttle compatible with nftables (seems so at least with NAT mode)?
- [ ] can we specify a list of compatible OpenWRT versions for the package?
- [ ] can we determine the correct package variant with the depencies and what is installed?

tcuje: Sshuttle works perfectly fine with nftables. I removed the dependencies. On OpenWRT 21.03 installations with iptables they might be mandatory though and must be installed manually. AFAIK the DEPENDS keyword does not provide such version checks.

net/sshuttle/Makefile
```diff
@@ -0,0 +1,72 @@
...
+  SUBMENU:=SSH
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=sshuttle: where transparent proxy meets VPN meets ssh
+  URL:=https://github.com/sshuttle/sshuttle
+  DEPENDS:= \
+  	+python3-light \
+  	+python3-ctypes \
+  	+python3-urllib
+  PKGARCH:=all
```
tcuje, line 35: Since this is a script only package should we use EXTRA_DEPENDS instead? All dependencies here are only needed at runtime

net/sshuttle/files/sshuttle.init
tcuje: Works like charm!

While checking I found that arguments from each instance are forwarded to the following ones. Fixed now, see my last commit.

net/sshuttle/Makefile
```diff
@@ -0,0 +1,72 @@
...
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=sshuttle
+PKG_VERSION:=1.1.1
+PKG_RELEASE:=1
+
+PYPI_NAME:=$(PKG_NAME)
```
BKPepe, line 14:
```suggestion
PYPI_NAME:=sshuttle
```

net/sshuttle/Makefile
```diff
@@ -0,0 +1,72 @@
...
+  URL:=https://github.com/sshuttle/sshuttle
+  EXTRA_DEPENDS:= \
+  	python3-light \
+  	python3-ctypes \
+  	python3-urllib
+  PKGARCH:=all
+endef
+
+define Package/sshuttle/description
+Transparent proxy server that works as a poor man's VPN. Forwards over ssh. Doesn't require admin. Works with Linux and MacOS. Supports DNS tunneling.
```
BKPepe, line 39:
```suggestion
  Transparent proxy server that works as a poor man's VPN. Forwards over ssh. Doesn't require admin. Works with Linux and MacOS. Supports DNS tunneling.
```

net/sshuttle/Makefile
```diff
@@ -0,0 +1,72 @@
...
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/python/python3-package.mk
+
+define Package/sshuttle
+  SUBMENU:=SSH
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=sshuttle: where transparent proxy meets VPN meets ssh
+  URL:=https://github.com/sshuttle/sshuttle
+  EXTRA_DEPENDS:= \
```
BKPepe, line 31: Why you used extra depends? Shouldn't it be proper dependencies?
You should check it with this guide: https://github.com/openwrt/packages/blob/96d2610994d4bb40c9ef8a63305ba1a97b6988af/lang/python/README.md#python-package-dependencies

net/sshuttle/files/sshuttle.init
```diff
@@ -0,0 +1,111 @@
...
+STOP=01
+USE_PROCD=1
+
+_log() {
+  logger -p daemon.info -t sshuttle "$@"
+}
+
+_err() {
+	logger -p daemon.err -t sshuttle "$@"
+}
```
BKPepe, line 20: This is specific for sshuttle or not? Because you should check how other packages are doing it.

BKPepe (changes requested):

net/sshuttle/Makefile
```diff
@@ -0,0 +1,72 @@
...
+define Package/sshuttle/conffiles
+/etc/config/sshuttle
+endef
+
+# Some sshuttle script files are copied to SSH remote hosts and are therefore mandatory.
+# This causes the otherwise pre-compiled package to contain the source package.
+# Compare with https://github.com/sshuttle/sshuttle/blob/v1.1.1/sshuttle/ssh.py#L95-L103
+# Since the source is very small in general, to reduce complexity
+# and support all architectures no explicit filtering is used.
+define Package/sshuttle/install
```
jefferyto, line 51: Which script files are needed exactly? By shipping the Python source files instead of precompiled bytecode, this forces target Python to compile the source files on device. This can be very slow and the reason why we ship bytecodes in the default packages.

net/sshuttle/files/sshuttle.init
```diff
@@ -0,0 +1,111 @@
...
+# See /LICENSE for more information.
+#
+
+# shellcheck disable=SC2034
+START=99
+STOP=01
+USE_PROCD=1
+
+_log() {
+  logger -p daemon.info -t sshuttle "$@"
```
jefferyto, line 15: Suggest you switch to tabs for indentation; disk space is a scarce resource for OpenWrt devices.

net/sshuttle/files/sshuttle.init
tcuje: This is not sshuttle specific, there are at least 7 other init files that use this function and I copied from there

net/sshuttle/Makefile
tcuje: I debugged into it and it's just the file listed there in the ssh.py source code for this version of sshuttle, but it may change in the future.
The files are needed in the first place to be sent and compiled on the SSH target device, not the OpenWRT device. To avoid compilation on the OpenWRT device, we would need to ship BOTH compiled and script version of the files, which I tried, but duplicates the required disk space, which is way worse IMO.

net/sshuttle/files/sshuttle.init
tcuje: sure, done

PR #20178, opened 2022-12-27
icecast: add libxml2 include directory

cotequeiroz:
Maintainer: @andrenarchy @thess 
Compile tested: aarch64/cortex-a53, master
Run tested: none

Description:

Fixes this error:
cfgfile.c:26:10: fatal error: libxml/xmlmemory.h: No such file or directory
   26 | #include <libxml/xmlmemory.h>
      |          ^~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/multimedia/icecast/Makefile
+++ b/multimedia/icecast/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=icecast
 PKG_VERSION:=2.4.4
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://downloads.xiph.org/releases/icecast/
@@ -55,6 +55,7 @@ CONFIGURE_ARGS+= \
 	--with-vorbis="$(STAGING_DIR)/usr" \
 	--with-xslt-config="$(STAGING_DIR)/usr/bin/xslt-config"
 
+TARGET_CFLAGS+= -I$(STAGING_DIR)/usr/include/libxml2
 
 # Manually edit configure in case both vorbis and tremor host packages are installed
 define Build/Configure


neheb:
do you know why it broke?

cotequeiroz:
I'm glad you asked.

From time to time I run builds with `CONFIG_ALL` to catch build problems.  I've done a quick fix for this one, without really going into the reason.

After your question, I looked into it a little further.  The change in libxslt by 0f9d554f1c34cf9169f5a6832a51acf7adf5f035 from autotools to cmake appears to have caused `xslt-config` to not expose libxml2 `CFLAGS`.

One thing is certain: you should not install different files depending which build system you choose.  The proper fix should be in libxslt.  This PR implements a workaround.  I will probably not find the time to propose a proper fix upstream.

Anyway, as an alternative, there's the possibility of building libxslt with autoconf instead of cmake.  I've tried it quickly here. 
 Apparently, it will not work with `PKG_FIXUP=autoreconf` somehow.  I did not go that deep into what's wrong.

I'll leave this open, if you think we should just work around the problem for the time being.

neheb:
It’s probably better to patch icecast to use pkgconfig.

cotequeiroz:
Icecast calls `xslt-config` to get the flags.  It does not handle libxml2 directly.  The problem is that libxslt's behavior is not constant across build system.  If you build it with autotools, it will include libxml2 with `CFLAGS`; if you use cmake, it will not add it.  This is definitely a bug in libxslt.
As for icecast's reliance on libxslt to add libxml2 flags, it is debatable.  I believe that, in principle, if you include a libxml2 header in your code, you should provide its location.  The counterargument is that libxslt might use a different libxml2 version than what you normally find, requiring different flags.  So I'm not sure I would change this behavior.

neheb:
I'd say a dependency on libxml2 is appropriate

hnyman:
Did you reach a conclusion on the proposed solution being the right one? 

There is has already been the libxml2 dependency and now the include dir flag for the compiler is added.)

cotequeiroz:
This whole discussion of which project needs to be fixed may be counterproductive, I know.

Personally, I think that the right thing to do would be reverting libxslt to using autotools.  As much as I like using cmake over autotools (I haven't started a single project from scratch using autotools), we must be aware that cmake support may be subpar.  This is not the first case we encounter here.

The problem appears to be that `LIBXML_CFLAGS` is not being set with `cmake`.  Then when `xslt-config` is generated from `xslt-config.in`, `cflags` is empty.  When autotools is used, they are set.   Also notice that libxslt cmake module, `libxslt-config.cmake`, adds libxml2 includes and libraries to their own.  I'm not a cmake expert, but I find it odd that `CMakefiles.txt` has `-lxml2` hardcoded, instead of using libxml2 variables from `find_package(LibXml2)`.

neheb:
If there's an issue with libxslt's cmake support, I'd rather patch that than revert to autotools.

hnyman:
Should this still be merged, while a "more perfect" solution is being considered?

neheb:
sure.

PR #20183, opened 2022-12-28
bind: start up with -4 if not listening on ipv6 addresses

pprindeville:
Maintainer: @nmeyerhans
Compile tested: x86_64, generic, HEAD (fb15cb4ce95)
Run tested: same, running on production router

Description:

On startup, my logging fills up with noise about not being able to reach the IPv6 addresses of the root name servers:

```
Dec 27 17:14:24 OpenWrt named[9965]: network unreachable resolving '_.org/A/IN': 2001:500:200::b#53
Dec 27 17:14:24 OpenWrt named[9965]: network unreachable resolving '_.org/A/IN': 2001:500:9f::42#53
Dec 27 17:14:24 OpenWrt named[9965]: network unreachable resolving '_.org/A/IN': 2001:dc3::35#53
Dec 27 17:14:24 OpenWrt named[9965]: network unreachable resolving '_.org/A/IN': 2001:500:1::53#53
```

This is normal.  My MSP (Sparklight) doesn't do IPv6.  To eliminate this noise, we can pass `-4` to named when it gets started.

### DIFF
--- a/net/bind/files/named.init
+++ b/net/bind/files/named.init
@@ -30,6 +30,10 @@ fix_perms() {
     done
 }
 
+no_ipv6() {
+    [ -z "$(ip -6 -o route show default)" ]
+}
+
 reload_service() {
     rndc -q reload
 }
@@ -68,8 +72,11 @@ start_service() {
 
     touch $conf_local_file
 
+    local args=
+    [ no_ipv6 ] && args="-4"
+
     procd_open_instance
-    procd_set_param command /usr/sbin/named -u bind -f -c $config_file
+    procd_set_param command /usr/sbin/named -u bind -f $args -c $config_file
     procd_set_param file $config_file \
 			 $config_dir/bind.keys \
 			 $named_options_file \
--- a/net/bind/Makefile
+++ b/net/bind/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bind
 PKG_VERSION:=9.18.10
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 USERID:=bind=57:bind=57
 
 PKG_MAINTAINER:=Noah Meyerhans <frodo@morgul.net>


nmeyerhans (changes requested):
I'm not sure I like this approach. It essentially changes the meaning of `listen-on-v6 {none;}` to also avoid sending queries over v6, which is not the documented or expected behavior of that setting. It makes the documented behavior of that setting, in which the server doesn't listen on v6 but will send queries over it, difficult to achieve without editing init scripts. (At least if I'm understanding the sed script correctly!)

In my opinion, if the goal is simply to avoid logging these messages in an IPv4-only network, it might just be preferable to suppress logging those messages without otherwise changing the behavior of the server.

Another option might be to pass the `-4` parameter if we don't have a default IPv6 route at startup.

pprindeville:
> Another option might be to pass the -4 parameter if we don't have a default IPv6 route at startup.

@nmeyerhans Done

pprindeville:
Thanks!

rucketeg:
The logic here seems reverted, I do have, and want to use IPv6, but bind is started with the `-4` option.
```
$ no_ipv6() { \[ -z "$(ip -6 -o route show default)" ] ; }
$ no_ipv6 && echo "no ipv6" || echo "ya ipv6"
ya ipv6
$ [ no_ipv6 ] && echo "no ipv6" || echo "ya ipv6"
no ipv6
```

rucketeg:
Or more generally:
```
$ /bin/false; echo $?
1
$ [ /bin/false ] ; echo $?
0
```

rucketeg:
So the line 76 should look like this:
```
no_ipv6 && args="-4"
```

net/bind/files/named.init
```diff
@@ -68,8 +72,11 @@ start_service() {
 
     touch $conf_local_file
 
+    local args=
+    [ no_ipv6 ] && args="-4"
```
rucketeg, line 76: Should be this instead:
```
no_ipv6 && args="-4"
```

PR #20205, opened 2023-01-03
byobu: fix missing dependency on tmux

ngehrsitz:
Maintainer: @jefferyto 
Compile tested: N/A
Run tested: FB7520@[r21588-2e0e25c9f7](https://github.com/openwrt/openwrt/commit/2e0e25c9f7)

Description:
You can´t launch byobu without having tmux installed:
```
root@FB7520:~# byobu
/usr/bin/byobu: exec: line 242: tmux: not found
```
Installing it fixes the issue so it should be a dependency:
```
root@FB7520:~# opkg install tmux
Installing tmux (3.3a-1) to root...
Downloading https://downloads.openwrt.org/snapshots/packages/arm_cortex-a7_neon-vfpv4/packages/tmux_3.3a-1_arm_cortex-a7_neon-vfpv4.ipk
Installing libevent2-core7 (2.1.12-1) to root...
Downloading https://downloads.openwrt.org/snapshots/packages/arm_cortex-a7_neon-vfpv4/base/libevent2-core7_2.1.12-1_arm_cortex-a7_neon-vfpv4.ipk
Configuring libevent2-core7.
Configuring tmux.
root@FB7520:~# byobu

Welcome to the light, powerful, text window manager, Byobu.
```

### DIFF
--- a/utils/byobu/Makefile
+++ b/utils/byobu/Makefile
@@ -36,7 +36,7 @@ endef
 define Package/byobu
 $(call Package/byobu/Default)
   TITLE:=Text-based window manager and terminal multiplexer
-  DEPENDS:=+python3-light +python3-newt
+  DEPENDS:=+python3-light +python3-newt +!PACKAGE_screen:tmux
 endef
 
 define Package/byobu-utils


paper42:
I think byobu can work with screen as well.

jefferyto (commented):
@paper42 is correct - byobu works with either screen or tmux. If there is a way to say “either tmux or screen” (as a dependency) then that could work, but I don’t see this being merged as it is.

ngehrsitz:
> @paper42 is correct - byobu works with either screen or tmux. If there is a way to say “either tmux or screen” (as a dependency) then that could work, but I don’t see this being merged as it is.

From my understanding of https://openwrt.org/docs/guide-developer/dependencies#special_notes this should work.

utils/byobu/Makefile
```diff
@@ -37,6 +37,7 @@ define Package/byobu
 $(call Package/byobu/Default)
   TITLE:=Text-based window manager and terminal multiplexer
   DEPENDS:=+python3-light +python3-newt
+  select tmxu if !screen
```
jefferyto, line 40: This has to be inside of a config section (`define Package/byobu/config`) to work.

But more importantly, this is a directive for the build system; it doesn't translate into dependency information for opkg, which is what you want (and why this isn't specified in the makefile already).

jefferyto:
> From my understanding of https://openwrt.org/docs/guide-developer/dependencies#special_notes this should work.

The last paragraph of that section (emphasis mine): 

> It is important to note that using `select bar` in `Package/foo/config` directives will select `bar` for building, but **it will not be included in the dependencies when `opkg` installs `foo`**, nor will it guarantee that `bar` will be built before `foo`. Using `DEPENDS` will take care of all that.

BKPepe:
Anything new, @ngehrsitz ? Otherwise this PR will be closed.

ngehrsitz:
I switched to the approach from https://github.com/openwrt/packages/commit/78b7bbbfa60bbe165b7b38025f2d5db887877191. I could verify that `tmux` is selected if `screen` is not present. Trying to test the other case with the [sdk container](https://github.com/openwrt/docker#sdk-example) failed with a compilation error for screen. But this error is also present without my changes so that appears to be unrelated.

jefferyto (commented):
You don't seem to understand my point in https://github.com/openwrt/packages/pull/20205#discussion_r1165282732 and https://github.com/openwrt/packages/pull/20205#issuecomment-1506671782.

If you copy your compiled ipk file onto a device and do `opkg install byobu.ipk` (replace with the actual file name), does it also install tmux or screen?

ngehrsitz:
> You don't seem to understand my point in [#20205 (comment)](https://github.com/openwrt/packages/pull/20205#discussion_r1165282732) and [#20205 (comment)](https://github.com/openwrt/packages/pull/20205#issuecomment-1506671782).
> 
> If you copy your compiled ipk file onto a device and do `opkg install byobu.ipk` (replace with the actual file name), does it also install tmux or screen?

You are correct, I tested some more and opkg does not pick up such complex configurations. Since the build constraint is not that useful to have I will close this.

PR #20206, opened 2023-01-03
dump1090: realigned dump1090.config and dump1090.init switches

darkglade:
Signed-off-by: Morgan Darkglade <openwrt@darkglade.com>

Maintainer: @Noltari
Compile tested: ramips (mipsel_24kc) HEAD (a7b770e)
Run tested: ramips (mipsel_24kc) HEAD (a7b770e)

Description:
Updated dump1090.config and dump1090.init to align with the command line switches for the in-tree version of dump1090 removing support for several ignored/obsolete switches.

This should resolve failures to start the dump1090 service which originate from the obsoleted --html-dir switch being passed to dump1090 when it is started via the init/rc system.

A companion PR will be forthcoming for luci-app-dump1090 to align it with these changes.


utils/dump1090/files/dump1090.config
```diff
@@ -2,49 +2,43 @@ config dump1090 main
...
+	option net_ro_size ''
+	option net_ro_interval ''
 	option net_sbs_port ''
 	option net_bi_port ''
 	option net_bo_port ''
-	option net_ro_size ''
-	option net_ro_interval ''
-	option net_heartbeat ''
+	option net_stratux_port ''
+    option net_heartbeat ''
```
hnyman, line 35: spaces instead of a tab.
The same also once in the init file.

darkglade: my vimrc must be off, should be fixed by the new commit.

darkglade: OK, this time for sure...

darkglade: Third time's a charm (hopefully)

darkglade:
OK, I've no idea what's going on with that Test Formalities test, the subject line DOES start with the package name...

darkglade:
Skimming through the output from that failed task it appears to be an upstream failure (base/ucode failing to build on PPC8540) so not much to be done on this end.

GeorgeSapkin:
@darkglade if this is still relevant, I'd say it needs a `PKG_RELEASE` bump and `Signed-off-by` is usually the last bit of the commit message.

openwrt-ai (commented):
## Commit checks

Reviewed 2 new commits. Both messages match their changes: commit 1 realigns the config/init switch order and drops obsolete switches; commit 2 only converts leading spaces to tabs on the two `net_stratux_port` lines. The reordered `dump1090.config` and `dump1090.init` stay in exact option-for-option parity, and the removed switches (`html_dir`, `no_crc_check`, `phase_enhance`, `debug`, `oversample`, `dcfilter`, `measure_noise`) are consistent with the stated fix, so the realignment itself is internally sound.

Two items before merge:

- **`PKG_RELEASE` bump missing.** The change touches installed files (`files/dump1090.init` and the `/etc/config/dump1090` conffile) but `utils/dump1090/Makefile` still has `PKG_RELEASE:=1`, so the package version does not increment and existing installs won't receive the fix via opkg/apk. Bump `PKG_RELEASE`. (Consistent with @GeorgeSapkin's earlier note.)
- **nit: squash the fixup.** Commit 2 is a pure whitespace correction of lines introduced in commit 1; consider folding it into commit 1 so the series lands as a single clean commit.

---

PR #20212, opened 2023-01-03
lighttpd: update to lighttpd 1.4.68 release hash

gstrauss:
Maintainer: @flyn-org

Description:
* update to lighttpd 1.4.68 release hash
* modify build cmd for type: feature opts


gstrauss:
`Makefile` properly declares dependency on libmaxminddb
`$(eval $(call BuildPlugin,maxminddb,MaxMind DB,+PACKAGE_lighttpd-mod-maxminddb:libmaxminddb,30))`

neheb:
```
../../../home/build/openwrt/build_dir/target-arc_archs_glibc/lighttpd-1.4.68/src/meson.build:718:0: ERROR: File t/test_mod_evasive.c does not exist.
```

I get this locally too.

neheb:
I changed maxminddb upstream to use dependency() instead of find_library(). The former requires a pkgconfig file, which is probably not being installed.

gstrauss:
`ERROR: File t/test_mod_evasive.c does not exist.`

Yes, I saw that, too.  For some reason `git format-patch` produced something which the openwrt CI thinks is a dirty patch.  I'll try to fix that up.

> I changed maxminddb upstream to use dependency() instead of find_library(). The former requires a pkgconfig file, which is probably not being installed.

What do you recommend I do for lighttpd?  I added a commit which disables maxminddb so that I could fix up any other errors, but I think we should fix maxminddb detection rather than disabling it.  Should I modify lighttpd src/meson.build to use find_library() or should I look into modifying libmaxminddb to install .pc file, if that is even an option?

neheb:
after https://github.com/openwrt/packages/pull/20215 is merged maxminddb should work again.

gstrauss:
> after https://github.com/openwrt/packages/pull/20215 is merged maxminddb should work again.
> edit: dirty patches are fixed with make package/lighttpd/{clean,refresh}

Thanks!  I'll refreshed patch and re-push.

gstrauss:
Apparently quilt is not smart enough to ignore git format-patch annotations.  Note the header and footer removed:
```
--- a/net/lighttpd/patches/020-meson-tests.patch
+++ b/net/lighttpd/patches/020-meson-tests.patch
@@ -7,8 +7,6 @@ Subject: [PATCH] [meson] remove t/test_mod_evasive.c
  src/meson.build | 1 -
  1 file changed, 1 deletion(-)
 
-diff --git a/src/meson.build b/src/meson.build
-index ff472c32..77089fc3 100644
 --- a/src/meson.build
 +++ b/src/meson.build
 @@ -721,7 +721,6 @@ test('test_mod', executable('test_mod',
@@ -19,6 +17,3 @@ index ff472c32..77089fc3 100644
                't/test_mod_evhost.c',
                't/test_mod_indexfile.c',
                't/test_mod_simple_vhost.c',
--- 
-2.39.0
-
```

neheb:
I doubt it can ever be taught that.

gstrauss:
New challenge:
```
Collected errors:
 * check_data_file_clashes: Package lighttpd-mod-staticfile wants to install file /usr/lib/lighttpd/mod_staticfile.so
	But that file is already provided by package  * lighttpd
 * opkg_install_cmd: Cannot install package lighttpd-mod-staticfile.
```
lighttpd 1.4.68 release notes include:
```
lighttpd 1.4.68 builds common modules into the lighttpd base executable.
Separate dynamic modules are still built for the benefit of existing
packaging scripts in various distributions, but those modules are not used.
A future version of lighttpd will omit building separate modules for:
mod_access mod_alias mod_evhost mod_expire mod_fastcgi mod_indexfile
mod_redirect mod_rewrite mod_scgi mod_setenv mod_simple_vhost mod_staticfile
```

@neheb do you have any suggestions on how best to proceed?  Should I remove those packages?  That might trip up people who install those packages explicitly.  Should I create stub `.c` files for those packages so that the packages continue to exist (for now)?  Other suggestions?

neheb:
I'd ask @BKPepe .

gstrauss:
This might be newly detected by the openwrt build, even though it has been there for a while.  The `Makefile`  sets `BASE_MODULES:=dirlisting indexfile staticfile` and ships those modules in the `lighttpd` package.  Of those three, only staticfile also had a rule to make package `lighttpd-mod-staticfile`.  I am going to try removing the `lighttpd-mod-staticfile` package.

gstrauss:
@BKPepe @flyn-org @neheb this passes CI and is ready for review. :)  Thanks.

flyn-org:
Looks good.

PR #20215, opened 2023-01-04
libmaxminddb: install pkgconfig file

neheb:

Maintainer: @ja-pa 

ping @gstrauss


BKPepe:
We should backport this to the stable releases.

neheb:
Backported to 22.03.

hnyman:
@neheb  @BKPepe 
I fixed the fallout from this commit for "bind" both in master and 22.03.

But looking at the 22.03 (and master) faillogs shows that this libmaxminddb change also broke knot and ocserv both in master and 22.03.

PR #20220, opened 2023-01-05
tvheadend: Fix ipv6 URL support 

sKyissKy:
Compile tested: x86_64
Run tested: x86_64

Description: recover tvheadend ipv6 IPTV URL support

### DIFF
--- a/libs/uriparser/Makefile
+++ b/libs/uriparser/Makefile
@@ -1,0 +1,56 @@
+#
+# Copyright (C) 2015-2018 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=uriparser
+PKG_VERSION:=0.9.7
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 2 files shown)


libs/uriparser/Makefile
```diff
@@ -0,0 +1,56 @@
...
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=uriparser
+PKG_VERSION:=0.9.7
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/uriparser/uriparser/releases/download/$(PKG_NAME)-$(PKG_VERSION)?
```
zdz2019, line 15: ? should be changed to /

Otherwise a 404 error will occur.

libs/uriparser/Makefile
```diff
@@ -0,0 +1,56 @@
...
+PKG_HASH:=11553b2abd2b5728a6c88e35ab08e807d0a0f23c44920df937778ce8cc4d40ff
+
+PKG_MAINTAINER:=Weijia Song <songweijia@gmail.com> & Sebastian Pipping <sebastian@pipping.org>
```
hnyman, line 18: The ampersand & should be changed to to comma "," so that the field values can be passed to mail clients etc.

sKyissKy: I don't know how to modify the pull request, I modify it directly on the web page.

PR #20222, opened 2023-01-05
[22.03] syncthing: update to 1.23.0

brvphoenix:
Backport #20221.


BKPepe:
In the commit description, there should be said that it is cherry picked.

PR #10745, opened 2019-12-10
stubby: switch to ca-bundle

mstorchak:
Signed-off-by: Maxim Storchak <m.storchak@gmail.com>

Maintainer: @jonathanunderwood 
Compile tested: ath97, WNDR3800, r11638+3-3ee767086d
Run tested: ath97, WNDR3800, r11638+3-3ee767086d. stubby runs just fine, I've been using this setup for about a year with no issues.

Description:
ca-bundle provides the same set of CA certificates as ca-certificates, but is more compact in terms of filesystem space.

PR #20236, opened 2023-01-07
stubby: actually switch to ca-bundle

ghost:
Fixes commit 11731c81c35b6271c92097cb4edf4259aa45e376.

Signed-off-by: Aquila Cooper <aquila@cpr.is>

### DIFF
--- a/net/stubby/Makefile
+++ b/net/stubby/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=stubby
 PKG_VERSION:=0.4.3
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


hnyman:
> Fixes commit https://github.com/openwrt/packages/commit/11731c81c35b6271c92097cb4edf4259aa45e376

As discussed in the PR #10745 about that commit, the "ca-certs" is meant as a flexible alternative that enables the user to use either ca-certificates or ca-bundle, as some other package might explicitly require one of them and you shouldn't be forced to install both if not necessary. So, stubby accepts either of them.

The flexibility is provided by https://github.com/openwrt/openwrt/commit/dd299805ad18472a8245b4524a25e4381e166057

You proposal would force to use ca-bundle.

NAK from me.

PR #20239, opened 2023-01-07
shadowsocks-libev: Add server IPs to dst_bypass set, fix looping redirect issue

luisliou:

Description:
shadowsocks server IPs were not added to the bypass set, this would cause a looping redirection for ss_redir service, this PR would fix this issue.


1715173329:
Maintainer @yousong

yousong:
Applied with "minor fixup on commit title, version bump, etc.".  Thank you for your contribution!

PR #20244, opened 2023-01-08
ddns-scripts: add strato ipv6 support

stargieg:
Maintainer: ?
Compile tested: x86 ath79 ramips
Run tested: x86 ath79 ramips

Description:
ddns-scripts: add strato ipv6 support


feckert:
thanks

PR #20268, opened 2023-01-11
gcc: add support for gcc 12

Ansuel:
Copy patch from gcc 11 to gcc 12 and add missing hash.

---

@hnyman lets see... we asked for gcc 12 support ages ago and here we are....


hnyman:
CI will fail for the AUTORELEASE  that you did not remove.

Ansuel:
@hnyman are you aware if tests try to test compile even if patch are not refreshed?

hnyman:
The refresh test is done at the end of the compilation tests, I think.
The tests are first run, and then will finally fail at the end due to dirty patches.

(I thought  that you might actually separate it to a separate item, but some of the patches might be architecture specific)

Ansuel:
Ok in theory now this should be correct

Ansuel:
For some reason gcc package fail in the postrm... But that was already there in old gcc 11 version so nothing new...

PR #20275, opened 2023-01-12
efivarfs-mount: new package

oskarirauta:

Description:
simple init.d mount script to mount efivarfs on boot

### DIFF
--- a/utils/efivarfs-mount/Makefile
+++ b/utils/efivarfs-mount/Makefile
@@ -1,0 +1,31 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=efivarfs-mount
+PKG_VERSION:=1.0
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=OSkari Rauta <oskari.rauta@gmail.com>
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/efivarfs-mount
+  SECTION:=utils
+  CATEGORY:=Utilities
+  SUBMENU:=Filesystem
+  TITLE:=efivarfs mouting script
+  DEPENDS:=@TARGET_x86_64 +mount-utils +kmod-fs-efivarfs
+endef
+
+define Package/efivarfs-mount/description
+  Mounts efivarfs automaticly on boot as a init script
+endef
+
+define Build/Compile
+endef
+
+define Package/efivarfs-mount/install
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_BIN) ./files/efivarfs $(1)/etc/init.d/
+endef
+
+$(eval $(call BuildPackage,efivarfs-mount))
--- a/utils/efivarfs-mount/files/efivarfs
+++ b/utils/efivarfs-mount/files/efivarfs
@@ -1,0 +1,21 @@
+#!/bin/sh /etc/rc.common
+# Copyright (C) 2014 OpenWrt.org
+
+START=99
+
+start() {
+	[ -d /sys/firmware/efi/efivars ] && {
+		findmnt -n --pseudo --mountpoint /sys/firmware/efi/efivars > /dev/null || {
+			echo Mounting efivarfs
+			mount -t efivarfs none /sys/firmware/efi/efivars
+		}
+	}
+}
+
+stop() {
+	[ -d /sys/firmware/efi/efivars ] && {
+		findmnt -n --pseudo --mountpoint /sys/firmware/efi/efivars > /dev/null && {
+			umount /sys/firmware/efi/efivars
+		}
+	}
+}


oskarirauta:
@1715173329

Please, take a look at this. Simple init script.

1715173329:
I'm sorry that I'm not familiar with efivarfs, but the script itself lgtm.

oskarirauta:
@1715173329 

efivarfs is a pseudo filesystem (like debugfs for example), which contains variables from efi firmware. With these, it's possible to modify settings, for example, change boot order. Manipulating efi generally requires mounted efivarfs - as it has come a pretty standard way to do it.

Yes, totally different things- this indeed is useful with my other efi related packages, but not requirement at all, one can just mount efivarfs by hand if he wants to. This is just a simple script to mount efivarfs on boot time.

But those failing tests... What is going on? How can this script fail, as in the end, it is just a shell script without real linked depends, etc.. (I made it depend only on what it requires to be working)

1715173329:
> efivarfs is a pseudo filesystem (like debugfs for example), which contains variables from efi firmware. With these, it's possible to modify settings, for example, change boot order. Manipulating efi generally requires mounted efivarfs - as it has come a pretty standard way to do it.

Thank you so much!

> But those failing tests... What is going on? How can this script fail, as in the end, it is just a shell script without real linked depends, etc.. (I made it depend only on what it requires to be working)

Please just ignore them.
This package is only available on TARGET_x86_64, so no ipk will be generated on other targets. And because of this, ci runs into "no such file or directory" error.

oskarirauta:
@1715173329 

I was wondering that maybe it's something like that, but error message was not quite clear to me.. Thanks.

Maybe then review....? :)

dangowrt:
Imho this should be done by procd, then we won't need an extra package, on EFI systems it is reasonable to expect efivarfs tp be mounted on `/sys/firmware/efi/efivars`.

Can you try this patch for procd instead, please it in `package/system/procd/patches` (you have to create that folder):
```diff
--- a/initd/early.c
+++ b/initd/early.c
@@ -58,13 +58,14 @@ early_mounts(void)
 	unsigned int oldumask = umask(0);
 
 	if (!is_container()) {
-		mount("proc", "/proc", "proc", MS_NOATIME | MS_NODEV | MS_NOEXEC | MS_NOSUID, 0);
-		mount("sysfs", "/sys", "sysfs", MS_NOATIME | MS_NODEV | MS_NOEXEC | MS_NOSUID, 0);
+		mount("proc", "/proc", "proc", MS_NOATIME | MS_NODEV | MS_NOEXEC | MS_NOSUID, NULL);
+		mount("sysfs", "/sys", "sysfs", MS_NOATIME | MS_NODEV | MS_NOEXEC | MS_NOSUID, NULL);
+		mount("efivars", "/sys/firmware/efi/efivars", "efivarfs", MS_NOATIME | MS_NODEV | MS_NOEXEC | MS_NOSUID, NULL);
 		mount("cgroup2", "/sys/fs/cgroup", "cgroup2",  MS_NODEV | MS_NOEXEC | MS_NOSUID | MS_RELATIME, "nsdelegate");
 		mount("tmpfs", "/dev", "tmpfs", MS_NOATIME | MS_NOEXEC | MS_NOSUID, "mode=0755,size=512K");
 		ignore(symlink("/tmp/shm", "/dev/shm"));
 		mkdir("/dev/pts", 0755);
-		mount("devpts", "/dev/pts", "devpts", MS_NOATIME | MS_NOEXEC | MS_NOSUID, 0);
+		mount("devpts", "/dev/pts", "devpts", MS_NOATIME | MS_NOEXEC | MS_NOSUID, NULL);
 
 		early_dev();
 	}
```

oskarirauta:
@dangowrt thanks Daniel, indeed that would be a better solution... Hope it comes to mainstream soon enough, although, I guess there needs to be some checking that system supports it on procd size, most likely on run-time, as afaik, only x86-64 has EFI, but if they weren't booted with EFI, efivarfs cannot be mounted.

dangowrt:
EFI is also used on some Aarch64 platforms and (in theory) could also be used on 32-bit ARM.
In the patch I suggested, procd attempts to mount efivarfs and doesn't care if that succeeds or not as the return value of the mount syscall is ignored. Hence this is fine and will not disrupt non-EFI systems, but just silently result in not having anything in `/sys/firmware/efi/efivars`.

Did you test if this patch fixes mounting efivarfs on boot? I don't have any EFI system to (easily) test this atm.

oskarirauta:
I am going to test it, but my build is still on-going, as it's rather large scale build- nearly all (about 8 packages cannot be compiled no matter what I do to fix them, 5 of these are variants, numbers are approx), with extended attributes to max/near max functionality, with full iconv, custom kernel options + own custom kernel patches (just prettifies some boot output..):

current build stats:
```
[user@buildroot /usr/src/openwrt]$ pkgcount
1.8G total
5792 files
[user@buildroot /usr/src/openwrt]$

```

about 5 packages still un-compiled, some of their builds are extremely slow (hwinfo, left that to last for a reason..) and I just refreshed to 5.15.88 version of kernel which is building, so testing is about to happen in tomorrow morning if everything goes as planned. About 30-35 packages have also been modified to build correctly with gcc 12 + full iconv, as there's issues with that kind of setup with soooo many packages. Necessary fix is to just include nls.mk and add -liconv to TARGET_LDFLAGS, evend depends can stay as-is, but about 3 packages have sort of problems that are difficult to identify.

But I am very confident that patch produces good results. 

Sidenote: If those mentioned other than x86-64 architechtures (in theory) really support EFI, there **must** be something else as bootloader, than uboot and other similar solutions. Well actually, boot must in that case be handled- with (U)EFI firmware. I really like EFI in many ways- but it's very far yet from standard, I mean, I have seen so many of those, I have x86-64 routers with different kinds of EFI firmwares, some lack "shell" completely, luckily it's available as EFI binary, but some have it as a in baked boot option. Some have it as a boot option, but actual PE32 code lacks from firmware and "starting shell" causes error: not found. And propably this is just a tip of mountain from differences between systems. First time I got familiar with it, was ofcourse Macs.. Which always worked flawlessly, but I am very pleased to see EFI on PC side so often and how it is becoming standard. For most of us, it's just a system that allows one to have multiple different boot loaders simultaneously (and in some cases, you can chainload them if feature is supported), but there are some nice tools too which used to require a DOS boot disk, which might be tricky these days, like BIOS (bios+efi) updates. Or ami's dmiedit.... Apparently there's also version of gdisk, but I haven't tried it ever..

And about those differencies: I would recommend to avoid bootloader called clover- it's something between refind and opencore, but it's risky- there's about 50% risk that it won't boot. On a virtual environment (ESXi, VMWare Fusion on Mac); there's 100% risk that it does not boot. But on some systems it does, this is where EFI firmware's differ these days, and stay out of certain standards. As for the end-user it seems like a starter of boot manager, it's kinda crucial that all common ones would work on all manufacturers and versions; think of it as linux command ls wouldn't work on some hardware.....

But if you wish, I will report on this thread how it went as soon as I have tested it (leaving compilation to run over night..), and close this PR with that report..

oskarirauta:
Oh, virtual efi (duet) is also tricky, and unfortunately forgotten- it was great starter of virtual EFI which actually on some cases even worked. There is ONE build of clover ISO that boots in BIOS mode even on virtual environment (r4769) at https://sourceforge.net/projects/cloverefiboot/files/Bootable_ISO/ but unfortunately has PS/2 + ISA support removed, so it's not very useful. All others fail, that build actually starts as supposed to (without input support :( ) - I've been trying to build one by myself, but unfortunately it seems impossible to replicate it. If I could, I would re-enable legacy input support and could have same EFI boot manager for bios and EFI. And as a success, I would have superior compatibility, best so far of what I have found. That would be really cool. But unfortunately, world seems to think that it's not necessary/useful, as bios is dead... with bios being dead, I ofcourse mean just bios booting, but so often people don't realize that pc's usually (unlike macs) have bios anyway, there's just EFI firmware that is booted with BIOS, and that EFI is a small (DOS-like in my opinion) mini OS, used mainly to boot what you wanted (but there are some tools available as well, and there are ways for anyone to build some new ones also)

oskarirauta:
btw. some packages that fail to compile:
 - syslog-ng
 - jose, libjose, tang (all linked together, and jose or libjose failed to comply)
 - coova-chilli's kernel module, therefor, also coova-chilli was left out, even though it won't depend on that module
 - gensio, when python3-gensio is enabled (rest of gensio builds just fine)
 - avahi's libavahi-compat-libdnssd fails from time to time, sometiems succeeds- unknown reason. So shairplay not compiled as it depends on compat libavahi, and ofcourse luci-app-shairplay, as it depends on shairplay
 - dovecot
 - baresip
 - byobu

All other packages were compiled, some with variants, but mostly without- if choosable, always the most full featured set, or ssl enabled version, always when possible to choose: with openssl, compiled with crypto device support. Left out packages: all example/test packages. Without SELinux is exception. Packages built from provided sources, some patched with not yet approved PRs, and some patched/changed by me to enable functionality not provided by default, then some packages from my own feeds (includes mostly what I have sent PR for + some extra for own use..)

Built on a docker container running on TrueNAS Scale machine - image source is available at here: https://github.com/oskarirauta/openwrt-buildroot

It is designed for my own build environment. I used to use TrueNAS CORE, but decided to switch to scale to get containerization support as I can share resources much more efficiently than doing the same thing in a VM- also container deployment is pretty fast, so when all goes south with build environment... Easy to renew it.

Sidenote: building PERL was a real pain in the ass, I bet it has something to do with alpine as building host. Finally I got it to work, but it wasn't pretty. It looked for files from all the wrong places and refused to compile them and then some- but I worked that out. Also bunch of others I had to change due to -liconv problem mentioned before.

Built target is x86-64, rootfs.tar.gz, somewhat modified kernel opts, extra patches for kernel, modified busybox setup, in general, openwrt with all the bells and whistles. Really shouldn't had seen the trouble to build all that I built. As I propably don't use more than 2 percent of it. But for some reason, I wanted to go through it. All listed packages that refused to build on realistic amount of effort to fix the issues, are packages that I have no use at all personally. This report is definetly in wrong place, but atleast it's now in place at somewhere :) - Hopefully there's some good of listing problematic builds for the future generations to come ;)

Last build running before I am ready to make your test. When I wake up.

oskarirauta:
That pkgcount function by the way is not accurate, but it does not need to be for me.

@dangowrt

I can verify that your patch works as expected. efibootmgr immediately after boot is able to display boot entries, and efivarfs is mounted properly. I can recommend and approve your patch for mainstream :) - Make a PR or something :)

oskarirauta:
5800pkg's built and efivarfs mount via procd works perfectly on target device.

PR #20276, opened 2023-01-12
efivar: new package

oskarirauta:
Maintainer: me / @oskarirauta 

Description:

efivar library. Required by my another PR #20277 (efibootmgr)

### DIFF
--- a/libs/efivar/Makefile
+++ b/libs/efivar/Makefile
@@ -1,0 +1,66 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=efivar
+PKG_VERSION:=38
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
+PKG_SOURCE_URL:=https://github.com/rhboot/efivar/releases/download/$(PKG_VERSION)
+PKG_HASH:=f018ed6e49c5f1c16d336d9fd7687ce87023276591921db1e49a314ad6515349
+
+PKG_LICENSE:=LGPL-2.1-only
+PKG_LICENSE_FILES:=COPYING
+
+PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/efivar
+  SECTION:=libs
+  CATEGORY:=Libraries
+  TITLE:=Tools and libraries to work with EFI variables
+  DEPENDS:=@TARGET_x86_64
+  URL:=https://github.com/rhboot/efibootmgr
+endef
+
+define Package/efivar/description
+  Tools and libraries to work with EFI variables
+endef
+
+MAKE_VARS += \
+	ERRORS= \
+	HOSTCC="$(HOSTCC)" \
+	HOST_CFLAGS="$(HOST_CFLAGS)" \
+	HOST_LDFLAGS="$(HOST_LDFLAGS)" \
+	LIBDIR="/usr/lib"
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefiboot.so* $(1)/usr/lib/
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefisec.so* $(1)/usr/lib/
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefivar.so* $(1)/usr/lib/
+
+	$(INSTALL_DIR) $(1)/usr/include/efivar
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/efivar/*.h $(1)/usr/include/efivar/
+	$(INSTALL_DIR) $(1)/usr/lib/pkgconfig/
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/pkgconfig/*.pc $(1)/usr/lib/pkgconfig/
+endef
+
+define Package/efivar/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/efisecdb $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/efivar $(1)/usr/bin/
+
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefiboot.so* $(1)/usr/lib/
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefisec.so* $(1)/usr/lib/
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefivar.so* $(1)/usr/lib/
+endef
+
+$(eval $(call BuildPackage,efivar))
--- a/libs/efivar/patches/002-musl-compat.patch
+++ b/libs/efivar/patches/002-musl-compat.patch
@@ -1,0 +1,199 @@
+From cece3ffd5be2f8641eb694513f2b73e5eb97ffd3 Mon Sep 17 00:00:00 2001
+From: Natanael Copa <ncopa@alpinelinux.org>
+Date: Fri, 28 Jan 2022 12:13:30 +0100
+Subject: [PATCH 1/2] efisecdb: fix build with musl libc
+
+Refactor code to use POSIX atexit(3) instead of the GNU specific
+on_exit(3).
+
+Resolves: #197
+Resolves: #202
+Signed-off-by: Natanael Copa <ncopa@alpinelinux.org>
+
+--- a/src/compiler.h
++++ b/src/compiler.h
+@@ -7,8 +7,6 @@
+ #ifndef COMPILER_H_
+ #define COMPILER_H_
+ 
+-#include <sys/cdefs.h>
+-
+ /* GCC version checking borrowed from glibc. */
+ #if defined(__GNUC__) && defined(__GNUC_MINOR__)
+ #  define GNUC_PREREQ(maj,min) \
+--- a/src/efisecdb.c
++++ b/src/efisecdb.c
+@@ -25,6 +25,10 @@
+ extern char *optarg;
+ extern int optind, opterr, optopt;
+ 
++static efi_secdb_t *secdb = NULL;
++static list_t infiles;
++static list_t actions;
++
+ struct hash_param {
+ 	char *name;
+ 	efi_secdb_type_t algorithm;
+@@ -187,12 +191,11 @@ add_action(list_t *list, action_type_t a
+ }
+ 
+ static void
+-free_actions(int status UNUSED, void *actionsp)
++free_actions(void)
+ {
+-	list_t *actions = (list_t *)actionsp;
+ 	list_t *pos, *tmp;
+ 
+-	for_each_action_safe(pos, tmp, actions) {
++	for_each_action_safe(pos, tmp, &actions) {
+ 		action_t *action = list_entry(pos, action_t, list);
+ 
+ 		list_del(&action->list);
+@@ -202,12 +205,11 @@ free_actions(int status UNUSED, void *ac
+ }
+ 
+ static void
+-free_infiles(int status UNUSED, void *infilesp)
++free_infiles(void)
+ {
+-	list_t *infiles = (list_t *)infilesp;
+ 	list_t *pos, *tmp;
+ 
+-	for_each_ptr_safe(pos, tmp, infiles) {
++	for_each_ptr_safe(pos, tmp, &infiles) {
+ 		ptrlist_t *entry = list_entry(pos, ptrlist_t, list);
+ 
+ 		list_del(&entry->list);
+@@ -216,27 +218,12 @@ free_infiles(int status UNUSED, void *in
+ }
+ 
+ static void
+-maybe_free_secdb(int status UNUSED, void *voidp)
++maybe_free_secdb(void)
+ {
+-	efi_secdb_t **secdbp = (efi_secdb_t **)voidp;
+-
+-	if (secdbp == NULL || *secdbp == NULL)
++	if (secdb == NULL)
+ 		return;
+ 
+-	efi_secdb_free(*secdbp);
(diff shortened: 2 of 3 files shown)


libs/efivar/Makefile
```diff
@@ -0,0 +1,67 @@
...
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
+PKG_SOURCE_URL:=https://github.com/rhboot/efivar/releases/download/$(PKG_VERSION)
+PKG_HASH:=f018ed6e49c5f1c16d336d9fd7687ce87023276591921db1e49a314ad6515349
+
+PKG_LICENSE:=LGPL-2.1-only
+PKG_LICENSE_FILES:=COPYING
+PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=1
+
+PKG_MAINTAINER:=OSkari Rauta <oskari.rauta@gmail.com>
```
1715173329, line 16:
```suggestion
PKG_LICENSE:=LGPL-2.1-only
PKG_LICENSE_FILES:=COPYING
PKG_MAINTAINER:=OSkari Rauta <oskari.rauta@gmail.com>

PKG_BUILD_PARALLEL:=1
PKG_INSTALL:=1
```

libs/efivar/files/efivar-guids.h
```diff
@@ -0,0 +1,80 @@
+#ifndef EFIVAR_GUIDS_H
```
1715173329, line 1: where do these files come from? can't them be upstreamed?

libs/efivar/patches/001-workarounds.patch
```diff
@@ -0,0 +1,28 @@
...
+ src/include/workarounds.mk | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+--- a/src/include/workarounds.mk
++++ b/src/include/workarounds.mk
+@@ -4,12 +4,12 @@
+ 
+ LD_FLAVOR := $(shell $(LD) --version | grep -E '^(LLD|GNU ld)'|sed 's/ .*//g')
+ LD_VERSION := $(shell $(LD) --version | grep -E '^(LLD|GNU ld)'|sed 's/.* //')
+-# I haven't tested 2.36 here; 2.35 is definitely broken and 2.37 seems to work
```
1715173329, line 18: we're at 2.39 already, is it still required?

libs/efivar/patches/002-musl-compat.patch
```diff
@@ -0,0 +1,187 @@
+--- a/src/compiler.h
```
1715173329, line 1: I found this patch at https://git.alpinelinux.org/aports/tree/main/efivar/musl-compat.patch.
Please keep his patch header here.

1715173329 (changes requested):

libs/efivar/files/efivar-guids.h
oskarirauta: normal linux efivar headers....

libs/efivar/patches/001-workarounds.patch
oskarirauta: You are correct :)

libs/efivar/patches/002-musl-compat.patch
oskarirauta: Sure. Header was lost during patch refreshing for quilt compatibility (dirty patches...)

libs/efivar/patches/001-workarounds.patch
oskarirauta: Worked on my end without patch. I removed it.

libs/efivar/files/efivar-guids.h
oskarirauta: They came from somewhere - but I cannot find them now afterwards from anywhere. And I though that they were needed for a reason.. Maybe they were for older version of efivar, hard to say. I removed them and patches equipping them.

oskarirauta:
It says that 1 change is requested, but I can't find it from anywhere.... What is it?

1715173329:
> It says that 1 change is requested, but I can't find it from anywhere.... What is it?

It just meant somebody left a review.

oskarirauta:
> > It says that 1 change is requested, but I can't find it from anywhere.... What is it?
> 
> It just meant somebody left a review.

Sometimes, when I don't press "Resolve" **before** pushing commit, platform seems to remember that there's a request for changes, but it becomes inaccessible since content has changed, it's propably that then...

1715173329:
Nah.

oskarirauta:
Could you verify if this is ok. I accidentally seem to have removed request for neheb's review.

@neheb - please, review.

neheb (approved):

oskarirauta:
I can't remember where I got those files- but seems that more or less they are needed. Problem with finding out where I got those headers, is that I created this package for personal use ages ago, had difficulties to build it and found out that I needed files that I included- but source... Well, it seems to have disappeared..

It is rather difficult to "test" this locally, as I build on a system that is compatible with target (alpine x86-64)- so most built software can be executed on a host- then there occurs issues when I push the commit, as build environment is different. Sorry about that, I'll try to get this to build properly again.

oskarirauta:
Ok, now it should be able to build again.

@1715173329 now you might want to ask again where did these files come from- answer is that there is util included: makeguids - which generates these files. Ofcourse it will be cross-compiled, so there is no way to use it on the host system, except in my case since my host system has no trouble to execute it. So, I generated those files and added them manually and removed generation of files from Makefile with patch 007.

missing qsort_r patch and files- I did not add them, as they are not necessary (anymore), like I said; I first added this package some time ago, and at that time, musl did not have qsort_r, so I found a patch and these files from somewhere, and then added them; now we run with newer musl, which has qsort_r, so that problem is in the past. Propably these files are still in somewhere, but google can't find them- I just found other issues with similar/same solution conversations on some github projects, which revealed history behind the solution.

oskarirauta:
Okay, now it's fixed. I'll rebase and then this is ready for reviewing.

1715173329:
> Ofcourse it will be cross-compiled, so there is no way to use it on the host system, except in my case since my host system has no trouble to execute it.

https://github.com/rhboot/efivar/blob/1753149d4176ebfb2b135ac0aaf79340bf0e7a93/src/Makefile#L62-L69
it calls host gcc to build - in theory it should work.

oskarirauta:
> > Ofcourse it will be cross-compiled, so there is no way to use it on the host system, except in my case since my host system has no trouble to execute it.
> 
> https://github.com/rhboot/efivar/blob/1753149d4176ebfb2b135ac0aaf79340bf0e7a93/src/Makefile#L62-L69 it calls host gcc to build - in theory it should work.

Yes. In theory. Still in real life it failed on checks here, because it was not possible to execute makeguids.

oskarirauta:
@dangowrt please, take a look at this you too. As you were more familiar with EFI.

oskarirauta:
@1715173329 
@dangowrt 

What would it take to get this merged?

dangowrt (changes requested):
Imho it's better to generate host-builds for what ever is needed to generate `efivar-guids.h` from source instead of carrying a built version in the tree. Ie. add a host-build like it is done for e.g. for postgresql.

oskarirauta:
@dangowrt

> Imho it's better to generate host-builds for what ever is needed to generate `efivar-guids.h` from source instead of carrying a built version in the tree.

ccdf3c6f18475a48b1aab08f7b9585d15f0064bc should do it..

oskarirauta:
@dangowrt 

I have updated accordingly to your request. Would you review this again with these changes made?

libs/efivar/Makefile
```diff
@@ -0,0 +1,88 @@
...
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
+PKG_SOURCE_URL:=https://github.com/rhboot/efivar/releases/download/$(PKG_VERSION)
+PKG_HASH:=f018ed6e49c5f1c16d336d9fd7687ce87023276591921db1e49a314ad6515349
+
+PKG_LICENSE:=LGPL-2.1-only
+PKG_LICENSE_FILES:=COPYING
+PKG_BUILD_PARALLEL:=1
+PKG_BUILD_DEPENDS:=efivar/host
+PKG_INSTALL:=1
```
1715173329, line 20:
```suggestion

PKG_BUILD_PARALLEL:=1
PKG_BUILD_DEPENDS:=efivar/host
PKG_INSTALL:=1
```

libs/efivar/Makefile
```diff
@@ -0,0 +1,88 @@
...
+
+include $(INCLUDE_DIR)/host-build.mk
```
1715173329, line 22:
```suggestion
```
just need to pass correct flags, host build can be removed.

libs/efivar/Makefile
```diff
@@ -0,0 +1,88 @@
...
+endef
+
+define Package/efivar/description
+  Tools and libraries to work with EFI variables
+endef
+
+HOST_MAKE_VARS += ERRORS=
+MAKE_VARS += \
+	MAKEGUIDS_BIN="$(STAGING_DIR_HOSTPKG)/usr/bin/efivar-makeguids" \
+	ERRORS=
```
1715173329, line 40:
```suggestion
MAKE_VARS += \
	ERRORS= \
	HOSTCC="$(HOSTCC)" \
	HOST_CFLAGS="$(HOST_CFLAGS)" \
	HOST_LDFLAGS="$(HOST_LDFLAGS)" \
	LIBDIR="/usr/lib"
```

libs/efivar/Makefile
```diff
@@ -0,0 +1,88 @@
...
+define Package/efivar/install
+	$(INSTALL_DIR) $(1)/usr/bin $(1)/usr/lib/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/{efisecdb,efivar} $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/lib/{libefiboot.so.1.38,libefisec.so.1.38,libefivar.so.1.38} $(1)/usr/lib/
+	$(LN) libefiboot.so.1.38 $(1)/usr/lib/libefiboot.so
+	$(LN) libefiboot.so.1.38 $(1)/usr/lib/libefiboot.so.1
+	$(LN) libefisec.so.1.38 $(1)/usr/lib/libefisec.so
+	$(LN) libefisec.so.1.38 $(1)/usr/lib/libefisec.so.1
+	$(LN) libefivar.so.1.38 $(1)/usr/lib/libefivar.so
+	$(LN) libefivar.so.1.38 $(1)/usr/lib/libefivar.so.1
```
1715173329, line 84:
```suggestion
	$(INSTALL_DIR) $(1)/usr/bin
	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/efisecdb $(1)/usr/bin/
	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/efivar $(1)/usr/bin/

	$(INSTALL_DIR) $(1)/usr/lib
	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefiboot.so* $(1)/usr/lib/
	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefisec.so* $(1)/usr/lib/
	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefivar.so* $(1)/usr/lib/
```

libs/efivar/patches/006-fix-lib-path.patch
```diff
@@ -0,0 +1,10 @@
+--- a/src/include/defaults.mk
++++ b/src/include/defaults.mk
+@@ -1,6 +1,6 @@
+ PREFIX	?= /usr
+ EXEC_PREFIX ?= $(PREFIX)
+-LIBDIR	?= $(PREFIX)/lib64
++LIBDIR	?= $(PREFIX)/lib
```
1715173329, line 7: replaced by `LIBDIR="/usr/lib"`. can be removed.

libs/efivar/patches/007-use-host-makeguids.patch
```diff
@@ -0,0 +1,19 @@
+--- a/src/Makefile
++++ b/src/Makefile
+@@ -31,6 +31,7 @@ GENERATED_SOURCES = include/efivar/efiva
+ MAKEGUIDS_SOURCES = makeguids.c util.c
+ MAKEGUIDS_OBJECTS = $(patsubst %.S,%.o,$(patsubst %.c,%.o,$(MAKEGUIDS_SOURCES)))
+ MAKEGUIDS_OUTPUT = $(GENERATED_SOURCES) guids.lds
++MAKEGUIDS_BIN ?= ./makeguids
```
1715173329, line 7: replaced by `HOSTCC="$(HOSTCC)" \`. can be removed.

1715173329 (changes requested):

libs/efivar/patches/006-fix-lib-path.patch
oskarirauta: patch removed

libs/efivar/patches/007-use-host-makeguids.patch
oskarirauta: patch removed

oskarirauta:
@1715173329 

That indeed was better solution than building a host package.

libs/efivar/Makefile
```diff
@@ -0,0 +1,67 @@
...
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/efisecdb $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/efivar $(1)/usr/bin/
+
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefiboot.so* $(1)/usr/lib/
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefisec.so* $(1)/usr/lib/
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libefivar.so* $(1)/usr/lib/
+endef
+
+$(eval $(call HostBuild))
```
1715173329, line 66: This should be removed too ;)

1715173329 (approved):
LGTM - Thank you!

libs/efivar/Makefile
oskarirauta: True. I'll remove that.. Just a second.

oskarirauta:
Okay, that should be all. Thanks @1715173329 - You really did put some effort it, it's now very nice and clean :)

PR #20278, opened 2023-01-12
podman: update to 4.3.1

oskarirauta:
Maintainer: me / @oskarirauta

Description:
Update podman to 4.3.1
catatonit is no longer required, works without.

[Release notes](https://github.com/containers/podman/blob/main/RELEASE_NOTES.md)

### DIFF
--- a/utils/podman/Makefile
+++ b/utils/podman/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=podman
-PKG_VERSION:=4.1.0
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_VERSION:=4.3.1
+PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://github.com/containers/podman/archive/v$(PKG_VERSION)
-PKG_HASH:=f814e12a7311d486c1ccdc4eb021bc6dd24499569de7a572e436342876f70e95
+PKG_HASH:=455c29c4ee78cd6365e5d46e20dd31a5ce4e6e1752db6774253d76bd3ca78813
 
 PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=LICENSE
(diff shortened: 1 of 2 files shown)


oskarirauta:
@1715173329 

I get a lot of these ci errors. Care to help me, as I have no idea what the problem might be..

1715173329:
https://stackoverflow.com/questions/58244095/gcc-7-error-unrecognized-command-line-option-m64

Looks like podman passed wrong cflags to the compiler.

1715173329:
> https://github.com/openwrt/packages/actions/runs/3900033826/jobs/6660255371#step:5:8211
> CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build \

So this is the cause. More work is needed if you want to use the upstream Makefile.

oskarirauta:
> https://stackoverflow.com/questions/58244095/gcc-7-error-unrecognized-command-line-option-m64
> 
> Looks like podman passed wrong cflags to the compiler.

Hey thanks, I'll look into it. Build process is a bit changed with this update, since I wanted to include rootless support to build as well and didn't know how to do it with go-lang.mk (how to build multiple targets) - so I used Makefile instead, which is a little tricky this time..

oskarirauta:
> > https://github.com/openwrt/packages/actions/runs/3900033826/jobs/6660255371#step:5:8211
> > CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build \
> 
> So this is the cause. More work is needed if you want to use the upstream Makefile.

Yes, I added bunch of vars from golang-package.mk - let's see if I got all necessary flags to succeed...

oskarirauta:
@1715173329 

Tough job- but I got it finally payed off :) - Atleast next update should be solved by current setup, so it should be much easier.

utils/podman/Makefile
```diff
@@ -82,14 +80,48 @@ else
...
+
 define Build/Prepare
 	$(call Build/Prepare/Default)
 	$(eval $(call Download,default-registries))
 	$(eval $(call Download,default-policy))
 endef
 
 define Package/podman/install
-	$(call GoPackage/Package/Install/Bin,$(1))
+	#$(call GoPackage/Package/Install/Bin,$(1))
```
dangowrt, line 121: Please remove this line rather than commenting it out.

oskarirauta: @dangowrt

Thanks, my bad :) I removed it. Please, review it again :)

dangowrt (approved):

PR #20283, opened 2023-01-12
autossh: return from instance function, not exit the script

ValdikSS:
In case of disabled configuration instances in the top of configuration file, enabled instances won't be started as the first disabled instance would result in init script termination.

Signed-off-by: ValdikSS <iam@valdikss.org.ru>

Maintainer: @bk138

### DIFF
--- a/net/autossh/Makefile
+++ b/net/autossh/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=autossh
 PKG_VERSION:=1.4g
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tgz
 PKG_SOURCE_URL:=https://www.harding.motd.ca/autossh/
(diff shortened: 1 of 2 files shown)


bk138:
Lgtm, though one of the openwrt guys should check this too.

1715173329:
maybe we should simply backport this commit d71bea3f19bd19488f9c36709dfdae4a4e8fad98?

ValdikSS:
> maybe we should simply backport this commit [d71bea3](https://github.com/openwrt/packages/commit/d71bea3f19bd19488f9c36709dfdae4a4e8fad98)?

It has its drawbacks, check the comment. Simple backport would result in broken autossh configuration as the new version changes configuration file, namely `localport` parameter is now required.

1715173329:
> > maybe we should simply backport this commit [d71bea3](https://github.com/openwrt/packages/commit/d71bea3f19bd19488f9c36709dfdae4a4e8fad98)?
> 
> It has its drawbacks, check the comment. Simple backport would result in broken autossh configuration as the new version changes configuration file, namely `localport` parameter is now required.

got it, thanks for clarification

ValdikSS:
Please also check https://github.com/openwrt/packages/pull/20269

PR #20284, opened 2023-01-13
mpd: update to 0.23.11

neheb:

Maintainer: nobody


cotequeiroz:
gcc12 is breaking this and udphp-client because C++ toolchain header `<tuple>` used to include `<array>`, but it does not anymore.  So any user of `std::array` will have to add `#include <array>`.

Backport https://github.com/MusicPlayerDaemon/MPD/commit/e4b055eb6d08c5c8f8d85828ce4005d410e462cb here; it works with the current package, and I don't see what could go wrong with the update.

PR #20287, opened 2023-01-13
elektra: get rid of pointless iconv patch

neheb:
A better solution was discovered.

Maintainer: @haraldg


neheb:
@Ansuel failure is with ucode. Can you take a look?

PR #20304, opened 2023-01-14
modemmanager: Allow qcom-soc modems

TravMurav:
qcom-soc is a special case modem on devices where it is part of the SoC, which is the case on some Qualcomm chipsets.
Don't fail if the device field is set to this value.

Signed-off-by: Nikita Travkin <nikita@trvn.ru>

Maintainer: @mips171 [[nicholas@nbembedded.com](mailto:nicholas@nbembedded.com)], @aleksander0m

Relevant PR: [openwrt/WIP: introduce Snapdragon 410 based devices](https://github.com/openwrt/openwrt/pull/11729)


TravMurav:
Seems fixed by dropping the checks in #21378 instead.

PR #20308, opened 2023-01-14
netperf: use git tarball

neheb:
netperf.org is gone now. Only github is left.

Maintainer: @tohojo


tohojo:
Guess we should update the package URL as well?

neheb:
Done.

PR #20322, opened 2023-01-15
atlas-probe: update to 2.6.3

neheb:
Switch to git tarballs.

Remove upstreamed patches and add OpenSSL patch.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @ja-pa 

ping @Ansuel

PR #20329, opened 2023-01-15
atlas-sw-probe: update to version 5080

BKPepe:
Maintainer: @ja-pa 
Run tested: not yet

Description:
Related to https://github.com/openwrt/packages/pull/20322


net/atlas-sw-probe/Makefile
```diff
@@ -7,15 +7,15 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=atlas-sw-probe
-PKG_VERSION:=5040
+PKG_VERSION:=5080
 PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/RIPE-NCC/ripe-atlas-software-probe.git
 PKG_SOURCE:=ripe-atlas-software-probe-$(PKG_VERSION).tar.gz
```
neheb, line 15: Not needed.

BKPepe: Yeah, but I haven't touched it. :-( We should do it in separate PR as your PR was merged, and usually, it is a good idea to have the same version of busybox for atlas-sw-probe to remain the compatibility and as it seems to be tested with that version.

Ref: https://github.com/RIPE-NCC/ripe-atlas-software-probe, and there is used 2.6.2 version, not 2.6.3, so hopefully, it will not cause any issues.

PR #20337, opened 2023-01-17
ddns-scripts: add hosting.de provider

bdrung:
Add hosting.de provider. To use dynamic DNS you have to create a DDNS host with a separate DDNS user.

Note: As of 2023-01-17 hosting.de does not work with wget which will fail with `400: Bad Request` (it will work with `--auth-no-challenge`). You should use curl instead. I have reported that to the provider.

Maintainer: ?


feckert:
thanks

Issue #20343, opened 2023-01-18
uvol: handle integral free sizes in autopart


pprindeville:
Maintainer: @dangowrt 
Environment: x86_64, generic, HEAD (0a35d3f992)

Description:

In the case of `sfdisk` returning an integral number of megabytes, the `case` statement will never match. Example:

```
root@pigpen:~# sfdisk -q -F /dev/vda 
 Start     End Sectors  Size
505856 1009667  503812  246M
root@pigpen:~# 
```

Also, the `end` sector is the *last* sector, so `size` should be computed as `$((end - start + 1))` ... or simply use `$sectors`.

dangowrt:
Should be fixed by https://github.com/openwrt/packages/commit/b1edbe349f9e70c72c4a48565a4fd37c94f20a0a and subsequent https://github.com/openwrt/packages/commit/adce04d840ea0dee718455a51a576122c7a460a7. Please re-open if not.

PR #20344, opened 2023-01-18
uvol: fix regression in handling fractional megabytes free

pprindeville:
Maintainer: @dangowrt
Compile tested: x86_64, generic, HEAD (0a35d3f992)

Description:

Handle case of integral free size in megabytes; also simplify free size computation.

Fixes issue #20343

### DIFF
--- a/utils/uvol/Makefile
+++ b/utils/uvol/Makefile
@@ -1,8 +1,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=uvol
-PKG_VERSION:=0.8
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_VERSION:=0.9
+PKG_RELEASE:=1
 
 PKG_MAINTAINER:=Daniel Golle <daniel@makrotopia.org>
 PKG_LICENSE:=GPL-2.0-or-later
--- a/utils/uvol/files/autopart.defaults
+++ b/utils/uvol/files/autopart.defaults
@@ -56,23 +56,14 @@ part_fixup() {
 
 get_free_area() {
 	local found=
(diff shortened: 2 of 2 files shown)


dangowrt:
Why not use sectors also to check the size, that would make it even easier:
```diff
diff --git a/utils/uvol/files/autopart.defaults b/utils/uvol/files/autopart.defaults
index 99ea09406..20e2640fb 100644
--- a/utils/uvol/files/autopart.defaults
+++ b/utils/uvol/files/autopart.defaults
@@ -62,17 +62,8 @@ get_free_area() {
                        continue
                        ;;
                [0-9]*)
-                       case "$size" in
-                               *"k" | *"b")
-                                       continue
-                                       ;;
-                               *"M")
-                                       [ "${size%%.*M}" -lt 100 ] && continue
-                                       ;;
-                               *"G" | *"T")
-                                       ;;
-                       esac
-                       [ "$found" ] || echo "start=$start, size=$((end - start))"
+                       [ "$sectors" -lt 204800 ] && continue
+                       [ "$found" ] || echo "start=$start, size=$sectors"
                        found=1
                        ;;
                esac
```

pprindeville:
> Why not use sectors also to check the size, that would make it even easier:

Using `$size` is more intuitive/easier to read.  Also doesn't depend on the sector size being 512 bytes.

dangowrt:
Thank you for taking care of this!

dangowrt:
Apparently this actually broke some other case:
https://forum.banana-pi.org/t/banana-pi-bpi-r3-openwrt-image/13236/111

dangowrt:
Fixed https://github.com/openwrt/packages/commit/adce04d840ea0dee718455a51a576122c7a460a7

PR #20350, opened 2023-01-20
bash: update to version 5.2.15

BKPepe:
Maintainer: @Naoir
Compile tested: Turris 1.1, powerpc8540/p2020, OpenWrt 21.02.05
Run tested: Turris 1.1, powerpc8540/p2020, OpenWrt 21.02.05


oskarirauta (changes requested):
Set PKG_RELEASE to 1.

BKPepe:
I will not do that until there is a different solution. Although it is deprecated, it works, and many packages, even in the base repository, are using it. In my opinion, it won't be removed so soon. 

This should be done by treewide commit. If there is no autorelease, you simply forgot to bump the version, which recently happened with ucode.

There was a reason why autorelease was implemented, and from the maintenance stuff, it is a good thing.

diizzyy:
LGTM (I have no opinion about PKG_RELEASE)

PR #20351, opened 2023-01-20
goeap_proxy: New package for EAP proxy written in GO

thagabe:
Maintainer: me
Compile tested: mvebu/wrt3200acm&wrt1900acv1/v22.03.3
Run tested: mvebu/wrt3200acm&wrt1900acv1/v22.03.3

Description:
New Package for goeap_proxy written by Matthew Gyurgyik (pyther). Allows router to bypass AT&T gateway and connect directly to ONT you still need to have a gateway.

Signed-off-by: Gabriel Rodriguez <<lifehacksback@gmail.com>>

### DIFF
--- a/net/goeap_proxy/Makefile
+++ b/net/goeap_proxy/Makefile
@@ -1,0 +1,52 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=goeap_proxy
+PKG_VERSION:=0.3.0
+PKG_RELEASE:=1
+PKG_MAINTAINER:= Gabriel Rodriguez <lifehacksback@gmail.com>
+PKG_LICENSE:=none
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/pyther/goeap_proxy.git
+PKG_MIRROR_HASH:=e4014a8dc317d3325d02c1ae81678d6552628bd7a3e253c8ad01a6912ec49dbf
+PKG_SOURCE_VERSION:=3abd66cf35c70b86fe75989ca64c7a72cc8fa797
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/pyther/goeap_proxy
+GO_PKG_LDFLAGS_X:=main.Version=$(PKG_VERSION) main.BuildStamp=$(shell date +%FT%T%z)
+
+include $(INCLUDE_DIR)/package.mk
+include $(TOPDIR)/feeds/packages/lang/golang/golang-package.mk
+
+define Package/goeap_proxy
+	SECTION=:net
+ 	CATEGORY:=Network
+	TITLE:=Proxy EAP packets between interfaces
+  	URL:=https://github.com/pyther/goeap_proxy
+	DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
+define Package/goeap_proxy/description
+Daemon written in GO to proxy EAP Authentication packets
+
+endef
+
+define Package/goeap_proxy/conffiles
+/etc/config/goeap_proxy
+endef
+
+define Package/goeap_proxy/install
+	$(call GoPackage/Package/Install/Bin,$(1))
+
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_BIN) ./files/etc/init.d/goeap_proxy $(1)/etc/init.d/goeap_proxy
+
+	$(INSTALL_DIR) $(1)/etc/config
+	$(INSTALL_CONF) ./files/etc/config/goeap_proxy $(1)/etc/config/goeap_proxy
+endef
+
+$(eval $(call GoBinPackage,goeap_proxy))
+$(eval $(call BuildPackage,goeap_proxy))
--- a/net/goeap_proxy/files/etc/init.d/goeap_proxy
+++ b/net/goeap_proxy/files/etc/init.d/goeap_proxy
@@ -1,0 +1,45 @@
+#!/bin/sh /etc/rc.common
+#
+
+START=25
+USE_PROCD=1
+
+PROG="/usr/bin/goeap_proxy"
+
+boot()
+{
+    config_load goeap_proxy # load config files
+    config_get wan devices wan # create a variable var; letting it know it's a device and assigning the 'wan'
+    config_get lan devices lan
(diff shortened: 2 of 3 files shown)


net/goeap_proxy/files/etc/init.d/goeap_proxy
```diff
@@ -0,0 +1,45 @@
...
+    config_get lan devices lan
+    ubus -t 30 wait_for "network.${wan}.device" "network.${lan}.device" # updated for DSA (v22.02+)
+    rc_procd start_service
+}
+
+start_service()
+{
+    config_load goeap_proxy # load config files
+    config_get wan devices wan
+    config_get lan devices lan4 # change 'lan4' to your switch's port
```
champtar, line 22: Would be cleaner to fail if config is missing, lan4 is not a viable default

net/goeap_proxy/files/etc/init.d/goeap_proxy
```diff
@@ -0,0 +1,45 @@
...
+    config_get_bool ignore_logoff interface 'ignore_logoff' '0'
+
+    local if_wan=$(uci get "network.${wan}.device")
+    local if_router=$(uci get "network.@device[4].name") # my device has 4 ports and I use the 4th port
```
champtar, line 26: you should use the config here

net/goeap_proxy/Makefile
```diff
@@ -0,0 +1,52 @@
...
+PKG_SOURCE_URL:=https://github.com/pyther/goeap_proxy.git
+PKG_MIRROR_HASH:=e4014a8dc317d3325d02c1ae81678d6552628bd7a3e253c8ad01a6912ec49dbf
+PKG_SOURCE_VERSION:=3abd66cf35c70b86fe75989ca64c7a72cc8fa797
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/pyther/goeap_proxy
+GO_PKG_LDFLAGS_X:=main.Version=$(PKG_VERSION) main.BuildStamp=$(shell date +%FT%T%z)
```
champtar, line 19: This is not reproducible

champtar (changes requested):
This is definitely doable in a few lines with nftables using 'fwd to'
https://www.netfilter.org/projects/nftables/manpage.html

net/goeap_proxy/files/etc/init.d/goeap_proxy
thagabe: Let me read up on nftables!

net/goeap_proxy/Makefile
thagabe: What do you mean?

thagabe:
> This is definitely doable in a few lines with nftables using 'fwd to' https://www.netfilter.org/projects/nftables/manpage.html

Wait, you mean this whole project can be done in nftables? That would be facinating! Let me read into it.

champtar:
Here a small example I had to write recently to convert unicast to 'multicast' (for some reasons I had to use the broadcast mac)
```
nft -f- <<EOF
table netdev u2m
flush table netdev u2m
table netdev u2m {
  chain eno5 {
    type filter hook ingress device eno5 priority 0; policy accept;
    ether type ip udp dport 5070 ether daddr set ff:ff:ff:ff:ff:ff ip daddr set 239.1.2.3 fwd to br-internal;
  }
}
EOF
```
I'll let you read up, but basically you need 2 chains (one for each directions), don't know if you need to change the source MAC but it's just 'ether daddr set', match either on the dst mac or packet type or both, and forward to the right interface

net/goeap_proxy/Makefile
```diff
@@ -0,0 +1,52 @@
...
+PKG_NAME:=goeap_proxy
+PKG_VERSION:=0.3.0
+PKG_RELEASE:=1
+PKG_MAINTAINER:= Gabriel Rodriguez <lifehacksback@gmail.com>
+PKG_LICENSE:=none
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/pyther/goeap_proxy.git
+PKG_MIRROR_HASH:=e4014a8dc317d3325d02c1ae81678d6552628bd7a3e253c8ad01a6912ec49dbf
+PKG_SOURCE_VERSION:=3abd66cf35c70b86fe75989ca64c7a72cc8fa797
```
jefferyto, line 12: Suggest to use codeload, e.g. [rclone](https://github.com/openwrt/packages/blob/3570a61cd330e95d98700c4ae5630406226a8026/net/rclone/Makefile#L12-L14), or check out the tag for `PKG_VERSION`, e.g. [irtt](https://github.com/openwrt/packages/blob/3570a61cd330e95d98700c4ae5630406226a8026/net/irtt/Makefile#L11-L14).

net/goeap_proxy/Makefile
jefferyto: Because the value for `main.BuildStamp` depends on the build date/time, if you compile this at two different times with the exact same source code you will get two slightly different binaries, hence not reproducible.

jefferyto, line 19:
```suggestion
GO_PKG_LDFLAGS_X:=main.Version=$(PKG_VERSION) $(if $(SOURCE_DATE_EPOCH),main.BuildStamp=$(SOURCE_DATE_EPOCH))
```

net/goeap_proxy/Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_USE_MIPS16:=0
+
+GO_PKG:=github.com/pyther/goeap_proxy
+GO_PKG_LDFLAGS_X:=main.Version=$(PKG_VERSION) main.BuildStamp=$(shell date +%FT%T%z)
+
+include $(INCLUDE_DIR)/package.mk
+include $(TOPDIR)/feeds/packages/lang/golang/golang-package.mk
```
jefferyto, line 22:
```suggestion
include ../../lang/golang/golang-package.mk
```

net/goeap_proxy/Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+define Package/goeap_proxy
+	SECTION=:net
```
jefferyto, line 25:
```suggestion
	SECTION:=net
```

BKPepe:
PR is stalled, closing.

PR #20353, opened 2023-01-20
treewide: quote CC and CXX

paulfertser:
Maintainer: various
Compile tested: ath79 09f313bfd753543b9030fe0bc25983224f21952d
Run tested: none

Description:
Discovered some packages are not buildable when ccache is used without wrappers, so fixed packaging errors in them.


paulfertser:
Dear maintainers, I'll need some guidance on making this mergeable. Do you I really need to split it into individual commits one per affected package and manually increment PKG_RELEASE in each?

paulfertser:
Cross-link to the change that motivated this work: https://lists.openwrt.org/pipermail/openwrt-devel/2023-January/040296.html

paulfertser:
I'd rather do some scripting and remove all AUTORELEASE instances in one go. But not today.

Issue #19500, opened 2022-10-01
python pyproject.toml


lucize:
what is needed so we can support pyproject.toml packages ?
I would like to update homeassistant but it needs orjson that needs rust and it seems that using a generated setup.py is not working.
I know rust is not in yet but I have it built and the hostfiles are there

karlp:
I'm running into this as well, for a new package: https://pypi.org/project/asyncio-mqtt
@commodo  and @BKPepe  You're the python maintainers?

karlp:
I've worked around it by using the "recommended stub" setup.py, but is that something that we might like to try and create internally if it's not available in the project itself?    @lucize here's an example of patching a package directly, that might be sufficient? https://github.com/etactica/owrt_pub_feeds/blob/wip/egm400-202211/lang/python/python-asyncio-mqtt/patches/0001-support-legacy-setup.py-installation.patch

lucize:
I know that workaround, but I guess a modern approach would be desired, but seems that rust is still kept in PR status

commodo:
> I'm running into this as well, for a new package: https://pypi.org/project/asyncio-mqtt @commodo and @BKPepe You're the python maintainers?

so, i "think" proposed a "workaround" (a while back) with the build infra-structure creating that `setup.py` stub;
now, there can be a debate about whether that's a good idea (and i think we've had it; but i am not so sure about it);

(personally, i think the `setup.py` stub-thingi is not a bad idea for a while, given that the inner-workings of `pyproject.toml` still elude me, and i didn't/don't have time to really dig into how that works)

officially, @jefferyto is the current maintainer of Python;
i've pinged him via email (a few days ago), but no reply yet;
he's probably busy with other stuff (which is totally fine)
i'm running into the pyproject.toml with updating `chardet`

i resigned from being an official maintainer of Python (due to some disagreement with Jeffery; which is not nice to say in public, but ¯\\\_(ツ)_/¯);

now, in the past, Jeffery has had some absences with some package stuff (and that's totally fine);
but usually, he would show up after a while (or me doing some things), and then disagree with me on some stuff;
(now that may be really overly simplified and me being an asshole)

but the point of the above is that i was thinking of maintaining Python again, seeing that it's still at 3.10.7 in the feeds, and 3.11.x is out already, and this `pyproject.toml ` thingi
and it may also cause Jeffery to re-appear :P   (and i don't know of other ways to do that)

commodo:
Oh.
[...]

PR #20366, opened 2023-01-22
python-build: add support for pyproject.toml files (reboot of #17130)

commodo:
Maintainer: @jefferyto
Compile tested: x86 https://github.com/openwrt/commit/6e3acebe35cfc584e145e495b175ae5308d2f2bc
Run tested: x86 https://github.com/openwrt/commit/6e3acebe35cfc584e145e495b175ae5308d2f2bc

------------------------------------------------------------------------

A new PEP 517 (https://www.python.org/dev/peps/pep-0517/) has defined that Python packages can be shipped without any `setup.py` file, and that a `pyproject.toml` file is sufficient.

For these cases, we will add a support in the OpenWrt build-system to provide the default `setup.py` shim layer in case this file does not exist, but there is a `pyproject.toml` file.

We also seem to need to tweak the shim layer with the PKG_VERSION, otherwise the detected version is 0.0.0.

### DIFF
--- a/lang/python/python3-host.mk
+++ b/lang/python/python3-host.mk
@@ -51,6 +51,12 @@ HOST_PYTHON3_PIP:=$(STAGING_DIR_HOSTPKG)/bin/pip$(PYTHON3_VERSION)
 
 HOST_PYTHON3_PIP_CACHE_DIR:=$(DL_DIR)/pip-cache
 
+define SetupPyShim
+	if [ -f $(1)/pyproject.toml ] && [ ! -f $(1)/setup.py ] ; then \
+		$(CP) $(python3_mk_path)setup.py.shim $(1)setup.py ; \
+	fi
+endef
+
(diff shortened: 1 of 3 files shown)


commodo:
Fixes https://github.com/openwrt/packages/issues/19500
Reboot of https://github.com/openwrt/packages/pull/17130

So, while I'm not a big fan of half-assing stuff, we do ocasionally get to the point, where something half-ass-ed will be better than nothing. Like for example this interim mechanism for providing a `setup.py` for all packages that have `pyproject.toml` but not `setup.py`.
The original idea proposed in PR #17130 did not get a better version, so for now that will have to work.
Until we have energy/time (or someone else wants) to rework this.

commodo:
Btw: tested with `chardet` 5.1.0

commodo:
let's merge this now

PR #20383, opened 2023-01-24
bccmd: new package

vtremblay:
Maintainer: me / @vtremblat
Compile tested: ip40xx - Linksys WHW03v2 - master
Run tested:  ip40xx - Linksys WHW03v2 - master

Description:
bccmd was a tool provided in BlueZ 5.55 and less. They removed it as there was not a lot of support.
It is still needed to configure bluetooth CSR chips with the BlueCore command interface. No other solution is currently available and we need to go fetch the source and compile it each time it is needed.
I extracted the code, created a new repo and based the tool on the latest version of the BlueZ lib.

Signed-off-by: Vincent Tremblay <vincent@vtremblay.dev>

### DIFF
--- a/utils/bccmd/Makefile
+++ b/utils/bccmd/Makefile
@@ -1,0 +1,45 @@
+# SPDX-Identifier-License: GPL-2.0-or-later
+#
+# Copyright (C) 2023 Vincent Tremblay <vincent@vtremblay.dev>
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=bccmd
+PKG_VERSION:=6.0
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/bccmd/bccmd.git
+PKG_SOURCE_VERSION:=v$(PKG_VERSION)
+PKG_MIRROR_HASH:=2cd7e6eb755785101a60e1da70de2d7788ba7bd2459b576e82b5851ed422588e
+
+PKG_MAINTAINER=Vincent Tremblay <vincent@vtremblay.dev>
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=COPYING
+
+PKG_BUILD_PARALLEL:=1
+
+PKG_FIXUP:=autoreconf
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/bccmd
+  TITLE:=CSR BlueCore command interface
+  SECTION:=utils
+  CATEGORY:=Utilities
+  URL:=https://github.com/bccmd/bccmd.git
+  DEPENDS:=+bluez-libs
+endef
+
+define Package/bccmd/description
+  BBCMD is used to send BlueCore commands to Cambridge Silicon Radio devices.
+endef
+
+CONFIGURE_ARGS += --disable-manpages
+
+define Package/bccmd/install
+	$(INSTALL_DIR) $(1)/usr/sbin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/bccmd $(1)/usr/sbin/
+endef
+
+$(eval $(call BuildPackage,bccmd))


vtremblay:
> 
> * utils/bccmd/Makefile: PKG_RELEASE:=3

fixed

utils/bccmd/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+PKG_MAINTAINER=Vincent Tremblay <vincent@vtremblay.dev>
+
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=COPYING
+
+PKG_FIXUP:=autoreconf
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/bccmd/${PKG_NAME}/tar.gz/v$(PKG_VERSION)?
```
neheb, line 19: Please use regular git, look at PKG_SOURCE_PROTO for examples,

vtremblay: Thanks for the hint. I have updated the file and re-tested it.

utils/bccmd/Makefile
```diff
@@ -0,0 +1,44 @@
...
+
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=COPYING
+
+PKG_FIXUP:=autoreconf
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/bccmd/bccmd.git
+PKG_SOURCE_VERSION:=v$(PKG_VERSION)
+PKG_MIRROR_HASH:=2cd7e6eb755785101a60e1da70de2d7788ba7bd2459b576e82b5851ed422588e
```
1715173329, line 21:
```suggestion
PKG_SOURCE_PROTO:=git
PKG_SOURCE_URL:=https://github.com/bccmd/bccmd.git
PKG_SOURCE_VERSION:=v$(PKG_VERSION)
PKG_MIRROR_HASH:=2cd7e6eb755785101a60e1da70de2d7788ba7bd2459b576e82b5851ed422588e

PKG_MAINTAINER=Vincent Tremblay <vincent@vtremblay.dev>
PKG_LICENSE:=GPL-2.0-or-later
PKG_LICENSE_FILES:=COPYING

PKG_FIXUP:=autoreconf
```

consider adding `PKG_BUILD_PARALLEL:=1` if this program can be compiled multithreaded.

utils/bccmd/Makefile
```diff
@@ -0,0 +1,44 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/bccmd
+	TITLE:=CSR BlueCore command interface
+	SECTION:=utils
+	CATEGORY:=Utilities
+	URL:=https://github.com/bccmd/bccmd.git
+	DEPENDS:=+bluez-libs
```
1715173329, line 30:
```suggestion
  SECTION:=utils
  CATEGORY:=Utilities
  TITLE:=CSR BlueCore command interface
  URL:=https://github.com/bccmd/bccmd.git
  DEPENDS:=+bluez-libs
```

use 2 \<space\> instead of \<tab\>

utils/bccmd/Makefile
```diff
@@ -0,0 +1,44 @@
...
+endef
+
+define Package/bccmd/description
+	BBCMD is used to send BlueCore commands to Cambridge Silicon Radio devices.
```
1715173329, line 34:
```suggestion
  BBCMD is used to send BlueCore commands to Cambridge Silicon Radio devices.
```

same here

1715173329 (changes requested):

utils/bccmd/Makefile
1715173329: > Please use regular git, look at PKG_SOURCE_PROTO for examples,

hmmm, may I ask if there's any problem with codeload?

utils/bccmd/Makefile
vtremblay: I'm not too sure about this one. From the spec, Makefiles requires tabs and not spaces. 
What is the reason to change it with spaces? (I really do want to understand!)

utils/bccmd/Makefile
vtremblay: done, I fixed the order as suggested and added the `PKG_BUILD_PARALLEL` config, thanks!

utils/bccmd/Makefile
1715173329: > https://stackoverflow.com/questions/28712585/when-to-use-space-or-tab-in-makefile
> In order for make to tell the difference between a recipe and things that are not a recipe, it uses TAB characters. So, lines that begin with TAB are assumed to be part of a recipe (so they are shell scripts and passed to the shell for parsing), and lines that do not begin with TAB cannot be part of a recipe (so they cannot be shell scripts: they must be make syntax).

Tabs also works here, but let's keep constant with other makefiles in tree.

vtremblay: thanks for the explanation, I really appreciate. 
It's fixed!

utils/bccmd/Makefile
vtremblay: fixed

vtremblay:
Thanks a lot for the feedback @1715173329 and @neheb. Do you see any other changes that are needed?
Do you know what is the next step to get this approved?
Thanks again!

BKPepe:
I am not sure, how to proceed it further because from your OP, I see:

> bccmd was a tool provided in BlueZ 5.55 and less. They removed it as there was not a lot of support.

I think this says a lot about why it should not be merged it here. It is good that you forked it, but it means that you need to take care of it, it is a new repository, there are not many commits and it is not packaged for any GNU/Linux distributions. 

I am so sorry for this, but this is NACK from me.

BKPepe:
Forked repository seems to be inactive, which I was afraid of. The last commit what I can see is from 24th January, thus closing PR here.

PR #20410, opened 2023-01-28
python3: Do not remove the platform so suffix

lu-zero:
Otherwise pip install would not work for the wheels.

The history of the file does not seem to explain why this patch exists, but it definitely breaks building maturin-based packages.


jefferyto:
It’s been a while since I looked into this, but I am pretty sure removing the platform triplet is necessary to compile target Python packages that use packages compiled by host Python (see #12754).

I suggest trying to compile python-cryptography or other more complex Python packages with this change to see if it succeeds.

I also suggest adding OpenWrt packages for Python packages that require compilation (these are the ones that produce platform-specific wheels) instead of using pip to install the wheel.

I suppose another to fix the issue you are having is to have Python recognize .so files with or without the platform triplet, or for Python to accept either the host’s or target’s platform triplet. I don’t know how difficult this will be and I’m also not sure how pressing of an issue this is.

lu-zero:
I built most of home-assistant with this and it worked well. I can prepare the set with more of it.

lu-zero:
My repos have also an up to date python-cryptography that uses rust.

jefferyto:
I don’t mean cryptography using rust, I mean the in-repo version, or another package that uses cffi or another complex package like numpy. Sorry but getting home assistant compiled doesn’t mean much.

lu-zero:
ha relies on cffi-using packages like lru-dict and also numpy so... If you have in mind something that you knew did not work I can try it. Feel free to check the contents of https://github.com/domo-iot/packages/commits/ha-cleanup it is yet to be cleaned up but did work.

lu-zero:
To support every platform `EXT_SUFFIX` needs to be extracted since upstream python is sort of creative regarding the naming and its logic is present only in a conftest, but on aarch64 locally everything works.

the cryptography package apparently got more patches between I'd have to undo.

jefferyto:
This was done in #21552 so I'm going to close this PR - thanks for opening this though :pray:

PR #20415, opened 2023-01-28
mv88e6xxx_dump: add new packages to debug switch issues

champtar:
Maintainer: me
Compile tested: mvebu turris omnia / master https://github.com/openwrt/openwrt/commit/2a20dc717a897efe54f241636ee1566b695dbc7d
Run tested: no (I don't have access to my router for now)
Description:
This might help debug issues like https://github.com/openwrt/openwrt/issues/11877

### DIFF
--- a/utils/mv88e6xxx_dump/Makefile
+++ b/utils/mv88e6xxx_dump/Makefile
@@ -1,0 +1,37 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=mv88e6xxx_dump
+PKG_RELEASE:=1
+
+PKG_FIXUP:=autoreconf
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/lunn/mv88e6xxx_dump
+PKG_MIRROR_HASH:=2f1bce89212a10be89e106398ccc2f7eedcc1a7299bf8a1a05b3baf061f779f5
+PKG_SOURCE_DATE:=2022.03.29
+PKG_SOURCE_VERSION:=a3f713f99e0004d17d9282e0cb253505b9ca35b6
+
+PKG_MAINTAINER:=Etienne Champetier <champetier.etienne@gmail.com>
+PKG_LICENSE:=GPL-2.0-only
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/mv88e6xxx_dump
+  SECTION:=utils
+  CATEGORY:=Utilities
+  URL:=https://github.com/lunn/mv88e6xxx_dump
+  TITLE:=mv88e6xxx_dump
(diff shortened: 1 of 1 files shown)


champtar:
@chealy thanks for pointing out the existence of this tool (https://github.com/openwrt/openwrt/issues/11077#issuecomment-1399146533), should it work out of the box / anything special to enable ? I just added libmnl as dependency, and I can't test as I don't have access to my router for ~3 weeks

robimarko:
Can't you just use mdio-tools which provide `mvls` tool for the Marvell Linkstreet switches?

chealy:
> Can't you just use `mdio-tools` which provide `mvls` tool for the Marvell Linkstreet switches?

mdio-tools will only provide register values and some of the items being accessed via devlink regions requires querying the contents of MAC database by banging on multiple indirect registers.  This would proof challenging for most people debugging on the platform.

Additionally, the tool I'm pointing to above decodes the register data to human readable data such as the MAC and VLAN databases and port configurations.

chealy:
> @chealy thanks for pointing out the existence of this tool ([openwrt/openwrt#11077 (comment)](https://github.com/openwrt/openwrt/issues/11077#issuecomment-1399146533)), should it work out of the box / anything special to enable ? I just added libmnl as dependency, and I can't test as I don't have access to my router for ~3 weeks

I'm not aware of any special kernel configuration necessary to enable said feature (beyond the Marvell DSA driver in the first place.)  On the userspace side, I'm not aware of the full scope of dependencies.  I wouldn't expect much though.

chealy (approved):
Reviewed-by: Chris Healy <cphealy@gmail.com>

robimarko:
LGTM, so:
`Reviewed-by: Robert Marko <robimarko@gmail.com>`

dhewg:
A DEPENDS on the kernel CONFIG_NET_DSA_MV88E6XXX would be nice, but I guess that doesn't work as it's compiled in and not a kmod package?
So maybe a DEPENDS on imx||kirkwood||mvebu||qoriq or something like that?

champtar:
@dhewg will add the depends later today / tonight
Still looking for someone to run test this before merging

ynezz:
@champtar package seems to work on Omnia and wrt3200acm.

chealy:
@ynezz  any chance you can share some output from the tool in pastebin or via some other method?

ynezz:
> @ynezz any chance you can share some output from the tool in pastebin or via some other method?

http://sprunge.us/2wsMKk from wrt3200acm

chealy:
@ynezz  Excellent, thank you!
That the VTU is missing may be because the kernel you are using is too old.  Was this tested with 5.10 or 5.15 kernel?

ynezz:
>  That the VTU is missing may be because the kernel you are using is too old. Was this tested with 5.10 or 5.15 kernel?

Indeed, it's running 5.10 kernel, here is output from Omnia running latest snapshot http://sprunge.us/9h6ZAu

chealy:
@ynezz  All the output looks correct.  It could be enhanced to provide more human readable output, but that's a problem for later.  This tool seems good to integrate as it is.

champtar:
@dhewg DEPENDS added
@ynezz thanks for testing
All, thanks for the reviews

PR #20426, opened 2023-02-01
Fixe uuid randomly generated after each startup

luochongjun:
Since /lib/functions.sh is not referenced in startup scripts, config_get will always fail and a new uuid will be set.

Signed-off-by: luochongjun <luochongjun@gl-inet.com>

Description:


stangri:
Could you please prepend the commit and PR titles with the package name as per contribution guidelines?

Also, please add maintainer and compile/run-test information as per template.

BKPepe:
Closing due to OP inactivity here.

PR #20427, opened 2023-02-01
tailscale: update to 1.36.0

oskarirauta:
Maintainer: Jan Pavlinec / @ja-pa

 - Update tailscale to version 1.36.0
 - Patch iptables support

Tailscale does not (yet) support nftables.
Tailscale allows running with --netfilter=off allowing end-user to create his own firewall rules, but this affects only tailscale cli, not tailscaled daemon, so connection cannot be made without error telling that tailscaled was unable to determine execute iptables for determining it's version.

There is a work-around for those who do not want nft-iptables compatibility package; they can create a script to /usr/bin/iptables which responds to --version argument and echos fake version string
and on any other arguments or no arguments, just exits.

After this procedure and starting tailscale cli with netfilter off- it works. Openwrt has moved on to
nftables, so iptables manipulation seems unnecessary. Especially for other reasons, on Openwrt, firewall should be configured on it's own, because firewall rules made by other software, such as tailscale,
loose their firewalling rules when firewall restarts.

So I patched it to allow "fake" iptables pointing to executable /bin/false and ignoring version
request. And I also set cli to default to netfilter off setting.

If still end-user wants to use iptables, this patch does not make it impossible; just install
iptables, or nft-iptables, and run tailscale with argument --netfilter=on and it works out
as it did before, tailscaled daemon still matches with iptables if it is found in $PATH.

Release notes for version 1.36.0:
 - tailscale debug daemon-logs
 - tailscale version now advertises when you're on the unstable (dev) track
 - tailscale version --json for JSON output mode
 - --json output for tailscale lock status and tailscale lock log commands.
 - Handle cases where a node expires and we don't receive an update about it from the control server (https://github.com/tailscale/tailscale/pull/6929 and https://github.com/tailscale/tailscale/pull/6937).
 - tailscale status --json now includes KeyExpiry time and Expired boolean.
 - Support UPnP port mapping of gateway devices where they are deployed as a highly available pair (https://github.com/tailscale/tailscale/pull/6946).
 - Support arbitrary IP protocols like EOIP and GRE (https://github.com/tailscale/tailscale/pull/6423).
 - Fix exit node handling of a large number of SplitDNS domains (https://github.com/tailscale/tailscale/pull/6875).
 - Accept DNS-over-TCP responses up to 4KBytes.
 - Unix platforms: when /etc/resolv.conf needs to be overwritten for lack of options, a comment in the file now links to https://tailscale.com/s/resolvconf-overwrite
 - Tailscale SSH: ssh to tailscaled as a non-root user now works again, as long as you only ssh to the same user that tailscaled is running as

### DIFF
--- a/net/tailscale/Makefile
+++ b/net/tailscale/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tailscale
-PKG_VERSION:=1.32.3
+PKG_VERSION:=1.36.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=tailscale-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/tailscale/tailscale/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=4cf88a1d754240ce71b29d3a65ca480091ad9c614ac99c541cef6fdaf0585dd4
+PKG_HASH:=25b293a7e65d7b962f0c56454d66fa56c89c3aa995467218f24efa335b924c76
 
 PKG_MAINTAINER:=Jan Pavlinec <jan.pavlinec1@gmail.com>
 PKG_LICENSE:=BSD-3-Clause
@@ -61,23 +61,43 @@ endef
 
 Package/tailscaled/description:=$(Package/tailscale/description)
 
+define Package/tailscaled/conffiles
+/etc/config/tailscale
+/etc/tailscale/tailscaled.state
+endef
+
+GO_IPTABLES_VERSION:=0.6.0
+GO_IPTABLES_FILE:=$(PKG_NAME)-go-iptables-$(GO_IPTABLES_VERSION).tar.gz
+
+define Download/go-iptables
+  URL:=https://codeload.github.com/coreos/go-iptables/tar.gz/v$(GO_IPTABLES_VERSION)?
+  URL_FILE:=$(GO_IPTABLES_FILE)
+  FILE:=$(GO_IPTABLES_FILE)
+  HASH:=a784cc17fcb17879f073eae47bc4c2e899f59f6906dac5a0aa7a9cc9f95ea66d
+endef
+
+define Build/Prepare
+	$(PKG_UNPACK)
+	[ ! -d ./src/ ] || $(CP) ./src/. $(PKG_BUILD_DIR)
+	$(eval $(call Download,go-iptables))
+	( \
+		mkdir -p $(PKG_BUILD_DIR)/patched/ ; \
+		gzip -dc $(DL_DIR)/$(GO_IPTABLES_FILE) | $(HOST_TAR) -C $(PKG_BUILD_DIR)/patched $(TAR_OPTIONS) ; \
+		mv $(PKG_BUILD_DIR)/patched/go-iptables-$(GO_IPTABLES_VERSION) $(PKG_BUILD_DIR)/patched/go-iptables ; \
+	)
+	$(Build/Patch)
+endef
+
 define Package/tailscale/install
 	$(INSTALL_DIR) $(1)/usr/sbin
 	$(INSTALL_BIN) $(GO_PKG_BUILD_BIN_DIR)/tailscale $(1)/usr/sbin
 endef
 
 define Package/tailscaled/install
-	$(INSTALL_DIR) $(1)/usr/sbin
+	$(INSTALL_DIR) $(1)/usr/sbin $(1)/etc/init.d $(1)/etc/config
 	$(INSTALL_BIN) $(GO_PKG_BUILD_BIN_DIR)/tailscaled $(1)/usr/sbin
-	$(INSTALL_DIR) $(1)/etc/init.d/
 	$(INSTALL_BIN) ./files//tailscale.init $(1)/etc/init.d/tailscale
-	$(INSTALL_DIR) $(1)/etc/config/
 	$(INSTALL_DATA) ./files//tailscale.conf $(1)/etc/config/tailscale
-endef
-
-define Package/tailscaled/conffiles
-/etc/config/tailscale
-/etc/tailscale/tailscaled.state
 endef
 
 $(eval $(call BuildPackage,tailscale))
--- a/net/tailscale/README.md
+++ b/net/tailscale/README.md
@@ -25,4 +25,9 @@ Run command and finish device registration with the given URL.
 tailscale up
 ```
 
+If you are running with nftables, it is not supported by tailscale,
+so disable it and configure firewall by yourself and add argument
+--netfilter-mode off
+to tailscale up command to disable iptables use.
+
 After that, you should see your router in tailscale admin page.
--- a/net/tailscale/patches/010-fake_iptables.patch
+++ b/net/tailscale/patches/010-fake_iptables.patch
@@ -1,0 +1,53 @@
+--- a/go.mod
++++ b/go.mod
+@@ -2,6 +2,8 @@ module tailscale.com
+ 
+ go 1.19
+ 
++replace github.com/coreos/go-iptables => ./patched/go-iptables
++
+ require (
+ 	filippo.io/mkcert v1.4.3
+ 	github.com/Microsoft/go-winio v0.6.0
+--- a/patched/go-iptables/iptables/iptables.go
++++ b/patched/go-iptables/iptables/iptables.go
+@@ -149,12 +149,39 @@ func New(opts ...option) (*IPTables, err
+ 	return ipt, nil
+ }
+ 
++func NewFake(opts ...option) (*IPTables, error) {
++
++	ipt := &IPTables{
++		path:		   "/bin/false",
++		proto:		   ProtocolIPv4,
++		hasCheck:	   false,
++		hasWait:	   false,
++		waitSupportSecond: false,
++		hasRandomFully:	   false,
++		v1:		   0,
++		v2:		   0,
++		v3:		   0,
++		mode:		   "legacy",
++		timeout:	   0,
++	}
++
++	for _, opt := range opts {
++		opt(ipt)
++	}
++
++	return ipt, nil
++}
++
+ // New creates a new IPTables for the given proto.
+ // The proto will determine which command is used, either "iptables" or "ip6tables".
+ func NewWithProtocol(proto Protocol) (*IPTables, error) {
+ 	return New(IPFamily(proto), Timeout(0))
+ }
+ 
++func NewFakeWithProtocol(proto Protocol) (*IPTables, error) {
++	return NewFake(IPFamily(proto), Timeout(0))
++}
++
+ // Proto returns the protocol used by this IPTables.
+ func (ipt *IPTables) Proto() Protocol {
+ 	return ipt.proto
--- a/net/tailscale/patches/020-tailscaled_fake_iptables.patch
+++ b/net/tailscale/patches/020-tailscaled_fake_iptables.patch
@@ -1,0 +1,32 @@
+--- a/wgengine/router/router_linux.go
++++ b/wgengine/router/router_linux.go
+@@ -129,7 +129,7 @@ func newUserspaceRouter(logf logger.Logf
+ 
+ 	ipt4, err := iptables.NewWithProtocol(iptables.ProtocolIPv4)
+ 	if err != nil {
+-		return nil, err
++		ipt4, err = iptables.NewFakeWithProtocol(iptables.ProtocolIPv4)
+ 	}
+ 
+ 	v6err := checkIPv6(logf)
+@@ -148,7 +148,7 @@ func newUserspaceRouter(logf logger.Logf
+ 		// if unavailable. We want that to be a non-fatal error.
+ 		ipt6, err = iptables.NewWithProtocol(iptables.ProtocolIPv6)
+ 		if err != nil {
+-			return nil, err
++			ipt6, err = iptables.NewFakeWithProtocol(iptables.ProtocolIPv6)
+ 		}
+ 	}
+ 
+@@ -1635,11 +1635,6 @@ func checkIPv6(logf logger.Logf) error {
+ 		return fmt.Errorf("kernel doesn't support IPv6 policy routing: %w", err)
+ 	}
+ 
+-	// Some distros ship ip6tables separately from iptables.
+-	if _, err := exec.LookPath("ip6tables"); err != nil {
+-		return err
+-	}
+-
+ 	return nil
+ }
+ 
--- a/net/tailscale/patches/030-default_to_netfilter_off.patch
+++ b/net/tailscale/patches/030-default_to_netfilter_off.patch
@@ -1,0 +1,11 @@
+--- a/cmd/tailscale/cli/up.go
++++ b/cmd/tailscale/cli/up.go
+@@ -143,7 +143,7 @@ func defaultNetfilterMode() string {
+ 	if distro.Get() == distro.Synology {
+ 		return "off"
+ 	}
+-	return "on"
++	return "off"
+ }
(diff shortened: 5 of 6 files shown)


oskarirauta:
@1715173329 Could you check the issue, reason of failure is little strange...

1715173329:
https://github.com/openwrt/packages/blob/190b436f4ef5e9bf7e91d5160036a1086ccc2107/net/tailscale/test.sh#L7
maybe `tailscaled version`?

oskarirauta:
-version output:
```
root@gateway:~# tailscaled -version
1.36.0
  go version: go1.19.5
root@gateway:~# 
```

What's with that?

1715173329:
> https://github.com/openwrt/packages/actions/runs/4062002705/jobs/6992565993#step:14:57
> Installing tailscaled (1.32.3-1) to root...

CI installed an old tailscaled package, we ever got similar error on another package (library vs binary).

this script should work:
```
#!/bin/sh

case "$1" in
	tailscale)
		tailscale version | grep "$2"
		;;
tailscaled)
		tailscaled -version | grep "$2"
		;;
esac
```

oskarirauta:
@1715173329 

Thanks. You are the best- let's see if it works now :)

1715173329:
sure thing ;)

oskarirauta:
@1715173329 How about a review..? I have new configuration system pending, but I'd rather see thoughts about this first..

1715173329:
> Tailscale allows running with --netfilter=off allowing end-user to create his own firewall rules, but this affects only tailscale cli, not tailscaled daemon, so connection cannot be made without error telling that tailscaled was unable to determine execute iptables for determining it's version.

Can you report this issue to upstream (or it's already there)?

I don't ever learn golang, but IMO we should avoid calling go-iptables when iptables is not available rather than patch it (with a dummy command).

oskarirauta:
@1715173329 
> Can you report this issue to upstream (or it's already there)?

It is already there, and has been for a while already.
On that issue, someone shared a pre-compiled binary (for x86-64) with iptables disabled in one way or other, he did not provide a patch, how he did it - so I did it like this.

I suspect that switching from iptables would require another depency, that does the similar job for nftables instead, but iptables is tangled so much to there, that it most likely is very unwanted task to fulfill as it would require a huge modification; removal isn't any easier, I tried that first and failed so many times.

> I don't ever learn golang, but IMO we should avoid calling go-iptables when iptables is not available rather than patch it (with a dummy command).

I have same problem with golang, but was able to manage the patch finally.
Problem **why** go-iptables had to be patched, was that it was not possible to create "fake" iptables in tailscale itself, as iptables structs members were unexported (private) - so I needed a function **in** go-iptables allowing to create faked iptables. It is that complex that removing all iptables usage from tailscale, would be a huge patch and out of my capabilities, because it's go..

Now with my patch, it is only tested once unless you make your connection up with --netfilter-mode=on and that query is for iptables version, which was that thing in go-iptables struct constructor, so it was not possible to override it without patching go-iptables. Now it just is compiled in, but not used, unless user wants to use it, I made sure it is possible to use it, if someone really wants to.

If you check my patch that adds the function you see how it does (even though go language is foreign, it's pretty understandable) - I had that first patched to tailscale code, but it failed because member's were not accessible from other modules except iptables-go - I searched and found something called monkey patching, but none of those solutions worked out.

Avoiding to call it, would require removing ipt4 and ipt6 members from router struct, but removing them results in a thousand errors as using it, is skipped with that netfilter is off by default, but support is compiled in anyway and called from time to time from here and there, it would be truly a massive patch to project, and very difficult to maintain- this is more like a mini patch with a lot easier way to maintain, which was to approach that I was looking forward to.

@1715173329 penny for your thoughts about this?

1715173329:
okay it  makes sense to me then
Thanks for the clarification!

net/tailscale/Makefile
```diff
@@ -61,6 +61,28 @@ endef
 
 Package/tailscaled/description:=$(Package/tailscale/description)
 
+GO_IPTABLES_VERSION:=0.6.0
+GO_IPTABLES_FILE:=$(PKG_VERSION)-go-iptables-$(GO_IPTABLES_VERSION).tar.gz
```
1715173329, line 65:
```suggestion
GO_IPTABLES_FILE:=go-iptables-$(GO_IPTABLES_VERSION).tar.gz
```

this source file is generic anyway... or use `$(PKG_NAME)-go...` instead?

oskarirauta:
No problem. I am available for questions and reason explanations on all my commits for most time of day :)

1715173329 (approved):

net/tailscale/Makefile
oskarirauta: I'll fix that.. that's a typo, it should had been pkg_name, not version..

oskarirauta:
If this gets merged, I have a follow-up commit that makes configuring this simpler, as it brings support for tailscale protocol. And I also have in that case that it is merged, a protocol package for LuCi as well.

I am currently using this protocol based configuration, so it is configured in /etc/config/network and as simplest looks like this:
```
config interface 'tailscale'
	option proto 'tailscale'
```

but there are options available, such as device option for renaming default tailscale0 ifd to what you want it to be, also netfilter setting is available through there, if you really want to use iptables... static port number customization.. logging options. But if you want to go with defaults, just put that to network config and interface comes available and up on daemon start. But yes, that's for the second commit.

```
root@gateway:/etc/config# ifstatus tailscale
{
	"up": true,
	"pending": false,
	"available": true,
	"autostart": true,
	"dynamic": false,
	"uptime": 84,
	"l3_device": "tailscale0",
	"proto": "tailscale",
	"updated": [
		"addresses"
	],
	"metric": 0,
	"dns_metric": 0,
	"delegation": false,
	"ipv4-address": [
		{
			"address": "100.122.70.4",
			"mask": 32
		}
	],
...
}
```

oskarirauta:
@ja-pa 
@dangowrt 

Possible to check this out?

oskarirauta:
@1715173329 I can't make pass the checks due to kernel mismatch on arm cortex a15.

1715173329:
Just ignore it. Known issue of CI.

oskarirauta:
@1715173329 

Thanks. And thanks for merging as well- I thought your previous words about this, and made a patch that doesn't require patching of go-iptables, and got it also working - though.... I am not sure will it work on every occasion, I think, what just got merged - is stable, and patching tailscale to tolerate ipt4 and ip6 = nil, might have issues, so I would recommend what you just merged - but I am open for conversation on topic if you want me to push another PR with that other method.

Currently somewhat working patch (tested to run, but does it keep running..?) is available at:
https://github.com/oskarirauta/tailscale/commit/0ad3490ff2856f54669419b1802e7f2990f1cb74

Let me know if that's what interests more than current workaround..

1715173329:
I've never used it, so I'd prefer your choice ;)

oskarirauta:
Thanks, I value your opinion :)

oskarirauta:
@1715173329

I have nearly completed that new protocol model for tailscale that I previously mentioned, but there are some details that I'd like to get opinion about what would be best method to do it.

As you never had used it, I'll brief you in..

Tailscale vpn is a mesh type vpn, meaning that there is a central server at tailscale.com where you log in (actually there's a way to host one yourself, but I am not covering that yet on next update, since I haven't yet tried to host one..) and all your connected devices are then in same private network with 100.10.0.0/64 addressing (random, unlike at zerotier's equillavence where you can define ip addresses for clients). Tailscale offers a free account, and with free account, 20 machines can be connected to your account.

Features are offered also: you can also set up a "router", meaning that when you have 1 machine (router) connected to tailscale, you can route your subnet through it; meaning if my lan is for example 192.168.0.0/16, and my tailscale ip is 100.78.38.101 for my router, and I advertise route 192.168.0.0/16 through the host - anyone (unless rules are defined to decline selectively, available at admin panel) connecting to my tailscale vpn, can access 192.168.0.0/16 through it - for example, my phone can have ip 100.75.253.4 from my tailscale network- now if I use ssh or for example just ping, from my phone and both phone and router are online, I can ping and ssh from my phone to both 100.78.38.101 or 192.168.0.1 through this setup. Handy, isn't it? And ofcourse, my phone is pingable also from router. Ssh cannot be done towards my phone since there is no ssh server on it.. But if would, it would work out too.

Devices also get a hostname, normally this is hostname of machine, but can be also defined, either when connecting - or from admin panel, and if allow-dns feature is set, also these map perfectly on system.

```
root@gateway:/lib/netifd/proto# tailscale status
100.78.38.101    gateway                 oskarirauta@ linux        -
100.111.253.14   imac                       oskarirauta@ macOS   offline
100.75.253.4     iphone-12-pro       oskarirauta@ iOS         idle, tx 656 rx 656
```

Hostnames can be seen here, so I can ping using these hostnames, and also with domain tailnet.name which is assigned (and changeable, but not customizable to what ever value..), in my case default value I got is tail3c9d0.ts.net
so pinging iPhone-12-pro.tail3c9d0.ts.net works.

Other features also are available, but I won't cover them in detail here-

### Details on how it works

All configuration is done under /etc/config/network - but initscript must be enabled, for vpn to work - as it consists of 2 parts, server/daemon(tailscaled) and cli commander(tailscale).
My protocol checks that server's status is running and is initially unavailable. When init script is started, and tailscale0 tunnel (name customizable) comes up, protocol comes available. If auto option is not set to 0, init script brings it up by executing ifup tailscale (or whatever you named it..).

Protocol depends on wan connection and then commands tailscale up with arguments that come from /etc/config/network options - and all works nicely.

except... tailscale stores a "tailscaled.state" file, containing your previous options AND login information in a unknown format. We cannot keep this in /tmp or erase it every now and then, because if we do - we need to re-login and we get different ip addresses, and basicly... It's the same that would start it all-over.
State also contains information about up/down state, if network was not put down properly, it will automaticly be brought up on server startup.

### The problem

Problem is that tailscale compares settings with previous settings, and refuses to start if options have changed, asking user to either start tailscale up with identical arguments as before, or reset options (not login stuff..) and then user can set options again to whatever they want....

This is the procedure to reset (one of the ways)
 - ensure initscript has started server
 - check that tailscale0 is down, if not - ifdown tailscale (or whatever you named it)
 - restart server (but not with initscript, unless you set auto to 0 in config, otherwise it starts interface)
 - execute tailscale up --reset
 - execute tailscale down
 - stop server
 - clean up, execute tailscaled --cleanup
 - start server
 - new settings are applied when interface comes up

A bit tricky- especially when there are difficulties to know that settings changed and tailscale refused. Also server goes to state where it no longer answers to cli calls until restarted (failsafe mecanism for unauthorized changes?)

### Some ideas

When I'm finished and this gets merged (if approved) - I have plans to write a wiki entry containing all the necessary things on usage - but yet, I think - I could make things easier. And I should. Just how to do it? Not yet sure about that, and I'd like to get someone elses opinion on this-

Few things came to my mind:
 - patch tailscale to auto reset - this would be best, but might be too complicated, especially when tailscale gets update, so this is more or less, last resort solution even though it would be most handy...
 - add new method/argument to initscript: reset

reset argument would more or less do all steps I described- there is unfortunately no way to completely automate this process, atleast method that I know about - It could be triggered to happen everytime /etc/config/network changes, but it seems a bit overkill.... because it would happen also everytime you change some other network interface as well, so I am thinking about manual way. After all, it is very rare that you want to change your setup when you've got it the way you like it.

# The question
Would you recommend this method - or do you have something else in mind? I really hope you won't say that it must be patched, that sounds, a lot of work with a high failure rate, but again, I have a idea how to do this, but wouldn't want to be the only one deciding about it. I also have some plans to detect failed tailscale up due to changed options- and log this to syslog some how- but I haven't implemented this yet, as I am still doing some further testing and clean up of my work so far..

So, what do you think of this?

### Available options

 - device (custom device name, default: tailscale0)
 - port (custom port for udp, default 41641)
 - log_stdout & log_stderr (tailscaled server log output to syslog, ugly...)
 - disable_support (don't send statistics to tailscale)
 - config_path (/etc/config/tailscale)
 - hostname (default: none, tailscale uses system's own hostname)
 - disable_dns (default: false, allows dns resolve feature)
 - disable_routes (default: false, allows routing through other nodes to their advertised routes)
 - exitnode (default: false, allows routing other connected devices all internet traffic through this node)
 - routes, array (default: none, list of subnets to route to through this server)
 - iptables (default: false, use legacy iptables, my patch to go-iptables made it possible to ignore iptables usage)
 - disable_incoming (default: false, allow only outgoing connections to vpn, block all incoming)

I also have a luci protocol support made, but reset is required after applying settings there- maybe I could integrate it also there, hmm..

1715173329:
> @1715173329
> 
> I have nearly completed that new protocol model for tailscale that I previously mentioned, but there are some details that I'd like to get opinion about what would be best method to do it.
> 
> As you never had used it, I'll brief you in..
> 
> < snip >

Thank you!
I primarily use ZeroTier to reach my simple requirement :P

> < snip >
> 
> Few things came to my mind:
> 
> * patch tailscale to auto reset - this would be best, but might be too complicated, especially when tailscale gets update, so this is more or less, last resort solution even though it would be most handy...

Basically, I would not recommend patching anything - it's really hard to maintain patches, especially for such a large project.

You may try to discuss it with upstream, and see if thry're interested in this.
If upstream rejects the idea, then it should be avoided AMAP, even it would be the best solution.

> * add new method/argument to initscript: reset
> 
> reset argument would more or less do all steps I described- there is unfortunately no way to completely automate this process, atleast method that I know about - It could be triggered to happen everytime /etc/config/network changes, but it seems a bit overkill.... because it would happen also everytime you change some other network interface as well, so I am thinking about manual way. After all, it is very rare that you want to change your setup when you've got it the way you like it.
> 
> # The question
> Would you recommend this method - or do you have something else in mind? I really hope you won't say that it must be patched, that sounds, a lot of work with a high failure rate, but again, I have a idea how to do this, but wouldn't want to be the only one deciding about it. I also have some plans to detect failed tailscale up due to changed options- and log this to syslog some how- but I haven't implemented this yet, as I am still doing some further testing and clean up of my work so far..
> 
> So, what do you think of this?
> 

I'm sorry that I cannot give any valid suggestion. I'm not familiar with tailscale, and also not interested in it.

Besides of this, I think here might not be a good place to discuss, as rare people (especially users) would check PRs. Try to open a new thread in OpenWrt forum, it may help.

Thanks.

> < snip >

oskarirauta:
@1715173329

zerotier indeed is a bit more simple solution, besides it gives you 4 advantages over zerotier:
 - instead of 20, you can have 25 clients on a free account
 - you can have more than one network
 - you can set those ip addresses statically
 - no strange networking solutions that are invisible outside application, like routing..

and ofcourse, zerotier's not written in Go, but whether it is a good or bad thing, is debatable. I also want to rather **not** patch it more. It's really unlikely that they would want to accept such patch in mainstream as it literally takes a way feature that they have designed :)

I know this wasn't a best place to ask question, but I thought this is more private here- if I ask this on the forum, it's likely that I get a hundred proposals that are 
 - impossible/difficult to archieve
 - suggestions to use zerotier instead
 - proposals that I would simply decline as they are really strange
 - proposals that would use other language/other software alongside, as it is used on X
 
There is 3rd solution, netbird - which I would choose over others, unless they would have a limited set of clients available, such as no mobile client at all. It's not that I would use it, but mobile client is often very crucial when you want to test that this stuff really works as intended, since you rarely have 2 independent networks in your hands with more or less direct access to internet.

I'll do some more futher testing and attempt to find a best solution to this. Current testing revealed a kind of a bug (or feature?) on tailscale, that needs some work as well- if tunnel is removed by other software (netifd) than tailscaled, it goes to unrecoverable state that requires server's restart. Tunnel is removed on everytime /etc/config/network has changed if tailscale was up.

Not bothering you any more further here :) Glad though I asked- there could had been a possibility of ultimate solution which would had never seen daylight without asking :)

PR #20428, opened 2023-02-01
mwan3: Handle iface reconnection while it's still offline

gentoo-root:
Fix the following use case:

1. Pings are lost, and mwan3 moves the interface to the offline state, score becomes 0.
2. The underlying netdev goes down and up (reconnects).
3. The initial_state of the interface is "online" (the default), so mwan3 moves the interface to the online state and continues tracking.
4. At this point, score is still 0. Pings continue being lost, so the interface should actually become offline, but as the score is already 0 and never passes the "up" watermark, the interface stays online.

By resetting score to its initial value before calling firstconnect, ensure that the tracking restarts from scratch after the netdev reconnects, fixing the above issue.

Maintainer: @feckert
Run tested: ipq806x, Netgear R7800, OpenWrt 045baca10b157fcc57461392cf27838d147ae5bd, tested the above scenario.

### DIFF
--- a/net/mwan3/files/usr/sbin/mwan3track
+++ b/net/mwan3/files/usr/sbin/mwan3track
@@ -405,6 +405,7 @@ main() {
 		fi
 		if [ "${IFUP_EVENT}" -eq 1 ]; then
 			LOG debug "Register ifup event on interface ${INTERFACE} (${DEVICE})"
+			score=$((down+up))
 			firstconnect
 			IFUP_EVENT=0
 		fi


feckert:
Sorry for the delay. I have a patch on my set, that addresses the same issue. Could you verify if this also fixes your issue?
```
 Eckert <fe@dev.tdt.de>
Date: Mon, 28 Nov 2022 10:13:25 +0100
Subject: [PATCH] mwan3: refactoring mwan3track action handling

Redmine-patch-id: 7396
Signed-off-by: Florian Eckert <fe@dev.tdt.de>
---
 net/mwan3/files/usr/sbin/mwan3track | 24 ++++++++++++------------
 1 file changed, 12 insertions(+), 12 deletions(-)

diff --git a/net/mwan3/files/usr/sbin/mwan3track b/net/mwan3/files/usr/sbin/mwan3track
index 89654f86ba05658cd23a6ab1008c8a48a137f1b5..dd60f51eaa358e75e222d3dfed69afe8b1a94fae 100755
--- a/net/mwan3/files/usr/sbin/mwan3track
+++ b/net/mwan3/files/usr/sbin/mwan3track
@@ -351,34 +351,34 @@ main() {
 			if [ $score -lt $up ]; then
 				score=0
 				[ ${keep_failure_interval} -eq 1 ] && sleep_time=$failure_interval
-			else
+			elif [ $score -eq $up ]; then
 				disconnecting
 				sleep_time=$failure_interval
-			fi
-
-			if [ $score -eq $up ]; then
 				disconnected
 				score=0
+			elif [ $score -gt $up ]; then
+				disconnecting
+				sleep_time=$failure_interval
 			fi
 		else
 			if [ $score -lt $((down+up)) ] && [ $lost -gt 0 ]; then
-				connecting
 				LOG info "Lost $((lost*count)) ping(s) on interface $INTERFACE ($DEVICE). Current score: $score"
 			fi
 
 			let score++
 			lost=0
 
-			if [ $score -gt $up ]; then
-				echo "online" > $MWAN3TRACK_STATUS_DIR/$INTERFACE/STATUS
-				score=$((down+up))
-			elif [ $score -le $up ]; then
+			if [ $score -lt $up ]; then
+				connecting
+				sleep_time=$recovery_interval
+			elif [ $score -eq $up ]; then
 				connecting
 				sleep_time=$recovery_interval
-			fi
-
-			if [ $score -eq $up ]; then
 				connected
+				score=$((down+up))
+			elif [ $score -gt $up ]; then
+				echo "online" > $MWAN3TRACK_STATUS_DIR/$INTERFACE/STATUS
+				score=$((down+up))
 			fi
 		fi
 
-- 
 #```

feckert:
I would like to apologies for the late reply. I am very busy at the moment. I am willing to merge this, as it makes sense. However, my use case is always that my initialized state is offline and not online. Can you please verify again that everything works in this state too and that we don't get a regression? I don't have a setup at hand right now.

feckert:
Thanks for your pullrequest, but this #20855 fixes also this issue.

feckert:
I apologise for hurting your feelings. There is no need to get personal! I have no problem with you.

Everyone here does it voluntarily or takes responsibility for a part of the software where they may not be paid.

Please let me explain my decision.
From my point of view, all I can say is that your fix always resets the score, regardless if the function firstconnect succeeded or not. So I think the other fix dose have less impact on other setups and configuration. Since I cannot check all mwan3 setups. If the problem persists, I am happy to incorporate your change.

feckert:
@gentoo-root I would like to apologies for my less than optimal behavior as a maintainer. I am also not pleased when a pull request is simply closed. 

Okay, let's leave it at that. How do we continue here? I now have two solutions to the same problem.

@gentoo-root:
The score is always reset when an up event comes from the netifd.
@anyuta1166:
The score is only reset when connected is called.

Wouldn't it make more sense if we reset the score directly in [connected](https://github.com/gentoo-root/openwrt-packages/blob/master/net/mwan3/files/usr/sbin/mwan3track#L130) function in the same way as we set the score in [disconnected](https://github.com/gentoo-root/openwrt-packages/blob/master/net/mwan3/files/usr/sbin/mwan3track#L114)?

Then we could define the `score` variable as globally `SCORE`?

anyuta1166:
> I agree that we should reset the score directly in connected function

I have created a new PR #20924 using this approach. Made some tests, seems to work fine. @feckert @gentoo-root what do you think about it?

anyuta1166:
> 3\. Do we want to try to fix as much as possible, or do we only care about this bug?

I agree with @feckert that a major rewrite should be done after a new OpenWRT release. Just fix the current bugs for now.

PR #20429, opened 2023-02-02
Fix (x86_64) squashfs builds

jmarcet:
Description:
Revert "python3: add 'ensurepip' to python3-pip sub-package"

This reverts commit 1f317dfb75216904c92c3c8a64e088fb6b4c551b.

I have to revert it or the squashfs partition of x86_64 images built from current master is not mountable.

I've done several clean builds so I'm pretty confident this causes the breakage.
I still don't understand the relation though.

### DIFF
--- a/lang/python/python3/Makefile
+++ b/lang/python/python3/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 include ../python3-version.mk
 
 PKG_NAME:=python3
-PKG_RELEASE:=3
+PKG_RELEASE:=2
 PKG_VERSION:=$(PYTHON3_VERSION).$(PYTHON3_VERSION_MICRO)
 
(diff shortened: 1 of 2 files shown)


commodo:
> I have to revert it or the squashfs partition of x86_64 images built from current master is not mountable.
> 
> I've done several clean builds so I'm pretty confident this causes the breakage. I still don't understand the relation though.

first thought is NAK;

with Python (being fat and all), it's easy to create partition sizes that won't work/mount/etc;

can you please share your config?
and i can look, and suggest something else;
unless i get more info/context, i would be against reverting this;

commodo:
ping?

i'd still like to see the `.config` produces the issue :)

commodo:
ping on this :)

BKPepe:
Let's close it. @commodo You tried, but you have not received any response. :)

PR #20433, opened 2023-02-02
adblock: fix gzip command in README

kidonng:
Maintainer: @dibdot
Compile tested: n/a
Run tested: n/a

Description: after the `gunzip` command above, the sources file no longer have a `.gz` extension.


dibdot:
@kidonng even for a plain text change we need a correct sign-off with your real first and last name.

dibdot:
superseded by https://github.com/openwrt/packages/commit/51634a960c8bd22fce5870462011cea6fbf4f0bc

Issue #20434, opened 2023-02-02
bind: no_ipv6 detection is broken


rucketeg:
Maintainer:  @nmeyerhans 
Environment: master

Description: bind is always starting with `-4` flag
a7b770eec4370087a5ccd27887386dac9266214e introduced code to detect if no usable ipv6 address are assigned and start bind with `-4` flag in such case. Unfortunately the code checking for that:
```
[ no_ipv6 ] && args="-4"
```
has a bug that always evaluates to true leading to bind starting with `-4` flag even if ipv6 addresses are routable. Bash `test` construct, aka `[ ]`, without any operator (as used above) is testing for string being not empty, the same as `-n` operator, more info [here](https://tldp.org/LDP/abs/html/testconstructs.html) and in the output of `$ help test`:
```
$ help test
(...)
    String operators:
    
      -z STRING      True if string is empty.
    
      -n STRING
         STRING      True if string is not empty.
(...)
```
This means that the code `[ no_ipv6 ]` is equivalent to `[ -n "no_ipv6" ]` which always evaluates to `true`.

nmeyerhans:
Fixed with #20435

PR #20435, opened 2023-02-02
bind: Fix ipv6 detection logic

rucketeg:
Bug was introduced in a7b770eec4370087a5ccd27887386dac9266214e and results in bind always stating with the `-4` flag.

Signed-off-by: Rucke Teg <rucketeg@protonmail.com>

Maintainer: @nmeyerhans
Compile tested: bcm2711, master
Run tested: bind starts without `-4` flag if ipv6 address is routable


rucketeg:
closes #20434

nmeyerhans (approved):

Issue #16311, opened 2021-08-10
ddns-script: njalla support


himekifee:
Maintainer: @chris5560 (find it by checking history of the package Makefile)
Environment: x86_64, VM, 21.02-rc3

Description:

[Njalla](https://njal.la/) is a privacy-aware domain service. It provides DDNS functionality, which can be found [here](https://njal.la/docs/ddns/). It would be great if this ddns can be supported since it is a perfect domain service for privacy.

tobiaspc:
See https://github.com/openwrt/packages/pull/20467

PR #20467, opened 2023-02-08
ddns-scripts: Add njal.la provider

tobiaspc:
Description: Add njal.la provider. Use the key as password. Username is not needed.

Maintainer: me / @chris5560
Compile tested: x86-64 (APU2), ath79 (Archer C7) 
Run tested: x86-64 (APU2)

Fixes https://github.com/openwrt/packages/issues/16311

Signed-off-by: Tobias Hilbig <web.tobias@hilbig-ffb.de>

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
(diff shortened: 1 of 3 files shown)


feckert:
@BKPepe it's a trival change but I don't know why the job is not selected by the runners!
I would merge this without successful CI/CD?

feckert:
I will merge it with CI/CD as it fails for another reason!

PR #20469, opened 2023-02-09
restic: update to 0.15.1

Torxgewinde:
Maintainer: Tom Stöveken <tom@naaa.de>
Compile tested: SDK for OpenWrt 22.03.3
Run tested: x86/64 @ Intel(R) Celeron(R) CPU N3160 @ 1.60GHz, OpenWrt 22.03.3, ran a backup to check it works

Description:
Updated to version 0.15.1

Signed-off-by: Tom Stöveken <tom@naaa.de>


hnyman:
PKG_RELEASE (the minor version item) should actually be reset to 1, as you are upgrading the whole main PKG_VERSION (the major version item).   The automatic github actions workflow gave you wrong advice, based on thinking that there was no upstream version upgrade.

You also have two commits, and they are titled similarly.

Please amend the PR and set PKG_RELEASE to 1 and squash the commits together, so that there is just one commit in the PR.  (`git rebase -i`  and `git push -f` work nicely with PRs.  Advice can be found https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#advice-on-pull-requests )

Torxgewinde:
Thanks @hnyman, I think I learned how to squash the commits and it should be Ok now.

PR #20475, opened 2023-02-10
treewide: prepare packages to OpenSSL 3.0 update

cotequeiroz:
Maintainers: @peter-stadler @acinonyx @ratkaj @fatbob313 

Description:
These are packages that need adjustments or version bumps to work with OpenSSL 3.0.  See openwrt/openwrt#11963 for its status.
I've compile-tested all of the packages, and run-tested gost_engine, nginx-util, and squid.

* Squid worries me some, since I had to do a major upgrade, which changed configuration files options.  It has uci integration, and I have not tried to change its configurations.  I did confirm it works when first installed, but that was all of it.
* The fix for nsd was submitted upstream in NLnetLabs/nsd#266.
* The fix for umurmur is one-liner from Gentoo.

If desired, I can move any of the packages into a separate PR.


neheb (approved):

cotequeiroz:
I left out boinc, libuhttpd, and apfree-wifidog by mistake.  I'm adding them now.
Ping @neheb @zhaojh329 @liudf0716 @N30dG @smoe

liudf0716:
@cotequeiroz get it

cotequeiroz:
@liudf0716 
If you tag a new wifidog-apfree release, then we can do a version bump instead of adding a patch.

fatbob313 (approved):

liudf0716:
@cotequeiroz 
Ok, we are now preparing a new release that will replace libiptext completely with nft, and at the same time we will fix this openssl3 problem.

cotequeiroz:
Thanks @liudf0716 

You did not mention a timeline, so I recommend we patch the current version and then you can take your time with the new release. The OpenSSL change is not a major one, as nft probably is. I'm not counting on it, but should we need a revert, it will be cleaner then.

As an alternative, I can take apfree-wifidog out of this PR, and mark openwrt/openwrt#11963 ready to merge. I don't control when it will be merged.

I intend to merge this PR on Feb 20, but it may have to be sooner if OpenWrt decide to update OpenSSL before then.

cotequeiroz:
I had to change the nsd patch because the original one did not work with FreeBSD.

liudf0716:
@cotequeiroz Ok, when I finish the release of apfree-wifidog, I will do all the work around it.

cotequeiroz:
nsd fixed the issue of building nsd-control without ssl in a different way: instead of removing unused object files, they added the missing definition to the source.  They went on to immediately add unix sockets functionality to nsd-control, so for them, their ssl fix was much less intrusive--when adding the functionality, they will need the extra objects, so no point in changing configure.ac to remove then and then revert that change.

However, for our use case, adding this functionality would be better left to the next release, and linking a list of object files to a program that prints an error message and exits is just a waste of space.  So, I'm leaving my original patch here, but will change its comments to indicate it should not be applied to the next release.

PR #20481, opened 2023-02-12
hamlib: Added init script and config files

ttytyper:
Enables configuring and starting rigctld in the same was as other OpenWrt services

Compile tested: x86/64, OpenWrt 22.03.3
Run tested: x86/64, OpenWrt 22.03.3, using an Elecraft KX3 radio and Kenwood TH-D72
Signed-off-by: Jesper Henriksen <mail-openwrt@jesper.io>


utils/hamlib/files/rigctld.conf
```diff
@@ -0,0 +1,24 @@
...
+	option serial_speed ''
+	option civaddr ''
+	option listen_addr ''
+	option port ''
+	option set_conf ''
+	option vfo ''
+	option verbose ''
+	option twiddle_timeout ''
+	option uplink ''
+	option debug_time_stamps ''
```
feckert, line 24: missing new line

utils/hamlib/files/rigctld.init
```diff
@@ -0,0 +1,76 @@
...
+
+		local i=0
+		while [ $i -lt "$value" ]; do
+			procd_append_param command "--$option"
+			let i++
+		done
+	done
+
+	procd_close_instance
+}
```
feckert, line 76: dito

ttytyper:
Well spotted. New lines should be in order now

PR #20484, opened 2023-02-12
transmission: retrieve boolean config opts using `config_get_bool`

salim-b:
Signed-off-by: Salim B <git@salim.space>

Maintainer: @dangowrt / CC: @neheb
Compile tested: -
Run tested: OpenWrt 22.03 snapshot

# Description

The transmission UCI config options

- `config_overwrite`
- `incomplete_dir_enabled`
- `watch_dir_enabled`

are all booleans, so we have to retrieve them using `config_get_bool` in order to make sure they are properly interpreted in case the user sets them to a keyword (`true`/`false`, `on`/`off` etc.) and not an integer (`0`/`1`), cf. https://openwrt.org/docs/guide-developer/config-scripting#reading_booleans

### DIFF
--- a/net/transmission/files/transmission.init
+++ b/net/transmission/files/transmission.init
@@ -58,15 +58,15 @@ transmission() {
 	local incomplete_dir
 	config_get incomplete_dir "$cfg" 'incomplete_dir' '/var/etc/transmission'
 	local incomplete_dir_enabled
-	config_get incomplete_dir_enabled "$cfg" 'incomplete_dir_enabled' 0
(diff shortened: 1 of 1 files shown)


neheb (approved):

neheb:
I don't use transmission anymore. Seems like a good change.

dangowrt:
Please add linebreaks to the commit message. Otherthise looks good to me.

salim-b:
> Please add linebreaks to the commit message.

Ok, done.

PR #20491, opened 2023-02-13
banip: release 0.8.0 (nft rewrite)

dibdot:
Maintainer: me / @dibdot
Compile tested: -
Run tested: Turris Omnia, OpenWrt SNAPSHOT r22040-eeba2a67ca. During the last 3 months different forum users have tested in sum 6 pre-releases to stabilize this piece of software. A big "Thank you" to all testers for their feedback and efforts.

Description:
- complete rewrite of banIP to support nftables
- all sets are handled in a separate nft table/namespace 'banIP'
- for incoming blocking it uses the inet input hook, for outgoing blocking it uses the inet forward hook
- full IPv4 and IPv6 support
- supports nft atomic set loading
- supports blocking by ASN numbers and by iso country codes
- 42 preconfigured external feeds are available, plus local allow- and blocklist
- supports local allow- and blocklist (IPv4, IPv6, CIDR notation or domain names)
- auto-add the uplink subnet to the local allowlist
- provides a small background log monitor to ban unsuccessful login attempts in real-time
- the logterms for the log monitor service can be freely defined via regex
- auto-add unsuccessful LuCI, nginx, Asterisk or ssh login attempts to the local blocklist
- fast feed processing as they are handled in parallel as background jobs
- per feed it can be defined whether the input chain or the forward chain should be blocked (default: both chains)
- automatic blocklist backup & restore, the backups will be used in case of download errors or during startup
- automatically selects one of the following download utilities with ssl support: aria2c, curl, uclient-fetch or wget
- supports a 'allowlist only' mode, this option restricts internet access from/to a small number of secure websites/IPs
- provides comprehensive runtime information
- provides a detailed set report
- provides a set search engine for certain IPs
- feed parsing by fast & flexible regex rulesets
- minimal status & error logging to syslog, enable debug logging to receive more output
- procd based init system support (start/stop/restart/reload/status/report/search)
- procd network interface trigger support
- ability to add new banIP feeds on your own
- add a readme with all available options/feeds to customize your installation to your needs
- a new LuCI frontend will be available soon


Neustradamus:
To follow :)

Ansuel:
Happy that this is finally ready... been waiting for this from a lot

PerkelSimon:
we just have to use that patch in our buildroot ?

PR #20497, opened 2023-02-14
igmpproxy: add support for whitelist/blacklist in UCI

Forst:
This permits allowing/denying specific multicast networks to be proxied.

* `whitelist` was added in 0.2 (pali/igmpproxy@65f777e7f66b55239d935c1cf81bb5abc0f6c89f)
* `blacklist` was added in 0.4 (pali/igmpproxy@cfad33001fc5031399bdd073048ca86889378678)

Maintainer: @nbd168 (as per Makefile)
Compile tested: ipq40xx / Linksys EA8300 / openwrt/openwrt@19988b66d07f44be757c4a0e41be899dbf8c72dd
Run tested: ipq40xx / Linksys EA8300 / openwrt/openwrt@19988b66d07f44be757c4a0e41be899dbf8c72dd

### DIFF
--- a/net/igmpproxy/Makefile
+++ b/net/igmpproxy/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=igmpproxy
 PKG_VERSION:=0.4
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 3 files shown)


Forst:
I think I'll redo with this a single `acl` list in UCI, instead of the two `whitelist` and `blacklist` lists.

Checked `igmpproxy` docs and source, and the relative order of `whitelist` and `blacklist` entries matters, and it should be configurable by the user. From what I can tell, having two lists in UCI doesn't provide this flexibility, or at least it's not straightforward.

Issue #5559, opened 2018-02-02
autossh work if manuel /etc/init.d/autossh start, but not on reboot, ssh exit error status 1


winfriedmuench:
tested with WRT54GS V2 and netgear WA801DN
LEDE 17.01.4 r3560-79f57e422d

If I login with ssh end run
```
/etc/init.d/autossh stop
/etc/init.d/autossh start
```
it works well.

After reboot, loop with
in syslog
```
user.info autossh[pid]: ssh exited with error status 1; restarting ssh 
```
after 5 seconds the same
No Idea

winfriedmuench:
Also restart over system/startup works.

winfriedmuench:
I tryed to move the private key to /etc/dropbear/id_dropbear. Its the same.
Tryed also:
```root@LEDE:/etc# cat rc.local 
```# Put your custom commands here that should be executed once
```# the system init finished. By default this file does nothing.
```sleep 180
```/etc/init.d/autossh restart
```exit 0
```root@LEDE:/etc#

same thing. AAAARG

dibdot:
@bk138

bk138:
I can just say it works for me... Would you mind posting your config @winfriedmuench ?

winfriedmuench:
sure
the hole one:
[backup-LEDE-2018-02-05.tar.gz](https://github.com/openwrt/packages/files/1695620/backup-LEDE-2018-02-05.tar.gz)

bk138:
Thank you! From a first glance, the `/443 ` in /etc/config/autossh looks weird to me. Have you tried running ssh with those args from the cmdline, with autossh stopped?

winfriedmuench:
Yes. It works perfect if i login with ssh and do /etc/init.d/autossh restart
i need 443 because of some firewall rules in other environments.

bk138:
I am not aware that ssh understand this slash syntax. Running `ssh -i /etc/dropbear/id_dropbear -N -T -R 5558:192.168.1.1:22 rlogin@asteriskpi1.dyn-ip24.de/443` (from you /etc/config/autossh) gives an error about the hostname.

winfriedmuench:
but it works well if startet on shell. The slash is implemented in the dropbear ssh. Have try it also with -p option. You are welcome to try it on my server. Key should be includet. Login not possible but remote port forwarding possible.

bk138:
[...]

Issue #5583, opened 2018-02-08
dropbear ssh no parameter for location of known_hosts file


winfriedmuench:
so I try to use autossh. It works well, but not on reboot. Diggin around found on reboot $HOME is / 
dropbear search known_hosts in $HOME/.ssh/known_hosts. There is no parameter to change that:
```
root@LEDE:~# ssh -o help
ssh: Available options:
	ExitOnForwardFailure
	UseSyslog

root@LEDE:~# ssh -V
Dropbear v2017.75
root@LEDE:~# #
```
would be nice to integrate parameter UserKnownHostsFile or GlobalKnownHostsFile. Or make shure that on boot $HOME is set to /root.
See original issue on [autossh](https://github.com/openwrt/packages/issues/5559#issuecomment-364055154)

dibdot:
seems to be a race condition in autossh package during boot. In my (booted) environment HOME always points to '/root'. I think it doesn't make sense to hardcode dropbear (which is not maintained at the package repo!) to use this directory. A fix should be implemented in autossh, e.g. switch init to procd, use network trigger and explicitly set the home dir environment variable or change the existing hotplug script.

@winfriedmuench please always add the current package maintainer to your issues 

for reference: @bk138 

Anyway I close this by now cause dropbear is not part of this repo. Issues with core components are tracked here: https://bugs.openwrt.org/

B0F1B0:
My content of `ls -all /root/.ssh/`
````
-rw-------    1 root     root            83 May 20 10:11 id_dropbear
-rw-r--r--    1 root     root           186 May 20 10:09 known_hosts
````
For me the following works 😎

````
cp /root/.ssh/* /.ssh/
````

````
service dropbear restart
````

PR #20499, opened 2023-02-15
autossh: bugfix: add env HOME for dropbear ssh

douniwan5788:
Fixed this 5 year old bug
https://github.com/openwrt/packages/issues/5559

Description:

### DIFF
--- a/net/autossh/files/autossh.init
+++ b/net/autossh/files/autossh.init
@@ -49,6 +49,9 @@ start_instance() {
 	procd_set_param command /usr/sbin/autossh ${forwarding} ${ssh}
 	procd_set_param respawn ${respawn_threshold:-3600} ${respawn_timeout:-5} ${respawn_retry:-5}
 	[ -n "$pidfile" ] && procd_set_param pidfile "$pidfile"
+	
+	#dropbear ssh seek the known_hosts file in $HOME/.ssh/known_hosts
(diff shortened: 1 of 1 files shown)


BKPepe:
CI is failing for you:
```
: === Checking commit '2dff3d84c129297d0e8c85b296b8d38d15fb62cc'
: Author name (douniwan5788) need to be your real name 'firstname lastname'
```

douniwan5788:
> 1. `PKG_RELEASE` needs to be increased in the Makefile
> 2. The commit description should be improved. It is not enough to say that you are fixing 5 years old bug, but also why. I see that the issue is closed and it was also locked as resolved. This means that you need to provide more details - what happened, why and what needs to be done about that.
> 
> Same as it is said here https://openwrt.org/submitting-patches

The issue has been closed as a solution was found. However, @winfriedmuench has opened [another issue](https://github.com/openwrt/packages/issues/5583) regarding dropbear. The bug and solution has also been mentioned in the [official documentation](https://openwrt.org/docs/guide-user/services/ssh/autossh), but it appears that nobody has fixed the issue yet.

BKPepe:
The other issue is also closed, most likely due to race condition. 

The page, which you refer to be official documentation, is a wiki, which means that anyone can modify it, create new pages, etc., and developers do not verify that the details written there are correct.

PR #20504, opened 2023-02-16
tvheadend: Fix ipv6 URL support 

sKyissKy:
Maintainer: me / @<github-user> (find it by checking history of the package Makefile)
Compile tested: x86_64
Run tested: x86_64

Description: Fix tvheadend ipv6 IPTV URL support


M95D:
Thanks for this PR.
I have a few things to discuss:

1) I think uriparser should be an option and turned off by default.
Tvheadend is already a very big package (for a router) and most people only use it in their LAN, where IPv6 is not needed.

2) I never used tvheadend with IPv6. Why is uriparser needed? What doesn't work without it?

3) The tests are failing. Download link doesn't seem to work.

BKPepe:
Closing due to inactivity as the comment was not adressed.

PR #20508, opened 2023-02-16
git: update to 2.39.2

paper42:
Maintainer: @tripolar
Run tested: armv7l, Turris Omnia, OpenWrt 21.02, tested git clone and gitweb

Description: GITWEB_* variable values don't work as they should since 2.38, so let's remove them and add a workaround. This issue was reported in https://lore.kernel.org/git/80eb3972-4960-5727-ce86-acc3a4425fd4@nic.cz/T/#u

* refresh patches

earlier today I opened #20506 which is a small security update, this PR replaces it, but I will leave whether you choose to merge the small security update first and then this one or merge this PR and close the small security update to you.

### DIFF
--- a/net/git/Makefile
+++ b/net/git/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=git
-PKG_VERSION:=2.34.7
+PKG_VERSION:=2.39.2
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@KERNEL/software/scm/git/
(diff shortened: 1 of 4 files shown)


oskarirauta (approved):

neheb:
conflicts

BKPepe:
Conflict is raised because I merged https://github.com/openwrt/packages/pull/20506 , but in this case, I am a little reluctant to merge this as by this update, we will break gitweb and nobody has replied to @paper42's email, yet.

paper42:
> I am a little reluctant to merge this as by this update, we will break gitweb and nobody has replied to @paper42's email, yet.

This won't break gitweb, I made sure to fix that in this PR. The fix is unfortunately a workaround and it would be much better if they could fix it upstream, but this is what we have for now.

I think it's important to have the latest git in master because if we keep using the years old version, a new openwrt release will be tagged with the old version and we will be stuck with it until EOL of the new OpenWrt version and during that time it might stop getting security updates.

PR #20515, opened 2023-02-17
cni-protocol: new package

oskarirauta:
simple protocol support script for netifd.

netifd protocol support for cni networks makes
defining network for podman and other similar
systems using cni networking much easier and simpler.

with cni protocol support, on a cni network, where firewall and portmapper is disabled, you may control firewalling with openwrt's standard firewall configuration.

for example, create a container that hosts web content on port 80 with static ip on your cni network, if your network is 10.88.0.0/16, use for eg. 10.88.0.101 as your containers static ip address. Create a zone, cni to your firewall and add your interface to it. You could manage all this without protocol support, but this makes it a bit easier, especially when this is requirement for my upcoming luci protocol support for same thing- so no tinkering necessary if someone feels more comfortable on LuCi, or just wants to see his interfaces page without warning about unsupported protocol extension.

Now you can easily set up redirectiong to 10.88.0.101:80 to expose it's port 80 to wan for serving your website.
Protocol has only one setting: _device_, on podman this often is **cni-podman0**.

Maintainer: me / @oskarirauta

### DIFF
--- a/net/cni-protocol/Makefile
+++ b/net/cni-protocol/Makefile
@@ -1,0 +1,51 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=cni-protocol
+PKG_VERSION:=20230217
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/cni-protocol
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=cni netifd protocol
+  PKGARCH:=all
+endef
+
+define Package/cni-protocol/description
+  protocol support for cni networks for netifd
+  makes defining network for podman and other similar
+  systems using cni networking much easier and simpler.
+
+  with cni protocol support, on a network, where firewall
+  and portmapper management is disabled, you may control
+  firewalling with openwrt's default firewall configuration.
+
+  for example, create a container that hosts web content on
+  port 80 with static ip on your cni network, if your
+  network is 10.88.0.0/16, use for eg. 10.88.0.101 as
+  your containers static ip address. Create a zone, cni
+  to your firewall and add your interface to it.
+
+  Now you can easily set up redirectiong to 10.88.0.101:80
+  to expose it's port 80 to wan for serving your website.
+
+  Protocol has one setting: device, on podman this often
+  is cni-podman0.
+endef
+
+define Build/Configure
+endef
+
+define Build/Compile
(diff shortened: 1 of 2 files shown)


oskarirauta:
@1715173329 

This a no-go?

1715173329:
I think this can be added into the existing cni-plugins package directly.

oskarirauta:
@1715173329 I could do that. Personally, I don't use it **with** cni anymore, as I have moved to netavark (podman moving away from cni...) - but for the moment rust-lang support is still making it's way to openwrt, so I haven't made a PR on netavark package, as it requires language support that doesn't exist yet.

Maybe I should put it on cni for now and later when rust-lang is *ready*, separate it..?

1715173329:
if you gonna reuse it with netavark, then I agree to keep this separated.

oskarirauta:
@1715173329 

Yes, I don't think we should need identical protocol for netavark, or any other other similar equillavent and I think it's not a real world problem if protocol is named as cni; it sounds more universal than "netavark", when it works for any solution like this.

My private version though for this reason isn't depending on cni-plugins as there's no reason to install it when I use alternative (podman preferred) solution. Should I then remove depency on this PR too, already now?

1715173329:
> My private version though for this reason isn't depending on cni-plugins as there's no reason to install it when I use alternative (podman preferred) solution. Should I then remove depency on this PR too, already now?

How about adding cni-protocol to cni-plugins's dependencies list? This script is tiny and does not take many bytes.
I think it could be easier than resolve dependeny issue in the cni-protocol itself.

oskarirauta:
What a splendid idea! I'll first remove depency from this....

oskarirauta:
I also added mention about netavark (and others if there are any others, for example for other container managers..) to commit message.

net/cni-protocol/Makefile
```diff
@@ -0,0 +1,54 @@
...
+
+PKG_NAME:=cni-protocol
+PKG_VERSION:=20230217
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/cni-protocol/Default
```
1715173329, line 11: this can be removed

1715173329 (approved):

net/cni-protocol/Makefile
1715173329: Yep!

oskarirauta: Done

PR #20529, opened 2023-02-19
ddns-scripts: enable IPv6 for easydns.com

braewoods:
easydns.com has supported IPv6 for awhile now using the same update URL as IPv4. This duplicates the IPv4 entry for IPv6 to enable support for it.

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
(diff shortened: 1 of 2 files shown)


feckert:
Please use a correct **Signed-off-by** with your real name and a valid email address.

braewoods:
This should suffice.

feckert:
Sorry i forgot, that you also have to bump the [PKG_RELEASE number](https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L11)

braewoods:
Ok. I have done that now.

feckert:
Thanks for your contribution!

PR #20531, opened 2023-02-19
curl: update to 1.88.0-1

stangri:
Maintainer: me
Compile tested: x86_64, Sophos XG-105w, OpenWrt 22.03.3
Run tested: x86_64, Sophos XG-105w, OpenWrt 22.03.3, use with https-dns-proxy

Description:
* https://curl.se/changes.html#7_88_0

### DIFF
--- a/net/curl/Makefile
+++ b/net/curl/Makefile
@@ -9,15 +9,15 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/nls.mk
 
 PKG_NAME:=curl
-PKG_VERSION:=7.86.0
-PKG_RELEASE:=$(AUTORELEASE)
+PKG_VERSION:=7.88.0
+PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


stangri:
@trippleflux can you test this with `perl-www-curl`?

BKPepe:
There is a typo in the commit subject and we don't need to have there PKG_RELEASE as it is obvious. We are not doing anything special about the curl versioning, so it should be ``curl: update to 7.88.0``

paper42:
1.88.0 apparently breaks HTTP/2, so it should not be merged as is https://lore.kernel.org/distributions/e3306ab8-3656-3128-40f8-f76e6871a3a3@footclan.ninja/T/#u https://github.com/curl/curl/issues/10525

1.88.1 was released a few hours ago which should fix this issue.

BKPepe:
In that case, closing.

paper42:
> In that case, closing.

I think this PR could have been just updated to 1.88.1 instead of closing it

BKPepe:
From my point of view, this prevents it could be merged by others who have commit access.

PR #20546, opened 2023-02-21
getdns: fix compilation with OPENSSL_NO_DEPRECATED

dhewg:
SSL_get_peer_certificate() is deprecated, OpenSSL v3.0 added SSL_get0_peer_certificate() and SSL_get1_peer_certificate().

Use the latter since the return value is explicitely X509_free()ed here, see [0].

[0] https://www.openssl.org/docs/manmaster/man3/SSL_get_peer_certificate.html


cotequeiroz (changes requested):
Hi.

I think this change should be fixed upstream.  While I do appreciate keeping up the code with current API, I don't think we should keep the patches here.  Instead, they should be sent straight upstream, then we can just update to a new version.

I don't mind submitting patches upstream and here at the same time; in that case, bump `PKG_RELEASE`.

PR #20548, opened 2023-02-22
ddns-scripts: fix URL endpoint to use ns1 latest API endpoints

kaypeter87:
Signed-off-by: Peter Kay <contact@kaypeter.com>

`NS1.com` uses a different API route endpoint to update records on their DNS domains. Currently, it uses `/v1/zones/{zoneName}/{domain}/{type}` found on their [documentation](https://ns1.com/api?docId=2185). The `ddns-script` used in openwrt instead uses  `/v1/zones/{username}/{domain}/{type}`.

I changed the shell script with the URL changes on my `openwrt` deployment and was able to get a successful API call:

```
 044132       : Detect local IP on 'network'
 044132       : Local IP 'REDACTED' detected on network 'wan'
 044132       : Forced Update - L: 'REDACTED' == R: 'REDACTED'
 044132       : parsing script '/usr/lib/ddns/update_ns1_com.sh'
 044134       : NS1 answered: {"domain":"REDACTED","meta":{},"filters":[],"ttl":3600,"zone_name":"REDACTED","networks":[0],"zone":"REDACTED","blocked_tags":[],"id":"REDACTED","regions":{},"type":"A","use_client_subnet":true,"tags":{},"answers":[{"answer":["REDACTED"],"id":"REDACTED"}],"link":null,"tier":1}
 044134  info : Forced update successful - IP: 'REDACTED' send
 044134       : Waiting 600 seconds (Check Interval)
 ```
 
 Should also probably update the `Password` field in the `ns1` DDNS Service `Basic Settings` to `API Token`. Password will not work here.

### DIFF
--- a/net/ddns-scripts/files/usr/lib/ddns/update_ns1_com.sh
+++ b/net/ddns-scripts/files/usr/lib/ddns/update_ns1_com.sh
@@ -25,7 +25,7 @@ json_close_array
 json_close_object
 json_close_array
 
-__URL="$__ENDPOINT/zones/$username/$domain/$__RRTYPE"
+__URL="$__ENDPOINT/zones/$domain/$lookup_host/$__RRTYPE"
 
 __STATUS=$(curl -s -X POST "$__URL" \
 	-H "X-NSONE-Key: $password" \


kaypeter87:
Hello!

Was wondering if I could get a quick review on this simple PR that fixes the `ns1` ddns script.

Thank you!

feckert:
We only need one commit for this change and not 4. PLease fixup your change locally and make a git push --force against this branche. Also please update the PKG_RELEASE by one.

Since I don't use this ddns provider, I wonder if there has been an API change? Does the current one no longer work or has it never worked? I don't want to commit something that might make an old configuration unusable just because we make this change. Maybe we need a v2 for the same ddns provider.

kaypeter87:
@feckert I'm not sure if it has ever worked. As you can see in the NS1 documentation, it uses `{zoneName}` in its API routes and has never used username for anything. NS1 has been on API `v1` for a while now.

feckert:
Could you please rewrite your [commit message](https://github.com/openwrt/packages/pull/20548/commits/4aac51df3f2022816b8ff23969a2617966fc87a3)? You have 2 Signed-off-by!

kaypeter87:
Apologies, thanks for the catch. It has been fixed!

feckert:
Sorry but this still does not fit. The Signed-of-by ist the last line in the commit message.
You could use `git commt --amend -s` to fix this in your branch.

kaypeter87:
Signed off line is on the last line of the commit message now. Thanks!

feckert:
I have now had a look at the [NS1 documentation](https://ns1.com/api?docId=2185). If I understand it correctly, the pull request is not quite finished. If the update fails, then it will try to create the [new record](https://github.com/openwrt/packages/blob/master/net/ddns-scripts/files/usr/lib/ddns/update_ns1_com.sh#L42). But then we have to change [that](https://github.com/openwrt/packages/blob/master/net/ddns-scripts/files/usr/lib/ddns/update_ns1_com.sh#L45) there too? You still use username and domain there.

Idependently:

- The `zoneName` must not be a FQDN but should (see documentation). It must only be a unique identifier. And we use in openwrt the `$username`.

- For the `recordDomain` we use the `$domain`.

- For the API key we use the `$password` value from the uci config. This is unpleasant, but the OpenWrt config doesn't allow more without having to rebuild everything.

From my point of view, that fits after all. Please describe again exactly why we should change this now

kaypeter87:
I believe there was some confusion on my part on what `$username` is being used for. From what you explained, there should be no changes to the script based on what you explained. Apologies for the confusion! Thanks for the explanations!

PR #20552, opened 2023-02-22
nfs-kernel-server: update to v2.6.2

oskarirauta:
Also added patch that is from alpine's same package to assist building on musl.
Some overrides for hostpkg configure were also necessary to make it possible to build **on** musl.

Maintainer: Peter Wagner / @tripolar 

### DIFF
--- a/net/nfs-kernel-server/Makefile
+++ b/net/nfs-kernel-server/Makefile
@@ -7,9 +7,9 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nfs-kernel-server
-PKG_VERSION:=2.5.4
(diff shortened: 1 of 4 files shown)


oskarirauta:
@1715173329 

Some tests are failing, problem with my PR or with tests?

1715173329:
see https://github.com/openwrt/openwrt/pull/12049

1715173329:
You don't need to add this dependency to nfs, it's caused by openssl and already fixed in latest tree.

oskarirauta:
Oh, sorry- I'll revert it :)

oskarirauta:
@1715173329 review maybe..?

PR #20553, opened 2023-02-22
open-vm-tools: update to 12.1.5

oskarirauta:
Maintainer: @neheb @srchack

Description:
patches refreshed and one mainstream merged patch removed.
added also --disable-glibc-check to configure args to allow building on hosts that use musl.


oskarirauta:
@1715173329

This failing some tests because TARGET_x86 depency or something else wrong?

1715173329:
Yep

PR #20554, opened 2023-02-22
https-dns-proxy: 2022-10-15-11 update

stangri:
Maintainer: me
Compile tested: x86_64, Sophos XG-105w, OpenWrt 22.03.3
Run tested: x86_64, Sophos XG-105w, OpenWrt 22.03.3, use https-dns-proxy for dns resolution

Description:
* config file update
* introduce boot() function

Signed-off-by: Stan Grishin [stangri@melmac.ca](mailto:stangri@melmac.ca)

### DIFF
--- a/net/https-dns-proxy/Makefile
+++ b/net/https-dns-proxy/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=https-dns-proxy
 PKG_VERSION:=2022-10-15
-PKG_RELEASE:=10
+PKG_RELEASE:=11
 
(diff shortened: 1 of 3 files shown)


BKPepe:
I would say that your commit subject is not exactly alright as you are saying update, but you're doing changes - **changing listen port** without saying why and adding boot function in init script. So, it should be improved.

BKPepe:
Maybe, you just forgot to force push it as I was talking about [contributing guidelines](https://openwrt.org/submitting-patches#submission_guidelines).

> it should explain to a competent reader why you made this commit.

It will also make a good git history, so you would, in the future, why you did the commit, as it is not clear now, though.

Issue #20558, opened 2023-02-23
acme renew should update haproxy.pem


AlvinZhu:
https://github.com/openwrt/packages/blob/0a15829d7377cb989694fb3cef5b789d6fbad1d0/net/acme/files/run.sh#L357
haproxy.pem only generated on the first cert Issuing.

https://github.com/openwrt/packages/blob/0a15829d7377cb989694fb3cef5b789d6fbad1d0/net/acme/files/run.sh#L270
haproxy.pem need update when cert renew

PR #20559, opened 2023-02-23
[22.03] acme: fix haproxy.pem renew

AlvinZhu:
fix issuse https://github.com/openwrt/packages/issues/20558

Maintainer: Toke Høiland-Jørgensen <toke@toke.dk>
Compile tested: (x84_64, OpenWrt 22.03.4)
Run tested: (x84_64, OpenWrt 22.03.4, tests done)

Description:

### DIFF
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -266,7 +266,7 @@ issue_cert() {
 			log "Found previous cert config. Issuing renew."
 			[ "$keylength_ecc" -eq "1" ] && acme_args="$acme_args --ecc"
 			run_acme --home "$STATE_DIR" --renew -d "$main_domain" $acme_args && ret=0 || ret=1
-			if [ -e /etc/init.d/haproxy ] && [ -w /etc/haproxy.cfg ] && [ "$update_haproxy" -eq "1" ]; then
+			if [ "$ret" -eq 0 ] && [ -e "${domain_dir}/${main_domain}-haproxy.pem" ]; then
 				cat "${domain_dir}/${main_domain}.key" "${domain_dir}/fullchain.cer" > "${domain_dir}/${main_domain}-haproxy.pem"
 			fi
 			post_checks
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -266,6 +266,9 @@ issue_cert() {
 			log "Found previous cert config. Issuing renew."
 			[ "$keylength_ecc" -eq "1" ] && acme_args="$acme_args --ecc"
 			run_acme --home "$STATE_DIR" --renew -d "$main_domain" $acme_args && ret=0 || ret=1
+			if [ -e /etc/init.d/haproxy ] && [ -w /etc/haproxy.cfg ] && [ "$update_haproxy" -eq "1" ]; then
+				cat "${domain_dir}/${main_domain}.key" "${domain_dir}/fullchain.cer" > "${domain_dir}/${main_domain}-haproxy.pem"
+			fi
 			post_checks
 			return $ret
 		fi


1715173329:
And I think this should go for the master branch firstly.
Cc @tohojo @hgl

hgl:
The acme package in the master branch already addresses this.

net/acme/files/run.sh
```diff
@@ -266,6 +266,9 @@ issue_cert() {
 			log "Found previous cert config. Issuing renew."
 			[ "$keylength_ecc" -eq "1" ] && acme_args="$acme_args --ecc"
 			run_acme --home "$STATE_DIR" --renew -d "$main_domain" $acme_args && ret=0 || ret=1
+			if [ -e /etc/init.d/haproxy ] && [ -w /etc/haproxy.cfg ] && [ "$update_haproxy" -eq "1" ]; then
```
tohojo, line 269: I think it's better to just make this a check for the file, and we should check the return code as well. I.e., `if [ "$ret" -eq 0 ] && [ -e "${domain_dir}/${main_domain}-haproxy.pem" ]`

tohojo (changes requested):
I'm OK with taking this as a bugfix for the 22.03 branch only (as this should not be an issue on master as @hgl remarked). However, we shouldn't just duplicate the check, see below...

net/acme/files/run.sh
AlvinZhu: OK

tohojo: Actually, looking at this again, I think it's better to move the combining of the certificates into `post_checks()` before the restart of haproxy (also moving the existing instance of it). That way it'll be called from both code paths and we won't have the duplicate code. Sorry, should have suggested that the first time around...

AlvinZhu: But there is no variable main_domain in the function post_checks(), so I don't know how to change it.

tohojo: Just add it :)

As in, add this to the top of post_checks:

```sh
	local main_domain
	main_domain="$1"
```

and then add the $main_domain as a parameter for both calls to the function :)

AlvinZhu: line 166 if run.sh
```bash
err_out() {
	post_checks
	exit 1
}

int_out() {
	post_checks
	trap - INT
	kill -INT $$
}
```
I still don't know how to move it.
There are many things to change. I think maybe it's simpler and clearer to change only one place to fix the bug.

tohojo: Something like this should work?

```diff
diff --git a/net/acme/files/run.sh b/net/acme/files/run.sh
index 615cbf22ab32..7f0955ea9ef5 100644
--- a/net/acme/files/run.sh
+++ b/net/acme/files/run.sh
@@ -135,6 +135,11 @@ pre_checks() {
 }
 
 post_checks() {
+       local main_domain
+       local domain_dir
+       main_domain="$1"
+       domain_dir="$2"
+
        log "Running post checks (cleanup)."
        # $NFT_HANDLE contains the string 'handle XX' so pass it unquoted to nft
        [ -n "$NFT_HANDLE" ] && nft delete rule inet fw4 input $NFT_HANDLE
@@ -154,6 +159,9 @@ post_checks() {
        fi
 
        if [ -e /etc/init.d/haproxy ] && [ "$UPDATE_HAPROXY" -eq 1 ] ; then
+                if [ -n "$main_domain" ]; then
+                       cat "${domain_dir}/${main_domain}.key" "${domain_dir}/fullchain.cer" > "${domain_dir}/${main_domain}-haproxy.pem"
+                fi
                /etc/init.d/haproxy restart
        fi
 
@@ -266,7 +274,7 @@ issue_cert() {
                        log "Found previous cert config. Issuing renew."
                        [ "$keylength_ecc" -eq "1" ] && acme_args="$acme_args --ecc"
                        run_acme --home "$STATE_DIR" --renew -d "$main_domain" $acme_args && ret=0 || ret=1
-                       post_checks
+                       post_checks "$main_domain" "$domain_dir"
                        return $ret
                fi
        fi
@@ -354,12 +362,11 @@ issue_cert() {
        fi
 
        if [ -e /etc/init.d/haproxy ] && [ -w /etc/haproxy.cfg ] && [ "$update_haproxy" -eq "1" ]; then
-               cat "${domain_dir}/${main_domain}.key" "${domain_dir}/fullchain.cer" > "${domain_dir}/${main_domain}-haproxy.pem"
                sed -i "s#bind :::443 v4v6 ssl crt .* alpn#bind :::443 v4v6 ssl crt ${domain_dir}/${main_domain}-haproxy.pem alpn#g" /etc/haproxy.cfg
                # commit and reload is in post_checks
        fi
 
-       post_checks
+       post_checks "$main_domain" "$domain_dir"
 }
 
 load_vars() {
```

PR #20563, opened 2023-02-23
clamav: update to 0.104.4

paper42:
Maintainer: @lucize @ratkaj 
Compile tested: armv7l, Turris Omnia, OpenWrt 22.03
Run tested: armv7l, Turris Omnia, OpenWrt 22.03

Description: 0.105.0+ needs rust, 0.103.x is the last lts version which doesn't need rust that is still getting fixes, but unfortunately OpenWrt doesn't use this version. 0.104 is not getting any fixes anymore because it's not an LTS branch, but at least this is the latest patch version that fixes some CVEs.

This should be backported to 22.03 and 21.03 since it's a patch release that fixes security issues.


lucize:
there is some work on rust, hopefully will be able to have the update in future

paper42:
The CI failure seems to be from openssl, so unrelated to this PR.

lucize:
LGTM

PR #20575, opened 2023-02-27
podman: update to v4.4.2

oskarirauta:
Security:
 - This release fixes CVE-2023-0778, which allowed a malicious user to potentially replace a normal file in a volume with a symlink while exporting the volume, allowing for access to arbitrary files on the host file system.

Bugfixes:
 - Fixed a bug where containers started via the podman-kube systemd template would always use the "passthrough" log driver.
 - Fixed a bug where pulls would unexpectedly encounter an EOF error. Now, Podman automatically transparently resumes aborted pull connections.
 - Fixed a race condition in Podman's signal proxy.

Misc:
 - Updated the containers/image library to v5.24.1.

And patch also refreshed

Compile tested: x86-64/latest git
Run tested: x86-64/latest git


oskarirauta:
@1715173329 @neheb

That security fix sounds important...

PR #20581, opened 2023-02-28
acme: merge cli into init script

### LINKED ISSUE #20568: acme: deprecate the CLI
Maintainer: @tohojo 

Description:

It seems the convention is to stick util-like commands to the init script. What do you think if we deprecate the CLI and use `/etc/init.d/acme get` instead? Could we simply dropped the CLI, since it's not really an API and the new version of the acme package is only available in snapshot?

hgl:
Maintainer: @tohojo 
Compile tested: x86 22.03.2
Run tested: x86 22.03.2

Fixes #20568


net/acme-common/Makefile
```diff
@@ -52,13 +50,13 @@ endef
...
-grep -q '/usr/bin/acme' /etc/crontabs/root 2>/dev/null && exit 0
-echo "0 0 * * * /usr/bin/acme get" >> /etc/crontabs/root
+grep -q '/etc/init.d/acme' /etc/crontabs/root 2>/dev/null && exit 0
+echo "0 0 * * * /etc/init.d/acme start" >> /etc/crontabs/root
 endef
 
 define Package/acme-common/prerm
 #!/bin/sh
-sed -i '\|/usr/bin/acme|d' /etc/crontabs/root
+sed -i '\|/etc/init.d/acme|d' /etc/crontabs/root
```
tohojo, line 59: Does this trigger on an update? Otherwise we need to add some logic to get rid of the old crontab entry...

hgl: If it does, it should trigger the one in the old package?

hgl: I've created a mock package to test it. Indeed the old prerm was run. So this should be good to merge.

BTW, I'm not sure if our uci-defaults necessary. I remember the reason for its existence is to support package being compiled into the firmware image, but if that's necessary, it seems to mean all packages should provide uci-defaults if they use postinst. But I don't see other packages do this, for example, mail/fdm.

Are you sure embedded packages won't run postinst? I don't compile firmware image myself, so I can't test right now.

tohojo: Alright, cool, thanks for checking. And yeah, I'm fairly sure postinst is not run by the image builder (it would have to execute in the context of the image, which may be an entirely different arch), so I think we'll need to keep it...

stangri:
@tgl @tohojo is there a transition guide for people who used standalone CLI?

tohojo:
Stan Grishin ***@***.***> writes:

> @tgl @tohojo is there a transition guide for people who used
> standalone CLI?

You mean you manually use the upstream acme.sh ? We still ship that, so
you just need to adjust the path you execute it at...

stangri:
@tohojo Thanks for your prompt reply, I may have misunderstood the intention of this PR.

PR #20586, opened 2023-03-03
acme: fix incompatibilty with image builder

hgl:
Maintainer: @tohojo 

Description:

I'm not sure if directly writing to `$IPKG_INSTROOT/etc/crontabs/root` works. Will just leave uci-defaults alone for the time being. Using image builder is on my todo list tho, will update uci-defaults once I get to that.

### DIFF
--- a/net/acme-common/files/acme.init
+++ b/net/acme-common/files/acme.init
@@ -9,7 +9,7 @@ HOOK=/usr/lib/acme/hook
 LOG_TAG=acme
 
 # shellcheck source=net/acme/files/functions.sh
-. /usr/lib/acme/functions.sh
+. "$IPKG_INSTROOT/usr/lib/acme/functions.sh"
 
 cleanup() {
 	log debug "cleaning up"
--- a/net/acme-common/Makefile
+++ b/net/acme-common/Makefile
@@ -50,13 +50,17 @@ endef
 
 define Package/acme-common/postinst
 #!/bin/sh
-grep -q '/etc/init.d/acme' /etc/crontabs/root 2>/dev/null && exit 0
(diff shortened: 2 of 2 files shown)


hgl:
Pretty strange that this issue didn't pop up when acme was first updated. Any idea why the init script change surfaced it? Seems unrelated.

net/acme-common/files/acme.init
```diff
@@ -9,7 +9,7 @@ HOOK=/usr/lib/acme/hook
 LOG_TAG=acme
 
 # shellcheck source=net/acme/files/functions.sh
-. /usr/lib/acme/functions.sh
+. "$IPKG_INSTROOT/usr/lib/acme/functions.sh"
```
tohojo, line 12: I don't think we should have this variable check in the init script.

Since we've removed that START variable anyway, how about just bailing out before the include if the parameter is `enable`? I believe that should fix the issue we were seeing...

hgl: I don't think I understand. The image builder will run the init script, although I'm not sure with what arguments. Why only check for `enable`?

If I understand it correctly, the image builder sets `IPKG_INSTROOT`, so `source` should find the `functions.sh` file, and in runtime, `IPKG_INSTROOT` is empty, `source` should also find the `functions.sh` file. Isn't that how it should be fixed?

tohojo: Yeah, it will, I just think it's ugly to leak that kind of package manager-specific variables into an init script.

However, it seems to be used in a number of other init files as well, so okay, let's go with that solution...

tohojo:
I agree it's weird that it surfaced all of a sudden, but no I don't have any good idea for why it would. Maybe it's just that no one built any images with the acme package in them before now?

PR #20590, opened 2023-03-05
upodman: new package

oskarirauta:
Maintainer: me / @oskarirauta

This software can be used as a mini "remote controller" to podman.

One potential use case would be container management in LuCi. LuCi already has a quite comprehensive add-on for docker which I once ported to support podman (all but load and ram stats along with logs that caused some unexpected difficulties) for most parts and added POD support. Still, I think that full container management is best done in a terminal with provided tools and I loved how it was done with Cockpit, so it inspired me and I started to build something of my own based on it and from that idea, upodman was born.

You cannot create/remove pods/containers/images/networks; mostly you can use upodman to retrieve information that would normally be available through Rest API socket, but this will help you get it through ubus; there are also methods available that can be used to start/stop/restart container/pod - creating new pod or container has so many available options that I see no point of trying to integrate them to tool like this; u in front of name is _micro_ - so that's what this is for, some small podman management..

Not relevant to here, but I also have a luci-app-upodman using this, which provides interface such as in this example screenshot; I basicly wrote this application for that luci extension.

upodman-screenshot

(In the image, there is nginx inside pod called servers, and then container agitated_ride that does not belong to any pod)

### DIFF
--- a/utils/upodman/Makefile
+++ b/utils/upodman/Makefile
@@ -1,0 +1,106 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=upodman
+PKG_VERSION:=0.98.1
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/oskarirauta/upodman/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=c05eddfe584fcb4e7e1a8893d676356b6332fd3ee6d292793097d16dcc33e6b9
+
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/upodman
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Podman manager via Ubus
+  URL:=https://github.com/oskarirauta/upodman
+  DEPENDS:=+libjson-c +libstdcpp +libubox +libuci +libblobmsg-json +libubus +podman
+endef
+
+define Package/upodman/description
+  upodman extends ubus with bindings to podman and allows
+  controlling and statistics over pods and containers through
+  ubus.
+
+  upodman is written in C++ except for ubus parts which are
+  externed from C. It uses staticly linked libcurl for unix
+  socket operations, because system's default libcurl might
+  not support unix sockets.
+endef
+
+MAKE_VARS += \
+	BUILD_DIR="$(PKG_BUILD_DIR)"
+
+CURL_VERSION:=7.87.0
+CURL_PATH:=$(PKG_BUILD_DIR)/curl-$(CURL_VERSION)
+CONFIGURE_PATH:=curl-$(CURL_VERSION)
+
+CONFIGURE_ARGS+= \
+		--disable-debug --enable-optimize --disable-curldebug --disable-ares \
+		--enable-rt --disable-ech --enable-shared=no --enable-static=yes --enable-http \
+		--disable-ftp --disable-file --disable-ldap --disable-ldaps --disable-rtsp \
+		--disable-proxy --disable-dict --disable-telnet --disable-tftp --disable-pop3 \
+		--disable-imap --disable-smb --disable-smtp --disable-gopher --disable-mqtt \
+		--disable-manual --disable-ipv6 --enable-pthreads --enable-verbose --disable-sspi \
+		--disable-crypto-auth --disable-ntlm --disable-ntlm-wb --disable-tls-srp \
+		--enable-unix-sockets --disable-cookies --disable-socketpair --disable-http-auth \
+		--disable-doh --disable-mime --disable-dateparse --disable-netrc --disable-progress-meter \
+		--disable-dnsshuffle --enable-get-easy-options --disable-alt-svc --disable-headers-api \
+		--disable-hsts --disable-websockets --without-schannel --without-secure-transport \
+		--without-amissl --without-ssl --without-zlib --without-brotli --without-zstd \
+		--without-gssapi --without-ca-bundle --without-libpsl --without-libgsasl \
+		--without-libssh2 --without-libssh --without-wolfssh --without-librtmp \
+		--without-winidn --without-libidn2 --without-nghttp2 --without-ngtcp2 \
+		--without-nghttp3 --without-quiche --without-msh3 \
+		--without-zsh-functions-dir --without-fish-functions-dir
+
+define Build/Configure
+endef
+
+define Build/Prepare
+	$(call Build/Prepare/Default)
+	mkdir -p $(CURL_PATH)
+	$(TAR) xzf $(PKG_BUILD_DIR)/dl/curl-$(CURL_VERSION).tar.gz -C $(PKG_BUILD_DIR)
+endef
+
+define Build/Configure
+	$(call Build/Configure/Default)
+	$(SED) 's#DIST_SUBDIRS = \$$(SUBDIRS) tests packages scripts include docs#DIST_SUBDIRS = \$$(SUBDIRS) packages scripts include#g' $(CURL_PATH)/Makefile
+	$(SED) 's#cd docs && \$$(MAKE) install#echo not-installing-docs#g' $(CURL_PATH)/Makefile
+	$(SED) 's#cd docs/libcurl && \$$(MAKE) install#echo not-installing-docs#g' $(CURL_PATH)/Makefile
+	$(SED) 's#SUBDIRS = ../docs#SUBDIRS =#g' $(CURL_PATH)/src/Makefile
+	$(SED) 's#SUBDIRS = ../docs#SUBDIRS =#g' $(CURL_PATH)/src/Makefile.am
+endef
+
+define Build/Compile/Curl
+	+$(MAKE_VARS) \
+	$(MAKE) $(PKG_JOBS) -C $(CURL_PATH) \
+		$(MAKE_FLAGS);
+endef
+
+define Build/Install/Curl
+	$(MAKE_VARS) \
+	$(MAKE) -C $(CURL_PATH) \
+		$(MAKE_FLAGS) \
+	DESTDIR="$(CURL_PATH)/inst" \
+	install ;
+endef
+
+define Build/Compile
+	$(call Build/Compile/Curl)
+	$(call Build/Install/Curl)
+	$(call Build/Compile/Default)
+endef
+
+define Package/upodman/install
+	$(INSTALL_DIR) $(1)/usr/bin $(1)/etc/init.d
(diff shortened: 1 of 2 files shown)


utils/upodman/Makefile
```diff
@@ -0,0 +1,106 @@
...
+		--without-zsh-functions-dir --without-fish-functions-dir
+
+define Build/Configure
+endef
+
+define Build/Prepare
+	$(call Build/Prepare/Default)
+	mkdir -p $(CURL_PATH)
+	$(TAR) xzf $(PKG_BUILD_DIR)/dl/curl-$(CURL_VERSION).tar.gz -C $(PKG_BUILD_DIR)
+endef
```
BKPepe, line 69: I don't like it. This makes hard dependency on curl.

utils/upodman/Makefile
```diff
@@ -0,0 +1,106 @@
...
+  socket operations, because system's default libcurl might
+  not support unix sockets.
+endef
+
+MAKE_VARS += \
+	BUILD_DIR="$(PKG_BUILD_DIR)"
+
+CURL_VERSION:=7.87.0
+CURL_PATH:=$(PKG_BUILD_DIR)/curl-$(CURL_VERSION)
+CONFIGURE_PATH:=curl-$(CURL_VERSION)
```
BKPepe, line 41: No, I doubt you are going to bump upodman when new curl version is released

utils/upodman/Makefile
```diff
@@ -0,0 +1,106 @@
...
+		$(MAKE_FLAGS) \
+	DESTDIR="$(CURL_PATH)/inst" \
+	install ;
+endef
+
+define Build/Compile
+	$(call Build/Compile/Curl)
+	$(call Build/Install/Curl)
+	$(call Build/Compile/Default)
+endef
```
BKPepe, line 98: No, no no, just no. Most likely, you are looking for HostBuild of curl. Why are you using curl at all?

BKPepe (changes requested):

utils/upodman/Makefile
oskarirauta: It does. Default curl (and self-built, if opted out), does not include support for unix sockets. It's required. I first had my own socket handling, but it broke on Podman 3 -> podman 4. Curl is included in source as a "internal" curl. It's linked statically.

utils/upodman/Makefile
oskarirauta: If I see it fit, I will update upodman when curl is updated. Rarely updates have anything to do with unix-sockets as it's unlikely to change, and required http capabilities are quite basic.

utils/upodman/Makefile
oskarirauta: No no no and just no. Host curl? Why on earth? You've got it all wrong. Original makefile builds curl, but unfortunately all flags don't seem to carry-on to makefile that is included with include directive, that's why I added built-in curl building to package Makefile. I need target built curl library, either dynamic or static, but it must have support for unix sockets.

oskarirauta:
@BKPepe read the source....

utils/upodman/Makefile
oskarirauta: changing CURL_VERSION here does only effect on failed build, as source provided curl source is not found if version is changed. So it must go hand to hand with source of software. Version/extraction would not be relevant at all if curl would ship with software as already extracted, but that would just grow released source file enormously. If it would be under renamed file as curl.tar.gz for example and it would had been re-packed to unextract to curl instead of curl-7.87.0 directory, this would make things a lot different, but it would be more difficult to update the program and I see no point in re-packing curl, just to make package Makefile please someone's eye. Software was originally meant to be built as native only due to it targeting personal use, but on request I'm making it publicly available without requirement to build it natively.

utils/upodman/Makefile
oskarirauta: Look further; that isn't openwrt dl dir, it's a directory under software's source code that just happens to be named as dl.

oskarirauta:
@neheb
Please, I'm requesting your opinion on comments above.

utils/upodman/Makefile
BKPepe: I see. I still don't like it. I don't know, why there should be two versions of curl and also you are compiling it twice. Normal curl and then yours curl. So you are not only adding your's forked package, which isnt in any GNU/Linux distribution, but also we are slowing down build system by compiling embedded curl in your package. :( It also increases the build resources (capacity, etc.).

utils/upodman/Makefile
BKPepe: Yup, I got it wrong, but that's caused that I don't know your package and it was not properly explained, why you put curl in your's upodman repository. 

If I understand it correctly, you want to add dependency to your's upodman package on [LIBCURL_UNIX_SOCKETS](https://github.com/openwrt/packages/blob/a79df49a5fb7fed5a525c72824c966261cbef62e/net/curl/Config.in#L140) instead of just trying to hack it in very bad way.

BKPepe: Also, you need to take it from the security point of view. We can not trust your sources that you put some curl tarball in your repository if you don't verify the checksum, it could be corrupted, your account could be hijacked, the tarball can be changed and we are screw up. Don't we?

oskarirauta: Many software included "built-in" software, it's as likely as any software being hijacked or whatever, why would someone instead attack included curl source, when they can put their code in very much simpler set of mine? It's verified **in** sha256sum of package as it's pre-downloaded, so that more or less breaks that assumption; go ahead, download source code and then download identical version from curl's site and verify they are the same, then you know that sha256sum of release covers that, if you need verification.

If I put dependency on LIBCURL_UNIX_SOCKETS which is not enabled by default, how can this package ever be available publicly with opkg, as I was asked to deliver it as? I ain't gonna start compiling it for anything else except x86-64...

utils/upodman/Makefile
oskarirauta: Because provided one doesn't offer a feature needed? My curl is no way forked, it is as official as any, downloaded and stored as-is directly from curl's site. There's something wrong in what you are saying, this builds in way less than minute on decent hardware, very much identical time to so many other software. Like I previously said, read the source first, check curl's tarball and verify before assumptions.

oskarirauta:
@1715173329
Any knowledge when new kernel's baked into ci?

BKPepe:
> @BKPepe read the source....

I am so sorry to tell you, but you can not ask anyone to review your source code. I wouldn't do anything else, man. There should be documentation, proper readme, etc. I am not going to review the source code and because of your reasoning this is NACK from me.

utils/upodman/Makefile
BKPepe: > go ahead, download source code and then download identical version from curl's site and verify they are the same, then you know that sha256sum of release covers that, if you need verification.

C'mon, imagine my situation, when I clone all the source code from OpenWrt, compile all the packages and I should verify manually each tarball, which was donwloaded as it is build-in? You can not be serious about that.  That's what the build system should do and it does, but that's not in your case. We are trusting you and anything might happen.

> If I put dependency on LIBCURL_UNIX_SOCKETS which is not enabled by default, how can this package ever be available publicly with opkg, as I was asked to deliver it as?

How it is done for other packages? 🤔 Did you check any other packages before arguing about that?

utils/upodman/Makefile
BKPepe: >  it is as official as any, downloaded and stored as-is directly from curl's site. 

How anyone can know that? You did not provide any hash in your repository. 

> read the source first,

Hm, thank you for your suggestion, but no. If you are going to argument like that, then we are not going anywhere else.

1715173329:
excuse me? what happend? please be friendly, always

utils/upodman/Makefile
oskarirauta: It's done the same way, **except** source code is pre-extracted. This way release tar ball takes less space, and it's also possible to verify the curl source. Though, I only designed it in the first place so that it would be a smaller download.

oskarirauta:
@1715173329 
> excuse me? what happened?

Some checks failed, due to kernel version mismatch, and I was looking forward to see if there's a problem.

> please be friendly, always

I try to. It's just a bit hard to reason with someone who in the first place decided it's a no go without further investigation and excepts me to re-invent the wheel since it doesn't look like something he likes. I'll propably close this anyway, it was in the first place supposed to be for personal need anyway; doing favors, might next mean feature requests as well and other extra work not necessary at all for my own need.

1715173329:
> Some checks failed, due to kernel version mismatch, and I was looking forward to see if there's a problem.

Well, it's a known issue of CI, iirc you have asked it many times, just ignore those errors.

oskarirauta:
> > Some checks failed, due to kernel version mismatch, and I was looking forward to see if there's a problem.
> 
> Well, it's a known issue of CI, iirc you have asked it many times, just ignore those errors.

I know- sorry. Just wondered if it's known when it works next time.

utils/upodman/Makefile
oskarirauta: Would it ease your mind if I'd include a curl-7.87.0.tar.gz.md5 or curl-7.87.0.tar.gz.sha256 file to the release, which would contain hash of file then?

oskarirauta:
Closed by de-approval.

1715173329:
> I know- sorry. Just wondered if it's known when it works next time.

Generally speaking, a few hours or a few days

utils/upodman/Makefile
diizzyy: Personally I think we're at a point where "we" (OpenWrt) needs to decide whether to micromanage package and hunt kbytes or just try to adapt sane defaults. I'm fully aware that old MIPS devices still are "supported" but at some point we need to move on rather than holding things back with little to no benefit overall.

diizzyy:
Submit a PR which changes curl's default settings, that's how it works in other package repos.

oskarirauta:
@diizzyy 

> Personally I think we're at a point where "we" (OpenWrt) needs to decide whether to micromanage package and hunt kbytes or just try to adapt sane defaults. I'm fully aware that old MIPS devices still are "supported" but at some point we need to move on rather than holding things back with little to no benefit overall.

I hope you are speaking more in general instead of this case, not so sure if one or two people's beneficial over **this** package is worth the trouble of being merged.

> Submit a PR which changes curl's default settings, that's how it works in other package repos.

Ok; one more time. Unfortunately I made necessary changes to upodman to link again with system's libcurl without being aware of PR's breaking on those occasions if they are closed, so I cannot re-open this anymore; so if I try again, I am breaking rules/recommendations since I would have to re-open this on a new PR, but in general, I think libcurl should as default, to support unix sockets as well; from time to time, I have wanted to just check the structure of json outputted by podman's REST API, and in my case, as I have enabled *all the bells and whistles* of curl, this is not a problem for me, but I can see frustration of anyone else who didn't check that box- overhead on ticking unix sockets on, is quite minimal; so even if this wouldn't be merged, I can also recommend enabling support for unix sockets; it can be useful over other things as well as just with podman's api; though only thing I can say for sure at the moment, would be docker's equillavent socket..

PR #20593, opened 2023-03-05
sing-box: update to 1.1.6

brvphoenix:
Maintainer: me
Compile tested: (x86_64, OpenWrt snapshot)
Run tested: (x86_64, OpenWrt 22.03, tests done)

Description:

1. Add `kmod-inet-diag` as a dependency (https://github.com/openwrt/packages/pull/20486#issuecomment-1430610086)
2. Remove redundant `default n` (https://github.com/openwrt/packages/pull/20486#discussion_r1113008854)

### DIFF
--- a/net/sing-box/Makefile
+++ b/net/sing-box/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=sing-box
-PKG_VERSION:=1.1.5
+PKG_VERSION:=1.1.6
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


BKPepe (changes requested):
I am afraid you can not hide mulitple changes under one commit, which says update, when you are doing more changes than that and in the commit description, it is not said.

My proposal to have it merged:

1. Bump version as a separate commit; changelog would be nice to have in the commit description, but that's not required, though.
2. Adding dependency as separate commit, but the commit description is required
3. Removing redundant ``default n`` could be in the 2nd commit, but to have a better commit history, I suggest keeping it as a separate commit and also, having there reason, why you are doing such changes. 

Nobody is going through the old closed pull request, where there are so many comments, to find the reason why you are doing such changes. You want to use ``git blame`` / ``git show`` or any other command to know why the author did such changes. You will not remember it after a week or month.

BKPepe (approved):
Thanks! Looks much better. 👏

PR #20600, opened 2023-03-06
curl: enable unix sockets by default

oskarirauta:
- changed Config.in to enable unix sockets support by default
 - release number bumped

Maintainer: Stan Grishin / @stangri 

Description:
socket support is very handy when communicating with
various REST APIs.

Size increases are very small, nearly unnoticiable.

Tested-by: Stan Grishin <stangri@melmac.ca>

### DIFF
--- a/net/curl/Config.in
+++ b/net/curl/Config.in
@@ -139,7 +139,7 @@ config LIBCURL_ZSTD
 
 config LIBCURL_UNIX_SOCKETS
 	bool "Enable unix domain socket support"
-	default n
+	default y
 	help
 		Enable HTTP over unix domain sockets.
 		To use this with the curl command line, you specify the socket path to the new --unix-domain option.
--- a/net/curl/Makefile
+++ b/net/curl/Makefile
@@ -10,7 +10,7 @@ include $(INCLUDE_DIR)/nls.mk
 
 PKG_NAME:=curl
 PKG_VERSION:=7.88.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://github.com/curl/curl/releases/download/curl-$(subst .,_,$(PKG_VERSION))/ \


BKPepe:
From my point of view, it would be good to add:
- What is the size difference when you are enabling unix sockets support
 - and why are you making such change in the commit description
 
 I don't want anything new, but just follow some practices, which we have here based on https://openwrt.org/submitting-patches

oskarirauta:
That would depend on arch you are using, but it is nominal. Reason is linked to description.

champtar:
@oskarirauta in the commit description (not github description) you've explained what you are doing, not why you are doing it. It doesn't need to be long, just a simple sentence "this is required by new package upodman" can be enough

stangri:
@oskarirauta could you please provide a binary and IPK size difference for the platform you've tested this on?

oskarirauta:
@champtar

Personally, I think this is more useful feature than whole upodman. Like I explained it in a longish explanation. There's a great tool available in openwrt, called sockread; that you can use to communicate with sockets, but if you have a protocol answering, in this case, when it's http; you could succeed way better with tool such as curl that already faciliates this feature; making a POST request is way easier when you just -X post... When on the other hand, sweat pasting pre-written text to interact with socket properly.

But for sure; easier if I just say that it's for upodman... But that would be a lie, I've been using curl sockets for years. I propably forgot to mention, that curl sockets support also expands to curl client program; so you don't have to write your own software to use it. Though, often you propably would need it for that mostly..

@stangri 

unfortunately I have to decline. I already abandoned this project once, it doesn't benefit myself at all; now I was re-asked to work on this again, and there's just so much that I can give. My build runs multiple times per day (number varies from changes that I made locally, and from changes that are made on mainstream, for all the packages and buildroot part) for a full-set and full feature, and is on a production use; disabling curl sockets, or any of **all** features, that I have enabled on my build would reduce functionality on multiple packages resulting to a huge disaster. You realize, that when feature X disappears from one package, another package might have to loose features A, B and C, and the list would go end-lessly.
Finally resulting in possibility of multiple package builds failing and possibly days of work. If this is a key question, I have to apologize and leave this as-is, as I don't have resources for that at the moment.

But as a bonus, that doesn't proof anything; but provides interesting twist..
I can point you to a c source code of similar unix socket code, [here](https://github.com/denehs/unix-domain-socket-example/blob/master/client.c) - it's not from curl project per se, but it uses same core functions as it does (would had pointed you to curl source, but it's a bit of mess, and if you want that; I can do it- some day) - main point here being that code for unix sockets in curl is very likely to be less than 100 lines of code; whilst using same headers that are already used with tcp/ip connection; so... It's nominal.

champtar:
@oskarirauta I just followed your link and saw upodman, if you have a longer explanation on why this is needed no problem ;)

oskarirauta:
@champtar

Thank you for patience. That had to be difficult. I could talk of benefits for hours, texts as long as the bible. And unfortunately I most likely will, sorry. I've been here for ~20 years and introduced myself to openwrt when the first wrt54g was marketed, I still got numerous of them at the attic, so is my nslu2, where I finally got a native toolchain and chance to get web camera support, and wrote my own web server that published picture from that camera on a hosted page that refreshed it's self eternally. Don't know, that might still be in the archives of old openwrt forums. What a joy it was. Camera ofcourse was facing towards nslu2 as it was proof of concept how small embedded devices can be so much more that comes out of the box. We.. I've gone a long way from that. Unfortunately one of good things that happened during that journey, is that I've grown my knowledge, and bad thing with that, is that I am happy to share; and older I get, more rambling I seem to get, mother of my children constantly reminds me about that. Sometimes multiple times on a same day.

Understanding benefits of unix sockets support in curl is difficult to explain to anyone who has never seen use for it. And then are those who need it. Unix sockets are propably most often used as paths to communicate with on-going software; for example ubus uses it. There are ready made protocols, such as http here, that I offered as example; all rest api's use it. You make a http request using get, to retrieve *data*, and usually post, when you request for an *action* of sorts.

Curl is for the moment, and very likely, for a long time in the future as well, the best library available for many features that it has. To feed my own curiosity, I made a quick search on the internet, and then solely on github, for different libraries for that provide *any kind* of socket support to find out, that based on the short search, curl is actually only available library properly supporting for example, http protocol over sockets; there was **one** other library (forgot the name) that also supported http over sockets, but could not be used for REST API of either podman or docker, at least not fully, as logs and stats come in *chunked* form - which is feature that libcurl has, but all others that I found, lack - including that only promising alternative.

If you do `make menuconfig` and go to libcurl to check available options, you find a handful of available protocols (and at near the bottom, you also find curl sockets..) - these *all besides ftp* are supported over unix sockets by libcurl. FTP isn't for technical reason of it needing a two-way stream... Actually if you study further, you find out that this list of available features, is actually quite short when you compare it to really full availability of options; I am not saying here that, we should enable all features, I am just telling a bit more about curl;

Okay, now that we all know that curl is **all mighty and above**.. If you come up to think about a new project, for example, project similar to upodman - that uses REST APIs, or what ever they want to call something that makes hand-shake in http; and you need to communicate over unix socket- you could.. for example, visit stackoverflow.. People there are very helpful (even I, those very rare times that I visit there, but mostly I'm there only with swift language tagged posts..) and very often solution is found; but when you ask assistance over such thing as communicating with http over a socket, it's most likely recommended in every thread, that *just use curl*.. Because there is no point in inventing the wheel again, as it's already done and we have a superior library to handle that.

Breaking my promise again; upodman, used to have it's own routines that helped me to avoid usage of libcurl completely, I previously left a link to curl communicating example, it's not that difficult... Right? Or wrong, because there are million of things that can go wrong. Decision if communication with socket takes too long (timeout), different ways to handle errors, and especially support for that chunked encoding, I had it- but something happened between update of podman 3 -> 4, and my code stopped working. For that reason, I went to libcurl. No doubt, that I couldn't had fixed my own socket/http code, but since there already exists a perfectly working library that very likely covers million of those things that I didn't cover (which came along after upgrading) - the recommendation to *just use curl* starts suddenly sound very wise choice, even for self-made man like myself.

I don't know. Here, as trying to prove benefit over enabling such feature as default, sounds like it's more important than water- I start to doubt myself, that maybe it's then just me who thinks it's useful. Like I said, I've used it for years, for so many things it can be useful, for example, you can use it with nginx, and it's possible that there is a long line of programs that could be used with socket interactions, but I don't have such list, and won't be gathering such list also;

Maybe I am then only one using sockets? Would it really hurt enabling unix sockets support? Everyone can bake their own openwrt and enable it. Though I do remeber one time when I forgot once that it didn't come with *stock image*- ugh.. What a nightmare..

Can't vouch it more than this. If it's truly so, that no one finds it useful besides me, is it worth then enabling it? I didn't go through issues to see if there's complaints about missing sockets support, but I am pretty sure there are none. And if I need to proove a why it's useful feature, I don't know- how would any of this infotainment make it more popular that it would be needed.

I promised myself that I'll keep this short and on this point, I see that failed my intention poorly. I maybe just should had said this:
https://docs.podman.io/en/latest/_static/api.html

Check the first paragraph for details on how to interact with rest-api, it instructs you to use curl, you can use either curl or libcurl from your own software; doesn't matter- it just is the best tool for that. *Almost* identical api is available for docker as well. REST(ful) API outputs JSON, unlike it's cousin SOAP that outputs XML, REST API is considered somewhat better, marketing? Who knows. REST API is used with atleast those two containerization systems and if I remember correctly, atleast one home automation system had REST API support, home assistant; one of those most biggest ones that you can host on your own (I liked it, unfortunately I went for tuya, and there I will go to Matter when time is right..)

So, enabling unix sockets on libcurl (this enables it then on the curl executable as well) - can and **is** useful.

Thanks for reader's patience while reading this. I hope you found some enjoyment on it.

Btw. Etienne.. I am feeling more and more confident, that I shouldn't be both maintainer and creator, of upodman; so more or less, if this is all about enabling this feature just for that, considering this might not be worth a while, unless you want to be maintainer of it by yourself, then give you all the kudos, but I ain't going that road again- I already am head over heels while trying to maintain podman and all it's parts, especially with new netavark.

oskarirauta:
One more thing.. I checked curl's code for unix sockets (quick oversight) - 
when unix sockets are enabled, USE_UNIX_SOCKETS is defined and by that definition, code is included into build.
Most of this code exists in curl's source in lib/url.c - I made a short calculation of rows of code that are appended with USE_UNIX_SOCKETS enabled. It was somewhere around 75 lines. [url.c](https://github.com/curl/curl/blob/master/lib/url.c)
http_proxy.c had 4 lines. And curl_addrinfo.c had 27 lines. I skipped empty lines and comments, but did calculate lines that are splitted to 2 or more lines, as invidual lines of code. Also setopt.c had some lines, but I didn't count those. Those lines are in the command-line parser and --help section. So, library grows about 100 lines of code, in C - using headers that it would use, whether unix sockets is enabled or not, so library file's size... I wouldn't dare to say that this would grow it too much.

@stangri
my libcurl.so.4.8.0 file's reported size is 590K. It has all features listed in Config.h enabled, and some that are left out from the list, with nls, and no stripping used at all. My daily re-built bin directory, containing packages and image, is 3.3gb and my base image, is 96mb in rootfs.tar.gz - I don't do images at all. Buildroot is 137.9gb and built packages count is 6117. It has basic repositories, and some of my own.

I've got no idea, how much it would shrink if I'd disable unix_sockets, but even if I would, difference is difficult to put in proportions, as my build is not stripped and is optimized always for speed, not for size.

stangri:
Since @oskarirauta couldn't find the time to test how much merging their PR affects the binaries size, I've done some tests on x86_64 and turning on UnixSockets increased the  file size of `libcurl4_7.88.1-1_x86_64.ipk` from 172061 bytes to 172638 bytes with no effect on the installed library size (361827 bytes).

Pending satisfaction of requests from other reviewers and proper filling of the PR template, this is an ACK from me.

oskarirauta:
@stangri

Thanks, it's though not due to time, it's more that I'd had needed to setup another host to build on, as my current building host is busy doing it's current task. But I appreciate your effort. All 8(4 really, threads just double it) cores on that router already run with a red glow constantly in +85c without doing another concurrent build.

Interesting stats though, I though that with reduced set of features and with stripping extracted library size would had been just 172K. I would guess from size that you built it for x86-64.

Spudz76:
Without patch:
```
 59313 bin/packages/mipsel_24kc/packages/curl_7.88.1-1_mipsel_24kc.ipk
163035 bin/packages/mipsel_24kc/packages/libcurl4_7.88.1-1_mipsel_24kc.ipk
132243 ./usr/bin/curl
336487 ./usr/lib/libcurl.so.4.8.0
```
With patch:
```
 59316 bin/packages/mipsel_24kc/packages/curl_7.88.1-2_mipsel_24kc.ipk
163040 bin/packages/mipsel_24kc/packages/libcurl4_7.88.1-2_mipsel_24kc.ipk
132243 ./usr/bin/curl
336487 ./usr/lib/libcurl.so.4.8.0
```

So it adds 3 bytes to the `curl` ipk and 5 bytes to the `libcurl` ipk, and no change at all to the actual executable sizes.

oskarirauta:
Nominal increase in size in that case seems to be understatement then. Quite large when considered as a build for mipsel, I thought it would had been smaller. Thanks for your effort as well.

oskarirauta:
Should we then enable it as default?

BKPepe (changes requested):
If you want to have this merged, you should:

1. Add a commit description to the commit itself.
 Reasoning:
    - By using ``git log``/``git blame``, you can see who is responsible for enabling unix domain socket support, but you don't see the reason for this change. That's why you need to add it, so your contribution follows [OpenWrt guidelines](https://openwrt.org/submitting-patches).
     - It is time-consuming to go to the pull request, if it exists, in this case, it exists, but nobody is going to check the reasoning if your's response has more than 1300 words.
   
  2. Take the size, what is different and what was done by others, request ``Tested-by`` (opt-in) tag and add it both to your commit description. 
  Reason:
      - OpenWrt still supports min. 8 MB as flash, min. 32MB as RAM.

oskarirauta:
champtar requested for longer explanation.
It was time consuming to write it as well.

BKPepe (approved):

mhei (approved):
LGTM

oskarirauta:
@BKPepe @mhei 
Thank you

PR #20608, opened 2023-03-07
tiff: force libdeflate support to off

micmac1:
Commit 81d2b72 added a package providing libdeflate. Tiff by default links to it, causing a build error.

Package libtiff is missing dependencies for the following libraries: libdeflate.so.0

This commit forces libdeflate use off to avoid this. No revision bump is done because the package is currently not compiling anyway.

Maintainer: @jslachta 

Compile tested: ath79 master
Run tested: N/A

Description: fix current packaging issue


jslachta (approved):

jslachta:
Thank you, Sebastian. Let's wait until CI passes.

PR #20609, opened 2023-03-07
libgpiod: update to 2.0 and introduce C++ binding

mhei:
Maintainer: me
Compile tested: mxs
Run tested: mxs, Duckbill

Description:

This updates this library to the latest major version.

For the Python binding switch to the Py3Package infrastructure.

And while at, also introduce the C++ binding, too.

### DIFF
--- a/libs/libgpiod/Makefile
+++ b/libs/libgpiod/Makefile
@@ -26,6 +26,12 @@ PYTHON3_PKG_BUILD:=0
 include $(INCLUDE_DIR)/package.mk
 include ../../lang/python/python3-package.mk
 
+ifneq ($(CONFIG_PACKAGE_libgpiodcxx),)
+CONFIGURE_ARGS += --enable-bindings-cxx
+else
+CONFIGURE_ARGS += --disable-bindings-cxx
+endif
+
 ifneq ($(CONFIG_PACKAGE_gpiod-tools),)
 CONFIGURE_ARGS += --enable-tools
 endif
@@ -59,6 +65,18 @@ endef
 define Package/gpiod-tools/description
   Tools for interacting with the linux GPIO character device
   (gpiod stands for GPIO device).
+endef
+
+define Package/libgpiodcxx
+  SECTION:=libs
+  CATEGORY:=Libraries
+  URL:=https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git
+  TITLE:=C++ binding for libgpiod
+  DEPENDS:=+libstdcpp +libgpiod
+endef
+
+define Package/libgpiodcxx/description
+  This package contains the C++ binding for libgpiod.
 endef
 
 define Package/python3-gpiod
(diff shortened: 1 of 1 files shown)


BKPepe:
Hello! :-)

I think, it would be good to split your changes into several commits, well at least three.
- update
- switch to Py3Package
- introduce the C++ binding

This allows us to have a better Git history, but not only history, but also it makes it easier to understand and know why you made such changes and if anything happens, we can just simply revert what is broken instead of reverting all the changes.

mhei:
I could split into 2 commits:
1. the update to 2.0 requires the switch to Py3Package - the older build stuff will just not work with the new version anymore - in other words, only the update would leave the package in a broken state
2. introduce the C++ binding

BKPepe:
Sounds good to me, thanks for your explanation and time!

mhei:
Thanks for your review, very appreciated. :smile:
I've force-pushed the branch. Maybe you could have a second look. I think a PKG_RELEASE bump is not strictly required for the second commit since there is not ABI change for the existing packages.

mhei:
I missed that the library update breaks openocd, need to sort this out first.

karlp:
I'd like to see this go forward, or "someone" work out what's wrong the 1.6.x branch.

v2 works properly, giving consistent timestamps with no rollover, but the current v1.6.x branch seems completely broken as far as timestamps go.  If I use the raw chardev myself, with the v1 api, I correctly get a full 64bit nanoseconds timestamp, which never rolls over.  If I use "gpiomon" or the python bindings from 1.6, I get this sort of thing... for a line that's giving me a rising edge every 20 ms.

```
# gpiomon -r -F "hoho %o => %s . %n" gpiochip1 11
... snip ...
hoho 11 => -1093714298 . 940445038
hoho 11 => -1093714298 . 960476916
hoho 11 => -1093714298 . 980382383
hoho 11 => -1093714298 . 452927
hoho 11 => -1093714298 . 20391309
hoho 11 => -1093714298 . 40427270
hoho 11 => -1093714298 . 60395818
hoho 11 => -1093714298 . 80364032
hoho 11 => -1093714298 . 100337788
hoho 11 => -1093714298 . 120343292
```

Without the custom formatting, it's even worse.  The lower digit rolls over here, slowly, but something's very wrong.
```
# gpiomon -r gpiochip1 11
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
event:  RISING EDGE offset: 11 timestamp: [   71001.000594410]
```

using the python bindings are ~equivalent.

The same thing with v2 merged (I cherry-picked this PR) works as expected, though the python api has changed.  (because of course...)

I'm going to talk to the openocd people about what they want to do, but the incompatible v1 and v2 chardevs are _highly_ unfriendly to downstreams :|

mhei:
I had a really fast look only at the openocd code and my guess is that it should not be that hard to adapt. However, the big question for me was, how the openocd people will handle the two different chardevs apis internally. For here, in "downstream", a simple patch which corrects the API usage would do, but for upstream, I'm unsure.
Sadly, I did not found the time to create such a patch and I need to think about a test setup...

Djfe:
Just stumbled across all of this because gpio sysfs isn't working for me and I need to trigger GPIOs manually 😅
I'm going to be building my device with [CONFIG_GPIO_CDEV_V1](https://zameermanji.com/blog/2022/1/18/ubuntu-21-10-makes-libgpiod-unusable/) now, but it would be very nice to see this solved eventually. Maybe with the release of OpenWrt 23 (one can dream)

Upstream issue of OpenOCD:
https://sourceforge.net/p/openocd/tickets/306/

Kind of offtopic:
I just added an info box to the wiki which is kind of outdated again by now, since v2 is right around the corner. Hopefully this keeps other from wasting their time with gpio-sysfs
https://openwrt.org/docs/techref/hardware/port.gpio#software
The wiki page definitely needs to be overhauled by someone more knowledgeable than me.

Issue #20596, opened 2023-03-05
openblas: compiling issue if using ccache


trippleflux:
Maintainer: @commodo
Environment: x86_64 , snapshot

Description:

~~Looks like the recent OpenWrt 5a1eb3bb2c27ede3046bb2977008c74a04681db9 commit causing this openblas compiling issues~~ : 
```
Y=x86_64-openwrt-linux-gnu-objcopy OBJDUMP=x86_64-openwrt-linux-gnu-objdump SIZE=x86_64-openwrt-linux-gnu-size CROSS="x86_64-openwrt-linux-gnu-" ARCH="x86_64" CROSS=1 HOSTCC=gcc CROSS_SUFFIX=x86_64-openwrt-linux-gnu- BINARY=64 MAKE_NB_JOBS=-1 NUM_THREADS=2 PREFIX=/usr COMMON_OPT="" TARGET=GENERIC NOFORTRAN=1 ;
make[3]: Entering directory '/home/username/works/openwrt/build_dir/target-x86_64_glibc_custom/OpenBLAS-0.3.21'
ccache: invalid option -- 'm'
C Compiler (ccache -m64) is something wrong.
ccache: invalid option -- 'm'
C Compiler (ccache -m64) is something wrong.
/usr/bin/ld: warning: /home/username/works/openwrt/tmp/ccDwe54L.o: missing .note.GNU-stack section implies executable stack
/usr/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
ccache: invalid option -- 'm'
C Compiler (ccache -m64) is something wrong.
make[3]: *** [Makefile.prebuild:63: config.h] Error 1
Makefile.system:284: Makefile.conf: No such file or directory
make[3]: *** No rule to make target 'Makefile.conf'.  Stop.
make[3]: Leaving directory '/home/username/works/openwrt/build_dir/target-x86_64_glibc_custom/OpenBLAS-0.3.21'
make[2]: *** [Makefile:108: /home/username/works/openwrt/build_dir/target-x86_64_glibc_custom/OpenBLAS-0.3.21/.built] Error 2
make[2]: Leaving directory '/home/username/works/openwrt/feeds/packages/libs/openblas'
time: package/feeds/packages/openblas/compile#0.72#1.06#1.67
    ERROR: package/feeds/packages/openblas failed to build.
make[1]: *** [package/Makefile:116: package/feeds/packages/openblas/compile] Error 1
[...]

PR #20624, opened 2023-03-09
openblas: don't allow ccache use

commodo:
Maintainer: me
Run tested: n/a

-----------------------------

Fixes https://github.com/openwrt/packages/issues/20596


trippleflux:
Compiled fine for x86_64

PR #20627, opened 2023-03-10
miniupnpd: bump to 2.3.3

stintel:
Maintainer: NA
Compile tested: NA
Run tested: NA

Untested as I don't run 22.03. CI should do compile tests, and if someone who does run 22.03 would be so kind to test this and report back, that would be appreciated.

### DIFF
--- a/net/miniupnpd/Makefile
+++ b/net/miniupnpd/Makefile
@@ -8,15 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=miniupnpd
-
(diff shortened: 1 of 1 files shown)


Strasedon:
Will test, I had already been running 2.3.2 for a bit over a month with no problems since I was seeing the nft_send_rule and send_batch errors on 22.03.

FWIW it compiled and started up fine, testing with OpenWrt 22.03-SNAPSHOT r19160+883-7ea2f3d6e2 on an Archer C7v2.

PR #20647, opened 2023-03-11
shadowsocks-libev: added support for ACL

fertkir:
Added ability to pass an Access Control List (--acl param) to forbid access to specified IP addresses

Maintainer: @yousong 
Compile tested: didn't compile, since this is only an init-file change
Run tested: MediaTek MT7628AN ver:1 eco:2, TP-Link TL-WR841N v13, OpenWrt 22.03.3 r20028-43d71ad93e

### DIFF
--- a/net/shadowsocks-libev/files/shadowsocks-libev.init
+++ b/net/shadowsocks-libev/files/shadowsocks-libev.init
@@ -81,6 +81,7 @@ ss_xxx() {
 	[ -z "$mtu" ] || json_add_int mtu "$mtu"
 	[ -z "$timeout" ] || json_add_int timeout "$timeout"
 	[ -z "$user" ] || json_add_string user "$user"
+	[ -z "$acl" ] || json_add_string acl "$acl"
 	json_dump -i >"$confjson"
 
 	procd_open_instance "$cfgtype.$cfg"
@@ -273,7 +274,8 @@ validate_server_section() {
 }
 
 validate_ss_local_section() {
-	validate_common_client_options_ ss_local "$1" "$2"
+	validate_common_client_options_ ss_local "$1" "$2" \
+	  'acl:file'
 }
 
 validate_ss_redir_section() {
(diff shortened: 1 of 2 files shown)


fertkir:
@yousong, please review

net/shadowsocks-libev/files/shadowsocks-libev.init
```diff
@@ -307,7 +308,8 @@ validate_ss_server_section() {
 		'local_address:ipaddr' \
 		'local_ipv4_address:ip4addr' \
 		'local_ipv6_address:ip6addr' \
-		'bind_address:ipaddr'
+		'bind_address:ipaddr' \
+		'acl:string'
```
yousong, line 312: Is the value supposed to be a path to the acl file?  Consider using something like `dst_ips_bypass_file:file`

yousong: My guess is that the option should be set at the client side.  That is `validate_common_client_options_`, right?

yousong:
Please also increment PKG_RELEASE variable for the package Makefile.

If you are also a user of luci-app-shadowsocks-libev, please consider adding support for this newly-added field there. 
 Maybe following the example of [`dst_ips_bypass_file `](https://github.com/openwrt/luci/blob/b683ff3ea2bbd49a38b12bab4225440ba3de5ff5/applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/rules.js#L95-L102), but for [client instances](https://github.com/openwrt/luci/blob/b683ff3ea2bbd49a38b12bab4225440ba3de5ff5/applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/instances.js#L102-L109).  This is optional and will not block merging of this pull request here.

net/shadowsocks-libev/files/shadowsocks-libev.init
fertkir: > Is the value supposed to be a path to the acl file? Consider using something like `dst_ips_bypass_file:file`

Changed type from `string` to `file`

> My guess is that the option should be set at the client side. That is `validate_common_client_options_`, right?

[According to docs](https://github.com/shadowsocks/shadowsocks-libev), ACL is only available in local and server mode. At least I use it for `ssserver`. So added to `validate_ss_server_section` and `validate_ss_local_section`.

fertkir:
PKG_RELEASE release incremented.
Pull-request for Luci will be ready soon.

Should I squash commits?

yousong:
Hi @fertkir , The change was just applied after an indention fix.  Thank you!

PR #20652, opened 2023-03-12
zerotier: do not allow executable stack

oskarirauta:
zerotier has executable stack.
`
[   11.343143] process '/usr/bin/zerotier-one' started with executable stack
`

executable stacks are not recommend, possibly provide a threat and there seems to be no real advantage of executable stack with zerotier - so let's build it without instead.

Some chatter about [executable stacks](https://security.stackexchange.com/questions/210785/how-and-why-is-executable-stack-dangerous) and a older feature request on [same subject](https://github.com/zerotier/ZeroTierOne/issues/1179) at zerotier's repository for reference.

Maintainer: Moritz Warning / @mwarning

### DIFF
--- a/net/zerotier/Makefile
+++ b/net/zerotier/Makefile
@@ -58,8 +58,8 @@ endif
 endef
 
 # Make binary smaller
-TARGET_CFLAGS += -ffunction-sections -fdata-sections
-TARGET_LDFLAGS += -Wl,--gc-sections,--as-needed
+TARGET_CFLAGS += -ffunction-sections -fdata-sections -Wl,-z,noexecstack
+TARGET_LDFLAGS += -Wl,--gc-sections,--as-needed -Wl,-z,noexecstack
 
 define Package/zerotier/conffiles
 /etc/config/zerotier


mwarning:
The stack seems to be non-executable on ramips already:
```
root@OpenWrt:~# ps  | grep -i zero
 2167 root     12756 S    /usr/bin/zerotier-one /var/lib/zerotier-one_sample_config
 2299 root      1308 S    grep -i zero
root@OpenWrt:~# cat /proc/12756/task/*/maps | grep stack
cat: can't open '/proc/12756/task/*/maps': No such file or directory
root@OpenWrt:~# cat /proc/2167/task/*/maps | grep stack
7f9fd000-7fa21000 rw-p 00000000 00:00 0          [stack]
7f9fd000-7fa21000 rw-p 00000000 00:00 0          [stack]
7f9fd000-7fa21000 rw-p 00000000 00:00 0          [stack]
7f9fd000-7fa21000 rw-p 00000000 00:00 0          [stack]
7f9fd000-7fa21000 rw-p 00000000 00:00 0          [stack]
```
@oskarirauta is it the same for x86_64?

oskarirauta:
@mwarning 

not same. rwxp here..

```
root@dev2:~# logread|grep stack
Sun Mar 12 22:10:53 2023 kern.info kernel: [    0.201618] mem auto-init: stack:off, heap alloc:off, heap free:off
Sun Mar 12 22:10:54 2023 kern.warn kernel: [   11.165031] process '/usr/bin/zerotier-one' started with executable stack
root@dev2:~# ps|grep zero
 2685 root     12992 S    /usr/bin/zerotier-one /var/lib/zerotier-one_rauta_eu
 3819 root      1252 S    grep zero
root@dev2:~# cat /proc/2685/task/*/maps | grep stack
7ffe4341f000-7ffe43459000 rwxp 00000000 00:00 0                          [stack]
7ffe4341f000-7ffe43459000 rwxp 00000000 00:00 0                          [stack]
7ffe4341f000-7ffe43459000 rwxp 00000000 00:00 0                          [stack]
7ffe4341f000-7ffe43459000 rwxp 00000000 00:00 0                          [stack]
7ffe4341f000-7ffe43459000 rwxp 00000000 00:00 0                          [stack]
root@dev2:~# 
```

Should I grep through VSZ as pid as well? Is that something related to executable stack?
Do you have that warning in your syslog?

mwarning:
> Do you have that warning in your syslog?
```
root@OpenWrt:~# logread|grep stack
Tue Jan  3 00:40:56 2023 kern.info kernel: [    0.000000] mem auto-init: stack:off, heap alloc:off, heap free:off
```
There is no warning there for ramips/mt7620.

As for VSZ it probably does not make sense to check it. There is a lot of executable stuff there in shared librarys anyway.

Btw., the errror for powerpc_8540 is not releated to this change at all.

@oskarirauta  The only change I would request is that we should mention that the stack is executable on x86_64 in the commit message.

oskarirauta:
I can add that.. Just a sec..

mwarning:
@oskarirauta let's update to 1.10.4 in a different MR. There is already one.

LGTM

oskarirauta:
@mwarning It seems his PR also covers refreshing patches, I didn't notice on mine that they weren't perfectly align.

1715173329:
Please rebase your PR onto the latest master branch

oskarirauta:
@mwarning rebased.

BKPepe:
> Stack is executable on x86_64, but not on all archs, such as ramips.

I think we might have much bigger problem here and we should look at defaults from OpenWrt build system and tweak it there. Because I dont think that we should do something special for individual each package, which we have here.

oskarirauta:
Leave stack enabled for x86_64 now and wait few years for someone to enable it for all archs?

BKPepe:
No, you should check it how OpenWrt handles compiles for x86_64 and if it should be revisited there. This needs to be done in OpenWrt main repository.

oskarirauta:
@BKPepe It's just that it is by design using executable stacks, so it's working correctly for x86_64 - not for (some) others.. So it shouldn't be visited for x86_64....

PR #20654, opened 2023-03-12
libjpeg-turbo: enable static library building

oskarirauta:
Allow build of libjpeg as a static library as well; one is provided for libpng and possibly for other
formats as well.

Maintainer: Rosen Penev / @neheb

Description:
I have a application that can draw graphical content to frame buffer, including few image formats and
jpg is amongst supported formats. As software is meant for displaying a _boot splash_ and started early
from initrd, static linking is encouraged. But this would require static archive from libjpeg to be possible.

Static archive is not added to package, only to development libraries by this PR.


mhei (approved):
LGTM

PR #20676, opened 2023-03-16
[openwrt-19.07] python3: Update to 3.7.16, refresh patches

jefferyto:
Maintainer: me
Compile tested: ~~relying on CI (buildroot requires Python 2)~~ armvirt-32, 19.07.10 sdk
Run tested: none

Description:
Includes fixes:

* 3.7.14:
  * CVE-2020-10735: Prevent DoS by large int<->str conversions
  * CVE-2021-28861: http.server: Open Redirection if the URL path starts with //

* 3.7.16:
  * CVE-2022-45061: Slow IDNA decoding with large strings
  * CVE-2022-37454: Buffer overflow in the _sha3 module
  * CVE-2015-20107: mailcap.findmatch: document shell command Injection danger in filename parameter


jefferyto:
CI wants to check out feeds from git.openwrt.org and is failing. Let me try to find a way to test compile this locally.

BKPepe:
Since OpenWrt 19.07 is End of Life, I don't know if we should merge it, but since you worked on it, let's do it.

jefferyto:
> Since OpenWrt 19.07 is End of Life, I don't know if we should merge it, but since you worked on it, let's do it.

I'm not sure this will even be built by buildbots, but I think it would be good to have this in case someone wants to manually build the package.

PR #20679, opened 2023-03-16
ddns-scripts: add hotplug trigger for device name

ttxs25830:
Maintainer: me
Compile tested: no
Run tested: Unfortunately not

Description:
Users can register an ddns updater and bind it on an device, not an interface. 
So we must also trigger start/stop event for device name, not just interface name, otherwise this kind of updaters will not able to auto start/stop with hotplug event.


GeorgeSapkin:
No updates in over two years and no run testing. Closing as stale. Feel free to reopen as necessary.

PR #20683, opened 2023-03-17
sscep: add package

johnfzc:
Maintainer: @johnfzc
Compile tested: rampis, glinet_gl-mt300n-v2, 22.03.3
Run tested: rampis, glinet_gl-mt300n-v2, 22.03.3

Description: SSCEP client from certnanny, it was previously packaged in OpenWrt and removed at some point, I've been maintaining it outside of the tree since.

### DIFF
--- a/net/sscep/Makefile
+++ b/net/sscep/Makefile
@@ -1,0 +1,46 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=sscep
+PKG_VERSION:=0.10.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/certnanny/sscep/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=489cc8e093986776eb3f15082bf766778f707176f3cd604bf0ef1008da06b8e5
+
+PKG_MAINTAINER:=John Marrett <johnf@zioncluster.ca>
+PKG_LICENSE:=BSD-3-Clause OpenSSL
+PKG_LICENSE_FILES:=COPYING
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+define Package/sscep
+  SECTION:=net
+  CATEGORY:=Network
+  DEPENDS:=+libopenssl +openssl-util
+  TITLE:=A Simple SCEP (Simple Certificate Enrollment Protocol) client
+  URL:=https://github.com/certnanny/sscep/
+endef
+
+define Package/sscep/description
+  sscep is a simple SCEP client published by the CertNanny project.
+
+  It is used to retrieve CA certificates and enroll devices with a
+  compatible SCEP server.
+endef
+
+define Package/sscep/conffiles
+/etc/sscep.conf
+endef
+
+define Package/sscep/install
+	$(INSTALL_DIR) $(1)/etc
+	$(INSTALL_CONF) $(PKG_BUILD_DIR)/sscep.conf $(1)/etc/
+
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/sscep $(1)/usr/bin/sscep
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/mkrequest $(1)/usr/bin/mkrequest
+endef
+
+$(eval $(call BuildPackage,sscep))
--- a/net/sscep/test.sh
+++ b/net/sscep/test.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+sscep | grep -q "$PKG_VERSION"


net/sscep/Makefile
```diff
@@ -0,0 +1,63 @@
+#
+# Copyright (C) 2006 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
```
BKPepe, line 6: Remove it or assign it to yourself.
Check: https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#makefile-contents-should-contain

net/sscep/Makefile
```diff
@@ -0,0 +1,63 @@
...
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=sscep
+PKG_VERSION:=0.10.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=v${PKG_VERSION}.tar.gz
```
BKPepe, line 14: This isn't good, because in dl folder, the tarball will be downloaded as v0.10.0 and it would be hard to know for which sw it was downloaded. It should be packageName-version.

net/sscep/Makefile
```diff
@@ -0,0 +1,63 @@
...
+PKG_SOURCE_URL:=https://github.com/certnanny/sscep/archive/
+PKG_HASH:=489cc8e093986776eb3f15082bf766778f707176f3cd604bf0ef1008da06b8e5
+
+PKG_LICENSE:=OpenSSL
```
BKPepe, line 18: Based on this https://github.com/certnanny/sscep/blob/master/COPYING, I see BSD-3-Clause, OpenSSL and others. 
Anyway, you should add the LICENSE file.

net/sscep/Makefile
```diff
@@ -0,0 +1,63 @@
...
+endef
+
+define Build/Compile
+	$(MAKE) -C $(PKG_BUILD_DIR) \
+		$(TARGET_CONFIGURE_OPTS) \
+		CFLAGS="-ggdb3 $(TARGET_CFLAGS)" \
+		CPPFLAGS="$(TARGET_CPPFLAGS)" \
+		LDFLAGS="$(TARGET_LDFLAGS) -lssl -lcrypto"
+#		LDFLAGS="$(TARGET_LDFLAGS)"
+#		LDFLAGS="$(TARGET_LDFLAGS) -lssl -lcrypto"
```
BKPepe, line 47: Remove it.

BKPepe (changes requested):
1. I wonder, why it was removed back?
   - Why do we need it to add it back?
 
 2. It would be nice to have ``test.sh`` file, which ensures that this package is run tested by CI.
 See: https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#continuous-integration
 
 3. Package description is missing.
 
 4. I think Prepare,Configure and Compile stages can be improved a lot.

net/sscep/Makefile
```diff
@@ -0,0 +1,63 @@
...
+$(call Build/Prepare/Default)
+	cd $(PKG_BUILD_DIR); ./bootstrap.sh
+endef
+
+define Build/Configure
+	cd $(PKG_BUILD_DIR); ./configure
+endef
+
+define Build/Compile
+	$(MAKE) -C $(PKG_BUILD_DIR) \
```
1715173329, line 41: It's better to switch to use CMake.

johnfzc:
> * I wonder, why it was removed back?

I am not certain, I suspect it's maintainer may have abandoned it.

I can find it here: https://git.openwrt.org/?p=openwrt/svn-archive/packages.git;a=commit;h=b7f060b6ee758612a7a15ae26faf649269362667

And in oldpackages: https://github.com/mhei/openwrt-oldpackages/commit/f8fd1c82f6ebd164272715d6012a5930ca714566

But I can't find a trace of it's removal in the current packages tree.

>   * Why do we need it to add it back?

OpenWrt includes another [SCEP client](https://openwrt.org/packages/pkgdata/strongswan-scepclient), it doesn't work with my SCEP server.

> * It would be nice to have `test.sh` file, which ensures that this package is run tested by CI.
>   See: https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#continuous-integration

I'll add one and address the other concerns and comments if the package will be accepted once they are addressed. Please let me know what you think of my response about the removal and why to add it back first.

Thanks very much for taking the time to review this PR.

net/sscep/Makefile
johnfzc: This name is based on the naming used by the upstream package; here's the release page:

https://github.com/certnanny/sscep/releases/tag/v0.10.0

And the tar ball that's linked from it:

https://github.com/certnanny/sscep/archive/refs/tags/v0.10.0.tar.gz

Is there a way that I can download it to a different name as the naming. while mandatory for the download,  is highly confusing in the dl folder?

net/sscep/Makefile
johnfzc: How can I do this? Can you please point me to an example package?

net/sscep/Makefile
johnfzc: @BKPepe this is the last remaining issue raised in the review, I've taken a look at the [documentation](https://openwrt.org/docs/guide-developer/packages) for the variables in a package Makefile and don't see how I can download the package to a different local name than the remote file name reflected in the `PKG_SOURCE`.

Please let me know if you see a way this can be solved. Thanks again!

net/sscep/Makefile
```diff
@@ -0,0 +1,53 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=sscep
+PKG_VERSION:=0.10.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=v${PKG_VERSION}.tar.gz
+PKG_SOURCE_URL:=https://github.com/certnanny/sscep/archive/
```
1715173329, line 8:
```suggestion
PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
PKG_SOURCE_URL:=https://codeload.github.com/certnanny/sscep/tar.gz/v$(PKG_VERSION)?
```

net/sscep/Makefile
```diff
@@ -0,0 +1,53 @@
...
+PKG_HASH:=489cc8e093986776eb3f15082bf766778f707176f3cd604bf0ef1008da06b8e5
+
+PKG_MAINTAINER:=John Marrett <johnf@zioncluster.ca>
+PKG_LICENSE:=
```
1715173329, line 12:
```suggestion
PKG_LICENSE:=BSD-3-Clause
```

from the COPYING file, looks like it's BSD 3-Clause

net/sscep/Makefile
```diff
@@ -0,0 +1,53 @@
...
+  CATEGORY:=Network
+  DEPENDS:=+libopenssl +openssl-util
+  TITLE:=A Simple SCEP (Simple Certificate Enrollment Protocol) client
+  URL:=https://github.com/certnanny/sscep/
+endef
+
+define Package/sscep/description
+  sscep is a simple SCEP client published by the CertNanny project.
+
+  It can be used to retrieve CA certificates and enroll devices with a compatible SCEP server.
```
1715173329, line 29:
```suggestion
  It can be used to retrieve CA certificates and enroll devices with
  a compatible SCEP server.
```

wrap new line for every 70 chars

net/sscep/Makefile
```diff
@@ -0,0 +1,53 @@
...
+endef
+
+define Build/Prepare
+$(call Build/Prepare/Default)
+	cd $(PKG_BUILD_DIR); ./bootstrap.sh
+endef
+
+define Build/Configure
+	cd $(PKG_BUILD_DIR); ./configure
+endef
```
1715173329, line 39:
```suggestion
```

it's for the autotools, can be removed for cmake

net/sscep/Makefile
```diff
@@ -0,0 +1,53 @@
...
+define Package/sscep/conffiles
+/etc/sscep.conf
+endef
+
+define Package/sscep/install
+	$(INSTALL_DIR) $(1)/etc
+	$(INSTALL_CONF) $(PKG_BUILD_DIR)/sscep.conf $(1)/etc/
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/sscep $(1)/usr/bin/sscep
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/mkrequest $(1)/usr/bin/
```
1715173329, line 50:
```suggestion

	$(INSTALL_DIR) $(1)/usr/bin
	$(INSTALL_BIN) $(PKG_BUILD_DIR)/sscep $(1)/usr/bin/sscep
	$(INSTALL_BIN) $(PKG_BUILD_DIR)/mkrequest $(1)/usr/bin/mkrequest
```

1715173329 (changes requested):

net/sscep/Makefile
jefferyto: From the link the file name looks like it will be v0.10.0.tar.gz, but if you click the link the download name is actually sscep-0.10.0.tar.gz.

net/sscep/Makefile
jefferyto: @BKPepe was right (see https://github.com/openwrt/packages/pull/20683#discussion_r1140684101), the file contains multiple licenses.

I think it would be most appropriate to list all of the licenses (space-separated) in `PKG_LICENSE`.

net/sscep/Makefile
johnfzc: That's very interesting @jefferyto, but I don't understand how it works. The actual URI used for the download is https://github.com/certnanny/sscep/archive/refs/tags/v0.10.0.tar.gz and the browser appears to be renaming the file.

Can you give me a raw uri that works and contains the sscep file name? I tried looking at the network tab in my browser to validate what was happening but couldn't find a URI or understand how the apparent rename is occurring.

net/sscep/Makefile
johnfzc: Ok, that makes sense, I'll do that.

johnfzc:
I haven't been able to work on this yet but will address all of the provided feedback. I also had tested this new version but apparently some remnants of my previous build made it (appear to?) work as the CI has all failed. I'll do a cleaner job of testing it to figure out why it's no longer building and fix that as well. If you see an obvious reason for the build failures please let me know.

Thanks for the comments that will make this package better and clean, really appreciate it.

net/sscep/Makefile
jefferyto: It's the `Content-Disposition: attachment; filename=sscep-0.10.0.tar.gz` header in the final 200 response, after all the redirections.

You can also use codeload as @1715173329 [suggested](https://github.com/openwrt/packages/pull/20683#discussion_r1148317548).

There are many packages here that download from GitHub, you can check other package makefiles to see how they do it.

net/sscep/Makefile
johnfzc: I wasn't able to add the OpenOSP license, I've put in the two that are defined in SPDX.

net/sscep/Makefile
johnfzc: Sorry, this request isn't clear to me. I tried to switch from tabs to spaces as I thought that was what you were suggesting but it prevented the Makefile from working. I looked at the msgpack-c Makefile and it appears to follow the same tab and space usage as is in my Makefile.

1715173329:
```patch
--- a/net/sscep/Makefile
+++ b/net/sscep/Makefile
@@ -37,6 +37,7 @@ endef
 define Package/sscep/install
 	$(INSTALL_DIR) $(1)/etc
 	$(INSTALL_CONF) $(PKG_BUILD_DIR)/sscep.conf $(1)/etc/
+
 	$(INSTALL_DIR) $(1)/usr/bin
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/sscep $(1)/usr/bin/sscep
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/mkrequest $(1)/usr/bin/mkrequest
```

1715173329 (approved):
lgtm from my side, but it would be nice to have a init script.

johnfzc:
>it would be nice to have a init script.

The sscep binary doesn't provide a running service. It's a client, like curl for http or an ftp client. You use it interactively, to request a certificate from a server on an on demand basis.

jefferyto:
The first license in COPYING is BSD-2-Clause not BSD-3-Clause (you can tell by the [number of clauses](https://github.com/certnanny/sscep/blob/v0.10.0/COPYING#L8-L13)), and the OpenOSP license is the [Vovida Software License v1.0](https://spdx.org/licenses/VSL-1.0.html) (also [named in the file](https://github.com/certnanny/sscep/blob/v0.10.0/COPYING#L148)).

johnfzc:
I can create a PR to clean that up, @jefferyto can you please confirm the line should be:

`PKG_LICENSE:=BSD-2-Clause OpenSSL VSL-1.0`

jefferyto:
> I can create a PR to clean that up, @jefferyto can you please confirm the line should be:
> 
> `PKG_LICENSE:=BSD-2-Clause OpenSSL VSL-1.0`

Yes that looks correct - thanks :pray:

johnfzc:
@jefferyto I've fixed up the license line and created PR #20846 please take a look at it and approve it or let me know if there are issues.

PR #20686, opened 2023-03-19
rust: refactor build system

1715173329:
Maintainer: @lu-zero
Compile tested: ipq807x/generic
Run tested: redmi-ax6

Description:
refactor the rust build system


lu-zero (approved):
Looks great! Thank you!

PolynomialDivision:
Awesome! Thanks.

Besides those changes, what's the status of the plan to reuse the llvm openwrt toolchain? AFAIK, it is already shipped with the SDK. So build time would reduce even more.

lu-zero:
this version is fetching the pre-built llvm from the rust-lang ci.

1715173329:
> Besides those changes, what's the status of the plan to reuse the llvm openwrt toolchain? AFAIK, it is already shipped with the SDK. So build time would reduce even more.

There're some issues with llvm-bpf:
- It's not defaultly enabled in buildroot
- We need to enable more targets in `LLVM_TARGETS_TO_BUILD` (looks like it should go to toolchain)
- Need to disable `LLVM_ENABLE_ZSTD` or rust will fail to lookup library.

1715173329:
The default configuration of llvm-bpf does not work for rust, I don't wanna add a broken option.
Given that you have to modify llvm-bpf manually, then just do it for rust again.

trippleflux:
Based on yoctoproject rust [Makefile](https://git.yoctoproject.org/poky/plain/meta/recipes-devtools/rust/rust_1.67.1.bb),  must-root is only valid for musl libc, while on glibc it's doesn't need it : 
```
 config.set(host_section, "cxx", e(d.expand("${RUST_TARGET_CXX}")))
    config.set(host_section, "cc", e(d.expand("${RUST_TARGET_CC}")))
    config.set(host_section, "linker", e(d.expand("${RUST_TARGET_CCLD}")))
    if "musl" in host_section:
        config.set(host_section, "musl-root", e(d.expand("${STAGING_DIR_HOST}${exec_prefix}")))

    # If we don't do this rust-native will compile it's own llvm for BUILD.
    # [target.${BUILD_ARCH}-unknown-linux-gnu]
    build_section = "target.{}".format(d.getVar('RUST_BUILD_SYS'))
```

So I proposing the following for OpenWrt Makefile : 
```
$(if $(CONFIG_USE_MUSL),--set=target.$(RUSTC_TARGET_ARCH).musl-root=$(TOOLCHAIN_DIR) \)
```

Also regarding CONFIG_PKG_CC_STACKPROTECTOR_REGULAR & CONFIG_PKG_CC_STACKPROTECTOR_STRONG in glibc, the "-lssp_nonshared" seems cannot be found.

What do you think of the following for rust-values.mk : 
```
ifeq ($(CONFIG_USE_MUSL),y)
  # Force linking of the SSP library
  ifdef CONFIG_PKG_CC_STACKPROTECTOR_REGULAR
    ifeq ($(strip $(PKG_SSP)),1)
      RUSTC_LDFLAGS += -lssp_nonshared
    endif
  endif
  ifdef CONFIG_PKG_CC_STACKPROTECTOR_STRONG
    ifeq ($(strip $(PKG_SSP)),1)
      TARGET_CFLAGS += -lssp_nonshared
    endif
  endif
endif

ifeq ($(CONFIG_USE_GLIBC),y)
  ifdef CONFIG_PKG_CC_STACKPROTECTOR_REGULAR
    ifeq ($(strip $(PKG_SSP)),1)
      TARGET_CFLAGS += --enable-stack-protector=yes
    endif
  endif
  ifdef CONFIG_PKG_CC_STACKPROTECTOR_STRONG
    ifeq ($(strip $(PKG_SSP)),1)
      TARGET_CFLAGS += --enable-stack-protector=strong
    endif
  endif
endif
```

lu-zero:
`--enable-stack-protector=strong` doesn't seem something you'd pass to `gcc`, TARGET_CFLAGS is being used to make sure that the stand-alone ssp gets linked in. glibc should have it as part of itself.

trippleflux:
Other finding, If I am enabled ccache, the HOSTCC will return as ccache then the Build/Compile/Cargo will failed, changing back **CC=cc** inside Build/Compile/Cargo will return successful rust app compilation.

1715173329:
Thanks for the test, it should be `$(HOSTCC_NOCACHE)`.

1715173329:
> So I proposing the following for OpenWrt Makefile :
> ```
> $(if $(CONFIG_USE_MUSL),--set=target.$(RUSTC_TARGET_ARCH).musl-root=$(TOOLCHAIN_DIR) \)
> ```

Looks good to me.


> ifeq ($(CONFIG_USE_GLIBC),y)
>   ifdef CONFIG_PKG_CC_STACKPROTECTOR_REGULAR
>     ifeq ($(strip $(PKG_SSP)),1)
>       TARGET_CFLAGS += --enable-stack-protector=yes
>     endif
>   endif
>   ifdef CONFIG_PKG_CC_STACKPROTECTOR_STRONG
>     ifeq ($(strip $(PKG_SSP)),1)
>       TARGET_CFLAGS += --enable-stack-protector=strong
>     endif
>   endif
> endif

https://github.com/openwrt/openwrt/blob/c98e09f01bc6788644d22ffd02cd7aaaef63f5ea/toolchain/glibc/common.mk#L64-L65

Looks like it's enabled already?

oskarirauta:
Seems like an improvement, I haven't tested yet; but looking at Makefile, it would seem that this fixes previous version's won't build if host and target are same arch. (x86_64 and x86_64 for example/most often).

I'll test later, but it looks like none of problems to build on a must host has not been fixed, so glibc hosts without changes only.

1715173329:
I just removed Host/PackageDist, seems useless at least for now

esaaprillia:
I still doubt rust will exist in openwrt

oskarirauta:
> I still doubt rust will exist in openwrt

May I ask why?

1715173329:
I'd say this guy just threw garbage everywhere, check his github activity

oskarirauta:
@1715173329 no need, I'll take your word on it.

PolynomialDivision:
@1715173329 I think you can merge? ;)

fuqiang03:
@1715173329 


gcc12 glibc2.37 master

PR #20687, opened 2023-03-19
strongswan: add support for pools section

### LINKED ISSUE #20663: strongswan: pools support for swanctl?
Environment: (put here arch, model, OpenWrt version)

Description:

It seems the swanctl init script doesn't support pools section. Maybe I missed something, but allowing a responder to allocate IP addresses and offer DNS etc seems like a pretty critical feature. I wonder if there is another option that can provide that feature. I don't understand why the strongswan packages have existed for so long without it.

If it turns out allocating IP addresses and offering DNS isn't a commonly used feature, I'd be happy to send a PR that adds support for pools. Let me know if you would accept such a PR.

hgl:
Compile tested: 22.03.2 x86-64
Run tested: 22.03.2 x86-64

Description:
Since I got no response, I went ahead and implemented it. This fixes #20663

### DIFF
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=strongswan
 PKG_VERSION:=5.9.10
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=https://download.strongswan.org/ https://download2.strongswan.org/
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -434,6 +434,7 @@ config_connection() {
 	local local_key
 	local ca_cert
 	local rekeytime
+	local remote_ca_certs
 	local pools
 
 	config_get_bool enabled "$1" enabled 0
@@ -458,6 +459,7 @@ config_connection() {
 	config_get overtime "$1" overtime
 
 	config_list_foreach "$1" local_sourceip append_var local_sourceip ","
+	config_list_foreach "$1" remote_ca_certs append_var remote_ca_certs ","
 	config_list_foreach "$1" pools append_var pools ","
 
 	case "$fragmentation" in
@@ -529,6 +531,7 @@ config_connection() {
 	swanctl_xappend2 "remote {"
 	swanctl_xappend3 "auth = $auth_method"
 	[ -n "$remote_identifier" ] && swanctl_xappend3 "id = \"$remote_identifier\""
+	[ -n "$remote_ca_certs" ] && swanctl_xappend3 "cacerts = \"$remote_ca_certs\""
 	swanctl_xappend2 "}"
 
 	swanctl_xappend2 "children {"
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -378,6 +378,41 @@ config_transport() {
 	config_child "$1" "transport"
 }
 
+config_pool() {
+	local conf="$1"
+	local addrs
+	local dns
+	local nbns
+	local dhcp
+	local netmask
+	local server
+	local subnet
+	local split_include
+	local split_exclude
+
+	config_get addrs "$1" addrs
+	config_list_foreach "$1" dns append_var dns ","
+	config_list_foreach "$1" nbns append_var nbns ","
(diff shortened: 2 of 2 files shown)


ptpt52:
mark!

hgl:
Also added support for `cacerts`

pprindeville (approved):
Looks good but it needs to be rebased.

Thermi:
What's still missing is the inclusion of arbitrary attributes. They are identified by a number and can have arbitrary values, as defined in the swanctl.conf configuration language. It's not a common use case but it is required to support some proprietary clients that expect particular attributes.

hgl:
Rebased.

> What's still missing is the inclusion of arbitrary attributes.

I'm aware of that, but I'm not sure what's the best way to implement that in UCI. Since this package has lived without it for a long time and no one currently request it, I'd say we could at least leave it for a future version.

ptpt52:
hi, how to setup a config for ikev1 + l2tp vpn ( or L2TP/IPsec PSK VPN)? could someone give an example?

1715173329:
hi, is this ready to go?

hgl:
Should be good to merge.

1715173329:
Kindly ping @pprindeville

pprindeville (approved):
LGTM

pprindeville:
It would be great if you could update the Wiki to describe usage as well.

Thermi:
> Rebased.
> 
> > What's still missing is the inclusion of arbitrary attributes.
> 
> I'm aware of that, but I'm not sure what's the best way to implement that in UCI. Since this package has lived without it for a long time and no one currently request it, I'd say we could at least leave it for a future version.

It's just a question of time when it will be requested. It's better to get it done now than later.

hgl:
@Thermi Any advice on how to get UCI options with dynamic name? `config_get <what?> "$1" <what?>` doesn't seem to work here.

Thermi:
@hgl you could try making these unknown keys all sections under the section "special-keys" and then have a key inside them called "value" that you can access. The key then has the value of the option assigned. You can iterate over sections using `config_foreach` according to the wiki [here](https://openwrt.org/docs/guide-developer/config-scripting#reading_options_config_get)

hgl:
You mean a config like this?

```
config 'special-keys' '1'
  option 'vaule' '...'
config 'special-keys' '2'
  option 'vaule' '...'
```
But then how do you assign these values to a particular pool? For example, how should such swanctl.conf be translated?

```
pools {
  a { 1 = '...' }
  b { 1 = '...' }
}
```

Thermi:
Something like that, yes. You could simply have the section "special-keys" just in the same place you have these other keys for the pool configuration.
For swanctl translation it just expands to 
pools {
  pool-a {
    1 = 'foobarbla'
  }
}
(just about what you already wrote yourself)

hgl:
I mean since sections are at the top level, how do you assign two sections with the same name to different pools?

For this conf
```
pools {
  a { 1 = 'x' }
  b { 1 = 'y' }
}
```
You need a config like 
```
config 'special-keys' '1'
  option 'vaule' '...'
config 'special-keys' '1'
  option 'vaule' '...'
```
How do we know which `'1'` is assigned to what pool? Should we also add a `pool` option in each `special-keys` section?

Thermi:
Oh, can you not just nest them within the tunnel configuration?

hgl:
How should we nest it? I'm not sure sections are nestable?

```
config pool 'a'
  config special-keys '1'
    option 'vaule' '...'
```
I'm not sure this works

Thermi:
Something like this, yes. I'm also not sure. An alternative would be to use an external configuration directory with per-tunnel files of the special keys as key-value assignments, one per line.

e.g.
/etc/uci/ipsec/special-keys/tunnel-a
```
1=foo[...]\n
2=bar[...]\n
...
```

hgl:
This is not really UCI any more.

I think users who really need such dynamic config still have the option to directly change the shipped swanctl.conf. I'm not sure it's worth it to abuse UCI to support it.

1715173329:
> @Thermi Any advice on how to get UCI options with dynamic name? `config_get <what?> "$1" <what?>` doesn't seem to work here.

https://openwrt.org/docs/guide-developer/config-scripting#option_cb_callback

These func may help

hgl:
@1715173329 

Thanks for the tip. 

However, it seems you to have to choose between the two parsing styles? If we use `config_cb`, then `config_foreach` doesn't work any more, which mean the whole init script should be rewritten. Again, not sure it's worth it just for dynamic pool keys.

1715173329:
Yes, it is indeed a huge work. From my side it's not that worth :-)

PR #20699, opened 2023-03-21
tor: Add CONFLICTS for tor-basic

jefferyto:
Maintainer: @hauke, @tripolar 
Compile tested: armvirt-32, 2023-03-19 snapshot sdk
Run tested: armvirt-32, 2023-03-19 snapshot

Description:
Currently, trying to install tor-basic while tor is installed results in data file clashes errors (for /etc/init.d/tor and /usr/sbin/tor).

This sets `CONFLICTS` for tor-basic so that proper package conflict checking takes place instead.

### DIFF
--- a/net/tor/Makefile
+++ b/net/tor/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tor
 PKG_VERSION:=0.4.7.13
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://dist.torproject.org/ \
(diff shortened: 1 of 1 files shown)


BKPepe (approved):
Yeah, thanks for spotting this and fixing it! We should verify that ``make menuconfig`` does not throw warnings about recursive dependencies.

jefferyto:
> We should verify that make menuconfig does not throw warnings about recursive dependencies.

Oops - you're right, there is a recursive dependency warning :frowning_face: 

But why doesn't nginx-ssl-util and nginx-ssl-util-nopcre (in [nginx-util](https://github.com/openwrt/packages/blob/ab573bdb4d7b30d5dc46e70941cb025800d4cabd/net/nginx-util/Makefile)) cause the same warning? Because of the extra comma?

jefferyto:
Closing due to recursive dependency warning.

PR #20710, opened 2023-03-23
treewide: refactor to use PKG_BUILD_FLAGS

dhewg:
This refactors packages to use the new PKG_BUILD_FLAGS feature added to the main repo.


jefferyto:
python3 supports lto for non-MIPS architectures (see https://github.com/openwrt/packages/commit/be9a554a26c6b9bcd7ed91550c16613d4c64f7b7). I wouldn't want it disabled for architectures where it works.

Also I think the main purpose of this PR is to switch to using `PKG_BUILD_FLAGS`; I think the change for python-greenlet should be in its own PR.

dhewg:
`--with-lto` is still used as it was before. The `no-lto` build flag added here just prevents the OpenWrt build system to add LTO to C/LDFLAGS on its own (so in addition to --with-lto). That would happen without the flag added here and CONFIG_USE_LTO=y.
It's been a while, but that didn't build for me, presumably because `--with-lto` handles LTO differently and apparently appropriate ;)
Anyway, python3 should just build with LTO as before.

I can move python-greenlet to it's own PR, but if the comments about the mips16 usage there are correct - and @commodo confirmed that, see the linked commit ^ - this is directly related, which is why I included it here. But both ways work for me

dhewg:
And having said that, I just added that as comment (duh)

jefferyto:
> Anyway, python3 should just build with LTO as before.

Okay - it looks confusing, and lto should actually be enabled for supported platforms if `CONFIG_USE_LTO` is selected (I can do this later), but this is fine for now.

> I can move python-greenlet to it's own PR, but if the comments about the mips16 usage there are correct - and @commodo confirmed that, see the linked commit ^ - this is directly related, which is why I included it here. But both ways work for me

I can argue how this one commit is not actually related to this PR as a whole, but in the end whether it is one PR or two all the changes will probably be merged anyway and I'm the only one that cares, so I'm going to stop typing now.

Thanks for taking care of this btw.

1715173329 (approved):

1715173329:
@dhewg could you please rebase this PR upon the latest git HEAD?

dhewg:
Done, and I removed the additional changes for python-greenlet like @jefferyto suggested - in the hope that this gets merged a bit faster.

dhewg:
Nice, thanks!

PR #20714, opened 2023-03-25
pax-utils: add new package

oskarirauta:
Description:
A suite of ELF tools to aid auditing systems.
Contains various ELF related utils for ELF32, ELF64 binaries useful for displaying PaX and security info on a large groups of binary files.

Maintainer: me / @oskarirauta

### DIFF
--- a/utils/pax-utils/Makefile
+++ b/utils/pax-utils/Makefile
@@ -1,0 +1,52 @@
+#
+# Copyright (C) 2019 Daniel Engberg <daniel.engberg.lists@pyret.net>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=pax-utils
+PKG_VERSION:=1.3.7
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=https://dev.gentoo.org/~sam/distfiles/app-misc/pax-utils/
+PKG_HASH:=108362d29668d25cf7b0cadc63b15a4c1cfc0dbc71adc151b33c5fe7dece939a
+
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+PKG_LICENSE:=GPL-2.0-only
+PKG_LICENSE_FILES:=LICENSE
+
+include $(INCLUDE_DIR)/package.mk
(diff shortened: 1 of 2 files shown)


utils/pax-utils/Makefile
```diff
@@ -0,0 +1,52 @@
...
+  PaX and security info on a large groups of binary files.
+endef
+
+define Package/pax-utils/config
+  source "$(SOURCE)/Config.in"
+endef
+
+MESON_ARGS += \
+	-Dlddtree_implementation=sh \
+	-Duse_seccomp=$(if $(CONFIG_PAX_UTILS_SECCOMP),true,false)
```
1715173329, line 45: I'm not sure if it requires `libseccomp` package, otherwise LGTM.

oskarirauta: @1715173329 

oops... I thought you wanted to remove sec comp support completely so I removed it :O 
You are probably right, it probably would had needed it as a depency - while testing, I didn't enable seccomp support so didn't notice that.

Want it back?

1715173329: I'm sorry if I made any confusion...
I just meant if it requires `libseccomp` to work then it should be added to dependencies

oskarirauta: Not your fault, I am just so busy currently so I speed readed it and look what happened when in haste. I added it back with libseccomp depency, and now I need to remove it :)

```
dev@media:/usr/src/openwrt/build_dir/target-x86_64_musl/pax-utils-1.3.7# cat meson_options.txt 
option('lddtree_implementation', type : 'combo',
  choices : ['python', 'sh', 'none'], value : 'python',
  description : 'Which lddtree implementation to install?')
option('use_libcap', type : 'feature', value : 'auto',
  description : 'Enable listing capabilities in pspax output (requires libcap)'
)
option('use_seccomp', type : 'boolean', value : true,
  description : 'Enable seccomp filters at runtime (does *not* require libseccomp, but does require kernel support)'
)
option('build_manpages', type : 'feature', value : 'auto',
  description : 'Build manuals via DocBook (requires xmlto)')
option('tests', type : 'boolean', value : true,
  description : 'Enable testing (not guaranteed to work)'
)
option('use_fuzzing', type : 'boolean', value : true,
  description : 'Also build LibFuzzer fuzzers as tests'
)

```

And just remembered that when I made this PR, I had seen that but no longer could remember it..
Ok, now it's back to what it was 10-15 minutes ago and good to go again :)

1715173329: Thanks for your time!

1715173329 (approved):

PR #20726, opened 2023-03-26
curl: update to 8.0.1

stangri:
Maintainer: me
Compile tested: x86-64, Sophos XG-135r3, OpenWrt 22.03.3
Run tested: x86-64, Sophos XG-135r3, OpenWrt 22.03.3, test functionality with `https-dns-proxy`

Description:
* https://curl.se/changes.html#8_0_1

### DIFF
--- a/net/curl/Makefile
+++ b/net/curl/Makefile
@@ -9,15 +9,15 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/nls.mk
 
 PKG_NAME:=curl
-PKG_VERSION:=7.88.1
-PKG_RELEASE:=2
+PKG_VERSION:=8.0.1
+PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


BKPepe:
Hi!

I think, this PR should be splitted into 2 commits. The 1st one with
enabling unix sockets and the 2nd with the update. It should not be
together and they are not connected to each other.

Right now, I am not sure, why there are opened 2 PRs for the same thing.
What was the reason?

And yes, according to OpenWrt contributing guidelines, we should mention
the size.

Dne ne 26. 3. 2023 21:38 uživatel Stan Grishin ***@***.***>
napsal:

champtar:
@stangri I just merged #20600, you can rebase on top

stangri:
@champtar rebased. @BKPepe Unix sockets change removed.

BKPepe (approved):

champtar (approved):

PR #20728, opened 2023-03-26
Created utils/random-numgen package, an alternative to $RANDOM

ilario:
Maintainer: me / @ilario
Compile tested: OpenWrt 22.03 and OpenWrt 19.07, YouHua WR1200JS
Run tested: OpenWrt 22.03 and OpenWrt 19.07, YouHua WR1200JS

Description: 
By default, the `$RANDOM` shell variable is not available as Busybox does not have the `BUSYBOX_CONFIG_ASH_RANDOM_SUPPORT` option activated.
This command gives an equivalent output, an integer from 0 to 65535.
Originally, we decided we needed such a command in LibreMesh to replace the usage of `$RANDOM`, as discussed here https://github.com/libremesh/lime-packages/issues/800 and here https://github.com/libremesh/lime-packages/pull/980.

I would be nice to have this also in the openwrt-22.03 branch.

### DIFF
--- a/utils/random-numgen/Makefile
+++ b/utils/random-numgen/Makefile
@@ -1,0 +1,39 @@
+# This is free software, licensed under the GNU General Public License v3.
+# See /LICENSES for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=random-numgen
+PKG_VERSION:=0.1
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=Ilario Gelmetti <ilario@sindominio.net>
+PKG_LICENSE:=GPL-3.0-or-later
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/random-numgen
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Generates a random number 0-65535
+endef
+
+define Package/random-numgen/description
+  Offers an alternative to the RANDOM shell variable,
+  generating a pseudo-random integer number from 0 to
+  65535 using /dev/urandom as a random data source.
+endef
+
+define Build/Compile
+endef
+
+define Build/Configure
+endef
+
+define Package/random-numgen/install
+	$(INSTALL_DIR) $(1)/usr/bin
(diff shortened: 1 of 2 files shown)


ilario:
Just saw the "Formalities" test failing, I am going to fix it ASAP

BKPepe (changes requested):
I am not sure... You should rather active ``BUSYBOX_CONFIG_ASH_RANDOM_SUPPORT`` in your build instead of adding your own custom package for it.

And no, this will not be backported to OpenWrt 22.03.

1715173329:
since this is a very simple wrapper, imo it should go to somewhere like `package/base-files/files/lib/functions.sh`

ilario:
Thanks @BKPepe for the corrections!
Is it better now?
Following @1715173329's suggestion, I opened an alternative pull request here: https://github.com/openwrt/openwrt/pull/12291

> You should rather active `BUSYBOX_CONFIG_ASH_RANDOM_SUPPORT` in your build instead of adding your own custom package for it.

That is what we were doing in LibreMesh, but that implies that we are not compatible with pristine OpenWrt's ImageBuilder.

> And no, this will not be backported to OpenWrt 22.03.

Ok

ilario:
> @ilario what is the size increase of enabling `BUSYBOX_CONFIG_ASH_RANDOM_SUPPORT` ?

The same was asked by @aparcar here: https://github.com/openwrt/openwrt/pull/12291#issuecomment-1490579190 :)

Is there a proper way to measure that?

I compiled and then run binwalk on the `-sysupgrade.bin` image.
The result is that the third part of the file is a Squashfs filesystem with a size of:

* 3453220 bytes for clean OpenWrt (compiled for YouHua WR1200JS)
* 3453332 bytes when selecting `BUSYBOX_CUSTOM`
* 3453408 bytes when selecting also `BUSYBOX_CONFIG_ASH_RANDOM_SUPPORT`

So in total, it increases by only 188 bytes, if I measured it correctly.

ilario:
Closing as fixed in https://github.com/openwrt/openwrt/pull/12465 and https://github.com/openwrt/openwrt/commit/8f427f1a058dd5dcff21246a9a6d91318f55f80a

PR #20731, opened 2023-03-26
isc-dhcp: Update to 4.4.3-P1

pprindeville:
Maintainer: me
Compile tested: x86_64, generic, HEAD (c98e09f01b)
Run tested: same, deployed on production router

Description:

Fixes CVE-2022-2928 and CVE-2022-2929


pprindeville:
@nbd168 @dangowrt Is this the correct way to handle a virtual package?

pprindeville:
Didn't get any reviews but the CVE's need to be mitigated.  If there are subsequent comments, I'll address them separately.

PR #20745, opened 2023-03-28
rust: update to 1.68.1

oskarirauta:
also add new variable CARGO_VARS to make it possible to pass environment variables for cargo process.
This is necessary when for example, cross-compiling netavark.

Maintainer: Luca Barbato / @lu-zero 

Description:
example of CARGO_VARS usage is available [here](https://github.com/oskarirauta/openwrt-rust/blob/main/net/netavark/Makefile) -  netavark requires pointing protoc binary to it (staging_dir/hostpkg/bin/protoc) as environment value PROTOC.

### DIFF
--- a/lang/rust/rust-package.mk
+++ b/lang/rust/rust-package.mk
@@ -27,7 +27,8 @@ define Build/Compile/Cargo
 		CARGO_HOME=$(CARGO_HOME) \
 		TARGET_CFLAGS="$(TARGET_CFLAGS) $(RUST_CFLAGS)" \
 		TARGET_CC=$(TARGET_CC_NOCACHE) \
-		CC=$(HOSTCC) \
+		CC=$(HOSTCC_NOCACHE) \
+		$(CARGO_VARS) \
 		cargo install -v \
 			--profile stripped \
 			--target $(RUSTC_TARGET_ARCH) \
--- a/lang/rust/Makefile
+++ b/lang/rust/Makefile
@@ -5,12 +5,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=rust
-PKG_VERSION:=1.68.0
+PKG_VERSION:=1.68.1
 PKG_RELEASE:=1
 
 PKG_SOURCE:=rustc-$(PKG_VERSION)-src.tar.gz
 PKG_SOURCE_URL:=https://static.rust-lang.org/dist/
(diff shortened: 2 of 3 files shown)


lu-zero:
The patch looks incomplete regarding the CARGO_VARS, did you forget to add it?

To make the CI happy add or bump a rust package.

`--set=target.$(RUSTC_TARGET_ARCH).crt-static=false` is definitely a good idea.

```
+	--set=rust.codegen-units=1 \
+	--set=rust.codegen-units-std=1 \
+	--set=rust.parallel-compiler=false \
```
this may help getting a bit more optimization but also slow down building rustc.

```
+	--tools="analysis,cargo,clippy,src,rustfmt" \
```

Probably only `cargo` is really needed and `src` might be useful.

Thank you for further refining it :)

oskarirauta:
@lu-zero
Thanks for noticing, seems that I forgot to add rust-package.mk and rust-values.mk to my commit.

Yes, I tried so many combinations until getting it right; fine-tuning could be in place- and maybe it could be slimmed down, but for that reason, I did not provide a PR on that subject; I am not expert enough on realising effects of my choices, but bringing down a sort-of-solution might assist someone else to start on it, but open conversation on topic is healthy. My aim is to get it more general to make it work on more platforms, but even though I succeeded, I don't say that this is perfect ten. Currently I run with cargo from my own repo with these changes, but hopefully time comes soon to it being no longer necessary.

oskarirauta:
@lu-zero no, you go ahead for that other stuff- I think this is as far as I should go with this for now :)
few other packages still need some of my caring so I'll check later what you cooked up. Still merge this though.

lang/rust/rust-package.mk
```diff
@@ -28,6 +28,7 @@ define Build/Compile/Cargo
 		TARGET_CFLAGS="$(TARGET_CFLAGS) $(RUST_CFLAGS)" \
 		TARGET_CC=$(TARGET_CC_NOCACHE) \
 		CC=$(HOSTCC) \
```
1715173329, line 30:
```suggestion
		CC=$(HOSTCC_NOCACHE) \
```

I overlooked this change in my previous PR. Would you mind assisting me in fixing it?

oskarirauta: Sure thing

oskarirauta: Done.

1715173329: Thank you!

PR #20747, opened 2023-03-30
respondd: add new package

blocktrron:
Maintainer: me
Compile tested: mpc85xx-p1020 / mediatek-filogic
Run tested: Extreme WS-AP3825 / ASUS TUF-AX4200
Description:

respondd is a server distributing information within a network.

For doing so, respondd spawns a UDP socket (in Gluon 1001/udp), optionally joining a multicast group. When a request is received, the information requested is transmitted to the requester.

All information is organized in a non-hierarchical namespace. Each entry identifies a request name (e.g. statistics, nodeinfo, ...) implemented by at least one "provider" C module. The respond is the result of merging the outputs of all providers for the given request name.

### DIFF
--- a/net/respondd/Makefile
+++ b/net/respondd/Makefile
@@ -1,0 +1,64 @@
+#
+# Copyright (C) 2023 David Bauer <mail@david-bauer.net>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=respondd
+PKG_SOURCE_DATE:=2019-05-01
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/freifunk-gluon/respondd.git
+PKG_SOURCE_VERSION:=58405297e76f97d4752dc7511dc15bbc7a01e586
+PKG_MIRROR_HASH:=22bc00df280fe611cfc895528f17269f03af804ea4c2ca2dc62da2457d3a1ee0
+
+PKG_MAINTAINER:=David Bauer <mail@david-bauer.net>
+PKG_LICENSE:=BSD-2-Clause
+
+CMAKE_SOURCE_SUBDIR:=src
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+define Package/respondd
+  SECTION:=net
+  CATEGORY:=Network
+  DEPENDS:=@IPV6 +libjson-c
+  TITLE:=Lightweight metadata multicast-responder
+endef
+
+define Package/respondd/description
+respondd is a server distributing information within a network.
+
+For doing so, respondd spawns a UDP socket (in Gluon 1001/udp),
+optionally joining a multicast group. When a request is received,
+the information requested is transmitted to the requester.
+
+All information is organized in a non-hierarchical namespace.
+Each entry identifies a request name (e.g. statistics, nodeinfo, ...)
+implemented by at least one "provider" C module. The respond is the
+result of merging the outputs of all providers for the given request
+name.
+endef
+
+define Package/respondd/conffiles
+/etc/config/respondd
+endef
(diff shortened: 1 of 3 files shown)


net/respondd/Makefile
```diff
@@ -0,0 +1,50 @@
...
+CMAKE_SOURCE_SUBDIR:=src
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+define Package/respondd
+  SECTION:=net
+  CATEGORY:=Network
+  DEPENDS:=@IPV6 +libjson-c
+  TITLE:=Responds to multicast queries with answers generated by standalone information providers
```
BKPepe, line 31: This seems too long for me as I think that this is not going to be shown in ``make menuconfig``.

net/respondd/Makefile
```diff
@@ -0,0 +1,50 @@
...
+endef
+
```
BKPepe, line 33: There is a missing description for this package.

net/respondd/Makefile
```diff
@@ -0,0 +1,50 @@
...
+define Package/respondd/conffiles
+/etc/config/respondd
+endef
+
+define Package/respondd/install
+	$(INSTALL_DIR) $(1)/usr/bin $(1)/etc/init.d $(1)/etc/config
+	$(CP) ./files/respondd.config $(1)/etc/config/respondd
```
BKPepe, line 40: Can we use ``INSTALL_CONF``?

net/respondd/Makefile
```diff
@@ -0,0 +1,50 @@
...
+PKG_SOURCE_DATE:=2019-05-01
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/freifunk-gluon/respondd.git
+PKG_SOURCE_VERSION:=58405297e76f97d4752dc7511dc15bbc7a01e586
+PKG_MIRROR_HASH:=22bc00df280fe611cfc895528f17269f03af804ea4c2ca2dc62da2457d3a1ee0
+
+PKG_MAINTAINER:=David Bauer <mail@david-bauer.net>
+PKG_LICENSE:=BSD-2-Clause
```
BKPepe, line 20: Hmm, I dont see any LICENSE in that repository.

BKPepe (changes requested):
I am just thinking a little bit loud. This package is primarily to be used with Freifunk or Gluon, right? In the past, there was added Freifunk feed to the OpenWrt, but it did not end up very well as it was lately removed. Shouldn't be this package  in that feed if users want to use it?

Also the repository is not quite active, though.

blocktrron:
@BKPepe

The package is not Freifunk-generic but it enables OpenWrt APs not primarily used for Freifunk nettworks to provide monitoring information for a setup with such infrastructure (YANIC).

What drives the conclusion the Freifunk feed is the appropriate place, given it is not even built anymore and thus serves no purpose when using imagebuilder / opkg?

net/respondd/Makefile
blocktrron: See the source-file. As well as the package Makefile contained in the Repo.

net/respondd/Makefile
blocktrron: Simplified.

net/respondd/Makefile
blocktrron: changed

net/respondd/Makefile
blocktrron: Added

net/respondd/Makefile
BKPepe: I've checked it, but none files in src includes license header, there is no separate license file and only where there is something about BSD-2-Clause is here https://github.com/freifunk-gluon/respondd/blob/58405297e76f97d4752dc7511dc15bbc7a01e586/Makefile#L6

blocktrron: How about https://github.com/freifunk-gluon/respondd/blob/master/src/respondd.c#L7 ?

PR #20749, opened 2023-03-30
fio: update to 3.34

TjeuKayim:
Maintainer: @neheb
Compile tested: with container image openwrtorg/sdk:ath79-generic-22.03.2
Run tested: installed mips_24kc/packages/fio_3.34-1_mips_24kc.ipk with opkg on the glinet router with OpenWRT 22.03.2, confirmed that `fio --version` returns the new version number, and that it can run jobs too

Description:
Upgrading to 3.34, because that includes a bugfix for https://github.com/axboe/fio/issues/1372


BKPepe (commented):
I am not sure if we should allow contribution with GitHub's no-reply email. And it would be quite nice to include release notes (at least the link for them)

TjeuKayim:
> not sure if we should allow contribution with GitHub's no-reply email

f4058c1c18138ae1011abfe0e7aca74659834b21 that already landed in master and https://git.openwrt.org/?p=project/iwinfo.git;a=commit;h=705d3b5cc30ed1ac3b2fed131259bcbc61177e41 also used this same email address.

> include release notes 

A changelog I could not find, though this is the email introducing the release: https://lore.kernel.org/fio/20230315120002.382EF1BC012D@kernel.dk/

BKPepe:
That's interesting. I think it would be good to have any response from OpenWrt core members e.g., @hauke. Would you mind telling us if that's ok? Since it was said that some commits with NoReply email address were already merged into OpenWrt main repository.

1715173329:
Anyway, please use a valid email address  even though there's a exception.

TjeuKayim:
> please use a valid email address

Of course, is it better [this way](https://github.com/TjeuKayim/openwrt-packages/commit/1f2bd146be21cb012d19cd8e279526eb860a1c97) where I changed it to the same email address as my iwinfo patches?

1715173329:
Looks good, thank you!

PR #20751, opened 2023-03-31
afalg_engine, gost_engine: adjust to new openssl module infrastructure

cotequeiroz:
Maintainer: me
Compile tested: mediatek/mt7622, Linksys E8450, master
Run tested: same, tested with openssl cmdline util

Description:
This should be merged after https://github.com/openwrt/openwrt/pull/12151.
The package include file was renamed from `openssl-engine.mk` to `openssl-module.mk`.  Also, the name of the config section for the engine is now the name of the engine with `_sect` appended.

This removes the hack to build different versions of gost_engine depending on the openssl version in the main repository.
Also, the "software fallback" feature of afalg_engine was removed, since it will not be enabled with OpenSSL 3.0.

### DIFF
--- a/libs/afalg_engine/Config.in
+++ b/libs/afalg_engine/Config.in
@@ -9,19 +9,6 @@ if PACKAGE_libopenssl-afalg_sync
 	    This increases memory usage, and has problems when process fork
 	    with open digest contexts (openssh will not work because of it).
 
-    config AFALG_FALLBACK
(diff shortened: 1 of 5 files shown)


cotequeiroz:
OpenSSL change was committed, but it takes a while to update the SDKs used by CI.  I've tested the packages locally, so I will merge this now, to minimize build failures.

PR #20754, opened 2023-04-01
syslog-ng: update to version 4.1.1

BKPepe:
Maintainer: me
Compile tested: Turris 1.1, mpc85xx/powerpc8540, OpenWrt 21.02.5
Run tested: Turris 1.1, mpc85xx/powerpc8540, OpenWrt 21.02.5

Description:
- Release notes: https://github.com/syslog-ng/syslog-ng/releases/tag/syslog-ng-4.1.1

- Updated version in config

### DIFF
--- a/admin/syslog-ng/Makefile
+++ b/admin/syslog-ng/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=syslog-ng
-PKG_VERSION:=3.38.1
+PKG_VERSION:=4.1.1
 PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_MAINTAINER:=Josef Schlehofer <josef.schlehofer@nic.cz>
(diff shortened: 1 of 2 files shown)


BKPepe:
I will keep autorelease, for now, as there are also many packages, which are using it, and I like the idea of autorelease. Otherwise, I could forget about bumping the release.

admin/syslog-ng/Makefile
```diff
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=syslog-ng
-PKG_VERSION:=3.38.1
+PKG_VERSION:=4.1.1
 PKG_RELEASE:=$(AUTORELEASE)
```
pprindeville, line 5: Should be `PKG_RELEASE:=1`

pprindeville:
> I will keep autorelease, for now, as there are also many packages, which are using it, and I like the idea of autorelease. Otherwise, I could forget about bumping the release.

It's in packages that haven't been touched/updated.

BKPepe:
That could be true, but still an issue there, I see that several packages are using autorelease, others are not and autorelease is still supported even though it is deprecated. This needs to be addressed by treewide commit. 

The reason why the autorelease is getting deprecated is because of increased Internet traffic, and if it would be a such issue, then it is addressed ASAP. That's not happening and right now, we do have, let's say situation since 2nd December (more than 4 months), when we are still determining what is going to be next.

BKPepe:
The main reason I insist on using autorelease is that we are in a situation where we need to check Makefile - if it is using autorelease or not and if it not, we need to request the contributor to bump it. The cases when someone forgets it are getting again increased. :-(

pprindeville:
> The main reason I insist on using autorelease is that we are in a situation where we need to check Makefile - if it is using autorelease or not and if it not, we need to request the contributor to bump it. The cases when someone forgets it are getting again increased. :-(

That should be easy enough to cover in a CI/CD check...

BKPepe:
I agree with you, but nobody has done it yet.

PR #20760, opened 2023-04-02
modemmanager: improve mobile router setup

dangowrt:
Maintainer: Nicholas Smith (in Makefile), @aleksander0m (in reality)
Compile tested: ramips/mt7620
Run tested: ZBT-WE826-E with Quectel EC25-EG

Description:
ModemManager currently only tries once to connect the a network. It doesn't retry in case the device or wwan interface was started outside of mobile network coverage. It also doesn't reconnect in case the connection was lost and network access was regained at a later point.
Address this by letting netifd track the connection state similar to how we do it for ppp connections. In addition also make sure restarting is only blocked in case of wrong credentials or settings.

### DIFF
--- a/net/modemmanager/files/modemmanager.proto
+++ b/net/modemmanager/files/modemmanager.proto
@@ -404,6 +404,8 @@ proto_modemmanager_setup() {
 		return 1
 	}
 
+	proto_run_command "$interface" mmcli --modem="${device}" --track-state
+
 	# check if Signal refresh rate is set
 	if [ -n "${signalrate}" ] && [ "${signalrate}" -eq "${signalrate}" ] 2>/dev/null; then
 		echo "setting signal refresh rate to ${signalrate} seconds"
--- a/net/modemmanager/files/modemmanager.proto
+++ b/net/modemmanager/files/modemmanager.proto
@@ -352,7 +352,7 @@ proto_modemmanager_setup() {
 	local interface="$1"
 
 	local modempath modemstatus bearercount bearerpath connectargs bearerstatus beareriface
-	local bearermethod_ipv4 bearermethod_ipv6 auth cliauth
+	local bearermethod_ipv4 bearermethod_ipv6 auth cliauth state
 	local operatorname operatorid registration accesstech signalquality
 
 	local device apn allowedauth username password pincode iptype metric signalrate
@@ -385,6 +385,22 @@ proto_modemmanager_setup() {
 	}
 	echo "modem available at ${modempath}"
 
+	state="$(modemmanager_get_field "${modemstatus}" "state")"
+	state="${state%% *}"
+	if [ "$state" = "locked" ]; then
+		if [ -n "$pincode" ]; then
+			mmcli --modem="${device}" -i any --pin=${pincode} || {
+				proto_notify_error "${interface}" MM_WRONG_PIN
+				proto_block_restart "${interface}"
+				return 1
+			}
+		else
+			proto_notify_error "${interface}" MM_PIN_REQUIRED
+			proto_block_restart "${interface}"
+			return 1
+		fi
+	fi
+
 	# always cleanup before attempting a new connection, just in case
 	modemmanager_cleanup_connection "${modemstatus}"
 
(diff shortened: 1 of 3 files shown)


net/modemmanager/files/modemmanager.proto
```diff
@@ -385,6 +385,22 @@ proto_modemmanager_setup() {
 	}
 	echo "modem available at ${modempath}"
 
+	state="$(modemmanager_get_field "${modemstatus}" "state")"
+	state="${state%% *}"
+	if [ "$state" = "locked" ]; then
```
feckert, line 390: How about making a switch case statement here? In my case I have a system that does not have a SIM card and therefore sets the state to `failed`. The value `state-failed-reason` is set to `sim-missing`. Currently it is always trying to configure the interface in an endless loop. 
But this makes no sense, because the SIM is missing.
It would be better to stop now.
With an error:
```
proto_notify_error "${interface}" MM_SIM_MISSING
proto_block_restart "${interface}"
return 1
```

net/modemmanager/files/modemmanager.proto
```diff
@@ -393,17 +409,30 @@ proto_modemmanager_setup() {
 		cliauth="${cliauth}${cliauth:+|}$auth"
 	done
 
+
```
feckert, line 412: additional new line

ptpt52:
@dangowrt 
This might be off-topic, but I'd like to inquire about something: Does ModemManager support QMAP, and if so, how can it be utilized?

lynxthecat:
@feckert is it apparent to you whether this is compatible with, or would conflict with:

https://github.com/openwrt/packages/commit/bc754f31cfdb004eefa43038f8f0827922107fc6

net/modemmanager/files/modemmanager.proto
```diff
@@ -393,17 +409,30 @@ proto_modemmanager_setup() {
...
+	mmcli --modem="${device}" --timeout 120 --enable || {
+		proto_notify_error "${interface}" MM_MODEM_DISABLED
+		return 1
+	}
+
+	mmcli --modem="${device}" --3gpp-register-home || {
```
patrakov, line 418: Some people use SIM cards while roaming. In mainland China, using a Hong Kong SIM is a way to get out of the firewall. Also, some virtual-only mobile operators (MNVOs) do not have a home network. So, please don't hard-code the requirement to register in the home network.

lynxthecat:
Any thoughts on how reconnection speed might compare with the dispatcher script method using ifdown and ifup? 

```

root@OpenWrt:~# cat /usr/lib/ModemManager/connection.d/10-report-down-and-reconnect
#!/bin/sh

# Automatically report to netifd that the underlying modem
# is really disconnected and reconnect if interface was up

# require program name and at least 4 arguments
[ $# -lt 4 ] && exit 1

MODEM_PATH="$1"
BEARER_PATH="$2"
INTERFACE="$3"
STATE="$4"

[ "${STATE}" = "disconnected" ] || exit 0

. /usr/share/ModemManager/modemmanager.common
. /lib/netifd/netifd-proto.sh
INCLUDE_ONLY=1 . /lib/netifd/proto/modemmanager.sh

MODEM_STATUS=$(mmcli --modem="${MODEM_PATH}" --output-keyvalue)
[ -n "${MODEM_STATUS}" ] || exit 1

MODEM_DEVICE=$(modemmanager_get_field "${MODEM_STATUS}" "modem.generic.device")
[ -n "${MODEM_DEVICE}" ] || exit 2

CFG=$(mm_get_modem_config "${MODEM_DEVICE}")
[ -n "${CFG}" ] || exit 3

IFUP=$(ifstatus ${CFG} | jsonfilter -e '@.up')

logger -t "modemmanager" "interface ${CFG} (network device ${INTERFACE}) ${STATE}"
proto_init_update $INTERFACE 0
proto_send_update $CFG

[ "${IFUP}" = "true" ] && ifup ${CFG}

exit 0
```

The latter is too slow for my liking and I wonder if this proto command and tracking approach will offer a faster reconnect or not?

patrakov:
> Any thoughts on how reconnection speed might compare with the dispatcher script method using ifdown and ifup?

You can test this without building the patched package, by switching the protocol to PPP (luci-proto-3g). Yes, on paper, it does not support LTE and actively disables it, but de-facto, if you edit the `3g.sh` file by removing the line below, it will not touch the mode and will effectively ignore the `service` option as long as it is not `evdo` or `cdma`.

```
[ -n "$MODE" ] && gcom -d "$device" -s /etc/gcom/setmode.gcom
```

Important: this line should be removed before the first connection attempt, otherwise, your modem will be crippled and hard to uncripple (use `mmcli -m 0 --set-allowed-modes='3G|4G' --set-preferred-mode=4G` or a variation thereof if it happens).

The purpose of this test is to force the modem to use the obsolete PPP protocol, which, just like the proposed solution, relies on a daemon that exits when the connection breaks, and should, therefore, have the same reconnection speed.

Yes, I know that PPP is slow due to the asyncmap-byte-escaping overhead.

jeffsf:
Testing the reconnect script here on 23.05.2 and a Sierra Wireless EM7565, T-Mobile US. Please let me know if any specific tests or information would be helpful.

feckert:
I have taken ModemManager changes by @dangowrt and opened a [merge request](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/merge_requests/1131) to the Modemmanager upstream repository .

I will close this. The pull request is replaced by https://github.com/openwrt/packages/pull/23388

PR #20776, opened 2023-04-05
ddns-scripts: Enable no-ip.com by default.

pleaseking:
ddns-scripts: Enable no-ip.com by default.

Signed-off-by: Michael Pliskin <michael.pliskin@gmail.com>


1715173329:
This doesn't seem alright, are you looking for package `ddns-scripts-noip`?

feckert:
As far as I can remember, only providers that work out of the box and therefore do not need an extra script should be listed in this file. If the provider needs an extra script, then this must be installed via the package managemnen system opkg.

PR #20633, opened 2023-03-10
Package dependencies for ble2mqtt python package

QuintinHill:
Maintainer: me 
Compile tested: Unfortunately not (I tried without success to build this out on my machine but I think this is more of a failure in my setup.)  Hopefully CI will test this.
Run tested: Unfortunately not

Description: [ble2mqtt](https://github.com/devbis/ble2mqtt) is an interesting python package unfortunately it requires [Bleak](https://github.com/hbldh/bleak) which requires [dbus-fast](https://github.com/Bluetooth-Devices/dbus-fast) and this latter package cannot be successfully installed on a router with 128MB of RAM using pip (it runs out of memory and crashes).  This PR packages the dependencies for ble2mqtt but dbus-fast is the priority here.

PR #20781, opened 2023-04-07
Python ble2mqtt

QuintinHill:
Maintainer: me
Compile tested: ath79 22.03.3 SDK
Run tested: Netgear r6260 snapshot successfully run ble2mqtt as a daemon and collected data over BLE.

Description:
Follow up to #20633.  This bumps the version of Bleak to get a bugfix, adds a missing dependency to ble2mqtt and also adds an init script (from the upstream README) to the ble2mqtt package.


QuintinHill:
CI seems to be failing while compiling OpenSSL.

Issue #15433, opened 2021-04-15
lxc-start: The container failed to start (21.02)


erdoukki:
https://forum.openwrt.org/t/lxc-mvebu-snapshot-21-02/94033

I have just installed the snapshot 21-02 for MVEBU EspressoBIN...
I am surprised that LXC looks like to be integrated in this future release (great news if it is) !
I am trying to make a LXC container but get an error : 
```
root@OWRT-DEV:/# lxc-create --name myBUS --template download -- --dist debian --
release buster --arch arm64                                                     
Failed to create lock for myBUS                                                 
lxc-create: myBUS: tools/lxc_create.c: main: 260 Failed to create lxc container 
```

```
root@OWRT-DEV:/# uname -ar                                                      
Linux OWRT-DEV 5.4.111 #0 SMP Wed Apr 14 06:42:38 2021 aarch64 GNU/Linux        
```

```
root@OWRT-DEV:/# lxc-checkconfig                                                
LXC version 4.0.5                                                               
--- Namespaces ---                                                              
Namespaces: enabled                                                             
Utsname namespace: enabled                                                      
Ipc namespace: enabled                                                          
Pid namespace: enabled                                                          
User namespace: enabled                                                         
Network namespace: enabled                                                      
                                                                                
--- Control groups ---                                                          
Cgroups: enabled                                                                
                                                                                
Cgroup v1 mount points:                                                         
/sys/fs/cgroup/cpuset                                                           
/sys/fs/cgroup/cpu                                                              
/sys/fs/cgroup/cpuacct                                                          
/sys/fs/cgroup/blkio                                                            
/sys/fs/cgroup/memory                                                           
/sys/fs/cgroup/pids                                                             
/sys/fs/cgroup/rdma                                                             
                                                                                
Cgroup v2 mount points:                                                         
                                                                                
                                                                                
Cgroup v1 systemd controller: missing                                           
Cgroup v1 freezer controller: missing                                           
Cgroup v1 clone_children flag: enabled                                          
Cgroup device: missing                                                          
Cgroup sched: enabled                                                           
Cgroup cpu account: enabled                                                     
Cgroup memory controller: enabled                                               
Cgroup cpuset: enabled                                                          
                                                                                
--- Misc ---                                                                    
Veth pair device: enabled, loaded                                               
Macvlan: enabled, not loaded                                                    
Vlan: enabled, not loaded                                                       
Bridges: enabled, not loaded                                                    
[...]

PR #20785, opened 2023-04-08
lxc: set RUNTIME_PATH to the /var/run

rmilecki:
```
The default runtime directory used by LXC is /run which doesn't exist
in OpenWrt. It causes errors like:

Failed to create lock for foo
lxc-create: foo: tools/lxc_create.c: main: 260 Failed to create lxc container

There has been workaround for that in the lxc-auto.init but it requires
installing "lxc-auto" package. Replacing that "ln -s" workaround with
Makefile specifying RUNTIME_PATH define allows using pure "lxc" in
OpenWrt (without the "lxc-auto").
```

### DIFF
--- a/utils/lxc/Makefile
+++ b/utils/lxc/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lxc
 PKG_VERSION:=5.0.2
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://linuxcontainers.org/downloads/lxc/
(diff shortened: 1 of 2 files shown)


rmilecki:
* https://forum.openwrt.org/t/solved-lxc-mvebu-snapshot-21-02/94033
* https://forum.openwrt.org/t/fail-creating-lxc-container-on-mvebu-turris-omnia/106062
* https://forum.openwrt.org/t/request-for-lxc-support-in-mvebu/58767/17
* https://forum.openwrt.org/t/python-numpy-and-or-other-libraries-not-installing/99948/37
* https://github.com/openwrt/packages/issues/15433

1715173329:
please rebase this pr upon the latest git HEAD

rmilecki:
> please rebase this pr upon the latest git HEAD

Sure, I can do that, but still I'd rather have it reviewed ;)

graysky2:
Currently, lxc-auto.init takes care of this, see: https://github.com/openwrt/packages/blob/master/utils/lxc/files/lxc-auto.init#L72-L74

rmilecki:
@graysky2: ah, I didn't realize that as I don't use "lxc-auto". I've updated my commit message to make it clear this change is important for "lxc" users that don't install "lxc-auto".

PR #20786, opened 2023-04-08
acme-acmesh: Add option to run the http challenge on a different port

ngehrsitz:
Maintainer: @tohojo 
Compile tested: N/A
Run tested: FB7520@[r21918-1a47f19080](https://github.com/openwrt/openwrt/commit/1a47f19080)

Description:
This change adds the ability to run the http challenge on a different port. This can be used to generate a certificate without having to stop the services running on the default http port. Redirecting the external traffic from your CA to this custom port has to be configured manually.

### DIFF
--- a/net/acme-acmesh/files/hook.sh
+++ b/net/acme-acmesh/files/hook.sh
@@ -120,6 +120,9 @@ get)
 		fi
 	elif [ "$standalone" = 1 ]; then
 		set -- "$@" --standalone --listen-v6
+		if [ "$http_port" ]; then
+    			set -- "$@" --httpport "$http_port"
+		fi
 	else
 		mkdir -p "$WEBROOT"
 		set -- "$@" --webroot "$WEBROOT"
--- a/net/acme-common/files/acme.init
+++ b/net/acme-common/files/acme.init
@@ -49,6 +49,8 @@ load_options() {
 	export days
 	config_get standalone "$section" standalone 0
 	export standalone
+	config_get http_port "$section" http_port
+	export http_port
 	config_get dns_wait "$section" dns_wait
 	export dns_wait
 


ngehrsitz:
The setup that I am running this in has luci/uhttpd running on port 80 and 443 on the router. Trying to issue certificates using the http challenge now means that the script has to stop uhttpd, issue the cert and the start it again.
But my firewall setup does not expose 80 or 443 externally anyway which is why with this option I have setup a firewall rule that redirects external traffic to port 80 to another port that acme.sh can use exclusively. Thus I don´t have any downtime impacting uhttpd. Also not all certificates that I issue this way are for my uhttpd instance in the first place, but for other network devices meaning that not every certificate renew requires a uhttpd restart.

stokito:
Sorry, maybe I didn't understood something but it looks like you are running the acme in a standalone mode with nc. And the 80 port is always blocked except of when issuing a cert.
So the key issue here is that you need simultaneously uhttpd on 80 port but for LAN and the acme on the 80 port but for WAN only.


Then as a workaround you can:
Start an additional uhttpd instance on port 0.0.0.0:8080 with the  www home folder `/var/run/acme/challenge/`  and it will be used as a webroot.

Create a port forward rule in the `/etc/config/firewall`:

```
config redirect
	option dest_port '8080'
	list proto 'tcp'
	option src_dport '80'
	option dest 'lan'
	option src 'wan'
	option target 'DNAT'
	option dest_ip '192.168.1.1'
	option name 'wan-http-to-local-8080'
```

So all packets from WAN will be redirected to the uhttpd.
This is a simplest way that you may use just today.

ngehrsitz:
Yes, that setup would work with the downside of having a second permanently running uhttpd instance.
I do prefer having the external webserver only accessible while a cert is being issued. Since adding this option would be a very small change I would still like to get this merged.

stokito:
You probably can start it with some pre-hook.
It's memory pages will be shared so it won't be a big overhead.
In fact you'll use it instead of the nc which is as additional dependency btw.

tohojo:
I don't think there's any harm in adding this option even if workarounds exist?

tohojo:
https://github.com/openwrt/packages/pull/27582

PR #20790, opened 2023-04-09
qrencode: Add support for generating PNG files

firewire10000:
I created a forum post on the forums found at the link below asking how to compile the qrencode library with support for PNG files. It's just a simple case of modifying two lines of the makefile and I personally think this should be included by default.

https://forum.openwrt.org/t/how-does-one-compile-qrencode-with-png-support/101027

Compile tested: mvebu, Linksys WRT1900ACS v2, OpenWrt 22.03.2


1715173329:
https://openwrt.org/submitting-patches

please squash your commits and add a valid `signed-off-by` block

BKPepe:
Why this is required to have it on router? What is the size difference with and without it?

firewire10000:
> Why this is required to have it on router? What is the size difference with and without it?

I use qrencode to generate PNG files of my Wireguard interfaces. I can SFTP or SCP into OpenWrt, download the PNG files and then open them up on a computer which can be used for scanning in the configuration to smartphones.

1715173329:
qrencode supports printing result to terminal, and that should be enough.
(though I would assume those size-sensitive devices won't have it installed at all)

firewire10000:
> https://openwrt.org/submitting-patches
> 
> please squash your commits and add a valid `signed-off-by` block

I'm extremel new to GitHub and Git. I'm happy using CLI but I have no idea where to start.

1715173329:
You may use [github cli](https://github.com/cli/cli#installation) for authentication.

```
# login and clone your repo
gh auth login
gh repo clone firewire10000/packages -- -b patch-3

cd packages
# rebase commit
git rebase upstream/master
git reset --soft HEAD^

# update your git settings
git config --local user.name "your name"
git config --local user.email "your@email.com"

# adjust commit title, message etc.
git commit --amend --reset-author -s

# for example:
> qrencode: add support for generating PNG files
>
> <the description for this change, remember to wrap a new line for every 70 characters>
>
> Signed-off-by: your name <your@email.com>
```

1715173329:
well if you don't mind I can do a manual correction for you...
please leave a vaild E-mail address, rather than _users.noreply.github.com_.

firewire10000:
> well if you don't mind I can do a manual correction for you... please leave a vaild E-mail address, rather than _users.noreply.github.com_.

Yh no worries. I've changed my assosiated GitHub email as I was using it with my personal email account so I've created a new one I can use with GitHub. The email is will.j.owen@gmx.com

1715173329:
maintainer @jp-bennett 
Cc @neheb

jp-bennett:
That size difference is a big deal for very small routers. But as OpenWRT doesn't support those old 4 meg devices anymore, this makes more sense.

Personally, I'd like to see this as a menuconfig option, to enable or disable png support. Can default to enabled, as that is the more common scenario. There are a few of us that still do super minimal images, and retaining support for the smaller package for custom images would be great.

neheb:
ACK the menuconfig suggestion.

dangowrt:
> Personally, I'd like to see this as a menuconfig option, to enable or disable png support.

Rather than a build-time option it would be better to have package variants, so also users of the binary distribution can choose to install either `qrencode` or `qrencode-png` with only the latter depending on libpng. We do this already for many packages, and imho it's a much better solution than compile-time options which require the user to build everything from source in order to have a choice.

GeorgeSapkin:
No updates in over two years and PR has merge conflicts. Closing as stale. Feel free to reopen if anything changes.

PR #20791, opened 2023-04-10
Apinger: backport apinger-rrd and the procd conversion to 22.03

hnyman:
@jempatel  @jow- @samm-git 

The LuCI app luci-app-apinger was recently backported&cherry-picked to 22.03 by @jow- , but that LuCI app apparently requires the apinger-rrd that does not exist in 22.03.  
(discussion in https://forum.openwrt.org/t/luci-app-apinger-in-22-03/156835 ) 

Cherry-pick the apinger-rrd commit from master, and also the preceding procd conversion by @jempatel as the rrd commit might build on that for some extent.

Would be great if somebody really using apinger would run-test it in 22.03

Maintainer: @samm-git 
Compile tested: CI testing in this PR
Run tested: -


hnyman:
Thanks for testing.

Issue #12354, opened 2020-05-29
pdns/pdns-recursor: best practice for generating sample config


James-TR:
Maintainer: me
Environment: Any

Description:
Currently I'm manually generating sample config for pdns/pdns-recursor on another of my installations. Normally, this gets generated by calling the compiled binary and had to be patched out since this isn't compatible with a cross-compiler.

I was looking into postinst hooks but not sure if that's the correct place for it or not.

The command that used to be run by the build is below.
```
./pdns_server --no-config --config 2>/dev/null > $wherever
```

Thanks.

neheb:
With gerbera I generate it at runtime: https://github.com/openwrt/packages/blob/master/multimedia/gerbera/files/gerbera.init

Not ideal but works.

James-TR:
I like that idea. I'm thinking of combining with a post-install hook so that it's generated on first install and updated with current version whenever they upgrade.

PR #20797, opened 2023-04-12
less: Ignore .lesshst in sysupgrade backups

pprindeville:
Maintainer: @Zokormazo 
Compile tested: x86_64, generic, HEAD (c3cb2d48da)

Description:

Don't include `.lesshst` files in `sysupgrade -b` output.

Blocked on https://github.com/openwrt/openwrt/pull/12354.

### DIFF
--- a/utils/less/Makefile
+++ b/utils/less/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=less
 PKG_VERSION:=643
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:= @GNU/less \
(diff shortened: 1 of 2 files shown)


GeorgeSapkin:
@pprindeville still relevant? Doesn't look like the base PR is going anywhere.

utils/less/Makefile
```diff
@@ -43,6 +43,8 @@ endef
 define Package/less/install
 	$(INSTALL_DIR) $(1)/usr/libexec
 	$(CP) $(PKG_INSTALL_DIR)/usr/bin/less $(1)/usr/libexec/less-gnu
+	$(INSTALL_DIR) $(1)/lib/upgrade/ignore.d
+	$(INSTALL_DATA) ./files/less.ignore $(1)/lib/upgrade/ignore.d/less
```
openwrt-ai, line 47: As the PR body notes, this feature depends on openwrt/openwrt#12354. For grounding: current openwrt main still has no consumer for `/lib/upgrade/ignore.d/`. `sysupgrade`'s `list_static_conffiles` reads only `/etc/sysupgrade.conf` and `/lib/upgrade/keep.d/*` (package/base-files/files/sbin/sysupgrade), and there is no `ignore.d` handling anywhere else in the tree. So as installed, `ignore.d/less` is inert and `.lesshst` would not actually be excluded from `sysupgrade -b` output.

To @GeorgeSapkin's open question above about whether this is still relevant: #12354 remains unmerged, so functionally this PR is still blocked. Is the intent to keep it open until that lands, or would an approach that works with today's mechanism (e.g. arranging for `.lesshst` to live outside the backed-up config paths) be preferable?

---

openwrt-ai (commented):
Reviewed 1 new commit. One inline note on the `ignore.d` dependency; the commit message accurately describes the diff.

---

PR #20801, opened 2023-04-13
python: Add proper support for pyproject.toml-based builds, update Python packages for new build process

jefferyto:
Maintainer:
* me
* @blocktrron (python-crcmod, python-pycrate)
* @BKPepe (python-astral, python-dns, python-yaml)
* @commodo (pillow, python-evdev)
* @Cynerd (python-intelhex)
* @dangowrt (i2c-tools)
* @dddaniel (python-docutils)
* @erdoukki (fail2ban)
* @ja-pa (onionshare-cli, python-packaging, python-pytest, python-pytest-forked, python-typing-extensions)
* @jmarcet (python-cached-property, python-jsonschema, python-docker, python-dockerpty, python-docopt, python-paramiko, python-pynacl, python-stem)
* @julienmalik (borgbackup, python-exceptiongroup, xxhash)
* @mmtj (python3-pyroute2)
* @nunojpg (nmap)
* @paulo-raca (python-evdev)
* @QuintinHill (python-bleak, python-dbus-fast)
* @valdi74 (python-curl)
* @whooo (python-ubus)
* @ysc3839 (python-cryptodome, python-cryptodomex)

Compile tested: armvirt-32/armvirt-64/malta-be, 2023-04-10 snapshot sdk
Run tested: armvirt-32/armvirt-64/malta-be (qemu), 2023-04-10 snapshot, ran basic tests with pynacl

Description:
The main commit in this series is: **python: Add proper support for pyproject.toml-based builds**

This removes the changes made in https://github.com/openwrt/packages/commit/61f202c0170785addbbc449e4de61cc5886f0833 and adds actual support for pyproject.toml-based (PEP 517) builds of Python packages.

Packages can force the use of the old build process by setting `PYTHON3_PKG_FORCE_DISTUTILS_SETUP:=1`; this should only be a temporary workaround until the package can be updated/fixed to use the new build process.

* * *

The other changes are to prepare other Python packages for the new build process:
* Most of the updated packages are ones that already have a pyproject.toml file and require build dependencies to be added
  * Some of these require updating the package to the latest version, to clear out incompatibilities
* Other updated packages are currently incompatible with the new build process (even though the new build process will fall back to calling setup.py when necessary) and so are configured to continue using the old build process

Some packages are modified in more than one commit; the changes are structured so that all packages should build no matter where the repo is checked out, i.e. before or after the main commit in the series.


commodo (approved):

julienmalik:
for python-exceptiongroup, it looks like my patch that added a dummy setup.py can be removed, now that it builds natively with flit. Nice work !

jefferyto:
Thanks @julienmalik - I added a commit to remove the patch.

BKPepe (approved):

nunojpg:
@jefferyto I've bumped nmap. Also please mind that Python is not required for most nmap packages, only for ndiff. Not sure if any of that has impacts on your changes.

jefferyto:
Thanks @nunojpg - I have rebased on master. The [changes](https://github.com/openwrt/packages/pull/20801/commits/a7bff5075c71398e54791a4634e03e53ecbdea5b) for nmap in this PR are minimal, basically continuing to use the old Python build process (for ndiff).

The CI is now failing due to dirty patches for nmap though; perhaps the patches were not refreshed when you updated the package?

nunojpg:
Now refreshed. Thanks

jefferyto:
Rebased on master (amended for removal of `AUTORELEASE`).

jefferyto:
@BKPepe why did you cherry pick this into openwrt-22.03? This was a very big change and not meant for a stable branch.

Also would you consider adding your own Signed-off-by if/when you cherry pick commits authored by others? (I don't add a Signed-off-by when I cherry pick my own commits because I don't think adding a second sign-off from the same person adds anything useful.)

BKPepe:
Indeed, it was a really big change and I agree with that. The most reason why I did it was to implement [PEP 517](https://peps.python.org/pep-0517/) within OpenWrt build system for 22.03 version. I wanted to have some basic support from other PR, but in the end, I backported it all since it did not end up very well and several packages were failing.

Actually, some packages already switched to use ``pyproject.toml`` instead of ``setup.py`` and that could not be used in OpenWrt 22.03. That was the problem at least for me, when I wanted to the latest version of several Python packages across OpenWrt versions or update just one one, and that one depend on something else. Usually, before doing it, I am asking. This time, I did not ask. My fault.  However, all packages should be fixed and there are no faillogs at least on buildbots, so it is safe to use. 

The other advantage that I can think of is dropping entirely ``host-pip-requirements`` folder and instead of playing in downstream playground, they can focus or work more with us in upstream to have it working. 

Now at least we do have same build structure together same accross all stable branches and most likely even same package versions that helps for me and ensure that it will work with all stable versions, so no breakage will appear.

Regarding SoB, okay. No problem to do that. Yeah, adding two same SoBs for single person is weird.

PR #20802, opened 2023-04-13
openssh: preserve authorized_keys

hgl:
Maintainer: @tripolar
Compile tested: x86-64 22.03.3
Run tested: x86-64 22.03.3

Description:
The root user is usually the user that clients ssh into with, so in most cases its authorized_keys determines what clients are allowed to ssh into this device. Without preserving this file, they could potentially be locked out after upgrading.

### DIFF
--- a/net/openssh/Makefile
+++ b/net/openssh/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openssh
 PKG_VERSION:=9.3p1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/ \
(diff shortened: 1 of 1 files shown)


hgl:
Forgot to mention, the dropbear package preserves this file:

https://github.com/openwrt/openwrt/blob/877ec78e2354e39a363bf257b7562a36d4b18e4a/package/base-files/Makefile#L52

I think it shouldn't be a terrible idea that openssh follows suit.

pprindeville (approved):
LGTM

SibrenVasse (approved):

hgl:
Any reason this PR is only approved but not merged?

BKPepe:
Yes, it was not approved by the maintainer.

hgl:
@BKPepe I actually manually tested this, OP updated.

It seems @tripolar is dormant. Previous openssh PRs were all merged by other maintainers.

pprindeville:
@neheb Do you want to do the merge?

neheb:
Hmmm how does dropbear do this?

hgl:
For dropbear it's `/etc/dropbear/authorized_keys`, and since the whole `/etc/dropbear` directory is kept, it works automatically.

neheb:
Guess something like that can't be done with openssh

hgl:
Don't think so, openssh is multi-user

PR #20803, opened 2023-04-13
cache-domains: Fixed config files being outside of jail

G-M0N3Y-2503:
Maintainer: me @G-M0N3Y-2503
Compile tested: x86_x64, QEMU/KVM, master
Run tested: x86_x64, QEMU/KVM, master

Description:
This fix dnsmasq failing to startup once configured since the cache-domains config files are not included in the jail that dnsmasq runs inside.
Also, I've reduces the writes by symlinking each file to the dnsmasq directory instead of writing them all to one big file.

### DIFF
--- a/utils/cache-domains/Makefile
+++ b/utils/cache-domains/Makefile
@@ -1,8 +1,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cache-domains
-PKG_VERSION:=2.3.1
-PKG_RELEASE:=2
+PKG_VERSION:=2.4.0
+PKG_RELEASE:=1
 
 PKG_MAINTAINER:=Gerard Ryan <G.M0N3Y.2503@gmail.com>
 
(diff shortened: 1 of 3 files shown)


.github/workflows/entrypoint.sh
```diff
@@ -13,7 +13,7 @@ mkdir -p /var/lock/
 
 opkg update
 
-[ -n "${CI_HELPER:=''}" ] || CI_HELPER="/ci/.github/workflows/ci_helpers.sh"
+CI_HELPER="${CI_HELPER:="/ci/.github/workflows/ci_helpers.sh"}"
```
feckert, line 16: Is this intended?

G-M0N3Y-2503: I probably should have left a seperate commit for this, but yes.

if `CI_HELPER` was null or empty it would be set to the literal 2 char string `''`.  
so
```bash
[ -n "''" ]
```
would always be greater than zero chars long (2) and never call the command after the `||`

G-M0N3Y-2503: Removed this line to alleviate any contention since it was a benign bug anyway

G-M0N3Y-2503:
I ran the tests again locally on https://github.com/openwrt/openwrt/commit/36746893ac690796bf7aad00a793ff4bf739d066 and this PR still seems good to me, was there anything else holding back this PR?

p.s I forgot that the tests don't work in the CI because there is no ubus

PR #20808, opened 2023-04-14
openssh: add respawn and reloading via signal

erikrk:
Configure the openssh server to respawn. Reload by sending SIGHUP

Maintainer: @SibrenVasse 

Description:
Improvements to the OpenSSH init script. Respawning is added so that the OpenSSH server gets restarted in case it crashes. Reloading is implemented by sending SIGHUP to the OpenSSH server process. Before it would not re-read for example /etc/ssh/sshd_config when "/etc/init.d/sshd reload" was run.


SibrenVasse (approved):
Not the maintainer, but LGTM. Changes work on my devices as advertised.

PR #20810, opened 2023-04-15
python-hatchling: Update to 1.14.0, add new host-only dependencies

jefferyto:
Maintainer: me
Compile tested: armvirt-32, 2023-04-10 snapshot sdk
Run tested: N/A

Description:
hatchling 1.14.0 added trove-classifiers as a dependency; trove-classifiers in turn depends on calver.

### DIFF
--- a/lang/python/python-calver/Makefile
+++ b/lang/python/python-calver/Makefile
@@ -1,0 +1,48 @@
+#
+# Copyright (C) 2023 Jeffery To
+#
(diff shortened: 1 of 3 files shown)


jefferyto:
CI is failing at the Test via Docker container step:

```
tar: can't open '/ci/*.ipk': No such file or directory
tar: invalid magic
tar: short read
```

This because all three packages are host-only packages; they do not produce any target packages (.ipk files).

Issue #20807, opened 2023-04-14
softethervpn5: vpnclient page faults after 21.x -> 22.x upgrade


tew42:
Maintainer: @Andy2244
Environment: ath79 - TP-Link Archer A7 - OpenWRT 22.03.3

Description:
Running a SoftetherVPN client on my router worked well for me on 21.x (most recently 21.02.4). I upgraded to 22.03.3 today, and encountered a loop of VPN start -> access violation -> reconnect -> access violation, as documented below. 

I tried both with importing the old config, and with setting up a new configuration from scratch; same behavior.

Happy to provide additional info, just not quite sure where/how... Thanks!

Log excerpt:
```
15:03:05 2023 kern.info kernel: [  xxx] do_page_fault(): sending SIGSEGV to vpnclient for invalid write access to 00000000
15:03:05 2023 kern.info kernel: [  xxx] epc = 77dbf96c in libc.so[77d8e000+a9000]
15:03:05 2023 kern.info kernel: [  xxx] ra  = 77dbf908 in libc.so[77d8e000+a9000]
15:03:05 2023 daemon.notice netifd: Network device 'vpn_xxx' link is down
15:03:05 2023 daemon.notice netifd: Interface 'VPN' has link connectivity loss
15:03:05 2023 daemon.notice netifd: VPN (4064): udhcpc: SIOCGIFINDEX: No such device
15:03:05 2023 daemon.notice netifd: VPN (4064): udhcpc: received SIGTERM
15:03:05 2023 daemon.notice netifd: VPN (4064): udhcpc: unicasting a release of 192.168.xxx.xxx to 192.168.xxx.xxx
15:03:05 2023 daemon.notice netifd: VPN (4064): udhcpc: sending release
15:03:05 2023 daemon.notice netifd: VPN (4064): udhcpc: can't bind to interface vpn_xxx: No such device
15:03:05 2023 daemon.notice netifd: VPN (4064): udhcpc: bindtodevice: No such device
15:03:05 2023 daemon.notice netifd: VPN (4064): udhcpc: entering released state
15:03:05 2023 daemon.notice netifd: VPN (4064): Command failed: ubus call network.interface notify_proto { "action": 0, "link-up": false, "keep": false, "interface": "VPN" } (Permission denied)
15:03:05 2023 daemon.notice netifd: Interface 'VPN' is now down
15:03:05 2023 daemon.notice netifd: Interface 'VPN' is disabled
[...]

PR #20813, opened 2023-04-15
Update softethervpn5 libraries to fix 22.03 branch kernel incompatibility

tew42:
Maintainer: ??? - since @Andy2244 apparently dropped this, I'm tagging @neheb as the other person with semi-recent commits to the module
Compile tested: compiled for ath79 - Archer c7 v5 - OpenWRT 22.03.4 using the SDK in an Ubuntu 22.04 WSL environment 
Run tested: ath79 - Archer c7 v5 - OpenWRT 22.03.4

Description:
As reported in #20807 the softethervpn5 libraries seem to have issues with keeping an open connection in 22.03 series builds, breaking module functionality. The root cause is not clear to me, though I'm guessing it's related to the newer kernel over 21.x. Regardless, building the package & libraries from a more recent upstream source code solved the issue. 

Note 1: The build should definitely be double-checked by someone with a complete source-build environment, because I initially encountered some issues with find_package(OpenSSL) only locating the toolchain-included LibreSSL version (which fails internal functionality checks), rather than the required libopenssl version. It resolved itself after some cleaning and restarting, but may also be related to WSL and/or using SDK rather than build source.

Note 2: This patch should also be considered for master branch, but I can't really test on snapshot builds right now.


tew42:
sorry, branch issue, please ignore

PR #20817, opened 2023-04-15
lighttpd: fix package DEPENDS syntax

gstrauss:
Maintainer: @flyn-org

Description:
* fix package DEPENDS syntax  (follow-up from #20799)
* adjust packages for built-in modules (save some disk space)

@1715173329 I believe this addresses issues uncovered in #20799 and is compatible with 21.02 and 22.03 branches.
Reviewing tmp/.packageinfo looks better to me.

### DIFF
--- a/net/lighttpd/Makefile
+++ b/net/lighttpd/Makefile
@@ -60,17 +60,17 @@ PKG_BUILD_DEPENDS:= \
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/meson.mk
 
-# choose crypto lib for lighttpd to use for crypto algorithms
+# choose crypto lib for lighttpd to use for crypto algorithms (default: nettle)
 # (separate from lighttpd TLS modules, which are each standalone)
-ifdef CONFIG_LIGHTTPD_CRYPTOLIB_NONE)
-  cryptolib=
-else ifdef CONFIG_LIGHTTPD_CRYPTOLIB_NETTLE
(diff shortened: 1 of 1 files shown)


gstrauss:
Hmmm.  This does not seem to build all the packages I expected it to build.

net/lighttpd/Makefile
```diff
@@ -57,6 +57,11 @@ PKG_BUILD_DEPENDS:= \
   LIGHTTPD_CRYPTOLIB_MBEDTLS:mbedtls \
   LIGHTTPD_CRYPTOLIB_WOLFSSL:wolfssl
 
+cryptolibdep= \
```
1715173329, line 60: I think cryptolib and cryptolibdep can be simply removed,
crypto dependency is defined in lighttpd already, there's no need to add it to submodules again

gstrauss: I'd like to keep the dependencies declared unless you think that will cause a problem.

It would be minimalist to remove the declared dependencies due to the transitive dependency on the crypto library via dependency on lighttpd which depends on the crypto library.

However, each module which depends on crypto libs does so directly.  The lighttpd base depends on crypto libs for cryptographic randomness in lighttpd `src/rand.c`, but other modules depending on crypto libs do so for digest functions such as MD5, SHA1, SHA256, SHA512, HMAC.

That said, I will adjust the patch to check more directly for dependency on nettle for build flag `-Dwith_nettle=true`

PR #20839, opened 2023-04-18
ruby: build hostpkg with rpath

micmac1:
Compile currently fails in case host system doesn't provide a fitting libyaml, which is unnecessary because ruby already depends yaml/host, so the lib is there, ruby hostpkg just can't find it.

This commit adds an rpath to the host LDFLAGS to resolve this. An alternative would be to build a static ruby hostpkg, but why waste the space.

Maintainer: @luizluca 
Compile tested: fresh SDK, ath97
Run tested: N/A, build fix only

Description:
Fix build on some hosts, OpenWrt build bots also fail to build ruby currently

### DIFF
--- a/lang/ruby/Makefile
+++ b/lang/ruby/Makefile
@@ -12,7 +12,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ruby
 PKG_VERSION:=3.2.2
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 # First two numbes
 PKG_ABI_VERSION:=$(subst $(space),.,$(wordlist 1, 2, $(subst .,$(space),$(PKG_VERSION))))
@@ -85,6 +85,10 @@ HOST_CONFIGURE_ARGS += --disable-jit-support
 # any output
 HOST_LDFLAGS += \
 	$(if $(CONFIG_HOST_OS_MACOS),-Wl$(comma)-w)
+
+# hostpkg ruby needs to find the hostpkg libyaml, otherwise it won't load if
+# host system doesn't provide an alternative
+HOST_LDFLAGS += -Wl,-rpath,$(STAGING_DIR_HOSTPKG)/lib
 
 TARGET_LDFLAGS += -L$(PKG_BUILD_DIR)
 


luizluca:
The issue seems that the issue is psych.so not finding libyaml-0.so.2.

```
.ext/x86_64-linux-gnu/psych.so:
        linux-vdso.so.1 (0x00007ffc8c568000)
        libyaml-0.so.2 => not found
        libm.so.6 => /lib64/libm.so.6 (0x00007f78fcb44000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f78fc949000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f78fcc4f000)
```

During that ext config step, conftest.c compiles an executable that uses the absolute path to the libyaml found by checking gcc -L options. However, when it uses the same -L for a shared library (psych.so), it will not include the absolute path. It seems to be by-design, so libraries dependencies will be resolved at runtime. The `-Wl,-rpath,$(STAGING_DIR_HOSTPKG)/lib` will add the staging dir library path to the list of paths psych.so will looks for libraries. It will workaround the issue but it would not survive a parent path rename. If someone renames/copies the current openwrt directory, that path will break. I believe a dynamic path solution using LD_LIBRARY_PATH would be more robust.

luizluca:
This might do:

```patch
--- a/lang/ruby/Makefile
+++ b/lang/ruby/Makefile
@@ -99,6 +99,9 @@ MAKE_FLAGS += \
        DESTDIR="$(PKG_INSTALL_DIR)" \
        SHELL="/bin/bash"
 
+HOST_MAKE_FLAGS += LD_LIBRARY_PATH=$(STAGING_DIR_HOSTPKG)/lib
+MAKE_FLAGS += LD_LIBRARY_PATH=$(STAGING_DIR_HOSTPKG)/lib
+
 define Build/InstallDev
        ( cd $(PKG_INSTALL_DIR); $(TAR) -cf - \
                . \
```

Let me know if you think the rpath solution is better.

micmac1:
Am 20. April 2023 21:49:25 UTC schrieb Luiz Angelo Daros de Luca ***@***.***>:
>This might do:
>
>```patch
>--- a/lang/ruby/Makefile
>+++ b/lang/ruby/Makefile
>@@ -99,6 +99,9 @@ MAKE_FLAGS += \
>        DESTDIR="$(PKG_INSTALL_DIR)" \
>        SHELL="/bin/bash"
> 
>+HOST_MAKE_FLAGS += LD_LIBRARY_PATH=$(STAGING_DIR_HOSTPKG)/lib
>+MAKE_FLAGS += LD_LIBRARY_PATH=$(STAGING_DIR_HOSTPKG)/lib
>+
> define Build/InstallDev
>        ( cd $(PKG_INSTALL_DIR); $(TAR) -cf - \
>                . \
>```
>
>Let me know if you think the rpath solution is better.
>

Why set MAKE_FLAGS? There won't be a problem on target, as the lib will be in standard path on the target device.

If your idea with LD_LIBRARY_PATH in HOST_MAKE_FLAGS works for the hostpkg then fine. You pick.

Have a good day!

neheb:
Why not compile the libyaml hostpkg as static?

luizluca:
> Why not compile the libyaml hostpkg as static?

Yes, it would be a valid solution. When libyaml source core was bundled in ruby, it was building it together with psych.so, not even using it as a lib.a. However, since it started to use the library provided by the system, it will use it as a runtime dependency, without a build option to change that behavior. I don't think that hacking the build script is a good solution.

It seems that now ruby can again build with ---with-static-linked-ext, statically linking all ext into ruby. That way, the linker will use the stagging path for libyaml instead of letting it be resolved at runtime. I seems to be a cleaner solution than hacking the search path, either using rpath or LD_LIBRARY_PATH. I'll create a PR for that.

> edit: seems [a951e16](https://github.com/openwrt/packages/commit/a951e16646f284bfadf6b458d30ba069eb7db006) was incomplete.

Sorry, @neheb . I couldn't find what I missed in the yaml change. It would, indeed, require `HOST_CONFIGURE_ARGS += --with-pic` for statically linking that into another library. Is that it? Anyway, for a statically built ruby, it would not matter.

neheb:
I’ve disabled shared libraries on most host packages specifically to avoid problems like these.

1715173329:
i would vote to static host-build (for both libyaml and ruby)

PR #20845, opened 2023-04-19
CI: Add local feed for CI-built packages

jefferyto:
Maintainer: ping @aparcar @neheb @BKPepe @1715173329 
Compile tested: PRs in my fork (with dummy packages to test the Test via Docker step: https://github.com/jefferyto/openwrt-packages/actions/runs/4739914724/jobs/8415178055?pr=9)
Run tested: N/A

Description:
To test each package, the CI-built target package (ipk) file is installed, but currently the target package's dependencies are installed from the standard opkg feeds.

There are cases when the CI-built target packages should be installed/tested together:

* If a pull request contains several new packages that depend on each other, the test step will fail as the new dependencies cannot be found in the current packages feed.

* If a pull request upgrades a source package that builds several target packages that depend on each other, the test step may fail due to the version/ABI mismatch between a newer target package and the older dependencies installed from the packages feed.

This sets up a local feed for the CI-built packages so that dependencies are also installed from the same set of packages.

### DIFF
--- a/.github/workflows/entrypoint.sh
+++ b/.github/workflows/entrypoint.sh
@@ -2,7 +2,12 @@
 
 # not enabling `errtrace` and `pipefail` since those are bash specific
 set -o errexit # failing commands causes script to fail
-set -o nounset # undefined variables causes script to fail 
+set -o nounset # undefined variables causes script to fail
+
+echo "src/gz packages_ci file:///ci" >> /etc/opkg/distfeeds.conf
+
+FINGERPRINT="$(usign -F -p /ci/packages_ci.pub)"
+cp /ci/packages_ci.pub "/etc/opkg/keys/$FINGERPRINT"
 
 mkdir -p /var/lock/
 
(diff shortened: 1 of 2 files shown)


jefferyto:
#20781 is an example of a PR with a set of new packages that depend on each other; the CI would have passed if the CI-built packages were available to opkg to fulfill package dependencies.

aparcar:
Excellent work

.github/workflows/entrypoint.sh
```diff
@@ -2,7 +2,12 @@
 
 # not enabling `errtrace` and `pipefail` since those are bash specific
 set -o errexit # failing commands causes script to fail
-set -o nounset # undefined variables causes script to fail 
+set -o nounset # undefined variables causes script to fail
+
+echo "src/gz packages_ci file:///ci" >> /etc/opkg/distfeeds.conf
+
+FINGERPRINT="$(usign -F -p /ci/packages_ci.pub)"
```
aparcar, line 9: Is usign found in path?

aparcar (approved):

1715173329 (approved):
Nice work!

PR #20849, opened 2023-04-20
nginx: update to 1.24.0

hgl:
Maintainer: @Ansuel 
Compile tested:x86-64 22.03.3
Run tested: x86-64 22.03.3
Description:

I know you're still considering the lua support, but I've successfully compiled and ran the new version in my local device with LuCI and with only the brotli module enabled. Only sending the PR to allow CI to test if it can compile successfully with all modules enabled. 

---

Commit message:

With lua module dropped, for multiple reasons:

1. The lua module doesn't officially support the latest nginx stable
2. The lua module warns if it doesn't live inside the OpenResty flavored nginx
3. LuCI is moving away from lua to ucode


tiagogaspar8:
Great work, I was meaning to do that but haven't found the time.
Can you explain to me or point me to somewhere about the lua support you were mentioning? I'm out of the loop

hgl:
https://github.com/openwrt/packages/pull/19881#issuecomment-1320945050

@Ansuel is working a new PR that uses dynamic modules though.

tiagogaspar8:
Maybe you can close this now @hgl

PR #20850, opened 2023-04-20
softethervpn5: Update libraries to fix 22.03 branch kernel incompatibility

tew42:
Maintainer: ???
Compile tested: compiled for ath79 - Archer a7 v5 - OpenWRT 22.03.4 using the SDK in an Ubuntu 22.04 WSL environment
Run tested: ath79 - Archer a7 v5 - OpenWRT 22.03.4

Description:
As reported in https://github.com/openwrt/packages/issues/20807 the softethervpn5 libraries seem to have issues with keeping an open connection in 22.03 series builds, breaking module functionality. The root cause is not clear to me, though I'm guessing it's related to the newer kernel over 21.x. Regardless, building the package & libraries from a more recent upstream source code solved the issue.

Note 1: Seems like simple file editing on the website is insufficient for this commit, so re-started the whole thing with a proper git environment... Maybe now this will go better than my first attempts - sorry for the multiple requests.

Note 2: This patch should also be considered for master branch, but I can't really test on snapshot builds right now.


tew42:
Closed in favor of 23.x branch.

PR #20855, opened 2023-04-21
mwan3: fix mwan3 doesn't detect the absence of the internet connection if an interface goes down and then up

anyuta1166:
After an interface goes down and then up, mwan3 doesn't detect the absence of the internet connection if the interface initial state is online. This commit fixes the issue.

Maintainer: @feckert

Description:

On boot everything works fine:

1. Initial state of the interface is set to online.
2. Router boots up
3. Interface goes up, the state is set to online
4. mwan3track is launched, initial score is set to $((down+up)) upon mwan3track initialization.
5. mwan3track performs ping tests. If there is no internet connection, the score decreases and the interface becomes disconnected.

But if later the interface goes down:

1. Interface goes down. disconnected() function is called and score is reset to 0.
2. Interface goes up, the state is set to online
3. mwan3track performs ping tests. But if there is no internet connection, nothing is done because the score is already 0. Interface stays in online state.

Fix this issue by setting the score to it's initial value before calling connected() from firstconnect()

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.11.4
+PKG_VERSION:=2.11.6
 PKG_RELEASE:=1
 PKG_MAINTAINER:=Florian Eckert <fe@dev.tdt.de>, \
 		Aaron Goodman <aaronjg@alumni.stanford.edu>
--- a/net/mwan3/files/usr/sbin/mwan3track
+++ b/net/mwan3/files/usr/sbin/mwan3track
@@ -184,6 +184,7 @@ firstconnect() {
 	if [ "$STATUS" = "offline" ]; then
 		disconnected 1
 	else
+		score=$((down+up))
 		connected 1
 	fi
 }


feckert:
Looks good. Can you increase the PKG_VERSION by one here too?

anyuta1166:
Oh, I didn't see your PR, I've encountered this issue independently of you, just a few days ago.
But your PR is slightly different. I think that the resetting the score inside "firstconnect", just before calling "connected 1", is more proper than resetting it before "firstconnect".

anyuta1166:
> Could you explain to me why it works at all? I honestly have no clue.

Because it is called from main()? This already works in disconnected() (score=0).

feckert:
@gentoo-root You're absolutely right, it's a mess! But it has simply grown in history. I haven't dared to do it yet, because the next OpenWRT release will be tagged soon and I don't want to change everything beforehand. We should change and  rebuild the whole thing later. It would also be better if we used a state machine to track the states of the tracker.

Also, I'm still using iptables and but I want to switch to nftables for a while now, as openwrt now uses this as the default firewall solution (fw4). I have also been thinking about using ebpf to mark the packages. I would like to get ride of the whole shell scripting!

feckert:
Superseded by https://github.com/openwrt/packages/commit/1f6bd672fecd1ffc9f6b1fea152edc62bcdca026 and https://github.com/openwrt/packages/commit/5f0461be8b149aeb5732481d38c9d69650996f8c

Neustradamus:
Replaced by PR:
- https://github.com/openwrt/packages/pull/20924

PR #20856, opened 2023-04-21
qemu: update to 8.0.0

vooon:
Signed-off-by: Vladimir Ermakov [vooon341@gmail.com](mailto:vooon341@gmail.com)

Maintainer: @yousong
Compile tested: x86_64, OpenWrt master (https://github.com/openwrt/openwrt/commit/1416b9bbe9d3455c7ab64543a65236fea213a978)
Run tested: x86_64 host, KVM for qemu-x86_64-softmmu + Bios VM.

Description:
- Version updated to 8.0.0
- Added option to use ZStd compression

### DIFF
--- a/utils/qemu/Makefile
+++ b/utils/qemu/Makefile
@@ -9,10 +9,10 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=qemu
-PKG_VERSION:=7.2.0
+PKG_VERSION:=8.0.0
 PKG_RELEASE:=1
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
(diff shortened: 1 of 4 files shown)


vooon:
@yousong , @neheb could you please check why i386 CI failed?

I haven't testes qcow2 with zstd compression, but it's likely to work. ldd shows libzstd.so.
Also i've tested on ubuntu x86_64 image - works well.

yousong:
> @yousong , @neheb could you please check why i386 CI failed?
> 
> I haven't testes qcow2 with zstd compression, but it's likely to work. ldd shows libzstd.so. Also i've tested on ubuntu x86_64 image - works well.

Likely caused by the fact that qemu is not enabled for i386 and as such no .ipk was generated and the test via docker container failed because of that.

The pull request LGTM ;)

Issue #20835, opened 2023-04-18
strongswan: use --disable-defaults?


hgl:
Maintainer: @pprindeville @Thermi 
Environment: (put here arch, model, OpenWrt version)

Description:

The errors from https://github.com/openwrt/packages/pull/19128#issuecomment-1207487710 can happen again when strongSwan add new plugins that are enabled by default in a new release. I wonder if it would be a bad idea if we add `--disable-defaults` to the configure args? This will make strongSwan only use explicitly enabled packages.

hgl:
Also, that PR unconditionally enabled `mgf1`, but according to ecdsa:

> It's basically only needed for RSA-PSS with the gmp plugin and for the deprecated bliss and ntru plugins, so if none of these are enabled, it's not needed

Can we remove it from the configure args so it can be conditionally enabled?

Thermi:
here on OpenWRT using gmp itself is a valid use case, so we should support that

pprindeville:
> Since the gmp plugin exists, I guess it's supported? I'm only suggesting that it doesn't need to be unconditionally enabled.

Where is the mgf1 patch?

PR #20887, opened 2023-04-26
Strongswan use disable-defaults

hgl:
Compile tested: x86-64 master
Run tested: x86-64 master

Description:

Without it, every time strongSwan introduces a new plugin that is enabled
by default, we will get a warning from charon for missing the plugin
files. Using --disable-defaults tightens up the plugin list where only
explicitly enabled plugins are compiled.

Reason for removed flags:

--disable-scripts --disable-fast
disabled by default now

--disable-static
is the default

--with-systemdsystemunitdir=no
systemd won't be enabled, useless flag


pprindeville (commented):
Did we want line 313 to go away?

hgl:
> Did we want line 313 to go away?
Not sure I understand. Line 313 seems to be strongswan-isakmp’s dep mfg1? Do you want it removed?

pprindeville:
I thought you said in a separate thread that mgf1 was only used in PKCS#2.1 (RSA-PSS) in certificate signing and therefore only relevant when using X.509 certificates for authentication with RSA keys?

If you're doing purely ECC, for instance, there's no dependency on it...

hgl:
Oh, I only concerned with not enabling mgf1 unconditionally, didn't mean to touch the presets, since I think the dependency list is subjective.

> Or do NTRU, BLISS, and GMP have dependencies on XOF_MGF1_SHA*?

If I understand it correctly, mgf1's dependency is like `rsa-pss && (gmp || bliss || ntru)`. I can remove it from the presets if you like.

hgl:
@pprindeville done.

Quick question: any reason strongswan-full listed mods explicitly instead of looping over PKG_MOD_AVAILABLE?

Thermi:
I believe it is not a good idea to disable the defaults because it only has potential impacts at runtime. AFAIR some of the dependencies within the strongSwan code base are not modeled as compile time dependencies.

I think a good solution would be to package all currently packaged plugins as they are, and all that are new or that will be added by defaults, until it causes problems, go into a defaults package.

In your opinions, would that be a reasonable solution?

hgl:
> AFAIR some of the dependencies within the strongSwan code base are not modeled as compile time dependencies.

Sorry, not sure I understand. Do you mean if defaults are disabled, strongSwan would miss some hidden dependencies? Could you provide one example?

> I think a good solution would be to package all currently packaged plugins as they are, and all that are new or that will be added by defaults, until it causes problems, go into a defaults package.

I guess this package is created by collecting the left over files in the plugin directories after all other plugin packages have been created? I'm not sure this is a good idea. The existence of this package seem to suggest that some plugins, which should be their own plugin packages (unless it's a hidden dependency like you mentioned, which I'd like to know if there is any), are missed to be packages.

And it seems a user should always install this defaults package, other wise charon would complain in runtime since they're enabled by default. This creates a window, before the default plugins are turned into their own packages, where such plugins are unconditionally enabled even if they're not needed. For example, if a user has openssl enabled, but strongSwan introduces a new default crypto plugin that is covered by openssl, enabling this crypto plugin is then redundant.

hgl:
And verdict? @pprindeville @Thermi 

Would appreciate that we don't leave this in limbo.

pprindeville:
> And verdict? @pprindeville @Thermi
> 
> Would appreciate that we don't leave this in limbo.

I'm waiting on consensus.  And a rebase.

hgl:
> I'm waiting on consensus. And a rebase.

I've run charon with disable-defaults enabled for years without issues, but I must admit that my use cases are pretty limited, would be great if @Thermi could elaborate on "potential impacts at runtime".

Rebased.

hgl:
@pprindeville Thoughts? It'd be a waste if the PR becomes stale again.

pprindeville:
@Thermi Are we at an impasse?

hgl:
Do you mind if I ping @tobiasbrunner to comment on if it's ok for downstream to use disable-defaults? Any potential risks?

pprindeville:
> Do you mind if I ping @tobiasbrunner to comment on if it's ok for downstream to use disable-defaults? Any potential risks?

I'll defer to @Thermi on this PR.

Thermi:
I think it'd be better if we, by default, packaged everything in the default packages, except of course the plugins/files that already have their own packages. Then at least nobody ends up with a broken installation in case we miss something, but just a slightly larger image.

hgl:
With all due respect, the current approach *fails* to package everything whenever a new strongswan version introduces a new default package, as can be seen from https://github.com/openwrt/packages/pull/19128#issuecomment-1207487710.  If we keep the current approach, this can happen again.

[I proposed `disable-defaults` exactly for this reason](https://github.com/openwrt/packages/issues/20835).

With the way strongswan is packaged is here, the only way to package everything is that we explicitly list all plugins. This has always been the goal, but without `disable-defaults` and without explicitly listing them, we still fail to package them. `disable-defaults` just allow us to upgrade to a new version of strongswan without the fear of introducing missing plugin files error, so it's really orthogonal to your concern, imho.

Let me know what you think.

Thermi:
The problem with --disable-defaults is that it disables all default modules. So we would have to make sure that for every new release, we explicitely add an --enable-x argument to the configure commands (and then of course also package it, but that is already a problem today). So with --disable-defaults the problem gets worse, not better.

hgl:
> So we would have to make sure that for every new release, we explicitely add an --enable-x argument to the configure commands

Without --disable-defaults, we already have to make sure we explicitly add x to `PKG_MOD_AVAILABLE`, `BuildPlugin` etc, otherwise we get runtime file missing error as we talked about. Are you suggesting that we rely on these errors to be aware of what plugins to add to the Makefile?

The problem is that:

1. It's very tempting to simply add --enable-x globally as we did with mgf1 (which is no longer a default module btw).
2. We aren't currently maintaining a 1:1 mapping between strongswan defaults and what's globally enabled here (e.g., we don't globally enable drbg which is a default), so it's not necessary to always do --enable-x

With --disable-defaults, if we forget to add a plugin to the Makefile, we don't get runtime error, and if someone needs the plugin, it's always one PR away. It's more like a safe guard, so I don't understand how "the problem gets worse"?

Thermi:
No. I suggest we install all into the default package first. Then we move out the stuff we want to package separately.

hgl:
> I suggest we install all into the default package first

1. As I mentioned, clearly we currently don't do that, drbg is not in the default package but is a default module. Are you saying it's a mistake?
2. [We explicitly list the plugins in the default package](https://github.com/openwrt/packages/blob/4cb6cd2d18e2d6319226e91a57c635c2d27aba53/net/strongswan/Makefile#L259), so I don't understand your "so we would have to make sure that for every new release, we explicitely add an --enable-x argument to the configure commands" argument. Don't we already do that?

Thermi:
No, I propose we start doing that. Currently, without --disable-defaults, any plugin the strongSwan devs enable by default are also built by default. I suppose that is because they are also required for a basic working installation. I propose that we only split out the stuff from the standard/default package of strongSwan that we know is not required for a working installation.

GeorgeSapkin:
No updates in over two years, not following guidelines and PR has merge conflicts. Closing as stale. Feel free to reopen if anything changes.

PR #20900, opened 2023-04-28
mwan3: fix addition of routes to mwan3_connected ipset

### LINKED ISSUE #20779: mwan3: Router becomes unreachable from LAN is all WANs are down at the boot time
Maintainer: @feckert
Environment: ipq806x, Netgear R7800, OpenWrt from master (https://github.com/openwrt/openwrt/commit/0dc5fc8fa53be55c3adc59055e7f6f150a5ab1ad).

Description:

1. Make sure that all WANs are offline (in fact, I performed the debugging with only one WAN, but the issue exists with multiple WANs as well).
2. Turn on the router.
3. After it boots, you can ping the router and SSH into it from LAN.
4. After a while, exactly when mwan3 moves the interface to offline (`mwan3track[1532]: Interface wan (eth0.2) is offline`), the issue appears.
5. No new connections are accepted at this point, but the established connections stay alive. For example, if you have active SSH sessions or ping in progress, it will continue, but new SSH connections will fail (the client never receives a SYN-ACK), and new instances of ping don't receive any replies.

I debugged the issue and found what's wrong at startup. At step 3, the mwan3_connected_ipv4 ipset is not populated, it contains only 224.0.0.0/3. At the same time, the mwan3_ifaces_in iptables chain is empty, so no one assigns fwmarks, and the "unreachable" rules are not triggered.

When mwan3 detects that the WAN interface is offline (step 4), it creates the mwan3_ifaces_in and mwan3_iface_in_wan chains, but it doesn't populate the mwan3_connected_ipv4 ipset, and all ingress packets from LAN belonging to new connections (only new connections hit mwan3_ifaces_in) start being dropped by the ip rule `2062:	from all fwmark 0x3e00/0x3f00 unreachable`.

During normal operation, the mwan3_connected_ipv4 ipset contains the IP address of the router (along with the LAN subnet), so new packets don't get marked to be dropped.

I believe, this should be fixed so that the mwan3_connected_ipv4 ipset is populated from the start, and the mwan3_ifaces_in* chains should also exist from the start.

The issue only happens on boot. If you manage to SSH into the router during that small time frame when it's still reachable, and restart mwan3, it will stay reachable, even though the WAN is down.

anyuta1166:
Addition of routes to mwan3_connected ipset was broken in 2.11.0. The commit fixes this issue.

Maintainer: @feckert 

Description:

Addition of routes to mwan3_connected ipset was broken in 2.11.0.

The ipset name was changed from mwan3_connected_v4/6 to mwan3_connected_ipv4/6, but this change was not reflected in mwan3rtmon.

Fixes  #20779

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.11.4
+PKG_VERSION:=2.11.7
 PKG_RELEASE:=1
 PKG_MAINTAINER:=Florian Eckert <fe@dev.tdt.de>, \
 		Aaron Goodman <aaronjg@alumni.stanford.edu>
--- a/net/mwan3/files/usr/sbin/mwan3rtmon
+++ b/net/mwan3/files/usr/sbin/mwan3rtmon
@@ -75,7 +75,7 @@ mwan3_rtmon_route_handle()
 
 	if [ "$route_line" = "$1" ]; then
 		action="replace"
-		$IPS -! add mwan3_connected_${route_family##ip} ${route_line%% *}
+		$IPS -! add mwan3_connected_${route_family} ${route_line%% *}
 	else
 		action="del"
 		mwan3_set_connected_${route_family}


feckert:
Good catch. Could you also bump the PKG_VERSION

feckert:
Pushed changes to master with commit https://github.com/openwrt/packages/commit/84b3de9eda29666bb96b656cdadaffd1c69897e7
Pushed changes to owrt-22.03 with commit https://github.com/openwrt/packages/commit/3b80d11e2b5a7a6f9aa74d13a2ff76f701c7bcc6

@gentoo-root thanks for testing

PR #20901, opened 2023-04-28
mwan3: fix addition of iptables rules for mwan3 sticky rules

anyuta1166:
Addition of iptables rules for mwan3 sticky rules was broken in 2.11.0, resulting in non-working sticky rules. This commit fixes this issue.

Maintainer: @feckert

Description:

Addition of iptables rules for mwan3 sticky rules was broken in 2.11.0. There an error in passing arguments to mwan3_set_sticky_iptables() function, as a result some necessary rules are not added.
Example:

2.11.0+:

```
-A mwan3_rule_test -m mark --mark 0x0/0x3f00 -j mwan3_policy_balanced
-A mwan3_rule_test -m mark ! --mark 0xfc00/0xfc00 -j SET --del-set mwan3_rule_ipv4_test src,src
-A mwan3_rule_test -m mark ! --mark 0xfc00/0xfc00 -j SET --add-set mwan3_rule_ipv4_test src,src
```

2.10.13 (and 2.11.0+ with this patch):

```
-A mwan3_rule_test -m mark --mark 0x0/0x3f00 -j MARK --set-xmark 0x300/0x3f00
-A mwan3_rule_test -m mark --mark 0x300/0x3f00 -m set ! --match-set mwan3_rule_ipv4_test src,src -j MARK --set-xmark 0x0/0x3f00
-A mwan3_rule_test -m mark --mark 0x0/0x3f00 -j MARK --set-xmark 0x200/0x3f00
-A mwan3_rule_test -m mark --mark 0x200/0x3f00 -m set ! --match-set mwan3_rule_ipv4_test src,src -j MARK --set-xmark 0x0/0x3f00
-A mwan3_rule_test -m mark --mark 0x0/0x3f00 -j MARK --set-xmark 0x100/0x3f00
-A mwan3_rule_test -m mark --mark 0x100/0x3f00 -m set ! --match-set mwan3_rule_ipv4_test src,src -j MARK --set-xmark 0x0/0x3f00
-A mwan3_rule_test -m mark --mark 0x0/0x3f00 -j mwan3_policy_balanced
-A mwan3_rule_test -m mark ! --mark 0xfc00/0xfc00 -j SET --del-set mwan3_rule_ipv4_test src,src
-A mwan3_rule_test -m mark ! --mark 0xfc00/0xfc00 -j SET --add-set mwan3_rule_ipv4_test src,src
```

### DIFF
--- a/net/mwan3/Makefile
+++ b/net/mwan3/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-PKG_VERSION:=2.11.4
+PKG_VERSION:=2.11.5
 PKG_RELEASE:=1
 PKG_MAINTAINER:=Florian Eckert <fe@dev.tdt.de>, \
 		Aaron Goodman <aaronjg@alumni.stanford.edu>
--- a/net/mwan3/files/lib/mwan3/mwan3.sh
+++ b/net/mwan3/files/lib/mwan3/mwan3.sh
@@ -735,8 +735,8 @@ mwan3_set_policies_iptables()
 
 mwan3_set_sticky_iptables()
 {
-	local rule="${1}"
-	local interface="${2}"
+	local interface="${1}"
+	local rule="${2}"
 	local ipv="${3}"
 	local policy="${4}"
 
@@ -879,7 +879,7 @@ mwan3_set_user_iptables_rule()
 		fi
 
 		mwan3_push_update -F "mwan3_rule_$1"
-		config_foreach mwan3_set_sticky_iptables interface $ipv "$policy"
+		config_foreach mwan3_set_sticky_iptables interface "$rule" $ipv "$policy"
 
 
 		mwan3_push_update -A "mwan3_rule_$1" \


feckert:
Good catch! Thanks.
 Could you please bump the PKG_VERSION to 2.11.5

feckert:
@anyuta1166 I have now taken a closer look. What you have done fits so far thanks will merge in a separate Commit with updating your commit message. Thanks for that.

But I almost despaired!
After a long debug session I discovered that this does not work for inital_state is `offline`. **This has nothing to do with your change. It didn't work before.** I assume that you have set the initial state of the interface to `online`. 

I think it is because the initial_state `offline` uses this [function](https://github.com/openwrt/packages/blob/eaed3acdebc3fbf28d4bd435236489ddf819a41b/net/mwan3/files/lib/mwan3/mwan3.sh#L659). And therefore 'out' is evaluated in awk.

I wonder if you'll run into the same bug. Can you please verify this?

feckert:
Pushed to master with commit https://github.com/openwrt/packages/commit/66257510d9613478a9fd99c41d745329476f7574

PR #20902, opened 2023-04-29
nginx: refactor

hgl:
Maintainer: @Ansuel 
Compile tested: x86-64 master
Run tested: x86-64 master

Description:

Since modules now follow a naming convention, there is an opportunity to eliminate redundant code.

However, during the process, I found the following:

1. `CONFIG_NGINX_GEOIP_MODULE` didn't exist in Config_ssl.in, so it was never compiled. I tried to compile it, but nginx complained about the missing GeoIP library, which currently doesn't seem to exist in openwrt/packages? I removed the corresponding code for now, can restore if the GeoIP library exists.
2. ~~In menuconfig, if `nginx-full` is selected, it won't set any `CONFIG_NGINX_*` variables, leading `nginx-mod-dav-ext` to miss its `+@NGINX_DAV` dependency~~.
3. `nginx-full` depends on `nginx-mod-*`, which in turn depend on `nginx-ssl`, leading to conflict (current build error comes from this I believe).
4. ~~`STREAM_SSL_MODULE` and `STREAM_PREREAD_MODULE` are never used.~~
5. Should `nginx-mod-luci` be renamed to `nginx-luci` or something? Since it's not really a module.
6. ~~Should we remove `NGINX_STREAM_CORE_MODULE` and just offer `nginx-mod-stream`? Currently they seem to do the same thing~~.

### DIFF
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -295,7 +295,7 @@ define Package/nginx-mod-luci
   SUBMENU:=Web Servers/Proxies
   TITLE:=Support file for Nginx
   URL:=http://nginx.org/
-  DEPENDS:=+uwsgi +uwsgi-luci-support +nginx +nginx-mod-ubus
+  DEPENDS:=+uwsgi +uwsgi-luci-support +nginx-ssl +nginx-mod-ubus
   # TODO: add PROVIDES when removing nginx-mod-luci-ssl
   # PROVIDES:=nginx-mod-luci-ssl
 endef
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -25,6 +25,22 @@ PKG_BUILD_PARALLEL:=1
 PKG_INSTALL:=1
 PKG_BUILD_FLAGS:=gc-sections
 
+# 3rd-party modules
+PKG_MOD_EXTRA := \
+	lua \
+	rtmp \
+	dav-ext \
+	naxsi \
+	brotli \
+	headers-more \
+	ts \
+	ubus
+# official and 3rd-party modules
+PKG_MOD_ALL := \
+	$(PKG_MOD_EXTRA) \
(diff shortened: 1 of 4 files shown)


hgl:
Build passed, I'm not sure why it's trying to install both nginx-ssl and nginx-full

hgl:
> Build passed, I'm not sure why it's trying to install both nginx-ssl and nginx-full

The reason is that nginx-full depends on nginx-mod-*, and they in turn depends on nginx-ssl. This issue also exists in @Ansuel's PR if I'm not wrong, but somehow the workflow build didn't try to install the nginx-full package.

Ansuel:
@hgl i'm ok with the change just some formal issue. It's ok if you can't split them but would be good to have the rafactor change described and a better more informative title

For the --with-compat, in theory it should be optional reading the wiki and i tested module without it but for some reason now they doesn't work...

Can you verify and make sure that --with-compat actually fix the problem?

hgl:
@Ansuel commit title and description updated to better reflect changes

@mmartinv @darekxan @droid4joy Since I don't have a device with the arch that can reproduce the issue, could you help test run?

mmartinv:
@hgl sure no problem.

hgl:
I've already run tested for x86 and with-compat is in place. Just need your verdict if the not binary compatible issue you encountered is solved.

peterwillcn (approved):

Ansuel:
@hgl in your opinion is better to merge the openresty stuff first and you rebase this or merge this first?

hgl:
@Ansuel I probably won't be able to work on this for a short while. It'd be great if you could merge this first. Sorry if other authors have to rebase. 

I guess currently the merge conflict comes from the compat change you made? Since this PR already has it, it should be resolved if you just take this PR's change.

PR #20903, opened 2023-04-29
ci, cache-domains: added pre-test.sh CI step

G-M0N3Y-2503:
Maintainer: me @G-M0N3Y-2503
Compile tested: x86_x64, QEMU/KVM, master
Run tested: x86_x64, QEMU/KVM, master

Description:

Some packages variants have conflicting dependencies with the base packages and the CI test will fail to install before anything can be done by the packages to setup the system for install.

This change adds a pre-test.sh that runs before the install so things like the default libustream variant can be swapped out as shown in the updated cache-domains.


G-M0N3Y-2503:
This fixes the CI for https://github.com/openwrt/packages/pull/20803

PR #20907, opened 2023-04-30
aria2:fix ssl errors

hyy-666:
Maintainer: @kuoruan 
Compile tested: x86 aarch64
Run tested: x86,19.07,x86

Description:
Sometimes use aria2 download file through https will cause error like this 'SSL/TLS handshake failure: unable to get local issuer certificate',it's due to ujail not mount ca-certificate.
And it also need chose ca-certificate by defalut.

### DIFF
--- a/net/aria2/Makefile
+++ b/net/aria2/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=aria2
 PKG_VERSION:=1.36.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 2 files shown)


neheb:
Need a PkG_RELEASE bump.

neheb:
Please use a real name and a real email address

hyy-666:
It's ok? I'm a newbie of git.

hyy-666:
Yes,but i don't know how to do let ujail auto mount ca_certificate when i build with ssl support,and in my view disable the certificate check isn't a good idea,maybe use two aira.init (one use in no ssl support,another in ssl support) to fix it？

hyy-666:
> 

Get it

PR #20909, opened 2023-04-30
lighttpd: build fixes

gstrauss:
Maintainer: @flyn-org

Description:
build fixes for lighttpd meson

One recently reported issue: https://forum.openwrt.org/t/bpi-r3-issue-install-lighttpd/158867

### DIFF
--- a/net/lighttpd/Makefile
+++ b/net/lighttpd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lighttpd
 PKG_VERSION:=1.4.69
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 # release candidate ~rcX testing; remove for release
 #PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)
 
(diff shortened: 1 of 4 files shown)


gstrauss:
@1715173329 in #20817 I exposed a bug in the upstream lighttpd src/meson.build which was not properly including builtin_mods in the build of lighttpd.exe.  (#20817 is only on the master branch; backporting is not needed.)

This PR fixes that, along with a dependency issue with mod_mbedtls when mbedtls is built without MBEDTLS_DEBUG_C defined.  The last patch allows testing the (non-default) option for lighttpd.exe to use wolfssl crypto libs instead of nettle.

CI passes on all platforms except aarch64_cortex-a53, which fails with errors unrelated to these lighttpd patches:
```
Collected errors:
 * pkg_hash_check_unresolved: cannot find dependency kernel (= 5.15.109-1-75a1f8e492d9c6fccf1395c3f046e587) for kmod-fs-autofs4
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for pgsql-server:
 * 	kernel (= 5.15.109-1-75a1f8e492d9c6fccf1395c3f046e587)
 * opkg_install_cmd: Cannot install package pgsql-server.
```

1715173329 (approved):
LGTM - Thanks!

PR #20914, opened 2023-05-01
CI: Enable runtime_test for mips_24kc

brvphoenix:
Maintainer: @aparcar @neheb @BKPepe @1715173329
Compile tested: CI modifications are tested on my local machine. Besides, the workflow has run successfully.
Run tested: N/A

Description:
The QEMU provided by Ubuntu supports emulating mips_24kc for a long time. Use it instead of the docker `aptman/qus` to enable runtime test for mips_24kc.

QEMU supports emulating mips_24kc since https://github.com/qemu/qemu/commit/77d119dd335f910c7f953a265726e3753c69a0bb. Both [`aptman/qus`](https://github.com/dbhi/qus) and [`multiarch/qemu-user-static`](https://github.com/multiarch/qemu-user-static) use an older version of `qemu-binfmt-conf.sh`.

Update sing-box to trigger the runtime test.

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -17,7 +17,7 @@ jobs:
 
           - arch: mips_24kc
             target: ath79-generic
-            runtime_test: false
+            runtime_test: true
 
           - arch: mipsel_24kc
             target: mt7621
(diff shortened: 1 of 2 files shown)


neheb:
sounds like it can be adapted for more platforms.

PR #20921, opened 2023-05-02
isc-dhcp: Support for coexistence of IPv4 and IPv6

wltjr:
Maintainer: @pprindeville 
Compile tested: mt7622-linksys_e8450 on snapshot from mid-march
Run tested: mt7622-linksys_e8450 on snapshot from mid-march

Description:

Initial modifications to support coexistence of dhcpd for both IPv4 and IPv6. If you try to enable both now, it will fail during build and/or install, plus you only have a single binary. In theory, the single binary with flags should work, but for some reason it does not, or needs further changes to the build system. Thus producing a dhcpd6 binary that still requires a -6 flag when run.

It is possible this can be improved upon, and I am open to making any changes as needed, or anyone else is welcome. I used to maintain this in [my own feed](https://github.com/Obsidian-StudiosInc/openwrt-feed/tree/master/packages/isc-dhcp), when it was removed, so I am pleased that it is back in now. You can see some commit history there as to past issues with the different binary names, etc. I have been using this for a few months now, just lacked time to submit PR.

### DIFF
--- a/net/isc-dhcp/Makefile
+++ b/net/isc-dhcp/Makefile
@@ -132,15 +132,33 @@ $(call Package/isc-dhcp-server/description)
  This package is compiled with IPv4 and IPv6 support.
 endef
 
-define Package/isc-dhcp-dyndns
+define Package/isc-dhcp-dyndns-ipv4
   $(call Package/isc-dhcp/Default)
   TITLE+= server dynamic DNS dependencies (meta)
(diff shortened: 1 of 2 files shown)


pprindeville:
The IPv6 variant is really an IPv4+IPv6 version.  This is not reflected in your changes.

wltjr:
The build system does not provide means for creating a single binary that supports both. You are welcome to try, I did not have any success in my attempts. If you enable both as is now, it will fail as it tries to overwrite the same dhcpd binary.

PR #20924, opened 2023-05-03
mwan3: fix mwan3 doesn't detect the absence of the internet connection if an interface goes down and then up

anyuta1166:
This commit addresses the following issues:
1) The interface goes down and score is reset to 0. Interface goes up, the interface initial state is online, and thus the state is set to online. mwan3track performs ping tests. But if there is no internet connection, nothing is done because the score is already 0 and the branch that handles disconnection does not call disconnecting/disconnected in that case. Interface stays in online state. 
2) The interface is connecting and score reaches score=up. The connected function is called but we do not bump the score. If we lose the internet connection right after that, we end up in disconnection branch with score<up, and disconnecting/disconnected function is not called, we just end up with score=0 and status=online like in previous scenario.

The solution is to reset the score in connected function in the same way as we set the score in disconnected.
Also remove the redundant setting of score=0 (already set in disconnected function).

Maintainer: @feckert 

Description:

### DIFF
--- a/net/mwan3/files/usr/sbin/mwan3track
+++ b/net/mwan3/files/usr/sbin/mwan3track
@@ -128,6 +128,7 @@ connected() {
 	echo "online" > $MWAN3TRACK_STATUS_DIR/$INTERFACE/STATUS
 	echo "0" > $MWAN3TRACK_STATUS_DIR/$INTERFACE/OFFLINE
 	get_uptime > $MWAN3TRACK_STATUS_DIR/$INTERFACE/ONLINE
+	score=$((down+up))
 	host_up_count=0
 	lost=0
 	turn=0
@@ -358,7 +359,6 @@ main() {
 
 			if [ $score -eq $up ]; then
 				disconnected
-				score=0
 			fi
 		else
 			if [ $score -lt $((down+up)) ] && [ $lost -gt 0 ]; then


feckert:
As I mentioned in the [post](https://github.com/openwrt/packages/pull/20428#issuecomment-1458315876), I have also refactored the score handling.

If we now add this change to set score to up+down on connected, the whole thing should be less error prone. As already mentioned, there are certainly conditions that are not covered. But we have to live with that if we don't want to refactor everything.

@gentoo-root @anyuta1166 Can someone please test / verify [this](https://github.com/TDT-AG/packages/tree/pr/20230504-mwan3) and provide feedback?

Please note the following use case of the tracking interface /etc/config/mwan3:
`option initial_state 'online'`
and
`option initial_state 'offline'`

anyuta1166:
> Can someone please test / verify [this](https://github.com/TDT-AG/packages/tree/pr/20230504-mwan3) and provide feedback?
> 
> Please note the following use case of the tracking interface /etc/config/mwan3: `option initial_state 'online'` and `option initial_state 'offline'`

@feckert I've performed some tests, seems to work fine.

feckert:
pushed directly with commits https://github.com/openwrt/packages/commit/5f0461be8b149aeb5732481d38c9d69650996f8c https://github.com/openwrt/packages/commit/1f6bd672fecd1ffc9f6b1fea152edc62bcdca026

@anyuta1166 @gentoo-root thanks for the fix and sorry @gentoo-root for the inconvenience

PR #20925, opened 2023-05-04
python-hatchling: Update to 1.14.1

jefferyto:
Maintainer: me
Compile tested: armvirt-32, 2023-05-01 snapshot sdk
Run tested: N/A

Description:


jefferyto:
CI is failing for the same reason as #20810. Since this is a host-only package, no ipk files are generated; the CI expects ipk files to install and test.

Issue #20904, opened 2023-04-29
nginx fail since last commit series.


anomeome:
@Ansuel 

_nginx_ failure to launch:
```
Sat Apr 29 15:48:23 2023 daemon.err nginx_init: 2023/04/29 15:48:23 [emerg] 2765#0: unknown directive "ubus_interpreter" in /etc/nginx/conf.d/luci.locations:21
Sat Apr 29 15:48:23 2023 daemon.err nginx_init: nginx: configuration file /etc/nginx/uci.conf test failed
Sat Apr 29 15:48:23 2023 daemon.err nginx_init: NOT using conf file!
Sat Apr 29 15:48:23 2023 daemon.notice procd: /etc/rc.d/S80nginx: nginx_init: NOT using conf file!
Sat Apr 29 15:48:23 2023 daemon.notice procd: /etc/rc.d/S80nginx: show config to be used by: nginx -T -c '/etc/nginx/uci.conf'
```
is there any manual changes needed. mvebu : rango

Ansuel:
Yes but in theory a migration script should have applied and added the new stuff

Arduous:
Encountering the same failure. What are the migration scripts? What are the triggers to these scripts? Thank you

bauw2008:
nginx_init: NOT using conf file!
show config to be used by: nginx -T -c '/etc/nginx/uci.conf'
root@OpenWrt:~# nginx -T
2023/05/01 23:01:07 [emerg] 7935#0: open() "/etc/nginx/nginx.conf" failed (2: No                   such file or directory)
nginx: configuration file /etc/nginx/nginx.conf test failed
root@OpenWrt:~# /etc/init.d/rpcd restart
root@OpenWrt:~# service nginx reload
Command failed: Not found

Encountering the same failure.

Ansuel:
Yep already know the problem plan to solve today

Il Lun 1 Mag 2023, 17:14 bauw2008 ***@***.***> ha scritto:

hgl:
This is addressed in #20902, but you will still encounter a 502 error when visiting LuCI. To fix that, the uwsgi package needs to be updated, as its `chdir` option currently points to `/usr/lib/lua/luci/`, which doesn't exist.

Ansuel:
@hgl can you point me how you fixed the issue?

hgl:
> @hgl can you point me how you fixed the issue?

[...]

PR #20928, opened 2023-05-04
nginx: fix missing module entry on package update

Ansuel:
Uci defaults script are run only on first installation. On package upgrade, the ubus module for the nginx-mod-luci is not installed in the nginx conf. Fix this by adding a post install script.

Fixes: 65a676ed56fb ("nginx: introduce support for dynamic modules")
Fixes: #20904

---

@anomeome can you test? I assume your problem was with upgrading the package right? or it was from a clean image?

### DIFF
--- a/net/nginx-util/Makefile
+++ b/net/nginx-util/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nginx-util
 PKG_VERSION:=1.6
-PKG_RELEASE:=16
+PKG_RELEASE:=17
 PKG_MAINTAINER:=Peter Stadler <peter.stadler@student.uibk.ac.at>
 
 include $(INCLUDE_DIR)/package.mk
--- a/net/nginx-util/files/uci.conf.template
+++ b/net/nginx-util/files/uci.conf.template
@@ -1,6 +1,7 @@
 # Consider using UCI or creating files in /etc/nginx/conf.d/ for configuration.
 # Parsing UCI configuration is skipped if uci set nginx.global.uci_enable=false
 # For details see: https://openwrt.org/docs/guide-user/services/webserver/nginx
+# UCI_CONF_VERSION=1.1
 
 worker_processes auto;
 
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -402,6 +402,7 @@ CONFIGURE_ARGS += \
 			--prefix=/usr \
 			--conf-path=/etc/nginx/nginx.conf \
 			--modules-path=/usr/lib/nginx/modules \
+			--with-compat \
 			$(ADDITIONAL_MODULES) \
 			--error-log-path=stderr \
 			--pid-path=/var/run/nginx.pid \
--- a/net/nginx/files/nginx.init
+++ b/net/nginx/files/nginx.init
@@ -8,11 +8,27 @@ USE_PROCD=1
 G_OPTS="daemon off;"
 
 NGINX_UTIL="/usr/bin/nginx-util"
+UCI_CONF_TEMPLATE="/etc/nginx/uci.conf.template"
+LATEST_UCI_CONF_VERSION="1.1"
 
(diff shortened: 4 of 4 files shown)


Ansuel:
> Edit: should also add that when issue first presented, I tried doing an install followed by restore of config and issue manifested.

can you detail this? Trying to find a good way to repro this...

anomeome:
When issue first presented on an upgrade of a _rango_, I did a test of install wo keeping config on _mamba_ and it worked. As a test, I did a backup of config of _rango_, install wo keeping config, nginx was fine, restored config, rebooted and nginx was borked. Checked what was being backup up to see if there was anything untoward beyond defaults:
```
/etc/vpn-policy-routing.netflix.user
/etc/vpn-policy-routing.aws.user
/etc/ssl/private
/etc/openvpn/
/root/
```
seemed fine and opened the issue at that point.

hgl:
@anomeome could you check if the refactor PR has the same issue? I mentioned it in the original issue.

Ansuel:
> When issue first presented on an upgrade of a _rango_, I did a test of install wo keeping config on _mamba_ and it worked. As a test, I did a backup of config of _rango_, install wo keeping config, nginx was fine, restored config, rebooted and nginx was borked. Checked what was being backup up to see if there was anything untoward beyond defaults:
> 
> ```
> /etc/vpn-policy-routing.netflix.user
> /etc/vpn-policy-routing.aws.user
> /etc/ssl/private
> /etc/openvpn/
> /root/
> ```
> 
> seemed fine and opened the issue at that point.

no presence of the nginx directory in the config backup?

anomeome:
Yes, _/ettc/nginx/*_ is there, but presumably that is by design OOTB default of the config backup.

@hgl , I saw that but did not want to conflate the firefight

Ansuel:
@anomeome anyway this is a corner case that is really difficult to fix... the nginx template changed

https://github.com/openwrt/packages/pull/16842/files#diff-396e25a1839e80e1e12c6465b7b2971c69c8f17357dfee4db8295ecab2f48318

So on config restore i assume the uci template have the include entry missing... Can you check that?

Also you should correctly have modules.d and the module

anomeome:
Yep, include missing, rest is good. Not sure this is really an corner case though, seems that everyone keeping a config is going to hit the issue.

Ansuel:
> Yep, include missing, rest is good. Not sure this is really an corner case though, seems that everyone keeping a config is going to hit the issue.

yep but the fix is not trivial... the only place i think is init.d but it's a big hack...

hgl:
Could an uci-defaults script grep and add the line help?

Ansuel:
@hgl the script is already there... problem is that config needs to be migrated at runtime and that doesn't happen so people restoring old config result in old template...

Ansuel:
@hgl @anomeome can you test? You should use the old config where you didn't change or add the thing to the template (or just remove the string, backup and restore the config)

hgl:
I think what I don't really understand, which is also why I kept asking you about uci-defaults, is that why we need to migrate every time nginx starts with the init.d script, instead of just once after package install with uci-defaults?

I checked your dynamic module PR, but I only found [the added load_module in directive in uci.conf.template](https://github.com/openwrt/packages/pull/16842/files#diff-396e25a1839e80e1e12c6465b7b2971c69c8f17357dfee4db8295ecab2f48318). I don't understand nginx-util well, but I assume that line doesn't get added if a preexisting conf exists? With uci-defaults, you should be able to inspect the file, looking for `load_module`, if it doesn't exist, add it. I don't think you even need to version it.

Ansuel:
The problematic case is people restoring an old config.... The template is
part of the back... UCI defaults are not run on config restore

Il Mar 16 Mag 2023, 04:44 Glen Huang ***@***.***> ha scritto:

hgl:
Oh, I didn't know config could be restored. OK, LGTM.

I don't have an old config so I can't really test.

Ansuel:
I would wait for @anomeome to test and then merge your refactor pr

Issue #20906, opened 2023-04-30
nginx-full fail to compile after last commit


zcy85611:
with everthing unchanged, nginx-full fail to compile, it gives "install: cannot stat '/workdir/openwrt/build_dir/target-mipsel_24kc_musl/nginx-all-module/nginx-1.24.0/ipkg-install/usr/lib/nginx/modules/ngx_stream_module.so': No such file or directory"

would you please take a look, thanks?

```
make[3]: Entering directory '/workdir/openwrt/feeds/packages/net/nginx-util'
mkdir -p /workdir/openwrt/staging_dir/target-mipsel_24kc_musl/root-ramips/stamp
SHELL= flock /workdir/openwrt/tmp/.root-copy.flock -c 'cp -fpR /workdir/openwrt/build_dir/target-mipsel_24kc_musl/nginx-util-1.6/.pkgdir/nginx-ssl-util/. /workdir/openwrt/staging_dir/target-mipsel_24kc_musl/root-ramips/'
touch /workdir/openwrt/staging_dir/target-mipsel_24kc_musl/root-ramips/stamp/.nginx-ssl-util_installed
echo "nginx-ssl-util" >> /workdir/openwrt/staging_dir/target-mipsel_24kc_musl/pkginfo/nginx-util.default.install
mkdir -p /workdir/openwrt/staging_dir/target-mipsel_24kc_musl/root-ramips/stamp
SHELL= flock /workdir/openwrt/tmp/.root-copy.flock -c 'cp -fpR /workdir/openwrt/build_dir/target-mipsel_24kc_musl/nginx-util-1.6/.pkgdir/nginx-util/. /workdir/openwrt/staging_dir/target-mipsel_24kc_musl/root-ramips/'
touch /workdir/openwrt/staging_dir/target-mipsel_24kc_musl/root-ramips/stamp/.nginx-util_installed
echo "nginx-util" >> /workdir/openwrt/staging_dir/target-mipsel_24kc_musl/pkginfo/nginx-util.default.install
make[3]: Leaving directory '/workdir/openwrt/feeds/packages/net/nginx-util'
time: package/feeds/packages/nginx-util/compile#0.17#0.06#0.22
make[3]: Entering directory '/workdir/openwrt/feeds/packages/utils/dbus'
make[3]: Leaving directory '/workdir/openwrt/feeds/packages/utils/dbus'
time: package/feeds/packages/dbus/compile#0.16#0.04#0.18
[...]

PR #20934, opened 2023-05-05
nginx: fix compilation error for nginx-full

Ansuel:
Fix compilation error for stream module not converted to use the PACKAGE config flag and a missing required dependency for the DAV ext module.

Drop additional config for STREAM module since they are now included and built by default.

Fixes: 65a676ed56fb ("nginx: introduce support for dynamic modules")
Fixes: #20906

---

@EdiTurn @zcy85611 Can you test this?

### DIFF
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nginx
 PKG_VERSION:=1.24.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=nginx-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://nginx.org/download/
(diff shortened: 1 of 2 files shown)


net/nginx/Makefile
```diff
@@ -382,7 +378,7 @@ endef
 $(eval $(call module,lua, +luajit,ngx_http_lua, Enable Lua module))
 $(eval $(call module,stream, +@NGINX_STREAM_CORE_MODULE,ngx_stream, Add support for NGINX request streaming.))
 $(eval $(call module,ubus, +libubus +libjson-c +libblobmsg-json +@NGINX_UBUS,ngx_http_ubus, Enable UBUS api support directly from the server.))
-$(eval $(call module,dav-ext, +@NGINX_DAV,ngx_http_dav_ext, Enable the WebDAV methods PROPFIND OPTIONS LOCK UNLOCK.))
+$(eval $(call module,dav-ext, +@NGINX_DAV +libxml2,ngx_http_dav_ext, Enable the WebDAV methods PROPFIND OPTIONS LOCK UNLOCK.))
```
hgl, line 381: Since dav-ext depends on `NGINX_DAV`, and `+NGINX_DAV:libxml2` exists for nginx-ssl, `+libxml2` exists for nginx-full, why is `+libxml2` here necessary?

Ansuel: .so dependency scan complain on creating the separate package for dav-ext. the dependency is present for dav for nginx but is also required for the separate module dav-ext

EdiTurn:
Confirm the compilation is successful.

PR #20967, opened 2023-05-08
zabbix: add patch for gnutls version handling

feckert:
Maintainer: @champtar 
Compile tested: x86/64 and lantiq/xrx00

Description:
Zabbix checks the version of gnutls by search in the header gnutls.h. This is done with 'cat' and 'egrep'. The problem here is now that the preprocess definition did changed in gnutls.h in gntuls version 3.8.0, so the regex does not match anymore. The following error message a course in the log.

`configure: error: GnuTLS library version requirement not met (>= 3.1.18)`

To fix this the, regex must be corrected with this commit, so the version check does work again

**This is only needed for the current gnutls version 3.8.0. The master branch of gnutls already fixed this, so on rebasing to a new version of gnutls this patch can be removed from zabbix, because the check does work again.**
https://github.com/gnutls/gnutls/commit/18345986e29c820e64daced78232f236fd4a4e6e

### DIFF
--- a/admin/zabbix/Makefile
+++ b/admin/zabbix/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=zabbix
 PKG_VERSION:=6.2.3
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://cdn.zabbix.com/zabbix/sources/stable/6.2/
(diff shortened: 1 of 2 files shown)


BKPepe (changes requested):
I don't like it also.

The patch itself is missing or lacks common detaiils like:
- why it is necessary
- who is the author (to reach, if it is not upstreamed)

And we should not add more hacks than they are necessary. Same applies for non upstreamed patches. Gnutls should be patched not this.

feckert:
@BKPepe @1715173329 thanks for the reply, I will open a pullrequest against the gnutls package to fix this issue.

PR #20980, opened 2023-05-11
lighttpd: update to lighttpd 1.4.70 release hash

gstrauss:
Maintainer: @flyn-org

Description:
* update to lighttpd 1.4.70 release hash
* select wolfssl for default cryptolib
* include mod_h2 in base package


gstrauss:
@flyn-org @1715173329 @BKPepe 

CI passes (once I remembered to build target `refresh QUILT=1` to fixup the patches)

lighttpd 1.4.70 and these patches now use wolfssl by default (instead of Nettle) as the base crypto library for digest functions (MD5, SHA256, etc).  After a couple weeks on the openwrt master branch, I plan to backport this to OpenWRT 21.02 and to OpenWRT 22.03.  Starting with OpenWRT 21.02, wolfssl is included in openwrt base.

Also in these patches is a change to move lighttpd HTTP/2 code into an optional module mod_h2, so that memory use can be reduced if HTTP/2 support is disabled in lighttpd.conf.  HTTP/2 support remains enabled by default in lighttpd, since lighttpd 1.4.59.

Note: I do have a concern about the startup performance of wolfssl, and so plan to omit the backport of changing to use wolfssl for digest functions.  I have previously observed that lighttpd starts up very quickly with nettle and mod_openssl on a very old NetGear router I have, but takes close to 10 seconds to start when using lighttpd with nettle and mod_wolfssl.

1715173329:
Okay, what about mbedtls? It's the default crypto lib we used in base now.

gstrauss:
> Okay, what about mbedtls? It's the default crypto lib we used in base now.

Neat.  I personally prefer mbedtls.  I'll update the PR.

neheb:
wolfSSL has been awful for OpenWrt 22.03. Too much ABI breakage.

gstrauss:
CI passes with small change to make mbedtls to default lib that lighttpd uses for digest functions and randomness.

When backporting to OpenWRT 21.02 and 22.03, I plan to leave the default cryptolib in lighttpd as Nettle.

PR #20983, opened 2023-05-11
Remove dependencies on sftp packages and use CDN for downloads.

daztucker:
Maintainer: @neheb 
Compile tested: mips_24k, gl.inet AR150
Run tested: mips_24k, gl.inet AR150, installed individual packages with and (in the case of sftp ones) without dependencies.

Description:
    OpenSSH 9.1p1 removed remaining dependencies and stopped linking sftp,
    sftp-server and scp against libcrypto or libz.  This change moves those
    package dependencies from the default to those that still need them.
    In particular, this will allow sftp-server to be installed for use with
    Dropbear without needing to install zlib or openssl.

### DIFF
--- a/net/openssh/Makefile
+++ b/net/openssh/Makefile
@@ -12,8 +12,8 @@ PKG_VERSION:=9.3p1
 PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


daztucker:
the pentium_mmx test failure looks like an infrastructure problem (it just stopped in the middle of the run).  If I was able to I'd rerun just the failing test.

PR #20989, opened 2023-05-11
Update mdns-repeater.init

0xF4CED:
Maintainer: @axkg (?)
Compile tested: not relevant
Run tested: OpenWrt 22.03.5 

Description:
Increase start init sequence to 99, because the service fails at boot with 50 when repeating between two vlans.


neheb:
Need a PKG_RELEASE bump

BKPepe:
It requires more than just ``PKG_RELEASE`` bump.
 - Real name in the Signed-off-by
  - Commit description should be nearly the same as described in OP
  - The commit subject should start with the package name
  - We know that it was only on some OpenWrt router without specifying it.

BKPepe:
Closing, this PR is stalled.

Issue #20971, opened 2023-05-09
syslog-ng: late start on boot prevents dnsmasq from starting


ardrabczyk:
Maintainer: @BKPepe
Environment: armv7l, InHand Networks VG710, OpenWrt 22.03.0-rc4

Description:

/etc/rc.d/S19dnsmasq tries to mount /dev/log in jail but fails because /dev/log doesn't exist yet because /etc/rc.d/S50syslog-ng hasn't been run yet. The end result is that dnsmasq is not running:

```
~# ps aux | grep '[d]nsmasq'
~#
```

The issue can be easily reproduced even without syslog-ng installed:

```
~# mv /dev/log /dev/log.bak
~# /sbin/ujail -t 5 -n test-ujail -l -- /bin/echo hi; echo $?
jail: stat(/dev/log) failed: No such file or directory
jail: mount_all() failed
jail: failed to build jail fs
jail: jail (9738) exited with exit: 1
1
```

I wonder if other users of OpenWRT also have this problem? I haven't seen any bug reports but it looks quite serious so I'm wondering. Would it be a bad idea to start syslog-ng earlier, S12 as with logd should be ok, in my test even S18 (so one less that S19 for dnsmasq) was enough?

brada4:
In principle yes, it conflicts logd on `logread` file, thus can take same place easily.

ardrabczyk:
I'll send a PR and see what [BKPepe](https://github.com/BKPepe) thinks

brada4:
I think there is no further benefit of this issue, you got additional backing point for your concept and can add to your PR.

PR #20990, opened 2023-05-11
syslog-ng: start early

ardrabczyk:
This is needed for services such as /etc/rc.d/S19dnsmasq that start
before S50syslog-ng and expect /dev/log to exist to mount it in jail.

Fixes https://github.com/openwrt/packages/issues/20971

Maintainer: @BKPepe
Compile tested:  x86-64 Ubuntu 22.04.1 LTS, CONFIG_TARGET_ipq40xx, OpenWrt 22.03.0-rc4
Run tested: armv7l, InHand Networks VG710, OpenWrt 22.03.0-rc4

The purpose of the test was to check if dnsmasq starts:

    root@OpenWrt:~# busybox ps w | grep '[d]nsmasq'
     4443 root      2408 S    {dnsmasq} /sbin/ujail -t 5 -n dnsmasq -u -l -r /bin/ubus -r /etc/TZ -r /etc/dnsmasq.conf -r /etc/ethers -
     4447 dnsmasq   1260 S    /usr/sbin/dnsmasq -C /var/etc/dnsmasq.conf.cfg01411c -k -x /var/run/dnsmasq/dnsmasq.cfg01411c.pid

Description:

### DIFF
--- a/admin/syslog-ng/files/syslog-ng.init
+++ b/admin/syslog-ng/files/syslog-ng.init
@@ -1,7 +1,7 @@
 #!/bin/sh /etc/rc.common
 # Copyright (C) 2006-2019 OpenWrt.org
 
-START=50
+START=12
 
 USE_PROCD=1
 


ardrabczyk:
This fixes one problem but introduces another, if S20network takes a lot of time and there is no localhost set up when syslog-ng starts:

    [2023-05-17T10:21:12.096129] WARNING: experimental behaviors of the future syslog-ng 4.0 are now enabled. This mode of operation is meant to solicit feedback and allow the evaluation of the new features. USE THIS MODE AT YOUR OWN RISK, please share feedback via GitHub, Gitter.im or email to the authors; config-version='4.0'
    [2023-05-17T10:21:17.649083] Error binding socket; addr='AF_INET6([::1]:514)', error='Address not available (99)'
    [2023-05-17T10:21:17.649473] Error initializing message pipeline; plugin_name='network', location='#buffer:3:1'

Maybe we should install interface trigger or wait for localhost interface to show up before starting?

ardrabczyk:
@BKPepe any opinion on this?

ynezz:
> This is needed for services such as /etc/rc.d/S19dnsmasq that start

It is as well needed in situations where you don't want to lost important system logs during startup phase.

> Maybe we should install interface trigger or wait for localhost interface to show up before starting?

Then you might be again risking loosing some important system messages during startup.

> [2023-05-17T10:21:17.649083] Error binding socket; addr='AF_INET6([::1]:514)', error='Address not available (99)'

General question is, does everyone need that `network_localhost()` feature enabled by default? Perhaps it should be configurable and disabled by default?

ardrabczyk:
On Tue, Aug 08, 2023 at 04:36:23AM -0700, Petr Štetiar wrote:
>     This is needed for services such as /etc/rc.d/S19dnsmasq that start
>
> It is as well needed in situations where you don't want to lost important
> system logs during startup phase.
>
>     Maybe we should install interface trigger or wait for localhost interface
>     to show up before starting?
>
> Then you might be again risking loosing some important system messages during
> startup.
>
>     [2023-05-17T10:21:17.649083] Error binding socket; addr='AF_INET6
>     ([::1]:514)', error='Address not available (99)'

Good points.

> General question is, does everyone need that network_localhost() feature
> enabled by default? Perhaps it should be configurable and disabled by default?

Sounds good to me, I have disabled it in my local tree some time ago.

ardrabczyk:
Starting syslog-ng earlier isn't enough, to make 100% sure that all services that rely on /dev/log being present we should add this to syslog-ng init:

    service_started() {
	while true
	do
	    [ -S /dev/log ] && break
	    sleep 1
	done
    }

PR #20991, opened 2023-05-11
pbr: bugfix: create IPv6 routes

stangri:
Maintainer: me
Compile tested: x86_64, Sophos XG-135r3, OpenWrt 22.03.5
Run tested: x86_64, Sophos XG-135r3, OpenWrt 22.03.5, test IPv6 routes

Description:
* add missing space in str_contains
* unquote variable to make sure IPv6 rotues are added
* add IPv6 routes display to status output in nft mode

### DIFF
--- a/net/pbr/Makefile
+++ b/net/pbr/Makefile
@@ -5,7 +5,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=pbr
 PKG_VERSION:=1.1.1
-PKG_RELEASE:=5
+PKG_RELEASE:=7
 PKG_LICENSE:=GPL-3.0-or-later
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.ca>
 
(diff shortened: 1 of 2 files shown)


schm0:
Doesn't work.
The pbr_<wg interface> route is just a duplicate of the default route.
This should have target ::/0 instead of the link local address of the wg interface.

stangri:
> Doesn't work. The pbr_wgInterface route is just a duplicate of the default route. It should have target ::/0 instead of the link local address of the wg interface?

This will carry over whatever routing was set up prior to pbr run into the pbr tables.

schm0:
hmm okay.
Then I will leave my manual created route in place, thank you!

stangri:
> hmm okay. Then I will leave my manual created route in place, thank you!

If you think that your case is applicable to others using IPv6 on WAN and tunnels, I'd welcome a PR to supplement/add necessary routes.

None of the WG tunnels I use provide any IPv6 address/routing and my only option for WAN6 is 6-in-4, so I have a very limited exposure to IPv6.

PR #21003, opened 2023-05-12
libvorbis: fix soname of shared libraries

robho:
When building with cmake we don't get libvorbis*.so.N which means that some applications may fail to dynamically load libvorbis libraries.

This seems to be a problem specific to cmake (autotools builds work).

Maintainer: @thess 
Compile tested: ath79, D-Link DIR-825, 22.03
Run tested: ath79, D-Link DIR-825, 22.03, I've verified that squeezelite now can dynamically load libvorbisfile.so.3

### DIFF
--- a/libs/libvorbis/Makefile
+++ b/libs/libvorbis/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libvorbis
 PKG_VERSION:=1.3.7
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


libs/libvorbis/patches/series
```diff
@@ -0,0 +1 @@
+010-cmake_soname.patch
```
1715173329, line 1: `series` file is basically unused by openwrt - simply remove it

1715173329:
upstream pr: https://github.com/xiph/vorbis/pull/98

PR #21005, opened 2023-05-13
prometheus-node-exporter-lua: Add thermal collector

joelpet:
Maintainer: @dhewg @champtar @jpds @PolynomialDivision @mweinelt (committed `Makefile` within the last two years)
Compile tested: N/A (the script is not compiled)
Run tested: aarch64, Raspberry Pi 4 Model B,  OpenWrt 22.03.0, r19685-512e76967f;
- Say `opkg install prometheus-node-exporter-lua` (plus some more exporters)
- Install the `thermal.lua` script in the `/usr/lib/lua/prometheus-collectors` directory
- Say `/etc/init.d/prometheus-node-exporter-lua restart`
- Manually inspect exported metrics with `wget -O- -4 http://localhost:9100/metrics`:
```
# TYPE node_thermal_temp gauge
node_thermal_temp{thermal_zone="0"} 50.147
node_scrape_collector_duration_seconds{collector="thermal"} 0.00015020370483398
node_scrape_collector_success{collector="thermal"} 1
```

Description:
This was the most straightforward way I could come up with to let me monitor the CPU temperature of my Raspberry Pi running OpenWrt, and it gets the job done for me.
- Do you think this would be useful enough for inclusion in an OpenWrt package like this?
- Does it belong in the main `prometheus-node-exporter-lua` package, as suggested in this pull request? Or is it better suited in its own package, similar to e.g. `prometheus-node-exporter-lua-wifi`?
- What do you think about the Prometheus metric and label naming?

### DIFF
--- a/utils/prometheus-node-exporter-lua/Makefile
+++ b/utils/prometheus-node-exporter-lua/Makefile
@@ -4,7 +4,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=prometheus-node-exporter-lua
-PKG_VERSION:=2022.08.08
+PKG_VERSION:=2023.05.14
 PKG_RELEASE:=1
 
 PKG_MAINTAINER:=Etienne CHAMPETIER <champetier.etienne@gmail.com>
@@ -169,6 +169,17 @@ define Package/prometheus-node-exporter-lua-textfile/install
 	$(INSTALL_BIN) ./files/usr/lib/lua/prometheus-collectors/textfile.lua $(1)/usr/lib/lua/prometheus-collectors/
 endef
 
+define Package/prometheus-node-exporter-lua-thermal
+  $(call Package/prometheus-node-exporter-lua/Default)
+  TITLE+= (thermal collector)
+  DEPENDS:=prometheus-node-exporter-lua
+endef
+
+define Package/prometheus-node-exporter-lua-thermal/install
+	$(INSTALL_DIR) $(1)/usr/lib/lua/prometheus-collectors
+	$(INSTALL_BIN) ./files/usr/lib/lua/prometheus-collectors/thermal.lua $(1)/usr/lib/lua/prometheus-collectors/
+endef
+
 define Package/prometheus-node-exporter-lua-ubnt-manager
   $(call Package/prometheus-node-exporter-lua/Default)
   TITLE+= (ubnt-manager collector)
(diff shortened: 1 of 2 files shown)


mweinelt:
From what I remember the thermal zone numbers aren't stable between reboots, so if you have multiple you'd need something else to better identify it.

joelpet:
> From what I remember the thermal zone numbers aren't stable between reboots, so if you have multiple you'd need something else to better identify it.

Thank you for your input, @mweinelt! I did a quick investigation, but unfortunately I was unable to find a way of getting a stable identifier. Instead, I had a look at how the official Prometheus node exporter works and imitated that behavior simply by including a few more labels, namely `type` and `policy`, plus `mode` and `passive` (if available). For the official node exporter implementation details, see [parseClassThermalZone](https://github.com/prometheus/procfs/blob/035b875548bab3743cdfd915b3564fab4ab61efe/sysfs/class_thermal.go#L63). These changes are available in [a55921c..e03c124](https://github.com/openwrt/packages/compare/a55921cf772542b1a384abafcdb65b4a014ed531..e03c124d1b0e9452c2fd349e1036d1b9db163d73).

I also updated the metric name to match the one in the official Prometheus node exporter. See [e03c124..c02fab3](https://github.com/openwrt/packages/compare/e03c124d1b0e9452c2fd349e1036d1b9db163d73..c02fab36648d184c4351e6250d20467a32004f0a).

Is there anything else that need to be addressed?

champtar:
Hi @joelpet 

Thanks for your contribution !

I would put this in a separate `-thermal` package
You need to bump the PKG_VERSION
`Test Formalities` need to pass
output must match exactly the official implementation, for example I see that you use `temp / 1000` but I don't see such unit change in the official implementation

joelpet:
Thank you for your feedback, @champtar!

> I would put this in a separate `-thermal` package

Does [this](https://github.com/openwrt/packages/pull/21005/commits/13b9accd56e188d2e7909cadf28a556fe62bcceb) look alright?

> You need to bump the PKG_VERSION

[Bumped](https://github.com/openwrt/packages/pull/21005/commits/d8d995d31f8f9a1f7609cc189cfb3f2c26210db5). Is today's date fine, or do I put a later one, anticipating the time it'll take to get the patch correct first?

> `Test Formalities` need to pass

Amended with `--signoff`. That was the only problem I saw.

> output must match exactly the official implementation, for example I see that you use `temp / 1000` but I don't see such unit change in the official implementation

I believe the output actually is identical to the official implementation; in my second iteration, I picked the labels based on the official node exporter `parseClassThermalZone()` function, and the division by 1000 is done in the [thermal zone collector](https://github.com/prometheus/node_exporter/blob/184a4e0893dd5c28e540ca3070f2e3a07f939f11/collector/thermal_zone_linux.go#L87).

Also, `prometheus-node-exporter/stable,now 1.1.2+ds-2.1 amd64` in Debian returns the following metrics, which confirms that the temperature is converted into Celsius:

```
# HELP node_thermal_zone_temp Zone temperature in Celsius
# TYPE node_thermal_zone_temp gauge
node_thermal_zone_temp{type="acpitz",zone="0"} 27.8
node_thermal_zone_temp{type="acpitz",zone="1"} 29.8
node_thermal_zone_temp{type="x86_pkg_temp",zone="2"} 46
```

Please let me know if there's anything I've missed, or anything else I can do to facilitate the merging of this. Thanks!

P.S.: I'll squash and force-push right after I've posted this message.

joelpet:
@champtar I have had a look through the logs from the failed checks (e.g. https://github.com/openwrt/packages/actions/runs/5260027758/jobs/9776037911?pr=21005), but I am unable to make a connection between the build failures and my proposed changes. Am I missing something? If so, could you point me in the right direction?

champtar:
Sorry for the delay and thanks again for your contribution !

Issue #20969, opened 2023-05-08
gnutls: build fails (gcc 13)


oskarirauta:
gnutls build fails with latest changes, **might** be related to gcc 13 - or not.

```
...
make[7]: Leaving directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src/gl'
Making all in tests
make[7]: Entering directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src/gl/tests'
## ---------------------------------------------------- ##
## ------------------- Gnulib tests ------------------- ##
## You can ignore compiler warnings in this directory.  ##
## ---------------------------------------------------- ##
make  all-recursive
make[8]: Entering directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src/gl/tests'
Making all in .
make[9]: Entering directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src/gl/tests'
make[9]: Nothing to be done for 'all-am'.
make[9]: Leaving directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src/gl/tests'
make[8]: Leaving directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src/gl/tests'
make[7]: Leaving directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src/gl/tests'
make[6]: Leaving directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src/gl'
make[5]: Leaving directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src/gl'
Making all in src
make[5]: Entering directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src'
make  all-am
make[6]: Entering directory '/usr/src/openwrt/build_dir/target-x86_64_musl/gnutls-3.8.0/src'
[...]

PR #21007, opened 2023-05-13
gnutls: load libunistring-optional gnulib module

micmac1:
Since a few days staging_dir/host/share/aclocal/ contains new m4 files (libunistring-base.m4, libunistring-optional.m4, libunistring.m4 etc.) that get applied during autoreconf. This changes the libunistring setup enough that we run into problem (builds fail).

Load the libunistring-optional gnulib module in configure.ac to be able to add/use the new configure switch "--with-included-libunistring".

This is at most a workaround. This should be followed up with gnutls upstream to clean this up.

Maintainer: @nmav 
Compile tested:sdk ath79
Run tested: N/A

Description:
This is regarding the gnutls build failure reported in issue #20969. It was suspected that this is related to gcc-13, but as it turns out it's just a broken interaction between gnutls and updated autoreconf macros.

### DIFF
--- a/libs/gnutls/Makefile
+++ b/libs/gnutls/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gnutls
 PKG_VERSION:=3.8.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_BUILD_FLAGS:=no-mips16
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
@@ -132,6 +132,7 @@ CONFIGURE_ARGS+= \
 	--without-idn \
 	--with-default-trust-store-dir=/etc/ssl/certs/ \
 	--with-included-unistring \
+	--with-included-libunistring \
 	--with-librt-prefix="$(LIBRT_ROOT_DIR)/" \
 	--with-pic \
 	--with-system-priority-file="" \
--- a/libs/gnutls/patches/030-unistring-optional.patch
+++ b/libs/gnutls/patches/030-unistring-optional.patch
@@ -1,0 +1,11 @@
+--- a/configure.ac
++++ b/configure.ac
+@@ -471,6 +471,8 @@ DEFAULT_VALGRINDFLAGS='-q --error-exitco
+ 
+ gl_VALGRIND_TESTS_DEFAULT_NO
+ 
++gl_LIBUNISTRING_OPTIONAL
++
+ dnl Note that g*l_INIT are run after we check for library capabilities,
+ dnl to prevent issues from caching lib dependencies. See discussion
+ dnl in https://bugs.gentoo.org/show_bug.cgi?id=494940 and


mpratt14:
I would recommend first trying to get rid of autoreconf for gnutls build, or secondly try to use the autogen.sh script with gnulib source, and if neither of those things are possible then this would be good

egorenar:
thanks! tested right now and it fixed the problem for me.

1715173329:
```patch
diff --git a/libs/gnutls/Makefile b/libs/gnutls/Makefile
index dc94591c2..262df231e 100644
--- a/libs/gnutls/Makefile
+++ b/libs/gnutls/Makefile
@@ -98,7 +98,9 @@ endef
 define Package/libgnutls
 $(call Package/gnutls/Default)
   TITLE+= (library)
-  DEPENDS+= +libnettle +!LIBNETTLE_MINI:libgmp +GNUTLS_EXT_LIBTASN1:libtasn1 +GNUTLS_PKCS11:p11-kit +GNUTLS_CRYPTODEV:kmod-cryptodev +libatomic
+  DEPENDS+= +libnettle +!LIBNETTLE_MINI:libgmp +GNUTLS_EXT_LIBTASN1:libtasn1 \
+    +GNUTLS_PKCS11:p11-kit +GNUTLS_CRYPTODEV:kmod-cryptodev +libatomic \
+    +libunistring
 endef
 
 define Package/libgnutls-dane
@@ -131,7 +133,7 @@ CONFIGURE_ARGS+= \
 	--enable-local-libopts \
 	--without-idn \
 	--with-default-trust-store-dir=/etc/ssl/certs/ \
-	--with-included-unistring \
+	--with-libunistring-prefix=$(STAGING_DIR) \
 	--with-librt-prefix="$(LIBRT_ROOT_DIR)/" \
 	--with-pic \
 	--with-system-priority-file="" \
```

what about linking to our own libunistring package?

peterwillcn (approved):

mpratt14:
see the existing config `--with-included-unistring` which I believe was supposed to be "libunistring" instead of "unistring" but it is a typo or maybe something changed...

now that #12606 is merged you can try to use the "prefix" config instead of the "included" config in the same way that PR does and see if that works. If you do try it, first make sure that the libunistring package is built first, and to do that you probably need to delete the packages feed Makefile for libunistring because it is now a duplicate with the one in openwrt.git now.

mpratt14:
and I'm just repeating what @1715173329 already suggested haha

nmav:
@micmac1 thank you, it looks like a workaround that addresses the build issue. Have you tried the alternative recommended in the comments above to remove the PKG_FIXUP autoreconf step?

mpratt14:
ok, what if you just link to local libunistring without changing anything else?

see post in the forum
https://forum.openwrt.org/t/build-failure-gnutls-3-7-8-latest-snapshot/159426/14

but instead of deleting the config option try to change it to "with-___-prefix=___"

`./configure --help` after autoreconf runs is useful to see if that config option is available

micmac1:
The --with-libunistring-prefix=... option doesn't work for me. Builds fail with it. If one leaves it out and drops --with-included-unistring gnutls actually links to external libunistring. So exactly like the forum post @mpratt14 mentioned.

But in my opinion doing this would cause some users grief.

This is the size of the ipks when gnutls links against external libunistring:

762K libgnutls_3.8.0-2_mips_24kc.ipk
625K libunistring_1.1-1_mips_24kc.ipk

This is the size of gnutls using the included libunistring:

817K libgnutls_3.8.0-2_mips_24kc.ipk

So linking external libunistring adds 570K. That's a big difference one small devices.

mpratt14:
I'm fine with either way, just wanted to see if that worked

1715173329:
that worked for me though.

so is this ready to go? anyway let's fix the build issue firstly

micmac1:
I tried 1715173329's again and this time it worked. I must've copied something wrongly earlier.

Anyway, I have nothing else to add. I don't have write access in this repo so whether to merge or not is not up to me.

1715173329:
@nmav are you okay with it?

oskarirauta (approved):

oskarirauta:
I can verify that this fixes the issue.

PR #21011, opened 2023-05-14
xtables-addons: bump to 3.24, fix build errors since b2d1eb7

httpstorm:
Use `kmalloc` and remove conflicting `#include <stdarg.h>`, to fix the following build warnings treated as errors since https://github.com/openwrt/openwrt/commit/b2d1eb717b6597e0145c98fc5f9f88a7dc219513

```
error: ISO C90 forbids variable length array 'buf' [-Werror=vla]
error: "va_start" redefined [-Werror]
error: "va_arg" redefined [-Werror]
error: "va_copy" redefined [-Werror]
getstr(s)==NULL is always false
```

Maintainer: @jow-
Compile tested: WRT3200ACM

@Ansuel @aparcar @neheb @pprindeville

### DIFF
--- a/net/xtables-addons/Makefile
+++ b/net/xtables-addons/Makefile
@@ -8,7 +8,7 @@ include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=xtables-addons
 PKG_VERSION:=3.21
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_HASH:=2e09ac129a14f5e9c23b115ebcdfff4aa84e2aeba1268dbdf39b2d752bd71e19
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
--- a/net/xtables-addons/Makefile
+++ b/net/xtables-addons/Makefile
@@ -7,9 +7,9 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=xtables-addons
-PKG_VERSION:=3.21
-PKG_RELEASE:=2
-PKG_HASH:=2e09ac129a14f5e9c23b115ebcdfff4aa84e2aeba1268dbdf39b2d752bd71e19
+PKG_VERSION:=3.24
+PKG_RELEASE:=1
+PKG_HASH:=3e823f71720519ced31c4c7d2bfaf7120d9c01c59a0843dfcbe93c95c64d81c1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://inai.de/files/xtables-addons/
--- a/net/xtables-addons/patches/001-fix-kernel-version-detection.patch
+++ b/net/xtables-addons/patches/001-fix-kernel-version-detection.patch
@@ -1,6 +1,6 @@
 --- a/configure.ac
 +++ b/configure.ac
-@@ -42,7 +42,7 @@ regular_CFLAGS="-Wall -Waggregate-return
+@@ -41,7 +41,7 @@ regular_CFLAGS="-Wall -Waggregate-return
  
  AS_IF([test -n "$kbuilddir"], [
  	AC_MSG_CHECKING([kernel version that we will build against])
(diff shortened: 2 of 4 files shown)


Ansuel:
Kmalloc is wrong please take my patch.

httpstorm:
@aparcar 
Thanks. I'm on it...

Ansuel:
mhhh you had a different solution for va_list ? you stropped the include?

httpstorm:
@aparcar 
> mhhh you had a different solution for va_list ? you stropped the include?

Yes, I think it's better to remove the include. It was the most simple solution. Especially considering that `stdarg.h` is included from two places. Trouble is one place includes the linux version, and the other includes the version from the mvebu compiler, so that's the source of conflict. My change removes the second. The other way would be far more complicated to impossible.
If you like I can remove the include and add my self as contributor? In my patch I also edit the original patches, to keep it simple and less code. Would you prefer two patches: fix and bump, or one that does both?

Ansuel:
@httpstorm think you are using the wrong tag... anyway
can you provide the patch? I don't really like mine with the define so i would like to see your solution! (and ideally also apply that to broadcom-wl

rmandrad:
note that I am also having this issue (c90 .. -Werror=vla)  on a non-mvebu device(ipq807x) ... happy to test the fix when available

httpstorm:
@aparcar 
Below is a link to my first version of the patch. I will restore it, but change to `kcmalloc` as you advised, and keep your comments. Note that I am an embedded-kernel engineer. I created my own nano RTOS, where all API is safe in any context including RAW ISR. So while I know about the limitations in kernel-mode, my experience with Linux kernel is poor. Please analyse the patch carefully after I update this PR. I'll post back in 30-60 minutes.
https://github.com/httpstorm/openwrt.packages/commit/bd3f0de0944042f7ecbd82887bc6ecccc8655e5a

BTW is this LUA code still in use or is it legacy? I remember at some point I saw that JavaScript is used. I do not see a package called `xtables-addons` installed on my router, and I am not sure how to test the changes. I use `xtables` for the tarpit, as this wasn't available in `nft`.

httpstorm:
@Ansuel 
I'm ready. Successful build and flash on WRT3200ACM. Please review my changes!

@ptpt52 
> I am not sure but if this code run in softirq context, GFP_ATOMIC flag should be used in kcalloc

I'm not sure either in what context `byte_array_to_string` is used. Considering it is exported in a table to be called from LUA, I would assume it isn't called from an interrupt handler, rather from user space. I wonder if this code is used at all, please check my previous comment.

Ansuel:
@httpstorm i checked... when the string is pushed LUA operates under lock but not before so ATOMIC can't be used i think..

httpstorm:
@Ansuel 
> and ideally also apply that to broadcom-wl
Please explain!

Ansuel:
@httpstorm would you hate me if I ask you to split this? Tell me if it's problematic for. (and we can skip this) 
The commit is OK with the change, but I would split the package bump from the fixes part

Also I would add in credits me in co developed tag?

httpstorm:
@Ansuel 
From what I see here
https://www.unix.com/man-page/suse/9/KCALLOC/

- GFP_KERNEL - Allocate normal kernel ram. May sleep.
- GFP_ATOMIC - Allocation will not sleep. May use emergency pools. For example, use this inside interrupt handlers.

httpstorm:
@Ansuel 
> @httpstorm would you hate me if I ask you to split this? Tell me if it's problematic for. (and we can skip this)
> The commit is OK with the change, but I would split the package bump from the fixes part

Ha hah, I promise I won't! 😀 I was curious which way is better. My first implementation actually had fix + bump.
Then I merged them, because I had a bump commit with empty body. A couple of refreshes should do the job.

> Also I would add in credits me in co developed tag?
ok, I'm still learning about these rules.

Ansuel:
> ok, I'm still learning about these rules.

thing is confusing since 3 guys proposed a very similar fix and the patch got merged by multiple author ahahah

Ansuel:
@httpstorm since you are rebasing Fixes tag right above the co-developed-by tag (no new line between them)

httpstorm:
@Ansuel 
I'm ready. It was nice working with you! 🍻 I see you merged it just as I was installing the firmware for a final test. What about the `broadcom-wl`?

BTW how can I become part of the OpenWRT team? I like creating fixes every now and then, and for the most part it's pretty easy to get them accepted, since the people here are friendly and guide me to create good PRs. A couple of days ago however one of my fixes introduced an issue, so I prepared an urgent fix, but since I have no write access, to took a day to merge it. At that moment I wish I had write access to resolve it faster. I also see that even people with write access generally allow a couple of days so that others can check their work before they merge it. Nice practice.

PR #21026, opened 2023-05-15
dmx_usb_module: use upstream commit to fix compilation with kernel 6.2+

BKPepe:
Maintainer: no one
Compile tested: N/A, CI will do its job
Run tested: N/A

Description:
Recently, there was pushed patch, which fixed only part to get it working with Linux kernel 6.2+, but it was only local change and it did not use the changes, which are going to be included in the upcoming dmx_usb_module release.

This commit use the commit from the upstream repository: https://github.com/lowlander/dmx_usb_module/commit/eeecf40c3eb3dbdf1981c508080ab3aa61e105e2

Fixes: e15d9f75dac2888b9d863c6f2d1c19637631c1fb ("dmx_usb_module: fix compilation warning")
Reference below this commit ⬆️ inside comments

### DIFF
--- a/libs/dmx_usb_module/patches/100-fix-compilation-warning-wrong-cast.patch
+++ b/libs/dmx_usb_module/patches/100-fix-compilation-warning-wrong-cast.patch
@@ -1,6 +1,23 @@
+From eeecf40c3eb3dbdf1981c508080ab3aa61e105e2 Mon Sep 17 00:00:00 2001
+From: Erwin Rol <erwin@erwinrol.com>
+Date: Fri, 13 Jan 2023 17:57:36 +0100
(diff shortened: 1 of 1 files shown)


Ansuel:
@BKPepe you are missing the new line at the end of the patch so each test is failing

BKPepe:
Yep, noticed that too, but still I want to prepare the build env to be sure that there are no other cases (well and there are... :D more PRs are coming), which I miss while trying to fix it.

BKPepe:
You are right about that! I saw it, but I completely forget about it. Yeah, we should add it here as well.

neheb:
Rebased I think

BKPepe:
Patch refreshed, branch rebased, this should be ready to go.

PR #21027, opened 2023-05-15
openvpn: Add more hotplug events

miska:
Maintainer: @mkrkn 
Compile tested: mvebu, Turris Omnia, TOS 7.0 ~ OpenWrt 22.03
Run tested: mvebu, Turris Omnia, TOS 7.0 ~ OpenWrt 22.03, custom scripts reacting to those events work

Description:

OpenVPN supports more hooks than just 'up' and 'down'. Especially reacting to 'route-up' and 'route-pre-down' events could be important. When routing table changes, it can make sense to adapt firewall, run some tests or change even more routes. This change passes those events to hotplug, so it is easy to react to them without changing configuration files provided by VPN provider.

### DIFF
--- a/net/openvpn/Makefile
+++ b/net/openvpn/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=openvpn
 
 PKG_VERSION:=2.5.8
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


BKPepe:
There are some issues with this pull request and that's why there was created this issue:https://github.com/openwrt/packages/issues/21200 and hopefully, it is going to be fixed with this PR (https://github.com/openwrt/packages/pull/21245).

PR #21046, opened 2023-05-16
arp-scan: fix missing libcap dependency

PolynomialDivision:
arp-scan complains about missing libcap dependency. Add it as package dependency to fix compilation.

Maintainer: @urusha
Compile tested: mt7622
Run tested: no


neheb:
Sounds like a dependency not disabled.

PR #21064, opened 2023-05-17
mariadb: fix compilation with musl 1.2.4

1715173329:
Maintainer: @miska
Run tested: n/a

Description:

Manually pass -D_LARGEFILE64_SOURCE to allow to keep using LFS64 definitions.


1715173329:
whoa, github exploded again

1715173329:
Not sure why "CI_HELPER" variable are missing in test script. It should be exported by entrypoint.sh already

PR #21068, opened 2023-05-17
lttng: Update to 2.13.9

1715173329:
Maintainer: nobody / cc @neheb
Run tested: n/a

Description:
update lttng packages to the latest version and fix build with musl 1.2.4


devel/lttng-modules/Makefile
```diff
@@ -8,37 +8,37 @@
...
 endef
 
 define Build/Compile
-	$(MAKE) -C "$(LINUX_DIR)" \
-		ARCH="$(LINUX_KARCH)" \
-		CROSS_COMPILE="$(TARGET_CROSS)" \
+	+$(KERNEL_MAKE) $(PKG_JOBS) \
 		M="$(PKG_BUILD_DIR)" \
-		V="$(V)" \
+		CONFIG_LTTNG=m \
```
PolynomialDivision, line 41: Why do you add here CONFIG_LTTNG?

1715173329: This is an out-of-tree module, here is the right place to pass Kconfig to the Makefile.

https://github.com/lttng/lttng-modules/blob/da1f5a264fff33fc5a9518e519fb0084bf1074af/Makefile#L32

PolynomialDivision: I see. Thanks!

PR #21069, opened 2023-05-17
perl: fix compilation with musl 1.2.4

robimarko:
Maintainer: @neheb 
Compile tested: ipq807x/master

Description:

_LARGEFILE64_SOURCE has to be defined in the source, or CFLAGS can be used to pass -D_LARGEFILE64_SOURCE to allow to keep using LFS64 definitions.

### DIFF
--- a/lang/perl/Makefile
+++ b/lang/perl/Makefile
@@ -11,7 +11,7 @@ include perlver.mk
 
 PKG_NAME:=perl
 PKG_VERSION:=$(PERL_VERSION)
-PKG_RELEASE:=7
+PKG_RELEASE:=8
 
 PKG_SOURCE_URL:=\
 		https://cpan.metacpan.org/src/5.0 \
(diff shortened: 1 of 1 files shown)


BKPepe (approved):

PolynomialDivision:
Maintainer is @pprindeville

PolynomialDivision (approved):

1715173329 (approved):

neheb:
Perl needs an update to the latest version honestly.

pprindeville:
> Perl needs an update to the latest version honestly.

It absolutely does, but as it gets ported to more and more platforms I have to guess what the correct pre-salted parameters are since it leans heavily on native compilation and run-time detection of the host's characteristics...

What would be great is if we had a CI/CD job that did all of this detection for us, every time a new release came out.

PR #21071, opened 2023-05-17
python-attrs: Update to 23.1.0

jefferyto:
Maintainer: me
Compile tested: armvirt-32, 2023-05-15 snapshot sdk
Run tested: armvirt-32 (qemu), 2023-05-15 snapshot

Description:
The package now uses pyproject.toml-based builds with additional build dependencies.

### DIFF
--- a/lang/python/python-attrs/Makefile
+++ b/lang/python/python-attrs/Makefile
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2016-2018 OpenWrt.org
+# Copyright (C) 2016, 2018-2023 Jeffery To
 #
 # This is free software, licensed under the GNU General Public License v2.
 # See /LICENSE for more information.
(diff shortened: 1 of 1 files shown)


lang/python/python-attrs/Makefile
```diff
@@ -8,16 +8,18 @@
...
 
 PYPI_NAME:=attrs
-PKG_HASH:=626ba8234211db98e869df76230a137c4c40a12d72445c45d5f5b716f076e2fd
+PKG_HASH:=6279836d581513a26f1bf235f9acd333bc9115683f14f7e8fae46c98fc50e015
 
 PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=LICENSE
 PKG_MAINTAINER:=Jeffery To <jeffery.to@gmail.com>
 
+PKG_BUILD_DEPENDS:=python-hatchling/host python-hatch-vcs/host python-hatch-fancy-pypi-readme/host
```
BKPepe, line 21: I have one small nitpick. Would it be possible to describe such change in the commit description? Maybe, you can also add that you are assigning the copyright to yourself.

jefferyto: I can mention in the commit message that the build dependencies have changed though I don't recall seeing this as a usual practice.

AFAIK I've never explicitly assigned copyright to OpenWrt so changing that line is correcting it

jefferyto:
Updated commit message (for https://github.com/openwrt/packages/pull/21071#discussion_r1196180708).

lang/python/python-attrs/Makefile
BKPepe: Much better this time. Yeah, but it is also not a good idea to hide addding more build dependencies in the simple bump commit, so that's why I think, it was and it is still good idea to include it like you did. ;-)

BKPepe (approved):

PR #21077, opened 2023-05-17
xfsprogs: fix compilation with musl 1.2.4

1715173329:

Manually pass -D_LARGEFILE64_SOURCE to allow to keep using LFS64 definitions.

Maintainer: nobody
Run tested: n/a


PolynomialDivision:
Thanks!

PR #21079, opened 2023-05-17
acme: improve

hgl:
Maintainer: @tohojo 
Compile tested: x64 master
Run tested: x64 master

Description:

Please check the commit messages for details. Two changes:

1. Don't let clients deal with deprecated values.
2. Standardize key type values.

### DIFF
--- a/net/acme-common/files/acme.init
+++ b/net/acme-common/files/acme.init
@@ -53,9 +53,9 @@ load_options() {
 	export dns_wait
 
 	config_get webroot "$section" webroot
-	export webroot
 	if [ "$webroot" ]; then
 		log warn "Option \"webroot\" is deprecated, please remove it and change your web server's config so it serves ACME challenge requests from $CHALLENGE_DIR."
+		CHALLENGE_DIR=$webroot
 	fi
 }
 
--- a/net/acme-common/files/acme.init
+++ b/net/acme-common/files/acme.init
@@ -39,8 +39,17 @@ load_options() {
 	export domains
 	export main_domain
 	main_domain="$(first_arg $domains)"
-	config_get keylength "$section" keylength ec-256
-	export keylength
+	config_get keylength "$section" keylength
+	if [ "$keylength" ]; then
+		log warn "Option \"keylength\" is deprecated, please use key_type (e.g., ec256, rsa2048) instead."
+		case $keylength in
+		ec-*) key_type=${keylength/-/} ;;
+		*) key_type=rsa$keylength ;;
+		esac
+	else
+		config_get key_type "$section" key_type ec256
+	fi
+	export key_type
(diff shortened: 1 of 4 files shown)


stokito:
I see. And the keylength was renamed to keytype is now?
How the param is called in the acmesh itself?

I just trying to figure out what was changed  and where it's going

net/acme-common/files/acme.init
```diff
@@ -39,8 +39,17 @@ load_options() {
...
-	export keylength
+	config_get keylength "$section" keylength
+	if [ "$keylength" ]; then
+		log warn "Option \"keylength\" is deprecated, please use key_type (e.g., ec256, rsa2048) instead."
+		case $keylength in
+		ec-*) key_type=${keylength/-/} ;;
+		*) key_type=rsa$keylength ;;
+		esac
+	else
+		config_get key_type "$section" key_type ec256
```
tohojo, line 50: Hmm, if someone actually sets the new value, the old one probably shouldn't override it, should it?

hgl: I assumed the convention was that if the old value exists, a user probably hasn’t completely migrated yet.

If you want to change that to “as soon so the new value comes into being“ I’m fine with that too.

Let me know.

hgl:
> Oh, I see. And the keylength was renamed to keytype is How the param is called in the acmesh itself?
> 
> I just trying to figure out what was changed and where it's going

Old value is how acmesh calls it.

net/acme-common/files/acme.init
tohojo: Hmm, I guess we also do the "old name first, then new" for the `staging` option, so let's just keep it here as well to stay consistent...

tohojo:
Do you actually have another client in mind that will get confused by the current keylength values? Doesn't really seem like one is inherently superior to the other?

Also, luci-app-acme still uses the old config values, not just for this but also some of the other deprecated config names (use_staging and webroot). We should update that as well...

stokito:
@tohojo if you don't have a time I can update the luci-app on next week.

tohojo:
@stokito that would be awesome, thanks! :)

hgl:
> Do you actually have another client in mind that will get confused by the current keylength values?

I'm writing a client using lego, but it's a WIP. Strictly speaking, it won't get confused, since we will write value converter, if the client's supported values are different. I think the point is, if the values have to be converted (which are very likely for non-acmesh clients), let's convert from ones that are more intuitive. Converting from pure numbers to rsa is not a good design IMHO.

> if you don't have a time I can update the luci-app on next week.

Thanks for the help @stokito.

tohojo:
Ping @stokito on the luci update? :)

PR #21081, opened 2023-05-17
pcre/prce2: move pcre from base to here and pcre2 to base

PolynomialDivision:
The package is needed by selinux, so it was moved to the base repository. Remove it from openwrt packages feed.

Maintainer: @InBetweenNames 
Compile tested: mt7622, x86 and CI
Run tested: none


BKPepe:
If you are moving something between repositories, there should be some cross reference in the commit description or at least in the PR (if they are...)

PR #21099, opened 2023-05-19
borgbackup: Move from lang/python into utils

jefferyto:
Maintainer: @julienmalik 
Compile tested: armvirt-32, 2023-05-15 snapshot sdk
Run tested: none

Description:
lang/python is meant for Python libraries and other packages closely related to the Python language. It makes more sense for borgbackup to be in utils instead.

### DIFF
--- a/utils/borgbackup/Makefile
+++ b/utils/borgbackup/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=borgbackup
 PKG_VERSION:=1.2.4
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PYPI_NAME:=borgbackup
 PKG_HASH:=a4bd54e9469e81b7a30a6711423115abc818d9cd844ecb1ca0e6104bc5374da8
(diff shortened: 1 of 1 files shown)


1715173329:
@BKPepe could you review this?

BKPepe:
Sure, thanks for pinging me.

White checking on this website https://repology.org/project/borgbackup/versions , I see many distros are not quite sure, where it belongs. Some of them listed this package as utils, others as admin stuff for backups and one distro put it into lang. 

I agree with Jeffery to put it somewhere else. This is purely, I would say cosmetic and we can probably drop bumping PKG_RELEASE for that to avoid reinstalling it, but for both things, I dont have strong opinion on this.

jefferyto:
> we can probably drop bumping PKG_RELEASE

I checked the control file (of the ipk) and `Section: lang` is in it, that's why I thought it was safer to increment `PKG_RELEASE`. If that's not a concern I can remove the `PKG_RELEASE` change.

julienmalik:
FWIW : https://borgbackup.readthedocs.io/en/stable/internals/frontends.html#all-about-json-how-to-develop-frontends

borg does not provide any stable API at the Python level, so yes I agree it shall be better seen as an util than a python library.

PR #21105, opened 2023-05-19
ipfs-http-client: remove package

1715173329:
This package lacks proper maintenance in both upstream and here.

Maintainer: @LLE8


1715173329:
cc @neheb

BKPepe:
Hey @paper42, can you share with us some statistics from Turris survey? It would be good to know, how many Turris users has this specific package installed.

BKPepe:
I went through Git history and see @LLE8 proposed to add it here.

1715173329:
IRC discussion here:
https://oftc.irclog.whitequark.org/openwrt-devel/2023-05-19#32161857

> is from 2022

No update has been made in upstream since the **beginning** of last year.

Last update for this package is 2 years ago, made by @neheb. @LLE8 is not active on GitHub.

> only one pull request open

There were 40 PRs, most of them were just made by the upstream maintainers themselves.

This package doesn't build for 2 months, and nobody has made a actual PR for fixing it. I guess you don't notice it too.

PolynomialDivision (approved):

BKPepe:
>  I guess you don't notice it too.

Since I am not working for Turris anymore, I don't have access for full builds for each Turris OS version based on different OpenWrt version. Still it is interesting to see that I am still most active than Turris guys, here... :) Thats explain it. I am not using this package and thats why I was curious why it is getting dropped.

PolynomialDivision:
@BKPepe Is that an ack?

1715173329:
Plan to merge it tonight if there's no objection / no valid PR for fixing the build issue.

BKPepe:
You should wait more than 2 days, because at least in Europe, there is a weekend.

1715173329:
I'd say it's already opened for 2 days - and today is weekend anywhere.

BKPepe:
Many people will not have any chance to go through this as it has been here only for 2 days since Friday. It is not even one business day. We can wait a little bit. I understand you want to deal with it in a hurry, but it always ends up very badly.

- Do we have some statistics how many people are using this package?
  - So it will help us to know, if people are using this package or not. 
     - [Turris Survey](https://docs.turris.cz/basics/sentinel/threat-detection/#turris-survey) can provide some statistics from Turris devices, though.

1715173329:
I'm not interested in Turris, here is the upstream OpenWrt. If you want this package in Turris, it should be maintained in Turris too. 
At least I don't ever see you tried to update/fix it.

https://forum.openwrt.org/t/what-is-ipfs-http-client/153937
this is the only post in forum, and the user didn't even know what it is.

1715173329:
also with the comment from ipfs ex-staff:
> ah that used to be my dayjob long time ago :) not this ipfs-http-client package, but ipfs in general. i'd say go for it. it's a package for cpp-ipfs-http-client, which was never "the" ipfs client and looks pretty incomplete. and the package is already >1y out-of-date with upstream

1715173329:
@PolynomialDivision could you please cherry pick this package to 23.05? My development VM is broken currently

BKPepe:
> I'm not interested in Turris, here is the upstream OpenWrt. If you want this package in Turris, it should be maintained in Turris too. At least I don't ever see you try to update/fix it.

Don't get me wrong on this, but we already have this discussion before and it seems like you did not take anything from that. I have not said that this package is used in Turris. Our motivation as open source developers should be that everything ideally is upstreamed to us. Not playing on our own play ground with forked repository, which does not sync with upstream changes and it is getting outdated. 

That's why threads like this: 
 - https://forum.openwrt.org/t/do-any-vendors-work-with-upstream-openwrt/131571
  - https://forum.openwrt.org/t/industry-supporting-openwrt/144578/12
  
are going to take place and still popping out because many vendors and manufacturers have their own repo. 

My biggest worry here was about that many people, inc. developers, could be active during business days, not weekends. 2-3 days to get some packages remove is a too short period to handle it. 

You discussed it somewhere on IRC, hm... imagine that many people are not using that and as well they can not be using snapshots. Since we don't have any statistics, it is just about theory...

PR #21111, opened 2023-05-19
ocserv: add otp config option

THLIVSQAZ:
Maintainer: @nmav
Compile tested: arm cortex-a7 ipq40xx, p2w_r619ac-128m, OpenWrt 22.03
Run tested: arm cortex-a7 ipq40xx, p2w_r619ac-128m, OpenWrt 22.03

Description:
Add option enable OTP feature and also fix build error in 
https://github.com/openwrt/packages/pull/21035#issuecomment-1554311549

when enable OCSERV_LIBOATH :
```
configure:
Summary of build options:
  version:              1.1.7
  Host type:            arm-openwrt-linux-gnu
  Install prefix:       /usr
  Compiler:             arm-openwrt-linux-muslgnueabi-gcc
  CFlags:               -Os -pipe -fno-caller-saves -fno-plt -fhonour-copts -mfloat-abi=hard -fmacro-prefix-map=/mnt/openwrtZone/opmaster/build_dir/target-arm_cortex-a7+neon-vfpv4_musl_eabi/ocserv-1.1.7=ocserv-1.1.7 -Wformat -Werror=format-security -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro  -Wall -Wno-strict-aliasing -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-missing-field-initializers -Wno-implicit-fallthrough -Wno-stringop-truncation
  CWrap testing:        no
  CWrap PAM testing:    no
  CWrap NSS testing:    no

  PAM auth backend:     no
  Radius auth backend:  no
  GSSAPI auth backend:  no
  OIDC Auth backend:    no
  Anyconnect compat:    yes
  TCP wrappers:         no
  namespaces:           yes
  systemd:              no
  (socket activation)
  worker isolation:     none
  Compression:          yes
  LZ4 compression:      no
  readline:             yes
  libnl3:               
  liboath:              yes
  libgeoip:             no
  libmaxminddb:         no
  glibc (sha2crypt):    no
  local talloc:         yes
  local protobuf-c:     no
  local PCL library:    yes
  local http-parser:    no
  seccomp trap:		no
  capture latency stats no
```

Default: 
```
configure:
Summary of build options:
  version:              1.1.7
  Host type:            arm-openwrt-linux-gnu
  Install prefix:       /usr
  Compiler:             arm-openwrt-linux-muslgnueabi-gcc
[...]

PR #21114, opened 2023-05-19
ocserv: disable liboath detection

1715173329:
Disable liboath detection to fix a build error due to missing dependency.
(the liboath library is now detected, but is unncessary.)

Maintainer: @nmav
Run tested: n/a


1715173329:
oh looks like #21111 is better

PR #21129, opened 2023-05-21
ragel: new package to build vectorscan

graysky2:
This is a new package for ragel which is a dependency for another new package vectorscan


graysky2:
My goal is to add vectorscan support to snort3 which should improve performance and reduce CPU use.  To do so, I need to create two new packages:
1. vectorscan
2. ragel (this PR) which is required by vectorscan

PR #21145, opened 2023-05-23
curl: update to 8.1.0

stangri:
Maintainer: me
Compile tested: x86_64, Sophos XG-135r3, OpenWrt 22.03.5
Run tested: x86_64, Sophos XG-135r3, OpenWrt 22.03.5, test resolution from https-dns-proxy


dhewg:
A little late to the party, but what's that gc-sections change about?

PR #21150, opened 2023-05-24
xtables-addons: fix xt_DNETMAP load when netfilter NAT is module

bitthief:
Maintainer: @jow-
Compile tested: ipq807x/Xiaomi AX3600 on snapshot
Run tested: ipq807x/Xiaomi AX3600 on snapshot

Description:
The xt_DNETMAP module fails to load when the netfilter NAT is built as a module.
This fixes the ifdef compile-time check to also account for this scenario.

@Ansuel @pprindeville

### DIFF
--- a/net/xtables-addons/patches/300-xt-DNETMAP-fix-CONFIG-NF-NAT-module-dependency.patch
+++ b/net/xtables-addons/patches/300-xt-DNETMAP-fix-CONFIG-NF-NAT-module-dependency.patch
@@ -1,0 +1,20 @@
+--- a/extensions/xt_DNETMAP.c
++++ b/extensions/xt_DNETMAP.c
+@@ -20,7 +20,7 @@
+ 
+ #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+ #include <linux/module.h>
+-#ifdef CONFIG_NF_NAT
++#if defined(CONFIG_NF_NAT) || defined(CONFIG_NF_NAT_MODULE)
+ #include <linux/inet.h>
+ #include <linux/ip.h>
+ #include <linux/netdevice.h>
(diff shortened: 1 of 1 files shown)


ptpt52:
LGTM

jefferyto:
CI is failing because the patch needs to be refreshed - run `make package/xtables-addons/refresh`.

Also your commit is missing a `Signed-off-by` line, see [Sign your work](https://openwrt.org/submitting-patches#sign_your_work).

Ansuel:
@bitthief can you fix these thing so we can merge this?

bitthief:
> @bitthief can you fix these thing so we can merge this?

@Ansuel @jefferyto 

Done, refreshed/rebased, signed the commit and pushed the fix, would be great if you can merge?

jefferyto:
> Signed-off-by: bitthief <bitthief@protonmail.ch>

A real name is required for `Signed-off-by`.

Ovear:
Hi,

As xt_DNETMAP still does not work in the latest version of OpenWrt and this PR remains unmergred, I reached out to upstream for support. Fortunately, this feature has now been introduced upstream.

Therefore, we might be able to fix this by updating the version of xtables-addon.

PS: the kernel feature checking fixed are also introduced to other xtables-addon modules according upstream's [commit](https://codeberg.org/jengelh/xtables-addons/commit/8e8785f6f9bd9427da7e6836b220dd462c42affb).

feckert:
We are on [3.27](https://github.com/openwrt/packages/commit/b093eea601195184ea1364c4678c35f96269f7f5) since 4 days now and as mentioned by @Ovear this was fixed in upstream with this [commit](https://codeberg.org/jengelh/xtables-addons/commit/8e8785f6f9bd9427da7e6836b220dd462c42affb). I will therefore close this. If it is to be fixed for stable branches, then separate PRs should be opened.

PR #21159, opened 2023-05-24
gstreamer1: update to 1.22.3

flyn-org:
Maintainer: me
Compile tested: x86_64

Description:

* gstreamer1
* gst1-plugins-base
* gst1-plugins-good
* gst1-plugins-bad
* gst1-plugins-ugly
* gst1-libav


flyn-org:
It appears the Meson update (https://github.com/openwrt/openwrt/pull/10543) has not yet made itself into all of the build hosts. This is why some of the checks failed.

PR #21165, opened 2023-05-25
https-dns-proxy: fix time_t logging (#19366)

schmaller:
There is a problem
"related to the change in time_t (32 bits to 64 bits on a 32 bit system)". This patch fixes fprintf() calls by explicitly casting to 64bit integers.

Maintainer: Stan Grishin <stangri@melmac.ca>
Compile tested: mips_24kc, TP-Link Archer C6 v2, r22998-066441b5e4
Run tested: mips_24kc, TP-Link Archer C6 v2, r22998-066441b5e4

### DIFF
--- a/net/https-dns-proxy/patches/030-fix-time_t-logging.patch
+++ b/net/https-dns-proxy/patches/030-fix-time_t-logging.patch
@@ -1,0 +1,13 @@
+--- a/src/logging.c
++++ b/src/logging.c
+@@ -77,8 +77,8 @@ void _log(const char *file, int line, in
+ 
+   struct timeval tv;
+   gettimeofday(&tv, NULL);
+-  fprintf(logf, "%s %8ld.%06ld %s:%d ", SeverityStr[severity], tv.tv_sec,
+-          tv.tv_usec, file, line);
++  fprintf(logf, "%s %8lld.%06lld %s:%d ", SeverityStr[severity], (long long)tv.tv_sec,
++          (long long)tv.tv_usec, file, line);
+ 
+   va_list args;
+   va_start(args, fmt);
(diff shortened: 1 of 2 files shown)


net/https-dns-proxy/patches/030-fix-time_t-logging.patch
```diff
@@ -0,0 +1,13 @@
+--- a/src/logging.c
```
BKPepe, line 1: This patch does not seems to be present in the upstream.

The proper solution for this is to backport this patch as it is:
https://github.com/aarond10/https_dns_proxy/commit/a5c4428ef963a74b8a142a6f9f3a77cf6ee0bdc5

BKPepe (changes requested):

net/https-dns-proxy/patches/030-fix-time_t-logging.patch
stangri: The upstream developed their solution based on findings/patch suggested by @schmaller. 

@schmaller given that your patch was improved on by the original developer, you can (if you want) modify this PR to include their patch or just submit the PR for the newer https-dns-proxy version from [this commit](https://github.com/aarond10/https_dns_proxy/commit/a5c4428ef963a74b8a142a6f9f3a77cf6ee0bdc5). Or just use the version you've built until I update the OpenWrt repo with the newer upstream version.

schmaller: since upstream has been fixed there's no need to patch the package anymore. right?

stangri: @schmaller as far as I understood @aarond10 , this PR was superceded by https://github.com/openwrt/packages/pull/21177

PR #21186, opened 2023-05-26
mstpd: drop package (replaced by ustp in core)

commodo:
Maintainer: me
Compile tested: n/a
Run tested: n/a

OpenWrt core has a package called ustp which is an OpenWrt adaptation (from mstpd) for OpenWrt (using libubox, libubus, etc).

No sense in keeping mstpd anymore.
We can just update ustp.
Also, if mstpd has any updates, they can be ported over to ustp too.

Abandoned PR:
  https://github.com/openwrt/packages-abandoned/pull/30

Signed-off-by: Alexandru Ardelean <alex@shruggie.ro>


BKPepe:
I'm thinking if this package is removed, shouldn't be in the package makefile ``ustp`` provides for this ``mstpd`` package?

commodo:
> I'm thinking if this package is removed, shouldn't be in the package makefile `ustp` provides for this `mstpd` package?

umm, could be;
no idea about this;

@nbd168  thoughts ?

nbd168:
I don't think it needs a "provides" entry for mstpd, since nothing in packages depends on it.

PR #21190, opened 2023-05-26
python3: Split venv into separate package, pip and setuptools into separate source packages

jefferyto:
Maintainer: me, @commodo for python-pip-conf
Compile tested: armvirt-32, 2023-05-21 snapshot sdk
Run tested: armvirt-32 (qemu), 2023-05-21 snapshot

Description:
* Split venv into a separate package (python3-venv)
  * It makes more sense to ship ensurepip in this package rather than with pip
* Split pip and setuptools into separate source packages (python-pip, python-setuptools)
  * Host versions continue to be installed as part of host Python
  * Target versions can follow upstream
  * Reduces complexity of the python3 Makefile
* Include pip.conf with python3-pip instead of in a separate package (python-pip-conf)
  * The separate package made sense when Python 2 and 3 could be installed together, now there is no need


commodo:
> * Split pip and setuptools into separate source packages (python-pip, python-setuptools) 

should be fine these days; it's probably less hassle to maintain with just a Python3 (and no Python2) package

>   * Reduces complexity of the python3 Makefile

yes, but setuptools & pip may occasionally break;
though (again) it should be less breakage now with just a single Python3 version; 

> * Include pip.conf with python3-pip instead of in a separate package (python-pip-conf)
> 
>     * The separate package made sense when Python 2 and 3 could be installed together, now there is no need

ah; forgot about this; good point about removing it;

commodo (approved):

sevmonster:
I am not seeing `python3-venv` package on the index. Is there a problem with the build? It looks like the [makefile](https://github.com/openwrt/packages/blob/master/lang/python/python3/files/python3-package-venv.mk) is still there.

jefferyto:
@sevmonster what arch/target are you using?

sevmonster:
> @sevmonster what arch/target are you using?

I checked updates for [Linksys E8450](https://openwrt.org/toh/linksys/e8450) and all the indexes [on the site](https://openwrt.org/packages/start), and didn't see it.

jefferyto:
> I checked updates for [Linksys E8450](https://openwrt.org/toh/linksys/e8450) and all the indexes [on the site](https://openwrt.org/packages/start), and didn't see it.

https://downloads.openwrt.org/snapshots/packages/aarch64_cortex-a53/packages/python3-venv_3.11.5-1_aarch64_cortex-a53.ipk

Did you actually try to install the package on your device (`opkg install python3-venv`) or did you just search the wiki?

sevmonster:
> The venv package was added for master and openwrt-23.05, if you are using 22.03 or older then it isn't available.

That makes sense, thank you. I did not upgrade to the latest RC or snapshot. I suppose the package indexes are for the latest stable release.

PR #21194, opened 2023-05-27
python-greenlet: Update to 2.0.2

1715173329:
Fix build with Python 3.11.

Maintainer: @ja-pa / cc @commodo 
Run tested: n/a

### DIFF
--- a/lang/python/python-greenlet/Makefile
+++ b/lang/python/python-greenlet/Makefile
@@ -8,11 +8,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-greenlet
-PKG_VERSION:=1.1.2
-PKG_RELEASE:=2
+PKG_VERSION:=2.0.2
+PKG_RELEASE:=1
 
 PYPI_NAME:=greenlet
-PKG_HASH:=e30f5ea4ae2346e62cedde8794a56858a67b878dd79f7df76a0767e356b1744a
+PKG_HASH:=e7c8dc13af7db097bed64a051d2dd49e9f0af495c26995c00a9ee842690d34c0
 
 PKG_MAINTAINER:=Jan Pavlinec <jan.pavlinec1@gmail.com>
 PKG_LICENSE:=MIT


commodo:
i wonder how i missed this for so long;

thanks :)

lang/python/python-greenlet/Makefile
```diff
@@ -8,11 +8,11 @@
...
-PKG_RELEASE:=2
+PKG_VERSION:=2.0.2
+PKG_RELEASE:=1
 
 PYPI_NAME:=greenlet
-PKG_HASH:=e30f5ea4ae2346e62cedde8794a56858a67b878dd79f7df76a0767e356b1744a
+PKG_HASH:=e7c8dc13af7db097bed64a051d2dd49e9f0af495c26995c00a9ee842690d34c0
 
 PKG_MAINTAINER:=Jan Pavlinec <jan.pavlinec1@gmail.com>
 PKG_LICENSE:=MIT
```
commodo, line 18: since doing this PR, can you also remove these 2 lines below?
```
# FIXME: remove when GCC10 is the oldest supported compiler, or the issue goes away
PKG_BUILD_FLAGS:=no-mips16
```

commodo: oh; and also the lines lower:
```
# FIXME: remove when GCC10 is the oldest supported compiler, or the issue goes away
# This is required in addition to PKG_BUILD_FLAGS:=no-mips16 because otherwise MIPS16
# flags are inherited from the Python base package (via sysconfig module)
ifdef CONFIG_USE_MIPS16
TARGET_CFLAGS += -mno-mips16 -mno-interlink-mips16
endif
```

feel free to ignore my request;
i can do this later;

thanks again :)

commodo (approved):

lang/python/python-greenlet/Makefile
1715173329: Emmm in this version it seems failed again with mips16 instructions.
Ref https://github.com/openwrt/packages/actions/runs/5098008403

commodo: hmm; ok

i will try later maybe

Issue #20949, opened 2023-05-07
libwebsockets: fails to build with gcc-13


oskarirauta:
but adding `TARGET_CFLAGS += -Wno-error=enum-int-mismatch` to Makefile allows building..

```
...
PKG_BUILD_DEPENDS:=libubox

CMAKE_INSTALL:=1

include $(INCLUDE_DIR)/package.mk
include $(INCLUDE_DIR)/cmake.mk

TARGET_CFLAGS += -Wno-error=enum-int-mismatch

CMAKE_OPTIONS += -DLWS_IPV6=$(if $(CONFIG_IPV6),ON,OFF)
CMAKE_OPTIONS += -DISABLE_WERROR=ON
CMAKE_OPTIONS += -DCMAKE_BUILD_TYPE=Release
CMAKE_OPTIONS += -DLWS_WITH_EXTERNAL_POLL=ON
CMAKE_OPTIONS += -DLWS_WITH_ULOOP=ON

...
```

oskarirauta:
@blogic

karlp:
Please track upstream's fixes for this as well, looks like https://github.com/warmcat/libwebsockets/pull/2824 is the path they're taking.

oskarirauta:
FYI

https://github.com/openwrt/packages/issues/20961#issuecomment-1538111054

similar issue fixed like this.
But surely libwebsockets comes up with a new release with proposed fix soon..

oskarirauta:
@karlp warm cat/libwebsockets#2824 is not sufficient. I tried it. It works on no-ssl variant, but fails on mbedtls variant. Probably with openssl as well.

oskarirauta:
PR #21206 is available

karlp:
closed as superseded

PR #21206, opened 2023-05-28
libwebsockets: Fix -Werror=enum-int-mismatch in lws_tls_server_abort_connection()

oskarirauta:
Maintainer: @karlp

Description:

with gcc 13 libwebsockets fails with -Werror=enum-int-mismatch with both SSL variants. These 2 patches work out that issue. Problem is mentioned as a issue in openwrt/packages#20949 and instead of overriding issue with appropriate CFLAGS, this approach was suggested for me as it's been made as a PR (for openssl) at warmcat/libwebsockets#2824

Second patch for mbedtls, was made by me using same approach as was used of openssl.

Eventually these propably will be merged into libwebsockets mainstream.


oskarirauta:
@karlp

ping

PR #21213, opened 2023-05-28
speedtestcpp: add new package

oskarirauta:
Maintainer: me

Description:
speedtestcpp is a fork of Taganaka's speedtest, rewritten.

It has some improvements such as
 - interactive result show
 - use server recommended profiles, which makes it faster (can be disabled)
 - and more..

It also provides it's functions in shared and static libraries and offers development headers for exposing speedtest features for another projects.

I have some plans to provide client that uses ubus and does periodical tests; these details could be used to whatever user would want to use them for. My usage would be displaying periodically updated Speedtest results on a lcd4linux display..

This replaces package speedtestpp because this version has all the same features + fixes and improvements, and is actively developed.


1715173329:
> speedtestcpp is a fork of Taganaka's speedtest

so i think you can update the existing speedtestpp package to your fork directly ;)

oskarirauta:
My work indeed has all the features + some more, so it can replace previous package.. Hold a minute..

net/speedtestcpp/Makefile
```diff
@@ -0,0 +1,66 @@
...
+  shared library that provides support for ookla's speedtest
+endef
+
+define Package/speedtestcpp
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=SpeedTest++
+  DEPENDS:=+libspeedtestcpp +libstdcpp
+  URL:=https://github.com/oskarirauta/speedtestcpp
+  CONFLICTS:=speedtestpp
```
BKPepe, line 35: As it was pointed by @1715173329, your forked repository has a few features than original repository. There should be ``PROVIDES``, not ``CONFLICTS``.

net/speedtestcpp/Makefile
```diff
@@ -0,0 +1,66 @@
...
+define Package/libspeedtestcpp
+  SECTION:=libs
+  CATEGORY:=Libraries
+  TITLE:=library for ookla's speedtest
+  DEPENDS:=+libcurl +libstdcpp
+  URL:=https://github.com/oskarirauta/speedtestcpp
+endef
+
+define Package/libspeedtestcpp/description
+  shared library that provides support for ookla's speedtest
```
BKPepe, line 26:
```suggestion
  Shared library that provides support for ookla's speedtest.
```

net/speedtestcpp/Makefile
```diff
@@ -0,0 +1,66 @@
...
+  TITLE:=SpeedTest++
+  DEPENDS:=+libspeedtestcpp +libstdcpp
+  URL:=https://github.com/oskarirauta/speedtestcpp
+  CONFLICTS:=speedtestpp
+endef
+
+define Package/speedtestcpp/description
+  Yet another unofficial speedtest.net client cli interface
+  forked from taganaka's SpeedTest with few improments and
+  lesser depends.
```
BKPepe, line 41: Did you try to send it to the original repository? In this case, taganaka's SpeedTest repository?

BKPepe (commented):
We do already have multiple packages related to SpeedTest or speed network measurements, if I should be correct. What I wouldn't say I like that we need to add to the repository multiple packages, which are just forked from the original repository without any attempt to upstream it.

I appreciate what you do, but I do not think that it is a good way to include every package, which you have forked and improved it somehow. Why it should be accepted to this repository? How long would the development be active? How many users will use this package? Don't take me wrong on this one, but I just want to avoid cases when:

- We will accept the packages, the development goes down and we will remove it after 3-4 months

- The original repository will be more active and we will need to do the switch vice versa again.

- Why it is not possible to use already existing network measurements, which we do have here?

We should rather focus more on quality instead of quantity because there are not many active reviewers.

oskarirauta:
@BKPepe

Yes. Taganaka does not maintain his work anymore, it was already patched by 3rd party patch to get rid of libxml depency. I did not try to send it to him, because I think my fork should indeed be a fork as it deviates in that it provides it's functionality in shared library, why I also renamed it. But I did try to contact him few months back without answer, there are also some PRs, but he seems to have vanished in 2021.

Version that I made, follows more the _official_ cli client written in Python as Taganaka's work is a bit out-dated. Partially. I was not able to re-construct actual test for speeds to identical as I have always tried to avoid scripting languages such as Python, as much as possible and I am not fan of python; couldn't decipher that, at least yet. So like taganaka's and so many other's, this uses secondary method. Go version also is using this method.

I am not fully happy on my work yet either; and plan to streamline it in the future. Not the app so much, but the library. There actually isn't so many existing speedtesters (using Ookla) in our repository, there is _the official one_, but it depends on python - why I originally added Taganaka's version. In my experience, it's actually more often that python software breaks and fails, than c/c++ when something it depends on is updated. Sometimes also no one has maintained python software in ages, and you even need to use old version of python, such as 2.x, as major components might not be available for newer version and still need to modify it heavily to get it to work if components are not available anymore at anywhere, which sometimes does happen. Go, is similar and also creates quite a large binaries. I attempt to avoid python and such to this reason for longevity. Library is there to provide anyone to use results in wherever needed.

In my description I told why I need it, but have made it available for everyone to use it as they please. Since I found this necessary, I can't wonder if someone else wouldn't and as I share almost all my code that isn't for sale, for purposes to avoid people to not need to re-invent the wheel, I decided to do this - and if anyone ever wants to contribute, it's always welcome - even adopting whole project is available option in those cases.

Using exec in separate thread to run python script, or even a Go program is something that you always should avoid if possible. I haven't searched but I guess there are no ubus servers written in Python. Testing also takes some time, especially on lower speeds, where this gives whole lot of new problems. Quality is hard to measure if you take these details in account, one's preferring to python, etc and have hard time with c/c++/etc of course see this from other perspective. Quantity is not increased from the current as like @1715173329 proposed, this should replace Taganaka's version, which isn't bad at all as it works just fine also, if you forget those few options that never were implemented.

net/speedtestcpp/Makefile
oskarirauta: Or should there be no PROVIDES and CONFLICTS both at all?

net/speedtestcpp/Makefile
oskarirauta: No. Past PR's are not under discussion either. User seems to have vanished. I attempted to contact with e-mail with poor results.

oskarirauta:
@1715173329 ?

1715173329 (approved):

oskarirauta:
version bumped.

BKPepe:
Following up on this pull request after some time, as it seems to have stalled. We added the package, but since the initial update in https://github.com/openwrt/packages/pull/22569, there hasn't been any activity.

Are there any plans for a major bump to the latest upstream release (https://github.com/oskarirauta/speedtestcpp/releases)? According to [Repology](https://repology.org/project/speedtestcpp/versions), we are currently the only distribution packaging this. Are there any plans to expand support to other operating systems as well?

oskarirauta:
There has been updates when ever it has been necessary - are you requesting a major version bump without changes to actual code? If it works, I don't see a reason to update it, there hasn't been also changes to Ookla's API, so I don't see what is the feature that you are requesting. More fancy graphs?

Software works fine on other platforms as well, but it has been designed for Openwrt in mind. Calculating internet speed on router that is connected directly to WAN gives best results as there are as little of bottle necks as possible.

But no, I have not updated the package on openwrt packages repository, I decided to resign after being a contributor for over 2 decades as it became too difficult and vision of where Openwrt should be going diverged too much from vision that I had been following for ages.

Anyway, I still keep developing to 2 Linux platforms which are Openwrt and Alpine Linux; as others are overbloated and non-console oriented, if I need a browser, my choice is Safari on OS X, and that platform most likely will be only platform which I ever will be contributing anything graphics related.

So you do you and remove it - like you said, Openwrt supports too much in your opinion and I no longer participate on upstream levels, there is no need for a contributor or developer in such society when ideology has this much difference.

All the best and so-on.
Your's truly, Oskari Rauta.

PR #21233, opened 2023-05-31
ddns: Prevent clearing of desec.io entries

bateast:
When using both ipv4 and ipv6 entries on the same host, ddns is clearing A (or AAAA) record depending on the connection (ipv4 or ipv6)

see https://desec.readthedocs.io/en/latest/dyndns/update-api.html#determine-ip-addresses

Description:


hardfalcon:
Shouldn't this also be updated from HTTP to HTTPS URLs?

Issue #21200, opened 2023-05-27
openvpn: new hotplug events breaks server mode configs


mtpi:
Please make sure that the issue subject starts with `<package-name>: `

Also make sure that the package is maintained in this repository and not in base which should be submitted at https://bugs.openwrt.org or in the LuCI repository which should be submitted at https://github.com/openwrt/luci/issues.

Issues related to releases below 18.06 and forks are not supported or maintained and will be closed.

# Issue template (remove lines from top till here)

Maintainer: @miska (not sure about the maintainer)
Environment: OpenWrt SNAPSHOT (r23104-ef98dc3b3e) on Linksys WRT3200ACM

Description:
On the latest snapshot build i cannot start anymore my openvpn server.
The system log gives this error:
`Sat May 27 19:21:19 2023 daemon.err openvpn(server_tun)[10827]: Options error: --ipchange cannot be used with --mode server (use --client-connect instead)
Sat May 27 19:21:19 2023 daemon.warn openvpn(server_tun)[10827]: Use --help for more information.`
The config for this istance is a simple server config.
I think the error has been introduces by [this commit](https://github.com/openwrt/packages/commit/f8a8b71e26b9bdbf86fbb7d4d1482637af7f3ba4). This introduces the hotplug event for ipchange that is not supported for server istances, but only for client ones.

BKPepe:
cc: @miska ⬆️

ptpt52:
same issue for my openvpn server setup.

BKPepe:
@neheb I think we should revert this until someone has time to look at it. We can not and we really don't want to have a broken OpenVPN server.

PR #21245, opened 2023-05-31
openvpn: fix ipchange hotplug event

AuthorReflex:
Maintainer:  @mkrkn 
Compile tested: ramips/mt7621
Run tested:  ramips/mt7621 Xiaomi MI router 3 Pro

Description:
In f8a8b71e26b9bdbf86fbb7d4d1482637af7f3ba4 introduced new hotplug events. For server config, ipchange hotplug event produces an error. So, make ipchange hotplug event for client only

Fixes https://github.com/openwrt/packages/issues/21200

### DIFF
--- a/net/openvpn/files/openvpn.init
+++ b/net/openvpn/files/openvpn.init
@@ -140,10 +140,11 @@ openvpn_get_credentials() {
 openvpn_add_instance() {
 	local name="$1"
 	local dir="$2"
-	local conf="$3"
+	local conf=$(basename "$3")
 	local security="$4"
 	local up="$5"
 	local down="$6"
+	local client=$(grep -qEx "client|tls-client" "$dir/$conf" && echo 1)
 
 	procd_open_instance "$name"
 	procd_set_param command "$PROG"	\
@@ -155,7 +156,7 @@ openvpn_add_instance() {
 		--down "/usr/libexec/openvpn-hotplug down $name" \
 		--route-up "/usr/libexec/openvpn-hotplug route-up $name" \
 		--route-pre-down "/usr/libexec/openvpn-hotplug route-pre-down $name" \
-		--ipchange "/usr/libexec/openvpn-hotplug ipchange $name" \
+		${client:+--ipchange "/usr/libexec/openvpn-hotplug ipchange $name"} \
 		${up:+--setenv user_up "$up"} \
 		${down:+--setenv user_down "$down"} \
 		--script-security "${security:-2}" \
--- a/net/openvpn/Makefile
+++ b/net/openvpn/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=openvpn
 
 PKG_VERSION:=2.5.8
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_URL:=\
 	https://build.openvpn.net/downloads/releases/ \


turtlewax:
When executing this new script to start/restart openvpn on the CLI of openwrt, I observe an "sh: out of range" error message.  I did some basic debugging and see where it's occurring (below).  Some kind of variable value comparison error going on here?

```
+ get_openvpn_option /etc/openvpn/client.ovpn client client
+ local 'config=/etc/openvpn/client.ovpn'
+ local 'variable=client'
+ local 'option=client'
+ sed -rne 's/^[ \t]*client[ \t]+'"'"'([^'"'"']+)'"'"'[ \t]*$/\1/p' /etc/openvpn/client.ovpn
+ tail -n1
+ local 'value='
+ '[' -n  ]
+ sed -rne 's/^[ \t]*client[ \t]+"(([^"\\]|\\.)+)"[ \t]*$/\1/p' /etc/openvpn/client.ovpn
+ sed -re 's/\\(.)/\1/g'
+ tail -n1
+ value=
+ '[' -n  ]
+ sed -rne 's/^[ \t]*client[ \t]+(([^ \t\\]|\\.)+)[ \t]*$/\1/p' /etc/openvpn/client.ovpn
+ tail -n1
+ sed -re 's/\\(.)/\1/g'
+ value=
+ '[' -n  ]
+ return 1
+ '['  -eq 0 ]
sh: out of range
```

AuthorReflex:
> When executing this new script to start/restart openvpn on the CLI of openwrt, I observe an "sh: out of range" error message. I did some basic debugging and see where it's occurring (below). Some kind of variable value comparison error going on here?
> 
> ```
> + get_openvpn_option /etc/openvpn/client.ovpn client client
> + local 'config=/etc/openvpn/client.ovpn'
> + local 'variable=client'
> + local 'option=client'
> + sed -rne 's/^[ \t]*client[ \t]+'"'"'([^'"'"']+)'"'"'[ \t]*$/\1/p' /etc/openvpn/client.ovpn
> + tail -n1
> + local 'value='
> + '[' -n  ]
> + sed -rne 's/^[ \t]*client[ \t]+"(([^"\\]|\\.)+)"[ \t]*$/\1/p' /etc/openvpn/client.ovpn
> + sed -re 's/\\(.)/\1/g'
> + tail -n1
> + value=
> + '[' -n  ]
> + sed -rne 's/^[ \t]*client[ \t]+(([^ \t\\]|\\.)+)[ \t]*$/\1/p' /etc/openvpn/client.ovpn
> + tail -n1
> + sed -re 's/\\(.)/\1/g'
> + value=
> + '[' -n  ]
> + return 1
> + '['  -eq 0 ]
> sh: out of range
> ```
What is the value of the parameter ```client``` in your  /etc/openvpn/client.ovpn?
I tested different options for fix this error, cannot reproduce 
Then I will commit another version of check

turtlewax:
Simply the word “client” on a single line by itself in my config file.  Per the OpenVPN 2.5 reference manual:

```
--client
A helper directive designed to simplify the configuration of OpenVPN's client mode. 

This directive is equivalent to:
pull
tls-client
```

AuthorReflex:
> Simply the word “client” on a single line by itself in my config file. Per the OpenVPN 2.5 reference manual:
> 
> ```
> --client
> A helper directive designed to simplify the configuration of OpenVPN's client mode. 
> 
> This directive is equivalent to:
> pull
> tls-client
> ```

Check modified, local test shows that the issue should be fixed

turtlewax:
I still see the error in the same place unfortunately when starting both an OpenVPN client and OpenVPN server configuration:

```
+ get_openvpn_option /etc/openvpn/client.ovpn client client
+ local 'config=/etc/openvpn/client.ovpn'
+ local 'variable=client'
+ local 'option=client'
+ sed -rne 's/^[ \t]*client[ \t]+'"'"'([^'"'"']+)'"'"'[ \t]*$/\1/p' /etc/openvpn/client.ovpn
+ tail -n1
+ local 'value='
+ '[' -n  ]
+ sed -rne 's/^[ \t]*client[ \t]+"(([^"\\]|\\.)+)"[ \t]*$/\1/p' /etc/openvpn/client.ovpn
+ tail -n1
+ sed -re 's/\\(.)/\1/g'
+ value=
+ '[' -n  ]
+ sed+  -rnetail 's/^[ \t]*client[ \t]+(([^ \t\\]|\\.)+)[ \t]*$/\1/p' -n1 /etc/openvpn/client.ovpn

+ sed -re 's/\\(.)/\1/g'
+ value=
+ '[' -n  ]
+ return 1
+ '['  -eq 0 ]
sh: out of range
```

I know I have the newer version as I see your latest changes:
`[ "x$client" != x -a "x$client" = "x0" ]  && unset client`

Here is a snippet from my client config file showing how/where client stanza appears:

```
script-security 2
client
dev tun0
resolv-retry infinite
nobind
persist-key
persist-tun
```

Again, I also see the error when restarting a server, which has no client stanza present.  Thanks for your continued efforts.

net/openvpn/files/openvpn.init
```diff
@@ -144,6 +144,7 @@ openvpn_add_instance() {
 	local security="$4"
 	local up="$5"
 	local down="$6"
+	local client="$7"
```
ptpt52, line 147: on some case we have to check the config file, to determine it is client or not.
use code:
```
local client=$(grep -q "mode.*server" "/var/etc/$conf" || echo 1)$7
```

net/openvpn/files/openvpn.init
```diff
@@ -144,6 +144,7 @@ openvpn_add_instance() {
 	local security="$4"
 	local up="$5"
 	local down="$6"
+	local client=$(grep -q "mode.*server" "/var/etc/$conf" || echo 1)
```
ptpt52, line 147: maybe we have to check `server .*` line
```
local client=$(grep -q "mode.*server" "/var/etc/$conf" || grep -q "^server .*" "/var/etc/$conf" || echo 1)
```
on some case, `mode server` could be missing but `server .*` exist in config file.

ptpt52:
I am not really sure about it.
it is 100% that all the client config files contain `client` line? 
if that is the case, check the `client` line in config is good idea.

net/openvpn/files/openvpn.init
```diff
@@ -140,10 +140,11 @@ openvpn_get_credentials() {
 openvpn_add_instance() {
 	local name="$1"
 	local dir="$2"
-	local conf="$3"
+	local conf=$(basename "$3")
```
ptpt52, line 143: great! but it is good to use `$dir/$conf` since it is name not path in var `conf`
so basename is not needed, I think.

ptpt52: oops. looks like you are correct. since there is case where conf is absolute path.

ptpt52 (commented):
LGTM

PR #21251, opened 2023-06-01
zabbix: Add "oldstable" source URL

jefferyto:
Maintainer: @champtar 
Compile tested: N/A
Run tested: N/A

Description:
Zabbix moved the 6.2 directory from "stable" into "oldstable". This adds the "oldstable" URL to `PKG_SOURCE_URL`.

### DIFF
--- a/admin/zabbix/Makefile
+++ b/admin/zabbix/Makefile
@@ -12,7 +12,8 @@ PKG_VERSION:=6.2.3
 PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://cdn.zabbix.com/zabbix/sources/stable/6.2/
+PKG_SOURCE_URL:=https://cdn.zabbix.com/zabbix/sources/stable/$(basename $(PKG_VERSION))/ \
+	https://cdn.zabbix.com/zabbix/sources/oldstable/$(basename $(PKG_VERSION))/
 PKG_HASH:=2be7e57fb33a55fee71480598e317ffa6a8ee5a39639a7e1b42b2ea6872107b5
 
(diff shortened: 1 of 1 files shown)


admin/zabbix/Makefile
```diff
@@ -12,7 +12,8 @@ PKG_VERSION:=6.2.3
 PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://cdn.zabbix.com/zabbix/sources/stable/6.2/
+PKG_SOURCE_URL:=https://cdn.zabbix.com/zabbix/sources/stable/6.2/ \
```
champtar, line 15: Why keep the broken url ?

jefferyto: If the package is updated to 6.4 (the current stable) then the stable URL would apply.

1715173329:
```suggestion
PKG_SOURCE_URL:=https://cdn.zabbix.com/zabbix/sources/stable/$(basename $(PKG_VERSION))/ \
```

This looks better.

jefferyto:
Changes:
* Use `$(basename $(PKG_VERSION))` in URLs
* Rebase on master

admin/zabbix/Makefile
jefferyto: > This might be better.

I have made this change for both URLs, though I prefer the way it is done in [kernel-version.mk](https://github.com/openwrt/openwrt/blob/abcb30d36cfe65e3ed7786c929c1a2350dd2a9c2/include/kernel-version.mk#L46).

jefferyto:
Perhaps this can be merged for now and the URLs updated later if Zabbix comes back with a response?

PR #21258, opened 2023-06-02
cudy-bdinfo: add package

blocktrron:
Maintainer: me
Compile tested: mediatek-filogic / master
Run tested: mediatek-filogic / master / Cudy WR3000

Description:

This program can be used to obtain information stored on the bdinfo parition found on routers from Shenzhen Cudy Technology.

While this tool is not necessary for operation with OpenWrt, it can be helpful to decrypt information generated by the stock firmware as well as to obtain additional device-specific information.

### DIFF
--- a/utils/cudy-bdinfo/Makefile
+++ b/utils/cudy-bdinfo/Makefile
@@ -1,0 +1,42 @@
+#
+# Copyright (C) 2023 David Bauer <mail@david-bauer.net>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=cudy-bdinfo
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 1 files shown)


utils/cudy-bdinfo/Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+PKG_MAINTAINER:=David Bauer <mail@david-bauer.net>
+PKG_LICENSE:=GPL-2.0-or-later
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/cudy-bdinfo
+  SECTION:=utils
+  CATEGORY:=Utilities
+  DEPENDS:=+libopenssl
```
BKPepe, line 27: By default, this is going to be available for all routers supported within OpenWrt. I dont think this is desired behavior. You should add here also depends on the specific target or even subtarget for Cudy routers.

neheb: Interesting that openssh is used instead of mbedtls. The latter is default with OpenWrt.

blocktrron: @neheb I was not able to decrypt the content using mbedtls. I won't rule out this is possible though.

PR #21262, opened 2023-06-03
btop: add libatomic dependency for riscv64

1715173329:
Fixes the following build error:
```
Package btop is missing dependencies for the following libraries:
libatomic.so.1
```

Maintainer: me

### DIFF
--- a/admin/btop/Makefile
+++ b/admin/btop/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=btop
 PKG_VERSION:=1.2.13
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


neheb:
how strange that it's riscv only...

1715173329:
Not sure why... and mpd metadata etc. also has this issue: https://downloads.openwrt.org/snapshots/faillogs/riscv64_riscv64/packages/mpd/mini/compile.txt

And netifyd php8 etc. look like missing -latomic flag
https://downloads.openwrt.org/snapshots/faillogs/riscv64_riscv64/packages/netifyd/compile.txt

neheb:
Hmmm should be implicit with boost. I'll check later.

1715173329:
[the link](https://blog.jiejiss.com/A-RISC-V-gcc-pitfall-revealed-by-a-glibc-update) metioned by @mhei looks good, maybe we should backport that gcc fix?

1715173329:
fixed by https://github.com/openwrt/openwrt/commit/7b4a966de88aa0e1f9b7faa62a4d6cb7b01e9f8f

PR #21268, opened 2023-06-04
packr: remove package

1715173329:
Maintainer: @jefferyto 

Description:
In Go 1.18+ embedded files is natively supported, so this package is useless now. It has been archived on Jun 2022 and failed to build on the riscv64 target.


jefferyto (approved):
Thanks!

Issue #17352, opened 2021-12-16
packages broken with Python 3.10 because of collections.abc


paper42:
Maintainer: @dddaniel @BKPepe @jmarcet
Environment: master

Description:

Python 3.10 removed some deprecated names from collections, so we should either update packages which use them or import upstream patches (which often can't be applied because versions in master are too old, 3/4 of these packages didn't get a single update since they were added o this repository):

- [ ] python-awscli @dddaniel 
https://github.com/aws/aws-cli/commit/0c9c08856bff1e60cb14ea9a2fdfa640c9af9eeb
- [ ] python-botocore @dddaniel
https://github.com/boto/botocore/commit/ebd12d8d539b2135db32c84b94f33319179e2eb3
- [x] python-dns (dnspython) @BKPepe
https://github.com/rthalley/dnspython/commit/62d3d41382d919af4662c12876f3de6ca854fd54
- [x] python-stem @jmarcet
https://github.com/torproject/stem/commit/36bcb170ba9097885902513640075eac2e6ce38

BKPepe:
>  python-dns (dnspython) @BKPepe

I tried a few examples mentioned on [their website](https://www.dnspython.org/examples/) and it seems that it works. :thinking: 

```
root@omnia:~# opkg install python3-dns
Installing python3-dns (1.16.0-3.10-2) to root...
Downloading https://repo.turris.cz/crashlab/omnia/packages/packages/python3-dns_1.16.0-3.10-2_arm_cortex-a9_vfpv3-d16.ipk
Configuring python3-dns.
root@omnia:~# python3
Python 3.10.0 (default, Dec 29 2021, 00:36:57) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import dns.resolver
>>> n = dns.reversename.from_address("127.0.0.1")
>>> print(n)
1.0.0.127.in-addr.arpa.
>>> print(dns.reversename.to_address(n))
127.0.0.1
```

But I will update it anyway shortly. Thanks for letting me know.

BKPepe:
PR sent - https://github.com/openwrt/packages/pull/17458

jefferyto:
Patch backported for python-stem in #17460.

commodo:
closing;

no idea if this is still an issue; 

we're at Python 3.13 in 25.12 and 3.14 in master

PR #21293, opened 2023-06-06
fixing aws-cli package solving issue #17352

UMB8998:
Maintainer: @UMB8998
Compile tested: ARCH64, GlobalScale Mochabin, Openwrt-23.05
Run tested: ARCH64, GlobalScale Mochabin, Openwrt-23.05
The repository compiled with success and after installing in the system we where able to run the command 'aws --version'

Description:
This fixes the issue #17352 that happens after python update


lang/python/python-awscli/Makefile
```diff
@@ -1,13 +1,19 @@
...
 PYPI_NAME:=$(PKG_NAME)
-PKG_HASH:=8d96ec0de325ea8271cc6aa95b7392bbf548ec4aabd3ffbcdc0619b64edd4a45
 
-PKG_MAINTAINER:=Daniel Danzberger <daniel@dd-wrt.com>
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE:=aws-cli-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/aws/aws-cli.git
+PKG_SOURCE_DATE:=2023-06-02
+PKG_SOURCE_VERSION:=e5f2007f27b067d34fb882bd907e2ad50a60823d
+PKG_MIRROR_HASH:=24b2f233140376ebb6aec6fb2dd3a9734bc1934896069f90d00785c81d121ced
```
jefferyto, line 14: There is no need to switch to downloading from GitHub - you can find the `PKG_HASH` value from pypi.org (search for the `PYPI_NAME` value, go to "Download files" on the left side nav, select "view hashes" next to the tar.gz link, copy the SHA256 value)

lang/python/python-awscli/Makefile
```diff
@@ -1,13 +1,19 @@
...
+
+PKG_MAINTAINER:=Daniel Danzberger <daniel@dd-wrt.com>  ExtraIoT <soporte@extraiot.com.mx>
```
jefferyto, line 16: Is `ExtraIoT <soporte@extraiot.com.mx>` you? We will need a real name here. (Also please add a comma between maintainers.)

jefferyto (commented):
Also:

* Please read the Contributing Guidelines, specifically [this section](https://github.com/openwrt/packages/blob/5b965e4d79afae179080a8bf3b370e650045fe05/CONTRIBUTING.md#commits-in-your-pull-requests-should)
* I suggest splitting each package update into its own commit (so two commits instead of one)

UMB8998:
OK, i will be updating this commit to fix only the aws package and make a new for the botocore, I'm testing now the PyPI pkg just to see that everything runs smoothly. Hopefully by tomorrow I will have it fixed. Thank you for your comments.
About signing the package I'm using a company email as I'm not going to be the only maintainer. Is that OK or should I change to my personal email?

jefferyto:
> About signing the package I'm using a company email as I'm not going to be the only maintainer. Is that OK or should I change to my personal email?

I'm not sure - @neheb @BKPepe @1715173329 @hauke do we have any guidelines on company maintainers?

BKPepe:
Indeed, that's a good question. I can only tell what I was doing when I was working for CZ.NIC (Turris).

- If I was personally interested in the package, then I put mine personal e-mail address.
- If I was not personally interested and we wanted to have the package included in upstream instead of having it downstream, then I put there company e-mail address. This has one disadvantage is that if you are stopped working for the organization, then no one can reach you on your old work address, but on the other hand, customers/users will see that you are working for that company and makes you more known that your company do it better than others, which is good.

If you are not going to be the only maintainer, then decide who will have the main responsibility leading on that. It could be you, and if someone wants to update it, improve it, then you have the final word on that. It is simple. There is a special word that I learned recently it is the abbreviation **SPoC** known as **single point of contact**, which was actually me in the case of Turris (and now they are ignoring me as they are quite understaffed and one small release took them one more and I am still active here more than they. It's a pity, though!).

**In any case, this is your first contribution here and we don't need to reinvent the wheel, and it is too early to discuss is as I see more issues right now**:

- Don't use the master branch of your repository to send the stuff to any upstream repository. There are feature branches for that.
- You need to have an agreement with the previous maintainer that he agrees that you will be coop maintaining the package or become a new one.

UMB8998:
Ok, perfect, is my first time making a pull request so I'm a little out of my depth, from what I see because I used my master branch I will need to close the pull request and make a two new ones under the new names.
The only thing is the agreement with the previous maintainer, as the issue I'm referencing as been open for quite some time. If I don't change the maintainer will the pull request will be accepted?

BKPepe:
Usually, we are merging such changes after 2 weeks to have sufficient time to review it from others if maintainer is not active.

PR #21296, opened 2023-06-07
rust: bump libc to 0.2.146

1715173329:
Fixes build with musl 1.2.4.
Upstream commit: https://github.com/rust-lang/rust/commit/032857e7e403f654129c45dc7e6718a9ad49e377

Maintainer: @lu-zero 
Run tested: n/a


lu-zero:
It looks good, thank you for taking care of it :)

PR #21317, opened 2023-06-09
acme-acmesh: update to 3.0.6

1715173329:
Fix important remote exec bug.
Ref: https://github.com/acmesh-official/acme.sh/issues/4659

Maintainer: @tohojo @hgl


1715173329:
Oh thanks! l missed that ^^

PR #21325, opened 2023-06-10
transmission: Fix env variables passing

vonZeppelin:
Maintainer: @dangowrt
Compile tested: N/A
Run tested: OpenWrt 22.03.5 on WD My Book Live

Description:
It's not possible to configure custom Transmission web home as corresp. environment variable gets overwritten by the command that sets CA bundle environment variable.

The fix is to replace `procd_set_param` with `procd_append_param`.

### DIFF
--- a/net/transmission/files/transmission.init
+++ b/net/transmission/files/transmission.init
@@ -158,8 +158,8 @@ transmission() {
 		logger -t transmission "Starting with $USE virt mem"
 	fi
 
-	[ -d "$web_home" ] && procd_set_param env TRANSMISSION_WEB_HOME="$web_home"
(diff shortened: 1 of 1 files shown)


dangowrt:
> It's not possible to configure custom Transmission web home as corresp. environment variable gets overwritten by the command that sets CA bundle environment variable.

Good catch.
@vonZeppelin Can you put this sentence in the actual commit description as well and do a force-push to your transmission-env branch?

vonZeppelin:
> Can you put this sentence in the actual commit description as well and do a force-push to your transmission-env branch?

@dangowrt Done!

PR #21327, opened 2023-06-11
treewide: change references from armvirt to armsr

1715173329:
Maintainer: @hnyman @jow- for collected, @yousong for qemu, @aparcar for ci 
Compile tested: n/a
Run tested: n/a

Description:
armvirt target has been renamed to armsr (Arm SystemReady),
so the dependency need to be changed as well.

Related to https://github.com/openwrt/openwrt/pull/12832

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -15,12 +15,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - arch: aarch64_cortex-a53
-            target: mvebu-cortexa53
+          - arch: aarch64_generic
+            target: armsr-armv8
             runtime_test: true
 
           - arch: arm_cortex-a15_neon-vfpv4
-            target: armvirt-32
+            target: armsr-armv7
             runtime_test: true
 
           - arch: arm_cortex-a9_vfpv3-d16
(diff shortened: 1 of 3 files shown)


.github/workflows/multi-arch-test-build.yml
```diff
@@ -15,12 +15,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - arch: aarch64_cortex-a53
-            target: mvebu-cortexa53
+          - arch: aarch64_generic
+            target: armsr-armv8
```
BKPepe, line 19: I am quite curious about this change. Maybe I have not follow it carefully, but is there any reference for proposing this change? While looking at https://downloads.openwrt.org/snapshots/packages/aarch64_cortex-a53/packages/ I can not see such change there since all these packages were build yesterday with ``aarch64_cortex-a53``.

Also mvebu-cortexa53 is still there: https://github.com/openwrt/openwrt/tree/9519a625499a963508755b7fc08fb825b3d8cfde/target/linux/mvebu/cortexa53 under mvebu.

BKPepe (changes requested):
It would be good if commits like as it is done in https://github.com/openwrt/packages/pull/21341 includes details such as git commit in OpenWrt main repository.

1715173329:
added commit reference

.github/workflows/multi-arch-test-build.yml
BKPepe: Interesting finding, thank you for your explanation. Even though, I would rather see to have updated ``aarch64_cortex-a53`` rootfs container, but LGTM

hnyman:
Collectd change is ok

PR #21331, opened 2023-06-11
fatresize: bump release

oskarirauta:
use newer commit, fixes problem with manually
specified sizes.

Maintainer: me


utils/fatresize/Makefile
```diff
@@ -2,13 +2,13 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=fatresize
 PKG_VERSION:=1.1.0
-PKG_RELEASE:=1
```
neheb, line 5: Hmm? Why bump?

1715173329: As `PKG_VERSION` is specified, the download script will still generate fatresize-1.1.0.tar.xz file.

I would suggest removing PKG_VERSION directly and resetting PKG_RELEASE to 1.

oskarirauta: @neheb
Because previous version is same, shouldn't release then be bumped?

@1715173329 

I'm ok, with all available resolutions. Newer commits just seemed important enough to provide update. You to decide together what solution should we proceed here, and then I'll make necessary changes to my PR, ok?

oskarirauta: Version is 1.1.0 + some patches that haven't yet drafter to a release.. So version probably should be there- or not?
So, which approach would be correct?

1715173329: from my point
```patch
-PKG_VERSION:=1.1.0
-PKG_RELEASE:=2
+PKG_RELEASE:=1
```

oskarirauta: @1715173329 
Changes pushed.

@neheb - what do you think of it?

neheb: Looks correct

oskarirauta:
Hash fixed.

1715173329:
Could you please create a PR for the 23.05 branch? Thanks!

PR #21341, opened 2023-06-12
treewide: change armvirt target references to armsr (qemu,collectd,CI)

mcbridematt:
A couple of days ago, the armvirt target was renamed to 'armsr' (Arm SystemReady) after the introduction of EFI and support for 'real' hardware.

There are a few `DEPENDS:=@TARGET_armvirt||@TARGET_etc` references in the OpenWrt feeds which need to be changed, as well as the GitHub workflow.

See http://lists.openwrt.org/pipermail/openwrt-devel/2023-June/041151.html and https://github.com/openwrt/openwrt/pull/12832
The same change is likely to occur for the 23.05 branch as well: https://github.com/openwrt/openwrt/pull/12817

Marking this is a DRAFT until the buildbot (and downloads.openwrt.org) catches up with the target name change.

qemu:
Maintainer: @yousong 

collectd (cpufreq module):
Maintainer: @akosiaris

Compile tested: armvirt/armv8
Run tested: armvirt/armv8 VM (qemu)

### DIFF
--- a/.github/workflows/multi-arch-test-build.yml
+++ b/.github/workflows/multi-arch-test-build.yml
@@ -20,7 +20,7 @@ jobs:
             runtime_test: true
 
           - arch: arm_cortex-a15_neon-vfpv4
-            target: armvirt-32
+            target: armsr-armv7
             runtime_test: true
 
           - arch: arm_cortex-a9_vfpv3-d16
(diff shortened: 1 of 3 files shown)


mcbridematt:
Oops, was already beaten to it
https://github.com/openwrt/packages/pull/21327

PR #21347, opened 2023-06-12
[openwrt-21.02] python3: Update to 3.9.17

jefferyto:
Maintainer: me
Compile tested: armvirt-32, 21.02.7 sdk
Run tested: none

Description:
This includes an updated patch for pip, as the bundled pip was also updated with this release.


BKPepe:
@paper42 You should give it a try.

PR #21349, opened 2023-06-12
gpsd: update to 3.25

PolynomialDivision:
Release Information:
https://gitlab.com/gpsd/gpsd/-/blob/release-3.25/NEWS?ref_type=tags

Latest olsrd version needs this bump.

Maintainer: none?
Compile tested: mt7621
Run tested: mt7621


BKPepe:
I see that you removed the patch, usually, it means that it was upstreamed or rather backported patch is no longer necessary, right? Could you be more talkative about that in the commit description? 🙏

PolynomialDivision:
Done :)

BKPepe:
Good, LGTM

PR #21350, opened 2023-06-12
watchcat: fix misleading prints about restarting network

timmarjo:
Maintainer: @roger-
Compile tested: ipq40xx, VG710, OpenWrt 22.03.0-rc4
Run tested: ipq40xx, VG710, OpenWrt 22.03.0-rc4

Description:
Noticed that watchcat prints it will restart an interface or networking (or run script) when pinging any configured host fails even though the action (restart/script run) is committed only if none of the hosts can be pinged.

### DIFF
--- a/utils/watchcat/files/watchcat.sh
+++ b/utils/watchcat/files/watchcat.sh
@@ -219,6 +219,7 @@ watchcat_monitor_network() {
 
 	ping_family="$(get_ping_family_flag "$address_family")"
 
+	about_to_restart_iface=1
 	while true; do
 		# account for the time ping took to return. With a ping time of 5s, ping might take more than that, so it is important to avoid even more delay.
 		time_now="$(cat /proc/uptime)"
@@ -231,6 +232,7 @@ watchcat_monitor_network() {
 		time_now="${time_now%%.*}"
 		time_lastcheck="$time_now"
 
+		end_result=1
 		for host in $ping_hosts; do
 			if [ "$ping_iface" != "" ]; then
 				ping_result="$(
@@ -246,16 +248,31 @@ watchcat_monitor_network() {
 
 			if [ "$ping_result" -eq 0 ]; then
 				time_lastcheck_withinternet="$time_now"
-			else
-				if [ "$script" != "" ]; then
-					logger -p daemon.info -t "watchcat[$$]" "Could not reach $host via \"$iface\" for \"$((time_now - time_lastcheck_withinternet))\" seconds. Will run the script after \"$failure_period\" seconds of failed reachability"
-				elif [ "$iface" != "" ]; then
-					logger -p daemon.info -t "watchcat[$$]" "Could not reach $host via \"$iface\" for \"$((time_now - time_lastcheck_withinternet))\" seconds. Will restart \"$iface\" after \"$failure_period\" seconds of failed reachability"
-				else
-					logger -p daemon.info -t "watchcat[$$]" "Could not reach $host for \"$((time_now - time_lastcheck_withinternet))\" seconds. Will restart networking after \"$failure_period\" seconds of failed reachability"
+				end_result=0
+				if [ "$about_to_restart_iface" -eq 0 ]; then
+					if [ "$script" != "" ]; then
+						logger -p daemon.info -t "watchcat[$$]" "$host replies to ping, running script cancelled"
+					elif [ "$iface" != "" ]; then
+						logger -p daemon.info -t "watchcat[$$]" "$host replies to ping, restarting \"$iface\" cancelled"
+					else
+						logger -p daemon.info -t "watchcat[$$]" "$host replies to ping, restarting networking cancelled"
+					fi
+					about_to_restart_iface=1
 				fi
+			else
+				logger -p daemon.info -t "watchcat[$$]" "Could not reach $host for \"$((time_now - time_lastcheck_withinternet))\" seconds."
 			fi
 		done
+		if [ "$end_result" -ne 0 ]; then
+			if [ "$script" != "" ]; then
+				logger -p daemon.info -t "watchcat[$$]" "Could not reach any configured host for \"$((time_now - time_lastcheck_withinternet))\" seconds. Will run the script after \"$failure_period\" seconds of failed reachability"
+			elif [ "$iface" != "" ]; then
+				logger -p daemon.info -t "watchcat[$$]" "Could not reach any configured host via \"$iface\" for \"$((time_now - time_lastcheck_withinternet))\" seconds. Will restart \"$iface\" after \"$failure_period\" seconds of failed reachability"
+			else
+				logger -p daemon.info -t "watchcat[$$]" "Could not reach any configured host for \"$((time_now - time_lastcheck_withinternet))\" seconds. Will restart networking after \"$failure_period\" seconds of failed reachability"
+			fi
+			about_to_restart_iface=0
+		fi
 
 		[ "$((time_now - time_lastcheck_withinternet))" -ge "$failure_period" ] && {
 			recovery_started="$time_now"
@@ -273,6 +290,7 @@ watchcat_monitor_network() {
 				fi
 			fi
 			/etc/init.d/watchcat start
+			about_to_restart_iface=1
 			# Optionally start a fresh failure window after the recovery action
 			# finishes instead of continuing to count the original outage.
 			if [ "$reset_failure_timer" = "1" ]; then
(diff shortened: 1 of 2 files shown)


BKPepe:
It requires to bump ``PKG_RELEASE``. Oh no, here we go again. We don't use AUTORELEASE anymore. :-(

GeorgeSapkin:
@roger- @timmarjo still relevant?

timmarjo:
Sorry for late response but yes, this is still relevant.

BKPepe:
There are still some merge conflicts, which needs to be solved.

utils/watchcat/files/watchcat.sh
```diff
@@ -149,16 +151,31 @@ watchcat_monitor_network() {
...
+					fi
+					about_to_restart_iface=1
 				fi
+			else
+				logger -p daemon.info -t "watchcat[$$]" "Could not reach $host for \"$((time_now - time_lastcheck_withinternet))\" seconds."
 			fi
 		done
+		if [ "$end_result" -ne 0 ]; then
+			if [ "$script" != "" ]; then
+				logger -p daemon.info -t "watchcat[$$]" "Could not reach any configured host via \"$iface\" for \"$((time_now - time_lastcheck_withinternet))\" seconds. Running script after reaching \"$failure_period\" seconds"
```
openwrt-ai: nit: In the `script` branch, `$iface` is typically empty when a script action is configured, so this logs `... via "" ...`. Since this PR is specifically about fixing misleading prints, consider dropping the `via "$iface"` fragment here (the "Running script" text already conveys the action). This mirrors a pre-existing quirk, but it is cheap to clean up while touching these lines.

---

openwrt-ai (commented):
## Commit checks

- `9c96550` ("watchcat: Add more prints when ping fails/succeeds") is missing a `Signed-off-by:` trailer. OpenWrt requires every commit to be signed off (see CONTRIBUTING.md); the other three commits have it. Please add it, e.g. rebase and `git commit --amend -s` on that commit.
- The series includes a merge commit `4d25b1a` ("Merge branch 'master' into master"). OpenWrt expects PRs to be rebased onto master rather than merged into it; please rebase to drop the merge commit — this is also what surfaced the reported conflicts. While rebasing, the `PKG_RELEASE` bump in `0ce0dab` ("Update Makefile") is conventionally squashed into the functional commit rather than kept separate.

The core logic change (tracking `end_result`/`about_to_restart_iface` so the "restarting" message is only emitted once all configured hosts fail, with a matching "cancelled" message when a host recovers) reads correctly and achieves the stated goal. One minor inline note.

---

timmarjo:
Regarding the `FormalityCheck / Git & Commits` failure ("Pull request must originate from a feature branch"): this PR's head is the `master` branch of the fork, which predates that check (this PR was opened in 2023, before the formality bot existed). GitHub does not allow changing a pull request's head branch after creation, so the only way to satisfy this specific check would be to close this PR and re-open an equivalent one from a properly named feature branch, losing this thread's history and comments in the process.

The check is not a required status check on this repo (no branch protection enforces it, and `mergeStateStatus` is `UNSTABLE` rather than `BLOCKED`), so it shouldn't block merging. Flagging this here so it isn't mistaken for something still broken in the diff.

_Generated by [Claude Sonnet 5](https://claude.ai/code)_

utils/watchcat/files/watchcat.sh
```diff
@@ -246,16 +248,31 @@ watchcat_monitor_network() {
...
+						logger -p daemon.info -t "watchcat[$$]" "$host replies to ping, running script cancelled"
+					elif [ "$iface" != "" ]; then
+						logger -p daemon.info -t "watchcat[$$]" "$host replies to ping, restarting \"$iface\" cancelled"
+					else
+						logger -p daemon.info -t "watchcat[$$]" "$host replies to ping, restarting networking cancelled"
+					fi
+					about_to_restart_iface=1
 				fi
+			else
+				logger -p daemon.info -t "watchcat[$$]" "Could not reach $host for \"$((time_now - time_lastcheck_withinternet))\" seconds."
```
openwrt-ai, line 263: nit (optional): the per-host line loses the `via "$iface"` context that two of the three old branches carried, so on a multi-WAN box the log no longer says which interface the failed ping went out of. The aggregate line below keeps it only for the `iface`-without-script case. Was dropping it here intentional, or would keeping the interface on the per-host line be preferable since that is the line that identifies the individual failure?

Separately, for the common single-host configuration this line plus the aggregate line below now produce two near-identical entries per ping interval; suppressing the per-host line when only one host is configured (or when all hosts failed) would keep the log volume where it was.

---

utils/watchcat/files/watchcat.sh
```diff
@@ -246,16 +248,31 @@ watchcat_monitor_network() {
...
 			fi
 		done
+		if [ "$end_result" -ne 0 ]; then
+			if [ "$script" != "" ]; then
+				logger -p daemon.info -t "watchcat[$$]" "Could not reach any configured host for \"$((time_now - time_lastcheck_withinternet))\" seconds. Will run the script after \"$failure_period\" seconds of failed reachability"
```
openwrt-ai, line 268: The interface is still meaningful in `run_script` mode, so dropping it here removes information rather than removing a misleading value: `run_script` accepts an `interface` option, it is used as the ping source (`ping -I "$ping_iface"`), and it is handed to the user script by [`watchcat_run_script`](https://github.com/openwrt/packages/blob/43afa2aabb244da88f14a6cc5f70bf8d74e32caf/utils/watchcat/files/watchcat.sh#L166-L169), which itself logs `Running script "$1" for network interface: "$2"`. Consider keeping the interface here the way the `elif` branch below does (it is empty only when the user did not configure one):

```suggestion
				logger -p daemon.info -t "watchcat[$$]" "Could not reach any configured host via \"$iface\" for \"$((time_now - time_lastcheck_withinternet))\" seconds. Will run the script after \"$failure_period\" seconds of failed reachability"
```

The same applies to the "running script cancelled" message on line 254.

---

utils/watchcat/files/watchcat.sh
```diff
@@ -246,16 +248,31 @@ watchcat_monitor_network() {
...
+					else
+						logger -p daemon.info -t "watchcat[$$]" "$host replies to ping, restarting networking cancelled"
+					fi
+					about_to_restart_iface=1
 				fi
+			else
+				logger -p daemon.info -t "watchcat[$$]" "Could not reach $host for \"$((time_now - time_lastcheck_withinternet))\" seconds."
 			fi
 		done
+		if [ "$end_result" -ne 0 ]; then
```
openwrt-ai, line 266: nit (optional): `ping_reboot` mode has exactly the same misleading print and is not touched by this PR — [`watchcat_ping()` at watchcat.sh:370](https://github.com/openwrt/packages/blob/43afa2aabb244da88f14a6cc5f70bf8d74e32caf/utils/watchcat/files/watchcat.sh#L370) logs `Could not reach $host for N. Rebooting after reaching $failure_period` per failing host, while the reboot on the line below it only fires when no configured host replied. Was leaving that mode alone deliberate (keeping the diff narrow), or is it worth applying the same `end_result` guard there so both modes report consistently?

---

openwrt-ai (commented):
The branch has been rebased since my last review: the four-commit series plus the `Merge branch 'master' into master` commit are gone, and everything is now one signed-off commit. The earlier findings (missing `Signed-off-by`, merge commit in the series, separate `PKG_RELEASE` bump) are all resolved, and the `PKG_RELEASE` 25 -> 26 bump is the right call for a change to the shipped script.

## Commit checks

- `43afa2a` "watchcat: fix misleading prints about restarting network" — the message justifies the last hunk with "Drop the interface name from the script-path warning since a script action does not target an interface", but `run_script` mode does accept an `interface`: it is the ping source and it is passed to the user script as `$1` by `watchcat_run_script`, which logs it as well. The rest of the body matches the diff accurately. See the inline note on line 268.

The core change reads correctly: `end_result` is reset per check cycle, only cleared by a successful ping, and the "will restart"/"will run the script" line is emitted once per cycle after every host has failed; `about_to_restart_iface` is cleared exactly where the warning is emitted and re-armed both on recovery and after the recovery action runs, so the "cancelled" notice cannot fire spuriously.

Nothing here blocks a merge — the inline items are all optional.

---

timmarjo:
Re-opened as #30573 from a dedicated feature branch (`watchcat-fix-misleading-prints`) to satisfy the `Pull requests must originate from a dedicated feature branch` formality check, which this PR's fork branch (named `master`) could never pass. Same change, rebased onto current master.

_Generated by [Claude Sonnet 5](https://claude.ai/code)_

PR #21356, opened 2023-06-12
liburiparser: new package

M95D:
Maintainer: @sKyissKy
Compile tested: Gentoo Linux x86-64
Run tested: mvebu, Linksys WRT-1900ACv1

Description:
This is based on original pull request by @sKyissKy : #20504
- Added liburiparser.
- Added tvheadend dependency on liburiparser. I made it configurable by user, since most users won't need this. Default is off.

Tested:
Tvheadend works when built with the new lib option, but I did not test the actual functioning of liburiparser.

### DIFF
--- a/libs/uriparser/Makefile
+++ b/libs/uriparser/Makefile
@@ -1,0 +1,56 @@
+#
+# Copyright (C) 2015-2018 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=uriparser
+PKG_VERSION:=0.9.7
+PKG_RELEASE:=1
+
+PKG_SOURCE:=uriparser-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/uriparser/uriparser/releases/download/uriparser-$(PKG_VERSION)/
+PKG_HASH:=11553b2abd2b5728a6c88e35ab08e807d0a0f23c44920df937778ce8cc4d40ff
+
+PKG_MAINTAINER:=Weijia Song <songweijia@gmail.com>,Sebastian Pipping <sebastian@pipping.org>
+PKG_LICENSE:=New BSD license
+PKG_LICENSE_FILES:=COPYING
+PKG_CPE_ID:=cpe:/a:uriparser:uriparser
+
+CMAKE_INSTALL:=1
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+define Package/liburiparser
+  SECTION:=libs
+  CATEGORY:=Libraries
+  TITLE:=RFC 3986 URI parsing library
+  URL:=https://uriparser.github.io/
+endef
+
+define Package/liburiparser/description
+ uriparser is a strictly RFC 3986 compliant URI parsing and handling library written in C89 ("ANSI C").
+endef
+
+CMAKE_OPTIONS += \
+	-DBUILD_SHARED_LIBS:BOOL=ON \
+	-DURIPARSER_BUILD_DOCS:BOOL=OFF \
+	-DURIPARSER_BUILD_TESTS:BOOL=OFF \
+	-DURIPARSER_BUILD_TOOLS:BOOL=OFF \
+	-DURIPARSER_BUILD_CHAR:BOOL=ON \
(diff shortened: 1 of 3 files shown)


libs/uriparser/Makefile
```diff
@@ -0,0 +1,56 @@
...
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=uriparser
+PKG_VERSION:=0.9.7
+PKG_RELEASE:=1
+
+PKG_SOURCE:=uriparser-$(PKG_VERSION).tar.gz
```
BKPepe, line 14:
```suggestion
PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
```

libs/uriparser/Makefile
```diff
@@ -0,0 +1,56 @@
...
+PKG_SOURCE_URL:=https://github.com/uriparser/uriparser/releases/download/uriparser-$(PKG_VERSION)/
+PKG_HASH:=11553b2abd2b5728a6c88e35ab08e807d0a0f23c44920df937778ce8cc4d40ff
+
+PKG_MAINTAINER:=Weijia Song <songweijia@gmail.com>,Sebastian Pipping <sebastian@pipping.org>
+PKG_LICENSE:=New BSD license
```
BKPepe, line 19: This is no SPDX License Identifier. Please check it here: https://spdx.org/licenses/

libs/uriparser/Makefile
```diff
@@ -0,0 +1,56 @@
...
+	-DURIPARSER_BUILD_CHAR:BOOL=ON \
+	-DURIPARSER_BUILD_WCHAR_T:BOOL=ON \
+	-DURIPARSER_ENABLE_INSTALL:BOOL=ON \
+	-DURIPARSER_WARNINGS_AS_ERRORS:BOOL=OFF
+
+define Package/liburiparser/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) \
+		$(PKG_INSTALL_DIR)/usr/lib/liburiparser.so.* \
+		$(1)/usr/lib/
```
BKPepe, line 53:
```makefile
	$(CP) $(PKG_INSTALL_DIR)/usr/lib/liburiparser.so* $(1)/usr/lib/

```

libs/uriparser/Makefile
```diff
@@ -0,0 +1,56 @@
+#
+# Copyright (C) 2015-2018 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
```
BKPepe, line 6: Hmmm, no. Check it here: https://github.com/openwrt/packages/blob/81d53813d63dcbedf1489946af9854070c3b80a0/CONTRIBUTING.md

BKPepe (changes requested):
You should consider to add also uriparser binary.
You can see it here: https://gitlab.nic.cz/turris/os/packages/-/blob/master/libs/uriparser/Makefile

libs/uriparser/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+PKG_NAME:=uriparser
+PKG_VERSION:=0.9.7
+PKG_RELEASE:=1
+
+PKG_SOURCE:=uriparser-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/uriparser/uriparser/releases/download/uriparser-$(PKG_VERSION)/
+PKG_HASH:=11553b2abd2b5728a6c88e35ab08e807d0a0f23c44920df937778ce8cc4d40ff
+
+PKG_MAINTAINER:=Weijia Song <songweijia@gmail.com>,Sebastian Pipping <sebastian@pipping.org>
```
BKPepe, line 18: Who are these maintainers? Are there aware that they will be maintainers?

libs/uriparser/Makefile
M95D: OK.

libs/uriparser/Makefile
M95D: Will fix.

libs/uriparser/Makefile
BKPepe: It would be if he is active, tries to contribute and take cares of the feedback, which does not happen. So, the maintainer should be someone, who is interested in it and will update it in the future. We can not add new packages without maintainer.

M95D:
Unrelated question:
Why are there gettext errors in the tests? Is there something wrong with the package? It builds fine on my system.

PR #21357, opened 2023-06-12
cloudreve: Update to 3.8.0

1715173329:
Maintainer: me
Run tested: nanopi-r2s

Description:
Updated cloudreve and added missing web frontend assets archive.

### DIFF
--- a/net/cloudreve/Makefile
+++ b/net/cloudreve/Makefile
@@ -5,13 +5,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cloudreve
-PKG_VERSION:=3.7.1
+PKG_VERSION:=3.8.0
 PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/cloudreve/Cloudreve.git
 PKG_SOURCE_VERSION:=$(PKG_VERSION)
-PKG_MIRROR_HASH:=b70601111e727d879a709884bc28237861216b99abac02b6f542f5984ddd8c0d
+PKG_MIRROR_HASH:=4c05afb2193ab83bf144d75963f4757c1445317e68b2690727375a8e01f7ba3d
 
 PKG_LICENSE:=GPL-3.0-only
 PKG_LICENSE_FILES:=LICENSE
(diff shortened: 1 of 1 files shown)


utils/7z/Makefile
```diff
@@ -0,0 +1,68 @@
...
+PKG_MAINTAINER:=Tianling Shen <cnsztl@immortalwrt.org>
+PKG_LICENSE:=LGPL-2.1-or-later BSD-3-Clause
+PKG_LICENSE_FILES:=DOC/License.txt DOC/copying.txt
+PKG_CPE_ID:=cpe:/a:7-zip:7-zip
+
+HOST_BUILD_PARALLEL:=1
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/host-build.mk
```
neheb, line 20: Why?

1715173329: cloudreve needs it to create zip archive.

neheb: This is a host package already provided by tools/7zip.

1715173329: Yep, but it's 7zr, which can only handle .7z archives.

neheb: If the compilation options for tools/7zip are inadequate, submit a patch to fix those.

neheb: Wait a minute, you're creating a zip file. Why not use libdeflate instead?

1715173329: > Wait a minute, you're creating a zip file. Why not use libdeflate instead?

If I understand correctly libdeflate can only create gzip archive, it's not suitable here.

And I see zip is already in tools https://github.com/openwrt/openwrt/tree/main/tools/zip, I must have missed something

neheb: Ah right. libdeflate is only for creating gzip files.

net/cloudreve/Makefile
```diff
@@ -35,7 +35,7 @@ define Package/cloudreve
   SUBMENU:=Cloud Manager
   TITLE:=A project helps you build your own cloud in minutes
   URL:=https://cloudreve.org
-  DEPENDS:=@(aarch64||arm||i386||i686||x86_64) +ca-bundle
+  DEPENDS:=@(aarch64||arm||i386||i686||riscv64||x86_64) +ca-bundle
```
neheb, line 38: Wasn’t there a GO_ARCH_DEPENDS variable or something like that?

1715173329: `mips/mips64/mips64el/mipsel/powerpc64` are unsupported by cloudreve, though i can use `@!(...)` but this line would be even longer

PR #21358, opened 2023-06-12
tvheadend: update, patch for busybox tar

M95D:
Maintainer: me
Compile tested: Gentoo Linux x86-64
Run tested: mvebu / Linksys WRT1900ACv1

Description:
- Update to git master 2023-06-07 and removed no longer needed compatibility patch.
- Patched tar command to use gzip instead of bzip2. This is used by tvheadend to back up old configuration after a version update. The backup was already disabled by init script command line, but some users may want it enabled.


multimedia/tvheadend/Makefile
```diff
@@ -1,14 +1,13 @@
...
-PKG_SOURCE_PROTO:=git
-PKG_SOURCE_URL:=https://github.com/tvheadend/tvheadend.git
-PKG_MIRROR_HASH:=9c1bb3eea3f3539454d17e4c6aabc774a104a772aa34ed39f248521cf6488ce5
-PKG_SOURCE_VERSION:=0ff96106aa2e0f9a384c3a2662ca005797a6b399
-PKG_SOURCE_DATE:=2022-11-20
+PKG_VERSION:=9df7d2d6bc37b8aa25ac63be7b0a5d69be10c892
+PKG_RELEASE:=4
+
+PKG_SOURCE_URL:=https://github.com/tvheadend/tvheadend/archive/
+PKG_SOURCE:=$(PKG_VERSION).zip
```
BKPepe, line 8: No, the tarball is going to be downloaded as 9df7d2d6bc37b8aa25ac63be7b0a5d69be10c892.zip that's not wanted. Keep the previous solution.

BKPepe (changes requested):

multimedia/tvheadend/patches/060-busybox-tar.patch
```diff
@@ -0,0 +1,11 @@
+--- a/src/config.c
++++ b/src/config.c
+@@ -1481,7 +1481,7 @@ dobackup(const char *oldver)
```
BKPepe, line 3: I don't like patches, where there are no details like who is author, why it is necessary, etc. If you are using Tvheadend, then you should use proper tar, but if you want to use busybox tar, then please try to discuss it with upstream to include the solution there.

multimedia/tvheadend/Makefile
M95D: I tried that twice. The hash differs from my system. I don't know how to fix it.

multimedia/tvheadend/patches/060-busybox-tar.patch
M95D: I can remove it completely. It won't give any error unless someone changes the startup script.

BKPepe: There is also another possibility add tar as dependency to tvheadend until someone changes the startup script.

BKPepe:
No need to close it, you could use force push with the update. Just change the values in variables and thats it

M95D:
I know, but the package hash is still not working when cloning with git vs. downloading the zip. I'll try to clean everything, including my github fork.

M95D:
Can I set the name of the downloaded file to something other than gihub hash?

PR #21374, opened 2023-06-15
c-ares: bump to 1.19.1

nxhack:
Maintainer: @karlp
Compile tested: head, aarch64, arm, i386, mipsel, x86_64
Run tested: aarch64 (qemu 8.0.2)

Description:
This is a security and bugfix release.

Security
 o CVE-2023-32067. High. 0-byte UDP payload causes Denial of Service
 o CVE-2023-31147. Moderate. Insufficient randomness in generation of DNS query IDs
 o CVE-2023-31130. Moderate. Buffer Underwrite in ares_inet_net_pton()
 o CVE-2023-31124. Low. AutoTools does not set CARES_RANDOM_FILE during cross compilation

Fixing libcares.pc
 The pkg-config file libcares.pc in version 1.19.1 has been changed to be unsuitable for OpenWrt
 and causes build errors with Openwrt packages that use libcares.
 For this reason, libcares.pc was replaced.

### DIFF
--- a/libs/c-ares/Makefile
+++ b/libs/c-ares/Makefile
@@ -9,12 +9,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=c-ares
-PKG_VERSION:=1.18.1
+PKG_VERSION:=1.19.1
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://c-ares.org/download
-PKG_HASH:=1a7d52a8a84a9fbffb1be9133c0f6e17217d91ea5a6fa61f6b4729cda78ebbcf
+PKG_HASH:=321700399b72ed0e037d0074c629e7741f6b2ec2dda92956abe3e9671d3e268e
 
 PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=LICENSE.md
@@ -47,6 +47,13 @@ CMAKE_OPTIONS += \
 	-DCARES_STATIC_PIC=ON \
 	-DCARES_BUILD_TESTS=OFF \
 	-DCARES_BUILD_TOOLS=OFF
+
+define Build/InstallDev
+	$(call Build/InstallDev/cmake,$(1))
+	$(SED) 's,/usr/bin,$$$${prefix}/lib,g' $(1)/usr/lib/pkgconfig/libcares.pc
+	$(SED) 's,/usr/include,$$$${prefix}/include,g' $(1)/usr/lib/pkgconfig/libcares.pc
+	$(SED) 's,/usr/lib,$$$${prefix}/lib,g' $(1)/usr/lib/pkgconfig/libcares.pc
+endef
 
 define Package/libcares/install
 	$(INSTALL_DIR) $(1)/usr/lib


libs/c-ares/patches/001-fix_libcares_pc_path.patch
```diff
@@ -0,0 +1,16 @@
+--- a/libcares.pc.cmake
```
BKPepe, line 1: I see this patch is added. Did you try to reach upstream if we can get rid of in the future? Maybe, it would be nice to have some slight description about this.

nxhack: Difference of $(STAGING_DIR)/usr/lib/pkgconfig/libcares.pc

1.18.1
```
prefix=/usr
exec_prefix=${prefix}/bin
libdir=${prefix}/lib
includedir=${prefix}/include
```

1.19.1 w/o patch
```
exec_prefix=/usr/bin
libdir=/usr/lib
includedir=/usr/include
```
1.19.1 with patch
```
prefix=/usr
#exec_prefix=${prefix}/bin
libdir=${prefix}/lib
includedir=${prefix}/include
```

nxhack: Additional information: ${prefix} is given externally.

nxhack: Ah, yes, if you give it to them in the c-ares configure, they might be fine without the patch. I'll give it a try.

nxhack: I still think this is reasonable.

1715173329: There's a similar PR in upstream https://github.com/c-ares/c-ares/pull/530

nxhack: CMAKE build, so it depends on libcares.pc.cmake.

nxhack: Past discussions about pkg-config.

https://github.com/openwrt/openwrt/pull/1615

1715173329: well, main question is if it can be fixed on upstream side so we won't have to keep this patch forever...

nxhack: I know that some openwrt packages have a sed replacement method.

nxhack: It seems complicated to convince pkg-config to use "${prefix}" instead of absolute paths. In normal use, absolute paths do not cause problems.

nxhack: Remove patch. Changed to replace with sed.

BKPepe: Looks much better this time, but my original proposal still remains. In the commit message, there is no sight why it is necessary, because from the commit, it looks like usual package bump, which actually is not as there is added InstallDev section. One or two sentence would be handy as it helps in the future to look, why it was added (``git blame``), and it helps to understand it.

nxhack: OK. Added to the commit message.

BKPepe: Awesome! 👏

1715173329 (approved):

BKPepe (approved):

1715173329:
ping @karlp

karlp:
I'm between roles right now, earliest I'll get to my earlier process of run testing c-ares updates is early next week, I'd hoped sooner.  If it builds, and someone has run tested it, it's fine by me.  (I only tangentially maintain this as a dependency of mosquitto)

1715173329:
@nxhack tested it already so maybe it's fine to merge.

nxhack:
@karlp @1715173329 
I have confirmed that mosquitto and node.js can be built successfully and that it works.

PR #21378, opened 2023-06-15
modemmanager: do not cache virtual device events and fix alias check

feckert:
Maintainer: @aleksander0m
Run tested: x86/64, APU3, OpenWrt latest

Description:

1. On small systems with many virtual devices, the modem manager sometimes could not start because it took too long until all devices for the modem were recognised. This is because all system events that are stored in the file events.cache have to be processed. To speed up the processing, all devices under /sys/devices/virtual are now filtered out so that they do not have to be processed.
2. If an alias name is used for the modem, then a check if the device exists in sysfs does not work. To fix this remove the check if the sysfs device exists. The protocoll handler already checks if the modemmanger is responsible for this device on the next line.

### DIFF
--- a/net/modemmanager/Makefile
+++ b/net/modemmanager/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=modemmanager
 PKG_SOURCE_VERSION:=1.20.6
-PKG_RELEASE:=2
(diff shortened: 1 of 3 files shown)


aleksander0m:
Hey! I'm not a maintainer of the package, so cannot approve, but both changes seem fine to me.

PR #21385, opened 2023-06-15
linuxptp: move KBUILD_OUTPUT to MAKE_ARGS

cotequeiroz:
Maintainer: Wojciech Dubowik
Compile tested: mediatek/mt7622, aarch64_cortex-a53
Run tested: none

Description:
It was being set in the environment, but it must be a make variable.

The bots are building this fine inspite of this.  Locally, the sure indication that the variable was not being picked up is these non-fatal errors right after running make, and again after the DEPEND lines:

grep: /include/uapi/linux/net_tstamp.h: No such file or directory
grep: /include/uapi/linux/net_tstamp.h: No such file or directory

The paths should have been prefixed by the value of KBUILD_OUTPUT.

The fatal error then is this:
```
In file included from clock.c:35:
missing.h:61:9: error: redeclaration of enumerator 'HWTSTAMP_TX_ONESTEP_P2P'
   61 |         HWTSTAMP_TX_ONESTEP_P2P = 3,
      |         ^~~~~~~~~~~~~~~~~~~~~~~
In file included from clock.c:21:
staging_dir/toolchain-aarch64_cortex-a53_gcc-12.3.0_musl/include/linux/net_tstamp.h:113:9: note: previous definition of 'HWTSTAMP_TX_ONESTEP_P2P' with type 'enum hwtstamp_tx_types'
  113 |         HWTSTAMP_TX_ONESTEP_P2P,
      |         ^~~~~~~~~~~~~~~~~~~~~~~
make[3]: *** [<builtin>: clock.o] Error 1
```

---

The reason why building locally fails, while the bots don't, is because I run a kernel without modules, so `/lib/modules/'uname -r'/build` does not exist (see incdefs.sh in the package build dir).  The path will otherwise be present, but it would point to the host kernel, not the target's--definitly a bug.


cotequeiroz:
It seems I need to review my theory here.  CI is failing just as I described locally.

PR #21388, opened 2023-06-16
dnsdist: Split in two packages `dnsdist` and `dnsdist-full`

rgacogne:
Maintainer: @Habbie 
Compile tested: TP-Link Archer C7 AC1750 v5, 19.07, 21.02 and master
Run tested: TP-Link Archer C7 AC1750 v5, 19.07 and 21.02

Description:

- `dnsdist-full` has all optional features enabled, but is a big package in term of both flash and memory footprint.
- `dnsdist` only keeps the features that make the most sense on embeded devices, but can also be customised to match the user's needs, up to the point where it matches `dnsdist-full`.

Signed-off-by: Remi Gacogne remi.gacogne@powerdns.com


net/dnsdist/Makefile
```diff
@@ -25,6 +24,35 @@ PKG_CONFIG_DEPENDS:= \
...
+  users while shunting or blocking abusive traffic.
+endef
+
+define Package/dnsdist/conffiles/Default
+/etc/dnsdist.conf
+/etc/config/dnsdist
+/etc/init.d/dnsdist
+endef
+
 define Package/dnsdist/config
```
BKPepe, line 56: I think, it would be good to move this to the ``Config.in`` as it is getting very long.

rgacogne: That makes complete sense to me, will do! Thanks!

rgacogne: Done!

Habbie:
I have also tested this (docker, x86_64) and everything looks good to me.

PR #21398, opened 2023-06-17
mhz: add new package

robimarko:
mhz is a tool for mathematically calculating the current CPU frequency, it has proven to be a really good help while developing CPU frequency scaling solutions as it allows to independently prove that scaling actually works.

Now that the author has added a license we can package it for all to use.

Maintainer: me 
Compile tested: ipq807x
Run tested: ipq807x

### DIFF
--- a/utils/mhz/Makefile
+++ b/utils/mhz/Makefile
@@ -1,0 +1,33 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=mhz
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL=https://github.com/wtarreau/mhz.git
+PKG_SOURCE_DATE:=2023-06-17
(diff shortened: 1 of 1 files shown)


utils/mhz/Makefile
```diff
@@ -0,0 +1,33 @@
...
+PKG_MAINTAINER:=Robert Marko <robimarko@gmail.com>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/mhz
+	SECTION:=utils
+	CATEGORY:=Utilities
+	TITLE:=CPU frequency measurement utility
```
1715173329, line 21:
```suggestion
  SECTION:=utils
  CATEGORY:=Utilities
  TITLE:=CPU frequency measurement utility
```

robimarko: Done

1715173329 (approved):
LGTM - Thanks!

PR #21402, opened 2023-06-18
isc-dhcp: force sending unrequested parameters

pprindeville:
Maintainer: me
Compile tested: x86_64, generic, HEAD (9519a62549)

Description:

Certain IoTs (like Amazon Echos, or Amazon smart light switches), game consoles (Nintendo Switch), SIP handsets, etc. don't request a hostname, and either end up using a static name (like `Amazon` for switches), or else synthesize a patterned name (like `amazon-s/n-here` for Echo's) which makes it hard to figure out what is what when looking at mDNS browsing.


pprindeville:
Should this be per-host for reservations, or is it useful to define on a per subnet basis as well?

PR #21403, opened 2023-06-18
net/iodine: upgrade to version 0.8.0

Tabiskabis:
Maintainer: Uwe Kleine-König <uwe+openwrt@kleine-koenig.org>
Compile tested: x86_64, Debian bookworm, OpenWRT v23.05.0-rc1
Run tested: x86/64, PC Engines APU2, OpenWRT v23.05.0-rc1
Tests done: installed/started iodine server+client, exchanged a few bytes.

Description:
It's been almost 9 years since the last upstream release. Relevant changes on OpenWRT:
- Add support for IPv6 in the server (Traffic inside tunnel is still IPv4)
- Change external IP lookup to using myip.opendns.com via DNS
- Option to choose IPv4 listen address from external IP lookup
- Server support for handling multiple domains via wildcard


net/iodine/patches/101-musl-workaround-incomplete-nameser-h.patch
```diff
@@ -14,14 +14,20 @@ the `osflags` file.
 An upstream fix for musl libc has been submitted with
 http://www.openwall.com/lists/musl/2017/12/04/3
 
+The fix had since been merged, and 22 months later been removed again,
+noting that it should not be required anymore in the future:
+https://git.musl-libc.org/cgit/musl/commit/include/arpa/nameser.h?id=71d23fbeacdd1f0c5e33cffb0979250d576fcf9c
+
+So this patch can likely be removed. Testing is required, though.
```
BKPepe, line 21: Then drop it and we will see.

jefferyto:
CI is failing as the patches need to be refreshed - run `make package/iodine/refresh`.

Also your commit is missing a `Signed-off-by` line, see [Sign your work](https://openwrt.org/submitting-patches#sign_your_work).

neheb:
Needs refreshing

jefferyto:
Refreshing doesn't mean changing `PKG_RELEASE`, it means you should run `make package/iodine/{clean,refresh}` to refresh the patch.

Tabiskabis:
I admit defeat before 

> Refreshing doesn't mean changing `PKG_RELEASE`, it means you should run `make package/iodine/{clean,refresh}` to refresh the patch.

I have no idea what went wrong with in my previous attempts. Make refresh kept restoring file/contents from previous internal git commits. That pkg_release update was fat-finger, though.

net/iodine/patches/101-musl-workaround-incomplete-nameser-h.patch
Tabiskabis: Dropped and incremented pkg_release version. Is that right? Or would the pkg_release need to be incrementend anyways, in case the version without that patch does not correctly work on big endian systems?

BKPepe: It depends on the situation. You did it it alright.

feckert:
@Tabiskabis I would suggest to merge this first.
Please  merge this commit in the pullrequest do one, and set the PKG_RELEASE to 1. So we have the commit message `iodine: upgrade to version 0.8.0`. After that we can merge https://github.com/openwrt/packages/pull/21465 with the PKG_RELEASE number set to 2.

Explanation: When updating the version PKG_VERSION, the PKG_RELEASE number is set to 1. If the PKG_VERSION does not change, but something is done to the package (e.g. updating the uci config or adding a compile fix), then the PKG_RELEASE version is increased by one.

BKPepe:
Hmm, even the previous variant was alright and if you would be using Git in CLI, then you will be confused why your patch does not increase PKG_VERSION. In any case, not sure, if we prefer NoReply address from GitHub.

Otherwise, don't reinvent a wheel here. It is not necessary.

Tabiskabis:
> not sure, if we prefer NoReply address from GitHub.

I see. I'll have to set up some anti spam address (with a mechanism that is not mv /dev/null), then. This will take me some time.

BKPepe:
Is there any progress?

PR #21404, opened 2023-06-18
isc-dhcp: refuse to add empty DHCP ranges

yogo1212:
Maintainer: @pprindeville
Compile tested: no
Run tested: no

Description:

With [#12925](https://github.com/openwrt/openwrt/pull/12925), 'BROADCAST' will no longer be set if there is no local broadcast address (instead of setting the global broadcast address). Prepare for the merger but stay compatible with the old version of ipcalc.

### DIFF
--- a/net/isc-dhcp/files/dhcpd.init
+++ b/net/isc-dhcp/files/dhcpd.init
@@ -373,7 +373,9 @@ gen_dhcp_subnet() {
 		echo " range $START $END;"
 	fi
 	echo " option subnet-mask $netmask;"
-	if [ "$BROADCAST" != "0.0.0.0" ] ; then
+	# check for 0.0.0.0 until all active releases of ipcalc.sh omit it
+	# for small networks:
+	if [ -n "$BROADCAST" ] && [ "$BROADCAST" != "0.0.0.0" ] ; then
 		echo " option broadcast-address $BROADCAST;"
 	fi
 	if [ "$dynamicdhcp" -eq 0 ] ; then
--- a/net/isc-dhcp/files/dhcpd.init
+++ b/net/isc-dhcp/files/dhcpd.init
@@ -365,6 +365,12 @@ static_hosts() {
 	config_foreach static_host_add host "$@"
 }
 
+eval_ipcalc() {
+	local output
+	output="$(ipcalc.sh "$@")" || return
+	eval "$output"
+}
+
 gen_dhcp_subnet() {
 	local cfg="$1"
 
@@ -438,7 +444,10 @@ dhcpd_add() {
 
 	dhcp_ifs="$dhcp_ifs $ifname"
 
-	eval "$(ipcalc.sh $subnet $start $limit)"
+	if ! eval_ipcalc "$subnet" "$start" "$limit"; then
+		echo "invalid range params: subnet: $subnet start: $start limit: $limit" >&2
+		return 1
+	fi
 
 	config_get netmask "$cfg" "netmask" "$NETMASK"
 	config_get leasetime "$cfg" "leasetime"


Ansuel:
@yogo1212 please bump pkg release and mention in the other pr that this should be merged first.

yogo1212:
@Ansuel Oops!
Thanks - will do!

yogo1212:
A second commit uses the return value of ipcalc.sh to abort generating the config from a UCI section if the range wouldn't be valid.

net/isc-dhcp/files/dhcpd.init
```diff
@@ -350,7 +350,9 @@ gen_dhcp_subnet() {
 	echo "subnet $NETWORK netmask $NETMASK {"
 	echo " range $START $END;"
 	echo " option subnet-mask $netmask;"
-	if [ "$BROADCAST" != "0.0.0.0" ] ; then
+	# check for 0.0.0.0 until all active releases of ipcalc.sh omit it
+	# for small networks:
+	if [ -n "$BROADCAST" -a "$BROADCAST" != "0.0.0.0" ] ; then
```
pprindeville, line 360: Should be `[ ... ] && [ ... ]` not `[ ... -a ... ]`.

net/isc-dhcp/files/dhcpd.init
```diff
@@ -344,6 +344,11 @@ static_hosts() {
 	config_foreach static_host_add host "$@"
 }
 
+eval_ipcalc() {
```
pprindeville, line 347: Why is this necessary?  Also, see PR #21410

pprindeville (changes requested):

net/isc-dhcp/files/dhcpd.init
yogo1212: This change is coming from https://github.com/openwrt/openwrt/pull/12925 and is only slightly related.
Instead of enabling modes of configuration, it is concerned with error handling.

The synopsis is: ipcalc.sh now fails if the range doesn't fit into the subnet.
The purpose of the function is to pass through the return code when using `eval`. While we're at it, don't call eval if ipcalc.sh fails.
A similar patch is pending for dnsmasq.

pprindeville: Hmm... why not wrap it with `export -n "$(...)"` instead?

yogo1212: Hmm, the quotes won't work. `IP` would be set to the entire output of `ipcalc.sh`.
xargs has no `-d` in busybox, also xargs won't work with the export built-in.

A more correct solution would be `while read LINE ; do export -n "$LINE"; done < <(ipcalc.sh 11.53.100.0 30 1 2)`.
If you want, you can grap that and pour it into a commit.
`ipcalc.sh 11.53.100.0 30 1 2 | while read LINE ; do export -n "$LINE"; done` doesn't work because the `export` runs in a subshell (because of the pipe).

Tbh, I don't think the added safety is worth the added complexity. The output of ipcalc.sh is intended for something like `eval`.

john-tho:
Ping on this PR. @pprindeville are there any remaining problems here?
It would be great to get at least the first commit (`BROADCAST` not null) applied, so we can proceed with getting the ipcalc fixes applied in the main repo.

yogo1212:
the `local var=$(false)` bug was fixed and the PR rebased.

yogo1212:
@pprindeville Is there something i can help with?
We'd like to get the PR moving in the main repo because it fixes ipcalc for IPv4 prefixes >30..

pprindeville (commented):
I still there might be a better solution that what `eval_ipcalc()` does, but we can circle back and fix that later.  I won't hold this up.

pprindeville (approved):
LGTM

pprindeville:
See PR #22414.

PR #21409, opened 2023-06-19
ser2net yaml config

morganchristiansson:
Maintainer: @mhei
Compile tested: n/a
Run tested: 23.05.0-rc1, init script generates expected ser2net.yaml

Description:
I'm trying to use [Slaesh CC2652 zigbee usb adaptor](https://slae.sh/projects/cc2652/) with OpenWRT which requires yaml config to work.

This PR is a bit rough but should be OK for some broad comments/feedback. I know there's some nitpicking, consistency and testing to address.

Some unresolved issues:
- parity is undocumented ~~/ not entirely supported but it should accept the values~~ [gensio parses all parity types from speed arg](https://github.com/cminyard/gensio/blob/ded2b618543d38e56c27d4f3ee67caad0b661b80/lib/sergensio_serialdev.c#L164-L168)
- ~~protocol rawlp is not supported~~ now `connector: serialdev(wronly=true)`
- ~~controlport is now admin ??~~ yes readconfig.c and yamlconfig.c [both call controller_init](https://github.com/search?q=repo%3Acminyard%2Fser2net+controller_init+language%3Ac&type=code)
- syntax for proxy "Extra options" `ser2net.@proxy[0].options` need to be updated by user. Otherwise should remain compatible with current uci config
- protocol=OFF - should remove this option. Use proxy.enable to set enable property in yaml instead.

TODO for luci-app-ser2net future PR:
- Remove protocol=off
- Rename title control port to admin port

btw thanks for upgrading to ser2net 4.3.5 in #18105

this is my first contribution from a longtime openwrt user


karlp:
Sorry, why does this other thing need a yaml config?  This is just a different config format for ser2net?

morganchristiansson:
For my zigbee usb device to work I need to set `dtr=off,rts=off` or the device goes into firmware mode and doesn't work.

This requires zigbee2mqtt 4.x and yaml config. See these issues for more background on my usecase:
- https://github.com/Koenkk/zigbee2mqtt/issues/6437
- https://github.com/cminyard/ser2net/issues/46

ser2net.conf is legacy/deprecated so we should switch at some point ...

mhei:
I really appreciate your work on this. I agree, that switching to yaml is important since the older config format will probably be dropped one day. Will have a deeper look at upcoming weekend, but the current PR already looks quite promising.

morganchristiansson:
Thank you. I'll keep chipping away at remaining issues to try to match existing uci config as close as possible.

Need to make some minor changes to luci-app-ser2net also.

It could make sense to make larger changes and re-think the UI to match the new ser2net.yaml structure, but I assume we want to keep backwards compatibility. There's a lot of new options available in ser2net.yaml.

Also since ser2net is not in any images I suppose we can push incremental upgrades if needed?

morganchristiansson:
Is ipmi sol supported? I notice deassert_CTS_DCD_DSR_on_connect option in ser2net.init which only applies to impi sol. 

Need to set yaml `connector: ipmisol,$device,...` for this instead of `connector: serialdev,$device,...`. Can maybe detect if $device is not a file as there's no option for connector type in uci config currently.

morganchristiansson:
The init script WorksForMe and it's a relief to have it work reliably without hacks or manual commands.

My use-case is simple and I haven't tested admin control or all options. Need feedback.

I've kept compat with existing uci config and luci settings. Old init script has settings that are not in luci ui, not sure if they need to be supported or what requirements are.

Perhaps in future release these should change to better match new yaml config. 

I appreciate this is not a quick PR :smile:  thanks for all help to progress.

mhei:
Sorry for being not as responsive as desired.

> It could make sense to make larger changes and re-think the UI
> to match the new ser2net.yaml structure, but I assume we want
> to keep backwards compatibility.

I think we should focus on backwards compat for the moment. The new gensio backends provide so many new possiblities, creating a UI which covers all use-cases will be hard. I fully agree with the incremental approach.

> Is ipmi sol supported?  ... deassert_CTS_DCD_DSR_on_connect option
> .. was added by @mhei...

Yes, I added it together with a bunch of other options. But personally, I don't use IPMI SOL. I'd say that we leave it out for now to keep things simple. Users could add it in the /etc/ser2net.yaml part if needed.

I enabled admin control in my setup and it seems to work as before.

But I stumpled over an empty "options:" part in the generated YAML file:
ser2net (or yaml library) complains about "Unexpected scalar value" when a line
```
connector: ...
  ...
  options:
connector: ...
  ...
```

is present. I think we need to suppress this line when no options are added.

In the makefile, we need to switch from ser2net.conf to ser2net.yaml in the
install section.

And I'm thinking about a migration path of existing `/etc/ser2net.conf` when the package is upgraded. If an unmodified default file is present, then we don't need to care about. But in case the file is modified by user, then it should be at least copied over/kept - maybe as ser2net.conf.old or so. Add more logic for migration would be overkill, what do you think?

morganchristiansson:
> Sorry for being not as responsive as desired.

I know the feeling :heart: 

> personally, I don't use IPMI SOL.

Ok - was it requested? Is noone likely to be using it? Dropping it sounds great would rather add proxy connector type in luci.

> admin control in my setup seems to work as before

Amazing when things work without testing :laughing: 

> I stumpled over an empty "options:"

I noticed this too, can't think of a clean way to detect this. It's followed by 2 for loops with nested config_get[_bool]. A "dirty" fix is quite obvious to implement...
I think it's just a warning? I ended up adding kickolduser: true as a workaround... Agree we should fix it.

Maybe something like this?
```
local echo_options=1
for ....
  [ -n "$val" ] || continue
  if [ -n "$echo_options" ]; then echo "  options:" ; echo_options=0 ; fi
```
> makefile, we need to switch from ser2net.conf to ser2net.yaml

Added commit.

> existing `/etc/ser2net.conf`

I did not realize init script allows using /etc/ser2net.conf. I guess it you enable ser2net but leave defaults&proxies empty then prepends it to /tmp/ser2net.conf/yaml. Can we even detect if ser2net.conf has been modified?

ser2net already parses ser2net.conf I don't think migration tool is in scope for us to implement :-) Can we flag breaking changes when upgrading? We already bumped major version.. hmm

mhei:
Your idea was great, I just used it. I also adjusted some other parts - maybe you can have a look.
I had no time yet to re-think about the update procedure etc.  and what user expectations could be...
Just wanted to keep you in the loop here.

net/ser2net/files/ser2net.init
```diff
@@ -21,34 +21,41 @@ list_cb_append() {
...
+			case "$stopbits" in
+				1) ;;
+				2) ;;
+				*) return 1
+			esac
+		fi
+
+		echo "default:"
+		echo "  name: speed"
+		echo "  value: $baudrate${parity:+$parity$databits$stopbits}"
```
morganchristiansson, line 51: more strict handling looks good.

net/ser2net/files/ser2net.init
```diff
@@ -21,34 +21,41 @@ list_cb_append() {
...
-	echo "default:"
-	echo "  name: speed"
-	echo "  value: $baudrate$parity$databits$stopbits"
-
 	for key in chardelay_scale chardelay_min; do
 		config_get val "$cfg" "$key"
 		[ -n "$val" ] || continue
 		key=`echo "$key" | tr '_' '-'`
-		echo "default: &$cfg_$key"
+		echo "default:"
```
morganchristiansson, line 58: I'm a bit nervous about removing the &$cfg_$key anchor.
Strictl yaml doesn't allow duplicate keys and I once took down a prod environment when the yaml parser overwrote the previous values.
Should be OK Just test that duplicate keys works as expected I guess.

net/ser2net/files/ser2net.init
```diff
@@ -90,90 +100,114 @@ ser2net_led() {
...
 	echo "  timeout: $timeout"
 	[ "$enabled" -eq 0 ] && echo "  enable: off"
-	echo "  connector: serialdev,$device,$baudrate$parity$databits$stopbits$options"
-	echo "  options:"
+	echo "  connector: serialdev,$device${baudrate:+,$baudrate${parity:+$parity$databits$stopbits}}${options:+,$options}${custom_options:+,$custom_options}"
 
 	for key in led_tx led_rx; do
 		config_get val "$cfg" "$key"
 		[ -n "$val" ] || continue
+		[ "$echo_options" -eq 1 ] && echo "  options:" && echo_options=0
```
morganchristiansson, line 197: Looks better than I thought it would. great!

net/ser2net/files/ser2net.init
```diff
@@ -90,90 +100,114 @@ ser2net_led() {
...
 		key=`echo "$key" | tr '_' '-'`
-		echo "    $key: $val"
+		echo "    $key: *$val"
```
morganchristiansson, line 199: Is this correct? the * refers to an anchor per yaml spec. (Examples under `define:` key in [stock ser2net.yaml](https://github.com/cminyard/ser2net/blob/master/ser2net.yaml))

morganchristiansson:
Thanks! I ran `git reset --hard mhei/ser2net-mhei && git push -f` to add your commit to PR.

I added some comments not much to add. Looks good to me! Should start a PR on luci-app-ser2net also.

The new init script has worked to communicate with my zigbee usb stick without issue since last comment 3 weeks ago. I guess I should try with your changes now...

net/ser2net/files/ser2net.init
mhei: To be honest, I just had a look at the shipped template of ser2net and there are no anchors - this is why I thought that they are not strictly required. But I have no strong opinion - we can also leave it in if you prefer.

net/ser2net/files/ser2net.init
mhei: Yes, tested it. I used a Raspi with two GPIO LEDs. The UCI config looks like:
```
config proxy
         ...
        option led_tx 'tx'
        option led_rx 'rx'

config led 'tx'
        ...

config led 'rx'
        ...
```
Works for me(tm). :smile:

mhei:
In my eyes, this is in a really good state now. What do you think? I think you could unflag as draft...

net/ser2net/files/ser2net.init
morganchristiansson: You are likely right. I'm just cautious from a mistake I made 10+ years ago.
The yaml spec does require unique non-duplicate keys.
And alias/anchors shouldn't matter if they are duplicate.
The ser2net repo [ser2net.yaml](https://github.com/cminyard/ser2net/blob/master/ser2net.yaml
) uses anchors for all blocks except `default:`. 

Anyhow, ser2net uses [libyaml](https://pyyaml.org/wiki/LibYAML) which is a stream parser. It's likely only an issue if parsing into memory and then accessing memory by key...

Nonetheless I'll quickly tag @cminyard for his comment. Tho maybe not worth making any fuss over.

morganchristiansson:
It's good with me. Just need to do a PR for luci-app-ser2net to update admin UI slightly should be quick to do. And it's not blocking.

cminyard:
This pretty much all looks good.  If a second define: is done with the same name, it new one will replace it, you won't get an error.

A few notes on other things:

On
```
connector: serialdev(wronly=true)
```
As of gensio 2.6.7, there is a "dev" gensio that lets you open a device without doing termios on it.  That is preferred for rawlp now, and can be used for other things, like /dev/null and such.

deassert_CTS_DCD_DSR_on_connect is for IPMI SOL only.  Unless you are compiling against a relatively recent version of OpenIPMI, that's not going to have much meaning.  Of course, IPMI is a pretty common thing nowadays on servers, so maybe it's useful.

ser2net can do encrypted and authenticated connections.  That is something I'd like to see people using.  The trouble is that you have to use ssl on the other end, too.  This is 2023, I'd like to move people to using encryption on all connections.  Any suggestions would be welcome.

I have contemplated what it would take to build a full GUI front-end for configuring ser2net.  It would be fairly daunting to get everything represented.  A hybrid approach is probably the best, but there are things you could add that could lead to security holes, like doing "connector: pty,/bin/sh".

morganchristiansson:
Thank you! Sorry, I have been occupied elsewhere lately.

PR #21427, opened 2023-06-24
liburing: Update to v2.4

gladiac:
Maintainer: Christian Lachner <gladiac@gmail.com> (**/me**)
Compile tested: mipsel, mips74k, ipq806x, x86, arc700

Description: Update to v2.4
- Updated download URL and hash

### DIFF
--- a/libs/liburing/Makefile
+++ b/libs/liburing/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=liburing
-PKG_VERSION:=2.3
+PKG_VERSION:=2.4
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


jefferyto:
Any reason why downloading the tar.gz is preferred over tar.bz2?

gladiac:
> Any reason why downloading the tar.gz is preferred over tar.bz2?

No particular reason. Do you want me to change to tar.bz2 in this PR or when the next release lands?

jefferyto:
It would be nice to change it now but it's up to you.

gladiac:
> It would be nice to change it now but it's up to you.

Alright, I just force-pushed a new commit with the download link changed to using bz2.

PR #21444, opened 2023-06-26
knot: update to version 3.2.8

Payne-X6:
Compile tested: ARM Cortex-A9 vfpv3, Turris Omnia, TurrisOS 7.0.0 (hbl)
Run tested: ARM Cortex-A9 vfpv3, Turris Omnia, TurrisOS 7.0.0 (hbl)

Description:

- Release upgrade (https://www.knot-dns.cz/2023-06-26-version-328.html)

### DIFF
--- a/net/knot/Makefile
+++ b/net/knot/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=knot
-PKG_VERSION:=3.2.7
+PKG_VERSION:=3.2.8
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://secure.nic.cz/files/knot-dns/
-PKG_HASH:=d3b7872ac8aa80f7f54ddb1bb3b1e2f90ec55f7270a2c4a9338eab42b7d2767b
+PKG_HASH:=ef419a428f327def77780bc90eda763b51e6121fe548543da84b9eb96a261a6e
 
 PKG_MAINTAINER:=Daniel Salzman <daniel.salzman@nic.cz>
 PKG_LICENSE:=GPL-3.0 LGPL-2.0 0BSD BSD-3-Clause OLDAP-2.8


salzmdan:
I don't know if I should approve the PR when I don't have permissions for that.
LGTM anyway.

BKPepe:
@salzmdan Yyou can approve any PR even if you think you don't have permissions. Anyone can do that. The button is under ``Files changed`` tab.

salzmdan:
> @salzmdan Yyou can approve any PR even if you think you don't have permissions. Anyone can do that. The button is under `Files changed` tab.

Screenshot_2023-06-26_16-51-47

:thinking:

BKPepe:
Eeehh... that's weird. Yeah, you are sure. We should allow it (again), if it is possible. 

BTW: Sorry about that! I am kinda suprised that it does not work anymore.

BKPepe:
Hey @aparcar or anyone else with the access to ``Settings tab`` for the repository. Not sure, who configured it as it should be disabled by default. 

Can you uncheck under ``Moderations options - ``Code Review limits`` this option:

image

More details here: https://github.blog/2021-11-01-github-keeps-getting-better-for-open-source-maintainers/

Also, while at it, can you check under ``General`` in menu to **require sign off**?
image

Ansuel:
@BKPepe the sign-off it's already like that
For the review it's a rather recent change and we decided to enable that as we notice many user approving pr randomly just to give attention to it.

People can still add comments to code or find other feature. Only approve and code suggestion are disabled.

BKPepe:
>  we decided to enable that as we notice many user approving pr randomly just to give attention to it.

Who decided that? Was there any vote about that? Too bad that I was not invited to participate in that vote, as we should really have it enabled.

Since I know that @salzmdan is in charge of Knot DNS, can we give him permission to approve related PRs to his project, please?

Ansuel:
> Who decided that? Was there any vote about that? Too bad that I was not invited to participate in that vote, as we should really have it enabled.

no vote, me and ynezz discussed on IRC as we notice this problem on many pr on main. (it was a bigger discussion in search of an idea to trigger CI only on pr approval) Sorry if this was a bad move but I didn't tought it was a big change that required a vote. (did the same for other very minor thing like requiring SoB and other security change like read by default on every CI action for the org)

About permission it might not be a bad idea but I have to check what github can offer on role and stuff.

BKPepe:
@Ansuel is there something new about this? I don't feel that is right when maintainer of the package is liking the OP instead of giving approval. That feels like a weird approach, though, because anyone can like or do emojis for each post and if there are going to be many people, you can miss the maintainer username or as well it can be withdrawn anytime.

Ansuel:
I wonder if a solution might be create a role on GitHub to permit approval.
We would use a simple grep and sed to add the members for pkg maintainers.
Think this would require a vote and first an email about this idea. If I
have silent feedback or negative ones I will just revert the block. What do
you think?

Il Lun 4 Set 2023, 14:00 Josef Schlehofer ***@***.***> ha
scritto:

Issue #13909, opened 2020-11-13
mwan3: Ping echo reply going over the wrong WAN interface from external hosts


jamesmacwhite:
Maintainer: @feckert 
Environment: 19.07.3 (mwan3 2.8.14-2)

Description:

I originally posted about an interesting issue I observed with ICMP and replies going over the wrong interface on the forums a while ago. As a bit of background there are services that can track your connection by sending ICMP data every second or at regular intervals as a way of monitoring uptime. In my case, due to having multiple WANs I have multiple monitors pinging different IPs but they are terminating on the same router still. I don't block ICMP inbound, contrary to all the sites that say ICMP should be stealth (it's total BS). 

https://forum.openwrt.org/t/wan-ping-echo-reply-going-over-the-wrong-interface/75317

The behaviour observed with tcpdump I'm assuming is because mwan3 and the default rule will be to use a primary WAN and fallback to the next one in the policy if not available, however this will mess up the replies for a WAN interface that isn't the primary. I tend to find sometimes the ICMP traffic is correctly routed then it might just suddenly show request timeouts, because the echo reply is then going over the wrong interface as seen by tcpdump example.

Is the solution to define a specific mwan3 rule before the default rule that targets ICMP specifically and the src_ip being the external IP address being pinged from whatever host?

```
config rule 'isp2_icmp'
	option proto 'icmp'
	option src_ip '87.xx.xx.xx'
	option family 'ipv4'
	option use_policy 'isp2_only'

config rule 'isp3_icmp'
	option proto 'icmp'
	option src_ip '81.xx.xxx.xxxx'
	option family 'ipv4'
	option use_policy 'isp3_only'
```

Or is there a better way? Right now I'm looking at a way to try and lock the ICMP traffic to properly go over the right interface, as without handling is seems to be OK sometimes, then will start to go wrong, doing mwan3 restart sometimes will help, but not all the time, so forcing with a specific rule seems like the best way, unless it can be specifically handled in mwan3 itself?

aaronjg:
The mwan3 conntrack rules should be handling that use case. (this is a slightly different issue than another bug report regarding to pings originating at the gateway).

What do you get with:

```
conntrack -F
conntrack -L|grep <ip of tracking server>
conntrack -E|grep <ip of tracking server>
```

The last command will show updates to the table, so you may need to wait for some failed pings before anything shows up.

jamesmacwhite:
[...]

PR #21445, opened 2023-06-26
[RFC] mwan3: Fix packet routing when WAN interface is partially up

tim-nordell-nimbelink:
Maintainer: @feckert @aaronjg 

Description:

We had some observed issues, which might be like #13909, where the wrong connection tracking mask could sometimes occur while an interface was being brought up, and thus the wrong routing table applied  This was in regards to incoming pings while bringing up a WAN interface.  I don't think the conclusion that it's a conntrack issue is really the culprit - but rather the interplay of the order events are captured during bringup of an interface within the `mwan3` package.

I have a detailed description within the commit which is re-explaind below, but what was happening is that the `rtmon` rules process much faster than the interface hotplug events.  As such you end up in a state where the `rtmon` added the new ranges into `ipset`, which in turn marks those packets (and thus connection) with the system wide routing table instead of matching the WAN specific table.

There are probably a couple of valid approaches, really, to mitigate this:
1. Have `rtmon` be aware if it's tracking interfaces that belong to WAN or non-WAN devices
2. Modify the middle check where the firewall currently checks that a packet is destined for a local subnet to also include checking that they were sourced from a local subnet

For the first option, the `rtmon` code would need to additionally keep track of if a given interface belongs to a WAN interface or not.  If any interface is in the middle of being brought up, this might get tricky as openwrt may not know if an interface belongs to a non-WAN device just yet since these `rtmon` events are fired off immediately rather than when OpenWRT is done configuring something.  As such, I'm not sure I would trust the `l3_device` or `device` nodes within `ubus call network.interface dump`.

The second option seems to be match the intent, for what I can tell, of how mwan3 operates and this is easy to add in.  The main purpose of applying the system wide routing table is for packets where there aren't any confusions in regards to a default route for the system.  By adding in the additional constraint that these packets are also *sourced* from a local subnet, this mitigates this. 

After adding in this additional check that the packet was sourced from the local subnet, these packets end up with the "last_resort" method applied.  Objectively, any packets that end up in the "last_resort" don't actually have a connection started (other than when "default" is set as the option) since these packets end up being dropped or rejected.  To allow these "connections" to actually establish once the WAN interface has fully gotten up, the rules were modified to make the blackhole and unreachable options as non-sticky.  That is, these specifically are cleared now every time a packet comes into the firewall to allow it to re-evaluate which interface these should go out of.

This patch was tested against an older version of mwan3 (2.10.13), but reformated to match the current version for inclusion.

I marked this as a draft since I wanted to make sure the approach was okay before marking it as a full PR, and I also wasn't sure if the version should be bumped in the Makefile directly as a part of this commit.


tim-nordell-nimbelink:
I'm realizing this solves the edge case where the very first WAN interface is coming up, but not the second edge case where a subsequent WAN interface is coming up while any other WAN interfaces are already up.  I'll have to think on how to additionally solve that.  E.g. the firewall needs to know if it's safe to apply the "last_resort" policy or not versus known WAN interfaces.

tim-nordell-nimbelink:
I've reworked the patch to be optional and less intrusive to the overall mwan3 flow.  I realized I could match packets originating from a WAN interface just post the WAN interface check by using matching packets that the source address doesn't match any of the ipsets set locally.  Additionally, I caught a bug in the interface removal code introduced; this is addressed in the first commit in this series now.

GeorgeSapkin:
No updates in over two years and still marked as draft. Closing as stale. Feel free to reopen as necessary.

PR #21453, opened 2023-06-28
Added checks to determine if crontabs have changed

hickey:
Maintainer: Florian Eckert <fe@dev.tdt.de> / @feckert 
Compile tested: x86_64
Run tested: x86_64

Description:

This change allows `micrond` to determine that there was a modification (or deletion) of a crontab file. At the completion of each event loop, the modification times of each file in the crontab directory is checked to determine if it is greater than the last time the files were read in. A final check is performed to determine if the list of crontab files has decreased (i.e. files were deleted). If any of these conditions are true, the current jobs will be dumped and all the files will be read in again. 

Notes about how change was tested: I created and tested this change on a OSX box. There is a difference in the string library--specifically the `sscanf()` function. While everything compiled and worked as expected, the `sscanf()` function was not able to match the `crontab` lines correctly due to the `m` modifier in the `sscanf()` calls. So even though the files had valid crontab entries, the `handle_line()` call would fail to identify the line as a crontab entry. Since there are no changes to the way that files are being parsed I am considering this as nothing that would break existing functionality. 

The new `check_for_updates()` function was tested by running `micrond` against a sample crontab entries. The creation of files, modification of files and the deletion of files triggered the `refresh_jobs()` function just as expected.

### DIFF
--- a/utils/micrond/src/micrond.c
+++ b/utils/micrond/src/micrond.c
@@ -36,6 +36,7 @@
 #include <syslog.h>
 #include <time.h>
 #include <unistd.h>
+#include <sys/stat.h>
 
 
 typedef struct job {
@@ -64,6 +65,9 @@ static const char *crondir;
 
 static job_t *jobs = NULL;
 
+/* used to determine if crontabs have been changed */
+static int num_checked_files = 0;
+static time_t last_refresh = 0;
 
 static void usage(void) {
 	fprintf(stderr, "Usage: micrond <crondir>\n");
@@ -151,7 +155,7 @@ static int handle_line(const char *line) {
 	int matches = sscanf(line, "%ms %ms %ms %ms %ms %n", &columns[0], &columns[1], &columns[2], &columns[3], &columns[4], &len);
 	if (matches != 5 && matches != 6) {
 		if (matches <= 0)
-			ret = 0;
+			ret = 0;		/* return we have no crontab entry */
 
 		goto end;
 	}
@@ -190,7 +194,7 @@ static int handle_line(const char *line) {
 	jobp->next = jobs;
 	jobs = jobp;
 
-	ret = 0;
+	ret = 1;		/* return that we have valid crontab */
 
   end:
 	for (i = 0; i < matches && i < 5; i++)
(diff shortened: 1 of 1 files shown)


feckert:
@NeoRaider could you take a look at this?

Until now, if someone changes the micrond directory, the service/person who made the change has to restart micrond. This would no longer be the case if we merge this.

stangri:
@hickey Thank you for the contribution. What's the CPU/RAM penalty for this change compared to pre-PR micrond?

hickey:
@stangri Doh! Was not thinking that C did not have garbage collection. My bad for being a polyglot. 

I have added a loop to free all the job structs prior to refreshing the jobs. Beyond that, there is an `int` and a `time_t` used to detect when there has been a change in the crontabs. That is about the only real memory being used. 

As for CPU, yes, there is a bit of CPU used every minute to perform a `stat` on each file and compare its modification time, but that can not be avoided. 

I was just running some basic tests with this new loop and I am seeing a segfault when the files change. Not sure what is happening at the moment. It might be a couple of days before I get a debugger looking at the execution to get a good idea of the problem. I will let you know what I find and when it is been fixed.

stangri:
@feckert can you pull someone else for review? I believe this is a welcome update.

neocturne:
Two thoughts I have on the matter:

- I don't expect micrond crontabs to be touched by humans - Busybox's crond should be fine for admin-managed jobs. Maybe packages installing crontabs should just restart micrond in their postinst/postrm?
- Full-featured cron daemons have an extra command `crontab` to manage crontabs that will allow you to edit a copy of a crontab and then copy it back atomically, so crond will never see a half-edited version. If micrond is used for admin-managed jobs after all, re-reading the crontabs automatically might result in an incomplete version (if the file is saved in the middle of editing)

feckert:
@hickey Thanks for your contribution, and your work to add this feature to `micrond`.
But the objections raised by the author @neocturne of  mircond are entirely justified and I see this the same way.

The `mircond` is only a solution for system tasks, and should not be edited by a user. Each service that places or edit a file in the `micrond` directory should restart the `micrond` service on its own.

The crontab of the busybox was supposed to take over this task for system tasks at some point.
But it cannot do this at the moment. So we are using `mircond`.

PR #21465, opened 2023-06-28
iodine: enhance uci configuration

Tabiskabis:
Maintainer: @ukleinek
Compile tested: x86_64, Debian bookworm, OpenWrt 23.05.0-rc1
Run tested: x86_64, QEMU, OpenWRT 23.05.0-rc1: Tested start & stop functions

Description:
 * apply more arguments/options via init script
 * include enhanced config file w/ annotations
 * very small chance to breaking previous configs, since address no longer defaults to '0.0.0.0'
  iodined provides its own mechanism to do pretty much the same thing, which should be preferred
  iodined defaults to using port 53 if not specified

### DIFF
--- a/net/iodine/Makefile
+++ b/net/iodine/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=iodine
 PKG_VERSION:=0.8.0
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=http://code.kryo.se/iodine/
--- a/net/iodine/files/iodined.config
+++ b/net/iodine/files/iodined.config
@@ -1,5 +1,50 @@
+# Example configuration
+
 config iodined
-        option address     ''
-        option password    ''
-        option tunnelip    '10.0.0.1'
-        option tld     	   ''
+
+	# Enable this section? Default is 'yes' (compatible with older conffiles).
+	option enabled no
+
+	### Minimal required configuration:
+
+	# DNS traffic will be sent as queries for subdomains under this top domain.
+	option tld 'jod.example.net'
+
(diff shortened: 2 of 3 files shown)


feckert (commented):
While we are at it, refactor the init script.
We should also port it to the PROCD?

And please use the following pattern to add arguments on service start.
```
procd_open_instance
procd_set_param command /usr/sbin/iodined
[ -n "$password" ] && procd_append_param command -P "$password"
[ -n "$address" ] && procd_append_param command -l "$address"
[ -n "$port" ] && procd_append_param command -p "$port"
...
procd_set_param respawn
procd_close_instance


```

Tabiskabis:
> While we are at it, refactor the init script. We should also port it to the PROCD?
This is what you meant, i suppose? Great susggestion.
It'll probably also solve an issue where the process dies because DHCP or something.

feckert (changes requested):
Looks good.

Could you also add the following lines to the init script:
```
service_trigger() {
	procd_add_reload_trigger iodined
}
```
So the service gets restarted on a config change.

Tabiskabis:
> Could you also add the following lines to the init script:
> ...
> So the service gets restarted on a config change.

This, plus I've added the ability to handle named instances, too.

feckert (changes requested):
Unfortunately, I overlooked that in the review. But can you also please bump the PKG_RELEASE number by one?
https://github.com/openwrt/packages/blob/master/net/iodine/Makefile#L12

Tabiskabis:
> ... can you also please bump the PKG_RELEASE number by one? 

I should mention that I opened a [pull request to update iodine to v0.8](21403) (appears stuck) about a week before this one.
This being my first contribution to OpenWRT, I wasn't sure how long things take and whether they get acceptet at all.
Should I have merged this requestion into the other one, instead?

GeorgeSapkin:
@Tabiskabis still relevant? If so, there are merge conflicts and you can't use GitHub noreply email.

PR #21466, opened 2023-06-28
irqbalance: add banned_cpulist option

Hentioe:
Maintainer: Hannu Nyman <hannu.nyman@iki.fi>
Compile tested: x86_64, Ubuntu 22.04, OpenWrt 22.03.5 
Run tested: ARMv7, ASUS RT-AC42U, OpenWrt 22.03.5

Description:

`IRQBALANCE_BANNED_CPULIST`: Provides a cpulist which irqbalance should ignore and never assign interrupts to. If not specified, irqbalance use mask of isolated and adaptive-ticks CPUs on the system as the default value.

Related documents (I couldn't find the official project ones): https://manpages.debian.org/unstable/irqbalance/irqbalance.1.en.html#IRQBALANCE_BANNED_CPULIST

Note: The `IRQBALANCE_BANNED_CPULIST` variable replaces the deprecated `IRQBALANCE_BANNED_CPUS` variable, and we do not need to add configuration options for the latter.

### DIFF
--- a/utils/irqbalance/files/irqbalance.config
+++ b/utils/irqbalance/files/irqbalance.config
@@ -8,6 +8,9 @@ config irqbalance 'irqbalance'
 	# The default value is 10 seconds
 	#option interval '10'
 
+	# Specify excluded cpulist
+	#option banned_cpulist '0'
+
 	# List of IRQ's to ignore
 	#list banirq '36'
 	#list banirq '69'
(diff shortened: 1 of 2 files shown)


neheb:
ping @hnyman

hnyman:
New features should be added to the development master, not to the old stable 22.03.

Hentioe:
I found out that the name structure of 'firstname lastname' is mandatory. I had to close this pr because I couldn't pass this check anyway.

PR #21471, opened 2023-07-01
hyperscan: new package for speeding up regex ops

graysky2:
Hyperscan is a high performance regular expression matching library from Intel that runs on x86 platforms and offers support for Perl Compatible Regular Expressions (PCRE) syntax, simultaneous matching of groups of regular expressions, and streaming operations.

This has utility in speeding up snort3.

### DIFF
--- a/libs/hyperscan/Makefile
+++ b/libs/hyperscan/Makefile
@@ -1,0 +1,82 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=hyperscan
+PKG_VERSION:=5.4.2
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/intel/hyperscan/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=32b0f24b3113bbc46b6bfaa05cf7cf45840b6b59333d078cc1f624e4c40b2b99
+
+PKG_MAINTAINER:=John Audia <therealgraysky@proton.me>
+PKG_LICENSE:=BSD-3-Clause BSD-2-Clause BSL-1.0
+PKG_LICENSE_FILES:=LICENSE
+PKG_BUILD_DEPENDS:=ragel/host python3/host boost/host
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+# A minimum of SSSE3 support is required to use hyperscan
+# We need to define the C(XX)FLAGS to avoid -march=native being
+# used as this is hardcoded by upstream, see:
+# https://github.com/intel/hyperscan/blob/master/doc/dev-reference/getting_started.rst
+CMAKE_OPTIONS += \
+	-DCMAKE_INSTALL_PREFIX=/usr \
+	-DCMAKE_INSTALL_LIBDIR=lib \
+	-DBUILD_SHARED_LIBS=ON \
+	-DCMAKE_C_FLAGS="-march=x86-64-v2" \
+	-DCMAKE_CXX_FLAGS="-march=x86-64-v2" \
+	-Wno-dev
+
+define Package/hyperscan-headers
+  CATEGORY:=Libraries
+  SECTION:=libs
+  TITLE:=Hyperscan Headers
+  URL:=https://github.com/intel/hyperscan
+  DEPENDS:=@TARGET_x86_64
+endef
+
+define Package/hyperscan-runtime
+  CATEGORY:=Libraries
+  SECTION:=libs
+  TITLE:=Hyperscan Runtime
+  URL:=https://github.com/intel/hyperscan
+  DEPENDS:=@TARGET_x86_64 +libstdcpp
+endef
+
+define Package/hyperscan-headers/description
+  This package contains the headers for Hyperscan.
+endef
+
+define Package/hyperscan-runtime/description
+  This package contains the shared objects for Hyperscan.
+endef
+
+# This installs files into ./staging_dir/. so that you can cross compile from the host
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/include/hs
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/include/hs/* $(1)/usr/include/hs/
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/lib/libhs* $(1)/usr/lib/
+	$(INSTALL_DIR) $(1)/usr/lib/pkgconfig
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/libhs.pc $(1)/usr/lib/pkgconfig/libhs.pc
+endef
+
+# These install files on the target.  Compare with Build/InstallDev
+define Package/hyperscan-headers/install
+	$(INSTALL_DIR) $(1)/usr/include/hs
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/include/hs/*.h $(1)/usr/include/hs/
+endef
+
+define Package/hyperscan-runtime/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/lib/libhs* $(1)/usr/lib/
+endef
+
+$(eval $(call BuildPackage,hyperscan-headers))
+$(eval $(call BuildPackage,hyperscan-runtime))


graysky2:
@flyn-org - can you help me to enable hyperscan in snort3?  I tried adding the following but I think there is another cmake arg to pass to define the missing headers.

```
diff --git a/net/snort3/Makefile b/net/snort3/Makefile
index bd888ec47..22ce732a7 100644
--- a/net/snort3/Makefile
+++ b/net/snort3/Makefile
@@ -25,7 +25,7 @@ define Package/snort3
   SUBMENU:=Firewall
   SECTION:=net
   CATEGORY:=Network
-  DEPENDS:=+libstdcpp +libdaq3 +libdnet +libopenssl +libpcap +libpcre +libpthread +libuuid +zlib +libhwloc +libtirpc +luajit +libatomic
+  DEPENDS:=+libstdcpp +libdaq3 +libdnet +libopenssl +libpcap +libpcre +libpthread +libuuid +zlib +libhwloc +libtirpc +luajit +libatomic +hyperscan-runtime
   TITLE:=Lightweight Network Intrusion Detection System
   URL:=http://www.snort.org/
   MENU:=1
@@ -52,6 +52,8 @@ CMAKE_OPTIONS += \
        -DMAKE_PDF_DOC:BOOL=NO \
        -DMAKE_TEXT_DOC:BOOL=NO \
        -DHAVE_LIBUNWIND=OFF \
+       -DHS_INCLUDE_DIRS=$(STAGING_DIR)/usr/lib/pkgconfig \
+       -DHAVE_HYPERSCAN=ON \
        -DHAVE_LZMA=OFF

 TARGET_CFLAGS  += -I$(STAGING_DIR)/usr/include/daq3 -I$(STAGING_DIR)/usr/include/tirpc
```

When I try building snort3 with this:
```
[135/982] Building CXX object src/main/CMakeFiles/main.dir/shell.cc.o
FAILED: src/main/CMakeFiles/main.dir/shell.cc.o 
/scratch/union/staging_dir/toolchain-x86_64_gcc-12.3.0_musl/bin/x86_64-openwrt-linux-musl-g++ -DHAVE_CONFIG_H -Dinline=inline -Drestrict=__restrict -I/scratch/union/build_dir/target-x86_64_musl/snort3-3.1.64.0/src/network_inspectors -I/scratch/union/build_dir/target-x86_64_musl/snort3-3.1.64.0/src -I/scratch/union/staging_dir/target-x86_64_musl/usr/include/luajit-2.1 -I/scratch/union/build_dir/target-x86_64_musl/snort3-3.1.64.0 -I/scratch/union/staging_dir/host/include -I/scratch/union/staging_dir/target-x86_64_musl/usr/lib/pkgconfig -I/scratch/union/staging_dir/target-x86_64_musl/usr/include/uuid -I/scratch/union/build_dir/target-x86_64_musl/snort3-3.1.64.0/src/main -Os -pipe -fno-caller-saves -fno-plt -fhonour-copts -fmacro-prefix-map=/scratch/union/build_dir/target-x86_64_musl/snort3-3.1.64.0=snort3-3.1.64.0 -Wformat -Werror=format-security -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro -I/scratch/union/staging_dir/target-x86_64_musl/usr/include/daq3 -I/scratch/union/staging_dir/target-x86_64_musl/usr/include/tirpc  -fvisibility=hidden   -DNDEBUG    -DNDEBUG -std=c++14 -MD -MT src/main/CMakeFiles/main.dir/shell.cc.o -MF src/main/CMakeFiles/main.dir/shell.cc.o.d -o src/main/CMakeFiles/main.dir/shell.cc.o -c /scratch/union/build_dir/target-x86_64_musl/snort3-3.1.64.0/src/main/shell.cc
/scratch/union/build_dir/target-x86_64_musl/snort3-3.1.64.0/src/main/shell.cc:38:10: fatal error: hs_compile.h: No such file or directory
   38 | #include <hs_compile.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
```

graysky2:
@flyn-org - Once this is merged, how can we modify `net/snort3/Makefile` to include the following for x86_64?  This will allow snort3 to be compiled against hyperscan.
```
diff --git a/net/snort3/Makefile b/net/snort3/Makefile
index bd888ec47..7440802f0 100644
--- a/net/snort3/Makefile
+++ b/net/snort3/Makefile
@@ -25,7 +25,7 @@ define Package/snort3
   SUBMENU:=Firewall
   SECTION:=net
   CATEGORY:=Network
-  DEPENDS:=+libstdcpp +libdaq3 +libdnet +libopenssl +libpcap +libpcre +libpthread +libuuid +zlib +libhwloc +libtirpc +luajit +libatomic
+  DEPENDS:=+libstdcpp +libdaq3 +libdnet +libopenssl +libpcap +libpcre +libpthread +libuuid +zlib +libhwloc +libtirpc +luajit +libatomic +hyperscan-runtime
   TITLE:=Lightweight Network Intrusion Detection System
   URL:=http://www.snort.org/
   MENU:=1
@@ -51,6 +51,7 @@ CMAKE_OPTIONS += \
        -DMAKE_HTML_DOC:BOOL=NO \
        -DMAKE_PDF_DOC:BOOL=NO \
        -DMAKE_TEXT_DOC:BOOL=NO \
+       -DHS_INCLUDE_DIRS=$(STAGING_DIR)/usr/include/hs \
        -DHAVE_LIBUNWIND=OFF \
        -DHAVE_LZMA=OFF
```
Tested on x86_64:
```
# snort -V

   ,,_     -*> Snort++ <*-
  o"  )~   Version 3.1.64.0
   ''''    By Martin Roesch & The Snort Team
           http://snort.org/contact#team
           Copyright (C) 2014-2023 Cisco and/or its affiliates. All rights reserved.
           Copyright (C) 1998-2013 Sourcefire, Inc., et al.
           Using DAQ version 3.0.12
           Using LuaJIT version 2.1.0-beta3
           Using OpenSSL 3.0.9 30 May 2023
           Using libpcap version 1.10.4 (with TPACKET_V3)
           Using PCRE version 8.45 2021-06-15
           Using ZLIB version 1.2.13
           Using Hyperscan version 5.4.2 2023-07-01
```

john-tho:
You may need to change the depend in this package to include `TARGET_`, example:
https://github.com/openwrt/packages/blob/6a897e22b0cd6d06fe99f9e651b6dc4347eb8caf/utils/acpica-unix/Makefile#L30

> @flyn-org - Once this is merged, how can we modify `net/snort3/Makefile` to include the following for x86_64? This will allow snort3 to be compiled against hyperscan.

Example for the conditional depend package: https://github.com/openwrt/packages/blob/6a897e22b0cd6d06fe99f9e651b6dc4347eb8caf/utils/flashrom/Makefile#L50
Example for the conditional additional config option: https://github.com/openwrt/packages/blob/6a897e22b0cd6d06fe99f9e651b6dc4347eb8caf/net/samba4/Makefile#L222

graysky2:
@john-tho - thank you for these suggestions.  Just changed the `DEPENDS` per your guidance.  Let's see how others like this PR, if it gets merged, I will work with @flyn-org to get the `snort3` package updated correctly.

BKPepe:
What is the difference between this package and https://github.com/openwrt/packages/pull/21140 ? Do we really need to have two packages doing the same stuff? How many users will use this package?

graysky2:
@BKPepe - in theory, the draft for vectorscan you linked above works on ARM and ARM64 but there are upstream issues that have not been fixed in months.  Hyperscan (this package) is the original project, not a fork and works only for x86.  I do not recommend vectorscan at this time.

graysky2:
@BKPepe - what do you think about merging this into master?  I have been using it for the past week with stable results as have other users, see: [recent postings](https://forum.openwrt.org/t/ips-mode-of-snort3-is-not-dropping-traffic/161479/98) in this thread.

libs/hyperscan/Makefile
```diff
@@ -0,0 +1,83 @@
...
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=hyperscan
+PKG_VERSION:=5.4.2
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/intel/hyperscan/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=32b0f24b3113bbc46b6bfaa05cf7cf45840b6b59333d078cc1f624e4c40b2b99
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)
```
BKPepe, line 15: Isn't this already default?

libs/hyperscan/Makefile
```diff
@@ -0,0 +1,83 @@
...
+define Package/hyperscan-headers/description
+  This package contains the headers for Hyperscan.
+endef
+
+define Package/hyperscan-runtime/description
+  This package contains the shared objects for Hyperscan.
+endef
+
+define Package/hyperscan-runtime/install
+endef
```
BKPepe, line 59: Ehm? Why this is empty and below there is the same define, but with some files?

BKPepe (changes requested):

libs/hyperscan/Makefile
graysky2: Mistake, fixed.

libs/hyperscan/Makefile
graysky2: You are correct, removed.

graysky2:
Thanks for the review, @BKPepe - fixed and pushed

graysky2:
@BKPepe - Use xxxx in [the forum](https://forum.openwrt.org/t/some-help-with-a-makefile-gperftools/165656/45) pointed out the hardcoding of `-march=native` by upstream.  Please review my current commit which uses `-march=x86-64-v2`

The plus side is that it should allow for the most supported CPUs.  The down side is that AVX and AVX2 processing will not be built in to the package.  Do you have an opinion about which to go with?  More supported hardware or better theoretical preformance?

For reference:
```
-march=x86-64: CMOV, CMPXCHG8B, FPU, FXSR, MMX, FXSR, SCE, SSE, SSE2
-march=x86-64-v2: (close to Nehalem) CMPXCHG16B, LAHF-SAHF, POPCNT, SSE3, SSE4.1, SSE4.2, SSSE3
-march=x86-64-v3: (close to Haswell) AVX, AVX2, BMI1, BMI2, F16C, FMA, LZCNT, MOVBE, XSAVE
-march=x86-64-v4: AVX512F, AVX512BW, AVX512CD, AVX512DQ, AVX512VL
```
Source: https://www.phoronix.com/news/GCC-11-x86-64-Feature-Levels

graysky2 (commented):
Review complete/changes made

graysky2:
@john-tho 
> Example for the conditional depend package:

Would the syntax to have two args be to list the thing twice or to use a comma separated construct?
```
  DEPENDS:= \
        +(TARGET_x86||TARGET_x86_64):hyperscan-runtime \
        +(TARGET_x86||TARGET_x86_64):gperftools-runtime \
        +libstdcpp +libdaq3 +libdnet +libopenssl +libpcap +libpcre +libpthread \
        +libuuid +zlib +libhwloc +libtirpc +luajit +libatomic
```

graysky2:
@BKPepe - I applied your suggestions about install macros from https://github.com/openwrt/packages/pull/21627 to this PR.

graysky2:
@BKPepe - could you take a look at the changes?  Is the code in this PR acceptable for merging?

libs/hyperscan/Makefile
```diff
@@ -0,0 +1,85 @@
...
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/lib/libhs* $(1)/usr/lib/
+	$(INSTALL_DIR) $(1)/usr/lib/pkgconfig
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/libhs.pc $(1)/usr/lib/pkgconfig/libhs.pc
+endef
+
+# These install files on the target.  Compare with Build/InstallDev
+define Package/hyperscan-headers/install
+	$(INSTALL_DIR) $(1)/usr/include/hs
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/include/hs/* $(1)/usr/include/hs/
```
BKPepe, line 76: Do you really want to install everything?

libs/hyperscan/Makefile
```diff
@@ -0,0 +1,85 @@
...
+	-DCMAKE_INSTALL_LIBDIR=lib \
+	-DBUILD_SHARED_LIBS=ON \
+	-DCMAKE_C_FLAGS="-march=x86-64-v2" \
+	-DCMAKE_CXX_FLAGS="-march=x86-64-v2" \
+	-Wno-dev
+
+define Package/hyperscan-headers
+  CATEGORY:=Libraries
+  SECTION:=libs
+  TITLE:=Hyperscan Headers
```
BKPepe, line 42: I think, we can do this as default and then you can call it once again.  E.g. something what is in ``ffmpeg`` Makefile.

libs/hyperscan/Makefile
graysky2: Yes, there are only four header files.
```
% tree ./build_dir/target-x86_64_musl/hyperscan-5.4.2/.pkgdir/hyperscan-headers
[  60]  ./build_dir/target-x86_64_musl/hyperscan-5.4.2/.pkgdir/hyperscan-headers
└── [  60]  usr
    └── [  60]  include
        └── [ 120]  hs
            ├── [2.0K]  hs.h
            ├── [ 20K]  hs_common.h
            ├── [ 49K]  hs_compile.h
            └── [ 23K]  hs_runtime.h
```

libs/hyperscan/Makefile
graysky2: Sorry, I am confused.  What do you want me to change here?

diizzyy:
Since OpenWrt targets multiple archs it probably makes more sense to package https://github.com/VectorCamp/vectorscan

graysky2:
@diizzyy - I attempted to do so initially but there is an upstream bug preventing that from building.  Further, one needs a pretty powerful CPU to make use of IPS mode on a fast connection.  ARM devices will not cut it/throughput drops.  Even an Intel N95-based x86 machine cannot do gigabit speeds in IPS mode.  It makes the most sense to stick with this code and keep the x86 only.

libs/hyperscan/Makefile
graysky2: @BKPepe - ping, can you please clairify?

graysky2:
@neheb - would you mind reviewing this PR?

libs/hyperscan/Makefile
```diff
@@ -0,0 +1,85 @@
...
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/intel/hyperscan/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=32b0f24b3113bbc46b6bfaa05cf7cf45840b6b59333d078cc1f624e4c40b2b99
+
+PKG_MAINTAINER:=John Audia <therealgraysky@proton.me>
+PKG_LICENSE:=BSD-3-Clause BSD-2-Clause BSL-1.0
+PKG_LICENSE_FILES:=LICENSE
+PKG_BUILD_PARALLEL:=1
```
neheb, line 19: No need.

libs/hyperscan/Makefile
```diff
@@ -0,0 +1,85 @@
...
+CMAKE_INSTALL:=1
```
neheb, line 20: No need.

libs/hyperscan/Makefile
graysky2: To be clear - you are suggesting that I remove myself as the `PKG_MAINTAINER`?

libs/hyperscan/Makefile
graysky2: Removed these 4 lines.

neheb:
No I meant just build parallel and CMake install.

libs/hyperscan/Makefile
graysky2: Per your comment, only removed the parallel and cmake install lines

graysky2:
Got it, adjusted.

BKPepe:
> I attempted to do so initially but there is an upstream bug preventing that from building.

Have you tried to consult it with vectorscan developers and filled a bug report?

libs/hyperscan/Makefile
BKPepe: Yeah, if you look at ``fmpeg`` Makefile, you can see that something like this:
```
define Package/hyperscan/Default
  CATEGORY:=Libraries
  SECTION:=libs
  URL:=https://github.com/intel/hyperscan
  DEPENDS:=@TARGET_x86_64
endef

define Package/hyperscan-headers
  $(call Package/hyperscan/Default)
  TITLE:=Hyperscan Headers
endef

define Package/hyperscan-runtime
  $(call Package/hyperscan/Default)
  TITLE:=Hyperscan Runtime
  DEPENDS+= +libstdcpp
endef
```


could be better. :)

libs/hyperscan/Makefile
```diff
@@ -0,0 +1,82 @@
...
+PKG_BUILD_DEPENDS:=ragel/host python3/host boost/host
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+# A minimum of SSSE3 support is required to use hyperscan
+# We need to define the C(XX)FLAGS to avoid -march=native being
+# used as this is hardcoded by upstream, see:
+# https://github.com/intel/hyperscan/blob/master/doc/dev-reference/getting_started.rst
+CMAKE_OPTIONS += \
```
BKPepe, line 28: We should define here BUILD_TYPE. Is there any specific reason, why do you need by default to have debug symbols enabled? Wouldn't it be sufficient just to use ``Release`` type without debug symbols?
Is there any size difference? There should be right? Could you check how much and submit pull request, please if it is worth doing it?

libs/hyperscan/Makefile
```diff
@@ -0,0 +1,82 @@
...
+	-DCMAKE_INSTALL_PREFIX=/usr \
+	-DCMAKE_INSTALL_LIBDIR=lib \
+	-DBUILD_SHARED_LIBS=ON \
+	-DCMAKE_C_FLAGS="-march=x86-64-v2" \
+	-DCMAKE_CXX_FLAGS="-march=x86-64-v2" \
```
BKPepe, line 33: This is not right, why do you need to explicitly specified it? Why while doing cross-compilation the defaults are not working? Have you looked at it further?

PR #21472, opened 2023-07-01
inadyn: build both openssl and mbedtls variants

robho:
Maintainer: none
Compile tested: mipsel_24kc, Netgear R6220, master
Run tested: mipsel_24kc, Netgear R6220, 23.05 branch, verified that updating dynamic dns address works

### DIFF
--- a/net/inadyn/Makefile
+++ b/net/inadyn/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=inadyn
 PKG_VERSION:=2.11.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://github.com/troglobit/inadyn/releases/download/v$(PKG_VERSION)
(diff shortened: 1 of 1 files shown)


net/inadyn/Makefile
```diff
@@ -24,32 +24,40 @@ PKG_BUILD_PARALLEL:=1
...
+  --disable-static \
+  --with-pic
 
+ifneq ($(CONFIG_PACKAGE_inadyn-mbedtls),)
 CONFIGURE_ARGS += \
-	--enable-shared \
-	--disable-static \
-	--enable-openssl \
-	--with-pic
+  MbedTLS_CFLAGS="-I$(STAGING_DIR)/usr/include" \
```
neheb, line 49: curious about this. why?

BKPepe:
Since you are introducing breaking change, there should be ``provides`` to one of the package to have inadyn package.

net/inadyn/Makefile
robho: inadyn's configure script fails to automatically find mbedtls because it looks like mbedtls doesn't provide a pkg-config file.

Is there some better way than manually specifying these compile flags?

robho:
I added `PROVIDES` and hopefully fixed the build problems.

BKPepe (changes requested):

net/inadyn/Makefile
```diff
@@ -24,32 +24,43 @@ PKG_BUILD_PARALLEL:=1
...
+define Package/inadyn/Default
+  TITLE:=A Dynamic DNS client with SSL/TLS support (uses $(1))
+  VARIANT=$(2)
   SECTION:=net
   CATEGORY:=Network
-  DEPENDS:=+confuse +libopenssl +ca-certificates
-  TITLE:=A Dynamic DNS client with SSL/TLS support
-  URL:=http://troglobit.com/project/inadyn/
   SUBMENU:=IP Addresses and Names
+  PROVIDES:=inadyn
```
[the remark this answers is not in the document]
robho: Thanks for the pointer!

I still don't quite understand how PROVIDES is used. Is the idea that it should work to do "opkg install inadyn" also in the future? Or, is it used by package dependencies (if a package would depend on inadyn)? Or, some other situation?

The documentation [here](https://openwrt.org/docs/guide-developer/packages) says:
> PROVIDES - (optional) allow to define a virtual package that might be provided by multiple real-packages

.. which to me contradicts what you said above:
> Right now as it is done inadyn-openssl and inadyn-mbedtls provides the same package inadyn. That's not good.

Am I reading the documentation wrong or is that documentation for some other use-case?

---

Anyway, I tried to implement the suggestions you had:
* Add PROVIDES only to inadyn-openssl
* Add CONFLICTS to inadyn-mbedl-tls
* Expand commit message with information on why an Mbed TLS variant of inadyn is built

Does it look ok?

net/inadyn/Makefile
neheb: modify the original configure script to look for libraries instead of using pkg-config.

PR #21475, opened 2023-07-01
irqbalance: add banned_cpulist option

Hentioe:
Maintainer: @hnyman <hannu.nyman@iki.fi>
Compile tested: x86_64, Ubuntu 22.04, OpenWrt 22.03.5 
Run tested: ARMv7, ASUS RT-AC42U, OpenWrt 22.03.5

Description:

`IRQBALANCE_BANNED_CPULIST`: Provides a cpulist which irqbalance should ignore and never assign interrupts to. If not specified, irqbalance use mask of isolated and adaptive-ticks CPUs on the system as the default value.

Related documents (I couldn't find the official project ones): https://manpages.debian.org/unstable/irqbalance/irqbalance.1.en.html#IRQBALANCE_BANNED_CPULIST

Note: The `IRQBALANCE_BANNED_CPULIST` variable replaces the deprecated `IRQBALANCE_BANNED_CPUS` variable, and we do not need to add configuration options for the latter.

This is a re-started PR because I made a mistake that required closing the old PR (I tried to merge a new feature into the old branch). So I apologize for the inconvenience once again.

### DIFF
--- a/utils/irqbalance/files/irqbalance.config
+++ b/utils/irqbalance/files/irqbalance.config
@@ -8,6 +8,9 @@ config irqbalance 'irqbalance'
 	# The default value is 10 seconds
 	#option interval '10'
 
+	# Specify excluded cpulist
+	#option banned_cpulist '0'
+
 	# List of IRQ's to ignore
 	#list banirq '36'
 	#list banirq '69'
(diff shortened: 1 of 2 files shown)


hnyman:
Declaring a function in the middle of the procd invocation seems strange/wrong to me.

Hentioe:
@hnyman I just want to put the logic of setting an option that may become complex in the future into one code block (within a unit). Of course, if you think it's better to remove the function and function call and extract the if statement directly, I will make the modification accordingly.

Actually this option doesn't usually get any more complicated, maybe I really shouldn't.

hnyman:
Merged as e1ff9619a  (with PKG_RELEASE bump added)

PR #21478, opened 2023-07-02
nyx: fix python 3.11 compatibility

jmarcet:
Maintainer: me
Compile tested: master x86_64
Run tested: master x86_64

Description:
Upstream has not released a new stable version since Sat Jan 12 15:06:37 2019 -0800

This version breaks with python 3.11. The last commit in the [development repo](https://git.torproject.org/nyx.git), from Mon Jun 27 19:38:34 2022 +0200 includes a fix `Replace inspect.getargspec usage to support python 3.11`

Here I include a patch to get version `2.1.0` in sync with the devel repo. I haven't tested whether the last commit is enough to get it working. I've only had to drop this chunk:

```
diff --git a/nyx/uninstall b/nyx/uninstall
index 5448f73..52cbf7d 100755
--- a/nyx/uninstall
+++ b/nyx/uninstall
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Copyright 2010-2017, Damian Johnson and The Tor Project
+# Copyright 2010-2019, Damian Johnson and The Tor Project
 # See LICENSE for licensing information
 
 files="/usr/bin/nyx /usr/share/man/man1/nyx.1.gz /usr/share/nyx"
```

I was unsure what to do with the package version, since I'm only increasing `PKG_RELEASE`.


admin/nyx/patches/nyx-2.1.0-devel.patch
```diff
@@ -0,0 +1,1526 @@
+diff --git a/README.md b/README.md
```
BKPepe, line 1: Ehmm... this patch looks really weird to me!

You should replace it by this one: 
https://github.com/torproject/nyx/commit/dcaddf2ab7f9d2ef8649f98bb6870995ebe0b893.patch

admin/nyx/Makefile
```diff
@@ -11,6 +11,8 @@ PKG_MAINTAINER:=Javier Marcet <javier@marcet.info>
 PKG_LICENSE:=GPL-3.0-only
 PKG_LICENSE_FILES:=LICENSE
 
+PYTHON3_PKG_WHEEL_VERSION=2.1.0.dev0
```
BKPepe, line 14: drop

BKPepe (changes requested):

admin/nyx/patches/nyx-2.1.0-devel.patch
jmarcet: It is a `git diff 2.1.0..HEAD`. I'll replace it.

jmarcet:
@BKPepe

BKPepe:
To be honest, I am thinking about what we should do here. In my view, we can not supply work from others or even keep alive software, which is no longer maintained[^1], and the developer/or the project itself does not care about it anymore[^2], but e.g., GNU/Linux distribution Debian does that [^3].

While looking at stem (also from The Tor Project) and this commit https://gitlab.torproject.org/tpo/network-health/stem/-/commit/6131577af5aeb79edb79bd6c5297c52a92769680, I am not quite sure what they are going to do with ``nyx``, because ``tpo gitweb`` should be removed, but nyx is still there - https://gitweb.torproject.org/nyx.git 🤷 

[^1]: https://github.com/torproject/stem/issues/130#issuecomment-1445476190
[^2]: https://github.com/torproject/nyx/issues/63 (they did not release a new version, but they did it for stem)
[^3]: https://sources.debian.org/patches/nyx/2.1.0-2.2/python3.11.patch/

commodo:
> Let me ask our Python maintainers (@jefferyto and @commodo) about their thoughts, here.

well, if @jmarcet is happy to keep this package maintained, even if upstream no longer maintains it, i am fine to let him do it;
(from my point of view) this issue is less with this being a Python package, and more with the willingness of someone to keep this package maintained downstream;

on a similar note, i have a `stress` package (in `utils/stress`) that i keep around;
i wanted to drop it for a while;
but then @neheb found this page ( https://github.com/resurrecting-open-source-projects ) and wouldn't-you-know-it?, stress is also there;

so,  an alternative idea would be for @jmarcet to start an upstream fork and point this Makefile to it (just a thought);

similarly, i've maintained for a while https://github.com/mstpd/mstpd/ upstream (and as a package here), and dropped it, because i have no time to do proper STP/RSTP/MSTP support (I even forgot a lot of fine details about this) and then funny enough OpenWrt forked this (  https://github.com/openwrt/ustp  ) from mstpd, so then I really no longer need to keep mstpd in the feed here :)

that's my view at least on this :)

jefferyto:
I also think this package can be kept as long as @jmarcet (or someone else) continues to maintain it.

stem 1.8.2 was released about a month ago, it fixes the Python 3.11 compatibility issue.

jmarcet:
As long as it is only about making sure it keeps working, I'm fine with it, one way or another I know I'm going to keep it in my local server. I don't think I'll have the time, or the willingness, to further develop it.

So, what do you say, shall I create the fork right away and modify the package accordingly? @BKPepe @commodo @jefferyto

BKPepe (approved):

BKPepe:
> start an upstream fork and point this Makefile to it (just a thought);

That's also possible solution, but right now, not needed I would say, because the compatibility Python 3.11 commit is in the developer's repository, but they did not release a new version as they did with stem.

Let's keep this as it is for now and we will see how it goes in the future. ;-)

PR #21496, opened 2023-07-04
net-mtools: Add new package to debug multicast setups

Ansuel:
Add new package to debug multicast setups. This is required to use kselftests script for network testing.

net-mtools is used instead of mtools as it does conflicts with another package that is also called mtools.

### DIFF
--- a/net/net-mtools/Makefile
+++ b/net/net-mtools/Makefile
@@ -1,0 +1,40 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=net-mtools
+PKG_VERSION:=2.3
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL=https://github.com/troglobit/mtools
+PKG_SOURCE_VERSION:=db665a4303c38cee908eba4dac50873c3f1d899c
+PKG_MIRROR_HASH:=687e3743e46c8ddd23f03168b4021ed08b1a858b2a6743db3b62cb3d4c3592a0
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/net-mtools
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Debug multicast setups with mtools (msend and mreceive)
+  URL:=https://github.com/troglobit/mtools
+endef
+
+define Package/net-mtools/description
+  The tools msend and mreceive can be particulary useful
+  when debugging multicast setups.
+
+  msend continuously sends UDP packets to the multicast
+  group specified by the -g and -p options.
+
+  mreceive joins a multicast group specified by the -g and
+  -p options, then receives and displays the multicast
+  packets sent to this group:port combination by the msend
+  command.
+endef
+
+define Package/net-mtools/install
+	$(INSTALL_DIR) $(1)/usr/sbin
+	$(CP) $(PKG_BUILD_DIR)/msend $(1)/usr/sbin/
+	$(CP) $(PKG_BUILD_DIR)/mreceive $(1)/usr/sbin/
+endef
+
+$(eval $(call BuildPackage,net-mtools))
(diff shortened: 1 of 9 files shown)


net/net-mtools/Makefile
```diff
@@ -0,0 +1,44 @@
...
+include $(INCLUDE_DIR)/package.mk
+
+define Package/net-mtools
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=mtools
+  URL:=https://github.com/vladimiroltean/mtools/
+endef
+
+define Package/net-mtools/description
```
BKPepe, line 25: This section has two spaces as indentation

BKPepe:
> net-mtools is used instead of mtools as it does conflicts with another package that is also called mtools.

Hmm. Shouldn't be there ``conflict`` in the Makefile for this package?

Ansuel:
> > net-mtools is used instead of mtools as it does conflicts with another package that is also called mtools.
> 
> Hmm. Shouldn't be there `conflict` in the Makefile for that package?

mhhh no because i'm calling the package with a different name. Using the correct name and using the CONFLICT won't solve the problem that on package index we have 2 package with the same name. (with my testing using the same name resulted in only the current mtools compiled and this never parsed)

net/net-mtools/Makefile
Ansuel: you are referring to Package/net-mtools ?

BKPepe:
I see, [mtools](https://github.com/openwrt/packages/blob/f0ef3925d4247234bc585ea8a3e7d212daaf8b27/utils/mtools/Makefile#L35) does not have these binaries, so that's okay. That's why I was confused in the first sight.

Anyway, if you look into ``test.sh`` files in this repository and add one to your related package, then it will be run tested inside CI/CD and it will be just fine.

Ansuel:
@BKPepe do you have any example for test.sh? Also I'm sure how we can test this as it would require major setup like knowing the container network port and tcpdump to scan if the package is actually sent

BKPepe:
Well, we do have pretty dump run testing check, but at least it is does something. It checks, if the binary prints the version.
Basically, something like this https://github.com/openwrt/packages/commit/3f00d28adf0fb0a154385a9ef96189f4d9e21fc6 will do the job.

However, it is not enabled still by default, thus adding test.sh manually is required (https://github.com/openwrt/packages/issues/13589 and https://github.com/openwrt/packages/pull/13785)

BKPepe:
I see run testing inside the CI looks great!

```
Installing net-mtools (2.3-1) to root...
Configuring net-mtools.
Use package specific test.sh
msend version 2.3
mreceive version 2.3
Test successful
```

Anyway, I do have one question, though. During my 1st round of review, I saw that you were using [forked mtools repository](https://github.com/vladimiroltean/mtools/), which is still left as URL, but you are downloading it from https://github.com/troglobit/mtools. Is there any reason for that? 

Except from this one thing, which could be nitpicking, as I would prefer I think the forked repository (without patches here, fewer patches = less maintainance), it looks very good. Thank you @Ansuel for doing such good work and also for providing feedback during the reviews.

Ansuel:
Eh I forgot to update the other URL.

The project seems to be very old but it does have some release. And to have the CI test to work we need PKG_RELEASE.

So I thought it's better to use the original repo where the release were created and add the additional patch. 

If it's a problem I can totally use the forked repo from Vladimir but it might be confusing and someone may think they are release from him (and the versioning from him) while in reality it's just forking the repo as using the original release tag.

But np, it's really a second for me to revert to the old link! Just tell me what do you prefer!

BKPepe:
Nope, it is not an issue at all. I just asked about it. :)

It's up to you to decide what you want. I can live with both solutions.

BKPepe (approved):

BKPepe:
Once CI/CD finishes its job, I will merge it.

PR #21500, opened 2023-07-05
shonenjump: add new package

graysky2:
This is a lightweight autojump clone written in Go.

From upstream's readme:
Once you have cd into a directory, shonenjump will save it in a list. The next time you can use the j shortcut to visit it.

For example, suppose that you have cd into a directory called /usr/local/Very-Long-Dir-Name/Sub-Dir/target after shonenjump is enabled. You can then use j long or j target or j vldn to visit it.

Sometimes the first matched directory is not what you want, you can type j <your key word> and then type Tab to trigger auto completion and see the options.

Build system: x86_64
Build-tested: x86_64/5800U
Run-tested: x86_64/5800U

### DIFF
--- a/utils/shonenjump/Makefile
+++ b/utils/shonenjump/Makefile
@@ -1,0 +1,51 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=shonenjump
+PKG_VERSION:=0.8.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/suzaku/$(PKG_NAME)/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=ff02e0872c96dbaecdb64e8ad977a98260ec1a105869d8cf4b988ac579dcf789
+
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=John Audia <therealgraysky@proton.me>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_BUILD_FLAGS:=no-mips16
+
+GO_PKG:=github.com/suzaku/shonenjump
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/shonenjump
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Lightweight autojump clone written in Go
+  URL:=https://github.com/suzaku/shonenjump
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
+define Package/shonenjump/description
+Quote from the description of autojump:
+
+autojump is a faster way to navigate your filesystem. It works by maintaining a database of the directories you use the most from the command line.
+
+Directories must be visited first before they can be jumped to.
+endef
+
+define Package/shonenjump/install
+	$(call GoPackage/Package/Install/Bin,$(1))
+	$(INSTALL_DIR) $(1)/usr/share/shonenjump
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/scripts/shonenjump.bash $(1)/usr/share/shonenjump
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/scripts/shonenjump.fish $(1)/usr/share/shonenjump
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/scripts/shonenjump.zsh $(1)/usr/share/shonenjump
+	$(INSTALL_DIR) $(1)/usr/share/zsh/site-functions
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/scripts/_j $(1)/usr/share/zsh/site-functions
+endef
+
+$(eval $(call GoBinPackage,shonenjump))
+$(eval $(call BuildPackage,shonenjump))


graysky2:
Currently, bash, fish, and zsh are supported whereas ash is not.  I do not know how to define this.  The package can be installed without these shells, it just will not work with ash.

utils/shonenjump/Makefile
```diff
@@ -0,0 +1,47 @@
...
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com//suzaku/$(PKG_NAME)/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=ff02e0872c96dbaecdb64e8ad977a98260ec1a105869d8cf4b988ac579dcf789
+
+PKG_LICENSE:=Satoru Logic and contributors
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=
+
```
BKPepe, line 14: We can not add packages without maintainer.

utils/shonenjump/Makefile
```diff
@@ -0,0 +1,47 @@
...
+PKG_NAME:=shonenjump
+PKG_VERSION:=0.8.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com//suzaku/$(PKG_NAME)/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=ff02e0872c96dbaecdb64e8ad977a98260ec1a105869d8cf4b988ac579dcf789
+
+PKG_LICENSE:=Satoru Logic and contributors
```
BKPepe, line 11: That LICENSE looks like BSD-3-Clause

utils/shonenjump/patches/0001-store.go-do-not-call-Age.patch
```diff
@@ -0,0 +1,11 @@
+--- a/jump/store.go
++++ b/jump/store.go
```
BKPepe, line 2: Patch will be upstreamed or it was already send to upstream?

BKPepe (changes requested):

utils/shonenjump/Makefile
graysky2: Fixed

utils/shonenjump/patches/0001-store.go-do-not-call-Age.patch
graysky2: Upstream is thinking about how to handle the code, see: https://github.com/suzaku/shonenjump/issues/10

Haven't seen anything in a while.  If you want, I will remove that patch and build pure upstream.

BKPepe: Okay, from your patch it is not clear for me (as I am not involved either using shonenjump anyhow) that it was related to some issue, which you are dealing with. Could you please improve this patch to be more upstreamable based on wiki - https://openwrt.org/submitting-patches ?

So it has Git header - author, commit subject, description. Otherwise, we would end up with the situation as it is in the main repository where these patches are, but no author, no description and we can just guess what they are doing.

graysky2: Done, let me know if it is acceptable.  Will keep this unresolved until I hear back.

BKPepe: Thanks, much better. 👍

BKPepe:
Since, this is the Go package, let me ask @jefferyto for review!

utils/shonenjump/patches/0001-store.go-do-not-call-Age.patch
graysky2: Great, thanks for you help to review.

graysky2:
> Since, this is the Go package, let me ask @jefferyto for review! 

This is a good idea.  It build fine but I based it on several other go packages I inspected.  I not by no means expert :smile:

utils/shonenjump/Makefile
```diff
@@ -0,0 +1,47 @@
...
+define Package/shonenjump/description
+ A lightweight autojump clone written in Go.
+endef
+
+define Package/shonenjump/install
+	$(call GoPackage/Package/Install/Bin,$(1))
+	$(INSTALL_DIR) $(1)/etc/profile.d
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/scripts/shonenjump.bash $(1)/etc/profile.d
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/scripts/shonenjump.fish $(1)/etc/profile.d
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/scripts/shonenjump.zsh $(1)/etc/profile.d
```
jefferyto, line 41: I don't think you can just copy these scripts into /etc/profile.d and be done. Most if not all shells will load these scripts - at best these pollute the environment, at worst they cause errors due to incompatible syntax (e.g. busybox sh loading the fish/zsh scripts)

I think as a minimum you need to add stub script(s) that load the right shonenjump script for the user's shell. Better would be to split the scripts into separate packages, so if I use bash I just need to install the bash package and not have to install the other scripts.

utils/shonenjump/patches/0001-store.go-do-not-call-Age.patch
```diff
@@ -0,0 +1,24 @@
...
+Disable aging function to more closely approximate autojump behavior.
+
+See: https://github.com/suzaku/shonenjump/issues/10
+
+---
+ jump/store.go | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+--- a/jump/store.go
++++ b/jump/store.go
```
jefferyto, line 15: I don't think this patch should be applied unless it is accepted into upstream. This currently represents your preference; other users of shonenjump may not share this preference and there is no way for them to reverse this change.

utils/shonenjump/Makefile
```diff
@@ -0,0 +1,47 @@
...
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/suzaku/$(PKG_NAME)/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=ff02e0872c96dbaecdb64e8ad977a98260ec1a105869d8cf4b988ac579dcf789
+
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=John Audia <therealgraysky@proton.me>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
```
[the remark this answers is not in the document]
graysky2: Done

utils/shonenjump/Makefile
```diff
@@ -0,0 +1,47 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/shonenjump
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Shonenjump
+  URL:=https://github.com/suzaku/shonenjump
+  DEPENDS:=$(GO_ARCH_DEPENDS) +libstdcpp
```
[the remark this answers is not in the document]
graysky2: It depends on glibc, is +libstdcpp not right for that?

utils/shonenjump/Makefile
graysky2: I like these suggestions (split package and have a stub script automatically source these on a per-shell basis).  I have a few questions:
1. If not `/etc/profile.d/` where should these script go?
2. Can you suggest a stub script that would load these automatically?  I am not familiar enough with how OpenWrt would call these.

utils/shonenjump/Makefile
```diff
@@ -0,0 +1,47 @@
...
+
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=John Audia <therealgraysky@proton.me>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+
+GO_PKG:=github.com/suzaku/shonenjump
+GO_PKG_BUILD_PKG:=$(GO_PKG)
```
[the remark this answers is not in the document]
graysky2: Removed

utils/shonenjump/Makefile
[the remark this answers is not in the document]
jefferyto: libstdcpp is standard library for C++.

Every package has an implicit dependency on libc, there is no need to add `+libc` to `DEPENDS`.

utils/shonenjump/Makefile
jefferyto: > If not /etc/profile.d/ where should these script go? (/usr/share/shonenjump/)?

/usr/share/shonenjump would be fine.

> Can you suggest a stub script that would load these automatically?

Test for a specific shell, if it is found then source the corresponding shonenjump script - something like [this](https://github.com/openwrt/openwrt/blob/afa229038c05ba0ca20595d7f73bea94db21d3a6/package/base-files/files/etc/shinit#L6).

Ensure the stub script(s) only use POSIX shell features so they can be safely loaded with any POSIX-compatible shell.

> How can we make selecting the base file automatically select the one for bash but allow the user to deselect that one if another is selected?

You don't - users should install the shell-specific package (which pulls in the base package), not the base package directly. The base package description can include instructions for users to do this.

utils/shonenjump/Makefile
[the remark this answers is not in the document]
graysky2: Done, removed.

utils/shonenjump/Makefile
graysky2: > Test for a specific shell, if it is found then source the corresponding shonenjump script - something like [this](https://github.com/openwrt/openwrt/blob/afa229038c05ba0ca20595d7f73bea94db21d3a6/package/base-files/files/etc/shinit#L6).

I will look at the example you provided.

> users should install the shell-specific package (which pulls in the base package), not the base package directly. The base package description can include instructions for users to do this.

Can you suggest the proper syntax for a base package and for the three shell-specific packages?  I am thinking this split might be more trouble than it's worth; savings is ~5 K:
```
-rw-r--r-- 1 root root 2.9K Jan 14  2023 shonenjump.bash
-rw-r--r-- 1 root root 2.5K Jan 14  2023 shonenjump.fish
-rw-r--r-- 1 root root 2.1K Jan 14  2023 shonenjump.zsh
```
What do you think?

jefferyto: > Where would such a file go such that upon login it is called like /etc/shinit gets called?
> What if the user is not using that particular shell? Just searching for the executable seems like a bad idea to check.
> Can you suggest the proper syntax for a base package and for the three shell-specific packages?

You should do some work and find the answers for yourself.

> I am thinking this split might be more trouble than it's worth; savings is ~5 K:

Having shell-specific packages is optional, loading the right script for the right shell is the key point.

graysky2: > You should do some work and find the answers for yourself.

My thought was to use `/etc.profile.d/` but that will cause the default shell to source the needed files whereas we need each of the respective shells (bash, fish, or zsh) to source.  I am not sure how to do this without studying each of the respective shell's packages.

jefferyto: Actually I was thinking about this today

> Test for a specific shell, if it is found then source the corresponding shonenjump script - something like [this](https://github.com/openwrt/openwrt/blob/afa229038c05ba0ca20595d7f73bea94db21d3a6/package/base-files/files/etc/shinit#L6).
> 
> Ensure the stub script(s) only use POSIX shell features so they can be safely loaded with any POSIX-compatible shell.

This is wrong.

> that will cause the default shell to source the needed files whereas we need each of the respective shells (bash, fish, or zsh) to source.

You are correct.

> I am not sure how to do this without studying each of the respective shell's packages.

Not sure how much you will get from studying each shell's packages, but you will at least need to read the man pages for each shell to find out which files they load on startup (for all interactive shells, not just login shells).

graysky2:
In reviewing the man pages for zsh, it seems that placing the source line in `/etc/zsh/zshrc` would get the job done.  It works in this fashion on Arch Linux but our zsh package seems to ignore that file entirely.  I have not looked into fish or bash.  Since upstream's [docs](https://github.com/suzaku/shonenjump#other-distros) calls out a manual modification to `~/.shellprofile`, I think shipping this without trying to modify things is the right thing to do.

utils/shonenjump/Makefile
graysky2: In reviewing the man pages for zsh, it seems that placing the source line in /etc/zsh/zshrc would get the job done. It works in this fashion on Arch Linux but our zsh package seems to ignore that file entirely. I have not looked into fish or bash. Since upstream's [docs](https://github.com/suzaku/shonenjump#other-distros) calls out a manual modification to ~/.shellprofile, I think shipping this without trying to modify things is the right thing to do.

jefferyto: Whether you want to do the work is up to you.

If I am installing software from source, then I would expect that I might have to do some integration work, e.g. write an init.d script/systemd unit file, edit my .bashrc file, etc. If I am installing software from a distro package, then I would expect most of that integration work to already be done.

graysky2: I understand your expectation.  If OpenWrt has this philosophy for packaging, it makes sense to continue digging for the replace place for a script to place a file.  I am out of ideas that don't involve directly editing another package's file.

Issue #21300, opened 2023-06-07
nginx-full (nginx-all-module): installation issue on snapshot


iganeshk:
Maintainer: @hgl @Ansuel 
Environment:
  ```bash
  root@hostname:~# cat /etc/openwrt_release
  DISTRIB_ID='OpenWrt'
  DISTRIB_RELEASE='SNAPSHOT'
  DISTRIB_REVISION='r23293-98e6ea32a4'
  DISTRIB_TARGET='mvebu/cortexa9'
  DISTRIB_ARCH='arm_cortex-a9_vfpv3-d16'
  DISTRIB_DESCRIPTION='OpenWrt SNAPSHOT r23293-98e6ea32a4'
  DISTRIB_TAINTS=''
  ```
## Description: Unable to install nginx-full (nginx-all-module) on a fresh snapshot sysupgrade

Plausible issue:
```makefile 
define Package/nginx-ssl
  $(Package/nginx/default)
  TITLE += with SSL support
  VARIANT:=ssl
  DEPENDS+= +NGINX_PCRE:libpcre \
	+NGINX_PCRE:nginx-ssl-util +!NGINX_PCRE:nginx-ssl-util-nopcre \
	+NGINX_HTTP_GZIP:zlib +NGINX_DAV:libxml2
  EXTRA_DEPENDS:=nginx-ssl-util$(if $(CONFIG_NGINX_PCRE),,-nopcre) (>=1.5-1) (<2)
  CONFLICTS:=nginx-full
endef

define Package/nginx-full
  $(Package/nginx/default)
  TITLE += with ALL module selected
  DEPENDS+=+libpcre +nginx-ssl-util +zlib +libxml2 \
    $(foreach m,$(PKG_MOD_ALL),+nginx-mod-$(m))
  EXTRA_DEPENDS:=nginx-ssl-util (>=1.5-1) (<2)
  VARIANT:=full
  PROVIDES += nginx-ssl
endef
```

PS: Shouldn't there be a placeholder package `nginx-all-module` which points to `nginx-full` when a package is deprecated?

Log:
  ```bash
  root@hostname:~# opkg list-installed
  addrwatch - 1.0.2-3
  addrwatch-mysql - 1.0.2-3
  avahi-dbus-daemon - 0.8-7
  avahi-utils - 0.8-7
  base-files - 1532-r23293-98e6ea32a4
  busybox - 1.36.1-1
  ca-bundle - 20230311-1
[...]

PR #21502, opened 2023-07-05
nginx: don't install all module for FULL variant

Ansuel:
We currently have a more or less circular dependency with nginx ssl and
full variant.

FULL variant depends on every nginx module. Every nginx module depends
on nginx-ssl.

Since nginx-full depends on an nginx module, nginx-ssl is installed as
module depends on it and then the installation fails as nginx-full
conflicts with nginx-ssl.

nginx-full in it's meaning is nginx built with every config selected and
it should not have module as dependency. In fact an user should always
install them separetly as while other things, local modification to the
nginx config file are required to include the just installed module.

To fix this circular dependency problem, drop the dependency on every
nginx module for FULL variant.

Fixes: https://github.com/openwrt/packages/issues/21300

### DIFF
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nginx
 PKG_VERSION:=1.25.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=nginx-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://nginx.org/download/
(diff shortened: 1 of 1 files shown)


Ansuel:
@hgl funny bug we triggered!

SopaDeMacaco-UmaDelicia:
I think I experienced the same bug in 23.05-rc2, after trying to install nginx-full (which failed) I couldn't install nginx-ssl, opkg told me that it conflicted with nginx-full. Trying to "delete" nginx-full just told me there's no such package installed.

Ansuel:
@SopaDeMacaco-UmaDelicia i hope it's not a opkg bug and when this problem is solved, and opkg update should fix the situation

hgl:
> @hgl funny bug we triggered!

Yeah, and it originally only happened in CI, not sure how it started to surface.

PR #21521, opened 2023-07-09
luaposix: update to 36.2.1

mstorchak:
Maintainer: me
Compile tested: ipq807x, r23300+3-86bc525d00
Run tested: ipq807x, r23300+3-86bc525d00, `posix = require "posix"` works, `posix.version` reports 36.2.1, `posix.{stat,open,write,close}` work

Description:
- update to 36.2.1
- remove the patch that has been already applied to the release branch


BKPepe:
I see that the patch was not refreshed, but removed. Was it on purpose?

mstorchak:
Yes, the release tgz has it applied. So yeah, refreshed by removal :)

BKPepe:
Thats could be quite confusing. Because usually if they are offsets or
fuzzes, then the patch is refreshed, but in this case, I would rather see:
* Patch removed as it was applied by upstream

or something like that

Dne po 10. 7. 2023 8:37 uživatel mstorchak ***@***.***>
napsal:

mstorchak:
@BKPepe , I see. I'll update the description, thanks for the notice!

PR #21529, opened 2023-07-10
transmission: fix depends on libmbedtls

jjm2473:
If a firmware build with curl without mbedtls, transmission installed from openwrt official repo will fail to start

Maintainer: Daniel Golle <daniel@makrotopia.org>

Description:
For example, a firmware build with curl with libopenssl, and firmware build without libmbedtls, install transmission from openwrt official repo, then transmission will fail to start.


1715173329:
cc @dangowrt

PR #21533, opened 2023-07-10
cpupower: add new package

graysky2:
Linux kernel userspace tool to examine and to tune power saving related features of the processor


graysky2:
I might not have the DEPENDS setup correctly.  This should only build for x86 or x86/64

1715173329:
iirc sdk doesn't contain kernel source

jefferyto:
Need to get the source directory included in [`USERSPACE_UTILS_FILES`](https://github.com/openwrt/openwrt/blob/cd82fc2235d4a11fc33e5b590d8f3bc6a7b13b78/target/sdk/Makefile#L87-L91) before it can be built by sdk.

graysky2:
@jefferyto - Can you suggest a modification to the Makefile to allow it to grab the source?  If the source is present in the build_dir, this Makefile works to build and to package.

jefferyto:
> Need to get the source directory included in [`USERSPACE_UTILS_FILES`](https://github.com/openwrt/openwrt/blob/cd82fc2235d4a11fc33e5b590d8f3bc6a7b13b78/target/sdk/Makefile#L87-L91) before it can be built by sdk.

This already contains everything you need to do.

Here is another hint: https://openwrt.org/submitting-patches

utils/cpupower/Makefile
```diff
@@ -0,0 +1,58 @@
...
+PKG_RELEASE:=1
+PKG_LICENSE:=GPL-2.0-only
+
+# Since kernel 2.6.39.1 userspace tools are inside the kernel tree
+# Package Automatic match version in kernel
+# HASH is not useful kernel package already check it
+PKG_VERSION:=$(shell basename $(LINUX_DIR) | sed 's/linux-//')
+PKG_SOURCE:=
+PKG_SOURCE_URL:=
+PKG_HASH:=unknown
```
BKPepe, line 14: Why this is empty?

utils/cpupower/Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+define Build/Compile/cpupower
+	$(MAKE) CC="$(TARGET_CC)" \
+		CFLAGS="$(TARGET_CFLAGS)" \
+		LDFLAGS="$(TARGET_LDFLAGS)" DESTDIR=fake install-tools
+endef
+
+define Build/Compile
+	(cd $(PKG_BUILD_DIR); $(call Build/Compile/cpupower))
+endef
```
BKPepe, line 49: This looks like default to me. Isn't it?

utils/cpupower/Makefile
```diff
@@ -0,0 +1,58 @@
...
+define Package/cpupower
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Shows and sets processor power related values
+  DEPENDS:=+libpci
+  URL:=https://www.kernel.org
+endef
+
+define Package/cpupower/description
+Linux kernel tool to examine and to tune power saving related features of the processor
```
BKPepe, line 38: Two spaces as indentation here.

utils/cpupower/Makefile
```diff
@@ -0,0 +1,58 @@
+include $(TOPDIR)/rules.mk
+include $(INCLUDE_DIR)/kernel.mk
+
+PKG_NAME:=cpupower
+PKG_RELEASE:=1
+PKG_LICENSE:=GPL-2.0-only
+
+# Since kernel 2.6.39.1 userspace tools are inside the kernel tree
+# Package Automatic match version in kernel
+# HASH is not useful kernel package already check it
```
BKPepe, line 10: We don't need to have it in the Makefile, I suppose. Documentation is elsewhere. :)

utils/cpupower/Makefile
graysky2: It's the linux kernel source... I modeled this after https://github.com/openwrt/packages/blob/master/net/usbip/Makefile#L15

Is that not OK?

utils/cpupower/Makefile
graysky2: I modeled this after https://github.com/openwrt/packages/blob/master/net/usbip/Makefile#L15 that contains the comment.  I deleted it.

PR #21536, opened 2023-07-11
treewide: make all GNOME URLs consistent

neheb:
Maintainer: @flyn-org


flyn-org:
This is fine, as long as we apply it consistently over the OpenWrt tree. I do sometimes forget to update that part of the path, so this will sometimes save a few failed local builds when I do my own updates. Recommend merge, assuming there is consensus across the tree.

PR #21540, opened 2023-07-11
nginx-util: fix compilation with GCC13

neheb:
Maintainer: @peter-stadler


peter-stadler:
Thanks :-)

neheb:
return std::move is almost always wrong.

PR #21543, opened 2023-07-11
shairport-sync: fix init script

jjm2473:
'name' may contains '%h' or '%v', printf will fail on that

Maintainer: Ted Hess <thess@kitschensync.net>, Mike Brady <mikebrady@eircom.net>

Description:


BKPepe:
Squash your commits to single one, which has ``SoB``.

PR #21544, opened 2023-07-11
rust: compile host package per target

### LINKED ISSUE #21530: rust: can't find crate for `core` (when using multiple targets using ./scripts/env)
Maintainer: @1715173329 @lu-zero
Environment: Raspberry Pi 4 (`bcm27xx/bcm2709`) + Raspberry Pi Zero W (`bcm27xx/bcm2708`) 

### Summary

I'm creating custom OpenWrt based images for several Raspberry Pi models. To manage the configuration I'm using `./scripts/env` to switch between configuration. The first target (`bcm27xx/bcm2709`) compiled fine. After switching my configuration to the `bcm27xx/bcm2708` target, I got the following error:

```
error[E0463]: can't find crate for `core`
  |
  = note: the `arm-unknown-linux-musleabihf` target may not be installed
  = help: consider downloading the target with `rustup target add arm-unknown-linux-musleabihf`
```

I see that `rustc` is invoked with `--target arm-unknown-linux-musleabihf`.

Looking at the `cargo/config`, I find the following:

```
❯ cat ./staging_dir/hostpkg/cargo/config
[target.armv7-unknown-linux-musleabihf]
linker = "arm-openwrt-linux-muslgnueabi-gcc"
rustflags = ["-Ctarget-feature=-crt-static", "-Clink-args=-lssp_nonshared"]

[profile.stripped]
inherits = "release"
opt-level = "s"
strip = true
```

### Assumption

I think this issue is related to the fact that I'm using `./scripts/env` for working with multiple environments. Compiling my Rust source-code for the first target works fine because this will compile the `rust/host` package. My assumption is that if `rust/host` compiles, that this will include the toolchain for the target that is present in the OpenWrt configuration (in my case `bcm27xx/bcm2709`).

Once I switch my configuration to `bcm27xx/bcm2708`, then this does not invalidate the `rust/host` package and thus the `--target` flag points to a target which is not present in the already compiled `rust/host`, causing the above error.

### Possible solution

Currently the `cargo` binary is installed in:

```
./staging_dir/hostpkg/cargo/bin/cargo
```

Wouldn't it be better to prefix this with the OpenWrt target, like for example the `gcc` binaries:

```
./staging_dir/toolchain-arm_cortex-a7+neon-vfpv4_gcc-12.3.0_musl_eabi/bin/arm-openwrt-linux-muslgnueabi-gcc
./staging_dir/toolchain-arm_arm1176jzf-s+vfp_gcc-12.3.0_musl_eabi/bin/arm-openwrt-linux-muslgnueabi-gcc
```

brocaar:
Maintainer: @lu-zero 
Compile tested: Raspberry Pi bcm27xx/bcm2709, bcm27xx/bcm2708. OpenWrt `openwrt-23.05` branch
Run tested: Raspberry Pi bcm27xx/bcm2709, bcm27xx/bcm2708. OpenWrt `openwrt-23.05` branch

Description:

While the compiled binaries are intended to run on the host system, the rust/host package does include the target matching the configured OpenWrt target.

If using (for example) ./scripts/env to switch between different OpenWrt configurations, this will cause issues if the different configuration is for a different target. In such case there will be a mismatch between the available Rust target and OpenWrt target and the following error will be printed:

> error[E0463]: can't find crate for `core`
> note: the `XXX` target may not be installed

This fix will add the RUSTC_TARGET_ARCH as HOST_BUILD_DIR and CARGO_HOME suffix, such that rust/host will be compiled in case an OpenWrt configuration change causes the RUSTC_TARGET_ARCH to change.

Fixes #21530.

### DIFF
--- a/lang/rust/Makefile
+++ b/lang/rust/Makefile
@@ -3,6 +3,7 @@
 # Copyright (C) 2023 Luca Barbato and Donald Hoskins
 
 include $(TOPDIR)/rules.mk
+include ./rust-values.mk
 
 PKG_NAME:=rust
 PKG_VERSION:=1.70.0
@@ -11,7 +12,7 @@ PKG_RELEASE:=2
 PKG_SOURCE:=rustc-$(PKG_VERSION)-src.tar.gz
 PKG_SOURCE_URL:=https://static.rust-lang.org/dist/
 PKG_HASH:=b2bfae000b7a5040e4ec4bbc50a09f21548190cb7570b0ed77358368413bd27c
-HOST_BUILD_DIR:=$(BUILD_DIR_HOST)/rustc-$(PKG_VERSION)-src
+HOST_BUILD_DIR:=$(BUILD_DIR_HOST)/rustc-$(PKG_VERSION)-src-$(RUSTC_TARGET_ARCH)
 
 PKG_MAINTAINER:=Luca Barbato <lu_zero@luminem.org>
 PKG_LICENSE:=Apache-2.0 MIT
@@ -22,7 +23,9 @@ PKG_HOST_ONLY:=1
 
 include $(INCLUDE_DIR)/host-build.mk
 include $(INCLUDE_DIR)/package.mk
-include ./rust-values.mk
+
+TAR_OPTIONS+= --strip-components 1
+TAR_CMD=$(HOST_TAR) -C $(1) $(TAR_OPTIONS)
 
 define Package/rust
   SECTION:=lang
--- a/lang/rust/rust-values.mk
+++ b/lang/rust/rust-values.mk
@@ -5,7 +5,6 @@
 # Rust Environmental Vars
 CONFIG_HOST_SUFFIX:=$(word 4, $(subst -, ,$(GNU_HOST_NAME)))
 RUSTC_HOST_ARCH:=$(HOST_ARCH)-unknown-linux-$(CONFIG_HOST_SUFFIX)
-CARGO_HOME:=$(STAGING_DIR_HOSTPKG)/cargo
 CARGO_VARS:=
 
 ifeq ($(CONFIG_USE_MUSL),y)
@@ -51,3 +50,5 @@ endif
 ifeq ($(ARCH),aarch64)
     RUST_CFLAGS:=-mno-outline-atomics
 endif
+
+CARGO_HOME:=$(STAGING_DIR_HOSTPKG)/cargo-$(RUSTC_TARGET_ARCH)


lang/rust/Makefile
```diff
@@ -22,7 +23,6 @@ PKG_HOST_ONLY:=1
 
 include $(INCLUDE_DIR)/host-build.mk
 include $(INCLUDE_DIR)/package.mk
-include ./rust-values.mk
```
lu-zero, line 25: Why that?

lu-zero:
This is not the cleanest way but shouldn't impact at all the normal usage.

If you want to fully support your use-case please split the rust package between host and target ideally the amount of overhead would be further reduced.

lang/rust/Makefile
brocaar: It was moved up, because:

> `HOST_BUILD_DIR:=$(BUILD_DIR_HOST)/rustc-$(PKG_VERSION)-src-$(RUSTC_TARGET_ARCH)`

needs the `RUSTC_TARGET_ARCH` variable, which is provided by `./rust-values.mk`

lu-zero:
@1715173329 I defer to your expertise regarding if it is acceptable or not in the bigger picture. I'm not against it if you are fine with it.

lang/rust/Makefile
```diff
@@ -73,6 +73,12 @@ HOST_CONFIGURE_ARGS = \
 	--set=llvm.download-ci-llvm=true \
 	$(TARGET_CONFIGURE_ARGS)
 
+define Host/Prepare
```
1715173329, line 76: please use tar options instead:
https://github.com/openwrt/packages/blob/7709b85f86f7ce2da5a86f2d5d1fcdd77c5089a1/net/librespeed-go/Makefile#L28-L29

1715173329:
> @1715173329 I defer to your expertise regarding if it is acceptable or not in the bigger picture. I'm not against it if you are fine with it.

this might be the simplest way to go.
ideally this should go to toolchain (with llvm) but i guess rust is not that "important" for whole openwrt and cannot move into base currently...

lang/rust/Makefile
brocaar: I pushed a fix for this yesterday.

brocaar:
@1715173329 I have updated the commit message as I noticed the tests were failing because of missing package prefix.

1715173329:
> @1715173329 I have updated the commit message as I noticed the tests were failing because of missing package prefix.

Please squash your commits, just keep the first one

@lu-zero this is your package so it's up to you ;) I'll merge this pr if you are okay with it

> i dont see any feed package does it before, but can we install rust into `$(TOOLCHAIN_DIR)` directly or any better idea?

Also cc @BKPepe @jefferyto

lu-zero:
For the specific usecase it is needlessly builds many times the host tools since they stay exactly the same, but given that overall the normal usage is not impacted I'm fine with that if it solves @brocaar problems already.

brocaar:
> Please squash your commits, just keep the first one

@1715173329  done :heavy_check_mark:

lang/rust/Makefile
```diff
@@ -11,7 +12,7 @@ PKG_RELEASE:=2
 PKG_SOURCE:=rustc-$(PKG_VERSION)-src.tar.gz
 PKG_SOURCE_URL:=https://static.rust-lang.org/dist/
 PKG_HASH:=b2bfae000b7a5040e4ec4bbc50a09f21548190cb7570b0ed77358368413bd27c
-HOST_BUILD_DIR:=$(BUILD_DIR_HOST)/rustc-$(PKG_VERSION)-src
+HOST_BUILD_DIR:=$(BUILD_DIR_HOST)/rustc-$(PKG_VERSION)-src-$(RUSTC_TARGET_ARCH)
```
jefferyto, line 15:
```suggestion
HOST_BUILD_DIR:=$(BUILD_DIR_HOST)/rust-$(RUSTC_TARGET_ARCH)/rustc-$(PKG_VERSION)-src
```

This way you won't need to need to override `TAR_OPTIONS`/`TAR_CMD`.

jefferyto:
> can we install rust into $(TOOLCHAIN_DIR) directly or any better idea?

I think `$(STAGING_DIR)/host` may be a better place.

1715173329:
> > can we install rust into $(TOOLCHAIN_DIR) directly or any better idea?
> 
> I think `$(STAGING_DIR)/host` may be a better place.

Indeed

1715173329:
Merged via #21571, thank you!

PR #21550, opened 2023-07-12
php8: update to 8.2.8

mhei:
Maintainer: me
Compile tested: mxs, riscv64


1715173329:
Upstream has merged the [atomic workaround for riscv64](https://github.com/php/php-src/commit/1dfa277a966cff1988e720146de21d1bcc09e0f5) into 8.2.8+ stable branch, maybe we need to add libatomic dependency for riscv64 again, or create a patch to revert this change.

PR #21552, opened 2023-07-13
python3: Restore platform triplet to paths

jefferyto:
Maintainer: me
Compile tested: most/all architectures, 2023-07-09 snapshot sdks
Run tested: armsr-armv7/armsr-armv8/malta-be/x86-generic/x86-64 (qemu), 2023-07-09 snapshot

Description:
The main purpose of these changes are to restore the platform triplet to paths. The direct benefit of this would be to have setuptools-rust build successfully (and so enable building of Python packages with Rust extensions), as PyO3 expects to find the platform triplet in the _sysconfigdata file name. This should also enable maturin-based packages (see #20410) though I have not verified this / know enough about maturin to say.

I have not been able to reproduce the issues I found in #12754 regarding the platform triplet and host/target module loading; I'm not sure if this is result of changes in Python itself (there has been work done recently to improve cross-compilation), changes in our Python package build process (i.e. moving away from pip), or something else.

I'm not sure if these changes affect installing host Python packages using the previous/deprecated method (with host pip requirements files); I plan on submitting changes to downstream projects to move them toward the current process (including adding host packages here if necessary).

Please see the individual commit messages for details.


lang/python/python-package-install.sh
```diff
@@ -1,45 +1,138 @@
...
+	error "Missing source directory"
 	exit 1
-}
+fi
+if [ -z "$dest" ]; then
+	error "Missing destination directory"
+	exit 1
+fi
+
+while IFS='|' read cmd path file_mode dir_mode; do
```
commodo, line 47: phew...
quite the rewrite :)

commodo:
so, the platform triplet patches worry me a bit;

they seem like we're starting our own Python fork downstream;
i would NAK them, but ultimately you are the official Python maintainer

i would suggest pushing these patches upstream, and then let them trickle down back

if you go AWOL again for too long, and someone else needs to pick Python up again (whether it's me or someone else), the patches will be ¯\\\_(ツ)\_/¯
if that happens, i would not mind dropping them (during an upgrade of Python) and keeping the  "old" (current) way

commodo (changes requested):
I would prefer to upstream the new patches

commodo:
the rest of the changes are fine from my side

jefferyto:
I would prefer to upstream the patches as well but https://github.com/python/cpython/pull/96001 has been stuck in draft status for months. I'll try pinging them again.

Without the patches, the triplet will have `-gnu` instead of `-musl`; this is because configure is run with arguments like `--target=arm-openwrt-linux --host=arm-openwrt-linux --build=x86_64-pc-linux-gnu` (set by default by OpenWrt buildroot). So some patch or change would still necessary on our side.

I also intentionally split my changes into separate patches to make them as understandable as possible. If someone else wants to remove them without a good explanation or demonstrating they understand what they are doing, then I would question the value of their contribution.

jefferyto:
BTW the various `-musl` values (`muslsf`, `musln32`, etc.) come from https://github.com/richfelker/musl-cross-make#supported-targets. That project (musl-cross-make) is linked to from the musl home page: https://musl.libc.org/

I will update commit message for "python3: Fix/expand platform triplet detection" with a reference to that list of targets.

jefferyto:
Changes to filespec script:
* Fix typos in `file_dir_chmod`
* Add `-r` option to `read`
* Use here document (`<<`) instead of here string (`<<<`), removing the only Bashism in the script
* Change shebang to `/bin/sh`

jefferyto:
Updates:
* Changed "python3: Fix/expand platform triplet detection" to backport https://github.com/python/cpython/commit/c163d7f0b67a568e9b64eeb9c1cbbaa127818596
* Rebased on master

jefferyto:
If there are no more comments, I will merge this in a week :pray:

PR #21571, opened 2023-07-15
rust: Update to 1.71.0

1715173329:
Maintainer: @lu-zero 
Compile tested: rockchip/armv8
Run tested: NanoPi R2S

Description:
- Merged PR #21544 with @jefferyto's suggestion
- rust: Update to 1.71.0

Cc: @brocaar

PR #21574, opened 2023-07-15
rust: update to 1.71.0

lu-zero:
Maintainer: me 
Compile tested: aarch64
Run tested: N/A

Description: Usual version bump


1715173329:
Please refresh patches with `make package/rust/host/refresh V=s`
And the libc version in rust 1.71.0 is 0.2.143, while the minimum version that contains fixes for musl 1.2.4 is 0.2.145, so I think 0002 patch is still required.

btw I did this in #21571 too ;)

lu-zero:
Ops I didn't notice it. :)

PR #21580, opened 2023-07-16
python-pytz: Add host build

jefferyto:
Maintainer: @commodo 
Compile tested: armsr-armv7, 2023-07-09 snapshot sdk
Run tested: N/A

Description:
pytz is a dependency of Babel, so to add a host build of Babel a host build of pytz is necessary.


jefferyto:
With newer versions of Babel, pytz is not a requirement when using Python 3.9 or higher - closing.

PR #21593, opened 2023-07-17
CI: Fix finding test script

jefferyto:
Maintainer: @aparcar
Compile tested: N/A
Run tested: N/A

Description:
Currently, the run-test code tries to find the package source directory based on the directory name only. This fails for the Go compiler package because there is more than one directory named "golang".

This uses the full path listed in the "Source:" line of the control file to find the package source directory.

This also:

* Checks for the test script earlier, to avoid installing and removing ipk files when there is no test script to be run

* Makes `PKG_VERSION` parsing more lenient, as the package may not have a `PKG_RELEASE`, e.g. attendedsysupgrade-common


aparcar (approved):
Looks good to me, thanks for the work!

PR #21598, opened 2023-07-18
cpusage: remove AUTORELEASE

kuanyili:
Maintainer: @thuehn

Description:

Remove deprecated AUTORELEASE.

Commit 0c10c224be81 only handles the case where AUTORELEASE is used in PKG_RELEASE thus this package was left behind. Let's fix this up.

This packages is untouched since 23.05 branching. It should be straightforward applying this patch also to that branch.


thuehn:
ACK from my side
Grez Thomad

PR #21622, opened 2023-07-21
tailscale: Update to 1.48.0

mochaaP:
Maintainer: @ja-pa @oskarirauta
Compile tested: `aarch64_cortex-a53` @ SNAPSHOT
Run tested: `aarch64_cortex-a53` @ SNAPSHOT

This update brings nftables support to Tailscale.

Signed-off-by: Zephyr Lykos <git@mochaa.ws>


BKPepe:
This really needs to be run tested as there are huge changes.

mochaaP:
Agree that, but I temporarily have no access to my testing device. Feedbacks welcomed!

mochaaP:
race cond in CI, please rerun

BKPepe:
Still, to be merged, it requires to be run tested. I will not merge this unless there is confirmation or some indication of run testing. The changes are huge.

cmtsij:
@mochaaP 
Do we have to set TS_DEBUG_USE_NETLINK_NFTABLES=true to enable nftables ?
ref: https://forum.openwrt.org/t/tailscale-with-nftables-support/166872

mochaaP:
Run tested: `aarch64_cortex-a53` @ SNAPSHOT

Everything works including routing with nft.

Users who set up the rules with fw4 can try running with `--netfilter-mode=on` and check the rules with `nft list ruleset`.

mochaaP:
cc @BKPepe

net/tailscale/patches/010-remove_iptables.patch
```diff
@@ -0,0 +1,1967 @@
+--- a/cmd/derper/depaware.txt
```
BKPepe, line 1: Ok, so... this patch... 

- Who is the author?
   - Who did those changes and when we can find them?
- Where is the commit subject and commit description for this patch?
- Why do we need this patch at all? Why it is not included in tailscale repository?
- Will you send it to upstream?

net/tailscale/Makefile
```diff
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tailscale
-PKG_VERSION:=1.44.0
+PKG_VERSION:=1.46.1
 PKG_RELEASE:=1
 
 PKG_SOURCE:=tailscale-$(PKG_VERSION).tar.gz
```
BKPepe, line 14: PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz

net/tailscale/patches/010-remove_iptables.patch
mochaaP: > Who is the author?

Me.

> Where is the commit subject and commit description for this patch?

It's in the main commit message: `the go-iptables support is no longer compiled in.`

> Why do we need this patch at all? Why it is not included in tailscale repository?

I removed iptables related code and ran `go mod tidy` on the dependencies. This eliminates the go-iptables dependency to reducing binary size and the need to add a flag for this.

> Will you send it to upstream?

No, as upstream will not accept those changes, and we only need this patch in the branches without iptables support.

net/tailscale/Makefile
mochaaP: resolved in <https://github.com/openwrt/packages/commit/aeedbf9c101e3c2fe653ab9f19a0855514b7bdcb>

net/tailscale/patches/010-remove_iptables.patch
BKPepe: Sorry, but no. This patch is so huge, then we can not accept it here. Nobody will look into it and refreshing or editing this patch will be pain.  You should reach upstream to check what needs to be done here.

Even it is in the commit message, it is not in the patch itself.

There is even iptables support in OpenWrt master branch. Not enabled by default, but still there is.

1715173329:
Ping @oskarirauta
Please take a look, thanks

net/tailscale/patches/010-remove_iptables.patch
mochaaP: But I don't see a point where we should maintain the old patch for go-iptables while we have support for nftables. The old fake_iptables patch is only required to workaround detections in go-iptables as it would check for the existence of the `iptables` binary on system.

The huge patch size is only because deleting files have to count every line of code and updating the indirect dependencies required by both go-iptables and other packages.

I could add a script to automate refreshing the patch if you wish me to.

:heart:

BKPepe: It was not removed upstream, why should we do that if others GNU/LInux distros did not do it as well? Nftables was added only 3 weeks ago and now it seems like you are eager to see iptables support drop, then discuss that it upstream, please.

mochaaP: I should outline the problem here:

**The old fake_iptables patch is only required to workaround detections in go-iptables as it would check for the existence of the iptables binary on system and refuse to run if none is provided.**

iptables support is already broken a while ago, as it does nothing but executes `/bin/false`. The "support" is just to make the program run and the routing functionality is not provided anyways. For me it doesn't make much sense to continue supporting the fake iptables patch, if the actual feature is working via nftables.

mochaaP:
We finally don't need any patches on our side. :smiley:

cc @1715173329 @BKPepe

mochaaP:
Squashed into a single commit as per Shen suggests.

1715173329 (approved):
LGTM - Thanks!

PR #21624, opened 2023-07-21
rust: put host build dir to target build dir

1715173329:
Maintainer: @lu-zero 
Run tested: n/a

Description:
```
When user runs `make clean` command, everything in `$(STAGING_DIR)`
(where we installed rust) will be removed, but `$(BUILD_DIR_HOST)`
(where we compiled rust and stored build stage) is untouched.

So when user starts a new build after that, OpenWrt buildroot will
still consider `rust` is installed already, resulting the build error
"cargo: command not found".

Fix this by moving to target build dir as well.

```

cc @brocaar


lu-zero:
I do not have knowledge to suggest a different solution, go for it :)

PR #21627, opened 2023-07-22
gperftools: add new package

graysky2:
Thread-caching malloc provided by this package improves snort3 performance.  I have been running with this for over five months now without issues.  Avg CPU usage is down.  Another user reported higher throughput achieved with snort3 compiled with this on samba transfers on system with CPU-limited snort performance.[1]

1. https://forum.openwrt.org/t/some-help-with-a-makefile-gperftools/165656/22

Build system: x86/64
Build-tested: x86/64
Run-tested: x86/64

Maintainer: 

@flyn-org - just a heads up that this and hyperscan (see https://github.com/openwrt/packages/pull/21471) improve snort3 performance.  I have a draft Makefile for snort3 that incorporates both.  Sneak peak: https://github.com/graysky2/packages/commit/a017d4bf1fe7d2ccf7c1664e0d064d25c78f4c1c

### DIFF
--- a/libs/gperftools/Makefile
+++ b/libs/gperftools/Makefile
@@ -1,0 +1,72 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gperftools
+PKG_VERSION:=2.15
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/gperftools/gperftools/tar.gz/$(PKG_NAME)-$(PKG_VERSION)?
+PKG_HASH:=3918ff2e21bb3dbb5a801e1daf55fb20421906f7c42fbb482bede7bdc15dfd2e
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_NAME)-$(PKG_VERSION)
+
+PKG_MAINTAINER:=John Audia <therealgraysky@proton.me>
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=COPYING
+
+PKG_BUILD_FLAGS:=no-mips16
+PKG_INSTALL:=1
+PKG_BUILD_PARALLEL:=1
+
+PKG_FIXUP:=autoreconf
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/gperftools-headers
+  CATEGORY:=Libraries
+  SECTION:=libs
+  TITLE:=Gperftools Headers
+  URL:=https://github.com/gperftools/gperftools
+  DEPENDS:= @!mips @!mipsel @!powerpc
+endef
+
+define Package/gperftools-runtime
+  SECTION:=libs
+  CATEGORY:=Libraries
+  TITLE:=Gperftools Runtime
+  URL:=https://github.com/gperftools/gperftools
+  DEPENDS:= +libunwind +libstdcpp @!mips @!mipsel @!powerpc
+endef
+
+define Package/gperftools-headers/description
+  Gperftools is a collection of a high-performance multi-threaded malloc() implementation, plus some pretty nifty performance analysis tools.
+  This package contains the headers.
+endef
+
+define Package/gperftools-runtime/description
+  Gperftools is a collection of a high-performance multi-threaded malloc() implementation, plus some pretty nifty performance analysis tools.
+  This package contains the shared objects and bins.
+endef
+
+CONFIGURE_ARGS += \
+	--enable-frame-pointers \
+	--enable-libunwind \
+	--disable-deprecated-pprof
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libtcmalloc.so* $(1)/usr/lib/
+endef
+
+define Package/gperftools-headers/install
+	$(INSTALL_DIR) $(1)/usr/include/google
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/include/google/tcmalloc.h $(1)/usr/include/google
+	$(INSTALL_DIR) $(1)/usr/include/gperftools
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/include/gperftools/tcmalloc.h $(1)/usr/include/gperftools
+endef
+
+define Package/gperftools-runtime/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libtcmalloc.so* $(1)/usr/lib/
+endef
+$(eval $(call BuildPackage,gperftools-headers))
+$(eval $(call BuildPackage,gperftools-runtime))


libs/gperftools/Makefile
```diff
@@ -0,0 +1,80 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gperftools
+PKG_VERSION:=
```
BKPepe, line 4: ?

libs/gperftools/Makefile
```diff
@@ -0,0 +1,80 @@
...
+PKG_RELEASE:=1
+
+#PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+#PKG_SOURCE_URL:=https://codeload.github.com/$(PKG_NAME)/$(PKG_NAME)/tar.gz/$(PKG_NAME)-$(PKG_VERSION)?
+#PKG_HASH:=83e3bfdd28b8bcf53222c3798d4d395d52dadbbae59e8730c4a6d31a9c3732d8
```
BKPepe, line 9: Why it is commented?

libs/gperftools/Makefile
```diff
@@ -0,0 +1,80 @@
...
+
+PKG_SOURCE_URL:=https://github.com/gperftools/gperftools
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_DATE:=2023-07-22
+PKG_SOURCE_VERSION:=42dde1b883b29921ac89234369f88b133821690c
```
BKPepe, line 14: Do we really need to check out the sources to have the latest version instead of properly tagged version? If you want to have tagged the latest version, just ask the developer.

libs/gperftools/Makefile
```diff
@@ -0,0 +1,80 @@
...
+CONFIGURE_ARGS += \
+	--enable-frame-pointers \
+	--enable-libunwind
+
+define Build/Configure
+	( cd $(PKG_BUILD_DIR); ./autogen.sh );
+	$(call Build/Configure/Default)
+endef
+
+# This installs files into ./staging_dir/. so that you can cross compile from the host
```
BKPepe, line 60: Just drop this.

libs/gperftools/Makefile
```diff
@@ -0,0 +1,80 @@
...
+define Package/gperftools-headers/install
+	$(INSTALL_DIR) $(1)/usr/include/google
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/google/*.h $(1)/usr/include/google
+	$(INSTALL_DIR) $(1)/usr/include/gperftools
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/gperftools/*.h $(1)/usr/include/gperftools
+endef
+
+define Package/gperftools-runtime/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(CP) $(PKG_INSTALL_DIR)/usr/bin/* $(1)/usr/bin/
```
BKPepe, line 75: **You decided to install everything to /usr/bin**. Not sure, if this is a great idea, but please avoid using CP. There are also other installation macros: https://openwrt.org/docs/guide-developer/packages#file_installation_macros

libs/gperftools/Makefile
```diff
@@ -0,0 +1,80 @@
...
+  URL:=https://github.com/gperftools/gperftools
+  DEPENDS:=@TARGET_x86_64 +libc +libunwind +libgcc +libstdcpp
+endef
+
+define Package/gperftools-headers/description
+  This package contains the headers for gperftools.
+endef
+
+define Package/gperftools-runtime/description
+  This package contains the shared objects and bins for gperftools.
```
BKPepe, line 48: This is kinda short description, though.

BKPepe (changes requested):

libs/gperftools/Makefile
graysky2: I modeled this after another git Makefile where `PKG_VERSION:=` was undefined.  I assume it was needed.

libs/gperftools/Makefile
graysky2: I commented this because upstream will eventually create a tagged release containing the needed fixes.  For now, the fixes are in the branch untagged.

libs/gperftools/Makefile
graysky2: I asked: https://github.com/gperftools/gperftools/issues/1406#issuecomment-1658987755

libs/gperftools/Makefile
graysky2: Done.

libs/gperftools/Makefile
graysky2: I updated `$(CP)` to `$(INSTALL_DATA)` throughout.

libs/gperftools/Makefile
graysky2: I expanded this.

libs/gperftools/Makefile
graysky2: I switched to a tag now that the developer included one

libs/gperftools/Makefile
graysky2: Fixed in latest commit

libs/gperftools/Makefile
graysky2: fixed in latest commit

graysky2:
I am not sure why this is failing now.

libs/gperftools/Makefile
```diff
@@ -0,0 +1,77 @@
...
+	--disable-deprecated-pprof
+
+define Build/Configure
+	( cd $(PKG_BUILD_DIR); ./autogen.sh );
+	$(call Build/Configure/Default)
+endef
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/lib/*.so* $(1)/usr/lib/
```
BKPepe, line 60: Do you want to really install every .so files, which it will found? 🤔

graysky2: I'm not totally sure which of them are needed for the dev package.  Since it is for the cross compiler thought it best to include them all.  I can strip them down for the runtime package.

PolynomialDivision:
Why is this package still in review?

BKPepe:
Hmmm, feel free to review it. ;-)

Dne út 19. 12. 2023 17:48 uživatel Nick Hainke ***@***.***>
napsal:

graysky2:
@neheb - Thank you for your time and assistance with hyperscan.  Can I ask you to please review this PR as well?  I am eager to update snort to use both of these.

libs/gperftools/Makefile
```diff
@@ -0,0 +1,75 @@
...
+  URL:=https://github.com/gperftools/gperftools
+  DEPENDS:=@TARGET_x86_64
+endef
+
+define Package/gperftools-runtime
+  SECTION:=libs
+  CATEGORY:=Libraries
+  TITLE:=Gperftools Runtime
+  URL:=https://github.com/gperftools/gperftools
+  DEPENDS:=@TARGET_x86_64 +libc +libunwind +libgcc +libstdcpp
```
neheb, line 35: Aren’t libc and libgcc added by default?

graysky2: I am not sure.  I found some other Makefiles listing them out.  I removed them.

graysky2:
@neheb - do you have any objections to merging this as-is or are additional changes required?

BKPepe:
Sure, there is a new version.

libs/gperftools/Makefile
```diff
@@ -0,0 +1,75 @@
...
+endef
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/lib/libtcmalloc.so* $(1)/usr/lib/
+endef
+
+define Package/gperftools-headers/install
+	$(INSTALL_DIR) $(1)/usr/include/google
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/include/google/*.h $(1)/usr/include/google
```
BKPepe, line 65: Do you really need everything? Are you sure about it?

libs/gperftools/Makefile
```diff
@@ -0,0 +1,75 @@
...
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/gperftools-headers
+  CATEGORY:=Libraries
+  SECTION:=libs
+  TITLE:=Gperftools Headers
+  URL:=https://github.com/gperftools/gperftools
+  DEPENDS:=@TARGET_x86_64
```
BKPepe, line 27: This is only x86_x64?  Shouldnt this work also on aarch64 and riscv?

libs/gperftools/Makefile
```diff
@@ -0,0 +1,75 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=gperftools
+PKG_VERSION:=2.13
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/gperftools/gperftools/tar.gz/gperftools-$(PKG_VERSION)?
```
BKPepe, line 8:
```suggestion
PKG_SOURCE_URL:=https://codeload.github.com/gperftools/gperftools/tar.gz/$(PKG_NAME)-$(PKG_VERSION)?
```

libs/gperftools/Makefile
```diff
@@ -0,0 +1,75 @@
...
+
+define Build/Configure
+	( cd $(PKG_BUILD_DIR); ./autogen.sh );
+	$(call Build/Configure/Default)
+endef
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/lib/libtcmalloc.so* $(1)/usr/lib/
+endef
```
BKPepe, line 61: Do you need this? If yes for what?

graysky2: Yes, without this, building snort3 against this package will fail:
```
...
ninja: Entering directory `/scratch/union/build_dir/target-x86_64_musl/snort3-3.1.77.0'
ninja: error: '/scratch/union/staging_dir/target-x86_64_musl/usr/lib/libtcmalloc.so', needed by 'src/snort', missing and no known rule to make it
make[2]: *** [Makefile:161: /scratch/union/build_dir/target-x86_64_musl/snort3-3.1.77.0/.built] Error 1
```

BKPepe: Yeah, then I realized once I saw https://github.com/openwrt/packages/pull/23056, otherwise this was useless and no needed.

libs/gperftools/Makefile
graysky2: You are right, these can be trimmed down.

libs/gperftools/Makefile
graysky2: I believe you are correct in that others can use this.

graysky2:
@BKPepe - I optimized based on your suggestions.  I verified that it works and builds a functional snort3.  Do you want this squashed into a single commit?

BKPepe:
Yup, it should be squashed into the single commit.

libs/gperftools/Makefile
BKPepe: Hmm, are you sure that it is gonna to be compiled and useful for each target, which we do here? Lets see, though.

graysky2:
> Yup, it should be squashed into the single commit.

Done, thanks.

graysky2:
@BKPepe - any thoughts about the failing checks?

BKPepe:
1. Both mips should be reported to upstream
Error is here:
```
In file included from src/heap-checker.cc:46:
src/heap-checker.cc: In function 'std::pair<bool, unsigned int> CaptureRegs(pid_t, const Body&)':
src/heap-checker.cc:110:42: error: 'EF_REG29' was not declared in this scope
  110 |   sp_offset = offsetof(struct user, regs[EF_REG29]);
      |                                          ^~~~~~~~
```

2. And it looks like that for powerpc, there are several things, most likely gperftools does not support it. You can reach upstream as well for this, maybe they will tell you more.

graysky2:
For now, shall we exclude building for them?  Is the `PKG_BUILD_FLAGS` variable the correct way?  If so, what is syntax for multiple values (space or comma) and are the correct values  for all mips24 and powerpc targets?  Thanks.

BKPepe:
Why not exclude them in depends like you have it for x86_64? :) Sure, it works. We can exclude them for now. Sounds good. But try to reach upstream about that.

graysky2:
> Why not exclude them in depends like you have it for x86_64

Does it accept the syntax of not like this: `!@TARGET_mpc85xx`?  I do not know how to see a list of valid `@TARGET*` values.  Can you point me to a file or document?

BKPepe:
You can look how it is done in other Makefiles always. Right? Also, you can
search it. :)

Dne pá 5. 1. 2024 22:36 uživatel graysky ***@***.***> napsal:

graysky2:
You're right... based my latest commit on https://github.com/openwrt/packages/blob/master/utils/reptyr/Makefile

Did I catch too many or to few?

graysky2:
@BKPepe - does the CI honor the `DEPENDS`?  It still ran the affected arches.

graysky2:
@BKPepe - I needed to define them in both packages.  All checks are good now.  Is this OK to merge as-is?

libs/gperftools/Makefile
```diff
@@ -0,0 +1,75 @@
...
+  This package contains the shared objects and bins.
+endef
+
+CONFIGURE_ARGS += \
+	--enable-frame-pointers \
+	--enable-libunwind \
+	--disable-deprecated-pprof
+
+define Build/Configure
+	( cd $(PKG_BUILD_DIR); ./autogen.sh );
```
BKPepe, line 54: @neheb Would it be better here to use ``PKG_FIXUP:=autoreconf``?

libs/gperftools/Makefile
```diff
@@ -0,0 +1,75 @@
...
+define Package/gperftools-headers/install
+	$(INSTALL_DIR) $(1)/usr/include/google
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/include/google/tcmalloc.h $(1)/usr/include/google
+	$(INSTALL_DIR) $(1)/usr/include/gperftools
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/include/gperftools/tcmalloc.h $(1)/usr/include/gperftools
+endef
+
+define Package/gperftools-runtime/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DATA) $(PKG_INSTALL_DIR)/usr/lib/libtcmalloc.so* $(1)/usr/lib/
```
BKPepe, line 72: @neheb I think that we could set ABI_VERSION in package define. What do you think?

BKPepe:
> @BKPepe - does the CI honor the `DEPENDS`? It still ran the affected arches.

Yes, you did it correctly! 👍 Thanks!

graysky2:
@neheb @BKPepe - I switch to the `$(CP)` macro, rationale in: https://github.com/openwrt/packages/pull/23512

libs/gperftools/Makefile
graysky2: Note that I changed the `$(INSTALL_DATA)` macro to the `$(CP)` macro which correctly handles the symlinks.

BKPepe:
> 1. Both mips should be reported to upstream
>    Error is here:
> 
> ```
> In file included from src/heap-checker.cc:46:
> src/heap-checker.cc: In function 'std::pair<bool, unsigned int> CaptureRegs(pid_t, const Body&)':
> src/heap-checker.cc:110:42: error: 'EF_REG29' was not declared in this scope
>   110 |   sp_offset = offsetof(struct user, regs[EF_REG29]);
>       |                                          ^~~~~~~~
> ```
> 
> 2. And it looks like that for powerpc, there are several things, most likely gperftools does not support it. You can reach upstream as well for this, maybe they will tell you more.

It looks like, no one reported it to upstream, right? :-)

For powerpc:
https://github.com/gperftools/gperftools/issues/1602

PR #21642, opened 2023-07-23
tor-hs: use named sections

stokito:
Maintainer: @ja-pa
Compile tested: no
Run tested: TurrisOS 6.3.3

Description: minor changes to improve usability, see commit messages.


stokito:
I also wanted to discuss how the tor
```
config tor 'conf'
	option default '/etc/tor/torrc'
	option generated '/tmp/torrc'
	list tail_include '/etc/tor/torrc_generated'
```

The /tmp/torrc looks like
```
## This file was automatically generated please do not edit here !
%include /etc/tor/torrc
%include /etc/tor/torrc_generated
```

And the `/etc/config/tor-hs`
```
config tor-hs common
	option GenConf "/etc/tor/torrc_generated"
```

It looks for me like too many of moving parts. Why would I need to edit the `tor.conf.default` option?
Essentially this all cen be replaced with a few includes directly in the torrc file. By default it already has a line
`%include /etc/torrc.d/*.conf` but it's commented out.
Maybe just add a patch to uncomment the line directly in the torrc and this will simplify a lot of things and many options and creation of a temp file won't be needed.

I created a topic on the https://forum.torproject.org/ where proposed to the Tor team to do that on their side. Also to create a standard location for hidden services.
The topic wait's for moderation yet.

The tor-hs itself has a problem that the generated file is stored to `/etc/tor/torrc_generated` which may kill NAND flash eventually. Instead we may generate the file into `/var/run/tor` older which is an in-memory temp folder.
The generation itself may be skipped if last modify time of uci config file is before the creted time of the  `/etc/tor/torrc_generated`. The simple optimization but may safe some time on restart.

BKPepe:
I am not sure if @ja-pa continues using his Turris router together with tor, but to be safe about this. Let's also ping guys from @turris-cz, especially packagers (@paper42 , @miska), to look into it further because they could be interested in it. They do have Tor listed in [reForis](https://docs.turris.cz/basics/reforis/intro/) under package management:
image

stokito:
I tested on my Turris but I installed the tor from opkg. Now I enabled the Tor in packages but looks like nothing changed. How to check what exactly was installed?

Currently the `tor` package itself doesn't have any UCI options. From what I understand the tor by default creates a socks proxy on localhost:9050 but this can be changed to `SOCKSPort 192.168.1.1:9050` so that anyone from a local network can use it. I created a proxy PAC file so that all `.onion` domains are resolved with the local tor. Maybe add the `SOCKSPort` as UCI option?
Basically other settings are for bridge or exit node and looks like not needed.

stokito:
Also it may be better to introduce a new option `Target` or `Destination` instead of the `Ipv4` because it can be also IPv6 or even a domain. The `IPv4` name is highly confusing.

stokito:
I added 83e19fb to prevent error when Name is empty

BKPepe:
Let's ping anyone from @turris-cz to look into it.

stokito:
the changes are really minimal and can be reviewed just by eyes. The only questionable change is 83e19fb where onion services without a required Name are skipped.
I think it would be better to add a uci validation instead. But if the validation failed then as far I remember the whole service won't start. This may broke existing configuration if just one hidden service is broken.
Honestly I don't want to waste time on this because anyway this highly unlikely to happen. If a user will use a luci-app-tor then the Name will be always set anyway.

stokito:
@ja-pa @BKPepe please merge

BKPepe:
No one reviewed it yet and it seems like Turris guys are not interested in this anymore. I am not going to merge this unless I will or anyone else will review this.

stokito:
I made additional cleanup and added validation.

stangri:
LGTM. Not sure if it's still required to squash commits, otherwise GTG.

BKPepe:
Because Jan seems no longer be involved in package, which he created and
added, feel free to retake maintainership.

Dne čt 8. 2. 2024 16:26 uživatel Stan Grishin ***@***.***>
napsal:

BKPepe:
I know, but maybe we misunderstood each other. :) You dont need to be 2nd
maintainer as you seems to be using this package. So take maintainership

Dne čt 8. 2. 2024 16:39 uživatel Sergey Ponomarev ***@***.***>
napsal:

stokito:
I don't see any problem to have two maintainers

codekow:
Tested with various systems using version: 23.05.5

Works. I'd like to see this merged.

@stokito @ja-pa

PR #21697, opened 2023-08-01
flip: add new package

ktetzlaff:
Maintainer: me
Compile tested: x86_64_musl
Run tested: x86_64_musl

Description:

  The program converts line endings of text files between MS-DOS and
  *nix formats. It detects binary files in a nearly foolproof way and
  leaves them alone unless you override this. It will also leave files
  alone that are already in the right format and preserves file
  timestamps. User interrupts are handled gracefully and no garbage or
  corrupted files left behind.

(copied from the Debian package description)

### DIFF
--- a/utils/flip/Makefile
+++ b/utils/flip/Makefile
@@ -1,0 +1,53 @@
+#
+# Copyright (C) 2023 Kai Tetzlaff
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=flip
+PKG_VERSION:=1.20
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)_$(PKG_VERSION).orig.tar.gz
+PKG_SOURCE_URL:=@DEBIAN/pool/main/f/flip
+PKG_HASH:=4cd45e581c71d7bcf1ab824a47fb9263fe5371ce702879a7d2efa08d27253471
+
+PKG_MAINTAINER:=Kai Tetzlaff <openwrt@tetzco.de>
+PKG_LICENSE:=GPL-2.0-or-later
+
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/flip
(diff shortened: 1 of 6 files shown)


utils/flip/Makefile
```diff
@@ -0,0 +1,54 @@
...
+  alone that are already in the right format and preserves file
+  timestamps. User interrupts are handled gracefully and no garbage or
+  corrupted files left behind.
+
+  The program does not convert files to a different character set, and
+  it can not handle old Apple Macintosh line endings that use CR only.
+endef
+
+EXTRA_CPPFLAGS += -DBSD -DIX -DSTDINCLUDE
+Build/Compile=$(call Build/Compile/Default,flip)
```
feckert, line 45: I think this is not necessary or why do you need this?

ktetzlaff: Ok, yes - the `Build/Compile` line can be removed. But `EXTRA_CPPFLAGS` line is required since the flags are used to select between different build alternatives (OS, OS Variant, C standards supported by compiler). There is no working default. Without the flags, the Makefile would require additional patches.

feckert: Fits for me. I would only be interested in the line `Build/Compile=$(call Build/Compile/Default,flip)`

feckert: Please update your PR and remove the line Build/compile

ktetzlaff: Done

ktetzlaff:
@feckert: The last workflow run failed in an author name check in a merge commit from upstream/master (resulting from something I inadvertently triggered in the GH web UI). I think I fixed this by removing this merge commit (by a rebase onto upstream master). Could you please trigger the workflow to run once again?

BKPepe:
Don't take me wrong, but do we really need to have this software on router? What is your use case, if it is not secret?

@diizzyy , @dibdot What do you think about this package?

diizzyy:
Seems redundant to utils/dos2unix ?

1715173329:
Indeed. And dos2unix supports skipping binary files as well.

BKPepe:
In that case, let's close it?

1715173329:
> In that case, let's close it?

+1 by me

PR #21732, opened 2023-08-05
OpenVPN: add missing script event options to be used in the openvpn config file

egc112:
Maintainer: @neheb
Run tested: Dynalink DL-WRX36

Description:

The purpose of this patch is to add these script event options so that those can be used again when set in the openvpn config file (openvpn.ovpn) 
Discussion in [this thread](https://forum.openwrt.org/t/openvpn-custom-route-up-script-in-23-05-rc2/167105/13)

It might need some tidying up but I tested it and it appears to work.

Please have a look and consider implementing or make it possible to use all script event options in the openvpn config file in another way.

Signed-off-by: Erik Conijn <egc6774@gmail.com>

### DIFF
--- a/net/openvpn/files/openvpn.init
+++ b/net/openvpn/files/openvpn.init
@@ -144,6 +144,9 @@ openvpn_add_instance() {
 	local security="$4"
 	local up="$5"
 	local down="$6"
+	local route_up="$7"
+	local route_pre_down="$8"
+	local ipchange="$9"
 	local client=$(grep -qEx "client|tls-client" "$dir/$conf" && echo 1)
 
 	procd_open_instance "$name"
@@ -159,6 +162,9 @@ openvpn_add_instance() {
 		${client:+--ipchange "/usr/libexec/openvpn-hotplug ipchange $name"} \
 		${up:+--setenv user_up "$up"} \
 		${down:+--setenv user_down "$down"} \
+		${route_up:+--setenv user_route_up "$route_up"} \
+		${route_pre_down:+--setenv user_route_pre_down "$route_pre_down"} \
+		${ipchange:+--setenv user_ipchange "$ipchange"} \
 		--script-security "${security:-2}" \
 		$(openvpn_get_dev "$name" "$conf") \
 		$(openvpn_get_credentials "$name" "$conf")
@@ -182,9 +188,12 @@ start_instance() {
 		return 1
 	}
 
-	local up down script_security
+	local up down route_up route_pre_down ipchange script_security
 	config_get up "$s" up
 	config_get down "$s" down
+	config_get route_up "$s" route_up
+	config_get route_pre_down "$s" route_pre_down
+	config_get ipchange "$s" ipchange
 	config_get script_security "$s" script_security
 
 	[ ! -d "/var/run" ] && mkdir -p "/var/run"
@@ -193,7 +202,10 @@ start_instance() {
 		append UCI_STARTED "$config" "$LIST_SEP"
 		[ -n "$up" ] || get_openvpn_option "$config" up up
 		[ -n "$down" ] || get_openvpn_option "$config" down down
-		openvpn_add_instance "$s" "${config%/*}" "$config" "$script_security" "$up" "$down"
+		[ -n "$route_up" ] || get_openvpn_option "$config" route_up route-up
+		[ -n "$route_pre_down" ] || get_openvpn_option "$config" route_pre_down route-pre-down
+		[ -n "$ipchange" ] || get_openvpn_option "$config" ipchange ipchange
+		openvpn_add_instance "$s" "${config%/*}" "$config" "$script_security" "$up" "$down" "$route_up" "$route_pre_down" "$ipchange"
 		return
 	fi
 
@@ -203,7 +215,7 @@ start_instance() {
 	append_params "$s" $OPENVPN_PARAMS
 	append_list "$s" $OPENVPN_LIST
 
-	openvpn_add_instance "$s" "/var/etc" "openvpn-$s.conf" "$script_security" "$up" "$down"
+	openvpn_add_instance "$s" "/var/etc" "openvpn-$s.conf" "$script_security" "$up" "$down" "$route_up" "$route_pre_down" "$ipchange"
 }
 
 start_service() {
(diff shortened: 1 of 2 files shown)


AuthorReflex:
@egc112 
Ipchange event was added in same commit and already exists, but in server config this option produces an error (see https://github.com/openwrt/packages/issues/21200). Other events may be added

egc112:
Thanks @AuthorReflex for your comment, I had seen [ commit 808f67d](https://github.com/openwrt/packages/commit/808f67d6152fcac09dcda4d66e2bb285878fb3d4) which should solve  issue #21200  by making the `ipchange` event on the command line conditional as it is not compatible with `server mode`.

My patch should not change that behaviour, and I tested it also with `server mode` which worked as advertised.

My patches should take care of the parsing of the openvvpn config file to add the missing script events so that those are executed just like --up and --down are working now.

net/openvpn/files/openvpn.init
```diff
@@ -159,6 +162,9 @@ openvpn_add_instance() {
 		${client:+--ipchange "/usr/libexec/openvpn-hotplug ipchange $name"} \
 		${up:+--setenv user_up "$up"} \
 		${down:+--setenv user_down "$down"} \
+		${route_up:+--setenv user_route_up "$route_up"} \
+		${route_pre_down:+--setenv user_route_pre_down "$route_pre_down"} \
+		${ipchange:+--setenv user_ipchange "$ipchange"} \
```
AuthorReflex, line 167: ipchange parameter duplicated

AuthorReflex:
> Thanks @AuthorReflex for your comment, I had seen [ commit 808f67d](https://github.com/openwrt/packages/commit/808f67d6152fcac09dcda4d66e2bb285878fb3d4) which should solve issue #21200 by making the `ipchange` event on the command line conditional as it is not compatible with `server mode`.
> 
> My patch should not change that behaviour, and I tested it also with `server mode` which worked as advertised.
> 
> My patches should take care of the parsing of the openvvpn config file to add the missing script events so that those are executed just like --up and --down are working now.

in this configuration, parameter ipchange is duplicated. You can modify existing parameter behavior

egc112:
> 
> in this configuration, parameter ipchange is duplicated. You can modify existing parameter behavior

No it is not.

This line `${client:+--ipchange "/usr/libexec/openvpn-hotplug ipchange $name"} ` conditionally sets the ipchange event on the **openvpn command line**. (To view the whole command line you can use: `cat /proc/<PID>/cmdline > /tmp/psopenvpn` where PID is the pid of the openvpn process)

My patch sets an environmental variable with the name of the ipchange script as parsed from the openvpn config file (so if there is none it is empty). 
The script, as added in the openvpn config, is then eventually executed by the ipchange event in the hotplug script.
But it only works  if the --ipchange event is added to the command line which is not the case in `server mode`

Besides you do not set an ipchange event in server mode in the config file, you can try (I did to test) and OpenVPN warns that it is not compatible and simply removing it from the config file will let the OpenVPN server run.

As said my patches also work when using `server mode`  :)

AuthorReflex:
> As said my patches also work when using `server mode` :)

No. I made a test - if the ipchange script is defined in server mode - we will get the same error. 

<details>
<summary>Config file (generated from /etc/config/openvpn)</summary>
```
fast-io
float
ca /etc/openvpn/server.ca
cert /etc/openvpn/server.cert
cipher AES-256-GCM
cipher AES-128-CBC
client-config-dir /etc/openvpn/ccd
dev tun0
dev-type tun
dh /etc/openvpn/server.dh
ipchange /etc/openvpn/ipchange.sh
keepalive 10 60
key /etc/openvpn/server.key
port 1194
proto udp
server 192.0.6.0 255.255.255.0
verb 3
```
</details>

<details>
<summary>Log output</summary>
```
Mon Aug  7 10:31:59 2023 daemon.warn openvpn(server)[11053]: WARNING: --topology net30 support for server configs with IPv4 pools will be removed in a future release. Please migrate to --topology subnet as soon as possible.
Mon Aug  7 10:31:59 2023 daemon.warn openvpn(server)[11053]: DEPRECATED OPTION: --cipher set to 'AES-128-CBC' but missing in --data-ciphers (AES-256-GCM:AES-128-GCM:CHACHA20-POLY1305). OpenVPN ignores --cipher for cipher negotiations.
Mon Aug  7 10:31:59 2023 daemon.notice openvpn(server)[11053]: Note: NOT using '--topology subnet' disables data channel offload.
Mon Aug  7 10:31:59 2023 daemon.err openvpn(server)[11053]: Options error: --ipchange cannot be used with --mode server (use --client-connect instead)
Mon Aug  7 10:31:59 2023 daemon.warn openvpn(server)[11053]: Use --help for more information.
```
</details>


This is an incorrect configuration, but nevertheless, such a case cannot be ruled out. 
The right way is to modify the existing ipchange behavior to support custom script only in client mode
```
${client:+--setenv user_ipchange "$ipchange"} \
```

egc112:
> > As said my patches also work when using `server mode` :)
> 
> No. I made a test - if the ipchange script is defined in server mode - we will get the same error.
> Config file (generated from /etc/config/openvpn)
> Log output
> 
> This is an incorrect configuration, but nevertheless, such a case cannot be ruled out. The right way is to modify the existing ipchange behavior to support ipchange custom script only in client mode
> 
> ```
> ${client:+--setenv user_ipchange "$ipchange"} \
> ```

Thanks for testing and looking into it.
As you remarked, the ipchange event is not allowed in server mode and if you add it to the config file you get a warning as expected: 
`Mon Aug  7 11:27:29 2023 daemon.err openvpn(ovpnserver1)[11593]: Options error: --ipchange cannot be used with --mode server (use --client-connect instead)`

The warning is there both with my patch and also with your solution of:
`${client:+--setenv user_ipchange "$ipchange"} \`
which is of course as it should be.

Your solution does not add the $user_ipchange at all in server mode which is indeed not necessary and not wanted so that is fine.
My solution does add the  $user_ipchange variable but as there is no ip_change event handler in server mode (because of the conditional placement on the command line) it has no consequence.
Besides in both solutions OpenVPN server will not run at all when there is an ipchange event, which is of course the intended behaviour.

But changing the patch with your solution is fine by me, the important thing is that this is resolved one way or the other :)

egc112:
Another thought just combine it an have the best of both worlds?
`${client:+${ipchange:+--setenv user_ipchange "$ipchange"}} \`

AuthorReflex:
You need to squash commits into one.

egc112:
> You need to squash commits into one and increment build number

Thanks will do tomorrow.

egc112:
will close this and make a new fresh one as it would not let me squash the commits

PR #21736, opened 2023-08-06
rust: Update to 1.71.1

1715173329:
Maintainer: @lu-zero 
Run tested: n/a

Description:

Version 1.71.1 (2023-08-03)
\===========================

- Fix CVE-2023-38497: Cargo did not respect the umask when extracting dependencies
- Fix bash completion for users of Rustup
- Do not show `suspicious_double_ref_op` lint when calling `borrow()`
- Fix ICE: substitute types before checking inlining compatibility
- Fix ICE: don't use `can_eq` in `derive(..)` suggestion for missing method
- Fix building Rust 1.71.0 from the source tarball

### DIFF
--- a/lang/rust/Makefile
+++ b/lang/rust/Makefile
@@ -6,12 +6,12 @@ include $(TOPDIR)/rules.mk
 include ./rust-values.mk
 
 PKG_NAME:=rust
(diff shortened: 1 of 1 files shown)


lu-zero:
Thank you :)

PR #21744, opened 2023-08-07
flashrom: Update to 1.3.0, package libflashrom

computersforpeace:
Notably, the meson args for programmers changed from a series of individual 'config_$foo>, config_$bar' to a list of 'programmer=$foo,$bar'

The patches were all either already merged upstream (or similar ones, at least), or aren't useful (bitbang_spi is no longer a separated programmer; it's included in the defaults).

Maintainer: @neheb 
Compile tested: ipq806x and ipq40xx, openwrt.git head as of 06 Aug 2023
Run tested: ipq40xx / Google WiFi and ipq806x / TP-Link OnHub, tested basic flashrom read/write; also built applications against libflashrom


neheb:
Why is cmocka being built? Should be disabled IMO.

computersforpeace:
> Why is cmocka being built? Should be disabled IMO.

dunno why I didn't notice that one locally.

I updated to disable tests, which should avoid cmocka.

PR #21753, opened 2023-08-07
nvme-cli: Fix categorization

lu-zero:
Maintainer: me 
Compile tested: N/A
Run tested: N/A

Description: Change the categories as per @jefferyto suggestion.


jefferyto (approved):
Thanks!

PR #21755, opened 2023-08-07
openvpn: add missing script event options to be used in the openvpn config file

egc112:
Maintainer: @neheb @AuthorReflex
Run tested: Dynalink DL-WRX36

Description:

The purpose of this patch is to add these script event options so that those can be used again when set in the openvpn config file (openvpn.ovpn) 
Discussion in [this thread](https://forum.openwrt.org/t/openvpn-custom-route-up-script-in-23-05-rc2/167105/13)

[Previous Pull request](https://github.com/openwrt/packages/pull/21732 ) has been discussed and improved with the help of @AuthorReflex 

Signed-off-by: Erik Conijn <egc6774@gmail.com>

### DIFF
--- a/net/openvpn/files/etc/hotplug.d/openvpn/01-user
+++ b/net/openvpn/files/etc/hotplug.d/openvpn/01-user
@@ -7,10 +7,14 @@
 		$*
 }
 
-# Wrap user defined scripts on up/down events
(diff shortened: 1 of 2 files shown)


egc112:
If the commit is approved, is it possible to backport it to 23.05?
I am not allowed to put a label on it (understandably)

PR #21758, opened 2023-08-08
openvpn: add missing script event options to be used in the openvpn config file

egc112:
Maintainer: @mkrkn  @neheb
Run tested: Dynalink DL-WRX36

Description:

This PR fixes a regression that makes it problematic to specify certain event options in the OpenVPN configuration file.

Discussion in [this thread](https://forum.openwrt.org/t/openvpn-custom-route-up-script-in-23-05-rc2/167105/13) and [here](https://forum.openwrt.org/t/openvpn-route-up-and-route-pre-down-broken-in-23-05/176568)

Please have a look and consider implementing or make it possible to use all script event options in the openvpn config file in another way.

Pull request has been discussed and improved with the help of @AuthorReflex, see: https://github.com/openwrt/packages/pull/21732 

Signed-off-by: Erik Conijn <egc112@msn.com>


egc112:
I am not allowed to put a label on it (understandably) but can this also be backported to 23.05?

net/openvpn/files/openvpn.init
```diff
@@ -159,6 +162,9 @@ openvpn_add_instance() {
 		${client:+--ipchange "/usr/libexec/openvpn-hotplug ipchange $name"} \
```
AuthorReflex, line 162: this line not needed if you redefined new behavior below

BKPepe:
Lets ping @miska about this since he introduced those options as you refenced.

net/openvpn/files/openvpn.init
egc112: If we remove it the ipchange action in the hotplug script (`/etc/hotplug.d/openvpn/01-user`) is never triggered.

The beauty of the solution is that it combines OpenWRT's hotplug actions via the hotplug script but makes it also compatible with the old way of placing script events in the openvpn config file by parsing the openvpn config file and add those scripts to the hotplug actions by setting it as an environmental variable e.g. `$user_ipchange`.

By removing  `${client:+--ipchange "/usr/libexec/openvpn-hotplug ipchange $name"} \ ` we are breaking this.

Sure the ipchange script in the openvpn config file is exectued but by openvpn and not by the hotplug action as it is intended to be.

AuthorReflex: If parameter specified more once, only last occurence will be accepted to execute. So, even without removing, hotplug event do not exeuted

ThomasCr:
I also miss the route-up and route-pre-down option, because I setup extra routing table/rules based on a list of hostnames I translate at runtime and for a specific client on my network.  Now I am a little bit lost, that my setup is broken.
Is there any change, that this get fixed in the next months?

supersebbo:
This behaviour introduced in 23.05 has completely broken my VPN routes, when will this fix be merged?

zorxd:
A possible workaround is to add the following to /etc/openvpn.user

`
if [ "$ACTION" == "route-up" ]
then
        /sbin/ip route whatever argument here
fi
`

net/openvpn/files/openvpn.init
vgaetera: The current code already overrides the options from the config file with command-line options, so OpenVPN ignores the `route-up`, `route-pre-down`, and `ipchange` options in the config file.

The proposed code exports the relevant options as environment variables which are used in the hotplug script to call the action-specific scripts and make the options from the config file work again.

PR #21787, opened 2023-08-11
collectd: fix implicit dependence for collect-mod-disk to libudev

feckert:
Maintainer: @hnyman 
Compile tested: x86_64, APU3
Run tested: no

Description:

Fix missing dependencies for 'collectd-mod-disk' to 'libudev' if collectd-mod-smart is not enabled.
```
Package collectd-mod-disk is missing dependencies for the following libraries:
libudev.so.1
```
The package 'collect-mod-disk' is now build always with 'libudev' support, independent of the enable/disable build state of 'collectd-mod-smart'.

### DIFF
--- a/utils/collectd/Makefile
+++ b/utils/collectd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=collectd
 PKG_VERSION:=5.12.0
-PKG_RELEASE:=46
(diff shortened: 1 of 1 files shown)


feckert:
@hnyman thanks for merging. I have also backported this to[ openwrt-23.05](https://github.com/openwrt/packages/commit/7ceceaec904d61cd39df2c6f3d12c16910b24e6b)

PR #21801, opened 2023-08-12
python3 dbus-fast/aio-mqtt-mod/ble2mqtt update to latest releases

QuintinHill:
Maintainer: me 
Compile tested: pending
Run tested: pending

Description: update to latest releases


lang/python/python-dbus-fast/patches/001-unpin-setuptools.patch
```diff
@@ -1,6 +1,6 @@
 --- a/pyproject.toml
 +++ b/pyproject.toml
-@@ -97,5 +97,5 @@ module = "docs.*"
+@@ -96,5 +96,5 @@ module = "docs.*"
```
jefferyto, line 3: This patch can be removed now that host setuptools is new enough ([65.5.0](https://github.com/openwrt/packages/blob/162ca6ad7f12378a1e9bb13c4835c0d06ddc90ea/lang/python/python3-version.mk#L18)).

QuintinHill: Removed.  Can this be removed when backporting to 23.05 too?

jefferyto: Yes, 23.05 is still in sync with master.

QuintinHill:
This should be good to go now.

PR #21803, opened 2023-08-13
natmap: Add some new features

muink:
Maintainer: Richard Yu @ysc3839 , Ray Wang @heiher 
Compile tested: x86_64, 22.03.5
Run tested: x86_64, 22.03.5

Description:
Add some new features and adjust some uci options.
- Add global stun_server, http_server and interval
- Auto configure the Firewall
- NAT Loopback support
- uci: interface --> bind_ifname, notify_script --> custom_script
- forward port = 0 support (use public Port as forward port)
- refresh client listen port (when forward port = 0)

### DIFF
--- a/net/v2raya/Makefile
+++ b/net/v2raya/Makefile
@@ -5,12 +5,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=v2rayA
-PKG_VERSION:=2.1.0
+PKG_VERSION:=2.1.3
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/v2rayA/v2rayA/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=97cffd5dd568fc3be22fa73768bbc218a0650548ddb0849506eeb168dde0291a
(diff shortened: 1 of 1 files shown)


ysc3839:
The code is bloated, especially `natmap.init`. Personally think it's unmaintainable, and even buggy. There are `eval` in that script, which is likely vulnerable to remote code execution.

Is there a "temporary firewall rules" feature? When firewall reload (or system reboot?), these rules would be automatically deleted. Just like firewalld's "without `--permanent`". Ping @jow-

muink:
@ysc3839 Yes, firewall rules are added at runtime and cleared at stop.
After restarting the firewall, the added rules will be cleared and will not be re-added

ysc3839:
@muink I'm not asking you. I'm asking if fw3/fw4 itself has such feature. If so, there's no need to write such bloated code.

muink:
```
procd_open_instance
procd_open_data
json_add_array firewall
...
json_close_array
procd_close_data
procd_close_instance

service_started() { procd_set_config_changed firewall; }
service_stopped() { procd_set_config_changed firewall; }
```
The above code seems to be able to add/clear firewall rules on instance started/stoped
But don't know how to replace/remove the added rules at runtiming

muink:
The main features have been added, can they be merged? @ysc3839 @heiher

PR #21812, opened 2023-08-13
ruby: fix riscv64 build

1715173329:
Maintainer: @luizluca 
Compile tested: sifiveu/generic
Run tested: n/a

Description:
Fix build by adding missing macros.

The patch is taken from:
https://lists.openembedded.org/g/openembedded-core/message/161168

See also:
https://gitlab.alpinelinux.org/alpine/aports/-/issues/13667

### DIFF
--- a/lang/ruby/Makefile
+++ b/lang/ruby/Makefile
@@ -12,7 +12,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ruby
 PKG_VERSION:=3.2.2
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 # First two numbes
 PKG_ABI_VERSION:=$(subst $(space),.,$(wordlist 1, 2, $(subst .,$(space),$(PKG_VERSION))))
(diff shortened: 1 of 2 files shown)


luizluca:
Thanks for the patch. I just don't know if it will not bite us in the future. The patch mentions it is a musl bug. Once they fix it, won't it break this patch?

Is it reported to or fixed in musl upstream? If we have it fixed upstream, we could backport the musl fix or, at least, version-target this patch.

1715173329:
> The patch mentions it is a musl bug. Once they fix it, won't it break this patch?

Once these macros are defined in musl, this patch needs to be removed.

> Is it reported to or fixed in musl upstream? If we have it fixed upstream, we could backport the musl fix or, at least, version-target this patch.

The author made a patch for musl: kraj/musl@9ec2f12207b4, but it's not upstreamed yet.

luizluca:
> The author made a patch for musl: kraj/musl@9ec2f12207b4, but it's not upstreamed yet.

If it was upstream, I would suggest to submit that as an openwrt backport for musl. I might ping the author about upstreaming that patch.

From ruby side, the patch is still hacky. It assumes "riscv and not glibc" as "musl with missing macros". Musl does not define a __MUSL__ macro exactly for that reason. It should have checked if the required macros are defined or not and not that musl will always be broken.

I'm not willing to accept a patch that I know will break in the future. You can update this PR checking for the missing macros or, in a couple of days, I can do that in a new PR.

1715173329:
> You can update this PR checking for the missing macros

Okay I updated it to detect riscv and missing macros.

luizluca:
Thanks. With that changes, there is no rush to drop this patch when OpenWrt got a fixed musl

PR #21827, opened 2023-08-15
golang: Update to 1.21.0, remove patch

jefferyto:
Maintainer: me
Compile tested: armsr-armv7, 2023-08-13 snapshot sdk
Run tested: armsr-armv7 (qemu), 2023-08-13 snapshot

Description:
Upstream has updated the Go compiler to not use gold when building for arm, and is waiting for a fix to binutils (released in 2.41) before doing the same for aarch64. ([upstream issue][1])

Based on the above, it does not appear that https://github.com/golang/go/pull/49748 will be merged. This removes the patch from that pull request.

[1]: https://github.com/golang/go/issues/22040


hnyman:
This seems to have broken lots of go packages, especially those that use quic. e.g. adguardhome
E.g. https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a72/packages/


> ./../../../../dl/go-mod-cache/github.com/quic-go/quic-go@v0.35.1/internal/qtls/go121.go:5:13: cannot use "The version of quic-go you're using can't be built on Go 1.21 yet. For more details, please see https://github.com/quic-go/quic-go/wiki/quic-go-and-Go-versions." (untyped string constant "The version of quic-go you're using can't be built on Go 1.21 yet. F...) as int value in variable declaration

jefferyto:
> This seems to have broken lots of go packages,

Yes, this is why 23.05 hasn't been updated yet.

> especially those that use quic.

A Go 1.21-compatible version of quic-go has already been released ([v0.37.4](https://github.com/quic-go/quic-go/releases/tag/v0.37.4)). It's up to each Go program/package to update their own dependencies.

In general, every major Go update causes these packages to fail to build for this same exact reason, i.e. quic-go has to be updated, then each Go program has to update their dependencies. I don't see a point in waiting for most/all Go programs to be updated first.

hnyman:
Yeah, apparently AdGuard Home v0.107.36 is needed for go 1.21 compatibility

PR #21834, opened 2023-08-16
dnscrypt-proxy2: update to 2.1.5

SakuraFallingMad:
Maintainer: @BKPepe
Compile tested: n/a
Run tested: n/a
Description:
https://github.com/DNSCrypt/dnscrypt-proxy/commit/1b65fe62b01965bfee3f7ed8ba852c15de1255eb

### DIFF
--- a/net/dnscrypt-proxy2/Makefile
+++ b/net/dnscrypt-proxy2/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnscrypt-proxy2
-PKG_VERSION:=2.1.4
+PKG_VERSION:=2.1.5
 PKG_RELEASE:=1
 
 PKG_SOURCE:=dnscrypt-proxy-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/DNSCrypt/dnscrypt-proxy/tar.gz/$(PKG_VERSION)?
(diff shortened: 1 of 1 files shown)


1715173329:
Please sign your work with vaild name and E-mail with:
- `git config user.name '<your name>'`
- `git config user.email '<your email>'`
-  `git commit --amend -s`

BKPepe:
Sorry, but I am not going to accept such PRs, when you did not bother to compile it and test it locally. Because anyone can do that, no special knowledge or skills are requires to do so, which results to care about quantity over quality. What if you break something?

1715173329:
well if possible let's give a little bit more patience for our new contributor

BKPepe:
I might be a little bit rude, even though everything is about attitude, but... there is PR template for some reason and if you indicate in the first place  in both things that it is not compile or run tested, then why it should be merged. 

Also, there must be reason, why the package must be updated. There was released a new version and we need to have it updated ASAP? In some opinion, if there is fix for security vulnerability then yes, but referenced commit goes to bump version, when there are no details, etc. Proper link would be https://github.com/DNSCrypt/dnscrypt-proxy/releases/tag/2.1.5 and from that it does not looks like we need to have it in short manner.

1715173329:
> and from that it does not looks like we need to have it in short manner.

Well we do need it now, as the Go compiler was recently updated to 1.21 and we need this new version to build.

PR #21935, opened 2023-08-27
libssh: compile against MbedTLS instead of OpenSSL

G10h4ck:
Maintainer:  @neheb @PolynomialDivision 
Compile tested: librerouter-v1 ath79 OpenWrt main development branch
Run tested: librerouter-v1 ath79 OpenWrt, successfully run tmate which depends on libssh

Description:
Since 2017 libssh supports to be compiled against libmbedtls instead of libopessl, OpenWrt ships mbedtls by default, while depending on OpenSSL increase image size by almost 1.5MB unnecessarily.

### DIFF
--- a/libs/libssh/Makefile
+++ b/libs/libssh/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libssh
 PKG_VERSION:=0.10.4
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


neheb:
Hmm I wonder what the reason for OpenSSL was. Maybe dependent packages use it.

neheb:
there's a new version.

BKPepe:
Also, it requires to bump PKG_RELEASE.

G10h4ck:
> there's a new version.

shouldn't we keep that change for a separate PR ?

G10h4ck:
> Also, it requires to bump PKG_RELEASE.

Done, thanks!

PR #21937, opened 2023-08-28
tailscale: enable autodect of fw type

youngt2:
Maintainer: @ja-pa @oskarirauta
Compile tested: aarch64_cortex-a53 @ SNAPSHOT
Run tested: aarch64_cortex-a53 @ SNAPSHOT

Description: Starting with v1.48.1 tailscale reverted to using iptables as default. This change enables the TS_DEBUG_FIREWALL_MODE=auto to enable autodectect of nftables / iptables. Without this change, tailscale start up will fail because the iptable command cannot be found on systems using nftables.

### DIFF
--- a/net/tailscale/files/tailscale.init
+++ b/net/tailscale/files/tailscale.init
@@ -23,6 +23,9 @@ start_service() {
   procd_open_instance
   procd_set_param command /usr/sbin/tailscaled
 
+  # starting with v1.48.1 ENV variable is required to enable autodetection of iptables / nftables
+  procd_set_param env TS_DEBUG_FIREWALL_MODE=auto
+
   # Set the port to listen on for incoming VPN packets.
   # Remote nodes will automatically be informed about the new port number,
   # but you might want to configure this in order to set external firewall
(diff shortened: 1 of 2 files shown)


1715173329:
Merged, thank you!

net/tailscale/files/tailscale.init
```diff
@@ -23,6 +23,9 @@ start_service() {
   procd_open_instance
   procd_set_param command /usr/sbin/tailscaled
 
+  # starting with v1.48.1 ENV variable is required to enable autodetection of iptables / nftables
+  procd_set_param env TS_DEBUG_FIREWALL_MODE=auto
```
alaviss, line 27: It might be better to just set `nftables` here.

Detection is only supported for arm64 and amd64: https://github.com/tailscale/tailscale/blob/dc7aa98b768bf82017aa5cc82a62dd4d685f811d/util/linuxfw/linuxfw_unsupported.go#L4C58-L4C58

On MIPS this shows up:

```
router: detect iptables rule: linuxfw:unsupported
router: detect nftables rule: linuxfw:unsupported
router: nftables rule count: 0, iptables rule count: 0
router: using iptables
```

alaviss: > I wonder why it is not finding nftables?

The detection code only supports arm64 and amd64 architectures.

> Perhaps default should be nftables, and there can be a variable in [tailscale.conf](https://github.com/openwrt/packages/blob/master/net/tailscale/files/tailscale.conf) to go back to iptables.

I think this is the way to go.

PR #21953, opened 2023-08-29
knot: update to version 3.3.0

Payne-X6:
Compile tested: ARM Cortex-A9 vfpv3, Turris Omnia, TurrisOS 7.0.0 (hbl)
Run tested: ARM Cortex-A9 vfpv3, Turris Omnia, TurrisOS 7.0.0 (hbl)

Description:

- Release upgrade (https://www.knot-dns.cz/2023-08-28-version-330.html)
- Enable QUIC support

### DIFF
--- a/net/knot/Makefile
+++ b/net/knot/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=knot
-PKG_VERSION:=3.2.9
+PKG_VERSION:=3.3.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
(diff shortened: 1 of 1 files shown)


salzmdan:
:+1:

BKPepe:
Okay, I'm merging this even though I would prefer to know why there was enabled QUIC support, because in the future while looking in the past (_Back to the future, ehm, ehm_), then you will not know why it was enabled by default, what was the size difference, etc.

salzmdan:
> Okay, I'm merging this even though I would prefer to know why there was enabled QUIC support, because in the future while looking in the past (_Back to the future, ehm, ehm_), then you will not know why it was enabled by default, what was the size difference, etc.

The DNS over QUIC support is a new feature in 3.3.0 version. If enabled, `knotd` and `kdig` can use it.

PR #21964, opened 2023-08-30
collectd-mod-ipstatistics: fix handling of long `/proc` lines

alex3d:
`/proc/net/netstat` `TcpExt:` line is 2064 chars long on linux 6.1

### DIFF
--- a/utils/collectd/patches/932-add-ipstatistics.patch
+++ b/utils/collectd/patches/932-add-ipstatistics.patch
@@ -41,7 +41,7 @@
 +
 +int ipstatistics_read() {
 +  FILE *fh;
-+  char buffer[1024];
++  char buffer[4096];
 +  char *fields[19];
 +  int numfields;
 +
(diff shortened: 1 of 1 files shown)


hnyman:
cc @PolynomialDivision who has authored the plugin

alex3d:
> Ps. Your commit is missing the signed-off-by line

fixed

hnyman:
Applied to master as b5bf5927f702  (with PKG_RELEASE also bumped)

Not sure if that is needed for 23.05. When has the size grown? 5.10 or 5.15 or 6.1?

alex3d:
> Not sure if that is needed for 23.05. When has the size grown? 5.10 or 5.15 or 6.1?

Currently I don't have OpenWrt with 5.15 on hand, but on ubuntu kernel 5.15 size has already grown. So I assume the patch is needed for 23.05.

In fact line size was more than 1024 a long time ago. Previously line was split into two "long" parts so it was not trigger `if (numfields < 8)` check in this plugin.
```
root@OpenWrt:~# uname -a
Linux OpenWrt 5.4.188 #0 Sat Apr 16 12:59:34 2022 mips GNU/Linux
root@OpenWrt:~# cat /proc/net/netstat | head -n 1 | wc
        1       120      1942
```

Issue #21956, opened 2023-08-29
mhz: Compilation fails due to missing -fPIC flag on mvebu


kesslerm:
Maintainer: @robimarko
Environment: (mvebu, turris omnia, 23.05 / main)

Description:

Without `-fPIC` linking fails with the OpenWRT mvebu toolchain:

```
/home/ubuntu/development/TurrisBuild/staging_dir/toolchain-arm_cortex-a9+vfpv3-d16_gcc-12.3.0_musl_eabi/lib/gcc/arm-openwrt-linux-muslgnueabi/12.3.0/../../../../arm-openwrt-linux-muslgnueabi/bin/ld: mhz.o: relocation R_ARM_MOVW_ABS_NC against `a local symbol' can not be used when making a shared object; recompile with -fPIC
/home/ubuntu/development/TurrisBuild/staging_dir/toolchain-arm_cortex-a9+vfpv3-d16_gcc-12.3.0_musl_eabi/lib/gcc/arm-openwrt-linux-muslgnueabi/12.3.0/../../../../arm-openwrt-linux-muslgnueabi/bin/ld: mhz.o(.text+0x75c): unresolvable R_ARM_CALL relocation against symbol `__aeabi_l2d@@GCC_3.5'
```

In my build environment I solved this by adding a patch to the build that fixes up the upstream mhz Makefile:
```
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 CC         := gcc
-CFLAGS     := -O3 -Wall -fomit-frame-pointer
+CFLAGS     := -O3 -Wall -fomit-frame-pointer -fPIC
 
 all: mhz
 
```

PR #22001, opened 2023-09-02
mhz: respect CC and CFLAGS passed by buildsystem

robimarko:
Maintainer: me
Compile tested: mvebu, a9/Turris Omnia and ipq807x/Qnap 301W
Run tested: ipq807x/Qnap 301W

Description:
It seems that the Makefile has both CC and CFLAGS hardcoded and does not allow overriding them by ones being passed by the buildsystem.

This works fine until CONFIG_PKG_ASLR_PIE_ALL is selected, then building will fail with:
```
arm-openwrt-linux-muslgnueabi/bin/ld.bfd: mhz.o: relocation R_ARM_MOVW_ABS_NC against `a local symbol' can not be used when making a shared object; recompile with -fPIC
arm-openwrt-linux-muslgnueabi/bin/ld.bfd: mhz.o(.text+0x75c): unresolvable R_ARM_CALL relocation against symbol `__aeabi_l2d@@GCC_3.5
```

So, lets add a patch pending upstream that allows both CC and CFLAGS to be overriden so that ones passed by the buildsystem are actually respected.

Fixes: 89123b308f98 ("mhz: add new package")
Fixes: #21956

### DIFF
--- a/utils/mhz/Makefile
+++ b/utils/mhz/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mhz
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=https://github.com/wtarreau/mhz.git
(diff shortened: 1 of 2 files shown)


robimarko:
@kesslerm This should sort everything out

BKPepe (approved):

kesslerm:
@robimarko, @BKPepe: The fix works for me. Thanks very much for taking care of this issue.

BKPepe:
Cool. Merging!

PR #22004, opened 2023-09-03
syslog-ng: update to version 4.3.1

BKPepe:
Maintainer: me
Compile tested: MacBook Pro, A1990, OpenWrt 22.03 for Turris Omnia, mvebu/cortexa9

Run tested: N/A, since I could not compile it for OpenWrt 21.02 due to Darwin platform, which resulted in cherry-picking a few commits to that branch, hopefully to be compiled, because the latest Turris OS version, which is now 6.3.1 still runs EoL, thus in the meantime, I prepared locally build env for OpenWrt 22.03, where it works.
Once, I confirm it that it works on the router, I will set my PR as ready instead of draft.

Description:
Makefile changes:
- Since version 4.3.0, there is required to use pcre2 instead of pcre 
Reference: https://github.com/syslog-ng/syslog-ng/pull/4537 (not mentioned in the release notes in tag :( )

- Disable c++ support by default to avoid picking and including libstdcpp dependency
Reference: https://github.com/syslog-ng/syslog-ng/pull/4484

Config changes:
- Bump version in config file

### DIFF
--- a/admin/syslog-ng/Makefile
+++ b/admin/syslog-ng/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=syslog-ng
-PKG_VERSION:=4.2.0
+PKG_VERSION:=4.3.1
 PKG_RELEASE:=1
 
 PKG_MAINTAINER:=Josef Schlehofer <josef.schlehofer@nic.cz>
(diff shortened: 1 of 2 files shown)


BKPepe:
Run tested - Turris Omnia, mvebu/cortexa9, OpenWrt 22.03:

  ```
  root@turris:~# syslog-ng --version
syslog-ng 4 (4.3.1)
Config version: 4.2
Installer-Version: 4.3.1
Revision:
Compile-Date: Sep  2 2023 14:02:20
Module-Directory: /usr/lib/syslog-ng
Module-Path: /usr/lib/syslog-ng
Include-Path: /usr/share/syslog-ng/include
```

PR #22008, opened 2023-09-03
Remove package: Bandwidthd

BKPepe:
Maintainer: @padre-lacroix 
Compile and run tested: N/A

Description:
This package is not maintained by upstream anymore as they replace it by Ntopng 3. [1]
Upstream repository is archived and read only since February 2022 [2]

[1] https://community.nethserver.org/t/bandwidthd-for-nethserver-7/4071/18
[2] https://github.com/NethServer/bandwidthd


1715173329 (approved):

padre-lacroix:
There are potentially other sources of bandwidthd that can be used.  Originally when I took bandwidthd I moved the source from the original bandwidthd on sourceforge to the current one.  Now that the current one has stopped maintaining it, I may look for another source, and if lucky, one that will support IPv6 also (there is one that I investigated a few years back, but I do not know if it is still maintained).

BKPepe:
I don’t see a reason to keep this package here or even find different sources for it when upstream claimed that it is not going to be maintained and supported as well because they replaced it completely with a different package. What is the purpose? Some software is getting buried and we should not try to keep it alive for a short notice and then leave it also dead. 

Even common GNU/Linux distributions don’t have this package. We don’t need to be special. We are not. We do have a lack of maintainers. A few maintainers are raising policy quantity over quality. 

Let's not reinvent the wheel once again and keep it simple:
- If anyone still wants to have dead project installed on their routers, fine. They can compile it locally. 
- If they want to have something that gets new features and updates, they can use any other alternatives, which we do have in the repository.

PR #22025, opened 2023-09-04
[openwrt-23.05] python-voluptuous-serialize: Update to 2.6.0

jefferyto:
Maintainer: @BKPepe
Compile tested: none (cherry picked from #21948)
Run tested: none

Description:
(cherry picked from commit 92d49899e448efb1e362568184a5dab82c171542)

### DIFF
--- a/lang/python/python-voluptuous-serialize/Makefile
+++ b/lang/python/python-voluptuous-serialize/Makefile
@@ -7,16 +7,16 @@
 
 include $(TOPDIR)/rules.mk
 
-PKG_NAME:=voluptuous-serialize
-PKG_VERSION:=2.5.0
+PKG_NAME:=python-voluptuous-serialize
+PKG_VERSION:=2.6.0
 PKG_RELEASE:=1
 
-PYPI_NAME:=$(PKG_NAME)
-PKG_HASH:=5359f2e0a4f972ae03066e0777b4f0755c9226b2af099ca4fc55432efd1a447b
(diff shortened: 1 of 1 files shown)


BKPepe:
Guys, excuse me for bringing off-topic here!

Hey @jefferyto,

Nice to see you back here! You are doing really awesome work. Keep going. :)
You rocks. So many PRs and notifications these days. You were really missing. I think OpenWrt should somehow motivate you to be more involved in it. What we would be doing without you.

Hopefully, I will continue my contributions as I did in past, so I can help you with updating these Python packages, which would result that you will have less work or you can focus on something else within OpenWrt.

jefferyto:
> Nice to see you back here! You are doing really awesome work. Keep going. :)

Thanks :smile: 

> So many PRs and notifications these days.

Mainly because these are things I want to get done before 23.05 final is released.

> You were really missing. I think OpenWrt should somehow motivate you to be more involved in it.

When I'm missing it's usually because I need to focus on work. When I'm here it is either because I am procrastinating from doing work or I have no work :sweat_smile: 

> What we would be doing without you.

Someone else will step in / step up. I'm only a cog in the machine lol

> Hopefully, I will continue my contributions as I did in past, so I can help you with updating these Python packages, which would result that you will have less work or you can focus on something else within OpenWrt.

More help is always welcome. It would be even better if the folks at Turris/CZ.NIC can help to update packages, many of the Python packages were added to support their work.

PR #22027, opened 2023-09-04
libedit: update to version 20230828-3.1

Payne-X6:

Description:

- update of the library on which `knot` depends on


salzmdan:
:+1:

PR #22029, opened 2023-09-04
treewide: change my no longer used email

BKPepe:
Since February 2023, I decided to no longer work with Turris, I mean CZ.NIC company due to some reasons how the development goes and since that day my work address is not available and not sure if there is some redirect to someone else, but if anyone wants to reach me, use my email address, where they can find me.

____

Also, it does not require to bump PKG_VERSION. My first thought of doing this was like, hey, I am not going to use these packages personally except syslog-ng and they are needed for Turris, so in the future, guys from Turris show up to take maintainership of these packages or there will be some alias, but... nope. That didn't happen and I haven't noticed that someone from Turris is active here as I was. Lets hope for brighter future now.

### DIFF
--- a/admin/syslog-ng/Makefile
+++ b/admin/syslog-ng/Makefile
@@ -4,7 +4,7 @@ PKG_NAME:=syslog-ng
 PKG_VERSION:=4.2.0
 PKG_RELEASE:=1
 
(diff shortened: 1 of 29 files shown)


BKPepe:
Ok, dnscrypt-proxy2 is failing, will be addressed shortly. However, not related to this PR. Merging.

PR #22045, opened 2023-09-05
acme-acmesh: Honor dns_wait option in ACME config for renewals

Fail-Safe:
The dns_wait option can be set in the ACME config within OpenWrt. However, it was not handle on cert renewal for those who choose to set the option.

Maintainer: @tohojo
Compile tested: Linksys E8450/Belkin RT3200
Run tested: Linksys E8450/Belkin RT3200

### DIFF
--- a/admin/syslog-ng/Makefile
+++ b/admin/syslog-ng/Makefile
@@ -4,7 +4,7 @@ PKG_NAME:=syslog-ng
 PKG_VERSION:=4.2.0
 PKG_RELEASE:=1
 
-PKG_MAINTAINER:=Josef Schlehofer <josef.schlehofer@nic.cz>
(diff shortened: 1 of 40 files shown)


tohojo:
You'll need to add the package name prefix to the commit description itself to make the bot happy :)

What about the other dns-related options (like --challenge-alias)? Are they preserved automatically from the state dir?

tohojo:
Hmm, so looking at the acme.sh code, AFAICT all those values are actually saved in the state dir and supplying them on renew should not be necessary. This includes the `--dnssleep` parameter, so I'm a little puzzled why this PR is necessary at all?

Fail-Safe:
I'm sure you're right--I probably have something odd going on with my test scenario. I'll close this PR and review things on my side to see what is going on. Sorry for any time lost on this one! 😞

PR #22057, opened 2023-09-06
dawn: Update to latest upstream version (2023-05-14)

schuettecarsten:
Compile and run tested on Belkin 3200/Linksys 6450 (mt7622) and TP-Link Archer C7 v5 (ath79)
@PolynomialDivision


net/dawn/Makefile
```diff
@@ -5,13 +5,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dawn
-PKG_SOURCE_DATE:=2022-07-24
-PKG_RELEASE:=3
+PKG_SOURCE_DATE:=2023-05-14
+PKG_RELEASE:=4
```
feckert, line 9: Please set PKG_RELEASE to 1 this is a new upstream version. PKG_RELEASE should only increment if we do not change the upstream version. And if we have a new upstream version the we reset the PKG_RELEASE to 1

schuettecarsten: Fixed and force-pushed for better commit message.

schuettecarsten:
Had to force-push again because of missing "Signed-off-by" in commit messages

1715173329:
Cc maintainer @PolynomialDivision

PolynomialDivision:
Thanks!

PR #22090, opened 2023-09-11
openvpn: fix recursive dependency error

AuthorReflex:
Maintainer: @mkrkn 
Compile tested: ramips/mt7620, ramips/mt7621
Run tested: Xiaomi mi router 3 pro, Xiaomi mi-mini, Xiaomi miwifi-r3

When updating the openvpn version, a dependency on the ovpn-dco module was added. A clean build throws a recursive dependency error. So, the recursive dependency will be removed for now


AuthorReflex:
PR #22099 is best solution. Close.

PR #22095, opened 2023-09-13
rust: Move RUST_ARCH_DEPENDS into rust-values.mk

jefferyto:
Maintainer: @lu-zero
Compile tested: armsr-armv7, 2023-09-09 snapshot sdk
Run tested: N/A

Description:
This allows other packages access to `RUST_ARCH_DEPENDS` by including rust-values.mk, instead of rust-package.mk which also sets `Build/Compile`.


lu-zero:
Sounds fine

PR #22096, opened 2023-09-13
[22.03] nextdns: Update to version 1.40.0

rs:
Update to version 1.40.0


rs:
We just pushed a 1.40.1

PR #22120, opened 2023-09-16
Update ariang from 1.3.2 to 1.3.6

ArielHeleneto:

Description:

### DIFF
--- a/net/ariang/Makefile
+++ b/net/ariang/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ariang
-PKG_VERSION:=1.3.2
-PKG_RELEASE:=1
+PKG_VERSION:=1.3.6
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).zip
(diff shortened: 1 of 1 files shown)


net/ariang/Makefile
```diff
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ariang
-PKG_VERSION:=1.3.2
-PKG_RELEASE:=1
+PKG_VERSION:=1.3.6
+PKG_RELEASE:=2
```
1715173329, line 5:
```suggestion
PKG_RELEASE:=1
```

This should be reset to `1`, see:
https://openwrt.org/docs/guide-developer/package-policies#versioning

1715173329 (commented):
Commit title should be `ariang: Update to 1.3.6`
And it would be nice to include Changelog in commit message.

Cc maintainer @Ansuel

net/ariang/Makefile
ArielHeleneto: fixing

ArielHeleneto: @1715173329 something went wrong when sysncing from openwrt and I need to open an new pull #22125 22125

PR #22125, opened 2023-09-17
ariang: Update from 1.3.2 to 1.3.6

ArielHeleneto:
Maintainer: @Ansuel 
Compile tested: AXT1800 (an qualcommax/ipq60xx) and x86
Run tested: AXT1800 (an qualcommax/ipq60xx) and x86

Description:

    Fix the problem that cannot create new task after clearing checksum option
    Other bug fix and user interface optimization

    Support overriding aria2 global settings with empty content when create new download task (#712)
    Other bug fix and user interface optimization

    Update Traditional Chinese translation (#705, thx @ChiaYen-Kan)
    Other bug fix and user interface optimization

    Add check-integrity to task settings tab (#693, thx @raytrap)
    Fix a spelling mistake (#696, thx @rusq)
   
    Some issues occuered and I have to re open a new Pull (old one is #22120 )

### DIFF
--- a/net/ariang/Makefile
+++ b/net/ariang/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ariang
-PKG_VERSION:=1.3.2
+PKG_VERSION:=1.3.6
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).zip
 PKG_SOURCE_URL:=https://github.com/mayswind/AriaNg/releases/download/$(PKG_VERSION)
-PKG_HASH:=2186dacf57c9d1650e00084c0454f2227e910f3203d89c6190f547b40cac7243
+PKG_HASH:=2d36e1a39d95867b8e0cdb3cde96d04d40117bd37e8742d639da92496e07cc7b
 UNPACK_CMD=unzip -q -d $(1) $(DL_DIR)/$(PKG_SOURCE)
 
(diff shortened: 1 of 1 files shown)


BKPepe:
CI is failing for you, can you check it, why it fails for you and also fill the PR template?

ArielHeleneto:
> CI is failing for you, can you check it, why it fails for you and also fill the PR template?

Oh hullo I tried to fix it and could you please approval the CI?

It may be failed again due to my lack experience for updating a package. This is my first time I think.

Strongly thank you.

ArielHeleneto:
seems that CI works fine and risc-v luajit problems has no connection to my pr

BKPepe:
1st commit looks good now.

ArielHeleneto:
> 1st commit looks good now. 2nd commit:
> 
>     * nitpick: Commit subject should not contain the dot at the end
>       Why it is required? Can you describe such change in the commit description? It is not enough to say just "include two more png files".

fix commit message and add detailed message. (those png is added because of the update to 1.3.6)

BKPepe:
I am thinking about this. Since those two PNGs were introduced in 1.3.6, then it should be only in one commit instead of two because if you decide to checkout or have only one commit, then it will not work, because of missing png files, right?

ArielHeleneto:
> I am thinking about this. Since those two PNGs were introduced in 1.3.6, then it should be only in one commit instead of two because if you decide to checkout or have only one commit, then it will not work, because of missing png files, right?

fixed

1715173329 (approved):
The build errors seem irrelevant, so LGTM.

PR #22130, opened 2023-09-17
dockerd: busybox compatibility

oskarirauta:
build hosts with busybox fail with long options on rm command. Short versions are supported by all, so this makes it script compatible with busybox hosts as well.

Maintainer: Gerard Ryan / @G-M0N3Y-2503 


G-M0N3Y-2503:
Thanks, LGTM!

Issue #22133, opened 2023-09-18
ripgrep: Failing to build for x86-generic/x86-legacy/x86-geode


jefferyto:
Maintainer: @lu-zero
Environment: x86-geode/x86-generic/x86-legacy, snapshot sdks

Description:

Faillogs:
* https://downloads.openwrt.org/snapshots/faillogs/i386_pentium-mmx/packages/ripgrep/compile.txt
* https://downloads.openwrt.org/snapshots/faillogs/i386_pentium4/packages/ripgrep/compile.txt

Error message:

```
[...]

PR #22134, opened 2023-09-18
rust: Set rustflags by environment variable

jefferyto:
Maintainer: @lu-zero
Compile tested: x86-generic, 2023-09-09 snapshot sdk
Run tested: x86-generic (qemu, ran `ripgrep --version`), 2023-09-09 snapshot

Description:
In order to use `$(RUSTC_TARGET_ARCH)` in `HOST_BUILD_DIR`, the line to include rust-values.mk in the Rust makefile was moved in f489e019ac4a15e974518d9928ef913a157d135a, causing it to be included before package.mk is included.

This had the side effect of preventing "-lssp_nonshared" from being added to `RUSTC_LDFLAGS`, because `PKG_SSP` is indirectly set by package.mk (package.mk includes hardening.mk, hardening.mk sets `PKG_SSP`).

There is a deeper issue; it is the Rust package's `PKG_SSP` value that causes `RUSTC_LDFLAGS` to be set and written to the Cargo config file. For packages that use Rust to build, their `PKG_SSP` value does not affect the linker flag.

This sets rustflags with the `RUSTFLAGS` environment variable, instead of writing the value to the Cargo config file, allowing `PKG_SSP` from the package being built to be used and for the package being built to modify the rustflags used.

This also:

* Fix "-lssp_nonshared" being added to `TARGET_CFLAGS` instead of `RUSTC_LDFLAGS`, when `CONFIG_PKG_CC_STACKPROTECTOR_STRONG` is set.

* Remove the use of `$(RUSTC_TARGET_ARCH)` in `HOST_BUILD_DIR` and move the include line for rust-values.mk back to after package.mk.

  Since the host build directory was moved under the target build directory in efdbac38dc8b649ca26b49fac27abeb5cf76cd28, it is no longer necessary to separate build directories with `RUSTC_TARGET_ARCH`; `$(BUILD_DIR)` already separates build directories by target.

* Add `BUILDONLY:=1`, as the Rust package does not build a target package.

* Install the Cargo config file as "config.toml" instead of "config", as this is the [preferred form][1].

* Rename `RUST_CFLAGS` to `RUSTC_CFLAGS` and `CONFIG_HOST_SUFFIX` to `RUSTC_HOST_SUFFIX`, for consistency.

* Allow `CARGO_VARS` to be set before rust-values.mk is included.

[1]: https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure

Fixes: 83785a7ce016 ("rust-lang: Add the rust language support")
Fixes: https://github.com/openwrt/packages/issues/22133

### DIFF
--- a/lang/rust/Makefile
+++ b/lang/rust/Makefile
@@ -3,16 +3,15 @@
 # Copyright (C) 2023 Luca Barbato and Donald Hoskins
 
 include $(TOPDIR)/rules.mk
-include ./rust-values.mk
 
 PKG_NAME:=rust
 PKG_VERSION:=1.72.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=rustc-$(PKG_VERSION)-src.tar.gz
 PKG_SOURCE_URL:=https://static.rust-lang.org/dist/
 PKG_HASH:=ea9d61bbb51d76b6ea681156f69f0e0596b59722f04414b01c6e100b4b5be3a1
-HOST_BUILD_DIR:=$(BUILD_DIR)/host/rust-$(RUSTC_TARGET_ARCH)/rustc-$(PKG_VERSION)-src
+HOST_BUILD_DIR:=$(BUILD_DIR)/host/rustc-$(PKG_VERSION)-src
 
 PKG_MAINTAINER:=Luca Barbato <lu_zero@luminem.org>
 PKG_LICENSE:=Apache-2.0 MIT
@@ -23,6 +22,7 @@ PKG_HOST_ONLY:=1
 
 include $(INCLUDE_DIR)/host-build.mk
 include $(INCLUDE_DIR)/package.mk
+include ./rust-values.mk
 
 define Package/rust
   SECTION:=lang
@@ -31,6 +31,7 @@ define Package/rust
   TITLE:=Rust Programming Language Compiler
   URL:=https://www.rust-lang.org/
   DEPENDS:=$(RUST_ARCH_DEPENDS)
+  BUILDONLY:=1
 endef
 
 define Package/rust/description
@@ -95,8 +96,7 @@ define Host/Install
 		\
 		sed -e 's|@RUSTC_TARGET_ARCH@|$(RUSTC_TARGET_ARCH)|g' \
 		    -e 's|@TARGET_CC_NOCACHE@|$(TARGET_CC_NOCACHE)|g' \
-		    -e 's|@RUSTC_LDFLAGS@|$(RUSTC_LDFLAGS)|g' \
-			$(CURDIR)/files/cargo-config > $(CARGO_HOME)/config ; \
+			$(CURDIR)/files/cargo-config > $(CARGO_HOME)/config.toml ; \
 	)
 endef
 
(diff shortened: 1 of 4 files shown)


lang/rust/Makefile
```diff
@@ -3,16 +3,15 @@
...
 PKG_NAME:=rust
 PKG_VERSION:=1.72.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=rustc-$(PKG_VERSION)-src.tar.gz
 PKG_SOURCE_URL:=https://static.rust-lang.org/dist/
 PKG_HASH:=ea9d61bbb51d76b6ea681156f69f0e0596b59722f04414b01c6e100b4b5be3a1
-HOST_BUILD_DIR:=$(BUILD_DIR)/host/rust-$(RUSTC_TARGET_ARCH)/rustc-$(PKG_VERSION)-src
+HOST_BUILD_DIR:=$(BUILD_DIR)/host/rustc-$(PKG_VERSION)-src
```
lu-zero, line 14: this seems unrelated

jefferyto: I think it is better to include rust-values.mk after package.mk for the Rust package itself, in case code in rust-values.mk needs to access `PKG_*` variables to build rustc in the future. To do this it is necessary to remove the reference to `$(RUSTC_TARGET_ARCH)` here, because this variable would not be defined yet.

lu-zero: You need to have that otherwise you break some people use-case :(

jefferyto: When @1715173329 changed the build dir from `$(BUILD_DIR_HOST)` to `$(BUILD_DIR)/host` in efdbac3, this had the same effect as including `$(RUSTC_TARGET_ARCH)` in the path. `$(BUILD_DIR)` is different for each target.

lu-zero: Then it seems good :) Thank you for doublechecking.

1715173329:
I'm not familiar with rust buildsystem tbh
Currently we just somehow manage to make it "work" and it does need to be improved

lu-zero:
The CARGO_HOME thing mainly to avoid an annoying problem with rpath assumptions that caused the `cargo` binary to not find all the host libraries.

The rest is just a side effect, most of the config.toml can be replaced with a bit more verbose env vars indeed.

Potentially if we want to use the `cargo vendor` feature to freeze the dependencies and make the build phase not use anything else we'll need a place to store the sources, but it is quite a bit of work.

jefferyto:
> The CARGO_HOME thing is mainly to avoid an annoying problem with rpath assumptions that caused the `cargo` binary to not find all the host libraries.

How can I reproduce this issue?

lu-zero:
Replace the CARGO_HOME with the normal host path and then try to call `cargo`, had been a while so I'm not 100% sure.

lu-zero:
ldd the binary just to make sure it is picking the right openssl. If it does, then probably we could rework it that way safely :)

jefferyto:
Cargo doesn't link to openssl:

```
$ ldd cargo
        linux-vdso.so.1 (0x00007ffff4948000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fb8ba21f000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fb8b8b17000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb8b8800000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fb8ba267000)
```

lu-zero:
Used to link to libgit (and thus also curl) and that was bringing in libssl and quite a bit of pain.

PR #22153, opened 2023-09-20
python3-networkx: Update to 3.1, rename source package

jefferyto:
Maintainer: @flyn-org
Compile tested: armsr-armv7, 2023-09-16 snapshot sdk
Run tested: armsr-armv7 (qemu), 2023-09-16 snapshot

Description:
This renames the source package to python-networkx to match other Python packages.

This also updates the list of dependencies.

### DIFF
--- a/lang/python/python-networkx/Makefile
+++ b/lang/python/python-networkx/Makefile
@@ -5,12 +5,12 @@
 
 include $(TOPDIR)/rules.mk
 
-PKG_NAME:=python3-networkx
-PKG_VERSION:=2.8.8
+PKG_NAME:=python-networkx
+PKG_VERSION:=3.1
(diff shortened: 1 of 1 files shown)


flyn-org:
This looks like a major-number change. Does your setools 4.4.3 update work with this? If so, then I recommend merging.

jefferyto:
I have run `sedta --help` and `seinfoflow --help` but not tested any actual functionality using networkx. There has been at least one change upstream for the major version change (https://github.com/SELinuxProject/setools/commit/fea650045d6ab8359ec583a86977dec784d16345) and no bug reports upstream regarding networkx, so I would assume it works.

flyn-org:
Upstream commit looks good. Recommend merge. Thank you.

PR #22159, opened 2023-09-20
sshtunnel: support Dropbear

stokito:
**Maintainer:** @nunojpg

**Compile tested:** OpenWrt 23
**Run tested:** tested in VirtualBox and Turris

**Description:**
This is the same PR as was in the #21263. 
I just removed the branch accidentally and now have to recreate the PR.

The only difference is that I changed dependency line: now it will depend ether on dbclient or openssh-client.
I copied the depends line from pppossh package but improved it to be more precise and depend on the dbclient and not just on the dropbear package.
The PR also contains update for the pppossh with the same depends change so @yousong please have a look.

### DIFF
--- a/net/sshtunnel/Makefile
+++ b/net/sshtunnel/Makefile
@@ -8,8 +8,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=sshtunnel
-PKG_VERSION:=4
-PKG_RELEASE:=5
+PKG_VERSION:=5
+PKG_RELEASE:=1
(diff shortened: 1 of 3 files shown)


nunojpg:
Hi @stokito can you move the pppossh change into another PR?

stokito:
You can merge if we have a commit to other app?

stokito:
Ok, I dropped the last commit

nunojpg:
Thanks for all the effort!

stokito:
@nunojpg thank you! Could you also check the https://github.com/openwrt/luci/pull/6424

PR #22160, opened 2023-09-20
vim: fix renamed defaults config file

leonarddevries:
Maintainer: @ratkaj 
Compile tested: arm_cortex-a7_neon-vfpv4, bcm27xx/bcm2709, r23389-5deed175a5 (23.05.0-rc3)
Run tested: Same arch, platform and openwrt version.

Before this fix:
```
root@router:/usr/share/vim# vim
E1187: Failed to source defaults.vim
Press ENTER or type command to continue
```
After this fix the error disappeared.

Description:
Since vim version 9 the config file has been renamed from vimrc to defaults.vim. The syntax also has been changed however this old version remains the default, therefore the config doesn't need a change.


BKPepe:
You are fixing it only in OpenWrt 23.05 branch. What about master branch? There is same issue.

leonarddevries:
Ah ok, I figured this would only matter for the 23.05 branch, but I see it's also on master. Should I create a new pull request with the changes based on master?

BKPepe:
Yes.

Dne čt 21. 9. 2023 9:20 uživatel Leonard de Vries ***@***.***>
napsal:

PR #22161, opened 2023-09-20
openvswitch: backport patch to fix build with recent groff

neocturne:
Maintainer: @yousong 
Compile tested: x86-64, latest openwrt-23.05
Run tested: No run test needed, this fixes the build of a manpage that isn't even installed

Description:

On build hosts with groff 1.23 or newer, the openvswitch build fails:

    troff:vswitchd/ovs-vswitchd.8:1298: warning: cannot select font 'CW'
    make[1]: *** [Makefile:6761: manpage-check] Error 1

Backport the upstream commit to fix the issue.


neocturne:
Hmm, should we increment `PKG_RELEASE`? The commit shouldn't have an effect on the built packages at all, so maybe it makes sense to leave it unmodified?

PR #22166, opened 2023-09-21
[21.02] libdrm: install all headers (cherry-picking)

bl4kraven:
Hi, 

Unbreak using libdrm program compile in 21.02. The error is  missing <drm.h>，so i just cherry-picking the same fix commit cb8c0ba for libdrm as in master.

### DIFF
--- a/libs/libdrm/Makefile
+++ b/libs/libdrm/Makefile
@@ -58,7 +58,7 @@ MESON_ARGS += \
 
 define Build/InstallDev
 	$(INSTALL_DIR) $(1)/usr/include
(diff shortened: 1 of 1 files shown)


BKPepe:
In the commit description, there is missing any indication that it was cherry-picked. That's can be done manually (you can look how it is done in stable branches) or you can use ``git cherry-pick -x hash``.  Then you need to force push it.

bl4kraven:
OK, done. I force push it.

By the way, the CI for 21.02 maybe some problem, some download URL is missing and report error.

BKPepe:
That's expected, because OpenWrt 21.02 is already end of life.

Dne čt 21. 9. 2023 6:14 uživatel Leo Chung ***@***.***>
napsal:

PR #22178, opened 2023-09-22
[openwrt-23.05] golang: Update to 1.21.1

jefferyto:
Maintainer: me
Compile tested: none (cherry picked from #21827 and #22083)
Run tested: none

Description:
At this point only cloudflared and ooniprobe fail to build with Go 1.21. It seems to me the benefit of having 1.21 in 23.05 outweighs not having these two packages.


BKPepe:
> It seems to me the benefit of having 1.21 in 23.05 outweighs not having these two packages.

Yeah, I do have the same thoughts as you do! Let's goooo.
However, we will need to backport (I can do that by cherry-picks) those packages from master branch, which was updated due to Go 1.21 to 23.05.

BKPepe:
Okay, I think I cherry-picked all the necessary updates for Go packages, which we do have from the ``master`` branch to ``openwrt-23.05`` branch.

BKPepe:
Yes, you are right! I will cherry-pick it.

jefferyto:
Thanks for your help!

PR #22180, opened 2023-09-22
python-cryptography: Update to 41.0.4, add new dependencies

jefferyto:
Maintainer: me
Compile tested: armsr-armv7/armsr-armv8/malta-be/x86-generic/x86-64, 2023-09-16 snapshot sdks
Run tested: armsr-armv7/armsr-armv8/malta-be/x86-generic/x86-64 (qemu), 2023-09-16 snapshot

Description:


lang/python/python-cryptography/Makefile
```diff
@@ -32,18 +32,18 @@ define Package/python3-cryptography
   URL:=https://github.com/pyca/cryptography
   DEPENDS:= \
       +libopenssl \
+      +libopenssl-legacy \
```
pprindeville, line 35: Is this really necessary?  I'd prefer to make due without... they've been marked for deprecation for a reason.

pprindeville (changes requested):
Drop legacy algorithms.

lang/python/python-cryptography/Makefile
jefferyto: cryptography package deliberately continues to support legacy algorithms and requires an [environment variable](https://cryptography.io/en/latest/openssl/#legacy-provider-in-openssl-3-x) to be set to bypass this.

I don't think it would be a good idea to patch this out because other programs/packages that use cryptography will also expect legacy algorithms to be available.

pprindeville: I guess it will get retired upstream in due time.

pprindeville (approved):

PR #22184, opened 2023-09-22
uwsgi: add experimental pcre2 patch and drop pcre

Ansuel:
Add experimental pcre2 patch (https://github.com/unbit/uwsgi/pull/2543) and drop pcre in favor of pcre2 library.


Ansuel:
@BKPepe for reference uwsgi repo is a bit problematic with patching... the tagged release is not sourced out of master branch but they use funny commit... So we have to use manual patching anyway. And I don't think they will make a new release in the next few days.

BKPepe:
To your OP, I added reference for the PR from which is PCRE2 support. Anyway, this has green light from me!

PR #22185, opened 2023-09-22
nginx-util: move to pcre2

Ansuel:
Convert to pcre2 library as pcre is EOL. No functional change intended.

@peter-stadler Can you check this? Not strictly pcre2 usage but at least c++ format :D

### DIFF
--- a/net/nginx-util/Makefile
+++ b/net/nginx-util/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nginx-util
 PKG_VERSION:=1.6
-PKG_RELEASE:=19
+PKG_RELEASE:=20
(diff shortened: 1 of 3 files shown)


neheb:
Why does this use PCRE instead of std::regex?

Ansuel:
Different syntax and from what I can read pcre2 is more stable and
robust/faster than the integrated one.

Il Dom 24 Set 2023, 03:40 Rosen Penev ***@***.***> ha scritto:

neheb:
Doesn’t make much sense to use it over std::regex IMO

peter-stadler:
Looks good, although I cannot test it right now. From my side we can merge it like thus, or switch to using std::regex only ...

I looked into switching to pcre2 a while ago, but was stuck at getting to run the tests again. And then I forgot about it, while there were (or are ongoing) efforts for using a new dynamic config for nginx (replacing nginx-util) ...

Thank you very much :-)

Issue #22170, opened 2023-09-21
ffmpeg: build failure with binutils 2.41


oskarirauta:
Maintainer: @thess
Environment: x86_64, latest git

Description:

Build failed, following error repeated numerous times:

```
Error: operand type mismatch for `shr'
```

I found a patch for it from here: [https://www.linuxquestions.org/questions/slackware-14/regression-on-current-with-ffmpeg-4175727691/](https://www.linuxquestions.org/questions/slackware-14/regression-on-current-with-ffmpeg-4175727691/) file: [https://git.videolan.org/?p=ffmpeg.git;a=patch;h=effadce6c756247ea8bae32dc13bb3e6f464f0eb](https://git.videolan.org/?p=ffmpeg.git;a=patch;h=effadce6c756247ea8bae32dc13bb3e6f464f0eb)

I can verify that this patch fixes the issue.

BKPepe:
I can not see that binutils 2.41 is in the master branch. I double-checked it and in the master branch, there is 2.40. It seems that you are using the latest git also with something else.

You can see that the latest ffmpeg update was successfully compiled on all targets: https://github.com/openwrt/packages/pull/22128

In this case as you seems to be using something else then the latest git, I suggest to create pull request  with that patch or wait until a new version ffmpeg is released. Also, you can usd binutils 2.40

oskarirauta:
what? I am using plain version.
Check here: https://github.com/openwrt/openwrt/tree/main/toolchain/binutils

There's 2.41

BKPepe:
Not used as default version - https://github.com/openwrt/openwrt/blob/587f94eb1adb2eb65ff7b87a78c4641d2d15dc2c/toolchain/binutils/Config.in#L5.

thess:
@oskarirauta - Thanks for flagging this issue. Apparently binutils introduced this change in:
https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=c34d1cc9200ae24dc7572aaf77d80276c0490e9b
and, the corresponding trac item: https://trac.ffmpeg.org/ticket/10405 documents a patch for mathops.h in libavcodec.

ffmpeg (5.1 branch) fixed on 16-Jul-2023 by:
https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/25cd95a9dc3510c3cc0d7aad6f9d83f6a1078c7e
[...]

PR #22191, opened 2023-09-23
ffmpeg: fix assembling with binutils v2.41

oskarirauta:
fix addresses issue when building ffmpeg with binutils v2.41 that is documented [here](https://trac.ffmpeg.org/ticket/10405)

Fixes: #22170

Remarks:
Patch is merged to ffmpeg upstream and will be included in next release of ffmpeg, so patch applies only to current version and is removed when package is updated to next version of ffmpeg.

Upstream patch refreshed.

Maintainer: @thess

### DIFF
--- a/multimedia/ffmpeg/Makefile
+++ b/multimedia/ffmpeg/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ffmpeg
 PKG_VERSION:=5.1.3
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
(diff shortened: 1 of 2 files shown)


multimedia/ffmpeg/060-fix-assembling-with-binutils-2.41.patch
```diff
@@ -0,0 +1,72 @@
+From 25cd95a9dc3510c3cc0d7aad6f9d83f6a1078c7e Mon Sep 17 00:00:00 2001
```
BKPepe, line 1: Drop this one. Patch under patches folder is just fine

oskarirauta: Thanks. Just pushed. Let's see...

BKPepe (approved):
LGTM

thess:
Thanks for the patch and figuring out a clean PR.

Issue #22192, opened 2023-09-23
new `luajit2` package Lead to Recursive loop!!!


zxlhhyccc:
````
tmp/.config-package.in:106579:error: recursive dependency detected!
tmp/.config-package.in:106579:	symbol PACKAGE_dnsdist depends on PACKAGE_dnsdist
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"

tmp/.config-package.in:106611:error: recursive dependency detected!
tmp/.config-package.in:106611:	symbol PACKAGE_dnsdist-full depends on PACKAGE_dnsdist-full
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"

tmp/.config-package.in:105249:error: recursive dependency detected!
tmp/.config-package.in:105249:	symbol PACKAGE_snort depends on PACKAGE_snort
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"

tmp/.config-package.in:105280:error: recursive dependency detected!
tmp/.config-package.in:105280:	symbol PACKAGE_snort3 depends on PACKAGE_snort3
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"

tmp/.config-package.in:107124:error: recursive dependency detected!
tmp/.config-package.in:107124:	symbol PACKAGE_knot-resolver depends on PACKAGE_knot-resolver
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"

tmp/.config-package.in:105113:error: recursive dependency detected!
tmp/.config-package.in:105113:	symbol PACKAGE_openappid depends on PACKAGE_openappid
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"
``````

BKPepe:
You dont need to spam it everywhere. You posted it already on three places:
1. Pull request, where it was merged
2. Commit, which was merged
and now here.

BKPepe:
Next time, please fill the proper issue template.

zxlhhyccc:
> Next time, please fill the proper issue template.

OK!

ptpt52:
just to say, same issue for my build.

Ansuel:
Fixed

PR #22195, opened 2023-09-23
treewide: add HAS_LUAJIT_ARCH dependency to luajit user

Ansuel:
Add HAS_LUAJIT_ARCH dependency to any user of luajit to fix circular
dependency limitation.

Fixes: https://github.com/openwrt/packages/issues/22192

### DIFF
--- a/lang/luajit/Makefile
+++ b/lang/luajit/Makefile
@@ -24,12 +24,19 @@ define Package/luajit
  CATEGORY:=Languages
  TITLE:=LuaJIT
  URL:=https://www.luajit.org
- DEPENDS:=@(i386||x86_64||arm||armeb||aarch64||powerpc||mips||mipsel||mips64)
+ DEPENDS:=@HAS_LUAJIT_ARCH
 endef
 
 define Package/luajit/description
  LuaJIT is a Just-In-Time (JIT) compiler for the Lua programming language. *** Requires GCC Multilib on host system to build! ***
 endef
+
+define Package/luajit/config
+config HAS_LUAJIT_ARCH
+	bool
+	default y if i386||x86_64||arm||armeb||aarch64||powerpc||mips||mipsel||mips64
+endef
+
 ifeq ($(HOST_ARCH),$(filter $(HOST_ARCH), x86_64 mips64))
   ifeq ($(CONFIG_ARCH_64BIT),)
     HOST_BITS := -m32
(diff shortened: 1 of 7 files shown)


BKPepe:
I don't like it. You can not have under one Makefile two different sources for two different packages. There should be better solution than this. We should keep it as it was. It means:
- Luajit has its own Makefile
- Luajit2 has its own Makefile

Ansuel:
I know and I don't like it too but this is not core and we have no other
way to add a flag in a centralized way.

Tried to find a solution with nbd and this is the least "bad" one.

Alternative is ignore consistency and declare flag in one of the package
and reference then in the other.

Il Sab 23 Set 2023, 20:07 Josef Schlehofer ***@***.***> ha
scritto:

dhewg:
Pardon my sarcasm, but why do we _add_ lua stuff for an 12 years old interpreter version?
So we can have faster insecure processes, in a web server no less?

Ansuel:
Lua and luajit are very different thing.

dhewg:
Okay.

FYI: I just checked on my debian system, and most packages that require 
this have "depends on libluajit | libluajit2" and some have "depends on 
libluajit2 | libluajit" (where the first one gets picked if neither is 
installed).

So they're binary compatible, and we could just drop one and be done 
with it?

BKPepe:
> "depends on libluajit2 | libluajit" (where the first one gets picked if neither is installed).

This is why I assigned the PR to myself to do it exactly as you are saying.

> So they're binary compatible, and we could just drop one and be done with it?

Hmm, no. Luajit2 is fork by Luajit repository, which was inactive for some while and now it is getting some attention. Both repositories these days include separate changes. E.g. the latest master in Luajit dropped support for PowerPC, thus it does not support Knot Resolver anymore on Turris 1.x routers, which is not good. However, the latest master in Luajit2 and as well the stable (well beta) of Luajit includes that support.

Ansuel:
They are not the same tho. Luajit2 is luajit + openresty fixes and
optimization + additional thing. Openresty try to stay in sync but they
don't provide patch so it might have more changes than expected. Also
Debian system have both right?

About solving the concerns for this pr I think I will just split again and
introduce the flag in luajit ad be done with it

Il Dom 24 Set 2023, 10:00 dhewg ***@***.***> ha scritto:

BKPepe:
>  Also Debian system have both right?

Many GNU/Linux distributions have both.

Ansuel:
@BKPepe well this is the way to fix this... with keeping the package split. A comment :(

Ansuel:
@BKPepe sorry :( I tought it was good as I correctly split the thing. Also there was the dependency problem.

PR #22205, opened 2023-09-25
openthread-br: switch to avahi

neheb:
mDNSResponder links to dns_sd, which is also provided by avahi and does not include needed functions. Just avoid this mess.

Maintainer: @stintel

### DIFF
--- a/net/openthread-br/Makefile
+++ b/net/openthread-br/Makefile
@@ -40,8 +40,7 @@ define Package/openthread-br
 	+libstdcpp \
 	+libubox \
 	+libubus \
-	+mdnsd \
-	+mdnsresponder
+	+libavahi-client
 endef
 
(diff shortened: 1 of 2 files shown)


neheb:
https://downloads.openwrt.org/snapshots/faillogs/mips64_octeonplus/packages/openthread-br/compile.txt

stintel:
NAK. The original PR mentions why avahi was not chosen.

neheb:
Aaand this is still broken on the buildbot. You want a new PR fixing just that?

neheb:
@stintel alright so the reason this fails is because both avahi and mdnsresponder install libdns_sd.so. Currently there are only two packages (including this one) that use it. The other one I want to transition to avahi as it's better maintained, which leaves only this one. Can you look into getting avahi working again?

PR #22206, opened 2023-09-25
libdmapsharing: update to 3.9.13

neheb:
Maintainer: @flyn-org


flyn-org:
What led you to move to the libavahi-glib backend? I thought the mDNSresponder backend used less storage space. (Granted, libdmapsharing itself pulls in glib2 and a lot of other things.) Was there a clear reason for the change?

I think we should separate the version update from the mDNS backend change.

neheb:
There are only two packages that use the former. In this case, there’s probably a slight increase. Avahi is also better maintained. Future versions of mdnsresponder don’t compile on non macOS.

PR #22211, opened 2023-09-25
adguardhome: wait for interfaces to be up at boot

### LINKED ISSUE #21868: adguardhome 0.107.21-1 doesn't wait for ip6 address and fails to start after reboot
Environment: 
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='22.03.5'
DISTRIB_REVISION='r20134-5f15225c1e'
DISTRIB_TARGET='rockchip/armv8'
DISTRIB_ARCH='aarch64_generic'

Description:

adguardhome doesn't start after reboot: `[fatal] couldn't start forwarding DNS server: listening to udp socket: list en udp [fda3:ca1f:2ac2::1]:53: bind: cannot assign requested address`

Moving adguard init from S25 to S97 doesn't help. Adding respawn 300 3 3 doesn't help giving same result several times. Starting manually after logging in with ssh works. Above address is bridge interface, does it not get created fast enough? It's using static ip4 but I don't see any options for ip6 when editing it in web ui.

alaviss:
Maintainer: @dobo90 
Compile tested: n/a
Run tested: @iljah and @mangkoran 

Description:

This should allow the service to be activated even earlier during the boot process and also avoids race condition against network.

~~I've not tested this change yet since I'm not affected by this issue.
@iljah if you can compile your own packages, please test this.~~

Fixes #21868.

### DIFF
--- a/net/adguardhome/files/adguardhome.init
+++ b/net/adguardhome/files/adguardhome.init
@@ -4,12 +4,22 @@ PROG=/usr/bin/AdGuardHome
 
 USE_PROCD=1
 
-# starts just after network starts to avoid some network race conditions
-START=25
+# matches dnsmasq
+START=19
 # stops before networking stops
 STOP=89
 
+boot() {
+  adguardhome_boot=1
+  start "$@"
+}
+
 start_service() {
+  if [ -n "$adguardhome_boot" ]; then
+    # Do not start yet, wait for triggers
+    return 0
+  fi
+
   config_load adguardhome
   config_get WORK_DIR config workdir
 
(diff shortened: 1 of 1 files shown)


alaviss:
You can just copy the edited init script and put it in /etc/init.d/adguardhome and see if it works.

Remember to backup the old one just in case.

iljah:
Oh right. Did that but didn't seem to help, same result. Running `service adguardhome restart` helped again.

alaviss:
It seems like your network interface got up really slow then.

Can you try adjusting the `3000` to a bigger number and see if that helps? `3000` here is in milliseconds so it's 3 seconds.

alaviss:
Updated the PR to use 5000 for the delay, thanks for testing!

mangkoran:
I just would like to inform that I also encountered AGH fail to start on boot. I tried the updated init script from this PR and now AGH start on boot!

Is there anything left before it could be merged?

stangri:
@alaviss -- if you could update original commit/PR comment with the run-test confirmation by @iljah and @mangkoran, that would take care of one of the requirements. Please mention/tag me when it's done.

I'd feel confident merging this without compile-testing as it's a trivial change and without blessing from @dobo90 as it's been almost 6 months.

alaviss:
@stangri I updated the PR description

alaviss:
It seems to me that some users on the forums also found the PR to work for them: https://forum.openwrt.org/t/adguardhome-dont-start-at-boot/185518

PR #22228, opened 2023-09-27
atftp: bump to release 0.8.0 + PCRE2

Ansuel:
Bump atftp to release 0.8.0 and move to PCRE2.

Maintainer: @dddaniel

---

@BKPepe


BKPepe:
Are we using it anyhow? I was rather thinking about dropping this package.

Ansuel:
We should check download stats for this

BKPepe:
My idea is to drop it, people will be asking why it was removed and someone
from them will bring it back and become a new maintainer.

Dne st 27. 9. 2023 19:16 uživatel Christian Marangi <
***@***.***> napsal:

BKPepe:
In your first commit, there is a typo in autoreconf.

> Bump to release 0.8.0. Autorecong is now needed to correctly compile the
package.

I checked the [Git history](https://github.com/openwrt/packages/commits/master/net/atftp/Makefile), and recent updates were done only by @neheb, thus it is a question if we want to keep this package in our repository. It seems like @dddaniel is no longer interested in it.

On the other hand, I need to say that this update seems smooth.

This is rather question, if we prefer quantity over quality... 🤔 

Do we have some other statistics about installed packages than this: https://downloads.openwrt.org/stats/awstats.downloads.openwrt.org.allextra2.html ?

BKPepe:
What about this https://github.com/BKPepe/packages/commit/843459ab47bf25d2bfc546c39211c100e0d817fa ?

Ansuel:
For me the package looks simple enough so it can stay. And won't probably
receive any other update soo maybe nothing to maintain

Il Mar 3 Ott 2023, 15:21 Josef Schlehofer ***@***.***> ha
scritto:

BKPepe:
Hmmm, let's see.

PR #22229, opened 2023-09-27
openthread-br: fix actually building

neheb:
Maintainer: @stintel

### DIFF
--- a/net/openthread-br/Makefile
+++ b/net/openthread-br/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=openthread-br
 PKG_SOURCE_DATE:=2023-08-01
 PKG_SOURCE_VERSION:=1738d8cd8b42106c2ef1262fbbac2f06beab83ba
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=https://github.com/openthread/ot-br-posix.git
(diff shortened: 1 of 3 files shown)


stintel:
I don't maintain mdnsresponder...

net/openthread-br/Makefile
```diff
@@ -94,10 +96,8 @@ endef
 
 define Package/openthread-br/install
 	$(INSTALL_DIR) \
-		$(1)/etc/init.d \
```
stintel, line 97: Why?

net/openthread-br/Makefile
```diff
@@ -94,10 +96,8 @@ endef
...
 		$(1)/lib/netifd/proto \
-		$(1)/usr/sbin \
-		$(1)/var/lib/thread
```
stintel, line 100: Why?

stintel (changes requested):

neheb:
Nothing is being installed to those locations. Make error on my end.

net/openthread-br/Makefile
stintel: Right. Forgot I moved things to a netifd proto handler. This one can be removed indeed.

net/openthread-br/Makefile
stintel: But this one should be kept. It's where otbr-agent writes its data file.

neheb: that's in conffiles. this change is in INSTALL_DIR. I don't think they're related.

stintel: Having it in conffiles is to make sure it is kept during sysupgrade. Afaik it doesn't ensure the dir is there.

neheb: OK added back after testing.

1715173329: Everything in `/var` can't be really kept as it's stored in the memory by default. Everything inside it will be lost after reboot.

https://github.com/openwrt/packages/pull/22252#issuecomment-1761024651
https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a53/base/base-files/compile.txt

Now we have a race condition issue because of it. In `base-files`, `/var` is soft linked to `/tmp`, but this Makefile creates a actual directory.

stintel: Please don't necro old/closed issues/PRs. If this is currently a problem, open a new issue and mention the maintainer.

neheb: In any case, this issue needs fixing.

PR #22237, opened 2023-09-28
ddns-scripts:  add new DDNS provider ipnodns.ru

IgorBai:
Here I am adding a new DDNS provider name - ipnodns.ru

I will try to add a "ipnodns.ru.json" file in the next request.

P.S. sorry for possible errors or misunderstanding: this is my first experience in github coworking.

Maintainer: me / @\<github-user> (find it by checking history of the package Makefile)
Compile tested: (put here arch, model, OpenWrt version)
Run tested: (put here arch, model, OpenWrt version, tests done)

Description:

PR #22238, opened 2023-09-28
Create ipnodns.ru.json - a new DDNS provider

IgorBai:
Here is the file required to work with ipnodns.ru
I have checked it for a number of times already.
Regards
Igor

Description:


feckert:
Please rework your [pullrequest](https://github.com/openwrt/packages/pull/22237) I will close this one because of:

- We do not needed for each change a pullrequest. The other was first opened
- Update the PKG_RELEASE of the [ddns-scripts Makefile](https://github.com/openwrt/packages/blob/master/net/ddns-scripts/Makefile#L11), se we can track the change int the package verstion.
- Change the commit header of your change to `ddns-scripts: add new DDNS provider ipnodns.ru`, so we see in the commit history of this repository what package was changed in this commit.

PR #22254, opened 2023-09-30
net-snmp: move to PCRE2 library

Ansuel:
Add upstream patch adding support for pcre2 and update dependency to require libpcre2 instead of libpcre.

--without-pcre is now needed to exclude support for pcre and only require pcre2.

Maintainer: @stintel

### DIFF
--- a/net/net-snmp/Makefile
+++ b/net/net-snmp/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=net-snmp
 PKG_VERSION:=5.9.1
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=@SF/net-snmp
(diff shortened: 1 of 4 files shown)


Ansuel:
Bumping the package is waaay too much effort and it seems there are some strange compilation error. For now we just backport. Also the current release doesn't include pcre2 support so we would have to patch the source anyway.

stintel:
```
Package libnetsnmp is missing dependencies for the following libraries:
libpcre.so.1
```

I guess I need to disable libpcre but ideally it would not pick it up at all.

Ansuel:
@stintel it's funny net-snmp requiring the library... i have to do some magic or check the backport patch.

Ansuel:
@stintel ok fixed... it's just that they made the config args a bit confusing.... patch 201 was clear and we need to set explicit support for pcre2 or pcre is still used by default. (and having funny situation with linking both library)

BKPepe:
Is it okay to cherry-pick it to OpenWrt 23.05 branch?

stintel:
> Is it okay to cherry-pick it to OpenWrt 23.05 branch?

No objections.

PR #22167, opened 2023-09-21
vim: fix renamed defaults config file

leonarddevries:
Maintainer: @ratkaj 
Compile tested: arm_cortex-a7_neon-vfpv4, bcm27xx/bcm2709, r23389-5deed175a5 (23.05.0-rc3)
Run tested: Same arch, platform and openwrt version.

Before this fix:
```
root@router:/usr/share/vim# vim
E1187: Failed to source defaults.vim
Press ENTER or type command to continue
```
After this fix the error disappeared.

Description:
Since vim version 9 the config file has been renamed from vimrc to defaults.vim. The syntax also has been changed however this old version remains the default, therefore the config doesn't need a change.

PR #22274, opened 2023-10-02
vim: add empty defaults.vim

raenye:
**Description**:
When installing, `touch /usr/share/vim/vim90/defaults.vim` to eliminate the error message
```
E1187: Failed to source defaults.vim
```
that appears every run. The problem exists for `vim` and `vim-full`, but not for `vim-fuller`.

The fix was suggested here (last comment). https://bugs.launchpad.net/ubuntu/+source/vim/+bug/1968912
See also https://bugs.debian.org/1004118.

**Maintainer**: @ratkaj
**Compile and run tested**: bcm53xx/generic Linksys EA9200 snapshot (cafa8804a4c5fd)

### DIFF
--- a/utils/vim/Makefile
+++ b/utils/vim/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=vim
 PKG_VERSION:=9.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 VIMVER:=90
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
@@ -207,6 +207,8 @@ define Package/vim/install
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/vim_tiny $(1)/usr/bin/vim
 	$(INSTALL_DIR) $(1)/usr/share/vim
 	$(INSTALL_CONF) ./files/vimrc $(1)/usr/share/vim/
+	mkdir -p $(1)/usr/share/vim/vim$(VIMVER)
+	touch $(1)/usr/share/vim/vim$(VIMVER)/defaults.vim
 endef
 
 define Package/vim-full/install
@@ -215,6 +217,8 @@ define Package/vim-full/install
 	$(INSTALL_DIR) $(1)/usr/share/vim
 	$(LN) vim $(1)/usr/bin/vimdiff
 	$(INSTALL_CONF) ./files/vimrc.full $(1)/usr/share/vim/vimrc
+	mkdir -p $(1)/usr/share/vim/vim$(VIMVER)
+	touch $(1)/usr/share/vim/vim$(VIMVER)/defaults.vim
 endef
 
 


BKPepe:
Isn't this duplicate from https://github.com/openwrt/packages/pull/22167 ?

raenye:
> Isn't this duplicate from #22167 ?

Yes it is, sorry for not searching. But I wonder whether the other solution works, since in my case vim was looking for the file in a slightly different directory.

SuperSandro2000:
I can confirm that running the following commands fixes the error

```
mkdir -p /usr/share/vim/vim90
touch /usr/share/vim/vim90/defaults.vim
```

leonarddevries:
Looking at the Makefile in the vim repository I also come to the conclusion the directory you chose is the correct one. 
See below the distilled rows of the [vim Makefile](https://github.com/vim/vim/blob/master/src/Makefile) that builds the path to the defaults file in the `/vim/vim90` directory:

```Makefile
VIMMAJOR = 9
VIMMINOR = 0
VIMRTDIR = /vim$(VIMMAJOR)$(VIMMINOR)
VIMRTLOC	= $(DATADIR)$(VIMDIR)$(VIMRTDIR)
SCRIPTLOC	= $(VIMRTLOC)
VIM_DEFAULTS_FILE = $(DESTDIR)$(SCRIPTLOC)/defaults.vim
```

In the change you propose an empty `defaults.vim` is created, but I am not sure if that would be the best thing to do. The defaults.vim [supplied by vim](https://github.com/vim/vim/blob/master/runtime/defaults.vim) has a lot of content. Also the default vimrc in this OpenWRT package has a lot of default settings. Vim starts without error now but we might miss a few good defaults?

raenye:
What you're suggesting is the approach of the other PR (which still uses wrong paths). Not sure if I should just withdraw this one.

joez:
According to the source code (src/misc1.c):

```c
    // handling $VIMRUNTIME and $VIM is below, bail out if it's another name.
    vimruntime = (STRCMP(name, "VIMRUNTIME") == 0);
    if (!vimruntime && STRCMP(name, "VIM") != 0)
        return NULL;

    /*
     * When expanding $VIMRUNTIME fails, try using $VIM/vim<version> or $VIM.
     * Don't do this when default_vimruntime_dir is non-empty.
     */
```

It will be better to add empty defaults.vim in /usr/share/vim/ ($VIM) folder, so that once there is an official defaults.vim installed into the /usr/share/vim/vim* ($VIMRUNTIME), the empty defaults.vim will be ignored.

Here is the patch, and it has been verifed:

```diff
diff --git a/utils/vim/Makefile b/utils/vim/Makefile
index 125acbf..3eb648b 100644
--- a/utils/vim/Makefile
+++ b/utils/vim/Makefile
@@ -209,6 +209,7 @@ define Package/vim/install
        $(INSTALL_BIN) $(PKG_BUILD_DIR)/vim_tiny $(1)/usr/bin/vim
        $(INSTALL_DIR) $(1)/usr/share/vim
        $(INSTALL_CONF) ./files/vimrc $(1)/usr/share/vim/
+       touch $(1)/usr/share/vim/defaults.vim
 endef

 define Package/vim-full/install
@@ -217,6 +218,7 @@ define Package/vim-full/install
        $(INSTALL_DIR) $(1)/usr/share/vim
        $(LN) vim $(1)/usr/bin/vimdiff
        $(INSTALL_CONF) ./files/vimrc.full $(1)/usr/share/vim/vimrc
+       touch $(1)/usr/share/vim/defaults.vim
 endef

```

totkeks:
Is there any chance this PR can replace #22167 or is there actually some more thought needed here how to deal with vim9.0?

Ra2-IFV:
I agree that #22167 is a ultimate solution for this.

PR #22275, opened 2023-10-02
samba4: update smb.conf.template to disable guest logons by default

jirijanata:
Maintainer: @Gingernut1978
Compile tested: not tested
Run tested: ARM, Netgear R7800, OpenWrt 23.05.0-rc4

Description:
Insecure guest logons are on UpToDate Windows 10 1709 and later disabled by default. The option `map to guest` should be set to Samba default value = `Never`, because only then the Windows will ask to enter the password when accesing the share (when there are no valid user logged-in in Windows). If is the option `map to guest` is set to `Bad User` then Windows Client will not connect (when the user, which is logged-in in Windows is not also configured on OpenWrt) and fail with message, that insecure logons are disabled.

https://learn.microsoft.com/en-us/troubleshoot/windows-server/networking/guest-access-in-smb2-is-disabled-by-default

Typo also corrected.

### DIFF
--- a/net/samba4/files/smb.conf.template
+++ b/net/samba4/files/smb.conf.template
@@ -19,7 +19,7 @@
 
 	## This parameter controls whether a remote client is allowed or required to use SMB encryption.
 	## It has different effects depending on whether the connection uses SMB1 or SMB2 and newer:
-    ## If the connection uses SMB1, then this option controls the use of a Samba-specific extension to the SMB protocol introduced in Samba 3.2 that makes use of the Unix extensions.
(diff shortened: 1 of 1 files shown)


jefferyto:
The first line of the commit message should be prefixed with the name of the package ("samba4:"), and a real name is required for the Signed-off-by line - see [CONTRIBUTING.md](https://github.com/openwrt/packages/blob/8ce092cb1280464b0069d188394e9bdbe616e55b/CONTRIBUTING.md#commits-in-your-pull-requests-should).

jefferyto:
You haven't pushed any new changes to this PR/branch.

Issue #22250, opened 2023-09-29
`pymysql`  Recursive dependency！


zxlhhyccc:
@commodo
````
Collecting package info: done
tmp/.config-package.in:36491:error: recursive dependency detected!
tmp/.config-package.in:36491:	symbol PACKAGE_python3-pymysql depends on PYTHON3_PYMYSQL_SHA_PASSWORD_SUPPORT
feeds/packages/lang/python/pymysql/Config.in:4:	symbol PYTHON3_PYMYSQL_SHA_PASSWORD_SUPPORT depends on PACKAGE_python3-pymysql
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"
````

BKPepe:
Again reported twice? Please, try to wait for a while before screaming and shouting everywhere. That should help you to be more patient. You already it reported in https://github.com/openwrt/packages/commit/9fdff3ea94662653188c3902840e97c35e9f138f#commitcomment-128680904, which is not good place to report such things. You need to create issue, if you want to get resolved (this one you did today) and as well keep the issue template to be able to reproduce (need to be done).

PR #22279, opened 2023-10-03
pymysql: fix circular dependency

commodo:
Maintainer: me
Compile tested: x86  https://github.com/openwrt/openwrt/commit/ff95f859ebf710d2914472a3feeeb0d187d14459
Run tested: n/a

---------------------------------------

Fixes: https://github.com/openwrt/packages/issues/22250

The error is:
```
Collecting package info: done
tmp/.config-package.in:36491:error: recursive dependency detected!
tmp/.config-package.in:36491:	symbol PACKAGE_python3-pymysql depends on PYTHON3_PYMYSQL_SHA_PASSWORD_SUPPORT
feeds/packages/lang/python/pymysql/Config.in:4:	symbol PYTHON3_PYMYSQL_SHA_PASSWORD_SUPPORT depends on PACKAGE_python3-pymysql
```


cotequeiroz:
I **think** this will cause trouble if the arch does not build python3-cryptography (i.e. powerpc_8540), and you select `PYTHON3_PYMYSQL_SHA_PASSWORD_SUPPORT`.  Even though it will not try to build python3-cryptography, I **think** it will, register it as a dependency.  If my assumption is correct, when you try to install the resulting pymysql package, opkg will try to install `python3-cryptography` and it should fail.

zxlhhyccc:
I changed it so that the dependencies are automatically configured when a configuration item is selected：
https://github.com/immortalwrt/packages/pull/924

1715173329:
> I **think** this will cause trouble if the arch does not build python3-cryptography (i.e. powerpc_8540), and you select `PYTHON3_PYMYSQL_SHA_PASSWORD_SUPPORT`. Even though it will not try to build python3-cryptography, I **think** it will, register it as a dependency. If my assumption is correct, when you try to install the resulting pymysql package, opkg will try to install `python3-cryptography` and it should fail.

I think it's unrelated to this PR itself. If someone try to enable `PYTHON3_PYMYSQL_SHA_PASSWORD_SUPPORT` on a unsupported arch, `python3-pymysql` will just disappear due to unmeet dependency.

jefferyto:
I agree with @cotequeiroz in https://github.com/openwrt/packages/commit/9fdff3ea94662653188c3902840e97c35e9f138f#commitcomment-128936778:

> So, my recommendation would be to create a separate empty package inside `python3-pymysql` Makefile that will just install both `python3-pymysql` and `python3-cryptography`, with the added bonus of making the SHA password option available in the official feeds, without a custom build.

cotequeiroz:
I believe #22301 should take care of this.

PR #22297, opened 2023-10-04
add liboqs and rosenpass

aparcar:
Maintainer: me
Compile tested: broken
Run tested: -

Description:

Rosenpass adds an extra layer to wireguard tunnels to harden them against quantum computers. This is for experimental use only and this PR to check compilation on multiple architectures.


aparcar:
@jefferyto hej, could you kindly have a look at why this doesn't build? I looks like clang is required, however I added the missing BPF stuff, at least to my understanding.

PR #22300, opened 2023-10-04
nerdctl: New package

lu-zero:
contaiNERD CTL - Docker-compatible CLI for containerd,
 with support for Compose, Rootless, eStargz, OCIcrypt, IPFS, ...

Maintainer: me 
Compile tested: aarch64 23.05-rc4 and master
Run tested: bpi-r3

Description: First step for using directly containerd

### DIFF
--- a/utils/nerdctl/Makefile
+++ b/utils/nerdctl/Makefile
@@ -1,0 +1,38 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=nerdctl
+PKG_VERSION:=1.6.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/containerd/nerdctl/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=a369b1c517d9c3d53d00b29633a6176a05811214a44dd25d339c32cc6a901579
+
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=Luca Barbato <lu_zero@gentoo.org>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_BUILD_FLAGS:=no-mips16
+
+GO_PKG:=github.com/containerd/nerdctl
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/nerdctl
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=contaiNERD CTL - Docker-compatible CLI for containerd
+  URL:=https://containerd.io
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
+define Package/nerdctl/description
+  Docker-compatible CLI for containerd, with support for Compose, Rootless,
+  eStargz, OCIcrypt, IPFS, ...
+endef
+
+$(eval $(call GoBinPackage,nerdctl))
+$(eval $(call BuildPackage,nerdctl))


utils/nerdctl/Makefile
```diff
@@ -0,0 +1,52 @@
...
+PKG_BUILD_PARALLEL:=1
+PKG_BUILD_FLAGS:=no-mips16
+
+GO_PKG:=$(PKG_GIT_URL)
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
```
BKPepe, line 27: Drop this. This is duplicated from above.

utils/nerdctl/Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+PKG_NAME:=nerdctl
+PKG_VERSION:=1.6.0
+PKG_RELEASE:=1
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_GIT_URL:=github.com/containerd/nerdctl
+PKG_GIT_REF:=v$(PKG_VERSION)
```
BKPepe, line 11: Why do we need to have new variables? Please take a look how it is done for other packages.

utils/nerdctl/Makefile
```diff
@@ -0,0 +1,52 @@
...
+PKG_SOURCE_URL:=https://codeload.$(PKG_GIT_URL)/tar.gz/$(PKG_GIT_REF)?
```
BKPepe, line 12: This is getting more confusing and dont reintroduce back something which was removed in the past.

utils/nerdctl/Makefile
```diff
@@ -0,0 +1,52 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=nerdctl
+PKG_VERSION:=1.6.0
+PKG_RELEASE:=1
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE
```
BKPepe, line 7: This belongs below, where you have PKG_MAINTAINER variable.

utils/nerdctl/Makefile
```diff
@@ -0,0 +1,52 @@
...
+define Package/nerdctl
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=contaiNERD CTL - Docker-compatible CLI for containerd
+  URL:=https://containerd.io/
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
+define Package/nerdctl/description
+Docker-compatible CLI for containerd, with support for Compose, Rootless, eStargz, OCIcrypt, IPFS, ...
```
BKPepe, line 38: Two spaces as indentation in this define.

BKPepe (changes requested):

utils/nerdctl/Makefile
lu-zero: I started from the docker Makefile and another go package, looks like they are leftovers and I got lucky ^^;

BKPepe: Yeah. The Docker Makefile is awful, though. Not very great example here.

utils/nerdctl/Makefile
lu-zero: If you can point me to a good Makefile to use as template I'd be grateful :)

utils/nerdctl/Makefile
```diff
@@ -0,0 +1,47 @@
...
+  TITLE:=contaiNERD CTL - Docker-compatible CLI for containerd
+  URL:=https://containerd.io/
+  DEPENDS:=$(GO_ARCH_DEPENDS)
+endef
+
+define Package/nerdctl/description
+  Docker-compatible CLI for containerd, with support for Compose, Rootless, eStargz, OCIcrypt, IPFS, ...
+endef
+
+GO_PKG_BUILD_VARS += GO111MODULE=auto
```
BKPepe, line 36: I haven't look at it closely, but why do we need to have it? We do have Go 1.12 in the master and openwrt-23.05 branches.

utils/nerdctl/Makefile
```diff
@@ -0,0 +1,47 @@
...
+TARGET_LDFLAGS += $(if $(CONFIG_USE_GLIBC),-lc -lgcc_eh)
```
BKPepe, line 37: Not sure about this one. Maybe @jefferyto could help on this one?

utils/nerdctl/Makefile
lu-zero: I copied what's in the docker makefile, I wanted to see if nerdctl + cni-plugins would be lighter than dockerd+docker+docker-compose

BKPepe: Okay, so we just copied something from somewhere and we don't know even why and what it does. Am I wrong?

lu-zero: My faulty assumption is that the docker Makefile could be copied as-is and I forgot to make this as draft. If you have a good Makefile to point me to I'd be grateful.

utils/nerdctl/Makefile
BKPepe: Okay, this was merged by @neheb and I see only 4 packages in our repository have exactly the same.

https://github.com/openwrt/packages/pull/18428#issuecomment-1748363981

Let's see...

utils/nerdctl/Makefile
BKPepe: https://github.com/openwrt/packages/commit/0187a8d55ab22e4a77659659b2032586c4bbf894#r129204807

BKPepe:
As an example, you took very complex and not very good Makefile. You should not stick with only one example to write your own Makefile for Go. Keep in mind that you can not even blindly copy&paste code without knowing what it does. That's not how it works.

Good examples could be ``dnscrypt-proxy2``, ``v2ray-core``, ``ooniprobe``, ``cloudflared``, etc. There are many good Makefiles, hard to choose which one is better than others. :-) I just picked a few.

lu-zero:
> As an example, you took very complex and not very good Makefile. You should not stick with only one example to write your own Makefile for Go. Keep in mind that you can not even blindly copy&paste code without knowing what it does. That's not how it works.
> 
> Good examples could be `dnscrypt-proxy2`, `v2ray-core`, `ooniprobe`, `cloudflared`, etc. There are many good Makefiles, hard to choose which one is better than others. :-) I just picked a few.

Thank you a lot :) I started with `docker` since `nerdctl` is a replacement for it and...

utils/nerdctl/Makefile
lu-zero: I removed all the noise inherited from the docker Makefile and it seems still working fine from my limited testing.

1715173329 (approved):

PolynomialDivision (approved):

lu-zero:
Is there anything left to be done @BKPepe ?

PR #22303, opened 2023-10-04
ddns-scripts: add samples to package

### LINKED ISSUE #20800: ddns-scripts: include samples inside the package
Maintainer: @feckert 
Environment: OpenWrt 22.03.3 r20028-43d71ad93e

Description:

I am interested in using the [`slaac_sample.sh`](https://github.com/openwrt/packages/blob/master/net/ddns-scripts/samples/slaac_sample.sh) inside my own OpenWrt box, but interestingly it's just not included in the distributed package. I think it would be useful to include it, since if there are updates, we don't need to keep track of the file inside the git repo tree, and because the file is useful as-is.

perrin4869:
Maintainer: @feckert

Description: closes #20800
I am not sure how to run this `Makefile`. I tried to drop it inside the `openwrt` root tree, then run `make`, but it didn't work, so couldn't test how this works...

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
-PKG_RELEASE:=39
+PKG_RELEASE:=40
 
 PKG_LICENSE:=GPL-2.0
 
@@ -55,6 +55,16 @@ endef
 
 define Package/ddns-scripts-services/description
   Dynamic DNS Client definitions for supported services
+endef
+
+define Package/ddns-scripts-utils
+  $(call Package/ddns-scripts/Default)
+  TITLE:=Utility scripts for ddns configs
+  DEPENDS:=ddns-scripts
+endef
+
+define Package/ddns-scripts-utils/description
+  Dynamic DNS Client utility scripts for config files
 endef
 
 
@@ -359,6 +369,15 @@ define Package/ddns-scripts-services/install
 	rm $(1)/usr/share/ddns/default/transip.nl.json
 	rm $(1)/usr/share/ddns/default/ns1.com.json
 	rm $(1)/usr/share/ddns/default/one.com.json
+endef
+
+
+define Package/ddns-scripts-utils/install
+	$(INSTALL_DIR) $(1)/usr/lib/ddns
+	$(INSTALL_BIN) ./samples/slaac_sample.sh \
+		$(1)/usr/lib/ddns/slaac.sh
+	$(INSTALL_BIN) ./samples/getlocalip_sample.sh \
+		$(1)/usr/lib/ddns/getlocalip.sh
 endef
 
 
@@ -668,6 +687,7 @@ endef
 
 $(eval $(call BuildPackage,ddns-scripts))
 $(eval $(call BuildPackage,ddns-scripts-services))
+$(eval $(call BuildPackage,ddns-scripts-utils))
 $(eval $(call BuildPackage,ddns-scripts-luadns))
 $(eval $(call BuildPackage,ddns-scripts-cloudflare))
 $(eval $(call BuildPackage,ddns-scripts-gcp))


net/ddns-scripts/Makefile
```diff
@@ -316,6 +316,11 @@ define Package/ddns-scripts/install
 	$(INSTALL_DIR) $(1)/usr/bin
 	$(INSTALL_BIN) ./files/usr/bin/ddns.sh \
 		$(1)/usr/bin/ddns
+
+	$(INSTALL_BIN) ./samples/slaac_sample.sh \
+		$(1)/usr/lib/ddns
+	$(INSTALL_BIN) ./samples/getlocalip_sample.sh \
+		$(1)/usr/lib/ddns
```
BKPepe, line 323: **Why do we need to ship also samples in our package?** That's not clear for me. It should be described in the commit description.  Please check: https://openwrt.org/submitting-patches#submission_guidelines

Also, you need to bump PKG_RELEASE.

BKPepe (changes requested):

net/ddns-scripts/Makefile
perrin4869: Right, sorry about that! I updated the commit message and PKG_RELEASE, thanks!

BKPepe: Then another question came to my mind. Why do we really need to include samples in every installation? There should be subpackage for examples, so it does not take more space than it should have. If you want examples/samples (whatever), you should follow the documentation.

perrin4869: well, they are samples, but you can also consider them as utils...
maybe a `ddns-scripts-utils` package?

perrin4869: ok, I put together a simple `ddns-scripts-utils` package, and managed to build it (`menuconfig` with the packages repo inside the `buildroot` and then `make package/ddns-scripts/compile`, and confirmed it appeared within my `bin` directory).

BKPepe: Okay, now it looks good.

net/ddns-scripts/Makefile
```diff
@@ -57,6 +57,17 @@ define Package/ddns-scripts-services/description
   Dynamic DNS Client definitions for supported services
 endef
 
+define Package/ddns-scripts-utils
+  $(call Package/ddns-scripts/Default)
+  TITLE:=Utility scripts for ddns configs
+  DEPENDS:=ddns-scripts
+  PROVIDES:=ddns-scripts_utils
```
feckert, line 64: I would say this is not necessary. Please remove `PROVIDES:=ddns-scripts_utils`, this is only needed if there are two packages that provides the same functionality, but build with different feature sets.

net/ddns-scripts/Makefile
```diff
@@ -362,6 +373,15 @@ define Package/ddns-scripts-services/install
 endef
 
 
+define Package/ddns-scripts-utils/install
+	$(INSTALL_DIR) $(1)/usr/lib/ddns
+	$(INSTALL_BIN) ./samples/slaac_sample.sh \
+		$(1)/usr/lib/ddns
+	$(INSTALL_BIN) ./samples/getlocalip_sample.sh \
+		$(1)/usr/lib/ddns
```
feckert, line 381: Could you also prefix the files with `ddns`, and remove `_sample.sh` postfix.
So the following files get installed on the target.
* `ddns-getlocalip`
* `ddns-slaac`

The advantage is it feels more like a normal scripts with out `_sample.sh` and  with the prefix it could be find easier on the system.

feckert (changes requested):
Everything else looks good! Thanks

net/ddns-scripts/Makefile
perrin4869: I see! Just out of curiosity, why are the other packages in this makefile using `PROVIDES`?

net/ddns-scripts/Makefile
perrin4869: makes sense! done!

net/ddns-scripts/Makefile
feckert: As far as I can remember, this is because the names were changed and in order not to break downstream projects, PROVIDES was added with the old name

net/ddns-scripts/Makefile
```diff
@@ -362,6 +372,15 @@ define Package/ddns-scripts-services/install
 endef
 
 
+define Package/ddns-scripts-utils/install
+	$(INSTALL_DIR) $(1)/usr/lib/ddns
+	$(INSTALL_BIN) ./samples/slaac_sample.sh \
+		$(1)/usr/lib/ddns/ddns-slaac.sh
+	$(INSTALL_BIN) ./samples/getlocalip_sample.sh \
+		$(1)/usr/lib/ddns/ddns-getlocalip.sh
```
feckert, line 380: Now I haven't read properly where it is installed. I thought it was installed under `/bin` or `/usr/bin`. If it is installed under `/usr/lib/ddns`, we don't necessarily have to change the name.

The question now is whether it should be installed under `/bin` or `/usr/bin`, or whether we should put it under `/usr/lib/ddns` as you already have done.

For which use case are the scripts useful?

perrin4869: So, I have a raspberry pi connected to my openwrt router, and it gets its ipv6 address via slaac.
In order to have the ddns record point to my raspberry pi I have:

```
config service 'myddns_ipv6'
        option use_ipv6 '1'
        option enabled '1'
        option lookup_host '...'
        option service_name 'no-ip.com'
        option domain '...'
        option username '...'
        option password '...'
        option use_syslog '2'
        option check_unit 'minutes'
        option force_unit 'minutes'
        option retry_unit 'seconds'
        # this points to openwrt's public ip address
        # option ip_source 'network'
        # option ip_network 'wan6'
        option ip_source 'script'
        option ip_script '/usr/lib/ddns/slaac_sample.sh wan b8:27:eb:13:9a:2b' // <-- this
        option interface 'wan6'
```

I think `/usr/lib` makes more sense, they are not meant to be used outside the context of being an argument to the `ip_script` option. The [comment](https://github.com/openwrt/packages/blob/f3a6491674ab451535e392b63e94a545abf3f292/net/ddns-scripts/samples/slaac_sample.sh#L9) on the files also makes this suggestion.
I'm also thinking what's the best naming here... maybe dropping the `ddns-` prefix will make most sense?

feckert: Make it so

perrin4869: done!

feckert (approved):

feckert:
Thanks!

perrin4869:
hi, was just wondering, will this package land in 22.03? What is the pipeline until it is released? 🙂

feckert:
As I know this was only merged into the master branch. The 22.03 is the old stable, so I think it makes only sense to merge this to the new stable 23.05 ?

BKPepe:
I agree with @feckert.

perrin4869:
I see, sure. Something to look forward to when upgrading then
Out of curiosity, does master get merged into 23.05 when 23.05 gets the final release?

feckert:
> I see, sure. Something to look forward to when upgrading then Out of curiosity, does master get merged into 23.05 when 23.05 gets the final release?

The first version of openwrt-23.05 was released last week. This new stable release was branched in may, so if we want some changes from master branch to 23.05 branch, then we have to cherry-pick them.

perrin4869:
Oh I see, I can't believe I missed it!
Sure, I'll really it appreciate if you can cherry pick it

feckert:
Done https://github.com/openwrt/packages/commits/openwrt-23.05/net/ddns-scripts. The package `ddns-scripts` version is now the same in the master and openwrt-23.05 branch.

PR #22305, opened 2023-10-04
openssh: bump to 9.5p1

graysky2:
Chanelog: https://www.openssh.com/txt/release-9.5

Build system: x86/64

Maintainer: @tripolar

### DIFF
--- a/net/openssh/Makefile
+++ b/net/openssh/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openssh
-PKG_VERSION:=9.4p1
+PKG_VERSION:=9.5p1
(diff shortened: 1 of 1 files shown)


BKPepe:
Would be great if you can add release notes to the description. Something like this https://www.openssh.com/txt/release-9.5 would be sufficient. Link is okay. We don't need copy&paste it.

graysky2:
@BKPepe - Added... thinking it might be more tidy to just link them?  Quite a bit of text.  Thoughts?

BKPepe:
That was my thought in the first place. URL, which points to the changelog is ok. No need to take the text from it.

graysky2:
Fixed

BKPepe (approved):

PR #22319, opened 2023-10-06
python: add missing STAGING_DIR/host/bin to PATH for PYTHON3_VARS

Ansuel:
Add missing STAGING_DIR/host/bin to PATH for PYTHON3_VARS. This is
needed if some packages require some special tool to identify
specific cflags and include dir for dependency.

An example of this case is uwsgi python plugin that require pcre2-config
to detect pcre2 usage and what to include to be correctly built.


neheb:
why is uwsgi not using pkgconfig?

Ansuel:
@neheb they use a python script that try to call pcre2-config. If not found try to use pcre-config.

jefferyto:
`$(STAGING_DIR)/host/bin` is already in the path when building packages: https://github.com/openwrt/openwrt/blob/9536446965e853231e34c4e5dc4cf13f838b9e90/rules.mk#L218

Ansuel:
@jefferyto sure but they are not in the PATH exposed by PTYHON3_VARS. Read here... I posted what PATH is produced
https://github.com/openwrt/openwrt/pull/13621#issuecomment-1751501383

jefferyto:
You're right that there is a problem with the path but this PR isn't the right solution. The issue with this line in `PYTHON3_VARS`:

```
	PATH="$(CARGO_HOME)/bin:$(PATH)" \
```

is that `$(PATH)` is the Makefile variable `PATH`, not the shell variable `PATH`, that's why the value is wrong.

I'm working on a patchset that has the side effect of removing that line so I would prefer to try to get that done rather than doing a separate fix for this right now. If I'm not done in a few days feel free to submit a fix.

Ansuel:
@jefferyto sure! consider tho that currently uwsgi package is broken (and indirectly luci-nginx) on buildbot so If you have only part of the changes to deprecate the strange PATH handling, feel free to create a pr!

Ansuel:
Suppressed by #22339

PR #22339, opened 2023-10-08
rust: Improve build performance

jefferyto:
Maintainer: @lu-zero, me (for python-setuptools-rust)
Compile tested: armsr-armv7/armsr-armv8/malta-be/x86-generic/x86-64, 2023-10-03 snapshot sdk
Run tested: armsr-armv7/armsr-armv8/malta-be/x86-generic/x86-64 (qemu), 2023-10-03

Description:
The overall goal of these changes is to improve build performance (caching downloads, caching compilation results, not building host Python). This also improves overall build "consistency" (using the make jobserver, using the same options when building Python extensions written in Rust) and better optimizes target binary sizes.

I wasn't able to get rust/cargo to use the make jobserver when building rust itself. I opened https://github.com/rust-lang/rust/issues/116101 but so far I am not optimistic about this being resolved.

These changes have the side effect of fixing a bug I introduced earlier to enable building of Python extensions written in Rust (https://github.com/openwrt/packages/pull/22319), so I hope this can be merged soon :sweat_smile: 

Please see the individual commit messages/changes for details.


lang/rust/rust-values.mk
```diff
@@ -5,7 +5,7 @@
 # Rust Environmental Vars
 RUSTC_HOST_SUFFIX:=$(word 4, $(subst -, ,$(GNU_HOST_NAME)))
 RUSTC_HOST_ARCH:=$(HOST_ARCH)-unknown-linux-$(RUSTC_HOST_SUFFIX)
-CARGO_HOME:=$(STAGING_DIR)/host/cargo
+CARGO_HOME:=$(DL_DIR)/cargo
```
Ansuel, line 8: Mhhh... I don't like bloating the DL_DIR. From documentation the CARGO_HOME can also have cached bin and other configuration file. DL_DIR should really only have downloaded stuff.

Maybe a better place would be TMP_DIR?

Also consider that an user might use the same buildroot to compile the different ARCH. We should make sure if that might cause problems with using the same CARGO_HOME directory for this case...

jefferyto: > From documentation the CARGO_HOME can also have cached bin and other configuration file. DL_DIR should really only have downloaded stuff.

The bin directory holds binaries the user has built, where under normal circumstances `CARGO_HOME` is `~/.cargo`. Neither target or host packages should install binaries there, and no configuration file should be created there, so there is no problem to put `CARGO_HOME` in `$(DL_DIR)`.

> Also consider that an user might use the same buildroot to compile the different ARCH. We should make sure if that might cause problems with using the same CARGO_HOME directory for this case...

Under normal circumstances `CARGO_HOME` is `~/.cargo`, which would be shared among all of the architectures the user is building, so I would assume they have solved this issue.

lu-zero (approved):
It looks very nice, I'm still a bit concerned on not having rustc/cargo in CARGO_HOME, but if it is surviving a CI run we should be fine.

lang/rust/rust-host-build.mk
```diff
@@ -17,7 +25,7 @@ define Host/Compile/Cargo
 		CC=$(HOSTCC_NOCACHE) \
 		cargo install -v \
 			--profile stripped \
-			$(if $(RUST_PKG_FEATURES),--features "$(RUST_PKG_FEATURES)") \
```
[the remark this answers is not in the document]
jefferyto: It was changed to `RUST_HOST_FEATURES` - in general I think host and target build options should be separate, in this case `RUST_HOST_FEATURES` can default to `RUST_PKG_FEATURES` if not set, if that is preferred.

jefferyto:
Updates:
* "Fixed" CI build

jefferyto:
CI failed for riscv64 because ripgrep failed to build - it appears unrelated to this PR (same error causing it to fail to build for buildbots: https://downloads.openwrt.org/snapshots/faillogs/riscv64_riscv64/packages/ripgrep/compile.txt).

lang/rust/rust-host-build.mk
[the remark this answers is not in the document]
lu-zero: Not sure which way is more straightforward, this way. Overall your set is very cool :) Thank you a lot

lu-zero:
Looks like jemallocator and/or jemalloc have to be updated to work on risc-v. I guess it is fine to drop ripgrep on riscv until upstream fixes it, if somebody have access to a risc-v system would be nice to track down why that constant is missing there...

lang/rust/rust-values.mk
Ansuel: Ok something to check sometime just in case we miss some option to not use CARGO_HOME. Thanks for the explainations.

Ansuel:
@jefferyto looks ok to me... Also take notice that this (or at least the fix to PATH) should be backported to 23.05 as uwsgi is also broken there.

oskarirauta:
Few suggestions..

Rust 1.73.0 is available, if you add [this patch](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-lang/rust/files/1.72.0-bump-libc-deps-to-0.2.146.patch?id=515b5920046117355d88b3494c74da269ce9b30a) from gentoo, it's also buildable on/for musl 1.2.4 without openwrt/openwrt#12667 - but you need to add `--set=target.$(RUSTC_TARGET_ARCH).crt-static=false` to TARGET_CONFIGURE_ARGS if you are building **on** musl 1.2.4. I suggest there could be a Config.in file with option to enable it for build hosts that are musl, such as alpine or openwrt (both need some minor changes but build openwrt very well in the end, changes are needed for example for perl to build).

I have a sample [here](https://github.com/oskarirauta/openwrt-rust/blob/main/lang/rust/Makefile) (without Config file) of rust 1.73.0 that is the first to build **on** musl 1.2.4 without compatibility patch from openwrt/openwrt#12667. That Makefile has mentioned configuration arg to succeed on musl build host.

Ansuel:
@oskarirauta mhhh but the problem this pr fix would still be there correct?

oskarirauta:
@Ansuel Sure - my proposal fixes other things. But I wanted to point out some findings, such as that patch and it finally working on musl 1.2.4 (been waiting since musl 1.2.4 was out) and ofcourse, new version (actually also required for musl 1.2.4).


Just thought, that there is no point for me making a PR, if this is about to succeed soon as this is a bit different. And also, I've pointed out earlier the extra TARGET_CONFIGURE_ARGS necessary on musl build hosts, and it hasn't got any wind under it's wings, so better maybe to propose to people who are actually maintaining project. Proposal to put it as a configurable option - is something that I wanted to avoid, but from Makefile, it is very difficult to detect if host system is using musl. ldd --version prints it out, but it's not capturable, for example..

Ansuel:
@oskarirauta actually no probably never used but we set config if musl is used instead of glibc. Check your .config file. I don't like using the configuration as we would face situation with glibc selected and rust musl option set. The option should be handled internally.

oskarirauta:
@Ansuel 

I tried that. I just couldn't figure out the way.. Here's few:
```
dev@media:~# ldd --version
musl libc (x86_64)
Version 1.2.4
Dynamic Program Loader
Usage: ldd [options] [--] pathname
```

This cannot be parsed even in shell, for e.g.  `ldd --version | head -n 1` fails.
output is somehow different from normal stdout.

```
#!/bin/sh
musl_lib=$(ldd $(whereis ls | awk '{print $2}') | grep 'musl' | head -1 | cut -d ' ' -f1)
[ -n "$musl_lib" ] && musl_lib=true || musl_lib=false
echo lib: $musl_lib
```

Later works in shell... But in Makefile, if I try:
`IS_MUSL:=$(ldd $(whereis ls | awk '{print $2}') | grep 'musl' | head -1 | cut -d ' ' -f1)`
or `ifneq ($(ldd $(whereis ls | awk '{print $2}') | grep 'musl' | head -1 | cut -d ' ' -f1),)`
both attempts fail.

I'd be interested to see what you come up, it seems that quite few people have searched for answer to this. I also think it should be handled internally, just couldn't figure out the way. Checking libc link isn't reliable. libc.so isn't always link to ld-musl-x86_64.so or it cannot be counted that it is. For example on openwrt:
```
dev@media:/usr/src/openwrt# ls -l /lib/libc.so
-rwxr-xr-x 1 root root 3256328 Sep 30 23:06 /lib/libc.so
```

But I am very pleased that it was taken to consideration, finally I don't need to fork rust part of repo :)

Ansuel:
```
#
# C Library
#
# CONFIG_LIBC_USE_GLIBC is not set
CONFIG_LIBC_USE_MUSL=y
# CONFIG_MUSL_DISABLE_CRYPT_SIZE_HACK is not set
```

oskarirauta:
@Ansuel it is about host system.. What you use to build it... config does not know what is building it..

Ansuel:
oh... so you want to use musl for host? well afaik that is totally not supported... last time i tried an alpine container to build stuff there were lots of problems and never manage to complete it in an usable state so I would first check and make that work.

Did the situation changed? Also instead of ldd can't you just parse the output of opening the libc .so?

jefferyto:
> Rust 1.73.0 is available, if you add [this patch](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-lang/rust/files/1.72.0-bump-libc-deps-to-0.2.146.patch?id=515b5920046117355d88b3494c74da269ce9b30a) from gentoo, it's also buildable on/for musl 1.2.4

I think updating Rust and enabling it to build on musl hosts would be best handled in separate pull requests; this PR is probably a bit too big already.

trippleflux:
Is there a new way to  invoke compilation in subdir of $(PKG_BUILD_DIR) like MAKE_PATH in for normal compilation?.

This is useful for example gping rust package which using gping subdir of its source code : https://github.com/orf/gping/tree/master/gping .

Normally I am doing the following in the previous rust package (before this PR) : 
```
define Build/Compile
	$(call Build/Compile/Cargo,gping)
endef
```

1715173329:
please rebase to update package [procs](https://github.com/openwrt/packages/commit/482e7170e884eeb8e98582df3cb12426b7e83bb4) as well, then let's merge it?

lang/rust/rust-package.mk
```diff
@@ -15,26 +15,22 @@ ifeq ($(origin RUST_INCLUDE_DIR),undefined)
...
-			--root $(PKG_INSTALL_DIR) \
-			--path "$(if $(strip $(1)),$(strip $(1)),.)" \
-			$(2) ; \
-	)
+	+$(CARGO_PKG_VARS) \
+	cargo install -v \
+		--profile $(CARGO_PKG_PROFILE) \
+		$(if $(strip $(RUST_PKG_FEATURES)),--features "$(strip $(RUST_PKG_FEATURES))") \
+		--root $(PKG_INSTALL_DIR) \
+		--path "$(if $(strip $(1)),$(strip $(1)),$(PKG_BUILD_DIR))" \
```
lu-zero, line 31: Keep the `cd $(PKG_BUILD_DIR)` so you can pass the relative paths to the `Build/Compile/Cargo` invocation.

jefferyto:
Changes:
* Rebase on master
* Add `PKG_BUILD_PARALLEL:=1` to procs
* Restore relative path argument to `$(call Build/Compile/Cargo)`
* Expanded rustc bootstrap patch to cache downloads made by download.rs, e.g. for llvm
* Add commit to speed up `Host/Install` (gzip instead of xz, libdeflate-gzip instead of gzip)

> Is there a new way to invoke compilation in subdir of $(PKG_BUILD_DIR) like MAKE_PATH in for normal compilation?.

Fixed :pray: 

> please rebase to update package [procs](https://github.com/openwrt/packages/commit/482e7170e884eeb8e98582df3cb12426b7e83bb4) as well, then let's merge it?

Done!

lang/rust/rust-package.mk
jefferyto: This is done, no need for `cd`.

lu-zero:
Hopefully ripgrep will have a new release soon, in the mean time either we distribute a snapshot or omit it from riscv

BKPepe:
We will wait a few days until we will backport it to OpenWrt 23.05 branch, right? :)

Ansuel:
@BKPepe I would create a pr just to check if the thing can be backported with no changes

jefferyto:
> Hopefully ripgrep will have a new release soon, in the mean time either we distribute a snapshot or omit it from riscv

Perhaps you can file an issue with ripgrep and/or jemallocator about riscv support?

> We will wait a few days until we will backport it to OpenWrt 23.05 branch, right? :)

I have opened a PR (#22376), of course can wait a few days before merging.

lu-zero:
> > Hopefully ripgrep will have a new release soon, in the mean time either we distribute a snapshot or omit it from riscv
> 
> Perhaps you can file an issue with ripgrep and/or jemallocator about riscv support?

it should work fine already, it is just the current ripgrep release is very old. Switching to the current master would make it work.

> 
> > We will wait a few days until we will backport it to OpenWrt 23.05 branch, right? :)
> 
> I have opened a PR (#22376), of course can wait a few days before merging.

PR #22341, opened 2023-10-08
cni-protocol: update protocol

oskarirauta:
Changes to protocol file and it's description.
Works better now and restarts firewall automaticly when tunnel comes available. More informative/guiding description.

Maintainer: me
Compile tested: x86_64/latest git
Run tested: x86_64/latest git


oskarirauta:
@1715173329

PR #22342, opened 2023-10-08
quectel-timesync: add package

blocktrron:
Maintainer: me
Compile tested: ath79-nand
Run tested: ath79-nand (GL.iNET GL-XE300)

Description:

This tool allows for acquiring the current time from the cellular network for configuring the local clock. Compared to NTP, this has the advantage of nut using up mobile traffic.

It takes advantage of the AT+QLTS command found on Quectel modems. This functionality depends on support of the mobile network.


utils/quectel-timesync/Makefile
```diff
@@ -0,0 +1,51 @@
+# SPDX-License-Identifier: GPL-2.0-only
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=quectel-timesync
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL=https://github.com/freifunk-darmstadt/quectel-timesync.git
```
BKPepe, line 9: This looks like it is really new. The repository was created 2 days ago.  Are we sure that we want to include this in our repository?  Shouldn't it be rather added to [freifunk's feed](https://github.com/freifunk/openwrt-packages)?

blocktrron: Did you just jump to the conclusion this package is Freifunk specific just because it was developed by a Freifunk group?

The package is not specific to Freifunk, it's for synchronizing time to the host system by using the network-provided time information. This helps when using IoT SIM cards with heavily limited data quotas, as this process is not billed.

Also please point me to the policy of repository age being relevant for adding a package to the packages feed.

BKPepe: > Also please point me to the policy of repository age being relevant for adding a package to the packages feed.

Why do we need to add new packages for a few users? We do already have in this repository a large amount of packages, which are not already maintained. That's the same case when upstream developers are not active or refuse to accept someone's changes, let's fork it and add it here. Then, questions like security arise. Who will review those changes? Who will be responsible for it? 

We don't even have a policy to add new packages at all! We can add everything. That's not good.

Yes, repository age is relevant because how long it will be maintained? Every single package takes hardware resources, and as you can see we do have lack of maintainers, we are still adding new packages and the situation is worse than it should be. I am removing packages as much as I can for whatever reason I said here.

 > Did you just jump to the conclusion this package is Freifunk specific just because it was developed by a Freifunk group?
 
 No, my conclusion is that as it developed by Freifunk group, it will be mostly used by them, right? In the past, we had Freifunk feed, which was added then removed. So, why do we need to have it here? That's my question. It can be compiled and used by them. I have not seen here many guys from Freifunk. If they would be active here, I don't have any issues with it.

PR #22344, opened 2023-10-09
python-zope-interface: Update to 6.1, refresh patch

jefferyto:
Maintainer: me
Compile tested: armsr-armv7, 2023-10-07 snapshot sdk
Run tested: armsr-armv7 (qemu), 2023-10-07 snapshot

Description:
This also updates the list of dependencies and adds a test.sh script for the packages feed CI.

### DIFF
--- a/lang/python/python-zope-interface/Makefile
+++ b/lang/python/python-zope-interface/Makefile
@@ -9,11 +9,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-zope-interface
-PKG_VERSION:=6.0
+PKG_VERSION:=6.1
 PKG_RELEASE:=1
 
 PYPI_NAME:=zope.interface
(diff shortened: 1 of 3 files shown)


lang/python/python-zope-interface/Makefile
```diff
@@ -29,7 +29,7 @@ define Package/python3-zope-interface
   SUBMENU:=Python
   TITLE:=Interfaces for Python
   URL:=https://github.com/zopefoundation/zope.interface
-  DEPENDS:=+python3-light
+  DEPENDS:=+python3-light +python3-logging
```
BKPepe, line 32: Maybe we can say in the commit description, that there is a new dependency together with the test.sh. Otherwise, it looks like usual bump and there are many changes.

jefferyto: Done!

jefferyto:
Updates:
* Refreshed patch
* Rebased on master
* Updated commit message

PR #22345, opened 2023-10-09
librespeed-go: Reload the daemon after modifying the tls certificate

muink:
Maintainer: Tianling Shen cnsztl@immortalwrt.org @1715173329 
Compile tested: x86_64, x86/64, 22.03.5
Run tested: x86_64, x86/64, 22.03.5

Description:
Make the daemon reload after the tls certificate is updated

### DIFF
--- a/net/librespeed-go/files/librespeed-go.init
+++ b/net/librespeed-go/files/librespeed-go.init
@@ -12,6 +12,7 @@ mount_jail_file() {
 	local cfg="$1"
 	local isdir="${2:-0}"
 	local rw="${3:-0}"
+	local reload="${4:-0}"
 
 	local value
 	config_get value "config" "$cfg"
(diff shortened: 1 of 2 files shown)


net/librespeed-go/files/librespeed-go.init
```diff
@@ -31,6 +32,10 @@ mount_jail_file() {
 	else
 		procd_add_jail_mount "$value"
 	fi
+
+	if [ "$refresh" = "1" ]; then
+		procd_append_param file "$value"
+	fi
```
1715173329, line 38:
```suggestion
	[ "$refresh" = "0" ] || procd_append_param file "$value"
```

btw i'd prefer `reload` over `refresh`

1715173329 (changes requested):
Please:
- Drop unnecessary description from your commit message
  `Make the daemon reload after the tls certificate is updated.` is enough
- Use a valid email address

1715173329 (approved):

BKPepe:
Requires to bump PKG_RELEASE in the Makefile.

1715173329:
Merged, thank you!

PR #22351, opened 2023-10-09
pdns: update to 4.8.3

Habbie:
Maintainer: me
Compile tested: github CI
Run tested: Docker openwrt/rootfs x86_64

Description:

### DIFF
--- a/net/pdns/files/pdns.conf-dist
+++ b/net/pdns/files/pdns.conf-dist
@@ -141,6 +141,11 @@
 # default-api-rectify=yes
 
 #################################
+# default-catalog-zone	Catalog zone to assign newly created primary zones (via the API) to
+#
+# default-catalog-zone=
+
+#################################
 # default-ksk-algorithm	Default KSK algorithm
 #
 # default-ksk-algorithm=ecdsa256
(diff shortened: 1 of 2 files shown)


net/pdns/files/pdns.conf-dist
```diff
@@ -140,6 +140,11 @@
 #
 # default-api-rectify=yes
 
+#################################
+# default-catalog-zone	Catalog zone to assign newly created primary zones (via the API) to
+#
+# default-catalog-zone=
```
BKPepe, line 146: There isn't said anything about this in the commit description. You should do these changes in separate commit.

BKPepe (changes requested):

1715173329:
ping @Habbie

net/pdns/files/pdns.conf-dist
Habbie: I can update the commit description, but doing it in a separate commit doesn't make sense. It's part of the version bump.

Habbie: commit description updated

PR #22352, opened 2023-10-09
stress-ng: bump to version 0.17.0

commodo:
Maintainer: me
Compile tested: x86 https://github.com/openwrt/openwrt/commit/e66eed033f9f9d27fc839d81d3a03d4fad1b9b5b
Run tested: x86 https://github.com/openwrt/openwrt/commit/e66eed033f9f9d27fc839d81d3a03d4fad1b9b5b

    Refreshed 001-disable-extra-stressors.patch
    Dropped 002-disable-compiler-test.patch
      - no longer needed since commit https://github.com/ColinIanKing/stress-ng/commit/a24c7f2048548e6e9ded652b0d16a7da37e4edf0


utils/stress-ng/patches/002-disable-compiler-test.patch
```diff
@@ -1,13 +0,0 @@
---- a/Makefile.config
```
BKPepe, line 1: Would be good to mention in the commit description that this patch was intentionally removed.

commodo: done

BKPepe (approved):

utils/stress-ng/patches/002-disable-compiler-test.patch
BKPepe: Thank you! :)

PR #22356, opened 2023-10-09
shadowsocks-libev: convert to PCRE2

Ansuel:
Convert package to PCRE2 by porting a pending patch from a closed PR.

The PR is old but the code never changed and is simple enough to check the changes. The patch apply directly with no changes (aside from commenting out the travis CI file)

The PR was never merged as PCRE2 at times was too new and they were trying to find a better regex lib.

---

@BKPepe look what I found. If the stats are high enough, we can consider using this patch... the code is very little and use basic pcre API.

### DIFF
--- a/net/shadowsocks-libev/Makefile
+++ b/net/shadowsocks-libev/Makefile
@@ -14,7 +14,7 @@ include $(TOPDIR)/rules.mk
 #
 PKG_NAME:=shadowsocks-libev
 PKG_VERSION:=3.3.5
-PKG_RELEASE:=9
+PKG_RELEASE:=10
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://github.com/shadowsocks/shadowsocks-libev/releases/download/v$(PKG_VERSION)
@@ -29,7 +29,7 @@ PKG_FIXUP:=autoreconf
 PKG_INSTALL:=1
 PKG_BUILD_FLAGS:=no-mips16 lto
 PKG_BUILD_PARALLEL:=1
-PKG_BUILD_DEPENDS:=c-ares pcre
+PKG_BUILD_DEPENDS:=c-ares pcre2
 
 include $(INCLUDE_DIR)/package.mk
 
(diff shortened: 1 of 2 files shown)


yousong:
The upstream repo is now in a bugfix-only state and the community there seems to have turned to another implementation.  This means very likely we are going to maintain the added code on our own.  That is not good if the package is to be gradually faded out.

BKPepe:
Guys, I am thinking. Let's bring here one more thought about this, maybe it will clear some things or not. You can decide on this one. It was said in the tracking list for removing PCRE2 that we (or someone else) should package ``shadowsocks-rust``.

Steps:
1. Keep shadowsocks-libev with conversion to PCRE2 in the master branch.
Period: 6 months / 1 year or until alternative is added. The rust one. 👀 

2. Remove PCRE from shadowsocks-libev in the stable branches
This helps us to remove deprecated PCRE in the repository. 

Also/or, we can stick to Ansuel's solution, because shadowsocks-libev as it says, is not maintained anymore. No new versions will be released.

yousong:
Sorry, besides the points already presented, other thing is that time budget is tight on me that I am reluctant to take the patch.  It requires proper reviewing and polishing.  The upstream project maintainers know it better if we are not to seriously invest time in it.

Ansuel:
@yousong for me reverting acl here and luci is way too much effort :( 

Also if you check the patch, 3/4 of it is dropping pcre configure detection and introducing pcre2 configure... we litterally change 2 function. the pcre source code is like 50 lines of code.

1715173329:
rust port is ready ([Makefile here](https://github.com/immortalwrt/packages/blob/d2c511d4a30a211a415b7cdfe58318f0cad831ac/net/shadowsocks-rust/Makefile)), pretty simple.

There're some incompatible arguments in the init script needs to be updated/checked, but I don't use that script anyway...

BKPepe:
> rust port is ready ([Makefile here](https://github.com/immortalwrt/packages/blob/d2c511d4a30a211a415b7cdfe58318f0cad831ac/net/shadowsocks-rust/Makefile)), pretty simple.
> 
> There're some incompatible arguments in the init script needs to be updated/checked, but I don't use that script anyway...

Maybe, you can add it here without the init script, so we do have simple package here to someone, who will be able to test it? Hard to say. What you think?

Regarding this PR, we need to do something with this. Merging is option, too! 😆

Ansuel:
@BKPepe imho ok to merge while we migrate to rust variant.

1715173329:
> Maybe, you can add it here without the init script, so we do have simple package here to someone, who will be able to test it? Hard to say. What you think?

Done in #22477.

PR #22358, opened 2023-10-09
shadowsocks-libev: convert to PCRE2

Ansuel:
Convert package to PCRE2 by porting a pending patch from a closed PR.

The PR is old but the code never changed and is simple enough to check the changes. The patch apply directly with no changes (aside from commenting out the travis CI file)

The PR was never merged as PCRE2 at times was too new and they were trying to find a better regex lib.

---

@BKPepe look what I found. If the stats are high enough, we can consider using this patch... the code is very little and use basic pcre API.


Ansuel:
(testing 1-2 thing with github actions)

PR #22359, opened 2023-10-10
cligen: Initial packaging

pprindeville:
Maintainer: me, @olofhagsand
Compile tested: x86_64, generic, HEAD (703d667a0c)

Description:

CLIgen is a Command-Line Interface generator

cc: @olofhagsand

### DIFF
--- a/utils/cligen/Makefile
+++ b/utils/cligen/Makefile
@@ -1,0 +1,60 @@
+#
+# Copyright (C) 2020-2023 Olof Hagsand and Rubicon Communications, LLC(Netgate)
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=cligen
+PKG_VERSION:=6.5.0
+PKG_RELEASE:=1
+PKG_MAINTAINER:=Olof Hagsand <olof@hagsand.se>, Philip Prindeville <philipp@redfish-solutions.com>
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/clicon/$(PKG_NAME)/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=8b3943430f7aa9eea6a5f7cf1ace5b68eb382380cf68f41ae3ef5e032e08816f
+
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE.md
+
+PKG_INSTALL:=1
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/cligen
+  SECTION:=utils
+  CATEGORY:=Utilities
+  URL:=https://www.cligen.se
+  TITLE:=CLIgen is a Command-Line Interface generator
+  DEPENDS:=libxml2
+endef
+
+define Package/cligen/description
+  CLIgen is a Command-Line Interface generator
+endef
+
+CONFIGURE_ARGS+= \
+	--exec-prefix=/usr
+
+CONFIGURE_ARGS:=$(filter-out --disable-dependency-tracking,$(CONFIGURE_ARGS))
+
+INSTALLFLAGS:=-s --strip-program=$(TARGET_CROSS)strip
+
+CONFIGURE_VARS+= \
+	INSTALLFLAGS="$(INSTALLFLAGS)"
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/include
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/cligen $(1)/usr/include/
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libcligen.so* $(1)/usr/lib/
+endef
+
+define Package/cligen/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libcligen.so* $(1)/usr/lib/
+endef
+
+$(eval $(call BuildPackage,cligen))


utils/cligen/Makefile
```diff
@@ -0,0 +1,66 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+
+# Package definition; instructs on how and where our package will appear in the overall configuration menu ('make menuconfig')
+define Package/cligen
+	SECTION:=utils
+	CATEGORY:=Utilities
+	URL:=http://www.cligen.se
+	TITLE:=CLIgen is a Command-Line Interface generator
+endef
```
BKPepe, line 32: 1. Two spaces as indentation.
2. No dependencies at all?
3. The website uses https, so we can use https here as well even though there is redirect.

utils/cligen/Makefile
```diff
@@ -0,0 +1,66 @@
...
+PKG_NAME:=cligen
+PKG_VERSION:=6.4.0
+PKG_RELEASE:=1
+PKG_MAINTAINER:=Olof Hagsand <olof@hagsand.se>, Philip Prindeville <philipp@redfish-solutions.com>
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE.md
+
+PKG_SOURCE_URL:=https://github.com/clicon/$(PKG_NAME)/archive/refs/tags
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL_FILE:=$(PKG_VERSION).tar.gz
```
BKPepe, line 18: Why? This does not look good.

utils/cligen/Makefile
```diff
@@ -0,0 +1,66 @@
...
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=cligen
+PKG_VERSION:=6.4.0
+PKG_RELEASE:=1
+PKG_MAINTAINER:=Olof Hagsand <olof@hagsand.se>, Philip Prindeville <philipp@redfish-solutions.com>
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE.md
+
+PKG_SOURCE_URL:=https://github.com/clicon/$(PKG_NAME)/archive/refs/tags
```
BKPepe, line 16:
```suggestion
PKG_SOURCE_URL:=https://github.com/clicon/cligen/archive/refs/tags
```

Not sure, where you seen that this variable PKG_NAME should be in the PKG_SOURCE_URL. It looks like to me that it is some bad habbit and in some examples should be corrected.

utils/cligen/Makefile
```diff
@@ -0,0 +1,66 @@
...
+# See /LICENSE for more information.
+#
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=cligen
+PKG_VERSION:=6.4.0
+PKG_RELEASE:=1
+PKG_MAINTAINER:=Olof Hagsand <olof@hagsand.se>, Philip Prindeville <philipp@redfish-solutions.com>
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE.md
```
BKPepe, line 14: This belongs below PKG_HASH.

utils/cligen/Makefile
```diff
@@ -0,0 +1,66 @@
...
+
+# Package definition; instructs on how and where our package will appear in the overall configuration menu ('make menuconfig')
+define Package/cligen
+	SECTION:=utils
+	CATEGORY:=Utilities
+	URL:=http://www.cligen.se
+	TITLE:=CLIgen is a Command-Line Interface generator
+endef
+
+# Package description; a more verbose description on what our package does
```
BKPepe, line 34: Ok, thanks for comment, dude! So why it is not more verbose description, which is in the TITLE? 🤔

utils/cligen/Makefile
```diff
@@ -0,0 +1,66 @@
...
+define Package/cligen/description
+	CLIgen is a Command-Line Interface generator
+endef
+
+CONFIGURE_ARGS:=$(filter-out --disable-dependency-tracking,$(CONFIGURE_ARGS))
+
+# Don't strip on arm/bcm/pi (INSTALLFLAGS="")
+define Build/Configure
+	$(call Build/Configure/Default, INSTALLFLAGS="")
+endef
```
BKPepe, line 44: This can be removed, right?

utils/cligen/Makefile
```diff
@@ -0,0 +1,66 @@
...
+
+define Build/Compile
+	$(call Build/Compile/Default, DESTDIR="$(PKG_INSTALL_DIR)")
+endef
```
BKPepe, line 48: Hmm. Not sure, if I like this. Isn't there better way to handle this?

utils/cligen/Makefile
```diff
@@ -0,0 +1,66 @@
...
+
+# Source packages defining binary packages that ship shared libraries should declare a Build/InstallDev
+# recipe that copies all resources required to discover and link the shared libraries into the staging
+# directory.
```
BKPepe, line 52: These comments can be removed, documentation is elsewhere.

BKPepe (changes requested):

utils/cligen/Makefile
pprindeville: Done

utils/cligen/Makefile
pprindeville: I just ran:

```
grep -r PKG_SOURCE_URL | grep PKG_NAME
```

and got 86 hits.  It's apparently not that rare.  What's the problem with it?  I like it because it's idiomatic and avoids duplication.

utils/cligen/Makefile
pprindeville: Because the `PKG_NAME` doesn't occur in the tarball name on GitHub, since these are tags, not releases.  When we download it then, we need to make sure it gets put into a file which includes the `PKG_NAME` or `cligen` and `clixon` with clash as they'll both be called `6.4.0.tar.gz`.

BKPepe: Hey man, c'mon. This is wrong. Don't give it even a try to convince me that there is no better way. Of course, there is. :) This is not the only package, which is using tags. Thus we don't need to reinvent a wheel once again. 

For example, if we want to use codeload:
https://codeload.github.com/clicon/cligen/tar.gz/6.4.0?

It will look in Makefile like this:
```
PKG_SOURCE_URL:=https://codeload.github.com/clicon/cligen/tar.gz/V$(PKG_VERSION)?
PKG_HASH:=sha256sum
```

If you don't want to use codeload, then it's fine. Feel free to look at OpenWrt documentation or you can try to use some Makefile as an example in this repository.

utils/cligen/Makefile
pprindeville: Deleting most of the comments as they seem to be reminders to self...

utils/cligen/Makefile
pprindeville: Done.

utils/cligen/Makefile
pprindeville: I think it can safely be removed.

utils/cligen/Makefile
pprindeville: Yup.

utils/cligen/Makefile
pprindeville: I hadn't used `codeload` before, surprisingly.  Done.

pprindeville:
@BKPepe I think I got all of your  comments addressed.  Thanks for the thoughtful review!

pprindeville:
@BKPepe Any idea why the ARM builds are failing on the `install ... -s ...` of the libraries?

BKPepe:
I will need to look into it.

Dne po 23. 10. 2023 20:37 uživatel Philip Prindeville <
***@***.***> napsal:

pprindeville:
If we can figure out this part:

```
/builder/staging_dir/host/bin/install -c -m 0755 -d /builder/build_dir/target-aarch64_generic_musl/cligen-6.4.0/ipkg-install/usr/lib
/builder/staging_dir/host/bin/install -c -m 0644 -s libcligen.so.6.4 /builder/build_dir/target-aarch64_generic_musl/cligen-6.4.0/ipkg-install/usr/lib
strip: Unable to recognise the format of the input file `/builder/build_dir/target-aarch64_generic_musl/cligen-6.4.0/ipkg-install/usr/lib/libcligen.so.6.4'
/builder/staging_dir/host/bin/install: strip process terminated abnormally
make[3]: *** [Makefile:157: install-lib] Error 1
```

we should be good to go.  Not sure what the stripping and the ELF header have to do with each other... unless install only understands native architecture and we're not running the cross-compiled version.

pprindeville:
> I will need to look into it. 

Anything?

LynxChaus:
/builder/staging_dir/host/bin/install is symlink to $(HOST)/usr/bin/install, default INSTALLFLAGS set in comfigure.ac to "-s" (strip).

You have two options:
1 - remove setting INSTALLFLAGS from configure.ac or Makefile.in
2 - tell install about rigth strip programm via '--strip-program=$STRIP'

LynxChaus:
Well. buildfarm configured with USE_SSTRIP, $(STRIP) set to "sstrip -z", install can't handle program with switches in --strip-program="..." switch.
So - remove -s from INSTALLFLAGS and call $(STRIP) $(PKG_INSTALL_DIR)/usr/lib/libcligen.so* naually in install section. 
And remove debugging "file $(PKG_INSTALL_DIR)/usr/lib/libcligen.so*" entry from Makefile

utils/cligen/Makefile
```diff
@@ -0,0 +1,73 @@
...
+CONFIGURE_ARGS+= \
+	--exec-prefix=/usr
+
+CONFIGURE_ARGS:=$(filter-out --disable-dependency-tracking,$(CONFIGURE_ARGS))
+
+INSTALLFLAGS:=-s
+ifneq ($(STRIP),:)
+INSTALLFLAGS+= \
+	--strip-program="$(STRIP)"
+endif
```
LynxChaus, line 47: Remove this block.

utils/cligen/Makefile
```diff
@@ -0,0 +1,73 @@
...
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/cligen $(1)/usr/include/
+	$(INSTALL_DIR) $(1)/usr/lib
+	file $(PKG_INSTALL_DIR)/usr/lib/libcligen.so*
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libcligen.so* $(1)/usr/lib/
+endef
+
+define Package/cligen/install
+	echo "STRIP=$(STRIP)"
+	$(INSTALL_DIR) $(1)/usr/lib
+	file $(PKG_INSTALL_DIR)/usr/lib/libcligen.so*
```
LynxChaus, line 69: replace to ```$(STRIP) $(PKG_INSTALL_DIR)/usr/lib/libcligen.so*```

utils/cligen/Makefile
```diff
@@ -0,0 +1,73 @@
...
+$(info CONFIGURE_VARS="$(CONFIGURE_VARS)")
+$(info INSTALLFLAGS="$(INSTALLFLAGS)")
+
+define Build/InstallDev
+	echo "STRIP=$$(STRIP)"
+	echo "INSTALLFLAGS=$$(INSTALLFLAGS)"
+	$(INSTALL_DIR) $(1)/usr/include
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/cligen $(1)/usr/include/
+	$(INSTALL_DIR) $(1)/usr/lib
+	file $(PKG_INSTALL_DIR)/usr/lib/libcligen.so*
```
LynxChaus, line 62: replace to ```$(STRIP) $(PKG_INSTALL_DIR)/usr/lib/libcligen.so*```

utils/cligen/Makefile
```diff
@@ -0,0 +1,73 @@
...
+
+CONFIGURE_VARS+= \
+	INSTALLFLAGS='$(INSTALLFLAGS)'
+
+$(info STRIP="$(STRIP)")
+$(info CONFIGURE_VARS="$(CONFIGURE_VARS)")
+$(info INSTALLFLAGS="$(INSTALLFLAGS)")
+
+define Build/InstallDev
+	echo "STRIP=$$(STRIP)"
```
LynxChaus, line 57: remove all debugging echo's

LynxChaus (changes requested):

utils/cligen/Makefile
pprindeville: Have a look at https://github.com/openwrt/openwrt/blob/main/rules.mk#L310C1-L322 in particular line 319.  The subsequent lines where `STRIP` gets changed, but `TARGET_CONFIGURE_OPTS` still passes in the previous value of `STRIP=$(TARGET_CROSS)strip` just seems highly problematic.

utils/cligen/Makefile
pprindeville: That was for debugging to see what happens in CI/CD.

utils/cligen/Makefile
pprindeville: Also for debugging of CI/CD.

utils/cligen/Makefile
pprindeville: The strip should actually be happening inside the package's own `Makefile`, not here.

pprindeville:
@BKPepe Please re-review.

pprindeville:
> Well. buildfarm configured with USE_SSTRIP, $(STRIP) set to "sstrip -z", install can't handle program with switches in --strip-program="..." switch.
> So - remove -s from INSTALLFLAGS and call (PKG_INSTALL_DIR)/usr/lib/libcligen.so* naually in install section.
> And remove debugging "file $(PKG_INSTALL_DIR)/usr/lib/libcligen.so*" entry from Makefile

It's building now.

pprindeville:
@BKPepe @LynxChaus Are we good to merge?

pprindeville:
> @BKPepe @LynxChaus Are we good to merge?

Didn't hear back, so proceeding...

utils/cligen/Makefile
```diff
@@ -0,0 +1,60 @@
...
+define Package/cligen
+  SECTION:=utils
+  CATEGORY:=Utilities
+  URL:=https://www.cligen.se
+  TITLE:=CLIgen is a Command-Line Interface generator
+  DEPENDS:=libxml2
+endef
+
+define Package/cligen/description
+  CLIgen is a Command-Line Interface generator
```
BKPepe, line 35: This is same as TITLE, imho, not good.

pprindeville: Okay, will push a separate PR with updates shortly.

pprindeville: Here: https://github.com/openwrt/packages/pull/22823

PR #22361, opened 2023-10-10
python-curl: fix missing curl-config path

oskarirauta:
python-curl wheel setup-tools failed to build because it wasn't able to find curl-config.

curl-config path is requested in setup.py, line 218: 
`curl_config = os.environ.get('PYCURL_CURL_CONFIG', "curl-config")`

and current Makefile did not set this variable. After adding it to build vars, package buils without issues.

curl-config can be found from 2 locations:
 - staging_dir/target/usr/bin/curl-config
 - staging_dir/target/host/bin/curl-config (which is link to first mentioned).

this commit uses first one. Could use either one.

Maintainer: @valdi74 

### DIFF
--- a/lang/python/python-curl/Makefile
+++ b/lang/python/python-curl/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=pycurl
 PKG_VERSION:=7.45.2
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PYPI_NAME:=$(PKG_NAME)
 PKG_HASH:=5730590be0271364a5bddd9e245c9cc0fb710c4cbacbdd95264a3122d23224ca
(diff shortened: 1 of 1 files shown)


jefferyto:
If you remove [this line](https://github.com/openwrt/packages/blob/93cbaacbfb13048ad378520a7afea7c9027dd1d6/lang/python/python3-package.mk#L50), does the build succeed?

oskarirauta:
Very propably, seems like a perfect reason for error. I encountered similar error with psycopg2- it is different conf though and didn’t have similar variable so I did not fix that while at it.But can’t test it at this moment, had to go to work. Anyway, proposal: add stagingdir/host/bin after cargo path to path environment. Can’t hurt and most likely fixes the issue.Lähetetty iPhonestaJeffery To ***@***.***> kirjoitti 11.10.2023 kello 5.38:﻿
If you remove this line, does the build succeed?

oskarirauta:
closing this- fixed by #22339, fixes same issue for python-psycopg2 also

PR #22370, opened 2023-10-11
unzip: add missing headers

feckert:
Maintainer: only an example for adding CVE informations
Compile tested: x86_64, APU3
Run tested: not needed

Description:
This commits add a valid patch header for each patch, so that additional information can be stored.
This is in this case and 'CVE:' tags. This tag can be used in the feature by CVE scanner to find out if the patch fixes a CVE

See pullrequest disussion:
https://github.com/openwrt/openwrt/pull/13646


BKPepe:
Why this is still draft?

BKPepe:
@feckers Could you please rebase your PR and then merge it? :) No need to have this blocked anymore.

BKPepe:
ping @feckert

PR #22377, opened 2023-10-12
h2o+dnsdist: fix for CVE-2023-44487

Habbie:
Maintainer: me 
Compile tested: github actions
Run tested: x86_64 docker openwrt/rootfs

Description:


BKPepe:
Since, we are adding ABI-breaking patch. What about adding ``ABI_VERSION``?
https://openwrt.org/docs/guide-developer/package-policies#abi_version

I think, it would help in your case.

PR #22380, opened 2023-10-12
rust: update to 1.73.0

oskarirauta:
changelog at https://github.com/rust-lang/rust/releases/tag/1.73.0

Maintainer: @lu-zero 

### DIFF
--- a/lang/rust/patches/0002-rustc-bootstrap-cache.patch
+++ b/lang/rust/patches/0002-rustc-bootstrap-cache.patch
@@ -1,6 +1,6 @@
 --- a/src/bootstrap/bootstrap.py
 +++ b/src/bootstrap/bootstrap.py
-@@ -543,7 +543,7 @@ class RustBuild(object):
+@@ -546,7 +546,7 @@ class RustBuild(object):
                  shutil.rmtree(bin_root)
  
              key = self.stage0_compiler.date
@@ -11,7 +11,7 @@
                  os.makedirs(rustc_cache)
 --- a/src/bootstrap/download.rs
 +++ b/src/bootstrap/download.rs
-@@ -507,7 +507,10 @@ impl Config {
+@@ -520,7 +520,10 @@ impl Config {
          key: &str,
          destination: &str,
      ) {
@@ -23,10 +23,10 @@
          let cache_dir = cache_dst.join(key);
          if !cache_dir.exists() {
              t!(fs::create_dir_all(&cache_dir));
-@@ -627,7 +630,10 @@ download-rustc = false
+@@ -647,7 +650,10 @@ download-rustc = false
          let llvm_assertions = self.llvm_assertions;
  
-         let cache_prefix = format!("llvm-{}-{}", llvm_sha, llvm_assertions);
+         let cache_prefix = format!("llvm-{llvm_sha}-{llvm_assertions}");
 -        let cache_dst = self.out.join("cache");
 +        let cache_dst = match env::var_os("OPENWRT_RUSTC_BOOTSTRAP_CACHE") {
 +            Some(v) => PathBuf::from(v),
--- a/lang/rust/Makefile
+++ b/lang/rust/Makefile
@@ -5,12 +5,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=rust
-PKG_VERSION:=1.72.0
-PKG_RELEASE:=3
+PKG_VERSION:=1.73.0
+PKG_RELEASE:=1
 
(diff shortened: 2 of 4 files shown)


oskarirauta:
Also preparation to conversation started [here](https://github.com/openwrt/packages/pull/22339#issuecomment-1753948111) about musl host support, but this just updates to 1.73.0.

@Ansuel 
rust has been long supported with musl(host) with various repositories, such as alpine. Also if openwrt is supported as a build host, by using openwrt/openwrt#12667 and as it seems, with gentoo as well now that 1.73.0 is out, by just patching it to use newer libc from gentoo at [here](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-lang/rust/files/1.72.0-bump-libc-deps-to-0.2.146.patch?id=515b5920046117355d88b3494c74da269ce9b30a)

support provided by alpine is a bit too complex, I was never able to adopt it to openwrt packages. With 1.73.0 and libc patch, it worked like a charm with `--set=target.$(RUSTC_TARGET_ARCH).crt-static=false` in `TARGET_CONFIGURE_ARGS` - using that with glibc does not hurt, but I think it would be nicer to retain it as it is and use it only if build host is musl.

But this is just conversation about topic. Hopefully idea gets some hope, but let's update rust first.

oskarirauta:
converted to draft for the moment to do some local checking. I'll get back to this..

lu-zero:
> support provided by alpine is a bit too complex, I was never able to adopt it to openwrt packages. With 1.73.0 and libc patch, it worked like a charm with `--set=target.$(RUSTC_TARGET_ARCH).crt-static=false` in `TARGET_CONFIGURE_ARGS` - using that with glibc does not hurt, but I think it would be nicer to retain it as it is and use it only if build host is musl.

Doesn't hurt and since OpenWrt favours shared libraries it is future proof for when we'll have to consider other libcs.

lang/rust/patches/0002-rustc-bootstrap-cache.patch
```diff
@@ -23,10 +23,10 @@
...
          if !cache_dir.exists() {
              t!(fs::create_dir_all(&cache_dir));
-@@ -627,7 +630,10 @@ download-rustc = false
+@@ -647,7 +650,10 @@ download-rustc = false
          let llvm_assertions = self.llvm_assertions;
  
-         let cache_prefix = format!("llvm-{}-{}", llvm_sha, llvm_assertions);
+         let cache_prefix = format!("llvm-{llvm_sha}-{llvm_assertions}");
 -        let cache_dst = self.out.join("cache");
 +        let cache_dst = match env::var_os("OPENWRT_RUSTC_BOOTSTRAP_CACHE") {
```
lu-zero, line 31: Probably we could float it towards upstream using a less os-specific name

lu-zero (approved):
Looks fine, thank you for updating it :)

oskarirauta:
> Doesn't hurt and since OpenWrt favours shared libraries it is future proof for when we'll have to consider other libcs.

Should I then add it to this already?

lang/rust/patches/0002-rustc-bootstrap-cache.patch
oskarirauta: I'll put there whatever you want- that is just refreshed patch that appeared in previous package release update (1.72.0-3) without me doing any actual changes.

lu-zero:
I'm fine with folding it in :) (and also your change for crt-static)

oskarirauta:
Okay, thanks.. I'll make the necessary adjustment to PR..

oskarirauta:
@lu-zero 

Where are we at with that os branding topic within patch 0002-rustc-bootstrap-cache.patch?

jefferyto:
Re: 0002-rustc-bootstrap-cache.patch - the bootstrap already "caches" downloaded files in the build directory. I imagine this is fine for normal users but for us the build directory is not persisted between builds. I'm not sure upstream would be interested in adding more customization (like an environment variable) for where to save this cache, but then again I never asked.

lu-zero:
I think it is a good idea overall, sometimes you may have more than a tree around even in other scenarios :)

oskarirauta:
So, how do we proceed?

jefferyto:
> I'd land this set, anybody willing to change the env name to `RUSTC_BOOTSTRAP_CACHE` and send upstream for the future?

I opened https://github.com/rust-lang/rust/pull/116697. I think the patch here can stay as it is until if/when the change is merged upstream.

oskarirauta:
Though I thought that I don't participate in decision about it, I have to say that I agree; maybe it's not topic of this PR. Do good.

PR #22382, opened 2023-10-13
ooniprobe: remove unused package

BKPepe:
Maintainer: @ja-pa 
Compile and run tested: NO

Description:

This package is not maintained anymore in the OpenWrt packages feed and since we updated Go to 1.21 version, it is not compiled either.

Let's hope that with removing this package from our feed, someone will step it and become a maintainer to take care of this package.


BKPepe:
> Is this package used by Turris? I assume that is the reason why the package was added in the beginning.

You are right, but it is not used by default there and @ja-pa is not active here and he is not working for CZ.NIC/Turris for almost 2 years from now and no one from Turris step in here to take care of this. If they wish to keep the package, they can do that in their own playground, I mean downstream. :-) Or... they can re-submit it here.

jefferyto:
> no one from Turris step in here to take care of this

Perhaps we can ping someone from Turris here, so they have a chance to respond?

BKPepe:
Sure, I tried it many times and still they dont respond. @turris-cz

BKPepe:
@miska 👀

BKPepe:
Okay, so I think this can be removed, right @jefferyto ? :) No answer, I think it counts. They are not interested in it anymore.

jefferyto:
Would you mind adding this to the abandoned packages repo?

BKPepe:
Sure, will do that.

Dne út 24. 10. 2023 6:41 uživatel Jeffery To ***@***.***>
napsal:

PR #22393, opened 2023-10-13
maturin: Update to 1.3.0, build as Python (host-only) package

jefferyto:
Maintainer: @lu-zero
Compile tested: armsr-armv7, 2023-10-07 snapshot sdk
Run tested: N/A

Description:
Python packages that use maturin to build do not call the maturin program directly; they use the [maturin build backend][1]. This build backend is a Python library provided with maturin that interfaces with the maturin program.

This changes the maturin package to use the Python build process so that the build backend is installed correctly.

This also renames the source package to python-maturin and moves it into the lang/python directory.

[1]: https://www.maturin.rs/#source-distribution

### DIFF
--- a/lang/python/python-maturin/Makefile
+++ b/lang/python/python-maturin/Makefile
@@ -4,40 +4,48 @@
 
 include $(TOPDIR)/rules.mk
 
-PKG_NAME:=maturin
-PKG_VERSION:=0.14.15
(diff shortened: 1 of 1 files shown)


lu-zero:
If it works better for building python programs sure :)

lu-zero (approved):

PR #22404, opened 2023-10-14
openconnect: add support for option --pfs

vgaetera:
Add support for the OpenConnect option `--pfs`.
Designed to require perfect forward secrecy.

Maintainer: @nmav


nmav (approved):
LGTM

PR #22427, opened 2023-10-17
modemmanager: bump to 1.22.0

Linaro1985:
Update modemmanager to 1.22.0 and necessary dependencies (libmbim and libqmi). 

- switch to builtin plugins.
- import upstream patch for fixing support of ublox LARA-R6001 / LARA-R6001D modules
  (see: https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/issues/779)

Maintainer: Nicholas Smith [nicholas@nbembedded.com](mailto:nicholas@nbembedded.com)

Compile tested:
on amd64 for mipsel_24kc on https://git.openwrt.org/openwrt/openwrt.git commit https://github.com/openwrt/packages/commit/7ed86d2eb7a58a6ed938b08fcacb4461a2876631

Run tested:
Zbtlink ZBT-WG3526


net/modemmanager/Makefile
```diff
@@ -64,6 +64,7 @@ MESON_ARGS += \
 	-Dintrospection=false \
 	-Dman=false \
 	-Dbash_completion=false \
+	-Dbuiltin_plugins=true \
```
feckert, line 67: I don't like the idea of including the plugins in the ModemManager binary. All plugins, whether they are needed or not, have to be loaded into RAM when the ModemManager is started. I think this is currently not the case for plugins that are not needed at runtime.

feckert (changes requested):
Also the ModemManager compilesl, but does not work.
Have a look at the [issue](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/issues/779) I have opened on the ModemManager page

net/modemmanager/Makefile
Linaro1985: The modemmanager author @aleksander0m wrote that this option works faster. Also, my total package size has become smaller.

Linaro1985:
> Also the ModemManager compilesl, but does not work

I have a Quectel EP06 modem. It worked on it. I'll look for other modems to test.

net/modemmanager/Makefile
aleksander0m: All plugins whether needed or not we're being loaded at runtime. Including all in the MM binary is much better, if there is no plan to allow third party plugins

feckert: Up to now, I think there is no plan to support third party plugins in openwrt. So I think this is OK. But I am little surprised, that the ModemManager is realy smaller, if the plugins are all build in! I would not have expect that.

feckert (approved):

feckert:
Thanks merged!

feckert:
@BKPepe   I know the first openwrt-23.05 was release last week, And I know that updating a package in a release is not desired, but the ModemManager  1.22 version was released 1 day after openwrt-23.05 release. Its not a major ModemManager update, but it has some nice new features and bug fixes.

BKPepe:
Go ahead! :)

PR #22440, opened 2023-10-18
atftpd: Add 'enable' config option

myszsoda:
Current version of atftpd daemon does automatically start when installed. This commit adds 'enable' option to config file to have control over atftpd daemon.

Maintainer: Daniel Danzberger <daniel@dd-wrt.com>
Compile tested: aarch64 / 2023-10-18 snapshot
Run tested: aarch64, raspberry Pi 4, 2023-10-18 snapshot

Description:
This patch disables atftpd daemon by default and adds config field to package settings


net/atftp/files/atftpd.conf
```diff
@@ -1,3 +1,4 @@
 
 config service 'service'
+	option en '0'
```
feckert, line 3: Please use enable if possible, that is more common?

net/atftp/files/atftpd.init
```diff
@@ -6,6 +6,10 @@ PIDFILE=/tmp/run/atftpd.pid
 
 start() {
```
feckert, line 7: Add a local variable there `local enable`

net/atftp/files/atftpd.init
```diff
@@ -6,6 +6,10 @@ PIDFILE=/tmp/run/atftpd.pid
...
 	config_load atftpd
+	config_get ENABLE service enable 0
```
feckert, line 9: Make variable `ENABLE` to lowercase `enable`. This is more common in this repository

myszsoda: Should I then refactor whole package? Since SRV and PORT and using uppercase also.

net/atftp/files/atftpd.init
myszsoda: Should I then refactor whole package? Since other variables in this script are not using local variables

net/atftp/files/atftpd.init
feckert: I didn't see that at all. I just looked at your change. If that's the case, then please yes.
Do this in a separate commit before your changes.

net/atftp/files/atftpd.init
myszsoda: Done

net/atftp/files/atftpd.init
myszsoda: Done.
I have not tested this new series but there shouldn't be possible in such simple script.
Implemented refactor + base change of this pull request.

feckert:
Sorry, but you still have to increase the [PKG_RELEASE](https://github.com/openwrt/packages/blob/master/net/atftp/Makefile#L11) by one, because something has changed.

feckert:
For whatever reason, you just have a conflict in the Makefile. Can you please rebase again.

feckert:
Something is wrong with a pull request, please take a look at the message from the CI.
https://github.com/openwrt/packages/actions/runs/6800258389/job/18555266562?pr=22440

feckert (approved):

PR #22445, opened 2023-10-18
softethervpn5: library update to December 2023 (vulnerability fixes)

tew42:
Maintainer: ???
Compile tested: compiled for ath79 - Archer a7 v5 - OpenWRT 23.05.0 using the SDK in a Debian WSL environment
Compile tested: compiled for ath79 - Archer a7 v5 - OpenWRT master using the SDK in a Debian WSL environment
Compile tested: compiled for x64 - OpenWRT 23.05.0 using the SDK in a Debian WSL environment
Compile tested: compiled for x64 - OpenWRT master using the SDK in a Debian WSL environment
Run tested: ath79 - Archer a7 v5 - OpenWRT 23.05.0 - VPN client tested
Run tested: x64 - Proxmox LXC - OpenWRT 23.05.0 - VPN client tested
Run tested: x64 - Proxmox LXC - OpenWRT master - VPN client tested

Description:
Here we go again... This is now the PR for master, with hopefully a correctly-formatted commit msg/sign-off. Should still be backported to 23.x due to security vulnerability fixes, PR to follow shortly.

The softethervpn5 package is due for an update from recent source. This PR implements a Makefile update to pull the October 15, 2023 github version, which includes fixes for recently-disclosed vulnerabilities. The build patches are also updated accordingly.

(Side note: I originally became interested in updating the package due to page faults on 22.x, as described in https://github.com/openwrt/packages/issues/20807 and https://github.com/openwrt/packages/pull/20850. That build was prevented by OpenSSL compatibility issues in 22.x toolchain, which are resolved in 23.x. Interestingly, though, the existing softethervpn5 package actually works fine on 23.x without page faults, so that issue seems to have been specific to only 22.x. Regardless the security vulnerability fixes still make this a worthwhile update to implement)


tew42:
... Bump?

tew42:
tagging @BKPepe since you had made the change requests on my initial PR, and also expressed some related interests over SoftEtherVPN/SoftEtherVPN#1916

BKPepe:
It looks like 5.02.5181 was released, this needs to be reworked.

1. Your commit subjects are too long!
2. Why do we need to have two commits. The second one could be squashed to the first one, right?

tew42:
> It looks like 5.02.5181 was released, this needs to be reworked.
> 
> 1. Your commit subjects are too long!
> 2. Why do we need to have two commits. The second one could be squashed to the first one, right?

Thanks, fixed/updated.

lortieav:
Hi tew42, I have successfully built the packages using your make and patch files using the OpenWRT 23.05.2 SDK. However, after installing the package on the system, I have noticed the softethervpn5 build version is still at 5180 in the source code, even though it builds using the 5181 source code. See attached screenshots.
capture1
Capture2

tew42:
As I understand it, the package manager descriptions are pulled from the OpenWRT package server, so I think the old build number in OPKG description is to be expected.
The self-reported version number is slightly more puzzling, but I'm 90% sure that this is due to hard-coded and not-updated build number [upstream](https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/CMakeSettings.json)

PR #22477, opened 2023-10-22
shadowsocks-rust: add new package

1715173329:
Maintainer: me
Compile tested: all targets
Run tested: ipq40xx/generic, mediatek/filogic, rockchip/armv8, x86/64 etc.

Description:
The rust port of shadowsocks, designed to replace the old libev implementation.

Ref: #22356

### DIFF
--- a/net/shadowsocks-rust/Makefile
+++ b/net/shadowsocks-rust/Makefile
@@ -1,0 +1,60 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2017-2020 Yousong Zhou <yszhou4tech@gmail.com>
+# Copyright (C) 2021-2023 ImmortalWrt.org
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=shadowsocks-rust
+PKG_VERSION:=1.17.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/shadowsocks/shadowsocks-rust/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=ef964ac60a499a0c8313d48284a7542c085f9c7672eb121a796b70d49163f35a
+
+PKG_MAINTAINER:=Tianling Shen <cnsztl@immortalwrt.org>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DEPENDS:=rust/host
+PKG_BUILD_PARALLEL:=1
+
+RUST_PKG_FEATURES:=aead-cipher-2022,local-http,local-redir,local-tun,local-tunnel
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/rust/rust-package.mk
+
+define Package/shadowsocks-rust/Default
+  define Package/shadowsocks-rust-$(1)
+    SECTION:=net
+    CATEGORY:=Network
+    SUBMENU:=Web Servers/Proxies
+    TITLE:=shadowsocks-rust $(1)
+    URL:=https://shadowsocks.org/
+    DEPENDS:=$$(RUST_ARCH_DEPENDS)
+  endef
+
+  define Package/shadowsocks-rust-$(1)/description
+    Shadowsocks is a fast tunnel proxy that helps you bypass firewalls.
+
+    This package contains the $(1).
+  endef
+
+  define Package/shadowsocks-rust-$(1)/install
+	$$(INSTALL_DIR) $$(1)/usr/bin
+	$$(INSTALL_BIN) $$(PKG_INSTALL_DIR)/bin/$(1) $$(1)/usr/bin/
+  endef
+endef
+
+SHADOWSOCKS_COMPONENTS:=sslocal ssmanager ssserver ssurl ssservice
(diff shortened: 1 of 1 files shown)


net/shadowsocks-rust/Makefile
```diff
@@ -0,0 +1,54 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2017-2020 Yousong Zhou <yszhou4tech@gmail.com>
+# Copyright (C) 2021-2023 ImmortalWrt.org
```
PolynomialDivision, line 4: Why?

1715173329: Hmmm I'm not sure what do you mean by "why".
This Makefile is modified from yousong's one, so included his copyright.

BKPepe:
Since I am not Makefile specialist for Rust, I think they are better people than me. 😅 @jefferyto what do you think? Will you have time to look at this? We do have gurus for Python packages, but not sure about Rust, though. Thinking about adding codeowners.

net/shadowsocks-rust/Makefile
PolynomialDivision: ah okay. forget it :) sorry

net/shadowsocks-rust/Makefile
```diff
@@ -0,0 +1,54 @@
...
+PKG_HASH:=ef964ac60a499a0c8313d48284a7542c085f9c7672eb121a796b70d49163f35a
+
+PKG_MAINTAINER:=Tianling Shen <cnsztl@immortalwrt.org>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DEPENDS:=rust/host
+PKG_BUILD_PARALLEL:=1
+
+RUST_PKG_FEATURES:=local-redir
```
jefferyto, line 23: I don't know shadowsocks too well, but would any of the other [optional features](https://github.com/shadowsocks/shadowsocks-rust#optional-features) be useful for users?

net/shadowsocks-rust/Makefile
```diff
@@ -0,0 +1,54 @@
...
+    SUBMENU:=Web Servers/Proxies
+    TITLE:=shadowsocks-rust $(1)
+    URL:=https://github.com/shadowsocks/shadowsocks-rust
+    DEPENDS:=$$(RUST_ARCH_DEPENDS)
+  endef
+
+  define Package/shadowsocks-rust-$(1)/install
+	$$(INSTALL_DIR) $$(1)/usr/bin
+	$$(INSTALL_BIN) $$(PKG_INSTALL_DIR)/bin/$(1) $$(1)/usr/bin/
+  endef
```
jefferyto, line 41: Would be nice to have a package description section as well.

jefferyto (commented):
Also would the init script/config file from shadowsocks-libev be useful for this package?

1715173329:
> Also would the init script/config file from shadowsocks-libev be useful for this package?

https://github.com/openwrt/packages/pull/22356#issuecomment-1756599482

net/shadowsocks-rust/Makefile
1715173329: shadowsocks is pretty simple and for users the package name is enough to tell what it is used for.
original libev one doesn't contain it too.

net/shadowsocks-rust/Makefile
1715173329: yes i guess, will think about adding more popular features.

net/shadowsocks-rust/Makefile
jefferyto: > shadowsocks is pretty simple and for users the package name is enough to tell what it is used for.

By this logic every package doesn't need a description because anyone that installs a package probably knows what it is for already.

> original libev one doesn't contain it too.

The original libev one should have had descriptions too. Just because a mistake was made in the past doesn't mean we have to continue making it.

jefferyto:
> > Also would the init script/config file from shadowsocks-libev be useful for this package?
> 
> [#22356 (comment)](https://github.com/openwrt/packages/pull/22356#issuecomment-1756599482)

> There're some incompatible arguments in the init script needs to be updated/checked, but I don't use that script anyway...

So how do you run the daemon (I assume something needs to be run to listen to connections) if not with that script? Would you expect other users to do it the same way?

1715173329:
> By this logic every package doesn't need a description because anyone that installs a package probably knows what it is for already.

I think so. It's just a simple proxy implementation. If you don't even know it, then it's basically - useless for you.

> Would you expect other users to do it the same way?

IMHO I don't care that. I have my own daemon, only for my personal purpose.
Those scripts from libev one are pretty complex, I'm not interested in wasting time on them.

I created this PR to see if there's any interest from the community, as @BKPepe suggested.
Feel free to take over it.

net/shadowsocks-rust/Makefile
1715173329: Added.

1715173329:
Closed for now.

satmandu:
Any chance of bringing this back? https://github.com/shadowsocks/shadowsocks-rust appears to be well supported now, receiving regular updates.

1715173329:
Yep it's well maintained, I closed this as I don't want to maintain those complex [daemon scripts](https://github.com/openwrt/packages/tree/openwrt-23.05/net/shadowsocks-libev/files).
If your router have enough spaces, try sing-box with its `auto_route`.

PR #22478, opened 2023-10-23
Update to 1.12.2

MicahZheng:

Description:


feckert:
The [commit message](https://github.com/openwrt/packages/pull/22478/commits/553adaee903422510137755334ac32119667f9bb) is not correct. Please add the name of the package as a prefix.
Have a look at the other commits in this repository. Also add your full name to the `Signed-of-by`

PR #22480, opened 2023-10-23
nerdctl: Version bump

lu-zero:
Maintainer: me
Compile tested: aarch64-musl
Run tested: bpi-r3

Description:


1715173329:
it would be better to take `nerdctl: update to 1.6.2` as the commit title

PR #22485, opened 2023-10-23
Modemmanager fix race condition

osedl:
Maintainer: @mips171. Please also have a look at this change @aleksander0m
Compile tested: x86_64, lantiq_xrx200 latest openwrt master
Run tested:  x86_64, lantiq_xrx200 latest openwrt master, test done = yes

On slower targets, it can happen that modems are not recognized correctly. During startup, the sysfs_cache is cleared, and then it waits for mmcli -L to complete. If a hotplug event occurs during this waiting time, the device attempts to report itself. However, the mmcli --report-kernel-event command fails, but it is still marked as "processed" in the sysfs cache. As a result, the modem is never further processed.
To address this issue, I propose two commits:

1. [remove sysfscache after dbus ready](https://github.com/openwrt/packages/commit/4125f74085dadc8bbc8a9133eda5246ab0a82d62)
2. [check status of report-kernel-event](https://github.com/openwrt/packages/commit/04e5ecbd7f0dff191a118d6653e021a439b9b8f6)

### DIFF
--- a/net/modemmanager/Makefile
+++ b/net/modemmanager/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=modemmanager
 PKG_SOURCE_VERSION:=1.22.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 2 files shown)


aleksander0m (commented):
Both commits seem reasonable from my POV

aleksander0m:
@feckert should approve this

PR #22486, opened 2023-10-23
modemmanager improve cleanup

osedl:
Maintainer: @mips171. Please also have a look at this change @aleksander0m
Compile tested: x86_64, lantiq_xrx200 latest openwrt master
Run tested: x86_64, lantiq_xrx200 latest openwrt master, test done = yes

There are situations where the cleanup of interfaces doesn't work as expected. For instance, when ModemManager crashes, cleanup is not triggered. Even if cleanup were to be triggered, the interfaces cannot be queried because the mmcli command is no longer available after a crash. To address this issue, I propose two commits:
1. [ improve cleanup of ifaces](https://github.com/openwrt/packages/commit/0d7a557462702617f73cb17b34aa2e2149ed3151)
2. [move iface cleanup to wrapper script](https://github.com/openwrt/packages/commit/c49aff8595dfaa765f8049dc59bee597eff64d56)

### DIFF
--- a/net/modemmanager/files/modemmanager.common
+++ b/net/modemmanager/files/modemmanager.common
@@ -235,24 +235,20 @@ mm_report_modem_wait() {
 # Cleanup interfaces
 
 mm_cleanup_interfaces() {
-	local modemlist modemlength idx modeminfo modemsysfspath
+	local sysfs_path status
 
-	modemlist=$(mmcli --list-modems --output-keyvalue)
-	[ -n "${modemlist}" ] || return 0
+	# Do nothing if there is no sysfs cache
+	[ -f "${MODEMMANAGER_SYSFS_CACHE}" ] || return
 
-	modemlength=$(modemmanager_get_field "${modemlist}" "modem-list.length")
+	while IFS= read -r sysfs_cache_line; do
+		sysfs_path=$(echo "${sysfs_cache_line}" | awk '{print $1}')
+		status=$(echo "${sysfs_cache_line}" | awk '{print $2}')
 
-	# do nothing if no modem reported
-	[ -n "${modemlength}" ] && [ "${modemlength}" -ge 1 ] && {
-		idx=1
-		while [ $idx -le "$modemlength" ]; do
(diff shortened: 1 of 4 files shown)


net/modemmanager/files/modemmanager.common
```diff
@@ -235,24 +235,20 @@ mm_report_modem_wait() {
...
+
+	# Do nothing if there is no sysfs cache
+	[ -f "${MODEMMANAGER_SYSFS_CACHE}" ] || return
+
+	while IFS= read -r sysfs_cache_line; do
+		sysfs_path=$(echo "${sysfs_cache_line}" | awk '{print $1}')
+		status=$(echo "${sysfs_cache_line}" | awk '{print $2}')
+
+		if [ "${status}" = "processed" ]; then
+			mm_log "info" "call cleanup for: ${sysfs_path}"
```
aleksander0m, line 248: I don't think this should be "info" level

net/modemmanager/files/usr/sbin/ModemManager-wrapper
```diff
@@ -28,6 +28,9 @@ main() {
 	mm_report_events_from_cache
 
 	wait "$CHILD"
+
+	# Set all configured interfaces as unavailable
+	mm_cleanup_interfaces
```
aleksander0m, line 33: There is another `mm_cleanup_interfaces` call just before launching the MM daemon. Was that call doing anything before these changes, given that it required `mmcli` to work?

aleksander0m (commented):
I think the changes are reasonable, but @feckert should review and approve.

net/modemmanager/files/modemmanager.common
feckert: Yes this should be a "debug" level

net/modemmanager/files/usr/sbin/ModemManager-wrapper
osedl: Since the MM daemon should be down at this point, the call likely did nothing before

PR #22501, opened 2023-10-25
Pulseaudio: update to 16.1

jmv2009:
Maintainer: @neheb
Compile tested:armsr/armv8
Run tested:armsr/armv8

Description:
Update to Pulseaudio 16.1 (latest)
For getting through meson, gio patch is generated (obsolete on head of pulseaudio)
https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/commit/de8b0c11242a49c335abdae292d0bb9f6d71d2dd
For getting through meson, bluez-daemon dependency is generated.
Obsolete patches removed.
-Ddoxygen passed.


sound/pulseaudio/Makefile
```diff
@@ -30,7 +30,7 @@ define Package/pulseaudio-daemon/Default
   SECTION:=sound
   CATEGORY:=Sound
   DEPENDS:=+libsndfile +libltdl +libpthread +librt +alsa-lib \
-	+libopenssl +libcap $(ICONV_DEPENDS) $(INTL_DEPENDS)
+	+libopenssl +libcap +glib2 $(ICONV_DEPENDS) $(INTL_DEPENDS)
```
neheb, line 33: $(ICONV_DEPENDS) $(INTL_DEPENDS) come with glib2. no need for them now.

sound/pulseaudio/patches/010-iconv.patch
```diff
@@ -10,6 +10,8 @@
 +if not have_iconv and cc.has_function('iconv_open')
```
neheb, line 10: I don't think this patch is needed anymore.

sound/pulseaudio/patches/020-doxygen.patch
```diff
@@ -1,10 +1,10 @@
...
- 
+@@ -827,7 +827,6 @@ check_dep = dependency('check', version
  # Subdirs
  
--subdir('doxygen')
- subdir('po')
- if get_option('man')
-   subdir('man')
+ if get_option('doxygen')
+-  subdir('doxygen')
```
neheb, line 7: this is definitely not needed anymore.

sound/pulseaudio/Makefile
neheb: +libpthread +librt are also not needed too.

neheb: libltdl might not be needed.

jmv2009: Glib2 (GIO) was needed for it to build.

sound/pulseaudio/patches/010-iconv.patch
jmv2009: I'll check.

sound/pulseaudio/patches/020-doxygen.patch
jmv2009: This patch was ready for it to build somehow.

neheb: Pass -Ddoxygen=false

sound/pulseaudio/Makefile
BKPepe: That should be specified in the commit description!

BKPepe:
I am really sad to see that this was merged without proper review. Not sure, what to think about this. :(
Commit subject says update, you would expect usual update, but while going through the commit, it contains so many things. No description, no one will know in the future why the hell it was done like this.

Please stop doing this, this is the worst approach ever. I am really thinking about reverting this commit and that it should be redone.

sound/pulseaudio/Makefile
neheb: glib2 was patched out.

jmv2009:
@BKPepe: Thank you for your comments.
To clarify: There were a couple of issues upgrading to Pulseaudio 16.1.
a) The doxygen patch was not working anymore. They changed the meson.build code to allow -Ddoxygen=disabled to be passed. So we pass -Ddoxygen instead, deleting the associated patch.
b) The gio-2.0 (part of glib2) is not required. However, due to a modification in the meson system, the dependency check stops the system when it fails. A patch already upstreamed already in the head of master of Pulseaudio was therefore also applied here: https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/commit/de8b0c11242a49c335abdae292d0bb9f6d71d2dd (in my initial version, I had included glib2 instead)
c) @Neheb clarified that +libpthread +librt were not required anymore, so they were taken out.
d) -Dbluez required enabled/disabled instead of true/false
e) @Nehab said that the 010-iconv.patch was not required anymore, so it was deleted.
f) Meson of Pulseaudio now requires bluez5 to be available when building against it. Therefore blues-daemon was included as a dependency.
g) As of Pulseaudio 16.1, the directory of the files changed to include "pulseaudio" instead of a directory name with the package version included.
https://www.freedesktop.org/wiki/Software/PulseAudio/Notes/16.0/
Thank you for your review, and please propose changes and problem areas, if needed.

PR #22514, opened 2023-10-26
i2pd: Bumped version to 2.49.0 and updated sha256 hash

sysfu:
Maintainer: @yangfl 
Compile tested: arch = mvebu/cortexa53 cpu model = ARMv8 Processor rev 4 OpenWRT = 22.03.4
Run tested: arch = mvebu/cortexa53 cpu model = ARMv8 Processor rev 4 OpenWRT = 22.03.4

Description: Update to version 2.49.0


BKPepe:
Your's master branch is too old and there is a conflict to prevent this to be merged.

Also, there are formality issues.
1. The commit subject should start with the PKG_NAME this in this case i2pd
2. The ``Signed-off-by`` should contain your real name.

PR #22516, opened 2023-10-26
iperf3-mt: new package

jonasjelonek:
This adds a multithreaded variant of iperf3 as a package. This variant is still experimental, developed in the mt branch of the iperf repository and expected to be merged when it is considered stable.

Maintainer: me / @jonasjelonek
Compile tested: aarch64_cortex-a53 (BananaPi R64) + x86/64 (APU2C4), OpenWrt snapshot
Run tested: aarch64_cortex-a53 (BananaPi R64) + x86/64 (APU2C4), OpenWrt snapshot

Description:
The main iperf3 that we already have as a package, has no support for multithreading yet. However, the developers are working already on that in the `mt` branch of the [iperf Github repository](https://github.com/esnet/iperf/tree/mt) and release this as 3.XX-mt versions. This multithreaded iperf3 is still considered experimental, however, in my tests I couldn't notice any instabilities so far and the load is distributed across multiple CPU cores when using multiple streams.

Starting point for this Makefile is the one of the normal iperf3. I switched to Github codeload because the developers only publish non-beta versions of the multithreaded variant in the other one. This PR starts with `3.15-mt-beta1`, the latest non-beta published in the location that the `iperf3` package uses is `3.13-mt1`. I think we can go with this beta to not lack behind the non-MT version.

### DIFF
--- a/net/iperf3-mt/Makefile
+++ b/net/iperf3-mt/Makefile
@@ -1,0 +1,104 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2023 Jonas Jelonek
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=iperf
+PKG_VERSION:=3.15-mt-beta1
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/esnet/iperf/tar.gz/$(PKG_NAME)-$(PKG_VERSION)?
+PKG_HASH:=4d5ad5bef9321adb832581a495c3cb1b5dec9d9678296f90bfc87166bbb7a43b
+
+PKG_MAINTAINER:=Jonas Jelonek <jelonek.jonas@gmail.com>
+PKG_LICENSE:=BSD-3-Clause
+PKG_CPE_ID:=cpe:/a:es:iperf3
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_NAME)-$(PKG_VERSION)
+PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=1
+
+PKG_FIXUP:=autoreconf
+
+include $(INCLUDE_DIR)/package.mk
+
+DISABLE_NLS:=
+
+define Package/iperf3-mt/default
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=iperf3 with multithreading
+  URL:=https://github.com/esnet/iperf
+  CONFLICTS:=iperf3 iperf3-ssl
+endef
+
+define Package/iperf3-mt
+$(call Package/iperf3-mt/default)
+  VARIANT:=nossl
+  DEPENDS:=+libiperf3-mt
+  CONFLICTS+=iperf3-mt-ssl
+endef
+
+define Package/iperf3-mt-ssl
+$(call Package/iperf3-mt/default)
+  TITLE+= and iperf_auth support
+  VARIANT:=ssl
+  DEPENDS:=+libopenssl +libatomic
+endef
+
+define Package/libiperf3-mt
+  SECTION:=libs
+  CATEGORY:=Libraries
+  TITLE:=libiperf3 with multithreading
+  URL:=https://github.com/esnet/iperf
+  CONFLICTS:=libiperf3
+  DEPENDS+=+libatomic
+endef
+
+TARGET_CFLAGS += -D_GNU_SOURCE
+TARGET_LDFLAGS += -latomic
+
+ifeq ($(BUILD_VARIANT),ssl)
+	CONFIGURE_ARGS += --with-openssl="$(STAGING_DIR)/usr" --disable-shared
+else
+	CONFIGURE_ARGS += --without-openssl
+endif
+
+MAKE_FLAGS += noinst_PROGRAMS=
+
+define Package/iperf3-mt/description
+ iPerf3 is a modern alternative for measuring TCP and UDP bandwidth
+ performance, allowing the tuning of various parameters and
+ characteristics.
+ iperf3-mt has experimental multithreading support.
+endef
+
+define Package/libiperf3-mt/description
+ Libiperf is a library providing an API for iperf3 functionality.
+endef
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libiperf.* $(1)/usr/lib/
+	$(INSTALL_DIR) $(1)/usr/include
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/* $(1)/usr/include/
+endef
+
+define Package/iperf3-mt/install/Default
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/iperf3 $(1)/usr/bin/
+endef
+
+Package/iperf3-mt/install = $(Package/iperf3-mt/install/Default)
+Package/iperf3-mt-ssl/install = $(Package/iperf3-mt/install/Default)
+
+define Package/libiperf3-mt/install
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libiperf.so.* $(1)/usr/lib
+endef
+
+$(eval $(call BuildPackage,iperf3-mt))
+$(eval $(call BuildPackage,iperf3-mt-ssl))
+$(eval $(call BuildPackage,libiperf3-mt))


jonasjelonek:
cc @PolynomialDivision

net/iperf3-mt/Makefile
```diff
@@ -0,0 +1,100 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2023 Jonas Jelonek
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=iperf
+PKG_VERSION:=iperf-3.15-mt-beta1
```
BKPepe, line 8:
```suggestion
PKG_VERSION:=3.15-mt-beta1
```

net/iperf3-mt/Makefile
```diff
@@ -0,0 +1,100 @@
...
+  CATEGORY:=Network
+  TITLE:=iperf3 with multithreading
+  URL:=https://github.com/esnet/iperf
+  CONFLICTS:=iperf3 iperf3-ssl
+endef
+
+define Package/iperf3-mt
+$(call Package/iperf3-mt/default)
+  VARIANT:=nossl
+  DEPENDS:=+libiperf3-mt
```
BKPepe, line 39: There should be conflict for iperf3-mt-ssl

net/iperf3-mt/Makefile
```diff
@@ -0,0 +1,100 @@
...
+endef
+
+define Package/libiperf3-mt/description
+ Libiperf is a library providing an API for iperf3 functionality.
+endef
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(INSTALL_DIR) $(1)/usr/include
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libiperf.* $(1)/usr/lib/
```
BKPepe, line 81: Can you move this row just above? So we have it like folder, file, folder, file? Thanks!

BKPepe:
This one fails for PowerPC and also for mips (same error):

```
OpenWrt-libtool: link: powerpc-openwrt-linux-musl-gcc -g -Os -pipe -mcpu=464fp -fno-caller-saves -fno-plt -fhonour-copts -ffile-prefix-map=/builder/build_dir/target-powerpc_464fp_musl/iperf-nossl/iperf-iperf-3.15-mt-beta1=iperf-iperf-3.15-mt-beta1 -Wformat -Werror=format-security -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z -Wl,now -Wl,-z -Wl,relro -D_GNU_SOURCE -Wall -g -fuse-ld=bfd -znow -zrelro -o .libs/iperf3 iperf3-main.o  -L/builder/staging_dir/toolchain-powerpc_464fp_gcc-12.3.0_musl/usr/lib -L/builder/staging_dir/toolchain-powerpc_464fp_gcc-12.3.0_musl/lib ./.libs/libiperf.so -lpthread -pthread
/builder/staging_dir/toolchain-powerpc_464fp_gcc-12.3.0_musl/bin/../lib/gcc/powerpc-openwrt-linux-musl/12.3.0/../../../../powerpc-openwrt-linux-musl/bin/ld.bfd: ./.libs/libiperf.so: undefined reference to `__atomic_store_8'
/builder/staging_dir/toolchain-powerpc_464fp_gcc-12.3.0_musl/bin/../lib/gcc/powerpc-openwrt-linux-musl/12.3.0/../../../../powerpc-openwrt-linux-musl/bin/ld.bfd: ./.libs/libiperf.so: undefined reference to `__atomic_fetch_add_8'
/builder/staging_dir/toolchain-powerpc_464fp_gcc-12.3.0_musl/bin/../lib/gcc/powerpc-openwrt-linux-musl/12.3.0/../../../../powerpc-openwrt-linux-musl/bin/ld.bfd: ./.libs/libiperf.so: undefined reference to `__atomic_load_8'
```

jonasjelonek:
> This one fails for PowerPC and also for mips (same error):
> 
> ```
> OpenWrt-libtool: link: powerpc-openwrt-linux-musl-gcc -g -Os -pipe -mcpu=464fp -fno-caller-saves -fno-plt -fhonour-copts -ffile-prefix-map=/builder/build_dir/target-powerpc_464fp_musl/iperf-nossl/iperf-iperf-3.15-mt-beta1=iperf-iperf-3.15-mt-beta1 -Wformat -Werror=format-security -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z -Wl,now -Wl,-z -Wl,relro -D_GNU_SOURCE -Wall -g -fuse-ld=bfd -znow -zrelro -o .libs/iperf3 iperf3-main.o  -L/builder/staging_dir/toolchain-powerpc_464fp_gcc-12.3.0_musl/usr/lib -L/builder/staging_dir/toolchain-powerpc_464fp_gcc-12.3.0_musl/lib ./.libs/libiperf.so -lpthread -pthread
> /builder/staging_dir/toolchain-powerpc_464fp_gcc-12.3.0_musl/bin/../lib/gcc/powerpc-openwrt-linux-musl/12.3.0/../../../../powerpc-openwrt-linux-musl/bin/ld.bfd: ./.libs/libiperf.so: undefined reference to `__atomic_store_8'
> /builder/staging_dir/toolchain-powerpc_464fp_gcc-12.3.0_musl/bin/../lib/gcc/powerpc-openwrt-linux-musl/12.3.0/../../../../powerpc-openwrt-linux-musl/bin/ld.bfd: ./.libs/libiperf.so: undefined reference to `__atomic_fetch_add_8'
> /builder/staging_dir/toolchain-powerpc_464fp_gcc-12.3.0_musl/bin/../lib/gcc/powerpc-openwrt-linux-musl/12.3.0/../../../../powerpc-openwrt-linux-musl/bin/ld.bfd: ./.libs/libiperf.so: undefined reference to `__atomic_load_8'
> collect2: error: ld returned 1 exit status
> ```

Yes, I already have a fix for this issue, libatomic is missing in TARGET_LDFLAGS and as dependency for libiperf3-mt. Will force-push this soon.

net/iperf3-mt/Makefile
jonasjelonek: Just this change doesn't work, I tried that before. The file that is downloaded from Github codeload seems to always have a name like `iperf-iperf-3.15-mt-beta1`, when extracted the directory will have the same name.
This means, the source is extracted to `iperf-iperf-3.15...` but the PKG_BUILD_DIR is generated by makefiles like `iperf-3.15...`, then it fails to build because it doesn't find the source.

Would you suggest to keep it as I did or overwrite PKG_BUILD_DIR?

net/iperf3-mt/Makefile
jonasjelonek: I basically agree, then we should also include this in the normal iperf3, cause there it's also missing :)

net/iperf3-mt/Makefile
BKPepe: Yes, you should rewrite PKG_BUILD_DIR, that's common here, where it is as you said PKG_NAME-PKG_NAME-PKG_VERSION.

net/iperf3-mt/Makefile
BKPepe: Yup.

net/iperf3-mt/Makefile
jonasjelonek: done, thanks for the hint!

net/iperf3-mt/Makefile
jonasjelonek: done, thanks!

net/iperf3-mt/Makefile
jonasjelonek: done

jonasjelonek:
@BKPepe 
CI fails for ssl variant, seems to be caused by an issue I already noticed with the normal iperf3. 

When the ssl variant is selected but libiperf3 is still selected (e.g. when nossl variant is deselected, libiperf3 doesn't get deselected automatically), the `--disable-shared` is added to the CONFIGURE_ARGS. I guess this then prevents the creation of `libiperf3.so.*`, causing this issue.

Can't we just remove this`--disable-shared`. As far as I understand this was introduced for iperf3 to avoid shipping two libiperf.so.* versions. Because I added the conflict between `iperf3-mt` and `iperf3-mt-ssl` as per your review, could this still be a problem anymore?

jonasjelonek:
Nevermind my previous message, CI succeeds now.

However the issue I described is still present, and comes up when previously iperf3(-mt) was selected, but then one deselects it, selects iperf3(-mt)-ssl and does not deselect libiperf3(-mt).

May I account for this in this PR or in a separate for both iperf3 and iperf3-mt?

BKPepe:
> When the ssl variant is selected, --disable-shared is added to the CONFIGURE_ARGS. When libiperf3 is selected at the same time, this seems to prevent the creation of libiperf3.so.*, causing the build to fail with the error libiperf.so.* not found.

I believe that this same issue is already present in iperf3. For now, I propose to create issue in this repository about it. So we would not forget about it and someone including me can look at it. I am not able to look at it now, sorry! I am quite busy with other things.

net/iperf3-mt/Makefile
```diff
@@ -0,0 +1,104 @@
...
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/* $(1)/usr/include/
+endef
+
+define Package/iperf3-mt/install/Default
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/iperf3 $(1)/usr/bin/
+endef
+
+Package/iperf3-mt/install = $(Package/iperf3-mt/install/Default)
+Package/iperf3-mt-ssl/install = $(Package/iperf3-mt/install/Default)
```
BKPepe, line 95: I am thinking about this. It might be nitpicking and I know the intention, why it is done like this, but wouldn't it be better to do it like it is done in iperf3? I mean, we will add one more row, but otherwise it might be better to read.

My proposal:
```
define Package/iperf3-mt/install
	$(INSTALL_DIR) $(1)/usr/bin
	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/iperf3 $(1)/usr/bin/
endef

define Package/iperf3-mt-ssl/install
	$(INSTALL_DIR) $(1)/usr/bin
	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/iperf3 $(1)/usr/bin/
endef
```

Nothing wrong about this, if you dont want it, I am fine with that.

BKPepe (approved):
Yeah, we reached the end here. 👍 Thanks. If there are no objections within 48hours, I will merge this. This looks good.

net/iperf3-mt/Makefile
jonasjelonek: I have no clear opinion on this. 
I swa that in the `mtr` Makefile, and it looked a bit cleaner to me :), but that might be quite subjective.

diizzyy:
While I haven't done any benchmarking at all I would guess that it makes sense to force ```-O2``` in this case if it improves performance that is.

jonasjelonek:
> While I haven't done any benchmarking at all I would guess that it makes sense to force `-O2` in this case if it improves performance that is.

I haven't done either so far. Probably this can in a separate PR, if I understand this correctly it could also be discussed for the normal iperf3.

jonasjelonek:
ping @BKPepe

PR #22523, opened 2023-10-28
owntone: Update to 28.8

ejurgensen:
Maintainer: me
Compile tested: x86, VM, OpenWrt trunk
Run tested: x86, VM, OpenWrt trunk

Description:
Update to 28.8, see [changelog](https://raw.githubusercontent.com/owntone/owntone-server/master/ChangeLog)

### DIFF
--- a/sound/owntone/Makefile
+++ b/sound/owntone/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=owntone
-PKG_VERSION:=28.5
+PKG_VERSION:=28.8
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://github.com/owntone/owntone-server/releases/download/$(PKG_VERSION)/
-PKG_HASH:=c9ee0152dc488f782a25a68e72d24c109882bef3dd2914315fe499c8415fd898
+PKG_HASH:=ebaee52ae617f08c41859522ba0a839d1865dcac7d6c0eb9e3fee81caf8fd47c
 
 PKG_FIXUP:=autoreconf
 PKG_BUILD_FLAGS:=no-mips16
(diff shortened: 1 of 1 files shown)


BKPepe:
Link to changelog, please

ejurgensen:
Sure: [ChangeLog](https://raw.githubusercontent.com/owntone/owntone-server/master/ChangeLog)

sound/owntone/Makefile
```diff
@@ -32,7 +32,7 @@ CATEGORY:=Sound
 TITLE:=iTunes (DAAP) server for Apple Remote and AirPlay
 URL:=https://github.com/owntone/owntone-server
 DEPENDS:=+libgpg-error +libgcrypt +libgdbm +zlib +libexpat +libunistring \
-	+libevent2 +libdaemon +confuse +alsa-lib +libffmpeg-full \
+	+libevent2 +libevent2-pthreads +libdaemon +confuse +alsa-lib +libffmpeg-full \
```
BKPepe, line 35: Why do we need new dependency?

sound/owntone/Makefile
```diff
@@ -54,8 +54,7 @@ CONFIGURE_ARGS += \
 	--disable-install_conf_file \
 	--disable-install_user \
 	--with-alsa \
-	--without-pulseaudio \
-	--without-libevent_pthreads
```
BKPepe, line 58: Why do you need to have it enabled?

BKPepe (changes requested):
Please stop slipping changes, which are not from upstream. Your commit subject as there is no commit description, not even in the pull request mentioned. 

This looks like usual update on the first sight and then you are enabling some stuff and no one knows why and it is not even in their changelog: https://github.com/owntone/owntone-server/releases/tag/28.8

ejurgensen:
> Please stop slipping changes

What do you mean by "stop slipping changes"? Is this something I have done before? I have added the "libevent2-pthreads" dependency simply because the build failed without it. That's all there is to it.

Maybe it's not your intent, but your feedback comes of as somewhat hostile. I am just keeping the package updated - as I suppose the primary duty of a maintainer is. I don't have ulterior motives by adding the dependency. What would they even be?

I get a bunch of PRs myself, and I understand the dread that can come with it, but remember that people are just trying to help.

BKPepe:
>  I have added the "libevent2-pthreads" dependency simply because the build failed without it. That's all there is to it.

Was it working previously without it? I can see that owntone is sucessfully compiled on OpenWrt buildbots, thus there is something wrong.

Torxgewinde:
Probably it is obvious to everyone, but to make sure i'd like to point out it is the main owntone author that is also maintaining the OpenWRT package here.

ejurgensen:
> Was it working previously without it? I can see that owntone is sucessfully compiled on OpenWrt buildbots, thus there is something wrong.

The dependency came in version 28.7, so that's why 28.5 can be built without. The changelog doesn't directly say it, but it is part of the "drop support for libevent < 2.1.4", ref. [this PR](https://github.com/owntone/owntone-server/pull/1576): "Support for libevent < 2.1.4 has been removed, and libevent pthreads is now a hard requirement." The configure option `--without-libevent_pthreads` also no longer exists.

BKPepe:
Okay, this one should be copy&pasted to the commit description.

Dne út 31. 10. 2023 1:47 uživatel ejurgensen ***@***.***>
napsal:

ejurgensen:
Ok, I have amended the commit message now

PR #22529, opened 2023-10-28
aircrack-ng: fix wrong inclusion of libbsd if detected

Ansuel:
Currently aircrack-ng try to link with libbsd if it does detect the library in staging_dir. This is the case with buildbot where every package is selected and compiled.

Fix this by adding a pending patch that permits to disable libbsd inclusion even if detected and set the related config flag.

aircrack-ng use 2 function of libbsd and it's not worth to include the entire library for 2 simple function for string manipulation.

Maintainer: Rick Farina <zerochaos@gentoo.org>


neheb:
I bet those two functions come with musl.

Ansuel:
@neheb good hint... indeed musl and glibc from one version and up also ship with the functions included. I added an extra patch to account for this!

PR #22532, opened 2023-10-29
freeradius3: switch to pcre2

crza:
switch to PCRE2 implementation https://github.com/openwrt/packages/issues/22006

Maintainer: @odmdas / @neheb
Compile tested: tested on mediatek unifi-6-lr-v1 with musl openwrt 23.05
Run tested: tested on mediatek unifi-6-lr-v1 with musl openwrt-23.05 with sample configuration (& internal setup ...)

Description:
libpcre is EOL with no further (security) fixes.
use libpcre2 instead

### DIFF
--- a/net/freeradius3/Makefile
+++ b/net/freeradius3/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freeradius3
 PKG_VERSION:=3.0.26
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=freeradius-server-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=https://github.com/FreeRADIUS/freeradius-server/releases/download/release_$(subst .,_,$(PKG_VERSION))/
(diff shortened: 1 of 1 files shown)


BKPepe:
Formality check is failing for you:
```
=== Checking commit '17f45bf5a6c26f8b860e946b8ef6f01370554845'
: Author name (crza) need to be your real name 'firstname lastname'
: Commit subject line seems ok (freeradius3: fix compile errors for pcre2)
: Signed-off-by is missing or doesn't match author (should be 'Signed-off-by: crza <crza@users.noreply.github.com>')
```

crza:
should be fixed.

BKPepe (changes requested):
Excuse me, but I don't understand the pull request description.

> libpcre is not available, compile fails. fix this by using libpcre2 instead.

What?  Do you have any output for those compile errors, which you mention?

Freeradius is actually correctly build and there are [no faillogs](https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a53/)

____

Then I am going back to your commit. Commit subject was said above.
Commit description should be imho revised. Please see it here: https://openwrt.org/submitting-patches
Use PCRE2, its okay, but it does not say why... which is important part here.

crza:
well ...
after repeating my steps from saturday (with fresh pulls from openwrt-23.05 branches) there is one difference in .config today (with a successful build): CONFIG_PACKAGE_libpcre=y exists, which was missing in saturdays .config

I will amend my commit & PR.

BKPepe:
Much better. :clap:

clutch2sft:
19ec302 also breaks [aarch64_cortex-a53](https://downloads.openwrt.org/releases/faillogs-23.05/aarch64_cortex-a53/)
image

BKPepe:
We know, we are working to fix it soon.

Dne po 6. 11. 2023 13:27 uživatel clutch2sft ***@***.***>
napsal:

clutch2sft:
@DKPepe Thank you.  I wasn't sure from reading the thread about the broad nature of the problem.  And thank you for working diligently to resolve it.  I also have a question if I may about the mechanics of this.

In my world a release is a point in time and it is never changed.  So 23.05.0 wouldn't revieve such change and it would come in the next stable release.  Indeed I see the commit was made to the master.  How did it get back-ported to the release?  To be clear this isn't a critisism - just a way to clarify my understanding of the release strategy for ongoing use.

jefferyto:
The build error was fixed in #22666 (#22670 for 23.05).

PR #22534, opened 2023-10-29
zabbix: move to PCRE2 library

Ansuel:
Move to PCRE2 library as PCRE is not EOL and won't receive any security updates anymore.

Maintainer: Etienne CHAMPETIER <champetier.etienne@gmail.com> @champtar


champtar:
If it builds let's merge this
Thanks @Ansuel !!

PR #22541, opened 2023-10-29
pulseaudio: disable compiling against libwrap (was Include +libwrap dependency)

jmv2009:
Reportedly was inadvertendly being compiled against libwrap per https://github.com/openwrt/packages/commit/aafc57c1ded5450419b9690056163acf3d19d51f#comments

Maintainer: @neheb

If libraries/libwrap (tcp wrapper library) was built before pulseaudio, 
Libwrap was detected at the meson configuration stage of pulseaudio and used since this commit:
https://github.com/pulseaudio/pulseaudio/commit/156e16f941789c5d53cdb0c46be480c64e3149f1
However, pulseaudio can not find libwrap.so.0 during the openwrt building process at some stage. 
Therefor, compilation against libwrap is disabled.

### DIFF
--- a/sound/pulseaudio/Makefile
+++ b/sound/pulseaudio/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=pulseaudio
 PKG_VERSION:=16.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://freedesktop.org/software/pulseaudio/releases
@@ -122,7 +122,9 @@ MESON_ARGS += \
 	-Dx11=disabled \
 	-Dadrian-aec=true \
 	-Dwebrtc-aec=disabled \
-        -Ddoxygen=false
+        -Ddoxygen=false \
+        -Dtcpwrap=disabled \
+        -Dbluez5-gstreamer=disabled 
 
 ifeq ($(BUILD_VARIANT),avahi)
 MESON_ARGS += \


BKPepe:
In your local build env, you need first compile the dependency, if you do have minimal build, which you apparently has or you need to have full build, then you need to compile pulseaudio to be able to reproduce the issue.

BKPepe:
Nitpick: it's not include, but add dependency and also your commit should have Fixes tag in the commit description.

jmv2009:
@bkpepe:
Can you spell out your proposed way of reproducing?
Is it a minimal build with makeconfig first.
Then build it again include pulseaudio-daemon included?

BKPepe:
My suggestion:
1. First compile libwrap - ``make package/libwrap/compile -jX V=s``
2. Then Compile pulseadio - ``make package/pulseaudio/{clean,compile} -jX V=s``
Clean should be not needed, but its better. No leftovers from previous build.

Hints:
``V=s`` - verbose output
``jX`` - how many threads will it use for compilation

jmv2009:
I can confirm that before this commit, pulseaudio was incompatible with libraries/libwrap (feeds/packages/libs/tcp_wrappers). Looking for other solutions.

BKPepe:
The commit is not correct so far.

1. You were able to reproduce the issue that adding the dependency works. Do you know why it is required? The feature which requires libwrap can not be disabled?
2. The commit subject should be rewritten. 
E.g. pulseaudio: ``add libwrap dependency``
3. In the commit description, there should be why.
Also, the Fixes tag is not correct. 

It should look like this: ``Fixes: hash ("commit subject")``. Should be near the SoB.

jmv2009:
@BKpepe : please review proposed changes then.

sound/pulseaudio/patches/020-disable_lib_tcp_wrapper_usage.patch
```diff
@@ -0,0 +1,12 @@
+--- a/meson.build
++++ b/meson.build
+@@ -733,9 +733,6 @@ if get_option('daemon')
+   endif
+ 
+   tcpwrap_dep = cc.find_library('wrap', required: get_option('tcpwrap'))
+-  if cc.has_header('tcpd.h') and cc.has_function('hosts_access', dependencies : tcpwrap_dep)
```
neheb, line 7: Why not just pass -Dtcpwrap=disabled

jmv2009: Done...

neheb:
I’ll let @BKPepe deal with this

egorenar:
i tested this commit, looks good.

jmv2009:
Fixed commit message.

jmv2009:
Pulseaudio packages are currently not building due to two other libraries not linking. Investigating.

jmv2009:
After merging this the linking problems disappeared in the fail log, and everything appears fine now.

PR #22545, opened 2023-10-29
gensio: update to 2.7.6

mhei:
Also disable probably rarely used features to
lower the library footprint.

Maintainer: @WereCatf 
Compile tested: mxs, Duckbill & Raspberry Pi
Run tested: mxs, Duckbill & Raspberry Pi (both with ser2net)


mhei:
@WereCatf : objections?

mhei:
Yes, I missed this. Will create a follow-up PR.

Issue #22330, opened 2023-10-07
python3-lxml: trying to link host python3 library


egorenar:
Maintainer: @Alexandru Ardelean <ardeleanalex@gmail.com>
Environment: ARMv7 Processor rev 4 (v7l), OpenWrt SNAPSHOT r24099-5dfd8dc7d4f5

python3-lxml fails to build because it tries to link host's Python3 library.
The compiler is using `-L/usr/lib`.

```
make V=sc -j$(($(nproc)+1)) package/feeds/packages/python-lxml/{clean,compile}
...
arm-openwrt-linux-muslgnueabi-gcc -shared -L/home/egorenar/Repositories/openwrt-rel/staging_dir/toolchain-arm_cortex-a15+neon-vfpv4_gcc-12.3.0_musl_eabi/usr/lib -L/home/egorenar/Repositories/openwrt-rel/staging_dir/toolchain-arm_cortex-a15+neon-vfpv4_gcc-12.3.0_musl_eabi/lib -fuse-ld=bfd -znow -zrelro -lpython3.11 -Os -pipe -fno-caller-saves -fno-plt -fhonour-copts -mfloat-abi=hard -fmacro-prefix-map=/home/egorenar/Repositories/openwrt-rel/build_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/pypi/lxml-4.9.3=lxml-4.9.3 -Wformat -Werror=format-security -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro -I/home/egorenar/Repositories/openwrt-rel/staging_dir/toolchain-arm_cortex-a15+neon-vfpv4_gcc-12.3.0_musl_eabi/usr/include -I/home/egorenar/Repositories/openwrt-rel/staging_dir/toolchain-arm_cortex-a15+neon-vfpv4_gcc-12.3.0_musl_eabi/include/fortify -I/home/egorenar/Repositories/openwrt-rel/staging_dir/toolchain-arm_cortex-a15+neon-vfpv4_gcc-12.3.0_musl_eabi/include -I/home/egorenar/Repositories/openwrt-rel/staging_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/usr/include/python3.11 build/temp.linux-arm-cpython-311/src/lxml/builder.o -L/usr/lib -o build/lib.linux-arm-cpython-311/lxml/builder.cpython-311-arm-linux-musleabihf.so
[...]

PR #22551, opened 2023-10-30
python3: Fix building C extensions with setuptools

jefferyto:
Maintainer: me
Compile tested: armsr-armv7, 2023-10-28 snapshot sdk
Run tested: N/A

Description:
setuptools provides a local copy of distutils and when building a C extension, this distutils will add the target `LIBDIR` (/usr/lib) to the list of library paths.

If the build system has a libpython3.11.so in /usr/lib, then the linker will try to link to this shared library and fail.

This adapts 008-distutils-use-python-sysroot.patch for host setuptools to add the correct library directory.

Fixes: https://github.com/openwrt/packages/issues/22330


commodo:
Interesting.

I have another issue with `python-yaml` that may be fixed by this.
Will check

PR #22553, opened 2023-10-30
ddns-scripts: Fix Route53 provider

maxberger:
Maintainer: @chris5560, @maxberger, @dibdot
Compile tested: : N/A
Run tested:  OpenWrt 23.05.0, r23497-6637af95aa, amd64 

Description:
This fixes the Invalid Resource Record: FATAL problem: ARRDATAIllegalIPv4Address error message described in https://forum.openwrt.org/t/route53v1-script-error/160068

Signed-off-by: Max Berger <max@berger.name>

PR #22554, opened 2023-10-30
ddns-scripts: Fix Route53 provider

maxberger:
This is the same as https://github.com/openwrt/packages/pull/22553 but backported to the 23.05 branch since it is broken there.

Maintainer: @chris5560, @maxberger, @dibdot
Compile tested: : N/A
Run tested: OpenWrt 23.05.0, r23497-6637af95aa, amd64

Description:
This fixes the Invalid Resource Record: FATAL problem: ARRDATAIllegalIPv4Address error message described in https://forum.openwrt.org/t/route53v1-script-error/160068

Signed-off-by: Max Berger <max@berger.name>

### DIFF
--- a/net/ddns-scripts/Makefile
+++ b/net/ddns-scripts/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ddns-scripts
 PKG_VERSION:=2.8.2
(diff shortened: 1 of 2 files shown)


dibdot:
This leads to a version mismatch with master branch. Let's cherrypick your original change #22553 once it has been applied to master.

PR #22564, opened 2023-10-31
micropython: Update to 1.21.0, refresh/redo patches

jefferyto:
Maintainer: me
Compile tested: armsr-armv7, 2023-10-28 snapshot sdk
Run tested: armsr-armv7 (qemu), 2023-10-28 snapshot

Description:
Removed 050-py-stackctrl-fix-gcc-13.patch as it has been merged in the new version.

This also:

* Add `STRIP=` (empty value) to skip micropython's stripping step, letting the OpenWrt build system control stripping.

* Add a workaround for "variable might be clobbered" warning leading to build error on riscv64 (https://github.com/micropython/micropython/issues/12838).

* Change Build/InstallDev to install host tools into an unversioned directory, and update micropython-lib to use the unversioned path. There is no need to keep available multiple versions of the host tools.

* Add a test.sh script for the packages feed CI.


jefferyto:
Updates:
* Add `STRIP=` make flag
* Add workaround for https://github.com/micropython/micropython/issues/12838

PR #22592, opened 2023-11-03
Dysk: Add new package

FacundoAcevedo:
Maintainer: me
Compile tested: x86_64, container image: docker.io/openwrt/sdk:latest (4c7b0adaed32)
Run tested: x86_64, Vm Openwrt 22.03.5

Description:
Dysk is a command-line tool designed for efficient file and directory management in Unix-like environments

### DIFF
--- a/utils/dysk/Makefile
+++ b/utils/dysk/Makefile
@@ -1,0 +1,41 @@
+# SPDX-License-Identifier: GPL-3.0-only
+#
+# Copyright (C) 2023 Facundo Acevedo
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=dysk
+PKG_VERSION:=2.8.2
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


1715173329 (approved):

utils/dysk/Makefile
```diff
@@ -0,0 +1,46 @@
...
+PKG_BUILD_DEPENDS:=rust/host
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/rust/rust-package.mk
+
+define Package/dysk
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Uility for eficient file and directory management
```
BKPepe, line 28:
```suggestion
  TITLE:=Utility for efficient file and directory management
```

jonasjelonek:
@FacundoAcevedo I wonder of overwriting the default Package/Install is necessary in this case? It looks equal to what the default would do.

FacundoAcevedo:
> @FacundoAcevedo I wonder if overwriting the default Package/Install is necessary in this case? To me it looks equal to what the default (defined by RustBinPackage) would do.

Just checked, you are right, thanks

jefferyto (approved):

Issue #22593, opened 2023-11-03
unbound: static leases are not used extra DNS entries


jcassette:
Maintainer: @EricLuehrsen
Environment: OpenWrt 23.05.0 r23497-6637af95aa / LuCI openwrt-23.05 branch git-23.236.53405-fc638c8

### Description:

When Unbound "Extra DNS" UCI option is set to "Host Records" or "Host/MX/SRV RR" or "Host/MX/SRV/CNAME RR", the hostnames of static leases are not used as extra DNS entries.

This is not in accordance with the description of the DHCP Hostname option:
> Hostnames are used to bind a domain name to an IP address. This setting is redundant for hostnames already configured with static leases, but it can be useful to rebind an FQDN.

### Reproduction steps:

```
uci add dhcp host
uci set dhcp.@host[-1].name='example'
uci set dhcp.@host[-1].dns='1'
uci set dhcp.@host[-1].mac='00:01:02:03:04:05'
uci set dhcp.@host[-1].ip='10.1.1.111'
uci commit dhcp

uci set unbound.ub_main.add_extra_dns='3'
uci commit unbound

/etc/init.d/unbound restart
```

### Actual behavior:

```
root@OpenWrt:~# nslookup example.lan
Server:         127.0.0.1
Address:        127.0.0.1:53

** server can't find example.lan: NXDOMAIN

** server can't find example.lan: NXDOMAIN

```

### Expected behavior:

```
root@OpenWrt:~# nslookup example.lan
Server:         127.0.0.1
Address:        127.0.0.1:53

Name:   example.lan
Address: 10.1.1.111

```

EricLuehrsen:
The question is do you want DNS when the device is not connected? DNS is pulled from the DHCP lease file.

DarkGhostHunter:
Maintainer: @EricLuehrsen

This problem persists on `OpenWrt SNAPSHOT r26318-f5e25ea04c / LuCI Master 24.112.72013~aa463d0`

DarkGhostHunter:
Nevermind, fixed by adding `option dns '1'` to the host config in `/etc/config/dhcp`

```
config host
        option name 'my-pc-1'
        option ip '192.168.0.200'
        option mac '00:00:00:00:00:00'
        option leasetime '12h'
        option dns '1'
```

PR #22594, opened 2023-11-03
unbound: create extra host records from DHCP static leases

jcassette:
Maintainer: @EricLuehrsen
Compile tested: n/a
Run tested: OpenWrt 23.05.0 r23497-6637af95aa / LuCI openwrt-23.05 branch git-23.236.53405-fc638c8

Description:

The "Extra DNS" option allows to create records from the DHCP "Hostnames" configuration entries.
This allows to create such records from the DHCP "Static leases" configuration entries too.

Fixes: #22593

### DIFF
--- a/net/unbound/files/dnsmasq.sh
+++ b/net/unbound/files/dnsmasq.sh
@@ -69,7 +69,7 @@ create_local_zone() {
 
 ##############################################################################
 
-create_host_record() {
+create_host_record_from_domain() {
   local cfg="$1"
   local ip name debug_ip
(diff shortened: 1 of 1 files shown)


EricLuehrsen:
Are we sure we want DNS entries for devices with transient connectivity? Even though they have a DHCP reservation, they may not be present. Unbound scripts read for the DHCP lease file to intentionally filter this condition.

EricLuehrsen (approved):
okay to merge

PR #22598, opened 2023-11-04
[22.03] dnsproxy: add more options

muink:
Maintainer: @1715173329 

add options:
- --http3
- --timeout


1715173329:
Have you tested with this old version?

muink:
> Have you tested with this old version?

Yes, These options also work in v0.52.0

PR #22599, opened 2023-11-04
wgsd: add package

vooon:
Maintainer: me / @vooon
Compile tested: ramips / Asus ax53u / OpenWrt master, 
Run tested: ramips / Asus ax53u / OpenWrt master - client, x64_64 - coredns

Description:

Adds [WGSD](https://github.com/jwhited/wgsd), a service that enables use cases such as:

- Building a mesh of WireGuard peers from a central registry
- Dynamic discovery of WireGuard Endpoint addressing (both IP address and port number)
- NAT-to-NAT WireGuard connectivity where [UDP hole punching](https://en.wikipedia.org/wiki/UDP_hole_punching) is supported.

Provides two packages:
- wgsd-coredns - a DNS-SD server
- wgsd-client - a client, which query DNS server and update wireguard peers

### DIFF
--- a/net/wgsd/Makefile
+++ b/net/wgsd/Makefile
@@ -1,0 +1,87 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=wgsd
+PKG_VERSION:=0.3.6
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/jwhited/wgsd/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=b64e7c93a98e444e6ef46871fb73349c485c501469758e459e10188d7abfec28
+
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+PKG_MAINTAINER:=Vladimir Ermakov <vooon341@gmail.com>
+
+PKG_BUILD_DEPENDS:=golang/host
+PKG_BUILD_PARALLEL:=1
+PKG_BUILD_FLAGS:=no-mips16
+
+GO_PKG:=github.com/jwhited/wgsd
+GO_PKG_EXCLUDES:=test
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/golang/golang-package.mk
+
+define Package/wgsd/Default
+  SECTION:=net
+  CATEGORY:=Network
+  SUBMENU:=VPN
+  TITLE:=WireGuard Service Discovery
+  URL:=https://github.com/jwhited/wgsd
+  DEPENDS:=+kmod-wireguard
+endef
+
+define Package/wgsd-coredns
+  $(call Package/wgsd/Default)
+  TITLE+= DNS-SD server
+endef
+
+define Package/wgsd-client
+  $(call Package/wgsd/Default)
+  TITLE+= Client
+endef
+
+define Package/wgsd/Default/description
+  wgsd is a CoreDNS plugin that serves WireGuard peer information via DNS-SD (RFC6763) semantics.
+  This enables use cases such as:
+
+  - Building a mesh of WireGuard peers from a central registry
+  - Dynamic discovery of WireGuard Endpoint addressing (both IP address and port number)
+  - NAT-to-NAT WireGuard connectivity where UDP hole punching is supported.
+endef
+
+define Package/wgsd-coredns/description
+  $(call Package/wgsd/Default/description)
+
+  CoreDNS binary.
+endef
+
+define Package/wgsd-client/description
+  $(call Package/wgsd/Default/description)
+
+  Client binary.
+endef
+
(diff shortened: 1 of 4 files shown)


net/wgsd/Makefile
```diff
@@ -0,0 +1,87 @@
...
+include ../../lang/golang/golang-package.mk
+
+define Package/wgsd/Default
+	SECTION:=net
+	CATEGORY:=Network
+	SUBMENU:=VPN
+	TITLE:=WireGuard Service Discovery
+	URL:=https://github.com/jwhited/wgsd
+	DEPENDS:=+kmod-wireguard
+endef
```
BKPepe, line 32: Two spaces as indentation

net/wgsd/Makefile
```diff
@@ -0,0 +1,87 @@
...
+	$(call Package/wgsd/Default/description)
+
+	CoreDNS binary.
+endef
+
+define Package/wgsd-client/description
+	$(call Package/wgsd/Default/description)
+
+	Client binary.
+endef
```
BKPepe, line 63: dtto - same spaces as indentation

net/wgsd/Makefile
```diff
@@ -0,0 +1,87 @@
...
+/etc/Corefile
+endef
+
+define Package/wgsd-coredns/install
+	$(call GoPackage/Package/Install/Bin,$(PKG_INSTALL_DIR))
+	$(INSTALL_DIR) $(1)/usr/sbin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/coredns $(1)/usr/sbin/wgsd-coredns
+
+	$(INSTALL_DIR) $(1)/etc/
+	$(INSTALL_CONF) $(CURDIR)/files/Corefile $(1)/etc/Corefile
```
BKPepe, line 75: Why upper case? All files at least in /etc/ does not begging with capital first letter.

net/wgsd/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Installation
+
+For Server
+
+    opkg install wgsd-coredns
+
+For Client
+
+    opkg install wgsd-client
```
BKPepe, line 17: Do we really NEED this README? I dont think so. This is kinda obvious, how to install packages. If not, look at the documentation. The usage? The same. Create documentation in OpenWrt wiki for that.

BKPepe (changes requested):
In the commit description, there should be - why are you adding these two packages and why they are useful (of course, reasons, why it should be merged.)

net/wgsd/Makefile
vooon: Fixed

net/wgsd/Makefile
vooon: That's convention used by CoreDNS, it by default expect `Corefile`, similar to Caddy's `Caddyfile`.

net/wgsd/README.md
vooon: I've used wg-installer readme as a template.

vooon:
And seems something has changed to golang package that i can't build it again on BuildPackage macro.

net/wgsd/README.md
BKPepe: Still, it does not mean it is right. For that, there is OpenWrt wiki. I think it is quite obvious how to install package for client or server. On the other hand, we can have that for each package in OpenWrt and I dont think it makes sense.

net/wgsd/files/wgsd-coredns.init
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh /etc/rc.common
+
+START=99
+
+USE_PROCD=1
+NAME=wgsd-coredns
```
BKPepe, line 6: This is not used anywhere.

vooon: Do you mean `NAME`, because i saw other flags in other packages.

BKPepe: Can you share with me some examples, please? I would like to check this out. Thanks!

vooon: @BKPepe whole idea is described at author's post: https://www.jordanwhited.com/posts/wireguard-endpoint-discovery-nat-traversal/

So basically i have a VPS which works as a VPN exit node, i call it Cloud Router (CR, simply because i run openwrt x86 somewhere at cloud), and two sites behind NAT that wants to talk directly, as the one does video streaming.

On a CR i run wgsd-coredns daemon and have WG interface with both places peers configured (let's call them Alice and the Bob).
Then on a client side (Alice and Bob) i just have WG interface with the config to connect to the CR and Bob's/Alice's  peers. Then wgsd-client query the central node for the other peers IP/Port and update configuration of your WireGuard interface.

That works fine for SNAT, but seems something has changed on one of my ISPs. So i'm searching for other site-to-site mesh tunneling...

vooon:
@BKPepe ah, ok. I dropped NAME from init.d, and also dropped installation section from readme.

PR #22604, opened 2023-11-05
mdnsresponder: add patch renaming lib to dns_sd-mdns

Ansuel:
Add patch renaming lib to dns_sd-mdns to permit coexist with avahi.

Also the 2 user of this package are updated to use the renamed library.

---

This is needed to try to solve a very specific problem of conflicts between mdnsresponder and avahi. Both package (avahi with the compat subpackage) export the same library and on buildbot compile or with an sdk compiling multiple package, the library can be deleted or overwrite during the compile project resulting in compile error...

One solution applied in this pr is to rename the library to a different name and patch the package to link to that. This also permits avahi and mdnsresponder to coexist.

An alternative more drastic function is to just drop mdnsresponder and switch these 2 package to avahi (they support avahi)
Problem of dropping this is that user might want to use mdnsresponder as a lighter utility and have some special configuration applied and don't want to reinvent the wheel with implementing the same thing with avahi.


Ansuel:
@ynezz look what I found? A package failing to compile as 2 package export the same library and for some reason it gets deleted.

This is hard to repro on a buildroot but I manage to repro by cloning an sdk and compiling each package... Eventually after a while the problem was repro.

I'm also having a similar problem with pcre to pcre2 conversion where package are finding pcre.h... I wonder if we should really invest some effort in implementing some kind of build isolation where only the header and library declared in dependency are exposed... (my idea is to have a staging_directory for each package and use simbolic links to insert all the files... but scanning the dependency tree might be very problematic)

Ansuel:
@stintel I wonder if you want to add a variant and introduce avahi support for these 2 package?

(it's dead easy... for openthread it's a config to change and for libdmapsharing is set a configure to force avahi use)

stintel:
> @stintel I wonder if you want to add a variant and introduce avahi support for these 2 package?

Honestly I prefer not to. OpenThread Border Router works fine with mDNSResponder and it doesn't depend D-Bus.

Ansuel:
@stintel yep i knew the reason of using mdns was having to skip dbus... actually it was one of the reason i chose this solution compared to the other. Thanks for the feedback!

Well guess this is the only way... Can you test this btw?

stintel:
I'll try to test somewhere next week. Feel free to remind me if I don't reply.

neheb:
I thought dbus was optional with avahi.

neheb:
ping @flyn-org

neheb:
@Ansuel maybe the proper solution is to get rid of avahi's dns-sd. Only shairplay uses avahi's compat library directly. And something tells me it doesn't need to...

stintel:
> I thought dbus was optional with avahi.

It depends. I don't recall the details, but using Avahi with OpenThread Border Router required the DBUS variant.

neheb:
@stintel yeah just saw. the compat library depends on libavahi-client which depends on dbus.

stintel:
Successfully tested openthread-br with this. Thanks!

Issue #22384, opened 2023-10-13
quassel-irssi is not compiled due to implicit declaration of function


BKPepe:
Maintainer: @TC01 
Environment: OpenWrt master, aarch64_cortex-a53

Description:

```
quassel-net.c: In function 'sig_connected':
quassel-net.c:108:17: warning: implicit declaration of function 'g_input_add'; did you mean 'i_input_add'? [-Wimplicit-function-declaration]
  108 |                 g_input_add(net_sendbuffer_handle(r->handle),
      |                 ^~~~~~~~~~~
      |                 i_input_add
quassel-net.c:109:29: error: 'G_INPUT_READ' undeclared (first use in this function); did you mean 'I_INPUT_READ'?
  109 |                             G_INPUT_READ,
      |                             ^~~~~~~~~~~~
      |                             I_INPUT_READ
quassel-net.c:109:29: note: each undeclared identifier is reported only once for each function it appears in
quassel-net.c:110:29: warning: cast between incompatible function types from 'void (*)(Quassel_SERVER_REC *)' {aka 'void (*)(struct Quassel_SERVER_REC_s *)'} to 'void (*)(void *, GIOChannel *, int)' [-Wcast-function-type]
  110 |                             (GInputFunction) quassel_parse_incoming, r);
      |                             ^
In file included from /builder/shared-workdir/build/sdk/staging_dir/target-aarch64_cortex-a53_musl/usr/include/irssi/src/core/commands.h:4,
                 from quassel-net.c:25:
quassel-net.c: In function 'quassel_net_init':
quassel-net.c:161:46: warning: cast between incompatible function types from 'void (*)(Quassel_SERVER_REC *)' {aka 'void (*)(struct Quassel_SERVER_REC_s *)'} to 'void (*)(const void *, const void *, const void *, const void *, const void *, const void *)' [-Wcast-function-type]
  161 |         signal_add_first("server connected", (SIGNAL_FUNC) sig_connected);
      |                                              ^
quassel-net.c: In function 'quassel_net_final_setup':
quassel-net.c:170:29: error: 'G_INPUT_READ' undeclared (first use in this function); did you mean 'I_INPUT_READ'?
  170 |                             G_INPUT_READ,
      |                             ^~~~~~~~~~~~
      |                             I_INPUT_READ
[...]

PR #7454, opened 2018-11-17
quasselc: Update to latest commit

neheb:
Signed-off-by: Rosen Penev <rosenp@gmail.com>

Maintainer: @TC01 
Compile tested: mvebu

PR #22605, opened 2023-11-05
quassel-irssi: fix compilation error with new version of irssi

Ansuel:
Add patch fixing compilation error with new version of irssi where the renamed some functions.

Fixes: #22384

### DIFF
--- a/net/quassel-irssi/Makefile
+++ b/net/quassel-irssi/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=quassel-irssi
 PKG_SOURCE_DATE:=2017-11-30
 PKG_SOURCE_VERSION:=079be662dde374a383646256108a4974c2bc7796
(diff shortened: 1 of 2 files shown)


BKPepe:
Hmm, shouldn't we rather remove this package? I checked opened issues in the upstream repository of quassel-irsi and found this one: https://github.com/phhusson/quassel-irssi/issues/36

neheb:
Relevant: https://github.com/openwrt/packages/pull/7454

Ansuel:
@BKPepe well ok for me to drop but wonder if we can find a single user that would still use this???

BKPepe:
So, the conclusion is that if we find only one single user, who use this package then it is ok to have this package in our repositories? Hmm, I dont like it. Its dead in upatream. No features will be developed, its not maintained anymore. If someone is using it, then he/she can compile it locally or rather switch to different package. There is no point that we will be maintaining dead projects. We don't even have people for that.

PR #22607, opened 2023-11-06
remove compat-libdns_sd

neheb:
Alternative to https://github.com/openwrt/packages/pull/22604


neheb:
anyone even use shairplay? seems it doesn't even work: https://github.com/juhovh/shairplay/pull/80

BKPepe:
Shairplay seems to be dead, so drop that package.

Ansuel:
@neheb main problem of switching to mdns is that people might have problem and notice the broken package. IMHO shairplay should not be dropped...

BKPepe:
Hmm, while checking it with Ubuntu or Debian, I see that https://salsa.debian.org/multimedia-team/shairplay/-/blob/master/src/lib/dnssd.c which is using this source https://github.com/juhovh/shairplay does not apply the referenced [patch](https://sources.debian.org/patches/shairplay/0.9.0~git20180824.096b61a+dfsg1-3/).  Hard to say, but it looks like it is not developed anymore and that happens. There are several issues opened in that repository and it does not look good, though.

However, they packaged and are using https://github.com/mikebrady/shairport-sync client, which works well with PulseAudio, which we do have here: https://github.com/openwrt/packages/blob/56e29ed7daf39b683be49f06cb4bc065b2bdfd35/sound/shairport-sync/Makefile

BKPepe:
I see that in the past @Noltari added and was using this package. Are you still using this or do you know some better alternatives?

neheb:
@Ansuel it’s broken already. No need to keep on life support

neheb:
@Ansuel merging soon if no objections. I want openthread-br built on the buildbots.

Ansuel:
Ok for me.

PR #22608, opened 2023-11-06
Remove two packages: quassel and quassel-irssi

BKPepe:
Maintainer: @TC01 
Compile and run tested: N/A

Description:
Drop two packages, which are dead. No longer maintained.

Superseeds: https://github.com/openwrt/packages/pull/22605
And also relevant https://github.com/openwrt/packages/pull/7454


Ansuel:
Ok for me... but what about 23.xx ? should we backport?

BKPepe:
Yup, for sure!

neheb (approved):

diizzyy:
Upstream is still active?

Ansuel:
@diizzyy the irssi seems dead... quasselc same... (this is not about quassel that can't be compiled on openwrt as it does require heavy qt framework... quasselc is a different thing and was a neat project to have a c program that implemented quassel api in pure c...)

I would move these 2 package in abbandoned... maybe someone might be still intersted in them? (I would also merge my patch that fix the compilation just for extra bonus)

1715173329 (approved):

neheb (approved):

BKPepe:
> I would move these 2 package in abbandoned... maybe someone might be still intersted in them? (I would also merge my patch that fix the compilation just for extra bonus)

So, lets merge your fix, I will rebase it and afterwards merge this PR and move this to abandoned repository.

Issue #22171, opened 2023-09-21
libsodium: build error, impossible constraint in 'asm'   104 |     __asm__ __volatile__("cpuid"


schuettecarsten:
libsodium fails to build on my config, I am using mt7622 with gcc-13, binutils-2.41 and kernel 6.1. The library is used by hostip which can be used by ddns as a busybox nslookup replacement.

```
In function '_cpuid',
    inlined from '_sodium_runtime_intel_cpu_features' at sodium/runtime.c:122:5,
    inlined from '_sodium_runtime_get_cpu_features' at sodium/runtime.c:239:12,
    inlined from 'sodium_init' at sodium/core.c:39:5:
sodium/runtime.c:104:5: error: impossible constraint in 'asm'
  104 |     __asm__ __volatile__("cpuid"
      |     ^
In function '_cpuid',
    inlined from '_sodium_runtime_intel_cpu_features' at sodium/runtime.c:126:5,
    inlined from '_sodium_runtime_get_cpu_features' at sodium/runtime.c:239:12,
    inlined from 'sodium_init' at sodium/core.c:39:5:
sodium/runtime.c:104:5: error: impossible constraint in 'asm'
  104 |     __asm__ __volatile__("cpuid"
      |     ^
make[7]: *** [/home/cschuette/openwrt/snapshot/openwrt/tmp/ccPxPqUP.mk:5: /home/cschuette/openwrt/snapshot/openwrt/tmp/cci6Piux.ltrans1.ltrans.o] Error 1
```

### OpenWrt version

r23990-4145ff4d8a

### OpenWrt target/subtarget

mediatek/mt7622

### Device

Linksys E8450 (UBI)

### Image kind

Self-built image

### Terms

- [X] I am reporting an issue for OpenWrt, not an unsupported fork.

PR #22612, opened 2023-11-07
libsodium: Update to 1.0.19 and fix patches

schuettecarsten:
Maintainer: @damianorenfer 
Compile and run-tested: bcm2711 (rpi4), mt7622 (Linksys E8450), ath79 (TP-Link Archer C7 v5)

Fixes https://github.com/openwrt/packages/issues/22171.

Unfortunately, latest libsodium-1.0.19.tar.gz extracts to libsodium-stable. That's why I had to update PKG_BUILD_DIR, maybe (hopefully) this can be reverted on next libsodium release.

### DIFF
--- a/libs/libsodium/Makefile
+++ b/libs/libsodium/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libsodium
-PKG_VERSION:=1.0.18
-PKG_RELEASE:=4
+PKG_VERSION:=1.0.19
+PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://download.libsodium.org/libsodium/releases \
 		https://github.com/jedisct1/libsodium/releases/download/$(PKG_VERSION)
-PKG_HASH:=6f504490b342a4f8a4c4a02fc9b866cbef8622d5df4e5452b46be121e46636c1
+PKG_HASH:=018d79fe0a045cca07331d37bd0cb57b2e838c51bc48fd837a1472e50068bbea
 
 PKG_MAINTAINER:=Damiano Renfer <x9w2n7xnu@relay.firefox.com>
 PKG_LICENSE:=ISC
@@ -26,6 +26,9 @@ PKG_INSTALL:=1
 PKG_BUILD_PARALLEL:=1
 
 include $(INCLUDE_DIR)/package.mk
+
+TAR_OPTIONS+= --strip-components 1
+TAR_CMD=$(HOST_TAR) -C $(1) $(TAR_OPTIONS)
 
 define Package/libsodium
   SECTION:=libs
--- a/libs/libsodium/patches/001-revert-f5076db5f8ef27.patch
+++ b/libs/libsodium/patches/001-revert-f5076db5f8ef27.patch
@@ -1,6 +1,6 @@
 --- a/src/libsodium/Makefile.am
 +++ b/src/libsodium/Makefile.am
-@@ -59,6 +59,7 @@ libsodium_la_SOURCES = \
+@@ -73,6 +73,7 @@ libsodium_la_SOURCES = \
  	crypto_scalarmult/curve25519/ref10/x25519_ref10.h \
  	crypto_scalarmult/curve25519/scalarmult_curve25519.c \
  	crypto_scalarmult/curve25519/scalarmult_curve25519.h \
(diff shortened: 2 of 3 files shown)


feckert:
While we're at it. We should add a valid patch header to each patch. So we have more information why we need this and what this patch do. If you are not the author of this patch, then you do not have to add a Signed-off-by and you could also replace the `From:` with `From: OpenWrt developers <openwrt-devel@lists.openwrt.org>` in the patch header.

schuettecarsten:
@feckert - updated. Unfortunately, formalities check fails now.

feckert:
The patch work is not from you. So I think this is OK because, you only added the missing header.
Lets ping other packages maintainers @neheb @ynezz  what is there opinion about this topic.

schuettecarsten:
@feckert @neheb @ynezz - That's correct, I only updated the patch to work with the latest version of libsodium. Quality checks require the Signed-Off-By text in the commit message. Please let me know what to do and I will update the commit message or the file header if neccessary.

libs/libsodium/Makefile
```diff
@@ -8,19 +8,20 @@
...
 		https://github.com/jedisct1/libsodium/releases/download/$(PKG_VERSION)
-PKG_HASH:=6f504490b342a4f8a4c4a02fc9b866cbef8622d5df4e5452b46be121e46636c1
+PKG_HASH:=018d79fe0a045cca07331d37bd0cb57b2e838c51bc48fd837a1472e50068bbea
 
 PKG_MAINTAINER:=Damiano Renfer <x9w2n7xnu@relay.firefox.com>
 PKG_LICENSE:=ISC
 PKG_LICENSE_FILES:=LICENSE
 
 PKG_BUILD_FLAGS:=no-mips16
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-stable
```
1715173329, line 24: please use `TAR_OPTIONS` instead.

add the following lines after `include $(INCLUDE_DIR)/package.mk`
```
TAR_OPTIONS+= --strip-components 1
TAR_CMD=$(HOST_TAR) -C $(1) $(TAR_OPTIONS)
```

BKPepe: I dont think this is a blocker, so we can merge this as it is. Yeah, it looks weird tha tit will be as libsodium-stable in the build dir, but it is what it is.

BKPepe:
> While we're at it. We should add a valid patch header to each patch. So we have more information why we need this and what this patch do. If you are not the author of this patch, then you do not have to add a Signed-off-by and you could also replace the `From:` with `From: OpenWrt developers <openwrt-devel@lists.openwrt.org>` in the patch header.

Uh oh... that should be done as treewide commit and we should not request users for doing that for us. Idea is nice, welcomed, appreciated. This was done in the past in the main repo and I like that, but asking users for not relevant change to do that for us, I find it... lets say strange.

libs/libsodium/Makefile
1715173329: This is confusing, IMO no.

feckert:
@BKPepe How do we continue here? Should I add the missing headers? We have already done the same with the kernel patches in the openwrt base repo.

BKPepe:
Yes, do that by treewide commit and for patches, which will miss this
header in the future simply refuse them and ask author to upstream it, if
possible of course.

Dne út 19. 12. 2023 10:23 uživatel Florian Eckert ***@***.***>
napsal:

BKPepe:
Thats can not be done like you said.

1. Why would someone should fix that for us now, when in the past, we
merged everything and did not care about that?
2. Fixing patches is a long way doing that step by step.
3. Users/contributors are here to help us, do something for us in their
free time. Dont bother them with completely unrelevant things. Why they
should do something with the patch, when it should be regular single
package bump?

You can see now how it can be time consuming, because CI requires SoB, etc.
Many things unrelated to this PR raised up and doing it per package is
honestly non sense.

The only way how this can be achieved is to create PR, when all patches
will be adressed.



Dne út 19. 12. 2023 10:40 uživatel Florian Eckert ***@***.***>
napsal:

feckert:
@BKPepe OK Thanks for the clarification. Then I'll have to see if I can write a little script that adds the missing headers if it doesn't have a valid patch header.

@schuettecarsten Now that we have clarified this, please update the pullrequest and remove the headers. So that we can merge this. I apologise for the back and forth.

libs/libsodium/Makefile
BKPepe: @schuettecarsten Can you look at it, please?

schuettecarsten:
Removed headers and added TAR_OPTIONS.

libs/libsodium/Makefile
```diff
@@ -27,6 +27,9 @@ PKG_BUILD_PARALLEL:=1
 
 include $(INCLUDE_DIR)/package.mk
 
+TAR_OPTIONS+= --strip-components 1
+TAR_CMD=$(HOST_TAR) -C $(1) $(TAR_OPTIONS)
```
BKPepe, line 31: Would be great if you can describe it in the commit description.

PR #22616, opened 2023-11-07
CI: Update sdk action, always upload artifacts

jefferyto:
Maintainer: @aparcar
Compile tested: N/A
Run tested: N/A

Description:
The updated version of gh-action-sdk will return compiled packages and build logs for both build success and build errors.

This ensures these artifacts are always uploaded. This also sets the V environment variable to enable verbose build output.


aparcar:
Thanks, as always!

PR #22622, opened 2023-11-08
modemmanager: add new proto options

feckert:
Maintainer: me @aleksander0m 

Description:

modemmanager: add possibilty for setting initial eps bearer
If no GSM but only 4G is available and a special APN must be used, it is necessary to set an inital EPS bearer beforehand. If this is not set, then modem cannot log in and register in the mobile network. To make this possible, the configuration options for setting up a mobile connection via 'simple-connect' could also be added to the init 3gpp bearer command. If the modem should add an initial EPS bearer, then following new boolean option 'gpp_epsbearer' must be set to true. If this option is set, then 'apn', 'iptype', 'allowedauth', 'user' and 'password' are used for setting the initial EPS bearer.

modemmanager: Request the modem register in its home network
This could be configured with the new 'gpp_home' boolean parameter. The command can only be executed if the modem is enabled.

@aleksander0m I have a special mobile phone setup here and would like to know your opinion on whether it can be done like this?

### DIFF
--- a/net/modemmanager/files/modemmanager.proto
+++ b/net/modemmanager/files/modemmanager.proto
@@ -364,6 +364,50 @@ modemmanager_set_allowed_mode() {
 	}
 }
 
+modemmanager_check_state() {
+	local device="$1"
+	local modemstatus="$2"
+	local pincode="$3"
+
+	local state reason
+
+	state="$(modemmanager_get_field "${modemstatus}" "state")"
+	state="${state%% *}"
+	reason="$(modemmanager_get_field "${modemstatus}" "state-failed-reason")"
+
+	case "$state" in
+		"failed")
+			case "$reason" in
+				"sim-missing")
+					echo "SIM missing"
+					proto_notify_error "${interface}" MM_FAILD_REASON_SIM_MISSING
+					proto_block_restart "${interface}"
+					return 1
+					;;
+				*)
+					proto_notify_error "${interface}" MM_FAILED_REASON_UNKNOWN
+					proto_block_restart "${interface}"
+					return 1
+					;;
+			esac
+			;;
+		"locked")
+			if [ -n "$pincode" ]; then
+				mmcli --modem="${device}" -i any --pin=${pincode} || {
+					proto_notify_error "${interface}" MM_PINCODE_WRONG
+					proto_block_restart "${interface}"
+					return 1
+				}
+			else
+				echo "PIN required"
+				proto_notify_error "${interface}" MM_PINCODE_REQUIRED
+				proto_block_restart "${interface}"
+				return 1
+			fi
+			;;
+	esac
+}
+
 modemmanager_set_preferred_mode() {
 	local device="$1"
 	local interface="$2"
@@ -430,6 +474,9 @@ proto_modemmanager_setup() {
 	}
 	echo "modem available at ${modempath}"
 
+	modemmanager_check_state "$device" "${modemstatus}" "$pincode"
+	[ "$?" -ne "0" ] && return 1
+
 	[ -z "${allowedmode}" ] || {
 		case "$allowedmode" in
 			"2g")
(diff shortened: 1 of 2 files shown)


net/modemmanager/files/modemmanager.proto
```diff
@@ -480,6 +481,25 @@ proto_modemmanager_setup() {
 		}
 	}
 
+	# set initial eps bearer settimgs (4G only)
+	[ -n "${gpp_epsbearer}" ] && [ "${gpp_epsbearer}" -eq 1 ] && {
+		echo "Setting inital EPS bearer with apn '${apn}'..."
+		# Append options to 'connectargs' variable
+		connectargs=""
```
aleksander0m, line 488: `attachargs` maybe?

net/modemmanager/files/modemmanager.proto
```diff
@@ -480,6 +481,25 @@ proto_modemmanager_setup() {
...
+		append_param "apn=${apn}"
+		append_param "${iptype:+ip-type=${iptype}}"
+		append_param "${cliauth:+allowed-auth=${cliauth}}"
+		append_param "${username:+user=${username}}"
+		append_param "${password:+password=${password}}"
```
aleksander0m, line 493: We recently added a new "force" boolean setting in this operation. If set, the settings update is always done, regardless of whether the current settings are the same. Without this "force" option, MM will treat the request as a no-op if the settings are the same.

net/modemmanager/README.md
```diff
@@ -45,3 +46,13 @@ The 'signalrate' option set's the signal refresh rate (in seconds) for the devic
 You can call signal info with command: mmcli -m 0 --signal-get
 
 If 'gpp_home' is set, the modem should register in the home network.
+
+If 'gpp_epsbearer' is set, then the modem uses the options apn, allowedauth,
+iptype, username and passowrd to set an initial EPS bearer. This is usefull
```
aleksander0m, line 51: typo: usefull -> useful

feckert:
@aleksander0m Thanks for the feedback!
I would like to avoid the user having to enter his data twice.
So I suggest the following implementation by adding a new option `init_eps_bearer`.
That could be `none`, `default` and `other`.
* `none`: No init eps bearer is set
* `default`: Use uci option `iptype`, `allowedauth`, `password`, `username` and `apn` from the already provided credential (also add new `force` option if needed)
* `other`: use other credentials that differ from the defaults. The new option would be `init_iptype`, `init_allowedauth`, `init_password`, `init_username` and `init_apn` (also add `force` option if needed)

However, I have seen that `allowedauth` could have set to more then one option at the same time in the uci (`pap`, `chap`, `mschap`, `mschapv2`, and `eap`). This is merged via `|` to a modemmanager [option](https://github.com/openwrt/packages/blob/master/net/modemmanager/files/modemmanager.proto#L465
) `allowed-auth` for simple connect. If the new `init_eps_bearer` option is set to `default`, then the `allowed-auth` option is also concatinated via `|` for `init_eps_bearer`. Could this be a problem, does the command `--3gpp-set-initial-eps-bearer-settings` is supporting this?

aleksander0m:
> @aleksander0m Thanks for the feedback! I would like to avoid the user having to enter his data twice. So I suggest the following implementation by adding a new option `init_eps_bearer`. That could be `none`, `default` and `other`.
> 
>     * `none`: No init eps bearer is set
> 
>     * `default`: Use uci option `iptype`, `allowedauth`, `password`, `username` and `apn` from the already provided credential (also add new `force` option if needed)
> 
>     * `other`: use other credentials that differ from the defaults. The new option would be `init_iptype`, `init_allowedauth`, `init_password`, `init_username` and `init_apn` (also add `force` option if needed)
> 

That seems reasonable.

> 
> However, I have seen that `allowedauth` could have set to more then one option at the same time in the uci (`pap`, `chap`, `mschap`, `mschapv2`, and `eap`). This is merged via `|` to a modemmanager [option](https://github.com/openwrt/packages/blob/master/net/modemmanager/files/modemmanager.proto#L465) `allowed-auth` for simple connect. If the new `init_eps_bearer` option is set to `default`, then the `allowed-auth` option is also concatinated via `|` for `init_eps_bearer`. Could this be a problem, does the command `--3gpp-set-initial-eps-bearer-settings` is supporting this?

I think that is fine.

feckert:
@aleksander0m Thanks for your comments. I have opened a follow up [PR](https://github.com/openwrt/packages/pull/22631). Would be helpful if you could have a quick look at it

PR #22624, opened 2023-11-08
pps-tools: use %lld in printf for 64-bit time_t

tofurky:
musl 1.2.0 switched to use 64-bit time_t everywhere, including 32-bit architectures, causing garbage values to be printed from ppswatch and ppstest.

Use the correct format string for the affected printf statements and explicitly cast to long long to avoid potential compatibility issues with 32-bit glibc.

Maintainer: @MASHtm 
Compile tested: arm, cloudengines_pogoe02, 23.05.0
Run tested: arm, cloudengines_pogoe02, 23.05.0

### DIFF
--- a/utils/pps-tools/patches/001-time_t_64bit.patch
+++ b/utils/pps-tools/patches/001-time_t_64bit.patch
@@ -1,0 +1,31 @@
+--- a/ppstest.c
++++ b/ppstest.c
+@@ -110,13 +110,13 @@ retry:
+ 	}
+ 
+ 	printf("source %d - "
+-	       "assert %ld.%09ld, sequence: %ld - "
+-	       "clear  %ld.%09ld, sequence: %ld\n",
++	       "assert %lld.%09ld, sequence: %ld - "
++	       "clear  %lld.%09ld, sequence: %ld\n",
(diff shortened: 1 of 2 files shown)


utils/pps-tools/patches/001-time_t_64bit.patch
```diff
@@ -0,0 +1,31 @@
+--- a/ppstest.c
++++ b/ppstest.c
```
BKPepe, line 2: Did you send this patch to upstream or is this included in upstream already?

tofurky: it's not upstream, and i haven't opened a PR there yet. seems kind of inactive, though.

BKPepe: Please open the PR there, otherwise, we would need to carry this forever, which is not good.

tofurky: https://github.com/redlab-i/pps-tools/pull/16

tofurky:
thanks @neheb, could we also please pull it into the 23.05 branch?

Issue #22566, opened 2023-10-31
iperf3: libiperf3 build fails when iperf3-ssl is selected


jonasjelonek:
Maintainer: @nbd168 
Also involved: @BKPepe 
Environment: (buildsystem issue)

Description:
I already mentioned this issue in my PR #22516 . When both `libiperf3` and `iperf3-ssl` are selected to be built, the build of libiperf3 fails with an error message `error: libiperf.so.*: No such file or directory`.

When digging through the Makefile one may already see what's the issue: Selecting `iperf3-ssl` adds the argument `--disable-shared` to `CONFIGURE_ARGS`, which configures the build of `iperf3` to not generate the `libiperf3` as shared library and just statically compile the `iperf3-ssl` package.
This was done this way by @PolynomialDivision in ae48be8e21 to - as far as I understand - avoid to ship multiple versions of `libiperf3` (with and without ssl support), as there is currently no package that depends on `libiperf3` with ssl support (`python3-iperf3` doesn't require the ssl support).

I already prepared two solutions for this issue in my own repo and I would like to get some comments on this.
Solution 1: [here](https://github.com/jonasjelonek/packages/commit/4262400341db364577956614abe9677e44df13de)
Solution 2: [here](https://github.com/jonasjelonek/packages/commit/5c970ea0911972633e08cbbc617ce85146b0a2b7)

Solution 1 just handles the ssl support as a config option instead of a second package/build variant. Besides fixing the issue, this also allows to build libiperf3 with or without ssl support.   
Solution 2 just evolved out of 'removing `--disable-shared` from `CONFIGURE_ARGS` and then fixing subsequent errors'. It adds a separate package `libiperf3-ssl` which separates both packages more, thus fixes the issue and would also allow to have `libiperf3` standalone with ssl support.

I would really vote for Solution 1, it is IMHO clean and solves the issue. Solution 2 would also work but seems quite messy to me, bloats both the Makefile and the package variants of iperf3. And then we would also have two versions of `libiperf3`, which was avoided before.

Any comments on this, in particular on my proposed Solution 1?
Maybe someone has an even better solution to this?

PR #22628, opened 2023-11-08
iperf3, iperf3-mt: fix failing build of libiperf3 with ssl variants

jonasjelonek:
Maintainer: @nbd168 for `iperf3`,  me for `iperf3-mt`
Compile tested: aarch64_cortex-a53 (BananaPi R64) + x86/64 (APU2C4), OpenWrt snapshot
Run tested: aarch64_cortex-a53 (BananaPi R64) + x86/64 (APU2C4), OpenWrt snapshot

Description:
In my PR #22516 and in the follow-up issue #22566 I opened, I already described the issue itself.
TLDR: Building `libiperf3` fails when `iperf3-ssl` is selected (same for `iperf3-mt`) because `iperf3-ssl` adds `--disable-shared` to CONFIGURE_ARGS which prevents generating `libiperf.so.*` file, which is obviously necessary for libiperf3.

My proposed changes fix the issue and in addition also allow to have a shared libiperf3 with SSL support, which afaik was not possible before. Instead of having separate package variants for with/without SSL support, let's just do that with a config option that can be chosen. Any objections to this?

@BKPepe Although you proposed to only create an issue for that and someone else (including you) will have a look at it later, I couldn't stop thinking about this ^^. So I thought I propose my attempt to solve this here, hopefully this is a good starting point.


jonasjelonek:
@BKPepe do you maybe have comments on this?

GeorgeSapkin:
@jonasjelonek still relevant?

jonasjelonek:
@GeorgeSapkin yes, the issue persists. Selecting iperf3-ssl and libiperf3 doesn't work as `iperf3-ssl` adds `--disable-shared` which causes `libiperf.so.*` to be missing.

GeorgeSapkin:
OK, well this is marked as draft for a while now. Are you planning on finishing this?

jonasjelonek:
Luckily, there's not really something to finish. It still applies cleanly and fixes the issue. Not sure though if there are any objections against how I fixed this.

GeorgeSapkin:
You should bump the `PKG_RELEASE`.

GeorgeSapkin:
I'm not using iperf3 so somebody else will need to check this. But after taking a brief look, does this change imply that if `PACKAGE_iperf3_SSL_SUPPORT=n` by default, SSL variant/support won't be able to be installed by the user? E.g. when building through the firmware selector or the image builder?

jonasjelonek:
> does this change imply that if `PACKAGE_iperf3_SSL_SUPPORT=n` by default, SSL variant/support won't be able to be installed by the user? E.g. when building through the firmware selector or the image builder?

If there's no way to set such config operation when using firmware selector/image builder, then yes. One of the reason I had this as an RFC/draft at first.

I'm also fine to drop this completely. I had this PR and even an issue some time ago with barely any replies. So it seems, no one actually came across it aka no need to fix something.

harrydleung (approved):
I've tested this and it works well on 24.10.5. I think this should be backported to the openwrt-24.10 branch as well.

PR #22631, opened 2023-11-09
modemmanager: add state check and set init_epsbearer

feckert:
Maintainer: me / @aleksander0m 

Description:
This is a follow up pullrequest with the request changes from this [PR](https://github.com/openwrt/packages/pull/22622)

This PR adds two new feature to the modemmanger proto handler:
1. Add add state and reason evaluation, so we could react on modem state
2. Add possibilty for setting an initial eps bearer

### DIFF
--- a/net/modemmanager/files/modemmanager.proto
+++ b/net/modemmanager/files/modemmanager.proto
@@ -364,6 +364,50 @@ modemmanager_set_allowed_mode() {
 	}
 }
 
+modemmanager_check_state() {
+	local device="$1"
+	local modemstatus="$2"
+	local pincode="$3"
+
+	local state reason
+
+	state="$(modemmanager_get_field "${modemstatus}" "state")"
+	state="${state%% *}"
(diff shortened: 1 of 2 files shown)


net/modemmanager/files/modemmanager.proto
```diff
@@ -364,6 +370,50 @@ modemmanager_set_allowed_mode() {
...
+	state="$(modemmanager_get_field "${modemstatus}" "state")"
+	state="${state%% *}"
+	reason="$(modemmanager_get_field "${modemstatus}" "state-failed-reason")"
+
+	case "$state" in
+		"failed")
+			case "$reason" in
+				"sim-missing")
+					echo "SIM missing"
+					proto_notify_error "${interface}" MM_FAILD_REASON_SIM_MISSING
```
aport, line 389: Typo `MM_FAILD` -> `MM_FAILED`

feckert: Nice catch. Will update on the next `git push --force`.

Issue #22632, opened 2023-11-09
tang: running but not doing anything (permissions issue)


tew42:
Maintainer: @Tiboris  (copying @nmav due to recent commits, including introduction of distinct "tang" user)
Environment: Archer A7 v5, OpenWRT 23.05

Description:
tang starts fine, keys are generated, but: default permissions are wrong (process running as user tang can't see jwk files that are generated by/owned by/readable by root)

Everything works as expected when removing `procd_set_param user tang` from the init script

Logs:
nothing, until I added following to init.d/tang (_this should be implemented as standard in any case, perhaps?_)
```
procd_set_param stdout 1
procd_set_param stderr 1
```

access attempt http://192.168.1.1:9090
```
Thu Nov  9 15:41:21 2023 daemon.err tangd[3596]: <unknown> GET / => 404 (../src/http.c:128)
Thu Nov  9 15:41:22 2023 daemon.err tangd[3596]: <unknown> GET / => 404 (../src/http.c:128)
```

access attempt http://192.168.1.1:9090/adv
```
Thu Nov  9 15:41:55 2023 daemon.err tangd[3596]: <unknown> GET /advCannot load JSON file (/usr/share/tang/db/xxx.jwk); skipping
Thu Nov  9 15:41:55 2023 daemon.err tangd[3596]: error text unable to open /usr/share/tang/db/xxx.jwk: Permission denied, line -1, col -1, pos 0
Thu Nov  9 15:41:55 2023 daemon.err tangd[3596]: Cannot load JSON file (/usr/share/tang/db/xxx.jwk); skipping
Thu Nov  9 15:41:55 2023 daemon.err tangd[3596]: error text unable to open /usr/share/tang/db/xxx.jwk: Permission denied, line -1, col -1, pos 0
Thu Nov  9 15:41:55 2023 daemon.err tangd[3596]: Error saving JWK to file (/usr/share/tang/db/xxx.jwk)
Thu Nov  9 15:41:55 2023 daemon.err tangd[3596]:  => 500 (../src/tangd.c:78)
```

Other relevant info:
```
top
3076     1 tang     S     5136   4%   0% /usr/sbin/tangd -p 9090 -l /usr/share/tang/db
```
```
/usr/share/tang/db# ls -l
-r--r-----    1 root     root           349 Nov  9 15:37 xxx.jwk
-r--r-----    1 root     root           354 Nov  9 15:37 xxx.jwk
```

brada4:
chown -R tang db/

[...]

PR #22634, opened 2023-11-09
tang: set the right permissions to keys

nmav:
Maintainer: me
Compile tested: -
Run tested: -

Description:
This corrects the permissions of generated files and sets the right maintainer.

Resolves: #22632

### DIFF
--- a/utils/tang/Makefile
+++ b/utils/tang/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tang
 PKG_VERSION:=14
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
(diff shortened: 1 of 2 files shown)


tew42:
Thanks for the quick response!

Weirdly enough, it doesn't work, however - neither with the updated init script, nor by directly issuing chown. It seems related to the fact that the tang user got created as:

cat /etc/passwd
tang:x:65536:65536:tang:/var/run/tang:/bin/false
If I manually update the UID to 65530, everything works fine, however?! I thought that's only supposed to be an issue for really old kernels...

nmav:
Thanks, it looks indeed strange the issue, the creation of a user shouldn't use this ID if it causes issues. Nevertheless, I'll merge the fix as it addresses the part of tang.

PR #22639, opened 2023-11-10
libwebsockets: Fix compile error due to uninitialized variable with gcc-13

schuettecarsten:
Fix compile error due to uninitialized variable with gcc-13.

Compile and run-tested (with gcc-13):
bcm2711 (rpi4), mt7622 (Linksys E8450), ath79 (TP-Link Archer C7 v5)

Issue also reported to https://github.com/warmcat/libwebsockets/issues/3004


libs/libwebsockets/patches/100-fix-uninitialized-variable-usage.patch
```diff
@@ -0,0 +1,22 @@
+From 3589cc48e28cd06c4143da244af5ea91b06f2e35 Mon Sep 17 00:00:00 2001
```
BKPepe, line 1: This commit was changed since you posted it here.

Can you please replace it by this one:
https://github.com/warmcat/libwebsockets/commit/bfce9f86187b0846c3eab359591e64ed4fa4bb0a.patch ?

schuettecarsten: @BKPepe updated.

BKPepe:
The merge commit should not be there. You should rebase your local branch on top of upstream branch.

schuettecarsten:
> The merge commit should not be there. You should rebase your local branch on top of upstream branch.

Also fixed, I do not know how this happened

BKPepe:
Patch needs to be refreshed.

BKPepe:
CI:

> Dirty patches detected, please refresh and review the diff

schuettecarsten:
This has been fixed upstream.

PR #22660, opened 2023-11-14
nut: update to 2.8.1

ruralroots:
Maintainer: None
Compile tested: yes mvebu, WRT1900ACS, Master/Main & Mediatek, Bananapi BPi-Rs, 23.05.0

Description: Adds additional features.


jefferyto:
The patch needs to be refreshed (`make package/nut/{clean,refresh}`). Also, rebase your branch on master instead of adding a merge commit.

hnyman:
Applied as 82f36e0 after fixing the "dirty patches" error removing the upstream patch that is already included in 2.8.1

PR #22662, opened 2023-11-14
[23.05] pick commits for modemmanager from master branch

dangowrt:
Maintainer: @mips171 
Compile tested: -
Run tested: -

Description:
Pick commits updating ModemManager from master branch to openwrt-23.05.


feckert:
Can we wait a little longer! I have problems with an EC25 from Quectel that works with the older modemmanger 1.20 version and now with the new version 1.22 it doesn't work anymore with a hotplug!

I am still trying to find out what is causing the problem. An update of the modem firmware does solve the problem. But an update in the field is not possible for me because I have too little memory for the image!!!

robimarko:
@feckert Doesn't it also support FOTA updates via direct HTTP link?

Personally, on RM520N I have seemingly random issues with MM registering it before it processes all of the events and thus does not see the AT and secondary AT ports at all.

robimarko:
> Yes the modem EC25 does have the possibility to upgrade via FOTA. But I am still in contact with the quectel support to get a valid upgrade file for FOTA. My modem does have current the software version R02M07 and I want to update to software version R06M16.

For me they are usually fast when providing FW updates, though you have to host those with a direct link yourself.

> 
> This is possibly a race condition with event processing during booting? Since the modem manager plays back all recorded events that happened before the modem manager was started and that are added during the boot phase after modem manager start. Since this is not locked, it can lead to a race condition from my point of view. You can switch off the modem manager during the boot and then start it manually when the system is booted. Just to see if my assumption is correct?

I tracked this down to a hotplug script added by my colleague who was waiting for the modem to enable GPS but the issue is that he was doing this sleeping in a while loop and blocking hotplug events which led to half of the ttyUSB events arriving 20+ seconds later than the first 2 thus getting ignored by MM since more than 1.5s have passed since the modem object was registered.

So, it was our own fault, nothing to do with OpenWrt, and now MM 1.22 works just fine again(With the exception of GPS configuration being broken over QMI)

feckert:
We have found the [commit](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/commit/65efdded070af57bac8171ed686706dbd447c9a4) that is responsible for the EC25 not working with the older firmware versions.

To summarise, a new generic QMI function has been added to the modem manager version 1.22, which adds event handling for profile management. With older firmware smaller than R06, the modem crashes when these commands are sent via QMI and bootstraps. We have therefore created a [pull request](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/merge_requests/1104) that checks which firmware the modem has and if this is older, then we overwrite the functionality and the profile handling is switched off.

There is also an [issue](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/issues/795) about this on the modemmanager Gitlab server.
As long as the problem is not fixed, I would wait to backport this modem manager version 1.22 to owrt-23.05. We only became aware of the problem through our test setup.

feckert:
@BKPepe With the new ModemManger version 1.22 there is an additional issue with the PLS8E. Since the new version, the ModemManager can no longer recognize the hotplug events for a SIM hotswap. We have already opened an [issue](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/issues/812) with the ModemManager repository.

I'm not quite sure whether we should continue with the approach of back-porting the 1.22 to the owrt-23.05 branch. We have already identified problems with our EC25 and now with PLS8E modems, which we first have to fix!

BKPepe:
Bugs are happening, there is no software without bugs. I understand that for you it is critical, but we need to consider what and if the benefits from the newer versions are suitable and handful in the stable branch. For now, it isnt and no one is hurry about this. This can sit here for some time until you are confident to be there.

BKPepe:
> We have already opened an [issue](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/issues/812) with the ModemManager repository.

I see, it is fixed, thus we can continue with this PR.

feckert:
@BKPepe This was only merged in the ModemManager master branch. The Fixes are not backported to the ModemManager stable branches 1.22 https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/commits/mm-1-22/?ref_type=heads

I do not know if this will ever happen!

If then we have to do it ourselves. Also in the Openwrt master branch the fixes still have to be backported as long as no new ModemManager version has been released by @aleksander0m

hnyman:
23.05 is EoL, so let's close this old PR

PR #22666, opened 2023-11-15
freeradius3: Fix build when pcre is present

jefferyto:
Maintainer: none
Compile tested: armsr-armv7, 2023-11-10 snapshot sdk
Run tested: none

Description:
If pcre is built before freeradius, then freeradius' configure will detect pcre and freeradius will be built with pcre functionality enabled. This causes a "missing dependencies" error at the end of package build.

This passes `--without-pcre` to configure to disable this autodetection.

This also removes the dependency on libpcre2 as freeradius v3 does not have support for pcre2.

Fixes: 19ec30255f13 ("freeradius3: switch to pcre2")
Fixes: https://github.com/openwrt/packages/issues/22574

### DIFF
--- a/net/freeradius3/Makefile
+++ b/net/freeradius3/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freeradius3
 PKG_VERSION:=3.0.26
-PKG_RELEASE:=2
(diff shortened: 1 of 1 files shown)


jefferyto:
Ping @odmdas, @neheb

PR #22667, opened 2023-11-15
gl-puli-mcu: Adaptation for GL.iNet XE3000 (Puli AX)

JeThWifirst:
Maintainer: @nunojpg 
Compile tested: GL.iNet GL-XE300, GL.iNet GL-XE3000
Run tested: GL.iNet GL-XE300, GL.iNet GL-XE3000

Description:
I'm working on the support for GL.iNet GL-XE3000 (Puli AX) in OpenWrt, and I adapted this package in order to be able to retrieve the battery status on this new target.

### DIFF
--- a/utils/gl-puli-mcu/Config.in
+++ b/utils/gl-puli-mcu/Config.in
@@ -1,0 +1,12 @@
+choice
+    depends on PACKAGE_gl-puli-mcu
+
+    prompt "GL.iNet target"
+    default GL_PULI_MCU_XE300
+
+    config GL_PULI_MCU_XE300
+            bool "GL.iNet XE300 (Puli)"
+
+    config GL_PULI_MCU_XE3000
+            bool "GL.iNet XE3000 (Puli AX)"
+endchoice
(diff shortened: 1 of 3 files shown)


nunojpg:
Hey. Thanks. 
I would parse the simple json without using any external libraries. The reason is that you are making this 100x the size by bundling a library you use only 1%. Parsing a single level json like this is trivial in C.

Also you need to sign the commits (search for `signed-off-by`)

JeThWifirst:
Hey, thanks for the quick reply!
I thought it would be safer and more robust to use a real JSON parser, just in case the format unexpectedly changes on the MCU side. My choice was also motivated by the 8Gb available on the target, as a few kB would not really matter. But I understand your point, let's not waste memory if not really necessary. 
I've made the changes you suggested, and now the JSON object is manually parsed.

nunojpg:
Very good in general, just one fix to go

JeThWifirst:
Which fix are you referring to? I don't see any comment about the code in the review, am I missing something ?

nunojpg:
I can see it just above in this thread, the comment is "+kmod-usb-serial-ch341 becomes only required for XE300"

JeThWifirst:
Indeed, good catch, `kmod-usb-serial-ch341` is now conditioned by XE300

PR #22684, opened 2023-11-17
upmpcli: update and fix with GCC13

neheb:
Maintainer: nobody

ping @medoc92


medoc92:
Thanks, the upmpdutils.hxx change will be in the next release.

PR #22695, opened 2023-11-18
Unbound reload fix

drujd:
Maintainer: @EricLuehrsen 
Run tested: Turris Omnia running OpenWrt master

Description:
Reloading unbound (`/etc/init.d/unbound reload` or `service unbound reload`) does not work at all, only restart does. This is due to default reload() implementation running start_service() and comparing states. Since the state is always the same, reload (and therefore triggers - other than the raw trigger that calls restart) only ever regenerates unbound  config files, but those files are never (re)loaded. This causes various problems as changes in trigger interfaces are ignored until manual restart.

Fix this by adding files instance parameter for the basic config files (changes in additional included files will not cause reload, however) and removing dates from config parts headers - that way, unbound is actually not restarted when the regenerated config is completely unchanged.


EricLuehrsen:
That might work if Unbound reliably allowed a soft reload. Some settings or their combinations do not take hold without a hard restart.

drujd:
Since reload is not overriden in the init script, reload does not work at all. With this modification, reload does the same thing as restart does as long as there is config changes. There are triggers in the init script that call reload instead of restart, which causes problems as without the fix restart has to be manually called if some interface status change was supposed to change the config.

As far as I know, Unbound does not support soft reload at all, with the correct signal it can only restart itself keeping the original PID.

EricLuehrsen:
I had to look at my notes (long time back). I ran out of patience trying to resolve an issue. If you can work it out, then we have something. `netifd` causes excessive triggers wrt. DHCPv6. `netifd` asserts WAN6 renewal-like events for mere lifetime and route metric updates in RA messages. `procd` doesn't provide a filter mechanism to only respect real RENEW, material address, or route changes. `procd` helps with false starts and crashes, and is better on the application side for that. Sadly the old hotplug scripts worked better for network triggers.

drujd:
Sorry for the late reply. This change obviously has nothing to do with the issue you mentioned, but at the very least will make Unbound _not_ restart when the generated configuration is identical to the previous one - so a 'useless' trigger should not restart Unbound with this change.

If that does not help: I am not using IPv6 yet since my ISP does not support it (and I could not be bothered with tunnels etc.), but isn't then using hotplug scripts instead the solution? That is waiting for the 'ifupdate' action and then checking whether that is a meaningful change (checking whether IFUPDATE_ADDRESSES differs or something).

EricLuehrsen:
OKTM. This falls in the arena of continuous improvement. Lets set it wild and see if it works for everyone.

EricLuehrsen (approved):
OKTM. This falls in the arena of continuous improvement. Lets set it wild and see if it works for everyone. Sorry I have been busy with other things lately.
cc: @BKPepe

BKPepe:
``PKG_RELEASE`` needs to be bumped in the Makefile.

net/unbound/files/unbound.init
```diff
@@ -40,6 +40,7 @@ start_service() {
   procd_open_instance "unbound"
   procd_set_param command $PROG -d -c $UB_TOTAL_CONF
   procd_set_param respawn
+  procd_set_param file "$UB_TOTAL_CONF" "$UB_SRV_CONF" "$UB_EXT_CONF"
```
BKPepe, line 43: Would be nice to have some commit description, what is the reason, why it needs to be added now.
Based on https://openwrt.org/submitting-patches#submission_guidelines

1715173329:
one small nit, please warp a new line for every 70 chars in your commit message.

EricLuehrsen:
This is acceptable. I don't know why its sitting.

EricLuehrsen (approved):
This is acceptable. I don't know why its sitting.

drujd:
Well, **I** cannot merge it... :)

1715173329:
Merged, thanks!

PR #22697, opened 2023-11-18
netbird: update to version 0.24.2

oskarirauta:
I decided to put some effort on netbird.io package as they finally published iOS client (for selected audience for testing currently..) to be able to provide detailed feedback on all available bells and whistles.

Release notes: https://github.com/netbirdio/netbird/releases/tag/v0.24.2

Build fix:
	- fixed issue of not being able to build with musl

Changes:
After serious tests, I created a new rather complex init script for netbird and a optional netifd protocol support which allows you to easily set up a zone for netbird interface, or join existing zone, such as "vpn".

Philosophy:
Create necessary accounts for joining netbird network, whether you used global, or your own. Install package and start service with /etc/init.d/netbird start - this will start the background service. Next you need to command netbird up - at this stage if this is a new installation, you get request to open a link in a browser that is already logged in to netbird.io web service to authenticate your new device to network. After this, you should see wt0 interface appear (not to netifd, but to system) and you are all set up. You might want to set your device's login not to expire at all at web panel at this point, otherwise you end up needing to do this authentication again.

Now, you can go with that, or if you want more permanent setup with firewalling support etc, you should make interface available for netifd. Add following to your network config:

config interface 'wt0'
	option proto 'netbird'
	option device 'wt0'

Optionally, you can add option delay, it's default value is 2 and I landed on this value after testing thoroughly.

After this you have it visible in netifd and can command it with ifstatus, ifdown and ifup.

Init script works hand in hand with protocol, but if protocol is not used, then those features are ignored. If you don't use protocol, it works exactly the same as it used to so features do not require any changes from end-users if they want to stick to what it used to be.. I also have soon finished my protocol package for luci as well.

Someone should make a wiki entry on netbird at some point.

Maintainer: me

### DIFF
--- a/net/netbird/Makefile
+++ b/net/netbird/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=netbird
-PKG_VERSION:=0.23.9
+PKG_VERSION:=0.24.2
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/netbirdio/netbird/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=1b037f35d3e426d8cbeba17e4d89d12265cd7e6fbd7c975ce552293e468db35a
+PKG_HASH:=7fd90c6004c7fde6393bd618e106410e708f2d84f7884055acda6b016be42858
 
 PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
 PKG_LICENSE:=BSD-3-Clause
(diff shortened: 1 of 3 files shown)


oskarirauta:
Strange. Just a while ago it worked perfectly and now there's issue with protocol. I'll have to look into it..

oskarirauta:
For testing protocol in longer-term phase (few hours), I came to conclusion that even when it's as good as it is, it is not stable enough; problems ahead with it. Someone could probably make it better, but for now, I have decided to remove it. Instead, use cni protocol in place of netbird protocol, you just can't "manage" it anymore with ifup. I'll push new commits later that will remove all protocol stuff. I probably should rename cni protocol as name is mis-leading, to something else such as emp (externally managed protocol) or emgp (externally managed general protocol), name suggestions are accepted..

oskarirauta:
Unfortunately I have to re-do this PR.. Sorry.

PR #22701, opened 2023-11-18
gensio: add patch to fix python detection

mhei:
Maintainer: @WereCatf 
Compile tested: mxs
Run tested: -

Description:

This hopefully resolves the buildbot issues finally.
Upstream PR: https://github.com/cminyard/gensio/pull/69

### DIFF
--- a/net/gensio/patches/0001-Ensure-that-ax_python_devel_found-is-defined.patch
+++ b/net/gensio/patches/0001-Ensure-that-ax_python_devel_found-is-defined.patch
@@ -1,0 +1,29 @@
+From 6bbc3056c4b9192010d888672d97810609ee23f9 Mon Sep 17 00:00:00 2001
+From: Michael Heimpold <mhei@heimpold.de>
+Date: Sat, 18 Nov 2023 21:46:15 +0100
+Subject: [PATCH] Ensure that $ax_python_devel_found is defined
+
+Otherwise in case of --without-python, it triggers an error like:
+-snip-
+...
+checking consistency of all components of python development environment... yes
+./configure: line 23729: test: =: unary operator expected
+...
+-snap-
+
+Signed-off-by: Michael Heimpold <mhei@heimpold.de>
+---
+ configure.ac | 2 ++
+ 1 file changed, 2 insertions(+)
+
+--- a/configure.ac
++++ b/configure.ac
+@@ -1997,6 +1997,8 @@ if test "x$trypython" = "xyes"; then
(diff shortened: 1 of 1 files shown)


hnyman:
Seems to fail the patch format cleanness test. "refresh" the patch

mhei:
Upstream did not like my approach either, I've updated now.

mhei:
I'm really lost: the builbot's faillogs e.g. for [i386_pentium4](https://downloads.openwrt.org/snapshots/faillogs/i386_pentium4/packages/gensio/compile.txt) shows that the patch is applied, but the error line is still present:
`./configure: line 23731: test: =: unary operator expected`
Further down, we still see ` python:		no` which should not happen. Good news is that I finally managed to reproduce it on my local machine. But when I look at the `configure` script, then I only see 2 occurrences of `ax_python_devel_found`: the one patched-in and the failing test. That means that `ax_python_devel_found` is not set in the if-branch? But this should be the case since `ax_python_devel.m4` sets it early to yes and at some points it sets it to no, but it is never undefined.
Do I miss anything completely?
@cminyard: Do you have any hints?

mhei:
I think, there are several problems here: you are right, the ax_python_devel.m4 seems to not be used from the tarball, but a system-wide (I mean here: toolchain wide global one) seems to be included. This is because autoreconf is called from the OpenWrt's recipe. I don't understand yet why the system-wide one is used. One of the problem might be that both files (local one and system-wide one) have the same serial number (aka version number) set - but when I patch the included one to have a higher serial, then it still not yet used - this still confuses me. I created [gensio PR 71](https://github.com/cminyard/gensio/pull/71) to sync the file in gensio with latest upstream version.
Another problem seems to be the swig detection. Here too, since autoreconf is used, the system-wide version of ax_pkg_swig.m4 is used, but not the patched local version. I created [gensio PR 72](https://github.com/cminyard/gensio/pull/72).
I have to run autoreconf within OpenWrt, otherwise it bails out because of a libtool version mismatch.
Last problem seems that we also need to pass PYTHON_LIBS environment variable not only while configuring, but also when the actual compiling is done. I created a [follow-up PR](https://github.com/openwrt/packages/pull/22723) that addresses all these points. Then the current version should finally compile on the buildbots.

cminyard:
All merged, thank you for your work on this.  When all is good from your point of view, I'll do a new release.

mhei:
Yes, until I'm told better, I consider this a bug in OpenWrt :smile: Will try to discuss this on then openwrt-devel list, hope that there are more experts who can judge whether this is by design or really an issue.

PR #22702, opened 2023-11-19
external-protocol: rename and update cni-protocol

oskarirauta:
cni-protocol can be used for both cni and netavark and also for many other things, such as vpn's that lack customized protocol supports for openwrt as a general externally managed protocol, so it was time due to rename it.

I also added one extra option, search domain, which is optional and updated scripts retrieving ip address and routing information.

After merging, I will make a PR to update protocol's luci counterpart accordingly.

Maintainer: me


oskarirauta:
@hnyman sorry that delayed. I plan do that in next 6 hours..

PR #22718, opened 2023-11-21
ddns-scripts: update_gandi_net: Add support for Personal Access Token (PAT).

kevbo-oh:
Maintainer: @feckert
Run tested: openwrt-23.05, applied patch, manually "broke" DNS A record and verified that both key and token still work and can update the A record.

Description:
Gandi's LiveDNS now has a new Personal Access Token (PAT) that can be used instead of the API Key.  APIKeys are actually deprecated. A used account can have many PATs, each limited to the domains they can act on, and what type of API calls they can do.  https://docs.gandi.net/en/managing_an_organization/organizations/personal_access_token.html
The downside is that PATs expire.  The longest you can make is 1 year.  (In reality, this is probably better than infinite API keys, but I'm a bit concerned about DDNS being able to report a failure because a key expired.)
The Authorization header needs to changed, based on the type.  APIKeys say "Authorization: Apikey".  PATs say "Authorization: Bearer".
acme.sh 3.0.7 was updated to be able to use both types.  It determines what the Authorization header should say by having two separate environment variables you can use.  (I have provided a PR to update acme.sh.)
DDNS is trickier because there's only one field to use: "password".
My attempt here is to prepend "PAT:" to the password in the password field if the password represents a PAT.  If this prefix is detected, it is stripped, and the Authorization header is changed to "Bearer".  If the prefix isn't there, the plugin behaves as before.  This way, we won't break any existing apikey installs.
I also made sure that either logged value will still be replaced with ***PW***.  This basically means I have two versions of the password, the original $password to be used for logging (which would be _not_ stripped, so that the output value matches the config password, so that write_log detects it and replaces it with ***PW***), and $_apipassword, which is sent to the API, and would have the "PAT:" prefix stripped.
I am fully aware that this "magic" behavior may be unwelcome.  I am also concerned that there's no way to prompt the user and explain the functionality.
The only other way I could think to do this is to just fork the whole provider, and make update_gandi_net_pat.sh. Should that seem to be the better case, I'd be more than willing to change the PR. I wanted to submit this and at least get a discussion going.


kevbo-oh:
I don't know for sure who the maintainer of ddns-scripts is.  I see that @feckert  tends to be the one that merges pull requests, so I hope that you are the maintainer.

pekare:
With my limited testing the API keys seems to be [a-zA-Z0-9]{24} and PAT seems be [a-e0-9]{40}.
Perhaps a simple check of the length of the password parameter (possibly in conjunction with character types) would suffice?

kevbo-oh:
A length check to determine which kind of token is being presented is a good idea.  It would be "magic", in that the script would act differently based upon the length of the input.  It makes sense though, other than being hard to explain.  I'll work on it.

kevbo-oh:
OK, done.  I put this script on my running box. mucked up my DNS, and force it to update.  I tried with both a PAT and an API key: the script made the correct choice and returned my DNS entry to the correct value both times.

kevbo-oh:
As a personal note: I'd really like this patch to go in, so I can use a PAT and stop using the way more open API Key. So, anything I can do...

kevbo-oh:
Also, I submitted a support ticket to Gandi asking if the 24/40 character assumption holds. Hopefully they will respond.

feckert:
I don't know. I am not satisfied with the whole situation in the ddns-script package. It is completely broken!

There are way too many exceptions and thus errors.

Can't we just give the scripts the uci section the script is working on and then read a new type option, and then read either the Apikey option or the Bearer option?

kevbo-oh:
I haven't delved deeply into uci/luci. I was just trying to get this one script working. I honestly don't know how easy it is to "add fields" to uci/luci...how to go about deciding that the ddns config needed additional fields.  I also figured that was a "higher pay grade" decision that just making one provider script work with a new type of key.

I'm willing to look at it, I just don't quite get how the larger OpenWRT project works with respect o things like that.

kevbo-oh:
I got a response from Gandi support:

```
Dear Kevin,
I can confirm that

    API keys are always composed of 24 characters,
    API PATs currently consist of 40 characters, although we cannot guarantee that this will remain unchanged in the future.

 
I hope this information is helpful to you.
 
Best regards,
Jeremie
Gandi.net
```

If we use the "auto-detect" logic as is, it might make sense to check for length 24 for API keys, and assume anything larger than that is a PAT.

kevbo-oh:
I have actually asked a follow up, regarding the "cannot guarantee that this will remain unchanged". I asked if there is any chance the PAT would _decrease_ in length. I also asked them to make a decision to not allow the PAT to ever be 24 characters. I'll include any response.

net/ddns-scripts/files/usr/lib/ddns/update_gandi_net.sh
```diff
@@ -13,6 +13,19 @@ local __STATUS
...
+_len=${#password}
+if [ "$_len" -eq "24" ]; then
+    _auth="Authorization: Apikey "
+elif [ "$_len" -eq "40" ]; then
+    _auth="Authorization: Bearer "
+else
+    write_log 14 "Password wasn't length 24 or 40, cannot determine type"
+    return 1
+fi;
+#write_log 7 $_auth
```
feckert, line 27: I think you can remove this line

net/ddns-scripts/files/usr/lib/ddns/update_gandi_net.sh
```diff
@@ -13,6 +13,19 @@ local __STATUS
...
+local _auth
+_len=${#password}
+if [ "$_len" -eq "24" ]; then
+    _auth="Authorization: Apikey "
+elif [ "$_len" -eq "40" ]; then
+    _auth="Authorization: Bearer "
+else
+    write_log 14 "Password wasn't length 24 or 40, cannot determine type"
+    return 1
+fi;
```
feckert, line 26: Please use `tabs `for the indentation and not `spaces`.

feckert (changes requested):
I do not use the `gandi_net` ddns provider.
So I can only say whether the source code is OK.

- Please fix my comments.
- Please bump the PKG_RELEASE number by one https://gitlab.dev.tdt.de/dev/sys6/upstream/packages-trunk/-/blob/master/net/ddns-scripts/Makefile?ref_type=heads#L11

GeorgeSapkin:
No updates in over two years and not following guidelines. Closing as stale. Feel free to reopen as necessary.

PR #22722, opened 2023-11-21
lua5.4: Add new lua version

Ansuel:
Add new lua version 5.4 required by new version of nmap.

Patch copied from lua 5.3. Readline patch has to be reworked as lua 5.4 now supports no readline for linux but still needs some tweaks for macos and bsd systems.

Patch shared lib required some rework.

---

@BKPepe is this ok? Required for nmap bump

### DIFF
--- a/lang/lua5.4/Makefile
+++ b/lang/lua5.4/Makefile
@@ -1,0 +1,165 @@
+#
+# Copyright (C) 2006-2023 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lua
+PKG_VERSION:=5.4.6
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://www.lua.org/ftp/ \
+	https://www.tecgraf.puc-rio.br/lua/ftp/
+PKG_HASH:=7d5ea1b9cb6aa0b59ca3dde1c6adcb57ef83a1ba8e5432c0ecd06bf439b3ad88
+PKG_BUILD_PARALLEL:=1
+
+PKG_MAINTAINER:=Christian Marangi <ansuelsmth@gmail.com>
+
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=COPYRIGHT
+PKG_CPE_ID:=cpe:/a:lua:lua
+
+HOST_PATCH_DIR := ./patches-host
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/host-build.mk
+
+define Package/lua5.4/Default
+  SUBMENU:=Lua
+  SECTION:=lang
+  CATEGORY:=Languages
+  TITLE:=Lua programming language (version 5.4)
+  URL:=https://www.lua.org/
+endef
+
+define Package/lua5.4/Default/description
+  Lua is a powerful, efficient, lightweight, embeddable scripting language. It
+  supports procedural programming, object-oriented programming, functional
+  programming, data-driven programming, and data description.
+endef
+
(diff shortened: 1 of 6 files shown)


lang/lua5.4/Makefile
```diff
@@ -0,0 +1,164 @@
...
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/host-build.mk
+
+define Package/lua5.4/Default
+  SUBMENU:=Lua
+  SECTION:=lang
+  CATEGORY:=Languages
+  TITLE:=Lua programming language
+  URL:=https://www.lua.org/
+  MAINTAINER:=Christian Marangi <ansuelsmth@gmail.com>
```
BKPepe, line 35: You can move this above ``PKG_LICENSE``.

lang/lua5.4/Makefile
```diff
@@ -0,0 +1,164 @@
...
+endef
+
+define Package/lua5.4/Default/description
+  Lua is a powerful, efficient, lightweight, embeddable scripting language. It
+  supports procedural programming, object-oriented programming, functional
+  programming, data-driven programming, and data description.
+endef
+
+define Package/liblua5.4
+$(call Package/lua53/Default)
```
BKPepe, line 45: Why do we call lua53 in this Makefile? 🤔

BKPepe (commented):
I do have one question. Why lua package are located in the main repo and this one goes to packages repository? 🤔

lang/lua5.4/Makefile
```diff
@@ -0,0 +1,164 @@
...
+endef
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(1)/usr/include/lua5.4 $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/lua5.4/lua{,lib,conf}.h $(1)/usr/include/lua5.4/
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/lua5.4/lua.hpp $(1)/usr/include/lua5.4/
+	$(CP) $(PKG_INSTALL_DIR)/usr/include/lua5.4/lauxlib.h $(1)/usr/include/lua5.4/
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/liblua5.4.{a,so*} $(1)/usr/lib/
+	$(LN) liblua5.4.so.0.0.0 $(1)/usr/lib/liblualib5.3.so
```
BKPepe, line 143: is this correct? Should be there conflict, though?

lang/lua5.4/Makefile
```diff
@@ -0,0 +1,164 @@
+#
+# Copyright (C) 2006-2014 OpenWrt.org
```
BKPepe, line 2: Since this is a new package, shouldnt you update the copyright bcs you are affiliated with OpenWrt?

Ansuel:
> I do have one question. Why lua package are located in the main repo and this one goes to packages repository? 🤔 

From what I notice other lua version have some dependency on main packages while this doesn't

BKPepe:
> > I do have one question. Why lua package are located in the main repo and this one goes to packages repository? 🤔
> 
> From what I notice other lua version have some dependency on main packages while this doesn't

Yeah, but lua5.3 is [already out of life](https://www.lua.org/versions.html). Shouldn't be there some plan to update/tweak packages to use newer Lua version?

Ansuel:
@BKPepe no the actual move is to drop lua as a requirement. (the main usage of lua is LuCI that is migrating to js usage, aka the Lu part is being dropped AHAHAH)

Also different version of lua are not just a bump in version... each lua version  have ABI changes and are not compatible with each other... 99% of the time an app written for lua 5.1 is not compatible with lua 5.3... same with 5.3 and 5.4... (app needs to be converted to the new version and is not trivial also they sometimes introduce higher ram consumption and bigger lib, or arch problem)

BKPepe (approved):

BKPepe:
Now the formality check should pass. Let's see. 🤞

Issue #22674, opened 2023-11-16
https-dns-proxy: ubus command failed


systemcrash:
@stangri ?

Just upgraded to 23.05.2, `logread` shows this: 
```
# logread | grep https-dns-
Tue Nov 14 14:38:58 2023 user.notice https-dns-proxy: Starting service instances on_boot ✓✓✓
Tue Nov 14 14:39:25 2023 user.notice https-dns-proxy: Restarting dnsmasq on_boot ✓
Tue Nov 14 14:39:26 2023 daemon.info https-dns-proxy[5301]: [F] 1699969166.206376 https-dns-proxy-2023-10-25/src/dns_server.c:51 Error binding 127.0.0.1:5055: Address in use (-1)
Tue Nov 14 14:39:26 2023 daemon.notice procd: /etc/rc.d/S90https-dns-proxy: Command failed: ubus call service add { "type": "config.change", "data": { "package": "firewall" } } (Invalid argument)
Tue Nov 14 14:39:26 2023 daemon.notice procd: /etc/rc.d/S90https-dns-proxy: Usage: ubus [<options>] <command> [arguments...]
Tue Nov 14 14:39:26 2023 daemon.notice procd: /etc/rc.d/S90https-dns-proxy: Options:
Tue Nov 14 14:39:26 2023 daemon.notice procd: /etc/rc.d/S90https-dns-proxy:  -s <socket>:		Set the unix domain socket to connect to
Tue Nov 14 14:39:26 2023 daemon.notice procd: /etc/rc.d/S90https-dns-proxy:  -t <timeout>:		Set the timeout (in seconds) for a command to complete
Tue Nov 14 14:39:26 2023 daemon.notice procd: /etc/rc.d/S90https-dns-proxy:  -S:			Use simplified output (for scripts)
Tue Nov 14 14:39:26 2023 daemon.notice procd: /etc/rc.d/S90https-dns-proxy:  -v:			More verbose output
Tue Nov 14 14:39:26 2023 daemon.notice procd: /etc/rc.d/S90https-dns-proxy:  -m <type>:		(for monitor): include a specific message type
Tue Nov 14 14:39:26 2023 daemon.notice procd: /etc/rc.d/S90https-dns-proxy: 			(can be used more than once)
Tue Nov 14 14:39:26 2023 daemon.notice procd: /etc/rc.d/S90https-dns-proxy:  -M <r|t>		(for monitor): only capture received or transmitted traffic
[...]

PR #22737, opened 2023-11-23
https-dns-proxy: bugfix: prevent erros from boot()

stangri:
Maintainer: me
Compile tested: x86_64, Sophos XG-135r3, OpenWrt 23.05.0
Run tested: x86_64, Sophos XG-135r3, OpenWrt 23.05.0

Description:
* fixes https://github.com/openwrt/packages/issues/22674
* rename resolver_health_check to is_resolver_running for readability
* reorder functions in the init file by name


net/https-dns-proxy/files/etc/init.d/https-dns-proxy
```diff
@@ -131,13 +130,11 @@ append_bootstrap() {
...
 
-resolver_health_check() { resolveip -t 3 one.one.one.one >/dev/null 2>&1; }
-
 boot() {
 	ubus -t 30 wait_for network.interface 2>/dev/null
 	on_boot_trigger=1
-	rc_procd start_service 'on_boot' && rc_procd service_started 'on_boot'
-	resolver_health_check || rc_procd stop_service 'on_boot'
+	rc_procd start_service 'on_boot' && service_started 'on_boot'
+	is_resolver_working || rc_procd stop_service 'on_boot' && service_stopped 'on_boot'
```
matjon, line 137: Hello,
`service_stopped 'on_boot'` will likely get executed every time.

```
BusyBox v1.30.1 (Ubuntu 1:1.30.1-4ubuntu6.4) built-in shell (ash)
Enter 'help' for a list of built-in commands.

~ $ true || echo foo && echo bar
bar
~ $ false || echo foo && echo bar
foo
bar
~ $ 
```
Is this intentional?

matjon: The above is from busybox on Ubuntu 20.04, I don't have the router handy.

stangri: Thank you for spotting it, not intentional, I'll push a fix!

PR #21786, opened 2023-08-10
netifyd: Updated to v4.4.7

dsokoloski:
Maintainer: Darryl Sokoloski / @dsokoloski
Compile tested: arm_cortex-a15_neon-vfpv4, TP-Link Archer C2600, master
Run tested: TP-Link Archer C2600

## Bug Fixes

- Fixed mbedTLS crashes and TLS handshake errors when the nDPI-bundled libgcrypt "lite" version conflicts with the system version (via libcurl).
- Fixed linking order issue with libini.
- Fixed non-portable static linking warning with libndpi.
- Write flows to sockets regardless if "add_flows" is true.
- Fixed compilation error if _DIRENT_HAVE_D_RECLEN isn't defined.
- Fixed Agent path.
- [OpenWrt] Switch to "grep -E" as "egrep" is deprecated.
- Fixed possible ndAddr crash: return a const reference for cached strings.

PR #22749, opened 2023-11-25
netifyd: Updated to v4.4.7

dsokoloski:

## Release Notes
https://gitlab.com/netify.ai/public/netify-agent/-/releases/v4.4.7

### DIFF
--- a/net/netifyd/Config.in
+++ b/net/netifyd/Config.in
@@ -1,0 +1,16 @@
+menu "Configuration"
+	depends on PACKAGE_netifyd
+
+config NETIFYD_BUILD_SHARED_LIBRARY
+	bool "Build the Agent as a shared library?"
+	default y
+	help
+		Build the Agent as a shared library when enabled, or statically link it otherwise.  Statically linking can make debugging easier.
+
+config NETIFYD_WITH_LOCAL_LIBGCRYPT
+	bool "Link nDPI with the local system libgcrypt?"
+	default y
+	help
+		Link the Agent (and nDPI) against the local system libgcrypt, or use the bundled gcrypt lite version.
+
+endmenu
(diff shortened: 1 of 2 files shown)


BKPepe (changes requested):
I think, I am going to be strict this time.

1. From the commit itself, it is not clear that it is cherry-picked from the master branch.
You should do that by ``git cherry-pick -x hash``

Not sure, if I should comment those changes as it was reviewed in https://github.com/openwrt/packages/pull/21786, but let me share my two cents.

However, I see that in your commit description and as well you copy&pasted things from your release https://gitlab.com/netify.ai/public/netify-agent/-/releases/v4.4.7
**Rather point to it next time**, because these are not relevant to Makefile or Config.in. 

E.g.
```
Release notes: https://gitlab.com/netify.ai/public/netify-agent/-/releases/v4.4.7
```
It is sufficient.

I see there many things, which somehow slipped in and these are not described anywhere.

1. Why do we have options, when those are enabled by default?
If we disable them, then what is the size difference?

2. Why there is now libatomic dependency?

And I can keep continue asking for that...

dsokoloski:
> 1. From the commit itself, it is not clear that it is cherry-picked from the master branch.
>    You should do that by `git cherry-pick -x hash`

Ok, done.  This is the first time I've submitted a branch PR.  I'll remember for next time.

> Not sure, if I should comment those changes as it was reviewed in #21786, but let me share my two cents.
> 
> However, I see that in your commit description and as well you copy&pasted things from your release https://gitlab.com/netify.ai/public/netify-agent/-/releases/v4.4.7 **Rather point to it next time**, because these are not relevant to Makefile or Config.in.

Ok I think I understand... but I didn't do it correctly.  In the future, my commit should **only** contain OpenWrt-specific changes... and it sufficient to add the link to my GitLab release for the application-specific changes?

> I see there many things, which somehow slipped in and these are not described anywhere.
> 
> 1. Why do we have options, when those are enabled by default?
>    If we disable them, then what is the size difference?
> 2. Why there is now libatomic dependency?

Ok, I've added more details regarding what the new options are for and why libatomic is now required.

Let me know if there is anything else you'd like to see.

PR #22751, opened 2023-11-25
telegraf: Update to version 1.26.0

barteo:

Description:

### DIFF
--- a/utils/telegraf/Makefile
+++ b/utils/telegraf/Makefile
@@ -3,12 +3,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=telegraf
-PKG_VERSION:=1.23.4
+PKG_VERSION:=1.26.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/influxdata/telegraf/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=05188b5f0c0dfa204dc6bd8429ebc5366b73e42c7bdd4f1a50fffa2a1e75616f
+PKG_HASH:=ee6933a16930dfd8b32832f0ed9e0393bb14cdb973abc1a773bfb58976470ea8
 
 PKG_MAINTAINER:=Jonathan Pagel <jonny_tischbein@systemli.org>
 PKG_LICENSE:=MIT
--- a/utils/telegraf/Makefile
+++ b/utils/telegraf/Makefile
@@ -59,6 +59,12 @@ define Package/telegraf-full/description
 	(Full build. including all plugins)
 endef
 
+define Package/telegraf/conffiles
+/etc/telegraf.conf
+endef
+
+Package/telegraf-full/conffiles = $(Package/telegraf/conffiles)
+
 define Build/Prepare
 	$(call Build/Prepare/Default)
 ifeq ($(BUILD_VARIANT),small)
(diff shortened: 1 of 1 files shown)


utils/telegraf/Makefile
```diff
@@ -59,12 +59,6 @@ define Package/telegraf-full/description
 	(Full build. including all plugins)
 endef
 
-define Package/telegraf/conffiles
-/etc/telegraf.conf
-endef
-
-Package/telegraf-full/conffiles = $(Package/telegraf/conffiles)
```
jefferyto, line 66: Including /etc/telegraf.conf in `conffiles` means it gets preserved by opkg during package upgrades. You probably don't want to remove this section.

utils/telegraf/Makefile
```diff
@@ -76,7 +70,7 @@ define Package/telegraf/install
 	$(call GoPackage/Package/Install/Bin,$(1))
 	$(INSTALL_DIR) $(1)/etc/init.d $(1)/etc/config
 	$(INSTALL_BIN) ./files/etc/init.d/telegraf $(1)/etc/init.d/telegraf
-	$(INSTALL_CONF) $(PKG_BUILD_DIR)/etc/telegraf.conf $(1)/etc/telegraf.conf
+	$(1)/usr/bin/telegraf config > $(1)/etc/telegraf.conf
```
jefferyto, line 73: `telegraf` here is a program cross-compiled for the target device. It almost certainly will not run on the build system.

utils/telegraf/Makefile
barteo: Starting from 1.26.0 version `etc/telegraf.conf` is no longer included in repo:
https://github.com/influxdata/telegraf/pull/12766

It can be generated using `telegraf config` command though, ether with: `make config` or `telegraf config`

I fully agree that `telegraf.conf` has to be somehow preserved in package upgrades. I haven't figured out how to do it using OpenWrt Makefile.

utils/telegraf/Makefile
barteo: Right, this was my unsuccessful experiment to generate telegraf.conf file facing my limited knowledge of OpenWrt Makefile system.

utils/telegraf/Makefile
jefferyto: The config file can be preserved in package upgrades by keeping the `Package/telegraf/conffiles` and `Package/telegraf-full/conffiles` definitions in the makefile.

utils/telegraf/Makefile
jefferyto: I suggest adding a [uci-defaults](https://openwrt.org/docs/guide-developer/uci-defaults) script that (after testing if the config file already exists) generates the config file. This script will be run on package install or on first image boot (if the package was integrated into a custom image).

A [postinst](https://openwrt.org/docs/guide-developer/packages#packagepostinst) script is usually another option but it gets run when a custom image is created, which is still on the build system (not on the target).

vortexilation:
My build system or host debian SID x86_64 glibc, I am targeting x86_64 musl, whenever : 
```
$(1)/usr/bin/telegraf config > $(1)/etc/telegraf.conf
```
is executed, the telegraf binary seems cannot be executed because looks like it being linked with glibc instead musl (target) ? : 
```
/home/user/works/openwrt/build_dir/target-x86_64_musl_siliwangi/telegraf-full/telegraf-1.28.1/ipkg-install/usr/bin/telegraf: error while loading shared libraries: /lib/x86_64-linux-gnu/libc.so: invalid ELF header
```

Given use case above, Isn't it's useful if we have host build for go lang?.

BKPepe:
Closing this pull request due to inactivity.

PR #22761, opened 2023-11-27
perl: update to version 5.38.1

predators46:
Compile tested: armsr, riscv64
Run tested: armv8

Description:

perl update to version 5.38.1

delete perlbase-bigint and move it to perlbase-bignum

updated one patch

removed one that no longer exists in perl-5.38.1

fixed missing builtin.pm


predators46:
> I don't completely understand the need to repackage `bigint`.

bigint.pm is part of bignum so we don't need the perlbase-bigint package, it's better to move bigint.pm to the perlbase-bignum package. If this is not accepted I will change it later

PR #22766, opened 2023-11-27
xmlrpc-c: update to 1.54.06 and fix the abyss feature

yegorich:
Maintainer: @thess
Compile tested: ARMv7, TI Sitara AM335x, OpenWrt v23.05.2
Run tested: ARMv7, TI Sitara AM335x, OpenWrt v23.05.2, tested with internal software

Description:

Update version to 1.54.06 i.e. the latest version available as a tarball.

This PR also fixes the following build error breaking the abyss components:

xmlrpc_server_abyss.c:783:13: warning: format not a string literal and no format arguments [-Wformat-security]
  783 |             xmlrpc_faultf(envP, error);

PR #22783, opened 2023-11-28
libs/xmlrpc-c: add menu configuration options

yegorich:
Maintainer: @thess
Compile tested: ARMv7, TI Sitara AM335x, OpenWrt v23.05.2
Run tested: ARMv7, TI Sitara AM335x, OpenWrt v23.05.2, tested with internal software

Description:

Convert the package to Config.in format.

Also enable libxml2 support via providing an XML backend selection menu.

### DIFF
--- a/libs/xmlrpc-c/Config.in
+++ b/libs/xmlrpc-c/Config.in
@@ -1,0 +1,29 @@
+# xmlrpc-c configuration
+menu "Configuration"
+	depends on PACKAGE_xmlrpc-c
+
+comment "XML backend"
+
+choice
+	prompt "Selected XML backend"
+	default XMLRPC_C_INTERNAL
+
+	config XMLRPC_C_INTERNAL
+		bool "Internal"
+
+	config XMLRPC_C_LIBXML2
+		bool "libxml2"
+
+endchoice
+
+comment "Features"
+
+config XMLRPC_C_CLIENT
+	bool "XMLRPC client"
+	default n
+
+config XMLRPC_C_SERVER
+	bool "XMLRPC server"
+	default y
+
+endmenu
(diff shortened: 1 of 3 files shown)


yegorich:
So far, there are two confilcts with the following packages:

WARNING: Makefile 'package/feeds/packages/rtorrent/Makefile' has a dependency on 'xmlrpc-c-server', which does not exist
WARNING: Makefile 'package/feeds/telephony/rtpengine/Makefile' has a dependency on 'xmlrpc-c-client', which does not exist
WARNING: Makefile 'package/feeds/telephony/rtpengine/Makefile' has a dependency on 'xmlrpc-c-client', which does not exist

So, they have to be updated too.

yegorich:
@predators46 could you ive it a try with both XML backends?

jefferyto:
What problem is this trying to solve, and/or why is this an improvement?

In the original makefile, both internal and libxml2 versions are built, and both server and client packages are built. Having to select options makes this less useful to those who install prebuilt packages from the packages feed.

yegorich:
I'm coming from the Buildroot project. I still haven't got the packages mindset. I was trying to solve libxml2 backend breakage.

Fow now, it is not possible to build xmlrpc-c server or client and libxml2 backend. Because if you select `xmlrpc-c-server`, it depends on `xmlrpc-c-internal` and thus, tries to also install `libxmlrpc_xmltok.so` that is of course not available if you build with libxml2 backend:

cp: cannot stat '/home/user/openwrt/build_dir/target-arm_cortex-a8+vfpv3_musl_eabi/xmlrpc-c-libxml2/xmlrpc-c-1.51.07/ipkg-install/usr/lib/libxmlrpc_xmltok.so*': No such file or directory

Perhaps, the best solution would be to introduce xmlrpc-c-full package with libxml2 and OpenSSL support?

neheb:
This branch has conflicts that must be resolved

yegorich:
Closed in favour of #22766.

PR #22785, opened 2023-11-28
uspot: add package

f00b4r0:
Maintainer: me
Compile tested: ath79, ramips - OpenWrt 23.05
Run tested: ath79, ramips - OpenWrt 23.05

Description:
A captive portal system for OpenWrt

If accepted, a backport to 23.05 would be nice: it's been fairly thoroughly tested there :)

### DIFF
--- a/net/uspot/Makefile
+++ b/net/uspot/Makefile
@@ -1,0 +1,96 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=uspot
+PKG_RELEASE:=1
+
+PKG_LICENSE:=GPL-2.0
+PKG_MAINTAINER:=Thibaut VARÈNE <hacks@slashdirt.org>
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/f00b4r0/uspot.git
+PKG_SOURCE_DATE:=2023-11-30
+PKG_SOURCE_VERSION:=7e1e21b0f8425205d719b99a392fa893b3e512e6
+PKG_MIRROR_HASH:=494c616159b16d978fe00348ebe50c77a48f1db98d624ed613f3cca2d39e3a6e
+
+CMAKE_SOURCE_SUBDIR:=src
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+define Package/uspot
+  SUBMENU:=Captive Portals
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=uspot hotspot daemon
+  DEPENDS:=+conntrack \
+	   +libblobmsg-json +liblucihttp-ucode +libradcli +libubox +libubus +libuci \
+	   +spotfilter \
+	   +ucode +ucode-mod-math +ucode-mod-nl80211 +ucode-mod-rtnl +uhttpd-mod-ucode +ucode-mod-uloop
+
+endef
+
+define Package/uspot/description
+  This package implements a captive portal supporting click-to-continue,
+  simple credential-based as well as RADIUS authentication.
+  It is UAM capable, and has limited support for RFC5176
+  RADIUS Dynamic Authorization Extensions.
+  It is meant to be a drop-in replacement for CoovaChilli,
+  leveraging the performance and flexibility of the nftables firewall
+  without the need for extra kernel modules.
+endef
+
+define Package/uspot/conffiles
+/etc/config/uspot
+endef
+
+define Package/uspot-www
+  SUBMENU:=Captive Portals
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=uspot default user interface files
+  DEPENDS:=+uspot
+endef
+
+define Package/uspot-www/description
+  This package provides CSS and HTML templates for uspot UI.
+  This package must be installed with uspot unless a local alternative is provided.
+endef
+
+define Package/uspotfilter
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=uspot implementation of spotfilter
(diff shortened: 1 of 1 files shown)


net/uspot/Makefile
```diff
@@ -0,0 +1,95 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=uspot
+PKG_RELEASE:=1
+
+PKG_LICENSE:=GPL-2.0
+PKG_MAINTAINER:=Thibaut VARÈNE <hacks@slashdirt.org>
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/f00b4r0/uspot.git
```
BKPepe, line 10: I am not so sure if we should accept newly created packages, because it will be used by a few users, but it will be available for all routers, etc. So, it will end up to provide quantity over quality and it will use so many resources.

stintel: This is the packages feed, and with uspot people can finally ditch the abomination that is CoovaChilli.

stintel (approved):
Minor nit: I'd order DEPENDS alphabetically, and if the list is long, one item per line. Other than that: LGTM.

stintel:
Oh and just maybe a little description of what it is in the commit message. And that it can replace CoovaChilli without the need for horrible external kernel modules.

net/uspot/Makefile
stintel: The packages feed has been a mess for years, but you completely ignore my valid argument that it can replace CoovaChilli.

BKPepe: If I will rephrase it:

1. Are you aware that we do have so many packages here, which are not maintained anymore here?
The reasons:
We do have lack of maintainers (here) - no new contributors
Upstream projects are dead - what we should do about it? Keep it here and act like nothing is happening?

== **Just look at PCRE2 transition.**

2. Usually new projects are like huray in the beginning and over the time, if there are no users or developers (changed priorities), etc. then it goes to maintenance phase and later, the project is archived / read - only.

3. What is the audience for this package? You said "ditch the Coovachilli" what you dont like it about? Why you raised it only in comment?

We are still adding new packages here without proper policy, that what I am repeating every single time here. Honestly, we can add every single package. Is that what we want? I don't think so. That's why we should vote for it like main repo does.

And I can keep asking more questions, if this is what you want. How many packages you reviewed recently or updated?

stintel: 1. Yes. But there's a truckload of packages that don't even make sense on OpenWrt's primary target (routers/APs). That is not true for this new submission.
2. So because some projects disappear, we should block all new package submissions? How is innovation supposed to happen?
3. People who want to use a captive portal, without the need for a an external kernel module (kmod-ipt-coova). Coova has been around for years, and all I heard about it is that it's unmaintainable because of that external kernel module requirement.

As for how many packages I've reviewed/updated recently: not many, because most of the packages make no sense to me to use on OpenWrt.

net/uspot/Makefile
```diff
@@ -0,0 +1,101 @@
...
+PKG_SOURCE_VERSION:=7e1e21b0f8425205d719b99a392fa893b3e512e6
+PKG_MIRROR_HASH:=494c616159b16d978fe00348ebe50c77a48f1db98d624ed613f3cca2d39e3a6e
+
+CMAKE_SOURCE_SUBDIR:=src
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+
+define Package/uspot
```
1715173329, line 21: Please remove blank lines and sort these packages:

```
define Package/one
endef

define Package/one/description
endef

define Package/two
endef

define Package/two/description
endef

...

define Package/one/install
endef

define Package/two/install
endef

...



```

1715173329 (changes requested):

net/uspot/Makefile
f00b4r0: In that ordering scheme, where do `define Package/one/conffiles` go?

1715173329 (approved):

f00b4r0:
> @f00b4r0
> 
> It would seem interesting if this could replace coovachilli. 

It can. It already does in field testing.

> have you tried daloradius using uspot?

No. Uspot works well with freeradius, for which daloradius appears to be a GUI frontend.

f00b4r0:
@1715173329 thanks a lot. Should I open a separate PR for backport into 23.05 (uspot has been fully tested there) or will you cherry-pick this commit?

1715173329:
> @1715173329 thanks a lot. Should I open a separate PR for backport into 23.05 (uspot has been fully tested there) or will you cherry-pick this commit?

I'm glad to cherry pick it ;)

1715173329:
done in d83ed18f8d78ea257a2177177ca16b3745bab722

f00b4r0:
> done in [d83ed18](https://github.com/openwrt/packages/commit/d83ed18f8d78ea257a2177177ca16b3745bab722)

Thanks again! :)

bauw2008:
tmp/.config-package.in:140041:error: recursive dependency detected!
tmp/.config-package.in:140041:	symbol PACKAGE_prometheus-node-exporter-ucode depends on PACKAGE_ucode
tmp/.config-package.in:39962:	symbol PACKAGE_ucode is selected by PACKAGE_uspotfilter
tmp/.config-package.in:126340:	symbol PACKAGE_uspotfilter is selected by PACKAGE_uspot
tmp/.config-package.in:95844:	symbol PACKAGE_uspot depends on PACKAGE_uhttpd
tmp/.config-package.in:118531:	symbol PACKAGE_uhttpd is selected by PACKAGE_prometheus-node-exporter-ucode

Issue #22469, opened 2023-10-21
domoticz package does not start in 23.05.0


rolandvans:
domoticz package does not work in 23.05.0

It will not start because the Python version 3.11 is not supported and this is the default Python version in 23.05.0.
`Status: PluginSystem: Failed dynamic library load, install the latest libpython3.x library that is available for your platform.`

Changing the file `990-python3.10_fix.patch`, replacing "python3.10" with "python3.11" 
```
#else
+				if (!shared_lib_) FindLibrary("python3.11", true);
 				if (!shared_lib_) FindLibrary("python3.9", true);
 				if (!shared_lib_) FindLibrary("python3.8", true);
 				if (!shared_lib_) FindLibrary("python3.7", true);
@@ -198,6 +192,9 @@ namespace Plugins {
...

```
Fixes the status message, but domoticz does not start properly on boot.
Manually starting domoticz does work:
`domoticz -sslwww 0 -approot /usr/share/domoticz/ -noupdates -userdata /var/lib/domoticz -syslog -daemon`

BKPepe:
Ping maintainer @dwmw2 about this issue.

rolandvans:
No response from the maintainer until now

WLKe001:
Try make symlink for libpython3.11.so and libpython3.11.a to libpython3.9.so and libpython3.9.a
This working for me ;-)

rolandvans:
> Try make symlink for libpython3.11.so and libpython3.11.a to libpython3.9.so and libpython3.9.a This working for me ;-)

That trick solves the libpython part, but in my setup, it still does not start on boot. Syslog shows this message:
`Fri Nov 17 07:00:40 2023 daemon.notice procd: /etc/rc.d/S95done: Command failed: ubus call service delete { "name": "domoticz" } (Not found)`
Starting it manually _does_ work. Any idea what could be the problem here?

BKPepe:
PIng @dwmw2.

Let's ping previous contributors, who was interested in it as well - @lucize, @stintel.

dwmw2:
Apologies for the delayed response. I've updated to 2023.2 on the master branch, will look at backporting just the libpython3.11 fix to stable, once I've got local builds of 23.05 actually working on my hardware.

Not sure about @rolandvans's issue with it not starting though. That seems to be something different.

dwmw2:
@rolandvans Could your other startup issue be caused by https://github.com/openwrt/openwrt/issues/9545 ?

[...]

PR #22786, opened 2023-11-28
domoticz: update to 2023.2

dwmw2:
Maintainer: me / @neheb / @lucize 
Compile tested: arm_cortex-a7_neon-vfpv4, u7623, OpenWrt 23.05
Run tested: arm_cortex-a7_neon-vfpv4, u7623, OpenWrt 23.05, configure with OpenZWave and see devices

Description:
Update to 2023.2 release.

This is the last release which will support libopenzwave. Trying to get zwave-js-ui working in https://github.com/nxhack/openwrt-node-packages/pull/1826 as it'll be needed for the next update.


BKPepe:
Does it fixes https://github.com/openwrt/packages/issues/22469 ? Shouldnt it be backported also to the stable branches?

dwmw2:
> Does it fixes #22469 ? Shouldnt it be backported also to the stable branches?

It'll fix the libpython part but not the other issue. I'm trying to get 23.05 to actually install on my real Domoticz server so I can do some more testing. I think the one-line fix is probably more appropriate for the stable branch than a full upgrade to 2023.2 though.

PR #22806, opened 2023-12-03
ddns-scripts: Fix ddns not starting when booting

suisuiniang:
Maintainer: @cschoenebeck
Compile tested: x86_64
Run tested: x86_64 snapshots

Description:
Remove invalid functions

### DIFF
--- a/net/ddns-scripts/files/etc/init.d/ddns
+++ b/net/ddns-scripts/files/etc/init.d/ddns
@@ -2,10 +2,6 @@
 START=95
 STOP=10
 
-boot() {
-	return 0
-}
-
 reload() {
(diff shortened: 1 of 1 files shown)


feckert:
The ddns script is event driven with the following script https://github.com/openwrt/packages/blob/master/net/ddns-scripts/files/etc/hotplug.d/iface/ddns. Therefore the script should not start when booting. 
I think that the boat is already right. It should prevent the service from starting at boot time.

lxhao61:
Why was this fix update closed without approval?

feckert:
@lxhao61 As already written before. This is not a service that is started at bootup but is started via hotplug from netifd interface events.

PR #22815, opened 2023-12-04
python-rpds-py: Update to 0.13.2

jefferyto:
Maintainer: me
Compile tested: armsr-armv7, 2023-12-03 snapshot sdk
Run tested: armsr-armv7 (qemu), 2023-12-03 snapshot

Description:


jefferyto:
The mips/mipsel CI error is https://github.com/PyO3/pyo3/issues/3614. This was fixed in https://github.com/PyO3/pyo3/pull/3619 but it may take a while for the fix to be released. I don't believe there is any easy way for us to patch in the fix.

The only packages that depend on rpds-py are jsonschema and referencing, but they don't need the newest version ([0.7.1](https://github.com/python-jsonschema/jsonschema/blob/v4.20.0/pyproject.toml#L40) and [0.7.0](https://github.com/python-jsonschema/referencing/blob/v0.31.1/pyproject.toml#L38) or newer, respectively). I'm going to close this and wait for the next Py03/rpds-py release.

PR #22831, opened 2023-12-07
Package iperf3-mt-ssl is missing dependencies for the following libra…

scott0122:
Package iperf3-mt-ssl is missing dependencies for the following libraries: libiperf.so.0

### DIFF
--- a/net/iperf3-mt/Makefile
+++ b/net/iperf3-mt/Makefile
@@ -45,7 +45,7 @@ define Package/iperf3-mt-ssl
 $(call Package/iperf3-mt/default)
   TITLE+= and iperf_auth support
   VARIANT:=ssl
-  DEPENDS:=+libopenssl +libatomic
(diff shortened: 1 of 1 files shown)


jonasjelonek:
This PR lacks a proper title, description, commit message and signed-off-by line.
I'm not clear how this problem comes up, the ssl variant is statically compiled without the shared libiperf.so.0 so it should not depend on libiperf3-mt. However, there's still a known build problem between iperf3's ssl variant and libiperf3 (same applies to iperf3-mt) when both are selected. But since you provided no further information on your problem I could only guess.

Anyway, I just created a PR to update iperf3 to the latest version in this branch (#22832), and another one to remove iperf3-mt (#22833) because the multithreading feature is now included in 3.16. 
So I suggest to close this, I couldn't notice any unexpected problem with the iperf3-ssl package.

1715173329:
closed in favor of #22833

PR #22843, opened 2023-12-08
lvm2: remove dependency on libreadline by default

sah-cdo:
Maintainer: @dangowrt 
Compile tested: arm/aarch64 / OpenWrt 22.05
Run tested: arm/aarch64 / OpenWrt 22.05

Description:
Remove dependency on GPLv3 package.

### DIFF
--- a/utils/lvm2/Makefile
+++ b/utils/lvm2/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=LVM2
 PKG_VERSION:=2.03.25
 PKG_VERSION_DM:=1.02.199
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME).$(PKG_VERSION).tgz
 PKG_SOURCE_URL:=https://sourceware.org/pub/lvm2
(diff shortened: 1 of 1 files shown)


dangowrt:
Imho we don't need variants with readline. We don't need a fully-featured interactive CLI, it's enough to use lvm inside normal shell and for that we don't need readline.

BKPepe (changes requested):
1. Squash commits
2. In the commit description, there should be reason, why libreadline was removed.
3. Of course, while adding/removing dependencies, then there should be release bump.

sah-cdo:
@BKPepe : could you provide a feedback on this MR?

Rondom (approved):
LGTM

dangowrt:
You should still add `--disable-readline` to CONFIGURE_ARGS, so libreadline won't picked up during build in case it is present.

Matthiasfranck:
Hi @BKPepe @Rondom @dangowrt, all comments are fixed and I think we should be good to go now :)

GeorgeSapkin:
@sah-cdo still relevant? If so there are merge conflicts.

dangowrt:
Superseded by 2a1e7734ed055a3107d673f3d147c0031ded532f

PR #22853, opened 2023-12-10
adblock-fast: bugfix: block domains from config when not using block-lists

stangri:
Maintainer: me

Description:
* remove empty lines from the combined list to allo optimization code to work properly


BKPepe:
Your commit subject is too long.

PR #22861, opened 2023-12-10
openssh-keygen: Make ssh-keygen as an alternative to dropbearkey

stokito:
Maintainer: @tripolar
Compile tested: master, VirtualBox
Run tested: VirtualBox
Description:

The PR is part of https://github.com/openwrt/openwrt/pull/14174 so please read motivation there.

If will be in a draft state unless a new version of the Dropbear is released.


systemcrash:
Dropbox? Or dropbear?

PR #22894, opened 2023-12-14
wifidog: make it compile again with recent version of WolfSSL

Ansuel:
Recent version of WolfSSL dropped CyaSSL shims and made the package not compilable again. Converting it to the WolfSSL is simple enough as the API used are very basic and can be converted directly. Add patch that fully conver the package to WolfSSL and doesn't use the compat shim anymore.

Maintainer: Michael Haas <haas@computerlinguist.org> @mhaas


Ansuel:
If this is OK. I will take care of backporting this also to 23.xx as it's currently broken also there.

neheb (approved):

PR #22898, opened 2023-12-15
strongswan: trigger reload when interfaces are specified

### LINKED ISSUE #20848: strongswan: "running with no instances"
Maintainer: @pprindeville 
Environment: aarch64 Linksys E8450, OpenWrt 22.03.4

Description:

I've noticed that when I specify `ipsec.interface` on `/etc/config/ipsec`, occasionally Charon doesn't actually listen for connections. `/etc/init.d/swanctl status` says `running with no instances`.

I tried to dig through the code and my main suspect is: https://github.com/openwrt/packages/blob/8686e69765d2e2886c7819eacb55bbd7bebe170e/net/strongswan/files/swanctl.init#L715

I believe that's a problem when swanctl-init runs before the interfaces are actually registered (maybe a race with interface bringup?)

I'm not sure where this problem originated, but maybe the fix is to provide a default /etc/hotplug.d/iface script that reloads the service every time an interface change is detected? If that seems to be the fix I'll be happy to provide a patch.

lowjoel:
Compile tested: na
Run tested: aarch64/mt7622 Linksys E8450

Description:

Add interface triggers if interfaces to listen to are specified in `/etc/config/ipsec`. This fixes the "running with no instances" scenario after rebooting a router.

I've been running this for a month to good effect.

Fixes #20848

### DIFF
--- a/net/strongswan/Makefile
+++ b/net/strongswan/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=strongswan
 PKG_VERSION:=5.9.13
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=https://download.strongswan.org/ https://download2.strongswan.org/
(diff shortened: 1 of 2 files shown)


pprindeville:
Needs `PKG_RELEASE` bump.

lowjoel:
Thanks @pprindeville -- is this what you meant?

pprindeville (approved):
LGTM

PR #22908, opened 2023-12-17
snort3: compile with lzma support

graysky2:
This PR adds the ability of snort to process rules that target swf and pdf files requiring lzma decompression to look for
malicious payloads therein.  This change only increases the size of the snort3 executable by a fraction of a KB and the added
dependency of liblzma (based on currently offered 5.4.4-1) is only a 169 KB shared object.  Based on CPU requirements of snort,
x86 users likely represent the majority user-base and space their rootfs is not an issue as it may be for lower-powered SoCs.

Size of snort3-3.1.76.0-2: 7354403 bytes
Size of snort3-3.1.76.0-3: 7354435 bytes

Build system: x86/64

Maintainer: me and @flyn-org

### DIFF
--- a/net/snort3/Makefile
+++ b/net/snort3/Makefile
@@ -25,7 +25,7 @@ define Package/snort3
   SUBMENU:=Firewall
   SECTION:=net
   CATEGORY:=Network
-  DEPENDS:=+libstdcpp +libdaq3 +libdnet +libopenssl +libpcap +libpcre +libpthread +libuuid +zlib +libhwloc +libtirpc @HAS_LUAJIT_ARCH +luajit +libatomic +kmod-nft-queue
+  DEPENDS:=+libstdcpp +libdaq3 +libdnet +libopenssl +libpcap +libpcre +libpthread +libuuid +zlib +libhwloc +libtirpc @HAS_LUAJIT_ARCH +luajit +libatomic +kmod-nft-queue +liblzma
   TITLE:=Lightweight Network Intrusion Detection System
   URL:=http://www.snort.org/
   MENU:=1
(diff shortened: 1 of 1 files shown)


graysky2:
Recommend merge.

BKPepe:
> Recommend merge.

?? :-) You created PR and you recommend it to merge, right? Did I understand it, right? 😆

graysky2:
I did not realize I had direct commit abilities.  Thought someone else had to do it.

graysky2:
@BKPepe - do permissions need to be granted for me to commit?

```
% git push
ERROR: Permission to openwrt/packages.git denied to graysky2.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

And
```
% cat .git/config
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
[remote "origin"]
	url = git@github.com:openwrt/packages.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[branch "master"]
	remote = origin
	merge = refs/heads/master
```

BKPepe:
I think we are still not understanding each other. Of course, commit rights
needs to be granted and it needs to be deserved. My point was something
entirely different.

You created PR and the same person, who created PR recommeds his changes to
be merged. Thats quite obvious, isnt it? Your comment is really pointless
and the second maintainer should write it. Not you.

Dne po 18. 12. 2023 22:37 uživatel graysky ***@***.***>
napsal:

graysky2:
Thank you for the clarification. @flyn-org - can you please review?

BKPepe (changes requested):
In the commit description, we got why do you need lzma support, but we need more details. Why it should be enabled by default? What is the size difference, etc? 

Shouldn't it be done rather as configuration option (e.g. similarly done as Config.in)?

graysky2:
@BKPepe - I think I addressed your questions in the commit message.  As to your configuration option question, I think this represents the most straight-forward implementation.  Perhaps @flyn-org can comment on that point.

BKPepe (approved):

PR #22910, opened 2023-12-17
snort3: add missing action-override option

efahl:
Allow use of rules as-defined, and don't override their actions.  This is generally the best way to use the ruleset, and overriding their actions should only be undertaken when you fully understand how it affects their use.

Run tested: x86/64


flyn-org:
Looks good to me.

efahl:
Rebased to 5cdf7d58836451ffa8ad0906f8a728d9216a9b5a

@graysky2 , look good to you?

PR #22919, opened 2023-12-18
Add missing python3-bleak dependency

QuintinHill:
Maintainer: me
Compile tested: fairly trivial will rely on CI for this.
Run tested: ath79 wndr3700v4 manual installed dependency

Description: fix up of 64fa106.

### DIFF
--- a/lang/python/python-bleak/Makefile
+++ b/lang/python/python-bleak/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=python-bleak
 PKG_VERSION:=0.21.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PYPI_NAME:=bleak
 PKG_HASH:=ec4a1a2772fb315b992cbaa1153070c7e26968a52b0e2727035f443a1af5c18f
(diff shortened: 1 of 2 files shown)


QuintinHill:
CI errors are:  unable to download https://downloads.openwrt.org/snapshots/packages/mips_24kc/telephony/Packages.gz and similar.

BKPepe (changes requested):
1. Adding new dependencies requires to bump PKG_VERSION.
2. Avoid vague commit subjects as updating it to the latest version, of course... you are updating it each time to the latest version, which is quite obvious. This just bloats the commit history and once you will dig it in the past, then it would be really hard, so please be more specific and concrete on this one. There isnt anything wrong to said: "update to version X.Y.Z". 
3. Also, it would be handy if you can add changelog (link is just fine) to the commit description as well.

QuintinHill:
@BKPepe sorry about those omissions.  Corrected now! Thank you very much for the review.

jefferyto:
Can you add a `Fixes:` tag to the commit message for python-bleak, referencing https://github.com/openwrt/packages/commit/64fa106b0711a28732495b9cd334cfd6dfad9683? See https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-changes (near the bottom of that section).

BKPepe:
Oh yeah, @jefferyto has very good idea, please do that. Maybe if I am thinking more, could be interesting to know why the missing dependency is required now, which Python package requires it, etc.

QuintinHill:
Updated.

QuintinHill:
@BKPepe /@jefferyto these should all be addressed including a link to the bleak commit that added the dependency requirement.

QuintinHill:
Also note @predators46 has tested the python3-dbus-fast change in https://github.com/openwrt/packages/pull/22965.

predators46:
> Also note @predators46 has tested the python3-dbus-fast change in #22965.

I don't see this. if you want to update python-dbus-fast i will close update python-dbus-fast if needed

QuintinHill:
I don't mind which commit goes in (but mine does have the changelog in the commit message at the request of reviewers)!

PR #22920, opened 2023-12-19
gstreamer1 suite update to 1.22.8

neheb:
Maintainer: @flyn-org


flyn-org:
Is this the last package that depends on libsoup rather than libsoup3? Would you be interested in instead modifying the GStreamer build to use libsoup3? If not, then we could merge this, and I could try getting libsoup3 to work later.

neheb:
It is indeed. Note the first commit which fixes compilation with newer libxml2. I have not looked into conversion to libsoup3.

flyn-org:
Lets merge, and I will do the libsoup3 work later.

PR #22924, opened 2023-12-19
syncthing: fix group in syncthing.init

wxjstz:
The goup does not always have the same name as user, and when I try to run syncthing with nobody I get an error. Since nobody belongs to the group nogroup, I added some code to find out which group user belongs to.

Maintainer:  @aparcar @brvphoenix

Description:


efahl:
I believe you need to increment the value of PKG_RELEASE in `Makefile` as this is a functional change and not a version update.

You should change Maintainer to reference @aparcar and @brvphoenix , above.

The init file changes look good to me.

PR #22938, opened 2023-12-20
libassuan: Fix license

ffontaine:
libassuan license is wrong since the addition of the package in commit https://git.openwrt.org/?p=feed/packages.git;a=commit;h=e24e8fa98c813911419271d64433deb2b453fa02

Indeed, libassuan has been licensed under LGPL-2.1+ since version 1.0.3 back in 2007 [1]:

> Noteworthy changes in version 1.0.3 (2007-08-24)
> ------------------------------------------------
> 
>  * Changed the license of the library code back to LGPLv2.1 to support a bunch of GPLv2(only) software which does not allow the use of LGPLv3.  Note that this is only a temporary change and authors of GPLv2(only) software are asked to switch to GPLv3 or to add an exception which allow the use of LPGLv3 software.

[1]: https://git.gnupg.org/cgi-bin/gitweb.cgi?p=libassuan.git;a=blob;f=NEWS;h=3a86eca4175fbcb12bb00722c2047062df67a46d;hb=HEAD

Maintainer: @dangowrt
Compile tested: Not needed
Run tested: Not needed


BKPepe:
I'd say that this requires to bump PKG_RELEASE. Am I right?

ffontaine:
Why? Because of the addition of `COPYING.LIB` to `PKG_LICENSE_FILES`?

1715173329 (approved):

PR #22941, opened 2023-12-20
openvpn: start openvpn connection located under '/etc/openvpn' not only on system start

feckert:
Maintainer: @jow- @AuthorReflex 

Description:

OpenVPN configurations that have a uci entry, the enable/enabled option can be used to control whether the OpenVPN connection should be started at system startup or not.

OpenVPN configurations that are located under `/etc/openvpn/` are always started at system boot. To ensure that these connections can also be started later, they must **not** be started automatically during system boot. This can be prevented with the following new uci section entry in the OpenVPN configuration.

```
config globals 'globals'
    option autostart '0'
```

These OpenVPN configurations can then be started later with the command `/etc/init.d/openvpn start <name>`.

### DIFF
--- a/net/openvpn/Makefile
+++ b/net/openvpn/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=openvpn
 
 PKG_VERSION:=2.6.8
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


AuthorReflex:
Tested on my live environmet. Works fine.

feckert:
Thanks for testing :+1:

PR #22944, opened 2023-12-21
python-cython: update to version 3.0.7

predators46:
Maintainer:@jefferyto 
Compile tested: armsr
Run tested: armv8

Description:

https://github.com/cython/cython/releases

i need to update python-cython 3.0.7 version to add python-gevent


jefferyto:
I held off of updating because when version 3 was first released, not all packages were compatible.

Have you verified that all packages in the repo that use Cython are compatible with version 3? (Ideally all run-tested but compile-tested would be acceptable.)

predators46:
@jefferyto 

later I will try it. python-cython related packages

```
python/numpy/Makefile:PKG_BUILD_DEPENDS:=python-cython/host
python/python-dbus-fast/Makefile:PKG_BUILD_DEPENDS:=python-poetry-core/host python-cython/host python-wheel/host
python/python-fnv-hash-fast/Makefile:PKG_BUILD_DEPENDS:=python-cython/host python-poetry-core/host
python/python-fnv-hash-fast/Makefile:(32) if cython is available, and will fallback to pure python from the fnvhash
python/python-msgpack/Makefile:PKG_BUILD_DEPENDS:=python-cython/host
python/python-sqlalchemy/Makefile:PKG_BUILD_DEPENDS:=python-cython/host
python/python-yaml/Makefile:PKG_BUILD_DEPENDS:=python-cython/host
python/python-zeroconf/Makefile:PKG_BUILD_DEPENDS:=python-cython/host python-poetry-core/host
```

jefferyto:
> later I will try it. python-cython related packages

Also utils/borgbackup and utils/setools.

predators46:
hy @jefferyto 

python-numpy should be updated to version 1.26.2 but that requires a dependency package

predators46:
@neheb 

can you please update meson. looks like numpy-1.26.2 requires the newer meson

```
../../../../../build_dir/target-aarch64_generic_musl/pypi/numpy-1.26.2/meson.build:1:0: ERROR: Meson version is 1.2.1 but project requires >=1.2.99
```

PR #22948, opened 2023-12-21
python-msgpack: update version 1.0.7

predators46:
Maintainer:Jan Pavlinec <jan.pavlinec1@gmail.com>
Compile tested:armsr
Run tested:armv8

Description:

python-msgpack update version 1.0.7


predators46:
@jefferyto 

the cause of this error is because it requires version cython-3.0.7 Waiting for Cython to be updated to version Cyhton-3.0.7

PR #22950, opened 2023-12-21
python-zeroconf: update version 0.131.0

predators46:
Maintainer:Josef Schlehofer <pepe.schlehofer@gmail.com>
Compile tested: armsr
Run tested: armv8

Description:

python-zeroconf update version 0.131.0


predators46:
@jefferyto 

the cause of this error is because it requires version cython-3.0.7 Waiting for Cython to be updated to version Cyhton-3.0.7

PR #22958, opened 2023-12-22
python-cython: update to version 3.0.7

predators46:
Maintainer: @jefferyto 
Compile tested: armsr
Run tested: armv8

Description:


predators46:
@jefferyto 

I have tested building packages that require the host python-cython, everything was built successfully except python-msgpack and python-zeroconf which require cython-3.0.7

PR #22961, opened 2023-12-22
pyproject-metadata: add pyproject-metadata

predators46:
Maintainer: @predators46 
Compile tested: armsr
Run tested: armv8

Description:

pyproject-metadata: add pyproject-metadata

https://pypi.org/project/pyproject-metadata/

source code
https://github.com/FFY00/python-pyproject-metadata

add the pyproject-metadata package that numpy requires

PR #22962, opened 2023-12-22
numpy: update to version 1.26.2

predators46:
Maintainer: @commodo 
Compile tested: armsr
Run tested: armv8

Description:

numpy: update to version 1.26.2

https://github.com/numpy/numpy/releases

NumPy 1.26.2 Release Notes
NumPy 1.26.2 is a maintenance release that fixes bugs and regressions
discovered after the 1.26.1 release. The 1.26.release series is the last
planned minor release series before NumPy 2.0. The Python versions
supported by this release are 3.9-3.12.


predators46:
pyproject-metadata ==> adds the pyproject-metadata package that numpy requires
numpy ==> requires host cython-3.0.7

predators46:
pyproject-metadata ==> pyproject-metadata package that numpy requires

commodo:
hey

thanks for the PR;
i'm a bit slow due to holiday/vacation;

lang/python/numpy/Makefile
```diff
@@ -21,7 +21,7 @@ PKG_CPE_ID:=cpe:/a:numpy:numpy
 PKG_CONFIG_DEPENDS:= \
         CONFIG_NUMPY_OPENBLAS_SUPPORT
 
-PKG_BUILD_DEPENDS:=python-cython/host
+PKG_BUILD_DEPENDS:=python-cython/host pyproject-metadata/host
```
commodo, line 24: this will not work (yet)
there is no `pyproject-metadata/host` package;

you could however try to use `HOST_PYTHON3_PACKAGE_BUILD_DEPENDS`;
there are some examples in the `openwrt-21.02` branch (if you grep there);
the infrastructure is still in place to run that logic
but we'd need @jefferyto to provide an opinion on whether to use that (or not)

lang/python/numpy/patches/003-without-vendored-meson.patch
```diff
@@ -0,0 +1,22 @@
+--- a/pyproject.toml
++++ b/pyproject.toml
+@@ -1,6 +1,6 @@
+ [build-system]
+-build-backend = "mesonpy"
+-backend-path = ['./vendored-meson/meson-python']
```
commodo, line 6: any hint as to why we need to comment out this (and the code)?
i'm assuming that this helps us use the build system of openwrt somehow; though i'm somewhat vague here

lang/python/numpy/Makefile
predators46: hy @commodo 

actually I wanted to update chython to version 3.0.7 no plans to update this but at the time I thought numpy needed a chython update and it turns out it didn't. if you think this pull request is bad i would close it let the python package maintainer update this

lang/python/numpy/patches/003-without-vendored-meson.patch
predators46: maybe you should try building this package without this patch. it will explain to you what happens in the absence of this patch. I also saw this issue reported in numpy but I forgot to link about this issue.

lang/python/numpy/Makefile
predators46: @commodo 

https://github.com/openwrt/packages/pull/22961

neheb:
rebase

predators46:
@neheb 

rebase done

Issue #22973, opened 2023-12-23
wifischedule: wrong permissions for /etc/init.d/wifi_schedule


raenye:
Maintainer: @newkit
Environment: all

Description: init script added in e0d7181a6d397e181cc is missing the executable bit (+x).
(edit: in fact it is set by the install method in the `Makefile`)

```
root@OpenWrt:~# service
Usage: service <service> [command]
/etc/init.d/boot                   enabled         stopped
/etc/init.d/bootcount              enabled         stopped
...
/etc/init.d/urngd                  enabled         running
/etc/init.d/vnstat                 enabled         running
/sbin/service: line 21: /etc/init.d/wifi_schedule: Permission denied
/etc/init.d/wifi_schedule         disabled         stopped
/etc/init.d/wpad                   enabled         running
```

PR #22974, opened 2023-12-23
wifischedule: fix startup problems

raenye:
Maintainer: @newkit
Compile tested: mediatek/fiilogic Sinovoip Banana Pi R3, snapshot kernel 6.1.69
Run tested: mediatek/fiilogic Sinovoip Banana Pi R3, snapshot kernel 6.1.69

Description:
1. Make the new `startup()` function in `/usr/bin/wifi_schedule.sh`
    respect the global `enabled` config flag; in particular, make no
    changes to `/etc/config/wireless` when wifi_schedule is disabled.
2. Make the new `/etc/init.d/wifi_schedule` service script executable.

Fixes: e0d7181a6
Closes: #22973
Closes: #22988

### DIFF
--- a/net/wifischedule/net/usr/bin/wifi_schedule.sh
+++ b/net/wifischedule/net/usr/bin/wifi_schedule.sh
@@ -277,14 +277,16 @@ _should_wifi_enabled()
 startup()
 {
     _log "startup"
-    local _enable_wifi=$(_should_wifi_enabled)
-    if [[ ${_enable_wifi} -eq 0 ]]
-    then
-        _log "enable wifi"
-        enable_wifi
-    else 
-        _log "disable wifi"
-        disable_wifi
+    local global_enabled=$(_get_uci_value ${GLOBAL}.enabled) || _exit 1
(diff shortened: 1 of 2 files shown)


net/wifischedule/net/usr/bin/wifi_schedule.sh
```diff
@@ -277,14 +277,17 @@ _should_wifi_enabled()
...
-    then
-        _log "enable wifi"
-        enable_wifi
-    else 
-        _log "disable wifi"
-        disable_wifi
+    local global_enabled=$(_get_uci_value ${GLOBAL}.enabled) || _exit 1
+    if [ ${global_enabled} -eq 1 ]; then
+        local _enable_wifi=$(_should_wifi_enabled)
+        if [[ ${_enable_wifi} -eq 0 ]]
```
neheb, line 283: [[]] is a bashism. Might want to run through shellcheck in any case.

raenye: Thanks. It's in the original code, but I can get rid of it anyway.

PR #22967, opened 2023-12-22
sshtunnel: Add support for "ProxyCommand" [v5] 

stokito:
Maintainer: @nunojpg
Compile tested: master, VirtualBox 
Run tested: master, VirtualBox, TurrisOS. Both OpenSSH and Dropbear

Description:
The PR replaces #20163.

It turned out that instead of generation of a temp file we can use `procd_append_param command -o "ProxyCommand=$ProxyCommand"` that will make a trick with quoting.

The replacing of placeholders `%h` and `%p` is not needed because it's done by ssh client itself.

Additionally I made the code simpler and faster.

@lars18th please review

PR #22975, opened 2023-12-23
sshtunnel: Add support for "ProxyCommand" [v6]

stokito:
Maintainer: @nunojpg
Compile tested: master, VirtualBox
Run tested: master, VirtualBox, TurrisOS. Both OpenSSH and Dropbear

Description:
The PR replaces https://github.com/openwrt/packages/pull/22967.

It turned out that instead of generation of a temp file we can use procd_append_param command -o "ProxyCommand=$ProxyCommand" that will make a trick with quoting.

The replacing of placeholders %h and %p is not needed because it's done by ssh client itself.

Additionally I made the code simpler and faster.

I removed the line of @lars18th from config so now the PR doesn't have his code but only mine.

### DIFF
--- a/net/sshtunnel/Makefile
+++ b/net/sshtunnel/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=sshtunnel
 PKG_VERSION:=5
-PKG_RELEASE:=1
+PKG_RELEASE:=2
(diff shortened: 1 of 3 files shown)


nunojpg:
need to bump pkg release

stokito:
Added a commit with v5.2

nunojpg (approved):

stokito:
Thank you!

PR #22976, opened 2023-12-23
qemu: update to 8.2.0

vooon:
Maintainer: @yousong
Compile tested: x86_64, OpenWrt master (https://github.com/openwrt/openwrt/commit/713dc12960dd6146706a364905904b58d4d586d4)
Run tested: x86_64 host, KVM for qemu-x86_64-softmmu + Bios VM.

Description:
- Version updated to 8.2.0

### DIFF
--- a/utils/qemu/Makefile
+++ b/utils/qemu/Makefile
@@ -9,10 +9,10 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=qemu
-PKG_VERSION:=8.1.2
+PKG_VERSION:=8.2.0
 PKG_RELEASE:=1
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_HASH:=541526a764576eb494d2ff5ec46aeb253e62ea29035d1c23c0a8af4e6cd4f087
+PKG_HASH:=bf00d2fa12010df8b0ade93371def58e632cb32a6bfdc5f5a0ff8e6a1fb1bf32
 PKG_SOURCE_URL:=http://download.qemu.org/
 PKG_LICENSE:=GPL-2.0-only
 PKG_LICENSE_FILES:=LICENSE tcg/LICENSE
@@ -336,7 +336,6 @@ CONFIGURE_ARGS +=			\
 
 # accel
 CONFIGURE_ARGS +=			\
-	--disable-hax			\
 	--disable-hvf			\
 	--disable-whpx			\
 	--disable-xen			\
@@ -378,6 +377,7 @@ CONFIGURE_ARGS +=			\
 #
 CONFIGURE_ARGS +=			\
 	--audio-drv-list=''		\
+	--disable-af-xdp                \
 	--disable-attr			\
 	--disable-auth-pam		\
 	--disable-brlapi		\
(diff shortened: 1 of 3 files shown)


utils/qemu/Makefile
```diff
@@ -378,6 +377,7 @@ CONFIGURE_ARGS +=			\
 #
 CONFIGURE_ARGS +=			\
 	--audio-drv-list=''		\
+	--disable-af-xdp                \
```
BKPepe, line 380: This is not said in the commit.

utils/qemu/Makefile
```diff
@@ -401,6 +401,7 @@ CONFIGURE_ARGS +=			\
 	--disable-libpmem		\
 	--disable-libssh		\
 	--$(if $(CONFIG_QEMU_DEV_USB),enable,disable)-libusb		\
+	--disable-libkeyutils           \
```
BKPepe, line 404: dtto

utils/qemu/Makefile
```diff
@@ -417,6 +418,7 @@ CONFIGURE_ARGS +=			\
 	--disable-qom-cast-debug	\
 	--disable-rbd			\
 	--disable-rdma			\
+	--disable-rutabaga-gfx          \
```
BKPepe, line 421: dtto

utils/qemu/Makefile
vooon: By defaul most of the fatures if disable in openwrt, so i decided to to enable feature we didn't use previosly.

utils/qemu/Makefile
vooon: Same as for new nic.

utils/qemu/Makefile
vooon: And again, we haven't use new graphics by default. Also as i know that more useful for the cloud, than wrt like single node appilcations.

vooon: @BKPepe also do you know that't is the sate of Rust libararies like libblkio or that rustabaga ones in the WRT?

utils/qemu/Makefile
1715173329: Okay so please mention it in your commit message like "disabled new options introduced in this version"

vooon: @1715173329 done.

1715173329 (approved):

utils/qemu/Makefile
1715173329: OpenWrt has rust toolchain, but these packages aren't added yet. You can create new Makefile for them.

vooon: @1715173329 thanks for the info. I'm not going to use features that requires that libraries, so i'm not the best candidate to maintain those packages.

vooon:
Fiendly ping @BKPepe @yousong @neheb

utils/qemu/Makefile
```diff
@@ -9,10 +9,10 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=qemu
-PKG_VERSION:=8.1.2
+PKG_VERSION:=8.2.0
 PKG_RELEASE:=1
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_HASH:=541526a764576eb494d2ff5ec46aeb253e62ea29035d1c23c0a8af4e6cd4f087
+PKG_HASH:=bf00d2fa12010df8b0ade93371def58e632cb32a6bfdc5f5a0ff8e6a1fb1bf32
 PKG_SOURCE_URL:=http://download.qemu.org/
```
dcousens, line 16:
```suggestion
PKG_SOURCE_URL:=https://download.qemu.org/
```
You could probably update this at the same time, it is easily checked by opening https://download.qemu.org/ in your browser

PR #22989, opened 2023-12-25
wifischedule: respect global enable during startup

raenye:
(combined into #22974)

### DIFF
--- a/net/wifischedule/net/usr/bin/wifi_schedule.sh
+++ b/net/wifischedule/net/usr/bin/wifi_schedule.sh
@@ -277,14 +277,17 @@ _should_wifi_enabled()
 startup()
 {
     _log "startup"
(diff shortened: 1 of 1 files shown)


BKPepe:
You need to bump ``PKG_RELEASE`` in the Makefile.

raenye:
Thanks @BKPepe, I'm just not sure how to deal with two concurrent PRs to the same package.

Maybe I should submit a single PR for both changes? (fix the two related bugs together)

1715173329:
> Thanks @BKPepe, I'm just not sure how to deal with two concurrent PRs to the same package (the other is #22974).
> 
> Maybe I should submit a single PR for both changes? (fix the two related bugs together)

Yes, to the same package, it's okay.

raenye:
Done; please see #22974.

PR #22993, opened 2023-12-26
Update shairport-sync and add nqptp for AirPlay2

bklang:
This commit updates shairport-sync to version 4.3.2, which adds support for AirPlay2. As a dependency, it also introduces a new port for nqptp.

Maintainer: me / @bklang
Compile tested: bcm27xx-bcm2710 (RaspberryPi 3 B+), OpenWrt 23.05
Run tested: bcm27xx-bcm2710 (RaspberryPi 3 B+), OpenWrt 23.05. Installed package on RPi 3, connected to speakers, played audio. Confirmed new AirPlay2 functionality works (streaming to multiple devices from iPhone)


net/nqptp/Makefile
```diff
@@ -0,0 +1,58 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+# updated to work with latest source from abrasive
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=nqptp
```
BKPepe, line 9: This one should be in separate commit

net/nqptp/Makefile
```diff
@@ -0,0 +1,58 @@
...
+  nqptp is a daemon that monitors timing data from PTP clocks it sees on ports
+  319 and 320. It maintains records for one clock, identified by its Clock ID.
+
+  It is a companion application to Shairport Sync and provides timing
+  information for AirPlay 2 operation.
+endef
+
+define Build/Configure
+  $(call Build/Configure/Default)
+endef
```
BKPepe, line 44: Why are you calling default?

net/nqptp/Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+define Build/Install
+	$(INSTALL_DIR) $(PKG_INSTALL_DIR)/usr/bin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/nqptp $(PKG_INSTALL_DIR)/usr/bin/
+endef
```
BKPepe, line 49: We dont need it, right?

sound/shairport-sync/Makefile
```diff
@@ -7,15 +7,16 @@
...
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/mikebrady/shairport-sync/tar.gz/$(PKG_VERSION)?
-PKG_HASH:=17990cb2620551caa07a1c3b371889e55803071eaada04e958c356547a7e1795
+PKG_HASH:=dfb485c0603398032a00e51f84b874749bbf155b257adda3d270d5989de08bfd
 
 PKG_MAINTAINER:=Ted Hess <thess@kitschensync.net>, \
-		Mike Brady <mikebrady@eircom.net>
+		Mike Brady <mikebrady@eircom.net>, \
+		Ben Klang <bklang@wirehack.net>
```
BKPepe, line 19: Those maintainers approved it?

sound/shairport-sync/Makefile
```diff
@@ -29,7 +30,7 @@ define Package/shairport-sync/default
   SECTION:=sound
   CATEGORY:=Sound
   TITLE:=AirPlay compatible audio player
-  DEPENDS:=@AUDIO_SUPPORT +libpthread +alsa-lib +libconfig +libdaemon +libpopt
+  DEPENDS:=@AUDIO_SUPPORT +libpthread +alsa-lib +libconfig +libdaemon +libpopt +libstdcpp +libplist +libsodium +libgcrypt +libffmpeg-full +libuuid +nqptp
```
BKPepe, line 33: So many dependencies right now. Do you really need full variant of ffmpeg?

net/nqptp/Makefile
bklang: We do need it. The installation task that comes from upstream includes a call to `setcap` that is not desired on the build host. Incidentally, this has been removed in upstream nqptp, but is not yet in a released version.

sound/shairport-sync/Makefile
bklang: Unfortunately, yes. I tested with both `libffmpeg-audio-dec` and `libffmpeg-mini`, but neither of those provide the required `libswresample.so`

sound/shairport-sync/Makefile
bklang: I was not sure of the etiquette here since I am upgrading the package and providing a dependency. I'll remove myself.

net/nqptp/Makefile
bklang: This was my ignorance of the build system. I have removed it.

net/nqptp/Makefile
1715173329: `make install` won't be called if you don't set `PKG_INSTALL:=1`.
Simply moving this to `Package/nqptp/install` would be okay.

BKPepe: But the same binary is already in the ``define Package/nqptp/install``, so my question is still valid. We dont need this.

net/nqptp/Makefile
```diff
@@ -0,0 +1,54 @@
...
+PKG_FIXUP:=autoreconf
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/nqptp
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Not Quite Peer-to-Peer
+  PROVIDES:=nqptp
```
BKPepe, line 30: Why do we need provides? O_o

net/nqptp/Makefile
```diff
@@ -0,0 +1,54 @@
...
+  URL:=https://github.com/mikebrady/nqptp
+endef
```
BKPepe, line 32: This package has no dependencies, I find it weird.

net/nqptp/Makefile
```diff
@@ -0,0 +1,54 @@
...
+PKG_NAME:=nqptp
+PKG_VERSION:=1.2.4
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/mikebrady/nqptp/tar.gz/$(PKG_VERSION)?
+PKG_HASH:=1df1d5edd5b713010d6495b3abca4c1cf4ad8fa6029df0abeb9e4de8e0eb707a
+
+PKG_MAINTAINER:=Ben Klang <bklang@wirehack.net>
+PKG_LICENSE:=GPL-2.0
```
BKPepe, line 18: Deprecated SPDX License Identifier.

net/nqptp/Makefile
```diff
@@ -0,0 +1,54 @@
...
+PKG_LICENSE_FILES:=COPYING LICENSES nqptp.c
```
BKPepe, line 19: nqptp.c is license file?

sound/shairport-sync/Makefile
BKPepe: Okay, but this file is also provided by libffmpeg-custom. So, this makes it as hard dependency for full. 
What about other dependencies? Why they are required now?

net/nqptp/Makefile
bklang: Yes, nqptp.c contains the specific version of GPL (2) that is used by the package. The LICENSE file contains only the GPL license text without the version specification. I followed this convention from the shairport-sync port. Would you prefer something different?

net/nqptp/Makefile
bklang: It's a very small program. I checked again by looking at the source code and the build instructions, neither list any dependencies. To be safe, I installed this into a fresh OpenWrt SDK with no dependencies installed and it compiled cleanly.

ldd reports:

```
# ldd /usr/bin/nqptp
	/lib/ld-musl-aarch64.so.1 (0x7fa7752000)
	libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x7fa7721000)
	libc.so => /lib/ld-musl-aarch64.so.1 (0x7fa7752000)
```

net/nqptp/Makefile
bklang: Thank you both for the advice, you are right. I have removed this.

sound/shairport-sync/Makefile
bklang: The newly added dependencies are because of the new AirPlay2 support. I went back and checked each one to be sure.

I'm new to packaging with OpenWrt. Can you recommend the best way to add a dependency on either libffmpeg-custom or libffmpeg-full?

net/nqptp/Makefile
bklang: Because I'm  new to packaging for OpenWrt and I thought it was required :) Removed, thanks for the feedback

PR #23016, opened 2024-01-01
v4l2camera: Update Makefile to fix mips build

mpromonet:
Maintainer: @mpromonet 
Compile tested: allwinner H3, nanopi neo air, openwrt latest
Run tested: allwinner H3, nanopi neo air, openwrt latest

Description:
fix build for [mipsel_24kc_24kf/packages/v4l2camera/compile.txt](https://downloads.openwrt.org/snapshots/faillogs/mipsel_24kc_24kf/packages/v4l2camera/compile.txt)

### DIFF
--- a/multimedia/v4l2camera/Makefile
+++ b/multimedia/v4l2camera/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=v4l2camera
 PKG_VERSION:=0.1.8
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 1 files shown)


1715173329:
You should raise `PKG_RELEASE`.
If possible please also provide details in your commit message: what's the error, and how this fixes it.

BKPepe:
There is still missing what @1715173329 wanted.

- Whats the error
- How this fixes it.

Only what we know is that it fixes mips build without any details. 😢

mpromonet:
@BKPepe 

Sorry for that, I was expecting github action will be able to check this. It is based on fact that v4l2rtspserver built with same compilation option. I did not took the time to install mips cross compiler and checked, I only checked it is still work on nanopi air.

Best Regards
Michel

PR #23023, opened 2024-01-01
gst1-plugins-good: use libsoup3 instead of libsoup

flyn-org:
Maintainer: me
Compile tested: x86_64

Description:
This removes the last dependency in OpenWrt on libsoup, replacing it with libsoup3.


neheb:
does that work? I would think API changes are needed.

flyn-org:
It looks like gst1-plugins-good can detect either libsoup or libsoup3 and select the correct code accordingly.

PR #23030, opened 2024-01-01
nfdump: update to 1.7.3

flyn-org:
Maintainer: me
Compile tested: x86_64

### DIFF
--- a/net/nfdump/Makefile
+++ b/net/nfdump/Makefile
@@ -6,12 +6,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nfdump
-PKG_VERSION:=1.7.2
+PKG_VERSION:=1.7.3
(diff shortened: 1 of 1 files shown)


hnyman:
New version might cause new dependencies unless you disable that library usage:

https://downloads.openwrt.org/snapshots/faillogs/arm_fa526/packages/nfdump/compile.txt
```
Package nfdump is missing dependencies for the following libraries:
libzstd.so.1
```

flyn-org:
> New version might cause new dependencies unless you disable that library usage:
> 
> https://downloads.openwrt.org/snapshots/faillogs/arm_fa526/packages/nfdump/compile.txt
> 
> ```
> Package nfdump is missing dependencies for the following libraries:
> libzstd.so.1
> ```

See https://github.com/openwrt/packages/pull/23064.

PR #23040, opened 2024-01-02
Remove libdaq and snort packages

flyn-org:
Maintainer: me

I no longer maintain libdaq/snort, as I have moved to libdaq3/snort3.


BKPepe:
I was thinking, if there is package without maintainer and it is replaced by newer version I mean snort3... shouldnt we remove it? I know, it would be somehow strict, but we are talking here about master branch and it will take one year (approx) that these changes in the master branch would be in the stable branch. So... what I am saying is that it is OK to remove those packages, because there are still in the stable branches.

I know, snort is still supported, it is not end of life, but we should rather focus on something, which we are using.

flyn-org:
I force-pushed to remove snort and libdaq instead of simply dropping the maintainer. I will add both to packages-abandoned shortly.

PR #23041, opened 2024-01-02
libsoup: remove package

flyn-org:
Remove libsoup package, as all of the packages that have depended on it moved to libsoup3. The last package to require libsoup is gst1-plugins-good, and https://github.com/openwrt/packages/pull/23023 migrates that package to libsoup3. Merge this after pull #23023.

Maintainer: me


neheb:
if you could open a PR to packages-abandoned, that would be great.

PR #23044, opened 2024-01-02
xmlrpc-c: update to 1.59.02

yegorich:
Maintainer: @thess
Compile tested: ARMv7, TI Sitara AM335x, OpenWrt v23.05.2
Run tested: ARMv7, TI Sitara AM335x, OpenWrt v23.05.2, tested with internal software

Description: update xmlrpc-c to 1.59.02

While on this, remove the upstreamed patch and add a patch fixing pkg-config files (001-fix-pkg-config-files.patch). This patch is a combination of two upstream revisions [r3203](https://sourceforge.net/p/xmlrpc-c/code/3203/) and [r3205](https://sourceforge.net/p/xmlrpc-c/code/3205/).


libs/xmlrpc-c/patches/001-fix-pkg-config-files.patch
```diff
@@ -0,0 +1,327 @@
+Index: lib/abyss/src/Makefile
+===================================================================
+diff --git a/lib/abyss/src/Makefile b/lib/abyss/src/Makefile
+--- a/lib/abyss/src/Makefile	(revision 3202)
++++ b/lib/abyss/src/Makefile	(revision 3205)
```
BKPepe, line 5: - Who is the author?
- Where is the commit subject, commit description
- Is this going to be upstreamed?

And more questions, I do have...

yegorich: I have updated the PR description describing the origin of the patch. As described in this commit log, this patch is already upstream.

yegorich:
How do I create a clean patch from SVN? I have already used the `--git` option.

yegorich:
A refresh was required to pass the tests. Fixed now.

PR #23049, opened 2024-01-03
mstflint: add new package

tk154:
Maintainer: me
Compile tested: x86_64, OpenWrt Snapshot r24403+320-c23b509d72
Run tested: x86_64, Mellanox Spectrum SN2100, OpenWrt Snapshot r24403+320-c23b509d72

Description:
A tool for querying and installing firmware for Mellanox Spectrum Switches and ConnectX NICs.


utils/mstfwmanager/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+PKG_NAME:=mstfwmanager
+PKG_VERSION:=4.26.0
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/Mellanox/mstflint
+PKG_MIRROR_HASH:=9ab90387f295e974e3e6f6e8b0ce7e5f37749e0b1a6977c1b5c2ee6c2c2b80f6
+PKG_SOURCE_VERSION:=ff014c1e873e86255c07a396d0f3588ad018cc2a
+PKG_SOURCE_DATE:=2023-11-08
```
BKPepe, line 18: You can use their tagged releases instead of using specific commit hash

utils/mstfwmanager/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+PKG_MAINTAINER:=Til Kaiser <mail@tk154.de>
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=LICENSE
```
BKPepe, line 22: License or copying? 🤔

utils/mstfwmanager/Makefile
```diff
@@ -0,0 +1,56 @@
...
+    the system for available Mellanox devices (only mst
+    PCI devices) and performs the necessary firmware updates.
+endef
+
+CONFIGURE_ARGS += \
+	--enable-fw-mgr \
+	--disable-inband
+
+TARGET_CFLAGS += \
+	-D_GNU_SOURCE
```
BKPepe, line 49: This is still required?

utils/mstfwmanager/Makefile
```diff
@@ -0,0 +1,56 @@
...
+  TITLE:=Mellanox firmware update and query utility
+  DEPENDS:=@TARGET_x86 \
+    +libcurl +liblzma +libopenssl +libstdcpp +libxml2 +zlib
+endef
+
+define Package/mstfwmanager/description
+    Mellanox firmware update and query utility which scans 
+    the system for available Mellanox devices (only mst
+    PCI devices) and performs the necessary firmware updates.
+endef
```
BKPepe, line 42: two spaces here as indentation

BKPepe (changes requested):
- Patches without Git header are no go from my side.
They should be in upstreamable state, thus commit subject, description, etc. If possible send them to upstream.

utils/mstfwmanager/Makefile
tk154: Copying and apparently GPL 2.0 only.

utils/mstfwmanager/Makefile
tk154: I believe so. The package accesses u_int8_t, u_int16_t, and u_int32_t which are protected by _GNU_SOURCE in musl libc <sys/types.h>: https://git.musl-libc.org/cgit/musl/tree/include/sys/types.h#n59

tk154:
> * Patches without Git header are no go from my side.
>   They should be in upstreamable state, thus commit subject, description, etc. If possible send them to upstream.

I updated them with the appropriate headers.

utils/mstfwmanager/patches/0003-mtcr_ul-use-linux-instead-of-sys-pci.h.patch
```diff
@@ -0,0 +1,29 @@
+From 3b1883e4a9e190d729345f3a4c52ecd59c691986 Mon Sep 17 00:00:00 2001
+From: Til Kaiser <mail@tk154.de>
+Date: Wed, 3 Jan 2024 18:02:07 +0100
+Subject: [PATCH] mtcr_ul: use linux instead of sys pci.h
+
+The musl toolchain doesn't have <sys/pci.h> but <linux/pci.h>.
+<sys/pci.h> seems to be just a wrapper around <linux/pci.h> anyway.
+
```
BKPepe, line 8: This can be upstreamed, right?

utils/mstfwmanager/patches/0002-configure-remove-check-for-std-regex-compatibility.patch
```diff
@@ -0,0 +1,51 @@
+From 05dc7c55e0653542f92ed4e791f5eecb0e83b99c Mon Sep 17 00:00:00 2001
+From: Til Kaiser <mail@tk154.de>
+Date: Wed, 3 Jan 2024 17:58:41 +0100
+Subject: [PATCH] configure: remove check for std::regex compatibility
+
+The configure script compiles and runs a simple program to check
+for std::regex compatibility. Remove the check since this
+isn't possible when cross-compiling and returns an error message.
```
BKPepe, line 8: Report this to upstream or send this to upstream what they will say about this.

tk154:
I have just opened a PR requesting the changes upstream: Mellanox/mstflint#874

utils/mstfwmanager/patches/0002-configure-remove-check-for-std-regex-compatibility.patch
BKPepe: Ok, I see that you were trying to solve this with upstream, which is nice and appreciated.

Anyway, could you please use the patch ,which is included in upstream?
https://github.com/Mellanox/mstflint/pull/871/files

utils/mstfwmanager/Makefile
```diff
@@ -0,0 +1,57 @@
...
+
+PKG_NAME:=mstfwmanager
+PKG_VERSION:=4.26.0-1
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/Mellanox/mstflint
+PKG_MIRROR_HASH:=2592515184fe229690f2567ee42b5e2a0c847c14a2d8924175384655caf649e7
+PKG_SOURCE_VERSION:=v$(PKG_VERSION)
+PKG_SOURCE_DATE:=2023-11-08
```
BKPepe, line 18: Can not we use tagged releases? Is there something why do you need to use Git sources?

tk154:
Meanwhile, they have accepted the patches upstream into their `master-devel` branch.
Therefore, I am using the commit hash for now and will switch to a tagged release when they release the new version (4.27).

I have also decided to add all `mstflint` tools rather than just the `mstfwmanager`.

utils/mstflint/Makefile
```diff
@@ -0,0 +1,93 @@
...
+
+define Package/mstflint/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/* $(1)/usr/bin/
+
+	$(INSTALL_DIR) $(1)/usr/lib
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/mstflint $(1)/usr/lib/
+
+	$(INSTALL_DIR) $(1)/etc
+	$(CP) $(PKG_INSTALL_DIR)/etc/mstflint $(1)/etc/
```
BKPepe, line 90: Shouldnt be some configuration file in this folder marked as configuration? (conffiles :eyes:)

utils/mstflint/Makefile
```diff
@@ -0,0 +1,93 @@
...
+	--enable-fw-mgr \
+	--disable-inband
+
+TARGET_CFLAGS += \
+	-D_GNU_SOURCE \
+	-I$(STAGING_DIR)/usr/include/libxml2
+
+define Package/mstflint/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/* $(1)/usr/bin/
```
BKPepe, line 84: This is approach like copy everything to /usr/bin. Wouldnt it be better to copy only binaries, which we need? I mean specify them?

utils/mstflint/Makefile
```diff
@@ -0,0 +1,93 @@
...
+PKG_INSTALL:=1
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/mstflint
+  SECTION:=Utilities
+  CATEGORY:=Utilities
+  TITLE:=Mellanox Firmware Burning and Diagnostics Tools
+  URL:=https://github.com/Mellanox/mstflint
+  DEPENDS:=@(aarch64||x86_64) \
```
BKPepe, line 35: I havent look at it closer, but only these two targets are supported?

utils/mstflint/Makefile
tk154: I could do that, but there are 16 executables (11 binaries, 5 scripts).

utils/mstflint/Makefile
tk154: I misinterpreted the configure script and thought only those two were supported. After testing, I discovered that only mips(el) is unsupported.

utils/mstflint/Makefile
```diff
@@ -0,0 +1,101 @@
...
+
+PKG_NAME:=mstflint
+PKG_VERSION:=4.26.0-1
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/Mellanox/mstflint
+PKG_MIRROR_HASH:=7d0168add5857f54e2f96edb6cc78ff309fad78c404e262f145b89c63f6db39c
+PKG_SOURCE_VERSION:=17e9b849ba57cfac1b28240a25ece300a0f74a45
+PKG_SOURCE_DATE:=2024-01-15
```
BKPepe, line 18: We dont need to use the latest commit and we could stick to the stable/tagged releases, right?

utils/mstflint/Makefile
```diff
@@ -0,0 +1,101 @@
...
+# Copyright (C) 2024 Til Kaiser
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=mstflint
+PKG_VERSION:=4.26.0-1
```
BKPepe, line 11: Since you are using the latest commit from the repository, this is no required.

aparcar:
Please rebase the branch so the CI no longer fails :)

BKPepe:
Ok, there is one thing, which needs to be discussed, otherwise this is ready to go. :) Thanks for your patience.

utils/mstflint/Makefile
```diff
@@ -0,0 +1,100 @@
...
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/* $(1)/usr/bin/
+
+	$(INSTALL_DIR) $(1)/usr/lib/mstflint
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/mstflint $(1)/usr/lib/
+
+	$(INSTALL_DIR) $(1)/usr/share/mstflint
+	$(CP) $(PKG_INSTALL_DIR)/usr/share/mstflint $(1)/usr/share/
+
+	$(INSTALL_DIR) $(1)/etc/mstflint
+	$(INSTALL_CONF) $(PKG_INSTALL_DIR)/etc/mstflint/ca-bundle.crt $(1)/etc/mstflint/
```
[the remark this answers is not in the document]
tk154: You are right; I have removed the two lines.

tk154:
There is a CI build error for the `powerpc` targets coming from the `cryptodev` package, which isn't a dependency of the `mstflint` package.

Yet, for me, the compilation of the `cryptodev` package also runs fine for those `powerpc` targets on my local machine.

PR #23053, opened 2024-01-03
openssh: bump to 9.6p1

rucketeg:
Release notes: https://www.openssh.com/txt/release-9.6

Maintainer:  @SibrenVasse, @tripolar 
Compile tested: bcm2711, rpi4, master
Run tested: bcm2711, rpi4, master

### DIFF
--- a/net/openssh/Makefile
+++ b/net/openssh/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openssh
-PKG_VERSION:=9.5p1
+PKG_VERSION:=9.6p1
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://cdn.openbsd.org/pub/OpenBSD/OpenSSH/portable/ \
 		https://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/
-PKG_HASH:=f026e7b79ba7fb540f75182af96dc8a8f1db395f922bbc9f6ca603672686086b
+PKG_HASH:=910211c07255a8c5ad654391b40ee59800710dd8119dd5362de09385aa7a777c
 
 PKG_LICENSE:=BSD ISC
 PKG_LICENSE_FILES:=LICENCE
--- a/net/openssh/Makefile
+++ b/net/openssh/Makefile
@@ -20,6 +20,8 @@ PKG_LICENSE:=BSD ISC
 PKG_LICENSE_FILES:=LICENCE
 PKG_CPE_ID:=cpe:/a:openssh:openssh
 
+#While bumping new version, make sure that it works without it, so it can be removed.
+PKG_FIXUP:=autoreconf
 PKG_REMOVE_FILES:=
 PKG_CONFIG_DEPENDS := \
 	CONFIG_OPENSSH_LIBFIDO2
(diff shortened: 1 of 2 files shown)


BKPepe:
Why this is draft?

rucketeg:
Correct, it's still in draft because of the failure mentioned in https://github.com/openwrt/packages/pull/23053#issuecomment-1877316274. This error is probably caused by https://github.com/openssh/openssh-portable/commit/59d691b886c79e70b1d1c4ab744e81fd176222fd, that was included in the latest openssh release. I want to test if adding https://github.com/openssh/openssh-portable/commit/1036d77b34a5fa15e56f516b81b9928006848cbd as a patch would help.

SibrenVasse:
Your build is failing because it detected a dirty patch. I've refreshed the patch here: https://github.com/SibrenVasse/packages/pull/26/commits/03aeaaeb8d57c242f066d1b89d19734044dddb10

Currently running build:
https://github.com/SibrenVasse/packages/actions/runs/7413900137

rucketeg:
I wonder if there is any way to remind us in the future about reverting b3afcf24c267892911d425dd05ae025632933a9a once it's included in the next release.

SibrenVasse:
When updating to a new version, the patch won't apply if the lines are already present in the new source. So it won't build without removing the patch.

rucketeg:
I was mostly worried about the `PKG_FIXUP:=autoreconf` part.

SibrenVasse:
I didn't think you would cherry pick my commit, only grab the patch. As it was just a test I didn't write a commit message worthy of the master branch.  Please use this new one: https://github.com/SibrenVasse/packages/commit/1a54ff4707038c5d7ee15e01f86191015ce76777

net/openssh/patches/010-better_fzero-call-detection.patch
```diff
@@ -0,0 +1,30 @@
+--- a/m4/openssh.m4
++++ b/m4/openssh.m4
```
BKPepe, line 2: We can not accept patches without Git header. We dont know if it is backport or not.

Could you please use it like this:
https://github.com/openssh/openssh-portable/commit/1036d77b34a5fa15e56f516b81b9928006848cbd.patch

and then refresh it, please?

net/openssh/Makefile
```diff
@@ -8,18 +8,19 @@
...
 PKG_SOURCE_URL:=https://cdn.openbsd.org/pub/OpenBSD/OpenSSH/portable/ \
 		https://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/
-PKG_HASH:=f026e7b79ba7fb540f75182af96dc8a8f1db395f922bbc9f6ca603672686086b
+PKG_HASH:=910211c07255a8c5ad654391b40ee59800710dd8119dd5362de09385aa7a777c
 
 PKG_LICENSE:=BSD ISC
 PKG_LICENSE_FILES:=LICENCE
 PKG_CPE_ID:=cpe:/a:openssh:openssh
 
+PKG_FIXUP:=autoreconf
```
BKPepe, line 23: What about making comment, here? Will it work? :)

#While bumping new version, make sure that it works without it, so it can be removed.

SibrenVasse:
https://github.com/SibrenVasse/packages/commit/d09ae0e1a3e6d558a19c9aa45821e21779138b2a for inclusion

net/openssh/Makefile
BKPepe: You should mention that in the commit description as well.

PR #22965, opened 2023-12-22
python-dbus-fast: update version 2.21.0

predators46:
Maintainer: @QuintinHill 
Compile tested: armsr
Run tested: armv8

Description:

PR #23054, opened 2024-01-03
[23.05] Add missing python3-bleak dependency

QuintinHill:
Maintainer: me
Compile tested: None for backport
Run tested: None for backport

Description: https://github.com/openwrt/packages/pull/22919 and https://github.com/openwrt/packages/pull/22965 backports to 23.05.

CC: @predators46


predators46:
there should be no problem because this package does not require the latest version of cython.

Issue #21364, opened 2023-06-14
mwan3: route deletion occurs if only one of multiple routes with an source address selector is deleted


l-jonas:
Maintainer: @feckert @aaronjg 
Environment: OpenWrt 22.03

Description: Already reported at <https://forum.openwrt.org/t/mwan3-feedback-on-ipv6-and-missing-features/29483/41>, but without any response yet:

I'm having issues with default routes that get lost.

mwan3 is configured to remove the source address selector when copying the routes from the main table to the interface specific ones. This works as it intended.

When routes from the main table are removed, the same happens: The source address selectors is removed and this adjusted route is removed from the interface specific tables. This works, too.

The problem occurs when the main routing table contains multiple different routes that are equal after removing the source address selector. Then, removing one of them from the main table deletes the derived routes from all of them from the interface specific tables.

If this happens with the default route, then the interface specific tables contain no default route and all internet traffic that should use this interface is not forwarded anymore.

And yes, this happens with the default route for me. When my IPv6 address changes, new routes are added and some old default routes are removed some time later.

I'm using the following workaround now: modify /usr/sbin/mwan3rtmon; between route_line=... and handle_route, add the following:

```
if [ "$action" = "del" ]; then
	if mwan3_get_routes | grep -qxF "$route_line"; then
		LOG debug "deleted but route still exists - $route_line"
		return
	fi
fi
```

This does not correctly update the list of connected networks, but it's only a workaround.

However, I think that this is very inefficient with calling mwan3_get_routes. I think that it would be simpler to use ip monitor just as a trigger to regenerate the interface specific tables - or use it as a trigger for a complete regeneration in case of a removed route that has a source address selector.

feckert:
> The problem occurs when the main routing table contains multiple different routes that are equal after removing the source address selector. Then, removing one of them from the main table deletes the derived routes from all of them from the interface specific tables.

What do you mean with `multiple different`. Can you give me an `ip route show` dump ?

l-jonas:
The relevant beginning of the output from ``ip -6 route show``. Note that the first two groups of the source addresses and the whole gateway addresses (after via) were replaced for privacy reasons, but there relations were kept.

```
default from 2000:1:a5ab:6400::/64 via fe80::1 dev eth0.4 proto static metric 512 pref medium
default from 2000:1:a5ab:64f0::/60 via fe80::1 dev eth0.4 proto static metric 512 pref medium
[...]

PR #23058, opened 2024-01-04
mwan3: Check removed route before removal

l-jonas:
Maintainer: @feckert 
Compile tested: -
Run tested: OpenWrt 23.05.2 x86

Description: This makes mwan3rtmon check if mwan3_get_routes returns a route before removing it. This helps with IPv6 routes with source address selector removal where multiple original routes are transformed to the same mwan3 route if one of the source routes is removed while the others are kept.

This fixes <https://github.com/openwrt/packages/issues/21364>


neheb:
rebase.

feckert:
Sorry for the late reply and thanks for the fix.
I have pushed your change directly via commit https://github.com/openwrt/packages/commit/e119a57e97e894f0636df999d82807ae5c494030. I have also update the PKG_VERSION with your commit

PR #23066, opened 2024-01-05
freeradius3: add totp unixodbc eap-sim cache cache-rbtree

predators46:
Maintainer:

Compile tested: armsr
Run tested: armv8

Description:

add totp unixodbc eap-sim cache cache-rbtree

freeradius3-mod-totp
Implement the TOTP algorithm as given in RFC 6238.

freeradius3-mod-sql-unixodbc
SQL driver for the unixodbc ODBC connector library.

freeradius3-mod-eap-sim
Implements [RFC 4186](https://tools.ietf.org/html/rfc4186) EAP-SIM authentication. EAP-SIM provides authentication and session keying material for 802.11i (WPA/2 Enterprise) using SIM triplets.

freeradius3-mod-cache
Stores attributes and/or lists and adds them back to a subsequent request, or to the current request on a later execution of the module.

freeradius3-mod-cache-rbtree
Stores cache entries in an internal rbtree. It is a submodule of rlm_cache and cannot be used on its own.

### DIFF
--- a/net/freeradius3/Makefile
+++ b/net/freeradius3/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freeradius3
 PKG_VERSION:=3_0_26
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 PKG_SOURCE:=release_$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/FreeRADIUS/freeradius-server/tar.gz/release_$(PKG_VERSION)?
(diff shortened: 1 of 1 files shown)


BKPepe:
1. If possible, I would really appreciate if we can specify those modules in the subject, if it fits, of course.

2. Commit description:
drop add modules, thats duplicate from the subject.

net/freeradius3/Makefile
```diff
@@ -672,11 +712,8 @@ CONFIGURE_ARGS+= \
 	--with-radacctdir=/var/db/radacct \
 	--with-logdir=/var/log \
 	--without-pcre \
-	--without-rlm_cache \
 	--without-rlm_cache_memcached \
-	--without-rlm_couchbase \
```
BKPepe, line 677: This is used by which module?

predators46:
@BKPepe 
can you see the changes I made

net/freeradius3/Makefile
predators46: openwrt doesn't have a couchbase package so I added it back ```--without-rlm_couchbase```

BKPepe:
Why do I get every day notifications about your force push? It's pretty annoying, because what you are doing is just rebasing your branch on the top of the master branch without introducing new changes.  That's why it is sitting here for some time and I think it will be. :-)

predators46:
Good. you don't get any more notifications 😃

PR #23068, opened 2024-01-06
kmod: add host build

jetm:
Maintainer: @jdub 
Compile tested: (armsr, armv8, OpenWrt ce4ee14a46)
Run tested: (armsr, armv8, OpenWrt ce4ee14a46, tests done in Qemu and RockPi4)

Description:

Target armsr requires the depmod host binary to generate module dependency data for several Linux kernel modules that are installed in the initrd image.

It installs two files: kmod and depmod, the latter as a symlink to kmod. There is no need to build a host package.

It's a dependency for https://github.com/openwrt/openwrt/pull/13684

### DIFF
--- a/utils/kmod/Makefile
+++ b/utils/kmod/Makefile
@@ -47,6 +47,25 @@ define Host/Clean
 	rm -f $(STAGING_DIR_HOST)/bin/modinfo
 endef
 
+HOST_CONFIGURE_VARS += ./configure --with-zlib
+
+define Host/Compile
+	$(MAKE) -C $(HOST_BUILD_DIR) tools/kmod
+endef
+
+define Host/Install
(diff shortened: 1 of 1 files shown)


ynezz:
> Target armsr requires the depmod host binary to generate module dependency data for several Linux kernel modules that are installed in the initrd image.

Changes you're proposing requires depmod on the host.

> It's a dependency for https://github.com/openwrt/openwrt/pull/13684

Specifically for "Don't use OpenWrt kmod-* packages infrastructure." part, which you've been told, that its a wrong approach and likely breaks several use cases.

BTW this is a wrong place as well, if anything is needed on the host for the build, then it needs to be imported into the [OpenWrt tree](https://github.com/openwrt/openwrt/pulls), like for example commit [c39b0646f3f2](https://github.com/openwrt/openwrt/commit/c39b0646f3f2) ("pcre2: import pcre2 from packages feed"). Its not going to build/work, so I'm closing this.

PR #23071, opened 2024-01-06
python-yaml: test cython3

predators46:
https://github.com/yaml/pyyaml/issues/736

https://github.com/yaml/pyyaml/issues/724

Maintainer: 
Compile tested: 
Run tested: 

Description:


BKPepe:
If you are going to look at referenced issue, I would say that you are some kind of spammer. :) Each force push of your commit, references it. Thats why I do those references only in pull requests. 

There is also no need to open another PR, you could use the closed one. If you want to have it compile tested, there is possibility you can do that in your forked repository and you dont need to open pull requests here.

lang/python/python-yaml/Makefile
```diff
@@ -38,7 +38,11 @@ define Package/python3-yaml/description
   PyYAML is a YAML parser and emitter for the Python programming language.
 endef
 
-PYTHON3_PKG_BUILD_VARS:=PYYAML_FORCE_LIBYAML=1
+PYTHON3_PKG_BUILD_VARS:= \
+	PYYAML_FORCE_LIBYAML=1
```
BKPepe, line 42: This could be left as it is, right? No need to touch it.

predators46:
I don't want to argue I just want to wait to see if this works to build the yaml.

lang/python/python-yaml/Makefile
predators46: according to your wishes, I have returned it according to your wishes. actually I don't want to fix this but I have a sense of responsibility because the cython update caused this problem. Thank you also for observing the closed pull request. maybe this is my last pull request thanks

BKPepe: Since I got notification and see that you opened 2 PRs for the same thing, then you could expect the response.

predators46: if this works I will change the message of the changes I made

1715173329:
I think it's done by #23481.

PR #23072, opened 2024-01-06
yggdrasil-jumper: add new package

one-d-wide:
#### Description:

Yggdrasil Jumper is independent project that aims to transparently reduce latency of a connection over Yggdrasil network, utilizing NAT traversal to bypass intermediary nodes. It periodically probes for active sessions and automatically establishes direct peerings over internet with remote nodes running Yggdrasil Jumper without requiring firewall or port configuration.

#### Changes:

- Add `yggdrasil-jumper` package.
- Add logic to start/stop `yggdrasil-jumper` within `netifd` interface script for `yggdrasil`.

Maintainer: me
Compile tested: x86_64 master
Run tested: x86_64 master


wfleurant:
awesome ⚙️ work @one-d-wide

stangri:
@one-d-wide -- great work. What's the reason for splitting the Makefiles for yggdrasil and yggdrasil-jumper instead of using the same Makefile to build both?

one-d-wide:
> @one-d-wide -- great work. What's the reason for splitting the Makefiles for yggdrasil and yggdrasil-jumper instead of using the same Makefile to build both?

Yggdrasil Jumper is optional and logically separate part. And mixing different build processes in one Makefile probably wouldn't benefit readability anyway.

net/yggdrasil-jumper/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+PKG_NAME:=yggdrasil-jumper
+PKG_VERSION:=0.3.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/one-d-wide/yggdrasil-jumper/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=d0fd59e9f7a110094c9189b49b812fb0014c6c45535b4d30effd7cf602961454
+
+PKG_MAINTAINER:=one-d-wide <one-d-wide@protonmail.com>
```
BKPepe, line 11: You need to have here real name, otherwise it wont be accepted.

net/yggdrasil-jumper/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+
+define Package/yggdrasil-jumper
+	SECTION:=net
+	CATEGORY:=Network
+	SUBMENU:=Routing and Redirection
+	TITLE:=Yggdrasil peer-to-peer firewall stun
+	URL:=https://github.com/one-d-wide/yggdrasil-jumper
+	DEPENDS:=$(RUST_ARCH_DEPENDS) @!arc @IPV6 +kmod-tun
+endef
```
BKPepe, line 28: Two spaces as indentation

net/yggdrasil-jumper/Makefile
```diff
@@ -0,0 +1,48 @@
...
+PKG_MAINTAINER:=one-d-wide <one-d-wide@protonmail.com>
+PKG_LICENSE:=LGPL-3.0-only
+
+PKG_BUILD_DEPENDS:=rust/host
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/rust/rust-package.mk
+
+
```
BKPepe, line 20: Only one blank row is enough

net/yggdrasil-jumper/Makefile
```diff
@@ -0,0 +1,48 @@
...
+  of routers found in active sessions to leverage NAT traversal for constructing direct internet
+  bridges.
+endef
+
+
+define Package/yggdrasil-jumper/install
+	$(INSTALL_DIR) $(1)/usr/sbin
+
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/bin/yggdrasil-jumper $(1)/usr/sbin
+endef
```
BKPepe, line 44:
```suggestion
define Package/yggdrasil-jumper/install
	$(INSTALL_DIR) $(1)/usr/sbin
	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/bin/yggdrasil-jumper $(1)/usr/sbin
endef
```

net/yggdrasil-jumper/Makefile
```diff
@@ -0,0 +1,48 @@
...
+PKG_NAME:=yggdrasil-jumper
+PKG_VERSION:=0.3.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/one-d-wide/yggdrasil-jumper/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=d0fd59e9f7a110094c9189b49b812fb0014c6c45535b4d30effd7cf602961454
+
+PKG_MAINTAINER:=one-d-wide <one-d-wide@protonmail.com>
+PKG_LICENSE:=LGPL-3.0-only
```
BKPepe, line 12:
```suggestion
PKG_LICENSE:=LGPL-3.0-only
PKG_LICENSE_FILES:=LICENSE
```

BKPepe (changes requested):

one-d-wide:
@BKPepe

Thank you. It should be fixed now.

net/yggdrasil-jumper/Makefile
systemcrash: Do we have a field for `PKG_MAINTAINER_TAG`? Finding maintainers via their github tag (username) is a drag sometimes. Or perhaps a way of squeezing it into the same field above? Sorry to hijack...

BKPepe:
Nope, I did not finish it. SoB also requires real name, etc

Dne po 8. 1. 2024 14:46 uživatel Paul Donald ***@***.***>
napsal:

one-d-wide:
> Nope, I did not finish it. SoB also requires real name, etc

I replaced it (both in the Makefile and commit details) so CI formalities check should pass.

net/yggdrasil-jumper/Makefile
one-d-wide: There isn't seems to be a special variable used for maintainer's github tag anywhere in this repository.
I could add link to by github page as a comment next to `PKG_MAINTAINER`, but it seems not to be a common practice.

stangri:
> > @one-d-wide -- great work. What's the reason for splitting the Makefiles for yggdrasil and yggdrasil-jumper instead of using the same Makefile to build both?
> 
> Yggdrasil Jumper is optional and logically separate part. And mixing different build processes in one Makefile probably wouldn't benefit readability anyway.

You're changing the yggdrasil proper code to accommodate yggdrasil-jumper, so they are evidently connected.

My suggestion would be to merge the makefiles, like other packages do, for example acme: https://github.com/openwrt/packages/blob/master/net/acme-acmesh/Makefile, nebula: https://github.com/openwrt/packages/blob/master/net/nebula/Makefile, etc.

net/yggdrasil-jumper/Makefile
BKPepe: Hey @systemcrash , why do you need to have ``PKG_MAINTAINER_TAG``? Its pointless, because you or someone else should be more strict while merging commits. 

1. The tag applies only for GitHub, right. What about other mirrors? Don't forget that OpenWrt uses primary their own Git instance.

2. Username can be changed anytime.

3. Usually, if the committer email is not linked with the GitHub, I asked them politely to do so, because then the real name is linked to the GitHub account and you can mention that anytime and you can find it.

4. While merging commits, which are done by PR, there is always history (in Git or even here at GitHub) and referenced PR, which can be found at anytime. You can reach contributor via email and now we can have discussion what we will do if the email is no longer used, haha! 😆 

No need to have variables for everything, when they should be rules. :) Dont forget about that.

net/yggdrasil-jumper/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+PKG_NAME:=yggdrasil-jumper
+PKG_VERSION:=0.3.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/one-d-wide/yggdrasil-jumper/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=d0fd59e9f7a110094c9189b49b812fb0014c6c45535b4d30effd7cf602961454
+
+PKG_MAINTAINER:=one d wide <one-d-wide@protonmail.com>
```
BKPepe, line 11: I am still not sure that this is real name, right?
1. Same applies here and as well in SoB.

net/yggdrasil-jumper/Makefile
```diff
@@ -0,0 +1,45 @@
...
+  Yggdrasil Jumper is independent project that aims to transparently reduce latency of
+  a connection over Yggdrasil network, utilizing NAT traversal to bypass intermediary
+  nodes. It periodically probes for active sessions and automatically establishes direct
+  peerings over internet with remote nodes running Yggdrasil Jumper without requiring
+  firewall or port configuration.
+endef
+
+define Package/yggdrasil-jumper/install
+	$(INSTALL_DIR) $(1)/usr/sbin
+
```
BKPepe, line 40:
```suggestion
```

net/yggdrasil-jumper/Makefile
one-d-wide: I'd like to advocate that "real name" (as of linux contributing guidelines, which was adopted by openwrt) has been intended only to make possible tracking change authors and, despite poor wording, had never been actually used to discriminate pseudonyms ([source], see commit description):

> Long long ago, in a more innocent time, Greg wrote the clarification for
> how the DCO should work and that you couldn't make anonymous
> contributions, because the sign-off needed to be something we could
> check back with.
> 
> It was 2006, and nobody reacted to the wording, the whole Facebook 'real
> name' controversy was a decade in the future, and nobody even thought
> about it.  And despite the language, we've always accepted nicknames and
> that language was never meant to be any kind of exclusionary wording.
> 
> In fact, even when it became a discussion in other adjacent projects,
> apparently nobody even thought to just clarify the language in the
> kernel docs, and instead we had projects like the CNCF that had long
> discussions about it, and wrote their own clarifications [1] of it.

The wording of this rule has been recently clarified to avoid causing unnecessary angst and pain ([source], see changes of the same commit):

```diff 
-using your real name (sorry, no pseudonyms or anonymous contributions.)
+using a known identity (sorry, no anonymous contributions.)
```

Also quote from [CNCF guideline] on Developer Certificate of Origin (DCO, which is the concept underlying SoB):

> ## DCO and Real Names
> 
> The DCO requires the use of a real name that can be used to identify someone in case there is an issue about a contribution they made. 
> 
> **A real name does not require a legal name, nor a birth name, nor any name that appears on an official ID (e.g. a passport). Your real name is the name you convey to people in the community for them to use to identify you as you. The key concern is that your identification is sufficient enough to contact you if an issue were to arise in the future about your contribution.**
> 
> Your real name should not be an anonymous id or false name that misrepresents who you are.

[source]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/Documentation/process/submitting-patches.rst?id=d4563201f33a022fc0353033d9dfeb1606a88330
[CNCF guideline]: https://github.com/cncf/foundation/blob/659fd32c86dc/dco-guidelines.md

wfleurant:
> You're changing the yggdrasil proper code to accommodate yggdrasil-jumper, so they are evidently connected.

hi @stangri -- the bits of netifd related shell code used by yggdrasil-jumper may be the only option, no? um.. 

i'm only finding one `PKG_SOURCE_URL` per Makefile. its late, may be i am missing something... sorry im missing a solution, ttyl :zzz:

net/yggdrasil-jumper/Makefile
BKPepe: I think, I will be strict on this one. I dont want to argue with you, everybody is not afraid of his known identity or real name. We follow what is done by OpenWrt core developers. If you want to argue about it, then feel free to raise issue in their mailing list. So, since you try to reinvent a wheel here, then how to get it merged:

1. Use your real name

Anything else won't be accepted until you discussed it elsewhere not here. Rules and policy are set. They exists for a while.

one-d-wide: Hi. Thank for your advice. I mentioned the issue on mailing list and, thanks to Paul Spooren, [Submission Guidelines] identity rule on openwrt wiki is now in accordance with linux's one.

[Submission Guidelines]: https://openwrt.org/submitting-patches#submission_guidelines

And also appropriate CI patch should hopefully land soon #23084 :sparkles:

net/yggdrasil-jumper/Makefile
systemcrash: Heh. OK, with #23084 merged, we can close this discussion.

net/yggdrasil-jumper/Makefile
systemcrash: Can you gents resolve this conversation now that #23084 is merged?

wfleurant:
@one-d-wide do we fastforward to recheck with ci changes: https://github.com/openwrt/packages/pull/23084

wfleurant:
@BKPepe on your review, please. identity / signature OK in CI https://github.com/openwrt/packages/pull/23084

systemcrash:
Is this ready?

systemcrash:
@BKPepe Please update your review

systemcrash:
ping @ynezz @981213

systemcrash:
@BKPepe Please update your review

ping @ynezz @981213

ynezz:
@systemcrash FYI https://github.com/openwrt/actions-shared-workflows/commit/12d9551f2d07ec34ac813da8612c8014fb393af6

BKPepe:
Interesting. Should we revert the recent CI changes here as well and again
require real name? I have not any idea that it was reverted, there. :-(

Dne po 25. 3. 2024 16:52 uživatel Petr Štetiar ***@***.***>
napsal:

1715173329:
+1 and this time even no PR was made before committing.

systemcrash:
The weasel word there is "should". No discussion or vote raised.

ynezz:
> and this time even no PR was made before committing.

@1715173329 can you expand a bit on that? There was a PR https://github.com/openwrt/openwrt/pull/14380 and there was a public discussion, two times already, see the last thread https://lists.openwrt.org/pipermail/openwrt-devel/2024-January/042063.html leading to this series of pull requests.

systemcrash:
> and there was a public discussion ([1/2024](https://lists.openwrt.org/pipermail/openwrt-devel/2024-January/042058.html))

Then evidently the following is erroneous and it should be reverted (revert the revert):

> @systemcrash FYI [openwrt/actions-shared-workflows@12d9551](https://github.com/openwrt/actions-shared-workflows/commit/12d9551f2d07ec34ac813da8612c8014fb393af6)

systemcrash:
And *this* PR can be merged

BKPepe:
> evidently the following is erroneou

I think, @nbd168 wants to do vote, which is done on openwrt-adm, when they are voting about adding new people to the core team.

systemcrash:
> I think, @nbd168 wants to do vote, which is done on openwrt-adm, when they are voting about adding new people to the core team.

I see no desire. Writing "should" two months ago is inconclusive.

systemcrash:
So @one-d-wide I propose that while everyone is *really busy* trying to get a vote off the ground, in the meantime you adopt some official looking pseudonym, like One D Wide, or John D Wade and force push. Done.

systemcrash (approved):

1715173329 (approved):

1715173329:
oh please also increase `PKG_RELEASE` for yggdrasil.

one-d-wide:
> oh please also increase `PKG_RELEASE` for yggdrasil.

Oops. Done.

net/yggdrasil/files/yggdrasil.sh
```diff
@@ -97,6 +106,52 @@ proto_yggdrasil_generate_keypair() {
...
+	# Collect already defined protocols
+	protocols=""
+	_add_address_protocol() {
+		protocols="${protocols}$(echo $1 | cut -d "://" -f1) "
+	}
+	config_list_foreach "$config" listen_address _add_address_protocol
+
+	# Add new address for each previously unspecified protocol
+	for protocol in "tls" "quic"; do
+		if ! echo "$protocols" | grep "$protocol" &>/dev/null; then
```
systemcrash, line 121: Do not recommend this redirection. 

Advise to: 
```sh
		if ! echo "$protocols" | grep "$protocol" >/dev/null 2>&1 ; then
```

net/yggdrasil/files/yggdrasil.sh
```diff
@@ -97,6 +106,52 @@ proto_yggdrasil_generate_keypair() {
...
+			# By default linux dynamically alocates ports in the range 32768..60999
+			# `sysctl net.ipv4.ip_local_port_range`
+			random_port=$(( ($RANDOM + $RANDOM) % 22767 + 10000 ))
```
systemcrash, line 124:
```shell
random_port=$(( (RANDOM + RANDOM) % 22767 + 10000 ))
```

1715173329:
Merged, thank you!

net/yggdrasil/files/yggdrasil.sh
```diff
@@ -97,6 +106,52 @@ proto_yggdrasil_generate_keypair() {
...
+}
+
+proto_yggdrasil_generate_jumper_config() {
+	local config="$1"
+	local ygg_sock="$2"
+	local ygg_cfg="$3"
+
+	# Autofill Yggdrasil listeners
+	config_get is_autofill_listeners "$config" "jumper_autofill_listen_addresses"
+	if [ "$is_autofill_listeners" == "1" ]; then
```
systemcrash, line 137:
```shell
	if [ "$is_autofill_listeners" = "1" ]; then
```

net/yggdrasil/files/yggdrasil.sh
```diff
@@ -97,6 +106,52 @@ proto_yggdrasil_generate_keypair() {
...
+		echo "yggdrasil_listen = ["
```
systemcrash, line 138: We generate a config by echoing to stdout?

net/yggdrasil/files/yggdrasil.sh
```diff
@@ -178,6 +233,14 @@ EOF
 
 	json_add_array "Listen"
 	config_list_foreach "$config" listen_address proto_yggdrasil_add_string
+
+	# If needed, add new address for each previously unspecified protocol
+	config_get is_jumper_enabled "$config" "jumper_enable"
+	config_get allocate_listen_addresses "$config" "allocate_listen_addresses"
+	if [ "$is_jumper_enabled" == "1" ] && [ "$allocate_listen_addresses" == "1" ]; then
```
systemcrash, line 240:
```shell
	if [ "$is_jumper_enabled" = "1" ] && [ "$allocate_listen_addresses" = "1" ]; then
``` 
or simply:

```shell
	[ "$is_jumper_enabled" = "1" ] && [ "$allocate_listen_addresses" = "1" ] && proto_yggdrasil_allocate_listen_addresses "$config"
``` 
?

net/yggdrasil/files/yggdrasil.sh
```diff
@@ -193,6 +256,16 @@ EOF
 	proto_add_ipv6_address "$(yggdrasil -useconffile "${ygg_cfg}" -address)" "7"
 	proto_add_ipv6_prefix "$(yggdrasil -useconffile "${ygg_cfg}" -subnet)"
 	proto_send_update "$config"
+
+	# Start jumper if needed
+	config_get is_jumper_enabled "$config" "jumper_enable"
+	if [ "$is_jumper_enabled" == "1" ] && [ -f /usr/sbin/yggdrasil-jumper ]; then
```
systemcrash, line 262:
```shell
	if [ "$is_jumper_enabled" = "1" ] && [ -f /usr/sbin/yggdrasil-jumper ]; then
``` 

or better

```shell
	if [ "$is_jumper_enabled" -ne 0 ] && [ -f /usr/sbin/yggdrasil-jumper ]; then
```

systemcrash (changes requested):

net/yggdrasil/files/yggdrasil.sh
one-d-wide: Yeah, this could have been even better described simply using `grep -q`.

Btw, what's wrong with `&>...` redirection? `man bash` says it is identical to `2>&1 >...` and imo it looks cleaner.

net/yggdrasil/files/yggdrasil.sh
one-d-wide: Agree. Do/how I should change it after it has been committed to the master branch?

net/yggdrasil/files/yggdrasil.sh
one-d-wide: I think explicit redirection on the caller's end looks more concise then repeatedly doing the same after each `echo` call.

Though I probably should have called it like something that "prints" rather than "generates".

net/yggdrasil/files/yggdrasil.sh
systemcrash: Follow-up commit should be trivial now.

net/yggdrasil/files/yggdrasil.sh
systemcrash: https://www.shellcheck.net/wiki/SC3020

systemcrash: Also, what is the `! echo "$protocols"` bit supposed to do? I can guess but I don't like guessing. Syntax is a bit vague. Checking empty return value or string?

one-d-wide: This whole condition checks if `$protocol` is not present in `$protocols`.
So `!` here acts as a shell keyword that inverting result of the pipeline as a whole (not just `echo`).
And thinking about it, `grep -q --invert-match` would probably fit better here.

Issue #20587, opened 2023-03-03
The erlang package does not include the no_dot_erlang.boot file


helloyan:
Maintainer: @cretingame
Environment: (put here arch, model, OpenWrt version)
OpenWrt R23.2.14 / LuCI Master (git-23.061.09355-88ea4f3)

Description:

```
The file /usr/lib/erlang/bin/no_dot_erlang.boot is not present
```

Neustradamus:
@antonlacon, @yousong, @champtar, @neheb, @ja-pa, @peterwillcn, @micmac1, @lipnitsk, @svlobanov, @dhewg, @paulfertser, @aparcar: Have you seen this ticket?

cretingame:
Sorry, I'm not maintaining this package anymore. I cannot help you.

PR #23073, opened 2024-01-06
erlang: update to 26.2.1

krant:
- Update to Erlang/OTP 26.2.1
- Remove obsolete patches, add new one
- Fix https://github.com/openwrt/packages/issues/20587
- Change URL to https


lang/erlang/patches/010-openssl-md5.patch
```diff
@@ -0,0 +1,421 @@
+--- a/erts/emulator/openssl/crypto/md5/md5_local.h
++++ b/erts/emulator/openssl/crypto/md5/md5_local.h
```
BKPepe, line 2: Why do we add patch without Git header? There are no details, why do we need to have it. Who is the author, etc.


Could you please take this patch as it should be: https://patch-diff.githubusercontent.com/raw/erlang/otp/pull/7989.patch ?

BKPepe (changes requested):

lang/erlang/Makefile
```diff
@@ -296,26 +296,15 @@ $(call Package/erlang/Default/description)
...
-
-HOST_CFLAGS += \
-        -DHAS_EVP_PKEY_CTX \
-        -DHAVE_EVP_CIPHER_CTX_COPY
-
 HOST_CONFIGURE_ARGS += \
 	--with-ssl="$(STAGING_DIR_HOST)" \
-	--disable-hipe \
-	--disable-pgo \
-	--disable-smp-support \
```
BKPepe, line 310: This is removed why?

lang/erlang/Makefile
```diff
@@ -296,26 +296,15 @@ $(call Package/erlang/Default/description)
...
 	--without-javac
 
 # Target
 
 CONFIGURE_ARGS += \
-	--disable-hipe \
-	--disable-smp-support \
```
BKPepe, line 317: Removed why?

lang/erlang/Makefile
```diff
@@ -35,7 +35,7 @@ define Package/erlang/Default
   SECTION:=lang
   CATEGORY:=Languages
   TITLE:=Erlang/OTP programming language
-  URL:=http://www.erlang.org/
+  URL:=https://www.erlang.org/
```
BKPepe, line 38: This change is described where?

lang/erlang/Makefile
```diff
@@ -296,26 +296,15 @@ $(call Package/erlang/Default/description)
...
-	--disable-smp-support \
 	--without-javac
 
 # Target
 
 CONFIGURE_ARGS += \
-	--disable-hipe \
-	--disable-smp-support \
 	--without-javac \
+	--disable-builtin-zlib \
```
BKPepe, line 307: This is necessary why?

lang/erlang/patches/010-openssl-md5.patch
krant: Totally makes sense, done.

lang/erlang/Makefile
krant: I believe CFLAGS modification was required for older versions of erlang/libressl. Latest Erlang builds and works without it.
Regarding removed configure options - there are no such switches in Erlang anymore. HiPE was completely removed, VM always builds in SMP mode, and `--disable-pgo` has never existed I think.

lang/erlang/Makefile
krant: Same as above

lang/erlang/Makefile
krant: Just to be sure that OpenWrt's zlib will be used, not vendored one. [Buildroot](https://github.com/buildroot/buildroot/blob/b37249581fa4a221923540478db7fa720e04035d/package/erlang/erlang.mk#L76-L77) or [Alpine](https://git.alpinelinux.org/aports/tree/community/erlang/APKBUILD#n73) has similar options.

lang/erlang/Makefile
krant: You're right, I've PR description has been updated

lang/erlang/Makefile
BKPepe: Soo about CLAFGS, shouldnt you do it in separate commit, so it is easier to revert it, if anything happens?

BKPepe:
Everything should be explained in the commit description, thats why the commit is for.

lang/erlang/Makefile
krant: It seems unnecessary indeed - Erlang is getting linked with host zlib  without it. I've removed this line.

krant:
@BKPepe commits have been re-structured

lang/erlang/Makefile
BKPepe: Ok, but you should mention that in the commit description, because from the commit, it looks like removing, but why? Because we can. :D

krant: Changed  "remove CFLAGS modification required for older versions" into "remove CFLAGS modification since it has been fixed in the upstream", hope it looks better for you. Here is the fix in the upstream - https://github.com/erlang/otp/commit/8e5170d79e387d7e898d948a78e0c70407a4030c.

krant:
Hey @BKPepe , it seems all issues has been fixed, could we have it merged?

PR #23078, opened 2024-01-08
Strongswan: Added support for EAP-MSCHAPv2 authentication

peci1:
Maintainer: @pprindeville , @Thermi 
Compile tested: NO, only changed init script
Run tested: mvebu/cortexa9/Turris Omnia/OpenWRT 21.02 (with only the init script updated to the upstream version).

Description:

This PR mainly adds support for configuring EAP-MSCHAPv2 user auth. This is the only one that is sanely supported by Windows clients.

I also fixed a bug with the pools section generation - it got generated once for each remote instead of just once.

And I also added option `send_cert` which forces sending server cert even if nobody asks for it - some clients are easier to configure this way.

I tested the changes on a swanctl-based server configured with two remotes - one IKEv2 PSK, one IKEv2 EAP-MSCHAPv2. The generated configs are valid and the server works.

### DIFF
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -610,10 +610,6 @@ config_remote() {
 		fatal "AuthenticationMode $auth_mode not supported"
 	fi
 
-	swanctl_xappend0 "pools {"
-	config_list_foreach "$conf" pools config_pool
-	swanctl_xappend0 "}"
-
 	swanctl_xappend0 ""
 }
 
@@ -689,6 +685,10 @@ prepare_env() {
 	config_load ipsec
 	config_foreach config_ipsec ipsec
 	config_foreach config_remote remote
+	
+	swanctl_xappend0 "pools {"
+	config_foreach config_pool pools
+	swanctl_xappend0 "}"
 
 	do_postamble
 }
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -466,6 +466,7 @@ config_remote() {
 	config_get ca_cert "$conf" ca_cert ""
 	config_get rekeytime "$conf" rekeytime
 	config_get overtime "$conf" overtime
+	config_get send_cert "$conf" send_cert
 
 	config_list_foreach "$conf" local_sourceip append_var local_sourceip ","
 	config_list_foreach "$conf" remote_ca_certs append_var remote_ca_certs ","
@@ -559,6 +560,8 @@ config_remote() {
 		keyexchange=
 		;;
 	esac
+
+	[ -n "$send_cert" ] && swanctl_xappend2 "send_cert = $send_cert"
 
 	[ $mobike -eq 1 ] && swanctl_xappend2 "mobike = yes" || swanctl_xappend2 "mobike = no"
 
--- a/net/strongswan/files/swanctl.init
+++ b/net/strongswan/files/swanctl.init
@@ -421,6 +421,21 @@ config_pool() {
 	swanctl_xappend1 "}"
 }
 
+config_mschapv2_secret() {
+	local conf="$1"
+
+	local id
+	local secret
+
+	config_get id "$conf" id
+	config_get secret "$conf" secret
+
+	swanctl_xappend1 "eap-${conf} {"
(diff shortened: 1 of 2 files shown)


peci1:
Thanks for the feedback. I actually did read the contribution guide before submitting the PR to craft it as well as possible to the chosen maintenance model, but apparently I did read the wrong one: https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md . I guess there should be at least a link to the "real" contribution guide in this document?

I'll address the feedback later. But even in the "real" contribution guide, there is no mention of PRs to bumb release versions. I thought that is something that will be decided by the maintainers, to e.g. give the possibility to batch up changes or doing some other strategic choices. But if the packages are always built off of the master branch, then in really makes sense to bump the version in every PR. It is just an uncommon practice for me...

BKPepe:
> I'll address the feedback later. But even in the "real" contribution guide, there is no mention of PRs to bumb release versions.

I can see it there:
```
PKG_RELEASE should be initially set to 1 or reset to 1 if the software version is changed. You should increment it if the package itself has changed. For example, modifying a support script, changing configure options like --disable_ or --enable_ switches, or if you changed something in the package which causes the resulting binaries to be different. Changes like correcting md5sums, changing mirror URLs, adding a maintainer field or updating a comment or copyright year in a Makefile do not require a change to PKG_RELEASE.
```

Mea culpa, it should be ``PKG_RELEASE`` instead of ``PKG_VERSION`` (which is version from upstream).

Yup, improving the contributing guidelines is on my to-do list to improve it.

peci1:
> I can see it there:
> 
> ```
> PKG_RELEASE should be initially set to 1 or reset to 1 if the software version is changed. You should increment it if the package itself has changed. For example, modifying a support script, changing configure options like --disable_ or --enable_ switches, or if you changed something in the package which causes the resulting binaries to be different. Changes like correcting md5sums, changing mirror URLs, adding a maintainer field or updating a comment or copyright year in a Makefile do not require a change to PKG_RELEASE.
> ```

Eh, that is in https://openwrt.org/docs/guide-developer/packages, which is only linked from the contribution guidelines. This is however so important that I think it belongs directly to the github contrib guide, too. Or could github just redirect to the openwrt page to keep things consistent? :)

BKPepe:
I found it here:
https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#makefile-contents-should-contain

peci1:
> I found it here:

So I'm just blind, sorry!

BTW. I've force-pushed the updated commits with hopefully a more satisfying descriptions.

BKPepe:
Looks better now, great improvement! 👏

feckert (commented):
Please check your indentation. When I look at the code, it is not correct everywhere! We use tabs and not spaces.

Neustradamus:
@peci1, @pprindeville: Have you seen the latest @systemcrash comment?

systemcrash:
I think I can answer my own question: that these changes do not depend on a kernel module but they won't do anything without it.

net/strongswan/files/swanctl.init
```diff
@@ -684,6 +707,17 @@ prepare_env() {
 	config_load ipsec
 	config_foreach config_ipsec ipsec
 	config_foreach config_remote remote
+	
+	swanctl_xappend0 "# Global config"
+	swanctl_xappend0 ""
+	
+	swanctl_xappend0 "pools {"
+	config_foreach config_pool pools
```
pprindeville, line 715: If I'm reading this correctly, we're changing the section that `pools` are defined in.  Won't that break existing configurations without a migration script being applied?

pprindeville (commented):
Does this break compatibility?

net/strongswan/files/swanctl.init
peci1: See the text of commit 23ccb705184de16aef8001bc183d0786e1e23cde . I don't remember exactly what I did 1 year ago, but if the commit message is right, this change is not breaking anything.

pprindeville:
Please rebase

peci1:
Rebased.

systemcrash:
OKTM

PR #23083, opened 2024-01-10
snort3: finish up several incomplete capabilities

efahl:
Reporting
 - Use json alert data for 10x speed improvement in report generation
 - Include both gid and sid, plus packet direction in report output
 - Add by-date incident filtering
 - Add verbose mode which displays actual rules triggered and their source
 - Attempt to look up host names from IPs in verbose mode
 - Clean up display of port number involved in incidents

Rules
 - Complete downloader for subscription rules using oinkcode (only tested with snort.org's "free" tier subscription)
 - Auto-detect multiple rules files and include them in lua 'ips.rules'
 - Add '--backup' option to copy out current rules before installing new
 - Add '--persistent' option to 'snort-rules', storing in persistent location

CLI interface
 - Completely rework command line option parsing in all user scripts
 - Allow options and commands to be in any order on command line
 - Add long-form names for all options ('--help' for '-h' and so on)
 - Detect errors properly in options, enhance help pages

Bug fixes
 - Use 'mkdir -p' on all directory creation
 - Use proper tmp directory from 'snort.snort.temp_dir' everywhere

Run tested: x86/64

### DIFF
--- a/net/snort3/Makefile
+++ b/net/snort3/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=snort3
 PKG_VERSION:=3.1.78.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://github.com/snort3/snort3/archive/refs/tags/
--- a/net/snort3/files/main.uc
+++ b/net/snort3/files/main.uc
@@ -1,6 +1,6 @@
 {%
 //------------------------------------------------------------------------------
-// Copyright (c) 2023 Eric Fahlgren <eric.fahlgren@gmail.com>
+// Copyright (c) 2023-2024 Eric Fahlgren <eric.fahlgren@gmail.com>
 // SPDX-License-Identifier: GPL-2.0
 //
 // The tables defined using 'config_item' are the source of record for the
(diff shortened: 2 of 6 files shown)


graysky2:
Looks good to me, recommend merge.

efahl:
In case anyone cares, I've started the documentation at https://openwrt.org/docs/guide-user/services/snort#auto-configuration

efahl:
Merged version update from 0b2728b2f2aa3f412469178828613e85c13fe74f

hnyman:
@efahl 
You should probably add yourself to the snort maintainer list as you are active with the package, also authoring ucode for it.  

@graysky2 has also been active with snort, 
but @flyn-org has not been active with snort for the past year, so perhaps it might be ok to replace him.

> PKG_MAINTAINER:=W. Michael Petullo <mike@flyn.org>, John Audia <therealgraysky@proton.me>

flyn-org:
I would rather not be replaced, but I would be pleased if @efahl became a comaintainer. My lack of activity is due only because @graysky2 is so quick to update the package, which I appreciate.

efahl:
Ok, I can add myself as co-maintainer. I'll add a separate commit with just that once this one gets merged (I'm still feel far from being git-proficient, so I like to keep things simple on my end).

I've also got more improvements pending, most notably using jow's fix to "jsonfilter -a"/array handling...

efahl:
And sorry about this massive commit.  I wanted to do it in about five pieces, but the dependencies got tangled and I just said screw it and tossed the whole mess into one big intermixed blob.  My future commits will be much easier to review. 😁

PR #23084, opened 2024-01-10
ci: no longer require real name

aparcar:
This goes in accordance with the Linux Kernel:

> using a known identity (sorry, no anonymous contributions.)

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=HEAD#n442

### DIFF
--- a/.github/workflows/formal.yml
+++ b/.github/workflows/formal.yml
@@ -34,14 +34,6 @@ jobs:
               RET=1
             fi
 
-            author="$(git show -s --format=%aN $commit)"
-            if echo $author | grep -q '\S\+\s\+\S\+'; then
-              success "Author name ($author) seems ok"
-            else
-              err "Author name ($author) need to be your real name 'firstname lastname'"
-              RET=1
-            fi
-
             subject="$(git show -s --format=%s $commit)"
             if echo "$subject" | grep -q -e '^[0-9A-Za-z,+/_-]\+: ' -e '^Revert ' -e '^CONTRIBUTING.md' -e '^README.md'; then
               success "Commit subject line seems ok ($subject)"


ynezz (approved):
Thanks!

one-d-wide:
Thank you for the quick response.

I submitted identical [pull request](https://github.com/openwrt/luci/pull/6825) to `openwrt/luci` repository.

1715173329:
So... Should we check email to prevent the use of `users.noreply.github.com`?

aparcar:
@1715173329 good call, but this could happen in a separate PR right?

1715173329 (approved):

BKPepe:
Honestly, I am not sure if this was right decision. I think that in the time manner it might seems like acceptable solution, but again we are getting commits, which dont have ``Signed-off-by`` at all. Doing this for one contributor, which was somehow loud then in my point of view benefits to have this merged does not exceed those benefits, which we had in the past.

Also, I somehow fail to see reasons, why it was merged. Okay, known identity, might sounds nice. But let me ask one question. What is the known identity? You know me, I know you. Is that known? Perhaps in my case, yes, because I contributed for some time here, but is this the same case in the new contributor, which you have done this? Linux kernel does it, we should do that, too. But do we have policy or some rules about the known identity? Take it from the lawyer perspective, they can have all different opinions about this.

ynezz:
> we are getting commits, which dont have Signed-off-by at all

Thats not intended and likely unrelated to this PR. Can you perhaps provide examples of such commits and relevant PRs? Maybe there is something broken with SoB check? In other words, just the Real Name 

> Also, I somehow fail to see reasons, why it was merged.

See https://lists.openwrt.org/pipermail/openwrt-devel/2024-January/042058.html for details and prior discussions. 

> What is the known identity?

There is quite good [official Linux Foundation clarification](https://github.com/cncf/foundation/issues/383#issuecomment-1178254458) of the intended meaning, to make it clear that real names are not required, only ability to identify the person in the community. Feel free to adjust the relevant wiki page if you think, that link to kernel documentation is not good enough. Thanks!

BKPepe:
Based on recent @ynezz's https://github.com/openwrt/packages/pull/23072#issuecomment-2018331887, we should revert this here as well.

systemcrash:
> Based on recent @ynezz's [#23072 (comment)](https://github.com/openwrt/packages/pull/23072#issuecomment-2018331887), we should revert this here as well.

First, if this requires a vote or 'discussion', then have one. But for heaven's sake, how substantial does it have to be more than the discussion already had here?

BKPepe:
It looks like that the conversation on the mailing list is somehow quiet about this. I think, we should revert this until the decision is made for all OpenWrt projects.

PR #23098, opened 2024-01-14
Use HTTPS if available for some package sources

dcousens:
**Maintainer:** N/A
**Compile tested:** N/A
**Run tested:** N/A

**Other**
I have verified that HTTPS is supported for the updated hosts.
I could not apply this to every `http://`, as although HTTPS is often supported, it is not configured properly for the host domain provided, and thus many packages have been omitted from this pull request.

### Description:
(~copy of https://github.com/openwrt/openwrt/pull/10799)
This pull request changes a number of `PKG_SOURCE_URL`s that were using `http:`, to use `https:` if they were available.

I understand that this doesn't impact the integrity of the downloaded package, as `PKG_HASH` is for that purpose.
None-the-less, this is a beneficial change as HTTPS is the modern baseline.


dcousens:
Happy to rebase if necessary

mhei (changes requested):
You should rebase and squash into a single commit with proper description etc and then force push to this PR. See contribution hints for details.

dcousens:
Rebased and squashed, I don't necessarily know what description that is wanted, as this touches a number of different packages.

mhei:
Yes, since it covers multiple packages, the first rule (prefixed package name) cannot apply here.
Let's just ignore it, don't use boilerplate like the star.

For the proper commit message, we can re-use wording from your PR. The following example is short and should be completely sufficient:

```
Use https for some PKG_SOURCE_URLs 

This changes a number of PKG_SOURCE_URLs that were using http:, to use https: if they were available.
It was verified that HTTPS is supported properly for the updated hosts.

Signed-off-by: ...
```

The SoB is missing in your commit, see https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#commits-in-your-pull-requests-should for details.

dcousens:
Thanks @mhei, I'll try and update everything this week

BKPepe:
Kind reminder.

dcousens:
Thanks @BKPepe, I will attempt to rebase this and re-sign soo

dcousens:
@BKPepe should be OK to go now

BKPepe:
I am not sure, why this was closed. However, it is merged to the master branch - https://github.com/openwrt/packages/commit/8a6a3d6f5d92897cf191a12e8264e22010758113

dcousens:
Thanks @BKPepe! I didn't have time to keep the branch up to date, that's all 💙

PR #23110, opened 2024-01-15
autossh: make forwarding optional

kkubicki:
Maintainer: @jempatel / @neheb 

Compile tested: x86_64 b7f9742da8
Run tested: Netgear R6220, OpenWrt 23.05.2, r23630-842932a63d 

Description:
Previous change made localport option required, which broke older configs and was not documented in any tutorials regarding autossh package (e.g. https://openwrt.org/docs/guide-user/services/ssh/autossh).
This change makes in-config forwarding only optional, restoring original configuration template valid.
New localport & remoteport options are still accepted, however extra test is performed to ensure both of these options are provided as expected by ssh command.

### DIFF
--- a/net/autossh/Makefile
+++ b/net/autossh/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=autossh
 PKG_VERSION:=1.4g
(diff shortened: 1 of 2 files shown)


kkubicki:
SoB should be fine now.

PR #23120, opened 2024-01-18
p910nd: hotplug script

systemcrash:
Maintainer: @ptpt52 (?) / Philipp Kerling <pkerling@casix.org>
Run tested: ath79, Archer c7v2, 23.05.2, extensive testing

See also #23094 

Description: p910nd hotplug script which auto-configures all printer info, and the loading of its driver blob. The hotplug script also attempts to load an appropriately named driver blob on the filesystem. Just BYOB (bring your own blob).

One can fix the luci frontend to assist with uploading a blob to match what this hotplug script sets.

Review welcomed:
@pprindeville @jow- @hnyman


gth-42:
In case it helps, some devices of the same vendor/product can be uniquely ID'd with iSerialNumber and a symlink created.  Adding the following to the "add" action around line 354 works for me:

'''
ln -s /dev/"$DEVNAME" /dev/print_$iSerialNumber
logger -t "$DAEMON_HOTPLUG" $DEVNAME, $iSerialNumber, /dev/print_$iSerialNumber
'''

p910nd config using /dev/print_## is able to use the same device no matter the plug/power order.

Thank you for your work.

systemcrash:
That's a great suggestion - it might be actual, since in the next main kernel v6.6 there's a patch which makes USB load order non-deterministic, and I'm not sure whether we will carry it.

PR #23123, opened 2024-01-18
libdaq3: update to 3.0.14

graysky2:
Update to latest version.

Changelog: https://github.com/snort3/libdaq/releases/tag/v3.0.14

Build system: x86/64

Maintainer: @flyn-org


BKPepe:
Would be great to include link for release notes and maybe avoid using "upstream bump", because... will there be "downstream bump"?

PR #23139, opened 2024-01-24
iputils: bump to 20240117

drujd:
Maintainer: @nmeyerhans
Run tested: Turris Omnia running OpenWrt master


krant:
Oops, I've missed this PR and created #23231 instead. It slightly simplifies build options, so I'm not sure which one is to keep :thinking:

PR #23141, opened 2024-01-24
libs/libvpx: fix PKG_CPE_ID

ffontaine:
There is not a single CVEs under cpe:/a:john_koleszar:libvpx so use cpe:/a:webmproject:libvpx:
https://nvd.nist.gov/products/cpe/search/results?namingFormat=2.3&keyword=cpe%3A2.3%3Aa%3Awebmproject%3Alibvpx

Maintainer: @luizluca
Compile tested: Not needed
Run tested: Not needed


luizluca:
LGTM. Thanks

PR #23150, opened 2024-01-25
net/tcpreplay: fix PKG_CPE_ID

ffontaine:
There is not a single CVE under cpe:/a:appneta:tcpreplay so use cpe:/a:broadcom:tcpreplay:
https://nvd.nist.gov/products/cpe/search/results?keyword=cpe:2.3:a:broadcom:tcpreplay

Maintainer: @commodo 
Compile tested: Not needed
Run tested: Not needed


commodo:
thanks :)

PR #23163, opened 2024-01-25
libs/libinput: assign PKG_CPE_ID

ffontaine:
https://nvd.nist.gov/products/cpe/search/results?keyword=cpe:2.3:a:freedesktop:libinput

Maintainer: @aparcar
Compile tested: Not needed
Run tested: Not needed


aparcar:
@dangowrt is actually the maintainer here :)

PR #23169, opened 2024-01-25
nebula: update to 1.8.2-2

stangri:
Maintainer: me

Description:
The following fixes have been applied to Makefile:
* fix the nebula license type
* add PKG_CPE_ID
* remove unneeded call to Build/Compile
* add leading spaces to descriptions
* add Package/nebula/conffiles definition
* remove unneeded /lib/upgrade/keep.d files
* no longer install actual license file

Kudos to @BKPepe for feedback on a similar package which lead to these fixes

### DIFF
--- a/net/nebula/Makefile
+++ b/net/nebula/Makefile
@@ -1,19 +1,20 @@
 # Copyright 2021-2023 Stan Grishin (stangri@melmac.ca)
-# This is free software, licensed under the MIT License.
+# This is free software, licensed under the Apache 2.0 License.
 
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nebula
 PKG_VERSION:=1.8.2
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/slackhq/nebula/tar.gz/v$(PKG_VERSION)?
 PKG_HASH:=203713c58d0ec8a10df2f605af791a77a33f825454911ac3a5313ced591547fd
 
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.ca>
-PKG_LICENSE:=MIT
+PKG_LICENSE:=Apache-2.0
 PKG_LICENSE_FILES:=LICENSE
+PKG_CPE_ID:=cpe:/a:slack:nebula
 
 PKG_BUILD_DEPENDS:=golang/host
 PKG_BUILD_PARALLEL:=1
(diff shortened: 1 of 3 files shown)


net/nebula/Makefile
```diff
@@ -69,53 +70,45 @@ define Package/nebula-service
...
-Nebula is a scalable overlay networking tool with a focus on performance, simplicity
-and security. It lets you seamlessly connect computers anywhere in the world.
-This package contains only OpenWrt-specific init.d script for nebula.
+  Nebula is a scalable overlay networking tool with a focus on performance, simplicity
+  and security. It lets you seamlessly connect computers anywhere in the world.
+  This package contains only OpenWrt-specific init.d script for nebula.
+endef
+
+define Package/nebula/conffiles
+/etc/nebula/
```
BKPepe, line 99: Do you really need to backup the entirely folder of /etc/nebula/? Wouldnt it be sufficient to mark as configuration file only config file?

Also, in the previous solution, there was conffiles for nebula and nebula-cert package. Right now, it is only for nebula. Is it wanted?

net/nebula/Makefile
```diff
@@ -69,53 +70,45 @@ define Package/nebula-service
   PKGARCH:=all
 endef
```
BKPepe, line 72: again the URL should point to projects URL, not your downstream URL project. 

Anything else is unrelevant, I checked the URL.

1. How to install packages
Oh well, here we go again. Thats why there is LuCI, if you are familiar with SSH, then you know how to install packages. If not, go look at documentation (OpenWrt wiki)

2. Default settings, hm... shouldnt it be default?
Kinda obvious.

I dont see benefits or advantages to do the documentation, which are you doing. Sorry.

stangri: You're entitled to your opinion, however an attentive reader would have noticed that how to install and default settings are not numbers 1 and 2, but numbers 3 and 4.

BKPepe: Oh boy, you should really stop reinvent a wheel again.

Look here:
https://openwrt.org/docs/guide-developer/packages

What is in PKG_URL? Oh... yeah! Upstream project URL.

stangri: And yet, the PKG_URL is not even present in the example Makefile.

I see no benefit for OpenWrt users to have the link pointing to a page which has very little relevance to the OpenWrt package.

I appreciate a lot of your other feedback on this and other packages, but your holy war against my documentation repository (of which you choose to ignore important sections) is bordering on harassment.

net/nebula/Makefile
```diff
@@ -69,53 +70,49 @@ define Package/nebula-service
...
```
1715173329, line 71: I found the same variable was defined multiple times, this can be avoided by creating a shared recipe like:
```
define Package/nebula/Default
  SECTION:=net
  CATEGORY:=Network
  URL:=https://docs.openwrt.melmac.net/nebula/
end
```

then call it:
```
define Package/nebula
  $(call Package/nebula/Default)
  TITLE:=nebula
  ...
endef
```

the same applies to descriptions:
```
define Package/nebula/description/Default
  Nebula is a scalable overlay networking tool with a focus on performance, simplicity
  and security. It lets you seamlessly connect computers anywhere in the world.
endef
```

```
define Package/nebula/description
  $(call Package/nebula/description/Default)
  This package contains only nebula binary. Unless you want to start nebula manually,
  you may want to also install *either* 'nebula-service' *or* 'nebula-proto' package.
endef
```

net/nebula/Makefile
```diff
@@ -58,64 +59,59 @@ define Package/nebula-proto
...
+  This package contains only OpenWrt-specific init.d script for nebula.
+endef
+
+define Package/nebula/conffiles
+/etc/nebula/
 endef
 
+Package/nebula-cert/conffiles = $(Package/nebula/conffiles)
+Package/nebula-proto/conffiles = $(Package/nebula/conffiles)
+Package/nebula-service/conffiles = $(Package/nebula/conffiles)
```
1715173329, line 103: Are these really necessary? At least `nebula-proto` and `nebula-service` depend on `nebula` which has conffiles defined already.

stangri: Good catch, thanks!

1715173329 (approved):
LGTM

PR #23170, opened 2024-01-26
bcm27xx-eeprom: switch to bcm27xx-utils

Noltari:
Maintainer: me
Compile tested: bcm2711, bcm2712
Run tested: bcm2711/RPi4, bcm2712/RPi5

Description: bcm27xx-userland is now deprecated and utils should be used instead.
https://github.com/raspberrypi/userland/commit/96a7334ae9d5fc9db7ac92e59852377df63f1848

### DIFF
--- a/utils/bcm27xx-eeprom/Makefile
+++ b/utils/bcm27xx-eeprom/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bcm27xx-eeprom
 PKG_VERSION:=v.2024.01.05-2712
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/raspberrypi/rpi-eeprom/tar.gz/$(PKG_VERSION)?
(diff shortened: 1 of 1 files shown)


BKPepe:
I havent look at closer, but I think this can be solved by better solution, if isnt already of course. 
Couldnt utils provide userland and it would be solved?

Noltari:
> I havent look at closer, but I think this can be solved by better solution, if isnt already of course. Couldnt utils provide userland and it would be solved?

@BKPepe yes, I could do that, but my intention is to completely remove bcm27xx-userland from the main repository

BKPepe:
Even in OpenWrt 23.05, you should replace it. It is the stable branch, of course, removal packages would not be good, but since OpenWrt does not release new versions very often, I suggest at least even in the stable branch to use ``PROVIDES``. In that case, you could remove it, there. :) Because there will be no package missing.

Noltari:
> Even in OpenWrt 23.05, you should replace it. It is the stable branch, of course, removal packages would not be good, but since OpenWrt does not release new versions very often, I suggest at least even in the stable branch to use `PROVIDES`. In that case, you could remove it, there. :) Because there will be no package missing.

Just to clarify, my intention is to have only bcm27xx-utils on main branch, removing bcm27xx-userland.
BTW, bcm27xx-utils already PROVIDES bcm27xx-userland: https://github.com/openwrt/openwrt/blob/75ad5c24142a18fe22bc7ee5f46759a58a832f7c/package/utils/bcm27xx-utils/Makefile#L30
And make no changes on 23.05, so bcm27xx-eeprom will still use bcm27xx-userland since it has been deprecated, but it supports the 23.05 RPi models with no issues.

Knowing that, is it OK to merge this @BKPepe?

1715173329 (approved):

PR #23176, opened 2024-01-27
miniupnpd: Update package to 2.3.6

Self-Hosting-Group:
and change title to term used in LuCi

Maintainer:
Compile tested: not tested on OpenWrt
Run tested: not tested on OpenWrt

### DIFF
--- a/net/miniupnpd/Makefile
+++ b/net/miniupnpd/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=miniupnpd
-PKG_VERSION:=2.3.3
-PKG_RELEASE:=2
+PKG_VERSION:=2.3.6
+PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


1715173329:
> but since the old name is still accepted, no change was made in the PR.

Please follow upstream change and migrate to new config name.

BKPepe:
Also, make sure that it is compile and run tested for OpenWrt.

BKPepe:
Ehh... for me it seems like the title is the same. Why do we need to have this change? Since, there is no commit description, it is not clear. I am in favor to close this PR.

stangri:
> @stangri: Thank you for your reply. PR updated. Can we possibly do the PR without the configuration change (as backword compatible) and do it in a later PR (including LuCi)? I am waiting for some more upstream changes.

That's up to maintainer, I just provided (hopefully helpful) feedback.

BKPepe:
CI is still failing for you. Would you mind to fix it?

PR #23191, opened 2024-01-30
udpxy: workaround bug introduced by SSM.

cgang:
There is a bug introduced by SSM support, the source address
is uninitialized if it's not specified in command line argument.
It will cause UDP socket waiting for packet from random address,
and result in timeout in that case.

This patch is trying to workaround the bug in OpenWrt.

Signed-off-by: Gang Chen <11877522+cgang@users.noreply.github.com>

Maintainer: @neheb 
Compile tested: Linksys E8450 mediatek/mt7622 OpenWrt 23.05.2
Run tested: Linksys E8450 mediatek/mt7622 OpenWrt 23.05.2

Description:
This bug can be reproduced with Linksys E8450 mediatek/mt7622 OpenWrt 23.05.2.

### DIFF
--- a/net/udpxy/patches/001-fix_uninitialized_source_address.patch
+++ b/net/udpxy/patches/001-fix_uninitialized_source_address.patch
@@ -1,0 +1,20 @@
+--- a/chipmunk/udpxrec.c
++++ b/chipmunk/udpxrec.c
+@@ -283,6 +283,7 @@ subscribe( int* sockfd, struct in_addr*
+ 
+     assert( sockfd && mcast_inaddr && s_address );
+ 
++    memset( s_address, 0, sizeof(struct sockaddr_in) );
+     if (strlen(source_ipaddr) != 0 && 1 != inet_aton( source_ipaddr, &s_address->sin_addr)) {
+         mperror( g_flog, errno,
+                 "%s: Invalid source address (SSM) [%s]: inet_aton",
+--- a/chipmunk/udpxy.c
++++ b/chipmunk/udpxy.c
+@@ -693,6 +693,7 @@ udp_relay( int sockfd, struct server_ctx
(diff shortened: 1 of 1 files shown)


neheb:
needs a PKG_RELEASE bump

BKPepe:
1. Needs Git header (Who is the author, commit description, commit subject), etc.

2. Why do wee need to have this patch? Upstream it, if possible.

cgang:
> 1. Needs Git header (Who is the author, commit description, commit subject), etc.
udpxy: workaround bug introduced by SSM.

There is a bug introduced by SSM support, the source address
is uninitialized if it's not specified in command line argument.
It will cause UDP socket waiting for packet from random address,
and result in timeout in that case.

This patch is trying to workaround the bug in OpenWrt.


> 2. Why do wee need to have this patch? Upstream it, if possible.
The udpxrec will crash on some device, such as Linksys E8450 without this fix.

net/udpxy/patches/001-fix_uninitialized_source_address.patch
```diff
@@ -0,0 +1,20 @@
+--- a/chipmunk/udpxrec.c
++++ b/chipmunk/udpxrec.c
```
BKPepe, line 2: 1. Please backport this patch as is from https://patch-diff.githubusercontent.com/raw/pcherenkov/udpxy/pull/32.patch it is possible that it will need to have refresh.

2. It requires to bump PKG_RELEASE in the Makefile.

cgang:
Hi @BKPepe 
Seems @pcherenkov accept the fix, but I just realize there is a mistake in it.
https://github.com/pcherenkov/udpxy/pull/33
Sorry for the regression issue and need to bother you again.

PR #23199, opened 2024-01-30
libs/libdaq3: assign PKG_LICENSE_FILES

ffontaine:
Maintainer: @flyn-org 
Compile tested: Not needed
Run tested: Not needed


flyn-org:
Looks good. Recommend merge.

It is odd that upstream includes both COPYING and LICENSE with the same contents, but I guess that it their choice.

PR #23209, opened 2024-01-31
Nmap: adjust SSL dependencies

rockdrilla:
Maintainer: @nunojpg
Compile-tested: ath79/generic (TP-Link Archer C7 v2), mediatek/filogic (GL.iNet GL-MT6000)
Run-tested: mediatek/filogic (GL.iNet GL-MT6000)

Description:
- nmap: ncat: use default CA bundle
- nmap: unify SSL dependencies

### DIFF
--- a/net/nmap/Makefile
+++ b/net/nmap/Makefile
@@ -14,7 +14,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nmap
 PKG_VERSION:=7.93
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 PKG_MAINTAINER:=Nuno Gonçalves <nunojpg@gmail.com>
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
--- a/net/nmap/Makefile
+++ b/net/nmap/Makefile
@@ -73,14 +73,14 @@ endef
 
 define Package/ncat-ssl
 $(call Package/nmap/default)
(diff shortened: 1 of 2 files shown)


rockdrilla:
NB: my ncat patch differs from patch in Debian ([permalink](https://salsa.debian.org/pkg-security-team/nmap/-/blob/279e5b078297818a08c364f9de2e3fae27596520/debian/patches/0002-Use-Debian-s-ca-certificates.patch)).

nunojpg:
For sure it's good to use the existing certificates, but do we need to force it as a dependency? That is not the behaviour in for example cURL on debian. You install ca-certificates if you want in addition.

rockdrilla:
Package "ca-bundle" (provides "ca-certs") is shipped as default package since openwrt/openwrt@e79df351 so this MR just declares dependency but not adds more packages. I think "ca-certs" dependency may be declared in "libopenssl" (and other SSL/TLS-related libraries) but this change should land in base and after proper discussion and tests.

rockdrilla:
Furthermore, visible change after this MR is that ncat-ssl/ncat-full will be 200kb thinner.

rockdrilla:
@nunojpg friendly ping. I've tested ncat with proposed changes.

nunojpg:
Sorry, 2 minutes was not enough since you completed it :) Sometimes we take up to 3 minutes to review.

nunojpg (approved):

nunojpg:
Cool. Thanks!

PR #23210, opened 2024-01-31
leptonica: update to 1.84.1

krant:
Maintainer: @vk496

Description:
- remove upstreamed patch
- explicitly disable openjpeg to ignore host-installed library
- fix .cmake and .pc paths


krant:
@BKPepe the package was run tested, why is the label?

hnyman:
Leptonica will need to be updated to 1.86.0 in order to be compatible with the current cmake.

PR #23213, opened 2024-01-31
tiff: update to 4.6.0

krant:
Maintainer: @jslachta 
Compile tested: ARMv7, Linksys WRT3200ACM, master branch

Description:
- Don't set CMake options which are on by default


jslachta:
Thank you!

PR #23217, opened 2024-01-31
libsndfile: update to 1.2.2

krant:
Maintainer: @tripolar 

Description:
- Update package URL to the official one
- Update source URL to the official one
- Modernize CMake options
- Fixup pkgconfig file
- Enable mpg123 support per users request (+7kB)

### DIFF
--- a/libs/libsndfile/Makefile
+++ b/libs/libsndfile/Makefile
@@ -8,14 +8,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libsndfile
+PKG_VERSION:=1.2.2
 PKG_RELEASE:=1
 
-PKG_SOURCE_PROTO:=git
-PKG_SOURCE_URL:=https://github.com/erikd/libsndfile.git
-PKG_SOURCE_DATE:=2019-04-21
-PKG_SOURCE_VERSION:=25824cb914fb3b79e18f31fb861e218c84be7d34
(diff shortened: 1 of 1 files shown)


libs/libsndfile/Makefile
```diff
@@ -27,8 +27,7 @@ define Package/libsndfile
   SECTION:=libs
   CATEGORY:=Libraries
   TITLE:=Library for reading/writing audio files
-  URL:=http://www.mega-nerd.com/libsndfile/
-  MAINTAINER:=Peter Wagner <tripolar@gmx.at>
+  URL:=https://libsndfile.github.io/libsndfile/
```
zdz2019, line 30: Need this:
  DEPENDS:=+lame-lib +libmpg123

krant: Makes sense, done

krant:
@BKPepe run-test has been successful

libs/libsndfile/Makefile
```diff
@@ -27,8 +27,8 @@ define Package/libsndfile
...
+  DEPENDS:=+lame-lib +libmpg123
```
BKPepe, line 31: Why do we need these dependencies right now? In the commit description, it says enable something, but without any details, why we need it, what is the size difference, etc.

krant: You're right, commit description has been updated

Issue #22308, opened 2023-10-05
update      zstd  1.5.5


Potterli20:
https://github.com/facebook/zstd/releases/tag/v1.5.5

BKPepe:
Feel free to submit pull request for it.

BKPepe:
In that case, reach upstream to get this working. Zstd in our repository does not have any maintainer, so someone from the OpenWrt community should step it and update it. Nothing what we can do here. There are plenty of issues here and we need to deal with that one by one. 

There isn't mentioned any CVEs between 1.5.2 (which we have here) and 1.5.5.

This is feature request and because of that closing. You might want to ask someone at OpenWrt forum or use IRC channels, which are available to use.

PR #23222, opened 2024-02-01
zstd: update to 1.5.5

krant:
Maintainer: N/A

Description:
- Don't set Meson options which are matching defaults

### DIFF
--- a/utils/zstd/Makefile
+++ b/utils/zstd/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=zstd
-PKG_VERSION:=1.5.2
-PKG_RELEASE:=2
+PKG_VERSION:=1.5.5
+PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://github.com/facebook/zstd/releases/download/v$(PKG_VERSION)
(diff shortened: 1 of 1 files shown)


systemcrash (approved):

utils/zstd/Makefile
```diff
@@ -64,12 +64,6 @@ endif
 MESON_ARGS += \
 	-Dlegacy_level=7 \
 	-Ddebug_level=0 \
-	-Dbacktrace=false \
-	-Dstatic_runtime=false \
-	-Dbin_programs=true \
-	-Dbin_tests=false \
-	-Dbin_contrib=false \
-	-Dmulti_thread=enabled \
```
BKPepe, line 72: I wonder, why we are removing this? Maybe it is matching the defaults, but anytime defaults can be changed and we do have at least some control about this.

cc: @neheb  - What do you think about this?

neheb: I wrote it to have all the options explicit. I have no real opinion on whether or not that should change.

krant: IMO having smaller and simpler packages is what makes them easier to maintain and keeping up to date. Duplicating default values goes against KISS principle and in the end makes a package more brittle, #22308 is a perfect example of this.
Of cause everyone is free to choose their habits and I don't want to spend your time with my cheap rants. But if we want solve the problem of outdated zstd in OpenWrt, I could step in as its maintainer and solve possible future issues with it - empty PKG_MAINTAINER suggests that this role is vacant.
Cheers!

PR #23226, opened 2024-02-01
graphicsmagick: update to 1.3.42

krant:
Maintainer: @val-kulkov 


krant:
@val-kulkov package has been adopted

PR #23230, opened 2024-02-01
procps-ng: update to 4.0.4

krant:
Maintainer: N/A

Description:
- Update the patch

### DIFF
--- a/utils/procps-ng/Makefile
+++ b/utils/procps-ng/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=procps-ng
-PKG_VERSION:=3.3.16
-PKG_RELEASE:=3
+PKG_VERSION:=4.0.4
+PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@SF/procps-ng
-PKG_HASH:=925eacd65dedcf9c98eb94e8978bbfb63f5de37294cc1047d81462ed477a20af
+PKG_HASH:=22870d6feb2478adb617ce4f09a787addaf2d260c5a8aa7b17d889a962c5e42e
 
 PKG_MAINTAINER:=
 PKG_LICENSE:=GPL-2.0-or-later
@@ -93,7 +93,7 @@ MAKE_FLAGS += \
 
 define Package/procps-ng/install
 	$(INSTALL_DIR) $(1)/usr/lib
-	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libprocps.so* $(1)/usr/lib/
+	$(CP) $(PKG_INSTALL_DIR)/usr/lib/libproc2.so* $(1)/usr/lib/
 endef
 
 define Build/InstallDev
--- a/utils/procps-ng/patches/100-no-tests-docs.patch
+++ b/utils/procps-ng/patches/100-no-tests-docs.patch
@@ -1,26 +1,26 @@
 --- a/Makefile.am
 +++ b/Makefile.am
-@@ -14,10 +14,7 @@ PACKAGE_VERSION = @PACKAGE_VERSION@
+@@ -18,10 +18,7 @@ PACKAGE_VERSION = @PACKAGE_VERSION@
  
  ACLOCAL_AMFLAGS = -I m4
  SUBDIRS = \
--	include \
--	man-po \
+-	local \
+-	po-man \
 -	po \
 -	testsuite
-+	include
++	local
  
- AM_CFLAGS = -Iproc
- LDADD = ./proc/libprocps.la $(CYGWINFLAGS)
+ AM_CFLAGS = -Ilibrary/include
+ LDADD = ./library/libproc2.la $(CYGWINFLAGS)
 --- a/configure.ac
 +++ b/configure.ac
-@@ -298,8 +298,5 @@ AC_CHECK_FUNCS([__fpending alarm atexit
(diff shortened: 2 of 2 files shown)


feckert:
@neheb @krant
I m not happy with this change howt this is done!!!
In my opinion, this library can be installed in parallel.
I have a downstream project that uses this library.
The update of the library from version 3.x.x to 4.x.x has changed the API.
My program does not compile anymore, because of this change.
The `proc` handling is different!

```
/home/bbworker/bbworker/owrt_master_x86_64/System6/build/openwrt/build_dir/target-x86_64_musl/rpcd-mod-rsu/rsu.c:9:10: fatal error: proc/readproc.h: No such file or directory
    9 | #include <proc/readproc.h>
      |          ^~~~~~~~~~~~~~~~~
````
 I would suggest that we add the library `procps-ng2` ?

Version 3.3.16 of procps-ng:
```
ipkg-install
└── usr
    ├── bin
    │   ├── free
    │   ├── kill
    │   ├── pgrep
    │   ├── pidof
    │   ├── pkill
    │   ├── pmap
    │   ├── ps
    │   ├── pwdx
    │   ├── skill
    │   ├── slabtop
    │   ├── snice
    │   ├── tload
    │   ├── top
    │   ├── uptime
    │   ├── vmstat
    │   ├── w
    │   └── watch
    ├── include
    │   └── proc
    │       ├── alloc.h
    │       ├── devname.h
    │       ├── escape.h
    │       ├── numa.h
    │       ├── procps.h
    │       ├── pwcache.h
    │       ├── readproc.h
    │       ├── sig.h
    │       ├── slab.h
    │       ├── sysinfo.h
    │       ├── version.h
    │       ├── wchan.h
    │       └── whattime.h
    ├── lib
    │   ├── libprocps.a
    │   ├── libprocps.la
    │   ├── libprocps.so -> libprocps.so.8.0.2
    │   ├── libprocps.so.8 -> libprocps.so.8.0.2
    │   ├── libprocps.so.8.0.2
    │   └── pkgconfig
    │       └── libprocps.pc

```
Version 4.0.4 of procps-ng:
```
ipkg-install/
└── usr
    ├── bin
    │   ├── free
    │   ├── kill
    │   ├── pgrep
    │   ├── pidof
    │   ├── pidwait
    │   ├── pkill
    │   ├── pmap
    │   ├── ps
    │   ├── pwdx
    │   ├── skill
    │   ├── slabtop
    │   ├── snice
    │   ├── tload
    │   ├── top
    │   ├── uptime
    │   ├── vmstat
    │   ├── w
    │   └── watch
    ├── include
    │   └── libproc2
    │       ├── diskstats.h
    │       ├── meminfo.h
    │       ├── misc.h
    │       ├── pids.h
    │       ├── slabinfo.h
    │       ├── stat.h
    │       ├── vmstat.h
    │       └── xtra-procps-debug.h
    ├── lib
    │   ├── libproc2.a
    │   ├── libproc2.la
    │   ├── libproc2.so -> libproc2.so.0.0.2
    │   ├── libproc2.so.0 -> libproc2.so.0.0.2
    │   ├── libproc2.so.0.0.2
    │   └── pkgconfig
    │       └── libproc2.pc

```

krant:
Hi @feckert, sorry for the inconvenience. I'm afraid procps-ng2 would cause more confusion, what about `procps-compat` name for the legacy version? It is [the same approach](https://pkgs.alpinelinux.org/packages?name=*procps*&branch=edge&repo=&arch=x86_64&maintainer=) used by Alpine, will it work for you?

feckert:
@krant Thank you for your feedback.
If we rename the package `procps-ng` for version 3.x.x to `procps-compat`, then I thing we get problems with the dependencies in downstream projects .

In my case, for example, I have a feed for my external software where I maintain the source code for my package that uses `procps-ng` for version 3.x.x. I compile this software in this feed for different openwrt branches. If we now rename the package for version 3.x.x to `procps-compat`, then I have a problem with other branches. Because in the other branches the package dependency is still needed set to `procps-ng` and not `procps-compat` in the packages feed.

I think we should/must give the `procps-ng` with version 4.x.x the new package name. I don't care what name it is. The name `procps-ng2` was only a suggestion. As I have seen in the source of `procps-ng` this is already the third incompatible API change.

What about `procps-ng3` for version 3.x.x and `procps-ng4` for version 4.x.x and `procps-ng3` provides also the old name `procps-ng` with the Makefile feature [PROVDIDES](https://github.com/search?q=repo%3Aopenwrt%2Fpackages+PROVIDES%3A%3D&type=code). That should all be taken into consideration.

krant:
What you're suggesting is basically reserve `procps-ng` for 3.x version forever. It is not the approach in other distros, but OpenWrt has quite unique peculiarities which I'm not that familiar with. Let's wait for @neheb to chime in and show us the way :slightly_smiling_face:

feckert:
@krant 

> What you're suggesting is basically reserve `procps-ng` for 3.x version forever.

Good point. I hadn't even thought about that.

I have to revert my suggestion. I think we should rename the `procps-ng` package of version 3.x.x `procps-compat` as you suggest and the `procps-ng` package is building the version 4.x.x. In the other packages feed branches we could then add the variable PROVIDES with `procps-combat`. Then my downstream project could add the dependency `procps-compat`. This would also solve my dependency issue.

krant:
@feckert
> In the other packages feed branches we could then add the variable PROVIDES with procps-combat

 that's an excellent idea, thank you for comprehension! Let me arrange corresponding PRs..

dcousens:
I think `procps-ng3` seems more reasonable than `procps-compat`, with `procps-ng` referring to the newest version

PR #23233, opened 2024-02-01
libs/vips: fix PKG_CPE_ID

ffontaine:
libvips:libvips is a better CPE ID than vips:vips as this CPE ID has the latest CVEs (whereas vips only has an old CVE from 2010):

  https://nvd.nist.gov/products/cpe/search/results?keyword=cpe:2.3:a:libvips:libvips

Fix: 299e5b0a9bce19d6e96cb9ff217028b36ee2dd36

Maintainer: @flyn-org
Compile tested: Not needed
Run tested: Not needed


flyn-org:
Looks good. Recommend merge.

Issue #6748, opened 2018-08-13
[RFC] Decision on preference (or not) for the use of codeload


cshoredaniel:
@neheb Has submitted a number of PR's (referenced in https://github.com/openwrt/packages/issues/6584), full list of his/her PR's (which probably include the codeload thing for any github pulls): https://github.com/openwrt/packages/pulls/neheb

@diizzyy has also mentioned codeload. [EDIT: I misunderstood karlp's comment to be related, it's not] @karlp appears to feel the change is for marginal gain (based on https://github.com/openwrt/packages/issues/6584#issuecomment-410452508).

I personally think that tree-wide changes, especially for something @yousong has been trying solve another way ought t be clearly discussed and not just brought in through the back door, hence this ticket and I'll be following up on openwrt-devel as well.

neheb:
Just to clarify, I do it for GitHub projects that lack a release tarball.

Andy2244:
I also only use it for medium, large sized github packages, without release tarballs. see #6577

PS: I'm fine either way, the git clone + local tarball creation also worked fine for me in the past.

karlp:
Daniel Dickinson <notifications@github.com> wrote:
> @neheb Has submitted a number of PR's (referenced in
> https://github.com/openwrt/packages/issues/6584), full list of
> his/her PR's (which probably include the codeload thing for any
> github pulls): https://github.com/openwrt/packages/pulls/neheb
> 
> @diizzyy has also mentioned codeload. @karlp appears to feel
> the change is for marginal gain (based on
> https://github.com/openwrt/packages/issues/6584#issuecomment-410452508).

That's out of context.  I have no objection to the use of a documented github apis.  

Cheers,
Karl P

cshoredaniel:
@karlp apologies then, I misunderstood (the possibility I had is why I tagged you so you could comment)

cshoredaniel:
Please note it is preferred if the responses could be made as followups to http://lists.infradead.org/pipermail/openwrt-devel/2018-August/013570.html (followups to those followups) on [OpenWrt-devel Mailing List](https://lists.openwrt.org/mailman/listinfo/openwrt-devel)

diizzyy:
@cshoredaniel 
I'm not sure why you insist on having it on the mailing list since it's not related to OpenWrt distribution itself. GitHub has always been where package related discussions occur so it's the preferred forum.

Anyhow, a few advantages

* Reduces data usage considerably in a lot of cases compared to git clone
[...]

PR #23238, opened 2024-02-01
libstrophe: update 0.13.0

krant:
Maintainer: @changeway 

Description:
- Use common Github URL
- Add missing zlib dependency

### DIFF
--- a/libs/libstrophe/Makefile
+++ b/libs/libstrophe/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libstrophe
-PKG_VERSION:=0.10.0
+PKG_VERSION:=0.13.0
 PKG_RELEASE:=1
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://codeload.github.com/strophe/libstrophe/tar.gz/$(PKG_VERSION)?
-PKG_HASH:=8d804e4c74cea1133203cc95a59a88f700fbdaead076e7959b495d734dd7936d
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=https://github.com/strophe/libstrophe/releases/download/$(PKG_VERSION)
+PKG_HASH:=8d7c17c694efff7d2e2d3632684537e7ba874ce3d7f0e750a539f2299021bbaa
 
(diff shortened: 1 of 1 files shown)


BKPepe:
> - Use common Github URL

Whats the reason to switch from codeload to GitHub?

libs/libstrophe/Makefile
```diff
@@ -8,12 +8,12 @@
...
 
 PKG_NAME:=libstrophe
-PKG_VERSION:=0.10.0
+PKG_VERSION:=0.13.0
 PKG_RELEASE:=1
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://codeload.github.com/strophe/libstrophe/tar.gz/$(PKG_VERSION)?
-PKG_HASH:=8d804e4c74cea1133203cc95a59a88f700fbdaead076e7959b495d734dd7936d
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
```
BKPepe, line 14: What is the size difference? There isnt anything about this in the commit description.

neheb:
@BKPepe codeload and such need to be deprecated.

neheb:
codeload is less reproducible than custom tarballs.

Anyway in this case an official tarball is used. This is preferable to custom ones.

BKPepe:
Based on your observations or do you have any statistics, which can be seen here? 🤔

krant:
I guess it's related to this infamous case - https://news.ycombinator.com/item?id=34586917

neheb:
@BKPepe https://github.com/openwrt/packages/issues/6748#issuecomment-475410881

the whole thread actually...

BKPepe:
4 years old, cmon. Thats not an argument at all. Codeload is using a lot of packages these days here, so no progress there. No statistics. We would observe more issues if it would be happening recently, which is not and if it would be such bad, then we would switch to official tarballs sooner.

So, do we have some recent statistics/feedback or based on the history, we thinks it is bad and it stays bad and that Microsoft did not improve it?

krant:
> 4 years old, cmon

HN link I provided is one year old. Also it gives a technical explanation why codeload and such are less reproducible (compared to custom uploaded tarballs) - they are generated dynamically on-demand from git tag (with some caching behind the scenes). So when GitHub changes compression algorithm - tag-generated tarballs are changed too. Real files (like those in `github.com/strophe/libstrophe/releases/download/`) give better guarantees.

PR #23243, opened 2024-02-01
libs/libpam: fix PKG_CPE_ID

ffontaine:
linux-pam:linux-pam is a better CPE ID than kernel:linux-pam as this CPE ID has the latest CVEs (whereas kernel:linux-pam only has a SUSE-specific CVE): https://nvd.nist.gov/products/cpe/search/results?keyword=cpe:2.3:a:linux-pam:linux-pam

Fix: 6f74b0c4f15a095b1069a8aaeb19a32dfbc7539a

Maintainer: @nmav
Compile tested: Not needed
Run tested: Not needed


nmav:
Thank you!

PR #23246, opened 2024-02-01
net/krb5: fix PKG_CPE_ID

ffontaine:
mit:kerberos_5 is a better CPE ID than mit:kerberos as this CPE ID has the latest CVEs (whereas mit:kerberos only has CVEs until 2018): https://nvd.nist.gov/products/cpe/search/results?keyword=cpe:2.3:a:mit:kerberos_5

Fix: 299e5b0a9bce19d6e96cb9ff217028b36ee2dd36

Maintainer: @flyn-org
Compile tested: Not needed
Run tested: Not needed


flyn-org:
Looks good. Recommend merge.

PR #23249, opened 2024-02-02
jq: update to 1.7.1

krant:
Maintainer: @ratkaj 

Description:
- Update package URLs
- Fix license
- Drop obsolete CFLAGS
- Drop obsolete patches


BKPepe:
Obsolete patches, does it mean that you are removing patches, which were backported and they are now included in 1.7.1, right? Then it is not obsolete, though.

krant:
I'm sure it's quite common interpretation - https://github.com/search?q=repo%3Aopenwrt%2Fpackages+obsolete+patch&type=commits

PR #23251, opened 2024-02-02
openldap: update to 2.6.7

krant:
Maintainer: @flyn-org 

Description:
- Remove dead mirror URL
- Clean-up configure args

### DIFF
--- a/libs/openldap/Makefile
+++ b/libs/openldap/Makefile
@@ -8,14 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openldap
-PKG_VERSION:=2.6.6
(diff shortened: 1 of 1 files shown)


libs/openldap/Makefile
```diff
@@ -95,14 +94,12 @@ TARGET_CFLAGS += $(FPIC) -lpthread \
 	-DURANDOM_DEVICE=\\\"/dev/urandom\\\"
 
 CONFIGURE_ARGS += \
-	--enable-shared \
-	--enable-static \
```
BKPepe, line 99: This is removed, because is it default? You know, I am trying to find out the reason for "clean-up" configure args.

PR #23252, opened 2024-02-02
crowdsec: new upstream release version 1.6.0

ne20002:
Update crowdsec to latest upstream release version 1.6.0

Signed-off-by: S. Brusch <ne20002@gmx.ch>

Maintainer: Kerma Gérald <gandalf@gk2.net>
Package tested: not able to test run due to limited space (package is big)

Description: update to latest version of upstream


BKPepe:
Since, this can not be updated on your router, why we should merge this or even why there is a need to have the latest version, which you can not run at all?

ne20002:
> Since, this can not be updated on your router, why we should merge this or even why there is a need to have the latest version, which you can not run at all?
Actually I only run the bouncer on my router. But there are other users running the crowdsec package. I probably need to figure out how to exclude build for devices not big enough to install the package. The BPi-R3 in the default layout does not allow installation but other routers (x64) do.

PR #23256, opened 2024-02-02
lang/python/python-requests: fix PKG_CPE_ID

ffontaine:
There is not a single CVE linked to python-requests:requests so use python:requests instead:
https://nvd.nist.gov/products/cpe/search/results?keyword=cpe:2.3:a:python:requests

Fixes: ceadbcbb64de727c3a974e552d9a723d532e4e40 (treewide: add PKG_CPE_ID for cvescanner)

Maintainer: @BKPepe 
Compile tested: Not needed
Run tested: Not needed


BKPepe:
Btw, Fix tag is not correct.

It should be ``Fixes: hash ("subject")``.

BKPepe (approved):

PR #23262, opened 2024-02-02
utils/squashfs-tools: fix PKG_CPE_ID

ffontaine:
There is not a single CVE linked to phillip_lougher:squashfs so use squashfs-tools_project:squashfs-tools instead:
https://nvd.nist.gov/products/cpe/search/results?keyword=cpe:2.3:a:squashfs-tools_project:squashfs-tools

Fixes: 299e5b0a9bce19d6e96cb9ff217028b36ee2dd36 (treewide: add PKG_CPE_ID for better cvescanner coverage)

Maintainer: @commodo
Compile tested: Not needed
Run tested: Not needed


commodo:
thanks

i wonder what is it with all these CPE ID changes

PR #23273, opened 2024-02-03
cloudflared: refine config.yml

stokito:
The config.yml is an example of a tunnel local configuration. But the cloudlfared treat it as a real config and fails to start. So to avoid problems let's comment all the statements.

The `url: http://localhost:8000` is not a valid config option.

Additionally add a smale of configuring ingres rules.

The cloudflared.config has missing option token.

Maintainer: @1715173329 
Compile tested: not needed
Run tested: VBox

### DIFF
--- a/net/cloudflared/files/cloudflared.config
+++ b/net/cloudflared/files/cloudflared.config
@@ -1,6 +1,7 @@
 
 config cloudflared 'config'
 	option enabled '0'
+	option token ''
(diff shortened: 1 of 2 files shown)


1715173329:
Merged, thank you!

BKPepe:
This is missing ``PKG_RELEASE`` bump.

stokito:
Can you add the commit yourself?

PR #23292, opened 2024-02-05
subversion: update to 1.14.3

krant:
Maintainer: me

Description:
- Adopt the package
- Remove default configure options
- Rebase the patch


krant:
@val-kulkov the package has been adopted

val-kulkov:
Looks good to me, thank you.

PR #23295, opened 2024-02-05
snort3: improve script reliability

efahl:
- Enable missing variable checking by default
 - Explicitly check variables are defined in all 'rm' commands

Compile tested: none, all scripts
Run tested: snapshot x86

Description:
General code improvement and runtime checks.

### DIFF
--- a/net/snort3/Makefile
+++ b/net/snort3/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=snort3
 PKG_VERSION:=3.1.78.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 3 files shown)


flyn-org:
I really appreciate these types of changes. Thank you!

Do you have a guide that is leading to these Bourne shell idioms, or is this from experience?

I am curious why you favor explicitly resetting shell variables rather than using unset?

flyn-org:
Thanks for sharing the guide!

I think the unset stuff is a toss up. The trick is to remember there is a difference between a missing (unset) variable and an existing variable with an empty value. Whichever approach highlights this is probably best.

Recommend merge.

efahl:
> an existing variable with an empty value

That's the case here, the values are used to populate the report table row and must exist (albeit possibly empty).  I'll leave it as-is.

PR #23328, opened 2024-02-07
Update Tailscale version to 1.58.2

janknieling:
Updated tailscale to version 1.58.2

Maintainer: @ja-pa  @mochaaP @BKPepe
Compile tested: aarch64_cortex-a53 @ SNAPSHOT
Run tested: aarch64_cortex-a53 @ SNAPSHOT

Description:
I have updated the tailscale package to the latest stable release v. 1.58.2

### DIFF
--- a/net/tailscale/Makefile
+++ b/net/tailscale/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tailscale
-PKG_VERSION:=1.24.2
+PKG_VERSION:=1.58.2
 PKG_RELEASE:=$(AUTORELEASE)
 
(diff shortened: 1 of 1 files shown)


mochaaP:
Not sure if our golang toolchain on 21.02 could build tailscale. Needs testing

janknieling:
There is a tailscale package for 21.02 already available (v.1.24.2) but its very outdated version with security issues.

BKPepe:
I am not the maintainer of tailscale as said by @mochaaP, this one is not going to happen, because the latest tailscale version requires newer version of Golang. That is not going to be upgraded, because OpenWrt 21.02 is End of Life branch.

Also, we can not accept PRs, which was run and compile tested for snapshots and your target is the old branch, which is not supported anymore.

PR #23334, opened 2024-02-08
glib2: update to 2.78.4

krant:
Maintainer: @tripolar 

Description:
- Use HTTPS for package URL
- Don't set default Meson options
- Remove upstreamed patch
- Refresh remaining patches


neheb:
```
../../../builder/build_dir/target-aarch64_generic_musl/glib-2.79.1/meson.build:2383:26: ERROR: /builder/staging_dir/host/bin/python3 is missing modules: packaging
```

hmmm?

neheb:
not a fan of updating to a beta version. stable only please (even versions).

krant:
Ha, that's why it was failing! Thanks for pointing it out, PR has been updated.

neheb:
@krant 

vala/host is failing

```
/home/mangix/devstuff/openwrt/build_dir/hostpkg/glib-2.78.4/openwrt-build/../glib/gregex.c:1514: undefined reference to `pcre2_substring_number_from_name_8'
```

krant:
@neheb I've tried to rollback glib to 2.74.0 and build vala/host with refreshed patch form #23361 - it fails the same :confused:

neheb:
Yeah I used a different solution.

BKPepe:
Hmm, it seems that this was not run tested completely. It causes multiple issues, though!
E.g. https://github.com/openwrt/packages/issues/23395

feckert:
@BKPepe I would suggest we revert the version update of glib2, then we can take a closer look at the issue?

BKPepe:
Yup, I agree. Thanks!

Dne čt 15. 2. 2024 15:09 uživatel Florian Eckert ***@***.***>
napsal:

feckert:
@neheb I tried to revert the commit of glib2 update. I have seen that you did another [change](https://github.com/openwrt/packages/pull/23362). I would also revert this commit. So we have a working system again. We should after this reverts, have a closer look what the issue is with ModemManager and the new glib2 update.

PR #23345, opened 2024-02-10
libextractor: update to 1.13

krant:
Maintainer: @dangowrt 

Description:
- Remove obsolete configure option
- Remove patch and var override since MEM_SRCDST_SUPPORTED is always on


dangowrt:
Thank you!

PR #23357, opened 2024-02-11
libvpx: update to 1.14.0

neheb:
Fixes CVE-2023-5217

Maintainer: @luizluca


luizluca:
LGTM. Thanks

luizluca:
The test build seems to be busy. Maybe we should wait until it finishes before merging.

PR #23362, opened 2024-02-11
glib2: use internal pcre2

neheb:
Maintainer: @flyn-org


neheb:
ping @eli-schwartz

that pcre2 handling looks extremely funky.

PR #23381, opened 2024-02-12
tio: update to 2.7

krant:
Maintainer: N/A

Description:
- Switch to Meson build
- Fix license file name
- Add libinih dependency

### DIFF
--- a/utils/tio/Makefile
+++ b/utils/tio/Makefile
@@ -8,27 +8,27 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tio
-PKG_VERSION:=1.32
-PKG_RELEASE:=2
+PKG_VERSION:=2.7
+PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


utils/tio/Makefile
```diff
@@ -37,7 +37,7 @@ endef
 
 define Package/tio/install
 	$(INSTALL_DIR) $(1)/usr/sbin
-	$(INSTALL_BIN) $(PKG_BUILD_DIR)/src/tio $(1)/usr/sbin/
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/openwrt-build/src/tio $(1)/usr/sbin/
```
neheb, line 40: uhhh this doesn't go in PKG_INSTALL_DIR?

krant: PKG_INSTALL_DIR has the same-size binary. Should I use it instead?

neheb: Yes

PR #23388, opened 2024-02-12
modemmanager: add track state handling

feckert:
Maintainer: me / @
Compile tested: x86_64, APU3, latest OpenWrt master
Run tested:  x86_64, APU3, latest OpenWrt master

Description:

modemmanager: refactor bearermethod static:
To improve code readability, the required variables for a static IP configuration are now  being extracted from the bearer status object within the method.

modemmanager: add netifd connection tracking
Similar to ppp, add an active process which tracks an established data connection. The process exits if the modem was removed or in case connection has been lost and can immediately be re-established. A [merge request](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/merge_requests/1131) is already pending in the ModemManager upstream repository for the source code changes.

modemmanager: integrate new conncection tracking into proto
The ModemManager currently doesn't reconnect in case the connection was lost and network access was regained at a later point. To solve this problem, the proto modemmanager tracks the connection state with the new 'mmcli' 'track-state' option, similar to how we do this for a ppp connections.

### DIFF
--- a/net/modemmanager/files/modemmanager.proto
+++ b/net/modemmanager/files/modemmanager.proto
@@ -124,14 +124,13 @@ modemmanager_disconnected_method_ppp_ipv4() {
 				;;
 		esac
 	} || echo "pppd result code not given"
-
-	proto_kill_command "$interface"
 }
 
 modemmanager_connected_method_dhcp_ipv4() {
 	local interface="$1"
 	local wwan="$2"
 	local metric="$3"
+	local device="$4"
 
 	proto_init_update "${wwan}" 1
 	proto_set_keep 1
@@ -145,6 +144,8 @@ modemmanager_connected_method_dhcp_ipv4() {
 	[ -n "$metric" ] && json_add_int metric "${metric}"
 	json_close_object
 	ubus call network add_dynamic "$(json_dump)"
+
+	modemmanager_start_watcher "${interface}" "${device}"
 }
 
 modemmanager_connected_method_static_ipv4() {
@@ -152,6 +153,7 @@ modemmanager_connected_method_static_ipv4() {
 	local wwan="$2"
 	local bearerstatus="$3"
 	local metric="$4"
+	local device="$5"
 
 	local address prefix gateway mtu dns1 dns2
 	local mask=""
@@ -194,12 +196,15 @@ modemmanager_connected_method_static_ipv4() {
 	}
 	[ -n "$metric" ] && json_add_int metric "${metric}"
 	proto_send_update "${interface}"
+
+	modemmanager_start_watcher "${interface}" "${device}"
 }
 
 modemmanager_connected_method_dhcp_ipv6() {
 	local interface="$1"
 	local wwan="$2"
 	local metric="$3"
+	local device="$4"
 
 	proto_init_update "${wwan}" 1
 	proto_set_keep 1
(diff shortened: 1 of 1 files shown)


lynxthecat:
Referencing: https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/merge_requests/1131

>> So if I read the code right... track-state only does something if the modem is already CONNECTED. If the modem changes state from CONNECTED to anything other than REGISTERED, mmcli continues tracking state. If the modem changes to REGISTERED and the previous state was ENABLED/SEARCHING/CONNECTED then mmcli exits. Otherwise it continues tracking state. Thus mmcli only exits if the modem was (a) removed or (b) now disconnected but still registered with the network.
> 
> Correct
> 
>> I guess I'm wondering what happens in interim states, like if the modem gets rfkilled or locked or manually disconnected. I think the mmcli would just wait forever then?
> 
> This depends on whether the ModemManager generates a change event for such events. Unfortunately, I don't know enough about how the ModemManager behaves.

So in many cases the ISP will manually disconnect say every 48 hours. If I understand correctly, this will get intercepted as state 'disconnected' by modemmanager from 1.18.8 given the dispatcher script functionality introduced with this commit: https://github.com/openwrt/packages/commit/bc754f31cfdb004eefa43038f8f0827922107fc6.

But will this new state tracking and auto-reconnect logic actually deal with this situation? That is, state shifts to 'disconnected', and we need to tell ModemManager to reconnect, else network connectivity will remain lost pending manual intervention.

I wonder if this logic has been properly thought through with the existing dispatcher script logic that can be relied upon for automatic reconnections like so:

```
root@OpenWrt:~# cat /usr/lib/ModemManager/connection.d/10-report-down-and-reconnect
#!/bin/sh

# Automatically report to netifd that the underlying modem
# is really disconnected and reconnect if interface was up

# require program name and at least 4 arguments
[ $# -lt 4 ] && exit 1

MODEM_PATH="$1"
BEARER_PATH="$2"
INTERFACE="$3"
STATE="$4"

[ "${STATE}" = "disconnected" ] || exit 0

. /usr/share/ModemManager/modemmanager.common
. /lib/netifd/netifd-proto.sh
INCLUDE_ONLY=1 . /lib/netifd/proto/modemmanager.sh

MODEM_STATUS=$(mmcli --modem="${MODEM_PATH}" --output-keyvalue)
[ -n "${MODEM_STATUS}" ] || exit 1

MODEM_DEVICE=$(modemmanager_get_field "${MODEM_STATUS}" "modem.generic.device")
[ -n "${MODEM_DEVICE}" ] || exit 2

CFG=$(mm_get_modem_config "${MODEM_DEVICE}")
[ -n "${CFG}" ] || exit 3

IFUP=$(ifstatus ${CFG} | jsonfilter -e '@.up')

logger -t "modemmanager" "interface ${CFG} (network device ${INTERFACE}) ${STATE}"
proto_init_update $INTERFACE 0
proto_send_update $CFG

[ "${IFUP}" = "true" ] && ifup ${CFG}

exit 0
```

dangowrt:
> [...] albeit it does not deal with the cell tower going offline then coming back sometime later. I wonder if the logic proposed in this merge request will deal with either of these situations?

Yes, I've designed the `--track-state` patch for ModemManager with exactly that (mobile) case in mind.

feckert:
Related to the discussion on the ModemManager merge request https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/merge_requests/1131

I have worked out another solution that does not require any changes in the ModemManager source and, according to @aleksander0m , should also work. Which is the case from my point of view.

@lynxthecat @dangowrt @osedlbauer Perhaps all those involved  in this reconnect can please take another look at this issue whether this procedure also solves the problem.
https://github.com/openwrt/packages/pull/23590

PR #23396, opened 2024-02-13
imagemagick: update to 7.1.1-28

krant:
Maintainer: me

Description:
- Use .xz for PKG_SOURCE
~- Re-enable mips16~

### DIFF
--- a/multimedia/imagemagick/Makefile
+++ b/multimedia/imagemagick/Makefile
@@ -7,13 +7,13 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=imagemagick
 PKG_VERSION:=7.1.1
-PKG_REVISION:=27
+PKG_REVISION:=28
 PKG_RELEASE:=1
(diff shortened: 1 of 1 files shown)


neheb:
why reenable mips16?

krant:
Just to check with CI if it's work. It was disabled for ages, may be new toolchain handles it all right. I have no MIPS experience, but if mips16 gives smaller binaries - it's a good thing, right?

neheb:
In the case of imagemagick, it might be for performance reasons. Maybe @val-kulkov knows more.

val-kulkov:
I do not recall why I disabled mips16, it was many years ago. But I suspect that because many other packages apparently had problems compiling with mips16 enabled, I decided not to bother finding the source of the problem and simply disabled mips16. See the output of "grep -r MIPS16 package/" in the build root, it is quite telling.

krant:
It seems proper way would be to measure the performance of imagemagick on a real hardware with/without MIPS16 and make conscious adjustments afterwards. I have no MIPS device at hand atm, so if anyone would like to step in - you're welcome!

Issue #22451, opened 2023-10-19
syslog-ng: conflicts with logd package


myszsoda:
Maintainer: @BKPepe 
Environment: aarch64, snapshot on 19.10.2023; but issue applies to all versions

Description:
There is a problem with syslog-ng package which has its own logread script to read logs. But this package does conflict with core package logd which has its own logread and it is possible to select both of them.
Thus, I propose to either add CONFLICTS between syslog-ng and logd package or implement PROVIDES for both of them like https://github.com/myszsoda/packages/commit/f680a11fafb1b08f63f96eb5fa2030d12bf7f100

I can implement this, but would like to ask opinion which would be preferred.

BKPepe:
Hi,

Hmm, it might works, but that's not a solution. I will look into it.

Dne čt 19. 10. 2023 17:07 uživatel Jakub Raczynski ***@***.***>
napsal:

Rondom:
This issue seems to be a duplicate of #11535

I have submitted a PR here. https://github.com/openwrt/packages/pull/23401 It might be more a workaround than solution, but it is better than the current state of things.

stokito:
I added a fix for this to my older PR for the syslog-ng https://github.com/openwrt/packages/pull/22763
Please check it.

Rondom:
Thanks

Issue #11535, opened 2020-03-09
syslog-ng: clash with ubox logd when both are required 


muddyfeet:
Maintainer: @BKPepe 
Environment: Linksys WRT1200ac trunk

Commit fe165eeb94d2e7fb35fee93e64336afa024a1283 prevents both ubox logd and syslog-ng coexisting in the same builds. Previously, I was using the vanilla openwrt logd to forward events to syslog-ng at 127.0.0.1 which allowed both to run on the one device. This allowed you to display the syslog in Luci and also take advantage of the extra capability of syslog-ng (I note that this change allows you to achieve this with just syslog-ng now). 

Happy to delete the ubox logd from my builds and rework the logging if that is required - but I'm not sure this was intended. The make -j1 V=s output is shown below:

```
Collected errors:
* check_data_file_clashes: Package syslog-ng wants to install file /home/muddyfeet/openwrt-trunk/build_dir/target-arm_cortex-a9+vfpv3_musl_eabi/root-mvebu/sbin/logread
But that file is already provided by package * logd
* opkg_install_cmd: Cannot install package syslog-ng.
package/Makefile:65: recipe for target 'package/install' failed
make[2]: *** [package/install] Error 255
make[2]: Leaving directory '/home/muddyfeet/openwrt-trunk'
package/Makefile:108: recipe for target '/home/muddyfeet/openwrt-trunk/staging_dir/target-arm_cortex-a9+vfpv3_musl_eabi/stamp/.package_install' failed
make[1]: *** [/home/muddyfeet/openwrt-trunk/staging_dir/target-arm_cortex-a9+vfpv3_musl_eabi/stamp/.package_install] Error 2
make[1]: Leaving directory '/home/muddyfeet/openwrt-trunk'
/home/muddyfeet/openwrt-trunk/include/toplevel.mk:225: recipe for target 'world' failed
make: *** [world] Error 2



```

muddyfeet:
Further information:

Building the firmware without logd and only syslog-ng completes successfully. 

I can run only syslog-ng to collect the logs but it presents a new problem. My syslog-ng files are being created on an attached USB disk with a new file created per day (e.g. /mnt/sda1/logs/syslog-2020-03-11.log). However, logread is hard coded to use the RAM based syslog at /var/log/messages - so this prevents Luci from displaying the logs in my config. 

[...]

PR #23401, opened 2024-02-13
syslog-ng: Install logread to /usr/sbin

Rondom:
Maintainer: @BKPepe 
Compile tested: x86 / 64 / master
Run tested: x86 / 64 / master

Description:

Instead of into /sbin/ install logread into /usr/sbin/logread. This works around a conflict with the logd package.

Closes: https://github.com/openwrt/packages/issues/11535
Closes: https://github.com/openwrt/packages/issues/22451

### DIFF
--- a/admin/syslog-ng/Makefile
+++ b/admin/syslog-ng/Makefile
@@ -123,7 +123,7 @@ define Package/syslog-ng/install
 	touch $(1)/etc/syslog-ng.d/.keep
 
 	$(INSTALL_DIR) $(1)/sbin
-	$(INSTALL_BIN) ./files/logread $(1)/sbin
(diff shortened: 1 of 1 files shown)


BKPepe:
That could work, but there is a disadvantage. You can not have installed
two binaries, which do almost the same thing in different path. System
would be confused.

I know that you are trying to fix that, but this should be done properly
through alternatives.

Dne út 13. 2. 2024 15:14 uživatel Andreas Gnau ***@***.***>
napsal:

stokito:
@Rondom here is an example of specifying alternatives https://github.com/openwrt/packages/pull/22861/files

BKPepe:
As said above, this needs to be handled by alternatives.

PR #23402, opened 2024-02-13
procps-ng3: new compatilibty package for old version of procps-ng

krant:
- Only library .so file is installed (no utilities)
- Latest 3.3.17 version of 3.x series is used
- Refresh existing patch
- Add new patch from Alpine Linux

This is the result of discussion in #23230. If community agree with this PR, I will create corresponding PRs to stable releases with `PROVIDES:=procps-ng3` in `procps-ng` package.

Attaching @feckert @neheb @dcousens


feckert:
@krant Can you also add in a separate commit the update of procps-ng to API version 4?
So we have the following commits.
* Rename procps-ng to procps-ng3 without any changes only rename where needed
* Then make the changes to procps-ng3
    - Only library .so file is installed (no utilities because this are supported by procps-ng API version 4
    - Latest 3.3.17 version of 3.x series is used
    - Refresh existing patch
    - Add new patch from Alpine Linux
* Add new packages procps-ng with API version 4

If this gets merged into packages feed master, we could change the packages feed stable branches by adding PROVIDES

krant:
@feckert here you are. Hope I understood correctly what you were requesting.

feckert:
@krant LGTM.
Now the only question is why the CI/CD breaks here?
@neheb I do not think this is related to this change.

feckert:
@krant I have now taken a closer look. I would suggest this [commit](https://github.com/TDT-AG/packages/tree/pr/20240216-procps-ng) history it like this. This approach is better from my point of view. Then others who were not involved know what is being done and how it came about.

If you don't have an objections, I would close your pull request and merge mine instead. In any case, you will remain in the histroy as commit author.

Thanks for your work.

krant:
> Revert "procps-ng: update to 4.0.4"

agree, now it makes more sense to me too

> I would close your pull request and merge mine instead

Sure, go ahead

PR #23412, opened 2024-02-14
snort3: improve date filtering in report

efahl:
- Reduce memory footprint while running reports allowed by recent jsonfilter bug fix.  
- Improve headers and footers on report.  
- Add new match operator for date filtering.

Compile tested: NA
Run tested: x86 snapshot


graysky2:
Reviewing the code by eye seems OK.  Full disclosure: my snort setup is using the legacy snort init script with legacy `homenet.lua` `local.lua` etc. so I am not able to test the code in this PR.

@flyn-org thoughts?

efahl:
> using the legacy snort init script with legacy ... `local.lua`

I'm pretty sure the only thing reporting depends on is the json alert files, so if you were to stick the following snippet in your `local.lua` at the end, then run with that for a little bit, you would have some test inputs (assuming snort sees some events to log).  Oh, yeah, make sure `/etc/config/snort` has `option log_dir` set to point to wherever your logs actually reside...

```
alert_json = {
  file = true,
  fields = [[ timestamp pkt_num pkt_gen pkt_len proto dir src_addr src_port dst_addr dst_port gid sid rev action msg ]],
}
```

efahl:
@flyn-org , got a minute to look this over?

flyn-org:
Looks good. Recommend merge.

graysky2:
@efahl - can you rebase?  the 3.1.84 update got merged

efahl:
Rebased.

graysky2:
Please merge

efahl:
Rebased.

Anything else needed to get this merged?

efahl:
@neheb, this should be an easy one if you're looking for lingering merges to clean up...

PR #23432, opened 2024-02-15
zmq: update to 4.3.5

neheb:
Remove GCC13 backport.

Refresh other patches.

Maintainer: @srdgame


drujd:
```
Assertion failed: rc == 0 || zmq_errno () == ETERM (src/zsock_option.inc: zsock_set_curve_serverkey: 2383)
fish: Job 1, 'dynfw-nfset-client --table "f...' terminated by signal SIGABRT (Abort)
```
This is with GCC12! So maybe the patch was needed for GCC12 as well? Turris Omnia / mvebu

PR #23433, opened 2024-02-15
Add metanoia mt5311 vdsl modem tools

jasle:
Maintainer: me, @jasle 
Compile tested:  mediatek filogic MT7986 bpi-r3
Run tested: mediatek filogic MT7986 bpi-r3

Description:
Utilities for working with Metanoia/Proscend VDSL2 SFP Modems from @jimdigriz.
This Package only contain the parts needed for using it from shell or lua, not the snmp part, because it is currently mentioned as work in progress.

Also contains new package `lua-struct`, a dependency for the mt5311 utilities.

### DIFF
--- a/lang/lua-struct/Makefile
+++ b/lang/lua-struct/Makefile
@@ -1,0 +1,42 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lua-struct
+PKG_VERSION:=0.9.2
+PKG_RELEASE:=1
+
+PKG_VERSION_TAG:=$(PKG_VERSION)-1
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION_TAG).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/iryont/$(PKG_NAME)/tar.gz/$(PKG_VERSION_TAG)?
+PKG_HASH:=30d4e3584e27caa504745fdf5e191f2469ae284dc51271292654856905a4603d
+
+PKG_MAINTAINER:=jasle <jasle@riseup.net>
+PKG_LICENSE:=MIT/X11
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION_TAG)
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/lua-struct
+  SUBMENU:=Lua
+  SECTION:=lang
+  CATEGORY:=Languages
+  TITLE:=Implementation of binary packing/unpacking in pure lua
+  URL:=https://github.com/iryont/lua-struct
+endef
+
+define Package/lua-struct/desription
+	lua-struct is a pure lua Implementation for packing and unpacking 
+	binary data. 
+endef
+
+define Build/Compile
+	echo "Nothing to compile, pure lua package"
+endef
+
+define Package/lua-struct/install
+	$(INSTALL_DIR) $(1)/usr/lib/lua
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/src/struct.lua $(1)/usr/lib/lua/
+endef
+
+$(eval $(call BuildPackage,lua-struct))
+
--- a/utils/mt5311/Makefile
+++ b/utils/mt5311/Makefile
@@ -37,6 +37,7 @@ define Package/mt5311/install
 	$(INSTALL_DATA) $(PKG_BUILD_DIR)/ebm.lua $(1)/usr/lib/lua/mt5311/
 	$(INSTALL_DATA) $(PKG_BUILD_DIR)/register.lua $(1)/usr/lib/lua/mt5311/
 	$(INSTALL_DATA) $(PKG_BUILD_DIR)/register.map $(1)/usr/lib/lua/mt5311/
+	$(INSTALL_DATA) ./files/init.lua $(1)/usr/lib/lua/mt5311/
 
 	$(INSTALL_DIR) $(1)/usr/bin
 	$(INSTALL_BIN) ./files/ebm-read $(1)/usr/bin/
--- a/utils/mt5311/Makefile
+++ b/utils/mt5311/Makefile
@@ -1,0 +1,46 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=mt5311
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/jimdigriz/mt5311.git
+PKG_SOURCE_VERSION:=2cb13045cd82474416e7f454b36f3a07b0de49d7
+PKG_MIRROR_HASH:=a1e1e9354fe9b9d6aa5da6b38ff6ef84458761533f9dbc55758dd403868c42ed
+
+PKG_MAINTAINER:=jasle <jasle@riseup.net>
+PKG_LICENSE:=AGPL-3.0-or-later
+PKG_LICENSE_FILES:=LICENSE
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/mt5311
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Utilities for working with Metanoia/Proscend VDSL2 SFP Modems
+  URL:=https://github.com/jimdigriz/mt5311
+  DEPENDS:=+lua +luaposix +lua-struct
+endef
+
+define Package/mt5311/description
+	mt5311 is a tool to read status informationen from Metanoia/Proscend 
+	VDSL2/Proscend VDSL2 SFP Modems.
+endef
+
+define Build/Compile
+	echo "Nothing to compile, pure lua package"
+endef
+
+define Package/mt5311/install
+	$(INSTALL_DIR) $(1)/usr/lib/lua/mt5311
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/ebm-read.lua $(1)/usr/lib/lua/mt5311/
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/ebm.lua $(1)/usr/lib/lua/mt5311/
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/register.lua $(1)/usr/lib/lua/mt5311/
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/register.map $(1)/usr/lib/lua/mt5311/
+
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) ./files/ebm-read $(1)/usr/bin/
+endef
+
+$(eval $(call BuildPackage,mt5311))
+
--- a/utils/mt5311/files/ebm-read
+++ b/utils/mt5311/files/ebm-read
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+exec lua /usr/lib/lua/mt5311/ebm-read.lua "$@"
+
--- a/utils/mt5311/files/init.lua
+++ b/utils/mt5311/files/init.lua
@@ -1,0 +1,11 @@
+-- mt5311 init file, makes mt5311 importable as module 
+
+local dir = '/usr/lib/lua/mt5311/'
+local file = dir .. 'ebm.lua'
+arg={}
+arg[0] = file
+
+mt5311 = assert(loadfile(file))(arg)
+
+return mt5311
+


lang/lua-struct/Makefile
```diff
@@ -0,0 +1,32 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lua-struct
+PKG_VERSION:=0.9.2
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/iryont/lua-struct.git
+PKG_SOURCE_VERSION:=a86046034c28ed1a85bd4f2a877b903f232ea3d6
+PKG_MIRROR_HASH:=9d0deb73b6a42b0b85a8c271fea75862c7961ee468fad76700cf09e941ed934f
```
BKPepe, line 10: Why do we need to add package, which is archived and thus only read-only for about an year?

No.

lang/lua-struct/Makefile
```diff
@@ -0,0 +1,36 @@
...
+PKG_VERSION:=0.9.2
+PKG_RELEASE:=1
+
+PKG_LICENSE:=MIT/X11
+PKG_URL:=https://github.com/iryont/lua-struct
+
+PKG_MAINTAINER:=jasle <jasle@riseup.net>
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/iryont/$(PKG_NAME)/tar.gz/$(PKG_VERSION)?
```
BKPepe, line 13: Why do we need to add package, which is no longer developed and also the repository is archived for almost an year, now?

jasle: `lua-struct` is a dependency of `mt5311`, so it is needed in order to use the tools for the metanoia sfp vdsl modems, which are very useful when using such a modem.

It is just a very small package, in the end just one lua file with about 200 lines. So not much that could be further developed.

If desired we could switch to a fork of the repo for the package, but then it would base on a different source than the package in luarocks.

lang/lua-struct/Makefile
jasle: See other comment.

lang/lua-struct/Makefile
BKPepe: I understand your need of mt5311, but I dont see why we should fork the repo of package, which is dead, no longer developed and I dont think that you will take care of the forked package, anyway. 

You should raise the issue with mt5311, if they can find the replacement or what they will do with it.

lang/lua-struct/Makefile
```diff
@@ -0,0 +1,39 @@
...
+PKG_NAME:=lua-struct
+PKG_VERSION:=0.9.2
+PKG_RELEASE:=1
+
+PKG_LICENSE:=MIT/X11
+PKG_URL:=https://github.com/iryont/lua-struct
+
+PKG_MAINTAINER:=jasle <jasle@riseup.net>
+
+PKG_VERSION_TAG:=$(PKG_VERSION)-1
```
BKPepe, line 12: Why do you need to introduce a new variable? No.

jasle: The version `0.9.2` has a little bit wired version tag `0.9.2-1`. I wanted to avoid adding the suffix in multiple places, so I put the correct tag into a variable.

If it is preferred to at it at multiple places, i can change the Makefile.

lang/lua-struct/Makefile
jasle: I've raised the issue with mt5311.

There is no real alternative to using `lua-struct`, as a pure lua package without c bindings and with support for lua 5.1 is desired.

The other struct modules for lua are also kind of old and have no recent releases. Seems like implementing structs for lua is just is not a thing, that need continuously development at all.

Could include `lua-struct` directly into the repository of `mt5311`, but this would just be a dirty hack to avoid having it as external dependency and directly including dependencies into the repository isn't a good way at all for various reasons.

I think we can refrain from the _**should** be maintained_ for this small package here.

lang/lua-struct/Makefile
jasle: I stick with having the tag that is different from the version in a variable. We have variables, so why not use them and repeat things?

BKPepe: Please dont mark this conversation as resolved since it is not resolved.

BKPepe: So... when it has wired version tag, why dont put it exactly to the PKG_VERSION instead of doing new variable? Its just guess, it should work, because basically you are doing the same thing with the PKG_VERSION_TAG. Still my reaosn is that we dont need to have a new variable.

utils/mt5311/Makefile
```diff
@@ -0,0 +1,47 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=mt5311
+PKG_VERSION:=238-2cb1304
+PKG_RELEASE:=1
+
+PKG_LICENSES:=AGPL-3.0-or-later
+PKG_LICENSE_FILES:=LICENSE
+PKG_URL:=https://github.com/jimdigriz/mt5311
```
BKPepe, line 9: This is already at line 24. No need to have it twice.

lang/lua-struct/Makefile
```diff
@@ -0,0 +1,39 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lua-struct
+PKG_VERSION:=0.9.2
+PKG_RELEASE:=1
+
+PKG_LICENSE:=MIT/X11
+PKG_URL:=https://github.com/iryont/lua-struct
```
BKPepe, line 8: Already at line 26, so drop this one.

utils/mt5311/Makefile
```diff
@@ -0,0 +1,47 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=mt5311
+PKG_VERSION:=238-2cb1304
```
BKPepe, line 4: Since you are using sources from GitHub and checking this out, this is not necessary.

utils/mt5311/Makefile
```diff
@@ -0,0 +1,47 @@
...
+  URL:=https://github.com/jimdigriz/mt5311
+  DEPENDS:=+lua +luaposix +lua-struct
+endef
+
+define Build/Compile
+	echo "Nothing to compile, pure lua package"
+endef
+
+define Package/mt5311/install
+	# Install packages
```
BKPepe, line 33: This is clear, drop. :)

utils/mt5311/Makefile
```diff
@@ -0,0 +1,47 @@
...
+	$(INSTALL_DIR) $(1)/usr/lib/lua/mt5311
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/ebm-read.lua $(1)/usr/lib/lua/mt5311/
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/ebm.lua $(1)/usr/lib/lua/mt5311/
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/register.lua $(1)/usr/lib/lua/mt5311/
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/register.map $(1)/usr/lib/lua/mt5311/
+	$(INSTALL_DATA) ./files/init.lua $(1)/usr/lib/lua/mt5311/
+
+	#Install wrapper script
```
BKPepe, line 41: Drop, no need.

utils/mt5311/Makefile
```diff
@@ -0,0 +1,47 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/mt5311
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Utilities for working with Metanoia/Proscend VDSL2 SFP Modems
+  URL:=https://github.com/jimdigriz/mt5311
+  DEPENDS:=+lua +luaposix +lua-struct
+endef
```
BKPepe, line 26: We need also package description for this package.

lang/lua-struct/Makefile
```diff
@@ -0,0 +1,39 @@
...
+include $(INCLUDE_DIR)/package.mk
+
+define Package/lua-struct
+  SUBMENU:=Lua
+  SECTION:=lang
+  CATEGORY:=Languages
+  TITLE:=Implementation of binary packing/unpacking in pure lua
+  URL:=https://github.com/iryont/lua-struct
+endef
+
```
BKPepe, line 28: Package description would be great to have.

lang/lua-struct/Makefile
jasle: Its version `0.9.2` but the tag is `0.9.2-1`. No idea why, but that’s the way it is.

utils/mt5311/Makefile
jasle: fixed.

utils/mt5311/Makefile
jasle: fixed.

utils/mt5311/Makefile
jasle: Was used to add the number to commits before the short commit hash by other packaging, but seems unusual for OpenWrt packages.

Removed it.

lang/lua-struct/Makefile
jasle: fixed.

utils/mt5311/Makefile
```diff
@@ -0,0 +1,47 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=mt5311
+PKG_VERSION:=238-2cb1304
+PKG_RELEASE:=1
+
+PKG_LICENSES:=AGPL-3.0-or-later
```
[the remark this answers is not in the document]
jasle: Thanks for the hint, fixed.

utils/mt5311/Makefile
jasle: fixed.

utils/mt5311/files/ebm-read
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+lua /usr/lib/lua/mt5311/ebm-read.lua "$@"
```
jimdigriz, line 3: if you use instead `exec lua /usr/lib/lua/mt5311/ebm-read.lua "$@"` it should consume less RAM...mostly

lang/lua-struct/Makefile
```diff
@@ -0,0 +1,43 @@
...
+PKG_LICENSE:=MIT/X11
+
+PKG_MAINTAINER:=jasle <jasle@riseup.net>
+
+PKG_VERSION_TAG:=$(PKG_VERSION)-1
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION_TAG).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/iryont/$(PKG_NAME)/tar.gz/$(PKG_VERSION_TAG)?
+PKG_HASH:=30d4e3584e27caa504745fdf5e191f2469ae284dc51271292654856905a4603d
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION_TAG)
```
BKPepe, line 16:
```suggestion
PKG_NAME:=lua-struct
PKG_VERSION:=0.9.2-1
PKG_RELEASE:=1

PKG_MAINTAINER:=jasle <jasle@riseup.net>
PKG_LICENSE:=MIT/X11

PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
PKG_SOURCE_URL:=https://codeload.github.com/iryont/$(PKG_NAME)/tar.gz/$(PKG_VERSION)?
PKG_HASH:=30d4e3584e27caa504745fdf5e191f2469ae284dc51271292654856905a4603d
```

utils/mt5311/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+PKG_LICENSE:=AGPL-3.0-or-later
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_MAINTAINER:=jasle <jasle@riseup.net>
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/jimdigriz/mt5311.git
+PKG_SOURCE_VERSION:=2cb13045cd82474416e7f454b36f3a07b0de49d7
+PKG_MIRROR_HASH:=77ae6423f9aaaa11c562823df285105d9069fad0ad6a0380b8ac52e2357f3a1f
```
BKPepe, line 14:
```suggestion
PKG_NAME:=mt5311
PKG_RELEASE:=1

PKG_SOURCE_PROTO:=git
PKG_SOURCE_URL:=https://github.com/jimdigriz/mt5311.git
PKG_SOURCE_VERSION:=2cb13045cd82474416e7f454b36f3a07b0de49d7
PKG_MIRROR_HASH:=77ae6423f9aaaa11c562823df285105d9069fad0ad6a0380b8ac52e2357f3a1f

PKG_MAINTAINER:=jasle <jasle@riseup.net>
PKG_LICENSE:=AGPL-3.0-or-later
PKG_LICENSE_FILES:=LICENSE
```

jasle: Resorted.

lang/lua-struct/Makefile
jasle: Resorted, but kept version.

PR #23443, opened 2024-02-16
openblas: enable ARM-specific optimizations

krant:
Maintainer: @commodo 
Run tested: arm_cortex-a9+neon, Linksys WRT3200ACM, master branch

### DIFF
--- a/libs/openblas/Makefile
+++ b/libs/openblas/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=OpenBLAS
 PKG_VERSION:=0.3.26
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


commodo:
PKG_RELEASE:=1 needs a bump

rest looks good

BKPepe:
Commit description is missing - why do we need to have this change, what was the reason, how this is improvement, etc.

krant:
@commodo nice catch, done

krant:
@BKPepe look at the summary - it's all there

commodo:
> @BKPepe look at the summary - it's all there

so a commit description is nice to have :)
as a general overview of that the change does;

OpenBLAS has per-processor (well, per-family processor) optimizations in assembly;
but they need to be enabled specifically;
i did not bother to do it for each family;
when people want them, they can enable them (like in this patchset)

krant:
@commodo thanks for being constructive. Commit has been updated.

BKPepe (approved):

PR #23457, opened 2024-02-16
procps-ng: update to version 4.0.4 and rename old version 3.3.16 to procps-ng3

feckert:
* Revert commit https://github.com/openwrt/packages/commit/81629ba5918f48a0886e6f601d63d0b016ef8c1e
* Rename procps-ng to procps-ng3 and only install library .so file (no utilities) for downstream packges that use procps-ng library feature.
   - Update procps-ng3 to the latest version 3.3.17 of 3.x series
   - Refresh existing patch
 * Readd procps-ng version 4.0.3 of 4.x series

This is the result of discussion in https://github.com/openwrt/packages/pull/23230. 
This pullrequest replaces pullrequest https://github.com/openwrt/packages/pull/23402


BKPepe:
Oh boy... another thing, which needs to be handled, which was caused that someone wants to have the latest version of everything without proper testing... :-/

krant:
@BKPepe your comments as usually are non-constructive, attacking someone personally and painfully wrong. This specific upgrade caused breakage of some unspecified out-of-tree software relying on development branch of this repo. How do you suppose we had to find this specific breakage without updating the master? You have no answer, since your common stance is to ignore the questions and intervene only to bully or to gatekeep someones work. This exact attitude has led to enormous number of outdated packages, some of which has not been updated for 2, 3 or even 5 years. If instead of rambling in each thread how busy you're you spend some time, you know, maintaining - OpenWrt packages had not been a such a disaster.

BKPepe:
Sorry that I prefer quality over quantity. I just wonder how you could
update more than 70 packages in one or maybe two weeks? 🤔 Anyone could do
that, but its your opinion. I could also blindly merge everything without
testing, but that leads to another failures, etc. We do have CI here to do
ONLY basic tests and I really hope that you dont rely on it. We dont know
how you run tested all of those packages.

We dont have crystal ball for what and why you are doing it. There is a
reason, why commit description exists.

Dne pá 16. 2. 2024 15:01 uživatel Aleksey Vasilenko <
***@***.***> napsal:

krant:
There are no quality in years-dated software with a lot of known bugs and CVEs. Having it on the edge devices which are primary target of OpenWrt is the opposite of quality.
Regarding my testing - you're again assumed bad intentions in someone, what a surprise (it is not). But I open you the secret - I have fast multicore CPU, so I could iterate quickly, and my run tests are very simple - I just.. run the software on my router. By the way, it is the same router I use for the Internet access, that's how reckless I'm.
Let me stop interacting with you - as I've predicted, you've ignored straight-to-the-topic question in my previous comment. If you want to lecture me about your views on software development - first point me to the specific clause of official OpenWrt documentation which my commits you think are violating.

PR #23459, opened 2024-02-16
glib2: revert latest changes to get back to working version 2.74.0

feckert:
Maintainer:  @neheb
Compile tested: lantiq_xrx200 and x86_64
Run tested: x86_64, APU3 with modemmanager

Description:
* revert latest changes to get back to working version 2.74.0
* do not set default meson options
* Update to  latest version 2.74.7 in series 2.74.x and refresh patches

### DIFF
--- a/libs/glib2/Makefile
+++ b/libs/glib2/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=glib2
 PKG_VERSION:=2.74.0
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
 PKG_SOURCE:=glib-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@GNOME/glib/$(basename $(PKG_VERSION))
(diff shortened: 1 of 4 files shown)


yegorich:
Is version 2.80.0 better in this regard?

feckert:
@yegorich Difficult to say. I haven't tried it yet. In principle, we should update. But if, for example, the modem manager no longer works, as this was the case on last update try, then we could not update the glib2.

feckert:
@yegorich thanks for the testing the upgrade to the latest glib2.
@aleksander0m Maybe you could help us and give us a hint in the right direction why does ModemManager could not register to the dbus with the latest glib2 version 2.80.x?

yegorich:
I opened the related issue: https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/issues/849

yegorich:
@feckert could you test my PR: https://github.com/openwrt/packages/pull/23951

Looks like the problem is the default runtime_dir introduced in https://gitlab.gnome.org/GNOME/glib/-/commit/30995f9dd7c24314c362a3e30102b89196bcef84

PR #23467, opened 2024-02-17
imagemagick: add missing libzip dependency

egorenar:
Package imagemagick is missing dependencies for the following libraries: libzip.so.5.

Fixes: 7b697342e9fc ("imagemagick: update to 7.1.1-28")

Maintainer: @krant
Compile tested: Netgear XR700 ARMv7 Processor rev 4 (v7l), OpenWrt SNAPSHOT r25160-388cb43b8773
Run tested: Netgear XR700 ARMv7 Processor rev 4 (v7l), OpenWrt SNAPSHOT r25160-388cb43b8773

### DIFF
--- a/multimedia/imagemagick/Makefile
+++ b/multimedia/imagemagick/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=imagemagick
 PKG_VERSION:=7.1.1
 PKG_REVISION:=28
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_MAINTAINER:=Aleksey Vasilenko <aleksey.vasilenko@gmail.com>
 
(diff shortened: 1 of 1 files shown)


krant:
Thanks for the bug and proposed fix, it seems it was hiding there since forever. This what was happening:
- libzip was selected manually or via some other package dependent on it
- imagemagick configure script found it in staging_dir and enabled it automatically
- check script has signaled missing DEPENDS

Another possible fix would be just add `--without-zip` into configure args, so imagemagick will ignore libzip inside the staging - it will make the package a little bit smaller but without zip support. But I've checked - code bloat is just +3.5kb for imagemagic and +41kb for libzip itself (cortex-a9 target). Looks tiny to me.

@egorenar please increase PKG_RELEASE and the PR will LGTM.

BKPepe:
In the commit description, there should be something like @krant described as it is said here: https://openwrt.org/submitting-patches#submission_guidelines

PR #23476, opened 2024-02-19
openvpn: fix init errors

dibdot:
Maintainer: n/a
Compile tested: n/a
Run tested: GL.iNet GL-MT3000, OpenWrt SNAPSHOT r25153-869df9ecdf

Description:
* mute possible sed, tail etc. errors in the 'get_openvpn_option' function, to make sense of the value check.
* check the conffile existance (with .conf extension), before starting the 'start_path_instance' function. This fixes errors with
  non-existing (e.g. wrong spelling) instances and with opvn configurations as well.

### DIFF
--- a/net/openvpn/Makefile
+++ b/net/openvpn/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=openvpn
 
 PKG_VERSION:=2.6.8
-PKG_RELEASE:=5
+PKG_RELEASE:=6
 
 PKG_SOURCE_URL:=\
 	https://build.openvpn.net/downloads/releases/ \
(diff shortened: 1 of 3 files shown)


dibdot:
@feckert atomic git commits as requested ...

feckert:
@dibdot thanks for the fix.
I have pushed the following changes https://github.com/openwrt/packages/pull/23476/commits/1ec65e0ea9a0c418aa1cc3e856145ca0e795ab93 and https://github.com/openwrt/packages/pull/23476/commits/cde62f1ac8c47a3807313ff95887c06a73d107de from this pullrequest with commit  https://github.com/openwrt/packages/commit/fe736b2f8c415e6d590931f1f1f504e1d18c579b into the master branch directly.

I have not taken the commit https://github.com/openwrt/packages/pull/23476/commits/5cf03e747f054b1376c5a99c0808a52241a6d5b3 from this pullrequest, as the handling needs to be refactored. And in my opinion it is not enough to redirect the error outputs to to `/dev/null`.

PR #23481, opened 2024-02-20
python-yaml: fix build with Cython 3

krant:
Maintainer: @BKPepe 

Description:
This is cleaned up version of #23071, thanks to @predators46 for the effort. I've decided not to describe whole history of the issue - it is already too much noise while the fix is really simple.


BKPepe:
@neheb why this was merged? Nothing from previous version was addressed.

There is missing everything formal things about patch.
Everything can be found: https://openwrt.org/submitting-patches (mising author, subject, description)
- Is it backport or not?

Issue #12185, opened 2020-05-16
[utils/sysstat] hard coded log path renders user conf option obsolete


n8v8R:
Maintainer: @ratkaj
Environment: Master, 19.07
___
#### Description:

with compiler CONFIGURE_VARS set to `sa_dir="/var/log/sysstat"` the log path is hardcoded into the binary and thereby rendering user conf option `SA_DIR=` in */etc/config/sysstat* obsolete.

neheb:
ping @lucize

lucize:
Hi, I'm not the maintainer but, maybe that config option can overwrite the binary hardcoded path ?

n8v8R:
The point of opening the issue is that the hardcoded path supersedes/obsoletes the user option specified in the conf file */etc/config/sysstat*.

If you initiate `/usr/lib/sysstat/sadc -S DISK -F -L -` manually and without `[ -d /var/log/sysstat ] || mkdir -p /var/log/sysstat` (as coded in the init script) it prints:

>Cannot open /var/log/sysstat/sa17: No such file or directory
___
on a marginal note the maintainer sited for the package does not appear actively maintaining the package for quite some time

PR #23494, opened 2024-02-21
sysstat: update to 12.7.5

krant:
Maintainer: @ratkaj 

Description:
- Update package URLs
- Add missing xz dependency
- Fix incorrectly set sa_dir
- Refresh the patch

#12185 was fixed backwards which led to non-existent `/var/log/sa` was hardcoded, which in turn led to `sadc` was not able to start due to the missing dir. I've run-tested this PR - now `/var/log/sysstat` is being populated properly and `sadc` starts successfully.


utils/sysstat/Makefile
```diff
@@ -31,8 +31,8 @@ define Package/sysstat
   SECTION:=utils
   CATEGORY:=Utilities
   TITLE:=Sysstat performance monitoring tools
-  URL:=http://pagesperso-orange.fr/sebastien.godard/index.html
-  DEPENDS:=$(INTL_DEPENDS)
+  URL:=https://sysstat.github.io/
+  DEPENDS:=$(INTL_DEPENDS) +xz
```
nxhack, line 35: Since the 'xz' package is only the host package, could this be 'xz-utils'?

nxhack: It seems that 'xz' can only be specified if the meta package 'xz-utils' is selected. What about 'DEPENDS' in this case?

krant: @nxhack you're right, thank you for spotting this! I've pushed the fix in #23514.

PR #23500, opened 2024-02-22
nghttp3: fix pkgconfig file

neheb:
CMake build is passing host paths in pkgconfig.

Maintainer: @stangri


stangri:
I have no opposition to this and would gladly transfer maintainership to you, alternatively, can you educate me on in which use cases this is needed? I didn't have issues building nghttp3 before.

neheb:
This is needed for packages that use nghttp3. Without this, host paths leak and build fails.

PR #23507, opened 2024-02-22
clamav: update to 1.3.0

krant:
Maintainer: @ratkaj @lucize 

Description:
- Add build-time Rust dependency
- Don't set default and rename changed CMake options


lucize:
long on to do! LGTM

PR #23514, opened 2024-02-23
sysstat: add missing xz-utils dependency

krant:
Depending only on 'xz' hides the package when 'xz-utils' is not selected.

Last update in #23494 introduced 'xz' dependency incorrectly. Thanks to @nxhack for spotting the issue.


neheb:
it depends on xz, and not the library?

krant:
Yes, it is just a [shell script](https://github.com/sysstat/sysstat/blob/master/sa2.in#L16) reading `ZIP` variable from config (which is set to `xz`) and running the program.

PR #23520, opened 2024-02-23
httping: update to 3.5

krant:
Maintainer: @marcin1j 

Description:
- Update package URLs
- Use local tarball for sources
- Switch to CMake
- Drop obsolete patches including 'minimize' (ipk size +3KB only)
- Add 3 new patches to fix musl, openssl and cmake issues


net/httping/Makefile
```diff
@@ -8,23 +8,22 @@
...
-PKG_HASH:=3e895a0a6d7bd79de25a255a1376d4da88eb09c34efdd0476ab5a907e75bfaf8
-
-PKG_INSTALL:=1
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/folkertvanheusden/HTTPing
+PKG_SOURCE_VERSION:=v$(PKG_VERSION)
+PKG_MIRROR_HASH:=70c2a9baeed9fdb69009fcf159a5b7ef133df0b7f6df198ae2ccd1691b4713b8
 
 include $(INCLUDE_DIR)/package.mk
-include $(INCLUDE_DIR)/nls.mk
```
neheb, line 21: why get rid of nls? Now INTL_DEPENDS resolves to nothing.

krant: Nice catch, the package uses `libintl.h` for real. Reverted.

net/httping/patches/002-fix-openssl-bundle.patch
```diff
@@ -0,0 +1,11 @@
+--- a/mssl.c
++++ b/mssl.c
```
BKPepe, line 2: Commit author, commit description, etc. is missing here.

net/httping/patches/004-fix-cmake.patch
```diff
@@ -0,0 +1,32 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
```
BKPepe, line 2: Why do we need to have this patch in our repository? This is not part of upstream repository.
Did you even tried to upstream it?

BKPepe (commented):
@neheb cmon... we are going to end up with the same situation like what is happening with OpenWrt main repository and thats not wanted. All patches should include formal things (Git header) otherwise those patches can not be applied if using out of tree... :-/

krant:
> Did you even tried to upstream it?

I understand when maintainers ask to upstream complex patches to be sure it doesn't break anything, but here we're just trying to make it work and changes are really simple and obvious.

> Why do we need to have this patch in our repository?

To make it work, I can't believe person asking this is a maintainer.

> Commit author, commit description, etc. is missing here

I'm the author and commit has been described, you're just selectively ignoring it.

> patches can not be applied if using out of tree

Let's not try to fix all problems in the world and concentrate on this repo. The package version hasn't been updated for 7 years (!). If this is not a failure of maintainership I don't know what it is. I'd like to see you helping others to change the situation for the better or update the packages you volunteered to maintain (I've counted 31 outdated packages of yours), but not making up the reasons to disregard other people's work.

BKPepe:
1. Your patches can not be used with git am. 
2. Why we should have patches and maintain them for ages?
Thats does not make sense even you said in your arguments. I dont also believe that you can not understand that there is some workflow, policies which you clearly completely ignored. Please take a look at: https://openwrt.org/submitting-patches#patch_merging_and_tree_life_cycle and that there is also: https://openwrt.org/docs/guide-developer/toolchain/use-patches-with-buildsystem#naming_patches and let me quote it from there:

> ALL patches must be in a way that they are potentially upstreamable, meaning:
> 
> - they must contain a proper subject
> - they must contain a proper commit message explaining what they change
> - they must contain a valid Signed-off-by line

Not so hard to follow, right?  Now it leads to 1st point. 
You are introducing smth like what is happening in main repo - https://github.com/openwrt/openwrt/issues/10272 , which most of the things were fixed in https://github.com/openwrt/openwrt/pull/10264 (the issue was created as follow-up). Please, try to read what is going on, what is the issue and try to remember it next time, when you add patches.

You are bringing more hassle and issues, because you have not done your homework properly and now you are acting like someone broke your things. I dont get one point. You said you are trying to improve some things, but why someone should fix your improving things? That's not even polite. We can have the situation in the real world, if would be better for you to understand what I mean. You should do it once and for ages. So, please upstream your patches, otherwise we will get to this discussion again and again. IMHO everyone else is saying you that you should have commit description, I was the bad guy, who told you it for the first time and then in some other discussion, you were asked for the same and you did it and you did not even argue. Thats the same thing. :-)

Oh, thanks for bringing statistics! Only 31 outdated packages. What is so urgent to update it? You need more features? Maybe if you look properly e.g. Netdata is kinda complicated, because it is getting huge and thats not what we want.

Now the last part. Why your changes can not be upstreamed or be included as part of the next release and we will need to take care of anything happens with your changes downstream? Yeah, because downstream can contain hacky, not reviewed patches. 🤦  Please just simply dont do that.

krant:
Thanks for the links provided, unfortunately I don't see them relevant to this repo, here is the quote from it:
```
valid for target/linux/generic
```

> We can have the situation in the real world, if would be better for you to understand what I mean

I don't understand what you mean, please improve your English.

>  Only 31 outdated packages. What is so urgent to update it?

I can't believe you're a maintainer.

BKPepe:
Oh, thats because you are missing the point and the bigger picture. You
should at least think, what you are doing and what others are trying to
say. I am sorry, but both of your reasonings are wrong. We are talking here
that your patches does not have commit description, etc. and you havent
even said that anything valid why you did not upstream it. In this point, I
am missing here the point to continue in this discussion if you use as an
argument the same thing - "oh, I can not believe that you are maintainer."

You did not try to cooperate with upstream, why we should cooperate with
you as downstream repo.

Dne út 27. 2. 2024 9:37 uživatel Aleksey Vasilenko ***@***.***>
napsal:

PR #23525, opened 2024-02-24
rtpmidi: update to 23.12

neheb:
Get rid of codeload and use local tarballs.

Fixes compilation with newer fmt.

Minor cleanups.

Maintainer: @dangowrt


dangowrt:
Commit title wrongly states version 23.10 while we actually bump to 23.12. LGTM otherwise.

PR #21476, opened 2023-07-02
ragel: new package to build hyperscan

graysky2:
This is a new package for ragel which is a dependency for another new package vectorscan

(cherry picked from commit fa76c4df5c857ee564bf2f2eaeedfea5b497b1e7)

PR #23530, opened 2024-02-25
hyperscan: new package for speeding up regex ops

graysky2:
Hyperscan is a high performance regular expression matching library from Intel that runs on x86 platforms and offers support for Perl Compatible Regular Expressions (PCRE) syntax, simultaneous matching of groups of regular expressions, and streaming operations.

This has utility in speeding up snort3.

### DIFF
--- a/libs/hyperscan/Makefile
+++ b/libs/hyperscan/Makefile
@@ -1,0 +1,82 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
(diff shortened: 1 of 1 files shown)


1715173329:
same here, please use `git cherry-pick -x`.

graysky2:
Done!

graysky2:
This will need to be merged for hyperscan to build: https://github.com/openwrt/packages/pull/21476

PR #23536, opened 2024-02-25
tor: add /etc/torrc.d/ to conffiles

stokito:
The /etc/tor/torrc may contain the line:

    %include /etc/torrc.d/*.conf

So users may put their own config files there.
We should preserve the files during an upgrade.

Maintainer: @hauke
Compile tested: Not needed
Run tested: Not needed

CC: @rsalvaterra

### DIFF
--- a/net/tor/Makefile
+++ b/net/tor/Makefile
@@ -114,6 +114,7 @@ endef
 
 define Package/tor/conffiles
 /etc/tor/torrc
+/etc/tor/torrc.d/*
 /var/lib/tor/fingerprint
 /var/lib/tor/keys/*
 /etc/config/tor


stokito:
If you'll have a time please also review another tor related PR #21642

We can simplify the tor package:

Remove the `head_include` option and migrate existing value to `tail_include` or just `include`. There are no differences if we prepend or append configs.

Just add `%include /etc/tor/torrc_generated` to the torrc.
Because now the tor-hs adds it to the `tail_include` and then the init script checks if the file exists and generates the `%include` directive (which will check again the file for existence). This will significantly reduce code and improve speed.

But we should change the `/etc/tor/torrc_generated` to `/tmp/torrc_generated` or `/var/run/tor` to keep flash disk.

Uncomment the `#%include /etc/torrc.d/*.conf` to allow users simply put files to that folder without need for editing the `/etc/tor/torrc`. This will simplify many things.

I asked a Tor devs to uncomment the line in their `torrc.sample.in` but also proposed make a dedicated location for all onion services (e.g. `/etc/tor/hidden_service`)
https://forum.torproject.org/t/default-folder-for-onion-services/8466

Please check and join that discussion.

BKPepe:
You should bump `PKG_RELEASE` in the Makefile.

hnyman:
Applied as 83737ed with the PKG_RELEASE bumped.

PR #23537, opened 2024-02-26
dnsdist: disable XSK to fix the build

krant:
Maintainer: @Habbie 

XSK support is set to `auto` by default and on some hosts it is detected as `on` and leads to (from [faillog](https://downloads.openwrt.org/snapshots/faillogs/arm_cortex-a5_vfpv4/packages/dnsdist/full/compile.txt)):
```
In file included from dnsdist-backend.cc:32:
xsk.hh:28:10: fatal error: bits/types/struct_timespec.h: No such file or directory
   28 | #include <bits/types/struct_timespec.h>
```
Here we disable XSK so configure will behave more deterministically and hopefully fix the builders.

### DIFF
--- a/net/dnsdist/Makefile
+++ b/net/dnsdist/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsdist
 PKG_VERSION:=1.9.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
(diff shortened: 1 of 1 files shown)


BKPepe (changes requested):
1. Commit description is missing.
2. PKG_RELEASE needs to be bumped.

krant:
What's the point of bumping PKG_RELEASE if no single builder has produced existing version?
What's the point of commit description if all relevant info fits into the subject?
Or should I dump all error logs into git history?

BKPepe (approved):

Habbie (approved):

Habbie:
approved. thanks!

PR #23540, opened 2024-02-26
chrony: update package description

regunakyle:
Maintainer: @mlichvar 
Compile tested: Not tested 
Run tested: Not tested 

Description:
Close this forum topic
[Suggestion: State the difference between chrony and chrony-nts in package description](https://forum.openwrt.org/t/suggestion-state-the-difference-between-chrony-and-chrony-nts-in-package-description/187392)

(First time doing pull request. Do I need to test this? I can spin up a VM if needed)

### DIFF
--- a/net/chrony/Makefile
+++ b/net/chrony/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=chrony
 PKG_VERSION:=4.5
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://chrony-project.org/releases/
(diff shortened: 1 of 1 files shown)


net/chrony/Makefile
```diff
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=chrony
-PKG_VERSION:=4.5
+PKG_VERSION:=4.6
```
efahl, line 11: I think you've got the PKG_VERSION and PKG_RELEASE changes backwards, should be the same version, but the release gets bumped for non-functional changes.

regunakyle: > I think you've got the PKG_VERSION and PKG_RELEASE changes backwards, should be the same version, but the release gets bumped for non-functional changes.
> 
> EDIT: oops, should also add that PKG_VERSION is dictated by upstream...

Thanks for the hint, fixed

mlichvar:
Please squash the commits into one and add a Signed-off-by line. Other than that, it looks good to me. Thanks.

regunakyle:
> Please squash the commits into one and add a Signed-off-by line. Other than that, it looks good to me. Thanks.

Done

Issue #23206, opened 2024-01-31
openthread-br cp: cannot overwrite non-directory failed to build


outbackdingo:
```
mkdir -p /home/dingo/openwrt-r4/staging_dir/target-aarch64_cortex-a53_musl/root-mediatek/stamp
SHELL= flock /home/dingo/openwrt-r4/tmp/.root-copy.flock -c 'cp -fpR /home/dingo/openwrt-r4/build_dir/target-aarch64_cortex-a53_musl/openthread-br-2023-08-01-1738d8cd/.pkgdir/openthread-br/. /home/dingo/openwrt-r4/staging_dir/target-aarch64_cortex-a53_musl/root-mediatek/'
cp: cannot overwrite non-directory '/home/dingo/openwrt-r4/staging_dir/target-aarch64_cortex-a53_musl/root-mediatek/./var' with directory '/home/dingo/openwrt-r4/build_dir/target-aarch64_cortex-a53_musl/openthread-br-2023-08-01-1738d8cd/.pkgdir/openthread-br/./var'
make[3]: *** [Makefile:109: /home/dingo/openwrt-r4/staging_dir/target-aarch64_cortex-a53_musl/root-mediatek/stamp/.openthread-br_installed] Error 1
make[3]: Leaving directory '/home/dingo/openwrt-r4/feeds/packages/net/openthread-br'
time: package/feeds/packages/openthread-br/compile#91.78#9.31#102.11
    ERROR: package/feeds/packages/openthread-br failed to build.
make[2]: *** [package/Makefile:128: package/feeds/packages/openthread-br/compile] Error 1
make[2]: Leaving directory '/home/dingo/openwrt-r4'
make[1]: *** [package/Makefile:122: /home/dingo/openwrt-r4/staging_dir/target-aarch64_cortex-a53_musl/stamp/.package_compile] Error 2
make[1]: Leaving directory '/home/dingo/openwrt-r4'
make: *** [/home/dingo/openwrt-r4/include/toplevel.mk:232: world] Error 2
```

brada4:
openthread-br Makefile is 107 lines long, could you help with changes you made in your package tree?

outbackdingo:
this is on a stock fresh openwrt trunk build, we have not changed or modified any packages

kaizirlewagen:
@outbackdingo i've got the same error. Randomly and really nothing changed at code. 

[...]

PR #23543, opened 2024-02-26
openthread-br: fix build and modernize

krant:
Maintainer: @stintel 

Description:
On some builders the package was failing with:
```
cp: cannot overwrite non-directory '[..]/var' with directory '[..]/var
```

Here we fix it by changing settings path from /var/lib/thread to /etc/openthread

Also:
  - Remove superseded patch
  - Modernize CMake options


net/openthread-br/Makefile
```diff
@@ -49,30 +49,26 @@ define Package/openthread-br/description
 endef
 
 define Package/openthread-br/conffiles
-/var/lib/thread
+/etc/openthread
 endef
 
 CMAKE_OPTIONS += \
```
stintel, line 55: Please maintain alphabetical order.

net/openthread-br/Makefile
```diff
@@ -49,30 +49,26 @@ define Package/openthread-br/description
 endef
 
 define Package/openthread-br/conffiles
-/var/lib/thread
+/etc/openthread
```
stintel, line 52: NAK. The daemon writes its state to this file, so it does not belong in /etc at all. I used /var/lib/thread because this is the default, and on by default on OpenWrt this end up on a tmpfs. According to the timestamp, this file was written to less than 6h ago. Setting the default to a path written to flash might wear flash chips. The user should explicitly decide to do that themselves if they want to.

stintel (changes requested):

stintel:
Aside from the inline comments:
- how do you figure the patch is superseded?
- PR should be split in multiple commits:
  - fix build
  - modernize CMake options (and please elaborate because I have no idea why this is needed)

krant:
>     * how do you figure the patch is superseded?

https://github.com/openthread/ot-br-posix/pull/1893#issuecomment-1634265192

stintel:
> > ```
> > * how do you figure the patch is superseded?
> > ```
> 
> [openthread/ot-br-posix#1893 (comment)](https://github.com/openthread/ot-br-posix/pull/1893#issuecomment-1634265192)

Then please replace the patch with the new one (in a separate commit). Just dropping the patch will break compatibility with Home Assistant.

net/openthread-br/Makefile
krant: > Setting the default to a path written to flash might wear flash chips

That's makes sense, I was not aware of that. How do you propose then to fix failing builds e.g., https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a72/packages/openthread-br/compile.txt?

krant:
> Then please replace the patch with the new one

It _is_ replaced in the current package version.

stintel:
> It _is_ replaced in the current package version. That's why I've used the term _superseded_.

Good catch. Please remove the patch in a separate commit and elaborate in the commit message.

krant:
> If using /var/lib/thread is the problem it should fail on all builders. Since it builds fine on other builders, we should find out why it doesn't build rather than changing the path.

That's a good question, I don't know why it fails. All I've wanted to do is to fix the build - other changes is just casual maintenance, they're not that critical.

net/openthread-br/Makefile
stintel: Good question. I'm asking in #openwrt-devel.

stintel:
I guess we could `mkdir -p /var/lib/thread` in the proto handler before `ubus call otbr threadstart`. Feels like a hack, but might be easier than figuring out why some builders trip over /var, and others not.

krant:
> I guess we could `mkdir -p /var/lib/thread` in the proto handler before `ubus call otbr threadstart`. Feels like a hack, but might be easier than figuring out why some builders trip over /var, and others not.

Awesome, feels like a fix to me :slightly_smiling_face: 
Let me re-arrange the PR considering all your feedback..

net/openthread-br/Makefile
```diff
@@ -48,31 +48,22 @@ define Package/openthread-br/description
   A Thread border router for POSIX-based platforms.
 endef
 
-define Package/openthread-br/conffiles
-/var/lib/thread
-endef
```
stintel, line 53: Please keep this so that the content of /var/lib/thread is kept during sysupgrade.

net/openthread-br/Makefile
```diff
@@ -48,31 +48,22 @@ define Package/openthread-br/description
...
+	-DOTBR_BORDER_ROUTING=ON \
+	-DOTBR_DNSSD_DISCOVERY_PROXY=ON \
+	-DOTBR_DUA_ROUTING=ON \
 	-DOTBR_MDNS=mDNSResponder \
-	-DOTBR_OPENWRT:BOOL=ON \
-	-DOTBR_REST:BOOL=ON \
-	-DOTBR_SRP_ADVERTISING_PROXY:BOOL=ON \
-	-DOTBR_SRP_SERVER_AUTO_ENABLE:BOOL=ON \
-	-DOTBR_TREL:BOOL=ON
+	-DOTBR_NAT64=ON \
```
stintel, line 61: This does not have the same effect as enabling OT_BORDER_ROUTING_NAT64. At the end of the build, it shows the enabled OPENTHREAD_CONFIG_ options:

Current:
OPENTHREAD_CONFIG_NAT64_BORDER_ROUTING_ENABLE=1
OPENTHREAD_CONFIG_NAT64_TRANSLATOR_ENABLE=1


Vs after applying your PR:
OPENTHREAD_CONFIG_NAT64_BORDER_ROUTING_ENABLE=0                                                                                                           OPENTHREAD_CONFIG_NAT64_TRANSLATOR_ENABLE=0

stintel (changes requested):

net/openthread-br/Makefile
stintel: Interesting. Ran clean/build again and now it does not show this behaviour.

krant: I've just re-checked it on a fully clean build - both options are `1`. OT_BORDER_ROUTING_NAT64 is not even exists anywhere.

krant:
Turns out there is an issue for it - #23206

stintel (approved):

PR #23545, opened 2024-02-26
mc: depend on libe2p

oskarirauta:
Now that libe2p is separated from e2fsprogs;
midnight commander needs it added to depends.

Maintainer: unknown

Description:
I recently requested libe2p package separation from e2fsprogs; it had impact on a single package: midnight commander.
This PR adds libe2p to depends, nothing more.


BKPepe:
Ref: https://github.com/openwrt/openwrt/pull/14629

PR #23533, opened 2024-02-25
inotify-tools: update to 4.23.9.0

krant:
Maintainer: @dangowrt 
Run tested: ARMv7, Linksys WRT3200ACM, master branch

Description:
- Workaround superfluous linking with libstdccp

PR #23549, opened 2024-02-27
inotify-tools: fix link error on x86

krant:
Maintainer: @dangowrt 
Compile tested: AMD Geode

Thanks to @feckert for reporting this bug.
Previous commit (#23533) removed unnecessary linking with libstdc++ but introduced another error on x86 platforms:
```
  undefined reference to `__stack_chk_fail_local'
```
Fix it by explicitly linking libssp_nonshared.a

### DIFF
--- a/utils/inotify-tools/Makefile
+++ b/utils/inotify-tools/Makefile
@@ -3,7 +3,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=inotify-tools
 PKG_VERSION:=4.23.9.0
 PKG_HASH:=1dfa33f80b6797ce2f6c01f454fd486d30be4dca1b0c5c2ea9ba3c30a5c39855
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


utils/inotify-tools/Makefile
```diff
@@ -25,6 +25,7 @@ endif
 
 ## Avoid linking with libstdcpp
 TARGET_CXXFLAGS+= -nodefaultlibs -lc -fno-exceptions
+TARGET_LDFLAGS += $(if $(CONFIG_USE_MUSL),-lssp_nonshared)
```
feckert, line 28: Cosmetic note only. Could you please remove the <space> between `TARGET_LDFLAGS` and `+=`

feckert (changes requested):
Thanks for the quick fix. This change fixes my issue for the geode target.
What I don't understand is why the other targets don't break. They also use the musl!

utils/inotify-tools/Makefile
krant: Good catch, done

krant:
>  why the other targets don't break. They also use the musl!

That's a riddle for me too (and also why new version links to libstd++ at all?)

feckert:
@krant thanks :+1:

PR #23557, opened 2024-02-28
irqbalance: Update init script to remove duplicate spaces

schuettecarsten:
I have some strange issues with irqbalance sometimes overwriting smp_affinity values for banned/ignored IRQs. The issue is reproduceable and is mitigated when I change theway how the irqbalance command line is built. The only difference between the resulting command is that there is only one space between the -t parameter and the first -i parameter value.

### DIFF
--- a/utils/irqbalance/files/irqbalance.init
+++ b/utils/irqbalance/files/irqbalance.init
@@ -16,28 +16,36 @@ start_service() {
 	config_get_bool enabled irqbalance enabled 0
 	[ "$enabled" -gt 0 ] || return 0
 
-	# 2 is the default
 	config_get deepestcache irqbalance deepestcache 2
-
-	# 10 is the default
 	config_get interval irqbalance interval 10
-
-	# empty is the default
 	config_get banned_cpulist irqbalance banned_cpulist ''
-
-	# A list of IRQ's to ignore
-	banirq=""
-	handle_banirq_value()
-	{
-		banirq="$banirq -i $1"
-	}
-	config_list_foreach irqbalance banirq handle_banirq_value
+	config_get debug irqbalance debug 0
 
 	procd_open_instance "irqbalance"
 	if [ -n "$banned_cpulist" ]; then
 		procd_set_param env IRQBALANCE_BANNED_CPULIST="$banned_cpulist"
 	fi
-	procd_set_param command /usr/sbin/irqbalance -f -c "$deepestcache" -t "$interval" "$banirq"
+
+	procd_add_jail_mount_rw /proc
+	procd_set_param command /usr/sbin/irqbalance
+	procd_append_param command -f
+	procd_append_param command "-c $deepestcache"
+	procd_append_param command "-t $interval"
+
+	# add banned IRQs
+	handle_banirq_value()
+	{
+		procd_append_param command "-i $1"
+	}
+	config_list_foreach irqbalance banirq handle_banirq_value
+
+	# debug
+	if [ "$debug" -gt 0 ]; then
+		procd_append_param command -d
+		procd_set_param stdout 1
+	fi
+	procd_set_param stderr 1
+
 	procd_set_param respawn
 	procd_close_instance
 }
--- a/utils/irqbalance/Makefile
+++ b/utils/irqbalance/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=irqbalance
 PKG_VERSION:=1.9.3
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/Irqbalance/irqbalance.git


schuettecarsten:
Why are the checks failing?
The error message have nothing to do with a modified init script.

BKPepe:
Missing ``PKG_RELEASE`` bump in the Makefile.

1715173329:
Please format your commits, thanks!
FYI: https://openwrt.org/submitting-patches

schuettecarsten:
> Please format your commits, thanks! FYI: https://openwrt.org/submitting-patches

Done.

utils/irqbalance/files/irqbalance.init
```diff
@@ -33,11 +33,23 @@ start_service() {
 	}
 	config_list_foreach irqbalance banirq handle_banirq_value
 
+	# debug
```
BKPepe, line 36: Do we need this comment?

schuettecarsten: No. Removed and rebased the whole thing.

utils/irqbalance/files/irqbalance.init
```diff
@@ -33,11 +33,22 @@ start_service() {
 	}
 	config_list_foreach irqbalance banirq handle_banirq_value
 
+	config_get debug irqbalance debug 0
+
 	procd_open_instance "irqbalance"
 	if [ -n "$banned_cpulist" ]; then
 		procd_set_param env IRQBALANCE_BANNED_CPULIST="$banned_cpulist"
 	fi
-	procd_set_param command /usr/sbin/irqbalance -f -c "$deepestcache" -t "$interval" "$banirq"
```
1715173329, line 40: why do you remove these quotes?
See https://www.shellcheck.net/wiki/SC2086

1715173329 (changes requested):
Also please wrap a new line for every 70 chars in your commit message.

1715173329:
cc @hnyman for review

utils/irqbalance/files/irqbalance.init
schuettecarsten: The removal of the quotes fixes https://github.com/Irqbalance/irqbalance/issues/297

1715173329: Oh I see it.

The real problem is `$banirq` gets quoted, then it is evaluated as a single parameter.
For example, `banirq="-i 1 -i 2"`, the expected behavior is that parameter `-i` is passed twice, with values of `1` and `2` respectively.
But due to `$banirq` gets quoted, `-i 1 -i 2` are passed as a single completed parameter.
Guessing irqbalance just ignores unknown parameters.

I think `handle_banirq_value` can call `procd_append_param command -i "$1"` directly.

schuettecarsten: That's an interesting idea, I will test that.

schuettecarsten:
@1715173329 - updated commit message and make use of procd_append_param for parameters, also added check for default values. There might be more options to optimize this...

utils/irqbalance/files/irqbalance.init
```diff
@@ -16,28 +16,44 @@ start_service() {
...
 
-	# A list of IRQ's to ignore
-	banirq=""
+	# 2 is the default for deepestcache
+	if [ "$deepestcache" -ne 2 ]; then
+		procd_append_param command "-c $deepestcache"
+	fi
+
+	# 10 is the default for interval
+	if [ "$interval" -ne 2 ]; then
```
1715173329, line 39: I guess `2` here is a typo error. Is it a default value defined by us, or upstream irqbalance? Anyway, this if statement seems unnecessary.

1715173329 (approved):
lgtm, thanks

schuettecarsten:
Thank you for merging, the luci patch is available at https://github.com/openwrt/luci/pull/6971

hnyman:
Merged ;-)

utils/irqbalance/files/irqbalance.init
```diff
@@ -16,28 +16,36 @@ start_service() {
...
+
+	procd_open_instance "irqbalance"
+	if [ -n "$banned_cpulist" ]; then
+		procd_set_param env IRQBALANCE_BANNED_CPULIST="$banned_cpulist"
+	fi
+
+	procd_add_jail_mount_rw /proc
+	procd_set_param command /usr/sbin/irqbalance
+	procd_append_param command -f
+	procd_append_param command "-c $deepestcache"
```
1715173329, line 32: I’m so sorry I have overlooked this. Does it work as expected? I think it will result the same problem.
Generally speaking it should be `procd_append_param command -c "$deepestcache"`

schuettecarsten: Passing "-t $interval" or "-i $1" work as expected, just tested to set the interval to 15, and also IRQ banning works (which was the original bug). So I assume that "-c $deepestcache" also works.

1715173329: Okay, thanks for clarification.

PR #23558, opened 2024-02-28
eza: add new package

oskarirauta:
eza is a substitute for ls command with icons and colors and more. Common for those who use ohmyzsh or similar shell stylers.

Maintainer: Jonas Jelonek / @jonasjelonek 

### DIFF
--- a/utils/eza/Makefile
+++ b/utils/eza/Makefile
@@ -1,0 +1,46 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=eza
+PKG_VERSION:=0.18.5
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/eza-community/eza/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=9229b2111063577a0cb8650db270d0ae6bcc1b437dbacf814786f77c67b1003d
+
+PKG_MAINTAINER:=Jonas Jelonek <jelonek.jonas@gmail.com>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DEPENDS:=rust/host
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/rust/rust-package.mk
+
+define Package/eza
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=A modern, maintained replacement for ls
+  DEPENDS:=$(RUST_ARCH_DEPENDS) +zlib
+  URL:=https://eza.rocks
+endef
+
+define Package/eza/description
+ eza is a modern, maintained replacement for the venerable
+ file-listing command-line program ls that ships with Unix
+ and Linux operating systems, giving it more features and
+ better defaults. It uses colours to distinguish file types
+ and metadata. It knows about symlinks, extended attributes,
+ and Git.
+
+ And it’s small, fast, and just one single binary.
+endef
+
+define Package/eza/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/bin/eza $(1)/usr/bin
+endef
+
+$(eval $(call RustBinPackage,eza))
+$(eval $(call BuildPackage,eza))


oskarirauta:
Could someone point me out, what is so great with rust? Simple ls command has huge amount of deps and resulting binary takes 1.5mb :O that's 3 times size of whole busybox....

jonasjelonek:
> Could someone point me out, what is so great with rust? Simple ls command has huge amount of deps and resulting binary takes 1.5mb :O that's 3 times size of whole busybox....

I guess there are enough articles explaining what concepts Rust uses to address several issues that e.g. a language like C has, and how Rust could really improve security, safety, etc. for quite some things. So I won't try to explain that here, Google is your friend ;)
I really like Rust, not only because of its features, but also because of its syntax and ecosystem. But that's just my opinion.

But yes, the binary size is still a thing, especially because OpenWrt targets to be rather small. You can do quite some tuning to reduce the size (as explained [here](https://github.com/johnthagen/min-sized-rust)), but `eza` already includes some things in their `Cargo.toml` for that. The remaining issue is, that the binary is statically compiled since Rust's libstd is not available on OpenWrt (yet). In contrast, a C standard library is included in OpenWrt so most packages are dynamically compiled, and thus significantly smaller.
And besides that, `eza` has more features so I would expect it to be bigger anyway.

oskarirauta:
Yes, I've read that all.  Optionals to avoid usage of pointers, which is a direct copy from Swift - which ofcourse is missing a missing system unless you are using Ubuntu, where it also is very limited due to that it is developed and designed by Apple for targeted audience and systems. I don't like the eco system, it's like Boost libraries on c++ - I wouldn't write my own curl library, but for most of stuff, I want to write my own as the advantage of using a centralised library in itself is a huge risk when considering to security, millions of programs become insecure when one of them has a issue- and due to statical linking, program outputting "Hello world" gives a huge binary - optionals are great, but it's just more a lack of coder's skills in remembering to free pointers. In my case it's a poor man's swift except everything written in it is huge and everyone should be using centralised libraries, where in my case, it reminds of language basic - great for beginners and those who can't. These days carbon footprint is a huge thing, and it should be considered in things like these as well, I am not that eco-friendly, but something has to be wrong if in assembler it would take less than a kb - and on something else, even if it's not linked with everything being statical and all, huge capacity of binary in it's self is a risk.

utils/eza/Makefile
```diff
@@ -0,0 +1,46 @@
...
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/eza-community/eza/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=2200bc1c07511c105961dcb3fe1682cc9067d55280dac990de7623aff86cab1a
+
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DEPENDS:= rust/host
```
BKPepe, line 15:
```suggestion
PKG_BUILD_DEPENDS:=rust/host
```

small nitpicking, nothing else, sorry!

utils/eza/Makefile
```diff
@@ -0,0 +1,46 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=eza
+PKG_VERSION:=0.18.4
```
BKPepe, line 4: My 2c: Hmm, hmm... they are releasing very often. I see there is a new version. This will be very hard to maintain in terms of keeping this up to date. One guy complained recently that so many packages are so old in this repo... Dont take it wrong. Just my thoughts.

BKPepe (approved):
The Makefile looks great! 👏 👏  As I said in different PR, you learned so much that now I dont have anything else to add. Thank you!

oskarirauta:
**Hold on, at least before merging!**
Some conversation about package is necessary due to facts unknown to me before.

If they push that often, I must admit that I am not going to be able to maintain it that often. It would be wrong for the community to me maintain this if they push new one once in a week; though I wonder, what they could improve with it every week, in the end; this is a replacement for command to list files. I reviewed changelogs for some of last releases for this; mostly they seem to have updated project's submodules (or crates in rust?) - so it doesn't change user experience a lot, so it probably doesn't matter if it won't be updated for every release, but that's my thought only.

Anyway, it's a shame that it's not the other way around; lsd (LS deluxe) that I pushed at same time with this, is very similar to this; but I like eza a lot more, how it displays directory listing with -l argument looks way better, so I would take this over lsd any day; but I do realise that if it is necessary to be updating this every week- it could become a issue, so should we drop this?

oskarirauta:
> The Makefile looks great! 👏 👏 As I said in different PR, you learned so much that now I dont have anything else to add. Thank you!

You are welcome.. I'll clean up that extra space and update this after decision on how to proceed with this due to too often updates in mainstream.

utils/eza/Makefile
oskarirauta: None taken.
Concern about this is in place. I cannot volunteer to update it once in a week if they keep up with that pace.

jonasjelonek:
If nobody minds, I would volunteer to take over ;).   
In my daily usage I mostly replaced ls with eza. Thus, it would be quite nice to have that package in OpenWrt and I wouldn't mind to update it that often.

oskarirauta:
@jonasjelonek 

Sounds great to me! I can add you as the maintainer for the package for you.

jonasjelonek:
Would you mind changing my email to 'jelonek.jonas@gmail.com'? I'm rather using that one ... sorry, my bad. (I should also change that on my github profile ;) ).

otherwise LGTM

oskarirauta:
> Would you mind changing my email to [jelonek.jonas@gmail.com](mailto:jelonek.jonas@gmail.com) ? I'm rather using that one ... sorry, my bad. (I should also change that on my github profile ;) ).
> 
> otherwise LGTM

Done.

1715173329 (approved):

PR #23559, opened 2024-02-28
lsd: add new package

oskarirauta:
ls deluxe. Substitute for ls command with style.
Popular amongst people who use ohmyzsh and similar shell stylers.

Maintainer: me

### DIFF
--- a/utils/lsd/Makefile
+++ b/utils/lsd/Makefile
@@ -1,0 +1,40 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lsd
+PKG_VERSION:=1.0.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/lsd-rs/lsd/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=ab34e9c85bc77cfa42b43bfb54414200433a37419f3b1947d0e8cfbb4b7a6325
+
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DEPENDS:=rust/host
+PKG_BUILD_PARALLEL:=1
+
+include $(INCLUDE_DIR)/package.mk
+include ../../lang/rust/rust-package.mk
+
+define Package/lsd
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=LSDeluxe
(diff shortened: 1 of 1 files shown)


oskarirauta:
Once upon a time there was era, when demos were supposed to fit in 1K... Now we have rust :D

utils/lsd/Makefile
```diff
@@ -0,0 +1,41 @@
...
+PKG_SOURCE_URL:=https://codeload.github.com/lsd-rs/lsd/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=ab34e9c85bc77cfa42b43bfb54414200433a37419f3b1947d0e8cfbb4b7a6325
+
+PKG_MAINTAINER:=Oskari Rauta <oskari.rauta@gmail.com>
+PKG_LICENSE:=Apache-2.0
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_BUILD_DEPENDS:= \
+	rust/host
+PKG_BUILD_PARALLEL:=1
```
BKPepe, line 17:
```suggestion
PKG_BUILD_DEPENDS:=rust/host
PKG_BUILD_PARALLEL:=1
```

Just nitpick

BKPepe (approved):
I need to say, @oskarirauta, have you been practicing? Because the Makefile looks very great! It's great to see improvement over here.

oskarirauta:
@BKPepe

No I have not. It's my standard issue Makefile that I always use, copied it from my other packages - that's why, eza that I pushed a PR as well, had first a wrong license.... Or was it this one, can't remember. You left me wondering, what has been wrong with my past Makefile's- since they almost always are modified copies of my previous push.. Well, it is not important, I am glad and appreciateif you liked it.

1715173329:
Okay... So I wonder why we need two similar packages (from my point of view)

oskarirauta:
> Okay... So I wonder why we need two similar packages (from my point of view)

I didn't want to leave some one just because in my opinion on looks better than other; in overall, lsd gives more details, producing longer lines when used with -l -- I rarely need those details, and am able to retrieve them if I want to. lsd also provides a bit more versatile configuration options, though so many of those might be confusing to someone else, because for the average user, enough is already provided with eza. I'd feel bad to be the judge of choosing correct one, most of linux-utils package is included in busybox; but yet we have linux-utils, same for procps, because busybox isn't sometimes enough.
Beauty is in eye of the beholder.

Should we vote if only either of them should be kept? In that case, I would vouch for eza for reason of personal opinion of it looking better, even despite I'd like it to have more configuration options, whether I'd use them or not.

1715173329:
> I'd feel bad to be the judge of choosing correct one, most of linux-utils package is included in busybox; but yet we have linux-utils, same for procps, because busybox isn't sometimes enough.

Sounds reasonable to me, thanks for clarification!

1715173329 (approved):

PR #23560, opened 2024-02-28
acme: use validation_method option instead of guessing

stokito:
Maintainer: @tohojo
Run tested: OpenWrt master

Description:
https://github.com/openwrt/luci/issues/6937


BKPepe:
Since you are touching both packages, in all of them you need to bump PKG_RELEASE / PKG_VERSION. If you don't do that, then new installations would have such changes, but existing ones do not.

stokito:
Sorry, forgot about this. Added and made a force push.

net/acme-acmesh/files/hook.sh
```diff
@@ -107,7 +107,8 @@ get)
 		set -- "$@" --days "$days"
 	fi
 
-	if [ "$dns" ]; then
+	case $validation_method in
```
tohojo, line 110: We should quote the variable here; otherwise this LGTM :)

stangri:
@tohojo sorry, could you please fill me in, why there's acme-acmesh and acme-common. Which other packages the code of acme-common is shared with?

stokito:
I made a force push with quoted `case "$validation_method" in`

tohojo (approved):
LGTM :)

stokito:
@tohojo probably you can merge it

systemcrash:
Recommend this gets cherry picked to at least 23, and maybe 22. Otherwise @stokito can make PRs for those cherry picks...?

BKPepe:
This was merged to the master today. We should wait at least one day or
two, then I can cherry pick it.

Dne pá 1. 3. 2024 21:26 uživatel Paul Donald ***@***.***>
napsal:

systemcrash:
OK

PR #19701, opened 2022-10-25
libwebsockets: full variant provides OpenSSL

BKPepe:
Maintainer: @karlp 
Compile tested: Turris 1.1, OpenWrt 21.02.5, mpc85xx/p2020
Run tested: 

- **conflicts OpenSSL and full variant:**

Before:
 ```
 root@turris:~# opkg list-installed | grep libwebsockets
libwebsockets-openssl - 3.1.0-2
root@turris:~# opkg install libwebsockets-full
Installing libwebsockets-full (3.1.0-2) to root...
Downloading https://repo.turris.cz/hbk/omnia/packages/packages/libwebsockets-full_3.1.0-2_arm_cortex-a9_vfpv3-d16.ipk
Collected errors:
 * check_data_file_clashes: Package libwebsockets-full wants to install file /usr/lib/libwebsockets.so
	But that file is already provided by package  * libwebsockets-openssl
 * check_data_file_clashes: Package libwebsockets-full wants to install file /usr/lib/libwebsockets.so.14
	But that file is already provided by package  * libwebsockets-openssl
 * opkg_install_cmd: Cannot install package libwebsockets-full.
 ```
 
 After:
 ```
 root@turris:~# opkg install libwebsockets-full_3.1.0-3_powerpc_8548.ipk 
Installing libwebsockets-full (3.1.0-3) to root...
Configuring libwebsockets-full.
root@turris:~# opkg install libwebsockets-openssl_3.1.0-3_powerpc_8548.ipk 
Installing libwebsockets-openssl (3.1.0-3) to root...
Collected errors:
 * check_conflicts_for: The following packages conflict with libwebsockets-openssl:
 * check_conflicts_for: 	libwebsockets-full * 
 * opkg_install_cmd: Cannot install package libwebsockets-openssl.

 ```
- **full variant provides OpenSSL:**
```
root@turris:~# opkg list-installed | grep libwebsockets
libwebsockets-full - 3.1.0-2
root@turris:~# opkg install ttyd
Installing ttyd (1.6.3-3) to root...
Downloading https://repo.turris.cz/hbt/turris1x/packages/packages/ttyd_1.6.3-3_powerpc_8548.ipk
Configuring ttyd.
root@turris:~# opkg install mosquitto-ssl
Package mosquitto-ssl (2.0.15-1) installed in root is up to date.

root@turris:~# opkg remove libwebsockets-full
No packages removed.
Collected errors:
 * print_dependents_warning: Package libwebsockets-full is depended upon by packages:
[...]

PR #23562, opened 2024-02-29
libwebsockets: update to 4.3.3

liudongmiao:
Maintainer: OpenWrt.org
Compile tested: ath79-generic-openwrt-23.05 (docker sdk tag)
Run tested: ath79/generic, OpenWrt 23.05.2, Netgear WNDR3800

Description: Update libwebsockets to latest stable 4.3.3, released on 2023/11/20


BKPepe (changes requested):
Master first then it could be backported to the stable branches.

liudongmiao:
@oskarirauta @neheb add two fix for lws-4.3.2 in https://github.com/openwrt/packages/commit/6a897e22b0cd6d06fe99f9e651b6dc4347eb8caf.

For 010-fix-enum-int-mismatch-openssl.patch, the upstream pr is https://github.com/warmcat/libwebsockets/pull/2824 and is merged in lws-4.3.3.
For 011-fix-enum-int-mismatch-mbedtls.patch, the upstream pr is https://github.com/warmcat/libwebsockets/pull/3027, and upstream said it's fixed already.

@BKPepe Then, how to do with master? Do we need to keep the 011-fix-enum-int-mismatch-mbedtls.patch?

BKPepe:
My idea:

1. Drop, try to reproduce, if the issue still persists.
2. Keep it, compile it and you will see how it goes
3. You can check the source code and I see that you did not look at it.
The patch was applied here: https://github.com/warmcat/libwebsockets/commit/a07699d26973eec85795eebef425212637fc21b9

So, while you will bump this to 4.3.3, you will see that both patches does not apply, because both of them are included in the 4.3.3 release.

You are welcome.

liudongmiao:
@BKPepe Thanks. (And sorry, I din't check the history of the file in upstream.)

And make a new pr for master: https://github.com/openwrt/packages/pull/23566

liudongmiao (commented):
make it base on master

neheb:
No need for these two commits. Merge into 1.

liudongmiao:
@1715173329 As I change the base from openwrt-23.05 to master, could you remove the label "OpenWrt 23.05"?

liudongmiao:
[oskarirauta](https://github.com/openwrt/packages/commits?author=oskarirauta) authored and [neheb](https://github.com/openwrt/packages/commits?author=neheb) committed Jul 1, 2023

@oskarirauta @neheb Update libwebsockets 4.3.3 revert your commit 6a897e22b0cd6d06fe99f9e651b6dc4347eb8caf based on 4.3.2 (as there were fixed in 4.3.3), please help to review. And, do we need to add wolfssl variant?

BKPepe:
I don't know what you are trying right now.  

This commit: https://github.com/openwrt/packages/commit/2b9b12fd7c7a6cc9841c5ddbcb932dbd5df2f8a5 was perfectly fine and nothing else was needed except that you need to run: ``make package/foo/refresh V=s``. Refresh does not mean to do revert commit, etc. It was not a complicated step, and now we are trying to reinvent the wheel.

Such a simple change, and now it is not even efficient for simple bumps. :-/

libs/libwebsockets/Makefile
```diff
@@ -67,6 +67,15 @@ define Package/libwebsockets-mbedtls
 	CONFLICTS:=libwebsockets-openssl
 endef
 
+define Package/libwebsockets-wolfssl
+	$(call Package/$(PKG_NAME)/Default)
+	TITLE += (wolfSSL)
+	DEPENDS += +libwolfssl
+	VARIANT:=wolfssl
+	PROVIDES:=libwebsockets
+	CONFLICTS:=libwebsockets-openssl
```
BKPepe, line 76: Conflict only with OpenSSL? What about mbedtls?

libs/libwebsockets/Makefile
```diff
@@ -67,6 +67,15 @@ define Package/libwebsockets-mbedtls
 	CONFLICTS:=libwebsockets-openssl
 endef
 
+define Package/libwebsockets-wolfssl
+	$(call Package/$(PKG_NAME)/Default)
```
BKPepe, line 71: Why reuse PKG_NAME variable here?

BKPepe:
The revert commit needs to be gone. Also, as you added WolfSSL variant (why not do that in the different PR, oh boy...)
You need to follow https://openwrt.org/submitting-patches#submission_guidelines , basically you need to add commit description, etc.

liudongmiao:
@BKPepe Thanks for Review. Just remove the wolfssl, and rebase into one commit.

BKPepe:
However, i still do have problem that you are using word revert. You dont
revert anything. You removed patch, because it was backport as a patch from
upstream.

Dne út 5. 3. 2024 16:22 uživatel Liu DongMiao ***@***.***>
napsal:

liudongmiao:
@BKPepe Thanks, I use other words to avoid reverts:

> The old patches are removed as there were fixed in 4.3.3.

BKPepe:
Much better. 👏

liudongmiao:
@BKPepe current failed either in build image, or in depends. I think it's nothing due to the libwebsockets itself.

And I add two more commits, as it's not merged yet.
- make openssl variant provides libwebsockets
- make all variants requires libuv, its default for distros like debian, ubuntu, fedora, centos, arch linux and macos

libs/libwebsockets/Makefile
```diff
@@ -55,24 +57,20 @@ define Package/libwebsockets-openssl
...
-	CONFLICTS:=libwebsockets-full
 endef
 
 define Package/libwebsockets-mbedtls
 	$(call Package/$(PKG_NAME)/Default)
 	TITLE += (mbedTLS)
 	DEPENDS += +libmbedtls
 	VARIANT:=mbedtls
-	PROVIDES:=libwebsockets
-	CONFLICTS:=libwebsockets-openssl
```
BKPepe, line 67: > remove conflicts as similar packages don't specify at all.

This is really not true at all. Did you figure out why there are conflicts and what is their purpose? It does not mean that those are not wrong. Please check this out: https://github.com/openwrt/packages/pull/19701

BKPepe:
Anyway, this is really getting huge and complicated. What are you doing is now that all your changes even though splitted in multiple commits are in pull request, which should only update it to 4.3.3. So it gets really difficult to review it and as well to get in. There was a reaosn, why it was not merged, because it was waiting for CI/CD to finish its job.

Also, if you are adding more commits, which are not relevant to PR, then it is get messy, because of the noise. Everyone could review different commits, give you feedback and it keeps reminder others, which might not be interested in other commits.

Please try to keep this simple as possible. 

1. Update: OK
2. Removing conflicts: not OK
3. Not understand the 3rd commit.

libs/libwebsockets/Makefile
liudongmiao: IMO, package should avoid to require specify variant.

`mosquitto-ssl` requires `libwebsockets-openssl`, and `ttyd` requires `libwebsockets-full`.
If there's other package requires `libwebsockets`, don't care about the variants at all, then which they should install?

Of course, if it's the real case, then make `libwebsockets-full` provides `libwebsockets` and extra `libwebsockets-openssl`.

And if `libwebsockets-full` provides `libwebsockets-openssl`, then it's ok:
```
root@OpenWrt:~# opkg list-installed | grep -E '(libwebsockets|ttyd|mosquitto-ssl)'
libwebsockets-full - 4.3.3-1
mosquitto-ssl - 2.0.17-1
ttyd - 1.7.3-1
```

liudongmiao:
@BKPepe For the third commit, remove the `-DLWS_WITH_PLUGINS=ON`.

It's because it's not binary compatible if we enable `LWS_WITH_PLUGINS`, refer:

https://github.com/warmcat/libwebsockets/blob/v4.3.3/include/libwebsockets/lws-context-vhost.h#L646-L650

It takes me 3 days to find the why my codes doesn't work with specify variant.

There are two solutions:
- For all variants, add `-DLWS_WITH_PLUGINS=ON`
- For all variants, remove `-DLWS_WITH_PLUGINS=ON`
As other distros don't add `-DLWS_WITH_PLUGINS=ON`, so I choose to remove.

liudongmiao:
@BKPepe I just remove other (conflicts / provides, libuv, `-DLWS_WITH_PLUGINS=ON`) in the PR.
Yes, it should be discussing in other PR, not this one.

If it's merged, then I will add other PR to discussing them.

PR #23569, opened 2024-03-01
ntpclient: remove

systemcrash:
There exist forks out there, but no mirrors. 

The best looking sibling project is https://github.com/troglobit/sntpd which instead daemonizes, integrates with syslog, and also provides IPv6. Its options are not synonymous so it is not a drop-in replacement. 

The only other source I that looks like the original 2015 release is here: https://github.com/dnetlab/ntpclient/commit/7e574989f736018530a2043b740f18cb101f9b0e 

I don't know how much ntpclient functionality ntpd can handle today. That's a separate discussion. 

Edit: http://doolittle.icarus.com/ntpclient is down - let's see if it comes back. If not, remove this package.


BKPepe:
This needs to be done different. 

1. subject should be clear e.g:
- ntpclient: remove
- ntpclient: drop the package

2. in the commit description, you need to describe why.

3. nitpicking: subject does not end with dot

Anyway, everything is described in CONTRIBUTING.md file located in this repository and even here: https://openwrt.org/submitting-patches

systemcrash:
OK - done. 

The site is not back up, but there exist sources out there somewhere, should we ever *need* to resurrect this.

BKPepe:
Nope. Leave it to die.

Dne po 4. 3. 2024 14:48 uživatel Paul Donald ***@***.***>
napsal:

BKPepe:
Nope, you are wrong. Even though, source host is no longer available, then the sources are still available here as it was once downloaded - https://sources.cdn.openwrt.org/.

PR #23577, opened 2024-03-03
qrencode: add support for PNG output

M95D:
Maintainer: @jp-bennett 
Compile tested: AMD64, Gentoo Linux, OpenWrt r24600
Run tested: mvebu, Linksys WRT-1900ACv1, OpenWrt r24600
Tests done:
- qrencode works from command line, generates valid PNG files
- Luci generates QR images for wireguard config

Description:
Added menuconfig option to enable PNG output. Default=disabled.


jp-bennett:
Looks about right.

libs/qrencode/Makefile
```diff
@@ -56,8 +66,8 @@ digits or 4000 characters, and is highly robust.
 endef
 
 CMAKE_OPTIONS += \
-	-DCMAKE_DISABLE_FIND_PACKAGE_PNG=ON \
-	-DWITHOUT_PNG=ON \
+	-DCMAKE_DISABLE_FIND_PACKAGE_PNG=O$(if $(CONFIG_QRENCODE_PNG),FF,N) \
+	-DWITHOUT_PNG=O$(if $(CONFIG_QRENCODE_PNG),FF,N) \
```
neheb, line 70: need PKG_CONFIG_DEPENDS I think.

libs/qrencode/Makefile
```diff
@@ -30,7 +30,9 @@ define Package/libqrencode
   SECTION:=libs
   CATEGORY:=Libraries
   TITLE:=Library for encoding data in a QR Code symbol
+  MENU:=1
```
1715173329, line 33: not required

libs/qrencode/Makefile
```diff
@@ -40,6 +42,14 @@ as a mobile phone with CCD. The capacity of QR Code is up to 7000
 digits or 4000 characters, and is highly robust.
 endef
 
+define Package/libqrencode/config
+  config QRENCODE_PNG
+	bool "Enable PNG output"
+	depends on PACKAGE_libqrencode
+	select PACKAGE_libpng
```
1715173329, line 49: this should be done via `DEPENDS` already

libs/qrencode/Makefile
```diff
@@ -40,6 +42,14 @@ as a mobile phone with CCD. The capacity of QR Code is up to 7000
...
+	default n
```
1715173329, line 50: not required

1715173329:
btw this commit seems to come from several months ago, please rebase it upon the HEAD.

libs/qrencode/Makefile
M95D: It looks really bad without it. Please leave it.

libs/qrencode/Makefile
M95D: Yes, but in that case the dependency will not be visible in menuconfig Help.

libs/qrencode/Makefile
M95D: OK, but I like things explicit. Let the package maintainer decide.

libs/qrencode/Makefile
M95D: I'm not exactly sure what that does and how to write it.
Is this ok?
`PKG_CONFIG_DEPENDS:=CONFIG_PACKAGE_qrencode CONFIG_PACKAGE_libpng`
If I write it like that, will it not trigger libpng to be built even when PNG option is off?

M95D:
Rebased.

libs/qrencode/Makefile
neheb: Not quite. You need to write the name of the option itself in terms of config to force rebuilds. There are examples in this repo.

M95D: Not helpful.

neheb: uhhhh OK

Look at your .config file and look for QRENCODE_PNG . Use the full string.

M95D: `PKG_CONFIG_DEPENDS:=CONFIG_PACKAGE_qrencode CONFIG_QRENCODE_PNG` ?

M95D: It's probably the first option.
Added.

systemcrash (approved):

libs/qrencode/Makefile
```diff
@@ -19,7 +19,7 @@ PKG_MAINTAINER:=Jonathan Bennett <JBennett@incomsystems.biz>
 PKG_LICENSE:=LGPL-2.1-or-later
 PKG_LICENSE_FILES:=COPYING
 
-PKG_CONFIG_DEPENDS:=CONFIG_PACKAGE_qrencode
+PKG_CONFIG_DEPENDS:=CONFIG_PACKAGE_qrencode CONFIG_QRENCODE_PNG
```
neheb, line 22: I don’t think the firs is needed.

M95D: Done.

neheb (approved):

M95D:
Thanks!

PR #23578, opened 2024-03-03
syslog-ng: improve logread

stokito:
Maintainer: @BKPepe @dibdot 

Make the logread script more stable and shorter.
The PR has a few commits, please review and I'll squash them into one with a short description. 

Previous review comments:

> printf to echo migration - why? "Simple" for whom?

The echo adds a new line itself. Code become shorter and easier to read.

> No longer check for log file existence
The original ubox logread never fails when no logs and basically this is a normal case. The logs may be absent and in this case nothing will be printed to stdout, only to stderr. Once the file created it will be immediately printed.

> An unknown option ends up with ?

Thank you, I wondered how I didn't noticed this during testing. We can't print the wrong option so I changed to just print a warning without the option.

> A non number count as input ends up with `tail: invalid number`

I agree that the message may be cryptic. Still the tail makes the check itself and we may not worry about injection here. Silently remove non digits also has disadvantages.
The script will be mostly called from other applications and garbage is unlikely.


dibdot:
"printf" after "echo": this change is completely unnecessary for me. If you really want to save a few bytes, simply shorten the GPL header.

I think it's negligent not to check whether the log file is there or whether numerical values are actually passed (the old script sets a sensible default if the user makes an incorrect entry). Logread is very often used interactively by many users and I consider it necessary, to check the input and print clear error/output messages with an appropriate return code.

The latest version of this PR no longer works at all, as it now does an "echo tail ..." without calling tail in a subshell, i.e. "$(...)". The whole construct is far too complicated and error-prone - please just check the input in advance and react accordingly, simple tail calls are then sufficient.

The old script has now been running under Turris OS for a good 5 years without any problems. In any case, the adjustments do not bring more stability at the moment.

stokito:
> do not bring more stability

1. The new version will process `

PR #23121, opened 2024-01-18
lualanes: Update to upstream v3.16.2

Fail-Safe:
Maintainer: Vladimir Malyutin <first-leon@yandex.ru> / @first-leon
Compile tested: x86_64, Generic x86_64, OpenWrt SNAPSHOT r24797
Run tested: x86_64, Generic x86_64, OpenWrt SNAPSHOT r24797

Description:
Update the PKG_VERSION to pull version 3.16.2 from upstream. This includes fixes for the `pthread_yield: symbol not found` issue. This update also removes a deprecated patch.

PR #23592, opened 2024-03-05
lualanes: update to version 3.16.3 and use tarball

BKPepe:
Maintainer: @first-leon 
Compile tested: MacBook A1990, Sonoma 14.3.1 for Turris Omnia router
Run tested: Turris Omnia, mvebu/cortexa9, OpenWrt 22.03

Description:
1. Update it to version 3.16.3
Release notes: https://github.com/LuaLanes/lanes/releases/tag/v3.16.3

2. Change to download tarball instead of checking out Git sources In the previous commit (in the Fixes tag), it was changed to Git sources without any reason. Let's revert it back. Let's use again tagged release. Fixes: https://github.com/openwrt/packages/pull/23121#pullrequestreview-1917612521

Fixes: b93e5b45b1daac827d429b51d8763226268f2b9a ("lualanes: Version bump to v3.16.2")

### DIFF
--- a/lang/lualanes/Makefile
+++ b/lang/lualanes/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lualanes
-PKG_VERSION:=3.16.2
+PKG_VERSION:=3.16.3
 PKG_RELEASE:=1
 
-PKG_SOURCE_PROTO:=git
-PKG_SOURCE_URL:=https://github.com/LuaLanes/lanes.git
-PKG_SOURCE_VERSION:=489f7caf9ed893c1e8efc5c1b6ecb757a2131932
-PKG_MIRROR_HASH:=667042a2d773bb4d6139abca1ed2b62c2b6b3f4dd7c5446914de548625ea08f8
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/LuaLanes/lanes/tar.gz/v$(PKG_VERSION)?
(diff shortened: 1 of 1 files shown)


Fail-Safe:
> 2\. it was changed to Git sources without any reason

Sorry, but not true. Don't assume the worst intent in people.

I had to switch to sources because there was newer code there that fixed an issue. That code was not incorporated in the v3.16.2 release, which was the latest available release at the time I submitted the original MR.

BKPepe:
So, why did we mistakenly say that it is some version like 3.16.2, which it was not, because you were checking out additional commits/patches? You haven't said anything about this in the commit description, and there is nothing said in the pull request, which was merged.  The only thing you need is to backport one commit as a patch to fix the compilation on musl. Nothing else.

1. If you want to use Git sources, then use it without PKG_VERSION. 
- Otherwise, as said, you will not use tagged release, and the version will be different from the sources. Thats not wanted.
- If you use Git sources, users would know exactly which sources you use.

2. If you want to use release, then you realize you need something more, then backport it as a patch. 

It is not so complicated and hard, though.

Fail-Safe:
I was honest from the start in making it known that I am new to this: https://github.com/openwrt/packages/pull/23121#discussion_r1460674165

Not trying to bend the rules, not trying to undermine the project--just new to it. Had someone like yourself been kind enough to take a moment, like you've done today, and give me some guidance as in your numbered/bulleted points above, I would have happily fixed things and complied.

> It is not so complicated and hard, though.

Now I just feel like you're being condescending, which I do not appreciate in the least. Maybe if this is something you do day-in and day-out, or simply have done for years, it might be second nature.

Anyway, whatever your MO is, I hope you are pleased with the result. At least the OpenWrt community gets an updated version of Lualanes at the end of the day.

Issue #8238, opened 2019-02-18
perl: spurious build failures on openwrt-18.06 branch


mweinelt:
Maintainer: @pprindeville
Environment: 
- Build Host: Debian 9.8 (x86_64)
- Build Target: brcm2708-bcm2710, mpc85xx-generic
- OpenWrt https://github.com/openwrt/openwrt/commit/62feabecd8a84cfe0d6a5aa3bad2720dbf328370
- packages.git https://github.com/openwrt/packages/commit/d05b98c6c86da58db5cbda3c945007be09583609

Description:
I'm regularly building OpenWrt 18.06 through the Gluon mesh framework. Building Perl spuriously fails every once in a while. I managed to capture a build log this time.

brcm2708-bcm2710: https://gist.github.com/mweinelt/50691b00d47dd1cb6107d1c183182f02
mpc85xx-generic: https://gist.github.com/mweinelt/1925156fbca42559f5dae24c7c1dc234

mweinelt:
5d3cb594e49132158ca7b41f4246421078e8f92a

brcm2708-bcm2708
https://gist.github.com/mweinelt/9bec3e1559b94bd34f683c120a4c0761

mweinelt:
Still failing on

mweinelt:
Still failing on 619961596091449e8c95d98e591e02745b086a71.
https://github.com/freifunk-gluon/gluon/runs/578339551?check_suite_focus=true

Meanwhile I've disabled parallel builds of perl and I've yet to encounter this build failure again:

```
From: Martin Weinelt <martin@darmstadt.freifunk.net>
Date: Wed, 19 Jun 2019 01:28:44 +0200
Subject: parallel: don't build in parallel

diff --git a/lang/perl/Makefile b/lang/perl/Makefile
index b92b00279631ae07a0f23b063571d97eb8ed3efc..81dbfcdd5bd6cad92b9552f491b200fc85fd0084 100644
--- a/lang/perl/Makefile
+++ b/lang/perl/Makefile
@@ -33,8 +33,8 @@ PKG_BUILD_DIR:=$(BUILD_DIR)/perl/$(PKG_NAME)-$(PKG_VERSION)
 HOST_BUILD_DIR:=$(BUILD_DIR_HOST)/perl/$(PKG_NAME)-$(PKG_VERSION)
 PKG_INSTALL:=1
 PKG_BUILD_DEPENDS:=perl/host
-PKG_BUILD_PARALLEL:=1
[...]

PR #23602, opened 2024-03-06
perl: Run make depend to fix build race condition

Rondom:
Compile tested: x86_64, qemu, master
Compile tested: aarch64, mt7986 rfb, fork based on 23.05.2.  <= This has removed the failures mentioned in the log messages, but did not fix all races on 23.02.
Run tested: x86_64, qemu, master, Ran ` perl -E 'say("hello")'`

I could not see any races on master after applying this patch. On 23.05.2 I still see other races, but this patch seems to fix all occurences of the race described in the commit message and in the linked Debian bug report. Admittedly the testing on master was less thorough compared to the testing on 23.05.2, but I can confirm that this patch successfuly fixes one race and does not create any known new issues, so I think it is good to merge

~~Currently in draft status, because I am still waiting to confirm that the fix really fixes all issues. So I would like to wait a week or so, so I can see whether we really do not see any parallel build failures anymore.~~ I would appreciate if others who are encountering the issue (@mweinelt @blocktrron ?) can test this as well and confirm that the issue is fixed for them.

# Description:

## perl: Run make depend to fix build race condition

Run make depend before building perl. This fixes parallel build failures on machines with a high number of cores.

Example error 1:

    /bin/ln -s /build/staging_dir/hostpkg/usr/bin/generate_uudmap generate_uidmap
    make[5]: ./generate_uudmap: Command not found
    make[5]: *** [Makefile:321: bitcount.h] Error 127

Example error 2:

    /bin/ln -s /build/staging_dir/hostpkg/usr/bin/generate_uudmap generate_udmap
    ./generate_uudmap uudmap.h bitcount.h mg_data.h
    /bin/ln: failed to create symbolic link 'generate_uudmap': File exists
    make[5]: *** [Makefile:325: generate_uudmap] Error 1

Link: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=996953
Link: https://salsa.debian.org/perl-team/interpreter/perl/-/commit/366bc98c916a3455f15c532aa33a5636d2fe2803
Closes: https://github.com/openwrt/packages/issues/8238

### DIFF
--- a/lang/perl/Makefile
+++ b/lang/perl/Makefile
@@ -11,7 +11,7 @@ include perlver.mk
 
 PKG_NAME:=perl
 PKG_VERSION:=$(PERL_VERSION)
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_URL:=https://www.cpan.org/src/5.0
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
@@ -117,6 +117,8 @@ define Build/Configure
 endef
 
 define Build/Compile
+	# make depend is required to avoid race conditions: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=996953
+	+$(MAKE) -C $(PKG_BUILD_DIR) depend
 	+$(MAKE) $(PKG_JOBS) -C $(PKG_BUILD_DIR)
 endef
 


BKPepe:
Ping @muzikr, because I know that Turris OS (which is using OpenWrt with its own additional feed) have same issues with perl regarding parallel builds.

mweinelt:
Sorry, not in a position to reproduce any longer.

Rondom:
I could not see any races on master after applying this patch. On 23.05.2 I still see other races, but this patch seems to fix all occurrences of the race described in the commit message and in the linked Debian bug report and build failures on 23.05.2 which were very common before  have become a very seldom occurrence after applying this patch.

Admittedly the testing on master was less thorough compared to the testing on 23.05.2, but I can confirm that this patch successfully fixes one race and does not create any known new issues, so I think it is good to merge.

pprindeville (changes requested):
You'd need to bump the `PKG_VERSION`.

Rondom:
> You'd need to bump the `PKG_VERSION`.
Done

pprindeville (approved):
LGTM

PR #23604, opened 2024-03-07
libwslay: remove, nothing depends on it since h2o is gone

Habbie:
Maintainer: me
Compile tested: n/a
Run tested: n/a

Description:


BKPepe:
Should we also backport this change to the stable releases? I cherry-picked several commits related to dnsdist to OpenWrt 23.05 together with removal h2o, so users would not by using it anymore.

Habbie:
> Should we also backport this change to the stable releases? I cherry-picked several commits related to dnsdist to OpenWrt 23.05 together with removal h2o, so users would not by using it anymore.

yes, will do!

1715173329:
Please also make a PR to packages-abandoned.

Habbie:
to add the package there? should i do the same for h2o?

1715173329:
> to add the package there?

Yes.

> should i do the same for h2o?

That's great ;)

Habbie:
> Please also make a PR to packages-abandoned.

it turns out this was not wanted.

1715173329:
Sorry I missed your PR. I'm not watching that repo.

PR #23609, opened 2024-03-07
nginx: Fix compilation with LTO

Rondom:
Maintainer: @Ansuel 
Compile tested: x86, x86_64 QEMU, master
Run tested: x86, x86_64 QEMU, master, `wget --no-check-certificate https://localhost`

Description:
When CONFIG_USE_LTO=y, the int-size detection script will fail because a variable gets optimised out. Mark it as volatile to fix the issue.

### DIFF
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nginx
 PKG_VERSION:=1.25.3
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=nginx-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://nginx.org/download/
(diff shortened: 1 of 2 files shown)


Ansuel:
@Rondom did you tried to propose this to nginx?

Rondom:
This is a downstream patch that I fixed the issue in. I did look at how OpenEmbedded solves it (IIRC correctly, they hardcode the results of the check).

I looked at the nginx upstream repo and it seems like cross compilation has not been considered in the code. I am not aware of any tries of anyone to fix.

I do not like our patch that much either, because it relies on parsing the binary which seems a bit dirty, but I currently do not have a better idea, either.

Rondom:
@Ansuel The patch used in OpenEmbedded [has been submitted upstream in 2011](https://mailman.nginx.org/pipermail/nginx-devel/2011-December/001665.html), but [it has not been accepted](https://mailman.nginx.org/pipermail/nginx-devel/2011-December/001666.html), so it requires further work to get upstream to support cross-compilation.

I do not think our patch is upstreamable as is either, because it relies on bytes of the variable being present verbatim in the binary.

I could imaging checking for common sizes at compile time using a static assert, which would satisfy 99 percent of the use cases but I am not sure if this is worth it, when time could better be spent on working together with upstream.

neheb:
needs a PKG_RELEASE bump.

Ansuel:
@Rondom didn't notice it was a downstream patch. Ok for me for this just bump pkg release

Rondom:
> @Rondom didn't notice it was a downstream patch. Ok for me for this just bump pkg release and maybe add the lto flag to enable compilation with it?

done

PR #23614, opened 2024-03-09
espeak: remove libstdcpp dependency

neheb:
Apparently it goes away when passing --as-needed.

Maintainer: nobody


diizzyy:
espeak is since long superseded by espeak-ng so you might aswell consider retiring the port

neheb:
True.

PR #23615, opened 2024-03-09
acme: add "option" in config file

zcy85611:
acme: add "option" in config file

without "option" in config file, luci will tell "RPC call to uci/get failed with ubus code 9: Unspecified error"

微信截图_20240309165310


systemcrash (approved):

zcy85611:
it has been solved

PR #23618, opened 2024-03-11
New static-device-names package

jstrot:
Maintainer: me / @jstrot
Compile tested: x86/64, generic, OpenWrt SNAPSHOT r25250-873b0ed42c
Run tested: x86/64, generic, OpenWrt SNAPSHOT r25250-873b0ed42c

## Description

The purpose of this package is to make sure that network device names are
always assigned to the correct devices on systems, like x86, where moving a NIC
to a different motherboard slot may device names to be assigned in a different
order during boot.

The static-device-names service matches network devices by a number of
criteria, like their MAC address or PCI ID.


wez:
I'd love to try this out; is there an easy and reversible way to do that?

hurzhurz:
@wez You can simply manually download and place the files in the right location like I did:
static-device-names.config to /etc/config/static-device-names
static-device-names.init to /etc/init.d/static-device-names (probably needs chmod +x)
static-device-names.hotplug to /etc/hotplug.d/net/static-device-names (though this one is not completely finished, needs a little modification if you need hotplug support. This is the file I use: [static-device-names.hotplug.txt](https://github.com/openwrt/packages/files/14965628/static-device-names.hotplug.txt))

@jstrot It seems to work great for me, I hope it is available as a package soon! :)

jstrot:
> I'd love to try this out; is there an easy and reversible way to do that?

Hi @wez,

It's reversible if you disable the service or remove the files and reboot: All your interfaces will be back to their default names.

If you mess up the interface names, just fix the mapping in the config and restart the service and it will re-rename them according to your updated config.

Let me know how it works out for you!

jstrot:
> @jstrot It seems to work great for me, I hope it is available as a package soon! :)

Thanks for the feedback.
It's been working great for me so far as well. I'll figure out how the CI/CD pipeline works and get this committed soon.

Cheers

jstrot:
Hotplug support is complete and I also added `pci_slot` and `usb_slot` criteria.
OpenWRT maintainers, please review and merge at your convenience.
Thanks.

stangri (approved):

jstrot:
Thanks @stangri. Please go ahead and merge upstream.

utils/static-device-names/README.md
milnepe, line 1: Great script - works well for me!

There are some erroneous uci commands in the readme:
```uci commit static-commit-names```
should be:
```uci commit static-device-names```

**Review by milnepe — COMMENTED** [review]
Great script - works well for me, thanks!

File: utils/static-device-names/README.md
Comment (jstrot) [reply utils/static-device-names/README.md:1]: Thanks @milnepe, Good catch. I pushed the fix.

**Comment by Timendus** [comment]
Thanks for doing this work! Will be nice to have a package for this.

One issue I ran into is that I needed to add these lines taken from @hurzhurz's file to the end of the hotplug script for it to start working:
```bash
[ "$ACTION" = "add" ] && add
[ "$ACTION" = "remove" ] && remove
```

Other than that, great stuff.

jstrot:
@hurzhurz, sorry I had missed the change you did to the hotplug script. I thought it was just for a manual install. I pushed it.
@Timendus , thanks for pointing it out.

jstrot:
Maintainers, if there are no further comments, please merge this in. Thanks!

utils/static-device-names/README.md
```diff
@@ -0,0 +1,173 @@
...
+### Configuring a static device name
+
+Each `device` section of the configuration should have a name to make them
+easier to configure. The section name can be anything, like "WAN", or a network
+device name like "eth0".
+
+Add a new `device` section called "WAN" to hold options for the network device
+named "eth0":
+
+    uci set static-device-names.WAN=device
```
feckert, line 54: Its more commen to use lower case device names `wan` could you please update?

utils/static-device-names/files/static-device-names.init
```diff
@@ -0,0 +1,359 @@
...
+[ -n "${IPKG_INSTROOT}" ] && return 0
+
+###################
+# LOCAL FUNCTIONS #
+###################
+
+N=$'\n'  # New line
+
+LOG() {
+  local level ; level=$1 ; shift
```
feckert, line 25: You should use a standardized variable definitions header in all your functions. The source code is read more often than it is written, so a standardized format should be selected.

```
function() {
local var1="$1"
local var2="$2"

local var3

...
}
```

utils/static-device-names/files/static-device-names.init
```diff
@@ -0,0 +1,359 @@
...
+  [ "$level" = "debug" ] && [ "$TRACE" = 0 ] && return
+  logger -s -t "$NAME" -p "syslog.$level" -- "$@"
+}
+
+target_devices=
+mac_to_device_mapping=
+pci_id_to_device_mapping=
+pci_slot_to_device_mapping=
+usb_slot_to_device_mapping=
```
feckert, line 34: Are this global variables? If so you should write them in capital letters.

utils/static-device-names/files/static-device-names.init
```diff
@@ -0,0 +1,359 @@
...
+
+## Read configuration using UCI
+read_static_device_names_config() {
+  local rc=0
+  local dev
+  local mac
+  local pci_id
+  local pci_slot
+  local usb_slot
```
feckert, line 43: New line

utils/static-device-names/files/static-device-names.init
```diff
@@ -0,0 +1,359 @@
...
+    target_devices=
+    mac_to_device_mapping=
+    pci_id_to_device_mapping=
+    pci_slot_to_device_mapping=
+    usb_slot_to_device_mapping=
+    local i=0
+    while uci -q get "static-device-names.@device[$i]" > /dev/null ; do
+      dev=$(uci -q get "static-device-names.@device[$i].name")
+      if [ -n "$dev" ] ; then
+        target_devices="$target_devices$N$dev"
```
feckert, line 54: What is this `$N`? I assume that this is a space character. Why do we use this? Can't we just use a space. It makes the source code unreadable.

feckert (commented):
I have the following comment:
* change the indentation from spaces to tabs
* pay attention to the 80 character boundary

In general, I am personally not happy that it is a shell script. Experience has shown that shell scripts are prone to errors and they need to be tested again and again to see if they still work when the environment changes!

jstrot:
> I have the following comment:
> 
> * change the indentation from spaces to tabs
> * pay attention to the 80 character boundary
> 
> In general, I am personally not happy that it is a shell script. Experience has shown that shell scripts are prone to errors and they need to be tested again and again to see if they still work when the environment changes!

Hi @feckert , Thanks for reviewing.

Re: tabs, space, width, etc: are there any project-specific "style guidelines" I can read and follow?

> not happy that it is a shell script

What language would you prefer?

utils/static-device-names/files/static-device-names.init
jstrot: That *is* my standard ;-)
Using `local` or `export` separately from assignment is a habit because the assignment masks return values of commands executed, such as in `local v=$(false)`. For straight assignments, like `local v=$1`, it indeed does not matter.

I'll adapt the code to your liking, no problem. If you have a specific style guide to follow, I can do that too.

utils/static-device-names/files/static-device-names.init
jstrot: `N` was a clean way to add newline -- not a space. I renamed it to `NL` to be clearer.

feckert:
> > I have the following comment:
> > 
> > * change the indentation from spaces to tabs
> > * pay attention to the 80 character boundary
> > 
> > In general, I am personally not happy that it is a shell script. Experience has shown that shell scripts are prone to errors and they need to be tested again and again to see if they still work when the environment changes!
> 
> Hi @feckert , Thanks for reviewing.
> 
> Re: tabs, space, width, etc: are there any project-specific "style guidelines" I can read and follow?

There are no guidlines, but most scripts in OpenWrt are with tabs, because it saves memory.

> 
> > not happy that it is a shell script
> 
> What language would you prefer?

I understand that shell scripting gets you there quickly and I personally often use shell too mostly for prototyping. It always depends on how **complex** the script is. Personally (as maintainer of mwan3, which I inherited and is a complex shell hell) I decided that I would do as much as possible in C.

But that's just my personal opinion.

You have implemented my comments :+1:. Lets wait if other packages feeds maintainers have also a comment on that.

jstrot:
> You have implemented my comments 👍. Lets wait if other packages feeds maintainers have also a comment on that.

Hi @feckert , If there are no more comments, please feel free to go ahead with the merge.

Thanks

the2masters (approved):
I've been testing this package for the past few days, and it's perfect! Of course, you could add automatic configuration file creation on first boot, but this is not udev/systemd. Without this package, multiple network interfaces receive random names on x86. The same thing happens when you add several USB dongles on other platforms. With this package you can create stable names and configure them reliable in /etc/config/network.

Maybe you can squash a few commits together.

Tested-by: Stefan Hellermann <stefan@the2masters.de>

feckert:
As already mentioned, I am not happy that this is written in shell, so I cannot and do not want to take responsibility for merging it. @BKPepe A second opinion would not be bad about this topic.

the2masters:
I think the best solution would be to integrate this into netifd. For wireless devices the default configuration uses path for a long time, but for ethernet this is not possible:
```
# head /etc/config/wireless
config wifi-device 'radio0'
	option type 'mac80211'
	option path 'platform/ahb/18100000.wmac'
	option channel '1'
	option band '2g'
	option htmode 'HT40'

config wifi-device 'radio1'
	option type 'mac80211'
	option path 'pci0000:00/0000:00:00.0'
```

I think it should be possible to integrate `option path` into device sections of /etc/config/network, but I don't have enough free time to develop this. The path is unique, so the devices can be setup correctly. Right now eth0-2 are randomly assigned on each bootup on a virtual x86 machine, and netifd can only match on this randomly assigned names.

jstrot:
> I think the best solution would be to integrate this into netifd. For wireless devices the default configuration uses path for a long time, but for ethernet this is not possible:
> ...

I'm all for integrating it into netifd, it's a good idea. It's also written in C (as @feckert prefers) -- and C is my daily bread-winner language so that's not an issue.

But I'm reticent to put too much functionality into netifd...

I also have no incentive to start from scratch. I implemented it over a year ago because I had a need for this functionality (and so did others, obviously and thanks for everyone else who contributed). Unfortunately, it's been stalled here ever since -- lately because the CI/CD checks have been failing with no particular reason and no explanation of the errors, if any. I don't have any spare time to investigate alien repos' CI/CD issues or to donate beyond this.

Maybe someone else wants to give it a shot?

In the mean time, should this package be integrated?

---

In terms of the syntax, it's more than just "path": `mac`, `pci_id`, `pci_slot`, and `usb_slot` + global options like `hotplug`.

Perhaps a set of options to "match"? e.g.:
/etc/config/network
```
config globals 'globals'
        option hotplug '1'

config device
        option name 'eth0'
        list match_mac 'aa:bb:cc:dd:ee:00'
        list match_mac 'aa:bb:cc:dd:ee:01'

config device
        option name 'eth1'
        option match_usb_slot 9-1:1.0
```

And similar for /etc/config/wireless's wifi-device.

nvandamme:
Continuing on you great approach @jstrot, does it make sense to also add ethtool tuning options directly as lists in /etc/config/network to be applied whenever a static name interface is triggered?

```
config device
        option name 'eth2'
        option pci_slot '01:00.0'
        list offload 'sg on'
        list offload 'gso on'
        list offload 'gro on'
        list offload 'tso off'
        list ring 'rx 4096'
        list ring 'tx 4096'
```

resulting in and applied upon boot/interface static renaming

`ethtool -K sg on gso on gro on tso off`
`ethtool -G rx 4096 tx 4096`

jstrot:
@nvandamme, good idea, these options are definitely needed. That would also need to be a feature of netifd, or a different package -- but not static-device-names.

RadxaNaoki:
Is it possible to use this package to set the initial state of a specific device?
Specifically, I would like to configure pci_slot 0003:31:00.0 to eth0 and pci_slot 0004:41:00.0 to eth1 in the image for the Radxa ROCK 5 ITX / ROCK 5T.
I am looking to resolve https://github.com/openwrt/openwrt/issues/20202.

jstrot:
> Is it possible to use this package to set the initial state of a specific device? Specifically, I would like to configure pci_slot 0003:31:00.0 to eth0 and pci_slot 0004:41:00.0 to eth1 in the image for the Radxa ROCK 5 ITX / ROCK 5T. I am looking to resolve [openwrt/openwrt#20202](https://github.com/openwrt/openwrt/issues/20202).

Yes, @RadxaNaoki, that's exactly what this static-device-names package was meant for.

stangri:
@jstrot you're including, but not installing the default config file. Is that intentional?

Have you tested how does this work if the package/modified config are bundled in an image? Which platform(s) have you tested?

What's the use case/guidance on a live system for changing names of the ethX interfaces -- should the network config be updated beforehand with the new device names for LAN and after the renaming for WAN? Could you update README to reflect that?

jstrot:
> @jstrot you're including, but not installing the default config file. Is that intentional?

No, certainly not intentional. I can't remember if I intended to add it or decided to remove it.

Enabling hotplug by default is probably a good thing as people using it would expect it to work out of the box. I'd prefer to be open that such a feature is enabled by providing it in the default config. You're a package developer, what is the preferred way to deal with default config files? I figure test if one exists in /etc/config, else install the default.

> Have you tested how does this work if the package/modified config are bundled in an image? Which platform(s) have you tested?

Yes, I have tested with the package manually installed and as bundled in an image -- though not with simply the default config.

Personally, I've only tested it on generic x86 images. I don't know what platforms other people have tried.

> What's the use case/guidance on a live system for changing names of the ethX interfaces -- should the network config be updated beforehand with the new device names for LAN and after the renaming for WAN? Could you update README to reflect that?

It's a good idea to add some example use cases in the README. Thanks, I'll do that over the weekend...

I'm not one to do a lot of manual changes to /etc/config/network so I'd personally add a card, fix its name, then config through LuCI. But I can see someone experienced doing the manual setup in advance, before even inserting a card. If you can decide the device name ahead of time, you definitely can do that.

I don't see this being commonly used on live systems but during "maintenance windows" where one is installing/replacing a NIC. If you know beforehand that you will be inserting a new card, you can setup the device name before inserting it (e.g., by matching the MAC). Once you know the future card's device name you can pre-configure it in other configs.

A common case, like @RadxaNaoki's above, is to deal with systems where the device names may be determined dynamically during boot up and that process can be timing sensitive. Being able to specify fixed device names based on MAC or physical position is a must in those cases.

An other use case for me was to swap a bad NIC. Remove the old, add the new, reboot, update the MAC address in the static-device-names config, and force renaming live such that the new NIC takes the name of the old NIC, then all functionality was restored.

An other use case I had was I wanted to swap NIC positions, I switched the NICs in the static-device-names config, shutdown, made the cable swap, rebooted.

One more use case was to prepare a new OpenWRT version and configs in a VM with auto-generated MACs, then install the image on my real firewall. The static-device-names config was set to match multiple MACs (virtual + real) per device name. Migrating configs from one system to another without the fuss of differing device naming schemes is quite powerful.

Lastly, I recently got a Flint 2 (not yet my primary, but soon) and plan to keep the vendor's image on it as long as possible but they have "eth0" and "lan1" through "lan5" device names set up out of the box. One thing that bugs me is that I want to use "lan5" for a DMZ so I'm thinking of forcing it to be renamed "dmz1" for clarity. I think static-device-names can achieve this without getting rid of the vendor image.

Cheers,
JST

stangri:
> You're a package developer, what is the preferred way to deal with default config files? I figure test if one exists in /etc/config, else install the default.

In addition to conffiles section, this is a snippet from adblock-fast's install section:
```
	$(INSTALL_CONF) ./files/etc/config/adblock-fast $(1)/etc/config/adblock-fast
```

If the config already exists, opkg/apk keep the existing copy and install the "package" copy side by side (as `/etc/config/{{package-name}}-opkg` for opkg for example. I'd let the package manager do its thing.

My use case was the device where renaming of ports to match the case markings doesn't happen, but where PCI paths are known. So for me it was a multi-step process:
1. Confirm paths as reported by static-device-names
2. Add the new port names to LAN bridge in the network config file, reload network
3. Create the static-device-names config, run the service
4. Update the wan port in the network config and reload network one last time

Now the uci commands/static-device-names config can be brought into custom image.

Also, I understand @feckert's concerns over large shell scripts, but I feel that this isn't the case (codebase is quite small). If config file and README improvements are pushed, I'd merge this unless there are objections other than "it's a shell script".

utils/static-device-names/Makefile
```diff
@@ -0,0 +1,46 @@
...
+define Build/Compile
+endef
+
+define Package/static-device-names/install
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_BIN) ./files/static-device-names.init $(1)/etc/init.d/static-device-names
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/net
+	$(INSTALL_BIN) ./files/static-device-names.hotplug $(1)/etc/hotplug.d/net/00-00-static-device-names
+	$(INSTALL_DIR) $(1)/usr/share/doc/static-device-names
+	$(INSTALL_DATA) ./README.md $(1)/usr/share/doc/static-device-names/
```
BKPepe, line 43: I don’t think the README should be part of the final package — that’s quite unusual. It would make more sense to place it somewhere else, for example on the OpenWrt wiki.


```suggestion
```

utils/static-device-names/Makefile
```diff
@@ -0,0 +1,46 @@
...
+PKG_LICENSE_FILES := LICENSE
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/static-device-names
+  SECTION := utils
+  CATEGORY := Network
+  TITLE := Support for static device names
+  PKGARCH:=all
+endef
```
BKPepe, line 23:
```suggestion
PKG_NAMe:=static-device-names
PKG_VERSION:=1.0.0
PKG_RELEASE:=1

PKG_MAINTAINER:=Jean-Sebastien Trottier <jst@qualipsoft.com>
PKG_LICENSE:=GPL-3.0-or-later
PKG_LICENSE_FILES:=LICENSE

include $(INCLUDE_DIR)/package.mk

define Package/static-device-names
  SECTION:=utils
  CATEGORY:=Network
  TITLE:=Support for static device names
  PKGARCH:=all
endef
```

utils/static-device-names/LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
BKPepe, line 1: Do we really need this license here? 🤔 I know only @stangri does it here, but I find it still odd.

utils/static-device-names/Makefile
```diff
@@ -0,0 +1,46 @@
...
+
+define Package/static-device-names/conffiles
+/etc/config/static-device-names
+endef
+
+define Build/Compile
+endef
+
+define Package/static-device-names/install
```
BKPepe, line 37: Where is the configuration file? I see you added that file, but not included it in the package.

utils/static-device-names/README.md
```diff
@@ -0,0 +1,173 @@
...
+## Configuration
+
+### Configuration file
+
+    /etc/config/static-device-names
+
+If configuration is missing, create a new one:
+
+```sh
+touch /etc/config/static-device-names
```
BKPepe, line 39: This is indeed missing in the Makefile. There’s no reason for someone to create their own configuration file — it shouldn’t be missing in the first place. 😇
You don’t need it at all if it’s included in the package, so it can be safely removed.

utils/static-device-names/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+```sh
+service static-device-names reload
+```
+
+To see the current status of network devices without making any changes:
+
+```sh
+service static-device-names status
+```
```
BKPepe, line 139: Isn’t this type of commit practice already covered in the OpenWrt documentation?

BKPepe:
And before merging, there shouldn’t be twenty commits in the PR. You should squash them first.

feckert:
I would like to bring this back to the agenda. @BKPepe. In my opinion, this belongs to `netfid` and should not be a script. 
@jstrot Perhaps this could be integrated into `netfid`. In addition, the script must always be tested to ensure that it still works. In my opinion, this would be easier to manage in `netifd`.

From my own experience (mwan3), using shell to do this is not recommended. It quickly becomes **confusing** and **messi.**

BKPepe:
Closing. :]

PR #23622, opened 2024-03-11
perl-dbd-pg: add perl-dbd-pg

predators46:
Maintainer: @predators46 

Compile tested: 
Run tested: 

Description:

https://metacpan.org/dist/DBD-Pg

Changes for version 3.18.0
Support new PQclosePrepared function, added in Postgres 17
Greg Sabino Mullane
Better docs about ping always returning a value


predators46:
bug

PR #23633, opened 2024-03-12
amneziawg-tools: add new package

leninalive:
Maintainer: me
Compile tested: x86_64, arm64
Run tested: x86_64, arm64, mipsle

Description:
AmneziaWG VPN kernel module makes it resource efficient to connect to Amnezia VPN services. It is the hardened version of WireGuard protocol that allows to change key WG header values and add junk to handshake. This is the tools (awg and watchdog) used to manage AmneziaWG interfaces.


stangri:
@leninalive please fix the signed-of-by as it's failing test formalities check:

```
: Signed-off-by is missing or doesn't match author (should be 'Signed-off-by: Iurii Egorov <ye@amnezia.org>')
```

BKPepe:
My question is why do we need to have another Wireguard fork supported? We do have currently plenty of them in OpenWrt and adding more and more, it does not seems right for me. Usually, it happens that the developers adds it, but then they dont simply take care of it and it leads that the package is abandoned afterwards and then removed.

egc112:
WireGuard as protocol is easily detected and blocked so some form of obfuscation could be handy.
I am currently using: https://github.com/infinet/xt_wgobfs to obfuscate WireGuard (this is also available in DDWRT, I use it to connect a Broadcom DDWRT router with an OpenWRT router), but it looks like this can also obfuscate so could be useful

net/amneziawg-tools/Makefile
```diff
@@ -0,0 +1,56 @@
...
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=amneziawg-tools
+PKG_VERSION:=1.0.20240213
+PKG_RELEASE:=1
+
+PKG_SOURCE:=v$(PKG_VERSION).tar.gz
```
BKPepe, line 15: This can not be accepted as it is. The PKG_SOURCE should always include PKG_NAME together with the PKG_VERSION. If you include only PKG_VERSION, this leads that you will download file: v1.0.20240213.tar.gz into the dl folder within OpenWrt build system. This file could be overwritten by other packages, which could have the same name (in this case version), but also if many projects will do that, then you will not know which file belongs to which files.

For example, how it should look with using codeload:
```suggestion
PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
PKG_SOURCE_URL:=https://codeload.github.com/amnezia-vpn/amneziawg-tools/tar.gz/v$(PKG_VERSION)?
```

net/amneziawg-tools/Makefile
```diff
@@ -0,0 +1,56 @@
...
+PKG_SOURCE_URL:=https://github.com/amnezia-vpn/amneziawg-tools/archive/refs/tags/
+PKG_HASH:=4bde122630c9ddb1ec013c3e958f2c613b9eea56834674dda92fcb423c6f4d10
+
+PKG_MAINTAINER:=Amnezia Admin <admin@amnezia.org>
```
BKPepe, line 19: I think, we will need to have here real name.

net/amneziawg-tools/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+
+MAKE_PATH:=src
+MAKE_VARS += PLATFORM=linux
+
+define Package/amneziawg-tools
+  CATEGORY:=Network
+  URL:=https://amnezia.org/
+  MAINTAINER:=Amnezia Admin <admin@amnezia.org>
```
BKPepe, line 33: This is already at line 19, so you can remove it.

net/amneziawg-tools/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+PKG_SOURCE:=v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/amnezia-vpn/amneziawg-tools/archive/refs/tags/
+PKG_HASH:=4bde122630c9ddb1ec013c3e958f2c613b9eea56834674dda92fcb423c6f4d10
+
+PKG_MAINTAINER:=Amnezia Admin <admin@amnezia.org>
+PKG_LICENSE:=GPL-2.0
+PKG_LICENSE_FILES:=COPYING
+
+PKG_BUILD_DIR=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)
```
BKPepe, line 23: This is already default, you can remove it and see if it works without it. :)

net/amneziawg-tools/Makefile
```diff
@@ -0,0 +1,56 @@
...
+PKG_NAME:=amneziawg-tools
+PKG_VERSION:=1.0.20240213
+PKG_RELEASE:=1
+
+PKG_SOURCE:=v$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/amnezia-vpn/amneziawg-tools/archive/refs/tags/
+PKG_HASH:=4bde122630c9ddb1ec013c3e958f2c613b9eea56834674dda92fcb423c6f4d10
+
+PKG_MAINTAINER:=Amnezia Admin <admin@amnezia.org>
+PKG_LICENSE:=GPL-2.0
```
BKPepe, line 20: This LICENSE is deprecated. 

Reference: https://spdx.org/licenses/ ("Deprecated License Identifiers")

You should use it with ``only`` or ``later``

net/amneziawg-tools/Makefile
```diff
@@ -0,0 +1,56 @@
+#
+# Copyright (C) 2016-2019 Jason A. Donenfeld <Jason@zx2c4.com>
+# Copyright (C) 2016 Baptiste Jonglez <openwrt@bitsofnetworks.org>
+# Copyright (C) 2016-2017 Dan Luedtke <mail@danrl.com>
```
BKPepe, line 4: This package was adopted from some other package, here? Would be good to mention that. Otherwise someone, who does not know it, they will write them.

BKPepe (changes requested):
Reviewed only Makefile, which needs to address some comments before merging.

BKPepe:
Closing due to inactivity.

Issue #23588, opened 2024-03-05
p910nd: drivers are missing when hotpluging


loss-and-quick:
Maintainer: @pkerling, @systemcrash, @neheb 

Environment: `ramips/mt7621`, `TP-Link Archer C6U v1`, `OpenWrt 23.05.0`

Description:
Recently encountered a problem when connecting the printer: 
`daemon.err p910nd hotplug: Missing driver file: /opt/p910nd_drivers/FUJI_XEROX_DocuPrint_CP105_b_0550_00f5.bin for /dev/usb/lp0 [ 0550/00f5 ] (please upload it).`

It is solved by restarting the p910nd, but there was no such problem before.
How to solve this problem normally?

brada4:
What do you mean connecting? Like you connect usb and it asks bin file?
Hotplug?
i.e can you detail your setup and explain what does not work and how you wanted it to work.

systemcrash:
@loss-and-quick The very important text: `please upload it`. 

hotplug sets up the environment for your printer - you supply the blob. Does your printer need a blob loading to it before it will print? Or do you just plug it in and dump data to `lp0` and it prints?

If not, please paste the output of the following commands:

```sh
ls /sys/class/usbmisc/lp0/device/
cat /sys/class/usbmisc/lp0/device/ieee1284_id
cat /sys/class/usbmisc/lp0/device/bInterfaceClass
cat /sys/class/usbmisc/lp0/device/bInterfaceProtocol
cat /sys/class/usbmisc/lp0/device/bInterfaceSubClass
cat /sys/class/usbmisc/lp0/device/bNumEndpoints
```

Might need a bit more data later.

loss-and-quick:
> What do you mean connecting?

I just turn the printer on, it's always off if I'm not using it.

loss-and-quick:
Temporarily I made this script for hotplug.d/usb:
```
#!/bin/sh
if [ "$PRODUCT" = "0550/00f5/100" ]
then
case "$ACTION" in
        add)
        sleep 11s
        /etc/init.d/p910nd restart
        ;;
esac
fi
```

systemcrash:
> Temporarily I made this script for hotplug.d/usb:

please paste the output of the following commands
```sh
ls /sys/class/usbmisc/lp0/device/
cat /sys/class/usbmisc/lp0/device/ieee1284_id
cat /sys/class/usbmisc/lp0/device/bInterfaceClass
[...]

PR #23635, opened 2024-03-12
[22.03] p910nd: restart daemon even if no driver file is needed

systemcrash:
Not all USB printers need a blob loading; restart the daemon independently of driver loading.

Closes openwrt/packages#23588

Signed-off-by: Paul Donald <newtwen@gmail.com>
Tested-by: minicx <minicx@disroot.org>
(cherry picked from commit 685ef7d97b345c09edd428250794dd9fce07a174)


BKPepe:
This got even into OpenWrt 22.03? Hmm. Usually, I got blame for that like why I am doing it, etc. But, for me, it is better when I dont have so much suprises while going to newer OpenWrt versions as they are not released very often.

PR #23637, opened 2024-03-13
[22.03] opensc: update to fix CVE-2023-40660, CVE-2023-40661 and CVE-2023-4535.

sah-cdo:
Maintainer: @dangowrt 

Description:

Brings security fixes for https://github.com/advisories/GHSA-7635-x5f9-5458, https://github.com/advisories/GHSA-2c2j-2pgv-gfgc and
https://github.com/advisories/GHSA-phh2-j3h6-vqr9.

See release notes for further details:
https://github.com/OpenSC/OpenSC/releases/tag/0.24.0
and
https://github.com/OpenSC/OpenSC/releases/tag/0.23.0

Cherry picked from commit b67923b334ed32a9e8c73ecb63e62b6f3942e15e.


utils/opensc/Makefile
```diff
@@ -8,9 +8,9 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=opensc
-PKG_VERSION:=0.22.0
-PKG_RELEASE:=$(AUTORELEASE)
-PKG_HASH:=8d4e5347195ebea332be585df61dcc470331c26969e4b0447c851fb0844c7186
+PKG_VERSION:=0.24.0
```
ynezz, line 11: Indeed, this is the fastest approach, but looking at the number of changes:

```
$ git diff --stat 0.22.0..0.24.0 | grep changed
 501 files changed, 42206 insertions(+), 12300 deletions(-)
```

it seems to me like a big regression potential, so I'm wondering if it would be possible to identify just the security fixes and backport them as individual patches instead?

utils/opensc/Makefile
```diff
@@ -137,14 +143,14 @@ PROFILES:= \
 	iasecc_generic_pki \
 	iasecc \
 	incrypto34 \
-	jcop \
-	miocos \
```
ynezz, line 141: This looks like a regression, just imagine, that you're a `jcop` or `miocos` user and upgrading to this would basically breaks your use cases, right?

utils/opensc/Makefile
sah-cdo: the upgrade is not only for CVE but also for the capability to use pkcs11-tool to encrypt/decrypt with sympetric keys.

utils/opensc/Makefile
sah-cdo: Is this support still needed?
On openwrt 23.05, the support of jcop and miocos was drop.
Should we still concider this support on 22.03 branch ?
If yes, a patch on 0.24.0 could add this support. what do you think on this solution?

utils/opensc/Makefile
ynezz: > for the capability to use pkcs11-tool to encrypt/decrypt with sympetric keys.

Quoting https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#release-branches

* Please ONLY cherry-pick or commit security and bug-fixes to these branches.
* Do NOT add new packages and do NOT do major upgrades of packages here.

sah-cdo: so we need a solution on PRPL for adding new pages or doing major upgrade?
Will we fork these repo for PRPL to have more flexibility ?

ynezz: > so we need a solution on PRPL for adding new packages or doing major upgrade?

This is OpenWrt project, so that is off-topic here.

>  Will we fork these repo for PRPL to have more flexibility ?

An "upstream first" policy refers to the practice of contributing code changes to the original or main source code repository ("upstream") before using or applying these changes in your own project or a derivative ("downstream"). This approach ensures that improvements and fixes benefit the entire community and maintain compatibility with the main codebase. Forks, or separate branches of the codebase, should be avoided when possible as they can lead to fragmentation, making it harder to maintain software and integrate future updates from the original source.

utils/opensc/Makefile
ynezz: > Is this support still needed?

This is a stable release branch, so the goal is to keep it stable, thus the changes should be minimal, ideally keep it as it is and apply only security (or eventually other bug) fixes. Removing features or adding new features should be avoided, there is always a risk of breaking some existing use cases.

Simply look at this from user's perspective, imagine that you've `jcop` or `miocos` tools used somewhere in your project/device which is based on this stable/maintained release, then you notice this update with the security fixes so as a responsible sysadmin you're going to update your devices using `opkg update && opkg upgrade opensc` , but then suddenly realize, that it breaks your use case. Not a great experience, right?

>  On openwrt 23.05, the support of jcop and miocos was drop.

That is OK, its a different stable release branch, so some hiccups are expected during upgrade process between releases.

Neustradamus:
@sah-cdo: Have you progressed on CVE fixes?

BKPepe:
OpenWrt 22.03 is already End of Life, closing.

PR #23639, opened 2024-03-13
p11-kit: install p11-kit server and p11-kit-client binaries

sah-cdo:
Run a server process that exposes PKCS#11 module remotely.
```
p11-kit server pkcs11:token1 pkcs11:token2 ...
p11-kit server --provider /path/to/pkcs11-module.so pkcs11:token1 pkcs11:token2 ...
```

This launches a server that exposes the given PKCS#11 tokens on a local socket. The tokens must belong to the same module. To access the socket, use p11-kit-client.so module. The server address and PID are printed as a shell-script snippet which sets the appropriate environment variable: P11_KIT_SERVER_ADDRESS and P11_KIT_SERVER_PID

Reviewer: @neheb

### DIFF
--- a/libs/p11-kit/Makefile
+++ b/libs/p11-kit/Makefile
@@ -38,6 +38,25 @@ define Package/p11-kit/description
   way that they are discoverable.
 endef
 
+define Package/p11-kit-utils
+  SECTION:=libs
+  CATEGORY:=Utilities
+  TITLE:=A library that provides a way to load and enumerate PKCS11 modules.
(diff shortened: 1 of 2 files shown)


nmav:
How would this work? Remoting a smart card requires the two systems to have the same architecture, so you cannot remote a card from the desktop to the router. What are your intended use cases?

GeorgeSapkin:
22.03 is EOL. Closing as stale.

PR #23654, opened 2024-03-14
lpac: add package

blocktrron:
Maintainer: me
Compile tested: ath79-generic / master
Run tested: todo

Description:

lpac is a eUICC eSIM LPA manager written in C. It allows to manage eSIM profiles on eUICC SIM cards or modules using multiple backends.


bam80:
Hello, could we proceed?
Thanks.

blocktrron:
This is ready as is. I wait for confirmation to integrate the uqmi backend to make it actualy usable w/o a smartcard reader. But we can add this at a later stage.

bam80:
> I wait for confirmation to integrate the uqmi backend to make it actualy usable w/o a smartcard reader.

Where are you waiting it?

> But we can add this at a later stage.

I think that should be better. Could we fix the checks, mark it as "ready" to prepare for the merge?

Thanks.

bam80:
Thank you for the proceed.

@BKPepe 

> I wait for confirmation to integrate the uqmi backend to make it actualy usable w/o a smartcard reader.

For that, I would appreciate if you create a separate issue to track.

utils/lpac/Makefile
```diff
@@ -0,0 +1,47 @@
...
+TARGET_CFLAGS += $(FPIC)
+
+define Package/lpac/install
+	$(INSTALL_DIR) $(1)/usr/bin $(1)/usr/lib $(1)/etc/config
+
+	$(CP) ./files/lpac.uci $(1)/etc/config/lpac
+	$(INSTALL_BIN) ./files/lpac.sh $(1)/usr/bin/lpac
+
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/lpac $(1)/usr/lib/lpac
+endef
```
BKPepe, line 45:
```suggestion
define Package/lpac/install
	$(INSTALL_DIR) $(1)/usr/bin
	$(INSTALL_BIN) ./files/lpac.sh $(1)/usr/bin/lpac
	
	$(INSTALL_DIR) $(1)/etc/config
	$(INSTALL_CONF) ./files/lpac.uci $(1)/etc/config/lpac

	$(INSTALL_DIR) $(1)/usr/lib
	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/lpac $(1)/usr/lib/lpac
endef
```

There are just a few files with file extension .uci in this repository.  Maybe it would be good to have some consistency across packages. While I was checking OpenWrt main repo, there is no such file with that extension. So, maybe it would be great to have some structure like in this package: https://github.com/openwrt/openwrt/tree/main/package/network/config/qos-scripts/files/etc as example

Also, the config file I think should be marked as conffile, right? If it wont be, then the config file will be overwritten by each update and I dont think thats wanted behaviour.

utils/lpac/Makefile
```diff
@@ -0,0 +1,47 @@
...
+PKG_NAME:=lpac
+PKG_VERSION:=2.0.0
+PKG_RELEASE:=1
+
+PKG_SOURCE:=lpac-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/estkme-group/lpac/tar.gz/refs/tags/v$(PKG_VERSION)?
+PKG_HASH:=686faad1cac49649aaa749be41b44f6ec303833d4f8ef586245de4ad545d4886
+
+PKG_MAINTAINER:=David Bauer <mail@david-bauer.net>
+PKG_LICENSE:=AGPL-3.0-only
```
BKPepe, line 14: Isn't the package dual licensed according to https://github.com/estkme-group/lpac/blob/f521c455dc18459a719e4e90548829a8c9b10dfc/CMakeLists.txt#L40

utils/lpac/Makefile
```diff
@@ -0,0 +1,47 @@
...
+
+TARGET_CFLAGS += $(FPIC)
+
+define Package/lpac/install
+	$(INSTALL_DIR) $(1)/usr/bin $(1)/usr/lib $(1)/etc/config
+
+	$(CP) ./files/lpac.uci $(1)/etc/config/lpac
+	$(INSTALL_BIN) ./files/lpac.sh $(1)/usr/bin/lpac
+
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/lpac $(1)/usr/lib/lpac
```
BKPepe, line 44: Is this even correct, though? It does not look right for me.  Why we would change binary path from ``/usr/bin`` to ``/usr/lib``?

utils/lpac/files/lpac.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
```
BKPepe, line 1: This can not be done via init script?

BKPepe (commented):
Since someone requested review from my, here it is.

1. Before merging, it would be great to add test.sh file to at least have it run tested by CI/CD. You might want to look at test.sh files, which are already in this repository.
2. There were just a few things (formal ones), which could be improved and I did suggestion for that.

IMHO, I wonder how many users will use this package in the current state. Why don't we wait until there is implemented uqmi support as you said?

bam80:
> IMHO, I wonder how many users will use this package in the current state. Why don't we wait until there is implemented uqmi support as you said?

I think there are quite a lot users already.
It is usable as is:
- for modems supporting required AT commands:
  > If your modem supports these AT+CCHO/AT+CGLA/AT+CCHC AT commands, you may be able to use lpac on your OpenWrt router.
  
  https://github.com/estkme-group/lpac/issues/64#issuecomment-2054283028
- for modems supporting MBIM:
  https://github.com/stich86/lpac-libmbim-wrapper/

blocktrron:
@bam80 Not the "MAY" here. It is usable for anything not involving the SM-DS / SM-DP+ as AT commands typically timeout after 200ms.

But yes, activating and deactivating profiles as well as listing does work with the AT command backend.

utils/lpac/Makefile
blocktrron: Because you call it using the wrapper script colocated the package files.

utils/lpac/files/lpac.sh
blocktrron: I did not want to do this in order to leave the option open to support multiple devices with stored configuration.

bam80:
@blocktrron please report upstream if you faced this behavior with AT commands. Maybe there is a solution.

Still, we have a MBIM approach above which is a good to go?

blocktrron:
@bam80 Please read my comment again and take the time to understand it correctly. This is NOT an issue with lpac at all but depends on the implementation of your modem.

image

See the column "Maximum Response Time".

bam80:
@blocktrron I read your comment very carefully from the first time and my reply still persists:
please report about this situation to `lpac` as it's something probably worth to be discussed there.

Anyway, let's please concentrate on merging thins PR in here.
Thank you.

bam80:
> I wait for confirmation to integrate the uqmi backend to make it actualy usable w/o a smartcard reader. But we can add this at a later stage.

FYI: this is happened now:
https://github.com/estkme-group/lpac/pull/70#event-12857560988

Would be good to update the package either.

utils/lpac/Makefile
bam80: This is incorrect indeed:
#25035

Neustradamus:
Replaces:
- https://github.com/openwrt/packages/pull/23654
- https://github.com/openwrt/packages/pull/23697

PR #23680, opened 2024-03-17
mmc-utils: update to upstream's e1281d4de916 commit

mhei:
Maintainer: me
Compile tested: mxs
Run tested: mxs, Duckbill


BKPepe:
Could we avoid using cryptic commit subjects, like updating to the latest upstream version? Why can we not specify the specific version? If you would use git blame or browse the Git history, you will see that every single time you update the upstream version and, it makes it hard to know which version was working just by looking at the commit subject.

mhei:
You are right, this was a bad tradition. Please review again, whether my new commit message makes more sense for you.

PR #23684, opened 2024-03-18
pbr: update to 1.1.4-5

stangri:
Maintainer: me

Description:
This update includes the following changes:
1. Makefile
  * update copyright
  * attempt to implement the proper variants to avoid luci-app dependency on both variants
  * quietly stop service on uninstall

2. Config-file
  * add the list of dnsmasq instances to target in supported dnsmasq modes
  * for default pbr variant, set the `resolver_set` to `dnsmasq.nftset`
  * for iptables pbr variant, set the `resolver_set` to `dnsmasq.ipset`
  * add the `nft_file_support` (disabled by default)
  * introduce `procd_boot_delay` to delay service start on boot
  * introduce the following nft set creation options:
    * nft_set_auto_merge
    * nft_set_counter 
    * nft_set_flags_interval 
    * nft_set_flags_timeout 
    * nft_set_gc_interval 
    * nft_set_policy 
    * nft_set_timeout
  * add the pbr.user.wg_server_and_client custom user script to allow running wg server and client at the same time
  * add the "Ignore Local Requests" sample policy

3. Hotplug firewall/interface scripts
  * better logged messages

4. The pbr and pbr-iptables uci defaults script
  * use functions from the init script
  * improve vpn-policy-routing migration

5. The pbr-netifd uci defaults script
  * use functions from the init script
  * improve uci operations

6. Introduce the firewall.include file

7. Improve pbr.user.aws custom user script

8. Improve pbr.user.netflix custom user script

9. Introduce pbr.user.wg_server_and_client custom user script

10. Update the init file:
  * refactor some code to allow the init script file to be sourced by the uci defaults scripts and the luci rpcd script for shared functions
  * add support for `nft_file_mode` in which service prepares the fw4-compatible atomic nft/include file for faster operations on service reload
  * improve Tor support (nft mode only)
  * implement support for nft set options
  * update validation functions for new options/parameters

### DIFF
--- a/net/pbr/files/usr/share/pbr/pbr.user.netflix
+++ b/net/pbr/files/usr/share/pbr/pbr.user.netflix
@@ -3,45 +3,63 @@
 # Credits to https://forum.openwrt.org/u/dscpl for api.hackertarget.com code.
 # Credits to https://github.com/kkeker and https://github.com/tophirsch for api.bgpview.io code.
 
-TARGET_SET='pbr_wan_4_dst_ip_user'
-TARGET_IPSET='pbr_wan_4_dst_net_user'
+TARGET_INTERFACE='wan'
+TARGET_NFTSET_4="pbr_${TARGET_INTERFACE}_4_dst_ip_user"
+TARGET_NFTSET_6="pbr_${TARGET_INTERFACE}_6_dst_ip_user"
+TARGET_IPSET_4="pbr_${TARGET_INTERFACE}_4_dst_net_user"
+TARGET_IPSET_6="pbr_${TARGET_INTERFACE}_6_dst_net_user"
 TARGET_TABLE='inet fw4'
 TARGET_ASN='2906'
-TARGET_DL_FILE="/var/pbr_tmp_AS${TARGET_ASN}"
-TARGET_NFT_FILE="/var/pbr_tmp_AS${TARGET_ASN}.nft"
-#DB_SOURCE='ipinfo.io'
+TARGET_DL_FILE_4="/var/pbr_tmp_AS${TARGET_ASN}.ipv4"
+# Uncomment the following line if you enabled ipv6 for pbr and want IPv6 entries added to the IPv6 set
+# TARGET_DL_FILE_6="/var/pbr_tmp_AS${TARGET_ASN}.ipv6"
+DB_SOURCE='ipinfo.io'
 #DB_SOURCE='api.hackertarget.com'
-DB_SOURCE='api.bgpview.io'
-[ -z "$nft" ] && nft="$(command -v nft)"
-_ret=1
+#DB_SOURCE='api.bgpview.io'
+REGEX_IPV4='[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\/[0-9]\{1,\}'
+REGEX_IPV6='.*::.*'
+_ret=0
 
-if [ ! -s "$TARGET_DL_FILE" ]; then
+if [ ! -s "$TARGET_DL_FILE_4" ]; then
 	if [ "$DB_SOURCE" = "ipinfo.io" ]; then
 		TARGET_URL="https://ipinfo.io/AS${TARGET_ASN}"
-		uclient-fetch --no-check-certificate -qO- "$TARGET_URL" 2>/dev/null | grep -E "a href.*${TARGET_ASN}\/" | grep -v ":" | sed "s/^.*<a href=\"\/AS${TARGET_ASN}\///; s/\" >//" > "$TARGET_DL_FILE"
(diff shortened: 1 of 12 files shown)


net/pbr/files/usr/share/pbr/pbr.user.netflix
```diff
@@ -3,45 +3,63 @@
...
+TARGET_DL_FILE_4="/var/pbr_tmp_AS${TARGET_ASN}.ipv4"
+# Uncomment the following line if you enabled ipv6 for pbr and want IPv6 entries added to the IPv6 set
+# TARGET_DL_FILE_6="/var/pbr_tmp_AS${TARGET_ASN}.ipv6"
+DB_SOURCE='ipinfo.io'
 #DB_SOURCE='api.hackertarget.com'
-DB_SOURCE='api.bgpview.io'
-[ -z "$nft" ] && nft="$(command -v nft)"
-_ret=1
+#DB_SOURCE='api.bgpview.io'
+REGEX_IPV4='[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\/[0-9]\{1,\}'
```
systemcrash, line 19: Nitpick... Are there ever cases when the mask is > 32? Perhaps it's sensible to close the range?:
```
REGEX_IPV4='[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\/[0-9]\{1,2\}'
```

net/pbr/files/usr/share/pbr/pbr.user.netflix
```diff
@@ -3,45 +3,63 @@
...
+REGEX_IPV6='.*::.*'
```
systemcrash, line 20: I know we like simple expressions, but... Does this match if someone/something provides a non-abbreviated IPv6?

`2001:0db8:0000:0000:0000:8a2e:0370:7334`? I don't know what the likelihood of that is here.

Is a workaround to look for `[` and `]`?

systemcrash: If iPv4 don't have ports in them, perhaps just check for `:`

systemcrash:
> @stangri I can not compile it, I get error: `install: cannot stat './files/usr/share/pbr/.keep': No such file or directory`
> 
> `$(INSTALL_DATA) ./files/usr/share/pbr/.keep $(1)/usr/share/pbr/.keep`
> 
> Take the .keep file from your repo and now it compile

Yeah, test cases fail at this stage.

net/pbr/files/usr/share/pbr/pbr.user.netflix
stangri: Paul, thank you for your review. As I don't use these scripts myself, I have no way of testing for the desired result, so I don't want to change things in them unless I have to. If things work as is, I'd prefer to keep code as is.

net/pbr/files/usr/share/pbr/pbr.user.netflix
stangri: Paul, thank you for your review. As I don't use these scripts myself, I have no way of testing for the desired result, so I don't want to change things in them unless I have to. If things work as is, I'd prefer to keep code as is.

systemcrash:
Currently failing with:
```
...
Failed to connect to ubus
Failed to connect to ubus
Failed to connect to ubus
Failed to connect to ubus
Failed to connect to ubus
ERROR: The pbr service is currently disabled!
WARNING: Resolver set (dnsmasq.ipset) is not supported on this system.
WARNING: Resolver set (dnsmasq.ipset) is not supported on this system.
touch: /var/run/pbr.lock: No such file or directory
```

stangri:
> Currently failing with:

It's only failing for some architectures and not others. I actually build and test on x86_64, and CI is failing that arch. I believe this was also the case with the previous PR.

Can anyone shed any light on why CI is trying to run the service and where failed ubus calls are coming from?

pesa1234:
> > Currently failing with:
> 
> It's only failing for some architectures and not others. I actually build and test on x86_64, and CI is failing that arch. I believe this was also the case with the previous PR.
> 
> Can anyone shed any light on why CI is trying to run the service and where failed ubus calls are coming from?

Where the build is fine seems that the test with docker container is skipped.

systemcrash:
Green across the board 🥳

pesa1234:
Can you please update to 1.1.4.-r7 thanks?

stangri:
@hnyman Hannu, could you please do me a favor and have a look at/comment on the Makefile? I believe I was following your instructions from the mail-list or forum trying to do the variants better, but some users on the forum still say that for them when installing `luci-app-pbr` which depends on `pbr`, the opkg picks the `pbr-iptables` variant. I was trying to make the nft-capable variant called just `pbr` and be a default variant. Thanks!

stangri:
> I think that the most clear naming and dependency might be something like "pbr-nftables" and "pbr-legacy-iptables", and both of them would provide "pbr". and pbr-nftables would be the DEFAULT_VARIANT

Thank you for your prompt reply. I really want to keep the name of the default package just `pbr`, hence the invention of the `pbr-service` as a dependency. I'll try moving the VARIANT definition up.

PR #23694, opened 2024-03-19
flashrom: strip leading whitespace from PROGRAMMER_ARGS

nemesifier:
Maintainer: @Noltari 
Compile tested: qualcommax-ipq807x
Run tested: edgecore_eap102

Description:

I can't compile the current OpenWrt master branch without this change. I'm not 100% sure if this is the best way to fix it, but I hope it may be useful to somebody who may encounter the same issue.

It seems that newer version of meson do not allow empty arguments but `PROGRAMMER_ARGS` begin with a comma, which is considered as an empty argument.

### DIFF
--- a/utils/flashrom/Makefile
+++ b/utils/flashrom/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=flashrom
 PKG_VERSION:=1.3.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-v$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=https://download.flashrom.org/releases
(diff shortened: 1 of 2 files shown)


1715173329 (approved):
btw I would like to do strip in `MESON_ARGS` directly, i.e.:
`-Dprogrammer=$(subst $() $(),$(comma),$(strip $(PROGRAMMER_ARGS))) \`

nemesifier (commented):
@1715173329 followed your suggestion, seems that `$(comma)` is also defined elsewhere so I removed its definition as it seems redundant.

1715173329:
Hi @Noltari any chance to take a look? Thanks!

Noltari (approved):
LGTM

nemesifier (commented):
@Noltari @1715173329 I've incremented `PKG_RELEASE` to `2` in accordance with the guidelines outlined in the [Creating packages, BuildPackage variables](https://openwrt.org/docs/guide-developer/packages#buildpackage_variables) documentation section.

PR #23702, opened 2024-03-20
apk: move package to core

aparcar:
This will become part of openwrt.git and used within the build system.

Maintainer: me


aparcar:
The party goes on over here https://github.com/openwrt/openwrt/pull/14917

aparcar:
@neheb do you think the patch added here is also relevant for the APK version 3.0 in core?

neheb:
Nah no need.

PR #23704, opened 2024-03-21
htop: build with CONFIG_HTOP_LMSENSORS=y

### LINKED ISSUE #23700: htop: compiling with support for cpu temp does not add libsensors5 as a dep
Maintainer: @champtar 
Environment: flogic/glinet_gl-mt6000, latest snapshot

Description:  Building `htop` with `Compile Htop with lm-sensors support` does not add a dependency of `libsensors5` which is needed to get the functionality.

graysky2:
Default CONFIG_HTOP_LMSENSORS to y so users just need to install libsensors, no need to build your own, but no need to pay the size price when you don't use it.  Closes #23700.

Maintainer: @champtar

### DIFF
--- a/admin/htop/Makefile
+++ b/admin/htop/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=htop
 PKG_VERSION:=3.3.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/htop-dev/htop/tar.gz/$(PKG_VERSION)?
@@ -48,11 +48,12 @@ define Package/htop/config
 	config HTOP_LMSENSORS
 		bool "Compile Htop with lm-sensors support"
 		depends on PACKAGE_htop
-		default y if TARGET_x86
+		default y
 		help
 			Build htop with lm-sensors support.
-			This doesn't add lm-sensors as dependency,
-			if present it'll loaded using dlopen().
+			This increases the binary by approx 5 kB.
+			Users wanting this functionality need to
+			install libsensors.
 endef
 
 CONFIGURE_ARGS += \


graysky2:
@BKPepe

BKPepe:
Thanks for creating this PR. I will look at it today.

Dne čt 21. 3. 2024 10:57 uživatel graysky ***@***.***> napsal:

admin/htop/Makefile
```diff
@@ -33,7 +33,7 @@ define Package/htop
   SECTION:=admin
   CATEGORY:=Administration
   TITLE:=Interactive processes viewer
-  DEPENDS:=+libncurses
+  DEPENDS:=+libncurses $(if $(CONFIG_HTOP_LMSENSORS),+libsensors)
```
BKPepe, line 36: Why so complicated? :-)

This one should work:
```suggestion
  DEPENDS:=+libncurses +HTOP_LMSENSORS:libsensors
```

Not tested, though.

BKPepe (changes requested):
1. You need to change descriptipon for the config HTOP_LMSENSORS
2. Do we need to have this dependency as it was pointed out in https://github.com/openwrt/packages/issues/23700#issuecomment-2011859237 by @brada4 , we do have it also in the commit description and by looking how Debian does it - they have the dependency as suggested, not required - https://packages.debian.org/trixie/htop

Nitpick: your commit description has several typos, could you fix that? :-)

graysky2:
> You need to change description for the config HTOP_LMSENSORS

> Do we need to have this dependency as it was pointed out in https://github.com/openwrt/packages/issues/23700#issuecomment-2011859237 by @brada4 , we do have it also in the commit description and by looking how Debian does it - they have the dependency as suggested, not required - https://packages.debian.org/trixie/htop

I am confused by this statement.  If I do not have `libsensors` installed, I do not get temp data.  If `dlopen()` is a function within htop, the user still has to manually install libsensors.

> Nitpick: your commit description has several typos, could you fix that? :-)

Fixed the spelling error

graysky2:
If I hard-code `--enable-sensors` or `--disable-sensors` into the `Makefile` the resulting binary does not change in size.  It is a constant 265720 bytes.

graysky2:
OK so what is the intended behavior of that option in our package?  To me, if I select it in my `nconfig`, I expect the needed depends to get built and to get installed with it.  What is your expectation as to its behavior?

BKPepe:
> what is the intended behavior of that option in our package?

Keep it as it is. The dependency is optional, it is even written in the config option description in the Makefile, which you modifies. Feel free to ask htop developers.

Anyway: I dont see that you introduced my change and also the config option description is untouched.

BKPepe:
> I couldn't easily find your email address.

Hm, it should not be so hard based on this https://github.com/openwrt/packages/commits?author=BKPepe

graysky2:
> Anyway: I dont see that you introduced my change and also the config option description is untouched. 

I did actually.  I do not understand why the changes are not reflected here.

```
% git diff HEAD~1
diff --git a/admin/htop/Makefile b/admin/htop/Makefile
index 155363106..9c5412077 100644
--- a/admin/htop/Makefile
+++ b/admin/htop/Makefile
 
 PKG_NAME:=htop
 PKG_VERSION:=3.3.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_URL:=https://codeload.github.com/htop-dev/htop/tar.gz/$(PKG_VERSION)?
@@ -33,7 +33,7 @@ define Package/htop
   SECTION:=admin
   CATEGORY:=Administration
   TITLE:=Interactive processes viewer
-  DEPENDS:=+libncurses
+  DEPENDS:=+libncurses +HTOP_LMSENSORS:libsensors
   URL:=https://hisham.hm/htop/
   MAINTAINER:=Etienne CHAMPETIER <champetier.etienne@gmail.com>
 endef
@@ -46,13 +46,11 @@ endef
 
 define Package/htop/config
        config HTOP_LMSENSORS
-               bool "Compile Htop with lm-sensors support"
+               bool "Enable support for reading temperature data"
                depends on PACKAGE_htop
                default y if TARGET_x86
                help
-                       Build htop with lm-sensors support.
-                       This doesn't add lm-sensors as dependency,
-                       if present it'll loaded using dlopen().
+                       This adds libsensors as dependency.
 endef
 
 CONFIGURE_ARGS += \
```

brada4:
Present package with sensors lib optional is universally usable in container and out, you make it diverge.

graysky2:
Should I close the PR?

brada4:
The literary part could explain better that th*s time we can read sensirs with libsensors. But extra dependency no no.

champtar:
https://downloads.openwrt.org/snapshots/packages/x86_64/packages/
htop is 113.4 KB
libsensors is 20.2 KB
sysfsutils is 9.3 KB

@graysky2 if the size change is minimal when can set CONFIG_HTOP_LMSENSORS default to y

brada4:
@graysky2 change description to address your concern, like what would address your initial struggle that led you to this point.

graysky2:
@champtar - how would you like to proceed if at all?

champtar:
(on my phone until Wednesday)
@graysky2 I would say just default CONFIG_HTOP_LMSENSORS to y so people just need to install libsensors, no need to build your own, but no need to pay the size price when you don't use it (I understand the 5k as uncompressed size, my sizes are compressed size).
I don't think APK has 'recommended' dependencies, opkg definitely does not, that would be a good solution.

graysky2:
@champtar - I modified the PR.  Good to merge?

champtar:
@graysky2 I only see PKG_RELEASE being changed, nothing else

graysky2:
Forgot to commit, please seen latest.  Test built as expected for me.

brada4:
IMO diff summarizes intent - lmsensors supporting htop is built where sensors are likely.

PR #23705, opened 2024-03-21
etherwake: add missing lincense information

feckert:
Maintainer: @neheb @tripolar 
Compile tested: not needed
Run tested: not needed

Description:

In the source there is a note that it is under `GPL`.
But it is not clear which version of the `GPL` (GPL-1.0, GPL-2.0, GPL-3.0).

At the time the source file was created, `GPL-2.0` was the latest GPL.
Also, in debian the package refers to `GPL` which is a symlink to `GPL-3.0`.

Therefore the `PKG_LICENSE` variable is set to `GPL-2.0-or-later`.

### DIFF
--- a/net/etherwake/Makefile
+++ b/net/etherwake/Makefile
@@ -10,6 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=etherwake
 PKG_VERSION:=1.09
 PKG_RELEASE:=5
+PKG_LICENSE:=GPL-2.0-or-later
 
 PKG_SOURCE:=$(PKG_NAME)_$(PKG_VERSION).orig.tar.gz
 PKG_SOURCE_URL:=http://ftp.debian.org/debian/pool/main/e/etherwake


feckert:
@BKPepe The mail could not be delivered to the author of this source. I have added you as CC to my request.

What should we do?

Answer from the Mail Delivery System to my mail:
```
A message that you sent could not be delivered to all of its recipients.
This is a permanent error. The following address(es) failed:

  [becker@scyld.com](mailto:becker@scyld.com)
    retry timeout exceeded
Reporting-MTA: dns; relay.expurgate.net

Action: failed
Final-Recipient: rfc822;[becker@scyld.com](mailto:becker@scyld.com)
Status: 5.0.0
Subject 	Which GPL license does etherwake have
From 	[Florian Eckert](mailto:fe@dev.tdt.de)
To 	[becker@scyld.com](mailto:becker@scyld.com)
Cc 	[pepe.schlehofer@gmail.com](mailto:pepe.schlehofer@gmail.com)
Date 	2024-03-28 10:35
Hello Mr. Donald Becker,

As you may know, the source of your written ether-wake tool
is used in various distributions. This is also the case for
OpenWrt.I am currently in the process of adding the missing
licenses tags in the OpenWrt buildsystem for tools the does
not have a valid PKG_LICENSE variable set in the build
Makefile[[1](https://github.com/openwrt/packages/pull/23705)].

This also applies to your source of the etherwake package [[2](https://git.openwrt.org/?p=feed/packages.git;a=blob;f=net/etherwake/Makefile;hb=HEAD)]
that compiles your source in OpenWrt.

In the source file it says that it is GPL but not exactly
which one.

I think we have the following to choose for GPL that are eligible [[3](https://spdx.org/licenses/)]
* GPL-1.0-only
* GPL-1.0-or-later
* GPL-2.0-only
* GPL-2.0-or-later
* GPL-3.0-only
* GPL-3.0-or-later

Thank you for your feedback.

Best regards

Florian

[1] https://github.com/openwrt/packages/pull/23705
[2] https://git.openwrt.org/?p=feed/packages.git;a=blob;f=net/etherwake/Makefile;hb=HEAD
[3] https://spdx.org/licenses/
```

PR #23714, opened 2024-03-23
pbr: prepare migration to APK

stangri:
Maintainer: me

Description:
* remove dependencies/references to opkg
* simplify wan/wan6 discovery


BKPepe:
?? This should not be merged. 

Commit subject is prepare to migration to APK and it looks like that based on commit description, you are trying to slip unrelevant changes to it. 

cc: @feckert - What do you think about this?

PR #23715, opened 2024-03-23
nebula: prepare migration to APK

stangri:
Maintainer: me

### DIFF
--- a/net/nebula/Makefile
+++ b/net/nebula/Makefile
@@ -5,7 +5,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nebula
 PKG_VERSION:=1.8.2
-PKG_RELEASE:=2
+PKG_RELEASE:=r2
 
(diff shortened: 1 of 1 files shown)


BKPepe:
I am thinking that there should be commit description, which leads or described, what you done. I know it, but in the future, we will look back in the history and we would not know why.

Otherwise commit itself and the changes are understandable for me. You are not trying to slip any more changes like in other PRs, which were not relevant to APK migration.

BKPepe:
Anyway, I think this transition or migration to APK should be done via treewide commit because otherwise, we will open PR to each package, and it would not be good.

BKPepe:
Hola... why are you creating pull requests and then merging it by your own? It is like #yolo. I should not waste my time on this until is being reflected. I kindly ask you to follow what is being written here: https://openwrt.org/submitting-patches and in the CONTRIBUTING.md. Thanks.

PR #23716, opened 2024-03-23
https-dns-proxy: prepare migration to APK

stangri:
Maintainer: me


net/https-dns-proxy/patches/020-src-options.c-add-version.patch
```diff
@@ -5,7 +5,7 @@
    return SW_VERSION;
  #else
 -  return "2023.10.10-atLeast";  // update date sometimes, like 1-2 times a year
-+  return "2023-11-19-1";  // update date sometimes, like 1-2 times a year
++  return "2023.11.19$(subst(.,-,$(PKG_VERSION)))-r1";  // update date sometimes, like 1-2 times a year
```
stangri, line 8: I'll fix that later today or tomorrow.

net/https-dns-proxy/Makefile
```diff
@@ -1,12 +1,13 @@
...
-PKG_VERSION:=2023-11-19
-PKG_RELEASE:=1
+PKG_VERSION:=2023.11.19
+PKG_RELEASE:=r1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/aarond10/https_dns_proxy/
-PKG_SOURCE_DATE:=$(PKG_VERSION)
+PKG_SOURCE_DATE:=$(subst(.,-,$(PKG_VERSION)))
+PKG_SOURCE_RELEASE:=$(subst(r,,$(PKG_RELEASE)))
```
BKPepe, line 10: @aparcar Is this wanted behavior? I think there will be several exceptions like this. 

While looking at your treewide commit in the main repo: https://github.com/openwrt/openwrt/commit/e8725a932e16eaf6ec51add8c084d959cbe32ff2

It sounds like we can drop those two lines as they are not required. Is that right?

net/https-dns-proxy/files/etc/init.d/https-dns-proxy
```diff
@@ -300,7 +300,9 @@ service_triggers() {
 	local wan wan6 i
 	local procd_trigger_wan6
 	if [ "$on_boot_trigger" = '1' ]; then
+		output "Setting $serviceName raw_trigger for 'interface.*.up'"
 		procd_add_raw_trigger "interface.*.up" 3000 "/etc/init.d/${packageName}" restart 'on_interface_up'
+		output_okn
```
BKPepe, line 305: Not relevant anyhow to APK migration. If it is not going to be addressed and you will still merge it, I will revert it.

BKPepe:
@1715173329 , @feckert Could you please look at this pull request? :) Would be great to have someone, who will share same opinion on this one.

PR #23717, opened 2024-03-23
adblock-fast: prepare migration to APK

stangri:
Maintainer: me


net/adblock-fast/files/etc/init.d/adblock-fast
```diff
@@ -56,6 +56,7 @@ readonly unboundFilter='s|^|local-zone: "|;s|$|" static|'
 readonly A_TMP="/var/${packageName}.a.tmp"
 readonly B_TMP="/var/${packageName}.b.tmp"
 readonly SED_TMP="/var/${packageName}.sed.tmp"
+readonly uciConfigFile="/etc/config/${packageName}"
```
BKPepe, line 59: How this is relevant to migration to APK? This should be done in separate commit imho.

BKPepe:
I am so sorry, but I will revert your commit. There are unrelevant changes in your commit, which are no way related to APK migration. 

cc: @aparcar See, we should not allow that in the first place. So many changes here... :-( :-( I am really sad about this one, though.

PR #23723, opened 2024-03-23
proto-bonding: Modify ipaddr as optional

findlayfeng:

Description:

### DIFF
--- a/net/bonding/files/lib/netifd/proto/bonding.sh
+++ b/net/bonding/files/lib/netifd/proto/bonding.sh
@@ -193,15 +193,8 @@ proto_bonding_setup() {
 
 	proto_init_update "$link" 1
 
-	# For static configuration we _MUST_ have an IP address
-	[ -z "$ipaddr" ] && {
-		echo "$cfg" "No local IP address defined"
-		proto_notify_error "$cfg" INVALID_LOCAL_ADDRESS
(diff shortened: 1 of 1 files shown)


feckert:
@findlayfeng Can you please add to your commit description, why this change is needed?
At some point in the future, if you want to see why this has been changed, a good description would not be bad.
Also please update the pullrequest description with correct values.

findlayfeng:
I prefer to rewrite the content of the bond interface into a bond device. Bond is more suitable as a device, but that seems to be a relatively large modification and destroys compatibility.
The current modification is quite appropriate. It retains the compatibility of the old configuration and allows for more flexible use of the bond interface.
In addition, I also submitted a PR on luci-proto-bonding to remove the ipaddr check.

findlayfeng:
@feckert Is this appropriate?
My description in English requires the help of Google Translate

feckert:
Thanks merged!

hitech95:
@findlayfeng I think that there was/is a limitation in netifd that requires static protocol interfaces to have an IP address unless they are virtual.
That said, netifd support bonding devices since https://github.com/openwrt/netifd/commit/5ba9744aac6d42da1e56357aca951b52f86cfacb but luci does not.

I've implemented luci support in https://github.com/openwrt/luci/pull/7086

PR #23730, opened 2024-03-23
gatling: Add procd files

Akira25:
Maintainer: me / @Akira25
Compile tested: No. There was no change in compilation related units
Run tested: Yes. I've copi ed the files by hand unto a Archer C5 and runtested them with an OpenWrt 21.02

Description:
This commit adds an uci configuration file and makes the gatling server controllable by procd.

I'd love to have these changes backported into 23.05 too.


Akira25:
There are two failing builds on powerpc architecture. They both fail, because cryptodev-linux cannot be built. This looks unrelated to me.

net/gatling/Makefile
```diff
@@ -35,6 +35,12 @@ endef
 define Package/gatling/install
 	$(INSTALL_DIR) $(1)/usr/bin
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/gatling $(1)/usr/bin/
+
+	$(INSTALL_DIR) $(1)/etc/config
+	$(INSTALL_CONF) ./files/gatling.conf $(1)/etc/config/gatling
```
BKPepe, line 40: Shouldnt this be marked as conffile?

Akira25: The file is a regular uci configuration file, that gets copied to `/etc/config/`. Do you have something specific in mind for marking this file?

I thought, using the `$(INSTALL_CONF)` macro should be sufficient.

BKPepe: That macro sets correct permissions as it is said here: https://openwrt.org/docs/guide-developer/packages

``conffiles``, on the other hands, prevents that the config is overwritten by updating this package.
- You can test on your router: modify config, reinstall it and config file was replaced by the version from the package.

Akira25: Ah, I see. Thank you very much for the hint. I added that section in the current version of the PR.

net/gatling/Makefile
```diff
@@ -32,9 +32,19 @@ define Package/gatling/description
   Gatling is particularly good in situations with very high load.
 endef
 
+define Package/conffiles
```
BKPepe, line 35:
```suggestion
define Package/gatling/conffiles
```

BKPepe:
I minor suggestion for Makefile and I will let others review the rest.

net/gatling/Makefile
Akira25: Added in the new version of the commit. Thank you for pointing out.

Akira25:
> I minor suggestion for Makefile and I will let others review the rest.

Thank you for reviewing so far. Could you please point out, whats missing in your opinion, to get this merged? I'd really like this PR to not lay around for too long.

net/gatling/files/gatling.init
```diff
@@ -0,0 +1,114 @@
...
+    procd_add_reload_trigger "gatling"
+}
+
+start_instance() {
+
+    local cfg="$1"
+	local listen http ftp_server
+	local enabled
+
+	config_get_bool enabled "$cfg" 'enabled' 1
```
mwarning, line 49: This file contains tabs and spaces as indentation. Please use tabs only (that is what I see is used in other scripts).

net/gatling/files/gatling.init
```diff
@@ -0,0 +1,114 @@
...
+    procd_open_instance
+
+    procd_set_param command /usr/bin/gatling
+    procd_set_param stdout 1
+    procd_set_param stderr 1
+    procd_set_param term_timeout 20
+
+    # get listen-address and slice it from back, to cut at port-delimiter
+    config_get http gatling listen_http
+    port=$(echo "$listen" | rev | cut -d':' -f 1 | rev )
```
mwarning, line 61: I do not see that the variable "listen" is set anywhere. It should be "http"?
Anyway, the extraction of a port here is rather fragile. How about using a separate host and port variable?

mwarning:
```
procd_append_param command -i "[::]"
```
gives me the following error
```
Wed Apr 17 11:02:04 2024 daemon.err gatling[2551]: gatling: warning: could not parse IP address ::].
```
@BKPepe do you have an idea how to fix this?

net/gatling/files/gatling.conf
```diff
@@ -0,0 +1,28 @@
+
+config gatling 'v4'
+	option listen_http '0.0.0.0:80'
+	option virtual_hosting on
```
mwarning, line 4: I would suggest to disable virtual hosts by default. The first thing I have tried is "http://192.168.1.1". But this won't work with virtual hosts.

net/gatling/files/gatling.init
```diff
@@ -0,0 +1,114 @@
...
+	local listen http ftp_server
+	local enabled
+
+	config_get_bool enabled "$cfg" 'enabled' 1
+	[ $enabled -gt 0 ] || return
+
+    procd_open_instance
+
+    procd_set_param command /usr/bin/gatling
+    procd_set_param stdout 1
```
mwarning, line 55: This will flood the logs during normal usage. How about we comment this out?

Akira25: Yep, will flood the logs. Otherwise, there is a option for that in net/gatling/files/gatling.conf, lines 7 and 21 respectively:
https://github.com/openwrt/packages/blob/fa4b3b36d3ac03fde88bb76b5d49157945ffe495/net/gatling/files/gatling.conf#L7

I would rather prefer setting this to off as default, instead of the need for adjusting the init-script.

net/gatling/files/gatling.conf
Akira25: This should work, unless the 192.168.1.1 is handled by gatling.

For vhosts to work properly, gatling expects a correspondant directory in the webroot, in wich it will `cd`. For your example given, there must be a dir `192.168.1.1:80` with an `index.html` located in it.

Virtual hosts can get disabled in the config file too. Although I see it as a major advantage over other webservers, I'd be fine with setting it default off.

net/gatling/files/gatling.init
Akira25: > I do not see that the variable "listen" is set anywhere. It should be "http"? Anyway, the extraction of a port here is rather fragile. How about using a separate host and port variable?

I chose this way, to stay somewhat compatible to the way, `uhttpd` does this (which I consider the reference webserver for OpenWrt).

I would be completely fine, to make separate options on this (one address, one port).

net/gatling/files/gatling.init
Akira25: Adjusted it to use tabs only in the most recent version of this PR.

net/gatling/files/gatling.init
Akira25: I've set the option to default off now.

net/gatling/files/gatling.init
mwarning: We can keep it the way it is. But we can make it simplier with the lines in my comment.

Akira25: I added the lines you do suggest in your comments.

net/gatling/files/gatling.conf
mwarning: Please set virtual hosts to off in the config file. This makes the first user experience much nicer (it was for me).

net/gatling/files/gatling.init
mwarning: After some tries I found out that gatling does not accept IPv6 addresses with "[]", you need to remove them:
````
ip="${listen_http%:*}"
ip="${ip:1:-1}"
````

Comment (Akira25) [reply net/gatling/files/gatling.init]: Thats a really cool thing. I added this in the recent version of the commit.

File: net/gatling/files/gatling.conf (hunk 5 above)
Comment (Akira25) [reply net/gatling/files/gatling.conf:4]: addressed in the current commits version.

**Comment by mwarning** [comment]
@Akira25 the package works now. IPv4&IPv6
test:
```
mkdir /var/www
echo "<h1>Hello World</h1>" > /var/www/index.html
/etc/init.d/gatling start
```

But now we need to fix the build in the pipeline.

mwarning:
@Akira25 please update the patch with the updated one I have send you. Since the x86_64 pipeline seems to be broken, we can then ask for a merge.

Akira25:
> @Akira25 please update the patch with the updated one I have send you. Since the x86_64 pipeline seems to be broken, we can then ask for a merge.

Thank you very much for sending the patch. I've included it and the pipeline went all-green now.

mwarning:
@BKPepe looks good now. lgtm

PR #23734, opened 2024-03-24
rust: update to 1.77.0

krant:
Maintainer: @lu-zero 
Run tested: ARMv7, Linksys WRT3200ACM, by ripgrep package

Description:
- Restore patch hunk mis-deleted in dccb910
- Refresh patches
- Remove --enable-missing-tools configure option deleted in the upstream

### DIFF
--- a/lang/rust/Makefile
+++ b/lang/rust/Makefile
@@ -5,12 +5,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=rust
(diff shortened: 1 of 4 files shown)


lu-zero:
Looks nice, I might try to upstream the xz2 patch so the next version we might have even less patches to carry around (the cache one landed but only in master).

Thank you!

1715173329:
is it possible to use/backport that upstreamed cache patch?

Issue #23706, opened 2024-03-21
Preparations to switch to APK package manager


aparcar:
Hi all, some fellow developers and me worked for some time now on making [APK](https://gitlab.alpinelinux.org/alpine/apk-tools) the new package manager for OpenWrt, replacing the unmaintained OPKG fork we've been using for the longest time.

APK is actively developed and used in multiple other distributions, e.g. Alpine Linux 🎉 

While there is till some work ahead, I'd like to prepare everyone who maintains a package to verify that the `PKG_VERSION` follows  ~~[Semantic Versioning](https://semver.org)~~ `<major>.<minor>.<fixup>[.<sub1>...]`. APK uses a deterministic algorithm to compare versions and does not like random strings, except a valid hash prefixed with a `~`.

If in doubt, please use the Docker container below to verify the used version is valid:

```
docker run --rm -it ghcr.io/aparcar/apk-valid-version <VERSION> [<VERSION> ...]
```

It will print whatever version is not valid, if you get a zero exit code, you're fine.

Please feel free to reach out for assistance and have a look at the [core migration of versions](https://github.com/openwrt/openwrt/pull/14918).

aparcar:
Below packages that would cause trouble. Please bear in mind that `PKG_RELEASE` is soonish prefixed with an `r` and may only contain a number. 

**Naughty list:**

- [ ] ack                              `3.7.0+perl5.38-r1`
- [ ] atlas-probe                      `0~2.6.3-r1`
- [ ] bcm2711-eeprom                   `v.2024.01.05-2712-r2`
- [ ] bcm2712-eeprom                   `v.2024.01.05-2712-r2`
- [ ] bcm27xx-eeprom                   `v.2024.01.05-2712-r2`
- [ ] bluez-tools                      `20201025.f653217-r2`
- [ ] convbin                          `2.4.3_b34-r1`
- [ ] debootstrap                      `1.0.128+nmu2+deb12u1-r1`
- [ ] faad2                            `0~2.11.1-r1`
- [ ] fft-eval                         `2019-11-27-r4d3b6faee428e3bd9f44ab6a3d70585ec50484a1`
- [ ] hfsfsck                          `540.1.linux3-r2`
- [ ] https-dns-proxy                  `2023-11-19-r1`
- [ ] isc-dhcp-client-ipv4             `4.4.3-P1-r7`
- [ ] isc-dhcp-client-ipv6             `4.4.3-P1-r7`
[...]

PR #23743, opened 2024-03-25
nghttp3: update to 1.2.0

stangri:
Maintainer: me

Description:
* update PKG_RELEASE to be apk-compatible
* update PKG_SOURCE/PKG_SOURCE_URL so that it builds

### DIFF
--- a/libs/nghttp3/Makefile
+++ b/libs/nghttp3/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nghttp3
-PKG_VERSION:=1.1.0
-PKG_RELEASE:=1
+PKG_VERSION:=1.2.0
+PKG_RELEASE:=r1
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://codeload.github.com/ngtcp2/nghttp3/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=b3ffb23a90442a0eafe8bfbefbc8b4ffb5179d68a7c0b8a416a34cf04b28d7c5
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=https://github.com/ngtcp2/$(PKG_NAME)/releases/download/v$(PKG_VERSION)/
+PKG_HASH:=d2e216bae7bd7362f850922e4237a5caa204853b3594b22adccab4c1e1c1d1aa
 
 PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=COPYING


libs/nghttp3/Makefile
```diff
@@ -1,12 +1,12 @@
...
 PKG_NAME:=nghttp3
-PKG_VERSION:=1.1.0
-PKG_RELEASE:=1
+PKG_VERSION:=1.2.0
+PKG_RELEASE:=r1
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://codeload.github.com/ngtcp2/nghttp3/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=b3ffb23a90442a0eafe8bfbefbc8b4ffb5179d68a7c0b8a416a34cf04b28d7c5
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
```
BKPepe, line 7: What is the size difference, when we are changing this from .gz to .xz?

libs/nghttp3/Makefile
```diff
@@ -1,12 +1,12 @@
...
+PKG_SOURCE_URL:=https://github.com/ngtcp2/$(PKG_NAME)/releases/download/v$(PKG_VERSION)/
```
BKPepe, line 8: Please dont reuse PKG_NAME variable here. You can find it in some pull request, why it is not a good idea.

If you will copy it to the browser, it will not work, so keep it as it was in the previous, so:
https://github.com/ngtcp2/nghttp3/releases/download/v$(PKG_VERSION)

There is nothing wrong with that.

You are also changing it from codeload to GitHub. It is not mentioned in the commit description.

libs/nghttp3/Makefile
```diff
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nghttp3
-PKG_VERSION:=1.1.0
-PKG_RELEASE:=1
+PKG_VERSION:=1.2.0
+PKG_RELEASE:=r1
```
BKPepe, line 5: This will be done by treewide commit as I told you in the other PRs.
Check this out: https://github.com/openwrt/packages/issues/23706

BKPepe (changes requested):

libs/nghttp3/Makefile
stangri: NACK: https://github.com/search?q=repo%3Aopenwrt%2Fpackages%20%24(PKG_NAME)%2Freleases&type=code

BKPepe: @stangri I am regret to say, but it looks like you are doing what you want. This does not look good. How many URL based on your findings are using PKG_NAME in the PKG_SOURCE_URL? Just a few packages. Not all. Why?

Sorry, but you posted almost random URL and said NACK. You havent do anything and simply merged, because you can. This is not how the things works.

BKPepe:
Please, avoid marking the conversation as resolved, if you did not do anything.

libs/nghttp3/Makefile
BKPepe, line 8: I wondered what was wrong that you needed to introduce such a change, and it seems you reported it to upstream, but in the commit description nor in this PR, you did not say why.  That's why I replied to the issue, which you created in upstream - https://github.com/ngtcp2/nghttp3/issues/202#issuecomment-2028722525

It would be great if you describe it more closely, so we know what is happening.

PR #23744, opened 2024-03-25
ngtcp2: update to 1.4.0

stangri:
Maintainer: me

Description:
* update PKG_RELEASE to be apk-compatible
* update PKG_SOURCE/PKG_SOURCE_URL so that it builds
* drop dependency on libopenssl as other SSL libs start to support HTTP/3

### DIFF
--- a/libs/ngtcp2/Makefile
+++ b/libs/ngtcp2/Makefile
@@ -1,12 +1,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ngtcp2
-PKG_VERSION:=1.1.0
-PKG_RELEASE:=1
+PKG_VERSION:=1.4.0
+PKG_RELEASE:=r1
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://codeload.github.com/ngtcp2/ngtcp2/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=987d784643edea4f2859c405f7dfbc53871a9f7ae5fcddf5fb12ec5dfce1ef22
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=https://github.com/ngtcp2/$(PKG_NAME)/releases/download/v$(PKG_VERSION)/
+PKG_HASH:=b5d1433b9f5c06ce249e1e390e97dcfa49bf7ada5cb7c8bed8e6cd4feaf1ca4a
 
 PKG_LICENSE:=MIT
 PKG_LICENSE_FILES:=COPYING
@@ -22,7 +22,7 @@ define Package/libngtcp2
   CATEGORY:=Libraries
   TITLE:=Implementation of QUIC protocol
   URL:=https://nghttp2.org/ngtcp2
-  DEPENDS:=+libnghttp3 +libopenssl
+  DEPENDS:=+libnghttp3
 endef
 
 define Package/libngtcp2/description


libs/ngtcp2/Makefile
```diff
@@ -1,12 +1,12 @@
...
-PKG_VERSION:=1.1.0
-PKG_RELEASE:=1
+PKG_VERSION:=1.4.0
+PKG_RELEASE:=r1
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://codeload.github.com/ngtcp2/ngtcp2/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=987d784643edea4f2859c405f7dfbc53871a9f7ae5fcddf5fb12ec5dfce1ef22
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=https://github.com/ngtcp2/$(PKG_NAME)/releases/download/v$(PKG_VERSION)/
```
BKPepe, line 8:
```suggestion
PKG_SOURCE_URL:=https://github.com/ngtcp2/ngtcp2/releases/download/v$(PKG_VERSION)/
```

libs/ngtcp2/Makefile
```diff
@@ -1,12 +1,12 @@
...
 PKG_NAME:=ngtcp2
-PKG_VERSION:=1.1.0
-PKG_RELEASE:=1
+PKG_VERSION:=1.4.0
+PKG_RELEASE:=r1
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://codeload.github.com/ngtcp2/ngtcp2/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=987d784643edea4f2859c405f7dfbc53871a9f7ae5fcddf5fb12ec5dfce1ef22
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
```
BKPepe, line 7: This was changed, but your commit description does not say why suddently you decided to use .xz instead of .gz.

BKPepe (changes requested):

libs/ngtcp2/Makefile
stangri: NACK: https://github.com/search?q=repo%3Aopenwrt%2Fpackages%20%24(PKG_NAME)%2Freleases&type=code

BKPepe: While looking at it, I am not sure what you would like to say.  There are just a few packages looking at the URL, which you posted. Others are not related for PKG_SOURCE_URL, but nice attempt, I need to say! :) This conversation is not resolved and also, why you are merging it when I didnt have any chance to respond?

Reference:
https://github.com/openwrt/packages/pull/4119

I am sorry to ask, but what was your reason to change it? Make it harder to copy paste it to browser? I think so, because: https://github.com/ngtcp2/$(PKG_NAME)/ says 404. Your change makes it harder to read, to copy paste, well, for everything. There was no need to do it.

libs/ngtcp2/Makefile
BKPepe: > @stangri marked this conversation as resolved. 
image

How? There wasn't any change, you have not add any description.

libs/ngtcp2/Makefile
1715173329: So yes, avoid using `$(PKG_NAME)` for repo URL.

stangri: @BKPepe @1715173329 do you gentlemen want to clean up all the other Makefiles containing `$(PKG_NAME)/releases` 
 in the `PKG_SOURCE_URL`?: https://github.com/search?q=repo%3Aopenwrt%2Fpackages%20%24(PKG_NAME)%2Freleases&type=code

1715173329: They were introduced for history reasons but let's not add new.

PR #23745, opened 2024-03-25
uspot: update to Git HEAD (2024-03-25)

f00b4r0:
Maintainer: me
Compile tested: ath79/generic, ramips/mt7621, x86/64
Run tested: ath79/generic, ramips/mt7621, x86/64 - OpenWrt 23.05

Description: uspot: update to Git HEAD (2024-03-25)

**Please cherry-pick into 23.05 once merged :)**

### DIFF
--- a/net/uspot/Makefile
+++ b/net/uspot/Makefile
@@ -8,9 +8,9 @@ PKG_MAINTAINER:=Thibaut VARÈNE <hacks@slashdirt.org>
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/f00b4r0/uspot.git
-PKG_SOURCE_DATE:=2024-01-09
-PKG_SOURCE_VERSION:=c4b6f2f0bb1e9d2da4adc8cb3523cd7e440d7584
-PKG_MIRROR_HASH:=fa6be10e0479a9dc71b8c21e57b07aac09c8938e8e7120045816a5cd4b949343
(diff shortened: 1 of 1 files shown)


net/uspot/Makefile
```diff
@@ -8,9 +8,9 @@ PKG_MAINTAINER:=Thibaut VARÈNE <hacks@slashdirt.org>
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/f00b4r0/uspot.git
-PKG_SOURCE_DATE:=2024-01-09
-PKG_SOURCE_VERSION:=c4b6f2f0bb1e9d2da4adc8cb3523cd7e440d7584
-PKG_MIRROR_HASH:=fa6be10e0479a9dc71b8c21e57b07aac09c8938e8e7120045816a5cd4b949343
+PKG_SOURCE_DATE:=2024-03-25
```
neheb, line 11: I believe this needs to be changed from - to . because of apk.

f00b4r0: Well I used `update_git_source_package.sh` from up-to-date maintainer-tools repo so if the output is wrong, the script must be updated.

That said, I sure hope that old style dates aren't breaking anything otherwise there will be a transition period where existing packages will be uninstallable, among other things.

f00b4r0: Also this comment seems to suggest nothing should be done for now: https://github.com/openwrt/packages/issues/23706#issuecomment-2016507940 ?

